PR #730, opened 2015-01-09
[WIP] Adding unit test for multiple conditions in query

kneth:
A user has reported an issue with query, and I have rewritten his queries as a unit test.

### DIFF
--- a/realm/src/androidTest/java/io/realm/RealmResultsTest.java
+++ b/realm/src/androidTest/java/io/realm/RealmResultsTest.java
@@ -29,6 +29,7 @@ import io.realm.entities.AllTypes;
 import io.realm.entities.Cat;
 import io.realm.entities.NonLatinFieldNames;
 import io.realm.entities.Owner;
+import io.realm.entities.TwoStrings;
 
 public class RealmResultsTest extends AndroidTestCase {
     protected final static int TEST_DATA_SIZE = 2516;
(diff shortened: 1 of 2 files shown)


simonlevy5:
Thanks for writing these up! I appreciate you guys looking into this. I found a subtle difference that was causing the issue, though. I was applying a .sort() to the initial RealmResult<?>, then calling where() and the other conditions before findAll(). After applying the sort(), only a subset of the results was appearing.

To recreate that in these tests you can simply call sort("string1") before adding the beginsWith and or clauses. 

I'm not sure if this is a subtle bug (i'd suggest an assert since this yields very unexpected results)  or just me not being very nitpicky when reading the documentation (it does state, "once you have done your query, you can sort the results like this"). Either way, thanks for taking the time!

kneth:
@simonlevy5 Thanks for your detailed description. I'll take a closer look.

kneth:
I have added a `sort()` but cannot reproduce the issue.

simonlevy5:
@kneth Again, thanks for digging deeper on this. Now, sadly, i'm lost for words. I've got the exact same code that you have in your unit tests running within my application and it is yielding zero results after calling findAll() on the RealmQuery object. My workaround of calling sort() after the query has resolved my issue though. Is there any more detail I can provide? Also, happy to meet one of your developers in SF if that would help, i'm across the street (literally) from your SF office.

kneth:
@simonlevy5 Which version of Realm are you using? Notice that we released a new version (0.77.0) but I don't it makes a difference in your case. Can you share your code? Or create a minimal app which can reproduce the issue?

I'm pretty sure that you are welcome at our SF office. @timanglade when is a good time to come by?

timanglade:
Hey @simonlevy5 :) All of our Android guys are in Copenhagen, so we won’t be much help to debug this in person, although we can always give it a try (most of the time you’ll be better served scheduling a Hangout with @kneth, @cmelchior or @emanuelez)

That said, we’re having another Android guest for lunch on Wednesday and we’re always up for good company around good food, so if you’re free, you should definitely come have lunch with us! You can email me for details tim@realm.io

kneth:
I'll close this PR as the unit test is covered by the other tests.

Issue #648, opened 2014-12-05
Is there a way to sort `RealmResults` by two columns?


zsavely:
I am trying to sort `RealmResults` by two columns:
```java
RealmResults<User> results = realm.where(User.class).equalTo("relevant", true).findAll();
results.sort("balance", RealmResults.SORT_ORDER_ASCENDING);
results.sort("visible", RealmResults.SORT_ORDER_ASCENDING);
setListViewData(getActivity(), results, true);
```
But in the listview the results are only sorted by the column `visible`.
So is there a way to sort by two columns? 

I am using version 0.75.1.

cmelchior:
This was released with 0.77.0

njzk2:
It would be nice if it was at least indicated in the documentation that the sort is not stable.

cmelchior:
@njzk2 Yes we should. That is an unfortunate oversight.

kneth:
@njzk2 Notice that objects in a Realm are unordered while objects in a `RealmResults` are ordered. The in-place `RealmResults.sort()` is stable (according to the classical definition: http://en.wikipedia.org/wiki/Sorting_algorithm#Stability).

To test/illustrate this, I have expanded the unit tests - please take a look at #778. We should definitely update the documentation to explain this clearly.

njzk2:
@kneth isn't that in contradiction with the initial report by @zsavely? Shouldn't stability inherently allow multiple successive sorts?

kneth:
@njzk2 Not really. The initial report was using single field sorting twice which is sorting two different lists. After that report, we have implemented multi-field sorting.

PR #734, opened 2015-01-10
Multi-field sorting

kneth:
Multi-field sorting is a popular feature request (for example #648). In this pull request, sorting Realm.allObjects(), RealmResults.sort() and RealmQuery.findAll() extended with multi-field sorting.

@emanuelez @bmunkholm @cmelchior


realm-jni/src/io_realm_internal_tableview.cpp
```diff
@@ -860,6 +860,40 @@ JNIEXPORT void JNICALL Java_io_realm_internal_TableView_nativeSort(
...
+                case type_Int:
+                case type_DateTime:
+                case type_Float:
+                case type_Double:
+                case type_String:
+                    indices.push_back( S(arr[i]) );
+                    ascendings.push_back(ascending);
+                    break;
+                default:
+                    ThrowException(env, IllegalArgument, "Sort is currently only supported on Integer, Float, Double, Boolean, Date, and String columns.");
```
emanuelez: We don't support the boxed types yet so use int, float, etc

realm-jni/src/io_realm_internal_tableview.cpp
```diff
@@ -860,6 +860,40 @@ JNIEXPORT void JNICALL Java_io_realm_internal_TableView_nativeSort(
...
+        std::vector<size_t> indices;
+        std::vector<bool> ascendings;
+        jsize arr_len = env->GetArrayLength(columnIndices);
+        jlong *arr = env->GetLongArrayElements(columnIndices, NULL);
+        for (int i=0; i<arr_len; ++i) {
+            if (!COL_INDEX_VALID(env, TV(nativeViewPtr), arr[i]))
+                return;
+            int colType = TV(nativeViewPtr)->get_column_type( S(arr[i]) );
+            switch (colType) {
+                case type_Bool:
```
emanuelez: How about byte, short and long?

realm/src/main/java/io/realm/Realm.java
```diff
@@ -955,6 +955,35 @@ boolean contains(Class<?> clazz) {
...
+    /**
+     * Get all objects of a specific Class sorted by specific field names.
+     *
+     * @param clazz the Class to get objects of.
+     * @param fieldNames an array of field names to sort by.
+     * @param sortAscending sort ascending if SORT_ORDER_ASCENDING, sort descending if SORT_ORDER_DESCENDING.
+     * @return  A sorted RealmResults containing the objects.
+     * @throws java.lang.IllegalArgumentException if a field name does not exist.
+     */
+    public <E extends RealmObject> RealmResults<E> allObjects(Class<E> clazz, String fieldNames[], boolean sortAscending) {
```
emanuelez: It looks a bit weird with the String array. I would either go with a List<String> or varargs, or both

realm/src/main/java/io/realm/RealmQuery.java
```diff
@@ -1129,6 +1130,61 @@ public long count() {
...
+    public RealmResults<E> findAll(String fieldNames[], boolean sortAscending) {
+        if (fieldNames == null) {
+            throw new IllegalArgumentException("fieldNames cannot be 'null'.");
+        } else if (fieldNames.length == 0) {
+            throw new IllegalArgumentException("At least one field name must be specified.");
+        } else if (fieldNames.length == 1) {
+            return findAll(fieldNames[0], sortAscending);
+        } else {
+            TableView tableView = query.findAll();
+            ArrayList<Long> columnIndices = new ArrayList<Long>();
```
emanuelez: Use the interface on the left side of the assignment. In this case `List<Long>`

realm/src/main/java/io/realm/Realm.java
cmelchior: I would prefer varargs, something like this:
```
public <E extends RealmObject> RealmResults<E> allObjects(Class<E> clazz, String... fieldNames, boolean... sortAscending)
```

It would be even better with we could interleave the strings and booleans so the sort order is just beside the column name, but that is not possible without introducing many more methods (as varargs doesn't support that) or a wrapper class which isn't particular nice either.

changelog.txt
```diff
@@ -1,4 +1,5 @@
 0.77
+ * Realm.allObjects(), RealmResults.sort() and RealmQuery.findAll() extended with multi-field sorting.
```
cmelchior: Perhaps "Realm.allObjects(), RealmResults.sort() and RealmQuery.findAll() now supports multi-field sorting" instead?

realm-jni/src/io_realm_internal_tableview.cpp
kneth: These are core's types, and `type_Int` is the general integer column type. I will check if core supports sorting binary (byte) data.

realm/src/main/java/io/realm/Realm.java
kneth: Vargs have to be the last in the argument list. This implies that the signature for `RealmResults.sort()` must be

    public void sort(boolean sortAscending, String... fieldNames)

Should we then change the signature for the other `sort` method to:

    public void sort(boolean sortAscending, String fieldName)

cmelchior: Gah, forgot about that.
That means it is impossible to create a really nice interface for both multi column and single column sort.
I think I would prefer 3 methods then (The first 2 probably covers 90+% of usecases):
```
public void sort(String fieldName, boolean sortAscending); // Current 
public void sort(String f1, boolean s1, String f2, boolean s2); // Most common multisort case. Not sure about good names
public void sort(String[] fieldNames, boolean[] sortAscending); // General case.

What do you think?

Comment (kneth) [inline realm/src/main/java/io/realm/Realm.java]: Looks right to me :-)

Comment (emanuelez) [inline realm/src/main/java/io/realm/Realm.java]: I agree... I would even make one for 3 fields, just in case. Guava libraries go all the way to 5 but I doubt we will need that.
I'm a bit concerned about the use of boolean. It makes the code less readable IMO compared to an Enum

Comment (cmelchior) [inline realm/src/main/java/io/realm/Realm.java]: Yes, I has been considering if we should change all our boolean constants to Enums as well. See eg. this discussion here: https://plus.google.com/105051985738280261832/posts/YDykw2hstUu and here https://twitter.com/JakeWharton/status/551876948469620737

Comment (emanuelez) [inline realm/src/main/java/io/realm/Realm.java]: Not to mention Effective Java item 30 :smiley:

**Comment by emanuelez** [comment]
:+1:

File: realm-jni/src/io_realm_internal_tableview.cpp [hunk 6]
```diff
@@ -854,12 +854,56 @@ JNIEXPORT void JNICALL Java_io_realm_internal_TableView_nativeSort(
...
+        if (!VIEW_VALID_AND_IN_SYNC(env, nativeViewPtr))
+            return;
+
+        jsize arr_len = env->GetArrayLength(columnIndices);
+        if (arr_len != env->GetArrayLength(ascending)) {
+            ThrowException(env, IllegalArgument, "Number of column indices and sort orders do not match.");
+            return;
+        }
+
+        jlong *long_arr = env->GetLongArrayElements(columnIndices, NULL);
```
bmunkholm: rename long_arr and bool_arr. Name variables after content and not type.

realm-jni/src/io_realm_internal_tableview.cpp
```diff
@@ -854,12 +854,56 @@ JNIEXPORT void JNICALL Java_io_realm_internal_TableView_nativeSort(
                 TV(nativeViewPtr)->sort( S(columnIndex), ascending != 0 ? true : false);
                 break;
             default:
-                ThrowException(env, IllegalArgument, "Sort is currently only supported on Integer, Float, Double, Boolean, Date, and String columns.");
+                ThrowException(env, IllegalArgument, "Sort is currently only supported on integer, float, double, boolean, Date, and String columns.");
```
cmelchior: What about short?

cmelchior:
Apart from comment :+1:

realm/src/main/java/io/realm/Realm.java
```diff
@@ -894,6 +894,59 @@ boolean contains(Class<?> clazz) {
...
+     * Get all objects of a specific Class sorted by specific field names.
+     *
+     * @param clazz the Class to get objects of.
+     * @param sortAscending sort ascending if SORT_ORDER_ASCENDING, sort descending if SORT_ORDER_DESCENDING.
+     * @param fieldNames field names to sort by.
+     * @return  A sorted RealmResults containing the objects.
+     * @throws java.lang.IllegalArgumentException if a field name does not exist.
+     */
+    public <E extends RealmObject> RealmResults<E> allObjects(Class<E> clazz, String fieldNames[], boolean sortAscending[]) {
+        // FIXME: This is not an optimal implementation. When core's Table::get_sorted_view() supports
```
bmunkholm: I assume this will be fixed in the core we are going to use?

realm/src/androidTest/java/io/realm/RealmQueryTest.java
```diff
@@ -305,4 +305,91 @@ public void testRealmQueryLink() {
...
+        AllTypes object2 = testRealm.createObject(AllTypes.class);
+        object2.setColumnLong(4);
+        object2.setColumnString("Brian");
+
+        AllTypes object3 = testRealm.createObject(AllTypes.class);
+        object3.setColumnLong(4);
+        object3.setColumnString("Adam");
+        testRealm.commitTransaction();
+
+        RealmResults<AllTypes> results1 = testRealm.where(AllTypes.class).findAll(new String[]{FIELD_STRING, FIELD_LONG}, new boolean[] {RealmResults.SORT_ORDER_ASCENDING, RealmResults.SORT_ORDER_ASCENDING});
```
bmunkholm: please wrap this and similar long lines

realm/src/main/java/io/realm/RealmResults.java
```diff
@@ -211,6 +219,75 @@ public void sort(String fieldName, boolean sortAscending) {
         }
     }
 
+    /**
+     * Sort existing @{link io.realm.RealmResults}.
+     *
+     * @param fieldNames an array of field names to sort by. Only fields of type boolean, short, int,
```
bmunkholm: This limitation should be copied to all the other methods that sorts.

changelog.txt
```diff
@@ -1,4 +1,5 @@
 0.77
+ * Realm.allObjects(), RealmResults.sort() and RealmQuery.findAll() now supports multi-field sorting.
```
kneth: The new method must be mentioned here. And that we deprecate another method.

omidraha:
With this merge, now realm support sorting on multiple columns/fields?
What is syntax for that?

kneth:
@omidraha You can use [`RealmQuery.sort()`](https://realm.io/docs/java/latest/api/io/realm/RealmQuery.html#sort-java.lang.String:A-io.realm.Sort:A-) or [`RealmResults.sort()`](https://realm.io/docs/java/latest/api/io/realm/RealmResults.html#sort-java.lang.String:A-io.realm.Sort:A-) with array of field names.

omidraha:
Thanks, @kneth,
 
I asked this because there is no doc about this on the main doc:
https://realm.io/docs/java/latest#sorting

kneth:
@omidraha The documentation only highlights some features of the SDK, and you need to dive into the API documentation to get all the details :smile:

bmunkholm:
That said, I think it could a good idea to mention it in the docs anyway.

bmunkholm:
It is actually mentioned as the last sentence: "It is possible to sort using multiple fields simultaneously.", while it doesn't have an example though.

PR #736, opened 2015-01-11
Added Realm.executeTransaction() method

cmelchior:
Added convenience method for using the begin/commit/cancel pattern for transactions. It also mirrors the functionality for RLMRealm.transactionWithBlock on Cocoa.

Note that while it actually looks like it is more difficult to write, IntelliJ folding + autocomplete means it actually require fewer keypresses plus folds nicely.

I added it as a inner public class in order to keep the global number of classes low.

@emanuelez @kneth @bmunkholm

### DIFF
--- a/realm/src/main/java/io/realm/Realm.java
+++ b/realm/src/main/java/io/realm/Realm.java
@@ -1078,6 +1078,29 @@ public final class Realm implements Closeable {
      }
 
     /**
+     * Executes a given transaction on the Realm. {@link #beginTransaction()} and
+     * {@link #commitTransaction()} will be called automatically. If any exception is thrown
+     * during the transaction {@link #cancelTransaction()} will be called instead.
+     *
+     * @param transaction Transaction to execute.
+     * @throws RealmException if any error happened during the transaction.
+     */
+    public void executeTransaction(Transaction transaction) {
+        if (transaction == null) return;
+        beginTransaction();
+        try {
+            transaction.execute(this);
+            commitTransaction();
+        } catch (RuntimeException e) {
+            cancelTransaction();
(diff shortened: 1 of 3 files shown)


emanuelez:
:+1: wonderful job!

realm/src/main/java/io/realm/Realm.java
```diff
@@ -1078,6 +1078,29 @@ public void cancelTransaction() {
      }
 
     /**
+     * Executes a given transaction on the Realm. {@link #beginTransaction()} and
+     * {@link #commitTransaction()} will be called automatically. If any exception is thrown
+     * during the transaction {@link #cancelTransaction()} will be called instead.
```
kneth: instead of commitTransaction().

realm/src/main/java/io/realm/Realm.java
```diff
@@ -1078,6 +1078,29 @@ public void cancelTransaction() {
...
+     *
+     * @param transaction Transaction to execute.
```
kneth: Link to class.

realm/src/main/java/io/realm/Realm.java
```diff
@@ -1232,5 +1255,15 @@ public static boolean compactRealmFile(Context context) {
     public String getPath() {
         return path;
     }
+
+    /**
+     * Encapsulates a Realm transaction.
+     *
+     * Using this class will automatically handle {@link #beginTransaction()} and {@link #commitTransaction()}
+     * If any exception is thrown while inside the transaction {@link #cancelTransaction()} will be called instead.
```
kneth: "thrown during the transaction"

kneth:
Besides my comments :+1:

Issue #733, opened 2015-01-09
working with RealmObject properly


ImGauravC:
Question: How to properly handle the case of an activity working with a RealmObject and a background thread writing to the same RealmObject including, potentially, deleting the object while the activity is running?

cmelchior:
Hi @ImGauravC 
Currently you have two solutions:

1) Add a RealmChangeListener and reload the object whenever something changes.

2) Encapsulate access to that object with try/catch for RealmException. If it has been closed or deleted an exception will be thrown and you can act accordingly.

Do any of these two cases work in your use case?

ImGauravC:
Ok, so I have an activity with a ListView adapter which extends the RealmBaseAdapter and uses the  autorefresh feature. My issue is that the background update probably causes issues with the position that I receive in the getView() of the adapter.

cmelchior:
Usually the position index is just used by ListView to render the appropriate items that are visible on the screen, and the whole list should be redrawn when notifyDatasetChanged() is called by the RealmBaseAdapter. Do you save the position index and use it for something? If yes, you should probably override notifyDatasetChanged() so you can modify/reset as appropriate.

ImGauravC:
I don't save the position index at all. 
When I do Log.d(TAG, "position is: " + position) in my getView(), below is a sample output I get 
position is: 8
position is: 8
position is: 8
position is: 8
position is: 8
position is: 8

cmelchior:
Have you tried looking at our AdapterExample: https://github.com/realm/realm-java/tree/master/examples/adapterExample/src/main/java/io/realm/examples/realmadapters

I tried to reproduce your behaviour there, and it iterates the positions correctly. Can you try to compare our implementation there with yours. Perhaps you can spot something that makes a difference?

ImGauravC:
I don't see any difference in the implementation.

I've posted the relevant code in my Adapter here:
http://stackoverflow.com/questions/27951923/not-getting-the-last-position-in-the-getview-on-android?noredirect=1#comment44299494_27951923

cmelchior:
@ImGauravC 
I am closing this as an Realm issue as we cannot reproduce it. The most likely cause is the adapter getView() code making the listview render the same position twice, which hopefully will get answered on Stack Overflow.
 
Feel free to reopen if you have any new information.

PR #738, opened 2015-01-12
Check if a RealmObject is still valid.

cmelchior:
This adds support for RealmObject.isValid() which mirrors Cocoas RLMObject.invalidated property. 
It returns false if the object is a standalone object, the Realm has been closed or the object has been deleted.

This also helps with cases like those described here: https://github.com/realm/realm-java/issues/733

@kneth @emanuelez @bmunkholm


realm/src/androidTest/java/io/realm/RealmObjectTest.java
```diff
@@ -477,4 +477,35 @@ public void testThreadModelClass() {
...
+        testRealm.commitTransaction();
+        testRealm.close();
+        assertFalse(allTypes.isValid());
+    }
+
+    public void testIsValidDeletedObject() {
+        testRealm.beginTransaction();
+        AllTypes allTypes = testRealm.createObject(AllTypes.class);
+        assertTrue(allTypes.isValid());
+        testRealm.clear(AllTypes.class);
```
bmunkholm: add test right after delete, before commit

realm/src/androidTest/java/io/realm/RealmObjectTest.java
```diff
@@ -477,4 +477,35 @@ public void testThreadModelClass() {
...
+        testRealm.commitTransaction();
+        assertFalse(allTypes.isValid());
+    }
+
+    public void testIsValidManagedObject() {
```
bmunkholm: A little too many almost identical tests imho. This can easily be merged with testIsValidClosedRealm(). But up to you.

bmunkholm:
Why revet the naming of "invalidated" used in Cocoa?

cmelchior:
invalidated is a fine name for a property, but in java the most natural name is the positive term, ie. isValid(), it is also shorter. I am not that familiar with iOS naming conventions so isInvalidated might make better sense there.

realm/src/androidTest/java/io/realm/RealmObjectTest.java
cmelchior: I am not a big fan of unit tests that test different things as the naming become too vague, and although the unit tests look kinda similar they test different code paths. So I will probably keep them as is, unless the others also object :)

kneth:
:+1:

realm-annotations-processor/src/main/java/io/realm/processor/RealmProxyClassGenerator.java
```diff
@@ -456,6 +456,9 @@ private void emitGetFieldNamesMethod(JavaWriter writer) throws IOException {
     private void emitToStringMethod(JavaWriter writer) throws IOException {
         writer.emitAnnotation("Override");
         writer.beginMethod("String", "toString", EnumSet.of(Modifier.PUBLIC));
+        writer.beginControlFlow("if (!isValid())");
```
emanuelez: Why only in toString?

cmelchior: I noticed that Cocoa did the same and it makes sense instead of throwing RealmException which might be a bit unexpected on a standard java method.

emanuelez:
:+1:

PR #739, opened 2015-01-13
Realm.close() fix + RealmLogger

cmelchior:
This PR fixes a bug resulting in the Realm not getting closed if you called close() too many times (see the unit test).

It has been fixed, and now logs a warning to the Android log. A generic RealmLog class has been added that should be usable on both Android and normal Java. Two logger classes are provided: DebugAndroidLogger and ReleaseAndroidLogger for the different types of releases configured in Androids BuildConfig.

@emanuelez @kneth @bmunkholm


kneth:
I would have preferred the logging class in a separate PR. And please mention it in the change log.

Otherwise :+1:

realm/src/main/java/io/realm/internal/android/AndroidLogger.java
```diff
@@ -0,0 +1,123 @@
...
+import android.util.Log;
+
+import io.realm.internal.log.Logger;
+import io.realm.internal.log.RealmLog;
+
+public class AndroidLogger implements Logger {
+
+    private static final int LOG_ENTRY_MAX_LENGTH = 4000;
+    private int minimumLogLevel = RealmLog.VERBOSE;
+    private String logTag = "REALM";
```
emanuelez: private static final

realm/src/androidTest/java/io/realm/log/RealmLog.java
```diff
@@ -0,0 +1,8 @@
+package io.realm.log;
+
+public interface RealmLog {
+
```
emanuelez: so many newlines :)

realm/src/main/java/io/realm/internal/android/AndroidLogger.java
cmelchior: There is a method for changing the tag :)

emanuelez: Which is never used. But it makes sense it's there :)

emanuelez:
:+1:

Issue #726, opened 2015-01-08
java.lang.IllegalStateException: Cannot use ImplicitTransaction after it or its parent has been closed.


ImGauravC:
```java
java.lang.IllegalStateException: Cannot use ImplicitTransaction after it or its parent has been closed.
            at io.realm.internal.ImplicitTransaction.assertNotClosed(ImplicitTransaction.java:66)
            at io.realm.internal.ImplicitTransaction.advanceRead(ImplicitTransaction.java:29)
            at io.realm.Realm$RealmCallback.handleMessage(Realm.java:194)
            at android.os.Handler.dispatchMessage(Handler.java:95)
            at android.os.Looper.loop(Looper.java:176)
```

I'm getting the above error while adding/updating realm objects in a background handler thread. What is the reason behind this?

cmelchior:
You have called close() on Realm and are probably trying to access a RealmResult or RealmObject afterwards.

ImGauravC:
Is this the only possibility? Because I don't see this happening anywhere in my code.

cmelchior:
It is by far the most likely. We will need more information to be more specific, can you share your code or try to reduce it to a testcase?

ImGauravC:
I got it finally. Thanks :)

ImGauravC:
Wait, encountering it again. Below is an explanation of what I'm trying to do. 

I need to constantly update the app with data from the server. So for polling data, I created a handler thread with a Looper.  On app start, in the run() method of the Runnable, I get a Realm instance, perform a transaction, close the Realm instance, and then post the Runnable again to the same thread with a delay of say, 60 seconds. 

Below is some pseudo code:

```java
Runnable r = new Runnable() {
     @Override
     public void run() {
         httpRequest.get(uri, new DataCallback() {
             public void onSuccess(JSONObject result) throws JSONException {
                 Realm realm = Realm.getInstance(context);
                 realm.setAutoRefresh(true);
                 try {
                     realm.beginTransaction();
                     <Some write with data from the server>
                     realm.commitTransation();
                 } catch (Exception e) {
                     realm.cancelTransaction();
                 } 
                 realm.close();
             }
        }
        handlerThread.postDelayed(r, 60 * 1000);
    } 
}  

handlerThread.post(r)
```

cmelchior:
That pattern looks pretty solid. You just have to be careful about catching Exception, as you might catch OutOfMemory which could potentially cause corruption on your database. Catching RuntimeException is much safer.

[...]

PR #740, opened 2015-01-13
Fixed bug causing refresh to be called on closed Realms

cmelchior:
This PR fixes a concurrency bug that could cause background threads with autorefresh enabled to try to refresh Realms that has been closed resulting in IllegalStateExceptions. See https://github.com/realm/realm-java/issues/726

Anybody who can find a way to simplify the unit test gets a beer or two from me :)

@emanuelez @kneth @bmunkholm


cmelchior:
The unit test seems to be quite fragile timewise. Perhaps the background is at a higer risk of getting paused. We could easily fix it by capturing the TimeoutException and treat it as as success, but that feels very wrong...

emanuelez:
The unit test looks scary but :+1:

bmunkholm:
:+1:

PR #721, opened 2015-01-07
Subqueries on RealmLists now use the proper tablespec 

cmelchior:
This is a fix for: https://github.com/realm/realm-java/issues/707

Also improved error reporting for mismatching types when doing queries.

@bmunkholm @kneth @emanuelez

PR #742, opened 2015-01-14
Fix RealmQuery scope issue

cmelchior:
This PR https://github.com/realm/realm-java/pull/721 seems to have introduced a regression causing subqueries on the RealmResult to use the entire table instead of the results of the previous query.


cmelchior:
@emanuelez @kneth @bmunkholm

realm/src/main/java/io/realm/RealmQuery.java
```diff
@@ -34,6 +34,7 @@
 public class RealmQuery<E extends RealmObject> {
 
     private Realm realm;
+    private Table tableSpec;
```
kneth: Please don't name it `tableSpec` as we have an internal class called `TableSpec`.

kneth:
Please update the change log.

cmelchior:
There shouldn't be any need. It is a regression that hasn't been in a public release yet. There is a changelog entry for the original error that triggered this ;)

realm/src/main/java/io/realm/RealmQuery.java
emanuelez: I agree. `table` would be a better name IMO

realm/src/main/java/io/realm/RealmQuery.java
```diff
@@ -79,14 +78,10 @@ public RealmQuery(RealmResults realmList, Class<E> clazz) {
...
-        TableOrView dataStore = getTable();
+        this.tableSpec = realm.getTable(clazz);
         this.columns = Realm.columnIndices.get(clazz.getSimpleName());
     }
 
-    TableOrView getTable() {
-        return realm.getTable(clazz);
-    }
-
     private boolean containsDot(String s) {
```
emanuelez: Not related to the PR but I wonder why this is here. And why not using s.contains('.'). The compiler should do all the needed optimizations

cmelchior: Yeah I stumble across it a few times now, and forgot to refactor it each time.

emanuelez:
:+1: apart from the comment. No need for line in changelog IMO

realm/src/main/java/io/realm/RealmQuery.java
bmunkholm: @cmelchior Then fix it now.

bmunkholm:
Otherwise :+1:

PR #746, opened 2015-01-14
More debug info when converting a string from to Java fails.

kneth:
@bmunkholm


realm-jni/src/utf8.hpp
```diff
@@ -0,0 +1,380 @@
...
+        }
+        uint_fast16_t v1 = uint_fast16_t(traits8::to_int_type(in[0]));
+        if (TIGHTDB_LIKELY(v1 < 0x80)) { // One byte
+            // UTF-8 layout: 0xxxxxxx
+            *out++ = Traits16::to_char_type(v1);
+            in += 1;
+            continue;
+        }
+        if (TIGHTDB_UNLIKELY(v1 < 0xC0)) {
+            invalid = true;
```
bmunkholm: give it a number

realm-jni/src/util.cpp
```diff
@@ -211,13 +211,22 @@ struct JStringCharsAccessor {
 
 } // anonymous namespace
 
-string string_to_hex(const string& message, StringData& str) {
+string string_to_hex(const string& message, StringData& str, const char* in_begin, const char* in_end,
+                     jchar* out_curr, jchar* out_end, size_t retcode) {
```
bmunkholm: print retcode ?

emanuelez:
:+1: from Brian :)

PR #748, opened 2015-01-15
Cut down on the size of the library a bit

tgoyne:
While poking at some other things I noticed a few easy spots to eliminate some bloat. Total savings is ~340 KB off realm.jar, most of which (~300 KB) is from the exception handling change. Should also improve performance insignificantly in non-error-cases due being more icache-friendly and doing less pointless refcounting at the cost of an also insignificant performance hit when exceptions are thrown.


bmunkholm:
@tgoyne :100: Looks good!
Similar things could also be done with parameter checking. But that would require benchmarking to see the effect.

kneth:
:+1:

PR #750, opened 2015-01-15
Small API fixes

cmelchior:
Fixes to RealmList.remove(), RealmList.first() and RealmList.last() found during review + additional unit tests to capture errors.

@bmunkholm @kneth @emanuelez

### DIFF
--- a/realm/src/androidTest/java/io/realm/RealmListTest.java
+++ b/realm/src/androidTest/java/io/realm/RealmListTest.java
@@ -358,12 +358,13 @@ public class RealmListTest extends AndroidTestCase {
     public void testRemoveByIndex() {
         Owner owner = testRealm.where(Owner.class).findFirst();
         RealmList<Dog> dogs = owner.getDogs();
+        Dog dog5 = dogs.get(5);
 
         testRealm.beginTransaction();
         Dog removedDog = dogs.remove(5);
         testRealm.commitTransaction();
 
-        assertNull(removedDog);
+        assertEquals(dog5, removedDog);
         assertEquals(TEST_OBJECTS - 1, dogs.size());
     }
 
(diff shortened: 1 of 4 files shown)


realm/src/androidTest/java/io/realm/RealmListTest.java
```diff
@@ -211,6 +211,35 @@ public void testMoveUp_nonManagedMode() {
...
+        RealmList<AllTypes> list = new RealmList<AllTypes>();
+        AllTypes object1 = new AllTypes();
+        AllTypes object2 = new AllTypes();
+        list.add(object1);
+        list.add(object2);
+
+        assertEquals(object1, list.first());
+    }
+
+    public void testFirstEmptyList_nonManagedMode() {
```
bmunkholm: I would merge this with the other testing last. So having the empty list, verify that all methods behaves correctly (also size()).

realm/src/androidTest/java/io/realm/RealmListTest.java
```diff
@@ -294,29 +323,48 @@ public void testGetObjects() {
...
+    }
+
+    public void testLast() {
         Owner owner = testRealm.where(Owner.class).findFirst();
         Dog dog = owner.getDogs().last();
 
         assertEquals("Dog " + (TEST_OBJECTS - 1), dog.getName());
     }
 
+    public void testLastEmptyList() {
```
bmunkholm: merge with testFirstEmptyList. 
Too much duplicate code. Same as above comment, but here the benefits are even more clear :-)

emanuelez:
:+1: I actually like the atomicity of the unit tests

cmelchior:
That would be my concern as well, we sacrifice transparency/readability for a smaller codebase/faster test execution.

kneth:
:+1:

cmelchior:
@bmunkholm I noticed a few native crashes when doing the empty list unit tests. And I just checked the LinkView JNI code. At least get() and set() doesn't validate their input properly, hence the Java checks. So yes, we should probably do another roll on the RealmList.

PR #753, opened 2015-01-16
Thread handler no longer gets removed too soon.

cmelchior:
This PR fixes the case where close() method removed the Thread handler to soon even though open instances still existed.

@bmunkholm @kneth @emanuelez

### DIFF
--- a/realm/src/main/java/io/realm/Realm.java
+++ b/realm/src/main/java/io/realm/Realm.java
@@ -1283,12 +1283,11 @@ public final class Realm implements Closeable {
 
     // Returns the Handler for this Realm on the calling thread
     Handler getHandler() {
-        for (Handler handler : handlers.keySet()) {
-            if (handlers.get(handler) == id) {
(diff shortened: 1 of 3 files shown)


bmunkholm:
:+1:

kneth:
:+1:

realm/src/main/java/io/realm/Realm.java
```diff
@@ -1277,6 +1277,21 @@ public void clear(Class<?> classSpec) {
         getTable(classSpec).clear();
     }
 
+    int getId() {
+        return id;
+    }
+
+    // Returns the Handler for this Realm on the calling thread
+    Handler getHandler() {
+        for (Handler handler : handlers.keySet()) {
```
emanuelez: This should iterate over handlers.entries instead.

cmelchior: Good catch, one less lookup.

PR #755, opened 2015-01-16
Optimizing Realm.allObjectSorted()

kneth:
The core team has adding multi-column sorting to `Table::get_sorted_view()` and we save a number of JNI calls by using it. Moreover, it is also attempted to reuse more code across classes.


realm/src/main/java/io/realm/Realm.java
```diff
@@ -1115,11 +1110,22 @@ boolean contains(Class<?> clazz) {
      */
     public <E extends RealmObject> RealmResults<E> allObjectsSorted(Class<E> clazz, String fieldNames[],
                                                                boolean sortAscending[]) {
-        // FIXME: This is not an optimal implementation. When core's Table::get_sorted_view() supports
-        // FIXME: multi-column sorting, we can rewrite this method to a far better implementation.
-        RealmResults<E> results = this.allObjects(clazz);
-        results.sort(fieldNames, sortAscending);
-        return results;
+        io.realm.internal.android.Util.validateMultiSortParameters(fieldNames, sortAscending);
```
cmelchior: Move import to top of file

realm/src/main/java/io/realm/Realm.java
```diff
@@ -1115,11 +1110,22 @@ boolean contains(Class<?> clazz) {
...
+
+        Table table = this.getTable(clazz);
+        List<TableView.Order> TVOrder = new ArrayList<TableView.Order>();
+        List<Long> columnIndices = new ArrayList<Long>();
```
cmelchior: No need to create an List here, just use primitive array to avoid autoboxing later.

realm/src/main/java/io/realm/Realm.java
```diff
@@ -1115,11 +1110,22 @@ boolean contains(Class<?> clazz) {
...
+        for (int i = 0; i < fieldNames.length; i++) {
+            String fieldName = fieldNames[i];
+            long columnIndex = table.getColumnIndex(fieldName);
+            if (columnIndex == -1) {
+                throw new IllegalArgumentException(String.format("Field name '%s' does not exist.", fieldName));
+            }
+            columnIndices.add(columnIndex);
```
cmelchior: Same...

realm/src/main/java/io/realm/RealmResults.java
```diff
@@ -230,17 +230,7 @@ public void sort(String fieldName, boolean sortAscending) {
...
-        } else if (sortAscending == null) {
-            throw new IllegalArgumentException("sortAscending must be provided.");
-        } else if (fieldNames.length == 0) {
-            throw new IllegalArgumentException("You must provide at least one field name.");
-        } else if (sortAscending.length == 0) {
-            throw new IllegalArgumentException("You must provide at least one sort order.");
-        } else if (fieldNames.length != sortAscending.length) {
-            throw new IllegalArgumentException(String.format("Number of field names (%d) and sort orders (%d) do not match.", fieldNames.length, sortAscending.length));
-        }
+        io.realm.internal.android.Util.validateMultiSortParameters(fieldNames, sortAscending);
```
cmelchior: Import to top

realm/src/main/java/io/realm/internal/Table.java
```diff
@@ -546,6 +547,24 @@ public TableView getSortedView(long columnIndex){
     protected native long nativeGetSortedView(long nativeTableViewPtr, long columnIndex, boolean ascending);
 
 
+    public TableView getSortedView(List<Long> columnIndices, List<TableView.Order> orders) {
```
cmelchior: Use primitive array for indices.

realm/src/main/java/io/realm/internal/Table.java
```diff
@@ -546,6 +547,24 @@ public TableView getSortedView(long columnIndex){
...
+        context.executeDelayedDisposal();
+        long indices[] = new long[columnIndices.size()];
```
cmelchior: Then this can be removed.

changelog.txt
```diff
@@ -1,3 +1,7 @@
+0.78
+ * Optimizing Realm.allObjectSorted().
```
emanuelez: Shall we agree to use the past tense for all changelog entries?

realm-jni/src/io_realm_internal_table.cpp
```diff
@@ -1219,12 +1219,59 @@ JNIEXPORT jlong JNICALL Java_io_realm_internal_Table_nativeGetSortedView(
                 return reinterpret_cast<jlong>(pTableView);
             } CATCH_STD()
         default:
-            ThrowException(env, IllegalArgument, "Sort is currently only supported on Integer, Boolean and Date columns.");
+            ThrowException(env, IllegalArgument, "Sort is currently only supported on integer, boolean, double, float, String, and Date columns.");
```
emanuelez: byte, short and long should be here as well, isn't it?

realm/src/main/java/io/realm/internal/android/Util.java
```diff
@@ -0,0 +1,40 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.realm.internal.android;
+
+public class Util {
```
cmelchior: Do you expect this class to hold other stuff that is more generic or has other concerns? Because this name is very generic right now.

realm-jni/src/io_realm_internal_table.cpp
```diff
@@ -1219,12 +1219,59 @@ JNIEXPORT jlong JNICALL Java_io_realm_internal_Table_nativeGetSortedView(
...
+    jsize arr_len = env->GetArrayLength(columnIndices);
+    if (arr_len != env->GetArrayLength(ascending)) {
+        ThrowException(env, IllegalArgument, "Number of column indices and sort orders do not match.");
+        return 0;
+    }
+
+    jlong *long_arr = env->GetLongArrayElements(columnIndices, NULL);
+    jboolean *bool_arr = env->GetBooleanArrayElements(ascending, NULL);
+
+    std::vector<size_t> indices;
```
emanuelez: We already know the size so we might as well initialize it with the proper size.

realm/src/main/java/io/realm/internal/android/Util.java
kneth: Yes, my idea was that we should use it for code reuse.

realm-jni/src/io_realm_internal_table.cpp
```diff
@@ -1219,12 +1219,59 @@ JNIEXPORT jlong JNICALL Java_io_realm_internal_Table_nativeGetSortedView(
...
+{
+    Table* pTable = TBL(nativeTablePtr);
+
+    jsize arr_len = env->GetArrayLength(columnIndices);
+    if (arr_len != env->GetArrayLength(ascending)) {
+        ThrowException(env, IllegalArgument, "Number of column indices and sort orders do not match.");
+        return 0;
+    }
+
+    jlong *long_arr = env->GetLongArrayElements(columnIndices, NULL);
```
emanuelez: Faster way to copy java primitive arrays to C++ vectors: http://stackoverflow.com/a/7318371/1213485

cmelchior:
:+1:

realm/src/main/java/io/realm/internal/android/Util.java
bmunkholm: this is not specific to android. Shouldn't it be located under /internal? We have other util files afaik.

realm/src/main/java/io/realm/internal/Table.java
```diff
@@ -546,6 +546,20 @@ public TableView getSortedView(long columnIndex){
     protected native long nativeGetSortedView(long nativeTableViewPtr, long columnIndex, boolean ascending);
 
 
+    public TableView getSortedView(long columnIndices[], boolean orders[]) {
+        context.executeDelayedDisposal();
+        long indices[] = new long[columnIndices.length];
```
bmunkholm: remove these two lines (indices and sortOrder) - they look unused.

realm/src/main/java/io/realm/internal/Table.java
```diff
@@ -546,6 +546,20 @@ public TableView getSortedView(long columnIndex){
...
+        boolean sortOrders[] = new boolean[orders.length];
+        long nativeViewPtr = nativeGetSortedViewMulti(nativePtr, columnIndices, orders);
+        try {
+            return new TableView(this.context, this, nativeViewPtr);
+        } catch (RuntimeException e) {
```
bmunkholm: why only catch runtime exception?

realm/src/main/java/io/realm/Realm.java
```diff
@@ -1115,11 +1112,23 @@ boolean contains(Class<?> clazz) {
      */
     public <E extends RealmObject> RealmResults<E> allObjectsSorted(Class<E> clazz, String fieldNames[],
                                                                boolean sortAscending[]) {
-        // FIXME: This is not an optimal implementation. When core's Table::get_sorted_view() supports
-        // FIXME: multi-column sorting, we can rewrite this method to a far better implementation.
-        RealmResults<E> results = this.allObjects(clazz);
-        results.sort(fieldNames, sortAscending);
-        return results;
+        validateMultiSortParameters(fieldNames, sortAscending);
```
bmunkholm: So now we are checking parameters both in java and in C++ jni - I suggest to keep the parameter checking in C++ only (as other methods). And only protect in java if java code requires it.

realm/src/main/java/io/realm/internal/Table.java
kneth: To make it consistent with 1+ year old code (see method above).

realm/src/main/java/io/realm/Realm.java
kneth: At C++ we only check that the lengths of the two arrays are equal but we can easily move all checks to C++.

PR #764, opened 2015-01-19
DOC: empty RealmResults

kneth:
Notice about the non-null nature of `RealmResults` and `RealmResults.size()` must be used always.


realm/src/main/java/io/realm/Realm.java
```diff
@@ -953,7 +952,8 @@ boolean contains(Class<?> clazz) {
     }
 
     /**
-     * Get all objects of a specific Class
+     * Get all objects of a specific Class. If no objects exist, the returned RealmResults will not
+     * be null. The RealmResults.size() to check the number of objects instead.
      *
      * @param clazz the Class to get objects of
      * @return A RealmResult list containing the objects
```
cmelchior: Usually you would provide such information in the @return description, something along the lines of "@return A RealmResult list containing all objects. If no objects exist, an empty list is returned".

Saying that .size() must be used is not needed as that is the standard method for lists.

realm/src/main/java/io/realm/Realm.java
```diff
@@ -965,7 +965,8 @@ boolean contains(Class<?> clazz) {
     }
 
     /**
-     * Get all objects of a specific Class sorted by a field.
+     * Get all objects of a specific Class sorted by a field.  If no objects exist, the returned
+     * RealmResults will not be null. The RealmResults.size() to check the number of objects instead.
```
cmelchior: Same...

realm/src/main/java/io/realm/Realm.java
```diff
@@ -988,7 +989,8 @@ boolean contains(Class<?> clazz) {
...
```
cmelchior: same...

realm/src/main/java/io/realm/Realm.java
```diff
@@ -1011,7 +1013,9 @@ boolean contains(Class<?> clazz) {
     }
 
     /**
-     * Get all objects of a specific class sorted by two field names.
+     * Get all objects of a specific class sorted by two field names.  If no objects exist, the
```
cmelchior: same...

realm/src/main/java/io/realm/Realm.java
```diff
@@ -1028,7 +1032,9 @@ boolean contains(Class<?> clazz) {
...
```
cmelchior: same...

realm/src/main/java/io/realm/Realm.java
```diff
@@ -1045,7 +1051,9 @@ boolean contains(Class<?> clazz) {
     }
 
     /**
-     * Get all objects of a specific class sorted by two specific field names.
+     * Get all objects of a specific class sorted by two specific field names.  If no objects exist,
```
cmelchior: same...

realm/src/main/java/io/realm/Realm.java
```diff
@@ -1065,7 +1073,9 @@ boolean contains(Class<?> clazz) {
...
```
cmelchior: same...

realm/src/main/java/io/realm/Realm.java
```diff
@@ -1085,7 +1095,9 @@ boolean contains(Class<?> clazz) {
     }
 
     /**
-     * Get all objects of a specific Class sorted by multiple fields.
+     * Get all objects of a specific Class sorted by multiple fields.  If no objects exist, the
```
cmelchior: same...

realm/src/main/java/io/realm/Realm.java
```diff
@@ -1104,7 +1116,9 @@ boolean contains(Class<?> clazz) {
...
```
cmelchior: same...

realm/src/main/java/io/realm/RealmQuery.java
```diff
@@ -1095,7 +1095,9 @@ public long count() {
     // Execute
 
     /**
-     * Find all objects that fulfill the query conditions.
+     * Find all objects that fulfill the query conditions. The returned RealmResults will never be
```
cmelchior: Same for this class :)

realm/src/main/java/io/realm/RealmResults.java
```diff
@@ -38,6 +38,8 @@
  * Realm that was queried.
  * Updates to objects must be done within a transaction and the modified object is persisted to the backing
  * Realm file during the commit of the transaction.
+ * Notice that a RealmResults is never null not even in the case where it contains no objects. You
```
cmelchior: Here it is okay as it describes the purpose and invariant of the entire class.

realm/src/main/java/io/realm/Realm.java
kneth: Will it be obvious that an empty `RealmResults` is not `null`? If so, we really don't have to say anything.

cmelchior: That depends if people understand "A RealmResult list containing the objects". Unfortunately it is not uncommon to find libraries that do return null, even though it is bad practice, so stating that the return value is not-null will be helpfull. Otherwise you will see some people checking for null.

Issue #765, opened 2015-01-19
RealmMigrationNeededException doesn't clear all caches


cmelchior:
It looks like following code, doesn't properly clear the Realm cache, resulting in Realms with old schemas to exist which can cause NPE like this:

Code to reproduce:
```
@Override
protected void onCreate(Bundle savedInstanceState) {
  super.onCreate(savedInstanceState);
  try {
            realm = Realm.getInstance(this);
        } catch (RealmMigrationNeededException e) {
            Realm.deleteRealmFile(this);
            realm = Realm.getInstance(this);
        }
  realm.allObjects(Person.class).findAll();
}
```

Error:
```
Caused by: java.lang.NullPointerException
                        E      at io.realm.PersonRealmProxy.setName(PersonRealmProxy.java:36)
                        E      at io.realm.examples.intro.IntroExampleActivity.basicCRUD(IntroExampleActivity.java:101)
                        E      at io.realm.examples.intro.IntroExampleActivity.onCreate(IntroExampleActivity.java:59)
                        E      at android.app.Activity.performCreate(Activity.java:5231)
                        E      at android.app.Instrumentation.callActivityOnCreate(Instrumentation.java:1087)
                        E      at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2148)
```

PR #768, opened 2015-01-20
Fix Realm cache during migration

cmelchior:
This makes sure that Realms aren't cached if a RealmMigrationNeededException is thrown. It could result in references to wrong or closed Realms being returned on next call to Realm.getInstance(). Reported here: https://github.com/realm/realm-java/issues/765

@emanuelez @kneth @bmunkholm


kneth:
:+1:

realm/src/main/java/io/realm/internal/Group.java
```diff
@@ -215,12 +215,15 @@ public String getTableName(int index) {
      */
     public Table getTable(String name) {
         verifyGroupIsValid();
-        if (name == null || name.equals(""))
+        if (name == null || name.equals("")) {
```
emanuelez: name.isEmpty() is better

emanuelez:
Minor comment, otherwise :+1:

changelog.txt
```diff
@@ -1,3 +1,7 @@
+0.78
+ * Fixed bug causing Realms to be cached during a RealmMigration, resulting in invalid realms being returned.
```
bmunkholm: The description is better in the description of this PR :-)
e.g. "realms being returned" - from where?

bmunkholm:
Otherwise :+1:

PR #770, opened 2015-01-20
Add support for encrypted copies of Realms

emanuelez:
Left to do:
 * [x] Implementation
 * [x] Unit tests
 * [x] Changelog


emanuelez:
Changelog:
 * Added support to write an encrypted copy of a Realm

emanuelez:
@kneth @cmelchior @bmunkholm

realm/src/androidTest/java/io/realm/RealmTest.java
```diff
@@ -73,13 +76,20 @@ protected void setColumnData() {
...
+    }
 
     @Override
     protected void tearDown() throws Exception {
         if (testRealm != null) {
             testRealm.close();
         }
+
+        // Delete the encrypted Realm if present
+        File encryptedRealm = new File(getContext().getFilesDir(), ENCRYPTED_REALM_FILE_NAME);
```
cmelchior: Only one unit test uses this. Better move it there. Just check at the start of the unit test and delete if needed?

realm/src/main/java/io/realm/internal/Group.java
```diff
@@ -236,34 +236,28 @@ public Table getTable(String name) {
...
      *            A File object representing the file.
+     * @param key A 64 bytes long byte array containing the key to the encrypted Realm file. Can be null if
+     *            encryption is not required.
      * @throws IOException
      */
-    public void writeToFile(File file) throws IOException {
+    public void writeToFile(File file, byte[] key) throws IOException {
         verifyGroupIsValid();
-        nativeWriteToFile(nativePtr, file.getAbsolutePath(), null);
+        if (file.isFile() && file.exists()) {
```
cmelchior: if file is null this code will fail.

cmelchior:
Apart from comments :+1:
And we should probably make a styleguide document somewhere. I noticed that core has one :)

realm/src/androidTest/java/io/realm/RealmTest.java
emanuelez: I tend to disagree. This makes the actual test cleaner and easier to read IMO.

realm/src/main/java/io/realm/internal/Group.java
emanuelez: We should really start using @Nullable and @NotNull: https://developer.android.com/reference/android/support/annotation/NonNull.html

cmelchior: Would be nice, except you would add Android dependencies all over the place.
This is also a scary read: http://stackoverflow.com/questions/4963300/which-notnull-java-annotation-should-i-use

emanuelez: Oops... this sneaked in already. Was probably me :) https://github.com/realm/realm-java/blob/master/realm/build.gradle#L29

realm/src/androidTest/java/io/realm/RealmTest.java
```diff
@@ -960,4 +978,25 @@ private void fileCopy(File src, File dst) throws IOException {
...
+            testRealm.writeEncryptedCopyTo(destination, key);
+        } catch(Exception e) {
+            fail();
+        }
+
+        Realm encryptedRealm = Realm.getInstance(getContext(), ENCRYPTED_REALM_FILE_NAME, key);
+        long after = encryptedRealm.where(AllTypes.class).count();
+        assertEquals(TEST_DATA_SIZE, after);
+        encryptedRealm.close();
+    }
```
bmunkholm: Add a test to verify that you can write the new file with a new key. Or from encrypted to non-encrypted.

kneth:
:+1:

Issue #763, opened 2015-01-19
Involuntary invocation of close()


gillde:
I have three AsyncTaskLoaders in which I make calls to a REST Api.

Loader A and B: Two of them are simple one-shot calls. I receive a json (JSONObject) and put that object into my Realm.
Loader C: The third one has a while loop. In this loop I retrieve a json (JSONArray) from my server and add them to my Realm. So, in short, the loop approx loops for ~80 times and within each cycle I begin a transaction, put a jsonstring consisting of 500 objects into my realm and commit the transaction.

My app is really small. I have a Realm.getInstance(this) in onStart and realm.close() in onStop of my Activity.
Within each loader I have the mandatory loadInBackground. At the beginning of that method I have Realm.getInstance(context) and at the end of it I have realm.close(). So I can see no duplicate or redundant close() call in my app.

When I start my app and start Loader C everything works fine. If I start my app and start loader A and B and after that C, I get "Calling close() on a Realm that is already closed: /data/data/*/files/default.realm" TWICE at the first two iterations of the loop, but everything is added correctly to my Realm.

Debugging calls to Realm.close() shows that it is called with the following stack (but not called by me)
close():181, Realm (io.realm)
finalize():169, Realm (io.realm)
doFinalize():190, Daemons$FinalizerDaemon (java.lang)
run():173, Daemons$FinalizerDaemon (java.lang)
run():818, Thread (java.lang)

That call to close() gives me the above mentioned output.

If I can investigate more, let me now :)

Realm Version: 0.77

petersnoopy:
Same problem happens when making a few asynchronous webservice calls using Retrofit.

cmelchior:
@gillde 
Thank you for the detailed bug report. However, I am having a bit of a problem reproducing it. I created a small sample Activity from your description that is listed below. Running that works as expected and doesn't output any close() warnings. Can you see where your code differs from mine or even better reduce your code to breaking example? 

```
public class IntroExampleActivity extends Activity implements LoaderManager.LoaderCallbacks<Boolean> {

    public static final String TAG = IntroExampleActivity.class.getName();
    private LinearLayout rootLayout = null;

    private Realm realm;

    @Override
    protected void onCreate(Bundle savedInstanceState) {
        super.onCreate(savedInstanceState);
        setContentView(R.layout.activity_realm_basic_example);
        rootLayout = ((LinearLayout) findViewById(R.id.container));
        rootLayout.removeAllViews();
[...]

PR #773, opened 2015-01-21
Call to close() replaced by warning in Realm finalizer.

cmelchior:
Calling close() in finalizer led to people getting warnings even though they correctly closed their Realm instance. Now a warning is emitted instead, much like Android cursors. Reported here: https://github.com/realm/realm-java/issues/763

No unit test accompanies this as unit testing finalisers is as far as I know impossible without incurring a huge and unacceptable time penalty.


cmelchior:
@emanuelez @kneth @bmunkholm

realm/src/main/java/io/realm/Realm.java
```diff
@@ -166,7 +166,9 @@ private Realm(String absolutePath, byte[] key, boolean autoRefresh) {
 
     @Override
     protected void finalize() throws Throwable {
-        this.close();
+        if (sharedGroup != null) {
+            RealmLog.w("Realm " + path + " is being finalized without being closed. This can lead to running out of memory.");
```
emanuelez: I would maybe specify *native* memory.

bmunkholm: Maybe add "Remember to call close() on all realm instances."

bmunkholm:
miss update of changelog

PR #775, opened 2015-01-21
[WIP] Include version in exceptions

kneth:
This pull request introduces two things:

* Version to exception thrown at JNI level. 
* A utility to set version (`tools/set-version.sh`)

Setting the version for a new release requires that `tools/set-version.sh` and `realm-jni/generate-jni-headers.sh` to be run.

@emanuelez @cmelchior @bmunkholm

### DIFF
--- a/realm-jni/src/io_realm_internal_Util.cpp
+++ b/realm-jni/src/io_realm_internal_Util.cpp
@@ -17,6 +17,7 @@
 #include "util.hpp"
 #include "mem_usage.hpp"
 #include "io_realm_internal_Util.h"
+#include "io_realm_internal_Version.h"
 
 using std::string;
 
@@ -50,7 +51,6 @@ JNIEXPORT jstring JNICALL Java_io_realm_internal_Util_nativeTestcase(
     JNIEnv *env, jclass, jint testcase, jboolean dotest, jlong)
 {
     string expect;
-
     switch (ExceptionKind(testcase)) {
         case ClassNotFound:
             expect = "java.lang.ClassNotFoundException: Class 'parm1' could not be located.";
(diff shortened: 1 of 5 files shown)


tools/set-version.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+
+if [ -z "$1" ]; then
+    echo "You must supply a version number"
```
cmelchior: Describe what a valid number looks like.

realm-jni/src/io_realm_internal_Util.cpp
```diff
@@ -127,6 +127,9 @@ JNIEXPORT jstring JNICALL Java_io_realm_internal_Util_nativeTestcase(
     if (dotest) {
         return NULL;
     }
-    return to_jstring(env, expect);
+    string version = "Realm version " + num_to_string<int>(io_realm_internal_Version_REALM_JAVA_MAJOR) + "."
```
cmelchior: Debatting with myself if it is too verbose having the "Realm version". Wouldn't just the version number be enough as it is for people copy / pasting errors to us?

tools/set-version.sh
```diff
@@ -0,0 +1,23 @@
...
+    exit 1
+fi
+
+realm_java_version="$1"
+realm_java_version_file="realm/src/main/java/io/realm/internal/Version.java"
+realm_java_ver_major="$(echo "$realm_java_version" | cut -f1 -d.)" || exit 1
```
cmelchior: Why not just use ```set -e``` at the top of the file?

realm-jni/src/io_realm_internal_Util.cpp
bmunkholm: maybe "realm-java-VERSION"

emanuelez:
I think this is not the best way to do this. We should only update version.txt and everything else should be handled by gradle

bmunkholm:
@kneth: Should this be closed

kneth:
@bm: Yes, it looks that we never really wished to add the version number.

PR #778, opened 2015-01-22
Sorting RealmResults is stable.

kneth:
@cmelchior


cmelchior:
:+1:

realm/src/androidTest/java/io/realm/SortTest.java
```diff
@@ -0,0 +1,202 @@
+/*
```
bmunkholm: All other files that test a class is called CLASSTest.java, so when we test cross class functionality, maybe we should use another naming practice.

realm/src/androidTest/java/io/realm/SortTest.java
```diff
@@ -0,0 +1,202 @@
...
+    }
+
+    public void testSortMultiFailures() {
+        RealmResults<AllTypes> allTypes = testRealm.allObjects(AllTypes.class);
+
+        // zero fields specified
+        try {
+            allTypes.sort(new String[]{}, new boolean[]{});
+            fail();
+        } catch (IllegalArgumentException ignored) {}
```
bmunkholm: Let's wrap the last }

realm/src/androidTest/java/io/realm/SortTest.java
```diff
@@ -0,0 +1,202 @@
...
+
+        // number of fields and sorting orders don't match
+        try {
+            allTypes.sort(new String[]{FIELD_STRING}, new boolean[]{RealmResults.SORT_ORDER_ASCENDING, RealmResults.SORT_ORDER_ASCENDING});
```
bmunkholm: The ascending/ascending bool array is duplicate all over. Suggest to setup and reuse once for simpler reading.

bmunkholm:
Should also test descending/ascending sort.

realm/src/androidTest/java/io/realm/SortTest.java
kneth: We do have a `NotificationTest.java` which tests functionality and not a class. I extrapolated a naming practice from this example :-)

kneth:
I have cleaned up the unit tests (increase code reuse) and added extra tests.

@bmunkholm

emanuelez:
:+1:

cmelchior:
:+1:

PR #779, opened 2015-01-22
Include the encryption example in the distribution package

emanuelez:
Also adds it for monkey tests


emanuelez:
@kneth @bmunkholm @cmelchior

tools/monkey-examples.sh
```diff
@@ -20,11 +20,6 @@ for example in examples/*/ ; do
     applicationId=$(grep applicationId "${example}build.gradle" | cut -d \" -f 2 | cut -d \' -f 2)
     project=$(basename ${example})
 
-    if [ ${project} == "encryptionExample" ] ; then
```
kneth: Nice job

distribution/RealmEncryptionExample/app/build.gradle
```diff
@@ -0,0 +1,24 @@
...
+    buildTypes {
+        release {
+            minifyEnabled false
+            proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
+        }
+    }
+}
+
+dependencies {
+    compile 'io.realm:realm-android:0.77.0'
```
cmelchior: This is different than the other examples where we use: 

```
tasks.preBuild {
    dependsOn ":realm:androidJar"
}

dependencies {
    compile files("../../realm/build/libs/realm-${version}.jar")
}
```

cmelchior:
Do we have any speed measurements between a encrypted and non-encrypted Realm?

distribution/RealmEncryptionExample/app/build.gradle
emanuelez: This is the distribution example so it uses the jcenter version

cmelchior: Doh :+1:

bmunkholm:
We have for cocoa. If I remember correctly it's around < 20% slower. That should be roughly the same for Java. But worth measuring since the binding overhead may be different.

bmunkholm:
:+1:

PR #783, opened 2015-01-23
Better error reporting opening Realms

cmelchior:
Added additional checks so invalid paths to Realm files are caught early.
This is currently done in Java to prevent adding to much unneeded complexity to the rest of the internal API.

@bmunkholm @emanuelez @kneth


realm/src/main/java/io/realm/internal/SharedGroup.java
```diff
@@ -234,15 +234,16 @@ public void reserve(long bytes) {
...
                                      int durabilityValue,
                                      boolean no_create,
                                      boolean enableReplication,
                                      byte[] key);
 
     private void checkNativePtrNotZero() {
-        if (this.nativePtr == 0)
-            throw new java.lang.OutOfMemoryError("Out of native memory.");
+        if (this.nativePtr == 0) {
+            throw new java.lang.OutOfMemoryError("Out of native memory. Realm could not be opened");
```
bmunkholm: We should remove the "out of memory" part - that's not the only reason.

realm/src/main/java/io/realm/internal/SharedGroup.java
```diff
@@ -234,15 +234,16 @@ public void reserve(long bytes) {
 
     private native void nativeRollback(long nativePtr);
 
-    private native long createNative(String databaseFile,
+    private native long nativeCreate(String databaseFile,
```
bmunkholm: renaming a jni method should be reflected in the header and cpp file as well - can't see that's updated?

realm/src/androidTest/java/io/realm/RealmTest.java
```diff
@@ -1055,5 +1110,4 @@ public void testWriteEncryptedCopy() throws Exception {
             }
         }
     }
-
-}
+}
```
bmunkholm: add newline

realm/src/androidTest/java/io/realm/RealmTest.java
```diff
@@ -109,6 +110,60 @@ private void populateTestRealm() {
     }
 
 
+    public void testGetInstanceNullFolder() {
```
bmunkholm: naming of all the methods should something like testXXXthrows

bmunkholm:
Did you try the testcases without the java code? Most of that should already be captured in cpp. But the error messages could maybe be improved. In that case we are just adding unnecessary code.

cmelchior:
I did yes, and core throws errors for all of them. But it is very "internal" errors and they are hard to interpret from a user standpoint. I guess that is why Cocoa also does its own error checking.

The error messages would currently need to be fixed in core itself, which makes it difficult to have a Java relevant error message. We could do it in JNI, but we would have to complicate a lot of internal methods for very little benefit. Also having them this close to user input makes it a lot easier for users to see what is going wrong, as the stacktrace becomes shorter.

realm/src/main/java/io/realm/internal/SharedGroup.java
cmelchior: Then we should probably change it to an IOError instead of OutOfMemory as well?

realm/src/main/java/io/realm/Realm.java
```diff
@@ -455,6 +456,22 @@ private static Realm createAndValidate(String absolutePath, byte[] key, boolean
         return realm;
     }
 
+    private static void assertValidRealmPath(File writableFolder, String filename) {
+        if (filename == null || filename.isEmpty()) {
+            throw new IllegalArgumentException("Non-empty filename must be provided");
+        }
+        if (writableFolder == null || !writableFolder.isDirectory()) {
+            throw new IllegalArgumentException(("A existing folder must be provided. Yours was " + (writableFolder != null ? writableFolder.getAbsolutePath() : "null")));
```
kneth: "A" -> "An"

realm/src/main/java/io/realm/Realm.java
```diff
@@ -413,6 +413,7 @@ public static Realm getInstance(File writeableFolder, String fileName, byte[] ke
     }
 
     private static Realm create(File writableFolder, String filename, byte[] key) {
+        assertValidRealmPath(writableFolder, filename);
```
emanuelez: I would avoid the word assert here. Maybe "check" would be better. Just a detail of course

realm/src/main/java/io/realm/Realm.java
```diff
@@ -455,6 +456,15 @@ private static Realm createAndValidate(String absolutePath, byte[] key, boolean
         return realm;
     }
 
+    private static void assertValidRealmPath(File writableFolder, String filename) {
```
emanuelez: How about using @Notnull and evoiding the check?

cmelchior: It is still a IntelliJ only annotation, so I would rather not.

cmelchior:
Ready for 2nd round

@emanuelez @kneth @bmunkholm

realm/src/androidTest/java/io/realm/RealmTest.java
```diff
@@ -109,6 +110,60 @@ private void populateTestRealm() {
...
+            fail("Assuming that subfolders are created automatically should fail");
+        } catch (IllegalArgumentException expected) {
+        }
+    }
+
+    public void testGetInstanceFolderNoWritePermissionThrows() {
+        File folder = new File("/");
+        try {
+            Realm realm = Realm.getInstance(folder);
+            fail("Pointing to a read-only folder should throw an error");
```
kneth: "read-only" is different from "NoWritePermission" :-)

kneth:
:+1:

emanuelez:
:+1:

PR #787, opened 2015-01-23
Throw an exception if deleting an open Realm file

emanuelez:


emanuelez:
Changelog:
 * Throw an exception if deleteRealmFile() is called when there is an open instance of the Realm

realm/src/main/java/io/realm/Realm.java
```diff
@@ -428,6 +437,15 @@ private static Realm createAndValidate(String absolutePath, byte[] key, boolean
         if (references == null) {
             references = 0;
         }
+        if (references == 0) {
+            AtomicInteger counter = openRealms.get(id);
```
cmelchior: This doesn't look threadsafe, two threads can call this before the value is inserted, making the counter miss one. Testing for null and inserting new AtomicInteger must be a atomic operation. I don't think there is any way around a synchronized block for this?

realm/src/main/java/io/realm/Realm.java
```diff
@@ -111,6 +112,9 @@
     };
     private static final int REALM_CHANGED = 14930352; // Just a nice big Fibonacci number. For no reason :)
     protected static final Map<Handler, Integer> handlers = new ConcurrentHashMap<Handler, Integer>();
+
+    // Maps ids to a boolean set to true if the Realm is open. This is basically only needed by deleteRealmFile
```
cmelchior: Remove basically.

realm/src/androidTest/java/io/realm/NotificationsTest.java
```diff
@@ -38,6 +38,9 @@
 
     @Override
     protected void setUp() throws Exception {
+        for (Integer key : Realm.openRealms.keySet()) {
```
cmelchior: Having this seem to indicate that some other unit tests are not behaving like they should? Perhaps add a comment about that.

cmelchior:
Apart from comments :+1:  Will be nice to get this check in.

realm/src/main/java/io/realm/Realm.java
```diff
@@ -111,6 +112,9 @@
     };
     private static final int REALM_CHANGED = 14930352; // Just a nice big Fibonacci number. For no reason :)
     protected static final Map<Handler, Integer> handlers = new ConcurrentHashMap<Handler, Integer>();
+
```
bmunkholm: (This class could use some reordering of all the contants and class variables.)

emanuelez:
Ready for review: @cmelchior @kneth @bmunkholm

cmelchior:
Our notification unit tests are getting really hairy, but I have no idea if we can simplify them somehow.
:+1:

kneth:
Is this a real problem? I mean, Linux does refcounting of files, and a file is deleted when no process have  it open. Deleting a file which is in use, shouldn't cause any problem (in theory).

On the other hand, the check - and the exception - might help a developer to understand why her app is failing as deleting an open file might be a side effect of a bug somewhere else.

realm/src/main/java/io/realm/Realm.java
```diff
@@ -191,6 +195,11 @@ public void close() {
             realmsCache.get().remove(id);
             sharedGroup.close();
             sharedGroup = null;
+            AtomicInteger counter = openRealms.get(id);
+            if (counter == null) {
+                throw new AssertionError("The counter should not be zero");
```
bmunkholm: I assume this assertion should not be possible, it's basically unneeded code as the decrementAndGet would fail with nullException anyway. So is there much value in having this?

realm/src/main/java/io/realm/Realm.java
```diff
@@ -421,13 +430,23 @@ private static Realm create(File writableFolder, String filename, byte[] key) {
         }
     }
 
-    private static Realm createAndValidate(String absolutePath, byte[] key, boolean validateSchema, boolean autoRefresh) {
+    private static synchronized Realm createAndValidate(String absolutePath, byte[] key, boolean validateSchema,
```
bmunkholm: Do you need to sync the whole method? I prefer only the code block needed for clarity of what needs synced.

realm/src/main/java/io/realm/Realm.java
emanuelez: Of course it can go. I just hate NPEs :)

realm/src/main/java/io/realm/Realm.java
emanuelez: @cmelchior and I have been discussing this for a while, not just for this PR.
If we don't synchronize this method concurrent initialization might occur, which would be very bad of course

realm/src/androidTest/java/io/realm/RealmTest.java
```diff
@@ -1056,4 +1056,29 @@ public void testWriteEncryptedCopy() throws Exception {
         }
     }
 
+    public void testRealmFileDeletion() {
```
bmunkholm: Add comments about what's being tested just before it's tested.

realm/src/androidTest/java/io/realm/RealmTest.java
```diff
@@ -1056,4 +1056,29 @@ public void testWriteEncryptedCopy() throws Exception {
...
+        final String otherRealmName = "yetAnotherRealm.realm";
+
+        try {
```
bmunkholm: add:
// Deleting open realm must throw.

realm/src/androidTest/java/io/realm/RealmTest.java
```diff
@@ -1056,4 +1056,29 @@ public void testWriteEncryptedCopy() throws Exception {
...
+            Realm.deleteRealmFile(getContext());
+            fail();
+        } catch (Exception ignored) {
```
bmunkholm: Only check IllegalStateException

realm/src/main/java/io/realm/Realm.java
```diff
@@ -1402,9 +1418,17 @@ public static boolean deleteRealmFile(Context context) {
...
-        List<File> filesToDelete = Arrays.asList(
-                new File(writableFolder, fileName),
-                new File(writableFolder, fileName + ".lock"));
+
+        File realmFile = new File(writableFolder, fileName);
+        int realmId = realmFile.getAbsolutePath().hashCode();
+
+        AtomicInteger counter = openRealms.get(realmId);
+        if (counter != null && counter.get() > 0) {
+            throw new IllegalStateException("It's not allowed to delete the file associated to an open Realm. " +
```
kneth: "to" -> "with"

realm/src/main/java/io/realm/Realm.java
```diff
@@ -1402,9 +1418,17 @@ public static boolean deleteRealmFile(Context context) {
...
         boolean result = true;
         File writableFolder = context.getFilesDir();
-        List<File> filesToDelete = Arrays.asList(
-                new File(writableFolder, fileName),
-                new File(writableFolder, fileName + ".lock"));
+
+        File realmFile = new File(writableFolder, fileName);
+        int realmId = realmFile.getAbsolutePath().hashCode();
+
+        AtomicInteger counter = openRealms.get(realmId);
```
kneth: What happens if two threads call `deleteRealmFile()` at the same time?

emanuelez: Very good point. This method (and many others!!) should be synchronized

realm/src/androidTest/java/io/realm/RealmTest.java
```diff
@@ -1056,6 +1056,35 @@ public void testWriteEncryptedCopy() throws Exception {
         }
     }
 
+    public void testOpenRealmFileDeletionShouldThrow() {
+        final String OTHER_REALM_NAME = "yetAnotherRealm.realm";
+
+        // This instance is already cached because of the setUp() method so this deletion should throw
```
bmunkholm: :100: Love the test "story" I can now read.

bmunkholm:
:+1:

tgoyne:
> Deleting a file which is in use, shouldn't cause any problem (in theory).

Deleting the Realm file while it's open works fine on unixy platforms, but you then run into trouble with the write logs if you open a new SharedGroup at the same path. When the last SharedGroup for the old file is closed, it deletes the files at the paths it's expecting its writelogs to be at, but they're actually the writelogs for the new SharedGroup.

Issue #668, opened 2014-12-12
Improve error handling to avoid "NullPointerException" when getters/setters are forgotten.


Queatz:
Unable to build project using realm due to an NPE during build:

```
23:51:03.024 [ERROR] [org.gradle.BuildExceptionReporter] Caused by: java.lang.NullPointerException
23:51:03.025 [ERROR] [org.gradle.BuildExceptionReporter] 	at com.squareup.javawriter.JavaWriter.beginMethod(JavaWriter.java:401)
23:51:03.025 [ERROR] [org.gradle.BuildExceptionReporter] 	at com.squareup.javawriter.JavaWriter.beginMethod(JavaWriter.java:378)
23:51:03.026 [ERROR] [org.gradle.BuildExceptionReporter] 	at io.realm.processor.RealmProxyClassGenerator.emitAccessors(RealmProxyClassGenerator.java:199)
23:51:03.026 [ERROR] [org.gradle.BuildExceptionReporter] 	at io.realm.processor.RealmProxyClassGenerator.generate(RealmProxyClassGenerator.java:171)
23:51:03.026 [ERROR] [org.gradle.BuildExceptionReporter] 	at io.realm.processor.RealmProcessor.process(RealmProcessor.java:239)
23:51:03.027 [ERROR] [org.gradle.BuildExceptionReporter] 	at com.sun.tools.javac.processing.JavacProcessingEnvironment.callProcessor(JavacProcessingEnvironment.java:793)
23:51:03.027 [ERROR] [org.gradle.BuildExceptionReporter] 	at com.sun.tools.javac.processing.JavacProcessingEnvironment.discoverAndRunProcs(JavacProcessingEnvironment.java:722)
23:51:03.028 [ERROR] [org.gradle.BuildExceptionReporter] 	at com.sun.tools.javac.processing.JavacProcessingEnvironment.access$1700(JavacProcessingEnvironment.java:97)
23:51:03.028 [ERROR] [org.gradle.BuildExceptionReporter] 	at com.sun.tools.javac.processing.JavacProcessingEnvironment$Round.run(JavacProcessingEnvironment.java:1029)
23:51:03.029 [ERROR] [org.gradle.BuildExceptionReporter] 	at com.sun.tools.javac.processing.JavacProcessingEnvironment.doProcessing(JavacProcessingEnvironment.java:1163)
[...]

PR #793, opened 2015-01-26
Missing accessors no longer crash annotation compiler

cmelchior:
Fix for this: https://github.com/realm/realm-java/issues/668 + unit test
Errors are reported for all files at once instead of aborting the annotation processor.

@emanuelez @bmunkholm @kneth


emanuelez:
:+1:

PR #803, opened 2015-01-30
Build the distribution package in Gradle

emanuelez:
This commit does the following:
 * converts the build-distribution script to Gradle
 * automates the Realm version of distribution examples
 * generates the distribution archive


emanuelez:
@bmunkholm @cmelchior @kneth

realm/build.gradle
```diff
@@ -138,3 +138,84 @@ tasks.preBuild {
     dependsOn compileAnnotations
     dependsOn buildApt
 }
+
+def examples = [
```
cmelchior: Do we really want to enumerate all the projects we want to include. Wouldn't it be better to name the ones we *don't* want. I would assume that would be easier to maintain as the most common use case would be to include them?

cmelchior:
Apart from comments :+1:

realm/build.gradle
emanuelez: My evil master-plan is to completely automate the generation of the distribution folder.
But right now it's nice to have the flexibility to call the example folders in any way, without having to follow a convention.

cmelchior: :+1:

PR #811, opened 2015-02-02
Fail when instantiating a Realm with the wrong key

emanuelez:


emanuelez:
Changelog:
 * Fixed error when instantiating a Realm with the wrong key

emanuelez:
@cmelchior @kneth @bmunkholm

realm/src/main/java/io/realm/Realm.java
```diff
@@ -114,12 +114,15 @@
     private static final String APT_NOT_EXECUTED_MESSAGE = "Annotation processor may not have been executed.";
     private static final String INCORRECT_THREAD_MESSAGE = "Realm access from incorrect thread. Realm objects can only be accessed on the thread they where created.";
     private static final String CLOSED_REALM = "This Realm instance has already been closed, making it unusable.";
+    private static final String INVALID_KEY_MESSAGE = "The provided key is illegal. It should either be null or be 64 bytes long";
```
cmelchior: This sentence could be clearer i think.  Perhaps rephrase to "Illegal key length. It should either be null or 64 bytes long"

realm/src/main/java/io/realm/Realm.java
```diff
@@ -114,12 +114,15 @@
...
+    private static final String DIFFERENT_KEY_MESSAGE = "Realm access with incorrect key";
```
cmelchior: "Incorrect" is not very clear imo. Perhaps rephrase to "Wrong key used to decrypt Realm".

realm/src/main/java/io/realm/Realm.java
emanuelez: That was my first version, but null has no length so I changed to this instead

realm/src/main/java/io/realm/Realm.java
emanuelez: Yes, I like yours better

realm/src/main/java/io/realm/Realm.java
```diff
@@ -114,12 +114,15 @@
...
 
     @SuppressWarnings("UnusedDeclaration")
     private static SharedGroup.Durability defaultDurability = SharedGroup.Durability.FULL;
     private boolean autoRefresh;
     private Handler handler;
 
+    private final byte[] key;
```
cmelchior: It is a nitpick, but this implies that the key is present in memory. If an attacker has access to the phone at this point he can read the key directly. He got access to the database anyway at this point, but having the key would grant future access as well. 

I would probably hash this value using SHA-256 or something. It doesn't mean you cannot intercept the key anyway, but It would make it a little harder.

cmelchior:
Apart from comments :+1:

realm/src/main/java/io/realm/Realm.java
```diff
@@ -114,12 +114,15 @@
     private static final String APT_NOT_EXECUTED_MESSAGE = "Annotation processor may not have been executed.";
     private static final String INCORRECT_THREAD_MESSAGE = "Realm access from incorrect thread. Realm objects can only be accessed on the thread they where created.";
     private static final String CLOSED_REALM = "This Realm instance has already been closed, making it unusable.";
+    private static final String INVALID_KEY_MESSAGE = "The provided key is illegal. It should either be null or be 64 bytes long.";
```
kneth: "illegal" -> "invalid"

kneth:
:+1:

realm/src/androidTest/java/io/realm/RealmTest.java
```diff
@@ -1056,4 +1056,27 @@ public void testWriteEncryptedCopy() throws Exception {
         }
     }
 
+    public void testWrongKeyShouldFail() {
```
bmunkholm: maybe rename to testWrongKeyShouldThrow

realm/src/androidTest/java/io/realm/RealmTest.java
```diff
@@ -1056,4 +1056,27 @@ public void testWriteEncryptedCopy() throws Exception {
...
+        final String WRONG_KEY_REALM = "wrong-key-realm.realm";
+
+        // Wrong key size
+        try {
+            Realm.getInstance(getContext(), WRONG_KEY_REALM, new byte[63]);
+            fail();
+        } catch (IllegalArgumentException ignored) {
+        } catch (Exception ignored) {
```
bmunkholm: you don't need to catch Exception - it would fail if we don't catch it. (although it would not test the other cases - but I'm fine with that)

realm/src/androidTest/java/io/realm/RealmTest.java
```diff
@@ -1056,4 +1056,27 @@ public void testWriteEncryptedCopy() throws Exception {
...
+            fail();
+        }
+
+        Realm.getInstance(getContext(), WRONG_KEY_REALM);
+
+        try {
+            Realm.getInstance(getContext(), WRONG_KEY_REALM, new byte[64]);
+            fail();
+        } catch (IllegalStateException ignored) {
+        } catch (Exception ignored) {
```
bmunkholm: same

bmunkholm:
otherwise :+1:

realm/src/androidTest/java/io/realm/RealmTest.java
emanuelez: I'm trying to make sure we throw the right exception and be explicit about it. If I don't catch exception and then somebody adds a "thows Exception" to the declaration of the testcase (as it is normally, even autogenerated test templates by Android Studio) the test would pass when it should not.

bmunkholm: so this is a new "standard" you propose we use for these kind of tests then?

bmunkholm: have we ever experienced this issue? It just complicates a lot of test cases if we have to do that in case someone change our tests. Surely if you change the tests, they work differently. Catching all exceptions in any testcase would make all our testcases fail? So then we need it for every single testcase - even the positive tests?

PR #826, opened 2015-02-06
Fix findFirst() for queried RealmList

chibatching:
I found the bug of RealmQuery#findFirst() for queried ReamList.
FindFirst() may return different result of FindAll().get(0).

Code to reproduce:

```Java
public class Book extends RealmObject {
    private String id;
    private RealmList<Page> pages;
}
public class Page extends RealmObject {
    private String id;
}

String book = "book_city";
String page = "city_japan";
RealmList<Page> pages = realm.where(Book.class)
                             .equalTo("id", book)
                             .findFirst()
                             .getPages();

RealmList<Page> page1 = pages.where()
                             .equalTo("id", page)
                             .findAll()
                             .get(0);

RealmList<Page> page2 = pages.where()
                             .equalTo("id", page)
                             .findFirst();
```

Data:

```
[book]      [page]
book_food   food_fast
book_food   food_chinese
book_food   food_japanese
book_city   city_america
book_city   city_japan
```

Results:

```
page1 -> city_japan   -- Expected
page2 -> food_chinese -- Not expected
```

I don't know whether my PR is best way.
But RealmQuery#findFirst() should look queried table or view because returned rowIndex is not source row index.

### DIFF
--- a/realm/src/main/java/io/realm/RealmList.java
+++ b/realm/src/main/java/io/realm/RealmList.java
@@ -22,7 +22,7 @@ import java.util.List;
 
 import io.realm.exceptions.RealmException;
 import io.realm.internal.LinkView;
-import io.realm.internal.TableQuery;
+import io.realm.internal.TableOrView;
 
 /**
  * RealmList is used to model one-to-many relationships in a {@link io.realm.RealmObject}.
@@ -236,8 +236,8 @@ public class RealmList<E extends RealmObject> extends AbstractList<E> {
      */
     public RealmQuery<E> where() {
         if (managedMode) {
-            TableQuery query = this.view.where();
-            return new RealmQuery<E>(this.realm, query, clazz);
+            TableOrView table = this.view.where().findAll();
+            return new RealmQuery<E>(this.realm, table, clazz);
         } else {
             throw new RealmException(ONLY_IN_MANAGED_MODE_MESSAGE);
         }
--- a/realm/src/main/java/io/realm/RealmList.java
+++ b/realm/src/main/java/io/realm/RealmList.java
@@ -22,7 +22,6 @@ import java.util.List;
 
 import io.realm.exceptions.RealmException;
 import io.realm.internal.LinkView;
-import io.realm.internal.TableOrView;
 
 /**
  * RealmList is used to model one-to-many relationships in a {@link io.realm.RealmObject}.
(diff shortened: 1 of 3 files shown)


cmelchior:
Hi @chibatching 
Thanks for finding and fixing the bug. A few initial comments:

- In order for us to accept it, you need to sign the contributors form first: https://github.com/realm/realm-java/blob/master/CONTRIBUTING.md

- Please add a unit test that catches the original error.

- Im a bit concerned about you executing a query in the RealmQuery.where() method, but I'll have to take a deeper look at the code first to understand if it is really required there.

realm/src/main/java/io/realm/RealmList.java
```diff
@@ -236,8 +236,8 @@ public int size() {
      */
     public RealmQuery<E> where() {
         if (managedMode) {
-            TableQuery query = this.view.where();
-            return new RealmQuery<E>(this.realm, query, clazz);
+            TableOrView table = this.view.where().findAll();
```
bmunkholm: We should definitely not do a query in where(). Youmust be able to build up a query and then execute it afterwards.

chibatching:
Hi @cmelchior 
Thank you for comment.
I sent CLA form and added unit test for checking this error.
Please check it.

@bmunkholm 
Thank you for reviewing my code.
I removed executing query in where(). Please re-review.

kneth:
This looks good to me, and I merge it now. We might have a few extra tricks in the core library to speed it up but we haven't exposed them in Java yet.

gillde:
I see that this was merged into master in february. But I think I am still have issues related to different results from findFirst() and findAll().first()

I have a RealmResults which I further want to query: ```mAllUsersForCompany``` is of type ```RealmResults```:

The following works. It gives me a user whos name has length of three and who was in the ```mAllUsersForCompany``` list:
```
mAllUsersForCompany.where().equalTo("nameLength", 3).findAll().first()
```

But the following seems to completely ignore the additional ```equalTo``` constraint and gives me a user which has namelength != 3 
```
mAllUsersForCompany.where().equalTo("nameLength", 3).findFirst()
```

cmelchior:
Hi @gillde 
I have moved your comment to a separate issue so it is more easily trackable as this has already been merged: https://github.com/realm/realm-java/issues/1018

PR #565, opened 2014-11-14
Primary keys

cmelchior:
~~WIP: Awaiting completion of https://github.com/realm/realm-java/issues/725~~
Primary key support added through @PrimaryKey annotation.
Primary keys are only allowed on String, short, int, long.
Only 1 field can be a primary key. 
Primary key is automatically indexed if possible.

Some performance numbers (each test was run 10 times):

Label | Result | All results
------------ | -------------| -------------
Insert(10K) - nopk | mean: 0.193 s. - avg: 0.200 s. | [275, 194, 202, 192, 187, 199, 188, 189, 198, 184]
Insert(10K) - pk,string | mean: 1.526 s. - avg: 1.540 s. | [1654, 1559, 1559, 1530, 1514, 1522, 1512, 1495, 1541, 1522]
Insert(10K) - pk,long | mean: 0.303 s. - avg: 0.305 s. | [338, 310, 306, 304, 295, 301, 308, 297, 302, 290]
Insert(10K) - pk,string,index | mean: 0.939 s. - avg: 0.941 s. | [975, 904, 928, 926, 940, 939, 927, 968, 956, 950]
Insert(50K) - nopk | mean: 0.935 s. - avg: 0.938 s. | [1006, 946, 932, 934, 936, 911, 916, 939, 922, 945]
Insert(50K) - pk,string | mean: 19.657 s. - avg: 19.840 s. | [19684, 19582, 19671, 19587, 19636, 19644, 19614, 20202, 20041, 20744]
Insert(50K) - pk,long | mean: 3.593 s. - avg: 3.642 s. | [3838, 3714, 3529, 3567, 3522, 3491, 3562, 3620, 3761, 3819]
Insert(50K) - pk,string,index | mean: 4.891 s. - avg: 4.887 s. | [5189, 4638, 4812, 4892, 4905, 4891, 4672, 4854, 5005, 5012]

PR #827, opened 2015-02-06
Adding support for createOrUpdate methods

cmelchior:
This PR extends https://github.com/realm/realm-java/pull/565 with createOrUpdate methods.
Json support has been refactored due to this to be able to support both standalone and managed objects.

The following methods has been added for JSON: 
```
Realm.createOrUpdateObjectFromJson(Class, JSONObject)
Realm.createOrUpdateObjectFromJson(Class, InputStream)
Realm.createOrUpdateObjectFromJson(Class, String)
Realm.createOrUpdateAllFromJson(Class, JSONArray)
Realm.createOrUpdateAllFromJson(Class, InputStream)
Realm.createOrUpdateAllFromJson(Class, String)
```

And for standalone objects
```
Realm.copyToRealmOrUpdate(RealmObject)
Realm.copyToRealmOrUpdate(Iterable RealmObject)
```

Especially the last two gets a bit long winded and could probably use some iteration on naming, other suggestions
```
Realm.createOrUpdate(RealmObject); // From iOS
Realm.addOrUpdate(RealmObject); 
Realm.copyOrUpdateToRealm(RealmObject); // similar naming scheme as JSON methods
Realm.copyOrUpdate(RealmObject); // shorter, would imply that copyToRealm gets renamed to copy()
Realm.updateOrCopyToRealm(RealmObject); // emphasis on updates, which is the primary use case.
Realm.copyToRealm(RealmObject,  Flag); // Flags could be UPDATE, , INSERT
Realm.update(RealmObject); // only updates, return null if object couldn't be updated.
```


astigsen:
> Realm.copyToRealmOrUpdate(RealmObject)
> Realm.copyToRealmOrUpdate(Iterable RealmObject)
> Especially the last two gets a bit long winded and could probably use some iteration on naming

How about just ´createOrUpdate()´. No reason to reiterate the that we are adding to a realm, given that the method is called on the realm.

Any specific reason that these methods are not on the RealmObject like in the cocoa binding?

cmelchior:
We discarded the use of ```create``` when working with standalone objects, as the only way of getting them into Realm is to create new proxy objects we can return to the user. Copy conveys the intent better that the returned object is different from the standalone object. The downside to that is that ```Realm.copy()``` is really unclear, hence ```Realm.copyToRealm()```.

I am however also a big fan of  ```Realm.create()``` and ```Realm.createOrUpdate()```.

I cannot remember the exact reason for putting them on Realm instead of RealmObject. A couple of guesses:

- The type system is different ie. we cannot prevent this from happening: ```FooClass.copyOrUpdate(BarClassInstance)```. Which would make for an very odd API. Having all those methods on Realm makes it more intuitive that you can throw any RealmObject at them.

- Working with collections get a bit awkward as you *have* to use RealmList. If you keep your standalone objects in any other collection (of which there is many) you have to convert it to a RealmList first.

astigsen:
> Copy conveys the intent better that the returned object is different from
the standalone object.

Well, if you call a method that creates a new object from an existing
object, you would probably expect to get the new object back. In some ways
this is even more clear than copy.

*Alexander Stigsen, CEO*
Realm Inc

148 Townsend St
San Fransisco, CA 94107
United States

cmelchior:
Some of the reasoning was also that we wanted a different name than Cocoa because the method behaves differently, ie. on Cocoa you convert the existing object, here we return a new object. However that distinction might be lost on people who only uses the API on a single platform.

But it is nice to get other eyes on the naming scheme, as many different viewpoints as possible is good, and I wouldn't be afraid of going back and renaming some of our existing methods if it made those from this PR fit in better.

cmelchior:
A few comments:

- Json methods has moved from RealmObject to their own class RealmJsonImpl. This is in order to support both standalone and proxy objects. It also cleans up RealmObject.

- Added Generator class for RealmJsonImpl.

- Refactored all annotation processor utility methods into Utils.

- Current name will be copyToRealmOrUpdate(), we go with readability before brevity.

- Added createOrUpdateJson* for all variants.

- Added unit tests.

Ready for review: @emanuelez @kneth 
Note that the failing unit test is unrelated and will be fixed.

realm-annotations-processor/src/main/java/io/realm/processor/RealmJSonImplGenerator.java
```diff
@@ -0,0 +1,127 @@
...
+        String qualifiedGeneratedClassName = String.format("%s.%s", REALM_PACKAGE_NAME, CLASS_NAME);
+        JavaFileObject sourceFile = processingEnvironment.getFiler().createSourceFile(qualifiedGeneratedClassName);
+        JavaWriter writer = new JavaWriter(new BufferedWriter(sourceFile.openWriter()));
+        writer.setIndent("    ");
+
+        writer.emitPackage(REALM_PACKAGE_NAME);
+        writer.emitEmptyLine();
+
+        writer.emitImports(
+                "android.util.JsonReader",
```
kneth: Are imports better to be in alphabetic order?

realm-annotations-processor/src/main/java/io/realm/processor/Utils.java
```diff
@@ -0,0 +1,133 @@
...
+    public static String stripPackage(String clazz) {
+        String[] parts = clazz.split("\\.");
+        if (parts.length > 0) {
+            return parts[parts.length - 1];
+        } else {
+            return clazz;
+        }
+    }
+
+    public static String joinStringList(List<String> strings, String separator) {
```
kneth: Why not use `String.join()`?

realm/src/androidTest/java/io/realm/RealmJsonTest.java
```diff
@@ -431,4 +433,148 @@ public void testCreateObjectFromJsonStream_nullValues() throws IOException {
...
     }
-}
+
+    public void testCreateOrUpdateObject_noPrimaryKeyThrows() {
+        try {
+            testRealm.createOrUpdateObjectFromJson(AllTypes.class, new JSONObject());
+        } catch (IllegalArgumentException expected) {
+            return;
+        }
+        fail();
```
kneth: Why not move `fail()` into `try` block?

realm/src/main/java/io/realm/Realm.java
```diff
@@ -755,6 +906,33 @@ private static void initializeRealm(Realm realm) {
     }
 
     /**
+     * Tries to update an existing object defined by its primary key with new JSON data. If no existing object could be
+     * found a new object will be saved in the Realm. This must happen inside a transaction.
```
kneth: "inside" -> "within"?

According to http://en.wiktionary.org/wiki/within "within" means "before a specific duration ends" and the transaction is the "specific duration".

realm/src/androidTest/java/io/realm/RealmJsonTest.java
cmelchior: This way it will fail on anything but the IllegalArgumentException.

realm-annotations-processor/src/main/java/io/realm/processor/RealmJSonImplGenerator.java
cmelchior: Yup, it has been changed

realm/src/main/java/io/realm/Realm.java
```diff
@@ -1138,6 +1372,7 @@ boolean contains(Class<?> clazz) {
      * @return A sorted RealmResults containing the objects.
      * @throws java.lang.IllegalArgumentException if a field name does not exist.
      */
+    @SuppressWarnings("unchecked")
```
emanuelez: If these methods are tested they should now appear as unchecked. Remove warning suppressor or implement unit test

realm/src/main/java/io/realm/Realm.java
```diff
@@ -1018,6 +1230,28 @@ public void writeEncryptedCopyTo(File destination, byte[] key) throws IOExceptio
...
+     * Updates a list of existing RealmObjects that is identified by their {@link io.realm.annotations.PrimaryKey} or create a
+     * new copy if no existing object could be found.
+     *
+     * @param objects   List of objects to update or copy into Realm.
+     * @return A list of all the new or updated RealmObjects.
+     *
+     * @throws java.lang.IllegalArgumentException if RealmObject is {@code null} or doesn't have a Primary key defined.
+     * @see {@link #copyToRealm(Iterable)}
+     */
+    public <E extends RealmObject> List<E> copyToRealmOrUpdate(Iterable<E> objects) {
```
emanuelez: Shouldn't we return a RealmList, so it can be queried if needed?

cmelchior: We should yes, but it is the same as for copyToRealm(Iterable). We need to be able to return a tableview containing all inserted values which currently isn't feasible, at least that was how I remember the conclusion from when we discussed it.

realm/src/main/java/io/realm/Realm.java
cmelchior: Even though they are private they are tested intensively through the copyOrUpdate unit tests, so it should be OK to mark them as unchecked.

PR #832, opened 2015-02-06
Thread example no longer has too long write transactions

cmelchior:
The thread example showcased some very bad behaviour by having a write transaction lasting 5 seconds. This blocks other threads from doing writes and could cause ANR on the UI thread.

@emanuelez @kneth @bmunkholm


kneth:
:+1:

Maybe the above observation is worth mentioning in the documentation (or FAQ).

cmelchior:
Yeah it probably is. I'll add a section to the docs as well.

PR #838, opened 2015-02-08
Make the native libraries way smaller.

emanuelez:
This is done reactivating link-time optimizations and using strip.

It is now also easier to switch between any supported version of gcc and clang

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,4 +1,5 @@
 0.79
+ * Make the native libraries much smaller. Arm went from 1.8MB to 800KB.
  * Better error reporting when trying to create or open a Realm file fails.
  * Improved error reporting in case of missing accessors in model classes.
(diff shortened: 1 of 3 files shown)


emanuelez:
It required the latest NDK (r10d) to be installed on the Jenkins slave. It will also be needed to support ARM64.

Local final results:

CPU | Size
--- | ----
ARM | 809K
ARMv7a | 789K
MIPS | 1.6M
X86 | 1.3M


The jar file is now 1.9 MB

cmelchior:
Totally awesome, but we also need to document the frontpage README with new NDK requirements.

emanuelez:
Using gcc 4.8 the requirements did not change ☺

emanuelez:
@cmelchior @kneth @bmunkholm @kspangsege

kneth:
Nice job!

:+1:

PR #840, opened 2015-02-09
Event on MainThread

leonardoxh:
Just remove dead code, on the event bus a name convention on onEvent is used and vuola :)

### DIFF
--- a/experimental/pathExperiment/src/main/java/io/realm/path/MainActivity.java
+++ b/experimental/pathExperiment/src/main/java/io/realm/path/MainActivity.java
@@ -52,17 +52,12 @@ public class MainActivity extends ActionBarActivity {
         jobManager.addJob(new AddPersonJob(this));
     }
 
-    // Callback from EventBus on the same thread as the job which posted the event. This is not executed on the UI thread.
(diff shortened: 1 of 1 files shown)


experimental/pathExperiment/src/main/java/io/realm/path/MainActivity.java
```diff
@@ -52,17 +52,12 @@ public void addPerson() {
         jobManager.addJob(new AddPersonJob(this));
     }
 
-    // Callback from EventBus on the same thread as the job which posted the event. This is not executed on the UI thread.
-    public void onEvent(Person person) {
+    // Callback from EventBus
+    public void onEventMainThread(Person person) {
```
cmelchior: Have you tried running the example afterwards?
You are correct that this way you automatically force the event to happen on the UI thread, but RealmObjects cannot be parsed between threads that way. This is also why the toString() method is called on the original thread and the value then sent to the UI thread.

leonardoxh:
That's a good question. I really don't imagine in that way.

I think this pull request can be closed :|

cmelchior:
No worries :)

PR #844, opened 2015-02-10
Parameter checking of LinkView methods in C++

kneth:
Parameter checking is better done at the JNI layer (C++) since these methods might be used from different parts of the Realm framework. 

As a bonus, tracing has been added to `LinkView` and `Row` C++ methods.

@emanuelez @cmelchior

### DIFF
--- a/realm/src/androidTest/java/io/realm/RealmListTest.java
+++ b/realm/src/androidTest/java/io/realm/RealmListTest.java
@@ -64,6 +64,7 @@ public class RealmListTest extends AndroidTestCase {
                 }
                 fail();
             } catch (IndexOutOfBoundsException expected) {
+            } catch (RealmException ignored) {
             }
         }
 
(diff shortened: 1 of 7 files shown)


realm/src/androidTest/java/io/realm/RealmListTest.java
```diff
@@ -64,6 +64,7 @@ private void testMethodsOnEmptyList(RealmList<Dog> list) {
                 }
                 fail();
             } catch (IndexOutOfBoundsException expected) {
+            } catch (RealmException ignored) {
```
cmelchior: Why add this? Now it will incorrectly parse if a RealmException is thrown?

realm/src/main/java/io/realm/RealmList.java
```diff
@@ -114,7 +114,9 @@ public boolean add(E object) {
     public E set(int location, E object) {
         assertValidObject(object);
```
cmelchior: I think we are converging to naming all these check* instead of assert*

cmelchior:
Apart from comments :+1:

Issue #705, opened 2014-12-31
Add ARM64 bit support (UnsatisfiedLinkError on Nexus 9 ( Android 5 Lolipop) )


toantran-ea:
```java.lang.UnsatisfiedLinkError: dalvik.system.PathClassLoader[DexPathList[[zip file "/data/app/com.xxx.yyy-2/base.apk"],nativeLibraryDirectories=[/data/app/com.xxx.yyy-2/lib/arm64, /vendor/lib64, /system/lib64]]] couldn't find "libtightdb-jni.so"
            at java.lang.Runtime.loadLibrary(Runtime.java:366)
            at java.lang.System.loadLibrary(System.java:989)
            at io.realm.internal.TightDB.loadLibrary(TightDB.java:144)
            at io.realm.internal.SharedGroup.<clinit>(SharedGroup.java:32)
            at io.realm.Realm.<init>(Realm.java:151)
            at io.realm.Realm.createAndValidate(Realm.java:484)
            at io.realm.Realm.create(Realm.java:464)
            at io.realm.Realm.getInstance(Realm.java:415)
            at io.realm.Realm.getInstance(Realm.java:265)```

We tried on Nexus 9 and Nexus 4. But this issue only happened on Nexus 9.

**Comment by emanuelez** [comment]
How do you build your app? Gradle or ADT?

**Comment by cmelchior** [comment]
@toantran-ea Did you get it to work?

**Comment by toantran-ea** [comment]
@emanuelez I was using Gradle with Android Studio 
@cmelchior I'm still having problem with it :(

**Comment by emanuelez** [comment]
Can you reproduce the issue in the emulator?

**Comment by bmunkholm** [comment]
Hi @toantran-ea : We would love to get to the buttom of your issue, but obviously require you feedback to do that.

**Comment by emanuelez** [comment]
Ah! The answer was staring us in the face all this time!

```
nativeLibraryDirectories=[/data/app/com.xxx.yyy-2/lib/arm64, /vendor/lib64, /system/lib64]
```

The Nexus 9 is a 64 bits library and we only provide binaries for 32 bits architectures.
Time to add support for ARM64

falkorichter:
+1 on ARM64 binaries...

ousenko:
+1
guys, could you shed some light on when to expect ARM64 support?

cmelchior:
Unless we run into unforeseen problems, this week.

toantran-ea:
+1 great news, thanks a lot ;)

[...]

PR #845, opened 2015-02-10
Add support for ARM64

emanuelez:


emanuelez:
@kneth @cmelchior @kspangsege

emanuelez:
Fixes #705

cmelchior:
If we are sure that all pointers have been changed to support 64 bit :+1:

kneth:
We are storing native pointers as a `long`. According to http://stackoverflow.com/questions/337268/what-is-the-correct-way-to-store-a-native-pointer-inside-a-java-object this is 64 bit safe.

cmelchior:
I was more referring to the use of the ```PRId64``` syntax, but I wasn't very precise on that :)

kneth:
The `PRId64` selects the appropriate format string for `int64_t` as we cast native pointers to that type  (using the `S64` macro) when logging.

PR #847, opened 2015-02-11
Removing usages of the column type Mixed.

kneth:
In the current object store we do not utilise core's column type Mixed. In order to reduce the resulting jar file size, Mixed is removed.

808380 libtightdb-jni-arm-v7a.so
824756 libtightdb-jni-arm.so
1386408 libtightdb-jni-arm64.so
1699684 libtightdb-jni-mips.so
1344964 libtightdb-jni-x86.so

@emanuelez @bmunkholm @cmelchior


timanglade:
Hm, considering this isn’t a major improvement on the file size, should we just leave it in? It does feel like we’ll need mixed at some point for the dynamic API, so maybe this creates more work than it solves?

astigsen:
We will definitely need mixed soon. We already have users who has to do
weird and cumbersome workarounds to represent their data because we don't
have it. But I don't think we should expose it to users until we have
extended it to be complete (primarily by including support for nulls, links
and linklists).

timanglade:
Yes we agree on not exposing anything yet. It removing it had reduced our file size down to the 500kb range, maybe I’d feel better about removing it temporarily but since the impact seems really minimal, it’s probably not worth the hassle.

Unless the numbers are erroneous somehow (I assume the numbers listed in the description of this PR reflect the situation AFTER @emanuelez’s improvements + removing Mixed), let’s just close this. It was worth a shot but not worth it after all.

kneth:
@timanglade The numbers are fine. The mixed column type is probably highly integrated in core so the linker cannot reduce the size.

PR #849, opened 2015-02-13
Child objects with PK are now updated/copied correctly

cmelchior:
A bug was discovered when using copyToRealmOrUpdate on child elements with primary keys: https://groups.google.com/forum/#!topic/realm-java/V7AVHgU5Oo4 which caused child objects to always be copied.

It wasn't caught due to a bug in some unit tests *cough*.

This PR fixes both the bug and the unit tests.

@emanuelez @kneth


kneth:
:+1:

thorbenprimke:
@cmelchior - Awesome! Thanks for the quick fix.

emanuelez:
:+1:

PR #850, opened 2015-02-13
Adding light HTML markup to JavaDoc so documentation website looks better

kneth:
In general JavaDoc isn't like markdown, and some HTML formatting must be added in order to generate some better formatted HTML for the website. In particular, this includes empty line (should be `<p>`) and unordered list.

@emanuelez @cmelchior

### DIFF
--- a/realm/src/main/java/io/realm/RealmChangeListener.java
+++ b/realm/src/main/java/io/realm/RealmChangeListener.java
@@ -18,11 +18,11 @@ package io.realm;
 
 /**
  * Using RealmChangeListener, it is possible to be notified when a Realm instance has been updated.
- *
- * Realm instances on a thread without a {@link android.os.Looper} (almost all background threads)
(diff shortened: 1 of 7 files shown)


realm/src/main/java/io/realm/RealmChangeListener.java
```diff
@@ -18,11 +18,11 @@
 
 /**
  * Using RealmChangeListener, it is possible to be notified when a Realm instance has been updated.
- *
- * Realm instances on a thread without a {@link android.os.Looper} (almost all background threads)
- * don't get updated automatically, but have to call {@link Realm#refresh()} manually. This will
+ * <p>
+ * Realm instances on a thread without an {@link android.os.Looper} (almost all background threads)
+ * doesn't get updated automatically, but have to call {@link Realm#refresh()} manually. This will
```
emanuelez: don't was actually correct since instances is plural

emanuelez:
:+1:

PR #853, opened 2015-02-16
Made Thread example more resilient to monkey test events.

cmelchior:
This fixes the Thread example which tried to access resources after fragment was detached.
It also now properly kills the background thread.

@emanuelez @kneth @bmunkholm

### DIFF
--- a/examples/threadExample/src/main/java/io/realm/examples/threads/ThreadFragment.java
+++ b/examples/threadExample/src/main/java/io/realm/examples/threads/ThreadFragment.java
@@ -18,6 +18,7 @@ package io.realm.examples.threads;
 
 import android.app.Activity;
 import android.os.Bundle;
+import android.os.SystemClock;
 import android.support.v4.app.Fragment;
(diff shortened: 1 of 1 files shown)


examples/threadExample/src/main/java/io/realm/examples/threads/ThreadFragment.java
```diff
@@ -135,14 +136,14 @@ public void run() {
                     Dot dot = backgroundThreadRealm.createObject(Dot.class);
                     dot.setX(random.nextInt(100));
                     dot.setY(random.nextInt(100));
-                    dot.setColor(getResources().getColor(R.color.realm_red));
+                    dot.setColor(redColor);
                     backgroundThreadRealm.commitTransaction();
 
                     // Wait 0.5 sec. before adding the next dot.
                     try {
```
emanuelez: If we used SystemClock.sleep() we wouldn't need to catch the exception.

PR #854, opened 2015-02-16
Cm annotation javadoc

emanuelez:


emanuelez:
@cmelchior @kneth @bmunkholm

cmelchior:
We also need to update frontpage README as this changes the target for generating JavaDoc, otherwise :+1:

PR #856, opened 2015-02-16
Case sensitive queries on relationships

kneth:
Core has implemented case sensitive queries on links (relationships) long time ago, and we forgot to follow up on that: all five predicates support it (equal, not equal, contains, begins with, and ends with).

I took the liberty to simplify the string predicates on the JNI level instead of having so much duplicate code.

I have to redo `generate_jni_headers.sh` as we more and more depend on Android. The workaround is to compile using Android Studio (and probably fail), run the script, and then recompile without errors. I hope that the upcoming native plugin for Gradle will solve this problem elegantly.


realm-jni/generate-jni-headers.sh
```diff
@@ -1,20 +1,18 @@
 #!/bin/bash
 #
 # Generate header file for JNI calls
+#
+# Assumption: the .java files have been compiled
+# 1. build project using Android Studio (or gradle); unit tests will probably fail due to missing native methods
```
bmunkholm: why is this (1.) build needed?

realm-jni/src/io_realm_internal_TableQuery.cpp
```diff
@@ -657,75 +657,93 @@ JNIEXPORT void JNICALL Java_io_realm_internal_TableQuery_nativeEqual__J_3JZ(
...
+            case StringEndsWith:
+                Q(nativeQueryPtr)->and_query(tbl->column<String>(size_t(arr[arr_len-1])).ends_with(StringData(value2), isCaseSensitive));
+                break;
+            }
         }
     } CATCH_STD()
     RELEASE_ARRAY()
 }
 
+JNIEXPORT void JNICALL Java_io_realm_internal_TableQuery_nativeEqual__J_3JLjava_lang_String_2Z(
```
cmelchior: This looks like a copy/paste error

realm/src/main/java/io/realm/internal/TableQuery.java
```diff
@@ -402,70 +402,70 @@ public TableQuery notEqualTo(long columnIndex[], String value){
     }
     protected native void nativeNotEqual(long nativeQueryPtr, long columnIndex[], String value, boolean caseSensitive);
 
-    public TableQuery beginsWith(long columnIndex, String value, boolean caseSensitive){
+    public TableQuery beginsWith(long columnIndices[], String value, boolean caseSensitive) {
```
bmunkholm: I don't think we should remove the option to just provide one sort parameter in a simple fashion - then better provide both options. (goes for all methods taking columnIndices[])

bmunkholm: Sorry, I see this in only for the dynamic API. We can look at that when exposing it publicly.

cmelchior: I would tend to agree, even though this class is not public facing, it still generates a lot of ugly boilerplate elsewhere.

realm/src/androidTest/java/io/realm/internal/JNIQueryTest.java
```diff
@@ -315,19 +315,19 @@ public void testNegativeColumnIndexContains() {
         TableQuery query = table.where();
 
         // String
-        try { query.contains(-1, "hey").findAll(); fail("-1 column index"); } catch (ArrayIndexOutOfBoundsException e) {}
-        try { query.contains(-1, "hey").findAll(); fail("-10 column index"); } catch (ArrayIndexOutOfBoundsException e) {}
-        try { query.contains(-1, "hey").findAll(); fail("-100 column index"); } catch (ArrayIndexOutOfBoundsException e) {}
+        try { query.contains(new long[]{-1}, "hey").findAll(); fail("-1 column index"); } catch (ArrayIndexOutOfBoundsException e) {}
```
bmunkholm: This is testing the same thing 3 times (previous copy/paste error it seams).
But testing with -1, -10, -100 are bad test values anyway. It should be edge values instead.

realm-jni/generate-jni-headers.sh
kneth: In order to create the `.class` files which `javah` can operate on.

realm/src/main/java/io/realm/internal/TableQuery.java
kneth: If we have both methods, they should be used in `RealmQuery`. This means that we have doubled the code paths to test (and we have many predicates in `RealmQuery`). The proposed change is simply making the API for `RealmQuery` more consistent as `contains`, `beginsWith`, and `endsWith` support link queries without any exception.

cmelchior: Fair enough. Simplicity also has clear advantages. Then we can always revisit at a later point if needed.

realm/src/androidTest/java/io/realm/internal/JNIQueryTest.java
kneth: You are absolutely right - the unit tests for the JNI functions could be much better and they haven't been touch for a very long time.

realm-jni/src/io_realm_internal_TableQuery.cpp
```diff
@@ -657,75 +657,93 @@ JNIEXPORT void JNICALL Java_io_realm_internal_TableQuery_nativeEqual__J_3JZ(
...
+    JNIEnv *env, jobject, jlong nativeQueryPtr, jlongArray columnIndexes, jstring value, jboolean caseSensitive)
```
bmunkholm: Any reason to keep these multiple JNI methods now? Why not just use the TableQuery_StringPredicate() from Java?

realm/src/main/java/io/realm/internal/TableQuery.java
bmunkholm: If it's ever going to be public - and it likely is, then the consistency should be that there is a shortcut method for a single sort as that is mostly used. And that goes for all sort methods. The consistency would then be between the high object interface and the dynamic interface. But I'm ok with leaving it for now.

realm-jni/src/io_realm_internal_TableQuery.cpp
kneth: I was thinking the same when I wrote it but decided to keep them in order to have the same structure as the other types. Could we refactor the code (for all types) in another branch later?

Issue #859, opened 2015-02-16
copyToRealmOrUpdate() results in stack overflow for models with cyclic relationships


decnorton:
Similar to #484 and #589. If a model has a cyclic relation with another model it'll continously try to `copy()` or `update()`, resulting in stack overflow.

cmelchior:
This is also an issue with CopyToRealm

cmelchior:
A fix has been merged and will be part of 0.79.1

cmelchior:
0.79.1 has been released and should be in bintray now.

decnorton:
Thanks!

PR #867, opened 2015-02-18
copyToRealm now supports cyclic data structures

cmelchior:
This fixes issue: https://github.com/realm/realm-java/issues/859

In order to do so a <object, RealmObject> map is now sent down the recursive method chain, if any object inside that map is found, the cached object is used instead of trying to parse it again.

This means that we now keep references to possible nested RealmObjects for the entire duration of the method call, where before they could be garbage collected. This is unavoidable as primary keys are not required for this method.


realm-annotations-processor/src/main/java/io/realm/processor/RealmProxyClassGenerator.java
```diff
@@ -528,20 +529,26 @@ private void emitCopyMethod(JavaWriter writer) throws IOException {
...
                 writer
                     .emitEmptyLine()
-                    .emitStatement("%s %s = newObject.%s()", Utils.getFieldTypeSimpleName(field), fieldName, getters.get(fieldName))
+                    .emitStatement("%s %s = newObject.%s()", fieldType, fieldName, getters.get(fieldName))
                     .beginControlFlow("if (%s != null)", fieldName)
-                        .emitStatement("realmObject.%s(%s.copyOrUpdate(realm, %s, update))",
-                                setters.get(fieldName),
-                                Utils.getProxyClassSimpleName(field),
-                                fieldName)
+                        .beginControlFlow("if (cache.containsKey(%s))", fieldName)
```
emanuelez: cache.containsKey() + cache.get() is slow.
Use cache.get() and check for null instead

realm/src/androidTest/java/io/realm/RealmTest.java
```diff
@@ -1000,6 +1002,38 @@ public void testCopyToRealmObject() {
...
+        CyclicType anotherCyclicType = new CyclicType();
+        anotherCyclicType.setName("Two");
+        oneCyclicType.setObject(anotherCyclicType);
+        anotherCyclicType.setObject(oneCyclicType);
+
+        testRealm.beginTransaction();
+        CyclicType realmObject = testRealm.copyToRealm(oneCyclicType);
+        testRealm.commitTransaction();
+
+        assertEquals("One", realmObject.getName());
```
kneth: Isn't it worth checking that `realmObject.getObject().getName()` is `"Two"`?

realm-annotations-processor/src/main/java/io/realm/processor/RealmProxyClassGenerator.java
```diff
@@ -478,11 +478,11 @@ private void emitCopyOrUpdateMethod(JavaWriter writer) throws IOException {
                 className, // Return type
                 "copyOrUpdate", // Method name
                 EnumSet.of(Modifier.PUBLIC, Modifier.STATIC), // Modifiers
-                "Realm", "realm", className, "object", "boolean", "update" // Argument type & argument name
```
emanuelez: Shouldn't it be `Map<? extends RealmObject,? extends RealmObject>` ?

realm-annotations-processor/src/main/java/io/realm/processor/RealmProxyClassGenerator.java
```diff
@@ -478,11 +478,11 @@ private void emitCopyOrUpdateMethod(JavaWriter writer) throws IOException {
...
+                "Realm", "realm", className, "object", "boolean", "update", "Map<RealmObject,RealmObject>", "cache" // Argument type & argument name
```
emanuelez: Shouldn't it be `Map<? extends RealmObject, ? extends RealmObject>` ?
Also, since now we are at a point where it would be nice to distinguish proxy objects from standalone objects, wouldn't it maybe make sense to have proxy objects implement an empty interface, for generics purposes?

cmelchior: No, that doesn't work.

I have been thinking about implementing a interface as well. All our static methods could move into it as well, which could reduce all our reflection calls in Realm atm. But that should probably happen as part of a another PR?

cmelchior: Also, gief java 8. It allows for static methods as part of the interface

emanuelez: Why doesn't it work?

cmelchior: The type system complains once you start adding elements to the map

emanuelez: That shouldn't happens, which makes me think something might be wrong. Hmmm...

realm/src/androidTest/java/io/realm/RealmTest.java
cmelchior: Updated

emanuelez:
:+1:

Issue #864, opened 2015-02-17
copyToRealmOrUpdate() Throws PrimaryKey Constraint Exception When PrimaryKey Is Field In Non-PrimaryKey Class


thorbenprimke:
copyToRealmOrUpdate() doesn't seem to pass the update flag through for objects that don't have a primaryKey and thus any fields within that object is only attempted to be copied. 

Example: https://github.com/thorbenprimke/realm-java/tree/nestedObjectCausesPrimaryKeyConflict
Example code: https://github.com/thorbenprimke/realm-java/blob/9f2992ef20e9b3241bac9a6e45f7fe034167c70f/examples/introExample/src/main/java/io/realm/examples/intro/IntroExampleActivity.java#L106

thorbenprimke:
Forgot the stacktrace: https://gist.github.com/thorbenprimke/bc59000b0a336f7b1351

cmelchior:
A fix has been merged and will be released as part of 0.79.1

cmelchior:
0.79.1 has been released and should be in bintray now.

PR #869, opened 2015-02-18
Fix crash when using copyToRealmOrUpdate with complex object graph

cmelchior:
This PR fixes: https://github.com/realm/realm-java/issues/864

Before, the update flag was cleared if an class didn't contain a primary key. This could cause duplicate entry errors if that object contained references to objects with primary keys as they where only copied.

@emanuelez @kneth @bmunkholm


FaisalAbid:
Hey @cmelchior - when will this be merged and released?

kneth:
:+1:

emanuelez:
:+1:

PR #876, opened 2015-02-19
Unit test realms from Cocoa

cmelchior:
This PR adds unit tests for all basic datatypes from a Realm created by the Cocoa binding.


cmelchior:
@emanuelez @kneth @bmunkholm

kneth:
Will you add an encrypted realm file too?

cmelchior:
It has been added.

cmelchior:
Bump @emanuelez @kneth @bmunkholm

realm/src/androidTest/java/io/realm/TestHelper.java
```diff
@@ -41,6 +47,27 @@ public static String streamToString(InputStream in) throws IOException {
         return sb.toString();
     }
 
+    // Copies a Realm file from assets to app files dir
+    public static void copyRealmFromAssets(Context context, String realmPath, String newName) throws IOException {
+        AssetManager assetManager = context.getAssets();
+        InputStream is = assetManager.open(realmPath);
```
emanuelez: Properly handle closeables:
```java
InputStream in = null;
try {
  in = openInputStream();
  OutputStream out = null;
  try {
    out = openOutputStream();
    // do something with in and out
  } finally {
    if (out != null) {
      out.close();
    }
  }
} finally {
  if (in != null) {
    in.close();
  }
}
```

realm/src/main/java/io/realm/Realm.java
```diff
@@ -167,6 +169,7 @@
     private final Map<Class<?>, Constructor> generatedConstructors = new HashMap<Class<?>, Constructor>();
     private final List<RealmChangeListener> changeListeners = new ArrayList<RealmChangeListener>();
     private final Map<Class<?>, Table> tables = new HashMap<Class<?>, Table>();
+    private static final Set<Class<? extends RealmObject>> customSchema = new HashSet<Class<? extends RealmObject>>();
```
emanuelez: Could this potentially be used by several threads at the same time?

realm/src/main/java/io/realm/Realm.java
```diff
@@ -1778,6 +1789,25 @@ public String getPath() {
...
+    /**
+     * Override the standard behavior of all classes extended RealmObject being part of the schema.
+     * Use this method to define the schema as only the classes given here.
+     *
+     * This class must be called before calling {@link #getInstance(android.content.Context)}
+     *
+     * If {@code null} is given as parameter, the Schema is reset to use all known classes.
+     *
+     */
+    static void setSchema(Class<? extends RealmObject>... schemaClass) {
```
emanuelez: Why varargs?

emanuelez:
Why are the realm files assets and not normal resources?

kneth:
Will you also test 32 bit vs. 64 bit architectures - or it is out-of-scope?

realm/src/main/java/io/realm/Realm.java
cmelchior: This method is package private, so only used for testing. The assumption was that the most important thing is to quickly add a small number of classes with less code.
```
Realm.setSchema(Foo.class, Bar.class); // reads better than
Realm.setSchema(Arrays.asList(Foo.class, Bar.class)); 
```

cmelchior:
@emanuelez  They are there because the JSON files was there also. The difference between ```assets``` and ```res/raw``` is marginal at best and probably boils down to personal preference. I do like that ```assets``` can have subfolders which makes it easier to organize test resources.  ```res/raw``` doesn't allow that, so there you only have file prefixes to work with.

cmelchior:
@kneth These tests are mostly fallout from the Primary key feature, and there is a LOT more we should test to ensure compatibility with Cocoa Realms. So I would prefer to keep this PR small, then we can expand the tests later.

kneth:
@cmelchior I understand what you're saying (title of the PR is a bit broader). Let's merge it, get it automated, and extend it ;-)

PR #887, opened 2015-02-23
Only do the version checker on the first round of the processor.

emanuelez:
This is done by making a singleton out of RealmVersionChecker.
It’s not doable in RealmProcessor since one new instance of the processor is created per round.

### DIFF
--- a/realm-annotations-processor/src/main/java/io/realm/processor/RealmProcessor.java
+++ b/realm-annotations-processor/src/main/java/io/realm/processor/RealmProcessor.java
@@ -17,35 +17,22 @@
 package io.realm.processor;
 
 
-import java.io.IOException;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
+import io.realm.annotations.Ignore;
+import io.realm.annotations.Index;
+import io.realm.annotations.PrimaryKey;
+import io.realm.annotations.RealmClass;
 
 import javax.annotation.processing.AbstractProcessor;
 import javax.annotation.processing.RoundEnvironment;
 import javax.annotation.processing.SupportedAnnotationTypes;
 import javax.lang.model.SourceVersion;
-import javax.lang.model.element.Element;
-import javax.lang.model.element.ElementKind;
-import javax.lang.model.element.ExecutableElement;
-import javax.lang.model.element.Modifier;
-import javax.lang.model.element.PackageElement;
-import javax.lang.model.element.TypeElement;
-import javax.lang.model.element.VariableElement;
+import javax.lang.model.element.*;
 import javax.lang.model.type.TypeKind;
 import javax.lang.model.type.TypeMirror;
 import javax.lang.model.util.Types;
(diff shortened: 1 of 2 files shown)


emanuelez:
@cmelchior @kneth @bmunkholm

realm-annotations-processor/src/main/java/io/realm/processor/RealmProcessor.java
```diff
@@ -17,35 +17,22 @@
...
 import javax.annotation.processing.SupportedAnnotationTypes;
 import javax.lang.model.SourceVersion;
-import javax.lang.model.element.Element;
-import javax.lang.model.element.ElementKind;
-import javax.lang.model.element.ExecutableElement;
-import javax.lang.model.element.Modifier;
-import javax.lang.model.element.PackageElement;
-import javax.lang.model.element.TypeElement;
-import javax.lang.model.element.VariableElement;
+import javax.lang.model.element.*;
```
cmelchior: I would rather prefer explicit import.

realm-annotations-processor/src/main/java/io/realm/processor/RealmProcessor.java
```diff
@@ -17,35 +17,22 @@
...
 import javax.lang.model.type.TypeMirror;
 import javax.lang.model.util.Types;
 import javax.tools.Diagnostic;
-
-import io.realm.annotations.Ignore;
-import io.realm.annotations.Index;
-import io.realm.annotations.PrimaryKey;
-import io.realm.annotations.RealmClass;
+import java.io.IOException;
+import java.util.*;
```
cmelchior: Same

realm-annotations-processor/src/main/java/io/realm/processor/RealmVersionChecker.java
```diff
@@ -33,10 +35,17 @@
 
     private ProcessingEnvironment processingEnvironment;
 
-    public RealmVersionChecker(ProcessingEnvironment processingEnvironment) {
+    private RealmVersionChecker(ProcessingEnvironment processingEnvironment) {
         this.processingEnvironment = processingEnvironment;
     }
 
+    public static RealmVersionChecker getInstance(ProcessingEnvironment processingEnvironment) {
```
cmelchior: Using a singleton pattern this is usually synchronized, but since it is only running as part of the annotation processor I guess it is OK.

realm-annotations-processor/src/main/java/io/realm/processor/RealmProcessor.java
emanuelez: Intellij IDEA does this automatically and is usually very good in determining when it's worth to import *. Also, being this code only run at compile time, it makes sense to optimize for speed and not for memory.

realm-annotations-processor/src/main/java/io/realm/processor/RealmVersionChecker.java
emanuelez: Yes, it is guaranteed that only one thread is executing a processing round.

cmelchior:
Good point with memory vs. speed
:+1:

kneth:
:+1:

Issue #881, opened 2015-02-21
Compilation error in the code generated by the annotation processor


jcapra:
I have the following simple model class...
```
@RealmClass
public class Category extends RealmObject {
   @PrimaryKey
    private String id;
    private I18Ntext name;
    private float order;
    private String imageUrl;
    private String iconUrl;
    private RealmList<Template> questionTemplates;
}
```

And this is the portion of the RealmProxy class that is not compiling after the annotation processing...
```
public static void populateUsingJsonStream(Category obj, JsonReader reader) throws IOException {
        boolean standalone = obj.realm == null;
        reader.beginObject();
        while (reader.hasNext()) {
            String name = reader.nextName();
            if (name.equals("iconUrl") && reader.peek() != JsonToken.NULL) {
                obj.setIconUrl((String) reader.nextString());
            } else if (name.equals("id")  && reader.peek() != JsonToken.NULL) {
                obj.setId((String) reader.nextString());
            } else if (name.equals("imageUrl")  && reader.peek() != JsonToken.NULL) {
                obj.setImageUrl((String) reader.nextString());
            } else if (name.equals("name")  && reader.peek() != JsonToken.NULL) {
                com.riotmill.justus.services.model.I18Ntext name = standalone ? new com.riotmill.justus.services.model.I18Ntext() : obj.realm.createObject(com.riotmill.justus.services.model.I18Ntext.class);
                I18NtextRealmProxy.populateUsingJsonStream(name, reader);
                obj.setName(name);
            } else if (name.equals("order")  && reader.peek() != JsonToken.NULL) {
                obj.setOrder((float) reader.nextDouble());
            } else if (name.equals("questionTemplates")  && reader.peek() != JsonToken.NULL) {
                reader.beginArray();
                if (standalone) {
                    obj.setQuestionTemplates(new RealmList<com.riotmill.justus.services.model.Template>());
                }
                while (reader.hasNext()) {
                    com.riotmill.justus.services.model.Template item = standalone ? new com.riotmill.justus.services.model.Template() : obj.realm.createObject(com.riotmill.justus.services.model.Template.class);
                    TemplateRealmProxy.populateUsingJsonStream(item, reader);
                    obj.getQuestionTemplates().add(item);
                }
                reader.endArray();
            } else {
                reader.skipValue();
            }
        }
        reader.endObject();
    }
```

[...]

PR #888, opened 2015-02-23
Allow all names in model classes

cmelchior:
This PR fixes this issue: https://github.com/realm/realm-java/issues/881

It makes sure that object references are properly suffixed so they don't conflict with internal variabel names (This only happened in one JSON import function).

@emanuelez @kneth @bmunkholm

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,5 +1,6 @@
 0.80
 * Queries on relationships can be case sensitive.
+* Fixed bug crashing annotation processor when using "name" in model classes for RealmObject references
 
 0.79.1
  * copyToRealm() no longer crashes on cyclic data structures.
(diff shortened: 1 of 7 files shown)


kneth:
:+1: for solving this case. I'm just wondering if the we other similar clashes.

cmelchior:
I don't think so. I looked over the rest of the potential methods copy/update/populateUsingJsonObject/populateUsingJsonStream. They should all be safe.

emanuelez:
Is the the `Obj` suffix unique enough for us to avoid getting in trouble again in the future?

cmelchior:
For that case, yes, since the only problematic variabel in that scope is called "name".
Only RealmObject and RealmList references makes internal variabels which can conflict. Normal datatype references use getter/setters directly.

emanuelez:
:+1:

Issue #875, opened 2015-02-19
Deleting from RealmList throws exception


cmelchior:
Hi @cmelchior 
I am probably experiencing the same issue. I am using ver. 0.79.0.
```Stacktrace
FATAL EXCEPTION: main
java.lang.ArrayIndexOutOfBoundsException: rowIndex > available rows.
at io.realm.internal.LinkView.nativeRemove(Native Method)
at io.realm.internal.LinkView.remove(LinkView.java:67)
at io.realm.RealmList.remove(RealmList.java:189)
at com.sampleapp.MergedCardsListClick$1.onClick(MergedCardsListClick.java:64)
...
at dalvik.system.NativeStart.main(Native Method)
```
This issue happens when I try to delete the last element in `RealmList`. Here is my code:
```Java
for (int i = 0; i < parent.getChildren().size(); i++) {
	if (parent.getChildren().get(i).get_id() == object.get_id()) {
		mRealm.beginTransaction();
		parent.getChildren().remove(i); // Line that throws ArrayIndexOutOfBoundsException
		mRealm.commitTransaction();
		break;
	}
}
```
The strange thing is that when I try to delete any element but not last, it does not crash. But when I try to delete the last element (again, **any** element), it crashes.
So I changed the code a bit to make it work: (maybe somebody will find it helpful)
```Java
for (int i = 0; i < size; i++) {
	if (parent.getChildren().get(i).get_id() == sms.get_id()) {
		mRealm.beginTransaction();
		if (size != 1) {
			parent.getChildren().remove(i); // does not crash
		} else {
			parent.getChildren().clear(); // does not crash
		}
		mRealm.commitTransaction();
		break;
	}
}
```
My model:
```Java
package com.sampleapp;

import io.realm.RealmList;
import io.realm.RealmObject;
import io.realm.annotations.Index;
import io.realm.annotations.RealmClass;

@RealmClass
public class Parent extends RealmObject {
	private long _id;

	@Index
	private String originalText;
	private Parent parent;
	private RealmList<Parent> children;

	public long get_id() {
		return _id;
	}

	public void set_id(long _id) {
		this._id = _id;
	}

	public String getOriginalText() {
		return originalText;
	}

	public void setOriginalText(String originalText) {
		this.originalText = originalText;
	}

	public Parent getParent() {
		return parent;
	}

	public void setParent(Parent parent) {
		this.parent = parent;
	}

	public RealmList<Parent> getChildren() {
		return children;
	}

[...]

PR #889, opened 2015-02-23
Bugfix for removing last element in RealmList

cmelchior:
Fix for bug reported here: https://github.com/realm/realm-java/issues/875
Also made error message output a bit more relevant info.

@kneth @emanuelez @bmunkholm


kneth:
:+1:

PR #890, opened 2015-02-23
Fix support for Json NULL values

cmelchior:
This PR fixes support for fields with explicit NULL with wasn't working correctly due to a faulty unit test.

### DIFF
--- a/realm-annotations-processor/src/main/java/io/realm/processor/RealmJsonTypeHelper.java
+++ b/realm-annotations-processor/src/main/java/io/realm/processor/RealmJsonTypeHelper.java
@@ -28,10 +28,8 @@ import java.util.Map;
 public class RealmJsonTypeHelper {
 
     private static final Map<String, JsonToRealmTypeConverter> JAVA_TO_JSON_TYPES;
-
     static {
         JAVA_TO_JSON_TYPES = new HashMap<String, JsonToRealmTypeConverter>();
-
         JAVA_TO_JSON_TYPES.put("byte", new SimpleTypeConverter("byte", "Int"));
         JAVA_TO_JSON_TYPES.put("short", new SimpleTypeConverter("short", "Int"));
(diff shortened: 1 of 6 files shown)


cmelchior:
Ready for review
@emanuelez @kneth @bmunkholm

realm-annotations-processor/src/main/java/io/realm/processor/RealmJsonTypeHelper.java
```diff
@@ -51,12 +49,15 @@
...
                 writer
-                    .emitStatement("long timestamp = json.optLong(\"%s\", -1)", fieldName)
-                    .beginControlFlow("if (timestamp > -1)")
-                        .emitStatement("obj.%s(new Date(timestamp))", setter)
+                    .beginControlFlow("if (!json.isNull(\"%s\"))", fieldName)
+                        .emitStatement("Object timestamp = json.get(\"%s\")", fieldName)
+                        .beginControlFlow("if (timestamp instanceof String)")
+                           .emitStatement("obj.%s(JsonUtils.stringToDate((String) timestamp))", setter)
+                        .nextControlFlow("else")
+                            .emitStatement("obj.%s(new Date(json.getLong(\"%s\")))", setter, fieldName)
```
kneth: Is it safe to assume that it is a `long`?

cmelchior: Hmm, good point. If somebody messed up their JSON it could be anything.

cmelchior: Reconsidering this. We don't check for the right type anywhere else in the JSON code. If people use the wrong type they get an exception, so this is actually intended behaviour.

emanuelez:
:+1:

PR #903, opened 2015-02-24
Properly check for key equality while using a cached Realm instance

### LINKED ISSUE #901: Wrong Key Exception while multiple access to encrypted Realm file
Hi guys!
I happened to stumble across bizarre situation while accessing encrypted database multiple times using different instances of Realm - the 'Wrong key used to decrypt Realm.' exception was thrown. It was weird because I was sure that i was passing matching byte array. (Of course there was no problem opening not encrypted database multiple times) . It occurs that exception is thrown when different instances of byte arrays are passed to Realm.getInstance(...), due to checking equality by checking references:
 
```
if (realm.key != key) {
   throw new IllegalStateException(DIFFERENT_KEY_MESSAGE);
}
```

My question is: is this behaviour planned? It does not seem much issue, but for me, it seems quite misleading to pass matching byte array, while getting 'Wrong key' exception or modify key byte array on the same reference, but still be able to access Realm.

The code to reproduce the situation:
```
@Override
        Realm.deleteRealmFile(this, "test");
        byte[] key = new byte[64];
        byte[] key2 = new byte[64];
        Realm realm = Realm.getInstance(this, "test", key);
        key[0] = 0xF; <-the most misleading fragment
        realm = Realm.getInstance(this, "test", key); <-opens
        realm = Realm.getInstance(this, "test", key2); <-Exception
    }
```

emanuelez:
Fixes #901

### DIFF
--- a/realm/src/androidTest/java/io/realm/RealmTest.java
+++ b/realm/src/androidTest/java/io/realm/RealmTest.java
@@ -1316,6 +1316,39 @@ public class RealmTest extends AndroidTestCase {
         outStream.close();
     }
 
+    public void testOpeningOfEncryptedRealmWithDifferentKeyInstances() throws Exception {
+        byte[] key1 = new byte[64];
+        byte[] key2 = new byte[64];
+        new Random(42).nextBytes(key1);
+        new Random(42).nextBytes(key2);
+
+        // Make sure the key is the same, but in two different instances
+        assertArrayEquals(key1, key2);
+        assertTrue(key1 != key2);
+
+        final String ENCRYPTED_REALM = "differentKeys.realm";
+
+        Realm realm1 = null;
+        Realm realm2 = null;
+        try {
+            realm1 = Realm.getInstance(getContext(), ENCRYPTED_REALM, key1);
+            try {
+                realm2 = Realm.getInstance(getContext(), ENCRYPTED_REALM, key2);
+            } catch (Exception e) {
+                fail();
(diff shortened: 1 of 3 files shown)


emanuelez:
Changelog: 
 * Fixed problem occurring when opening an encrypted Realm with two different instances of the same key.

emanuelez:
@cmelchior @kneth @bmunkholm

realm/src/androidTest/java/io/realm/RealmTest.java
```diff
@@ -1316,6 +1316,39 @@ private void fileCopy(File src, File dst) throws IOException {
...
+                if (realm2 != null) {
+                    realm2.close();
+                }
+            }
+        } finally {
+            if (realm1 != null) {
+                realm1.close();
+            }
+        }
+
```
cmelchior: Remove empty line

realm/src/androidTest/java/io/realm/RealmTest.java
```diff
@@ -1316,6 +1316,39 @@ private void fileCopy(File src, File dst) throws IOException {
         outStream.close();
     }
 
+    public void testOpeningOfEncryptedRealmWithDifferentKeyInstances() throws Exception {
```
cmelchior: There is no checked exceptions being thrown?

cmelchior:
Apart from comments :+1:

PR #880, opened 2015-02-20
Migrations API v2

cmelchior:
Work-in-progress. 

Design thoughts are in migration_design.md
API examples are in MigrationAPIExamples.java

This is just a very early draft, and eg. automatic migrations has not been considered in detail as it depends on Cocoa behaviour, but comments are welcome on the general direction or some use cases / features I completely forgot about.

**Pre-migrations (seperate PR's):**
- [x] 1) Introduce ClassMetaData class to processor to encapsulate fields/getters/setters etc. #918
- [x] 2) Introduce ColumnIndicies class, making it easier to map between name, internalname and index #920 
- [ ] 3) Add support for @RealmName annotation so names in classes/fields can be overriden.
- [ ] 4) Reintroduce RealmProxyMediator from ProGuard PR as groundwork for Modules (https://github.com/realm/realm-java/pull/938)

**Constructors**
- [ ] Introduce builder pattern for constructors
- [ ] Add default/namedRealm constructors 
- [ ] Deprecate existing constructors 

**RealmModules**
- [ ] Expand RealmConfiguration.Builder with support for RealmModules
- [ ] Expand annotation processor with support for modules (https://github.com/realm/realm-java/pull/938)

**Outstanding:**
- [ ] Writeup agreed upon design direction (stepwise)
- [ ] Implement DynamicRealmObject class
- [ ] Implement RealmObjectSpec class
- [ ] Implement RealmSpec/Migration class
- [ ] Investigate ways to diff a list of commands so only relevant code is executed

Other PR's that extend/integrate it into existing API

PR #920, opened 2015-03-02
Encapsulate column indicies

cmelchior:
Refactor for #880. 

This PR moves the column indicies map from ```Realm.java``` into its own class, making it easier to refactor when custom names for Migrations are introduced. Also column indices are now calculated as part of validateTable and cached, saving a hashmap lookup for each getter/setter.

@bmunkholm @emanuelez @kneth


realm/src/main/java/io/realm/internal/ColumnIndicies.java
```diff
@@ -0,0 +1,68 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+
+import io.realm.RealmObject;
+
+/**
+ * Utility class used to cache the mapping between object field names and their column indices.
+ */
+public class ColumnIndicies {
```
bmunkholm: Couldn't this be done much faster and simpler?
Instead of having to do two map lookup's for each getter/setter, couldn't we just make a simple long array with the indices in each class and make an initializer which does the lookups once?

cmelchior: I am not sure I understand that? 
This class is used by Realm.java, not the annotation processor, so it is not used to find getters and setters, but the column indexes for the fields of the RealmObject? These indexes are not known by the annotation processor so must be calculated at runtime. Also it is a map used by all RealmClasses not just a single one. So while it is true that you do two lookups, first to find the proper class data, then to find the index, it is two different lookups?

cmelchior: Yes, you are right. The indexes are now calculated and cached as part of validateTable, saving a hashmap lookup for each getter/setter.

kneth: Why not use simple constants? The cached indices calculated by `validateTable` will not change.

cmelchior: Because the order is based on the order of the fields in the source code. In order to create provide compile time constants you would also require migrations if you changed the order of the fields, and you would have to ensure the same order across platforms. That would be a horrible user experience.

So you are right that the order is fixed once we have run validateTable, but it still has to be calculated at runtime which means we cannot use constants.

emanuelez:
I think Brian is right. We can define and use the static fields (not final) in the proxy class and instantiate them at runtime only once (yes, with reflection, but only in the very first Realm.getInstance())

cmelchior:
I don't think my brain was working right when I wrote the first comment. You are of course right we can save the indexes directly like so:

public static int INDEX_COLUMNSTRING;
public static int INDEX_COLUMNINT;

validateTable() {
 ...
 // Fill constants
 INDEX_COLUMNSTRING = table.getColumnIndex("columnString")
 INDEX_COLUMNINT = table.getColumnIndex("columnInt")
}

row.getString(INDEX_COLUMNSTRING);
```

We still need the HashMap for query lookups though, but making the getter/setter lookups as fast as possible are important.

**Comment by cmelchior** [comment]
I have added constants to the ProxyClasses that are used by getters/setters, while keeping the columnIndices map for use by RealmQuery.

Speedup is about 20% compared to current master when using getters/setters.

**Comment by emanuelez** [comment]
Who's populating the new static fields?

**Comment by cmelchior** [comment]
validateTable() are. Any use of the accessors before that would be violating the contract of the class anyway.

**Comment by cmelchior** [comment]
Bump for review @emanuelez @kneth @bmunkholm

File: realm-annotations-processor/src/main/java/io/realm/processor/RealmProxyClassGenerator.java [hunk 2]
```diff
@@ -144,6 +144,7 @@ public void generate() throws IOException, UnsupportedOperationException {
         imports.add("android.util.JsonToken");
         imports.add("io.realm.RealmObject");
         imports.add("io.realm.exceptions.RealmException");
+        imports.add("io.realm.exceptions.RealmMigrationNeededException");
```
kneth: Indentation is missing.

kneth:
:+1:

cmelchior:
Bump @emanuelez

realm/src/main/java/io/realm/internal/ColumnIndicies.java
emanuelez: It's indices, not indicies

emanuelez: Do we still need this class at all?

cmelchior: It is groundwork for being able to override names on objects/fields using annotations. So for this PR it has little benefit, but it will be used when implementing that.

cmelchior:
@emanuelez We talked about perhaps populating the ColumnIndices using annotation processor. However that will not be possible as the cache will also have to handle classes from library projects, so the cache will have to be filled at runtime.

bmunkholm:
Looks much better now :-) One constant better than double map :-)

PR #922, opened 2015-03-04
Use phantom references to manage the closing of row accessors

### LINKED ISSUE #905: Remove finalizers and use PhantomReference

emanuelez:
Fixes #905

### DIFF
--- a/realm-jni/build.gradle
+++ b/realm-jni/build.gradle
@@ -1,4 +1,4 @@
-ext.coreVersion = '0.89.2'
+ext.coreVersion = '0.89.5'
 ext.clang = false // gcc is default for the NDK. It also produces smaller binaries
 
 def commonCflags = [ '-std=c++11', '-ffunction-sections', '-fdata-sections', '-flto' ]
--- a/realm/src/main/java/io/realm/Realm.java
+++ b/realm/src/main/java/io/realm/Realm.java
@@ -46,6 +46,8 @@ import java.util.Map;
 import java.util.Scanner;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
 import java.util.concurrent.atomic.AtomicInteger;
 
 import io.realm.exceptions.RealmException;
@@ -53,6 +55,7 @@ import io.realm.exceptions.RealmIOException;
 import io.realm.exceptions.RealmMigrationNeededException;
 import io.realm.internal.ColumnIndices;
 import io.realm.internal.ColumnType;
+import io.realm.internal.FinalizerRunnable;
 import io.realm.internal.ImplicitTransaction;
 import io.realm.internal.RealmObjectProxy;
 import io.realm.internal.RealmProxyMediator;
@@ -123,6 +126,12 @@ import io.realm.internal.migration.SetVersionNumberMigration;
 public final class Realm implements Closeable {
     public static final String DEFAULT_REALM_NAME = "default.realm";
 
+    // This single thread executor ensures that only one finalizer thread ever exists
+    private static final ExecutorService executorService = Executors.newSingleThreadExecutor();
+
+    // This does not need to be thread safe since it's only used in a synchronized method
+    private static volatile boolean isFinalizerStarted = false;
+
     protected static final ThreadLocal<Map<String, Realm>> realmsCache = new ThreadLocal<Map<String, Realm>>() {
         @SuppressLint("UseSparseArrays")
         @Override
@@ -144,6 +153,10 @@ public final class Realm implements Closeable {
     protected static final Map<Handler, String> handlers = new ConcurrentHashMap<Handler, String>();
 
     private static RealmConfiguration defaultConfiguration;
+
+    // Caches Class objects (both model classes and proxy classes) to Realm Tables
+    private final Map<Class<? extends RealmObject>, Table> classToTable =
+            new HashMap<Class<? extends RealmObject>, Table>();
 
     private static final String INCORRECT_THREAD_MESSAGE = "Realm access from incorrect thread. Realm objects can only be accessed on the thread they were created.";
     private static final String CLOSED_REALM_MESSAGE = "This Realm instance has already been closed, making it unusable.";
@@ -288,8 +301,13 @@ public final class Realm implements Closeable {
 
     // Public because of migrations
     public Table getTable(Class<? extends RealmObject> clazz) {
-        clazz = Util.getOriginalModelClass(clazz);
-        return transaction.getTable(proxyMediator.getTableName(clazz));
+        Table table = classToTable.get(clazz);
+        if (table == null) {
+            clazz = Util.getOriginalModelClass(clazz);
+            table = transaction.getTable(proxyMediator.getTableName(clazz));
+            classToTable.put(clazz, table);
+        }
+        return table;
     }
 
     /**
@@ -535,8 +553,16 @@ public final class Realm implements Closeable {
     }
 
     private static synchronized Realm createAndValidate(RealmConfiguration config, boolean validateSchema, boolean autoRefresh) {
+        // Start the finalizer thread if needed
+        if (!isFinalizerStarted) {
+            executorService.submit(new FinalizerRunnable());
+            isFinalizerStarted = true;
+        }
+
         byte[] key = config.getEncryptionKey();
         String canonicalPath = config.getPath();
+
+        // Check thread local cache for existing Realm
         Map<String, Integer> localRefCount = referenceCount.get();
         Integer references = localRefCount.get(canonicalPath);
         if (references == null) {
--- a/realm/src/main/java/io/realm/internal/Context.java
+++ b/realm/src/main/java/io/realm/internal/Context.java
@@ -30,28 +30,25 @@ class Context {
     private List<Long> abandonedTables = new ArrayList<Long>();
     private List<Long> abandonedTableViews = new ArrayList<Long>();
     private List<Long> abandonedQueries = new ArrayList<Long>();
-    private List<Long> abandonedRows = new ArrayList<Long>();
 
     private boolean isFinalized = false;
 
     public void executeDelayedDisposal() {
         synchronized (this) {
-            for (long nativePointer: abandonedTables) {
+            for (int i = 0; i < abandonedTables.size(); i++) {
+                long nativePointer = abandonedTables.get(i);
                 Table.nativeClose(nativePointer);
             }
             abandonedTables.clear();
 
-            for (long nativePointer: abandonedRows) {
-                Row.nativeClose(nativePointer);
-            }
-            abandonedRows.clear();
-
-            for (long nativePointer: abandonedTableViews) {
+            for (int i = 0; i < abandonedTableViews.size(); i++) {
+                long nativePointer = abandonedTableViews.get(i);
                 TableView.nativeClose(nativePointer);
             }
             abandonedTableViews.clear();
 
-            for (long nativePointer: abandonedQueries) {
+            for (int i = 0; i < abandonedQueries.size(); i++) {
+                long nativePointer = abandonedQueries.get(i);
                 TableQuery.nativeClose(nativePointer);
             }
             abandonedQueries.clear();
@@ -64,15 +61,6 @@ class Context {
         }
         else {
             abandonedTables.add(nativePointer);
-        }
-    }
-
-    public void asyncDisposeRow(long nativePointer) {
-        if (isFinalized) {
-            Row.nativeClose(nativePointer);
-        }
-        else {
-            abandonedRows.add(nativePointer);
         }
     }
 
--- a/realm/src/main/java/io/realm/internal/FinalizerRunnable.java
+++ b/realm/src/main/java/io/realm/internal/FinalizerRunnable.java
@@ -1,0 +1,53 @@
+/*
+ * Copyright 2015 Realm Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.realm.internal;
+
+import java.lang.ref.Reference;
+import java.lang.ref.ReferenceQueue;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+
+/**
+ * This runnable performs the garbage collection of the references stored in the reference queue
+ */
+public class FinalizerRunnable implements Runnable {
+
+    // Store the row references. Without this, objects would be garbage collected immediately so don't remove this! ;)
+    // A ConcurrentHashMap is used in lack of a ConcurrentHashSet in the Java API.
+    // ConcurrentLinkedQueue was tried instead, but the removal operation turned out to be too slow.
+    static final Map<Reference<?>, Boolean> references = new ConcurrentHashMap<Reference<?>, Boolean>();
+
+    // This is the actual reference queue in which the garbage collector will insert the row instances ready to be
+    // cleaned up
+    static final ReferenceQueue<NativeObject> referenceQueue = new ReferenceQueue<NativeObject>();
+
+
+@Override
+    public void run() {
+        NativeObjectReference reference;
+        while (true) {
+            try {
+                reference = (NativeObjectReference) referenceQueue.remove();
+                Row.nativeClose(reference.nativePointer);
+            } catch (InterruptedException e) {
+                //restore interrupted exception
+                Thread.currentThread().interrupt();
+            }
+        }
+    }
+}
--- a/realm/src/main/java/io/realm/internal/LinkView.java
+++ b/realm/src/main/java/io/realm/internal/LinkView.java
@@ -22,9 +22,9 @@ package io.realm.internal;
 public class LinkView {
 
     private final Context context;
-    private final long nativeLinkViewPtr;
-    private final Table parent;
-    private final long columnIndexInParent;
+    final long nativeLinkViewPtr;
+    final Table parent;
+    final long columnIndexInParent;
 
     public LinkView(Context context, Table parent, long columnIndexInParent, long nativeLinkViewPtr) {
         this.context = context;
@@ -34,8 +34,7 @@ public class LinkView {
     }
 
     public Row get(long pos) {
-        long nativeRowPtr = nativeGetRow(nativeLinkViewPtr, pos);
-        return new Row(context, parent.getLinkTarget(columnIndexInParent), nativeRowPtr);
+        return Row.get(context, this, pos);
     }
 
     public long getTargetRowIndex(long pos) {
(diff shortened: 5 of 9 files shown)


emanuelez:
@cmelchior @kneth @bmunkholm @kspangsege

realm/src/main/java/io/realm/internal/NativeObjectReference.java
```diff
@@ -0,0 +1,29 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.realm.internal;
+
+import java.lang.ref.PhantomReference;
+import java.lang.ref.ReferenceQueue;
+
+public class NativeObjectReference extends PhantomReference<NativeObject> {
```
cmelchior: Add some JavaDoc on what this class is for

realm/src/main/java/io/realm/internal/Context.java
```diff
@@ -55,6 +60,12 @@ public void executeDelayedDisposal() {
                 TableQuery.nativeClose(nativePointer);
             }
             abandonedQueries.clear();
+
+            NativeObjectReference reference;
+            while ((reference = (NativeObjectReference) rowReferenceQueue.poll()) != null) {
```
cmelchior: It would be really nice if we could move this to some kind of worker thread with a BlockingQueue as this seems kinda dangerous. It would also remove the ton of context.executeDelayedDisposal() that litters the current code, but I assume that is not possible as RowAccessors has to be closed on the thread they where created?
And calling dispose as often as possible help keep this list small?

cmelchior:
Has the code been testet on https://github.com/realm/realm-java/tree/cm-memory-leak ?

realm/src/main/java/io/realm/internal/NativeObject.java
```diff
@@ -0,0 +1,25 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.realm.internal;
+
+/**
+ * This abstract class represents a native object from core.
+ * It specifies the operations common to all such objects.
+ * All Java classes wrapping a core class should extend NativeObject.
```
cmelchior: Then we are missing some classes, only Row is extending this right now?

realm/src/main/java/io/realm/internal/Context.java
emanuelez: Not possible but not for the reason you mention.
Native destructors **can** be called from other threads, but the constraint is more subtle.
You cannot create and delete native objects belonging to the same context (shared group) concurrently. This is what prevents the creation of a worker thread.

emanuelez:
The code has been tested on a tight loop and no memory leak occurs at all. The same test is failing with the current setup.

realm/src/main/java/io/realm/internal/NativeObject.java
emanuelez: Yes, and right now we plan to keep it this way, but it's nice to have the infrastructure ready to accommodate more classes if we ever need to.

cmelchior:
Ok, apart from small comment about JavaDoc, awesome stuff. Have you tried running some of @kneth benchmark code on this? 
:+1:

realm/src/main/java/io/realm/internal/NativeObject.java
```diff
@@ -0,0 +1,25 @@
...
+ */
+public abstract class NativeObject {
+    long nativePointer;
```
kneth: Why not `nativePtr`? The `ptr` suffix is common in C++, and we use it heavily in Java too.

realm/src/main/java/io/realm/internal/Row.java
```diff
@@ -262,20 +277,8 @@ public void nullifyLink(long columnIndex) {
...
      */
     public boolean isAttached() {
-        return nativePtr != 0 && nativeIsAttached(nativePtr);
+        return nativePointer != 0 && nativeIsAttached(nativePointer);
     }
 
     protected native boolean nativeIsAttached(long nativeRowPtr);
-
-    @Override
-    protected void finalize() {
```
kneth: Yes!!!

kneth:
Great stuff.

realm/src/main/java/io/realm/internal/NativeObject.java
emanuelez: Abbreviations are not well accepted in Java for field names. They are tolerated for local variables: https://google-styleguide.googlecode.com/svn/trunk/javaguide.html

realm/src/main/java/io/realm/internal/NativeObjectReference.java
```diff
@@ -0,0 +1,34 @@
...
+ * limitations under the License.
+ */
+package io.realm.internal;
+
+import java.lang.ref.PhantomReference;
+import java.lang.ref.ReferenceQueue;
+
+/**
+ * This phantom reference extension is specialized in holding the pointer to the native objects that needs to be
+ * handled.s
```
cmelchior: A bit convoluted. Perhaps something like "This class is used for holding the reference to the native pointers present in NativeObjects. This is required as phantom references cannot access the original objects for this value."

Adding some links to external presentation(s)/doc would also be excellent. These concepts are a bit tricky to get right, even though you make it look easy with so little code ;)

cmelchior:
Apart from comment, party time :+1:

kspangsege:
So did you end up changing the handling of rows only?

emanuelez:
Yes I did. The GC is able to keep up for other objects so it's good we can delegate part of the work to the GC thread

kspangsege:
Cool. :+1:

cmelchior:
This needs an update once https://github.com/realm/realm-core/pull/826 is released.

emanuelez:
Ready for a new round of reviews @nhachicha @bmunkholm @cmelchior @kspangsege @kneth

realm/src/main/java/io/realm/internal/FinalizerRunnable.java
```diff
@@ -0,0 +1,47 @@
+package io.realm.internal;/*
```
kneth: Move `package` after preamble.

realm/src/main/java/io/realm/internal/LinkView.java
```diff
@@ -34,10 +34,11 @@ public LinkView(Context context, Table parent, long columnIndexInParent, long na
     }
 
     public Row get(long pos) {
-        long nativeRowPtr = nativeGetRow(nativeLinkViewPtr, pos);
-        return new Row(context, parent.getLinkTarget(columnIndexInParent), nativeRowPtr);
+        return Row.get(context, this, pos);
     }
 
+    native long nativeGetRow(long nativeLinkViewPtr, long pos);
```
kneth: Wasn't it supposed to be at the end?

kneth:
Needless to say but remember to update the change log :-)

emanuelez:
This change is not visible to users so it should not be in the changelog. Maybe only the performance improvement.

realm-jni/build.gradle
```diff
@@ -1,4 +1,4 @@
-ext.coreVersion = '0.89.2'
+ext.coreVersion = '0.89.5'
```
cmelchior: Nitpick: Shouldn't this have been a separate PR? :)

realm/src/main/java/io/realm/Realm.java
```diff
@@ -123,6 +126,12 @@
 public final class Realm implements Closeable {
     public static final String DEFAULT_REALM_NAME = "default.realm";
 
+    // This single thread executor ensures that only one finalizer thread ever exists
+    private static final ExecutorService executorService = Executors.newSingleThreadExecutor();
+
+    // This does not need to me thread safe since it's only used in a synchronized method
```
cmelchior: me -> be

realm/src/main/java/io/realm/internal/FinalizerRunnable.java
```diff
@@ -0,0 +1,47 @@
...
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+
+/**
+ * This runnable performs the garbage collection of the references stored in the reference queue
+ */
+public class FinalizerRunnable implements Runnable {
+
+    // Store the row references. Without this objects would be garbage collected immediately so don't remove this! ;)
```
cmelchior: without this, objects

realm/src/main/java/io/realm/internal/FinalizerRunnable.java
```diff
@@ -0,0 +1,47 @@
...
+    // A ConcurrentHashMap is used in lack of a ConcurrentHashSet in the Java API.
```
cmelchior: Perhaps a comment about the speed of this data structure compared to the others you tried.

realm/src/main/java/io/realm/internal/NativeObjectReference.java
cmelchior: This could still use an update.

realm/src/main/java/io/realm/internal/FinalizerRunnable.java
```diff
@@ -0,0 +1,47 @@
...
+    // cleaned up
+    static final ReferenceQueue<NativeObject> referenceQueue = new ReferenceQueue<NativeObject>();
+
+
+    @Override
+    public void run() {
+        while (true) {
+            NativeObjectReference reference;
+            while ((reference = (NativeObjectReference) referenceQueue.poll()) != null) {
+                Row.nativeClose(reference.nativePointer);
```
nhachicha: Wrap inside a `try .. finally` to make sure to remove from the Queue if not cleared properly from native

realm/src/main/java/io/realm/internal/Row.java
```diff
@@ -18,20 +18,33 @@
...
 
-public class Row {
+public class Row extends NativeObject {
 
     private final Context context;
     private final Table parent;
-    protected long nativePtr;
 
-    Row(Context context, Table parent, long nativePtr) {
+    private Row(Context context, Table parent, long nativePtr) {
```
cmelchior: Add some JavaDoc pointing to the static constructor methods instead.

realm/src/main/java/io/realm/internal/Row.java
```diff
@@ -18,20 +18,33 @@
...
         this.context = context;
         this.parent = parent;
-        this.nativePtr = nativePtr;
+        this.nativePointer = nativePtr;
+    }
+
+    public static Row get(Context context, Table table, long index) {
```
cmelchior: JavaDoc missing for public methods

realm/src/main/java/io/realm/internal/Row.java
```diff
@@ -18,20 +18,33 @@
...
+        long nativeRowPointer = table.nativeGetRowPtr(table.nativePtr, index);
+        Row row = new Row(context, table, nativeRowPointer);
+        FinalizerRunnable.references.put(new NativeObjectReference(row, FinalizerRunnable.referenceQueue), true);
+        return row;
+    }
+
+    public static Row get(Context context, LinkView linkView, long index) {
```
cmelchior: JavaDoc missing for public methods.

realm/src/main/java/io/realm/internal/FinalizerRunnable.java
```diff
@@ -0,0 +1,47 @@
...
+    static final Map<Reference<?>, Boolean> references = new ConcurrentHashMap<Reference<?>, Boolean>();
+
+    // This is the actual reference queue in which the garbage collector will insert the row instances ready to be
+    // cleaned up
+    static final ReferenceQueue<NativeObject> referenceQueue = new ReferenceQueue<NativeObject>();
+
+
+    @Override
+    public void run() {
+        while (true) {
```
cmelchior: Right now this thread will run until the process dies. Is this intended?

realm-jni/build.gradle
emanuelez: Yes I agree. But it's also a hard dependency so I decided to go for it since it's a one liner

cmelchior: Are there any changes to that core version that require a change log entry?

realm/src/main/java/io/realm/internal/FinalizerRunnable.java
emanuelez: Yes, we don't want to risk stopping to garbage collect objects IMO

realm/src/main/java/io/realm/internal/Row.java
emanuelez: There's not static method... it's just only used in `Row.get()`

cmelchior: Sorry for being imprecise, but it was those methods I was referring to :)

realm/src/main/java/io/realm/internal/Row.java
```diff
@@ -84,62 +97,62 @@ public Table getTable() {
...
     public double getDouble(long columnIndex) {
-        return nativeGetDouble(nativePtr, columnIndex);
+        return nativeGetDouble(nativePointer, columnIndex);
     }
 
     protected native double nativeGetDouble(long nativeRowPtr, long columnIndex);
 
     public Date getDate(long columnIndex) {
-        return new Date(nativeGetDateTime(nativePtr, columnIndex)*1000);
+        return new Date(nativeGetDateTime(nativePointer, columnIndex)*1000);
```
nhachicha: prefer `TimeUnit.SECONDS.toMillis()`

realm/src/main/java/io/realm/internal/FinalizerRunnable.java
nhachicha: Call `remove` to avoid busy waiting

realm/src/main/java/io/realm/internal/Row.java
emanuelez: We're actually converting milliseconds to seconds here

realm/src/main/java/io/realm/Realm.java
```diff
@@ -123,6 +126,12 @@
 public final class Realm implements Closeable {
     public static final String DEFAULT_REALM_NAME = "default.realm";
 
+    // This single thread executor ensures that only one finalizer thread ever exists
+    private static final ExecutorService executorService = Executors.newSingleThreadExecutor();
+
+    // This does not need to be thread safe since it's only used in a synchronized method
+    private static boolean isFinalizerStarted = false;
```
nhachicha: I think this should be `static volatile` to avoid other Threads caching the value

realm/src/main/java/io/realm/internal/FinalizerRunnable.java
```diff
@@ -0,0 +1,58 @@
...
+import java.lang.ref.Reference;
+import java.lang.ref.ReferenceQueue;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+
+/**
+ * This runnable performs the garbage collection of the references stored in the reference queue
+ */
+public class FinalizerRunnable implements Runnable {
```
nhachicha: This suppose to be a singleton AFAIK?
Can we refactor it to something like 
```java
public enum FinalizerRunnable implements Runnable {
    INSTANCE;
//...
}

File: realm/src/main/java/io/realm/internal/FinalizerRunnable.java [hunk 21]
```diff
@@ -0,0 +1,58 @@
...
+    // cleaned up
+    static final ReferenceQueue<NativeObject> referenceQueue = new ReferenceQueue<NativeObject>();
+
+
+    @Override
+    public void run() {
+        while (true) {
+            NativeObjectReference reference;
+            try {
+                while ((reference = (NativeObjectReference) referenceQueue.remove()) != null) {
```
nhachicha: Actually no need to check for `null` as remove will block until an entry is available, or wait forever.
 
this could be refactored as follow:
```java 
@Override
    public void run() {
        NativeObjectReference reference = null;
        while (true) {
            try {
                reference = (NativeObjectReference) referenceQueue.remove();
                 Row.nativeClose(reference.nativePointer);
            } catch (InterruptedException e) {
                //restore interrupted exception
                Thread.currentThread().interrupt();
            } 
        }
    }
```

realm/src/main/java/io/realm/internal/FinalizerRunnable.java
emanuelez: It is already ensured by the fact that it's only used by SingleThreadExecutor

realm/src/main/java/io/realm/internal/NativeObject.java
```diff
@@ -0,0 +1,25 @@
+/*
+ * Copyright 2014 Realm Inc.
```
cmelchior: 2015

realm/src/main/java/io/realm/internal/Row.java
cmelchior: I really hope we are doing seconds -> millis here :)

emanuelez:
Bump @cmelchior @nhachicha @kneth @bmunkholm @kspangsege

realm/src/main/java/io/realm/internal/FinalizerRunnable.java
```diff
@@ -0,0 +1,53 @@
...
+    // A ConcurrentHashMap is used in lack of a ConcurrentHashSet in the Java API.
+    // ConcurrentLinkedQueue was tried instead, but the removal operation turned out to be too slow.
+    static final Map<Reference<?>, Boolean> references = new ConcurrentHashMap<Reference<?>, Boolean>();
+
+    // This is the actual reference queue in which the garbage collector will insert the row instances ready to be
+    // cleaned up
+    static final ReferenceQueue<NativeObject> referenceQueue = new ReferenceQueue<NativeObject>();
+
+
+@Override
```
nhachicha: Code formatting :)

realm/src/main/java/io/realm/internal/Row.java
```diff
@@ -18,20 +18,47 @@
...
+     * Get the row object associated to an index in a Table
+     * @param context the Realm context
+     * @param table the Table that holds the row
+     * @param index the index of the row
+     * @return an instance of Row for the table and index specified
+     */
+    public static Row get(Context context, Table table, long index) {
+        long nativeRowPointer = table.nativeGetRowPtr(table.nativePtr, index);
+        Row row = new Row(context, table, nativeRowPointer);
+        FinalizerRunnable.references.put(new NativeObjectReference(row, FinalizerRunnable.referenceQueue), true);
```
nhachicha: please use the immutable `Boolean.TRUE` to avoid autoboxing :)

realm/src/main/java/io/realm/internal/Row.java
```diff
@@ -18,20 +18,47 @@
...
+     * Get the row object associated to an index in a LinkView
+     * @param context the Realm context
+     * @param linkView the LinkView holding the row
+     * @param index the index of the row
+     * @return an instance of Row for the LinkView and index specified
+     */
+    public static Row get(Context context, LinkView linkView, long index) {
+        long nativeRowPointer = linkView.nativeGetRow(linkView.nativeLinkViewPtr, index);
+        Row row = new Row(context, linkView.parent.getLinkTarget(linkView.columnIndexInParent), nativeRowPointer);
+        FinalizerRunnable.references.put(new NativeObjectReference(row, FinalizerRunnable.referenceQueue), true);
```
nhachicha: Ditto

realm/src/main/java/io/realm/internal/Row.java
cmelchior: I'm pretty sure the Boolean class does this internally?

cmelchior:
Apart from the last comments (TimeUnit, Boolean.True), this looks ready to merge. I know this is just an internal change, but isn't it big enough to warrant a changelog entry?

emanuelez:
I would prefer not to use TimeUnit and save the time for the method call

cmelchior:
Fair enough

nhachicha:
:+1:

cmelchior:
:+1: :shipit:

bmunkholm:
I think a changelog entry would be good. The end user will see a change in behaviour in tight loops.

PR #929, opened 2015-03-06
RealmConfiguration + new constructors

### LINKED ISSUE #511: Create an Android library depending on Realm
Hello Realm,

First of all, let me say that I find that you're doing a great work with Realm, which is really an elegant solution for database management. So keep up the good work!

I'm facing an issue, which is most a question regarding a limitation I found on Realm.

I'm trying to build an Android library integrating Realm, so that several Android applications can use a common data model system, thanks to this library.
The use case I want to achieve is that the developer of the application can have its own Realm database and can integrate the library which will deal with the common data model system at the same time.

Unfortunately, when trying to do so, I encounter the following error:
UNEXPECTED TOP-LEVEL EXCEPTION:
com.android.dex.DexException: Multiple dex files define Lio/realm/ValidationList; 

My understanding is that Realm creates this ValidationList class according to the classes that extend RealmObject inside the project, and so there is a ValidationList class created for the library and another one created for the application, which result in this Multiple dex files define error.
I also understand that the ValidationList has to be created at a unique location, inside the io.realm package.

So, my question is: is this a known limitation for Realm? Can we consider the use case feasible in a future version of Realm?

Once again, congratulations for the work done so far with Realm: it's very easy to understand and use!

Thanks,

Gauthier

cmelchior:
This fixes #511 
This PR is part of #880 

This PR changes how Realms are created. Our current approach with telescoping constructor arguments are not scalable especially as Migrations introduce new arguments and we have in-memory/read-only realms coming as well.  It also allows us to move the Android Context away from Realm.java.

### API Overview
This PR replaces the current constructors with a RealmConfiguration that are constructed using the Builder pattern:
```
// Creating a configuration, using a Builder. This makes it very scalable without having to add new constructors in Realm.java

// Full set of options are in this PR are:
RealmConfiguration config = new RealmConfiguration.Builder(getContext())
                        .name("default.realm")
                        .encryptionKey(getKey())
                        .schemaVersion(42)
                        .migration(new MyMigration())
                        .addModule(new MyModule())  // Add 1 module to already set modules
                        .setModules(new MyModule(),  new MyLibraryModule()) // Replace current modules
                        .deleteRealmIfMigrationNeeded()
                        .schema(Foo.class, Bar.class) // package protected method, only used for internal testing.
                        .build()

// Constructors are:
RealmConfiguration.Builder builder = new RealmConfiguration.Builder(Context ctx);
RealmConfiguration.Builder builder = new RealmConfiguration.Builder(File realmDir);

I don't think a directory is needed for in-memory Realms, but requiring a folder in the constructor makes it a lot less error prone.

// Introduce the concept of a default Realm (This is also the same on Cocoa where you can configure a default Realm)
Realm.setDefaultConfiguration(config); // A default configuration must be set first
Realm.getDefaultInstance(); // Making it even easier to get a default instance
Realm.getInstance(config); // Getting the instance for any given configuration

// These two are similar
Realm.getInstance(getContext())
Realm.getInstance(new RealmConfiguration.Builder(getContext()).build())

// Deprecated methods
Realm.getInstance(Context context, String fileName)
Realm.getInstance(Context context, byte[] key)
Realm.getInstance(Context context, String fileName, byte[] key)
Realm.getInstance(File writeableFolder)
Realm.getInstance(File writeableFolder, String fileName)
Realm.getInstance(File writeableFolder, byte[] key)
Realm.getInstance(File writeableFolder, String fileName, byte[] key)
Realm.deleteRealmFile(Context context, String fileName)
Realm.deleteRealmFile(Context context)
Realm.migrateRealmAtPath(String realmPath, RealmMigration migration)
Realm.migrateRealmAtPath(String realmPath, byte[] key, RealmMigration migration)
Realm.migrateRealmAtPath(String realmPath, RealmMigration migration, boolean autoRefresh)
Realm.migrateRealmAtPath(String realmPath, byte[] key, RealmMigration migration, boolean autoUpdate)
Realm.compactRealmFile(Context context, String fileName)
Realm.compactRealmFile(Context context)

// Note that the following constructor is kept to make it easier for Realm beginners? Is this just API bloat?
Realm.getInstance(Context context)

// New methods replacing them
Realm.getDefaultInstance();
Realm.getInstance(RealmConfiguration realmConfig)
Realm.deleteRealm(RealmConfiguration realmConfig);
Realm.migrateRealm(RealmConfiguration realmConfig);
Realm.migrateRealm(RealmConfiguration realmConfig, RealmMigration migration);
Realm.compactRealm(RealmConfiguration realmConfig);
```

### SchemaVersion and Migration
The current interaction between using ```.schemaVersion()``` and ```.migration()``` is the following: 

1) If only schemaVersion is set we make a no-op migration behind the scenes that try to upgrade the version number. Schema is validated afterwards and throw an exception if there is mismatch.

2) If newSchemaVersion > currentSchemaVersion and a migration block is set. This migration will be called when first opening the Realm. We also have a method on the Realm for manually triggering this migration. Useful if you want to upgrade on a background thread: Realm.migrateRealm(RealmConfiguration)

3) Setting a migration block without a schemaVersion is a no-op, ie. migration code is never called. We might want to log a warning or throw an exception for this.

4) Setting a newSchemaVersion < currentSchemaVersion will throw an exception when opening the Realm.

5) Setting newSchemaVersion == currentSchemaVersion and a schema mismatch is detected will throw an exception when opening the Realm.

6) The current RealmMigration interface will be kept until we introduce the new Migration API as it will introduce breaking changes. 

### Discussion
Some implementation details that could be discussed:
- A RealmConfiguration is immutable once created.
- How to handle multiple configurations pointing to the same Realm. Currently it is allowed (eg. needed for ReadOnly realms).

This PR will also potentially deprecate a lot of methods (all constructors + all migrateRealmAtPath() + deleteRealmFile) , so I would rather introduce it all at once, and then having the option of extending RealmConfiguration afterwards.

**Missing:**
- [x] Determine policy for when migrations are needed and interaction with schemaVersion. Check Cocoa: https://github.com/realm/realm-cocoa/issues/1584
- [x] Determine policy for multiple configurations pointing to the same Realm file. See #1156 
- [ ] Update examples (separate PR)
- [x] Add example for using library projects with Realm (separate PR)
- [ ] Update website/documentation (separate PR)

----


emanuelez:
I like the direction, but we're breaking the API drastically. I think we should maintain the old constructors while deprecating them until the following big release.

cmelchior:
Yeah that was my plan as well, this commit might have cleaned up a bit to drastically

timanglade:
@alazier Would love to get your thoughts on this. See also #java on slack for additional conversation about this PR.

emanuelez:
Also, we might want to keep static constructor for the most common case:

* Context or File
* Context of File + Migration + Version

alazier:
@timanglade this is something I have proposed for the cocoa apis multiple times (and think I have mentioned it to @cmelchior). As the number of global configuration parameters grows this will likely be a necessity.

My concerns with this proposal would be:
- `version` should be `schemaVersion` to remove any ambiguity
- In cocoa we will also need parameters for `readOnly`, `inMemoryIdentifier` - not sure if you support these in java 
- We might also want to consider support notifications so that all realm configuration can be done in one place. We would still want register/unregister methods though for users who want to change notifications after realm initialization
- Is `name` the same thing as `path`? Do we just use different terminologies on java vs cocoa?

cmelchior:
@alazier 
We have actually been talking about trying to move away from SQL terminology on the Android side, ie. 
version instead of schemaVersion, Modules/ClassCollection instead of schema etc. to avoid people associating Realm with SQL too much. However it has not been decided yet and should probably be agreeded on for both platforms. What is your stance on that?

We don't support ReadOnly and InMemory yet, but it is coming.

I don't think the configuration is a good place for adding notification setup as they are added/removed a lot more while the Realm is used. I would prefer to keep the setup methods to initialisation properties.

Name is filename, on android path is split into 2 methods, one for folder and one for filename as getting a default document folder is a bit more complicated.

cmelchior:
I have updated the top note with current API design. Feedback welcome. We still miss support for custom schemas but that will be included ons #938 has been merged.

@segiddins @bmunkholm @emanuelez @kneth

realm/src/androidTest/java/io/realm/RealmConfigurationTest.java
```diff
@@ -0,0 +1,255 @@
...
+                .schema(Dog.class)
+                .create());
+        realm.close();
+
+        try {
+            Realm.getInstance(new RealmConfiguration.Builder(getContext())
+                    .schemaVersion(42)
+                    .schema(AllTypesPrimaryKey.class)
+                    .build());
+            fail("A migration should be required");
```
kneth: It seems to be an inconsistency how you use `try`/`catch` and `fail()` in this unit test and other unit tests.

kneth:
Will you add an option to compact the Realm file before opening?

cmelchior:
I didn't consider that. It would be fairly easy adding an option like:

```config.compactRealmWhenClosing()``` or ```config.compactRealmWhenOpening()``` 

However I am a bit worried it would introduce a unpredictable slowdown as compaction could potentially take a lot of time + plus it require additional space on the disc ?

**Comment by kneth** [comment]
Realm Core might be able to provide information about how much space is saved before actually calling compact.

**Comment by cmelchior** [comment]
This PR is leaving WIP status and is ready for review. OP has been updated with current status and design thoughts.

@emanuelez @kneth @bmunkholm @segiddins

File: realm-annotations/src/main/java/io/realm/annotations/RealmModule.java [hunk 2]
```diff
@@ -0,0 +1,85 @@
...
+ * @see <a href="">TODO Example of a project using modules</a>
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.TYPE)
+@Inherited
+public @interface RealmModule {
+
+    /**
+     * Setting this to true will mark this module as a library module. This will prevent Realm from generating the
+     * {@code DefaultRealmModule} containing all classes. This is required by libraries so they do notintefer with
```
kneth: "notintefer" -> "not interfere"

emanuelez:
The moving or RealmModule should have been a different PR.
We really have to strive to try to make such big PRs as small as possible

realm/src/main/java/io/realm/RealmConfiguration.java
```diff
@@ -0,0 +1,383 @@
...
+            }
+            return this;
+        }
+
+        /**
+         * DEBUG method. This restricts the Realm schema to only consist of the provided classes without having to
+         * create a module. These classes must be available in the default module. Calling this will remove any
+         * previously configured modules.
+         */
+        Builder schema(Class<? extends RealmObject>... schemaClass) {
```
emanuelez: Even if it's just a method for debugging purposes, we should still use varargs properly

changelog.txt
```diff
@@ -1,3 +1,11 @@
+0.81
+ * Introduced RealmModules for working with custom schemas in libraries and apps.
+ * Introduced Realm.getDefaultInstance(), Realm.setDefaultInstance(RealmConfiguration) and Realm.getInstance(RealmConfiguration).
+ * Deprecated most constructors. They have been been replaced by RealmConfiguration objects.
+ * Deprecated Realm.migrateRealmAtPath(). It has been Replaced by Realm.migrateRealm(RealmConfiguration).
```
kneth: "Replaced" -> "replaced"

realm-annotations-processor/src/main/java/io/realm/processor/ModuleMetaData.java
```diff
@@ -82,16 +75,12 @@ public boolean generate(ProcessingEnvironment processingEnv) {
...
-                    }
-                    ClassMetaData metadata = classMetaData.get(clazz.getName());
+                Set<String> classNames = getClassMetaDataFromModule(classElement);
+                for (String fullyQualifiedClassName : classNames) {
+                    ClassMetaData metadata = classMetaData.get(fullyQualifiedClassName);
                     if (metadata == null) {
-                        Utils.error(Utils.stripPackage(qualifiedName) + " could not be added to the module. It is not " +
-                                "possible to add classes which are part of another library.");
+                        Utils.error(Utils.stripPackage(fullyQualifiedClassName) + " could not be added to the module. " +
+                                "Only classes extending RealmObject that are part of this project can be added.");
```
kneth: "RealmObject that are part of this project can be added" -> RealmObject, which are part of this project, can be added"

realm-annotations-processor/src/test/resources/some/test/AppModuleAllClasses.java
```diff
@@ -0,0 +1,29 @@
+/*
+ * Copyright 2014 Realm Inc.
```
kneth: 2015

realm-annotations-processor/src/test/resources/some/test/AppModuleCustomClasses.java
```diff
@@ -0,0 +1,29 @@
+/*
+ * Copyright 2014 Realm Inc.
```
kneth: 2015

realm-annotations-processor/src/test/resources/some/test/InvalidAppModuleMixedParameters.java
```diff
@@ -0,0 +1,30 @@
+/*
+ * Copyright 2014 Realm Inc.
```
kneth: 2015

realm-annotations-processor/src/test/resources/some/test/InvalidAppModuleWrongType.java
```diff
@@ -0,0 +1,30 @@
+/*
+ * Copyright 2014 Realm Inc.
```
kneth: 2015

realm-annotations-processor/src/test/resources/some/test/InvalidLibraryModuleMixedParameters.java
```diff
@@ -0,0 +1,30 @@
+/*
+ * Copyright 2014 Realm Inc.
```
kneth: 2015

realm-annotations-processor/src/test/resources/some/test/InvalidLibraryModuleWrongType.java
```diff
@@ -0,0 +1,30 @@
+/*
+ * Copyright 2014 Realm Inc.
```
kneth: 2015

realm-annotations-processor/src/test/resources/some/test/LibraryModuleAllClasses.java
```diff
@@ -0,0 +1,29 @@
+/*
+ * Copyright 2014 Realm Inc.
```
kneth: 2015

realm-annotations-processor/src/test/resources/some/test/LibraryModuleCustomClasses.java
```diff
@@ -0,0 +1,29 @@
+/*
+ * Copyright 2014 Realm Inc.
```
kneth: 2015

realm-annotations/src/main/java/io/realm/annotations/RealmModule.java
```diff
@@ -0,0 +1,85 @@
...
+package io.realm.annotations;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Inherited;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * By default a Realm can stores all RealmClasses in a project. However, if you want to restrict a Realm to only contain
```
kneth: Is `RealmClass` an well-defined term?

realm-annotations/src/main/java/io/realm/annotations/RealmModule.java
```diff
@@ -0,0 +1,85 @@
...
+ * certain classes or want to share them between a library project and an app project you must use a RealmModule.
```
kneth: "certain" -> "given"

realm-annotations/src/main/java/io/realm/annotations/RealmModule.java
```diff
@@ -0,0 +1,85 @@
...
+ * <p>
+ * A RealmModule is a collection of RealmClasses that can be combined with other RealmModules to create the schema for a
+ * Realm. This makes it easier to control versioning and migration of those Realms.
+ * <p>
+ * A RealmModule can either be a library module or an app module. This distinction is made by setting
```
kneth: "This" -> "The"

realm-annotations/src/main/java/io/realm/annotations/RealmModule.java
```diff
@@ -0,0 +1,85 @@
...
+ * {@code library = true}. A default Realm will automatically create a RealmModule called {@code DefaulRealmModule}
+ * that contains all RealmClasses in a project. This module will be an app module and is automatically known by Realm.
```
kneth: "that" -> "which"

realm-annotations/src/main/java/io/realm/annotations/RealmModule.java
```diff
@@ -0,0 +1,85 @@
...
+ * <p>
+ * This behavior is problematic when combining a library project and an app project that both uses Realm as
```
kneth: This section is too long, and should be rephrased.

realm-annotations/src/main/java/io/realm/annotations/RealmModule.java
```diff
@@ -0,0 +1,85 @@
...
+ * the {@code DefaultRealmModule} will be created for both the library project and the app project causing import
+ * conflicts.
+ * <p>
+ * Library authors are reponsible for avoiding this conflict by creating a explicit library module where
```
kneth: "reponsible" -> "responsible"

realm-annotations/src/main/java/io/realm/annotations/RealmModule.java
```diff
@@ -0,0 +1,85 @@
...
+ * {@code library = true} is set. This disable the generation of the DefaultRealmModule for the library project and
```
kneth: "This disable" -> "This disables"

realm-annotations/src/main/java/io/realm/annotations/RealmModule.java
```diff
@@ -0,0 +1,85 @@
...
+     * Realm classes in this project. This does not include classes from other libraries which must be exposed using
+     * their own module.
+     *
+     * Setting both {@code allClasses = true} and {@code classes()} will result in the annotation processor throwing
+     * an error.
+     */
+    boolean allClasses() default false;
+
+    /**
+     * Specify the Realm classes part of this module. Only classes in this project can be included. Realm classes
```
kneth: "Specify" -> "Specifies"

realm-annotations/src/main/java/io/realm/annotations/RealmModule.java
```diff
@@ -0,0 +1,85 @@
...
+     * from other libraries must be exposed using their own module.
+     *
+     * Setting both {@code allClasses = true} and {@code classes()} will result in the annotation processor throwing
+     * an error.
```
kneth: "an error" -> "an exception"

cmelchior:
Bump @kneth @emanuelez

realm-annotations/src/main/java/io/realm/annotations/RealmModule.java
kneth: @cmelchior See my previous comments

realm/src/androidTest/java/io/realm/RealmConfigurationTest.java
```diff
@@ -0,0 +1,288 @@
...
+        try {
+            new RealmConfiguration.Builder(getContext()).encryptionKey(null).build();
+            fail();
+        } catch (IllegalArgumentException expected) {
+        }
+    }
+
+    public void testWrongKeyLengthThrows() {
+        try {
+            new RealmConfiguration.Builder(getContext()).encryptionKey(new byte[63]).build();
```
kneth: What about 65?

realm/src/androidTest/java/io/realm/RealmConfigurationTest.java
```diff
@@ -0,0 +1,288 @@
...
+        } catch (IllegalArgumentException expected) {
+        }
+    }
+
+    public void testVersionLessThanDiscVersionThrows() {
+        realm = Realm.getInstance(new RealmConfiguration.Builder(getContext()).schemaVersion(42).build());
+        realm.close();
+
+        try {
+            Realm.getInstance(new RealmConfiguration.Builder(getContext()).schemaVersion(1).build());
```
kneth: What about 41?

realm/src/androidTest/java/io/realm/RealmTest.java
```diff
@@ -1586,7 +1601,11 @@ public void testWrongKeyShouldThrow() {
 
         // Wrong key size
         try {
-            Realm.getInstance(getContext(), WRONG_KEY_REALM, new byte[63]);
+            Realm.getInstance(new RealmConfiguration.Builder(getContext())
+                    .name(WRONG_KEY_REALM)
+                    .encryptionKey(new byte[63])
```
kneth: What about 65?

realm/src/androidTest/java/io/realm/TestHelper.java
```diff
@@ -74,6 +75,14 @@ public static void prepareDatabaseFromAssets(Context context, String realmPath,
         TestHelper.copyRealmFromAssets(context, realmPath, newName);
     }
 
+    // Returns a random used used by encrypted Realms.
```
kneth: "used used" -> "used"

realm/src/main/java/io/realm/Realm.java
```diff
@@ -459,61 +453,137 @@ public static Realm getInstance(File writeableFolder, byte[] key) {
...
+     */
+    public static Realm getInstance(RealmConfiguration configuration) {
+        if (configuration == null) {
+            throw new IllegalArgumentException("A non-null RealmConfiguration must be provided");
+        }
+        return create(configuration);
+    }
+
+    /**
+     * Set the {@link io.realm.RealmConfiguration} used when calling {@link #getDefaultInstance()}
```
kneth: "Set" -> "Sets"

realm/src/main/java/io/realm/RealmConfiguration.java
```diff
@@ -0,0 +1,390 @@
...
+            DEFAULT_MODULE_MEDIATOR = getModuleMediator(DEFAULT_MODULE.getClass().getCanonicalName());
+        } else {
+            DEFAULT_MODULE_MEDIATOR = null;
+        }
+    }
+
+    private final File realmFolder;
+    private final String realmFileName;
+    private final String canonicalPath;
+    private final byte[] key;
```
kneth: Are there any security concerns when we keep an encryption key in memory?

realm/src/main/java/io/realm/RealmConfiguration.java
```diff
@@ -0,0 +1,390 @@
...
+        }
+
+        /**
+         * Sets the 64 bit key used to encrypt and decrypt the Realm file.
+         */
+        public Builder encryptionKey(byte[] key) {
+            if (key == null) {
+                throw new IllegalArgumentException("A non-null key must be provided");
+            }
+            if (key.length != 64) {
```
kneth: The value 64 is used in many places. Maybe we should promote it to a constant?

realm/src/main/java/io/realm/RealmConfiguration.java
cmelchior: Not really. Core does the same thing. It is an attack vector, but the current policy is that we only want to guard against attacks on the filesystem.

realm/src/main/java/io/realm/RealmConfiguration.java
```diff
@@ -0,0 +1,390 @@
...
+         *
+         * <bold>WARNING!</bold> This will result in loss of data.
+         */
+        public Builder deleteRealmIfMigrationNeeded() {
+            this.deleteRealmIfMigrationNeeded = true;
+            return this;
+        }
+
+        /**
+         * Add a {@link RealmModule}s to the existing modules. RealmClasses in the new module is added to the schema
```
kneth: "Add" -> "Adds"

realm/src/main/java/io/realm/RealmConfiguration.java
```diff
@@ -0,0 +1,390 @@
...
+         * annotation.
+         */
+        public Builder addModule(Object module) {
+            checkModule(module);
+            modules.add(module);
+            return this;
+        }
+
+        /**
+         * Replace the existing module(s) with one or more {@link RealmModule}s. Using this method will replace the
```
kneth: "Replace" -> "Replaces"

kneth:
I believe it is ready to be merged (when my few last comments are addressed).

emanuelez:
Please make this merge-able and address @kneth 's comments. I think this is ready for prime time.

cmelchior:
Any final comments @kneth @emanuelez @nhachicha @bmunkholm before I merge this?

realm-annotations-processor/src/test/java/io/realm/processor/RealmProcessorTest.java
```diff
@@ -29,7 +32,6 @@
     private JavaFileObject simpleModel = JavaFileObjects.forResource("some/test/Simple.java");
     private JavaFileObject simpleProxy = JavaFileObjects.forResource("io/realm/SimpleRealmProxy.java");
     private JavaFileObject allTypesModel = JavaFileObjects.forResource("some/test/AllTypes.java");
-    private JavaFileObject allTypesProxy = JavaFileObjects.forResource("io/realm/AllTypesRealmProxy.java");
```
emanuelez: we should still check that we produce the expected model class proxy!

cmelchior:
RealmConfiguration.addModule() removed and Realm.getDefaultModule() introduced instead.

@emanuelez @nhachicha @kneth

emanuelez:
:shipit:

PR #942, opened 2015-03-10
Make Realm.getTable(class) work with proxy classes as well.

emanuelez:
This solves an issue occurring when trying to update a Realm backed
object.

### DIFF
--- a/realm/src/main/java/io/realm/Realm.java
+++ b/realm/src/main/java/io/realm/Realm.java
@@ -168,6 +168,9 @@ public final class Realm implements Closeable {
     private final Map<Class<?>, Method> initTableMethods = new HashMap<Class<?>, Method>();
     private final Map<Class<?>, Method> insertOrUpdateMethods = new HashMap<Class<?>, Method>();
     private final Map<Class<?>, Constructor> generatedConstructors = new HashMap<Class<?>, Constructor>();
+
+    // Maps classes to the name of the proxied class. Examples: Dog.class -> Dog, DogRealmProxy -> Dog
+    private final Map<Class<?>, String> proxiedClassNames = new HashMap<Class<?>, String>();
     private final List<RealmChangeListener> changeListeners = new ArrayList<RealmChangeListener>();
     private final Map<Class<?>, Table> tables = new HashMap<Class<?>, Table>();
     private static final Set<Class<? extends RealmObject>> customSchema = new HashSet<Class<? extends RealmObject>>();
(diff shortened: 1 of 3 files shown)


emanuelez:
Changelog:
 * Fixed an issue when trying to update a Realm backed object

emanuelez:
@bmunkholm @cmelchior @kneth

realm/src/main/java/io/realm/Realm.java
```diff
@@ -168,6 +168,9 @@
     private final Map<Class<?>, Method> initTableMethods = new HashMap<Class<?>, Method>();
     private final Map<Class<?>, Method> insertOrUpdateMethods = new HashMap<Class<?>, Method>();
     private final Map<Class<?>, Constructor> generatedConstructors = new HashMap<Class<?>, Constructor>();
+
+    // Maps classes to the name of the proxied class. Examples: Dog.class -> Dog, DogRealmProxy -> Dog
```
kneth: You only store the latter, right?

emanuelez: I store both. The Map goes from Class to String and it's purpose is to reduce the need for introspection.

realm/src/main/java/io/realm/Realm.java
```diff
@@ -317,12 +320,14 @@ public void setAutoRefresh(boolean autoRefresh) {
...
     public Table getTable(Class<?> clazz) {
-        String simpleClassName = simpleClassNames.get(clazz);
-        if (simpleClassName == null) {
-            simpleClassName = clazz.getSimpleName();
-            simpleClassNames.put(clazz, simpleClassName);
+        final String proxySuffix = "RealmProxy";
+        String proxiedClassName = proxiedClassNames.get(clazz);
+        if (proxiedClassName == null) {
+            String classSimpleName = clazz.getSimpleName();
+            proxiedClassName = classSimpleName.replace(proxySuffix, "");
```
cmelchior: The name seems a bit weird. Since this is the "real" model class name, not the proxy class name as you just removed the proxy suffix

cmelchior:
Apart from comment :+1:

realm/src/main/java/io/realm/Realm.java
emanuelez: you are correct, but please notice it's called proxiedClassNames, not proxyClassNames. Meaning it's the name of the class being proxied.

Issue #917, opened 2015-03-01
read only query require a transaction, exception is Objects cannot be changed outside a transaction; see beginTransaction() for details


eboudrant:
I am doing a simple query and Realm complain about write transaction :
```
  public List<Installed> getInstalled(Realm realm, String pid) {
    RealmResults<Installed> packageQuery = realm.where(Installed.class)
        .equalTo("pid", pid)
        .findAll();
    return packageQuery;
  }
```
Exception is
```
java.lang.IllegalStateException: Objects cannot be changed outside a transaction; see beginTransaction() for details.
       at io.realm.internal.Group.throwImmutable(Group.java:320)
       at io.realm.internal.Group.getTable(Group.java:220)
       at io.realm.Realm.getTable(Realm.java:323)
       at io.realm.RealmQuery.(RealmQuery.java:75)
       at io.realm.Realm.where(Realm.java:1279)
       at n.e.a.realm.RealmDatasource.getInstalled(RealmDatasource.java:247)
```

cmelchior:
Hmm, it seems our error message in that place is a bit wrong. That exception is actually thrown because Realm doesn't know about your Installed class.

Does it extend RealmObject?

eboudrant:
Yes, it extend RealmObject.

cmelchior:
Hi @eboudrant 
Looking into this, I cannot reproduce it. So I have a few follow up questions?

- Can you reproduce this behaviour if you try to uninstall the app and then reinstall it.
- Are you manually doing a RealmMigration when starting the app?
- Is your Installed class part of the app code or part of a library project?

cmelchior:
Better error reporting has been merged. @eboudrant Are you still seeing this behavior on 0.80?

cmelchior:
Closing this due to no response. Please add additional information if the problem stil persists.

eboudrant:
Hello, I just updated from 0.79 t0 0.80 and it crash instantly : https://github.com/realm/realm-java/issues/982

PR #948, opened 2015-03-11
Improved error message for unknown tables

cmelchior:
Fixes wrong error message described in https://github.com/realm/realm-java/issues/917

@emanuelez @kneth @bmunkholm

### DIFF
--- a/realm/src/main/java/io/realm/internal/Group.java
+++ b/realm/src/main/java/io/realm/internal/Group.java
@@ -217,7 +217,8 @@ public class Group implements Closeable {
         }
         if (immutable) {
             if (!hasTable(name)) {
-                throwImmutable();
+                throw new IllegalStateException("Requested table is not in this Realm. " +
+                        "Creating it require a transaction: " + name);
             }
         }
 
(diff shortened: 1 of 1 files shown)


realm/src/main/java/io/realm/internal/Group.java
```diff
@@ -217,7 +217,8 @@ public Table getTable(String name) {
         }
         if (immutable) {
             if (!hasTable(name)) {
-                throwImmutable();
+                throw new IllegalStateException("Requested table is not in this Realm. " +
+                        "Creating it require a transaction: " + name);
```
emanuelez: requires*

realm/src/main/java/io/realm/internal/Group.java
```diff
@@ -217,7 +217,8 @@ public Table getTable(String name) {
         }
         if (immutable) {
             if (!hasTable(name)) {
-                throwImmutable();
+                throw new IllegalStateException("Requested table is not in this Realm. " +
```
kneth: I think that the table name (`name`) should be place earlier. In the lines of: "Requested table 'name' is not in this Realm. Creating the table must be done within a transaction."

PR #962, opened 2015-03-15
Change `Random` to `SecureRandom` to generate encryption key

kishikawakatsumi:
Change `Random` to `SecureRandom` to generate encryption key.
I guess it is better as example.

### DIFF
--- a/examples/encryptionExample/src/main/java/io/realm/examples/encryptionexample/EncryptionExampleActivity.java
+++ b/examples/encryptionExample/src/main/java/io/realm/examples/encryptionexample/EncryptionExampleActivity.java
@@ -20,7 +20,7 @@ import android.app.Activity;
 import android.os.Bundle;
 import android.util.Log;
 
-import java.util.Random;
+import java.security.SecureRandom;
 
 import io.realm.Realm;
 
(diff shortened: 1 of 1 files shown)


cmelchior:
The example doesn't show a proper way of generating/keeping a proper key either way. So I would be hesitant adding code that might make it seem like it is a valid approach. You are right that SecureRandom is better than Random, but in this case I would rather keep Random as to not give people the wrong idea.

For proper key handling  we would either have to show how to use the secure keystore from API 18 or how to get a key from a server. Both would require a lot of code that might distract from the purpose of the example.

cmelchior:
@emanuelez @kneth

kneth:
We have to be aware of the fact that our examples and unit tests are going to be used by our users as part of the documentation. By using `SecureRandom` we might hint to the less knowledgable user where to look.

kishikawakatsumi:
And there is a little difference between the documentation and sample code, that makes users may be confused. 

http://realm.io/docs/java/0.80.0/#encryption

The documentation says:
```
byte[] key = new byte[32];
new SecureRandom().nextBytes(key);
Realm realm = Realm.create(this, key);
```

Using `SecureRandom` but key length is **32** on the document.


I would like to fix both sample code and documentation to same code.
Using `SecureRandom` and key length is **64**.

Related: https://github.com/realm/realm.io/pull/372

kneth:
@kishikawakatsumi Thanks for pointing out that the documentation is outdated.

Issue #987, opened 2015-03-23
Create object with values aka. Optimized insert API


kneth:
Creating a new object in a Realm is often done like:

    realm.beginTransaction();
    Person person = realm.createObject(Person.class);
    person.setName("Arthur");
    person.setAge(42);
    realm.commit();

It turns out that the calls to the setters are checking if the `person` object accessed from the correct thread (calling `Realm.checkIfValid())`. And this checking takes up 40-50 % of the time spend in the setters. Typically, creating object and setting the fields' values are always done together, and a simple method could be useful:

    realm.beginTransaction();
    Person person = realm.createObjectWithValues(Person.class, "Arthur", 42);
    realm.commit();

Besides a simpler API for creating objects with values, the method will also increase performance as only one call to `Realm.checkIfValue())` is done.

See also http://stackoverflow.com/questions/29214236/how-to-add-1-milion-items-in-realm-correctly/29216133#29216133

cmelchior:
Merging this into #1684 as it is effectively the same thing

PR #988, opened 2015-03-24
Adding Realm.createObjectWithValues()

kneth:
As outlined in #987 the performance of `Realm.createObject()` and setters might not be as desired. The new method `Realm.createObjectWithValues()` is implemented to give better performance for batch inserts.

To do:
- [ ] updating change log ("Added Realm.createObjectWithValues() for better performance when inserting many objects.")
- [ ] benchmarking the new method

### DIFF
--- a/realm/src/androidTest/java/io/realm/RealmTest.java
+++ b/realm/src/androidTest/java/io/realm/RealmTest.java
@@ -26,6 +26,7 @@ import java.nio.channels.FileChannel;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Date;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Random;
 import java.util.Scanner;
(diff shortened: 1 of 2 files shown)


realm/src/main/java/io/realm/Realm.java
```diff
@@ -1060,30 +1086,70 @@ public void writeEncryptedCopyTo(File destination, byte[] key) throws IOExceptio
...
+     */
+    public <E extends RealmObject> E createObjectWithValues(Class<E> clazz, HashMap<String, Object> values) {
+        checkIfValid();
+        Table table = tables.get(clazz);
+        if (table == null) {
+            table = createTable(clazz);
         }
-
         long rowIndex = table.addEmptyRow();
+        for (String key : values.keySet()) {
```
emanuelez: Iterate over entryset

realm/src/androidTest/java/io/realm/RealmTest.java
```diff
@@ -1511,4 +1512,50 @@ public void testUpdateObjectWithLinks() throws Exception {
...
+        assertEquals(true, allTypeses.first().isColumnBoolean());
+        assertEquals(3.1415d, allTypeses.first().getColumnDouble());
+        assertEquals("Hello World", allTypeses.first().getColumnString());
+    }
+
+    public void testCreateObjectWithWrongValues() {
+        // not existing field
+        testRealm.beginTransaction();
+        try {
+            HashMap<String, Object> hashMap = new HashMap<String, Object>();
```
emanuelez: The first HashMap should be Map instead

realm/src/main/java/io/realm/Realm.java
```diff
@@ -1060,30 +1086,70 @@ public void writeEncryptedCopyTo(File destination, byte[] key) throws IOExceptio
...
+            }
+            ColumnType columnType = table.getColumnType(columnIndex);
+            Object object = values.get(key);
+            if (columnType.matchObject(object)) {
+                switch (columnType) {
+                    case BOOLEAN:
+                        table.setBoolean(columnIndex, rowIndex, ((Boolean) object).booleanValue());
+                        break;
+                    case INTEGER:
+                        table.setLong(columnIndex, rowIndex, ((Long) object).longValue());
```
tmeunier: Column type of type INTEGER is matching Long, Integer, Short and Byte.
That line will cause a ClassCastException if your object is not a Long.

Issue #993, opened 2015-03-27
Realm.copyToRealm() should set PrimaryKey values directly


RobertoArtiles:
Hi guys,

It's known that null is not supported. But it seems like it could lead to the DB corruption if you try to write something with a null @PrimaryKey field. The best would be to just crash before putting something into the DB.
So what happened with me. I by mistake executed something like this:
```java
        Realm realm = Realm.getInstance(this);
        realm.beginTransaction();
        UserRealm userRealm = realm.createObject(UserRealm.class);
        realm.commitTransaction();
        realm.close();
```
UserRealm class:
```java
public class UserRealm extends RealmObject {
    @PrimaryKey private String token;
    
    // getter and setter
}
```
So the first time I executed it nothing crashed. But starting from the second execution it was throwing ```   Caused by: io.realm.exceptions.RealmException: Primary key constraint broken. Value already exists:```. And even when I tried to write something with a set primary key it still failed with the same exception:
```java
        Realm realm = Realm.getInstance(this);
        realm.beginTransaction();
        UserRealm userRealm = realm.createObject(UserRealm.class);
        userRealm.setToken("some_token");
        realm.commitTransaction();
        realm.close();
```
And if you do ```realm.allObjects(UserRealm.class)``` it returns one user with a null token. So basically DB is corrupted after that. And you can't write anything in this "table".

Looks like a great way to shoot yourself in both legs in production.

Full Stacktrace:
```
io.realm.exceptions.RealmException: Primary key constraint broken. Value already exists:
            at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2184)
            at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:2233)
            at android.app.ActivityThread.access$800(ActivityThread.java:135)
            at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1196)
            at android.os.Handler.dispatchMessage(Handler.java:102)
            at android.os.Looper.loop(Looper.java:136)
            at android.app.ActivityThread.main(ActivityThread.java:5001)
            at java.lang.reflect.Method.invokeNative(Native Method)
            at java.lang.reflect.Method.invoke(Method.java:515)
            at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:785)
            at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:601)
            at dalvik.system.NativeStart.main(Native Method)
[...]

PR #995, opened 2015-03-27
Fixed bug in Realm.copyToRealm() with primary key values

cmelchior:
Fixes https://github.com/realm/realm-java/issues/993

This PR adds a package protected method ```Realm.createObject(Class clazz, Object primaryKeyValue);```

This was needed to fix the above bug, and we have before talked about adding such a method as Cocoa has it. I am still not convinced we should make it public as using copyToRealm() is an acceptable workaround imho (See isse for details), but I am open for discussing it.

@emanuelez @kneth @bmunkholm

### DIFF
--- a/realm/src/main/java/io/realm/Realm.java
+++ b/realm/src/main/java/io/realm/Realm.java
@@ -213,8 +213,8 @@ public final class Realm implements Closeable {
     protected void finalize() throws Throwable {
         if (sharedGroup != null) {
             RealmLog.w("Remember to call close() on all Realm instances. " +
-                    "Realm " + path + " is being finalized without being closed, " +
-                    "this can lead to running out of native memory."
+                            "Realm " + path + " is being finalized without being closed, " +
+                            "this can lead to running out of native memory."
             );
         }
         super.finalize();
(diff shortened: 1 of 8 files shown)


realm/src/main/java/io/realm/Realm.java
```diff
@@ -1089,6 +1089,47 @@ public void writeEncryptedCopyTo(File destination, byte[] key) throws IOExceptio
         return get(clazz, rowIndex);
     }
 
+    /**
+     * Creates a new object inside the Realm with the Primary key value initially set.
+     * If the value violates the primary key constraint, no object will be added and and
```
emanuelez: and and -> and a

realm/src/main/java/io/realm/Realm.java
```diff
@@ -1089,6 +1089,47 @@ public void writeEncryptedCopyTo(File destination, byte[] key) throws IOExceptio
...
+     * {@link RealmException will be thrown}.
```
emanuelez: close brackets after RealmException

realm/src/main/java/io/realm/Realm.java
```diff
@@ -1089,6 +1089,47 @@ public void writeEncryptedCopyTo(File destination, byte[] key) throws IOExceptio
...
+                    method = generatedClass.getMethod("initTable", new Class[]{ImplicitTransaction.class});
+                } catch (NoSuchMethodException e) {
+                    throw new RealmException("Could not find the initTable() method in generated proxy class: " + APT_NOT_EXECUTED_MESSAGE);
+                }
+                initTableMethods.put(generatedClass, method);
+            }
+
+            try {
+                table = (Table) method.invoke(null, transaction);
+                tables.put(clazz, table);
```
emanuelez: This should be after the catches

realm/src/main/java/io/realm/Realm.java
```diff
@@ -1059,8 +1059,29 @@ public void writeEncryptedCopyTo(File destination, byte[] key) throws IOExceptio
...
-        Table table;
-        table = tables.get(clazz);
+        Table table = initTable(clazz);
+        long rowIndex = table.addEmptyRow();
+        return get(clazz, rowIndex);
+    }
+
+    /**
+     * Creates a new object inside the Realm with the Primary key value initially set.
+     * If the value violates the primary key constraint, no object will be added and and
```
bmunkholm: and and -> and a

realm/src/main/java/io/realm/internal/Table.java
```diff
@@ -390,6 +390,54 @@ public long addEmptyRow() {
...
+                    row.setLong(primaryKeyColumnIndex, pkValue);
+                    break;
+
+                default:
+                    throw new RealmException("Cannot check for duplicate rows for unsupported primary key type: " + type);
+            }
+
+            return rowIndex;
+
+        } else {
```
bmunkholm: Minor style comment: I would encourage to check for the error conditions first. It gives less indenting in general. It would also be consistent with the style used in the case statements.

bmunkholm:
:+1:

kneth:
:+1:

PR #996, opened 2015-03-28
Column indices are now cached pr. Realm

### LINKED ISSUE #846: Type mismatch after migration
In our migration we delete a field and afterwards we get in some cases a "type mismatch. Was BOOLEAN, expected INTEGER" for the field isUploaded

Our class looks like: 

```
public class PictureFile extends RealmObject {
    private int id;
    private String fileName;
    // private String description;
    private boolean isUploaded;

    // Getter and Setter for all fields
}
```

This happened to us in with multiple tables. We could see that the row indices caches in the static field columnIndices of the Realm class don't correspond to the actual indices of the table (checked with getTable for debugging).

For the migration we use the following code:

```
Table pictureFileTable = realm.getTable(PictureFile.class);
if (getIndexForProperty(pictureFileTable, "description") >= 0) {
    pictureFileTable.removeColumn(getIndexForProperty(pictureFileTable, "description"));
}
```

We have multiple migrations on other tables before, but that is the first column removal on this table.

How can we fix the migration to avoid this problem, because restarting doesn't fix the problem and app data has to be cleared. 

We also observed the inconsistency between the Realm.columnIndices field the actual file when we added a column in the migration. This is because the migration always adds it to the end, but we didn't but the new field after all the existing field. Therefore the indices were messed up. For example:

```
public static class Test extends RealmObject {
       private int oldField;
       private int oldField2;
       private int newField; // added by migration
       private int oldField3;
}
```

The indices for newField and oldField3 are messed up when the migration was executed, otherwise everything was fine.

Any help is appreciated

Greeting
Wicket

cmelchior:
Fixes #846 

Globally caching column indices could cause potentially issues if a Migration added fields in another order than the one in the actual java class. This error would only manifest if at least 2 realms is used and one is created using RealmProxy.initTable() and the other is created using a RealmMigration.


cmelchior:
@kneth @emanuelez @bmunkholm

bmunkholm:
Where does this leave #920? We now have two PR's for updating the indices.

cmelchior:
It is two separate things. #920 is about the internals of the ColumnIndices while this is about Realm caching it in a static variabel instead of a local one. I don't mind merging them tonight as long as they are both reviewed and approved.

realm/src/main/java/io/realm/Realm.java
```diff
@@ -177,7 +177,7 @@
     private static final long UNVERSIONED = -1;
 
     // Package protected to be reachable by proxy classes
-    static final Map<String, Map<String, Long>> columnIndices = new HashMap<String, Map<String, Long>>();
+    private final Map<String, Map<String, Long>> columnIndices = new HashMap<String, Map<String, Long>>();
```
kneth: The comment above isn't accurate.

kneth:
This is a nice catch! :+1: :+1: :+1:

emanuelez:
I think this PR should be on top of #920 and not on top of master.

cmelchior:
You are probably right, but basing a bug fix on a feature branch is a bad idea imo. Then I would rather take the merge conflict hit. I probably have some time today to merge #920 and fix any potential issues.

cmelchior:
#920 has been merged, so this is ready for a proper review

@emanuelez @kneth @bmunkholm

realm/src/main/java/io/realm/Realm.java
```diff
@@ -177,8 +177,7 @@
     private static final Set<Class<? extends RealmObject>> customSchema = new HashSet<Class<? extends RealmObject>>();
     private static final long UNVERSIONED = -1;
 
-    // Package protected to be reachable by proxy classes
-    static final ColumnIndices columnIndices = new ColumnIndices();
+    private final ColumnIndices columnIndices = new ColumnIndices();
```
emanuelez: Why not making this directly package protected and avoid the getter?

emanuelez:
I'm not sure about the scenario you describe...
Two realm instances: one created with initTable() and one with a migration.
How can the initTable run at all if a migration is found to be needed?

cmelchior:
It is possible if you create a completely new Realm file, then it will just run initTable(), while if you open an old file a Migration might be needed. The details are in the unit test (which might need further documentation :))

emanuelez:
:+1:

creativepsyco:
@cmelchior Does this also address this problem? I am on v0.80

```java
addColumns();
removeColumns();
List<Attachment> attachments = realm.where(Attachment.class).findAll();
attachments.get(0).setSomeField(SomeVal);

void addColumns(Table migrationTable, ColumnType columnType, String fieldName) {
        long index = MigrationUtils.getIndexForProperty(migrationTable, fieldName);
        if (index == -1) {
            // Does not exist, add it
            index = migrationTable.addColumn(columnType, fieldName);
        }
    }
}

void removeColumns() {
    long columnCount = migrationTable.getColumnCount();
    HashMap<String, Integer> deleteableCols = new HashMap<>();
    for (int i = 0; i < columnCount; i++) {
        String columnName = migrationTable.getColumnName(i);
        if (!columnMap.containsKey(columnName)) {
            deleteableCols.put(columnName, i);
        }
    }

    for (String column : deleteableCols.keySet()) {
        long index = MigrationUtils.getIndexForProperty(migrationTable, column);
        migrationTable.removeColumn(index);
    }
}
```
When I try to update my POJO using query after doing a migration, the columnIndex for a particular table does not seem to update. I nailed down the problem to be the difference in the column index of the desired property, using `table.getColumnIndex("imageWidth")` I am able to update successfully but with `row.setLong(Realm.columnIndices.get("Attachment").get("imageWidth"), (long) value);` it fails. 

```
Realm.columnIndices.get("Attachment").get("imageWidth"); // Returns 9 ==> Incorrect, throws exception when updating because Column Type is incorrect, it belongs to something else.
migrationTable.getColumnIndex("imageWidth"); // Returns 33 ==> Correct, I can update the underlying value using this column
```

Any Idea why? Or is this the normal behaviour? Should I just use `migrationTable` as a source of truth and not query after a migration?

creativepsyco:
Oh Apologies, I read the description of the ticket #846 afterwards. Please ignore.

cmelchior:
No problem :). As this PR has been merged to master you should be able to test it using our -SNAPSHOT build.

kneth:
Just to comment: third line of your snippet above should be `RealmResults<Attachment> attachments = ...`.

creativepsyco:
Oh I just checked with `SNAPSHOT` build to see if the problem went away, I think it still exists, with the piece of code above. This happens if I query immediately after migrations are done.

PR #1015, opened 2015-04-07
validateTable now also verifies that primary key and index

### LINKED ISSUE #1004: Primary key/Index not checked in validateTable
Primary keys and indexes are are not checked by RealmProxy.validateTable(), this means the following error scenario is possible:

1) Migrate a Realm: Add a new class with a primary key configured in the class, but forgetting to call table.setPrimaryKey(field) in the migration code.

2) Migration succeed, because validateTable doesn't check for primary key.

3) Using Realm.copyOrUpdate() will fail with a RealmException because it can't find the primary key column. 

Solution:
- Use ```long columnIndex =  Table.getPrimaryKey()``` and ```Table.hasIndex(columnIndex)``` in RealmProxy.validateTable() to verify that model class and table structures in Realm match.

cmelchior:
Fixes #1004 

@emanuelez @kneth @bmunkholm

### DIFF
--- a/realm-annotations-processor/src/main/java/io/realm/processor/RealmProxyClassGenerator.java
+++ b/realm-annotations-processor/src/main/java/io/realm/processor/RealmProxyClassGenerator.java
@@ -408,6 +408,21 @@ public class RealmProxyClassGenerator {
                 writer.emitStatement("throw new IllegalStateException(\"Invalid type '%s' for column '%s'\")",
                         fieldTypeSimpleName, fieldName);
                 writer.endControlFlow();
+
+                // Validate @PrimaryKey
+                if (field.equals(metadata.getPrimaryKey())) {
+                    writer.beginControlFlow("if (table.getPrimaryKey() != table.getColumnIndex(\"%s\"))", fieldName);
+                    writer.emitStatement("throw new IllegalStateException(\"Primary key not defined for field '%s'\")", fieldName);
+                    writer.endControlFlow();
+                }
+
+                // Validate @Index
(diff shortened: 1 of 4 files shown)


realm-annotations-processor/src/main/java/io/realm/processor/RealmProxyClassGenerator.java
```diff
@@ -408,6 +408,21 @@ private void emitValidateTableMethod(JavaWriter writer) throws IOException {
                 writer.emitStatement("throw new IllegalStateException(\"Invalid type '%s' for column '%s'\")",
                         fieldTypeSimpleName, fieldName);
                 writer.endControlFlow();
+
+                // Validate @PrimaryKey
+                if (field.equals(metadata.getPrimaryKey())) {
+                    writer.beginControlFlow("if (table.getPrimaryKey() != table.getColumnIndex(\"%s\"))", fieldName);
+                    writer.emitStatement("throw new IllegalStateException(\"Primary key not defined for field '%s'\")", fieldName);
```
kneth: Shouldn't it throw `RealmMigrationExceptionNeeded`?

cmelchior: Realm converts the IllegalStateException to a RealmMigrationNeededException. So these just throw the same IllegalStateException as the checks do in validateTable.

kneth: In that case :+1:

emanuelez:
:+1:

PR #1021, opened 2015-04-09
Realm.copyToRealm() now ignore referenced Realm objects

### LINKED ISSUE #1006: copyToRealm() tries to copy nested objects even if they are already in the Realm
Hey guys,

This piece of code leads to the exception ```Caused by: io.realm.exceptions.RealmException: Primary key constraint broken. Value already exists: brother```
```java
        realm.beginTransaction();
        UserRealm brother = realm.createObject(UserRealm.class);
        brother.setToken("brother");

        UserRealm user = new UserRealm();
        user.setToken("token");
        user.setBrother(brother);
        realm.copyToRealm(user);
        realm.commitTransaction();
        }
```

Changing ```copyToRealm``` to ```copyToRealmOrUpdate``` solves it. Using ```UserRealm user = realm.createObject(UserRealm.class)``` also solves it. It looks like even though ```brother``` is an object from the DB, Realm anyway treats it as a new object during the ```copyToRealm()```. It's either a bug, or a confusing design.

cmelchior:
Fixes #1006 

This PR fixes a bug in ```Realm.copyToRealm()``` that caused it to try to copy referenced RealmObjects that already existed in the Realm. This would either duplicate them or crash if that object contained a primary key.

@emanuelez @kneth @bmunkholm


realm/src/main/java/io/realm/Realm.java
```diff
@@ -1623,10 +1615,6 @@ void setVersion(long version) {
         }
     }
 
-    private <E extends RealmObject> boolean isObjectInRealm(E object) {
```
emanuelez: why not using this instead of replicating the code in many places? we could make it package protected instead of private.

kneth:
:+1:

realm/src/main/java/io/realm/Realm.java
cmelchior: It is a tradeoff. Now the method is only used from the proxy classes, so I would rather save the method call + package protected method in Realm vs. slightly longer code in a autogenerated class.

It is not something I feel strongly about, but we do have a lot of package protected methods which makes it hard to move more specialised API's to other packages (like .dynamic or .migration). In this case I didn't think it was worth it to keep a package protected method.

emanuelez:
:+1: assuming the only conflict is the changelog

PR #1022, opened 2015-04-09
Realm.copyOrUpdateFromJson shouldn't update unset properties

### LINKED ISSUE #933: createOrUpdateFromJson also updates fields not set
If you update a existing RealmObject using ```Realm.createOrUpdateFromJson()``` it correctly creates default values for fields not set, however if a existing RealmObject is found the expected behavior would be to only override the fields actually set, eg

```
public class Data extends RealmObject {
  @PrimaryKey
  private int id;
  private boolean bool;
  private String str;

  ...
}

// From empty database
realm.beginTransaction();
Data obj = realm.createOrUpdateFromJson(Data.class, "{ "id":1, "str": "foo" });
obj.setBool(true);
realm.commitTransaction();

realm.beginTransaction();
obj = realm.createOrUpdateFromJson(Data.class, "{ "id":1, "str": "foo" })
realm.commitTransaction();

obj.isBool() == false; // unexpected, it should maintain its value unless specifically set.
```

cmelchior:
Fixes #933 

```Realm.copyOrUpdateFromJson()``` shouldn't update properties if the property doesn't exist in the JSON. Right now it gets updated with the default value, which confuses users (and is also a bit weird).

### CODE REVIEW

**Comment by cmelchior** [comment]
@emanuelez @kneth @bmunkholm

File: realm/src/androidTest/java/io/realm/RealmJsonTest.java [hunk 1]
```diff
@@ -479,6 +493,37 @@ public void testCreateOrUpdateJsonObject() throws JSONException {
         assertEquals("bar", newObj.getColumnString());
     }
 
+    public void testCreateOrUpdateJsonObject_ignoreUnsetProperties() {
+        testRealm.beginTransaction();
+        AllTypesPrimaryKey obj = testRealm.createOrUpdateObjectFromJson(AllTypesPrimaryKey.class, "{ \"columnLong\":1, \"columnString\": \"foo\" }");
+        obj.setColumnBoolean(true);
```
kneth: If you swap line 498 and 499, you will check if `createOrUpdateObjectFromJson()` does not reset to default value (0/false). Or what am I missing?

cmelchior: That line is just for setting up the test data. It might become more readable if I add the boolean to JSON. The interesting test is the one further below, where you call createOrUpdate again with the same JSON (without the boolean).

kneth: In that case, setting the boolean to `true` is just fine as it is the non-default value.

realm-annotations-processor/src/main/java/io/realm/processor/RealmJSonImplGenerator.java
```diff
@@ -108,19 +108,19 @@ private void emitPopulateUsingJsonStream(JavaWriter writer) throws IOException {
...
+            writer.emitStatement(proxyStatement, proxyClasses.get(0));
             for (int i = 1; i < simpleModelClasses.size(); i++) {
-                writer.nextControlFlow("else if (classQualifiedName.equals(%s.class.getName()))", simpleModelClasses.get(i));
-                writer.emitStatement(proxyStatement, proxyClasses.get(i), simpleModelClasses.get(i));
+                writer.nextControlFlow("else if (clazz.equals(%s.class))", simpleModelClasses.get(i));
+                writer.emitStatement(proxyStatement, proxyClasses.get(i));
             }
             writer.nextControlFlow("else");
-            writer.emitStatement("throw new RealmException(%s)", EXCEPTION_MSG);
+            writer.emitStatement("throw new RealmException(\"Boom\")", EXCEPTION_MSG);
```
emanuelez: Seriously?

cmelchior: Ups, test code

realm/src/main/java/io/realm/Realm.java
```diff
@@ -813,17 +809,21 @@ private static void initializeRealm(Realm realm) {
         }
         checkHasPrimaryKey(clazz);
 
-        JsonReader reader = new JsonReader(new InputStreamReader(in, "UTF-8"));
+        // As we need the primary key value we have to first parse the entire input stream as in the general
```
emanuelez: Whoa... I understand the reasons, but doesn't this defeat the purpose of an InputStream?

cmelchior: Yes it does. Which is why there would be totally valid points for removing this method completely. Right now it is there to provide symmetry with the other methods, and I couldn't make up my mind if we should just remove it.

What do you think @kneth @bmunkholm ?

kneth: We shouldn't have a method just for symmetry reasons. If it is not useful, please remove it.

cmelchior: It is not completely useless. It does make client code easier (ie. they don't have to manually have to write a Scanner and convert to a JSONObject). But I think general consensus is to keep it for now until we decide in what direction to move our JSON support?

cmelchior:
Bump @emanuelez @kneth

realm-annotations-processor/src/main/java/io/realm/processor/RealmJSonImplGenerator.java
```diff
@@ -28,7 +28,7 @@
 
     private static final String REALM_PACKAGE_NAME = "io.realm";
     private static final String CLASS_NAME = "RealmJsonImpl";
-    private static final String EXCEPTION_MSG = "\"Could not find the generated proxy class for \" + classQualifiedName";
+    private static final String EXCEPTION_MSG = "\"Could not find the generated proxy class for \"";
```
emanuelez: Does it really need the escaped quotes?

realm-annotations-processor/src/main/java/io/realm/processor/RealmProxyClassGenerator.java
```diff
@@ -768,15 +768,39 @@ private void emitEqualsMethod(JavaWriter writer) throws IOException {
...
+        if (!metadata.hasPrimaryKey()) {
+            writer.emitStatement("%s obj = realm.createObject(%s.class)", className, className);
+        } else {
+            String pkType = Utils.isString(metadata.getPrimaryKey()) ? "String" : "Long";
+            writer
+                .emitStatement("%s obj = null", className)
+                .beginControlFlow("if (update)")
+                    .emitStatement("Table table = realm.getTable(%s.class)", className)
+                    .emitStatement("long pkColumnIndex = table.getPrimaryKey()")
+                   .beginControlFlow("if (!json.isNull(\"%s\"))", metadata.getPrimaryKey().getSimpleName())
```
emanuelez: indentation

realm-annotations-processor/src/main/java/io/realm/processor/RealmJSonImplGenerator.java
cmelchior: Not really. Then I just need to add them to where this is used. But that would probably fit the pattern we are using elsewhere.

realm/src/androidTest/java/io/realm/RealmJsonTest.java
```diff
@@ -452,6 +457,15 @@ public void testCreateOrUpdateObjectStream_noPrimaryKeyThrows() throws IOExcepti
         fail();
     }
 
+    public void testCreateOrUpdateObjectStream_invalidJSonThrows() throws IOException {
```
emanuelez: So this test (and others) passes if you get an IOException. Is that wise?

realm/src/main/java/io/realm/Realm.java
```diff
@@ -982,25 +971,19 @@ private static void initializeRealm(Realm realm) {
...
-    }
-
-    private <E extends RealmObject> E createStandaloneRealmObjectInstance(Class<E> clazz) {
-        try {
-            return clazz.newInstance();
-        } catch (InstantiationException e) {
-            throw new RealmException("Could not create an object of class: " + clazz, e);
-        } catch (IllegalAccessException e) {
-            throw new RealmException("Could not create an object of class: " + clazz, e);
+            scanner = new Scanner(in, "UTF-8").useDelimiter("\\A");
```
emanuelez: This looks a lot like the code above... any way it can be refactored?

realm/src/androidTest/java/io/realm/RealmJsonTest.java
cmelchior: Actually they don't. Junit3 will still report an error in that case (I just tested it). Perhaps you are thinking about Junit4 and its expected annotation?

kneth:
:+1:

emanuelez:
:+1:

dadino:
Is there a way to use Retrofit and Realm together and avoid the following scenario?
I have a list of objects where I only get titles of these objects from the server (and save them). Then when I get inside the detail page for that object I download all the infos and save them. BUT, if I refresh the list of objects, I will get a JSON with just titles and Retrofit will provide me with these almost empty objects. When I save them, all other infos on my database are wiped out.

snowpong:
@dadino One solution might be in the Retrofit onResponse(..) callback you serialize back to JSON (sic) from the Realm Model / POJO and then call [createOrUpdateObjectFromJson(..)](https://realm.io/docs/java/latest/api/io/realm/Realm.html#createOrUpdateObjectFromJson-java.lang.Class-org.json.JSONObject-) on it - as it claims: " If updating a RealmObject and a field is not found in the JSON object, that field will not be updated". Now, beware: Serializing back with GSON does not work out of the box - [read more here](https://realm.io/docs/java/latest/#gson) and look at the Gist.

Note: I haven't tried this solution myself yet.

PR #1028, opened 2015-04-14
Fix a potential NPE

emanuelez:


emanuelez:
@cmelchior @kneth @bmunkholm

cmelchior:
Good catch, but if we are throwing an error on ```new RealmList(null)``` (which makes sense) should we not also do it on ```new RealmList(null, null)```? 

The later will get caught if you are trying to use ```Realm.copyToRealm()```, but not before?

cmelchior:
Also, is worth a changelog entry?

kneth:
:+1:

bmunkholm:
: No additional comments.

PR #1029, opened 2015-04-14
Fix wrong String comparison

emanuelez:


emanuelez:
@cmelchior @kneth @bmunkholm

cmelchior:
:+1:

kneth:
:+1:

But we might other comparisons like this one.

emanuelez:
Not according to findbugs

cmelchior:
A quick search-in-path says no to that one.

PR #955, opened 2015-03-12
Made Realm.compactRealmFile() more failure resilient.

cmelchior:
- [x] Awaits a new core release with fix for compact() on empty Realms.
Fixes #953 

 ```Realm.compactRealmFile()``` now uses the methods provided by core which are more failure resilient.

@emanuelez @kneth @bmunkholm

PR #1030, opened 2015-04-14
Updating to Realm Core 0.89.0

kneth:
Realm Core 0.89.0 is a bug fix and stability release, and has changed the namespace to `realm`. Moreover, to complete the renaming of the namespace, a number of Java files have been updated too.

The `CHANGELOG.md` is on purpose not updated. 

@cmelchior @emanuelez @bmunkholm

### DIFF
--- a/realm-jni/src/io_realm_internal_Version.cpp
+++ b/realm-jni/src/io_realm_internal_Version.cpp
@@ -19,16 +19,16 @@
 
 #include "util.hpp"
 #include "io_realm_internal_Version.h"
-#include <tightdb/version.hpp>
+#include <realm/version.hpp>
 
-static int tightdb_jni_version = 23;
+static int realm_jni_version = 23;
 
 
-using namespace tightdb;
+using namespace realm;
 
 JNIEXPORT jint JNICALL Java_io_realm_internal_Version_nativeGetAPIVersion(JNIEnv*, jclass)
 {
-    return tightdb_jni_version;
+    return realm_jni_version;
 }
 
 JNIEXPORT jstring JNICALL Java_io_realm_internal_Version_nativeGetVersion(JNIEnv *env, jclass)
--- a/realm-jni/src/util.hpp
+++ b/realm-jni/src/util.hpp
@@ -444,7 +444,7 @@ std::string concat_stringdata(const char *message, StringData data);
 // UTF-8 where U+0000 is stored as 0xC0 0x80 instead of 0x00 and
 // where a character in the range U+10000 to U+10FFFF is stored as
 // two consecutive UTF-8 encodings of the corresponding UTF-16
-// surrogate pair. Because TightDB uses proper UTF-8, we need to
+// surrogate pair. Because Realm uses proper UTF-8, we need to
 // do the transcoding ourselves.
 //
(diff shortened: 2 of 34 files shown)


realm-jni/src/io_realm_internal_Version.cpp
```diff
@@ -19,16 +19,16 @@
 
 #include "util.hpp"
 #include "io_realm_internal_Version.h"
-#include <tightdb/version.hpp>
+#include <realm/version.hpp>
 
-static int tightdb_jni_version = 23;
+static int realm_jni_version = 23;
```
cmelchior: Not related to naming, but is this actually used for anything?

realm-jni/src/util.hpp
```diff
@@ -19,6 +19,7 @@
 
 #include <string>
 #include <sstream>
+#include <memory>
```
cmelchior: Why is this suddenly needed?

realm/src/main/java/io/realm/internal/RealmCore.java
```diff
@@ -117,7 +117,7 @@ public void run(){
         });
     }
 
-    private static void initTightDB() {
+    private static void initRealm() {
```
cmelchior: Perhaps rename to either ```initCore()``` or just ```init()```. Calling it ```initRealm()``` is slightly confusing.

cmelchior:
Apart from comments. Farewell TightDB, it has been fun, see you on the dark side of Git.
:+1:

realm-jni/src/util.hpp
kneth: This is the `std::unique_ptr`.

realm-jni/src/io_realm_internal_Version.cpp
kneth: Probably not - found by `find`/`sed`.

realm-jni/src/util.hpp
emanuelez: I think it's because now we use the unique pointers from the standard library instead of our own

realm/src/main/java/io/realm/internal/RealmCore.java
emanuelez: And redundant. `init()` is better IMO

kneth: Since the class is `RealmCore`, the method should be named `init`: the class will be `RealmCore.init()` which is not confusing.

emanuelez:
:+1:

realm-jni/src/io_realm_internal_Version.cpp
bmunkholm: It was used when linking dynamically to ensure the API was matching. Not used in Android, but may be in Java at a point. Not a problem to just keep it.

bmunkholm:
Why not update the changelog @kneth ? There are changes from core that needs to be noticed.

kneth:
@bmunkholm #955 will take advantage of the major bug fix in the new core release (and update the change log).

bmunkholm:
Makes no sense :-) The only rule that works in my mind is that the PR that actually updates the version of the core, also updates the changes that introduces.

PR #1054, opened 2015-04-20
Realm IDs are now its absolute path instead of its hashcode

### LINKED ISSUE #1040: expecting absolutePath.hashCode() not to collide each other
'absolutePath.hashCode()' is used as key of some HashMaps(realmsCache, referenceCount, etc.).

https://github.com/realm/realm-java/blob/746083df06e9a7a581fb3e769960f30a96b4ff01/realm/src/main/java/io/realm/Realm.java#L495

If hash code of absolutePath collides with each other, wrong Realm instance is returned on Realm.getInstance().

cmelchior:
Fixes #1040 

@emanuelez @bmunkholm @kneth

### DIFF
--- a/realm/src/main/java/io/realm/Realm.java
+++ b/realm/src/main/java/io/realm/Realm.java
@@ -124,27 +124,26 @@ public final class Realm implements Closeable {
 
     private static final String TAG = "REALM";
     private static final String TABLE_PREFIX = "class_";
-    protected static final ThreadLocal<Map<Integer, Realm>> realmsCache = new ThreadLocal<Map<Integer, Realm>>() {
+    protected static final ThreadLocal<Map<String, Realm>> realmsCache = new ThreadLocal<Map<String, Realm>>() {
         @SuppressLint("UseSparseArrays")
         @Override
-        protected Map<Integer, Realm> initialValue() {
-            return new HashMap<Integer, Realm>(); // On Android we could use SparseArray<Realm> which is faster,
+        protected Map<String, Realm> initialValue() {
+            return new HashMap<String, Realm>(); // On Android we could use SparseArray<Realm> which is faster,
                                                   // but incompatible with Java
         }
     };
-    private static final ThreadLocal<Map<Integer, Integer>> referenceCount
-            = new ThreadLocal<Map<Integer,Integer>>() {
+    private static final ThreadLocal<Map<String, Integer>> referenceCount = new ThreadLocal<Map<String,Integer>>() {
         @SuppressLint("UseSparseArrays")
         @Override
-        protected Map<Integer, Integer> initialValue() {
-            return new HashMap<Integer, Integer>();
+        protected Map<String, Integer> initialValue() {
+            return new HashMap<String, Integer>();
         }
     };
     private static final int REALM_CHANGED = 14930352; // Hopefully it won't clash with other message IDs.
-    protected static final Map<Handler, Integer> handlers = new ConcurrentHashMap<Handler, Integer>();
+    protected static final Map<Handler, String> handlers = new ConcurrentHashMap<Handler, String>();
 
     // Maps ids to a boolean set to true if the Realm is open. This is only needed by deleteRealmFile
-    private static final Map<Integer, AtomicInteger> openRealms = new ConcurrentHashMap<Integer, AtomicInteger>();
+    private static final Map<String, AtomicInteger> openRealms = new ConcurrentHashMap<String, AtomicInteger>();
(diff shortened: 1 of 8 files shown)


realm/src/main/java/io/realm/Realm.java
```diff
@@ -158,7 +157,7 @@
     private Handler handler;
 
     private final byte[] key;
-    private final int id;
+    private final String id;
```
emanuelez: I guess this can be renamed to `absolutePath`

realm/src/main/java/io/realm/Realm.java
```diff
@@ -1680,7 +1679,7 @@ public static synchronized boolean deleteRealmFile(File realmFile) {
         File realmFolder = realmFile.getParentFile();
         String fileName = realmFile.getName();
 
-        int realmId = realmFile.getAbsolutePath().hashCode();
+        String realmId = realmFile.getAbsolutePath();
```
emanuelez: rename variable

realm/src/main/java/io/realm/Realm.java
```diff
@@ -1533,14 +1532,14 @@ public void clear(Class<?> classSpec) {
         getTable(classSpec).clear();
     }
 
-    int getId() {
+    String getId() {
```
emanuelez: rename

cmelchior:
We already have a path variabel we can use instead. I was just weary to overload the meaning of the name.

zaki50:
if we use path as ID, it's better to use canonical path instead of absolute path.
is it too strict?

emanuelez:
:+1: for @zaki50 proposal

cmelchior:
```File.getCanonicalPath()``` is 1300x times slower than calling ```File.getAbsolutePath()``` on my Nexus 5. Is there a use case where using getAbsolutePath would result in the same ID for two different Realms? 

ID's are only local and used on runtime. If we cached them on disk, I would probably agree, but right now I don't think we need the extra security or is there something I am missing?

**Comment by emanuelez** [comment]
The problem is the opposite. Two absolute paths can point to the same file.

**Comment by cmelchior** [comment]
True, but wouldn't you really have to go out of the way to make that happen, eg. 
```
File f1 = new File(getContext().getFilesDir(), "new_dir/../default.realm");
File f2 = new File(getContext().getFilesDir(), "default.realm");
assertFalse(f1.getAbsolutePath(), f2.getAbsolutePath())
assertTrue(f1.getCanonicalPath(), f2.getCanonicalPath())
```
I am not against it per se, but at least on Android I doubt this will be an issue, so introducing a slowdown for a really small corner case?

**Comment by emanuelez** [comment]
Maybe we can check if the string contains `..` and throw if it does. I kinda like the idea of making this hard to misuse

**Comment by cmelchior** [comment]
Throwing seems kinda harsh though. We could also add a check for ".." and then fallback to canonical path if needed.

```
String path = file.getAbsolutePath();
if (path.contains("..")) {
   path = file.getCanonicalPath();
}
```

This introduces a slowdown of about 10x for paths not containing "..". Neither of these solutions work with symlinks though, but we shouldn't encounter them on Android at least (we probably will on the server side)

emanuelez:
You last solution looks fine. We can document the performance drop in the Javadoc.

Only Java8 has proper support for symlinks so there's that

zaki50:
there are some other corner cases.

        final String path1 = new File(getActivity().getFilesDir(), "foo.realm").getAbsolutePath();
        final String path2 = new File(getActivity().getFilesDir(), "./foo.realm").getAbsolutePath();

        final String path2_1 = new File(getActivity().getFilesDir(), "foo/bar.realm").getAbsolutePath();
        final String path2_2 = new File(getActivity().getFilesDir(), "foo/./bar.realm").getAbsolutePath();

"./" or simply "/" is better for fallback condition, I think.

cmelchior:
Good point. Thinking about this a bit more, then ti would probably be cleaner just to use canonicalPath() and then take the performance hit. Once we introduce RealmConfiguration #929 , this becomes less of an issue anyway since we are only going to calculate it once.

cmelchior:
Note, I deprecated ```getPath()``` in favour of ```getCanonicalPath()``` but I am not totally convinced it is a good idea, since the later is probably way less know. Maybe we should just document that getPath() returns the canonical path instead?

zaki50:
I think getPath() is enough because we have just single getter of path.

cmelchior:
Bump @emanuelez

emanuelez:
:+1:

kneth:
:+1:

PR #1055, opened 2015-04-20
Linebreaks in version.txt no longer breaks build

cmelchior:
We can now edit the version.txt from GitHub. Yay

@emanuelez @kneth

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -17,7 +17,7 @@ allprojects {
   apply plugin: 'android-sdk-manager'
 
   group = 'io.realm'
-  version = new File("${rootDir}/version.txt").text
+  version = new File("${rootDir}/version.txt").text.replaceAll("\\r|\\n","")
 
   repositories {
     jcenter()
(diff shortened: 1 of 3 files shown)


build.gradle
```diff
@@ -17,7 +17,7 @@ allprojects {
   apply plugin: 'android-sdk-manager'
 
   group = 'io.realm'
-  version = new File("${rootDir}/version.txt").text
+  version = new File("${rootDir}/version.txt").text.replaceAll("\\r|\\n","")
```
emanuelez: How about using `trim()`?

cmelchior: I tried initially and it failed. At first I thought it was perhaps trim() wasn't available: http://www.groovy-lang.org/gdk.html, but after remembering to use it in all 3 projects, it does actually work, so much better than custom regexp.

kneth:
:+1:

emanuelez:
:+1:

PR #1056, opened 2015-04-21
Throw exception if using copyToRealmOrUpdate with a null primary key value

### LINKED ISSUE #1051: copyToRealmOrUpdate SIGSEGV
Hi,

Following an update in my code, I randomly found a native crash (SIGSEGV 0x00000008) when using
`copyToRealmOrUpdate` and an object with a declared `@PrimaryKey` which was not set before copy.
Using `Realm 0.80.1`

To reproduce with a testCase:

```java
public class PrimaryKeyTest extends RealmObject {

    @PrimaryKey
    private String key;

    public String getKey() {
        return key;
    }

    public void setKey(String key) {
        this.key = key;
    }
}
```
```java
public void testPrimKeyMissing() {
        testRealm.beginTransaction();
        try {
            testRealm.copyToRealmOrUpdate(new PrimaryKeyTest());
        } catch (Exception e) {
            // OK
            testRealm.cancelTransaction();
        }
}
```

==> 'Instrumentation run failed due to 'Native crash''

cmelchior:
Fixes #1051 

We now throw a proper exception instead of crashing core.

@emanuelez @bmunkholm @kneth

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,3 +1,6 @@
+0.81
+ * Trying to using Realm.copyToRealmOrUpdate() with a an object with a null primary key now throws a proper exception.
+
 0.80.1
  * Realm.createOrUpdateWithJson() no longer resets fields to their default value if they are not found in the JSON input.
  * Realm.compactRealmFile() now uses Realm Core's compact() method which is more failure resilient.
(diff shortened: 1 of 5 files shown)


kneth:
:+1:

changelog.txt
```diff
@@ -1,3 +1,6 @@
+0.81
+ * Trying to using Realm.copyToRealmOrUpdate() with a an object with a null primary key now throws a proper exception.
```
emanuelez: there's an extra a

emanuelez:
:+1: apart from comment

PR #1065, opened 2015-04-23
RealmMigrationNeededException can now return the path to the Realm

### LINKED ISSUE #1060: RealmMigrationNeededException should have getAbsolutePath()
I'd like to write schema check code as follows.

        try {
            Realm.getInstance(context, "foo.realm").close();
        } catch (RealmMigrationNeededException e) {
            Realm.migrateRealmAtPath(e.getAbsolutePath(), new MyRealmMigration());
        }

due to lacking of RealmMigrationNeededException#getAbsolutePath(), we must write 

    Realm.migrateRealmAtPath(new File(context.getFilesDir(), "foo.realm").getAbsolutePath(),
                             new MyRealmMigration());

cmelchior:
Fixes #1060 

@emanuelez @kneth @zaki50

### DIFF
--- a/realm/src/main/java/io/realm/internal/SharedGroup.java
+++ b/realm/src/main/java/io/realm/internal/SharedGroup.java
@@ -213,8 +213,6 @@ public class SharedGroup implements Closeable {
         return nativePtr == 0;
     }
 
-    static native String nativeGetDefaultReplicationDatabaseFileName();
-
     public boolean hasChanged() {
         return nativeHasChanged(nativePtr);
     }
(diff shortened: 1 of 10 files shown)


zaki50:
:+1:

emanuelez:
A unit test would be nice, otherwise :+1:

realm/src/main/java/io/realm/internal/SharedGroup.java
```diff
@@ -233,6 +231,18 @@ public boolean compact() {
         return nativeCompact(nativePtr);
     }
 
+
+    /**
+     * Returns the absolute path to the file backing this SharedGroup.
+     *
+     * @return Absolute path to the Realm file.
+     */
+    public String getPath() {
```
kneth: Since the Java documentation (https://docs.oracle.com/javase/7/docs/api/java/io/File.html) discuss `getPath()` and `getAbsolutePath()` as two different methods, I think this method should be called `getAbsolutePath()`.

kneth:
Besides my comment: :+1:

cmelchior:
Hmm, I am not sure I agree. Right now the method is called the same on Realm as on this exception. You are right that Java has 3 different ways for this (getPath/getAbsolutePath/getCanonicalPath), but if we don't have that distinction getPath is the shortest and most readable?

PR #1066, opened 2015-04-23
Change StackOverflow tag from realm to realm.io

erichkleung:
Apparently the realm tag is for "a collection of usernames and passwords that identify valid users..." 

StackOverflow question got edited and re-tagged. 

You may want to change it on http://realm.io/contact/ as well


timanglade:
Thanks but our name is Realm not realm.io. If we were to change domain names, that tag would have no meaning whatsoever and would be confusing to our users. Thanks for taking the time to submit a PR though!

erichkleung:
Hm, you may want to consider taking over that SO tag so that the tag doesn't get removed by people submitting realm SO posts in the future? (My "realm" tag was removed from my question)

timanglade:
Good point! I'm talking to some SO friends about best steps here, I know
right now it's a bit confusing.

PR #1073, opened 2015-04-28
Primary key table should match Cocoa primary key table

### LINKED ISSUE #1059: Unable to use a realm database file created with realm 0.80.0 with 0.80.1
I have a realm database file created using realm 0.80.0 that works fine under that version.  After changing gradle to pull in 0.80.1 I get an exception thrown that a migration is needed.  Using the following I create a migration:

		try {
			realm = Realm.getInstance(context);
		} catch (RealmMigrationNeededException ex) {
			Log.i(TAG, "migration needed");
			Realm.migrateRealmAtPath(new File(context.getFilesDir(), "default.realm").getAbsolutePath(), new Migration());
			Log.i(TAG, "migration complete");
		}

the next error occurred was that I need a primary key which was already defined in the models but I made the migration set the primary key using the following on all of the database tables

			Table personTable = realm.getTable(Person.class);
			personTable.setPrimaryKey("id");

where the id attribute for the Person class is a long;

After the migration has completed and I attempt to open the realm I get  an error

    Expected columnType 0, but got 2.
    jni: ThrowingException 3, ColumnType invalid., .
    Exception has been throw: Illegal Argument: ColumnType invalid.

It would be helpful if the column type error would actually state what table the error was in to help with debugging. However, the real problem is that I cannot use the realm file.  All of the tables "id" field are defined as longs so I am unable to determine why the ColumnType.STRING (2) factors in anywhere.

cmelchior:
Fixes #1059 

This silently upgrades any Realms that are wrong, but adds a little extra overhead first time we check for primary keys (they are cached after that). We should consider for how long we should support this code snippet, as it is both an overhead plus code we have to maintain.

@emanuelez @kneth


kneth:
Let's keep it for 3 releases and then remove it again.

bmunkholm:
update changelog

cmelchior:
And it is green: @emanuelez @kneth

realm-jni/src/io_realm_internal_table.cpp
```diff
@@ -1389,8 +1389,9 @@ JNIEXPORT jlong JNICALL Java_io_realm_internal_Table_createNative(JNIEnv *env, j
 
 // Checks if the primary key column contains any duplicate values, making it ineligible as a
 // primary key.
-bool check_valid_primary_key_column(JNIEnv* env, Table* table, size_t column_index) // throws
+bool check_valid_primary_key_column(JNIEnv* env, Table* table, StringData column_name) // throws
 {
+    int column_index = table->get_column_index(column_name);
```
kneth: The returned type is `size_t`.

realm-jni/src/io_realm_internal_table.cpp
```diff
@@ -1484,3 +1484,33 @@ JNIEXPORT jlong JNICALL Java_io_realm_internal_Table_nativeSetPrimaryKey(
...
+    (JNIEnv* env, jobject, jlong groupNativePtr, jlong privateKeyTableNativePtr)
+{
+    Group* group = G(groupNativePtr);
+    Table* pk_table = TBL(privateKeyTableNativePtr);
+    if (pk_table->get_column_type(io_realm_internal_Table_PRIMARY_KEY_FIELD_COLUMN_INDEX) == type_Int) {
+        StringData tmp_col_name = StringData("tmp_field_name");
+        std::size_t tmp_col_ndx = pk_table->add_column(DataType(type_String), tmp_col_name);
+
+        // Create tmp string column with field name instead of column index
+        int rows = pk_table->size();
```
kneth: The returned type is `size_t`.

realm-jni/src/io_realm_internal_table.cpp
```diff
@@ -1484,3 +1484,33 @@ JNIEXPORT jlong JNICALL Java_io_realm_internal_Table_nativeSetPrimaryKey(
...
+        for (int row_ndx = 0; row_ndx < rows; row_ndx++) {
+            StringData table_name = pk_table->get_string(io_realm_internal_Table_PRIMARY_KEY_CLASS_COLUMN_INDEX, row_ndx);
+            int64_t col_ndx = pk_table->get_int(io_realm_internal_Table_PRIMARY_KEY_FIELD_COLUMN_INDEX, row_ndx);
```
kneth: You have to be careful here: an integer column is always `int64_t` but column index is `size_t` (which is an `unsigned long`). They differ on a 32 bit architecture.

emanuelez:
Should this be double checked by somebody from core and/or cocoa?

cmelchior:
Good idea: @tgoyne ?

tgoyne:
Looks correct to me.

realm-jni/src/io_realm_internal_table.cpp
```diff
@@ -1484,3 +1484,33 @@ JNIEXPORT jlong JNICALL Java_io_realm_internal_Table_nativeSetPrimaryKey(
     } CATCH_STD()
     return 0;
 }
+
+// Fixes interop issue with Cocoa Realm where the Primary Key table had different types.
+// This effects:
```
bdash: affects?

emanuelez:
:+1:

realm-jni/src/io_realm_internal_table.cpp
```diff
@@ -1389,9 +1389,10 @@ JNIEXPORT jlong JNICALL Java_io_realm_internal_Table_createNative(JNIEnv *env, j
 
 // Checks if the primary key column contains any duplicate values, making it ineligible as a
 // primary key.
-bool check_valid_primary_key_column(JNIEnv* env, Table* table, size_t column_index) // throws
+bool check_valid_primary_key_column(JNIEnv* env, Table* table, StringData column_name) // throws
 {
-    int column_type = table->get_column_type(column_index);
+    size_t column_index = table->get_column_index(column_name);
+    DataType column_type = table->get_column_type(column_index);
```
bmunkholm: You have to check the index returned is valid before using it, and throwing if it is invalid.

realm-jni/src/io_realm_internal_table.cpp
```diff
@@ -1448,33 +1449,32 @@ JNIEXPORT jlong JNICALL Java_io_realm_internal_Table_nativeSetPrimaryKey(
...
+            // No primary key set => Remove any previous set keys
             if (row_index != realm::not_found) {
                 pk_table->remove(row_index);
             }
             return jlong(io_realm_internal_Table_NO_PRIMARY_KEY);
         }
         else {
-            JStringAccessor columnName2(env, columnName);
-            size_t primary_key_column_index = table->get_column_index(columnName2);
+            JStringAccessor NewPrimaryKeyColumn(env, columnName);
```
bmunkholm: Respect case convention in C++: lowercase with underscores: not "NewPrimaryKeyColumn"

realm/src/main/java/io/realm/internal/Table.java
```diff
@@ -677,7 +677,8 @@ public long getPrimaryKey() {
             if (pkTable == null) return NO_PRIMARY_KEY; // Free table = No primary key
             long rowIndex = pkTable.findFirstString(PRIMARY_KEY_CLASS_COLUMN_INDEX, getName());
             if (rowIndex != NO_MATCH) {
-                cachedPrimaryKeyColumnIndex = pkTable.getRow(rowIndex).getLong(PRIMARY_KEY_FIELD_COLUMN_INDEX);
+                String pkColumnName = pkTable.getRow(rowIndex).getString(PRIMARY_KEY_FIELD_COLUMN_INDEX);
```
bmunkholm: respect variable naming convention here - and everywhere in this method.
Also newline for return statements.

realm/src/androidTest/java/io/realm/internal/JNITransactions.java
```diff
@@ -376,4 +378,13 @@ public void testAddEmptyRowWithPrimaryKeyLong() {
         assertEquals(1, t.size());
         assertEquals(42, t.getRow(rowIndex).getLong(0));
     }
+
+    public void testPrimaryKeyTableMigration() throws IOException {
```
bmunkholm: Is this test sufficient? Look at the migration code and consider how you ensure that changing a single line of code there would get checked in this test.
- I can't see that any of the row copying is being tested.
- What if the Primary key column was at various different indexes (front, middle, end)?

realm/src/main/java/io/realm/internal/Table.java
cmelchior: What convention are you referring to? It looks like a normal Java variable name to me?

bmunkholm: ahh, sorry - ignore. Just came from the cpp file :-)

realm/src/androidTest/java/io/realm/internal/JNITransactions.java
cmelchior: Just being able to ask for the primary key without crashing actually tests most of the logic, but you are right that testing the primary key hasn't changed wasn't among them. It has been added.

changelog.txt
```diff
@@ -1,4 +1,7 @@
 0.81
+ * Trying to using Realm.copyToRealmOrUpdate() with an object with a null primary key now throws a proper exception.
```
kneth: using -> call

realm-jni/src/io_realm_internal_table.cpp
```diff
@@ -1405,9 +1405,15 @@ JNIEXPORT jlong JNICALL Java_io_realm_internal_Table_createNative(JNIEnv *env, j
...
 // Checks if the primary key column contains any duplicate values, making it ineligible as a
 // primary key.
-bool check_valid_primary_key_column(JNIEnv* env, Table* table, size_t column_index) // throws
+bool check_valid_primary_key_column(JNIEnv* env, Table* table, StringData column_name) // throws
 {
-    int column_type = table->get_column_type(column_index);
+    size_t column_index = table->get_column_index(column_name);
+    if (column_index == realm::not_found) {
+        std::ostringstream error_msg;
+        error_msg << table->get_name() << " does not contain the field \"" << column_name << "\"";
```
kneth: The `class_` is prefixed to the table name at this point. But is probably ok.

realm-jni/src/io_realm_internal_table.cpp
```diff
@@ -1464,33 +1470,32 @@ JNIEXPORT jlong JNICALL Java_io_realm_internal_Table_nativeSetPrimaryKey(
         const char* table_name = table->get_name().data();
         size_t row_index = pk_table->find_first_string(io_realm_internal_Table_PRIMARY_KEY_CLASS_COLUMN_INDEX, table_name);
 
-        // I
         if (columnName == NULL || env->GetStringLength(columnName) == 0) {
-            // No primary key set. Remove any previous set keys
+            // No primary key set => Remove any previous set keys
```
kneth: The comment should be "No primary key provided => remove previous set keys".

realm-jni/src/io_realm_internal_table.cpp
```diff
@@ -1500,3 +1505,33 @@ JNIEXPORT jlong JNICALL Java_io_realm_internal_Table_nativeSetPrimaryKey(
...
+// - All Realms created by Realm-Android 0.80.1 and below
+// See https://github.com/realm/realm-java/issues/1059
+JNIEXPORT void JNICALL Java_io_realm_internal_Table_nativeMigratePrimaryKeyTableIfNeeded
+    (JNIEnv* env, jobject, jlong groupNativePtr, jlong privateKeyTableNativePtr)
+{
+    Group* group = G(groupNativePtr);
+    Table* pk_table = TBL(privateKeyTableNativePtr);
+    if (pk_table->get_column_type(io_realm_internal_Table_PRIMARY_KEY_FIELD_COLUMN_INDEX) == type_Int) {
+        StringData tmp_col_name = StringData("tmp_field_name");
+        std::size_t tmp_col_ndx = pk_table->add_column(DataType(type_String), tmp_col_name);
```
kneth: You don't need the `std` name space.

realm-jni/src/io_realm_internal_table.cpp
```diff
@@ -1500,3 +1505,33 @@ JNIEXPORT jlong JNICALL Java_io_realm_internal_Table_nativeSetPrimaryKey(
...
+
+        // Create tmp string column with field name instead of column index
+        size_t number_of_rows = pk_table->size();
+        for (int row_ndx = 0; row_ndx < number_of_rows; row_ndx++) {
```
kneth: `row_ndx` should be of type `size_t`.

realm-jni/src/io_realm_internal_table.cpp
```diff
@@ -1500,3 +1505,33 @@ JNIEXPORT jlong JNICALL Java_io_realm_internal_Table_nativeSetPrimaryKey(
...
+            StringData table_name = pk_table->get_string(io_realm_internal_Table_PRIMARY_KEY_CLASS_COLUMN_INDEX, row_ndx);
+            size_t col_ndx = static_cast<size_t>(pk_table->get_int(io_realm_internal_Table_PRIMARY_KEY_FIELD_COLUMN_INDEX, row_ndx));
+            StringData col_name = group->get_table(table_name)->get_column_name(col_ndx);
+            pk_table->set_string(tmp_col_ndx, row_ndx, col_name);
+        }
+
+        // Delete old int column, and rename tmp column to same name
+        // Index should stay the same as we control creation of table
```
bmunkholm: What is meant with this?

realm-jni/src/io_realm_internal_table.cpp
cmelchior: I think the core team is moving to a policy of of not importing the std namespace in the header. Should we do the same?

bmunkholm: Not for size_t anyway. But in general I prefer not to.

realm-jni/src/io_realm_internal_table.cpp
```diff
@@ -1500,3 +1505,33 @@ JNIEXPORT jlong JNICALL Java_io_realm_internal_Table_nativeSetPrimaryKey(
...
+        size_t number_of_rows = pk_table->size();
+        for (size_t row_ndx = 0; row_ndx < number_of_rows; row_ndx++) {
+            StringData table_name = pk_table->get_string(io_realm_internal_Table_PRIMARY_KEY_CLASS_COLUMN_INDEX, row_ndx);
+            size_t col_ndx = static_cast<size_t>(pk_table->get_int(io_realm_internal_Table_PRIMARY_KEY_FIELD_COLUMN_INDEX, row_ndx));
+            StringData col_name = group->get_table(table_name)->get_column_name(col_ndx);
+            pk_table->set_string(tmp_col_ndx, row_ndx, col_name);
+        }
+
+        // Delete old int column, and rename tmp column to same name
+        // The column index for the renamed column will then be the same as the deleted old column
```
kneth: This is not generally true but you know the exact layout of the table (two columns: string, integer). It might be worth describing the old layout and the new layout (string, string) in a comment at the top of the function.

cmelchior: Yup. That was what my old comment was trying to convey (unsuccessfully). I have added a description about the formats to the top of the method.

realm/src/androidTest/java/io/realm/internal/JNITransactions.java
kneth: It is sufficient to test with two different old primary key layouts: 0 primary keys and 1 primary key (triggering the loop).

PR #1038, opened 2015-04-16
Adding Table.removeSearchIndex() to remove search index, and renaming other index methods

kneth:
No updates to change log since it's an internal method.

@emanuelez @cmelchior

PR #1075, opened 2015-04-29
Updating to Realm Core 0.89.1

kneth:
This is required for #1038.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,5 +1,7 @@
 0.81
  * Fixed bug where creating a Realm instance with a hashcode collision no longer returned the wrong Realm instance.
+ * Updated Realm Core to version 0.89.1
+   - Fixed bug in "index rebuilding" (would delete the wrong column, causing crash).
 
 0.80.1
  * Realm.createOrUpdateWithJson() no longer resets fields to their default value if they are not found in the JSON input.
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,7 +1,6 @@
 0.81
  * Fixed bug where creating a Realm instance with a hashcode collision no longer returned the wrong Realm instance.
- * Updated Realm Core to version 0.89.1
-   - Fixed bug in "index rebuilding" (would delete the wrong column, causing crash).
+ * Updated Realm Core to version 0.89.1.
 
 0.80.1
  * Realm.createOrUpdateWithJson() no longer resets fields to their default value if they are not found in the JSON input.
--- a/realm-jni/src/util.hpp
+++ b/realm-jni/src/util.hpp
@@ -58,7 +58,7 @@ JNIEXPORT jint JNICALL JNI_OnLoad(JavaVM *vm, void *reserved);
 
 #define CATCH_FILE(fileName) \
     catch (InvalidDatabase&) { \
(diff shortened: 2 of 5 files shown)


changelog.txt
```diff
@@ -1,5 +1,7 @@
 0.81
  * Fixed bug where creating a Realm instance with a hashcode collision no longer returned the wrong Realm instance.
+ * Updated Realm Core to version 0.89.1
+   - Fixed bug in "index rebuilding" (would delete the wrong column, causing crash).
```
cmelchior: Is this needed since we have not exposed that functionality on Java yet?

realm-jni/src/util.hpp
```diff
@@ -58,7 +58,7 @@ JNIEXPORT jint JNICALL JNI_OnLoad(JavaVM *vm, void *reserved);
 
 #define CATCH_FILE(fileName) \
     catch (InvalidDatabase&) { \
-        ThrowException(env, IllegalArgument, "Invalid Group file format."); \
+        ThrowException(env, IllegalArgument, "Invalid format of Realm file."); \
```
cmelchior: Does "format" makes any sense to users, it is out of their hands anyway. Wouldn't "Invalid Realm file" or "File is not a valid Realm file" be better?

realm/src/androidTest/java/io/realm/RealmTest.java
```diff
@@ -190,6 +190,8 @@ public void testGetInstanceClearsCacheWhenFailed() {
 
     public void testInstanceIdForHashCollision() {
         // Ea.hashCode() == FB.hashCode()
+        Realm.deleteRealmFile(getContext(), "Ea");
```
cmelchior: It's a bit scary this is needed. I agree it makes a better test case as state is reset, I am just wondering what bug this exposed?

changelog.txt
bmunkholm: No that can be removed. The bug has not been exposed in Java.

realm-jni/src/util.hpp
emanuelez: +1 for "Invalid Realm file"

realm/src/androidTest/java/io/realm/RealmTest.java
emanuelez: Also, how relevant is this with the PR?

kneth: No, really. I was using a custom compiled prerelease of Realm Core in my previous run, and forward compatibility shouldn't be a concern.

changelog.txt
kneth: We should mention updates of Realm Core, right? And when #1038 is merged, this bug fix is important.

realm-jni/src/util.hpp
kneth: You are right: "File is not a valid Realm file." is better.

changelog.txt
cmelchior: I like that we write updates to core as it might effect if users want to upgrade, but from a realm-android point of view there hasn't been any bug, as we have not exposed the functionality yet, and they cannot be exposed to it? So writing we fixed a bug doesn't make any sense (from a realm-android point of view)

PR #1087, opened 2015-05-05
RealmChangedListener now gets called on the same thread doing the commit

cmelchior:
This PR makes the OnRealmChanged callback behave more as expected and puts it in line with Realm-Cocoa behaviour.

Previously, callbacks registered on the thread doing the commit was not triggered, this is now the case. In order avoid weird cases like be below. The OnRealmChangedListener is called immediately instead of using the Handler mechanism.

```
// If using Handler messages, the following would be possible. By calling listeners at once, this is avoided.
realm.beginTransaction();
...
realm.commitTransaction();
realm.addChangeListener(new RealmChangedListener() {
  // This would get triggered even though it was added after the commit.
});
```

@emanuelez @kneth @jpsim @bmunkholm

### DIFF
--- a/realm/src/androidTest/java/io/realm/NotificationsTest.java
+++ b/realm/src/androidTest/java/io/realm/NotificationsTest.java
@@ -32,13 +32,24 @@ import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
 
+import io.realm.entities.AllTypes;
 import io.realm.entities.Dog;
 
 public class NotificationsTest extends AndroidTestCase {
 
+    private Realm realm;
+
     @Override
     protected void setUp() throws Exception {
         Realm.deleteRealmFile(getContext());
+    }
+
+    @Override
+    protected void tearDown() throws Exception {
+        super.tearDown();
+        if (realm != null) {
+            realm.close();
+        }
     }
 
     public void testFailingSetAutoRefreshOnNonLooperThread() throws ExecutionException, InterruptedException {
@@ -116,12 +127,13 @@ public class NotificationsTest extends AndroidTestCase {
             }
         });
 
-        // Wait until the looper is started
+        // Wait until the looper in the background thread is started
         while (!isReady.get()) {
             Thread.sleep(5);
         }
         Thread.sleep(100); 
 
+        // Trigger OnRealmChanged on background thread
         Realm realm = Realm.getInstance(getContext());
         realm.beginTransaction();
         Dog dog = realm.createObject(Dog.class);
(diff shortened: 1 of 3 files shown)


realm/src/androidTest/java/io/realm/NotificationsTest.java
```diff
@@ -304,4 +316,33 @@ public void testHandlerNotRemovedToSoon() {
...
+                success.set(true);
+            }
+        });
+        realm.beginTransaction();
+        realm.createObject(AllTypes.class);
+        realm.commitTransaction();
+        assertTrue(success.get());
+    }
+
+    public void testEmptyCommitTriggerOnChangeHandler() {
```
emanuelez: Confusing test name. We don't use the word handler anywhere else, unless to specify the Android's Handler class

emanuelez:
I'm not sure I like the calling of callbacks even if setup after a commit. It sounds counter-intuitive

cmelchior:
Not sure what you mean?  You mean that triggering the listener immediately is wrong?

emanuelez:
No, I'm referring to what you describe in the code snippet of your first comment.

kneth:
Just thinking about timing/ordering of operations: will this change imply that the change listener will be called when the Realm is initially opened (`Realm.initializeRealm()` might do a commit).

cmelchior:
It will, which might be a bit odd (although technically correct). I checked Realm-Cocoa and as far as I could tell they do the same.

tgoyne:
We had to do some non-trivial refactoring to eliminate spurious notifications from files being opened. The actual schema initialization happens before the user gets the chance to add a listener, and when Realms are opened with no initialization/migration needed we're careful to avoid committing a write transaction.

tgoyne:
Also, for the multi-process case we have to avoid even starting a write transaction when schema changes aren't needed, so that extensions don't hang on load when the main app is suspended during a write transaction (writing in that scenario obviously still doesn't work).

cmelchior:
Good points @tgoyne. We already grab a transaction and abort it if the schema is detected to be up to date. By executing the listeners immediately on the same thread we also go clear of the issue of getting RealmChange notifications for just opening (and creating) the initial Realm.

We have not implemented multiprocess support just yet, so any issues in this area will have to be fixed when that is introduced. I especially suspect that we need to begin a write transaction earlier to avoid  race conditions when creating the initial schema.

emanuelez:
:+1:

cmelchior:
Bump @kneth

kneth:
:+1:

beeender:
Hi

I still prefer the previous implementation.
Now, it would trigger ConcurrentModificationException if user is not careful enough, and actually some very common user cases  would trigger ConcurrentModificationException.

eg.
sendNotifications triggered the pageAdapter to refresh in onChange. And pageAdapter created some fragment which also has its realm instance and listener since the fragment wants to keep the instance in its lifecycle. Then when the addListener called in the fragment, ConcurrentModificationException happens.

Yes, post the fragment running would solve the problem, but sometimes this is not easy to reproduce & debug.

PR #1089, opened 2015-05-06
Introduce a Gradle plugin for Realm

emanuelez:
This can substitute having to define the dependency manually.
In the long run this will help us to split Realm and the annotation processor, saving a few KB from installed Realm-based apps.
Also this will allow byte-code manipulation.


emanuelez:
@cmelchior @kneth @bmunkholm

cmelchior:
You don't need 4 different versions of the launcher icon and we use a Realm logo as app icon for all our example apps, so I would add that.

But apart from that :+1:

gradle-plugin/sample/src/androidTest/java/io/realm/example/sample/ApplicationTest.java
```diff
@@ -0,0 +1,13 @@
+package io.realm.example.sample;
```
kneth: Standard header is missing.

gradle-plugin/sample/src/main/java/io/realm/example/sample/models/Person.java
```diff
@@ -0,0 +1,39 @@
+/*
+ * Copyright 2014 Realm Inc.
```
kneth: 2014 -> 2015

kneth:
Besides my comments: :+1:

emanuelez:
Ready for the second round @cmelchior @kneth @bmunkholm

gradle-plugin/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -0,0 +1,6 @@
+#Mon May 04 12:41:37 CEST 2015
+distributionBase=GRADLE_USER_HOME
+distributionPath=wrapper/dists
+zipStoreBase=GRADLE_USER_HOME
+zipStorePath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-2.2-all.zip
```
kneth: Are we going for version 2.4?

emanuelez: Yes, depending on which PR gets merged first, the other will have to adapt

bmunkholm:
:+1:

Issue #989, opened 2015-03-26
Finegrained notifications


Trikke:
Hi Realm devs,

we now have the RealmChangeListener to listen for changes if some background thread updates the Realm instance, which is great. But i'd like to put a feature request forward for more fine grained control when a listener can be called. It would be nice if it were possible to add a RealmChangeListener that only fires if a certain class of RealmObject is updated.

The issue i have now is that i have multiple fragments for multiple classes of RealmObject (eg. RealmUser, RealmMessage, RealmBeer) open at once, then a background services updates one of the RealmUser instances. Now all the RealmChangeListener are fired, and all fragments are updated individually. While ideally only the one that deals with RealmUser should be updating.

cmelchior:
Hi @Trikke
Yes, more finegrained notifications are also very high on our priority list, and we already have someone working on it. Unfortunately it require some changes to the core database, but you will hopefully see some progress on this soon.

feugatos:
Hi @cmelchior

Will this enhancement also include some sort of support for adding [RealmChangeListener](http://realm.io/docs/java/0.80.0/api/io/realm/RealmChangeListener.html)s to [RealmResults<>](http://realm.io/docs/java/0.80.0/api/io/realm/RealmResults.html) as well and get notifications when a [RealmObject](http://realm.io/docs/java/0.80.0/api/io/realm/RealmObject.html) referenced by the [RealmResults<>](http://realm.io/docs/java/0.80.0/api/io/realm/RealmResults.html) gets updated?

This would be a very useful feature and would surely improve the performance of [RealmBaseAdapter](http://realm.io/docs/java/0.80.0/api/io/realm/RealmBaseAdapter.html).
For instance the number of draw calls/redraws to a bound [ListView](https://developer.android.com/reference/android/widget/ListView.html) would be vastly reduced.

Thanks in advance,
Dimitris.

rynti:
+1

treylon:
+1

bdbergeron:
I'm also anxiously awaiting this feature. In my opinion, this is definitely one of the features that needs to be implemented before a 1.0 release!

yolapop:
I'm also excited with this feature :+1: 

Btw @feugatos, can you explain what you mean by this?

> This would be a very useful feature and would surely improve the performance of RealmBaseAdapter. For instance the number of draw calls/redraws to a bound ListView would be vastly reduced.

beeender:
Comments from #1378 by mshusek.

[...]

PR #1090, opened 2015-05-06
RealmCursor

### LINKED ISSUE #438: Android component: RealmCursor
Realm for Android looks awesome; however there are some Android specific use cases where certain elements work better when backed by a Cursor (http://developer.android.com/reference/android/database/Cursor.html) and when using a CursorLoader (http://developer.android.com/reference/android/content/CursorLoader.html). Is there any plans for supporting this interface?

cmelchior:
Fixes #438 

**Missing**
ContentProviders exposes a cursor, which works fine if used within a app. However it breaks if shared with another app:

- Accessing an ContentProvider in another app is done using IPC (Blocked by #1091)
- The ContentProvider is started on App1 main thread
- App2 is accessing ContentProvider using a Binder thread, meaning that opening the Realm in onCreate causes Thread exceptions.

We can ship RealmCursor with the explicit caveat that it doesn't work with ContentProviders across apps, but it is not ideal.

**Description**
This adds a Cursor implementation for Realm data. The main challenge is mapping the very limited types in SQLite to the more expressive types in Realm.

The current approach is to respect the Cursor interface as much as possible as that is what will be exposed in ContentProviders and other framework classes.

This has the following implications:
```
// Simple types
Boolean:  becomes a Cursor.INTEGER and will return 0 or 1 when calling Cursor.getInt()
Date: becomes a Cursor.INTEGER. getLong() return ms. since epoch. getInt() and getShort() will throw exception.

// Links/LinkLists: No immediately good solution is available. Suggestion is adding support  for the following syntax. Disallowing accessing RealmLists might also be an option as it doesn't map well to a table layout anyway?
int calculatedIndex = cursor.getIndex("foo.columnString");
int calculatedIndex = cursor.getIndex("fooList[0].columnString");
```

I have a single new method ```setIdColumn(String columnName)``` that creates a virtual "_id" field used by many framework classes. This should be acceptable as it is defined when the Cursor is created and the developer know it is a ReamCursor. Adding other nice methods like ```getDate()``` should be avoided as client code should only interact with the cursor using the Interface.

Note that RealmCursor just reuses the tableview backing the RealmResults. This means that any changes to RealmResults will be reflected in the RealmCursor. I am not sure this behaviour is developer friendly but AFAIK we cannot lock the tableview until we start implementing Async queries.

@emanuelez @kneth @bmunkholm


realm/src/androidTest/java/io/realm/RealmCursorTest.java
```diff
@@ -0,0 +1,595 @@
+/*
+ * Copyright 2015 Realm Inc.
```
kneth: The correct year :-)

realm/src/androidTest/java/io/realm/RealmCursorTest.java
```diff
@@ -0,0 +1,595 @@
...
+    @Override
+    protected void setUp() throws Exception {
+        super.setUp();
+        Realm.deleteRealmFile(getContext());
+        realm = Realm.getInstance(getContext());
+        populateTestRealm(realm, SIZE);
+        cursor = realm.allObjects(AllTypes.class).getCursor();
+    }
+
+    private void populateTestRealm(Realm realm, int objects) {
```
kneth: We use this is almost all our test modules. Why don't we move to a helper class?

realm/src/androidTest/java/io/realm/RealmCursorTest.java
```diff
@@ -0,0 +1,595 @@
...
+            Dog dog = realm.createObject(Dog.class);
+            dog.setName("Foo " + i);
+            allTypes.setColumnRealmObject(dog);
+        }
+        realm.commitTransaction();
+    }
+
+    @Override
+    protected void tearDown() throws Exception {
+        super.tearDown();
```
kneth: Shouldn't you tear down after closing?

realm/src/androidTest/java/io/realm/RealmCursorTest.java
```diff
@@ -0,0 +1,595 @@
...
+    public void testGetCount() {
+        assertEquals(AllTypes.COL_COUNT, cursor.getColumnCount());
+    }
+
+    public void testGetPosition() {
+        assertEquals(-1, cursor.getPosition());
+        cursor.moveToFirst();
+        assertEquals(0, cursor.getPosition());
+        cursor.moveToLast();
+        assertEquals(9, cursor.getPosition());
```
kneth: `SIZE -1` instead of 9.

realm/src/androidTest/java/io/realm/RealmCursorTest.java
```diff
@@ -0,0 +1,595 @@
...
+    }
+
+    public void testMoveOffsetValid() {
+        assertTrue(cursor.move(SIZE / 2));
+    }
+
+    public void testMoveOffsetInvalid() {
```
kneth: What about testing -1 and `SIZE`?

realm/src/androidTest/java/io/realm/RealmCursorTest.java
```diff
@@ -0,0 +1,595 @@
...
+
+    public void testGetPosition() {
+        assertEquals(-1, cursor.getPosition());
+        cursor.moveToFirst();
+        assertEquals(0, cursor.getPosition());
+        cursor.moveToLast();
+        assertEquals(9, cursor.getPosition());
+    }
+
+    public void testMoveOffsetValid() {
```
kneth: What about testing 0 and `SIZE-1`?

realm/src/androidTest/java/io/realm/RealmCursorTest.java
```diff
@@ -0,0 +1,595 @@
...
+    }
+
+    public void testMoveToFirst() {
+        assertTrue(cursor.moveToFirst());
+        assertEquals(0, cursor.getPosition());
+    }
+
+    public void testMoveToLast() {
+        assertTrue(cursor.moveToLast());
+        assertEquals(9, cursor.getPosition());
```
kneth: Use `SIZE - 1` instead.

realm/src/androidTest/java/io/realm/RealmCursorTest.java
```diff
@@ -0,0 +1,595 @@
...
+
+    public void testGetColumnNames() {
+        String[] names = cursor.getColumnNames();
+        assertEquals(AllTypes.COL_COUNT, names.length);
+        assertEquals("columnString", names[AllTypes.COL_INDEX_STRING]);
+        assertEquals("columnRealmList", names[AllTypes.COL_INDEX_LIST]);
+    }
+
+    public void testGetColumnCount() {
+        assertEquals(9, cursor.getColumnCount());
```
bmunkholm: SIZE

realm/src/androidTest/java/io/realm/RealmCursorTest.java
```diff
@@ -0,0 +1,595 @@
...
+
+    public void testRequeryThrows() {
+        try {
+            cursor.requery();
+            fail();
+        } catch (UnsupportedOperationException expected) {
+        }
+    }
+
+    public void testClose() {
```
bmunkholm: Test order could be a bit better. testClose() seems unrelated here. Collect the tests for "unsupported" methods.

I would personally have collected them all in one method. Literally compressing lines by a factor of 10 down to less than 10 lines total - completely disregarding normal structure and max line size :-) e.g.:
```
try { cursor.requery();       fail(); } catch (UnsupportedOperationException expected) {} }
try { cursor.deactivate();    fail(); } catch (UnsupportedOperationException expected) {} }
// etc..
```
Makes it easier to overview which methods to are unsupported.

realm/src/main/java/io/realm/RealmResults.java
```diff
@@ -490,6 +491,18 @@ public boolean add(E element) {
     public void add(int index, E element) {
         throw new UnsupportedOperationException();
     }
+
+    /**
+     * Returns this result set as a {@link RealmCursor}. All changes done to to this RealmResults afterwards will
```
bmunkholm: Wording wrong:
"to to"
"this RealmResults"

realm/src/main/java/io/realm/android/RealmCursor.java
```diff
@@ -0,0 +1,548 @@
...
+import android.os.Bundle;
+
+import io.realm.Realm;
+import io.realm.RealmChangeListener;
+import io.realm.RealmResults;
+import io.realm.internal.ColumnType;
+import io.realm.internal.TableOrView;
+
+/**
+ * This classes exposes {@link io.realm.RealmResults} as a cursor.
```
bmunkholm: classes -> class

realm/src/main/java/io/realm/android/RealmCursor.java
```diff
@@ -0,0 +1,548 @@
...
+ * <p>
+ * It is possible to traverse links using dot notation to access data in referenced objects, ie.
+ * {@code cursor.getInt("foo.bar"} will return the integer from the {@code bar} field in the {@code foo} object
+ * referenced the class represented by the RealmResult.
```
bmunkholm: wording: "referenced the"

realm/src/androidTest/java/io/realm/RealmCursorTest.java
kneth: Use `SIZE - 1`.

realm/src/androidTest/java/io/realm/RealmCursorTest.java
```diff
@@ -0,0 +1,595 @@
...
+                    case INT: cursor.getInt(AllTypes.COL_INDEX_STRING); break;
+                    case LONG: cursor.getLong(AllTypes.COL_INDEX_STRING); break;
+                    case FLOAT: cursor.getFloat(AllTypes.COL_INDEX_STRING); break;
+                    case DOUBLE: cursor.getDouble(AllTypes.COL_INDEX_STRING); break;
+                    case BLOB: cursor.getBlob(AllTypes.COL_INDEX_STRING); break;
+                }
+                fail(cursorGetter + " should throw an exception");
+            } catch (IllegalArgumentException expected) {
+            } catch (Exception wrongException) {
+                throw new RuntimeException(cursorGetter + " threw the wrong exception: ", wrongException);
```
kneth: `fail()` is probably enough.

realm/src/main/java/io/realm/android/RealmCursor.java
```diff
@@ -0,0 +1,548 @@
...
+        }
+    };
+
+    /**
+     * Exposes a RealmResults object as a cursor. Use {@link RealmResults#getCursor()} instead of this
+     * constructor
+     *
+     * @param table Table view representing the query results.
+     */
+    public RealmCursor(Realm realm, TableOrView table) {
```
bmunkholm: Why make it public if it should only be constructed from getCursor()?

realm/src/main/java/io/realm/android/RealmCursor.java
```diff
@@ -0,0 +1,548 @@
+/*
+ * Copyright 2014 Realm Inc.
```
kneth: The year should be 2015.

realm/src/main/java/io/realm/android/RealmCursor.java
cmelchior: Because they are in different packages. RealmResults are in ```io.realm``` while this is located in ```io.realm.android```.  #1093 is about the same.

realm/src/main/java/io/realm/android/RealmCursor.java
```diff
@@ -0,0 +1,548 @@
...
+        rowIndex = position;
+        return true;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public boolean moveToFirst() {
+        return moveToPosition(0);
```
bmunkholm: Reduce to as there is no reason to do all the parameter checks:
````
rowIndex = 0;
return true;
````

File: realm/src/main/java/io/realm/android/RealmCursor.java [hunk 17]
```diff
@@ -0,0 +1,548 @@
...
+     * {@inheritDoc}
+     */
+    @Override
+    public long getLong(int columnIndex) {
+        return mapRealmTypeToCursor(columnIndex, true);
+    }
+
+    private long mapRealmTypeToCursor(int columnIndex, boolean acceptLong) {
+        ColumnType type = table.getColumnType(columnIndex);
+        switch(type) {
```
bmunkholm: space after "switch"

realm/src/main/java/io/realm/android/RealmCursor.java
```diff
@@ -0,0 +1,548 @@
...
+     * {@inheritDoc}
+     */
+    @Override
+    public double getDouble(int columnIndex) {
+        return table.getDouble(columnIndex, rowIndex);
+    }
+
+    /**
+     * Returns the SQLite Cursor type for a column index. Realm has more types than SQLite and some of them cannot be
+     * mapped to any meaningful SQLite type. These will return -1 and are cannot be accessed using a cursor.
```
bmunkholm: List the unsupported types

realm/src/main/java/io/realm/android/RealmCursor.java
```diff
@@ -0,0 +1,548 @@
...
+        @Override
+        public void onChange() {
+            if (dataSetObservable != null) {
+                dataSetObservable.notifyChanged(); // TODO Should be replaced with something more fine grained.
+            }
+        }
+    };
+
+    /**
+     * Exposes a RealmResults object as a cursor. Use {@link RealmResults#getCursor()} instead of this
```
kneth: Documentation says `RealmResults` but signature says `TableOrView`. Code and doc should be consistent.

kneth: Saving a reference to the class (or the proxy class) could be useful.

realm/src/main/java/io/realm/android/RealmCursor.java
```diff
@@ -0,0 +1,548 @@
...
+    @Override
+    public String getColumnName(int columnIndex) {
+        return table.getColumnName(columnIndex);
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public String[] getColumnNames() {
```
kneth: The proxy class has this information.

realm/src/main/java/io/realm/android/RealmCursor.java
```diff
@@ -0,0 +1,548 @@
...
+     * @return One of the {@code FIELD_TYPE_*}'s described in {@link Cursor}
+     */
+    @Override
+    public int getType(int columnIndex) {
+        ColumnType realmType = table.getColumnType(columnIndex);
+        switch (realmType) {
+            case BOOLEAN: return Cursor.FIELD_TYPE_INTEGER;
+            case INTEGER: return Cursor.FIELD_TYPE_INTEGER;
+            case FLOAT: return Cursor.FIELD_TYPE_FLOAT;
+            case DOUBLE: return Cursor.FIELD_TYPE_FLOAT;
```
kneth: Will this lead to a loss of precision?

realm/src/androidTest/java/io/realm/RealmCursorTest.java
```diff
@@ -0,0 +1,595 @@
...
+            }
+        }
+    }
+
+    // Test that all getters fail when the cursor is closed
+    public void testGetXXXFailWhenCursorClosed() {
+        cursor.close();
+        for (CursorGetter cursorGetter : CursorGetter.values()) {
```
bmunkholm: Reuse code above.
What else is not allowed when the cursor is closed? Everything else I assume?

realm/src/main/java/io/realm/android/RealmCursor.java
```diff
@@ -0,0 +1,548 @@
...
+    }
+
+    /**
+     * Realm doesn't support {@link ContentResolver}. Calling this will throw an exception.
+     *
+     * @throws UnsupportedOperationException
+     */
+    @Override
+    public void setNotificationUri(ContentResolver cr, Uri uri) {
+        throw new UnsupportedOperationException("Notification URI's are not supported by RealmCursor");
```
kneth: URI's -> URIs

realm/src/main/java/io/realm/android/RealmCursor.java
```diff
@@ -0,0 +1,548 @@
...
+    }
+
+    /**
+     * Realm doesn't support {@link ContentResolver}. Calling this will throw an exception.
+     *
+     * @throws UnsupportedOperationException
+     */
+    @Override
+    public Uri getNotificationUri() {
+        throw new UnsupportedOperationException("Notification URI's are not supported by RealmCursor");
```
kneth: URI's -> URIs

realm/src/main/java/io/realm/android/RealmCursor.java
```diff
@@ -0,0 +1,548 @@
...
+
+    /**
+     * RealmCursors are not accessible across processes. This method will always return false.
+     *
+     * @return Will always return {@code false}.
+     * @see Cursor#getWantsAllOnMoveCalls()
+     */
+    @Override
+    public boolean getWantsAllOnMoveCalls() {
+        return false; // Realm TableViews doesn't support access from multiple processes.
```
kneth: doesn't -> don't

realm/src/androidTest/java/io/realm/RealmCursorTest.java
```diff
@@ -0,0 +1,595 @@
...
+
+    public void testGetColumnIndexOrThrowNotFoundThrows() {
+        try {
+            cursor.getColumnIndexOrThrow("foo");
+            fail();
+        } catch (IllegalArgumentException expected) {
+        }
+    }
+
+    public void testGetColumnNameInvalidIndexThrows() {
```
bmunkholm: Also test SIZE+1.

bmunkholm:
Every single method that takes a columnIndex should have tests that out of bounds -1 and SIZE.
Prefer to collect those tests together.

bmunkholm:
We should add an example that uses Cursor in a scenario that couldn't be done without it.

kneth:
How does the virtual id column react to sorting? I mean, is there a difference between

    RealmResults<A> rr = realm.allObjects();
    RealmCursor<A> cursor = rr.getCursor();
    rr.sort("foo");

and

    RealmResults<A> rr = realm.allObjects();
    rr.sort("foo");
    RealmCursor<A> cursor = rr.getCursor();

when it comes to these ids?

realm/src/androidTest/java/io/realm/RealmCursorTest.java
cmelchior: Nope, it is perfectly safe to close the Realm after the super class has been called

realm/src/main/java/io/realm/android/RealmCursor.java
cmelchior: Not really since the cursor still have Cursor.getFloat() and Cursor.getDouble(), but SQLite doesn't have a way to expose the difference. So ultimately it is a developer responsibility :(

realm/src/androidTest/java/io/realm/RealmCursorTest.java
cmelchior: Yes, but testing all methods seemed like going kinda of overboard.

cmelchior:
Yes, a example is a good idea. I will add that as a separate PR.
The virtual Id is just an alias, and as the Cursor is not "stable", it will change if you change the RealmResults. This might not be optional, and we might want to consider cloning the RealmResults for the Cursor, hopefully there is a way to do that without having to rerun the query.

realm/src/androidTest/java/io/realm/RealmCursorTest.java
bmunkholm: You don't have any checks if the cursor is closed in any methods, so it would expose random NullPointerExceptions everywhere. If that's the intended behaviour there is no reason to actually test it. And here it's also just a NullPointerException anyway. What's the norm with cursors?

It's not that big a thing to test actually, if you use my "compact" form of writing the tests.

realm/src/androidTest/java/io/realm/RealmCursorTest.java
cmelchior: It has been added

realm/src/androidTest/java/io/realm/RealmCursorTest.java
cmelchior: Now testing for close (like SQLiteCursor) and throwing proper exception. Unit test now tests all methods that fail if closed.

realm/src/androidTest/java/io/realm/RealmCursorTest.java
cmelchior: It has been added.

cmelchior:
PR ready for review: @emanuelez @kneth @bmunkholm 
First post has been updated with current design.

An example using this will be a separate PR.

realm/src/main/java/io/realm/android/RealmCursor.java
```diff
@@ -0,0 +1,729 @@
...
+     *</ol>
+     *
+     * @param columnIndex Get the data type for this column index or -1 if the column isn't accessible.
+     * @return One of the {@code FIELD_TYPE_*}'s described in {@link Cursor}.
+     */
+    @Override
+    public int getType(int columnIndex) {
+        checkClosed();
+        ColumnType realmType = table.getColumnType(columnIndex);
+        switch (realmType) {
```
emanuelez: Is this faster than a Map lookup?

realm/src/androidTest/java/io/realm/entities/AllTypes.java
```diff
@@ -22,6 +22,19 @@
 import io.realm.RealmObject;
 
 public class AllTypes extends RealmObject {
+
+
+    public static final int COL_COUNT = 9;
```
kneth: Maybe we should prefix them with `FIELD_` instead of `COL_`?

realm/src/androidTest/java/io/realm/RealmCursorTest.java
```diff
@@ -0,0 +1,573 @@
...
+        assertEquals(Cursor.FIELD_TYPE_FLOAT, cursor.getType(AllTypes.COL_INDEX_DOUBLE));
+        assertEquals(Cursor.FIELD_TYPE_INTEGER, cursor.getType(AllTypes.COL_INDEX_BOOLEAN));
+        assertEquals(Cursor.FIELD_TYPE_INTEGER, cursor.getType(AllTypes.COL_INDEX_DATE));
+        assertEquals(-1, cursor.getType(AllTypes.COL_INDEX_OBJECT));
+        assertEquals(-1, cursor.getType(AllTypes.COL_INDEX_LIST));
+    }
+
+    public void testUnsupportedMethods() {
+        cursor.moveToFirst();
+        // TODO We do support isNull for Objects.
```
bmunkholm: want to fix this TODO now?

realm/src/androidTest/java/io/realm/RealmResultsTest.java
```diff
@@ -74,15 +67,20 @@ private void populateTestRealm(int objects) {
...
+    @Override
+    protected void setUp() throws InterruptedException {
+        Realm.deleteRealmFile(getContext());
+        testRealm = Realm.getInstance(getContext());
+        populateTestRealm();
+    }
+
     private void populateTestRealm() {
         populateTestRealm(TEST_DATA_SIZE);
+//        TestHelper.populateTestRealm(testRealm, TEST_DATA_SIZE);
```
bmunkholm: why is this commented out? Fix or remove.

realm/src/main/java/io/realm/android/RealmCursor.java
bmunkholm: You could do with a simple array lookup if you wanted.

realm/src/main/java/io/realm/android/RealmCursor.java
```diff
@@ -0,0 +1,729 @@
...
+     * @return Will always return {@code false}.
+     * @see Cursor#getWantsAllOnMoveCalls()
+     */
+    @Override
+    public boolean getWantsAllOnMoveCalls() {
+        return false; // Realm TableViews don't support access from multiple processes.
+    }
+
+    /**
+     * Realm doesn't support extra metadata in the form a bundle. This method will always return the empty bundle.
```
bmunkholm: Missing "of"

realm/src/main/java/io/realm/android/RealmCursor.java
```diff
@@ -0,0 +1,729 @@
...
+     *
+     * @return {@code Bundle.EMPTY}
+     */
+    @Override
+    public Bundle getExtras() {
+        return Bundle.EMPTY;
+    }
+
+    /**
+     * Realm doesn't support extra metadata in the form a bundle. This method will always return the empty bundle.
```
bmunkholm: Missing "of"

realm/src/main/java/io/realm/android/RealmCursor.java
```diff
@@ -0,0 +1,729 @@
...
+     * @return {@code Bundle.EMPTY}
+     */
+    @Override
+    public Bundle respond(Bundle extras) {
+        return Bundle.EMPTY;
+    }
+
+    /**
+     * Map a field name to also act as the "_id" column. Such a column is required by a number of Android framework
+     * classes that uses cursors. The field must be able to mapped to a long so {@link #getLong(int)} can return a
```
bmunkholm: "to mapped" -> "to be mapped"

realm/src/main/java/io/realm/android/RealmCursor.java
```diff
@@ -0,0 +1,729 @@
...
+     * Returns the value of the requested column as a byte array.
+     *
+     * @param columnIndex the zero-based index of the target column.
+     * @return the value of that column as a byte array.
+     * @throws IllegalArgumentException if the requested column is not a byte array.
+     */
+    @Override
+    public byte[] getBlob(int columnIndex) {
+        checkClosed();
+        checkPosition();
```
bmunkholm: Do you ever call checkPosition() without also checkClosed()?
Consider making a checkClosedAndPosition()

realm/src/main/java/io/realm/android/RealmCursor.java
cmelchior: Switch statements on enums have specific support in bytecode as well as being very readable and using less memory. A map would definitely not be faster as that would involve multiple method calls as well as memory overhead. An primitive array *might* be faster, but would also increase memory usage as we would have to keep an array around. In any case this is not a critical path method.

realm/src/main/java/io/realm/RealmResults.java
```diff
@@ -490,6 +491,18 @@ public boolean add(E element) {
     public void add(int index, E element) {
         throw new UnsupportedOperationException();
     }
+
+    /**
+     * Returns this result set as a {@link RealmCursor}. All changes done to this RealmResults afterwards will
```
kneth: "Returns the RealmResults as a RealmCursor. All changes applied to this RealmResults ..."

realm/src/main/java/io/realm/RealmResults.java
```diff
@@ -490,6 +491,18 @@ public boolean add(E element) {
...
+     * also be automatically reflected in the Cursor without having to fetch it again.
+     *
+     * @return A cursor representing the query results.
```
kneth: query -> RealmResults

realm/src/main/java/io/realm/android/RealmCursor.java
```diff
@@ -0,0 +1,757 @@
...
+/**
+ * This class exposes {@link io.realm.RealmResults} as a cursor.
+ * <p>
+ * Many Android framework classes require the presences of an "_id" field. Instead of adding such a field to your
+ * model class it is instead possible to use {@link #setIdColumn(String)}.
+ * <p>
+ * A RealmCursor has the same thread restrictions as RealmResults, so it is not possible to move a RealmCursor between
+ * threads.
+ * <p>
+ * A RealmCursor is currently limited to only being able to display data from a single RealmClass, ie. it is not
```
kneth: ", ie." -> "i.e., "

realm/src/main/java/io/realm/android/RealmCursor.java
```diff
@@ -0,0 +1,757 @@
...
+import io.realm.Realm;
+import io.realm.RealmChangeListener;
+import io.realm.RealmResults;
+import io.realm.internal.ColumnType;
+import io.realm.internal.TableOrView;
+
+/**
+ * This class exposes {@link io.realm.RealmResults} as a cursor.
+ * <p>
+ * Many Android framework classes require the presences of an "_id" field. Instead of adding such a field to your
```
kneth: I think you should move this section to the end.

realm/src/main/java/io/realm/android/RealmCursor.java
```diff
@@ -0,0 +1,757 @@
...
+    public int getCount() {
+        checkClosed();
+        return (int) table.size();
+    }
+
+    /**
+     * Returns the current position of the cursor in the row set.
+     * The value is zero-based. When the row set is first returned the cursor
+     * will be at position -1, which is before the first row. After the
+     * last row is returned another call to next() will leave the cursor past
```
kneth: Can you link to the methods?

realm/src/main/java/io/realm/android/RealmCursor.java
```diff
@@ -0,0 +1,757 @@
...
+     * @return the current cursor position.
+     */
+    @Override
+    public int getPosition() {
+        checkClosed();
+        return rowIndex;
+    }
+
+    /**
+     * Move the cursor by a relative amount, forward or backward, from the
```
kneth: "Move" -> "Moves"

realm/src/main/java/io/realm/android/RealmCursor.java
```diff
@@ -0,0 +1,757 @@
...
+     * current position. Positive offsets move forwards, negative offsets move
+     * backwards. If the final position is outside of the bounds of the result
+     * set then the resultant position will be pinned to -1 or count() depending
```
kneth: "resultant" -> "new" or "final" (the word "resultant" isn't wrong but sounds a bit old-fashioned to me).

realm/src/main/java/io/realm/android/RealmCursor.java
```diff
@@ -0,0 +1,757 @@
...
+     * on whether the value is off the front or end of the set, respectively.
+     *
+     * <p>This method will return true if the requested destination was
```
kneth: "was" -> "is"

realm/src/main/java/io/realm/android/RealmCursor.java
```diff
@@ -0,0 +1,757 @@
...
+     * reachable, otherwise, it returns false. For example, if the cursor is at
```
kneth: "is at currently" -> is currently at"

realm/src/main/java/io/realm/android/RealmCursor.java
```diff
@@ -0,0 +1,757 @@
...
+     * @return whether the requested move fully succeeded.
+     */
+    @Override
+    public final boolean move(int offset) {
+        checkClosed();
+        return moveToPosition(rowIndex + offset);
+    }
+
+    /**
+     * Move the cursor to an absolute position. The valid
```
kneth: "Move" -> "Moves"

realm/src/main/java/io/realm/android/RealmCursor.java
```diff
@@ -0,0 +1,757 @@
...
+     * range of values is -1 &lt;= position &lt;= count.
+     *
+     * <p>This method will return true if the request destination was reachable,
```
kneth: "was" -> "is"

realm/src/main/java/io/realm/android/RealmCursor.java
```diff
@@ -0,0 +1,757 @@
...
+            rowIndex = -1;
+            return false;
+        }
+
+        rowIndex = position;
+        return true;
+    }
+
+    /**
+     * Move the cursor to the first row.
```
kneth: "Move" -> "Moves"

realm/src/main/java/io/realm/android/RealmCursor.java
```diff
@@ -0,0 +1,757 @@
...
+     *
+     * @return whether the move succeeded.
+     */
+    @Override
+    public boolean moveToFirst() {
+        return moveToPosition(0);
+    }
+
+    /**
+     * Move the cursor to the last row.
```
kneth: "Move" -> "Moves"

realm/src/main/java/io/realm/android/RealmCursor.java
```diff
@@ -0,0 +1,757 @@
...
+     *
+     * @return whether the move succeeded.
+     */
+    @Override
+    public boolean moveToLast() {
+        return moveToPosition(getCount() - 1);
+    }
+
+    /**
+     * Move the cursor to the next row.
```
kneth: "Move" -> "Moves"

realm/src/main/java/io/realm/android/RealmCursor.java
```diff
@@ -0,0 +1,757 @@
...
+     *
+     * @return whether the move succeeded.
+     */
+    @Override
+    public boolean moveToNext() {
+        return moveToPosition(rowIndex + 1);
+    }
+
+    /**
+     * Move the cursor to the previous row.
```
kneth: "Move" -> "Moves"

realm/src/main/java/io/realm/android/RealmCursor.java
```diff
@@ -0,0 +1,757 @@
...
+    public boolean isAfterLast() {
+        checkClosed();
+        if (getCount() == 0) {
+            return true;
+        }
+        return rowIndex == getCount();
+    }
+
+    /**
+     * Returns the zero-based index for the given column name, or -1 if the column doesn't exist.
```
kneth: I understand that SQLite will talk about columns, but Realm model classes have fields and not columns.

realm/src/main/java/io/realm/android/RealmCursor.java
```diff
@@ -0,0 +1,757 @@
...
+        if (DEFAULT_ID_COLUMN.equals(columnName)) {
+            return (int) idColumnIndex;
+        }
+        return (int) table.getColumnIndex(columnName);
+    }
+
+    /**
+     * Returns the zero-based index for the given column name, or throws
+     * {@link IllegalArgumentException} if the column doesn't exist. If you're not sure if
+     * a column will exist or not use {@link #getColumnIndex(String)} and check for -1, which
```
kneth: "not use" -> "not, then use"
"-1, which" -> "-1. This"

realm/src/main/java/io/realm/android/RealmCursor.java
```diff
@@ -0,0 +1,757 @@
...
+
+    /**
+     * Realm currently doesn't support null. Calling this method will always throw an exception.
+     *
+     * @throws UnsupportedOperationException
+     */
+    @Override
+    public boolean isNull(int columnIndex) {
+        ColumnType realmType = table.getColumnType(columnIndex);
+        if (realmType == ColumnType.LINK) {
```
kneth: The `LINK_LIST` type is supported as well.

realm/src/main/java/io/realm/android/RealmCursor.java
```diff
@@ -0,0 +1,757 @@
...
+     *
+     * @throws UnsupportedOperationException
+     */
+    @Override
+    public boolean requery() {
+        throw new UnsupportedOperationException("Request a new cursor from the RealmResults or call Realm.refresh() instead");
+    }
+
+    /**
+     * Calling this will close this cursor. The original RealmResult will not be affected. Trying to access any data in
```
kneth: "RealmResult" -> "RealmResults"

realm/src/main/java/io/realm/android/RealmCursor.java
```diff
@@ -0,0 +1,757 @@
...
+     * the cursor after calling this is illegal an the behavior is undefined.
```
kneth: "this" -> "this method"

realm/src/main/java/io/realm/android/RealmCursor.java
```diff
@@ -0,0 +1,757 @@
...
+     *
+     * @throws UnsupportedOperationException
+     */
+    @Override
+    public void unregisterContentObserver(ContentObserver observer) {
+        throw new UnsupportedOperationException("Content observers not supported");
+    }
+
+    /**
+     * Register an observer that is called when changes happen to the contents
```
kneth: "Register" -> "Registers"

realm/src/main/java/io/realm/android/RealmCursor.java
```diff
@@ -0,0 +1,757 @@
...
+     * @param observer the object that gets notified when the cursors data set changes.
+     * @see #unregisterDataSetObserver(DataSetObserver)
+     */
+    @Override
+    public void registerDataSetObserver(DataSetObserver observer) {
+        dataSetObservable.registerObserver(observer);
+    }
+
+    /**
+     * Unregister an observer that has previously been registered with this
```
kneth: "Unregister" -> "Unregisters"

realm/src/main/java/io/realm/android/RealmCursor.java
```diff
@@ -0,0 +1,757 @@
...
+     *
+     * @return {@code Bundle.EMPTY}
+     */
+    @Override
+    public Bundle respond(Bundle extras) {
+        return Bundle.EMPTY;
+    }
+
+    /**
+     * Map a field name to also act as the "_id" column. Such a column is required by a number of Android framework
```
kneth: "Map" -> "Maps"

realm/src/main/java/io/realm/android/RealmCursor.java
```diff
@@ -0,0 +1,757 @@
...
+     * classes that uses cursors. The field must be able to be mapped to a long so {@link #getLong(int)} can return a
```
kneth: "uses" -> "use"

realm/src/main/java/io/realm/android/RealmCursor.java
cmelchior: You are probably right, but I can't seem to find any documentation on that?

realm/src/main/java/io/realm/android/RealmCursor.java
cmelchior: Never check without closed, but do call close without calling check. So merging them would not help much unless we duplicated exception handling, which isn't ideal either. I feel this approach here is more readable for the time being.

realm/src/main/java/io/realm/android/RealmCursor.java
emanuelez: Actually an enum could do all of the work without the need of a switch statement. Something like this:
https://schneide.wordpress.com/2010/12/13/avoid-switch-use-enum/

realm/src/main/java/io/realm/android/RealmCursor.java
kneth: `RealmQuery.isNull()` states that it supports `RealmList`.

cmelchior: DOH :)...but I looked at the query methods in C++ and it wasn't documented there at all as far as I could tell?

kneth: Core's query engine requires that you know where to look, and the native code for between Java and Core is documented by source code only :-) The JavaDoc is clear.

realm/src/main/java/io/realm/android/RealmCursor.java
```diff
@@ -0,0 +1,776 @@
...
+    }
+
+    /**
+     * Realm doesn't support {@link #requery()}. Calling this will throw an exception.
+     *
+     * @throws UnsupportedOperationException
+     */
+    @Override
+    public boolean requery() {
+        throw new UnsupportedOperationException("Request a new cursor from the RealmResults or call Realm.refresh() instead");
```
nhachicha: In case of a `ContentProvider`, this message could be ambiguous as the caller is not aware about the persistence layer backing this cursor. The generic message should be fine IMO

realm/src/main/java/io/realm/android/RealmCursor.java
```diff
@@ -0,0 +1,776 @@
...
+            case INTEGER:
+            case FLOAT:
+            case DOUBLE:
+            case STRING:
+            case BINARY:
+            case DATE:
+            case TABLE:
+            case MIXED:
+            default:
+            throw new UnsupportedOperationException("isNull not yet supported by Realm for this type: " + realmType);
```
nhachicha: IMO This is a *NotYetImplemented* kind of exception. Once it's available we may start using @Since/@Version

realm/src/main/java/io/realm/android/RealmCursor.java
```diff
@@ -0,0 +1,776 @@
...
+     * @throws UnsupportedOperationException
+     */
+    @Override
+    public boolean requery() {
+        throw new UnsupportedOperationException("Request a new cursor from the RealmResults or call Realm.refresh() instead");
+    }
+
+    /**
+     * Calling this method will close this cursor. The original RealmResults will not be affected. Trying to access any data in
+     * the cursor after calling this is illegal an the behavior is undefined.
```
nhachicha: typo *and*

realm/src/main/java/io/realm/android/RealmCursor.java
```diff
@@ -0,0 +1,776 @@
...
+     */
+    @Override
+    public void close() {
+        closed = true;
+        table = null; // Instead of closing the table, we just release it. Original RealmResults would also be closed otherwise.
+        realm.removeChangeListener(changeListener);
+        dataSetObservable.notifyInvalidated();
+    }
```
nhachicha: It's safe to call `contentObservable.unregisterAll()` at this point.

realm/src/main/java/io/realm/android/RealmCursor.java
```diff
@@ -0,0 +1,776 @@
...
+
+    /**
+     * Realm doesn't support {@link ContentResolver}. Calling this will throw an exception.
+     *
+     * @throws UnsupportedOperationException
+     */
+    @Override
+    public void setNotificationUri(ContentResolver cr, Uri uri) {
+        throw new UnsupportedOperationException("Notification URIs are not supported by RealmCursor");
```
nhachicha: Is this why we're not inheriting from `AbstractCursor`?

realm/src/main/java/io/realm/android/RealmCursor.java
cmelchior: It is. I am not a big fan of using @Since/@Version before 1.0 as the API is still pr. definition unstable at that point, but those annotations are definitely worth keeping in mind.

realm/src/main/java/io/realm/android/RealmCursor.java
cmelchior: Good catch. It has been added.

realm/src/main/java/io/realm/android/RealmCursor.java
cmelchior: Partly. The primary reason is that AbstractCursor is really tightly coupled to the SQLite implementation. So instead of working around that it was easier to only use the interface (although some of the method implementations has been copied from AbstractCursor).

cmelchior:
Reverting to WIP status. There are some details about ContentObservers and NotificationUri's that needs to be discussed as it is probably also part of  #989

realm/src/main/java/io/realm/android/RealmCursor.java
```diff
@@ -0,0 +1,777 @@
...
+    private static final String DEFAULT_ID_COLUMN = "_id";
+
+    // Needed because constants are only available from API 11
+    private static final int FIELD_TYPE_NULL;
+    private static final int FIELD_TYPE_INTEGER;
+    private static final int FIELD_TYPE_FLOAT;
+    private static final int FIELD_TYPE_STRING;
+    private static final int FIELD_TYPE_BLOB;
+    static {
+        if (android.os.Build.VERSION.SDK_INT >= 11) {
```
zaki50: This branch can be removed.

Constant values are always inlined(JLS 13.1 The Form of a Binary (3)).
This means there will be no reference to Cursor.FIELD* in class file.

yuriy-larin:
Is Cursor example still not ready?

cmelchior:
No, there is work on it here though: https://github.com/realm/realm-android-adapters/pull/53 
Anything in particular you are missing?

yuriy-larin:
I'm trying to switch from SQLite to Realm while I'm on early stage of project.
I have code with :
implements LoaderCallbacks<Cursor>
SimpleCursorAdapter
static class MyCursorLoader extends CursorLoader
trying to rewrite it for Realm.
As I understand, I have to use something like:
RealmResults<User> results = realm.where(User.class).findAll();
in Fragment class instead. Is that correct?

yuriy-larin:
What is also not quite clear for me from documentation...
How to fetch limited number of instances, something equal to 'LIMIT 10' functionality?

cmelchior:
`realm.where(User.class).findAllAsync()` is better as that will run the query on a background thread. No need for loaders in that case.

We don't support LIMIT yet, but since RealmResults are lazy-loaded, you normally don't need it anyway. You can use `results.subList()` instead though if you want to restrict the result.

yuriy-larin:
Thank you for suggestion.

Today I also stumbled upon the limitation trying to extend class from external library.
/realm/realm-java/issues/761
Inheritance / Polymorphism
I had to make copy of code instead.

cmelchior:
Closing this PR as it is now being tracked https://github.com/realm/realm-android-adapters/issues/6 instead

PR #1096, opened 2015-05-07
Fix Date and byte[] being reset when using createOrUpdateFromJson

### LINKED ISSUE #1095: Date field gets reset to default value on createOrUpdateAllFromJson
Unlike suggested by the changelog ("0.80.1 Realm.createOrUpdateWithJson() no longer resets fields to their default value if they are not found in the JSON input") createOrUpdateAllFromJson resets Date fields to 0 (beginning of time) when the corresponding value is missing in the input JSON

cmelchior:
Fixes #1095 

Date and byte[] fields where incorrectly reset even though this bug was fixed in the last release.
Lesson learned: Remember to unit test all types :(

@emanuelez @kneth

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,3 +1,6 @@
+0.81.0
+ * Fixed bug where Realm.createOrUpdateWithJson() reset Date and Binary data to default values if not found in the JSON output.
+
 0.80.2
  * Trying to use Realm.copyToRealmOrUpdate() with an object with a null primary key now throws a proper exception.
  * RealmMigrationNeedException can now return the path to the Realm that needs to be migrated.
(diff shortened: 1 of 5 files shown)


kneth:
:+1:

emanuelez:
Did you actually add tests for all the types? It's not obvious from the diff.

cmelchior:
Yes, I piggybacked on another test, so it is perhaps a bit hidden: https://github.com/realm/realm-java/pull/1096/files#diff-ae6501e798ea7a8fb5767a69c5a9532cR507

emanuelez:
In that case :+1:

PR #1097, opened 2015-05-07
Update Gradle to version 2.4

emanuelez:
@cmelchior @kneth


kneth:
@emanuelez did you measure the speedup for our builds?

emanuelez:
I did on my local machine. The actual build is not improved by much, but the initial projects evaluation is waaayyyyy faster, especially in conjunction with the Gradle daemon (which is on by default when using Android Studio).

On a full no-op build it goes from 7 to 4 seconds

kneth:
Sounds really good. Will a multicore build machine make it possible to enable parallel builds?

emanuelez:
Parallel builds only affects native code, but we still use make for JNI so it does not affect us

kneth:
:+1:

cmelchior:
:+1:

Issue #777, opened 2015-01-22
Default values for fields / assignment in default constructor


erichkleung:
Question/request: 
Will the java version have default values for fields similar to the iOS one? 

i.e.
+defaultPropertyValues can be overriden to provide default values every time an object is created.

This might also helps to alleviate the lack of null support in the interim (and with the standalone object support now, helps with gson deserialization a bunch) 

Thanks for all the awesome work on this!

cmelchior:
Hi @erichkleung 
We actually havn't considered it yet. It might be slightly more tricky to implement in Java especially for stand alone objects as we don't do any magic on those. I will add it to our TODO to look into, but can't make any promises.

Ralphilius:
+1

I got a question. If my JSON does not contain a boolean field, what is the value in the database? Null? Current verison is 0.84

cmelchior:
@Ralphilius It depends on what methods you are using

1) If you are creating a new object, it will get the default value for that datatype, e.g. `false` for `boolean` and `null` for Boolean.

2) If you are updating an existing object, it will just ignore that field, so it will keep whatever value it already had.

Ralphilius:
Thanks @cmelchior for clarifying

My case turns into number 1.

cmelchior:
Proposed API:

```
@DefaultValue(42)
private int foo

@DefaultValue("foo")
private String foo
```

Would not work for `RealmObject` and `RealmList` references. Should probably be disallowed for fields annotated with `@PrimaryKey`.

Cocoa doc is here: https://realm.io/docs/objc/latest/#default-property-values

We need to align on the behaviour across bindings. Especially if the default values should be part of the Realm file schema information

beeender:
Does annotation work with `Date` as well?

cmelchior:
Ideally yes, but it looks like there is a major limitation for annotations. It looks like we cannot do:
```
public @interface DefaultValue {
    Object value();
}
```

That pretty much makes my proposal impossible :(

beeender:
More thoughts for the default value
* Consistency with the creating standalone object by `new`
* Do we have to maintain a history of default values for migration? Especially for the primary key fields.

cmelchior:
After we merged our RealmTransformer we should be able to just use normal Java constructors + field assignment for default values instead of adding a new annotation:

The below is a request for such a feature:

----


#### Expected Results

[...]

PR #1098, opened 2015-05-07
Support for null

### LINKED ISSUE #478: Support for Null values
I want to be able to soft-delete items with a created and a deleted field, both of which being Date fields. When I tried creating an object (with no deleted flag), I received this error message:

```
java.lang.IllegalArgumentException: Null Date is not allowed.
            at io.realm.internal.Row.setDate(Row.java:195)
            at io.realm.RealmDataFileRealmProxy.setEnd(RealmDataFileRealmProxy.java:53)
```

Soft deletes like this are very common in databases, so it seems reasonable to support them. Are there any plans to support this already in the works?

kneth:
Fixes #478

Support for null values is a popular feature request (see #478). Realm Core has an experimental/work-in-progress version with support for null strings. In order to validate the concept of null strings, Realm for Android is experimentally adding support for it.

A new annotation (`@Required`) is introduced. `String`s and binary data are nullable by default, but you can specify that it is required (null is not an allowed value). Moreover, primary keys cannot be null.

Querying with null has a special semantic. Every string begins with, ends with and contains null strings.

If a field is nullable, the default value when creating an object is null while a required string field is the empty string (`""`) in Realm. The JSON methods in Realm and the proxy classes follow these rules too. The default value of a field can be changed in the model class when #777 is implemented.

**Note**: This branch will _not_ be merged into master for a while.

TODO:
 - [x] Basic support for strings (insert, retrieve, query)
 - [x] Basic support for binary data (insert, retrieve)
 - [ ] Basic support for booleans
 - [ ] Basic support for integers
 - [ ] Basic support for dates
 - [x] Migration
 - [x] JSON
 - [ ] Sorting on nullable String columns doesn't work


realm-annotations-processor/src/main/java/io/realm/processor/ClassMetaData.java
```diff
@@ -55,6 +56,7 @@
     private List<String> fieldNames = new ArrayList<String>();
     private List<String> ignoreFieldNames = new ArrayList<String>();
     private List<VariableElement> indexedFields = new ArrayList<VariableElement>(); // list of all fields marked @Index.
+    private List<String> nullableFieldNames = new ArrayList<String>(); // list of all fields marked @Nullable
```
emanuelez: I think this would benefit in being a `Set<VariableElement>` instead

realm-annotations-processor/src/main/java/io/realm/processor/ClassMetaData.java
```diff
@@ -289,6 +291,23 @@ private boolean categorizeClassElements() {
                     }
                 }
 
+                if (variableElement.getAnnotation(NotNullable.class) == null) {
+                    // The field has not the @NotNullable annotation. It's only valid for:
+                    // * String
+                    String elementTypeCanonicalName = variableElement.asType().toString();
```
emanuelez: The proper way to do this comparison would be to get the TypeMirror object for String like this:
```
TypeMirror stringType = env.getElementUtils().getTypeElement("java.lang.String").asType();
```
and then using that for comparison instead of comparing strings everywhere

realm-annotations-processor/src/main/java/io/realm/processor/ClassMetaData.java
```diff
@@ -289,6 +291,23 @@ private boolean categorizeClassElements() {
...
+                    if (elementTypeCanonicalName.equals("java.lang.String")) {
+                        nullableFieldNames.add(variableElement.getSimpleName().toString());
+                    }
+                } else {
+                    // The field has the @NotNullable annotation. It's only valid for:
+                    // * String
+                    String elementTypeCanonicalName = variableElement.asType().toString();
+                    String variableElementName = variableElement.getSimpleName().toString();
+                    if (elementTypeCanonicalName.equals("java.lang.String") && nullableFieldNames.contains(variableElementName)) {
```
emanuelez: Same as above

vvictor10:
Thanks for providing this update, very helpful. However "This branch will not be merged into master for a while." still keeps it ambiguous. There are so many key features that have these ambiguous status and I am in a deadlock situation now with many projects. Can you guys please share some resemblance of a roadmap to address these key issues. Things which are blocking me right now are the Robolectric support, NULL support, Migration API and cascade deletes. Thanks a lot again, appreciate the great responsiveness.

realm-annotations-processor/src/main/java/io/realm/processor/ClassMetaData.java
emanuelez: Split this line

realm-annotations-processor/src/main/java/io/realm/processor/ClassMetaData.java
```diff
@@ -299,6 +318,11 @@ private boolean categorizeClassElements() {
                         return false;
                     }
 
+                    // A field cannot be nullable and primary key. Currently only String can be nullable
+                    if (Utils.isString(variableElement) && nullableFieldNames.contains(variableElement.getSimpleName().toString())) {
```
emanuelez: Split this line

realm-annotations-processor/src/main/java/io/realm/processor/ClassMetaData.java
```diff
@@ -386,6 +410,10 @@ public String getPrimaryKeyGetter() {
         return getters.get(primaryKey.getSimpleName().toString());
     }
 
+    public boolean isNullable(String fieldName) {
```
emanuelez: maybe rename to `isFieldNullable` and consider if it should take a VariableElement instead

realm-jni/src/io_realm_internal_Row.cpp
```diff
@@ -284,6 +285,12 @@ JNIEXPORT void JNICALL Java_io_realm_internal_Row_nativeSetString
         return;
 
     try {
+        if (value == NULL) {
+            if (!(ROW(nativeRowPtr)->get_table()->is_nullable( S(columnIndex) ))) {
```
emanuelez: is it more efficient having both condition in the same statement?

realm-jni/src/io_realm_internal_table.cpp
```diff
@@ -628,6 +632,13 @@ JNIEXPORT void JNICALL Java_io_realm_internal_Table_nativeSetString(
     if (!TBL_AND_INDEX_AND_TYPE_VALID(env, TBL(nativeTablePtr), columnIndex, rowIndex, type_String))
         return;
     try {
+        if (value == NULL) {
+            if (!TBL(nativeTablePtr)->is_nullable(S(columnIndex))) {
+                ThrowException(env, IllegalArgument, "Trying to set field to null but field is not nullable.");
```
emanuelez: "Trying to set a non-nullable field to null" sounds better to me

realm-jni/src/io_realm_internal_tableview.cpp
```diff
@@ -385,6 +385,10 @@ JNIEXPORT void JNICALL Java_io_realm_internal_TableView_nativeSetString(
         if (!VIEW_VALID_AND_IN_SYNC(env, nativeViewPtr) ||
             !INDEX_AND_TYPE_VALID(env, TV(nativeViewPtr), columnIndex, rowIndex, type_String))
             return;
+        if (!TV(nativeViewPtr)->get_parent().is_nullable(S(columnIndex))) {
+            ThrowException(env, IllegalArgument, "Trying to set field to null but field is not nullable.");
```
emanuelez: Same as above

realm/src/androidTest/java/io/realm/RealmObjectTest.java
```diff
@@ -510,4 +511,40 @@ public void testIsValidManagedObject() {
...
+
+    // try to store null string in non-nullable field
+    public void testNotNullableField() {
+        try {
+            testRealm.beginTransaction();
+            NullTypes nullTypes = testRealm.createObject(NullTypes.class);
+            nullTypes.setFieldStringNotNull(null);
+            fail();
+        }
+        catch (IllegalArgumentException ignore) {
```
emanuelez: rename it to "expected" for readability

realm/src/androidTest/java/io/realm/RealmObjectTest.java
```diff
@@ -510,4 +511,40 @@ public void testIsValidManagedObject() {
...
     }
+
+    // store and retrieve null strings
+    public void testNullString() {
+        testRealm.beginTransaction();
+        NullTypes nullTypes = testRealm.createObject(NullTypes.class);
+        nullTypes.setFieldStringNull(null);
+        testRealm.commitTransaction();
+
+        assertNull(testRealm.where(NullTypes.class).findAll().first().getFieldStringNull());
```
emanuelez: Why `.finalAll().first()` instead of `.findFirst()`?

realm/src/androidTest/java/io/realm/RealmQueryTest.java
```diff
@@ -459,4 +473,45 @@ public void testGeorgian() {
             assertEquals(sorted[i], stringOnlies2.get(i).getChars());
         }
     }
+
+    public void testQueryNullStringNotNullableField() {
+        try {
+            RealmResults<NullTypes> nullTypeses = testRealm.where(NullTypes.class).equalTo(NullTypes.FIELD_STRING_NOT_NULL,
+                    (String)null).findAll();
```
emanuelez: Why the casting to String?

realm/src/androidTest/java/io/realm/RealmResultsTest.java
```diff
@@ -700,6 +714,30 @@ public void testSubList() {
...
 
+    public void testNullStringNotNullableField() {
+        populateTestRealmForNullTests();
+        RealmResults<NullTypes> list = testRealm.allObjects(NullTypes.class);
+        try {
+            testRealm.beginTransaction();
+            list.first().setFieldStringNotNull(null);
+            fail();
+        }
+        catch (IllegalArgumentException ignore) {
```
emanuelez: expected

emanuelez:
@vvictor10 As you can see we are working hard on the features you desire but support and bug fixing has a higher priority, which makes promising dates impossible.

To answer your specific concerns:
 * Roboelectric: this is a problem in roboelectric not supporting native libraries... not really a Realm problem in itself.
 * Null support: as you can see with this PR we're actively working on null support for Strings. This will be merged into master when complete and reviewed and released shortly after. Support for null for the other types should hopefully come in the following weeks
 * Migration API: Another hot topic we are working around the clock on. It is a big change which requires extra care but it's ongoing pretty fast as you can see if you are following the many PRs relative to it.
 * Cascading deletes: This needs to be implemented in core and will take longer to achieve

vvictor10:
@emanuelez Thanks a lot for the detailed update.

realm/src/androidTest/java/io/realm/RealmQueryTest.java
kneth: Yes, as the compiler cannot deduce if I'm calling the `String` or `Date` version of the method.

realm-annotations-processor/src/main/java/io/realm/processor/ClassMetaData.java
cmelchior: You can also use Utils.isString(VariableElement)

realm-annotations-processor/src/test/resources/some/test/NullTypes.java
```diff
@@ -0,0 +1,45 @@
+/*
+ * Copyright 2014 Realm Inc.
```
cmelchior: 2015

realm-annotations/src/main/java/io/realm/annotations/NotNullable.java
```diff
@@ -0,0 +1,21 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.realm.annotations;
+
+public @interface NotNullable {
```
cmelchior: Why @NotNullable instead of @NotNull? All other frameworks use @NotNull, so it would be more familiar to people.

realm-jni/src/io_realm_internal_Row.cpp
```diff
@@ -284,6 +285,12 @@ JNIEXPORT void JNICALL Java_io_realm_internal_Row_nativeSetString
...
+                ThrowException(env, IllegalArgument, "Trying to set field to null but field is not nullable.");
```
cmelchior: Also print field name for better debug info.

realm-jni/src/io_realm_internal_TableQuery.cpp
```diff
@@ -671,6 +671,12 @@ enum StringPredicate {
 void TableQuery_StringPredicate(JNIEnv *env, jlong nativeQueryPtr, jlongArray columnIndexes, jstring value, jboolean caseSensitive, StringPredicate predicate) {
     GET_ARRAY()
     try {
+        if (value == NULL) {
+            if (!(Q(nativeQueryPtr)->get_table()->is_nullable( S(arr[0]) ))) {
+                ThrowException(env, IllegalArgument, "Trying to query field with null but field is not nullable.");
```
cmelchior: Output fieldname for better debug info

realm-jni/src/io_realm_internal_TableQuery.cpp
```diff
@@ -696,6 +702,12 @@ void TableQuery_StringPredicate(JNIEnv *env, jlong nativeQueryPtr, jlongArray co
         }
         else {
             Table* tbl = getTableLink(nativeQueryPtr, arr, arr_len);
+            if (value == NULL) {
+                if (tbl->is_nullable(S(arr[arr_len-1]))) {
+                    ThrowException(env, IllegalArgument, "Trying to query field with null but field is not nullable.");
```
cmelchior: Same

realm-annotations/src/main/java/io/realm/annotations/NotNullable.java
emanuelez: It serves a different purpose. One might want to use both `NotNull` (for their own logic) and `NotNullable` (for Realm) so I think having different names will avoid some confusion and mistakes.

realm-jni/src/io_realm_internal_TableQuery.cpp
```diff
@@ -696,6 +702,12 @@ void TableQuery_StringPredicate(JNIEnv *env, jlong nativeQueryPtr, jlongArray co
         }
         else {
             Table* tbl = getTableLink(nativeQueryPtr, arr, arr_len);
```
cmelchior: It is very unclear what the macro GET_ARRAY actually returns. Is there a better name for it? including these two variables?

realm-jni/src/io_realm_internal_TableQuery.cpp
```diff
@@ -671,6 +671,12 @@ enum StringPredicate {
 void TableQuery_StringPredicate(JNIEnv *env, jlong nativeQueryPtr, jlongArray columnIndexes, jstring value, jboolean caseSensitive, StringPredicate predicate) {
     GET_ARRAY()
     try {
+        if (value == NULL) {
+            if (!(Q(nativeQueryPtr)->get_table()->is_nullable( S(arr[0]) ))) {
+                ThrowException(env, IllegalArgument, "Trying to query field with null but field is not nullable.");
+                return;
+            }
+        }
         bool isCaseSensitive = caseSensitive ? true : false;
```
cmelchior: What is the style guide for variables. So far I thought we used ```camelCase``` for java variabels and ```under_score_case``` for c++ variabels?

realm-jni/src/io_realm_internal_table.cpp
cmelchior: Output field name

realm-jni/src/io_realm_internal_tableview.cpp
cmelchior: Same as above :)

realm-jni/src/util.cpp
```diff
@@ -278,6 +278,10 @@ string concat_stringdata(const char *message, StringData strData)
 
 jstring to_jstring(JNIEnv* env, StringData str)
 {
+    if (str.is_null()) {
```
cmelchior: Whats the difference between ```str.is_null``` and ```str == NULL```?

realm-jni/src/util.cpp
```diff
@@ -278,6 +278,10 @@ string concat_stringdata(const char *message, StringData strData)
...
+        return NULL;
```
cmelchior: Shouldn't we set ```m_is_null = true``` here as well?

realm/src/androidTest/java/io/realm/RealmObjectTest.java
cmelchior: Pr out discussion on slack. This can also potentially throw a nullpointer :)

realm/src/androidTest/java/io/realm/RealmQueryTest.java
cmelchior: Instead we should allow ```RealmQuery.isNull()``` and ```RealmQuery.isNotNull()``` to also work on Strings.

realm/src/androidTest/java/io/realm/RealmQueryTest.java
```diff
@@ -459,4 +473,45 @@ public void testGeorgian() {
...
+    public void testQueryNullStringsContains() {
+        populateTestRealmForNullTests();
+        assertEquals("Fish", testRealm.where(NullTypes.class).contains(NullTypes.FIELD_STRING_NULL,
+                (String)null).findAll().first().getFieldStringNotNull());
+    }
+
+    public void testQueryNullStringsEndsWith() {
+        populateTestRealmForNullTests();
+        assertEquals("Fish", testRealm.where(NullTypes.class).endsWith(NullTypes.FIELD_STRING_NULL,
+                (String)null).findAll().first().getFieldStringNotNull());
```
cmelchior: use FindFirst() instead

realm/src/androidTest/java/io/realm/entities/NullTypes.java
```diff
@@ -0,0 +1,47 @@
+/*
+ * Copyright 2014 Realm Inc.
```
cmelchior: 2015

cmelchior:
The initial post that Json also now supports Null, but I see no changes in the code for Json?

realm-annotations/src/main/java/io/realm/annotations/NotNullable.java
cmelchior: How so? Eg. IntelliJ's @NotNull is a hint that this String can never be null, exactly the same as ours. I agree that if you want to use multiple definitions you have to be a bit careful, which could lead to very subtle errors:

```
public class Foo {
  @NotNull
  private String bar;

  private void setBar(@com.intellij.annotations.NotNulll bar) {
    this.bar = bar;
  }
}
```

Here are some other options: http://stackoverflow.com/questions/4963300/which-notnull-java-annotation-should-i-use

Another radical solution: We still have default values and renaming as upcoming features. Would it makes sense instead to do something like

````
@RealmField(defaultValue = "foo", nullable = false, name="internal_name_used_in_tables_and_queries") 
```

emanuelez: The first example you provide is exactly my point. Many developers will want to use both `NotNull` and `NotNullable` and even if it won't happen the same class, it still is confusing having two annotations with the same name but different function and use.

realm-jni/src/util.cpp
kneth: The `StringData` object is not null but it's payload (the string) might be.

realm-annotations-processor/src/main/java/io/realm/processor/ClassMetaData.java
```diff
@@ -289,6 +291,23 @@ private boolean categorizeClassElements() {
                     }
                 }
 
+                if (variableElement.getAnnotation(NotNullable.class) == null) {
+                    // The field has not the @NotNullable annotation. It's only valid for:
```
cmelchior: NotNullable -> Nullable

realm-annotations-processor/src/main/java/io/realm/processor/ClassMetaData.java
```diff
@@ -289,6 +291,23 @@ private boolean categorizeClassElements() {
...
+                    // * String
+                    if (Utils.isString(variableElement)) {
+                        nullableFieldNames.add(variableElement);
+                    }
+                } else {
+                    // The field has the @NotNullable annotation. It's only valid for:
```
cmelchior: NotNullable -> Nullable

realm-annotations-processor/src/main/java/io/realm/processor/ClassMetaData.java
```diff
@@ -289,6 +291,23 @@ private boolean categorizeClassElements() {
...
+                    // * String
+                    String elementTypeCanonicalName = variableElement.asType().toString();
+                    String variableElementName = variableElement.getSimpleName().toString();
+                    if (Utils.isString(variableElement)
+                            && nullableFieldNames.contains(variableElementName)) {
```
cmelchior: This is short enough to be a 1-liner?

realm-annotations-processor/src/main/java/io/realm/processor/ClassMetaData.java
```diff
@@ -299,6 +318,12 @@ private boolean categorizeClassElements() {
                         return false;
                     }
 
+                    // A field cannot be nullable and primary key. Currently only String can be nullable
+                    if (Utils.isString(variableElement)
+                            && nullableFieldNames.contains(variableElement)) {
+                        nullableFieldNames.remove(variableElement);
```
cmelchior: Is this correct. You are looping the variable elements and are inside a if/else. This particular variable element could not have been added yet?

realm-annotations-processor/src/test/java/io/realm/processor/RealmProcessorTest.java
```diff
@@ -80,6 +82,39 @@ public void compareProcessedSimpleFile() throws Exception {
     }
 
     @Test
+    public void compileSNullTypesFile() {
```
cmelchior: S -> s

realm-annotations-processor/src/test/java/io/realm/processor/RealmProcessorTest.java
```diff
@@ -80,6 +82,39 @@ public void compareProcessedSimpleFile() throws Exception {
...
+    @Test
+    public void compileProcessedNullTypesFile() throws Exception {
+        ASSERT.about(javaSource())
+                .that(nullTypesModel)
+                .processedWith(new RealmProcessor())
+                .compilesWithoutError();
+    }
+
+    @Test
+    public void compileNullTypesProxyFile() throws Exception {
```
cmelchior: We don't  usually test this, the compareProcessedNullTypesFile test should be enough

realm-annotations-processor/src/test/java/io/realm/processor/RealmProcessorTest.java
```diff
@@ -80,6 +82,39 @@ public void compareProcessedSimpleFile() throws Exception {
...
 
     @Test
+    public void compileSNullTypesFile() {
+        ASSERT.about(javaSource())
+                .that(nullTypesModel)
+                .compilesWithoutError();
+    }
+
+    @Test
+    public void compileProcessedNullTypesFile() throws Exception {
```
cmelchior: This is redundant. compareProcessedNullTypesFile tests everything

realm-annotations/src/main/java/io/realm/annotations/NotNullable.java
cmelchior: We need to make an decision here. Today would be a good time :)

realm/src/androidTest/java/io/realm/RealmQueryTest.java
```diff
@@ -459,4 +473,64 @@ public void testGeorgian() {
...
+        try {
+            RealmResults<NullTypes> nullTypeses = testRealm.where(NullTypes.class).equalTo(NullTypes.FIELD_STRING_NOT_NULL,
+                    (String)null).findAll();
+            fail();
+        }
+        catch (IllegalArgumentException ignore) {
+        }
+    }
+
+    public void testQueryNullStringsNotNullableFieldIsNull() {
```
cmelchior: This method name makes my head hurt, perhaps a comment above it?

realm-annotations-processor/src/main/java/io/realm/processor/ClassMetaData.java
kneth: Strings are by default nullable, and if I find a primary key, that particular field cannot be nullable. Therefore it is removed from the set of nullable fields.

If I process the annotation in the opposite order, I will have to check every string field if it is a primary key.

realm/src/androidTest/java/io/realm/RealmQueryTest.java
kneth: I tried to create methods with names explaining what it tested. But I agree: the name is really ugly. Adding comments to unit tests is good idea!

realm-annotations/src/main/java/io/realm/annotations/NotNullable.java
emanuelez: I really don't think we should add the the confusion of the `@NotNull` java annotation.

realm/src/androidTest/java/io/realm/RealmQueryTest.java
kneth: I have added string support for these methods.

realm-jni/src/io_realm_internal_TableQuery.cpp
kneth: I think that it's better to solve in another PR.

realm-jni/src/util.cpp
kneth: `to_string()` is an aux. function and not a method. It does not have access to `m_is_null`.

kneth:
**DO NOT USE THIS IN PRODUCTION**

I have created a jar file with partial null support. You can download it as http://static.realm.io/downloads/java/realm-0.80.3-null-support-alpha1.jar, and you have to manually copy the jar file to your app's libs folder and sync your Android Studio project (see also http://realm.io/docs/java/latest/#installation).

This is a early release of the null support: please take notice of the "alpha1" in the file name :-) As Realm Core still works on supporting null, it hasn't been fully tested, some feature might be missing, and performance might not be optimal.

With this release, string and byte arrays are nullable by default. If you wish to mark a string or byte array not nullable, you can use the annotation `@NotNullable`. You do not have to specify that a primary key is not-nullable; not nullable is implied by being a primary key.

Types like boolean, integer, float, double and dates are still being worked on by the Realm Core team, and fields of these types are always not-nullable. We still have some work on the JSON and migration parts of Realm.

The internal file format of Realm files are changed. Using this release will irreversibly change your Realm files!

**DO NOT USE THIS IN PRODUCTION**

realm-annotations-processor/src/test/resources/io/realm/SimpleRealmProxy.java
```diff
@@ -67,8 +67,8 @@ public void setAge(int value) {
     public static Table initTable(ImplicitTransaction transaction) {
         if (!transaction.hasTable("class_Simple")) {
             Table table = transaction.getTable("class_Simple");
-            table.addColumn(ColumnType.STRING, "name");
-            table.addColumn(ColumnType.INTEGER, "age");
+            table.addColumn(ColumnType.STRING, "name", true);
```
bmunkholm: Add constants NULLABLE, NOT_NULLABLE and use instead of false/true.

kneth:
I have uploaded a few version of Realm with partial null support (string and binary only). You find it at http://static.realm.io/downloads/java/realm-0.81-null-support-beta1.jar

bmunkholm:
Fixes #478

realm/src/androidTest/java/io/realm/RealmResultsTest.java
```diff
@@ -30,10 +30,11 @@
 import io.realm.entities.Cat;
 import io.realm.entities.Dog;
 import io.realm.entities.NonLatinFieldNames;
+import io.realm.entities.NullTypes;
 import io.realm.entities.Owner;
 
 public class RealmResultsTest extends AndroidTestCase {
-    protected final static int TEST_DATA_SIZE = 2516;
+    protected final static int TEST_DATA_SIZE = 100; // TODO: set to 2516 when background GC works
```
beeender: What does this mean?

emanuelez: I guess this should be changed back to 2516 since our GC has been merged in.

realm-jni/src/util.cpp
```diff
@@ -166,6 +166,16 @@ void jprint(JNIEnv *env, char *txt)
...
 }
 
+void ThrowNullValueException(JNIEnv* env, Table* table, size_t col_ndx) {
+    std::ostringstream ss;
+    ss << "Trying to set a non-nullable field '"
+       << table->get_column_name(col_ndx)
+       << "' in '"
+       << table->get_name()
+       << "' to null.";
+    ThrowException(env, IllegalArgument, ss.str());
```
beeender: What about throw NullPointerException instead? For boxed types, NullPotinerException will be send when cast null to boolean/long. For IllegalArgumentException, extra code needs to be written which can slow down the setter a little bit.

beeender:
JSON with null value is not handled properly in the Proxy. Now only newly created Object will have the expected value (null, or empty string). When update, the value is not changed since we are using `JSONObject.isNull()` without checking `JSONObject.has()` first.

beeender:
:+1: It would be easier to release null support for String and byte[] first. Boxed types support is on the way. Otherwise it will be a huge PR and really hard to review.

beeender:
retest this please

kneth:
@beeender I agree that reviewing all null support changes might be too much. As this PR lay the fundation, it might be a good idea to review it first and merge it. And then quickly thereafter, review the other types and merge them.

kneth:
@beeender And nice of you to fix the JSON part ;-)

HPIWicket:
@kneth Is there the possibility to get a snapshot of the NULL support via gradle?

kneth:
@HPIWicket Currently it is not possible to get a snapshot via Gradle.

cmelchior:
@kneth Has this been superseded by #1396 ?

beeender:
@cmelchior , yes all the changes here has been merged to #1396

cmelchior:
Ok, so lets just close this one?

beeender:
Close this one for easier tracking. All changes here have been merged to #1396 .

PR #1099, opened 2015-05-08
Accept null RealmResults for RealmBaseAdapter. refs #1086

### LINKED ISSUE #1086: RealmBaseAdapter with null RealmResults
Can you please handle null `realmResults` within `RealmBaseAdapter`? If the adapter has no content at all, and `RealmQuery` possibly returns `null` too, it doesn't make sense for `RealmBaseAdapter` to not accepting `null`.

Example:

    RealmResults<MyObject> results = realm.where("...").find(); // it's possible the result is null
    MyRealmAdapter adapter = new MyRealmAdapter(context, results); // if results is null, you get exception
    // if I catch the exception, then adapter is null
    // I couldn't do this
    spinner.setAdapter(adapter);
    // or
    adapter.updateRealmResults(notNullRealmResults);

zaki50:
Fixes #1086

### DIFF
--- a/realm/src/main/java/io/realm/RealmBaseAdapter.java
+++ b/realm/src/main/java/io/realm/RealmBaseAdapter.java
@@ -36,35 +36,40 @@ public abstract class RealmBaseAdapter<T extends RealmObject> extends BaseAdapte
     protected LayoutInflater inflater;
     protected RealmResults<T> realmResults;
     protected Context context;
+    private final RealmChangeListener listener;
 
     public RealmBaseAdapter(Context context, RealmResults<T> realmResults, boolean automaticUpdate) {
         if (context == null) {
             throw new IllegalArgumentException("Context cannot be null");
         }
-        if (realmResults == null) {
-            throw new IllegalArgumentException("RealmResults cannot be null");
-        }
-
         this.context = context;
         this.realmResults = realmResults;
         this.inflater = LayoutInflater.from(context);
-        if (automaticUpdate) {
-            realmResults.getRealm().addChangeListener(new RealmChangeListener() {
-                @Override
-                public void onChange() {
-                    notifyDataSetChanged();
-                }
-            });
+        this.listener = (!automaticUpdate) ? null : new RealmChangeListener() {
+            @Override
+            public void onChange() {
+                notifyDataSetChanged();
+            }
+        };
+
+        if (listener != null && realmResults != null) {
+            realmResults.getRealm().addChangeListener(listener);
         }
     }
 
     @Override
     public int getCount() {
+        if (realmResults == null) {
+            return 0;
+        }
         return realmResults.size();
     }
 
     @Override
     public T getItem(int i) {
+        if (realmResults == null) {
+            return null;
+        }
         return realmResults.get(i);
     }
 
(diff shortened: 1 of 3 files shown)


zaki50:
Should we skip unnecessary listener update in updateRealmResults for better performance if target Realm instance is not changed?

realm/src/main/java/io/realm/RealmBaseAdapter.java
```diff
@@ -89,7 +94,20 @@ public long getItemId(int i) {
...
+            }
+            if (realmResults != null) {
+                realmResults.getRealm().addChangeListener(listener);
+            }
+        }
+
         this.realmResults = realmResults;
-        notifyDataSetChanged();
+        if (this.realmResults == null) {
+            notifyDataSetInvalidated();
```
cmelchior: I am not sure that calling this here would be correct. JavaDoc says that once this has been called the adapter should not be used anymore, which is not correct in this case, it could become "valid" again by just updating the results agains.

realm/src/main/java/io/realm/RealmBaseAdapter.java
```diff
@@ -89,7 +94,20 @@ public long getItemId(int i) {
      * @param realmResults the new RealmResults coming from the new query.
      */
     public void updateRealmResults(RealmResults<T> realmResults) {
+        if (listener != null) {
```
cmelchior: This section could use a comment along the lines of "Making sure that Adapter is refreshed correctly if new RealmResults come from another Realm"

realm/src/main/java/io/realm/RealmBaseAdapter.java
```diff
@@ -89,7 +94,20 @@ public long getItemId(int i) {
...
+            if (this.realmResults != null) {
+                this.realmResults.getRealm().removeChangeListener(listener);
+            }
+            if (realmResults != null) {
```
cmelchior: Using both ```this.realmResults``` and ```realmResults``` is a bit confusing to read. A different name for the method variable would increase readability. Perhaps just "results" or "queryResults"?

realm/src/main/java/io/realm/RealmBaseAdapter.java
zaki50: You are right.
I'll update my PR.

cmelchior:
Hi @zaki50 
Thank you very much. I only had minor comments. You are allowing the adapter to be updated with data from another Realm, but I guess it is OK.

Also please update the changelog.

cmelchior:
You are right that the ChangeListener currently gets triggered on any change, which is not very optimal. We are working on more fine-grained notifications, so until they are implemented I don't think we should handle it in this class.

realm/src/main/java/io/realm/RealmBaseAdapter.java
zaki50: You mean parameter name of this method should be changed?

cmelchior: Yes

cmelchior:
Looks good now :+1: 

@emanuelez @kneth Do you have any comments?

emanuelez:
It looks very good. My only comment is purely about coding style: the empty lines in the unit tests.
Great job @zaki50! And thank you for your contribution.

realm/src/main/java/io/realm/RealmBaseAdapter.java
```diff
@@ -86,10 +91,20 @@ public long getItemId(int i) {
...
      * If the query does not change you might consider using the automaticUpdate feature
      *
-     * @param realmResults the new RealmResults coming from the new query.
+     * @param queryResults the new RealmResults coming from the new query.
      */
-    public void updateRealmResults(RealmResults<T> realmResults) {
-        this.realmResults = realmResults;
+    public void updateRealmResults(RealmResults<T> queryResults) {
+        if (listener != null) {
+            // Making sure that Adapter is refreshed correctly if new RealmResults come from another Realm
```
kneth: I think you should write "another Realm instance" as "another Realm" make it sounds like it can come from another Realm file.

kneth:
Besides my little comment on a comment, it looks fine to me.

:+1:

realm/src/main/java/io/realm/RealmBaseAdapter.java
cmelchior: Actually it can, which is why it is necessary to add and remove the listener ;)

cmelchior:
Hi @zaki50 
If you just remove the empty lines in your unit tests, this will be ready to merge.

zaki50:
@cmelchior I removed empty lines and rebased against master branch.

cmelchior:
Thanks a lot @zaki50. It has been merged.

PR #1115, opened 2015-05-12
Fix a memory leak in RealmBaseAdapter

### LINKED ISSUE #1109: Possibility of a memory leak due to a combination of `RealmChangeListener` and `automaticUpdate`
Reported by @zaki50 

`RealmChangeListener` is added to `Realm` when `automaticUpdate` is true. Then the listener instance was not released until the `Realm` instance is released because the `RealmChangeListener` is never deleted.

Since a listener instance is an anonymous inner class of `RealmBaseAdapter`, it holds a reference to the `RealmBaseAdapter` as an enclosing instance. `RealmBaseAdapter` usually holds a reference to the `Activity` that uses it as `Context`. So these instances will not be released until the `Realm` instance is released.

If you use `Realm` in only one Activity, there is no problem because `Realm#close()` is called when the `Activity`  is destroyed. (Strictly speaking, adapters may be accumulated if re-create a `RealmBaseAdapter` while the Activity is alive.)

On the other hand, if you use two activities, there are main activity and another activity that has `ListView`, each activity uses `Realm`, the `Realm` instance is held in cache until `onDestroy` of main activity is called. For example if you transition between the two Activities as follows, Main -> List-> back -> List -> back -> List... I think the `RealmChangeListener` may be accumulated in the `Realm`.

What do you think? @cmelchior @emanuelez

emanuelez:
Fixes #1109
We now use WeakReferences to store the change listeners.
This allows the Garbage Collector to clean up the adapters.

### DIFF
--- a/realm/src/main/java/io/realm/Realm.java
+++ b/realm/src/main/java/io/realm/Realm.java
@@ -34,6 +34,7 @@ import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
+import java.lang.ref.WeakReference;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
@@ -42,6 +43,7 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Scanner;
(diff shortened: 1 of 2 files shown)


emanuelez:
Changelog:
 * Fixed a memory leak when using RealmBaseAdapter

emanuelez:
@cmelchior @kneth @bmunkholm

realm/src/main/java/io/realm/Realm.java
```diff
@@ -1408,9 +1410,14 @@ public void removeAllChangeListeners() {
     }
 
     void sendNotifications() {
-        List<RealmChangeListener> defensiveCopy = new ArrayList<RealmChangeListener>(changeListeners);
-        for (RealmChangeListener listener : defensiveCopy) {
```
cmelchior: I guess removing the defensive copy is OK since Realm instances are thread local anyway?

cmelchior:
:+1:

realm/src/main/java/io/realm/Realm.java
emanuelez: Exactly

kneth:
:+1:

PR #1117, opened 2015-05-12
RealmList methods now fail if not within a write transaction

### LINKED ISSUE #1112: Possible to call RealmList.clear() outside a write transaction
It looks like it is possible to call RealmList.clear() outside a write transaction which can cause core to crash in some cases. Unit test exposing the issue is here: https://github.com/realm/realm-java/tree/cm-bug-clear.

There should be unit tests for all functions that modify Realm

cmelchior:
Fixes #1112 

All RealmList methods that require to be inside a write transaction now throw an exception if they are not. Also added unit tests for Realm and RealmResults.

@emanuelez @kneth


kneth:
Great to see the checks!

realm/src/androidTest/java/io/realm/RealmTest.java
```diff
@@ -1627,4 +1633,47 @@ public void testDeleteNonRealmFile() throws IOException {
...
+        InputStream jsonObjStream = TestHelper.stringToStream(jsonObjStr);
+        InputStream jsonObjStream2 = TestHelper.stringToStream(jsonObjStr);
+
+        String jsonArrStr = " [{ \"columnLong\" : 1 }] ";
+        JSONArray jsonArr = new JSONArray(jsonArrStr);
+        InputStream jsonArrStream = TestHelper.stringToStream(jsonArrStr);
+        InputStream jsonArrStream2 = TestHelper.stringToStream(jsonArrStr);
+
+        // Test all methods that should require a write transaction
+        try { testRealm.createObject(AllTypes.class);   fail("createObject() did not throw"); }             catch (IllegalStateException expected) {}
```
emanuelez: Whoa what a wall of text! Is this really more readable compared to its standard form?

realm/src/androidTest/java/io/realm/RealmListTest.java
```diff
@@ -53,7 +53,8 @@ protected void setUp() throws Exception {
     }
 
     // Check that all methods work correctly on a empty RealmList
-    private void testMethodsOnEmptyList(RealmList<Dog> list) {
+    private void testMethodsOnEmptyList(Realm realm, RealmList<Dog> list) {
```
emanuelez: This method name should not begin with "test". The instrumentator might be smart enough to recognize it's not an actual test, but it makes it a bit harder for humans IMO

realm/src/androidTest/java/io/realm/RealmTest.java
cmelchior: @bmunkholm thinks so, and I tend to agree. It depends on what is the most important thing in the code. In this case it is in a brief glance being able to see which methods are being tested. The standard formatting makes this a lot harder.

realm/src/androidTest/java/io/realm/RealmTest.java
```diff
@@ -1627,4 +1633,47 @@ public void testDeleteNonRealmFile() throws IOException {
...
+        try { testRealm.createObjectFromJson(AllTypesPrimaryKey.class, jsonObjStr);             fail("createObjectFromJson(str) did not throw"); }              catch (RealmException expected) {}
+        try { testRealm.createObjectFromJson(AllTypesPrimaryKey.class, jsonObjStream);          fail("createObjectFromJson(stream) did not throw"); }           catch (IllegalStateException expected) {}
+        try { testRealm.createOrUpdateObjectFromJson(AllTypesPrimaryKey.class, jsonObj);        fail("createOrUpdateObjectFromJson(obj) did not throw"); }      catch (IllegalStateException expected) {}
+        try { testRealm.createOrUpdateObjectFromJson(AllTypesPrimaryKey.class, jsonObjStr);     fail("createOrUpdateObjectFromJson(str) did not throw"); }      catch (IllegalStateException expected) {}
+        try { testRealm.createOrUpdateObjectFromJson(AllTypesPrimaryKey.class, jsonObjStream2);  fail("createOrUpdateObjectFromJson(stream) did not throw"); }   catch (IllegalStateException expected) {}
```
cmelchior: Indentation

realm/src/androidTest/java/io/realm/RealmTest.java
```diff
@@ -1627,4 +1633,47 @@ public void testDeleteNonRealmFile() throws IOException {
...
+
+        try { testRealm.createAllFromJson(AllTypesPrimaryKey.class, jsonArr);                   fail("createAllFromJson(arr) did not throw"); }                 catch (RealmException expected) {}
+        try { testRealm.createAllFromJson(AllTypesPrimaryKey.class, jsonArrStr);                fail("createAllFromJson(str) did not throw"); }                 catch (RealmException expected) {}
+        try { testRealm.createAllFromJson(AllTypesPrimaryKey.class, jsonArrStream);             fail("createAllFromJson(stream) did not throw"); }              catch (IllegalStateException expected) {}
+        try { testRealm.createOrUpdateAllFromJson(AllTypesPrimaryKey.class, jsonArr);           fail("createOrUpdateAllFromJson(arr) did not throw"); }         catch (RealmException expected) {}
+        try { testRealm.createOrUpdateAllFromJson(AllTypesPrimaryKey.class, jsonArrStr);        fail("createOrUpdateAllFromJson(str) did not throw"); }         catch (RealmException expected) {}
+        try { testRealm.createOrUpdateAllFromJson(AllTypesPrimaryKey.class, jsonArrStream2);     fail("createOrUpdateAllFromJson(stream) did not throw"); }      catch (IllegalStateException expected) {}
```
cmelchior: Indentation

emanuelez:
:+1:

realm/src/androidTest/java/io/realm/RealmTest.java
bmunkholm: I don't think the text in fail() is needed. That will reduce the wall a bit.
But yeah this would otherwise be hundreds LOC.

realm/src/main/java/io/realm/internal/LinkView.java
```diff
@@ -30,58 +33,52 @@ public LinkView(Context context, Table parent, long columnIndexInParent, long na
         this.nativeLinkViewPtr = nativeLinkViewPtr;
     }
 
-    protected static native void nativeClose(long nativeLinkViewPtr);
```
bmunkholm: I don't see the need to reorder the private methods. It's easier to have it just where it's used when you change it. And it's consistent with other files. (alternatively other files should also be reordered - which I don't think is warranted)

bmunkholm:
:1 except from noted.

realm/src/main/java/io/realm/internal/LinkView.java
kneth: We have discussed this a number of times now. @cmelchior finds it easier to read while @emanuelez and I have no strong opinion on the matter.

cmelchior: I was planing to update the other files as well but didn't wanted to do it as part of this PR. I will create a separate PR for the other files. 

Also there is multiple reasons for changing it:

1) IMO It becomes easier to scan a class when scrolling through it as flow isn't interrupted by one-liners.
2) While there normally is merit in keeping private methods close to where they are used, the native method declaration does not give any additional detail that can be used to reason about their behaviour. 
2) It becomes easier to compare c++ files with Java files.
3) Grouping them is consistent with the standard Java classes and every other framework I have seen that uses native code.

PR #1134, opened 2015-05-20
Fix sendNotifications called twice

beeender:
This is caused by 54501237b3 which doesn't seem to be right behaviour.

### DIFF
--- a/realm/src/androidTest/java/io/realm/NotificationsTest.java
+++ b/realm/src/androidTest/java/io/realm/NotificationsTest.java
@@ -323,6 +323,8 @@ public class NotificationsTest extends AndroidTestCase {
         realm.addChangeListener(new RealmChangeListener() {
             @Override
             public void onChange() {
+                // Listener should only be called once
+                assertFalse(success.get());
                 success.set(true);
             }
         });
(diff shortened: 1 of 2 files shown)


beeender:
And i would like to change the test case to cover this issue, but i could not run the test with android studio. 

    testCompile 'junit:junit:4.12'
    testCompile "org.mockito:mockito-core:1.9.5"

Those two lines fixed the mock issue, but still i got some 
java.lang.ClassNotFoundException: io.realm.DefaultRealmModuleMediator

I would appreciate any guide about running realm test cases.

emanuelez:
So you are trying to run the unit tests in the JVM instead on on-device. We don't really do that right now so I'm not sure about the hurdles ahead in that direction. Would it be possible for you to add the testcase in https://github.com/realm/realm-java/blob/master/realm/src/androidTest/java/io/realm/NotificationsTest.java ?

beeender:
Hmm, something wrong with my android studio. Spend hours on finding the place to run androidTest... Test case updated.

emanuelez:
Nice! Thank you! 
Could you please also update the changelog file?

Also, can you please sign our CLA? :)

https://github.com/realm/realm-java/blob/master/CONTRIBUTING.md

cmelchior:
Thank you for catching that @beeender 
@emanuelez There shouldn't be a need to update the changelog as the original change hasn't been released yet?

emanuelez:
@cmelchior true

beeender:
CLA signed

emanuelez:
:+1:

cmelchior:
:+1:

PR #1138, opened 2015-05-20
Added sorting and caseInsensitive limitations to JavaDoc.

### LINKED ISSUE #1069: [DOC] RealmQuery.CASE_INSENSITIVE is not insensitive for special character
When using the query below on the following text:
JØSANG SJØMAT

it does not get hits on jøs or Jøs , but hits on JØS and jØs

contains("name", searchEditText.getText().toString(), RealmQuery.CASE_INSENSITIVE)

Seems that it does not support non-english characters.

cmelchior:
Fixes #1069 

@emanuelez @kneth

### DIFF
--- a/realm/src/main/java/io/realm/RealmQuery.java
+++ b/realm/src/main/java/io/realm/RealmQuery.java
@@ -217,7 +217,8 @@ public class RealmQuery<E extends RealmObject> {
      * Equal-to comparison
      * @param fieldName   The field to compare
      * @param value       The value to compare with
-     * @param caseSensitive if true, substring matching is case sensitive
+     * @param caseSensitive if true, substring matching is case sensitive. Setting this to false only works for ASCII
+     *                      characters.
      * @return The query object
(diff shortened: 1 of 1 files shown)


realm/src/main/java/io/realm/RealmQuery.java
```diff
@@ -850,7 +850,8 @@ private boolean containsDot(String s) {
      * Condition that value of field contains the specified substring
      * @param fieldName  The field to compare
      * @param value      The substring
-     * @param caseSensitive if true, substring matching is case sensitive
+     * @param caseSensitive If true, substring matching is case sensitive. Setting this to false only works for ASCII
```
emanuelez: Why only ASCII only for this method? Is this true?

cmelchior: No, I forgot the other methods. beginsWith(), endsWith(), equalTo(). They are being added now

emanuelez: ASCII is the wrong term. English Locale is more correct

PR #1141, opened 2015-05-20
Do remove the RealmChangeListener weak reference

beeender:
Weak references on a same object are not same.

### DIFF
--- a/realm/src/androidTest/java/io/realm/NotificationsTest.java
+++ b/realm/src/androidTest/java/io/realm/NotificationsTest.java
@@ -19,6 +19,7 @@ import android.os.Handler;
 import android.os.Looper;
 import android.test.AndroidTestCase;
 
+import java.lang.ref.WeakReference;
 import java.util.Map;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ConcurrentHashMap;
@@ -114,7 +115,7 @@ public class NotificationsTest extends AndroidTestCase {
         assertEquals(0, counter.get());
     }
 
-    public void testAddDuplicatedListener() throws InterruptedException, ExecutionException {
+    public void testAddDuplicatedListener() {
         final AtomicInteger counter= new AtomicInteger(0);
         RealmChangeListener listener = new RealmChangeListener() {
             @Override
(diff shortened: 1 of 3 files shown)


beeender:
And no need to update the changelog.txt since the bug is not released yet.

cmelchior:
Hi @beeender 
Thanks for catching that. I mistakenly assumed that equals on WeakReference would use the referenced object when reviewing the first fix. Guess I should have followed my hunch of actually checking that :)

realm/src/androidTest/java/io/realm/NotificationsTest.java
```diff
@@ -94,6 +94,70 @@ public Boolean call() throws Exception {
         assertTrue(Realm.realmsCache.get().isEmpty());
     }
 
+    public void testRemoveNotifications () throws InterruptedException, ExecutionException {
```
cmelchior: This test case doesn't have to be this complicated. Something like this will also work:

```
        final AtomicInteger counter= new AtomicInteger(0);
        RealmChangeListener listener = new RealmChangeListener() {
            @Override
            public void onChange() {
                counter.incrementAndGet();
            }
        };

        realm = Realm.getInstance(getContext());
        realm.addChangeListener(listener);
        realm.removeChangeListener(listener);

        realm.beginTransaction();
        realm.commitTransaction();

        assertEquals(0, counter.get());
```

realm/src/main/java/io/realm/Realm.java
```diff
@@ -1206,7 +1206,12 @@ public void addChangeListener(RealmChangeListener listener) {
      */
     public void removeChangeListener(RealmChangeListener listener) {
         checkIfValid();
-        changeListeners.remove(new WeakReference<RealmChangeListener>(listener));
+        Iterator<WeakReference<RealmChangeListener>> iterator = changeListeners.iterator();
+        while (iterator.hasNext()) {
```
cmelchior: As you pointed out in a comment (I can't find anymore?) The current implementation has a chance of ConcurrentModificationException. This can still happen using this code:

eg. the following test case crashes:

        final AtomicInteger counter= new AtomicInteger(0);
        RealmChangeListener listener = new RealmChangeListener() {
            @Override
            public void onChange() {
                counter.incrementAndGet();
                realm.addChangeListener(this);
            }
        };

        realm = Realm.getInstance(getContext());
        realm.addChangeListener(listener);

        realm.beginTransaction();
        realm.commitTransaction();

        realm.beginTransaction();
        realm.commitTransaction();

        assertEquals(2, counter.get());

beeender:
It was pretty late yesterday, after i wrote the comments, i found it has nothing to do with that commit, so i deleted it :)

Actually i have some ideas to do something with te ConcurrentModificationException.
Choice 1. Throw a meaningful exception in addListener and removeListener, at least, user can easily figure out how does the ConcurrentModificationException happen.
Choice 2. Move the onChanged out of the listener iteration loop. This might be even more convenient  to user.

realm/src/main/java/io/realm/Realm.java
emanuelez: So I guess making a defensive copy of the list of weak references is still a good idea

beeender: Another thing, do you think check duplicated entries in addChangeListener would be a good idea? It would prevent some strange problem caused by user's mistake, and we can stop the loop immediately in removeChangeListener.

emanuelez: Yeah, I considered turning the List into a Set for this reason

beeender: Well... Using set doesn't solve the problem, since WeakReference doesn't override the equals method, iteration is still needed...

emanuelez: You are right

cmelchior:
Failing unit test:
```
io.realm.NotificationsTest > testNotificationsNumber[Nexus 9 - 5.0.1] [31mFAILED [0m
	junit.framework.AssertionFailedError: expected:<1> but was:<0>
	at io.realm.NotificationsTest.testNotificationsNumber(NotificationsTest.java:197)
```

realm/src/androidTest/java/io/realm/NotificationsTest.java
```diff
@@ -347,4 +388,76 @@ public void onChange() {
...
+
+        RealmChangeListener listener3 = new RealmChangeListener() {
+            @Override
+            public void onChange() {
+                counter3.incrementAndGet();
+                realm.removeChangeListener(this);
+            }
+        };
+
+        realm = Realm.getInstance(getContext());
```
cmelchior: Some comments here about the intermediates step would be nice, as the interaction between counters and listeners are very subtle.

realm/src/main/java/io/realm/Realm.java
```diff
@@ -159,7 +160,8 @@
     private SharedGroup sharedGroup;
     private final ImplicitTransaction transaction;
 
-    private final List<WeakReference<RealmChangeListener>> changeListeners = new ArrayList<WeakReference<RealmChangeListener>>();
+    private final List<WeakReference<RealmChangeListener>> changeListeners =
+            new CopyOnWriteArrayList<WeakReference<RealmChangeListener>>();
```
cmelchior: Can this not fit on a single line?

realm/src/main/java/io/realm/Realm.java
```diff
@@ -1195,6 +1197,13 @@ boolean contains(Class<? extends RealmObject> clazz) {
      */
     public void addChangeListener(RealmChangeListener listener) {
         checkIfValid();
+        for (WeakReference<RealmChangeListener> ref : changeListeners) {
+            if (ref.get() == listener) {
+                // It is already added before
```
cmelchior: "It has already been added before"

realm/src/main/java/io/realm/Realm.java
emanuelez: Two lines is fine. 120 chars should be our limit

cmelchior: I agree, couldn't tell from glance if this exceeded it

cmelchior:
Apart from a few stylistic comments, this looks good. Thanks @beeender

emanuelez:
Also please merge master into your branch to resolve the conflicts

realm/src/main/java/io/realm/Realm.java
beeender: Done

realm/src/androidTest/java/io/realm/NotificationsTest.java
beeender: Done

cmelchior:
:+1:  Awesome

emanuelez:
:+1:

PR #1146, opened 2015-05-21
Cursor example

cmelchior:
Supplements #1090 with a new example that shows how Realm with a Cursor can work together with Loaders and CursorAdapters as well as how one would go about creating auto-incremented primary keys.

This must not be merged before #1090 
- [x] Loader
- [x] CursorAdapter
- [ ] ContentProvider across apps. Might not work before multiprocess. Have to validate that.


cmelchior:
Outdated and moved to https://github.com/realm/realm-android-adapters/issues/6

PR #1148, opened 2015-05-22
Lazily create the toRemoveList

beeender:
To make it a bit faster.

### DIFF
--- a/realm/src/main/java/io/realm/Realm.java
+++ b/realm/src/main/java/io/realm/Realm.java
@@ -1250,18 +1250,22 @@ public final class Realm implements Closeable {
 
     void sendNotifications() {
         Iterator<WeakReference<RealmChangeListener>> iterator = changeListeners.iterator();
-        List<WeakReference<RealmChangeListener>> toRemoveList =
-                new ArrayList<WeakReference<RealmChangeListener>>(changeListeners.size());
+        List<WeakReference<RealmChangeListener>> toRemoveList = null;
         while (iterator.hasNext()) {
             WeakReference<RealmChangeListener> weakRef = iterator.next();
             RealmChangeListener listener = weakRef.get();
             if (listener == null) {
+                if (toRemoveList == null) {
+                    toRemoveList = new ArrayList<WeakReference<RealmChangeListener>>(changeListeners.size());
+                }
                 toRemoveList.add(weakRef);
             } else {
                 listener.onChange();
             }
         }
-        changeListeners.removeAll(toRemoveList);
+        if (toRemoveList != null) {
+            changeListeners.removeAll(toRemoveList);
+        }
     }
 
     @SuppressWarnings("UnusedDeclaration")


beeender:
Does small changes like this need to be in the changlog.txt?

emanuelez:
In theory every change that the user is exposed to needs to be there. In this case it would be something like:
 * Minor optimizations in the handling of notifications

or something like that

cmelchior:
In this case there is no user facing changes, so don't think it is needed. Just to be curious, have you benchmarked this? Because it seems you are a trading a constant cost (allocating memory) with a variable one (checking on each loop).

That said, I suspect that allocating memory far outweigh the boolean check for a realistic amount of listeners.

beeender:
Hmmm, i am curious as well. Some testing:
```java
    public void testLoop1() {
        long start = System.nanoTime();
        int runs = 1000*1000;
        for(int i=0;i<runs;i++) {
            // do test
            List<AllTypes> list;
            list = new ArrayList<AllTypes>(100);
        }
        long time = System.nanoTime() - start;
        System.out.printf("The average time for loop1 taken was %.1f ns%n", (double) time / runs);

        start = System.nanoTime();
        for(int i=0;i<runs;i++) {
            // do test
            List<AllTypes> list = null;
            if (list == null) {
                list = new ArrayList<AllTypes>(100);
            }
        }
        time = System.nanoTime() - start;
        System.out.printf("The average time for loop2 taken was %.1f ns%n", (double) time / runs);

        start = System.nanoTime();
        for(int i=0;i<runs;i++) {
            // do test
            List<AllTypes> list = null;
            if (list == null) {
            }
        }
        time = System.nanoTime() - start;
        System.out.printf("The average time for loop3 taken was %.1f ns%n", (double) time / runs);
    }
```
Testing was running on Android 5.0 with ART

```
I/System.out(32555): The average time for loop1 taken was 913.9 ns
I/System.out(32555): The average time for loop2 taken was 894.6 ns
I/System.out(32555): The average time for loop3 taken was 6.2 ns
I/System.out( 1104): The average time for loop1 taken was 931.6 ns
I/System.out( 1104): The average time for loop2 taken was 1226.4 ns
I/System.out( 1104): The average time for loop3 taken was 6.1 ns
I/System.out( 1332): The average time for loop1 taken was 1145.8 ns
I/System.out( 1332): The average time for loop2 taken was 1276.5 ns
I/System.out( 1332): The average time for loop3 taken was 6.1 ns
I/System.out( 1420): The average time for loop1 taken was 1115.2 ns
I/System.out( 1420): The average time for loop2 taken was 804.7 ns
I/System.out( 1420): The average time for loop3 taken was 6.1 ns
```

I think it could almost save 0.0009 ms for the loop without released weakreference and almost stays with no changes with loop with realease weakreference. (In the real world there should be less than 100 weakreference, right?)

Any way, this won't improve a lot, but a little bit better than nothing.

beeender:
Testing is not good enough, the null checking is so easy to be optimized. Let me think another way for this.

beeender:
Hope testing like this can speak the point
```
    public void testLoop1() {
        long start = System.nanoTime();
        int runs = 1000*1000;
        String notNullString = "";
        for(int i=0;i<runs;i++) {
            // do test
            List<AllTypes> list = null;
            String str = null;
            if (i%2 == 0) {
                str = notNullString;
            }
            if (str == null) {
                list = new ArrayList<AllTypes>(100);
            } else {
                list = new ArrayList<AllTypes>(101);
            }
        }
        long time1 = System.nanoTime() - start;
        System.out.printf("The average time for loop1 taken was %.1f ns%n", (double) time1 / runs);

        start = System.nanoTime();
        for(int i=0;i<runs;i++) {
            // do test
            List<AllTypes> list = new ArrayList<AllTypes>(100);
        }
        long time2 = System.nanoTime() - start;
        System.out.printf("The average time for loop2 taken was %.1f ns%n", (double) time2 / runs);

        System.out.printf("The average time diff was %.1f ns%n", (double) (time2-time1) / runs);
    }
```
```
I/System.out(13958): The average time for loop1 taken was 1094.8 ns
I/System.out(13958): The average time for loop2 taken was 881.5 ns
I/System.out(13958): The average time diff was -213.2 ns
I/System.out(14602): The average time for loop1 taken was 1114.8 ns
I/System.out(14602): The average time for loop2 taken was 1171.4 ns
I/System.out(14602): The average time diff was 56.6 ns
I/System.out(14870): The average time for loop1 taken was 1330.4 ns
I/System.out(14870): The average time for loop2 taken was 892.8 ns
I/System.out(14870): The average time diff was -437.5 ns
I/System.out(15067): The average time for loop1 taken was 1019.5 ns
I/System.out(15067): The average time for loop2 taken was 1036.4 ns
I/System.out(15067): The average time diff was 16.9 ns
I/System.out(15149): The average time for loop1 taken was 1140.8 ns
I/System.out(15149): The average time for loop2 taken was 1012.5 ns
I/System.out(15149): The average time diff was -128.4 ns
```
And i don't think the ART optimized my condition checking since
```
long time1 = System.nanoTime() - start;
long time2 = System.nanoTime() - start;
long diff = time1 - time2
```
always gives me diff = 0

emanuelez:
Yeah, micro-benchmarking in Java and Android is quite complicated and full of pitfalls

cmelchior:
Good enough for me. :+1:

emanuelez:
:+1:

PR #1156, opened 2015-05-26
RealmConfiguration used as cache key

cmelchior:
This changes the Realm caching to work on RealmConfiguration instead of path alone. This is needed to catch subtle errors that path alone cannot, eg. different schema versions in the configuration and different configured schemas.

This means that an IllegalArgumentException will be thrown if:
- Encryption key in two different configurations pointing to the same file are used
- Schema version is different in two different configurations pointing to the same file
- The configured schema is different in two different configurations pointing to the same file.

### DIFF
--- a/realm/src/androidTest/java/io/realm/NotificationsTest.java
+++ b/realm/src/androidTest/java/io/realm/NotificationsTest.java
@@ -135,7 +135,7 @@ public class NotificationsTest extends AndroidTestCase {
         assertEquals(1, counter.get());
     }
 
-    public void testNotificationsNumber () throws InterruptedException, ExecutionException {
+    public void testNotificationsNumber() throws InterruptedException, ExecutionException {
         final AtomicInteger counter = new AtomicInteger(0);
         final AtomicBoolean isReady = new AtomicBoolean(false);
         final Looper[] looper = new Looper[1];
@@ -173,10 +173,10 @@ public class NotificationsTest extends AndroidTestCase {
         while (!isReady.get()) {
             Thread.sleep(5);
         }
-        Thread.sleep(100); 
+        Thread.sleep(100);
 
         // Trigger OnRealmChanged on background thread
-        Realm realm = Realm.getInstance(getContext());
+        realm = Realm.getInstance(getContext());
         realm.beginTransaction();
         Dog dog = realm.createObject(Dog.class);
         dog.setName("Rex");
--- a/realm/src/main/java/io/realm/Realm.java
+++ b/realm/src/main/java/io/realm/Realm.java
@@ -16,7 +16,6 @@
 
 package io.realm;
 
-import android.annotation.SuppressLint;
 import android.annotation.TargetApi;
 import android.content.Context;
 import android.os.Build;
@@ -67,7 +66,6 @@ import io.realm.internal.Util;
 import io.realm.internal.android.DebugAndroidLogger;
 import io.realm.internal.android.ReleaseAndroidLogger;
 import io.realm.internal.log.RealmLog;
-
 
 /**
  * The Realm class is the storage and transactional manager of your object persistent store. It
@@ -131,25 +129,33 @@ public final class Realm implements Closeable {
     // This does not need to be thread safe since it's only used in a synchronized method
     private static volatile boolean isFinalizerStarted = false;
 
-    protected static final ThreadLocal<Map<String, Realm>> realmsCache = new ThreadLocal<Map<String, Realm>>() {
-        @SuppressLint("UseSparseArrays")
+    protected static final ThreadLocal<Map<RealmConfiguration, Realm>> realmsCache =
+            new ThreadLocal<Map<RealmConfiguration, Realm>>() {
         @Override
-        protected Map<String, Realm> initialValue() {
-            return new HashMap<String, Realm>(); // On Android we could use SparseArray<Realm> which is faster,
-                                                  // but incompatible with Java
+        protected Map<RealmConfiguration, Realm> initialValue() {
+            return new HashMap<RealmConfiguration, Realm>();
         }
     };
-    private static final ThreadLocal<Map<String, Integer>> referenceCount = new ThreadLocal<Map<String,Integer>>() {
-        @SuppressLint("UseSparseArrays")
+
+    private static final ThreadLocal<Map<RealmConfiguration, Integer>> referenceCount =
+            new ThreadLocal<Map<RealmConfiguration,Integer>>() {
         @Override
-        protected Map<String, Integer> initialValue() {
-            return new HashMap<String, Integer>();
+        protected Map<RealmConfiguration, Integer> initialValue() {
+            return new HashMap<RealmConfiguration, Integer>();
         }
     };
-    // Maps ids to a boolean set to true if the Realm is open. This is only needed by deleteRealmFile
-    private static final Map<String, AtomicInteger> openRealms = new ConcurrentHashMap<String, AtomicInteger>();
-    private static final int REALM_CHANGED = 14930352; // Hopefully it won't clash with other message IDs.
+
+    // Map between all Realm file paths and all known configurations pointing to that file.
+    private static final Map<String, List<RealmConfiguration>> globalPathConfigurationCache =
+            new HashMap<String, List<RealmConfiguration>>();
+
+    // Map how many times a Realm path has been opened across all threads.
+    // This is only needed by deleteRealmFile.
+    private static final Map<String, AtomicInteger> globalOpenInstanceCounter =
+            new ConcurrentHashMap<String, AtomicInteger>();
+
     protected static final Map<Handler, String> handlers = new ConcurrentHashMap<Handler, String>();
+    private static final int REALM_CHANGED = 14930352; // Hopefully it won't clash with other message IDs.
 
     private static RealmConfiguration defaultConfiguration;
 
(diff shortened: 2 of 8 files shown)


cmelchior:
With RealmConfiguration merged in this is ready for review.
@emanuelez @nhachicha @kneth

realm/src/androidTest/java/io/realm/NotificationsTest.java
```diff
@@ -173,10 +173,10 @@ public Boolean call() throws Exception {
         while (!isReady.get()) {
             Thread.sleep(5);
         }
-        Thread.sleep(100); 
+        Thread.sleep(100);
 
         // Trigger OnRealmChanged on background thread
-        Realm realm = Realm.getInstance(getContext());
+        realm = Realm.getInstance(getContext());
```
kneth: Why not get the instance in `setUp()`?

realm/src/main/java/io/realm/Realm.java
```diff
@@ -123,25 +122,33 @@
...
-            return new HashMap<String, Integer>();
+        protected Map<RealmConfiguration, Integer> initialValue() {
+            return new HashMap<RealmConfiguration, Integer>();
         }
     };
-    // Maps ids to a boolean set to true if the Realm is open. This is only needed by deleteRealmFile
-    private static final Map<String, AtomicInteger> openRealms = new ConcurrentHashMap<String, AtomicInteger>();
-    private static final int REALM_CHANGED = 14930352; // Hopefully it won't clash with other message IDs.
+
+    // Map between all Realm file path and all known configurations pointing to that file.
```
kneth: "path" -> "paths"

realm/src/androidTest/java/io/realm/NotificationsTest.java
cmelchior: Because these tests are so specialised that doing that would decrease readability. Only reason they are class variables in the first place is to make sure the Realm gets closed in case of a test failure which can impact other tests if the Realm is not closed.

cmelchior:
Bump @emanuelez @nhachicha @kneth

realm/src/main/java/io/realm/Realm.java
```diff
@@ -123,25 +122,33 @@
 public final class Realm implements Closeable {
     public static final String DEFAULT_REALM_NAME = "default.realm";
 
-    protected static final ThreadLocal<Map<String, Realm>> realmsCache = new ThreadLocal<Map<String, Realm>>() {
-        @SuppressLint("UseSparseArrays")
```
emanuelez: Won't this revive the lint warning?

realm/src/main/java/io/realm/Realm.java
```diff
@@ -123,25 +122,33 @@
...
+        protected Map<RealmConfiguration, Integer> initialValue() {
+            return new HashMap<RealmConfiguration, Integer>();
         }
     };
-    // Maps ids to a boolean set to true if the Realm is open. This is only needed by deleteRealmFile
-    private static final Map<String, AtomicInteger> openRealms = new ConcurrentHashMap<String, AtomicInteger>();
-    private static final int REALM_CHANGED = 14930352; // Hopefully it won't clash with other message IDs.
+
+    // Map between all Realm file paths and all known configurations pointing to that file.
+    private static final Map<String, List<RealmConfiguration>> globalPathConfigurationCache =
```
emanuelez: Isn't the path information already held in the configuration? Why do we need this extra book-keeping?

realm/src/androidTest/java/io/realm/NotificationsTest.java
```diff
@@ -173,10 +173,10 @@ public Boolean call() throws Exception {
         while (!isReady.get()) {
             Thread.sleep(5);
         }
-        Thread.sleep(100); 
+        Thread.sleep(100);
```
nhachicha: you can use `SystemClock.sleep(millis)` if you don't want to handle `InterruptedException`

realm/src/main/java/io/realm/Realm.java
```diff
@@ -595,23 +600,66 @@ private static synchronized Realm createAndValidate(RealmConfiguration config, b
...
+
+            // Check that encryption keys aren't different
+            if (!Arrays.equals(cachedConfiguration.getEncryptionKey(), newConfiguration.getEncryptionKey())) {
+                throw new IllegalArgumentException(DIFFERENT_KEY_MESSAGE);
+            }
+
+            // Check schema versions are the same
+            if (cachedConfiguration.getSchemaVersion() != newConfiguration.getSchemaVersion()) {
+                throw new IllegalArgumentException(String.format("Configurations cannot have different schema versions " +
+                                "if used to open the same file. %s vs. %s", cachedConfiguration.getSchemaVersion(),
```
kneth: `getSchemaVersion()` returns a `long` and not a `String`.

realm/src/main/java/io/realm/Realm.java
cmelchior: No, SparseArrays are only relevant if the key is an Integer, which we removed some time ago.

realm/src/main/java/io/realm/Realm.java
cmelchior: Because we need to detect conflicting configurations for the same path, ie. two different schemas for the same path or two different schema versions, but eg. having one configuration readonly and another as writeable is OK.

So we need to group configurations pr. path to be able to validate them.

realm/src/main/java/io/realm/Realm.java
```diff
@@ -595,23 +600,66 @@ private static synchronized Realm createAndValidate(RealmConfiguration config, b
         return realm;
     }
 
+    // Make sure that the new configuration doesn't clash with any existing configurations for the Realm
+    private static void validateAgainstExistingConfigurations(RealmConfiguration newConfiguration) {
```
kneth: This one is only called from `createAndValidate()`? And since `createAndValidate()` is `synchronized`, the operations in here are thread-safe, right?

realm/src/main/java/io/realm/Realm.java
```diff
@@ -212,18 +216,20 @@ protected void finalize() throws Throwable {
...
             references = 0;
         }
         if (sharedGroup != null && references == 1) {
-            realmsCache.get().remove(canonicalPath);
+            realmsCache.get().remove(configuration);
+            globalPathConfigurationCache.get(canonicalPath).remove(configuration);
             sharedGroup.close();
             sharedGroup = null;
-            AtomicInteger counter = openRealms.get(canonicalPath);
+            AtomicInteger counter = globalOpenInstanceCounter.get(canonicalPath);
```
nhachicha: According to the doc https://github.com/realm/realm-java/blob/cm-configuration-builder-cache-compare/realm/src/main/java/io/realm/Realm.java#L146 , this is only needed by `deleteRealmFile` !?

realm/src/androidTest/java/io/realm/NotificationsTest.java
cmelchior: In this case we do care though as that would be a test failure. Also as we just add ```throws InterruptedException``` to the method name it doesn't really effect readability of the test.

realm/src/main/java/io/realm/Realm.java
```diff
@@ -123,25 +122,33 @@
...
-            return new HashMap<String, Realm>(); // On Android we could use SparseArray<Realm> which is faster,
-                                                  // but incompatible with Java
+        protected Map<RealmConfiguration, Realm> initialValue() {
+            return new HashMap<RealmConfiguration, Realm>();
         }
     };
-    private static final ThreadLocal<Map<String, Integer>> referenceCount = new ThreadLocal<Map<String,Integer>>() {
-        @SuppressLint("UseSparseArrays")
+
+    private static final ThreadLocal<Map<RealmConfiguration, Integer>> referenceCount =
```
nhachicha: Are we using the Android naming convention? if yes `referenceCount` -> `sReferenceCount`

cmelchior: We are not :)

realm/src/main/java/io/realm/Realm.java
cmelchior: Yes, although globalPathConfigurationCache is actually also used in ```.close()``` so globalPathConfigurationCache has to use a threadsafe map. I'll update that.

realm/src/main/java/io/realm/Realm.java
cmelchior: Not really as this counts the number of open instances, while globalPathConfigurationCache keeps track of unique RealmConfigurations. You can have multiple open instances pr. configuration.

realm/src/main/java/io/realm/Realm.java
```diff
@@ -595,23 +600,66 @@ private static synchronized Realm createAndValidate(RealmConfiguration config, b
...
-    private static void initializeRealm(Realm realm, RealmConfiguration config) {
+    private static void initializeRealm(Realm realm) {
         long version = realm.getVersion();
         boolean commitNeeded = false;
         try {
             realm.beginTransaction();
             if (version == UNVERSIONED) {
                 commitNeeded = true;
-                realm.setVersion(config.getSchemaVersion());
+                realm.setVersion(realm.configuration.getSchemaVersion());
```
nhachicha: use `realm.getConfiguration()` for consistency

emanuelez: the method call is usually 7 times slower than the direct field access

realm/src/main/java/io/realm/Realm.java
```diff
@@ -595,23 +600,66 @@ private static synchronized Realm createAndValidate(RealmConfiguration config, b
...
             }
-            for (Class<? extends RealmObject> modelClass : realm.proxyMediator.getModelClasses()) {
+
+            RealmProxyMediator mediator = realm.getConfiguration().getSchemaMediator();
```
nhachicha: `realm.configuration` as it's faster :)

cmelchior:
Bump @emanuelez @kneth @nhachicha 
Note this does actually build, build server is just acting up.

nhachicha:
:+1:

kneth:
:+1:

realm/src/main/java/io/realm/Realm.java
emanuelez: Indeed

emanuelez:
:+1: apart from very minor comment

PR #1164, opened 2015-05-28
Querying RealmResults can crash

### LINKED ISSUE #1161: Native crash  signal 11 (SIGSEGV), code 2
Finally i can reproduce the native crash in my project
Here is the original native crash backtrace:
```
05-28 15:30:42.375 28102 28102 F libc    : Fatal signal 11 (SIGSEGV), code 2, fault addr 0xbe1faff8 in tid 28102 (.kiwian.jiaxiao)
05-28 15:30:42.492   237   237 I DEBUG   : *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
05-28 15:30:42.492   237   237 I DEBUG   : Build fingerprint: 'oneplus/bacon/A0001:5.0.2/LRX22G/YNG1TAS17L:user/release-keys'
05-28 15:30:42.493   237   237 I DEBUG   : Revision: '0'
05-28 15:30:42.493   237   237 I DEBUG   : ABI: 'arm'
05-28 15:30:42.493   237   237 I DEBUG   : pid: 28102, tid: 28102, name: .kiwian.jiaxiao  >>> com.kiwian.jiaxiao <<<
05-28 15:30:42.494   237   237 I DEBUG   : signal 11 (SIGSEGV), code 2 (SEGV_ACCERR), fault addr 0xbe1faff8
05-28 15:30:42.542   237   237 I DEBUG   :     r0 afdefd30  r1 afdefd30  r2 00000006  r3 afdefd38
05-28 15:30:42.542   237   237 I DEBUG   :     r4 96a827c0  r5 96a827c0  r6 afdefd30  r7 00000006
05-28 15:30:42.542   237   237 I DEBUG   :     r8 00000000  r9 96a827c0  sl 00000006  fp afdefcc4
05-28 15:30:42.542   237   237 I DEBUG   :     ip 0000001d  sp be1fb000  lr afdb5827  pc afdcfd06  cpsr 600f0030
05-28 15:30:42.542   237   237 I DEBUG   : 
05-28 15:30:42.542   237   237 I DEBUG   : backtrace:
05-28 15:30:42.542   237   237 I DEBUG   :     #00 pc 000a7d06  /data/app/com.kiwian.jiaxiao-1/lib/arm/librealm-jni.so
05-28 15:30:42.542   237   237 I DEBUG   :     #01 pc 00000d34  /data/app/com.kiwian.jiaxiao-1/lib/arm/librealm-jni.so
05-28 15:30:42.808   775  1505 W ActivityManager: getTasks: caller 10124 does not hold GET_TASKS; limiting output
05-28 15:30:43.765   237   237 I DEBUG   : 
```

And here is the crash backtrace with my test case:
```
05-28 16:04:10.208   237   237 I DEBUG   : *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
05-28 16:04:10.208   237   237 I DEBUG   : Build fingerprint: 'oneplus/bacon/A0001:5.0.2/LRX22G/YNG1TAS17L:user/release-keys'
05-28 16:04:10.208   237   237 I DEBUG   : Revision: '0'
05-28 16:04:10.208   237   237 I DEBUG   : ABI: 'arm'
05-28 16:04:10.208   237   237 I DEBUG   : pid: 14514, tid: 14551, name: ationTestRunner  >>> io.realm.test <<<
05-28 16:04:10.208   237   237 I DEBUG   : signal 11 (SIGSEGV), code 2 (SEGV_ACCERR), fault addr 0xaccd8ff8
05-28 16:04:10.212   237   237 I DEBUG   :     r0 afdefd30  r1 afdefd30  r2 00000006  r3 afdefd38
05-28 16:04:10.212   237   237 I DEBUG   :     r4 acbd3860  r5 acbd3860  r6 afdefd30  r7 00000006
05-28 16:04:10.212   237   237 I DEBUG   :     r8 00000000  r9 acbd3860  sl 00000006  fp afdefcc4
05-28 16:04:10.212   237   237 I DEBUG   :     ip 0000001d  sp accd9000  lr afdb5827  pc afdcfd06  cpsr 600f0030
05-28 16:04:10.213   237   237 I DEBUG   : 
05-28 16:04:10.213   237   237 I DEBUG   : backtrace:
05-28 16:04:10.213   237   237 I DEBUG   :     #00 pc 000a7d06  /data/app/io.realm.test-2/lib/arm/librealm-jni.so
05-28 16:04:10.213   237   237 I DEBUG   :     #01 pc 00000d34  /data/app/io.realm.test-2/lib/arm/librealm-jni.so
```

And the testcase to reproduce it:
```java
    private RealmResults<CatOwner> results = null;
    public void testModifyRealmListD() {
        //If it doesn't crash, increase this
        final int count = 100;
        // Enable the list for maintaining the reference will avoid crash!!!!!!!!!!!!!
        //final ArrayList<RealmResults<CatOwner>> arrayList = new ArrayList<RealmResults<CatOwner>>();

        results = testRealm.where(CatOwner.class).findAll();
        RealmChangeListener listener = new RealmChangeListener() {
            @Override
            public void onChange() {
                // Enable the list for maintaining the reference will avoid crash!!!!!!!!!!!!!
                //arrayList.add(results);
                results = results.where().notEqualTo("name", "asdfas" + count).findAll();
                assertEquals(count, results.size());
                for (int i=0; i<count; i++) {
                    CatOwner catOwner = results.get(i);
                    assertEquals(catOwner.getName(), "catOwner"+i);
                    Log.e("testModifyRealmListD", "loop count " + i);
                }
                System.gc();
            }
        };
        testRealm.addChangeListener(listener);
        for (int i=1; i<=count; i++) {
            doTestModifyRealmList(count);
        }
    }

    private void doTestModifyRealmList(final int count) {
        final RealmResults<CatOwner> results = testRealm.where(CatOwner.class).findAll();
        testRealm.beginTransaction();
        for (int i=0; i<count; i++) {
            CatOwner catOwner = new CatOwner();
            catOwner.setName("catOwner" + i);
            catOwner = testRealm.copyToRealmOrUpdate(catOwner);
        }
    }
```

I am almost 99% sure it is caused by wild native pointer which is released by the GC since if the ReleamResults's references are maitained by a ArrayList, the crash won't happen.
I guess something needs to be done if a query is created from a RealmResults

kneth:
A reported in #1161, querying `RealmResults` might lead to a crash. The reason seems to be that a `TableQuery` might be finalised prematurely, and the native resource is freed. When the native resource is freed, accessing it will cause a segmentation fault.

This fixes #1161 by adding references to TableView and TableQuery in order to prevent premature garbage collection.

@emanuelez @cmelchior @beeender

### DIFF
--- a/realm-jni/src/io_realm_internal_tableview.cpp
+++ b/realm-jni/src/io_realm_internal_tableview.cpp
@@ -962,9 +962,10 @@ JNIEXPORT jstring JNICALL Java_io_realm_internal_TableView_nativeRowToString(
     return NULL;
 }
 
-JNIEXPORT jlong JNICALL Java_io_realm_internal_TableView_nativeWhere
-  (JNIEnv *env, jobject, jlong nativeViewPtr)
+JNIEXPORT jlong JNICALL Java_io_realm_internal_TableView_nativeWhere(
+    JNIEnv *env, jobject, jlong nativeViewPtr)
 {
+    TR_ENTER_PTR(nativeViewPtr)
     try {
         if (!VIEW_VALID_AND_IN_SYNC(env, nativeViewPtr))
             return 0;
--- a/realm/src/androidTest/java/io/realm/RealmQueryTest.java
+++ b/realm/src/androidTest/java/io/realm/RealmQueryTest.java
@@ -5,6 +5,7 @@ import android.test.AndroidTestCase;
 import java.util.Date;
 
 import io.realm.entities.AllTypes;
+import io.realm.entities.CatOwner;
 import io.realm.entities.Dog;
 import io.realm.entities.NonLatinFieldNames;
 import io.realm.entities.Owner;
@@ -457,6 +458,22 @@ public class RealmQueryTest extends AndroidTestCase{
         stringOnlies2.sort("chars");
         for (int i = 0; i < stringOnlies2.size(); i++) {
             assertEquals(sorted[i], stringOnlies2.get(i).getChars());
+        }
+    }
+
+    // If the RealmQuery is built on a TableView, it should not crash when used after GC.
+    // See issue #1161 for more details.
+    public void testBuildQueryFromResultsGC() {
+        // According to the testing, setting this to 10 can almost certainly trigger the GC.
+        // Use 30 here can ensure GC happen. (Tested with 4.3 1G Ram and 5.0 3G Ram)
+        final int count = 30;
+        RealmResults<CatOwner> results = testRealm.where(CatOwner.class).findAll();
+
+        for (int i=1; i<=count; i++) {
+            @SuppressWarnings({"unused"})
+            byte garbage[] = TestHelper.allocGarbage(0);
+            results = results.where().findAll();
+            System.gc(); // if a native resource has a reference count = 0, doing GC here might lead to a crash
         }
     }
 }
--- a/realm/src/main/java/io/realm/internal/TableQuery.java
+++ b/realm/src/main/java/io/realm/internal/TableQuery.java
@@ -23,19 +23,36 @@ public class TableQuery implements Closeable {
     protected boolean DEBUG = false;
 
     protected long nativePtr;
-    protected final Table parent;
+    protected final Table table;
+    // Don't convert this into local variable and don't remove this.
+    // Core requests Query to hold the TableView reference which it is built from.
+    @SuppressWarnings({"unused"})
+    private final TableOrView origin; // Table or TableView which created this TableQuery
     private final Context context;
 
     private boolean queryValidated = true;
 
     // TODO: Can we protect this?
-    public TableQuery(Context context, Table parent, long nativeQueryPtr){
-        if (DEBUG)
+    public TableQuery(Context context, Table table, long nativeQueryPtr) {
+        if (DEBUG) {
             System.err.println("++++++ new TableQuery, ptr= " + nativeQueryPtr);
+        }
         this.context = context;
-        this.parent = parent;
+        this.table = table;
         this.nativePtr = nativeQueryPtr;
+        this.origin = null;
     }
+
+    public TableQuery(Context context, Table table, long nativeQueryPtr, TableOrView origin) {
+        if (DEBUG) {
+            System.err.println("++++++ new TableQuery, ptr= " + nativeQueryPtr);
+        }
+        this.context = context;
+        this.table = table;
+        this.nativePtr = nativeQueryPtr;
+        this.origin = origin;
+    }
+
 
     public void close() {
         synchronized (context) {
@@ -55,7 +72,7 @@ public class TableQuery implements Closeable {
     protected void finalize() {
         synchronized (context) {
             if (nativePtr != 0) {
-                context.asyncDisposeQuery(nativePtr); 
+                context.asyncDisposeQuery(nativePtr);
                 nativePtr = 0; // Set to 0 if finalize is called before close() for some reason
             }
         }
@@ -472,7 +489,7 @@ public class TableQuery implements Closeable {
         context.executeDelayedDisposal();
         long nativeViewPtr = nativeFindAll(nativePtr, start, end, limit);
         try {
-            return new TableView(this.context, this.parent, nativeViewPtr);
+            return new TableView(this.context, this.table, nativeViewPtr, this);
         } catch (RuntimeException e) {
             TableView.nativeClose(nativeViewPtr);
             throw e;
@@ -486,7 +503,7 @@ public class TableQuery implements Closeable {
         context.executeDelayedDisposal();
         long nativeViewPtr = nativeFindAll(nativePtr, 0, Table.INFINITE, Table.INFINITE);
         try {
-            return new TableView(this.context, this.parent, nativeViewPtr);
+            return new TableView(this.context, this.table, nativeViewPtr, this);
         } catch (RuntimeException e) {
             TableView.nativeClose(nativeViewPtr);
             throw e;
@@ -687,13 +704,13 @@ public class TableQuery implements Closeable {
     // Deletion.
     public long remove(long start, long end) {
         validateQuery();
-        if (parent.isImmutable()) throwImmutable();
+        if (table.isImmutable()) throwImmutable();
         return nativeRemove(nativePtr, start, end, Table.INFINITE);
     }
 
     public long remove() {
         validateQuery();
-        if (parent.isImmutable()) throwImmutable();
+        if (table.isImmutable()) throwImmutable();
         return nativeRemove(nativePtr, 0, Table.INFINITE, Table.INFINITE);
     }
 
--- a/realm/src/main/java/io/realm/internal/TableView.java
+++ b/realm/src/main/java/io/realm/internal/TableView.java
@@ -64,19 +64,40 @@ import java.util.List;
  */
 public class TableView implements TableOrView, Closeable {
     protected boolean DEBUG = false; //true;
+    // Don't convert this into local variable and don't remove this.
+    // Core requests TableView to hold the Query reference.
+    @SuppressWarnings({"unused"})
+    private final TableQuery query; // the query which created this TableView
 
-     /**
+    /**
+     * Creates a TableView. This constructor is used if the TableView is created from a table.
+     *
+     * @param context
+     * @param parent
+     * @param nativePtr
+     */
+    protected TableView(Context context, Table parent, long nativePtr) {
+        this.context = context;
+        this.parent = parent;
+        this.nativePtr = nativePtr;
+        this.query = null;
+    }
+
+    /**
      * Creates a TableView with already created Java TableView Object and a
      * native native TableView object reference. The method is not supposed to
      * be called by the user of the db. The method is for internal use only.
      *
+     * @param context
      * @param parent A table.
-     * @param nativePtr pointer to table.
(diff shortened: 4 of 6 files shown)


realm/src/main/java/io/realm/internal/TableView.java
```diff
@@ -73,19 +73,24 @@
      * @param parent A table.
      * @param nativePtr pointer to table.
      */
-    protected TableView(Context context, Table parent, long nativePtr){
+    protected TableView(Context context, Table parent, long nativePtr) {
```
bmunkholm: Is there any for this without the query?

realm/src/main/java/io/realm/internal/TableView.java
```diff
@@ -73,19 +73,24 @@
...
         this.context = context;
         this.parent = parent;
         this.nativePtr = nativePtr;
     }
 
+    protected TableView(Context context, Table parent, long nativePtr, TableQuery query) {
```
bmunkholm: Add comment to this to explain when it should be used.

beeender:
Changes in RealmResults.java missed?
RealmQuery.java:1155: error: no suitable constructor found for RealmResults(Realm,TableView,Class<E>,RealmQuery<E>)
        return new RealmResults<E>(realm, query.findAll(), clazz, this);

kneth:
@beeender Yes, I forgot to add it to my commit :-(

realm/src/androidTest/java/io/realm/RealmQueryTest.java
```diff
@@ -459,4 +461,43 @@ public void testGeorgian() {
...
+            @Override
+            public void onChange() {
+                // Enable the list for maintaining the reference will avoid crash!!!!!!!!!!!!!
+                // arrayList.add(results);
+                results = results.where().notEqualTo("name", "asdfas" + count).findAll();
+                //assertEquals(count, results.size());
+                for (int i=0; i<count; i++) {
+                    CatOwner catOwner = results.get(i);
+                    assertEquals(catOwner.getName(), "catOwner"+i);
+                    Log.e("testModifyRealmListD", "loop count " + i);
```
cmelchior: No need to log during a unit test.

kneth:
@bmunkholm Your comments suggest that we should review how the different Realm classes (including the `io.realm.internal` classes) interact.

realm/src/androidTest/java/io/realm/RealmQueryTest.java
```diff
@@ -459,4 +461,43 @@ public void testGeorgian() {
...
+                }
+                System.gc();
```
cmelchior: This is just a hint to the GC. You are not guaranteed that it runs. So adding a note about what this is trying to trigger would be nice.

realm/src/main/java/io/realm/internal/TableView.java
cmelchior: Do we really need a reference to both parent and query? They seem to be semantically the same as the query is the "parent"?

kneth: No, the parent is the original table which the table view is fetching it data from. If you do a query on a table view, the new table view will still point back to the table. A statement like `r = r.where().findAll()` will create a table query (`r.where()`) which holds a native resource (a `realm::Query` object). When this table query is garbage collected, the native resource will deallocated, and the table view is rerun the query.

OK, my explanation isn't clear - and it calls for a drawing ;-)

cmelchior: I think I understand that. I am just wondering why the TableQuery doesn't hold that reference already, saving the complicated constructor.

kneth: The TableQuery object might hold a reference, but I believe we need to have a reference to the TableQuery object in order to prevent premature deallocation.

beeender:
Tested with my project. Works fine, thanks!

realm-jni/src/io_realm_internal_tableview.cpp
```diff
@@ -962,16 +962,19 @@ JNIEXPORT jstring JNICALL Java_io_realm_internal_TableView_nativeRowToString(
...
+    TR_ENTER_PTR(nativeViewPtr)
     try {
         if (!VIEW_VALID_AND_IN_SYNC(env, nativeViewPtr))
             return 0;
 
         Query query = TV(nativeViewPtr)->get_parent().where(TV(nativeViewPtr));
         TableQuery* queryPtr = new TableQuery(query);
-        return reinterpret_cast<jlong>(queryPtr);
+        jlong native_query_ptr = reinterpret_cast<jlong>(queryPtr);
+        TR("native_query_ptr %p", VOID_PTR(native_query_ptr))
```
emanuelez: Do we want this in the final version?

realm/src/main/java/io/realm/internal/TableView.java
emanuelez: I think the clean way to do this would be to have an `Table originatingTable` field in `TableQuery`, which would allow us to get rid of `parent` in this constructor

realm-jni/src/io_realm_internal_tableview.cpp
kneth: I need to do a little clean-up before we can merge it.

realm/src/main/java/io/realm/internal/TableView.java
kneth: Yes, the `io.realm.internal.Table` class has a few methods which return a `TableView` object with going through a query. An example is `getSortedView()`.

kneth:
@emanuelez @cmelchior I have cleaned up my code a bit. Please take a look.

realm/src/androidTest/java/io/realm/RealmQueryTest.java
```diff
@@ -459,4 +460,36 @@ public void testGeorgian() {
             assertEquals(sorted[i], stringOnlies2.get(i).getChars());
         }
     }
+
+    private RealmResults<CatOwner> results = null;
+    public void testModifyRealmListD() {
```
cmelchior: Perhaps a small description saying that this test is nondeterministic and depends on the GC doing it's job at the wrong time.

cmelchior: Also trailing D

realm/src/main/java/io/realm/internal/TableView.java
```diff
@@ -64,19 +64,34 @@
...
 
-     /**
+    /**
+     * Creates a TableView. This constructor is used if the TableView is created from a table.
+     *
+     * @param context
+     * @param parent
+     * @param nativePtr
+     */
+    protected TableView(Context context, Table parent, long nativePtr) {
```
cmelchior: Is this constructor ever used?

realm/src/main/java/io/realm/RealmResults.java
```diff
@@ -57,6 +57,7 @@
     private Class<E> classSpec;
     private Realm realm;
     private TableOrView table = null;
+    private RealmQuery<E> query; // Query which created this TableView
```
cmelchior: We don't need to save it here when it is referenced in TableView?

realm/src/main/java/io/realm/internal/TableView.java
beeender: It is used by Table.

beeender:
@emanuelez @kneth @nhachicha I did some clean up on this. Please help to review. Thanks!

kneth:
:+1:

realm/src/main/java/io/realm/internal/TableQuery.java
```diff
@@ -24,12 +24,16 @@
 
     protected long nativePtr;
     protected final Table parent;
```
emanuelez: The naming is a bit confusing now. How about renaming parent to table?

realm/src/main/java/io/realm/internal/TableQuery.java
```diff
@@ -24,12 +24,16 @@
...
+    // Don't convert this into local variable and don't remove this.
+    // Core requests Query to hold the TableView reference which it is built from.
+    @SuppressWarnings({"unused"})
+    private TableOrView origin; // Table or TableView which created this TableQuery
```
emanuelez: final

realm/src/androidTest/java/io/realm/RealmQueryTest.java
```diff
@@ -459,4 +460,20 @@ public void testGeorgian() {
             assertEquals(sorted[i], stringOnlies2.get(i).getChars());
         }
     }
+
+    // If the RealmQuery is built on a TableView, it should not crash when use it after GC.
```
emanuelez: use it -> used

realm/src/androidTest/java/io/realm/RealmQueryTest.java
```diff
@@ -459,4 +460,20 @@ public void testGeorgian() {
...
+    // See issue #1161 for more details.
+    public void testBuildQueryFromResultsGC() {
```
emanuelez: I'm dead scared of non-deterministic unit test :(

beeender: Be optimistic, The GC surely will happen!

beeender:
@emanuelez @nhachicha @kneth Smalls fixes, help to review. Thanks!

nhachicha:
:+1:

PR #1166, opened 2015-05-29
Standalone objects can now be added to managed RealmLists

### LINKED ISSUE #1140: RealmList.add() throws misleading exception for standalone objects
From this SO question: http://stackoverflow.com/questions/30340752/cant-add-new-object-to-realmlist-due-to-null-object-reference

The following will fail with a NullPointer exception:
```
RealmList<Bar> list = realm.where(Foo.class).findFirst().getBarList();
list.add(new Bar()); 
// throws NullPointerException: Attempt to invoke virtual method 'long io.realm.internal.Row.getIndex()' on a null object reference
```

As a minimum we should check if the object added is a managed RealmObject, but even better might be to do a ```copyToRealm()``` or ```copyToRealmOrUpdate``` inside the RealmList if it is a standalone object.

The later is needed for the following corner case:
```
RealmResult<Foo> result;
RealmList<BarWithPrimaryKey> list1 = result.get(0).getBarList();
RealmList<BarWithPrimaryKey> list2 = result.get(1).getBarList();

BarWithPrimaryKey newBar = new BarWithPrimaryKey();
list1.add(newBar);
list2.add(newBar);
```

which would fail if just using ```copyToRealm```.

What do you think @emanuelez @kneth

cmelchior:
Fixes #1140 

Users can now transparently add standalone objects to RealmLists already in Realm without manually having to call copyToRealm() or copyToRealmOrUpdate(). This also mirrors the behaviour the other way where users already can add managed objects to unmanaged RealmLists.


cmelchior:
@emanuelez @kneth

kneth:
:+1:

realm/src/main/java/io/realm/RealmList.java
```diff
@@ -100,12 +99,27 @@ public RealmList(E... objects) {
     }
 
     /**
-     * {@inheritDoc}
+     * Inserts the specified object into this List at the specified location. The object is inserted before any previous
+     * element at the specified location. If the location is equal to the size of this List, the object is added at the
+     * end.
+     * <p>
+     * <b>Un-managed RealmLists:</b> It is possible to add both managed and un-managed objects. If adding managed
```
emanuelez: This could probably be an HTML dictionary list

realm/src/main/java/io/realm/RealmList.java
```diff
@@ -143,6 +181,18 @@ public E set(int location, E object) {
         return object;
     }
 
+    // Transparently copies a standalone object or managed object from another Realm to the Realm backing this RealmList.
+    private E copyToRealmIfNeeded(E object) {
+        if (object.row != null && object.realm.getPath().equals(realm.getPath())) {
```
emanuelez: If we stored `path` as protected in Realm instances we could access this as a field instead of an accessor, which would be 7 times faster (according to Google)

emanuelez:
:+1: apart from comments

PR #1168, opened 2015-05-29
Fixing some typo + small tweaks for the 'concurrency' test project

nhachicha:
Note: the gradle file for this module `/test/concurrencyExample` is broken. I fix it locally so I can check my PR but I didn't want to commit it (involve changing the parent `build.gradle`) I'm not sure if this module is still relevant.


cmelchior:
Hi @nhachicha. Thanks :) 
The concurrency example is pretty much deprecated though as it was too complex It currently eludes my why we didn't just delete it though. Maybe @kneth or @emanuelez can remember that.

realm-jni/src/util.hpp
```diff
@@ -190,7 +190,7 @@ extern const char *log_tag;
 
 #define ROW_INDEXES_VALID(env,ptr,start,end, range)             (true)
 #define ROW_INDEX_VALID(env,ptr,row)                            (true)
-#defibe ROW_INDEX_VALID_OFFSET(env,ptr,row)                     (true)
+#define ROW_INDEX_VALID_OFFSET(env,ptr,row)                     (true)
```
kneth: Nice catch!

kneth:
@cmelchior You're right, it is a far too complex example. We should either remove or rewrite it.

cmelchior:
It was replaced by our current ThreadExample, so I would think we should just remove the example but keep the JNI fix :)

emanuelez:
:+1: for removing it

cmelchior:
@nhachicha I think if you delete the concurrency example, but keep the JNI fixes this would be ready to be merged.

nhachicha:
done @cmelchior @kneth @emanuelez

kneth:
:+1:

PR #1171, opened 2015-05-30
Add support for Kotlin

### LINKED ISSUE #509: Support KotlinLang
Current version of Realm (0.72) not working with Kotlin (http://kotlinlang.org/).
A have an exception "Caused by: io.realm.exceptions.RealmException: Could not find the generated ValidationList class", I think it's because there is no private fields with getters and setters in my model.
Kotlin don't recommend to create default getters/setters for properties.
Is it possible to use Realm with Kotlin?

cypressious:
Fixes #509 

This enables better compatibility with Kotlin's new annotation processing capability.


cmelchior:
Hi @cypressious 
Thank you for this. The only reason for the current ```SOURCE``` retention policy is a slightly reduced file size, so I wouldn't have any problems changing to ```CLASS``` for now if that was the only blocker for Kotlin support. We can always change back if (when?) Kotlin improve their annotation processing support.

However I am guessing we need to change all our annotations to ```CLASS``` not just PrimaryKey.java. This means that before merging this we need to update the following classes:

- Ignore.java
- Index.java
- RealmClass.java
- Also ```changelog.txt``` should be updated to reflect this.

Ideally we should also add an example project in our our ```/example``` folder showing how it works, but this can be done in another PR.

cypressious:
@cmelchior Thanks for your comment.

`Ignore` and `Index` have no retention specified so it defaults to CLASS. `RealmClass` has retention CLASS already.

What should the modified changelog.txt look like? Should I introduce a new version number?

I'm on Windows and unfortunately I can't get the project to build. I can add a small test project but somebody will need to test if it compiles and runs correctly.

cmelchior:
Ah. Didn't check that, sounds like a mistake from our side :) But for consistency all annotations should either have it explicitly set or use the default behaviour. Either is fine.

You can add a new version to the changelog. The next release is 0.81, so something like:

```
0.81
 * Realm now works with Kotlin (M12+). 
```

If you have time to add an example project that would be great. We can check it afterwards. Copying the current introExample and converting it to Kotlin would probably be the best solution as that example probe most of the standard Realm functionality. Only change would probably be  to add ```@PrimaryKey``` and ```@Index``` to one of the models just to make sure that all annotations are used.

cypressious:
I've implemented your feedback. The provided example compiles on my machine using a jar dependency. I hope it also works as part of the integrated build process.

For the example, I tried to stay as close to the Java example as possible while writing idiomatic Kotlin code and highlighting some of the added benefits (like `executeTransaction()` with a lambda or Kotlin's equivalent of try-with-resources `use()`.

examples/kotlinExample/build.gradle
```diff
@@ -0,0 +1,51 @@
...
+    compile "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
+    compile 'org.jetbrains.anko:anko:0.6.2-19s'
+    compile files("../../realm/build/libs/realm-${version}.jar")
+
+    kapt files("../../realm/build/libs/realm-${version}.jar")
+}
+
+buildscript {
+    ext.kotlin_version = '0.12.213'
+    repositories {
```
cmelchior: jCenter mirrors MavenCentral and we already have JCenter on the dependency list, so this is redundant and can be removed.

examples/kotlinExample/build.gradle
```diff
@@ -0,0 +1,51 @@
...
+        mavenCentral()
+    }
+    dependencies {
+        classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
+    }
+}
+
+repositories {
```
cmelchior: Same as above. This can be removed.

examples/kotlinExample/src/main/java/io/realm/examples/kotlin/KotlinExampleActivity.kt
```diff
@@ -0,0 +1,217 @@
+/*
+ * Copyright 2014 Realm Inc.
```
cmelchior: Should be 2015

examples/kotlinExample/src/main/java/io/realm/examples/kotlin/model/Cat.kt
```diff
@@ -0,0 +1,25 @@
+/*
+ * Copyright 2014 Realm Inc.
```
cmelchior: 2015

examples/kotlinExample/src/main/java/io/realm/examples/kotlin/model/Dog.kt
```diff
@@ -0,0 +1,25 @@
+/*
+ * Copyright 2014 Realm Inc.
```
cmelchior: 2015

examples/kotlinExample/src/main/java/io/realm/examples/kotlin/model/Person.kt
```diff
@@ -0,0 +1,62 @@
+/*
+ * Copyright 2014 Realm Inc.
```
cmelchior: 2015

examples/kotlinExample/src/main/java/io/realm/examples/kotlin/model/Person.kt
```diff
@@ -0,0 +1,62 @@
...
+// https://youtrack.jetbrains.com/issue/KT-7890 for progress on this issue. When this is resolved,
+// the @RealmClass annotation can be omitted)
+
+// Furthermore, the class and all of the properties must be annotated with open
+// (Kotlin classes and methods are final by default).
+@RealmClass
+public open class Person(
+        // You can put properties in the constructor as long as all of them are initialized with
+        // default values. This ensures that an empty constructor is generated.
+
```
cmelchior: Remove whitespace.

examples/kotlinExample/src/main/java/io/realm/examples/kotlin/model/Person.kt
```diff
@@ -0,0 +1,62 @@
...
+        // All properties are by default persisted.
+
```
cmelchior: Remove whitespace

examples/kotlinExample/src/main/java/io/realm/examples/kotlin/model/Person.kt
```diff
@@ -0,0 +1,62 @@
...
+        // Properties can be annotated with PrimaryKey or Index
+
```
cmelchior: Remove whitespace

examples/kotlinExample/build.gradle
cypressious: This won't compile. For some reason Gradle needs a repositories block in every buildscript block. I'm changing it to jcenter though.

settings.gradle
```diff
@@ -8,7 +8,8 @@ include ':introExample',
         ':migrationExample',
         ':adapterExample',
         ':threadExample',
-		':jsonExample'
+		':jsonExample',
```
cmelchior: Whitespace here looks wrong. Perhaps tabs instead of spaces?

cmelchior:
Hi @cypressious 
Apart from some small style fixes this is looking good. Thank you for a complete Kotlin example.

settings.gradle
cypressious: The opposite :-) Changing it

cmelchior:
:-1:

Any comments @emanuelez @kneth @nhachicha ?

cmelchior:
Ups, fumble fingers.

cypressious:
How to interpret the thumbs down?

cmelchior:
As me hitting the wrong buttons. It looks good.

cypressious:
Alright :-)

kneth:
:+1:

emanuelez:
:+1:

corbt:
This is beautiful. Any timeline for when it will be in a release on Maven?

cmelchior:
It should already be available as a ```0.81-SNAPSHOT``` release: https://github.com/realm/realm-java#using-snapshots , so feel free to try it out :)

corbt:
Thanks. I'm running into an issue using the `0.81-SNAPSHOT` release with a Kotlin `RealmClass`. This is with a fresh install of the app after a full clean and rebuild.

Error:
```
    io.realm.exceptions.RealmException: Could not find the generated proxy class for class com.emberall.kiosk.model.User. Annotation processor may not have been executed.
```

My class:
```kotlin
@RealmClass
public open class User(
    PrimaryKey
    Expose public var id: String = UUID.randomUUID().toString(),
    Expose public var created_at: Date = Date(),
    Expose public var updated_at: Date = Date(),

    Index Expose public open var name: String? = "",
    Index Expose public open var email: String? = "",
    Expose open public var slug: String = "",
    Expose public open var address: Address? = null,
    Expose public open var album: Album? = null,
    public open var dirty: Boolean = false
) : RealmObject() {}
```

Any ideas what might be causing that?

cmelchior:
Maybe the Kotlin annotation processor is not running? You can see an example of a working project here: https://github.com/realm/realm-java/blob/master/examples/kotlinExample/build.gradle

cypressious:
Did you add the kapt entry to your build.gradle?
Am 03.06.2015 12:34 schrieb "Kyle Corbitt" <notifications@github.com>:

corbt:
Yep, the missing kapt directive was the issue. Once I added that to my build.gradle I was fine. Thanks!

```gradle
def realm_version = 'io.realm:realm-android:0.81-SNAPSHOT'
dependencies {
    ...
    compile realm_version 
    kapt realm_version 
}
```

PR #1177, opened 2015-06-02
Specified Realm.copyToRealm behaviour 

### LINKED ISSUE #1176: Docs not saying will method copyToRealm() deeply copy  object to Realm?
Docs not saying will method copyToRealm() deeply copy  object to Realm? By this I mean including other realm objects which are fields of copy object.

cmelchior:
Fixes #1176 

@kneth @emanuelez @nhachicha

### DIFF
--- a/realm/src/main/java/io/realm/Realm.java
+++ b/realm/src/main/java/io/realm/Realm.java
@@ -973,7 +973,8 @@ public final class Realm implements Closeable {
 
     /**
      * Copies a RealmObject to the Realm instance and returns the copy. Any further changes to the original RealmObject
-     * will not be reflected in the Realm copy.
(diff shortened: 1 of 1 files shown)


emanuelez:
:+1: :shipit:

kneth:
What happens when you have a cyclic data structure? Should that case be documented?

cmelchior:
It is copied as a cyclic datastructure. The recursive copy method keeps a reference to all copied data so we don't accidentally copy it twice. There is even a unit test for it ;)

Documenting it would not make much sense i think, as it would basically say "Also works with cyclic datastructures."

kneth:
In that case :+1:

Issue #729, opened 2015-01-08
Getting a weird behavior when saving a string


ImGauravC:
```java
java.lang.RuntimeException: Unspecified exception. Failure when converting short string to UTF-16 0x20 0xae 0x00 0x00 0x90 0xae 0x00 0x00 0x98 0x54 0x01 in io_realm_internal_Row.cpp line 126
            at io.realm.internal.Row.nativeGetString(Native Method)
            at io.realm.internal.Row.getString(Row.java:126)
            at io.realm.MessageModelRealmProxy.getMessage(MessageModelRealmProxy.java:108)
```

cmelchior:
Hi @ImGauravC 
Thank you very much. Can you provide some more information. Do you know what string you expected to get when calling getMessage()?

bmunkholm:
Can it be replicated?

ImGauravC:
@cmelchior : getMessage() should be a normal text message that includes emoji characters.
@bmunkholm : Sure. I'll send the realm file the next time I see this.

kneth:
@ImGauravC Can you send us your realm file? It might give us a hint where the problem is.

yolapop:
Hi all, I got similar exception after upgrading to version 0.77.

[...]

PR #1180, opened 2015-06-03
Introducing RealmError

kneth:
#729 reports a number of cases when an app is returning a wrong string value. Investigations by users suggest that a call to `mmap()` failed. When `mmap()` (and it sister kernel calls) fail, Realm Core throws a `runtime_error`. So far, we have thrown a `RuntimeException` in these cases but exceptions are often ignored. Ignoring a `mmap()` failure is likely to leave the Realm file in a corrupted state which could easily explain the wrong string values.

Realm will now throw a RealmError when Realm Core enters an unrecoverable error condition. As errors are typically not ignored, it feels safer rather to crash the app than leave the Realm file corrupt.

@bmunkholm @kspangsege @cmelchior @emanuelez @nhachicha 

TODO:
 - [x] changelog


realm-jni/src/util.cpp
```diff
@@ -118,8 +118,8 @@ void ThrowException(JNIEnv* env, ExceptionKind exception, const std::string& cla
             break;
 
         case Unspecified:
-            jExceptionClass = env->FindClass("java/lang/RuntimeException");
-            message = "Unspecified exception. " + classStr;
+            jExceptionClass = env->FindClass("io/realm/exceptions/RealmError");
+            message = "Unrecoverable error condition. " + classStr;
```
kspangsege: @kneth, as far as I remember, an `mmap()` failure would cause a `std::runtime_error()` to be thrown , which in turn would be categorized as `RuntimeError` rather than `Unspecified`. Isn't `Unspecified` only used for the `catch (...)` case?

realm-jni/src/io_realm_internal_Util.cpp
```diff
@@ -109,7 +109,7 @@ JNIEXPORT jstring JNICALL Java_io_realm_internal_Util_nativeTestcase(
                 ThrowException(env, OutOfMemory, "parm1", "parm2");
             break;
         case Unspecified:
-            expect = "java.lang.RuntimeException: Unspecified exception. parm1";
+            expect = "io.realm.exceptions.RealmError: Unrecoverable error condition. parm1";
```
cmelchior: "error condition" sounds a bit funny to me. Maybe just "Unrecoverable error"?

realm/src/main/java/io/realm/exceptions/RealmError.java
```diff
@@ -0,0 +1,38 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.realm.exceptions;
+
+/**
+ * RealmError is Realm specific unrecoverable errors.
+ */
+
```
cmelchior: Remove whitespace

realm/src/main/java/io/realm/exceptions/RealmError.java
```diff
@@ -0,0 +1,38 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.realm.exceptions;
+
+/**
+ * RealmError is Realm specific unrecoverable errors.
```
cmelchior: Perhaps adding a comment that these should never be caught  and ignored?

realm-jni/src/util.cpp
cmelchior: Throwing an error for any kind of unknown exception seems kinda harsh. I know the user can probably not do anything about it anyway, but still. 

I would rather check the error description and if it involves mmap throw an error, otherwise an exception. This is a bit of slippery slope though.

kneth: The `catch(...)` case will throw a `RuntimeError` while `std::runtime_error()` will still be a `Unspecified` but I change the behaviour of `Unspecified` from `RuntimeException` to `RealmError`

bmunkholm: We should not try to parse text in the exception  - that's a no go. Then we need more detailed exceptions from core instead.

realm/src/main/java/io/realm/exceptions/RealmError.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+
+public class RealmError extends Error {
```
nhachicha: I'd prefer to extends `RuntimeException` as by convention for unchecked exception `Error` is used internally by the JVM,

kneth:
The Realm Core can throw `runtime_error` in about 100 places, and it used to indicate very different errors. Besides failure of `mmap`, `runtime_error` is also thrown if you try to negate an empty query. When `mmap()` fails, an app shouldn't continue but if you negate an empty query, it might simply indicate some wrong user input, and the app can continue after dealing with it.

simonask:
Wrong user input should not result in `std::runtime_error` from Core — if it does (and it looks like it does, such as the example you mention with negating an empty query), I think that's a bug in Core. `std::runtime_error` should be considered an error that is "beyond the scope of the program" (see for instance http://en.cppreference.com/w/cpp/error/runtime_error), and as such could be considered always irrecoverable.

There are some debatable corners (for instance, `std::locale`'s constructor throws `std::runtime_error` if called with a string naming a locale that does not exist in the operating system — some users could conceivably want to handle that), but for now we generally consider `std::runtime_error` irrecoverable in Core. @kspangsege Please correct me if I'm wrong. I hope that clears things up a bit. :)

realm/src/main/java/io/realm/exceptions/RealmError.java
kneth: I wish to have the same behaviour as described in the "Class Overview" in http://developer.android.com/reference/java/lang/Error.html.

kneth: And not the behaviour as described in http://developer.android.com/reference/java/lang/Exception.html.

kneth:
@realm/java Please review.

realm/src/main/java/io/realm/exceptions/RealmError.java
```diff
@@ -0,0 +1,38 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.realm.exceptions;
+
+/**
+ * RealmError is Realm specific unrecoverable errors. An RealmError should never be
```
emanuelez: The first sentence does not make sense to me. How about something in the lines of:

RealmError is a Realm specific `Error` used when unrecoverable problems occur at core level.

realm/src/main/java/io/realm/exceptions/RealmError.java
```diff
@@ -0,0 +1,38 @@
...
+ * caught or ignored. By doing so, the Realm file is likely to be corrupted.
```
emanuelez: the Realm file is likely to be corrupted -> the Realm could possibly get corrupted

Just a suggestion

realm/src/main/java/io/realm/exceptions/RealmError.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+public class RealmError extends Error {
+    public RealmError() {
```
emanuelez: Do we really need to define so many constructors?

changelog.txt
```diff
@@ -1,5 +1,6 @@
 0.81.2
  * Closing realm on another thread different from where it was created now throws an exception.
+ * Realm will now throw a RealmError when Realm Core enters an unrecoverable error condition.
```
bmunkholm: Can you "enter a condition"?

realm-jni/src/util.cpp
```diff
@@ -117,9 +112,9 @@ void ThrowException(JNIEnv* env, ExceptionKind exception, const std::string& cla
             message = classStr + " " + itemStr;
             break;
 
-        case Unspecified:
-            jExceptionClass = env->FindClass("java/lang/RuntimeException");
```
cmelchior: I assume this is only temporary until core manages to make their thrown exceptions better, or is it assumed that going forward all recoverable error cases will result in a named exception we can handle specifically? And we then have to add those as they are implemented, e.g. QueryException?

realm/src/main/java/io/realm/exceptions/RealmError.java
cmelchior: I think that @timanglade didn't like us referring to the core like that, but it was better to use something like "underlying storage engine", but I cannot find the blog post to verify this.

realm-jni/src/util.cpp
kneth: We will leave it as it is but add new exceptions as core adds them.

realm/src/main/java/io/realm/exceptions/RealmError.java
```diff
@@ -0,0 +1,27 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.realm.exceptions;
+
+/**
+ * RealmError is Realm specific Error used when unrecoverable problems in the underlying storage engine. An RealmError
```
beeender: "when unrecoverable problems in the underlying storage engine" -> "when unrecoverable problems **happen** in the underlying storage engine"

nhachicha:
:+1:

beeender:
Apart from comment :+1:

beeender:
:+1:

PR #1182, opened 2015-06-03
Fixes #1169 unchecked cast

### LINKED ISSUE #1169: Unchecked cast warnings
I am getting the unchecked cast warning for the proxy class that gets generated for all my realm objects.

Here is an example of the warning generated when processing my simple User class:

required: E
  found:    UserRealmProxy
  where E is a type-variable:
    E extends RealmObject declared in method <E>newInstance(Class<E>)
[[directory to my build]]/intermediates/classes/release/io/realm/DefaultRealmModuleMediator.java:186: warning: [unchecked] unchecked cast
        Class<E> clazz = (Class<E>) ((obj instanceof RealmObjectProxy) ? obj.getClass().getSuperclass() : obj.getClass());

I like to run a **no** warning environment, so this makes things really difficult ;)

nhachicha:
Fixes https://github.com/realm/realm-java/issues/1169
Note: I updated the expected template test file `realm-annotations-processor/test/java/resources/io/realm.DefaultRealmModuleMediator.java` but apparently our annotation processor unit tests are not running (at least as part pf `gradle cAT`)
@cmelchior @kneth @emanuelez


realm-annotations-processor/src/main/java/io/realm/processor/RealmProxyMediatorGenerator.java
```diff
@@ -243,13 +243,14 @@ private void emitCopyToRealmMethod(JavaWriter writer) throws IOException {
                 EnumSet.of(Modifier.PUBLIC),
                 "Realm", "realm", "E", "obj", "boolean", "update", "Map<RealmObject, RealmObjectProxy>",  "cache"
         );
-
-        writer.emitStatement("Class<E> clazz = (Class<E>) ((obj instanceof RealmObjectProxy) ? obj.getClass().getSuperclass() : obj.getClass())");
+        writer.emitSingleLineComment("This cast is correct because obj is either ");
```
cmelchior: Cool, hadn't noticed you could do this with JavaWriter. We should probably use this a bit more.

cmelchior:
:+1: 
And yes, the annotation processor unit tests has to be run by doing ```/realm-annotation-processor/gradlew clean test```, we should probably make it easier to run both Realm unit tests and annotation processor unit tests because I forget to to do it all the time also :)

cmelchior:
Ups forgot. Missing changelog update

kneth:
:+1:

cmelchior:
@nhachicha We normally don't reference issue numbers in change notes. Instead just write what was changed. In this case perhaps something like "Removed warnings when building with realm" or similar.

emanuelez:
:+1: apart from changelog :)

cmelchior:
:+1:

emanuelez:
:+1: if the conflict is only the changelog file :)

PR #1191, opened 2015-06-08
Fixed migration example

cmelchior:
This PR updates the MigrationExample to work with RealmConfiguration. It also changes the semantics slightly from the original #929 as that allowed you to upgrade from eg. v0 to v42 without a migration block. 

This was slightly odd and caused e.g. the MigrationExample to crash, so it has been changed so any attempt to set a version above 0 now require a migration block.

No changelog update as RealmConfigurations hasn't been released yet.

@emanuelez @nhachicha @kneth


kneth:
:+1:

nhachicha:
:+1:

nhachicha:
:+1:

PR #1200, opened 2015-06-09
Dynamic API

cmelchior:
This PR is a prerequisite for the new Migration API. It adds support for manipulating RealmObjects/RealmList using a Dynamic API instead of the current type safe one.

Implementation details:
DynamicRealmList currently only extends AbstractList instead of RealmList. This is to avoid also having to make a dynamic query API as well, which seemed a bit overkill for migration purposes. For a full blown dynamic API it will be needed, but for the purpose of Migrations I think it can be left out.

```
public interface DynamicRealmObjectInterface {
    boolean getBoolean(String fieldName);
    int getInt(String fieldName);
    short getShort(String fieldName);
    long getLong(String fieldName);
    byte getByte(String fieldName);
    float getFloat(String fieldName);
    double getDouble(String fieldName);
    byte[] getBlob(String fieldName);
    String getString(String fieldName);
    Date getDate(String fieldName);
    DynamicRealmObject getObject(String fieldName);
    DynamicRealmList getList(String fieldName);
    void setBoolean(String fieldName, boolean value);
    void setShort(String fieldName, short value);
    void setInt(String fieldName, int value);
    void setLong(String fieldName, long value);
    void setByte(String fieldName, byte value);
    void setFloat(String fieldName, float value);
    void setDouble(String fieldName, double value);
    void setString(String fieldName, String value);
    void setBlob(String fieldName, byte[] value);
    void setDate(String fieldName, Date value);
    void setObject(String fieldName, DynamicRealmObject value);
    void setList(String fieldName, DynamicRealmList list);
    boolean isNull(String fieldName);
    boolean hasField(String fieldName);
    String[] getFieldNames();
    void removeFromRealm();
}
```
 How to support the difference between null fields and not-nullable fields? Solution: add ```isNull``` for getters. Override setters: ```setInt(int)``` and ```setInt(Integer)```.

### DIFF
--- a/realm-annotations-processor/src/main/java/io/realm/processor/ClassMetaData.java
+++ b/realm-annotations-processor/src/main/java/io/realm/processor/ClassMetaData.java
@@ -367,7 +367,11 @@ public class ClassMetaData {
      * model classes.
      */
     public boolean isModelClass() {
-        return (!classType.toString().endsWith(".RealmObject") && !classType.toString().endsWith("RealmProxy"));
+        String type = classType.toString();
+        if (type.equals("io.realm.dynamic.DynamicRealmObject")) {
+            return false;
+        }
+        return (!type.endsWith(".RealmObject") && !type.endsWith("RealmProxy"));
     }
 
     public String getFullyQualifiedClassName() {
--- a/realm-jni/src/io_realm_internal_UncheckedRow.cpp
+++ b/realm-jni/src/io_realm_internal_UncheckedRow.cpp
@@ -288,3 +288,10 @@ JNIEXPORT jboolean JNICALL Java_io_realm_internal_UncheckedRow_nativeIsAttached
     TR_ENTER_PTR(nativeRowPtr)
     return ROW(nativeRowPtr)->is_attached();
 }
+
+JNIEXPORT jboolean JNICALL Java_io_realm_internal_UncheckedRow_nativeHasColumn
+  (JNIEnv* env, jobject obj, jlong nativeRowPtr, jstring columnName)
+{
+    jlong ndx = Java_io_realm_internal_UncheckedRow_nativeGetColumnIndex(env, obj, nativeRowPtr, columnName);
+    return ndx != to_jlong_or_not_found(realm::not_found);
+}
--- a/realm/src/androidTest/java/io/realm/DynamicRealmListTest.java
+++ b/realm/src/androidTest/java/io/realm/DynamicRealmListTest.java
@@ -1,0 +1,209 @@
+/*
+ * Copyright 2015 Realm Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.realm;
+
+import android.test.AndroidTestCase;
+
+import java.util.Date;
+
+import io.realm.dynamic.DynamicRealmList;
+import io.realm.dynamic.DynamicRealmObject;
+import io.realm.entities.AllJavaTypes;
+import io.realm.entities.Dog;
+
+public class DynamicRealmListTest extends AndroidTestCase {
+
+    private Realm realm;
+    private DynamicRealmObject dynamicObject;
+    private DynamicRealmList dynamicList;
+
+    @Override
+    protected void setUp() throws Exception {
+        RealmConfiguration realmConfig = new RealmConfiguration.Builder(getContext()).schema(AllJavaTypes.class, Dog.class).build();
+        Realm.deleteRealm(realmConfig);
+        realm = Realm.getInstance(realmConfig);
+        realm.beginTransaction();
+        AllJavaTypes obj = realm.createObject(AllJavaTypes.class);
+        obj.setFieldString("str");
+        obj.setFieldShort((short) 1);
+        obj.setFieldInt(1);
+        obj.setFieldLong(1);
+        obj.setFieldByte((byte) 4);
+        obj.setFieldFloat(1.23f);
+        obj.setFieldDouble(1.234d);
+        obj.setFieldBinary(new byte[]{1, 2, 3});
+        obj.setFieldBoolean(true);
+        obj.setFieldDate(new Date(1000));
+        obj.setFieldObject(obj);
+        obj.getFieldList().add(obj);
+        dynamicObject = new DynamicRealmObject(obj);
+        dynamicList = dynamicObject.getList(AllJavaTypes.FIELD_LIST);
+        realm.commitTransaction();
+    }
+
+    @Override
+    protected void tearDown() throws Exception {
+        super.tearDown();
+        realm.close();
+    }
+
+    public void testAddNullObjectThrows() {
+        realm.beginTransaction();
+        try {
+            dynamicList.add(null);
+            fail();
+        } catch (IllegalArgumentException ignored) {
+        } finally {
+            realm.cancelTransaction();
+        }
+    }
+
+    public void testAddWrongTableClassThrows() {
+        realm.beginTransaction();
+        Dog dog = realm.createObject(Dog.class);
+        try {
+            dynamicList.add(new DynamicRealmObject(dog));
+            fail();
+        } catch (IllegalArgumentException ignored) {
+        } finally {
+            realm.cancelTransaction();
+        }
+    }
+
+    public void testAddFromWrongRealmThrows() {
+        RealmConfiguration otherConfig = new RealmConfiguration.Builder(getContext()).name("realm2").build();
+        Realm.deleteRealm(otherConfig);
+        Realm realm2 = Realm.getInstance(otherConfig);
+        realm2.beginTransaction();
+        AllJavaTypes realm2Object = realm2.createObject(AllJavaTypes.class);
+        realm2.commitTransaction();
+
+        realm.beginTransaction();
+        try {
+            dynamicList.add(new DynamicRealmObject(realm2Object));
+            fail();
+        } catch (IllegalArgumentException ignored) {
+        } finally {
+            realm.cancelTransaction();
+            realm2.close();
+        }
+    }
+
+    public void testAddObject() {
+        realm.beginTransaction();
+        AllJavaTypes obj = realm.createObject(AllJavaTypes.class);
+        assertEquals(1, dynamicList.size());
+        dynamicList.add(new DynamicRealmObject(obj));
+        realm.commitTransaction();
+
+        assertEquals(2, dynamicList.size());
+    }
+
+    public void testClear() {
+        realm.beginTransaction();
+        dynamicList.clear();
+        realm.commitTransaction();
+
+        assertEquals(0, dynamicList.size());
+    }
+
+    public void testGetIllegalIndexThrows() {
+        int[] indexes = new int[] { -1, 1 };
+        for (int i : indexes) {
+            try {
+                dynamicList.get(i);
+                fail("Could retrieve from index" + i);
+            } catch (IndexOutOfBoundsException ignored) {
+            }
+        }
+    }
+
+    public void testGet() {
+        DynamicRealmObject listObject = dynamicList.get(0);
+        assertEquals(dynamicObject, listObject);
+    }
+
+    public void testSetIllegalLocationThrows() {
+        int[] indexes = new int[] { -1, 1 };
+        for (int i : indexes) {
+            try {
+                realm.beginTransaction();
+                dynamicList.set(i, dynamicObject);
+                fail("Could set index out of bounds " + i);
+            } catch (IndexOutOfBoundsException ignored) {
+            } finally {
+                realm.cancelTransaction();
+            }
+        }
+    }
+
+    public void testSetNullThrows() {
+        realm.beginTransaction();
+        try {
+            dynamicList.set(0, null);
+            fail();
+        } catch (IllegalArgumentException ignored) {
+        } finally {
+            realm.cancelTransaction();
+        }
+    }
+
+    public void testSetWrongTableClassThrows() {
+        realm.beginTransaction();
+        Dog dog = realm.createObject(Dog.class);
+        try {
+            dynamicList.set(0, new DynamicRealmObject(dog));
+            fail();
+        } catch (IllegalArgumentException ignored) {
+        } finally {
+            realm.cancelTransaction();
+        }
+    }
+
+    public void testSetWrongRealmThrows() {
+        RealmConfiguration otherConfig = new RealmConfiguration.Builder(getContext()).name("realm2").build();
+        Realm.deleteRealm(otherConfig);
+        Realm realm2 = Realm.getInstance(otherConfig);
+        realm2.beginTransaction();
+        AllJavaTypes realm2Object = realm2.createObject(AllJavaTypes.class);
+        realm2.commitTransaction();
+
+        realm.beginTransaction();
+        try {
+            dynamicList.set(0, new DynamicRealmObject(realm2Object));
+            fail();
+        } catch (IllegalArgumentException ignored) {
+        } finally {
+            realm.cancelTransaction();
+            realm2.close();
+        }
+    }
+
+    public void testSet() {
+        realm.beginTransaction();
+        AllJavaTypes obj = realm.createObject(AllJavaTypes.class);
+        obj.setFieldLong(2);
+        dynamicList.set(0, new DynamicRealmObject(obj));
+        realm.commitTransaction();
+
+        assertEquals(1, dynamicList.size());
+        assertEquals(new DynamicRealmObject(obj), dynamicList.get(0));
+    }
+
+    public void testSize() {
+        assertEquals(1, dynamicList.size());
+    }
+}
--- a/realm/src/androidTest/java/io/realm/DynamicRealmObjectTest.java
+++ b/realm/src/androidTest/java/io/realm/DynamicRealmObjectTest.java
@@ -1,0 +1,325 @@
+/*
+ * Copyright 2015 Realm Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.realm;
+
+import android.test.AndroidTestCase;
+
+import java.util.Arrays;
+import java.util.Date;
+import java.util.List;
+
+import io.realm.dynamic.DynamicRealmList;
+import io.realm.dynamic.DynamicRealmObject;
+import io.realm.entities.AllJavaTypes;
+import io.realm.entities.Dog;
+
+import static io.realm.internal.test.ExtraTests.assertArrayEquals;
+
+public class DynamicRealmObjectTest extends AndroidTestCase {
+
+    private Realm realm;
+    private DynamicRealmObject dObj;
+
+    @Override
+    protected void setUp() throws Exception {
+        RealmConfiguration realmConfig = new RealmConfiguration.Builder(getContext()).build();
+        Realm.deleteRealm(realmConfig);
+        realm = Realm.getInstance(realmConfig);
+        realm.beginTransaction();
+        AllJavaTypes obj = realm.createObject(AllJavaTypes.class);
+        obj.setFieldString("str");
+        obj.setFieldShort((short) 1);
+        obj.setFieldInt(1);
+        obj.setFieldLong(1);
+        obj.setFieldByte((byte) 4);
+        obj.setFieldFloat(1.23f);
+        obj.setFieldDouble(1.234d);
+        obj.setFieldBinary(new byte[]{1, 2, 3});
+        obj.setFieldBoolean(true);
+        obj.setFieldDate(new Date(1000));
+        obj.setFieldObject(obj);
+        obj.getFieldList().add(obj);
+        dObj = new DynamicRealmObject(obj);
+        realm.commitTransaction();
+    }
+
+    @Override
+    protected void tearDown() throws Exception {
+        super.tearDown();
+        realm.close();
+    }
+
+    public enum SupportedType {
+        BOOLEAN, SHORT, INT, LONG, BYTE, FLOAT, DOUBLE, STRING, BINARY, DATE, OBJECT, LIST;
+    }
+
+    // Test that all getters fail if given invalid field name
+    public void testGetXXXIllegalFieldNameThrows() {
+
+        // Set arguments
+        String linkedField = AllJavaTypes.FIELD_OBJECT + "." + AllJavaTypes.FIELD_STRING;
+        List<String> arguments = Arrays.asList(null, "foo", AllJavaTypes.FIELD_STRING, linkedField);
+        List<String> stringArguments = Arrays.asList(null, "foo", AllJavaTypes.FIELD_BOOLEAN, linkedField);
+
+        // Test all getters
+        for (SupportedType type : SupportedType.values()) {
+            List<String> args = (type == SupportedType.STRING) ? stringArguments : arguments;
+            try {
+                callGetter(type, args);
+                fail();
+            } catch(IllegalArgumentException ignored) {
+            }
+        }
+    }
+
+    private void callGetter(SupportedType type, List<String> arguments) {
+        for (String fieldName : arguments) {
+            switch(type) {
+                case BOOLEAN: dObj.getBoolean(fieldName); break;
+                case SHORT: dObj.getShort(fieldName); break;
+                case INT: dObj.getInt(fieldName); break;
+                case LONG: dObj.getLong(fieldName); break;
+                case BYTE: dObj.getByte(fieldName); break;
+                case FLOAT: dObj.getFloat(fieldName); break;
+                case DOUBLE: dObj.getDouble(fieldName); break;
+                case STRING: dObj.getString(fieldName); break;
+                case BINARY: dObj.getBlob(fieldName); break;
+                case DATE: dObj.getDate(fieldName); break;
+                case OBJECT: dObj.getObject(fieldName); break;
+                case LIST: dObj.getList(fieldName); break;
+                default:
+                    fail();
+            }
+        }
+    }
+
+    // Test that all getters fail if given invalid field name
+    public void testSetXXXIllegalFieldnameThrows() {
+
+        // Set arguments
+        String linkedField = AllJavaTypes.FIELD_OBJECT + "." + AllJavaTypes.FIELD_STRING;
+        List<String> arguments = Arrays.asList(null, "foo", AllJavaTypes.FIELD_STRING, linkedField);
+        List<String> stringArguments = Arrays.asList(null, "foo", AllJavaTypes.FIELD_BOOLEAN, linkedField);
+
+        // Test all getters
+        for (SupportedType type : SupportedType.values()) {
+            List<String> args = (type == SupportedType.STRING) ? stringArguments : arguments;
+            try {
+                callSetter(type, args);
+                fail();
+            } catch(IllegalArgumentException ignored) {
+            }
+        }
+    }
+
+    private void callSetter(SupportedType type, List<String> arguments) {
+        for (String fieldName : arguments) {
+            switch(type) {
+                case BOOLEAN: dObj.setBoolean(fieldName, false); break;
+                case SHORT: dObj.setShort(fieldName, (short) 1); break;
+                case INT: dObj.setInt(fieldName, 1); break;
+                case LONG: dObj.setLong(fieldName, 1L); break;
+                case BYTE: dObj.setByte(fieldName, (byte) 4); break;
+                case FLOAT: dObj.setFloat(fieldName, 1.23f); break;
+                case DOUBLE: dObj.setDouble(fieldName, 1.23d); break;
+                case STRING: dObj.setString(fieldName, "foo"); break;
+                case BINARY: dObj.setBlob(fieldName, new byte[]{}); break;
+                case DATE: dObj.getDate(fieldName); break;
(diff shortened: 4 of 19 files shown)


realm/src/androidTest/java/io/realm/DynamicRealmListTest.java
```diff
@@ -0,0 +1,205 @@
+package io.realm;/*
+ * Copyright 2014 Realm Inc.
```
emanuelez: 2015

realm/src/androidTest/java/io/realm/DynamicRealmObjectTest.java
```diff
@@ -0,0 +1,334 @@
+package io.realm;/*
+ * Copyright 2014 Realm Inc.
```
emanuelez: 2015

realm/src/main/java/io/realm/dynamic/DynamicRealmList.java
```diff
@@ -0,0 +1,100 @@
+/*
+ * Copyright 2014 Realm Inc.
```
emanuelez: 2015

realm/src/main/java/io/realm/dynamic/DynamicRealmList.java
```diff
@@ -0,0 +1,100 @@
...
+        linkView.set(location, object.row.getIndex());
+        return object;
+    }
+
+    @Override
+    public int size() {
+        return ((Long)linkView.size()).intValue();
+    }
+
+    private void checkIsValidObject(DynamicRealmObject object) {
```
emanuelez: I would rename this to `isObjectValid()`

realm/src/main/java/io/realm/dynamic/DynamicRealmList.java
```diff
@@ -0,0 +1,100 @@
...
+            throw new IllegalArgumentException("Cannot add a object belonging already in another Realm");
+        }
+        if (!linkView.getTable().hasSameSchema(object.row.getTable())) {
+            String expectedClass = linkView.getTable().getName();
+            String objectClassName = object.row.getTable().getName();
+            throw new IllegalArgumentException("Object is of type " + objectClassName + ". Expected " + expectedClass);
+        }
+    }
+
+    private void checkValidIndex(int location) {
```
emanuelez: I would rename this to `isIndexValid()`

realm/src/main/java/io/realm/dynamic/DynamicRealmObject.java
```diff
@@ -0,0 +1,209 @@
...
+ */
+public class DynamicRealmObject {
+
+     Realm realm;
+     Row row;
+
+    /**
+     * Creates a dynamic Realm object based on a row entry.
+     */
+    public DynamicRealmObject(Realm realm, Row row) {
```
emanuelez: I'm not sure I like this constructor since it exposes Row. How about having `DynamicRealmObject` extend `RealmObject` or something from which RealmObject could extend from. That was we could have `realm` and `row` in `RealmObject` as `protected` and be able to use them here. So the constructor could simply take a RealmObject.

realm/src/androidTest/java/io/realm/DynamicRealmObjectTest.java
```diff
@@ -0,0 +1,334 @@
...
+    private Realm realm;
+    private DynamicRealmObject dObj;
+
+    @Override
+    protected void setUp() throws Exception {
+        RealmConfiguration realmConfig = new RealmConfiguration.Builder(getContext()).schema(AllJavaTypes.class).build();
+        Realm.deleteRealm(realmConfig);
+        realm = Realm.getInstance(realmConfig);
+        realm.beginTransaction();
+        AllJavaTypes obj = realm.createObject(AllJavaTypes.class);
```
bmunkholm: AllJavaTypes should have it's fields renamed replacing "Column" with "Field" - but in a separate PR.

realm/src/androidTest/java/io/realm/DynamicRealmObjectTest.java
```diff
@@ -0,0 +1,334 @@
...
+
+    @Override
+    protected void tearDown() throws Exception {
+        super.tearDown();
+        realm.close();
+    }
+
+    // Test invalid input: <empty, non-existing field, wrong field type>
+    public void testGetBooleanIllegalArguments() {
+        List<String> arguments = Arrays.asList(null, "foo", "columnString");
```
bmunkholm: use constants instead of "columnString" etc.

realm/src/androidTest/java/io/realm/DynamicRealmObjectTest.java
```diff
@@ -0,0 +1,334 @@
...
+        for (String argument : arguments) {
+            try {
+                dObj.getBoolean(argument);
+                fail(argument + " should throw exception.");
+            } catch (IllegalArgumentException expected) {
+            }
+        }
+    }
+
+    public void testGetBoolean() {
```
bmunkholm: All these tests are so repetitive but dispersed. I would suggest to make a single testcase testing all GetXX, a single test case testing GetLinkedXX() and a third testing GetXXIllegalArguments.
That would be so much easier to get an overview of.

realm/src/androidTest/java/io/realm/entities/AllJavaTypes.java
```diff
@@ -0,0 +1,137 @@
...
+
+import io.realm.RealmList;
+import io.realm.RealmObject;
+import io.realm.annotations.Ignore;
+import io.realm.annotations.Index;
+import io.realm.annotations.PrimaryKey;
+
+public class AllJavaTypes extends RealmObject{
+
+    @Ignore private String columnIgnored;
```
bmunkholm: update naming to "field" instead of "column"

bmunkholm: add constants to the names so they can be used allover

realm/src/main/java/io/realm/dynamic/DynamicRealmList.java
```diff
@@ -0,0 +1,100 @@
...
+    public int size() {
+        return ((Long)linkView.size()).intValue();
+    }
+
+    private void checkIsValidObject(DynamicRealmObject object) {
+        if (object == null) {
+            throw new IllegalArgumentException("DynamicRealmList does not accept null values");
+        }
+        if (!realm.getPath().equals(object.realm.getPath())) {
+            throw new IllegalArgumentException("Cannot add a object belonging already in another Realm");
```
bmunkholm: a -> an

realm/src/main/java/io/realm/dynamic/DynamicRealmObject.java
```diff
@@ -0,0 +1,209 @@
...
+        if (row.getIndex() != other.row.getIndex()) return false;
+
+        return true;
+    }
+
+    @Override
+    public String toString() {
+        return super.toString(); // TODO How to iterate across all fields?
+    }
+}
```
bmunkholm: missing newline

realm/src/main/java/io/realm/dynamic/DynamicRealmObject.java
```diff
@@ -0,0 +1,209 @@
...
+     * Creates a dynamic Realm object based on a row entry.
+     */
+    public DynamicRealmObject(Realm realm, Row row) {
+        this.realm = realm;
+        this.row = row;
+    }
+
+    public boolean getBoolean(String fieldName) {
+        long columnIndex = row.getColumnIndex(fieldName);
+        checkFieldExists(columnIndex, fieldName);
```
bmunkholm: Should the validation be here and not in C++?
Doesn't Row already do the checks?
If it should be here, make one method to validate all instead of 3 method calls.

realm/src/main/java/io/realm/dynamic/DynamicRealmObject.java
```diff
@@ -0,0 +1,209 @@
...
+        int result = 17;
+        result = 31 * result + ((realmName != null) ? realmName.hashCode() : 0);
+        result = 31 * result + ((tableName != null) ? tableName.hashCode() : 0);
+        result = 31 * result + (int) (rowIndex ^ (rowIndex >>> 32));
+        return result;
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
```
bmunkholm: return statements on separate line.

realm/src/main/java/io/realm/dynamic/DynamicRealmList.java
bmunkholm: everywhere else we use "check" for these checks - so keep that or change it everywhere else. "check" is better that "is" due to missing boolean return type. But "validateIndex" may be better than "checkValidIndex", but again requires that it's changed everywhere else. If that should be done it must be a separate PR.

realm/src/main/java/io/realm/dynamic/DynamicRealmList.java
```diff
@@ -0,0 +1,100 @@
...
+        return true;
+    }
+
+    @Override
+    public void clear() {
+        linkView.clear();
+    }
+
+    @Override
+    public DynamicRealmObject get(int location) {
```
bmunkholm: "location" -> "index" (that's also the standard for AbstractList). Update everywhere.

realm/src/main/java/io/realm/dynamic/DynamicRealmList.java
```diff
@@ -0,0 +1,100 @@
...
+    public DynamicRealmObject set(int location, DynamicRealmObject object) {
+        checkIsValidObject(object);
+        checkValidIndex(location);
+        linkView.set(location, object.row.getIndex());
+        return object;
+    }
+
+    @Override
+    public int size() {
+        return ((Long)linkView.size()).intValue();
```
bmunkholm: How to handle lists larger than int? If we don't we might as well limit everything in Java to only handle int sized Realms.

realm/src/main/java/io/realm/dynamic/DynamicRealmObject.java
cmelchior: I agree, exposing internal classes is bugging me as well. I considered a few other options but neither seem that great either:

1) Moving DynamicRealmObject/DynamicRealmList to ```io.realm``` instead of ```io.realm.dynamic``` -> I would prefer to keep the main package space as clean as possible, i.e. only primary classes/interfaces.

2) Extending RealmObject doesn't work if we keep these classes in ```.dynamic``` as they cannot access the package protected fields from RealmObject.

Generally the current approach where we rely package protected methods/fields to avoid exposing internals is hurting our ability to modularise the code base. I don't have any golden answer though. But we should probably give this some more thought...

emanuelez: About 2) If you make the fields `protected` (not `package protected`) it should work!

realm/src/main/java/io/realm/dynamic/DynamicRealmList.java
cmelchior: What do you mean handle it? Even if every row only used 1 byte, a list with size = max_int would still hold 2 GB of data, so this would not be relevant for a mobile application?    

Fun fact from the Collections JavaDoc: "size(): Returns the number of elements in this collection. If this collection contains more than Integer.MAX_VALUE elements, returns Integer.MAX_VALUE."

So we should actually do a ```Math.min(linkView.size(), Integer.MAX_VALUE)``` here if want to honour that contract.

realm/src/main/java/io/realm/dynamic/DynamicRealmObject.java
cmelchior: No, row is non-checking as it up until now has been hidden behind the typed API where it couldn't fail. So error checking should be done here avoid adding unnecessary overhead to the typed API. Checks could probably be merged though.

kneth: Getting a column index, the C++ function will return `not_found` (or -1) if the name if not a valid column name.

The C++ getters for `Row` do check the type, and an exception will be thrown.

realm/src/main/java/io/realm/dynamic/DynamicRealmObject.java
```diff
@@ -0,0 +1,219 @@
...
+        if (row.getIndex() != other.row.getIndex()) {
+            return false;
+        }
+
+        return true;
+    }
+
+    @Override
+    public String toString() {
+        return super.toString(); // TODO How to iterate across all fields?
```
kneth: for(int i = 0; i < row.getColumnCount(); i++) {
        ...
    }

kneth:
Why doesn't `RealmDynamicObject` have setters?

realm/src/main/java/io/realm/dynamic/DynamicRealmObject.java
cmelchior: Smart ass :) My problem was accessing the data since AFAIK there is no generic row.getObject(index) function, so you would have to check the column type for each index in order to use the proper getX function. It's doable sure, but seemed like a lot of work if we had a smarter function. Maybe using the json() function in core?

cmelchior:
I hadn't gotten around to it yet (WIP), and my driver use case didn't require them. But they are incoming :)

realm/src/main/java/io/realm/dynamic/DynamicRealmObject.java
cmelchior: Hmm, seems you are right @kneth. I wonder if we should rethink this? Because our typed API should really not need the overhead of checking row/col/type on each access, while the dynamic API would?

realm/src/main/java/io/realm/dynamic/DynamicRealmObject.java
cmelchior: Actually extending RealmObject doesn't work at all since you would have to parse the original RealmObject in as a parameter. We can't access the realm/row of that class unless they are in the same package or are public :(

```
public DynamicRealmObject(RealmObject obj) {
  // Boom! Unless RealmObject and DynamicRealmObject are in the 
  // same package or obj.row is public.   
  this.row = obj.row; 
}
```

Users shouldn't need to create these anyways. They should always be created using e.g. something like the below when doing migrations. So the JavaDoc should definitely reflect this when adding the new migration API. 

```
DynamicRealmObject obj = tableSpec.createObject();
```

emanuelez: Hmmm... I think we should really strive to avoid any reference to internal classes. More brains are needed! @nhachicha @beeender ? Any suggestion?

cmelchior:
This has left WIP and is ready for review: @realm/java 

Currently the constructor for DynamicRealmObject takes a Row object. This is not ideal, so any feedback on work arounds that doesn't involve us exposing internal API's would be helpful.

realm-annotations-processor/src/main/java/io/realm/processor/ClassMetaData.java
```diff
@@ -367,7 +367,11 @@ public String getSimpleClassName() {
      * model classes.
      */
     public boolean isModelClass() {
-        return (!classType.toString().endsWith(".RealmObject") && !classType.toString().endsWith("RealmProxy"));
+        String type = classType.toString();
+        if (type.equals("io.realm.dynamic.DynamicRealmObject")) {
+            return false;
+        }
+        return (!type.endsWith(".RealmObject") && !type.endsWith("RealmProxy"));
```
kneth: Can't you collapse it to one line?

realm/src/androidTest/java/io/realm/DynamicRealmListTest.java
```diff
@@ -0,0 +1,204 @@
...
+        } finally {
+            realm.cancelTransaction();
+            realm2.close();
+        }
+    }
+
+    public void testAddObject() {
+        realm.beginTransaction();
+        AllJavaTypes obj = realm.createObject(AllJavaTypes.class);
+        dynamicList.add(new DynamicRealmObject(obj));
```
kneth: Checking that `dynamicList` only has one element before`add` would ensure that `add` only adds one element.

realm/src/androidTest/java/io/realm/DynamicRealmObjectTest.java
```diff
@@ -0,0 +1,317 @@
...
+        List<String> arguments = Arrays.asList(null, "foo", AllJavaTypes.FIELD_STRING, linkedField);
+        List<String> stringArguments = Arrays.asList(null, "foo", AllJavaTypes.FIELD_BOOLEAN, linkedField);
+
+        // Test all getters
+        for (SupportedType type : SupportedType.values()) {
+            List<String> args = (type == SupportedType.STRING) ? stringArguments : arguments;
+            try {
+                callGetter(type, args);
+                fail();
+            } catch(IllegalArgumentException expected) {
```
kneth: `ignored`

realm/src/androidTest/java/io/realm/DynamicRealmObjectTest.java
```diff
@@ -0,0 +1,317 @@
...
+        List<String> arguments = Arrays.asList(null, "foo", AllJavaTypes.FIELD_STRING, linkedField);
+        List<String> stringArguments = Arrays.asList(null, "foo", AllJavaTypes.FIELD_BOOLEAN, linkedField);
+
+        // Test all getters
+        for (SupportedType type : SupportedType.values()) {
+            List<String> args = (type == SupportedType.STRING) ? stringArguments : arguments;
+            try {
+                callSetter(type, args);
+                fail();
+            } catch(IllegalArgumentException expected) {
```
kneth: `ignored`

realm/src/androidTest/java/io/realm/DynamicRealmObjectTest.java
```diff
@@ -0,0 +1,317 @@
...
+    public void testSetObjectWrongTypeThrows() {
+        realm.beginTransaction();
+        AllJavaTypes obj = realm.createObject(AllJavaTypes.class);
+        Dog otherObj = realm.createObject(Dog.class);
+        DynamicRealmObject dynamicObj = new DynamicRealmObject(obj);
+        DynamicRealmObject dynamicWrongType = new DynamicRealmObject(otherObj);
+        try {
+            dynamicObj.setObject(AllJavaTypes.FIELD_OBJECT, dynamicWrongType);
+            fail();
+        } catch (IllegalArgumentException ignore) {
```
kneth: `ignored`

realm-jni/src/io_realm_internal_UncheckedRow.cpp
```diff
@@ -288,3 +288,10 @@ JNIEXPORT jboolean JNICALL Java_io_realm_internal_UncheckedRow_nativeIsAttached
     TR_ENTER_PTR(nativeRowPtr)
     return ROW(nativeRowPtr)->is_attached();
 }
+
+JNIEXPORT jboolean JNICALL Java_io_realm_internal_UncheckedRow_nativeHasField
```
bmunkholm: This method seems to have no value. Why not just make this "helper" in Java

realm-annotations-processor/src/main/java/io/realm/processor/ClassMetaData.java
cmelchior: Yes
```return !type.equals("io.realm.dynamic.DynamicRealmObject") && (!type.endsWith(".RealmObject") && !type.endsWith("RealmProxy"))```

Which is horrible to read and understand IMO

File: realm-jni/src/io_realm_internal_UncheckedRow.cpp (hunk 23 above)
Comment (cmelchior) [inline realm-jni/src/io_realm_internal_UncheckedRow.cpp]: Actually it does. In order to consolidate error messages you can no longer use getColumnIndex(fieldName) with illegal field names, it will throw an exception instead of returning -1. So the implicit behaviour of getColumnIndex to check for existence now require an explicit function.

File: realm/src/main/java/io/realm/RealmObject.java [hunk 24]
```diff
@@ -106,4 +106,22 @@ public void removeFromRealm() {
...
+     * @return The Realm this object belongs to or {code null} if it is a standalone object.
+     */
+    protected static Realm getRealm(RealmObject obj) {
+        return obj.realm;
+    }
+
+    /**
+     * Returns the {@link Row} representing this object. Internal use only.
+     *
+     * @return The {@link Row} his object belongs to or {code null} if it is a standalone object.
```
nhachicha: _his object_ ==> this object

realm/src/main/java/io/realm/dynamic/DynamicRealmList.java
```diff
@@ -0,0 +1,139 @@
...
+    public int size() {
+        long size = linkView.size();
+        return size < Integer.MAX_VALUE ? (int) size : Integer.MAX_VALUE;
+    }
+
+    private void checkIsValidObject(DynamicRealmObject object) {
+        if (object == null) {
+            throw new IllegalArgumentException("DynamicRealmList does not accept null values");
+        }
+        if (!realm.getPath().equals(object.realm.getPath())) {
```
nhachicha: shouldn't we check for the configuration instead?

realm/src/androidTest/java/io/realm/TestHelper.java
```diff
@@ -30,6 +30,9 @@
 import java.util.Random;
 import java.nio.charset.Charset;
 
+import io.realm.internal.CheckedRow;
+import io.realm.internal.Row;
+
```
nhachicha: unused imports

cmelchior:
Bump @realm/java

realm/src/main/java/io/realm/dynamic/DynamicRealmObject.java
```diff
@@ -0,0 +1,513 @@
...
+public class DynamicRealmObject extends RealmObject {
+
+     Realm realm;
+     Row row;
+
+    /**
+     * Creates a dynamic Realm object based on a existing object.
+     */
+    public DynamicRealmObject(RealmObject obj) {
+        Row row = RealmObject.getRow(obj);
```
emanuelez: You should check for `obj` being `null` before using it

realm/src/main/java/io/realm/dynamic/DynamicRealmObject.java
```diff
@@ -0,0 +1,513 @@
...
+     */
+    public void setObject(String fieldName, DynamicRealmObject value) {
+        long columnIndex = row.getColumnIndex(fieldName);
+        if (value == null) {
+            row.nullifyLink(columnIndex);
+        } else {
+            if (value.realm == null || value.row == null) {
+                throw new IllegalArgumentException("Cannot link to objects that are not part of the Realm.");
+            }
+            if (value.realm != null && !realm.getPath().equals(value.realm.getPath())) {
```
emanuelez: Should configuration be used instead of path here

realm/src/main/java/io/realm/dynamic/DynamicRealmList.java
```diff
@@ -0,0 +1,139 @@
...
+            String expectedClass = linkView.getTable().getName();
+            String objectClassName = object.row.getTable().getName();
+            throw new IllegalArgumentException("Object is of type " + objectClassName + ". Expected " + expectedClass);
+        }
+    }
+
+    private void checkValidIndex(int index) {
+        long size = linkView.size();
+        if (index < 0 || index >= size) {
+            throw new IndexOutOfBoundsException(String.format("Invalid index: %s. Valid range is [%s, %s]", index, 0, size - 1));
```
kneth: The format should be `%d` and not `%s`.

realm/src/main/java/io/realm/dynamic/DynamicRealmObject.java
```diff
@@ -0,0 +1,516 @@
...
+
+    /**
+     * Creates a dynamic Realm object based on a existing object.
+     */
+    public DynamicRealmObject(RealmObject obj) {
+        if (obj == null) {
+            throw new IllegalArgumentException("Non-null object must be provided.");
+        }
+        Row row = RealmObject.getRow(obj);
+        if (obj == null || row == null) {
```
kneth: How can `obj` be `null`?

realm/src/main/java/io/realm/dynamic/DynamicRealmObject.java
```diff
@@ -0,0 +1,516 @@
...
+     */
+    public void setObject(String fieldName, DynamicRealmObject value) {
+        long columnIndex = row.getColumnIndex(fieldName);
+        if (value == null) {
+            row.nullifyLink(columnIndex);
+        } else {
+            if (value.realm == null || value.row == null) {
+                throw new IllegalArgumentException("Cannot link to objects that are not part of the Realm.");
+            }
+            if (value.realm != null && !realm.getConfiguration().equals(value.realm.getConfiguration())) {
```
kneth: When can `value.realm` be `null`?

realm/src/main/java/io/realm/internal/UncheckedRow.java
```diff
@@ -283,4 +288,5 @@ public boolean isAttached() {
     protected native void nativeNullifyLink(long nativeRowPtr, long columnIndex);
     protected static native void nativeClose(long nativeRowPtr);
     protected native boolean nativeIsAttached(long nativeRowPtr);
+    protected native boolean nativeHasField(long nativeRowPtr, String columnName);
```
kneth: Can be name be `nativeHasColumn`? It is consistent with the other column functions for example `nativeGetColumnCount()`.

realm/src/main/java/io/realm/dynamic/DynamicRealmObject.java
emanuelez: when it's a standalone object

realm/src/androidTest/java/io/realm/DynamicRealmListTest.java
```diff
@@ -0,0 +1,205 @@
...
+        obj.setFieldInt(1);
+        obj.setFieldLong(1);
+        obj.setFieldFloat(1.23f);
+        obj.setFieldDouble(1.234d);
+        obj.setFieldBinary(new byte[]{1, 2, 3});
+        obj.setFieldBoolean(true);
+        obj.setFieldDate(new Date(1000));
+        obj.setFieldObject(obj);
+        obj.getFieldList().add(obj);
+        dynamicObject = new DynamicRealmObject(obj);
```
beeender: Byte is missing in this Test.

realm/src/androidTest/java/io/realm/DynamicRealmListTest.java
```diff
@@ -0,0 +1,205 @@
...
+        }
+    }
+
+    public void testGet() {
+        DynamicRealmObject listObject = dynamicList.get(0);
+        assertEquals(dynamicObject, listObject);
+    }
+
+    public void testSetIllegalLocationThrows() {
+        int[] indexes = new int[] { -1, 1 };
```
beeender: Don't you have to beginTransaction first?

realm/src/main/java/io/realm/dynamic/DynamicRealmObject.java
kneth: Just a few lines above, an exception is thrown if `value.realm` is `null`. If we get to this line, `value.realm` cannot be `null`, right?

realm/src/androidTest/java/io/realm/DynamicRealmListTest.java
cmelchior: We are testing for index before checking for beginTransaction so the test is correct. Although you could argue that the current test depends on the internal implementation of ```set()```. I can change  it if you think it is necessary?

realm/src/androidTest/java/io/realm/DynamicRealmListTest.java
cmelchior: It has been added

realm/src/androidTest/java/io/realm/DynamicRealmListTest.java
beeender: Oh, yes, I want to argue that exactly. I still would like to have that to avoid confusion in the future  :smile:

cmelchior: I'll change it :)

cmelchior:
All comments should be fixed: @realm/java

realm/src/main/java/io/realm/dynamic/DynamicRealmList.java
```diff
@@ -0,0 +1,139 @@
...
+     *
+     * @param object the object to add.
+     * @return true
+     * @throws IllegalArgumentException if object is either {@code null} or has the wrong type.
+     */
+    @Override
+    public boolean add(DynamicRealmObject object) {
+        checkIsValidObject(object);
+        linkView.add(object.row.getIndex());
+        return true;
```
beeender: A rare case could happen here. If the `size()` has been reached `Integer.MAX_VALUE`, `add` will succeed and return true. But user can never get the Object. Should it be more clear that just throw a IndexOutOfBoundsException here?

cmelchior: Actually this is the specified behaviour. Acessing objects after Integer.MAX require an iterator. Blame the original Java API designer ☺

realm/src/main/java/io/realm/dynamic/DynamicRealmObject.java
```diff
@@ -0,0 +1,526 @@
...
+ * Using a DynamicRealmObject is slower than using the regular model class.
+ */
+public class DynamicRealmObject extends RealmObject {
+
+     Realm realm;
+     Row row;
+
+    /**
+     * Creates a dynamic Realm object based on a existing object.
+     */
```
beeender: Param doc missed

realm/src/main/java/io/realm/dynamic/DynamicRealmObject.java
```diff
@@ -0,0 +1,526 @@
...
+            return new DynamicRealmObject(realm, linkRow);
+        }
+    }
+
+    /**
+     * Returns the {@link io.realm.RealmList} of objects being linked to from this field. This list is returned
+     * as a {@link DynamicRealmList}.
+     *
+     * @param fieldName Name of field.
+     * @return The {@link DynamicRealmList} representation of the RealmList.
```
beeender: Lower case "The" following `@return`?

realm/src/main/java/io/realm/dynamic/DynamicRealmObject.java
```diff
@@ -0,0 +1,526 @@
...
+        if (fieldName == null || fieldName.isEmpty()) {
+            return false;
+        }
+        return row.hasColumn(fieldName);
+    }
+
+    /**
+     * Returns the list of field names on this object.
+     *
+     * @return List of field names on this objects or the empty list if the object doesn't have any fields.
```
beeender: Same, "List" -> "list"

realm/src/main/java/io/realm/dynamic/DynamicRealmList.java
beeender: Ha, got it.

kneth:
:+1:

nhachicha:
:+1:

emanuelez:
:+1: :shipit:

beeender:
+1

PR #1203, opened 2015-06-10
Use s3cmd to perform the deployment of releases

### LINKED ISSUE #893: Automate release in Jenkins

emanuelez:
Fixes #893 

This assumes that the S3 credentials have already been setup with s3cmd --configure on the hosting machine

### DIFF
--- a/realm/build.gradle
+++ b/realm/build.gradle
@@ -250,33 +250,31 @@ examples.each() { src, dest ->
     }
 }
 
-apply plugin: 's3'
+task uploadDistributionPackage(type: Exec) {
+    examples.each { src, dest ->
+        dependsOn "distributionMonkey${dest}"
+    }
+    commandLine 's3cmd', 'put', "${rootDir}/realm-java-${version}.zip", 's3://static.realm.io/downloads/java/'
+}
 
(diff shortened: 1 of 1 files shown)


emanuelez:
@cmelchior @nhachicha @kneth

kneth:
:+1:

nhachicha:
Does it make sense to override the default  `xms xmx` JVM options only for the upload task (since we're having an OOM :) https://docs.gradle.org/current/dsl/org.gradle.api.tasks.Exec.html#org.gradle.api.tasks.Exec:environment

emanuelez:
It was actually a Permgen error, so those flags would not really help. We could increase the permgen space, but that's quite a clear symptom that the library is including way too many classes and methods for our needs.

cmelchior:
:+1:

nhachicha:
@emanuelez are we using Java 8 we shouldn't get permgen error, it was removed AFAIK!?

emanuelez:
Yeah, it has been renamed/repurposed to metadata memory, which actually uses native memory.
But still, such a simple dependency should not be so make your permgen memory run out even on Java7

PR #1211, opened 2015-06-12
size() now respect the Collection.size() contract.

cmelchior:
Small fix for an issue that crept up when reviewing DynamicObject. 

Realm theoretically support collections larger that Integer.MAX_VALUE. Collections.size() has a return type of int and right now we just cast the size potentially overflowing the value. 

For reference: http://docs.oracle.com/javase/7/docs/api/java/util/Collection.html#size()

@emanuelez @nhachicha @kneth 

No changelog, because....really?


kneth:
:+1:

emanuelez:
:+1:

PR #1216, opened 2015-06-16
Allow the Realm Gradle plugin to modify the bytecode

emanuelez:
The bytecode is modified in two ways:
 * Custom accessors are added to model classes
    They are called .realm[GS]etter$<fieldName>
 * Such accessors are used instead of field accesses

This is part of making model classes feel like normal POJOs.

### DIFF
--- a/gradle-plugin/plugin/build.gradle
+++ b/gradle-plugin/plugin/build.gradle
@@ -1,6 +1,12 @@
 apply plugin: 'groovy'
 
+repositories {
+    jcenter()
+}
+
 dependencies {
     compile gradleApi()
     compile localGroovy()
+    compile 'com.android.tools.build:gradle:1.2.3'
+    compile 'org.javassist:javassist:3.18.2-GA'
 }
--- a/gradle-plugin/plugin/src/main/groovy/io/realm/gradle/Realm.groovy
+++ b/gradle-plugin/plugin/src/main/groovy/io/realm/gradle/Realm.groovy
@@ -1,13 +1,126 @@
 package io.realm.gradle
-
+import com.android.build.gradle.AppPlugin
+import com.android.build.gradle.LibraryPlugin
+import javassist.*
+import javassist.expr.ExprEditor
+import javassist.expr.FieldAccess
+import org.gradle.api.GradleException
 import org.gradle.api.Plugin
 import org.gradle.api.Project
+import org.gradle.api.file.ConfigurableFileTree
+import org.gradle.api.file.FileCollection
+import org.gradle.api.tasks.compile.JavaCompile
 
 class Realm implements Plugin<Project> {
 
     @Override
     void apply(Project project) {
-    	project.repositories.add(project.repositories.jcenter())
+        // Make sure the project is either an Android application or library
+        def isAndroidApp = project.plugins.withType(AppPlugin)
+        def isAndroidLib = project.plugins.withType(LibraryPlugin)
+        if (!isAndroidApp && !isAndroidLib) {
+            throw new GradleException("'android' or 'android-library' plugin required.")
+        }
+
+        // Configure the project
+        project.repositories.add(project.repositories.jcenter())
         project.dependencies.add('compile', 'io.realm:realm-android:0.82.0') // TODO: make version dynamic
+
+        // Find all the variants
+        final def variants
+        if (isAndroidApp) {
+            variants = project.android.applicationVariants
+        } else {
+            variants = project.android.libraryVariants
+        }
+
+        // Do the bytecode manipulation
+        variants.all { variant ->
+            variant.dex.doFirst() {// Get the jar files of the imported libraries
+                JavaCompile javaCompile = variant.javaCompile
+                FileCollection classpathFileCollection = project.files(project.android.bootClasspath)
+                classpathFileCollection += javaCompile.classpath
+
+                // Initialize the class container for Javassist
+                ClassPool classPool = new ClassPool();
+
+                // Add the classpath jar files
+                classpathFileCollection.each {
+                    classPool.appendClassPath(it.toString())
+                }
+
+                // Append the folder containing the .class files for the project
+                classPool.appendClassPath(javaCompile.destinationDir.path)
+
+                // Find the class files
+                ConfigurableFileTree classFiles = project.fileTree(javaCompile.destinationDir);
+                classFiles.include("**/*.class");
+
+
+                def classCache = [:]
+
+                // Add custom accessors to the model classes
+                classFiles.each { classFile ->
+                    CtClass clazz = openClass(classFile, classPool)
+                    classCache[classFile] = clazz
+                    addRealmAccessors(clazz)
+                }
+
+                // Replace access fields with using the custom accessors
+                classFiles.each { classFile ->
+                    def clazz = classCache[classFile]
+                    useRealmAccessors(clazz as CtClass)
+                    (clazz as CtClass).writeFile(javaCompile.destinationDir.absolutePath)
+                }
+            }
+        }
+    }
+
+    private static void useRealmAccessors(CtClass clazz) {
+        clazz.getDeclaredBehaviors().each { behavior ->
+            if (!behavior.name.startsWith('realmGetter$') && !behavior.name.startsWith('realmSetter$')) {
+                behavior.instrument(new ExprEditor() {
+                    @Override
+                    void edit(FieldAccess fieldAccess) throws CannotCompileException {
+                        if (fieldAccess.field?.declaringClass?.superclass?.name?.equals("io.realm.RealmObject")) {
+                            println("Realm: Manipulating ${clazz.simpleName}.${behavior.name}(): ${fieldAccess.fieldName}")
+                            def fieldName = fieldAccess.fieldName
+                            if (fieldAccess.isReader()) {
+                                fieldAccess.replace('$_ = $0.realmGetter$' + fieldName + '();')
+                            } else if (fieldAccess.isWriter()) {
+                                fieldAccess.replace('$0.realmSetter$' + fieldName + '($1);')
+                            }
+                        }
+                    }
+                })
+            }
+        }
+    }
+
+    private static void addRealmAccessors(CtClass clazz) {
+        if (clazz.superclass.name.equals("io.realm.RealmObject")) {
+            println("Realm: Adding accessors to ${clazz.simpleName}")
+            clazz.declaredFields.each { field ->
+                // TODO: eventually filter out ignored fields. This would add a dependency on the annotations project
+                clazz.addMethod(CtNewMethod.getter("realmGetter\$${field.name}", field))
+                clazz.addMethod(CtNewMethod.setter("realmSetter\$${field.name}", field))
+            }
+        }
+    }
+
+    private static CtClass openClass(File classFile, ClassPool classPool) {
+        InputStream stream = null
+        CtClass clazz = null
+        try {
+            stream = new DataInputStream(new BufferedInputStream(new FileInputStream(classFile)))
+            clazz = classPool.makeClass(stream);
+        } catch (Exception ignored) {
+            throw new GradleException("Could not load a class for modification")
+        } finally {
+            if (stream != null) {
+                stream.close()
+            }
+        }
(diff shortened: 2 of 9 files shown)


emanuelez:
@cmelchior @nhachicha @kneth

gradle-plugin/plugin/build.gradle
```diff
@@ -1,6 +1,12 @@
 apply plugin: 'groovy'
 
+repositories {
+    jcenter()
+}
+
 dependencies {
     compile gradleApi()
     compile localGroovy()
+    compile 'com.android.tools.build:gradle:1.2.3'
```
cmelchior: This should be pulled from the global build.gradle instead

gradle-plugin/plugin/src/main/groovy/io/realm/gradle/Realm.groovy
```diff
@@ -1,13 +1,126 @@
...
+        // Make sure the project is either an Android application or library
+        def isAndroidApp = project.plugins.withType(AppPlugin)
+        def isAndroidLib = project.plugins.withType(LibraryPlugin)
+        if (!isAndroidApp && !isAndroidLib) {
+            throw new GradleException("'android' or 'android-library' plugin required.")
+        }
+
+        // Configure the project
+        project.repositories.add(project.repositories.jcenter())
         project.dependencies.add('compile', 'io.realm:realm-android:0.80.3') // TODO: make version dynamic
```
cmelchior: Is this incoming?

gradle-plugin/plugin/src/main/groovy/io/realm/gradle/Realm.groovy
```diff
@@ -1,13 +1,126 @@
...
+        final def variants
+        if (isAndroidApp) {
+            variants = project.android.applicationVariants
+        } else {
+            variants = project.android.libraryVariants
+        }
+
+        // Do the bytecode manipulation
+        variants.all { variant ->
+            variant.dex.doFirst() {// Get the jar files of the imported libraries
```
cmelchior: How does this work with library projects and pre-dexing? Ie. if the app is accessing fields of a Realm class defined in a library project that has been pre-dexed?

gradle-plugin/plugin/src/main/groovy/io/realm/gradle/Realm.groovy
```diff
@@ -1,13 +1,126 @@
...
+                }
+
+                // Append the folder containing the .class files for the project
+                classPool.appendClassPath(javaCompile.destinationDir.path)
+
+                // Find the class files
+                ConfigurableFileTree classFiles = project.fileTree(javaCompile.destinationDir);
+                classFiles.include("**/*.class");
+
+
```
cmelchior: Remove whitespace or add comment to indicate new "section"

gradle-plugin/plugin/src/main/groovy/io/realm/gradle/Realm.groovy
```diff
@@ -1,13 +1,126 @@
...
+    }
+
+    private static void useRealmAccessors(CtClass clazz) {
+        clazz.getDeclaredBehaviors().each { behavior ->
+            if (!behavior.name.startsWith('realmGetter$') && !behavior.name.startsWith('realmSetter$')) {
+                behavior.instrument(new ExprEditor() {
+                    @Override
+                    void edit(FieldAccess fieldAccess) throws CannotCompileException {
+                        if (fieldAccess.field?.declaringClass?.superclass?.name?.equals("io.realm.RealmObject")) {
+                            println("Realm: Manipulating ${clazz.simpleName}.${behavior.name}(): ${fieldAccess.fieldName}")
```
cmelchior: Do we really want to flood the log with this?

gradle-plugin/plugin/src/main/groovy/io/realm/gradle/Realm.groovy
```diff
@@ -1,13 +1,126 @@
...
+                        }
+                    }
+                })
+            }
+        }
+    }
+
+    private static void addRealmAccessors(CtClass clazz) {
+        if (clazz.superclass.name.equals("io.realm.RealmObject")) {
+            println("Realm: Adding accessors to ${clazz.simpleName}")
```
cmelchior: Same

gradle-plugin/sample/src/main/java/io/realm/example/sample/models/Person.java
```diff
@@ -16,10 +16,12 @@
 package io.realm.example.sample.models;
 
 import io.realm.RealmObject;
+import io.realm.annotations.Ignore;
 
 public class Person extends RealmObject {
```
cmelchior: Perhaps also add a setter that mutates the data?
I assume that removing limitations on methods etc. will be another PR?

gradle-plugin/plugin/build.gradle
emanuelez: Not yet. The plugin is still its own project

gradle-plugin/plugin/src/main/groovy/io/realm/gradle/Realm.groovy
emanuelez: I had some discussions with Nabil about this... we will take it all on Thursday cause it's important and has dramatic consequences

cmelchior:
Looks like a great start :dancers: 
Things I would like to see:
- How does this effect build times?
- Does it effect the debugging experience?

I am a bit concerned about the interaction with library projects and Realm classes exposed from there. This probably needs to be tested more.

gradle-plugin/plugin/src/main/groovy/io/realm/gradle/Realm.groovy
emanuelez: It's a use-case I thought about and it will require some work. Another topic for Thursday

gradle-plugin/plugin/src/main/groovy/io/realm/gradle/Realm.groovy
emanuelez: We do for the annotation processor. Either we do both and none IMO

gradle-plugin/sample/src/main/java/io/realm/example/sample/models/Person.java
emanuelez: That will require changes to the annotation processor, which is not part of this PR.
The changes to the annotation processor will have a huge impact on the realm unit tests

emanuelez:
As far as I could see build times are not affected dramatically. We will be able to do proper testing when every piece will be in piece.

The debugging experience should not change since as far as the IDE is concerned those are still field accesses and cannot be stepped into. Again, more testing needed when the changes to the AP are in place.

I see some pretty elegant solutions about the Realm libraries use-case, which we can discuss F2F :)

gradle-plugin/plugin/src/main/groovy/io/realm/gradle/Realm.groovy
```diff
@@ -1,13 +1,126 @@
...
+        clazz.getDeclaredBehaviors().each { behavior ->
+            if (!behavior.name.startsWith('realmGetter$') && !behavior.name.startsWith('realmSetter$')) {
+                behavior.instrument(new ExprEditor() {
+                    @Override
+                    void edit(FieldAccess fieldAccess) throws CannotCompileException {
+                        if (fieldAccess.field?.declaringClass?.superclass?.name?.equals("io.realm.RealmObject")) {
+                            println("Realm: Manipulating ${clazz.simpleName}.${behavior.name}(): ${fieldAccess.fieldName}")
+                            def fieldName = fieldAccess.fieldName
+                            if (fieldAccess.isReader()) {
+                                fieldAccess.replace('$_ = $0.realmGetter$' + fieldName + '();')
```
cmelchior: We might want to consider replacing $ with _, I think i saw some discussion on this in Dagger repo at one point and some editors (perhaps Eclipse) has problems with this. I will see if I can dig the discussion up.

emanuelez: Ok. It should be fine according to the Java Language Specification: http://docs.oracle.com/javase/specs/jls/se7/html/jls-3.html#jls-3.8

gradle-plugin/sample/build.gradle
```diff
@@ -14,11 +14,11 @@ apply plugin: 'realm'
 
 android {
     compileSdkVersion 22
-    buildToolsVersion "21.1.2"
+    buildToolsVersion "22.0.0"
 
     defaultConfig {
         applicationId "io.realm.example.sample"
-        minSdkVersion 15
+        minSdkVersion 22
```
kneth: We have an issue arguing that we can use version 21.

kneth:
Please post the provide line for `changelog.md`.

emanuelez:
There's no user facing changes since there's no way for a user to apply the plugin yet

gradle-plugin/plugin/src/main/groovy/io/realm/gradle/Realm.groovy
```diff
@@ -1,13 +1,126 @@
 package io.realm.gradle
-
+import com.android.build.gradle.AppPlugin
+import com.android.build.gradle.LibraryPlugin
+import javassist.*
```
nhachicha: shouldn't be allowed with new checkstyle :)

gradle-plugin/plugin/src/main/groovy/io/realm/gradle/Realm.groovy
```diff
@@ -1,13 +1,126 @@
...
-    	project.repositories.add(project.repositories.jcenter())
+        // Make sure the project is either an Android application or library
+        def isAndroidApp = project.plugins.withType(AppPlugin)
+        def isAndroidLib = project.plugins.withType(LibraryPlugin)
+        if (!isAndroidApp && !isAndroidLib) {
+            throw new GradleException("'android' or 'android-library' plugin required.")
+        }
+
+        // Configure the project
+        project.repositories.add(project.repositories.jcenter())
```
nhachicha: What happen if the user already pulled `jcenter()`?

gradle-plugin/plugin/src/main/groovy/io/realm/gradle/Realm.groovy
nhachicha: we can use it with a Debug flag, otherwise it will slow down the gradle task

gradle-plugin/sample/build.gradle
```diff
@@ -33,5 +33,5 @@ android {
 
 dependencies {
     compile fileTree(dir: 'libs', include: ['*.jar'])
-    compile 'com.android.support:appcompat-v7:22.0.0'
+    //compile 'com.android.support:appcompat-v7:22.0.0'
```
nhachicha: remove dead code

gradle-plugin/sample/src/main/java/io/realm/example/sample/models/Person.java
```diff
@@ -16,10 +16,12 @@
...
     private String name;
     private int age;
+    @Ignore public String note;
```
nhachicha: new checkstyle `@Ignore` on new line

nhachicha:
since this is going to be a long WIP (tests, refactor) maybe we can merge against a feature branch rather than `master`. so you can iterate quickly without worrying to break/regression master

zaki50:
Should we consider using Transform API newly introduced by android gradle plugin 1.4.0-beta2?

http://tools.android.com/tech-docs/new-build-system/transform-api

cmelchior:
Probably yes. The good thing about having a Realm gradle plugin is that we can switch to using the transform API when it becomes available.

emanuelez:
I'm looking into it. It's very promising, but it would force our users to use Gradle 1.4 which might be a bit too much right now

dalinaum:
Should we add the `wip` tag on this?

Saketme:
@emanuelez Do not worry about forcing users to gradle 1.4. It's a one time download thing. And I guess most of us are already on 1.5

emanuelez:
Superseded by #2196

PR #1227, opened 2015-06-20
Optimised thread check.

cmelchior:
Our current check for testing for the correct thread is stupidly expensive and it is executed for each accessor. It should be enough to just compare thread ID's.

An informal benchmark showed a ~20% speed increase when running RealmResultsTest.

@emanuelez @kneth @nhachicha @beeender


beeender:
:+1:

kneth:
:+1:

PR #1228, opened 2015-06-21
Add support for in-memory Realm

### LINKED ISSUE #1007: In-Memory Realms android
Is there any way to save realm database only in memory?

beeender:
Fixes  #1007 
1. Add durability to createNativeWithImplicitTransactions.
2. Add durability to RealmConfiguration.
3. Support passing durability to SharedGroup constructor.
4. Add new static method getInMemoryRealm to Realm class.
5. Add test cases.


realm/src/main/java/io/realm/RealmConfiguration.java
```diff
@@ -326,6 +334,11 @@ public Builder deleteRealmIfMigrationNeeded() {
             return this;
         }
 
+        public Builder durability(SharedGroup.Durability durability) {
```
emanuelez: We cannot expose SharedGroup to the public facing API. Either we move the enum, or we wrap it, or we should use some other mechanism. @cmelchior @nhachicha @kneth what do you guys think?

cmelchior: I agree. If we need durability as a parameter it should be moved to a public enum. I am not a big fan of calling it ```durability``` though. The only reason we would need that if we at some point are going to support the ```Durability.ASYNC``` which I am not sure we are going to?

If not I would much prefer this to be called ``inMemory()```

beeender: Changed the public API to inMemory.

What would we do for the future ASYNC support? Add another method `async()`? And throw an exception if user call `async()` after calling `inMemory()`?

realm-jni/src/io_realm_internal_SharedGroup.cpp
```diff
@@ -83,15 +83,32 @@ JNIEXPORT jlong JNICALL Java_io_realm_internal_SharedGroup_nativeCreate(
 }
 
 JNIEXPORT jlong JNICALL Java_io_realm_internal_SharedGroup_createNativeWithImplicitTransactions
-  (JNIEnv* env, jobject, jlong native_replication_ptr, jbyteArray keyArray)
+  (JNIEnv* env, jobject, jlong native_replication_ptr, jint durability, jbyteArray keyArray)
```
cmelchior: createNativeWithImplicitTransactions and createNative shares a lot of code. They should be refactored so we don't duplicate that code.

realm/src/androidTest/java/io/realm/RealmConfigurationTest.java
```diff
@@ -343,4 +342,18 @@ public void testDifferentSchemasThrows() {
             realm1.close();
         }
     }
+
+    public void testDifferentDurabilityThrows() {
+        RealmConfiguration config1 = new RealmConfiguration.Builder(getContext()).inMemory().build();
+        RealmConfiguration config2 = new RealmConfiguration.Builder(getContext()).inMemory().build();
```
cmelchior: This is the same configuration :). One of them should leave out the `inMemory()` setting.

realm/src/androidTest/java/io/realm/RealmInMemoryTest.java
```diff
@@ -0,0 +1,215 @@
...
+    private final static String IDENTIFIER = "InMemRealTest";
+
+    protected Realm testRealm;
+
+    @Override
+    protected void setUp() throws Exception {
+
+        // Delete the same name realm file just in case
+        RealmConfiguration configuration = new RealmConfiguration.Builder(getContext())
+                .name(IDENTIFIER).build();
```
cmelchior: `.build()` should be on a separate line to increase readability.

realm/src/androidTest/java/io/realm/RealmInMemoryTest.java
```diff
@@ -0,0 +1,215 @@
...
+        Realm.deleteRealm(configuration);
+
+        testRealm = Realm.getInMemoryInstance(getContext(), IDENTIFIER);
```
cmelchior: We should not add new constructors. The point of RealmConfiguration was to avoid that. All instances must be gotten by doing `Realm.getInstance(configuration)` or `Realm.getDefaultInstance()`.

realm/src/androidTest/java/io/realm/RealmInMemoryTest.java
```diff
@@ -0,0 +1,215 @@
...
+        assertEquals(testRealm.allObjects(Dog.class).first().getName(), "DinoDog");
+        testRealm.close();
+
+        // After all references to the in-mem-realm destroyed,
+        // in-mem-realm with same identifier should create a fresh new instance.
+        testRealm = Realm.getInMemoryInstance(getContext(), IDENTIFIER);
+        assertEquals(testRealm.allObjects(Dog.class).size(), 0);
+    }
+
+    public void testSameIdentifierRealmOnDisk() {
```
cmelchior: This is the same test as testDifferentDurabilityThrows, except that here you set a name. For the first test "default.realm" is used as the default name, but thats the only difference as far as I can see. I would move this test to RealmConfigurationTests.

realm/src/androidTest/java/io/realm/RealmInMemoryTest.java
```diff
@@ -0,0 +1,215 @@
...
+        try {
+            Realm.deleteRealm(configuration);
+            fail("Realm.deleteRealm should fail with illegal state");
+        } catch (IllegalStateException iae) { //NOPMD
+        }
+
+        // Nothing should happen when delete a closed in-mem-realm.
+        testRealm.close();
+        testRealm = null;
+        Realm.deleteRealm(configuration);
```
cmelchior: assertTrue(Realm.deleteRealm(configuration))

realm/src/androidTest/java/io/realm/RealmInMemoryTest.java
```diff
@@ -0,0 +1,215 @@
...
+    }
+
+    public void testWriteCopyTo() throws IOException {
+        String fileName = IDENTIFIER + ".realm";
+        String encFileName = IDENTIFIER + ".enc.realm";
+        byte[] key = TestHelper.getRandomKey();
+
+        Realm.deleteRealm(new RealmConfiguration.Builder(getContext()).name(fileName).build());
```
cmelchior: Refactor this into a variable to avoid having to create it multiple times.

realm/src/androidTest/java/io/realm/RealmInMemoryTest.java
```diff
@@ -0,0 +1,215 @@
...
+        Realm.deleteRealm(new RealmConfiguration.Builder(getContext()).name(encFileName).build());
```
cmelchior: same

realm/src/androidTest/java/io/realm/RealmInMemoryTest.java
```diff
@@ -0,0 +1,215 @@
...
+
+        testRealm.beginTransaction();
+        Dog dog = testRealm.createObject(Dog.class);
+        dog.setName("DinoDog");
+        testRealm.commitTransaction();
+
+        // Test a normal realm file
+        testRealm.writeCopyTo(new File(getContext().getFilesDir(), fileName));
+        Realm onDiskRealm = Realm.getInstance(new RealmConfiguration.Builder(getContext()).name(fileName).build());
```
cmelchior: same

realm/src/androidTest/java/io/realm/RealmInMemoryTest.java
```diff
@@ -0,0 +1,215 @@
...
+        assertEquals(onDiskRealm.allObjects(Dog.class).size(), 1);
+        onDiskRealm.close();
+
+        // Test a encrypted realm file
+        testRealm.writeEncryptedCopyTo(new File(getContext().getFilesDir(), encFileName), key);
+        onDiskRealm = Realm.getInstance(new RealmConfiguration.Builder(getContext()).name(encFileName)
```
cmelchior: same

realm/src/androidTest/java/io/realm/RealmInMemoryTest.java
```diff
@@ -0,0 +1,215 @@
...
+        // Test with a wrong key
+        try {
+            Realm.getInstance(new RealmConfiguration.Builder(getContext()).name(encFileName)
+                    .encryptionKey(TestHelper.getRandomKey(42)).build());
+            fail("Realm.getInstance should fail with illegal argument");
+        } catch (IllegalArgumentException iae) { //NOPMD
+        }
+    }
+
+    public void testCreateWithConfiguration() {
```
cmelchior: What does this test?

realm-jni/src/io_realm_internal_SharedGroup.cpp
```diff
@@ -83,15 +83,32 @@ JNIEXPORT jlong JNICALL Java_io_realm_internal_SharedGroup_nativeCreate(
...
+
     try {
         KeyBuffer key(env, keyArray);
-#ifdef REALM_ENABLE_ENCRYPTION
-        SharedGroup* db = new SharedGroup(*reinterpret_cast<realm::Replication*>(native_replication_ptr), SharedGroup::durability_Full, key.data());
+#ifdef REALM_ENABLE_REPLICATION
+        SharedGroup* db = new SharedGroup(*reinterpret_cast<realm::Replication*>(native_replication_ptr), level, key.data());
 #else
-        SharedGroup* db = new SharedGroup(*reinterpret_cast<realm::Replication*>(native_replication_ptr));
+    #error "REALM_ENABLE_REPLICATION must be defined for compiling realm-java!"
```
bmunkholm: We should just have this check once in the top of the file.

realm/src/androidTest/java/io/realm/RealmInMemoryTest.java
```diff
@@ -0,0 +1,215 @@
...
+
+        RealmConfiguration configuration = new RealmConfiguration.Builder(getContext())
+                .name(IDENTIFIER).inMemory().build();
+        Realm realm = Realm.getInstance(configuration);
+
+        assertEquals(realm.allObjects(Dog.class).first().getName(), "DinoDog");
+        realm.close();
+    }
+
+    public void testMultiThread() throws InterruptedException, ExecutionException {
```
cmelchior: Add a comment about what this is testing. From what I can tell it is that a in-memory realm is not fully deleted until all threads has released their instance.

realm/src/androidTest/java/io/realm/RealmInMemoryTest.java
```diff
@@ -0,0 +1,215 @@
...
+import java.io.File;
+import java.io.IOException;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+import io.realm.entities.Dog;
+
+public class RealmInMemoryTest extends AndroidTestCase {
+
+    private final static String IDENTIFIER = "InMemRealTest";
```
bmunkholm: RealTest -> RealmTest

realm/src/main/java/io/realm/Realm.java
```diff
@@ -396,6 +397,36 @@ public static Realm getInstance(Context context, String fileName, byte[] key) {
...
+     * @return an instance of the Realm class
+     *
+     * @throws IllegalArgumentException      An realm instance with same identifier/filename and different
+     *                                       {@link io.realm.internal.SharedGroup.Durability} has been created already.
+     * @throws RealmMigrationNeededException The model classes have been changed and the Realm
+     *                                       must be migrated
+     * @throws RealmIOException              Error when accessing underlying file
+     * @throws RealmException                Other errors
+     */
+    public static Realm getInMemoryInstance(Context context, String identifier) {
```
cmelchior: Delete this constructor. People should use `Realm.getInstance(RealmConfiguration)` instead.

realm/src/androidTest/java/io/realm/RealmInMemoryTest.java
```diff
@@ -0,0 +1,215 @@
+/*
```
bmunkholm: I havn't checked, but how much copy/paste is there?
I would assume that testing the inMemory Realm or the file based can be done generically using just a parameter more for relevant tests?

bmunkholm: This will be even more important for other new durability tests as well.

realm/src/main/java/io/realm/Realm.java
```diff
@@ -656,6 +687,15 @@ private static void validateAgainstExistingConfigurations(RealmConfiguration new
                 throw new IllegalArgumentException("Two configurations with different schemas are trying to open " +
                         "the same Realm file. Their schema must be the same: " + newConfiguration.getPath());
             }
+
+            // Check if the durability is the same
+            SharedGroup.Durability cachedDurability = cachedConfiguration.getDurability();
+            SharedGroup.Durability newDurability = newConfiguration.getDurability();
+            if (cachedDurability.value != newDurability.value) {
+                throw new IllegalArgumentException("Configurations cannot have different durabilities for the " +
+                        "same realm filename/identifier: " + newConfiguration.getRealmFileName() + " " +
```
cmelchior: identifier is not a word we have used before. Until we add more durability settings I would rephrase this to something like "A Realm cannot be both in-memory and persisted. Two conflicting configurations pointing to <filename> are being used."

realm/src/androidTest/java/io/realm/RealmInMemoryTest.java
beeender: Not too much copy/paste here. It's only in-mem related cases here since the diff here is writing to the disk or not. But still case like write-copy-to is pretty the same with the on-disk-realm. Maybe it is still good idea to generalize all the cases we have in RealmTest to apply to different kinds of durability?

realm/src/main/java/io/realm/Realm.java
beeender: Shall we allign with ios-bindings? I got the "identifier" word from IOS binding...

realm/src/main/java/io/realm/RealmConfiguration.java
```diff
@@ -327,6 +335,15 @@ public Builder deleteRealmIfMigrationNeeded() {
         }
 
         /**
+         * Settings this will create a in-memory Realm instance.
```
cmelchior: Change this something like: 
"Setting this will create an in-memory Realm instead of of saving it to disk. In-memory Realms might still use disk space if memory is running low, but all files created by a in-memory Realm will be deleted when the Realm is closed."

realm/src/main/java/io/realm/internal/SharedGroup.java
```diff
@@ -40,9 +40,9 @@
         MEM_ONLY(1);
         //ASYNC(2); // TODO: re-enable when possible
 
-        final int value;
+        public final int value;
```
cmelchior: Does this really need to be public?

realm/src/main/java/io/realm/RealmConfiguration.java
bmunkholm: For documentation please always review docs from ObjC or Swift.

realm/src/main/java/io/realm/Realm.java
cmelchior: We should probably at least check why they decided to go with identifier instead of name. Maybe it is to avoid confusion with the path, but as an in-memory Realm still might write data to disc we cannot really separate the two IMO.

realm/src/main/java/io/realm/internal/SharedGroup.java
beeender: No. My fault. I saw Enum's equals method is declared as final but didn't realize for the Enum comparision `==` is actually good enough. Will change it.

bmunkholm: We should actually just get rid of the enum in favor of int constants. (a general Android guideline)

beeender: The enum part seems to be removed from android performance tips http://stackoverflow.com/questions/5143256/why-was-avoid-enums-where-you-only-need-ints-removed-from-androids-performanc/5264861#5264861 .

But i agree that it makes the code actually simpler and we do have value checking in jni. And more performance could be gained when calculate the hashCode.

I will modify it to use int constants.

cmelchior: There is a lot of arguments about that particular guideline, and while it might make sense for the Android framework team, in 99.9% of other cases it wont. What that guideline is for is warning that enums have a larger memory footprint than ints, which is important if you start start creating new ones all over the place. We are nowhere near that, and enums make the code much more readable and maintainable imo.

bmunkholm: It should be just as readable with proper constants. But no autocompletion help. In this case however, it's internal and has to be converted to a number anyways before passing it to JNI. For external API's I would tend to agree to keep it for anything that's not performance critical.

realm-jni/src/io_realm_internal_SharedGroup.cpp
beeender: To avoid the long list of params, the both functions are kept for now. Add one more static function to share common code.

realm/src/androidTest/java/io/realm/RealmInMemoryTest.java
beeender: Done

realm-jni/src/io_realm_internal_SharedGroup.cpp
beeender: Done.

realm/src/androidTest/java/io/realm/RealmConfigurationTest.java
beeender: Oooops. Fixed, thanks!

realm/src/androidTest/java/io/realm/RealmInMemoryTest.java
beeender: Done

realm/src/androidTest/java/io/realm/RealmInMemoryTest.java
beeender: Done

realm/src/androidTest/java/io/realm/RealmInMemoryTest.java
beeender: Done

realm/src/androidTest/java/io/realm/RealmInMemoryTest.java
beeender: Done

realm/src/androidTest/java/io/realm/RealmInMemoryTest.java
beeender: Done

realm/src/androidTest/java/io/realm/RealmInMemoryTest.java
bmunkholm: We should never copy code. (Almost) always reuse. Only make those tests that really are different. Otherwise it makes sense to reuse.

realm/src/androidTest/java/io/realm/RealmInMemoryTest.java
beeender: Done

realm/src/androidTest/java/io/realm/RealmInMemoryTest.java
beeender: Done

realm/src/androidTest/java/io/realm/RealmInMemoryTest.java
beeender: It is trying to test if a in-memory realm can be written to the disk with encryption and the encrytion does work by using a wrong key to open it.

realm/src/androidTest/java/io/realm/RealmInMemoryTest.java
beeender: Done

realm/src/androidTest/java/io/realm/RealmInMemoryTest.java
beeender: Done! sharp eyes!

realm/src/main/java/io/realm/Realm.java
beeender: Done

realm/src/main/java/io/realm/Realm.java
beeender: I will check with IOS guy. And the "identifier" has been removed from PR. Exception msg changed as well.

realm/src/main/java/io/realm/RealmConfiguration.java
beeender: Yes, checked relevant doc from IOS, and added some notes to `inMemory()` in the 2nd commit. Please help to check if it is clear enough.

realm/src/main/java/io/realm/internal/SharedGroup.java
beeender: I leave the enum for this PR. And If needed, I would like to create another PR to change all the internal enums to constants.

realm/src/androidTest/java/io/realm/RealmInMemoryTest.java
beeender: I just went through the RealmTest, most of those cases should not be affected by the durability at all. So I choose to Have those a few special cases here for now. And also did checking with the relevent IOS cases, in addition to those basic cases we have here, some other performance cases are also created. Maybe we should have same performance cases as well to avoid performance regression?

realm/src/main/java/io/realm/Realm.java
tgoyne: We were worried that talking about the path of an in-memory realm would confuse users, but we've ended up getting all of that confusion anyway. We've talked about switching from "in memory" to "transient" or something along those lines.

astigsen: I am not too happy about "transient" as that only highlights the part about the file being removed after use. In my opinion the main characteristic is that it will never hit disk during normal operations (i.e. a performance quality).

But of course that still demands that we specify when it _does_ hit disk (possibly when the os thinks it has some spare cycles and decides to sync, and on memory pressure that forces swap).

realm/src/androidTest/java/io/realm/RealmInMemoryTest.java
```diff
@@ -0,0 +1,225 @@
...
+
+        Realm.deleteRealm(conf);
+        Realm.deleteRealm(encConf);
+
+        testRealm.beginTransaction();
+        Dog dog = testRealm.createObject(Dog.class);
+        dog.setName("DinoDog");
+        testRealm.commitTransaction();
+
+        // Test a normal realm file
```
kneth: We normally write Realm with capital R.

realm/src/androidTest/java/io/realm/RealmInMemoryTest.java
```diff
@@ -0,0 +1,225 @@
...
+        Realm onDiskRealm = Realm.getInstance(conf);
+        assertEquals(onDiskRealm.allObjects(Dog.class).size(), 1);
+        onDiskRealm.close();
+
+        // Test a encrypted realm file
+        testRealm.writeEncryptedCopyTo(new File(getContext().getFilesDir(), encFileName), key);
+        onDiskRealm = Realm.getInstance(encConf);
+        assertEquals(onDiskRealm.allObjects(Dog.class).size(), 1);
+        onDiskRealm.close();
+        // Test with a wrong key to see if it is encrypted successfully
```
kneth: To be pedantic: you are testing a wrong key to see if it fails as expected.

kneth:
Looks good :+1:

realm/src/main/java/io/realm/Realm.java
cmelchior: "in-memory" is already a concept from the SQLite API's, so people are familiar with it. It is of course a bit problematic that we actually write to disk, but at least "in-memory" is a term people would be familiar with on Android.

realm/src/androidTest/java/io/realm/RealmInMemoryTest.java
beeender: Done

realm/src/androidTest/java/io/realm/RealmInMemoryTest.java
beeender: Done

realm/src/main/java/io/realm/Realm.java
beeender: To me, as non-native english speaker, it is much easier to understand "in-memory" than "transient" directly.

realm-jni/src/io_realm_internal_SharedGroup.cpp
```diff
@@ -83,16 +97,19 @@ JNIEXPORT jlong JNICALL Java_io_realm_internal_SharedGroup_nativeCreate(
...
+
+    SharedGroup::DurabilityLevel level;
+    if (!jintToDurabilityLevel(durability, level)) {
+        ThrowException(env, UnsupportedOperation, "Unsupported durability.");
+        return 0;
+    }
+
     try {
         KeyBuffer key(env, keyArray);
-#ifdef REALM_ENABLE_ENCRYPTION
```
cmelchior: Why was this check disabled? It is still in native_create, and I assume we want to be able to disable encryption when building Realm Core or has that changed?

realm/src/androidTest/java/io/realm/RealmInMemoryTest.java
```diff
@@ -0,0 +1,225 @@
...
+                .name(IDENTIFIER)
+                .build();
+        inMemConf = new RealmConfiguration.Builder(getContext())
+                .name(IDENTIFIER)
+                .inMemory()
+                .build();
+
+        // Delete the same name Realm file just in case
+        Realm.deleteRealm(onDiskConf);
+
```
cmelchior: remove whitespace

realm/src/androidTest/java/io/realm/RealmInMemoryTest.java
```diff
@@ -0,0 +1,225 @@
...
+    @Override
+    protected void tearDown() throws Exception {
+        if (testRealm != null) {
+            testRealm.close();
+        }
+    }
+
+    // Testing the in-memory Realm by Creating one instance, adding a record, then close the instance.
+    // By the next time in-memory Realm instance with the same name created, it should be empty.
+    // Two in-memory Realms with different names should not affect each other.
```
cmelchior: This should be a separate test.

realm-jni/src/io_realm_internal_SharedGroup.cpp
beeender: REALM_ENABLE_ENCRYPTION doesn't exsit in group_shared.hpp anymore. And that constructor has been removed from native header.

realm/src/androidTest/java/io/realm/RealmInMemoryTest.java
```diff
@@ -0,0 +1,225 @@
...
+        onDiskRealm.close();
+        // Test with a wrong key to see if it fails as expected.
+        try {
+            RealmConfiguration wrongKeyConf = new RealmConfiguration.Builder(getContext())
+                    .name(encFileName)
+                    .encryptionKey(TestHelper.getRandomKey(42))
+                    .build();
+            Realm.getInstance(wrongKeyConf);
+            fail("Realm.getInstance should fail with illegal argument");
+        } catch (IllegalArgumentException iae) { //NOPMD
```
cmelchior: We usually call this exception "expected" and remove the comment.

realm/src/androidTest/java/io/realm/RealmInMemoryTest.java
```diff
@@ -0,0 +1,225 @@
...
+        assertEquals(testRealm.allObjects(Dog.class).size(), 0);
+    }
+
+    // Test deleteRealm called on a in-memory Realm instance
+    public void testDelete() {
+        RealmConfiguration configuration = testRealm.getConfiguration();
+        try {
+            Realm.deleteRealm(configuration);
+            fail("Realm.deleteRealm should fail with illegal state");
+        } catch (IllegalStateException iae) { //NOPMD
```
cmelchior: "expected" and remove comment

realm-jni/src/io_realm_internal_SharedGroup.cpp
cmelchior: Ok, then it should probably be removed in nativeCreate as well

realm/src/androidTest/java/io/realm/RealmInMemoryTest.java
```diff
@@ -0,0 +1,225 @@
...
+                    .name(encFileName)
+                    .encryptionKey(TestHelper.getRandomKey(42))
+                    .build();
+            Realm.getInstance(wrongKeyConf);
+            fail("Realm.getInstance should fail with illegal argument");
+        } catch (IllegalArgumentException iae) { //NOPMD
+        }
+    }
+
+    // Test below scenario:
```
cmelchior: Great with comments like this

realm/src/main/java/io/realm/RealmConfiguration.java
```diff
@@ -327,6 +336,19 @@ public Builder deleteRealmIfMigrationNeeded() {
         }
 
         /**
+         * Setting this will create an in-memory Realm instead of of saving it to disk. In-memory Realms might still use
+         * disk space if memory is running low, but all files created by a in-memory Realm will be deleted when the
+         * Realm is closed.
+         * <p></p>
+         * Note that Because in-memory Realms are not persisted, you must be sure to hold on to at least one non-closed
```
cmelchior: Because -> because

realm/src/main/java/io/realm/RealmConfiguration.java
```diff
@@ -327,6 +336,19 @@ public Builder deleteRealmIfMigrationNeeded() {
         }
 
         /**
+         * Setting this will create an in-memory Realm instead of of saving it to disk. In-memory Realms might still use
+         * disk space if memory is running low, but all files created by a in-memory Realm will be deleted when the
+         * Realm is closed.
+         * <p></p>
```
cmelchior: <p />

cmelchior:
This is starting to look good. Only things missing seems be some smaller edits and we need to settle on the exact name for the public configuration. I still think ```inMemory()``` would make sense to most people given how you normally work with SQLite :)

Also you want to bump the version.txt to "0.82.0-SNAPSHOT" as merging this is a new feature.

changelog.txt
```diff
@@ -1,3 +1,6 @@
+0.82.0-SNAPSHOT
```
cmelchior: Remove SNAPSHOT. Changelog is only for "official releases". I meant that the file `version.txt` should be updated to `0.82.0-SNAPSHOT` as well as `README.md`. This is because SNAPSHOT releases are deployed automatically when committed to master and since this a new feature it require a minor bump from 0.81.1 to 0.82.0 :)

cmelchior:
If the changelog/version.txt is updated: :+1:

changelog.txt
beeender: Ha, OK. I was confused why did you want  to put SNAPSHOT in...

beeender:
@kneth @cmelchior @emanuelez @nhachicha Please help to review :mahjong:

realm/src/main/java/io/realm/internal/SharedGroup.java
```diff
@@ -40,9 +40,9 @@
         MEM_ONLY(1);
         //ASYNC(2); // TODO: re-enable when possible
 
-        final int value;
+        int value;
```
emanuelez: This can still be final right?

beeender: Yep, my fault...

realm/src/main/java/io/realm/RealmConfiguration.java
```diff
@@ -327,6 +336,19 @@ public Builder deleteRealmIfMigrationNeeded() {
         }
 
         /**
+         * Setting this will create an in-memory Realm instead of of saving it to disk. In-memory Realms might still use
```
nhachicha: typo _of of_

realm/src/main/java/io/realm/RealmConfiguration.java
```diff
@@ -327,6 +336,19 @@ public Builder deleteRealmIfMigrationNeeded() {
...
+         * disk space if memory is running low, but all files created by a in-memory Realm will be deleted when the
+         * Realm is closed.
+         * <p></p>
+         * Note that because in-memory Realms are not persisted, you must be sure to hold on to at least one non-closed
+         * reference to the in-memory Realm object with the specific name as long as you want the data to last.
```
nhachicha: i.e as long as we keep one instance open, your data is kept

nhachicha:
looking good so far :+1: 
I was wondering if it make sense to add a test using [StrictMode](http://developer.android.com/reference/android/os/StrictMode.html) with `.detectDiskReads()` and
`.detectDiskWrites()` after creating the DB using inMemory? although the _StrictMode_ apply only to the VM threads & not native code

realm/src/main/java/io/realm/RealmConfiguration.java
beeender: Note that because in-memory Realms are not persisted, you must be sure to hold on to at least one non-closed reference to the in-memory Realm object with the specific name. As long as we keep one instance open,  your data is kept.
@nhachicha I feel there are too many "one instance" . What do you think?

beeender:
@nhachicha Using StrictMode in the tests is quite interesting!
I did some tests with using `detectDiskReads()` and `.detectDiskWrites()` in the test case, at least it can help us to make sure the disk IO won't happen. I will add this to the case.

But there is one thing needs to be noticed, setting the StrictMode(For thread, didn't try with vm) without setting it back to default before exiting the case might cause some strange problems. I have seen only 2 of 6 cases in my TestClass has been ran without any error reported (And sometime it would randomly report only 2 of 6 cases has been found). Are you familiar with this? I am thinking would it be safe to add it, maybe not this time.

realm-jni/src/io_realm_internal_SharedGroup.cpp
```diff
@@ -53,27 +77,13 @@ JNIEXPORT jlong JNICALL Java_io_realm_internal_SharedGroup_nativeCreate(
...
-                level = SharedGroup::durability_MemOnly;
-            else if (durability == 2)
-#ifdef _WIN32
-                level = SharedGroup::durability_Full;   // For Windows, use Full instead of Async
-#else
-                level = SharedGroup::durability_Async;
-#endif
-            else {
+            if (!jintToDurabilityLevel(durability, level)) {
                 ThrowException(env, UnsupportedOperation, "Unsupported durability.");
```
bmunkholm: Suggest to move the ThrowException() to the jintToDurabilityLevel() function.

realm/src/main/java/io/realm/RealmConfiguration.java
```diff
@@ -327,6 +336,19 @@ public Builder deleteRealmIfMigrationNeeded() {
         }
 
         /**
+         * Setting this will create an in-memory Realm instead of of saving it to disk. In-memory Realms might still use
+         * disk space if memory is running low, but all files created by a in-memory Realm will be deleted when the
```
bmunkholm: a -> an

realm-jni/src/io_realm_internal_SharedGroup.cpp
beeender: That requires adding an env param to jintToDurabilityLevel which makes me feel ugly...

bmunkholm: never mind. It's unlikely to be called too more than twice.

realm/src/main/java/io/realm/RealmConfiguration.java
emanuelez: No need to close the paragraph

realm/src/main/java/io/realm/RealmConfiguration.java
nhachicha: was just confirming my understanding, the original rephrasing should be good

nhachicha:
@beeender each Testcase run in it's own thread, so the policy you apply should not affect other test cases. You can also disable the policy on `tearDown`. for the flaky behaviour I suspect some operations like _close_ or _commit_ (the current protocol flip a bit in a file to indicate a successful op) that could cause an IO?

beeender:
@nhachicha 
For the flaky behaviour, even if there is IO happening, the exception should be be thrown since i have set the `penaltyToDeath`. But no, most of time it just stopped with a successful message.

About the thread, what I observered is different cases in the same test class run in the same thread. Or maybe i missed something?

cmelchior:
@nhachicha @beeender It should be easy to test, but I am also pretty sure the AndroidTests run on the same thread.

beeender:
@nhachicha I will add the StrictMode to one of the case. Let's see if there is any random thing happen in the long run. After set the policy back to default, it should be fine. Otherwise google really screwed up something.

nhachicha:
@cmelchior Actually the `InstrumentationTestRunner` spawn a new `InstrumentationThread` for each test. you can check on AOSP

cmelchior:
@nhachicha Ok, cool. The source is always right :)

beeender:
@nhachicha Added a StrictMode check :laughing:

beeender:
@realm/java Please help to review this again. Some small fixes in the last commit.

changelog.txt
```diff
@@ -1,3 +1,6 @@
+0.82
+ * Added in-memory Realm suport.
```
kneth: "suport" -> "support"

realm/src/androidTest/java/io/realm/RealmInMemoryTest.java
```diff
@@ -0,0 +1,242 @@
...
+        if (testRealm != null) {
+            testRealm.close();
+        }
+    }
+
+    // Testing the in-memory Realm by Creating one instance, adding a record, then close the instance.
+    // By the next time in-memory Realm instance with the same name created, it should be empty.
+    // Use StrictMode to check no disk IO would happen in VM to this thread.
+    public void testInMemoryRealm() {
+        StrictMode.setThreadPolicy(new StrictMode.ThreadPolicy.Builder()
```
cmelchior: Does this actually test anything meaningful? As AFAIK this doesn't check native code, so all this tests is IO from Java (of which we have none).

realm/src/main/java/io/realm/Realm.java
```diff
@@ -657,6 +658,14 @@ private static void validateAgainstExistingConfigurations(RealmConfiguration new
                 throw new IllegalArgumentException("Two configurations with different schemas are trying to open " +
                         "the same Realm file. Their schema must be the same: " + newConfiguration.getPath());
             }
+
+            // Check if the durability is the same
+            SharedGroup.Durability cachedDurability = cachedConfiguration.getDurability();
+            SharedGroup.Durability newDurability = newConfiguration.getDurability();
+            if (!cachedDurability.equals(newDurability)) {
+                throw new IllegalArgumentException("A Realm cannot be both in-memory and persisted. Two conflicting " +
+                        "configurations pointing to " + newConfiguration.getRealmFileName() + " are being used.");
```
cmelchior: Use newConfiguration.getPath() instead. You might risk that two Realms are using the same filename in different directories.

cmelchior:
https://github.com/realm/realm-java/blob/master/version.txt should also be bumped to "0.82.0-SNAPSHOT" and https://github.com/realm/realm-java/blob/master/README.md section on SNAPSHOT releases should be updated to reflect this.  

This is because this actually adds new functionality which would require a minor bump i version numbering also for our SNAPSHOT releases.

realm/src/androidTest/java/io/realm/RealmInMemoryTest.java
beeender: On IO here:
```
        public Builder(Context context) {
            if (context == null) {
                throw new IllegalArgumentException("A non-null Context must be provided");
            }
        }
```
I am thinking to keep it to prevent from doing something wrong in the future?

cmelchior: Ah, cool. You are right. Keep the check

realm/src/main/java/io/realm/Realm.java
beeender: That remind me that do we need to restrict use from using DEFAULT_REALM_NAME to create in-memory Realm? It won't be harmful at all, but it has slightly differnt API with IOS which always ask user to give the in-memory Realm a name (Although it can be default if user wants.)

cmelchior: Hmm. My initial reaction would probably be that you would rarely mix inMemory and persisted Ream anyway, and even if you did it accidentally we have the configuration check in Realm.java that will throw an error in that case. So no, I don't think we need to explicitly require a name for in-memory Realms.

realm-jni/src/io_realm_internal_SharedGroup.cpp
beeender: I convinced myself to throw the exception in the local function since i found in the other PR i have to send a long description exception two times...
And also, the name changed to jint_to_durability_level since it seems to be the right naming style.

realm-jni/src/io_realm_internal_SharedGroup.cpp
```diff
@@ -52,28 +77,11 @@ JNIEXPORT jlong JNICALL Java_io_realm_internal_SharedGroup_nativeCreate(
...
-                level = SharedGroup::durability_Full;   // For Windows, use Full instead of Async
-#else
-                level = SharedGroup::durability_Async;
-#endif
-            else {
-                ThrowException(env, UnsupportedOperation, "Unsupported durability.");
-                return 0;
-            }
+            // Exception thrown for wrong durability value
+            SharedGroup::DurabilityLevel level = jint_to_durability_level(env, durability);
```
cmelchior: you should check for != null here and return, throwing a Java exception will not abort C++ code execution, so if this fails you will send an illegal durability to SharedGroup most likely crashing it hard.

realm-jni/src/io_realm_internal_SharedGroup.cpp
```diff
@@ -30,6 +30,31 @@ using namespace realm;
...
+        level = SharedGroup::durability_MemOnly;
+    else if (durability == 2)
+#ifdef _WIN32
+        level = SharedGroup::durability_Full;   // For Windows, use Full instead of Async
+#else
+        level = SharedGroup::durability_Async;
+#endif
+    else {
+        ThrowException(env, UnsupportedOperation, "Unsupported durability.");
+    }
```
cmelchior: return null to make return error value explicit.

realm-jni/src/io_realm_internal_SharedGroup.cpp
```diff
@@ -83,16 +91,16 @@ JNIEXPORT jlong JNICALL Java_io_realm_internal_SharedGroup_nativeCreate(
...
 
 JNIEXPORT jlong JNICALL Java_io_realm_internal_SharedGroup_createNativeWithImplicitTransactions
-  (JNIEnv* env, jobject, jlong native_replication_ptr, jbyteArray keyArray)
+  (JNIEnv* env, jobject, jlong native_replication_ptr, jint durability, jbyteArray keyArray)
 {
     TR_ENTER()
+
+    // Exception thrown for wrong durability value
+    SharedGroup::DurabilityLevel level = jint_to_durability_level(env, durability);
+
```
cmelchior: Check for valid return value before continuing

realm/src/androidTest/java/io/realm/RealmInMemoryTest.java
```diff
@@ -0,0 +1,242 @@
...
+
+                assertEquals(realm.allObjects(Dog.class).size(), 1);
+
+                // Wait until Realm instance closed in main thread
+                try {
+                    while (!isRealmInMainThreadClosed.get()) {
+                        Thread.sleep(5);
+                    }
+                } catch (InterruptedException e) {
+                    fail("Worker thread was interrupted.");
```
nhachicha: This fail will not report in JUnit, you need to capture it & throw it within the testcase's Thread 
ex: https://github.com/realm/realm-java/blob/nh-async-query/realm/src/androidTest/java/io/realm/RealmAsyncQueryTests.java#L241 You can also use `CountDownLatch` to avoid _"busy waiting"_  `while (condition) Thread.sleep();`

realm/src/androidTest/java/io/realm/RealmInMemoryTest.java
```diff
@@ -0,0 +1,242 @@
...
+
+        // refresh will be ran in the next loop, manually refresh it here.
+        testRealm.refresh();
+        assertEquals(testRealm.allObjects(Dog.class).size(), 1);
+
+        // Step 3.
+        // Release the main thread Realm reference, and the worker thread hold the reference still
+        testRealm.close();
+        // Step 4.
+        isRealmInMainThreadClosed.set(true);
```
nhachicha: Wouldn't this cause the `workerThread` to close its Realm? hence at this point all Realms are closed, why we are asserting this below for a fresh Realm?
`assertEquals(testRealm.allObjects(Dog.class).size(), 1);` ?
it should be 0 IMO

beeender: Hey, you are right! It could be 0 or 1 since no waiting lock for the `realm.close()`. I will move it before the `isRealmInMainThreadClosed.set(true);`.

beeender:
@realm/java Help to review. All commented issues fixed.

realm-jni/src/io_realm_internal_SharedGroup.cpp
```diff
@@ -83,16 +94,19 @@ JNIEXPORT jlong JNICALL Java_io_realm_internal_SharedGroup_nativeCreate(
...
 
 JNIEXPORT jlong JNICALL Java_io_realm_internal_SharedGroup_createNativeWithImplicitTransactions
-  (JNIEnv* env, jobject, jlong native_replication_ptr, jbyteArray keyArray)
+  (JNIEnv* env, jobject, jlong native_replication_ptr, jint durability, jbyteArray keyArray)
 {
     TR_ENTER()
+
+    SharedGroup::DurabilityLevel level;
+    // Exception thrown for wrong durability value
+    if(!jint_to_durability_level(env, durability, level)) {
```
kneth: `if(` -> `if (`

realm-jni/src/io_realm_internal_SharedGroup.cpp
```diff
@@ -53,27 +78,13 @@ JNIEXPORT jlong JNICALL Java_io_realm_internal_SharedGroup_nativeCreate(
...
-            else {
-                ThrowException(env, UnsupportedOperation, "Unsupported durability.");
+            // Exception thrown for wrong durability value
+            if (!jint_to_durability_level(env, durability, level)) {
                 return 0;
             }
 
             KeyBuffer key(env, keyArray);
-#ifdef REALM_ENABLE_ENCRYPTION
             db = new SharedGroup(file_name, no_create!=0, level, key.data());
```
cmelchior: ``` no_create != 0```

realm/src/androidTest/java/io/realm/RealmConfigurationTest.java
```diff
@@ -343,4 +342,46 @@ public void testDifferentSchemasThrows() {
...
+        RealmConfiguration config1 = new RealmConfiguration.Builder(getContext()).inMemory().build();
+        RealmConfiguration config2 = new RealmConfiguration.Builder(getContext()).build();
+
+        // Create In-memory Realm first.
+        Realm realm1 = Realm.getInstance(config1);
+        try {
+            // On-disk Realm then. Not allowed!
+            Realm.getInstance(config2);
+            fail();
+        } catch (IllegalArgumentException expected) {
```
cmelchior: ignored to avoid warning

realm/src/androidTest/java/io/realm/RealmConfigurationTest.java
```diff
@@ -343,4 +342,46 @@ public void testDifferentSchemasThrows() {
...
+            realm1.close();
+        }
+
+        // Create on-disk Realm first.
+        realm1 = Realm.getInstance(config2);
+        try {
+            // In-memory Realm then. Not allowed!
+            Realm.getInstance(config1);
+            fail();
+        } catch (IllegalArgumentException expected) {
```
cmelchior: same

realm-jni/src/io_realm_internal_SharedGroup.cpp
kneth: `REALM_ENABLE_ENCRYPTION` is an environment variable set by either the builder or Jenkins. It can to have the same value as when core was built. It is true that we for many releases only have had encrypted enabled but I think that by removing it (and unconditionally enable encryption), it will require an extra effort to disable it.

cmelchior:
Apart from minor style comments :+1:

realm-jni/src/io_realm_internal_SharedGroup.cpp
beeender: I think i missed something before. It is better i don't change this right now. I will revert this part of change. Thanks @kneth

beeender:
@realm/java Sorry to bother you guys again, but, yes, two more commits :sun_with_face:

cmelchior:
This issue wins as most commented on ever I think :)
:+1: :shipit:

emanuelez:
Please rebase or merge master in your branch so you will have checkstyle support, which is now executed by Jenkins :)

beeender:
#1257 has been merged. which the squashed version of this one.

PR #1239, opened 2015-06-24
New migration API

cmelchior:
This depends on #1200 

This PR adds the new Migration API. It introduces a breaking change as the Migration interface has changed. It also introduces 2 new classes: RealmSchema and RealmObjectSchema. The interface for those methods are below and should hopefully be mostly self explanatory (that was the whole point :) )

```
// New interface for writing migration blocks
public interface RealmMigration {
  void migrate(RealmSchema schema, long oldVersion, long newVersion);
}

// Public methods in RealmSchema
public interface RealmSchemaInterface {
    RealmObjectSchema getClass(String className);
    RealmObjectSchema addClass(String className);
    void removeClass(String className);
    RealmObjectSchema renameClass(String oldName, String newName);
}

// Public methods in RealmObjectSchema
public interface RealmObjectSchemaInterface {
    RealmObjectSchema addString(String fieldName);
    RealmObjectSchema addString(String fieldName, Set<RealmModifier> modifiers);
    RealmObjectSchema addShort(String fieldName);
    RealmObjectSchema addShort(String fieldName, Set<RealmModifier> modifiers);
    RealmObjectSchema addInt(String fieldName);
    RealmObjectSchema addInt(String fieldName, Set<RealmModifier> modifiers);
    RealmObjectSchema addLong(String fieldName);
    RealmObjectSchema addLong(String fieldName, Set<RealmModifier> modifiers);
    RealmObjectSchema addBoolean(String fieldName);
    RealmObjectSchema addBoolean(String fieldName, Set<RealmModifier> modifiers);
    RealmObjectSchema addByteArray(String fieldName);
    RealmObjectSchema addByteArray(String fieldName, Set<RealmModifier> modifiers);
    RealmObjectSchema addFloat(String fieldName);
    RealmObjectSchema addFloat(String fieldName, Set<RealmModifier> modifiers);
    RealmObjectSchema addDouble(String fieldName);
    RealmObjectSchema addDouble(String fieldName, Set<RealmModifier> modifiers);
    RealmObjectSchema addDate(String fieldName);
    RealmObjectSchema addDate(String fieldName, Set<RealmModifier> modifiers);
    RealmObjectSchema addObject(String fieldName, RealmObjectSchema objectSchema);
    RealmObjectSchema addList(String fieldName, RealmObjectSchema objectSchema);
    RealmObjectSchema removeField(String fieldName);
    RealmObjectSchema renameField(String currentFieldName, String newFieldName);
    RealmObjectSchema addIndex(String fieldName);
    RealmObjectSchema removeIndex(String fieldName);
    RealmObjectSchema addPrimaryKey(String fieldName);
    RealmObjectSchema setNonNullable(String fieldName); // Not supported yet
    RealmObjectSchema setNullable(String fieldName); // Not supported yet
    boolean hasPrimaryKey();
    boolean hasIndex(String fieldName);
    RealmObjectSchema removePrimaryKey();
    DynamicRealmObject createObject();
    DynamicRealmObject createObject(Object primaryKeyValue);
    DynamicRealmObject forEach(RealmSchemaObject.Iterator iterator);
}

// Example of usage
public class Migration implements RealmMigration {

    @Override
    public void migrate(RealmSchema schema, long oldVersion, long newVersion) {
        /*
            // Version 0
            class Person
                String fullName;
                String lastName;
                int    age;
            // Version 1
            class Person
                String fullName;        // combine firstName and lastName into single field
                int age;
        */
        if (oldVersion == 0) {
            schema.getClass("Person")
                    .addString("fullName")
                    .forEach(new RealmObjectSchema.Iterator() {
                        @Override
                        public void next(DynamicRealmObject obj) {
                            obj.setString("fullName", obj.getString("firstName") + " " + obj.getString("lastName"));
                        }
                    })
                    .removeField("firstName")
                    .removeField("lastName");
            oldVersion++;
        }

        /*
            // Version 2
            class Pet                   // add a new model class
                String name;
                String type;
            class Person
                String fullName;
                int age;
                RealmList<Pet> pets;    // add a RealmList field
        */
        if (oldVersion == 1) {

            final RealmObjectSchema petSchema = schema.addClass("Pet")
                    .addString("name")
                    .addString("type");

            schema.getClass("Person")
                    .addList("pets", petSchema)
                    .forEach(new RealmObjectSchema.Iterator() { // TODO Should we add query support now?
                        @Override
                        public void next(DynamicRealmObject obj) {
                            if (obj.getString("fullName").equals("JP McDonald")) {
                                DynamicRealmObject pet = petSchema.createObject();
                                pet.setString("name", "Jimbo");
                                pet.setString("type", "dog");
                                obj.getList("pets").add(pet);
                            }
                        }
                    });
            oldVersion++;
        }

        /*
            // Version 3
            class Pet
                @Index
                String name;            // name is indexed
                int type;               // type becomes int
            class Person
                String fullName;
                RealmList<Pet> pets;    // age and pets re-ordered
                int age;
        */
        if (oldVersion == 2) {
            schema.getClass("Pet")
                    .addIndex("name")
                    .addInt("typeTmp")
                    .forEach(new RealmObjectSchema.Iterator() {
                        @Override
                        public void next(DynamicRealmObject obj) {
                            String type = obj.getString("type");
                            if (type.equals("dog")) {
                                obj.setInt("typeTmp", 1);
                            } else if (type.equals("cat")) {
                                obj.setInt("typeTmp", 2);
                            } else if (type.equals("hamster")) {
                                obj.setInt("typeTmp", 3);
                            }
                        }
                    })
                    .removeField("type")
                    .renameField("typeTmp", "type");

            // TODO 4 instructions to change a type, is that acceptable? But the alternative would be an
            // API explosion or to expose something like a "RealmField" which would mean we would have to expose
            // the internal Realm types
            // schema.addField(new RealmField("type", ColumnType.INTEGER))
            oldVersion++;
        }
    }
}
```

Missing:
- [x] JavaDoc
- [x] Unit tests
- [ ] Feedback on missing methods for manipulating schema
- [ ] Feedback from Cocoa
- [ ] Add Null support
- [x] Class iterator: forEach, Java8 stream/RxJava like


cmelchior:
@realm/java This is still WIP, but any feedback on the API: names, methods, is anything obvious missing would be nice. Public methods has been included in the OP to make it easier to see the API. Still missing a lot of unit tests to show usage, but the API tries to promote a builder-pattern type of use.

realm/src/main/java/io/realm/dynamic/RealmObjectSchema.java
```diff
@@ -0,0 +1,491 @@
...
+
+    private static final String TABLE_PREFIX = "class_";
+    private final Realm realm;
+    private final Table table;
+    private final ImplicitTransaction transaction;
+
+    /**
+     * Creates a schema object for a given Realm class.
+     * @param table Table representation of the Realm class
+     */
```
beeender: Other param's doc missed?

beeender:
We actually support `byte` as long. But `byte` is missed in the PR.

realm/src/main/java/io/realm/dynamic/DynamicRealmObject.java
```diff
@@ -0,0 +1,479 @@
...
+     */
+    public void setObject(String fieldName, DynamicRealmObject value) {
+        long columnIndex = row.getColumnIndex(fieldName);
+        if (value == null) {
+            row.nullifyLink(columnIndex);
+        } else {
+            if (value.realm == null || value.row == null) {
+                throw new IllegalArgumentException("Cannot link to objects that are not part of the Realm.");
+            }
+            if (value.realm != null && !realm.getPath().equals(value.realm.getPath())) {
```
beeender: Maybe better to compare the RealmConfiguration here? Now it is guranteed same path realm must have same configuration. Would there be any special case in the future?

cmelchior:
Good catch. This is also missing from the Dynamic API.

realm-jni/src/io_realm_internal_Group.cpp
```diff
@@ -158,6 +158,27 @@ JNIEXPORT jstring JNICALL Java_io_realm_internal_Group_nativeGetTableName(
...
+}
+
+JNIEXPORT void JNICALL Java_io_realm_internal_Group_nativeRenameTable(
+    JNIEnv* env, jobject, jlong nativeGroupPtr, jstring oldName, jstring newName)
+{
+    TR_ENTER_PTR(nativeGroupPtr)
+    try {
+        JStringAccessor old_name(env, oldName);
+        JStringAccessor new_name(env, newName);
+        G(nativeGroupPtr)->rename_table(old_name, new_name);
```
kneth: What if a table with name `oldName` does not exist?

realm-jni/src/io_realm_internal_UncheckedRow.cpp
```diff
@@ -288,3 +288,10 @@ JNIEXPORT jboolean JNICALL Java_io_realm_internal_UncheckedRow_nativeIsAttached
     TR_ENTER_PTR(nativeRowPtr)
     return ROW(nativeRowPtr)->is_attached();
 }
+
+JNIEXPORT jboolean JNICALL Java_io_realm_internal_UncheckedRow_nativeHasField
```
kneth: Isn't the name inconsistent with the other methods? I would prefer it was called `Java_io_realm_internal_IncheckRow_nativeHasColumn`.

realm/src/main/java/io/realm/dynamic/RealmObjectSchema.java
```diff
@@ -0,0 +1,491 @@
+/*
+ * Copyright 2014 Realm Inc.
```
kneth: 2015

realm/src/main/java/io/realm/dynamic/RealmModifier.java
```diff
@@ -0,0 +1,50 @@
...
+     * @see io.realm.annotations.PrimaryKey
+     */
+    PRIMARY_KEY,
+
+    /**
+     * Mark a field as explicitly being able to contain {@code null} values. The default behavior for
+     * allowing {@code null} depends on the type of the field.
+     * @see io.realm.NonNullable
+     */
+    NULLABLE,
```
kneth: Preparing for null support - thanks!

realm/src/main/java/io/realm/dynamic/RealmObjectSchema.java
```diff
@@ -0,0 +1,491 @@
...
+     *
+     * @param currentFieldName Field name to rename.
+     * @param newFieldName The new field name.
+     * @return The updated schema.
+     * @throws IllegalArgumentException if field name doesn't exists or if the new field name already
+     * exists.
+     */
+    public RealmObjectSchema renameField(String currentFieldName, String newFieldName) {
+        checkEmpty(currentFieldName);
+        checkFieldExists(currentFieldName);
```
kneth: You do the check here instead in C++.

realm/src/main/java/io/realm/dynamic/RealmObjectSchema.java
```diff
@@ -0,0 +1,491 @@
...
+     * @throws IllegalStateException if the class doesn't have a primary key defined.
+     */
+    public DynamicRealmObject createObject(Object primaryKeyValue) {
+        long index = table.addEmptyRowWithPrimaryKey(primaryKeyValue);
+        return new DynamicRealmObject(realm, table.getCheckedRow(index));
+    }
+
+    private void setModifiers(long columnIndex, Set<RealmModifier> modifiers) {
+        if (modifiers != null && modifiers.size() > 0) {
+            if (modifiers.contains(RealmModifier.INDEXED)) {
```
kneth: It is possible to remove the search index. It could be done by saying that if `RealmModifier.INDEXED` isn't in `modifiers` and the field has a search index, then it is removed.

realm/src/main/java/io/realm/dynamic/RealmSchema.java
```diff
@@ -0,0 +1,129 @@
+/*
+ * Copyright 2014 Realm Inc.
```
kneth: 2015

realm/src/main/java/io/realm/dynamic/RealmSchema.java
```diff
@@ -0,0 +1,129 @@
...
+        String internalClassName = TABLE_PREFIX + className;
+        if (transaction.hasTable(internalClassName)) {
+            return new RealmObjectSchema(realm, transaction, transaction.getTable(internalClassName));
+        } else {
+            throw new IllegalArgumentException("Class does not exist in this Realm: " + className);
+        }
+    }
+
+    /**
+     * Add a new class to the Realm.
```
kneth: "Add" -> "Adds"

realm/src/main/java/io/realm/internal/Row.java
```diff
@@ -105,4 +105,11 @@
      * data. {@code false} otherwise.
      */
     boolean isAttached();
+
+    /**
+     * Returns {@code true} if the field name exists.
+     * @param fieldName Field name to check.
+     * @return {@code true} if field name exists, {@code false} otherwise.
+     */
+    boolean hasField(String fieldName);
```
kneth: I strongly feel that `io.realm.internal` should be table/column/row oriented.

realm-jni/src/io_realm_internal_UncheckedRow.cpp
cmelchior: Good point. I will change it.

realm/src/main/java/io/realm/internal/Row.java
cmelchior: You are right. It should be hasColumn for consistency.

realm/src/main/java/io/realm/internal/Table.java
```diff
@@ -711,6 +711,10 @@ private void throwDuplicatePrimaryKeyException(Object value) {
         throw new RealmException("Primary key constraint broken. Value already exists: " + value);
     }
 
+    public void removeIndex(long columnIndex) {
```
kneth: Is this similar to `io.realm.internal.Table.removeSearchIndex()`.

realm/src/main/java/io/realm/internal/migration/MigrationCommand.java
```diff
@@ -0,0 +1,23 @@
+/*
+ * Copyright 2014 Realm Inc.
```
kneth: 2015

realm/src/main/java/io/realm/internal/migration/MigrationController.java
```diff
@@ -0,0 +1,34 @@
+/*
+ * Copyright 2014 Realm Inc.
```
kneth: 2015

realm-jni/src/io_realm_internal_Group.cpp
cmelchior: That is being verified in the Java layer currently.

realm/src/main/java/io/realm/dynamic/RealmObjectSchema.java
cmelchior: There is a specific method for removing the index. This should probably be renamed to addModifiers to avoid confusion.

realm/src/main/java/io/realm/internal/Table.java
cmelchior: Yes, since the annotation is called ```@Index``` I feel this is a better name even though we use search index internally.

cmelchior:
Some thoughts. Would it be better to introduce the concept of DynamicRealm right way? Right now having methods like ```createObject``` on a schema object might not seem natural?

This would change the interface to:

```
public class MyMigration implements RealmMigration { 
  @Override
    schema.addClass("Dog").addString("name").addString("type");  // As before

    // but querying/creating/deleting objects would fit the normal pattern but with strings instead
    RealmObjectSchema dogSchema = schema.getClass("Dog");
    DynamicRealmObject dog = realm.createObject("Dog");   
    dog.setName("Fido");
    DynamicRealmResults results = realm.where("Dog").equalTo("name", "Fido");
    realm.removeFromRealm(dog);
  }
}

File: realm/src/androidTest/java/io/realm/RealmObjectSchemaTests.java [hunk 15]
```diff
@@ -0,0 +1,641 @@
...
+    public void testAddRemoveIndexedField() {
+        for (IndexFieldType fieldType : IndexFieldType.values()) {
+            String fieldName;
+            switch (fieldType) {
+                case STRING:
+                    fieldName = AllJavaTypes.FIELD_STRING;
+                    schema.addString(fieldName, EnumSet.of(RealmModifier.INDEXED));
+                    checkAddedAndRemovable(fieldName);
+                    break;
+//                case SHORT:
```
kneth: Why are these cases commented out?

cmelchior: Because they are not supported yet.

kneth: So this PR depends on #1246?

I don't like large blocks of code commented out.

kneth:
Despite my reservations on commented out code :+1:

realm/src/androidTest/java/io/realm/RealmObjectSchemaTests.java
cmelchior: Not really, but it will most likely be merged in just before this one. So call it prematurely optimisation while I was touching the code anyway :). It is the same with Null, these features kinda affect all ongoing work that depend on knowing what types are available in Realm.

vvictor10:
@cmelchior  @kneth Any idea when the new Migration APIs would be merged in?

cmelchior:
@vvictor10 Not yet. There as still some internal debate on how exactly the migration API is going to like. This is our current best idea, but it might still be subject to change.

alazier:
Is there a reason we cant just have one ```addField(String fieldName, FieldType type)``` method rather than having a version for each type?

cmelchior:
@alazier No, we could do that.

I just feel that that having the method for each type makes both a lot more readable and a lot faster to type as autocomplete helps a lot more.

```
schema
  .addStringField("foo");
  .addIntField("bar") 

vs.

schema
  .addField("foo", FieldType.STRING)
  .addField("bar", FieldType.INT)

or 

schema
  .addField("foo", new FieldType(String.class))
  .addField("bar", new FieldType(Integer.class))
```

One of the reasons is also that I really would like to hide the core types, i.e. people should not have to care how Core represents a particular Java type, so I would much prefer if any interface hid that implementation detail and only talked in terms of Java types.

But it is a tradeoff as the number of methods becomes quite large. For reference, e.g. the Cursor class on Android does the same thing: http://developer.android.com/reference/android/database/Cursor.html

But what are you thoughts?

mmvicsonvictor:
@cmelchior Can you please provide an update on this work? Is it close to being completed and pushed up to master?

cmelchior:
Hi @mmvicsonvictor 
This PR unfortunately got sidetracked by bug fixes and other more urgent tasks. Right now it is mostly dependant on https://github.com/realm/realm-java/pull/1369 being merged first.

cmelchior:
This is being superseded by a new PR against the Dynamic/Migration feature branch instead of against master.

Issue #1241, opened 2015-06-25
a NullPointerExpection came out when I call Realm.close


eeoa:
My Realm version is 0.81.1 and the log is 

java.lang.NullPointerExcepiton
at io.realm.Realm.close(REalm.java:243)
at com.cdv.myshare.uploadservice.PreUploadService.commonCommand(PreUploadService.java:413)

The I call Realm.getDefaultInstance() to get the Realm object and it worked fine.But when I closed it , a 
NullPointerExpection would come out occasionally.

eeoa:
In my Application class ,I have set default configuration like 

	public void onCreate() {

		RealmConfiguration config = new RealmConfiguration.Builder(this)
				.build();
		Realm.setDefaultConfiguration(config);

		super.onCreate();
	}

beeender:
Hi!
Is there any chance you created Realm instance in ThreadA but tried to close it in Thread B?

eeoa:
I create and close Realm instance in a function and I don't pass it to another thread exactly.

beeender:
Would you mind to share the function code you create/close the realm?

eeoa:
A few days ago I updated Realm to  0.81.1  and these NullPointerExpection came out. Before the upadating ,my functions worked well .

Now the NullPointerExpection came out in more than one function occasionally. one of these functions is 

		private boolean refreshProgress(Realm db) {

			boolean changed = false;
			boolean alive = false;

			DataProvider dataProvider = DataProvider.getInstance(activity);

			Map<String, Task> tasks = new HashMap<String, Task>();

			for (Task task : dataProvider.getTasks()) {
				tasks.put(task.getWorkID(), task);
			}

			RealmResults<Link> linkResults = null;
			
			if (TextUtils.isEmpty(activityID)) {
				linkResults = db.where(Link.class).equalTo("failure", false)
						.findAllSorted("timeStamp");
			} else {
				linkResults = db.where(Link.class).equalTo("failure", false)
						.equalTo("activityID", activityID)
						.findAllSorted("timeStamp");
			}

			for (int i = 0; i < linkResults.size(); i++) {

				Link link = linkResults.get(i);

				if (!link.isValid()) {
					continue;
				}

				Task task = tasks.get(link.getLinkID());

				if (task == null) {
					continue;
				}

				if (Link.isFailure(link)) {
					if (!task.isFailure()) {
						changed = true;
					}
					task.setFailure(true);
					continue;
				}
				
				alive = true;

[...]

PR #1245, opened 2015-06-26
Throw exception when close called on a diff thread.

beeender:
Without checking the thread in close, some weird logic could be
triggered which is really hard to think about.

I was thinking maybe this would cause #1241 , but according to the feedback, this might not the root cause. But anyway, it is something worth to fix.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,3 +1,6 @@
+0.81.2
+ * Closing realm on another thread different from which it was created on throws an exception.
+
 0.81.1
  * Fixed memory leak causing Realm to never release Realm objects.
 
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,5 +1,5 @@
 0.81.2
- * Closing realm on another thread different from which it was created on throws an exception.
+ * Closing realm on another thread different from where it was created now throws an exception.
 
 0.81.1
  * Fixed memory leak causing Realm to never release Realm objects.
--- a/realm/src/androidTest/java/io/realm/RealmTest.java
+++ b/realm/src/androidTest/java/io/realm/RealmTest.java
@@ -19,6 +19,8 @@ import android.content.Context;
 import android.os.SystemClock;
 import android.test.AndroidTestCase;
 
+import junit.framework.AssertionFailedError;
+
 import org.json.JSONArray;
 import org.json.JSONException;
 import org.json.JSONObject;
@@ -36,12 +38,12 @@ import java.util.Map;
 import java.util.Random;
 import java.util.Scanner;
 import java.util.concurrent.Callable;
+import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicBoolean;
 
 import io.realm.entities.AllTypes;
 import io.realm.entities.AllTypesPrimaryKey;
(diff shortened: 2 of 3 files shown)


changelog.txt
```diff
@@ -1,3 +1,6 @@
+0.81.2
+ * Closing realm on another thread different from which it was created on throws an exception.
```
cmelchior: Closing a Realm on a thread different from which it was created now throws an exception.

beeender:
@cmelchior @emanuelez @nhachicha

cmelchior:
Apart from comment :+1:

changelog.txt
beeender: should it be "**which** it was created **on**" or "**where** it was created"?

realm/src/main/java/io/realm/Realm.java
```diff
@@ -225,9 +226,16 @@ protected void finalize() throws Throwable {
      * <p>
      * It's important to always remember to close Realm instances when you're done with it in order
      * not to leak memory, file descriptors or grow the size of Realm file out of measure.
+     *
+     * @throws java.lang.IllegalStateException if trying to close Realm on a different thread than the
+     * one it was created on.
      */
     @Override
     public void close() {
+        if (this.threadId != Thread.currentThread().getId()) {
```
emanuelez: This logic is now used in two places. It might as well be a private method taking a string as a parameter IMO

beeender: Uh... It is only one line checking, and the `checkIfValid` would be called quite often. I prefer to keep away from another function call :smile:

cmelchior: Yeah, that method would be called for all accessors, better to inline them imo

changelog.txt
cmelchior: @ke says "where" so go with that :)

realm/src/androidTest/java/io/realm/RealmTest.java
```diff
@@ -1719,4 +1720,29 @@ public void testFinalizerThread() throws NoSuchFieldException, IllegalAccessExce
...
+    // Test close Realm in another thread different from where it is created.
+    public void testCloseRealmInDifferentThread() throws InterruptedException {
+        final AtomicBoolean threadReady = new AtomicBoolean(false);
+
+        final Thread thatThread = new Thread(new Runnable() {
+            @Override
+            public void run() {
+                try {
+                    testRealm.close();
+                    fail("Close realm in a different thread should throw IllegalStateException.");
```
nhachicha: This fail will not report in JUnit, you need to capture it & throw it within the testcase's Thread 
ex: https://github.com/realm/realm-java/blob/nh-async-query/realm/src/androidTest/java/io/realm/RealmAsyncQueryTests.java#L241

realm/src/main/java/io/realm/Realm.java
```diff
@@ -225,9 +226,16 @@ protected void finalize() throws Throwable {
...
+            throw new IllegalStateException(INCORRECT_THREAD_CLOSE_MESSAGE);
```
nhachicha: Should we throw a Realm exception instead `RealmIOException` or `RealmException`?

beeender: Now the simliar exceptions are using IllegalStateException. Maybe we fix them in another PR?

cmelchior: Yeah, IllegalState might be the wrong wording here. Agree with @beeender, better to make it consistent here, and then change all places in another PR.

beeender:
@realm/java Please help to review if this would cause anything worse than current.

nhachicha:
:+1:

cmelchior:
:+1:

emanuelez:
:+1: :shipit:

PR #1250, opened 2015-06-27
Type of `schemaVersion` of `RealmConfiguration` should be `long`

kishikawakatsumi:
Should be the type of `schemaVersion` of `RealmConfiguration` `long`?
It seems to assume `long` anywhere using other `schemaVersion`.

https://github.com/realm/realm-java/blob/master/realm/src/main/java/io/realm/Realm.java#L1534
https://github.com/realm/realm-java/blob/master/realm/src/main/java/io/realm/Realm.java#L601-L602
https://github.com/realm/realm-java/blob/master/realm/src/main/java/io/realm/RealmMigration.java#L62

@cmelchior @emanuelez @kneth

### DIFF
--- a/realm/src/main/java/io/realm/RealmConfiguration.java
+++ b/realm/src/main/java/io/realm/RealmConfiguration.java
@@ -72,7 +72,7 @@ public class RealmConfiguration {
     private final String realmFileName;
     private final String canonicalPath;
     private final byte[] key;
-    private final int schemaVersion;
+    private final long schemaVersion;
     private final RealmMigration migration;
     private final boolean deleteRealmIfMigrationNeeded;
     private final RealmProxyMediator schemaMediator;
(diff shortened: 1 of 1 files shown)


cmelchior:
While ```int``` is a more "natural" number time, you are probably right. To make it consistent with the core implementation we should make it a long.

realm/src/main/java/io/realm/RealmConfiguration.java
```diff
@@ -143,7 +143,7 @@ public int hashCode() {
         result = 31 * result + realmFileName.hashCode();
         result = 31 * result + canonicalPath.hashCode();
         result = 31 * result + (key != null ? Arrays.hashCode(key) : 0);
-        result = 31 * result + schemaVersion;
+        result = 31 * result + (int)schemaVersion;
```
kneth: From an academic point, this is a bit dangerous, but in practise no one should get into problem here.

kneth:
:+1:

beeender:
:+1:

PR #1256, opened 2015-07-01
Fixed incorrect old changelog's description

ainame:
`setDate` will check range of date object as Integer.
https://github.com/realm/realm-java/blob/master/realm/src/main/java/io/realm/internal/UncheckedRow.java#L194-L203

So, the lower bound is -2147483648 and will convert "Fri, 13 Dec 1901 20:45:52 GMT".
http://www.epochconverter.com/


cmelchior:
Hi @ainame 
Thanks for noticing and updating.
:+1: 
@realm/java

kneth:
:+1:

beeender:
:+1: :shipit:

PR #1257, opened 2015-07-01
Add support for in-memory Realm - squashed

beeender:
This is squashed commits of #1228 to make a better looking of git history.

I do guarantee this would be exact the same with SHA: 4d8e9df which is the last merge commit of #1228 .


beeender:
@realm/java This would be the last time I bother you tonight.

cmelchior:
I would just have squashed your original branch and force pushed it, but never mind :)
:+1:

kneth:
:+1:

emanuelez:
Yeah... just squash and force push next time :) :+1:

beeender:
But that would loose all of the inline comments... I want gerrit!

emanuelez:
I want Gerrit too! Let's be friends! :)

PR #1265, opened 2015-07-03
Dynamic/Migration design document

cmelchior:
This design document summarises the discussion on the dynamic/migration API. Below is the markdown document that is this PR for easy reference.

# Dynamic / Migration API

This contains a description of how a combined Dynamic/Migration could look like:

**Design goals**

- Should support stepwise migration.

- Introduce as few new classes as possible.

- Keep terminology as close to familiar concepts as possible.

*Secondary design goals*

- Migrations should just feel like switching to dynamic mode.

**Why**

Currently Cococa uses automatic migrations. This concept is great at first, as
it tries to hide the schema concept as much as possible, but it also has a lot
of maintainability issues. The general consensus is we should migrate to
stepwise migration, ie. specify all changes between each version.

This will increase the burden on the developer for simple changes like adding
new classes or fields, but it is code that can be written once and then
forgotten. Automatic migration code has to be maintained always.

At least Android is also going to create an IntelliJ plugin that contains the Realm Browser. Having a IntelliJ plugin will also make it easy to extend the standard "Generate" menu with an option for creating the initial migration code for new Realm classes.

**New API**

To support the above goals I suggest adding the following new classes:

```
io.realm

DynamicRealm       // Copy of Realm interfaces to allow dynamic access
DynamicRealmObject // Wrapper for all current model classes

io.realm

RealmSchema        // Object for controlling Realm-Core tables 
RealmObjectSchema // Object for controlling Realm-Core columns
```
It looks like we can reuse our current RealmList, RealmQuery and RealmResults
classes as they are. It would require adding new dynamic constructors.
This would probably be an acceptable tradeoff compared to introducing a whole
new class? This might change if we decide to remove the RealmObject class at one
point. This should be investigated.

## Examples

```
// Dynamic/Typed is chosen when opening the Realm
// Users should not be able to switch once it is open. Is there any use case for that?

RealmConfiguration config = new RealmConfiguration.Builder(context).build();
RealmConfiguration dynamicConfig = new RealmConfiguration.Builder(context).build();

Realm.setDefaultConfiguration(config);
DynamicRealm.setDefaultInstance(dynamicConfig);

Realm realm = Realm.getDefaultInstance(); // Typed version
DynamicRealm dynamicRealm = DynamicRealm.getDefaultInstance() // Dynamic version


// Creating objects
DynamicRealmObject obj = dynamicRealm.createObject("Dog"); // Dog is name of typed class
obj.setString("name", "fido");
obj.setInt("age", 42);

// Queries
RealmResults<DynamicRealmObject> results = dynamicRealm.where("Dog").equalsTo("name", "fido").findAll();
for (DynamicRealmObject obj : results) {
    obj.getString("name");
    obj.getInt("age");
}

RealmQuery<DynamicRealmQuery> query = new RealmQuery<DynamicRealmObject>(dynamicRealm, "Dog").equalTo("name", "Fido");
DynamicRealmObjet dog = query.findFirst();


// DynamicRealm allows schema access
RealmSchema schema = dynamicRealm.getSchema();
schema.addClass("Cat");

// Migrations now just become a special case of using the dynamic Realm
// The DynamicRealm is used to query, create, delete objects
// The Schema is used to manipulate tables/columns
// Schema.forEach() is the odd duck out as it allows you to iterate elements, but it 
// makes it more fluent.

new MyMigration implements Migration {
  @Override
    RealmSchema = realm.getSchema(); // Annyoing to do this all the time?

    if (oldVersion == 0) {

        schema.getClass("Person")
                .addStringField("fullName")
                .forEach(new DynamicRealm.Iterator() {
                    @Override
                    public void next(DynamicRealmObject obj) {
                        obj.setString("fullName", obj.getString("firstName") + " " + obj.getString("lastName"));
                    }
                })
                .removeField("firstName")
                .removeField("lastName");

        DynamicRealmObjet p = realm.createObject("Person");
        p.setString("name", "John");        
        p.setInt("age", 42);
         
        oldVersion++;
    }

    if (oldVersion == 1) {

        schema.createClass("Cat")
            .addStringField("name")
            .addIntField("age")
            .setObject("owner", schema.getClass("Person"));

        schema.getClass("Person")
            .addIndex("fullName")   

        oldVersion++;
    }
  }
}
```

## Naming

We have two new concepts. Right now we are calling them Dynamic And Schema, is
that accidental or is it the names we want for it?

**Suggestions for 1**
Dynamic 
Untyped

**Suggestions for 2**
Schema
Spec
Definition
Model

Some arguments already brought up:

- Dynamic sounds better than a normal Realm -> Developers might try to start there instead?

- Schema : Everybody hates schemas, but at the same time everybody also knows what the word means.

- Spec is already used by other Java libraries/framework classes to describe class files.

## Current action points

- Is this an acceptable approach?
- Determine naming of concepts
- Investigate impact of removing RealmObject on this suggestion

## Below is the complete list of interfaces for these methods

```
**DynamicRealm**

public class DynamicRealm {

    // Static methods
    static DynamicRealm getInstance(Context context);
    static DynamicRealm getInstance(RealmConfiguration config);
    static DynamicRealm setDefaultInstance(RealmConfiguration config);
    static void removeDefaultInstance();
    static Object getDefaultModule();
    static boolean compactRealm(RealmConfiguration configuration);
    static boolean deleteRealm(RealmConfiguration configuration);

    // Does migrating a Dynamic Realm makes sense?
    static void migrateRealm(RealmConfiguration configuration, RealmMigration migration)
    static synchronized void migrateRealm(RealmConfiguration configuration)

    // New method for accessing schema
    RealmSchema getSchema();            
        
    // All public methods on Realm converted to use Strings and/or DynamicRealmObject
    void close();
    boolean isAutoRefresh();
    void setAutoRefresh(boolean autoRefresh);
    void createAllFromJson(String clazz, JSONArray json);
    void createOrUpdateAllFromJson(String clazz, JSONArray json);
    void createAllFromJson(String clazz, String json);
    void createOrUpdateAllFromJson(String clazz, String json);
    void createAllFromJson(String clazz, InputStream inputStream) throws IOException;
    void createOrUpdateAllFromJson(String clazz, InputStream in) throws IOException;
    void createObjectFromJson(String clazz, JSONObject json);
    DynamicRealmObject createOrUpdateObjectFromJson(String clazz, JSONObject json);
    DynamicRealmObject createObjectFromJson(String clazz, String json);
    DynamicRealmObject createOrUpdateObjectFromJson(String clazz, String json);
    DynamicRealmObject createObjectFromJson(String clazz, InputStream inputStream) throws IOException;
    DynamicRealmObject createOrUpdateObjectFromJson(String clazz, InputStream in) throws IOException;
    void writeCopyTo(File destination) throws IOException;
    void writeEncryptedCopyTo(File destination, byte[] key) throws IOException;
    DynamicRealmObject createObject(String clazz);
    DynamicRealmObject copyToRealm(DynamicRealmObject object);
    DynamicRealmObject copyToRealmOrUpdate(DynamicRealmObject object);
    List<DynamicRealmObject> copyToRealm(Iterable<DynamicRealmObject> objects);
    List<DynamicRealmObject> copyToRealmOrUpdate(Iterable<DynamicRealmObject> objects);
    RealmQuery<DynamicRealmObject> where(String clazz);
    RealmResults<DynamicRealmObject> allObjects(String clazz);
    RealmResults<DynamicRealmObject> allObjectsSorted(String clazz, String fieldName,
                                                             boolean sortAscending);
    RealmResults<DynamicRealmObject> allObjectsSorted(String clazz, String fieldName1,
                                                             boolean sortAscending1, String fieldName2,
                                                             boolean sortAscending2);
    RealmResults<DynamicRealmObject> allObjectsSorted(String clazz, String fieldName1,
                                                             boolean sortAscending1,
                                                             String fieldName2, boolean sortAscending2,
                                                             String fieldName3, boolean sortAscending3);
    RealmResults<DynamicRealmObject> allObjectsSorted(String clazz, String fieldNames[],
                                                             boolean sortAscending[]);
    void addChangeListener(RealmChangeListener listener);
    void removeChangeListener(RealmChangeListener listener);
    void removeAllChangeListeners();
    void refresh();
    void beginTransaction();
    void commitTransaction();
    void cancelTransaction();
    void executeTransaction(DynamicRealm.Transaction transaction);
    void clear(String clazz);
    String getPath();
    RealmConfiguration getConfiguration();
}


**DynamicRealmObject**

public class DynamicRealmObject {
    void set(String fieldName, Object value); // General set that automatically converts to proper type
    Object get(String fieldName); // General get
    boolean getBoolean(String fieldName);
    int getInt(String fieldName);
    short getShort(String fieldName);
    long getLong(String fieldName);
    byte getByte(String fieldName);
    float getFloat(String fieldName);
    double getDouble(String fieldName);
    byte[] getBlob(String fieldName);
    String getString(String fieldName);
    Date getDate(String fieldName);
    DynamicRealmObject getObject(String fieldName);
    RealmList<DynamicRealmObjet> getList(String fieldName);
    void setBoolean(String fieldName, boolean value);
    void setShort(String fieldName, short value);
    void setInt(String fieldName, int value);
    void setLong(String fieldName, long value);
    void setByte(String fieldName, byte value);
    void setFloat(String fieldName, float value);
    void setDouble(String fieldName, double value);
    void setString(String fieldName, String value);
    void setBlob(String fieldName, byte[] value);
    void setDate(String fieldName, Date value);
    void setObject(String fieldName, DynamicRealmObject value);
    void setList(String fieldName, RealmList<DynamicRealmObjet> list);
    boolean isNull(String fieldName);
    boolean hasField(String fieldName);
    String[] getFieldNames();
    RealmFieldType getType(String fieldName); // Expose the underlying Realm type.
    void removeFromRealm();
}

**RealmSchema**

public class RealmSchema {
    RealmObjectSchema getClass(String className);
    RealmObjectSchema createClass(String className);
    void addClass(RealmObjectSchema);
    void removeClass(String className);
    RealmObjectSchema renameClass(String oldName, String newName);
    Set<RealmObjectSchema> getAllClasses();
}

**RealmObjectSchema**

public interface RealmObjectSchemaInterface {
    RealmObjectSchema addStringField(String fieldName);
    RealmObjectSchema addStringField(String fieldName, Set<RealmModifier> modifiers);
    RealmObjectSchema addShortField(String fieldName);
    RealmObjectSchema addShortField(String fieldName, Set<RealmModifier> modifiers);
    RealmObjectSchema addIntField(String fieldName);
    RealmObjectSchema addIntField(String fieldName, Set<RealmModifier> modifiers);
    RealmObjectSchema addLongField(String fieldName);
    RealmObjectSchema addLongField(String fieldName, Set<RealmModifier> modifiers);
    RealmObjectSchema addBooleanField(String fieldName);
    RealmObjectSchema addBooleanField(String fieldName, Set<RealmModifier> modifiers);
    RealmObjectSchema addByteArrayField(String fieldName);
    RealmObjectSchema addByteArrayField(String fieldName, Set<RealmModifier> modifiers);
    RealmObjectSchema addFloatField(String fieldName);
    RealmObjectSchema addFloatField(String fieldName, Set<RealmModifier> modifiers);
    RealmObjectSchema addDoubleField(String fieldName);
    RealmObjectSchema addDoubleField(String fieldName, Set<RealmModifier> modifiers);
    RealmObjectSchema addDateField(String fieldName);
    RealmObjectSchema addDateField(String fieldName, Set<RealmModifier> modifiers);
    RealmObjectSchema addObjectField(String fieldName, RealmObjectSchema objectSchema);
    RealmObjectSchema addListField(String fieldName, RealmObjectSchema objectSchema);
    RealmObjectSchema removeField(String fieldName);
    RealmObjectSchema renameField(String currentFieldName, String newFieldName);
    RealmObjectSchema addIndex(String fieldName);
    RealmObjectSchema removeIndex(String fieldName);
    boolean hasIndex(String fieldName);
    RealmObjectSchema addPrimaryKey(String fieldName);
    RealmObjectSchema removePrimaryKey();
    boolean hasPrimaryKey();
    RealmObjectSchema setNullable(String fieldName, boolean nullable);
    String[] getFieldNames();
    RealmFieldType getFieldType(String fieldName)
    RealmObjectSchema forEach(Iterator iterator);
}
```

### DIFF
--- a/DYNAMIC.md
+++ b/DYNAMIC.md
@@ -1,0 +1,314 @@
+# Dynamic / Migration API
+
+This contains a description of how a combined Dynamic/Migration could look like:
+
+
+*Design goals*
+
+- Should support stepwise migration.
+
+- Introduce as few new classes as possible.
+
+- Keep terminology as close to familiar concepts as possible.
+
+
+*Secondary design goals*
+
+- Migrations should just feel like switching to dynamic mode.
+
+
+
+* Why *
+
+Currently Cococa uses automatic migrations. This concept is great at first, as it tries to hide the schema concept as much as possible, but it also has a lot of maintainability issues. The general consensus is we should migrate to stepwise migration, ie. specify all changes between each version.
+
+This will increase the burden on the developer for simple changes like adding new classes or fields, but it is code that can be written once and then forgotten. Automatic migration code has to be maintained always.
+
+At least Android is also going to create a IntelliJ plugin for the Realm Browser. Adding functionality for automatically creating migration code that can be pasted in would be fairly straight forward.
+
+
+* New API *
+
+To support the above goals I suggest adding the following new classes:
+
+io.realm.dynamic
+
+DynamicRealm       // Copy of Realm interfaces to allow dynamic access
+DynamicRealmObject // Wrapper for all current model classes
+
+io.realm.schmea
+
+RealmSchema        // Object for controlling Realm-Core tables
+RealmObjectSchema  // Object for controlling Realm-Core columns
+
+It looks like we can reuse our current RealmList, RealmQuery and RealmResults classes as they are. It would propably require adding new dynamic constructors. This would probably be an acceptable tradeoff compared to introducing a whole new class? This might change if we decide to remove the RealmObject class at one point. This should be investigated.
+
+
+## Examples
+
+```
+// Dynamic/Typed is chosen when opening the Realm
+// Users should not be able to switch once it is open. Is there any use case for that?
+
+RealmConfiguration config = new RealmConfiguration.Builder(context).build();
+RealmConfiguration dynamicConfig = new RealmConfiguration.Builder(context).build();
+
+Realm.setDefaultConfiguration(config);
+DynamicRealm.setDefaultInstance(dynamic);
+
+Realm realm = Realm.getDefaultInstance(); // Typed version
+DynamicRealm dynamicRealm = DynamicRealm.getDefaultInstance() // Dynamic version
+
+
+// Creating objects
+DynamicRealmObject obj = dynamicRealm.createObject("Dog"); // Dog is name of typed class
+obj.setString("name", "fido");
+obj.setInt("age", 42);
+
+// Queries
+RealmResults<DynamicRealmObject> results = dynamicRealm.where("Dog").equalsTo("name", "fido").findAll();
+for (DynamicRealmObject obj : results) {
+    obj.getString("name");
+    obj.getInt("age");
+}
+
+RealmQuery<DynamicRealmQuery> query = new RealmQuery<DynamicRealmObject>(dynamicRealm, "Dog").equalTo("name", "Fido");
+DynamicRealmObjet dog = query.findFirst();
+
+
+// DynamicRealm allows schema access
+RealmSchema schema = dynamicRealm.getSchema();
+schema.addClass("Cat");
+
+
+
+// Migrations now just become a special case of using the dynamic Realm
+// The DynamicRealm is used to query, create, delete objects
+// The Schema is used to manipulate tables/columns
+// Schema.forEach() is the odd duck out as it allows you to iterate elements, but it 
+// makes it more fluent.
+
+new MyMigration implements Migration {
+  @Override
+  public void migrate(DynamicRealm realm, long oldVersion, long newVersion) {
+    RealmSchema = realm.getSchema(); // Annyoing to do this all the time?
+
+    if (oldVersion == 0) {
+
+        schema.getClass("Person")
+                .addStringField("fullName")
+                .forEach(new DynamicRealm.Iterator() {
+                    @Override
+                    public void next(DynamicRealmObject obj) {
+                        obj.setString("fullName", obj.getString("firstName") + " " + obj.getString("lastName"));
+                    }
+                })
+                .removeField("firstName")
+                .removeField("lastName");
+
+        DynamicRealmObjet p = realm.createObject("Person");
+        p.setString("name", "John");        
+        p.setInt("age", 42);
+         
+        oldVersion++;
+    }
+
+    if (oldVersion == 1) {
+
+        schema.addClass("Cat")
+            .addStringField("name")
+            .addIntField("age")
+            .setObject("owner", schema.getClass("Person"));
+
+        schema.getClass("Person")
+            .addIndex("fullName")   
+
+        oldVersion++;
+    }
+  }
+}
+```
+
+
+## Naming
+
+We have two new concepts. Right now we are calling them Dynamic And Schema, is that accidential or is it the names we want for it?
+
+* Suggestions for 1*
+Dynamic 
+Untyped
+
+* Suggestions for 2*
+Schema
+Spec
+Definition
+
+Some arguments already brough up:
+
+- Dynamic sounds better than a normal Realm -> Developers might try to start there instead?
+
+- Schema : Everybody hates schemas, but at the same time everybody also knows what the word means.
+
+- Spec is already used by other Java libraries/framework classes to describe class files.
+
+
+
+## Current action points
+
+- Is this an acceptable approach?
+- Determine naming of concepts
+- Investigate tradeoffs reusing RealmQuery, RealmList and RealmResults for dynamic access.
+- Investigate impact of removing RealmObject on this suggestion
+
+
+## Below is the complete list of interfaces for these methods
+
+
+*DynamicRealm*
+
+public class DynamicRealm {
+
+    // Static methods
+    static DynamicRealm getInstance(Context context);
+    static DynamicRealm getInstance(RealmConfiguration config);
+    static DynamicRealm setDefaultInstance(RealmConfiguration config);
+    static void removeDefaultInstance();
+    static Object getDefaultModule();
+    static boolean compactRealm(RealmConfiguration configuration);
+    static boolean deleteRealm(RealmConfiguration configuration);
+
+    // Does migrating a Dynamic Realm makes sense?
+    static void migrateRealm(RealmConfiguration configuration, RealmMigration migration)
+    static synchronized void migrateRealm(RealmConfiguration configuration)
+
+    // New method for accessing schema
+    RealmSchema getSchema();            
+        
+    // All public methods on Realm converted to use Strings and/or DynamicRealmObject
+    void close();
+    boolean isAutoRefresh();
+    void setAutoRefresh(boolean autoRefresh);
+    void createAllFromJson(String clazz, JSONArray json);
+    void createOrUpdateAllFromJson(String clazz, JSONArray json);
+    void createAllFromJson(String clazz, String json);
+    void createOrUpdateAllFromJson(String clazz, String json);
+    void createAllFromJson(String clazz, InputStream inputStream) throws IOException;
+    void createOrUpdateAllFromJson(String clazz, InputStream in) throws IOException;
+    void createObjectFromJson(String clazz, JSONObject json);
+    DynamicRealmObject createOrUpdateObjectFromJson(String clazz, JSONObject json);
+    DynamicRealmObject createObjectFromJson(String clazz, String json);
+    DynamicRealmObject createOrUpdateObjectFromJson(String clazz, String json);
+    DynamicRealmObject createObjectFromJson(String clazz, InputStream inputStream) throws IOException;
+    DynamicRealmObject createOrUpdateObjectFromJson(String clazz, InputStream in) throws IOException;
+    void writeCopyTo(File destination) throws IOException;
+    void writeEncryptedCopyTo(File destination, byte[] key) throws IOException;
+    DynamicRealmObject createObject(String clazz);
+    DynamicRealmObject copyToRealm(DynamicRealmObject object);
+    DynamicRealmObject copyToRealmOrUpdate(DynamicRealmObject object);
+    List<DynamicRealmObject> copyToRealm(Iterable<DynamicRealmObject> objects);
+    List<DynamicRealmObject> copyToRealmOrUpdate(Iterable<DynamicRealmObject> objects);
+    RealmQuery<DynamicRealmObject> where(String clazz);
+    RealmResults<DynamicRealmObject> allObjects(String clazz);
+    RealmResults<DynamicRealmObject> allObjectsSorted(String clazz, String fieldName,
+                                                             boolean sortAscending);
+    RealmResults<DynamicRealmObject> allObjectsSorted(String clazz, String fieldName1,
+                                                             boolean sortAscending1, String fieldName2,
+                                                             boolean sortAscending2);
+    RealmResults<DynamicRealmObject> allObjectsSorted(String clazz, String fieldName1,
+                                                             boolean sortAscending1,
+                                                             String fieldName2, boolean sortAscending2,
+                                                             String fieldName3, boolean sortAscending3);
+    RealmResults<DynamicRealmObject> allObjectsSorted(String clazz, String fieldNames[],
+                                                             boolean sortAscending[]);
+    void addChangeListener(RealmChangeListener listener);
+    void removeChangeListener(RealmChangeListener listener);
+    void removeAllChangeListeners();
+    void refresh();
+    void beginTransaction();
+    void commitTransaction();
+    void cancelTransaction();
+    void executeTransaction(DynamicRealm.Transaction transaction);
+    void clear(String clazz);
+    String getPath();
+    RealmConfiguration getConfiguration();
+}
+
+
+*DynamicRealmObject*
+
+public class DynamicRealmObject {
+    boolean getBoolean(String fieldName);
+    int getInt(String fieldName);
+    short getShort(String fieldName);
+    long getLong(String fieldName);
+    byte getByte(String fieldName);
+    float getFloat(String fieldName);
+    double getDouble(String fieldName);
+    byte[] getBlob(String fieldName);
+    String getString(String fieldName);
+    Date getDate(String fieldName);
+    DynamicRealmObject getObject(String fieldName);
+    RealmList<DynamicRealmObjet> getList(String fieldName);
+    void setBoolean(String fieldName, boolean value);
+    void setShort(String fieldName, short value);
+    void setInt(String fieldName, int value);
+    void setLong(String fieldName, long value);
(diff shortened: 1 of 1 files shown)


timanglade:
Many thanks for this.

1. Reading this you seem to imply only Dynamic gives you access to Schema, so why not make all 4 classes part of the package (e.g. io.realm.dynamic), and prefix them all with the same prefix (e.g. “Dynamic”): DynamicRealmSchema, DynamicRealmObjectSchema.
2. If this really a first-party, alternative interface with many uses cases, I don’t think having consistent parallels (i.e. DynamicRealmObject, DynamicRealmQuery, DynamicRealmResults, etc.) is a problem.
3. I don’t mind the word “Schema” per se. I think it’s the right word and if it’s 100% a subconcept of Dynamic then it won’t hurt us much.
4. I like “Untyped” better than “Dynamic” but it might be weird to see that pop up during Migrations, whereas Dynamic is more innocuous. Would love to hear more suggestions here.

DYNAMIC.md
```diff
@@ -0,0 +1,326 @@
...
+    void removeChangeListener(RealmChangeListener listener);
+    void removeAllChangeListeners();
+    void refresh();
+    void beginTransaction();
+    void commitTransaction();
+    void cancelTransaction();
+    void executeTransaction(DynamicRealm.Transaction transaction);
+    void clear(String clazz);
+    String getPath();
+    RealmConfiguration getConfiguration();
```
beeender: I would like to ask for a public `getVersion()` call some where. User sometimes will just check the version and delete the db if it is not what he/she wants.

DYNAMIC.md
```diff
@@ -0,0 +1,326 @@
...
+}
+
+
+*DynamicRealmObject*
+
+public class DynamicRealmObject {
+    boolean getBoolean(String fieldName);
+    int getInt(String fieldName);
```
beeender: Since we are going to have boxed type, the interface numbers will grow... IMO, make it `Integer getInt(String fieldName)` won't be a good choice, since it will slow down the primitive calls. `getIntBoxed`, this name doesn't look good either.

DYNAMIC.md
```diff
@@ -0,0 +1,326 @@
...
+    RealmObjectSchema addDateField(String fieldName, Set<RealmModifier> modifiers);
+    RealmObjectSchema addObjectField(String fieldName, RealmObjectSchema objectSchema);
+    RealmObjectSchema addListField(String fieldName, RealmObjectSchema objectSchema);
+    RealmObjectSchema removeField(String fieldName);
+    RealmObjectSchema renameField(String currentFieldName, String newFieldName);
+    RealmObjectSchema addIndex(String fieldName);
+    RealmObjectSchema removeIndex(String fieldName);
+    RealmObjectSchema addPrimaryKey(String fieldName);
+    RealmObjectSchema setNonNullable(String fieldName); 
+    RealmObjectSchema setNullable(String fieldName);
```
beeender: I feel `setNullable(String fieldName, boolean nullable)` would be better. Or corresponding th `addIndex`  and `removeIndex`, maybe it should be `enableNullable` or `disableNullable`?

DYNAMIC.md
cmelchior: We actually had a discussion on this I think the pattern was something like this

`Integer i = obj.isNull(fieldName) ? null : obj.getInt(fieldName)`

A pattern that even had its own name

DYNAMIC.md
cmelchior: Hmm, yeah. Good point.

DYNAMIC.md
cmelchior: Its already on `Realm` but just package protected. It would be fairly easy to make it public.

beeender: That one only works after user opened realm. If a migration needed, user doesn't have a chance to check version.

cmelchior:
Iterating a bit on this. Then `Dynamic` might be a dangerous name as that implies DocumentStore behaviour, i.e. "No schema", which is false. So maybe `Untyped` better convey what is actually happening.

timanglade:
1. Just reiterating my preference for having the same prefix in front of the schema classes e.g. DynamicRealmSchema (or UntypedRealmSchema) to ease discovery and minimize confusion from new users.
2. To me Untyped means “no type can be set” not “properties can be added/removed and types can be edited”. As much as I hate “Dynamic”, I think it’s the best proposal so far, and something people might actually guess is what they need if they go through our API docs looking for a way to access Realm in a way that lets them add/remove properties on the fly.

cmelchior:
@timanglade Personally I think that Adding "Dynamic" in front of "RealmSchema" and "RealmObjectSchema" makes them to long and verbose. Also it is slightly wrong, as the schema is applicable for both the typed and dynamic Realm.

```
RealmSchema
RealmObjectSchema

vs.

DynamicRealmSchema
DynamicRealmObjectSchema
```

kneth:
Interestingly, we talk about "model classes" and never schemas. Introducing the RealmObjectSchema (or DynamicRealmObjectSchema) is somewhat redundant as "model class" and RealmObjectSchema is about the same underlying concept.

We could keep the loosely "model class", and then introduce `RealmModel` and `RealmObjectModel`. The `RealmModel` is a collection of `RealmObjectModel`s, and the `RealmObjectModel` can be implemented using a "model class".

On the other hand, we go for the schema, I think we should stop saying "model class" and start saying "object schema" (which can be implemented as a Java class extending `RealmObject`).

timanglade:
OK, if Schema are accessible from both Dynamic and regular makes sense to avoid the prefix. How will the packages look? Looking at your feature pr it looks like we will have an io.realm.dynamic. Where will RealmSchema and RealmObjectSchema be located?

cmelchior:
After talking to @emanuelez. We agreeded that all public API classes for Realm should be in `io.realm`. 
This also makes it possible to use `package protected` methods better, which are not visible from the outside. So while the `io.realm` package get more classes than I would like, it really simplifies our code + makes it easy to communicate to the outside:

```
io.realm -> public API
io.realm.internal -> "Private" API
io.realm.android -> Android only classes
``
 
The feature PR still has the `io.realm.dynamic` package. It will be removed when #1498 is merged.

**Comment by cmelchior** [comment]
@kneth  I added "Model" as a suffix to the list in the OP, and you are right. We need to carefully consider how we talk about the different concepts in Realm. Not sure what the right answer is so the more input the better.

**Comment by timanglade** [comment]
Thanks for the thoughtful analysis + explanation.

Since our number of public classes is going through the roof, can we take the time to write a few short paragraph (+ maybe a helpful diagram) for https://realm.io/docs/java/latest/api/overview-summary.html (the first frame people see when they load https://realm.io/docs/java/latest/api/) to explain which classes people might actually want to check out first (`Realm`, `RealmObject`, etc.), a one line explanation of what concerns they represent. And additionally at the bottom, we could list other cluster of classes (`Dynamic*`, `*Schema`, etc.) that are for more advanced use-cases, and what they can be used for.

**Comment by cmelchior** [comment]
We don't have much wiggle room to customize the Javadoc, but having an intro in our documentation definitely makes a lot of sense.

**Comment by zaki50** [comment]
I think we should avoid 'Model' since the word 'model' is used for the structure defined by Java code.

The migration API treats the structure in the database engine and using different word makes us easy to explain what is migration.

**Comment by cmelchior** [comment]
Ok, so you like `RealmObjectSchema` better than `RealmModelSchema` (fair argument btw.), what about `RealmClassSchema` since `RealmObject` might be hard to associate with a whole type of elements? But then again people are already familiar with RealmObject since they are required to extend it?

**Comment by cmelchior** [comment]
E.g. this is difficult because we right now have a method called `RealmSchema.getObjectSchema()`. The naming of that method also depends on what we call the other things.

- `RealmSchema.getClassSchema()`
- `RealmSchema.getObjectSchema()`
- `RealmSchema.getSchema()`

are all options

**Comment by cmelchior** [comment]
Right now we also have a lot of methods on `RealmSchema`:
- `schema.createClass()`
- `schema.removeClass()`

Using `RealmClassSchema` seems more coherent in that context unless we rename those methods.

- `schema.createObject()` -> eew
- `schema.createObjectClass()` -> maybe, but still eew

**Comment by cmelchior** [comment]
Personally I am starting to lean towards

```
RealmSchema schema;
schema.addClass("Foo")
schema.removeClass("Bar")

RealmClassSchema fooSchema = schema.addClass("Foo")
fooSchema.addStringField("Baz")
```

This means we talk about a  `Realm class` as the model definition (=schema). It would also match our annotation. "Realm object class" could be used when in doubt talking about either top-level Realm or a model class. We should then clean our documentation of any reference to "model".

What do you think?

**Comment by timanglade** [comment]
My strong personal preference would be to map the structure to the classes they shadow (i.e. `Realm` and `RealmObject`) so I would name them `RealmSchema` and `RealmObjectSchema`.

This “class” stuff is confusing to me. In a database context you don’t add a class to a schema, so why use that word here? How about just `realmSchema.addObject("Foo")` ?

**Comment by cmelchior** [comment]
Yes, that is roughly what it is now, only problem I have with `object` is that it means something different when outside the schema context. `Class` maps more naturally to the java-file itself (= the schema)

**Comment by timanglade** [comment]
I think using the word “class” introduces another layer of thinking that is largely unnecessary (but maybe I’m missing something. if `object` is confusing how about just `realmSchema.add("Foo")` ?

**Comment by cmelchior** [comment]
I would be perfectly happy removing the `class` reference. The reason it has merit is that it is the name of of our variables, e.g.

```
DynamicRealm.createObject(String className)
```

**Comment by cmelchior** [comment]
Looking at Cocoa they also use the terms `RLMSchema` and `RLMObjectSchema`. So mirroring these concepts in Java with `RealmSchema` and `RealmObjectSchema` makes a lot of sense.

This also means that we should remove variables like `className` and instead use `objectName`
After some discussion we can also adjusted the `RealmObjectSchema` as things like:

```
addField(int.class, "foo")
addField(Integer.class, "foo", Required.class)
```

Is expressive enough for our current need. It will not be expressive enough to describe more advanced concepts, e..g if our annotations start setting variables like `@DefaultValue(String val)`. However I feel that is bridge we can cross when we get there, and those use cases are better solving by providing something like a `RLMProperty` in Cocoa which is a Class fully describing a single property. There is no need to expose that yet though.

cmelchior:
Digging a bit more through the Cocoa API, and they seem to use the term `ObjectClass` which has an `RLMObjectSchema`. 

I'll try experimenting going back to a collection-based sort of API, i.e. only `add`, `get`, `create`, `remove` to see how that feels.

cmelchior:
#1677 has been merged to the feature branch

Issue #1283, opened 2015-07-10
Annotation processor may not have been executed using Proguard


Ralphilius:
### Issue
I got Annotation issue while enabling Proguard. Without it, the package is working. Raw mesage:
```
Caused by: io.realm.exceptions.RealmException: Could not find the generated io.realm.NavDrawerItemRealmProxy class: Annotation processor may not have been executed.
```

### Gradle Build file
```
buildscript {
    repositories {
        maven { url 'https://maven.fabric.io/public' }
    }
 
    dependencies {
        classpath 'io.fabric.tools:gradle:1.+'
    }
}
apply plugin: 'com.android.application'
apply plugin: 'io.fabric'
 
repositories {
    maven { url 'https://maven.fabric.io/public' }
}
 
 
android {
    compileSdkVersion 22
    buildToolsVersion "22.0.1"
    defaultConfig {
        applicationId '### Stripped out ###'
        minSdkVersion 15
        targetSdkVersion 22
        versionCode 123
        versionName '1.2.3'
    }
    lintOptions {
        disable 'InvalidPackage'
    }
 
    packagingOptions {
        pickFirst 'META-INF/services/javax.annotation.processing.Processor'
    }
 
    buildTypes {
        release {
            minifyEnabled true
            proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
            signingConfig getSigningConfig()
            applicationVariants.all { variant ->
                variant.outputs.each { output ->
                    def date = new Date();
                    def file = output.outputFile
                    def formattedDate = date.format('yyyyMMddHHmmss')
                    def fileName = file.name.replace("app-release","app-release-"+defaultConfig.versionName)
                    output.outputFile = new File(file.parent,fileName)
                }
            }
        }
    }
    productFlavors {
    }
}
 
dependencies {
    compile fileTree(dir: 'libs', include: ['*.jar'])
    compile 'com.android.support:appcompat-v7:22.2.0'
    compile 'com.android.support:recyclerview-v7:21.0.0'
    compile 'com.android.support:support-v4:22.2.0'
    compile 'com.jakewharton:butterknife:7.0.1'
    compile 'io.realm:realm-android:0.80.1'
    compile 'com.google.code.gson:gson:2.3.1'
    compile 'com.mcxiaoke.volley:library:1.0.16'
    compile 'com.parse.bolts:bolts-android:1.2.0'
    compile 'com.squareup.picasso:picasso:2.5.2'
[...]

PR #1290, opened 2015-07-13
Keep RealmModule class from being obfuscated

beeender:
Issue #1283

### DIFF
--- a/examples/realmModuleExample/app/proguard-rules.pro
+++ b/examples/realmModuleExample/app/proguard-rules.pro
@@ -17,6 +17,7 @@
 #}
 
 # Realm ProGuard setup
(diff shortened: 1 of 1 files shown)


beeender:
@realm/java Help to review. And the PR for the doc is here https://github.com/realm/realm.io/pull/753

emanuelez:
:+1:

cmelchior:
:+1: 
Since it works correctly when combined with custom modules, it it probably related to ProGuard optimising something it shouldn't when only having the ```DefaultRealmModule``` (which is loaded through reflection). It needs further investigation, but lets just merge this fix now to avoid confusion.

PR #1295, opened 2015-07-15
Added new setDefaultConfiguration method.

GavinThePacMan:
I needed to add this to combat possible key change issues brought about by user's removing their credentials. I think this is a needed addition for anyone using keys stored in the android keystore.

### DIFF
--- a/realm/src/main/java/io/realm/Realm.java
+++ b/realm/src/main/java/io/realm/Realm.java
@@ -543,10 +543,28 @@ public final class Realm implements Closeable {
      * @see RealmConfiguration for details on how to configure a Realm.
      */
     public static void setDefaultConfiguration(RealmConfiguration configuration) {
+       setDefaultConfiguration(configuration, false);
+    }
+
+    /**
+     * Sets the {@link io.realm.RealmConfiguration} used when calling {@link #getDefaultInstance()}.  
+     * 
+     * @param configuration RealmConfiguration to use as the default configuration.
+     * @param removePastConfigurations Boolean value to determine whether or not we should remove cached configurations.
+     * @see RealmConfiguration for details on how to configure a Realm.
+     */
+    public static void setDefaultConfiguration(RealmConfiguration configuration, boolean removePastConfigurations) {
         if (configuration == null) {
             throw new IllegalArgumentException("A non-null RealmConfiguration must be provided");
         }
         defaultConfiguration = configuration;
+
+        if(removePastConfigurations){
+	        String realmPath = newConfiguration.getPath();
+	        List<RealmConfiguration> pathConfigurationCache = globalPathConfigurationCache.get(realmPath);
+	        pathConfigurationCache = new CopyOnWriteArrayList<RealmConfiguration>();
+	        globalPathConfigurationCache.put(realmPath, pathConfigurationCache);
+        }
     }
 
     /**
--- a/realm/src/main/java/io/realm/Realm.java
+++ b/realm/src/main/java/io/realm/Realm.java
@@ -560,7 +560,7 @@ public final class Realm implements Closeable {
         defaultConfiguration = configuration;
 
         if(removePastConfigurations){
-	        String realmPath = newConfiguration.getPath();
+	        String realmPath = configuration.getPath();
 	        List<RealmConfiguration> pathConfigurationCache = globalPathConfigurationCache.get(realmPath);
 	        pathConfigurationCache = new CopyOnWriteArrayList<RealmConfiguration>();
 	        globalPathConfigurationCache.put(realmPath, pathConfigurationCache);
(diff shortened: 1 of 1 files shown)


emanuelez:
Thank you for your contribution! :)

The code you provide does not compile because `newConfiguration` is not defined.
I suggest you to add at least one unit test as well.

GavinThePacMan:
My apologies for that. I had already built it from a clone of this repository and trusted myself to edit on my fork without compiling! Doh! -.-

I have fixed the typo now anyway and it is working as expected.

Regarding the unit test, I was looking at the ones that are already written and couldn't find one for the standard ``setDefaultConfiguration()`` so am unsure how to proceed. I would love a bit of direction however as I enjoy contributing to such a great project! 

Thanks!

emanuelez:
retest this please

emanuelez:
@GavinThePacMan the comment above is not for you. it's simply to trigger a new build on our CI server :)

realm/src/main/java/io/realm/Realm.java
```diff
@@ -543,10 +543,28 @@ public static Realm getInstance(RealmConfiguration configuration) {
...
+     * @see RealmConfiguration for details on how to configure a Realm.
+     */
+    public static void setDefaultConfiguration(RealmConfiguration configuration, boolean removePastConfigurations) {
         if (configuration == null) {
             throw new IllegalArgumentException("A non-null RealmConfiguration must be provided");
         }
         defaultConfiguration = configuration;
+
+        if(removePastConfigurations){
+	        String realmPath = configuration.getPath();
```
emanuelez: We don't allow TABs in the code

realm/src/main/java/io/realm/Realm.java
```diff
@@ -543,10 +543,28 @@ public static Realm getInstance(RealmConfiguration configuration) {
...
+     * @param removePastConfigurations Boolean value to determine whether or not we should remove cached configurations.
+     * @see RealmConfiguration for details on how to configure a Realm.
+     */
+    public static void setDefaultConfiguration(RealmConfiguration configuration, boolean removePastConfigurations) {
         if (configuration == null) {
             throw new IllegalArgumentException("A non-null RealmConfiguration must be provided");
         }
         defaultConfiguration = configuration;
+
+        if(removePastConfigurations){
```
emanuelez: We require a space between `if` and `(`

GavinThePacMan:
retest this please

emanuelez:
No need for you to invoke our CI. It will pick up the new commit automatically. The magic phrase is only used to force a rebuild :)

GavinThePacMan:
Ah thank you!

realm/src/main/java/io/realm/Realm.java
```diff
@@ -543,10 +543,28 @@ public static Realm getInstance(RealmConfiguration configuration) {
...
+     * @see RealmConfiguration for details on how to configure a Realm.
+     */
+    public static void setDefaultConfiguration(RealmConfiguration configuration, boolean removePastConfigurations) {
         if (configuration == null) {
             throw new IllegalArgumentException("A non-null RealmConfiguration must be provided");
         }
         defaultConfiguration = configuration;
+        
+        if (removePastConfigurations) {
+	        String realmPath = configuration.getPath();
```
emanuelez: This line still includes a TAB

emanuelez:
@GavinThePacMan Any news on this?

GavinThePacMan:
@emanuelez my apologies replying so late. I was surprised that my text editor was using tabs instead of spaces! I changed it now anyway. Thanks!

GavinThePacMan:
Hi @emanuelez, any update on this? Thanks!

cmelchior:
Hi @GavinThePacMan 
Sorry for the late reply. We have been hit by people going to and from vacation.

I am a bit surprised that adding this method is required as we should ideally clear the configuration cache when it is no longer needed (if not, that is probably a bug). We are really restrictive about adding new constructors as we just removed a bunch of them to avoid the mental overload. So I would be really interested in seeing a unit test that showed adding this would really be needed? 

E.g. if you already encrypted a file with the old credentials, you would have to do something like this

1) Open file with old configuration
2) `Realm.writeEncryptedCopyToFile(newPassword, backup location)
3) Close realm
4) Delete original file
5) Move backup to old location
6) Open with new configuration.

Which is covered by this unit test: https://github.com/realm/realm-java/blob/master/realm/src/androidTest/java/io/realm/RealmTest.java#L1790-Lundefined

Checking back in our history I think we might by accident already fixed this in this PR: https://github.com/realm/realm-java/pull/1356

Which means it should already work if you try our `0.83.0-SNAPSHOT`?

GavinThePacMan:
@cmelchior, many thanks for your reply.

Your PR: #1356 seems to have solved this problem more eloquently! 

Many thanks,
Gavin

PR #1297, opened 2015-07-16
Transaction should not block opening Realm

beeender:
Minor issue, but it could blocking Realm reading which is not what we expected.

This is the test case only, the problem has been fixed when we introduce
the schema caches.

Since we are caching the schemas for typed Realm, when a transaction
lock is holding by one type Realm, open Realm instance in another thread
should not be blocked.
But if there is no typed Realm instance opened, DynamicRealm transaction
should block the typed Realm openning.

### DIFF
--- a/realm/realm-library/src/androidTest/java/io/realm/RealmTests.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/RealmTests.java
@@ -2801,4 +2801,25 @@ public class RealmTests {
         assertEquals(0, realm.where(Cat.class).count());
         assertTrue(realm.isEmpty());
     }
+
+    // When there is a transaction holding by a typed Realm in one thread, getInstance from the
+    // other thread should not be blocked since we have cached the schemas already.
+    @Test
+    public void getInstance_shouldNotBeBlockedByTransactionInAnotherThread()
+            throws InterruptedException {
+        final CountDownLatch latch = new CountDownLatch(1);
+        realm.beginTransaction();
+
+        Thread thread = new Thread(new Runnable() {
+            @Override
+            public void run() {
+                Realm realm = Realm.getInstance(realmConfig);
+                realm.close();
+                latch.countDown();
+            }
+        });
+        thread.start();
+        TestHelper.awaitOrFail(latch);
+        realm.cancelTransaction();
+    }
 }


beeender:
@realm/java Small change, help to review. Thx!

kneth:
:+1:

realm/src/main/java/io/realm/Realm.java
```diff
@@ -689,11 +689,13 @@ private static void validateAgainstExistingConfigurations(RealmConfiguration new
     private static void initializeRealm(Realm realm) {
         long version = realm.getVersion();
         boolean commitNeeded = false;
```
emanuelez: `commitNeeded` looks pretty useless now

cmelchior:
Keep in mind that when we add multiprocess support, transactions are the only synchronisation mechanism between processes and Realm initialisation code must not run concurrently in two processes. I didn't check what effect this PR has on that (vacation and all :) ).

beeender:
@cmelchior Actually this problem was found when i were working on the multi process support.
See https://github.com/realm/realm-java/pull/1300

When one process holding a write lock, there is no chance for others to open a realm anymore until the write lock released.

realm/src/main/java/io/realm/Realm.java
beeender: I thought so. But there is a chance setVersion throw an exception and the transaction needs to be cancelled? Would this be a case?

beeender:
http://stackoverflow.com/questions/31521421/realm-causing-lots-of-anr This might be caused by this.

beeender:
@realm/java Hey, guys, we really should have this. Think about user might need to initialize a huge db from cloud in the background thread, and then user open a Realm in the UI thread to wait for refreshing. IMO, it is totally a reasonable scenario but it would cause ANR right now.

beeender:
We still have race condition problem here like below: 
Realm is not initialized before, version == UNVERSIONED

Thread A | Thread B
------------- | --------------
check the version | check the version
Start transaction to write the version | Start transaction to write the version
Get the lock and start write | Wait
Commit the transaction | Wait
Start another transaction immediately for writing data | Wait
Get the transaction lock | Wait
Commit transaction | wait
Do other stuff | Finally get the transaction lock, and try to write the initial version which is not need any more

For above case, `Real.getInstance()` could be blocked for a really long time, which is not necessary at all.
If we have a API like `beginTransactionNoWait` then the problem is quite easy to be solved.

```
while(true) {
    if (beginTransactionNoWait()) {
        version = getVersion();
        if (version == UNVERSIONED) {
            //writeVersion and return
        } else {
            cancelTransaction();
        }
    } else {
        sleep(5);
    }
}
```

emanuelez:
It looks like the way to solve this is to check the version inside the transaction, not outside.

beeender:
Yes and no. Inside the transaction, we surely need to check the version since it could be changed by previous transaction. And we have to check the version before transaction, since we might not to have begin the transaction at all.

emanuelez:
We would have to begin the transaction and cancel it. They do that in the Cocoa binding as well

beeender:
@realm/java

realm/realm-library/src/androidTest/java/io/realm/RealmTests.java
```diff
@@ -2801,4 +2801,27 @@ public void clear_all() {
         assertEquals(0, realm.where(Cat.class).count());
         assertTrue(realm.isEmpty());
     }
+
+    // When there is a transaction holding by a typed Realm in one thread, getInstance from the
+    // other thread should not be blocking since we have cached the schemas already.
```
zaki50: [other thread should not be blocking] -> [other thread should not be blocked]?

zaki50:
:+1: with minor comment

realm/realm-library/src/androidTest/java/io/realm/RealmTests.java
stk1m1: Nothing other than what @zaki50 says. :+1:

stk1m1:
Nothing other than what @zaki50 says. :+1:

emanuelez:
:+1:

cmelchior:
:+1:

PR #1300, opened 2015-07-17
Multiprocess support

### LINKED ISSUE #1091: Add multi process support
Currently Realm for Java only support multi threading but not multi process.

beeender:
This PR will test the possibility of current status of multi process first. Then maybe add more features to make multi process work better.

Fixes #1091
- [ ] Core bug https://github.com/realm/realm-core/issues/887 needs to be fixed
- [x] Permission control
- [x] Compact Realm when there is an opened Realm in other process
- [x] Delete
- [ ] Share Realm in the app directory with another apk
- [x] Example two processes in one apk
- [ ] Example two processes in two apks
- [x] Consider the following implementation with fine grained notification.
- [ ] Blog post.
- [ ] Decouple Android only logic from `BaseRealm` (Optional) 
- [x] Handle schema changes


beeender:
Right now, merging this PR means more virtual spaces are needed for the shared group in RealmDaemon. Need https://github.com/realm/realm-core/issues/838 to be implemented first.

realm/realm-library/src/main/java/io/realm/RealmChangeDaemon.java
```diff
@@ -0,0 +1,90 @@
...
+import java.util.concurrent.Executors;
+import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+
+import io.realm.internal.SharedGroup;
+import io.realm.internal.log.RealmLog;
+
+/**
+ * Daemon thread sleeps until one ore more commits have been made to the corresponding Realm. When change happens, the
```
nhachicha: `s/ore/or`

Zhuinden:
Native Wait for change method is notified by changes in other process? :surprised:

Zhuinden:
I guess the question is when multiprocess will be supported for migrations / schema changes / lock version change.

cmelchior:
@Zhuinden In theory with the things we have merged to master, interprocess should be fully supported now, but we still need to review it with that in mind. At least last we checked there was a few unfortunate race conditions when opening Realms, e.g opening a Realm for the first time in a process requires a write transaction and if another process is writing at the same time, you suddenly completely blocked one of the processes.

leisim:
@cmelchior Is there any news on this issue? I still need multiprocess support :/

Elvis10ten:
Hello, Is multi process now supported in realm 3 ?. I also have a question here about multi-process: http://stackoverflow.com/questions/42733931/realm-multi-process-best-practice

Zhuinden:
This PR is like, super-outdated :smile: there is no such thing as `HandlerController`!

beeender:
Replaced by #5473

PR #1303, opened 2015-07-17
Fix default module not generated

### LINKED ISSUE #1302: DefaultModule should be created while library modules in use.
Problem described in so http://stackoverflow.com/questions/31450084/realm-getdefaultmodule-returns-null?noredirect=1#comment50901688_31450084

When there is no module defined in app project and defined modules in lib project are in use, default module should still be created for app.

beeender:
When the app project is built with modules defined in the lib project,
default realm should still be generated if there is no module defined in
the app project.

Fix #1302

### DIFF
--- a/realm-annotations-processor/src/main/java/io/realm/processor/ModuleMetaData.java
+++ b/realm-annotations-processor/src/main/java/io/realm/processor/ModuleMetaData.java
@@ -16,12 +16,22 @@
 
 package io.realm.processor;
 
-import io.realm.annotations.RealmModule;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
 
 import javax.annotation.processing.ProcessingEnvironment;
 import javax.annotation.processing.RoundEnvironment;
-import javax.lang.model.element.*;
-import java.util.*;
+import javax.lang.model.element.AnnotationMirror;
+import javax.lang.model.element.AnnotationValue;
+import javax.lang.model.element.Element;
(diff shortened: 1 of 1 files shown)


beeender:
@emanuelez @kneth @cmelchior 

I am not quite sure if this is the desired behaviour. And also, is it necessary to check if the default module exists in the lib project?

vvictor10:
@emanuelez @kneth @cmelchior I am working on a project which has a 'lib' module which is a common module for the other app modules we want to have. 'mobile', 'wear' etc are examples. I want to have all my realm code in the 'lib' module and the way I got it working is to have the RealmModule, RealmObject(s) in the lib module and I did not tag the RealmModule as a library module. So the DefaultRealmModule is generated for the lib module and things seem to be going in the right direction. Are you guys going to break that pattern?

emanuelez:
@vvictor10 is your library module doing this properly? https://realm.io/docs/java/latest/#sharing-schemas

cmelchior:
This PR will as far as I can see also make it possible for Library projects (i.e. projects with @RealmModule(library=true)) to create the DefaultRealmModule. Introducing that possibility is not a good IMO as depending as library projects should really control their state as much as possible and not depend on some class that might get overridden by app projects.

It sounds really odd if referencing modules in libraries prevent the default module being generated, because the annotation processor can't see library code. So perhaps we should try to reproduce this behaviour?

@vvictor10 This also means that you shouldn't really depend on the default module in library projects, but instead use explicit modules.

cmelchior:
Superseded by #1325

PR #1316, opened 2015-07-23
Add analytics on annotation

timanglade:
Asynchronously submits build information to Realm when the annotation processor is running

To be clear: this does *not* run when your app is in production or on your end-user’s devices; it will only run when you build your app from source.

Why are we doing this? Because it helps us build a better product for you. None of the data personally identifies you, your employer or your app, but it *will* help us understand what Realm version you use, what host OS you use, etc. Having this info will help with prioritizing our time, adding new features and deprecating old features. Collecting an anonymized bundle & anonymized MAC is the only way for us to count actual usage of the other metrics accurately. If we don’t have a way to deduplicate the info reported, it will be useless, as a single developer building their app on Windows 10 times would report 10 times more than a single developer that only builds once from Mac OS X, making the data all but useless. No one likes sharing data unless it’s necessary, we get it, and we’ve debated adding this for a long long time. Since Realm is a free product without an email signup, we feel this is a necessary step so we can collect relevant data to build a better product for you.

Currently the following information is reported:
- What version of Realm is being used
- What OS you are running on
- An anonymised MAC address and bundle ID to aggregate the other information on.

### DIFF
--- a/realm-annotations-processor/src/main/java/io/realm/processor/ClassMetaData.java
+++ b/realm-annotations-processor/src/main/java/io/realm/processor/ClassMetaData.java
@@ -21,12 +21,24 @@ import io.realm.annotations.Index;
 import io.realm.annotations.PrimaryKey;
 
 import javax.annotation.processing.ProcessingEnvironment;
-import javax.lang.model.element.*;
+import javax.lang.model.element.Element;
+import javax.lang.model.element.ElementKind;
+import javax.lang.model.element.ExecutableElement;
+import javax.lang.model.element.Modifier;
+import javax.lang.model.element.PackageElement;
+import javax.lang.model.element.TypeElement;
+import javax.lang.model.element.VariableElement;
 import javax.lang.model.type.DeclaredType;
 import javax.lang.model.type.TypeKind;
 import javax.lang.model.type.TypeMirror;
 import javax.lang.model.util.Types;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
 
 /**
  * Utility class for holding metadata for RealmProxy classes.
--- a/realm-annotations-processor/src/main/java/io/realm/processor/ClassMetaData.java
+++ b/realm-annotations-processor/src/main/java/io/realm/processor/ClassMetaData.java
@@ -16,30 +16,17 @@
 
 package io.realm.processor;
 
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
+import io.realm.annotations.Ignore;
+import io.realm.annotations.Index;
+import io.realm.annotations.PrimaryKey;
 
 import javax.annotation.processing.ProcessingEnvironment;
-import javax.lang.model.element.Element;
-import javax.lang.model.element.ElementKind;
-import javax.lang.model.element.ExecutableElement;
-import javax.lang.model.element.Modifier;
-import javax.lang.model.element.PackageElement;
-import javax.lang.model.element.TypeElement;
-import javax.lang.model.element.VariableElement;
+import javax.lang.model.element.*;
 import javax.lang.model.type.DeclaredType;
 import javax.lang.model.type.TypeKind;
 import javax.lang.model.type.TypeMirror;
 import javax.lang.model.util.Types;
-
-import io.realm.annotations.Ignore;
-import io.realm.annotations.Index;
-import io.realm.annotations.PrimaryKey;
+import java.util.*;
 
 /**
  * Utility class for holding metadata for RealmProxy classes.
@@ -375,6 +362,10 @@ public class ClassMetaData {
             return false;
         }
         return (!type.endsWith(".RealmObject") && !type.endsWith("RealmProxy"));
+    }
+
+    public String getPackageName() {
+        return packageName;
     }
 
     public String getFullyQualifiedClassName() {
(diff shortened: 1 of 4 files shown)


realm-annotations-processor/src/main/java/io/realm/processor/ClassMetaData.java
```diff
@@ -16,30 +16,17 @@
...
+import javax.lang.model.element.*;
 import javax.lang.model.type.DeclaredType;
 import javax.lang.model.type.TypeKind;
 import javax.lang.model.type.TypeMirror;
 import javax.lang.model.util.Types;
-
-import io.realm.annotations.Ignore;
-import io.realm.annotations.Index;
-import io.realm.annotations.PrimaryKey;
+import java.util.*;
```
beeender: I think import with wild card is not a good practice.
http://stackoverflow.com/questions/187453/import-package-vs-import-package-specifictype

realm-annotations-processor/src/main/java/io/realm/processor/RealmAnalytics.java
```diff
@@ -0,0 +1,151 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.realm.processor;
+
+import java.io.IOException;
+import java.io.UnsupportedEncodingException;
+import java.net.*;
```
beeender: same

realm-annotations-processor/src/main/java/io/realm/processor/RealmAnalytics.java
```diff
@@ -0,0 +1,151 @@
...
+//
+// Currently the following information is reported:
+// - What version of Realm is being used
+// - What OS you are running on
+// - An anonymous MAC address and bundle ID to aggregate the other information on.
+public class RealmAnalytics {
+    private static RealmAnalytics INSTANCE;
+    private static final String ADDRESS_PREFIX = "http://api.mixpanel.com/track/?data=";
+    private static final String ADDRESS_SUFFIX = "&ip=1";
+    private static final String TOKEN = "ce0fac19508f6c8f20066d345d360fd0";
```
beeender: Any security risk to expose the TOKEN? I assume it is fine. Someone could use it to inject wrong data?

realm-annotations-processor/src/main/java/io/realm/processor/RealmAnalytics.java
```diff
@@ -0,0 +1,151 @@
...
+// long long time. Since Realm is a free product without an email signup, we
+// feel this is a necessary step so we can collect relevant data to build a
+// better product for you.
+//
+// Currently the following information is reported:
+// - What version of Realm is being used
+// - What OS you are running on
+// - An anonymous MAC address and bundle ID to aggregate the other information on.
+public class RealmAnalytics {
+    private static RealmAnalytics INSTANCE;
```
beeender: INSTANCE -> instance since it is not final

realm-annotations-processor/src/main/java/io/realm/processor/RealmProcessor.java
```diff
@@ -28,8 +25,8 @@
 import javax.lang.model.element.Element;
 import javax.lang.model.element.ElementKind;
 import javax.lang.model.element.TypeElement;
-
-import io.realm.annotations.RealmClass;
+import java.io.IOException;
+import java.util.*;
```
beeender: same

realm-annotations-processor/src/main/java/io/realm/processor/Utils.java
```diff
@@ -1,19 +1,17 @@
...
-import java.util.List;
-
 import javax.annotation.processing.Messager;
 import javax.annotation.processing.ProcessingEnvironment;
-import javax.lang.model.element.Element;
-import javax.lang.model.element.ExecutableElement;
-import javax.lang.model.element.Modifier;
-import javax.lang.model.element.TypeElement;
-import javax.lang.model.element.VariableElement;
+import javax.lang.model.element.*;
```
beeender: same

realm-annotations-processor/src/main/java/io/realm/processor/RealmProcessor.java
```diff
@@ -127,6 +128,7 @@ public boolean process(Set<? extends TypeElement> annotations, RoundEnvironment
                 continue;
             }
             Utils.note("Processing class " + metadata.getSimpleClassName());
+            packages.add(metadata.getPackageName());
```
beeender: The packages will have lots of duplicated entries. Use `Set` instead? And `getPackageName` doesn't work before calling `metadata.generate()` .

beeender:
:+1:

realm-annotations-processor/src/main/java/io/realm/processor/RealmAnalytics.java
```diff
@@ -0,0 +1,172 @@
...
+// once from Mac OS X, making the data all but useless. No one likes sharing
+// data unless it’s necessary, we get it, and we’ve debated adding this for a
+// long long time. Since Realm is a free product without an email signup, we
+// feel this is a necessary step so we can collect relevant data to build a
+// better product for you.
+//
+// Currently the following information is reported:
+// - What version of Realm is being used
+// - What OS you are running on
+// - An anonymous MAC address and bundle ID to aggregate the other information on.
```
cmelchior: ```an anonymous MAC``` mac address sounds wrong to me as from that line it isn't clear that the MAC is scrambled. Perhaps "anonymised" or "hashed"?

realm-annotations-processor/src/main/java/io/realm/processor/RealmAnalytics.java
```diff
@@ -0,0 +1,172 @@
...
+    private static final String EVENT_NAME = "Run";
+    private static final String JSON_TEMPLATE
+            = "{\n"
+            + "   \"event\": \"%EVENT%\",\n"
+            + "   \"properties\": {\n"
+            + "      \"token\": \"%TOKEN%\",\n"
+            + "      \"distinct_id\": \"%USER_ID%\",\n"
+            + "      \"Anonymized MAC Address\": \"%USER_ID%\",\n"
+            + "      \"Anonymized Bundle ID\": \"%APP_ID%\",\n"
+            + "      \"Binding\": \"java\",\n"
```
cmelchior: perhaps "android" instead, as we might have to use "java" later for the pure Java binding.

cmelchior:
:+1:

PR #1325, opened 2015-07-28
RealmConfiguration now accepts Null modules

### LINKED ISSUE #1302: DefaultModule should be created while library modules in use.

cmelchior:
Fixes #1302

We had a fair amount of issues over the last weeks where people did not properly understanding when to use Realm.getDefaultModule(), which means they used even when no DefaultModule was created. This is fairly common when all your RealmClasses are in a library and your application just uses the RealmClasses from that library. 

This PR relaxes the requirements and allows a smaller programmer error by ignoring null values parsed to `RealmConfiguration.setModules()`. It turned out that returning a empty module would require a lot more code.

See #1302 for possible other solutions.

@realm/java

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,4 +1,5 @@
 0.82
+ * RealmConfiguration.setModules() now accept ignore null values which Realm.getDefaultModule() might return.
  * Trying to access a deleted Realm object throw throws a proper IllegalStateException.
  * Added in-memory Realm support.
  * Closing realm on another thread different from where it was created now throws an exception.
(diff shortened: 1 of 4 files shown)


beeender:
:+1:

emanuelez:
:+1:

PR #1331, opened 2015-07-30
Annotations, processor and JNI are now subproject and not standalone

emanuelez:
This will make working with some project easier and make refactorings more effective


cmelchior:
Nice, but AFAIK there are some trickery with regard to the annotations as they are included in both the Realm and the annotation processor. This is also why it doesn't compile currently.

emanuelez:
retest this please

emanuelez:
Closing because it's been made obsolete by #1585

PR #1351, opened 2015-08-05
Update the AP to support the bytecode manipulation.

emanuelez:
This is a breaking change that will only allow to use Realm as a plugin

**THIS PR DEPENDS ON #1216 SO DO NOT MERGE IT!**


emanuelez:
Superseded by #2196

PR #1356, opened 2015-08-06
No longer caches the wrong encryption key

### LINKED ISSUE #1352: Cannot connect to realmdb when passed wrong encryption key for the first time
Hallo.
I have a problem. I use users password to encrypt databse. When opening application, I ask user for the password, generate encryption key (using SHA512) and try to use it to open database. Everything works perfect as long as user pass correct password. But it is impossible to open database, when user pass wrong password for the first time. 

I have ended in this test:

I have default realm database encrypted by password "a". "makeSHA1Hash" is method to generate encryption key based on the password. Source code:

        try {
            Realm.getInstance(context, makeSHA1Hash("x"));
            Log.d(this.getClass().getName(), "MV - x is ok");
        } catch (Exception e) {
            Log.d(this.getClass().getName(), "MV - x is wrong");
        }

        try {
            Realm.getInstance(context, makeSHA1Hash("a"));
            Log.d(this.getClass().getName(), "MV - a is ok");
        } catch (Exception e) {
            Log.d(this.getClass().getName(), "MV - a is wrong");
        }

        try {
            Realm.getInstance(context, makeSHA1Hash("b"));
            Log.d(this.getClass().getName(), "MV - b is ok");
        } catch (Exception e) {
            Log.d(this.getClass().getName(), "MV - b is wrong");
        }

        try {
            Realm.getInstance(context, makeSHA1Hash("a"));
            Log.d(this.getClass().getName(), "MV - a is ok");
        } catch (Exception e) {
            Log.d(this.getClass().getName(), "MV - a is wrong");
        }

        try {
            Realm.getInstance(context, makeSHA1Hash("d"));
            Log.d(this.getClass().getName(), "MV - d is ok");
        } catch (Exception e) {
            Log.d(this.getClass().getName(), "MV - d is wrong");
        }

This sample code ends in this (wrong?) result:
MV - x is wrong
MV - a is wrong
MV - b is wrong
MV - a is wrong
MV - d is wrong

When I delete the first try - catch code with "X" (so the fist db open is successful - using the password "a"), the (correct?) result is :
MV - a is ok
MV - b is wrong
MV - a is ok
MV - d is wrong

Could you please help me how to solve situation when user pass wrong password and cannot be than connected even the second password is correct?

Thank you.
M.

cmelchior:
Fixes #1352 

@realm/java


beeender:
:+1:

emanuelez:
:+1:

PR #1364, opened 2015-08-07
loadLibrary should be a synchronized operation

beeender:
Otherwise it might cause #1350

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,5 +1,6 @@
 0.83
  * BREAKING CHANGE: Removed deprecated methods and constructors from the Realm class.
+ * Fixed a bug which might cause failure when loading the native library.
 
 0.82.1
  * Fixed a bug where using the wrong encryption key first caused the right key to be seen as invalid.
--- a/realm/src/main/java/io/realm/internal/RealmCore.java
+++ b/realm/src/main/java/io/realm/internal/RealmCore.java
@@ -25,7 +25,6 @@ import java.io.Serializable;
 import java.lang.reflect.Field;
 import java.nio.ByteBuffer;
 import java.util.Locale;
-import java.util.concurrent.atomic.AtomicBoolean;
 
 /**
  * Utility methods for Realm Core.
@@ -39,7 +38,7 @@ public class RealmCore {
     private static final String JAVA_LIBRARY_PATH = "java.library.path";
 //*/
 
-    private static AtomicBoolean libraryIsLoaded = new AtomicBoolean(false);
+    private static volatile boolean libraryIsLoaded = false;
 
 /*
     private static String getJniFileName()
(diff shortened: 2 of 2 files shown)


beeender:
@cmelchior @emanuelez @nhachicha I didn't spend time on reproducing this, but it should be reasonable fix.

changelog.txt
```diff
@@ -1,5 +1,6 @@
 0.83
  * BREAKING CHANGE: Removed deprecated methods and constructors from the Realm class.
+ * Fixed a bug which might cause failure of loading native library.
```
cmelchior: failure when loading the native library.

realm/src/main/java/io/realm/internal/RealmCore.java
```diff
@@ -39,7 +38,7 @@
     private static final String JAVA_LIBRARY_PATH = "java.library.path";
 //*/
 
-    private static AtomicBoolean libraryIsLoaded = new AtomicBoolean(false);
+    private static boolean libraryIsLoaded = false;
```
cmelchior: I don't think this boolean is needed, at least not if you look at the JavaDoc for Java 7: http://docs.oracle.com/javase/7/docs/api/java/lang/Runtime.html#loadLibrary%28java.lang.String%29

The Android JavaDoc doesn't have the same clarity though, but it would be really odd behaviour if it loaded the same library twice?

http://developer.android.com/reference/java/lang/Runtime.html#loadLibrary(java.lang.String)

cmelchior: Found this in the Android source code btw:

```
    // TODO: should be synchronized, but dalvik doesn't support synchronized internal natives.
    private static native String nativeLoad(String filename, ClassLoader loader, String ldLibraryPath);
```

cmelchior: Tracked down the source in AOSP: http://osxr.org/android/source/dalvik/vm/Native.cpp#0318
Boolean is not required :)

If that is removed. It looks fine to me :+1:

beeender:
Interesting, one line above that is 
```java
// nativeLoad should be synchronized so there's only one LD_LIBRARY_PATH in use regardless
// of how many ClassLoaders are in the system, but dalvik doesn't support synchronized
// internal natives.
synchronized (this) {
    return nativeLoad(name, loader, ldLibraryPath);
}
```

And the synchronized block is add by 652abda3 which is release in 4.3. Almost can explain estornino's problem in #1350.

cmelchior:
Yes that sounds plausible.

beeender:
retest this please

bmunkholm:
I would keep the boolean. I recall there was some problems at some point, and that's why it was added. It's two lines and doesn't harm. We call it once for each public class, and it's just faster than having to search all libs a number of times.

beeender:
retest this please

realm/src/main/java/io/realm/internal/RealmCore.java
nhachicha: Either remove the Boolean or make it `static volatile` otherwise liveness may occur

realm/src/main/java/io/realm/internal/RealmCore.java
```diff
@@ -94,8 +93,10 @@ public static void print(String caption, AbstractCursor<?> cursor) {
     }
 */
 
-    public static void loadLibrary() {
-        if (libraryIsLoaded.get())
+    // Although loadLibrary is synchronized internally from AOSP 4.3, for the compatibility reason,
+    // KEEP synchronized here for the old devices!
+    public static synchronized void loadLibrary() {
+        if (libraryIsLoaded)
             // only load library once
```
nhachicha: actually the comment here may be misleading, the lib will be loaded once anyway, this check will only save us one more trip to native land.

cmelchior:
:+1:

PR #1343, opened 2015-08-03
Introduce RealmBase and SharedGroupManager

cmelchior:
**Description**
Two new abstractions has been added: SharedGroupManager and RealmBase

- *SharedGroupManager*: Encapsulates the SharedGroup and Transaction to encapsulate reuse across Realm/DynamicRealm.
- *RealmBase*: Abstract class for sharing code/logic between Realm/DynamicRealm. 

**Missing**:
- [x] SharedGroup should be shared between DynamicRealm/Realm for the same configuration on the    same thread.
- [x] Move all non-static methods from Realm to RealmBase (those that don't take Clazz<> as parameter)
- [x] Naming: RealmInstance -> RealmBase and RealmFileWriter -> SharedGroupManager.

- [x] Should we move RealmInstance to the public `io.realm` instead of `io.realm.internal`? It would solve problems with inheriting JavaDoc and make the code cleaner as we don't have to explicitly call super for a lot of methods. Downside is a useless file in the JavaDoc? Can't find a way to not include it while still allowing the JavaDoc to show up: ANSWER: RealmBase moved to `io.realm` as we are consolidating all public API's there.

PR #1369, opened 2015-08-07
Full Dynamic API

cmelchior:
This introduces a full blown Dynamic API (apart from access to Schema which will be part of #880)

It depends on #1343 being merged first.

Missing:
- [x] Generic getType() or get() in order to work with more generalised code ( #1394 ).
- [x] JavaDoc for new Dynamic/Base classes
- [ ] Unit tests for the functionality that isn't shared with the typed counter parts
  * [ ] DynamicRealm
  * [ ] DynamicRealmObject
- [x] Discuss where the current booleans should be placed. Suggestion -> Move to enums in seperate classes instead
- [x] Split Dynamic API and typed API sharing the SharedGroup. The amount of book-keeping for this is insane and it introduces lots of weird edge cases. Writes across instances, setting auto refresh effects both instances and so on.  
- [ ] Handle nullable and boxed types fields after #1396 merged

----
Know limitations
- DynamicRealm is a stripped down version of a regular Realm. In the first iteration it will only provide simple query/create capabilities, i.e. no `json*` methods and no `createOrUpdate` and friends.
- `deleteRealm`, `compactRealm`, `migrateRealm` are not available in DynamicRealm. 
- only `where` is available as a query method on DynamicRealm, i.e. no `allObjects` and `allObjectsSorted`.


cmelchior:
retest this please

jpiche:
Is there a reason why `getClass`, `addClass`, and `removeClass` take String arguments? Should they have overloads that take `Class<? extends RealmObject>` arguments?

cmelchior:
Closing this, so we can merge it into #1496 instead. All progress is copied to the new PR

Issue #1342, opened 2015-08-03
Timeout on Context.finalize()


HPIWicket:
Hey,

we started seeing following error with the update to Realm 0.82.0:
```
java.util.concurrent.TimeoutException: io.realm.internal.Context.finalize() timed out after 10 seconds
       at java.lang.Object.equals(Object.java:205)
       at java.util.ArrayList.remove(ArrayList.java:418)
       at io.realm.internal.Context.cleanRows(Context.java:69)
       at io.realm.internal.Context.executeDelayedDisposal(Context.java:61)
       at io.realm.internal.Context.finalize(Context.java:113)
       at java.lang.Daemons$FinalizerDaemon.doFinalize(Daemons.java:209)
       at java.lang.Daemons$FinalizerDaemon.run(Daemons.java:192)
       at java.lang.Thread.run(Thread.java:818)
```
We also see some longer access times to the Realm after we added some "realm.clear(RealmObject.class)" calls.

Is this a known issue or is there a better way to clear a whole table?

beeender:
@HPIWicket Can you share your project with us which can reproduce this issue? to help@realm.io .

HPIWicket:
I cannot do that :/. But I will try to create a project that can reproduce this issue and try to send it to you.

emanuelez:
Thank you! We're looking forward to that

TheRishka:
I've managed to catch such error as well. Can't understand the pattern to reproduce it, but it appears sometimes.
```
java.util.concurrent.TimeoutException: io.realm.internal.SharedGroup.finalize() timed out after 10 seconds
       at io.realm.internal.SharedGroup.nativeClose(SharedGroup.java)
       at io.realm.internal.Context.asyncDisposeSharedGroup(Context.java:90)
       at io.realm.internal.SharedGroup.finalize(SharedGroup.java:183)
       at java.lang.Daemons$FinalizerDaemon.doFinalize(Daemons.java:191)
       at java.lang.Daemons$FinalizerDaemon.run(Daemons.java:174)
       at java.lang.Thread.run(Thread.java:818)
```

tmeunier:
Hi,
I was trying to update to 0.82.0 but I get an error of the same kind on intensive usage of a large realm (70k objects).
Two threads are extracting data (statistics) and when one closes the realm this error occurs.

```java
java.util.concurrent.TimeoutException: io.realm.internal.TableQuery.finalize() timed out after 10 seconds
[...]

PR #1379, opened 2015-08-11
Unit test to expose timeout in finalizer

kneth:
#1342 is reporting a timeout of the finalizer watchdog. When you have many objects in a Realm, and you iterate through them in multiple threads, many temporary objects are created. Disposal of these temporary objects might take too long (Android's limit is 10 seconds), and the exception as mentioned in #1342 is thrown.

Relevant background info: http://stackoverflow.com/questions/24021609/how-to-handle-java-util-concurrent-timeoutexception-android-os-binderproxy-fin


kneth:
We should probably find a data structure more efficient than `HashMap` but for now it is a good fix.

I have tested locally on Unnecto U511 (256 MB memory) and OnePlus One (3 GB memory) with success.

:+1:

beeender:
:+1: :100:  And needs to disable the test before merge~

meierjan:
In which version was this fixed?

kneth:
The fix closing this issue goes back to version 0.89. We have gradually moved over to phantom references to get rid of similar issues.

meierjan:
Okay, because I have an app using version 0.9 and I am getting this - but maybe it's coming from something else. Thanks anyways for the very fast response

kneth:
@meierjan If you can, consider to upgrade to a recent version. As I said, we have fixed similar issues but in 0.89 we fixed the first major one.

Zhuinden:
> 2.2.2 (2017-01-16)
>
> Use phantom reference to solve the finalize time out issue (#2496).

But if you support devices < API 16, then 3.1.2+ is the way to go... so 3.1.4 or 3.3.1.

meierjan:
Thanks for the info guys. Rlly amazed by how helpful you are. Yeah maybe I have to take the time and update.

PR #1395, opened 2015-08-17
isValid should not throw

### LINKED ISSUE #1393: isValid() doesn't behave as expected
I'm using `realm-android:0.83.0-SNAPSHOT`.

When I run this simple code:
``` java
Realm realm = Realm.getInstance(this);
realm.beginTransaction();
Dog myDog = realm.createObject(Dog.class);
myDog.setDogName("myDog");
realm.commitTransaction();

realm.beginTransaction();
myDog.removeFromRealm();
realm.commitTransaction();

Timber.e("myDog still valid: " + myDog.isValid());
```

`myDog.isValid()` throws an `IllegalStateException`

```java
     Caused by: java.lang.IllegalStateException: Object is no longer managed by Realm. Has it been deleted?
            at io.realm.internal.InvalidRow.getStubException(InvalidRow.java:180)
            at io.realm.internal.InvalidRow.isAttached(InvalidRow.java:171)
            at io.realm.RealmObject.isValid(RealmObject.java:107)
```

How can I check if an object was removed from realm?
Or does my little example not work because I'm calling this function from the same thread? However, I don't think that `isValid()` should ever throw such an exception.

kneth:
#1393 reports a bug in `isValid()`. When an object is deleted from a Realm, the underlaying row is change to an instance of the invalid row. Methods of invalid rows throw an exception unconditionally.

Fixes #1393

@realm/java

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -2,6 +2,7 @@
  * BREAKING CHANGE: Removed deprecated methods and constructors from the Realm class.
  * Fixed a bug which might cause failure when loading the native library.
  * Fixed a bug which might trigger a timeout in Context.finalize().
+ * Fixed a bug which might cause RealmObject.isValid() to throw an exception if the object is deleted.
 
 0.82.1
  * Fixed a bug where using the wrong encryption key first caused the right key to be seen as invalid.
--- a/realm/src/androidTest/java/io/realm/RealmObjectTest.java
+++ b/realm/src/androidTest/java/io/realm/RealmObjectTest.java
@@ -543,4 +543,19 @@ public class RealmObjectTest extends AndroidTestCase {
         } catch (IllegalStateException ignored) {
         }
     }
+
+    public void testIsValid() {
+        testRealm.beginTransaction();
+        Dog dog = testRealm.createObject(Dog.class);
+        dog.setName("Fido");
+        testRealm.commitTransaction();
+
+        assertTrue(dog.isValid());
+
+        testRealm.beginTransaction();
+        dog.removeFromRealm();
+        testRealm.commitTransaction();
+
+        assertFalse(dog.isValid());
+    }
 }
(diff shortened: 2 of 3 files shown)


changelog.txt
```diff
@@ -2,6 +2,7 @@
  * BREAKING CHANGE: Removed deprecated methods and constructors from the Realm class.
  * Fixed a bug which might cause failure when loading the native library.
  * Fixed a bug which might trigger a timeout in Context.finalize().
+ * Fixed a bug which cause RealmObject.isValid() to throw an exception if the object is deleted.
```
cmelchior: which could cause

realm/src/main/java/io/realm/RealmObject.java
```diff
@@ -103,8 +103,13 @@ public void removeFromRealm() {
      * @return {@code true} if the object is still accessible, {@code false} otherwise or if it is a
      * standalone object.
      */
-    public boolean isValid() {
-        return row != null && row.isAttached();
+    public boolean
+    isValid() {
```
beeender: The declaration should be in one line?
`public boolean isValid() {`

realm/src/main/java/io/realm/RealmObject.java
```diff
@@ -103,8 +103,13 @@ public void removeFromRealm() {
...
+        try {
+            return row != null && row.isAttached();
+        } catch (IllegalStateException e) {
+            return false;
```
cmelchior: It would be much better to change InvalidRow so it returned `false`. Using exceptions to control flow is very bad for performance.

kneth: I assumed that the purpose of `InvalidRow` is to throw an exception for every single method (see the comment at the top of the file).

cmelchior: Well, yes and no. The description could perhaps be better. The purpose of it is to act as a replacement when the row accessor is no longer available. In that context returning false for `isAttached` makes sense IMO.

cmelchior:
:+1:

beeender:
:+1:

PR #1399, opened 2015-08-19
Reduce rowReferences HashMap autoboxing allocations.

diegomontoya:
HashMap value of rowReferences is never used but the HashMap.put() call is always using a Integer.MIN_VALUE as value arg which is small integer and causes autoboxing allocations to Integer().

Use static Integer object to remove autoboxing and Integer object allocations.


beeender:
:+1:  And thanks a lot!

cmelchior:
Looks great, thanks a lot. :+1: If CI is happy.

PR #1405, opened 2015-08-20
1) Remove unnecessary object modification in Context.finalize(). If c…

diegomontoya:
…ontext.finalize() is called by GC, the context is already out of scope. Existing code does modification to HashMap and ArrayLists which are only held by context and are going to be GCed anyways. 2) Realm calls Jni-GC aka executeDelayedDisposal on every RealmObject creation so this is a very hot spot. Optimize this by A) using optimized java range loop B) avoid a method call to ArrayList.Clear() 3) Declare a few vars as ArrayList. Doesn't hurt and can only help with performance even if by 0.000001 percent.

### DIFF
--- a/realm-jni/build.gradle
+++ b/realm-jni/build.gradle
@@ -147,7 +147,7 @@ targets.each { target ->
         dependsOn "generateNdkToolchain${target.toolchain.name.capitalize()}"
         environment PATH: "${buildDir}/standalone-toolchains/${target.toolchain.name}/bin:${System.env.PATH}"
         environment CC: "${target.toolchain.commandPrefix}-${clang?'clang':'gcc'}"
-        environment STRIP: "${target.toolchain.commandPrefix}-strip -o librealm-jni-${target.name}-stripped.so"
+        environment STRIP: "${target.toolchain.commandPrefix}-strip -o libearthworm-${target.name}-stripped.so"
         environment REALM_ANDROID: '1'
         commandLine = [
             'make',
@@ -158,22 +158,22 @@ targets.each { target ->
             "BASE_DENOM=${target.name}",
             "REALM_LDFLAGS=-lrealm-android-${target.name} -lstdc++ -lsupc++ -llog -L${projectDir}/../core-${project.coreVersion} -Wl,--gc-sections -flto",
             'LIB_SUFFIX_SHARED=.so',
-            "librealm-jni-${target.name}.so"
+            "libearthworm-${target.name}.so"
         ]
     }
 
     task "copyAndroidJni${target.name.capitalize()}"(dependsOn: "buildAndroidJni${target.name.capitalize()}") << {
         copy {
-            from "${projectDir}/src/librealm-jni-${target.name}-stripped.so"
+            from "${projectDir}/src/libearthworm-${target.name}-stripped.so"
             into "${projectDir}/../realm/src/main/jniLibs/${target.jniFolder}"
-            rename "librealm-jni-${target.name}-stripped.so", 'librealm-jni.so'
+            rename "libearthworm-${target.name}-stripped.so", 'libearthworm.so'
         }
 
         // Store the unstripped version
         copy {
-            from "${projectDir}/src/librealm-jni-${target.name}.so"
+            from "${projectDir}/src/libearthworm-${target.name}.so"
             into "${projectDir}/../build/output/jniLibs-unstripped/${target.jniFolder}"
-            rename "librealm-jni-${target.name}.so", 'librealm-jni.so'
+            rename "libearthworm-${target.name}.so", 'libearthworm.so'
         }
     }
 }
--- a/realm/build.gradle
+++ b/realm/build.gradle
@@ -4,12 +4,12 @@ apply plugin: 'maven-publish'
 apply plugin: 'com.jfrog.bintray'
 
 android {
-    compileSdkVersion 20
-    buildToolsVersion '20.0.0'
+    compileSdkVersion 22
+    buildToolsVersion '22.0.1'
 
     defaultConfig {
         minSdkVersion 9
-        targetSdkVersion 20
+        targetSdkVersion 22
     }
 
 // TODO: re-enable once bug-fix is released
@@ -239,7 +239,7 @@ task copyLibrariesToEclipseFolder(dependsOn: 'androidJar') << {
     }
     abi.each() { abiName ->
         copy {
-            from "build/intermediates/eclipse/lib/${abiName}/librealm-jni.so"
+            from "build/intermediates/eclipse/lib/${abiName}/libearthworm.so"
             into "../distribution/eclipse/${abiName}"
         }
     }
--- a/realm/build.gradle
+++ b/realm/build.gradle
@@ -10,7 +10,10 @@ android {
     defaultConfig {
         minSdkVersion 9
         targetSdkVersion 22
+        // Enabling multidex support.
+        multiDexEnabled true
     }
+
 
 // TODO: re-enable once bug-fix is released
 //    jacoco {
@@ -26,6 +29,8 @@ android {
 }
 
 dependencies {
+    compile 'com.android.support:multidex:1.0.0'
+    compile 'it.unimi.dsi:fastutil:7.0.7'
     compile 'com.intellij:annotations:+@jar'
     compile files("../realm-annotations/build/libs/realm-annotations-${version}.jar")
     androidTestApt files("../realm-annotations-processor/build/libs/realm-annotations-processor-${version}.jar")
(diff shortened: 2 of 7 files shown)


bmunkholm:
Thanks a lot for the contribution!
Did you validate any performance improvement with measurements?

In general, if the improvement is only 0.000001 % it is definitely not worth it. It actually doesn't come for free. The cost of improvements are the time we spend on reviewing/verifying/testing, which in then not spent on other features that adds more value to users. The other cost is risk of introducing bugs - which is why we spend the time to test :-)

That's just a general comment to any change and in particular performance improvements. It would be lovely if you could test the effect on performance, although it can be very very tricky to do right in Java!

diegomontoya:
This comment if for Part 2)  of my commit where every single RealmRow/Object allocation causes a manual JNI-GC into executeDelayedDisposal. Java method calls are not free so reducing the number of method call (when ArrayList.size() == 0, don't call ArrayList.Clear(), -1 method call) is worth the when it is called so many times. 

Benchmark results: (For cases when the 3 array lists are empty)

New is my commit, old is the original executeDelayedDisposal().

08-20 22:10:18.148    3339-3360/? D/executeDelayedNew﹕ 1625685
08-20 22:10:18.153    3339-3360/? D/executeDelayedOld﹕ 2398882
08-20 22:10:19.703    3392-3410/? D/executeDelayedNew﹕ 1704171
08-20 22:10:19.709    3392-3410/? D/executeDelayedOld﹕ 2412596
08-20 22:10:49.182    3450-3468/io.realm.test D/executeDelayedNew﹕ 1842271
08-20 22:10:49.198    3450-3468/io.realm.test D/executeDelayedOld﹕ 7074363
08-20 22:11:02.373    3500-3529/io.realm.test D/executeDelayedNew﹕ 1631586
08-20 22:11:02.379    3500-3529/io.realm.test D/executeDelayedOld﹕ 2481707
08-20 22:11:07.711    3603-3642/? D/executeDelayedNew﹕ 4448522
08-20 22:11:07.727    3603-3642/? D/executeDelayedOld﹕ 7326694
08-20 22:11:13.234    3699-3754/io.realm.test D/executeDelayedNew﹕ 1631959
08-20 22:11:13.240    3699-3754/io.realm.test D/executeDelayedOld﹕ 2369423
08-20 22:11:18.000    3801-3824/? D/executeDelayedNew﹕ 1746719
08-20 22:11:18.008    3801-3824/? D/executeDelayedOld﹕ 2307235
08-20 22:11:29.509    3863-3883/io.realm.test D/executeDelayedNew﹕ 4151791
08-20 22:11:29.524    3863-3883/io.realm.test D/executeDelayedOld﹕ 5272860
08-20 22:11:37.888    3914-3932/io.realm.test D/executeDelayedNew﹕ 5203016
08-20 22:11:37.900    3914-3932/io.realm.test D/executeDelayedOld﹕ 7052860
08-20 22:11:46.169    3966-3987/? D/executeDelayedNew﹕ 4926731
08-20 22:11:46.186    3966-3987/? D/executeDelayedOld﹕ 9636767
08-20 22:11:54.746    4029-4047/io.realm.test D/executeDelayedNew﹕ 1941791
08-20 22:11:54.755    4029-4047/io.realm.test D/executeDelayedOld﹕ 2766250

For the following I changed some var scopes to access the methods/vars directly to better reveal actual bench values instead of going through another proxy method.

Added code to below setup in JNITableTest.java.

    public void testBenchMethod1() {
        long start = System.nanoTime();

        for (int i = 0; i < 10000; i++) {
            t.context.executeDelayedDisposal();
        }

        long timeSpent = System.nanoTime() - start;

        Log.d("executeDelayedNew", String.valueOf(timeSpent));
    }

    public void testBenchMethod2() {
        long start = System.nanoTime();

        for (int i = 0; i < 10000; i++) {
            t.context.executeDelayedDisposalOld();
        }

        long timeSpent = System.nanoTime() - start;

        Log.d("executeDelayedOld", String.valueOf(timeSpent));
    }

diegomontoya:
@bmunkholm My 0.000001 comment was for part 3) of the commit which on second though does really nothing except make me feel better. =)

You guys can do some benchmarks regarding the part 1) which introduce a special JNI-GC path just for Finalize().

diegomontoya:
Btw. I see a failed test, though it passes all the tests on my end. What is the failed test?

cmelchior:
@diegomontoya Our CI server had some connection problems to our test device. It doesn't seem to a be a problem in your PR

realm/src/main/java/io/realm/internal/Context.java
```diff
@@ -44,25 +44,50 @@
 
     public void executeDelayedDisposal() {
         synchronized (this) {
-            for (int i = 0; i < abandonedTables.size(); i++) {
-                long nativePointer = abandonedTables.get(i);
+            if(abandonedTables.size() > 0) {
+                for (long nativePointer : abandonedTables) {
```
cmelchior: You should not use a `for...in` loop here as it will allocate a Iterator under the hood. It is safer to use:

```
int tableViewSize = abandonedTableViews.size()
 for (int i = 0; i < tableViewSize ; i++) {
  // ...
}
```

realm/src/main/java/io/realm/internal/Context.java
```diff
@@ -44,25 +44,50 @@
 
     public void executeDelayedDisposal() {
         synchronized (this) {
-            for (int i = 0; i < abandonedTables.size(); i++) {
-                long nativePointer = abandonedTables.get(i);
+            if(abandonedTables.size() > 0) {
```
cmelchior: Small style issue: Space between `if (`

realm/src/main/java/io/realm/internal/Context.java
```diff
@@ -44,25 +44,50 @@
...
+                for (long nativePointer : abandonedTables) {
+                    Table.nativeClose(nativePointer);
+                }
+
+                abandonedTables.clear();
+            }
+
+            if(abandonedTableViews.size() > 0) {
```
cmelchior: `if (`

realm/src/main/java/io/realm/internal/Context.java
```diff
@@ -44,25 +44,50 @@
...
+                for (long nativePointer : abandonedTableViews) {
```
cmelchior: Same...

realm/src/main/java/io/realm/internal/Context.java
```diff
@@ -44,25 +44,50 @@
...
+                    TableView.nativeClose(nativePointer);
+                }
+
+                abandonedTableViews.clear();
+            }
+
+            if(abandonedQueries.size() > 0) {
```
cmelchior: Same...

realm/src/main/java/io/realm/internal/Context.java
```diff
@@ -44,25 +44,50 @@
...
+                for (long nativePointer : abandonedQueries) {
```
cmelchior: Same...

realm-jni/build.gradle
```diff
@@ -147,7 +147,7 @@ targets.each { target ->
         dependsOn "generateNdkToolchain${target.toolchain.name.capitalize()}"
         environment PATH: "${buildDir}/standalone-toolchains/${target.toolchain.name}/bin:${System.env.PATH}"
         environment CC: "${target.toolchain.commandPrefix}-${clang?'clang':'gcc'}"
-        environment STRIP: "${target.toolchain.commandPrefix}-strip -o librealm-jni-${target.name}-stripped.so"
+        environment STRIP: "${target.toolchain.commandPrefix}-strip -o libearthworm-${target.name}-stripped.so"
```
kneth: Why do we have to change the name of the file?

realm/build.gradle
```diff
@@ -26,6 +29,8 @@ android {
 }
 
 dependencies {
+    compile 'com.android.support:multidex:1.0.0'
+    compile 'it.unimi.dsi:fastutil:7.0.7'
```
kneth: Is adding a dependency to an external library a good idea? @cmelchior @nhachicha do you have experiences with it?

diegomontoya:
@kneth Ignore thhose commits. Didn't know the changes after my pull commit is actually going to show up here. Assumed the pull request is "just this commit". The following commits in my fork are actually just my local test/changes.  I might have to fork another just to keep it clean. I will close this pull now since this is commit tree is pretty polluted.

realm/build.gradle
diegomontoya: Ooops. This was just my local test changes. Never meant them to show up here.

kneth:
@diegomontoya OK. At least I got to study `fastutil`which looks like an interesting library :-)

Issue #1406, opened 2015-08-20
Main looper not getting all change events


cmelchior:
It seems it is possible to cause a race condition in Realm which will cause the Main Looper to miss updates from background threads:

```
public class MainActivity extends AppCompatActivity {

    private RealmConfiguration config;
    private Realm realm;
    private RealmResults<OperationalOutputMessage> objects;

    @Override
    protected void onCreate(Bundle savedInstanceState) {
        super.onCreate(savedInstanceState);
        setContentView(R.layout.activity_main);
        config = new RealmConfiguration.Builder(this).build();
        Realm.deleteRealm(config);
        realm = Realm.getInstance(config);
        realm.addChangeListener(new RealmChangeListener() {
            @Override
            public void onChange() {
                printSize();

            }
        });
        objects = realm.allObjects(OperationalOutputMessage.class);
        printSize();

        // Execute lots of small commits on a background thread
        new Thread(new Runnable() {
            @Override
            public void run() {
                for (int i = 0; i < 1000; i++) {
                    OperationalOutputMessage msg = new OperationalOutputMessage();
                    msg.setmTimeStamp(i + 1);
                    try(Realm realm = Realm.getInstance(MainActivity.this)) {
                        realm.beginTransaction();
                        realm.copyToRealm(msg);
                        realm.commitTransaction();
                    }
                }
            }
        }).start();
    }

    private void printSize() {
        ((TextView)findViewById(R.id.text_size)).setText("Size: " + objects.size());
    }

    @Override
    protected void onDestroy() {
        super.onDestroy();
        realm.close();
    }
}
```

Expected output on the screen would be 1000 after the background thread is done. However that is not the case (usually around 600 on GenyMotion). This indicate that messages are not properly sent to the UI thread.

fuwaneko:
I don't know if it's related, but I write to realm in a background thread a model with many2one relation. After that I try to get that data in UI thread, it works but the relation is gone and becomes null.

kneth:
@fuwaneko Is it possible for you to share the code of the background thread?

fuwaneko:
@kneth 

[...]

PR #1407, opened 2015-08-20
Cleanup SharedGroup.java and Group.java

cmelchior:
Cleanup native methods to match earlier discussed style (=grouped at the bottom of the class).
Preparing for bugfix for #1406

### DIFF
--- a/realm/src/main/java/io/realm/internal/Group.java
+++ b/realm/src/main/java/io/realm/internal/Group.java
@@ -26,7 +26,12 @@ import java.nio.ByteBuffer;
  * of a collection of tables.
  */
 public class Group implements Closeable {
-    
+
+    // Below values must match the values in realm::group::OpenMode in C++
+    public static final int MODE_READONLY = 0; // Open in read-only mode. Fail if the file does not already exist.
+    public static final int MODE_READWRITE = 1; // Open in read/write mode. Create the file if it doesn't exist.
+    public static final int MODE_READWRITE_NOCREATE = 2; // Open in read/write mode. Fail if the file does not already exist.
+
     protected long nativePtr;
     protected boolean immutable;
     private final Context context;
@@ -34,10 +39,6 @@ public class Group implements Closeable {
     static {
         RealmCore.loadLibrary();
     }
-
-    //
-    // Group construction and destruction
-    //
 
     private void checkNativePtrNotZero() {
         if (this.nativePtr == 0)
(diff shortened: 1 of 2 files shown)


realm/src/main/java/io/realm/internal/Group.java
```diff
@@ -287,17 +253,27 @@ public String toJson() {
...
-
     private void throwImmutable() {
         throw new IllegalStateException("Objects cannot be changed outside a transaction; see beginTransaction() for details.");
     }
+
+    protected native long createNative();
+    protected native long createNative(String filepath, int value);
+    protected native long createNative(byte[] data);
+    protected native long createNative(ByteBuffer buffer);
+    protected static native void nativeClose(long nativeGroupPtr);
```
bmunkholm: looks strange with two static methods. They are all static actually. Not sure it changes anything in the method prototype.

bmunkholm:
:+1:

realm/src/main/java/io/realm/internal/Group.java
```diff
@@ -55,18 +51,16 @@ public Group() {
...
         private int value;
-        private OpenMode(int value) {
+        OpenMode(int value) {
             this.value = value;
         }
-    };
+    }
 
     public Group(String filepath, OpenMode mode) {
         this.immutable = mode.equals(OpenMode.READ_ONLY);
```
nhachicha: use `(mode  == OpenMode.READ_ONLY)`+ it's null safe

realm/src/main/java/io/realm/internal/Group.java
```diff
@@ -55,18 +51,16 @@ public Group() {
         checkNativePtrNotZero();
     }
 
-    protected native long createNative();
-
     public enum OpenMode {
         // Below values must match the values in realm::group::OpenMode in C++
         READ_ONLY(0),
```
nhachicha: you can get rid of `value` by using `ordinal()`

realm/src/main/java/io/realm/internal/Group.java
nhachicha: @bmunkholm the methods that needs `nativeGroupPtr` shouldn't be static (they use a private attribute of the instance)

realm/src/main/java/io/realm/internal/Group.java
cmelchior: `ordinal()` depends on the order, which means that if someone changed the order at some point the code would no longer be valid.

realm/src/main/java/io/realm/internal/Group.java
cmelchior: For some reason close is called from Context in static way, so unfortunately there is no way around static here. Otherwise I would agree with @nhachicha.

cmelchior:
Retest this please

cmelchior:
retest this please

cmelchior:
Ready for review @realm/java

beeender:
:+1:

realm/src/main/java/io/realm/internal/Group.java
```diff
@@ -55,29 +51,25 @@ public Group() {
         checkNativePtrNotZero();
     }
 
-    protected native long createNative();
-
     public enum OpenMode {
```
bmunkholm: I would just get rid of the enum for these internal classes and use int constants.

realm/src/main/java/io/realm/internal/Group.java
kneth: A `native` is not always static. According to the JNI specification, the second argument (the `jobject`) in the C/native implementation of the method depends on wether it is a static method or not. Source: http://docs.oracle.com/javase/1.5.0/docs/guide/jni/spec/design.html#wp715

kneth:
:+1:

PR #1408, opened 2015-08-20
Fixes change events getting lost in fast write scenarios

### LINKED ISSUE #1406: Main looper not getting all change events

cmelchior:
Fixes #1406 

It has not been possible to write a unit test that captures this problem, maybe due to how Handlers being less stressed in unit tests. The code presented in #1406 captures it every time though and it has been validated that this indeed fix the problem.

### DIFF
--- a/realm-jni/src/io_realm_internal_SharedGroup.cpp
+++ b/realm-jni/src/io_realm_internal_SharedGroup.cpp
@@ -295,3 +295,20 @@ JNIEXPORT jboolean JNICALL Java_io_realm_internal_SharedGroup_nativeCompact(
     CATCH_STD()
     return false;
 }
+
+JNIEXPORT jlongArray JNICALL Java_io_realm_internal_SharedGroup_nativeGetVersionID
+        (JNIEnv *env, jobject, jlong native_ptr)
+{
+    TR_ENTER()
+
+    SharedGroup::VersionID versionID = SG(native_ptr)->get_version_of_current_transaction();
+
+    jlong versionArray [2];
+    versionArray[0] = static_cast<jlong>(versionID.version);
+    versionArray[1] = static_cast<jlong>(versionID.index);
+
+    jlongArray versionData = env->NewLongArray(2);
+    env->SetLongArrayRegion(versionData, 0, 2, versionArray);
+
+    return versionData;
+}
--- a/realm-jni/src/io_realm_internal_SharedGroup.cpp
+++ b/realm-jni/src/io_realm_internal_SharedGroup.cpp
@@ -158,7 +158,7 @@ JNIEXPORT jlongArray JNICALL Java_io_realm_internal_SharedGroup_nativeAdvanceRea
         return Java_io_realm_internal_SharedGroup_nativeGetVersionID(env, obj, native_ptr);
     }
     CATCH_STD()
-    return 0;
+    return NULL;
 }
 
 JNIEXPORT void JNICALL Java_io_realm_internal_SharedGroup_nativePromoteToWrite
--- a/realm/src/main/java/io/realm/Realm.java
+++ b/realm/src/main/java/io/realm/Realm.java
@@ -280,8 +280,9 @@ public final class Realm implements Closeable {
         @Override
         public boolean handleMessage(Message message) {
             if (message.what == REALM_CHANGED) {
-                transaction.advanceRead();
-                sendNotifications();
+                if (transaction.advanceRead()) {
+                    sendNotifications();
+                }
             }
             return true;
         }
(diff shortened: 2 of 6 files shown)


realm/src/main/java/io/realm/internal/SharedGroup.java
```diff
@@ -81,34 +82,34 @@ public SharedGroup(String databaseFile, Durability durability, boolean fileMustE
...
-
-    private native long nativeCreateReplication(String databaseFile, byte[] key);
-
-    void advanceRead() {
+    /**
+     * Advances the transaction the to latest available data.
+     *
+     * @return {@code true} if version was advanced, {@code false} if already at the latest version.
+     */
+    boolean advanceRead() {
```
bmunkholm: if `nativeGetVersionID()` is always called after `nativeAdvanceRead()` and we currently don't expect it otherwise in the near future, I would suggest to let `nativeAdvanceRead()` return the VersionID instead to save a JNI call.

cmelchior:
Ready for review

realm-jni/src/io_realm_internal_SharedGroup.cpp
```diff
@@ -149,14 +149,16 @@ JNIEXPORT jlong JNICALL Java_io_realm_internal_SharedGroup_nativeBeginImplicit
...
+JNIEXPORT jlongArray JNICALL Java_io_realm_internal_SharedGroup_nativeAdvanceRead
+  (JNIEnv *env, jobject obj, jlong native_ptr)
 {
     TR_ENTER_PTR(native_ptr)
     try {
         LangBindHelper::advance_read( *SG(native_ptr) );
+        return Java_io_realm_internal_SharedGroup_nativeGetVersionID(env, obj, native_ptr);
     }
     CATCH_STD()
+    return 0;
```
kneth: Is `null` more appropriate for an array?

realm/src/main/java/io/realm/Realm.java
```diff
@@ -280,8 +280,9 @@ private void removeHandler(Handler handler) {
         @Override
         public boolean handleMessage(Message message) {
             if (message.what == REALM_CHANGED) {
-                transaction.advanceRead();
-                sendNotifications();
+                if (transaction.advanceRead()) {
```
kneth: With my comment above, you should check for `NULL`.

realm/src/main/java/io/realm/internal/SharedGroup.java
```diff
@@ -81,8 +82,19 @@ public SharedGroup(String databaseFile, Durability durability, boolean fileMustE
...
 
-    void advanceRead() {
-        nativeAdvanceRead(nativePtr);
+    /**
+     * Advances the transaction the to latest available data.
+     *
+     * @return {@code true} if version was advanced, {@code false} if already at the latest version.
+     */
+    boolean advanceRead() {
+        long[] nextVersion = nativeAdvanceRead(nativePtr);
```
kneth: What if `nativeAdvanceRead` returns 0?

realm-jni/src/io_realm_internal_SharedGroup.cpp
kneth: Isn't `null` more appropriate?

emanuelez:
Does this mean that if we do 1000 commits in a BG thread the UI thread will receive 1000 notifications? Isn't that a bit overkill? IMO one per loop iteration is fine.

cmelchior:
Yes. The background thread will send 1000 notifications. Before this fix we checked the queue before sending, but that resulted in the UI thread loosing updates, which would resultat in versions getting out of sync. I linked to some testcode that will showcase the problem.

cmelchior:
Actually, What we might do is clear the handler from the receiving Looper thread. I will try that

cmelchior:
After some more digging. It turns out we are not loosing updates and the Looper/Handler works as expected. My listener in the test case was not saved in a field so because it is only saved in a WeakReference inside Realm it gets GC'ed :( .... want custom lint checks

PR #1410, opened 2015-08-21
Reintroducing lenient UTF8 -> UTF16 transcoding 

kneth:
This has been merged before but was removed when we saw many native crashes which were believe to be related to strings back then. The crashes often came from developers ignoring Out-of-Memory, which lead to a Realm in an inconsistent state. In order to track improvements of Realm's underlaying storage engine, I wish to go back to use the methods providing by the storage engine instead of the under-maintained version here.

This is based on the work done in #647.

@realm/java @kspangsege 

TODO:
- [x] Remove `string_to_hex()` functions


kspangsege:
Great move :-)

:+1:

cmelchior:
retest this please

realm-jni/src/util.cpp
```diff
@@ -235,6 +236,19 @@ struct JStringCharsAccessor {
...
+    ret << message << " ";
+    ret << "StringData.size = " << str.size() << "; ";
+    ret << "StringData.data = " << str.data() << "; ";
+    ret << "StringData as hex = ";
+    for (string::size_type i = 0; i < str.size(); ++i)
+        ret << " 0x" << std::hex << std::setfill('0') << std::setw(2) << (int)s[i];
+    return ret.str();
+}
+
 string string_to_hex(const string& message, StringData& str, const char* in_begin, const char* in_end,
```
bmunkholm: Is this (old) string_to_hex(....error_code) used anymore?

realm-jni/src/util.cpp
```diff
@@ -256,6 +270,16 @@ string string_to_hex(const string& message, StringData& str, const char* in_begi
...
+    ostringstream ret;
+
+    ret << message << "; ";
+    for (size_t i = 0; i < size; ++i)
+        ret << " 0x" << std::hex << std::setfill('0') << std::setw(4) << (int)str[i];
+    return ret.str();
+}
+
+
 string string_to_hex(const string& message, const jchar *str, size_t size, size_t error_code) {
```
bmunkholm: similar?

bmunkholm:
How about testcases to justify this has any effect?
Any performance difference?

realm-jni/src/util.cpp
kneth: No, the old version isn't used anymore (please see TODO list in description).

bmunkholm: I'll read that next time :-)

kneth:
@bmunkholm A quick benchmark indicates that the difference is below the timer's resolution.

realm-jni/src/util.cpp
```diff
@@ -271,63 +250,73 @@ string concat_stringdata(const char *message, StringData strData)
...
+    jchar *out_begin = out;
+    jchar *out_end = out_begin + stack_buf_size;
+
+    for (; ;) {
+        typedef Utf8x16 <jchar, JcharTraits> Xcode;
+        Xcode::to_utf16(in, in_end, out, out_end);
+        bool end_of_input = in == in_end;
+        if (end_of_input)
+            break;
+        bool bad_input = out != out_end;
```
zaki50: `out != out_end` does not always mean bad input.

In `Xcode::to_utf16(in, in_end, out, out_end);` above, if last processed UTF-8 character consists of four bytes sequence and rest space of `out` is 1, that four bytes sequence is translated into two jchars(surrogate pair) and those jchars can't be stored into `out`.
In that case, we should not skip input bytes.

To detect that case, we should check return value of `Xcode::to_utf16(in, in_end, out, out_end);`

cmelchior: Hi @zaki50 
We are currently all in CPH for a company get-together, so response on this might be a bit slow, but i'll poke @kneth :)

zaki50: @cmelchior Thank you for your response. I heard about get-together from @kishikawakatsumi and saw many beautiful pictures of CPH on facebook.

I'm not in a hurry.

Anyway, thank you!

kneth: @zaki50 @cmelchior I haven't had time to look closer. But I will poke @kspangsege as he wrote the original version :-)

zaki50: @kspangsege @cmelchior @kneth I wrote test case for this bug which is intended to be located in JNIRowTest.java.

This test case fails at kg-revert-to-core-utf-handling branch.

```
public void testSurrogatePairAtTheEndOfBuffer() {
    Table table = new Table();
    table.addColumn(ColumnType.STRING, "string");

    final StringBuilder sb = new StringBuilder();
    // 48 is value of 'stack_buf_size' in to_jstring(JNIEnv*,StringData) in util.cpp
    for (int i = 0; i < (48 - 1); i++) {
        sb.append((char) ('0' + (i % 10)));
    }
    final String nonBmpCharMOYAI = "\uD83D\uDDFF"; // MOYAI(U+1F5FF)
    sb.append(nonBmpCharMOYAI);
    final String originalString = sb.toString();

    table.add(originalString);
    final UncheckedRow row = table.getUncheckedRow(0);

    assertEquals(originalString, row.getString(0));
}
```

zaki50: if the value in `for (int i = 0; i < (48 - 1); i++) {` is other than 48, this test does not fail.

kspangsege: > `out != out_end` does not always mean bad input.

@zaki50, you are right. That is a bug.

I am actually working on an improved version of `Xcode::to_utf16()`; one that lets you whether it stopped because of lack of output buffer space. That will make it easier to fix this bug. It will probably be another week before I get time to finish it though.

zaki50: @kspangsege Thank you for confirming that.

You might have noticed that to_utf16() in utf8.hpp has a bug whose return value is different from what the comment says.
The comment says "true if all input transcoded or ...
                                 false if transcoding stopped due to invalidinput."
In fact, to_utf16() returns the opposite.
It seems that last line of to_utif16() shold be "return !invalid;"

cmelchior: @kspangsege Have you had time to look further into this?

diegomontoya:
Any update on this PR? We still have reoccurring short string utf8 to utf16 crashes and with this pr, we will at least know/distinguish if the the entire row is corrupted or if just the one specific row string field contains invalid utf8. 

https://github.com/realm/realm-java/issues/1811
https://github.com/realm/realm-java/issues/2492

kneth:
@diegomontoya This PR will probably get rid of those error. That is, you will not see them as a Realm user but the underlying issues are still here. That means, it will mask a potential bug in either Realm Java or Realm Core. Long story short, we stopped working on this PR as we prefer to see crashes rather than silently corrupt files.

bmunkholm:
@diegomontoya So if you can help us with a reproducible case we love to fix it if it's a bug in Realm.

diegomontoya:
@kneth @bmunkholm We will try to provide a corrupted db for inspection hopefully by next week. Thankfully we always have 1 or 2 crashes per day on this front so we just need to find a way to get those remote dbs to us.

cmelchior:
Closing this PR. We really want to catch these kinds of errors in the future as well. 
Also @kneth asked me to summarize with "I told you so" 😄

PR #1415, opened 2015-08-24
Test case fix

beeender:
@realm/java Strange, why didn't it crash on CI before?

### DIFF
--- a/realm/src/androidTest/java/io/realm/RealmTest.java
+++ b/realm/src/androidTest/java/io/realm/RealmTest.java
@@ -16,12 +16,9 @@
 package io.realm;
 
 import android.content.Context;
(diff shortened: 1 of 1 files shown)


cmelchior:
retest this please

cmelchior:
I am more worried, that it is actually possible to call `commit` twice in a row without the binding complaining about it. This PR is :+1:, but we should create an issue for it as it should throw an exception IMO.

emanuelez:
:+1:

beeender:
#1417 Created for tracking that issue. @cmelchior

kneth:
One could argue that committing a read transaction is equivalent to do a refresh.

PR #1424, opened 2015-08-26
[README][LICENSE] added Crimea to list of prohibited regions

jpsim:
Same as https://github.com/realm/realm-cocoa/pull/2440./cc @timanglade


timanglade:
lgtm! Thanks @jpsim

PR #1425, opened 2015-08-26
Fixes Process listeners after refresh()

### LINKED ISSUE #1401: Realm.addChangeListener doesn't work on background threads
If you do 

```
new Thread(new Runnable() {
  public void run() {
    realm.addChangeListener(listener)
    // ...
  }

}).start();
```

on a non-looper thread. That change listener will never be called unless you execute a transaction on that thread.

That behaviour seems a bit strange. The problem is that we are triggering onChange after updating the the Realm on looper-threads. That isn't possible on non-looper threads as there is no point in time updating the Realm behind users back would make sense.

We should consider doing one of the following:

1) Prevent people from adding change listeners on a non-looper thread (throw exception). People would have to notify these threads in some other way.

2) Document the behaviour better. 

3) Call onChange on non-looper threads, but without updating the Realm on that thread first.

4) Something else.... ?

emanuelez:
Fixes #1401

### DIFF
--- a/realm/src/androidTest/java/io/realm/RealmTest.java
+++ b/realm/src/androidTest/java/io/realm/RealmTest.java
@@ -1828,7 +1828,7 @@ public class RealmTest extends AndroidTestCase {
         final AtomicBoolean listenerWasCalled = new AtomicBoolean(false);
 
         // Used by the background thread to wait for the main thread to do the write operation
-        final CountDownLatch latch = new CountDownLatch(1);
+        final CountDownLatch bgThreadLatch = new CountDownLatch(1);
 
         Thread backgroundThread = new Thread() {
             @Override
@@ -1841,7 +1841,7 @@ public class RealmTest extends AndroidTestCase {
                     }
                 });
                 try {
-                    latch.await(); // Wait for the main thread to do a write operation
+                    bgThreadLatch.await(); // Wait for the main thread to do a write operation
                     bgRealm.refresh(); // This should call the listener
                     assertTrue(listenerWasCalled.get());
                 } catch (InterruptedException e) {
@@ -1856,6 +1856,6 @@ public class RealmTest extends AndroidTestCase {
         testRealm.beginTransaction();
         testRealm.createObject(Dog.class);
         testRealm.commitTransaction();
-        latch.countDown();
+        bgThreadLatch.countDown();
     }
 }
--- a/realm/src/androidTest/java/io/realm/RealmTest.java
+++ b/realm/src/androidTest/java/io/realm/RealmTest.java
@@ -27,6 +27,7 @@ import org.json.JSONObject;
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
+import java.lang.Thread;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Date;
@@ -39,6 +40,7 @@ import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 import io.realm.dynamic.DynamicRealmObject;
 import io.realm.entities.AllTypes;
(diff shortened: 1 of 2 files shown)


emanuelez:
@realm/java

realm/src/androidTest/java/io/realm/RealmTest.java
```diff
@@ -1818,4 +1822,40 @@ public void testDeletingRealmAlsoClearsConfigurationCache() throws IOException {
         testRealm = Realm.getInstance(newConfig);
         assertNotNull(testRealm);
     }
+
+    public void testProcessLocalListenersAfterRefresh() {
+        // Used to validate the result
+        final AtomicBoolean listenerWasCalled = new AtomicBoolean(false);
+
+        // Used by the background thread to wait for the main thread to do the write operation
+        final CountDownLatch latch = new CountDownLatch(1);
```
cmelchior: Perhaps call this `bgThreadReady` or something. `latch` is a bit too generic imo.

realm/src/androidTest/java/io/realm/RealmTest.java
```diff
@@ -1818,4 +1822,40 @@ public void testDeletingRealmAlsoClearsConfigurationCache() throws IOException {
...
+                    bgRealm.close();
+                }
+            }
+        };
+        backgroundThread.start();
+
+        testRealm.beginTransaction();
+        testRealm.createObject(Dog.class);
+        testRealm.commitTransaction();
+        latch.countDown();
```
cmelchior: Does this actually work? Doesn't the test just complete now, then ignoring any failures from running background threads?

realm/src/main/java/io/realm/Realm.java
```diff
@@ -1276,6 +1276,7 @@ boolean hasChanged() {
     public void refresh() {
         checkIfValid();
         transaction.advanceRead();
+        sendNotifications();
```
beeender: Then every thread will be notified? Which means the thread with looper will be notified more than once with nothing changed.

cmelchior:
Also. It would probably be nice to expand the JavaDoc to document the slightly different behaviour for listeners in `addChangeListener` and `refresh`

realm/src/androidTest/java/io/realm/RealmTest.java
emanuelez: It does work, in fact the test fails without the change in `Realm.java`

realm/src/main/java/io/realm/Realm.java
emanuelez: no, sendNotifications operates on the thread local list of changeListeners

realm/src/androidTest/java/io/realm/RealmTest.java
emanuelez: it's a latch, not a boolean. How about `bgThreadLatch`?

realm/src/androidTest/java/io/realm/RealmTest.java
cmelchior: :+1:

realm/src/androidTest/java/io/realm/RealmTest.java
cmelchior: :+1:

realm/src/main/java/io/realm/Realm.java
beeender: I see...
Is it possible to check the version then call `sendNotification`? I have seen @cmelchior made a change to return the version by `advanceRead()`

emanuelez: Right now `advanceRead()` returns `void`, but that would be a nice addition once available!

cmelchior: As requested :) https://github.com/realm/realm-java/pull/1408

beeender:
:+1: 
Just a reminder, need one more test case (refresh doesn't trigger when nothing changed.) when #1408 merged.

emanuelez:
retest this please

emanuelez:
Please ignore the failing dco-signed step... it's my experiments with Leeroy :)

emanuelez:
Changelog:
 * Fixed an issue preventing the call of listeners on refresh()

emanuelez:
Bump @realm/java

beeender:
:+1: again

kneth:
:+1: But I would like to see #1408 merged first.

cmelchior:
#1408 is not correct. After some digging, i discovered that the reason my test case fail is because I just use an anonymous change listener instance, which then gets GC'ed at a random point in time because we use WeakReferences....I hate WeakReferences :(

:+1:

emanuelez:
I assume this is ok to merge then, right @cmelchior?

cmelchior:
Yup :+1:

PR #1434, opened 2015-08-28
Schemas are now only validated once across threads.

### LINKED ISSUE #1373: Validating Realm schema should only be done once
**Problem**:
1) Background thread opens Realm and begins a long living write transaction.
2) UI thread calls `Realm.getInstance()` for the first time.
3) Result: UI thread is locked until background thread completes because `initializeRealm` requires a write transaction.

**Reason**:
Our Realm cache is thread local and doesn't take into account that a Realm on another thread might have validated the schema already.

**Solution**:
We should cache the result of `mediator.validateTable(modelClass, realm.sharedGroup.getTransaction());`. Doing so will not only prevent unnecessary locking of the UI thread, but it will also speed up opening Realms considerable when doing so for the first time on new threads.

cmelchior:
Fixes #1373 

Prevents a long living write transaction on a background thread blocking opening Realm on the UI thread if it is the same file, as we then are sure that the schema has already been validated.

@realm/java

### DIFF
--- a/realm/src/androidTest/java/io/realm/RealmTest.java
+++ b/realm/src/androidTest/java/io/realm/RealmTest.java
@@ -1823,6 +1823,8 @@ public class RealmTest extends AndroidTestCase {
 
     public void testValidateSchemasOverThreads() throws InterruptedException, TimeoutException, ExecutionException {
         final RealmConfiguration config = TestHelper.createConfiguration(getContext(), "foo");
+        Realm.deleteRealm(config);
+
         final CountDownLatch bgThreadLocked = new CountDownLatch(1);
         final CountDownLatch mainThreadDone = new CountDownLatch(1);
 
@@ -1852,6 +1854,6 @@ public class RealmTest extends AndroidTestCase {
         });
 
         bgThreadLocked.await(2, TimeUnit.SECONDS);
-        assertTrue(future.get(2, TimeUnit.SECONDS));
+        assertTrue(future.get(10, TimeUnit.SECONDS));
     }
 }
--- a/realm/src/androidTest/java/io/realm/RealmTest.java
+++ b/realm/src/androidTest/java/io/realm/RealmTest.java
@@ -1821,6 +1821,7 @@ public class RealmTest extends AndroidTestCase {
         assertNotNull(testRealm);
     }
 
+    // Tests that if the same Realm file is opened on multiple threads, we only need to validate the schema on the first thread.
     public void testValidateSchemasOverThreads() throws InterruptedException, TimeoutException, ExecutionException {
         final RealmConfiguration config = TestHelper.createConfiguration(getContext(), "foo");
         Realm.deleteRealm(config);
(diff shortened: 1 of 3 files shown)


realm/src/androidTest/java/io/realm/RealmTest.java
```diff
@@ -27,6 +27,7 @@
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
+import java.sql.Time;
```
emanuelez: Are you sure you are importing the right class here?

realm/src/main/java/io/realm/Realm.java
```diff
@@ -545,6 +551,7 @@ private static void initializeRealm(Realm realm) {
                 mediator.validateTable(modelClass, realm.transaction);
                 realm.columnIndices.addClass(modelClass, mediator.getColumnIndices(modelClass));
             }
+            validatedRealmFiles.put(realm.getPath(), true);
```
emanuelez: Use `Boolean.TRUE` so you don't have to rely on autoboxing

realm/src/main/java/io/realm/Realm.java
```diff
@@ -128,6 +128,9 @@
         }
     };
 
+    // List of Realm files that has already been validated
+    private static final Map<String, Boolean> validatedRealmFiles = new ConcurrentHashMap<String, Boolean>();
```
emanuelez: How about using `Collections.newSetFromMap`? http://docs.oracle.com/javase/6/docs/api/java/util/Collections.html#newSetFromMap%28java.util.Map%29

beeender:
Any good idea to handle the multi-process case?

cmelchior:
This should not really effect multi-process (nor solve it). I would much rather take that discussion in #1300. This does however prevent the number of times we do require a write transaction which will also help with multi-process support.

emanuelez:
:+1:

realm/src/androidTest/java/io/realm/RealmTest.java
```diff
@@ -1818,4 +1820,40 @@ public void testDeletingRealmAlsoClearsConfigurationCache() throws IOException {
         testRealm = Realm.getInstance(newConfig);
         assertNotNull(testRealm);
     }
+
+    public void testValidateSchemasOverThreads() throws InterruptedException, TimeoutException, ExecutionException {
```
bmunkholm: Please add a oneliner comment that explains what the test does.

realm/src/main/java/io/realm/Realm.java
```diff
@@ -1477,6 +1487,7 @@ public static void migrateRealm(RealmConfiguration configuration, RealmMigration
             realm.beginTransaction();
             realm.setVersion(realmMigration.execute(realm, realm.getVersion()));
             realm.commitTransaction();
+            validatedRealmFiles.remove(configuration.getPath());
```
kneth: `validatedRealmFiles.add()` is done within a write transaction. Should `remove()` also be done within a write transaction?

cmelchior: While there is a race condition in close that could cause this to be called twice, that doesn't really matter as calling remove on something that isn't there would be a no-op. Also the backing set is Thread safe, so no there shouldn't be any need for a write transaction.

cmelchior:
Updated with feedback from @nhachicha

beeender:
:+1:

Issue #896, opened 2015-02-23
Async queries + write  transactions


bmunkholm:
Realm should offer a API akin to this, where queries are run on a background thread and then transferred afterwards.

```
realm.where(Foo.class).findAllAsync(new RealmQuery.Callback() {
  public void onSuccess(RealmResults<Foo> results) {
    // Use results
 }
});

realm.where(Foo.class).findFirstAsync(new RealmQuery.Callback() {
  public void onSuccess(Foo result) {
    // Use result
 }
});
```

Topics for discussion
- How to handle version differences between threads?
- Should we expose a error callback as well?
- How should we handle that the UI is no longer available (i.e. don't be AsyncTask)

kneth:
Core issue: https://github.com/realm/realm-core/issues/725

Ruxton:
+100!

lucamtudor:
YES!

kevincwuu:
please please implement this...

nhachicha:
https://github.com/realm/realm-java/pull/1214

PR #1443, opened 2015-09-01
Replace constants with enums

cmelchior:
This PR replaces the `RealmQuery.CASE_*` and `ReamResults.SORT_ORDER_*` constants with enum classes instead. Names are kept deliberate short.

Question: Is naming OK?

`SortOrder` is already used by the Java framework and  might be more appropriate name, although a bit longer? (http://docs.oracle.com/javase/7/docs/api/javax/swing/SortOrder.html)

Couldn't find any examples of Case. Closest is probably `FieldNamingPolicy` from GSON (https://google-gson.googlecode.com/svn/trunk/gson/docs/javadocs/com/google/gson/FieldNamingPolicy.html) which however is something slightly different.  `Pattern` also has a flag called  CASE_INSENSITIVE (http://docs.oracle.com/javase/7/docs/api/java/util/regex/Pattern.html)

```
Case.INSENSITIVE
Case.SENSITIVE
Sort.ASCENDING
Sort.DESCENDING
```


cmelchior:
Ready for review @realm/java

emanuelez:
Why the new query package?

bmunkholm:
@cm I'm sure you noticed that Pattern uses an int :-)

cmelchior:
@emanuelez To keep `io.realm` as clean as possible. I kinda like that it only contain the high level classes that define the API (RealmBaseAdapter is moving to `io.realm.android`)

kneth:
I like the changes as it will make our API safer to use.

beeender:
One question, what is the elegant way to handle multiple options? for example:
```
sort(stringField, SORT_ORDER_ASCENDING | SORT_ORDER_NULL_FIRST);
```

cmelchior:
Enums won't do that nicely no. So if we think that the above is a use case we probably have to rethink enums.

Do you have a use case where the order will matter? Personally I don't think the order will matter in a Object Store abstraction model, so it would just be something we configure behind the scenes.

cmelchior:
Also the above approach would really only work with bit-flags, which are really hard to make easy to use across e.g. a DynamicRealmQuery and a RealmQuery as you wouldn't have one logical class to put the flags in. That was what prompted the enums in the first place, moving them to a single location that can be reused across classes.

beeender:
@cmelchior Check this issue in core https://github.com/realm/realm-core/issues/1078

Think about the `RealmBaseAdapter`, user might need this to nicely put null values before everything or after. If we don't give such an option, it requires some extra work from user.

beeender:
If there are only those two options, i guess something like `SORT_ORDER_ASCENDING_NULL_FIRST` and `SORT_ORDER_ASCENDING_NULL_LAST` would just solve the issue.

cmelchior:
Fair enough. I can see the use case (how big it is up for discussion though :) ). 

I don't really like combinatorial enums as it would require 4 enums. It would be overly long, it doesn't scale and most won't care. Also did you see the comment from @simonask about the order mattering more in a Relational model (which we are not).

One other solution could be varargs, as #896 will most likely inline the `sort` method anyway. So it would be something like:

```
realm.query(Foo.class).equalTo("name", "foo").sort("name", Sort.DESCENDING, Sort.NULL_FIRST).findAll();
```

That would also be something we could add later if a need arises.

cmelchior:
For what it is worth, SQLite doesn't allow changing this either.

changelog.txt
```diff
@@ -1,4 +1,8 @@
 0.83
+ * BREAKING CHANGE: RealmResults.SORT_ORDER_ASCENDING and RealmResults.SORT_ORDER_DESCENDING constants
+   has been replaced by Sort.ASCENDING and Sort.DESCENDING enums.
```
beeender: We used to not having line wraps in changelog. But be honest, i'd like to have it when line exceed some chars long (90 maybe?)

beeender:
:+1:

bmunkholm:
Did you consider the alternative using @IntDef? (http://tools.android.com/tech-docs/support-annotations)

cmelchior:
Yes, but it only work on Android, it doesn't work in normal Java. 
There are a lot of cool annotations in the support library, like null checks and thread annotations, but if we start using them it becomes much harder to make a Java only variant of Realm.

cmelchior:
Before merging, these classes should be moved back to `io.realm`

cmelchior:
Classes moved to `io.realm` pr. agreement with @emanuelez, but apart from that everything should be the same so just need a couple of thumbs up to merge @realm/java

beeender:
:+1: again

realm/src/main/java/io/realm/internal/TableView.java
```diff
@@ -867,18 +866,18 @@ public void sort(long columnIndex) {
         nativeSort(nativePtr, columnIndex, true);
     }
 
-    protected native void nativeSort(long nativeTableViewPtr, long columnIndex, boolean ascending);
+    protected native void nativeSort(long nativeTableViewPtr, long columnIndex, boolean sortOrder);
 
-    public void sort(List<Long> columnIndices, List<Order> order) {
+    public void sort(List<Long> columnIndices, Sort[] sortOrders) {
```
emanuelez: Have you considered an EnumSet instead of an array?

cmelchior: It has to be a list as each `SortOrder` correspond to the same entry in `columnIndices` (for multi-field sorting), so a Set wouldn't really work in this case.

realm/src/main/java/io/realm/Case.java
```diff
@@ -0,0 +1,28 @@
+package io.realm;
```
kneth: Copyright notice is missing

realm/src/main/java/io/realm/Sort.java
```diff
@@ -0,0 +1,26 @@
+package io.realm;
```
kneth: Copyright notice is missing

realm/src/main/java/io/realm/Realm.java
```diff
@@ -1128,17 +1127,17 @@ boolean contains(Class<? extends RealmObject> clazz) {
...
+     * @param sortOrder2 sort order for second field.
      * @return A sorted RealmResults containing the objects.
      * @throws java.lang.IllegalArgumentException if a field name does not exist.
      */
     public <E extends RealmObject> RealmResults<E> allObjectsSorted(Class<E> clazz, String fieldName1,
-                                                               boolean sortAscending1, String fieldName2,
-                                                                    boolean sortAscending2) {
-        return allObjectsSorted(clazz, new String[]{fieldName1, fieldName2}, new boolean[]{sortAscending1,
-                sortAscending2});
+                                                               Sort sortOrder1, String fieldName2,
```
bmunkholm: indenting bad

realm/src/main/java/io/realm/Case.java
cmelchior: Added

realm/src/main/java/io/realm/Sort.java
cmelchior: Added

kneth:
:+1:

cmelchior:
All comments fixed. Closing this here, so it can be merged against the "feature-dynamic-migration-api" branch instead

Issue #1394, opened 2015-08-17
missing field type getter of DynamicRealmObject


zaki50:
In order to access model object by generic way, we have to know the type of each field.

I'd like to write as follows.
```
DynamicRealmObject obj = ...;
for (String fieldName : obj.getFieldNames()) {
    switch(obj.getFieldType(fieldName)) { // getFieldType(String) is missing
        case BOOLEAN:
            boolean b = obj.getBoolean(fieldName);
            // do something
            break;
        case STRING:
            String str = obj.getString(fieldName);
            // do something
            break;
        ...
    }
}
```

cmelchior:
Yes, that is a problem. It does however mean we need to expose the underlying Realm types in the binding which was something we hoped to avoid.

We have been discussing adding a generic `get()` method as well:

```
Object value = obj.get(field);
```

Would this also solve your use case?

cmelchior:
In any case it is a use case we should have a solution for when #1200 is implemented

zaki50:
generic get() method is acceptable for my use case.

I'm writing a code to export all models into CSV file.

cmelchior:
I am merging this issue into #1369. Closing this now.

PR #1444, opened 2015-09-02
Renamed ColumnType to RealmFieldType and moved to io.realm.dynamic.

cmelchior:
It order to fix https://github.com/realm/realm-java/issues/1394 as part of the Dynamic API refactoring `ColumnType` has been renamed `RealmType` and moved from `io.realm.internal` to `io.realm.dynamic`

While ideally I would avoid exposing the underlying Realm types, with a dynamic API that becomes impossible. So instead I have beautified `ColumnType` and pulled it into the public API.

Small change/big impact as apparently we reference ColumnType all over the place.


bmunkholm:
Align with Cocoa team on naming things like this.
"RealmType" seems quite broad. RealmDataType or RealmFieldType) seems more distinct, although longer. Could maybe skip the "Realm" part so it becomes Realm.FieldType?

cmelchior:
Ready for review @realm/java 

Any other ideas for `RealmType` would also be welcome (as long as the it doesn't contain `Column` as that concept is not visible in the public API)

realm/src/main/java/io/realm/dynamic/RealmType.java
```diff
@@ -0,0 +1,117 @@
...
+ */
+
+package io.realm.dynamic;
+
+import java.nio.ByteBuffer;
+
+import io.realm.internal.Mixed;
+
+/**
+ * List of the types used by Realms underlying storage engine.
```
kneth: "Realms" -> "Realm's"

kneth:
I appreciate the idea behind the change. The proposal is `io.realm.dynamic.RealmType.INTEGER`, and I also find two times `realm` one too many. `io.realm.dynamic.FieldType.INTEGER` is more attractive to me.

cmelchior:
Note that you will normally never see the whole package name in code as it would be automatically imported, so:

```
new RealmMigration() {
    DynamicRealmObject foo = realm.allObjects("Foo").findFirst();
    FieldType type = foo.getFieldType("name");
 
    // Instead of
    RealmFieldType type = foo.getFieldType("name");

    // or
    io.realm.dynamic.FieldType = foo.getFieldType("name");
  }
}
```

I very much like shorter names, but I am not sure it will be entirely consistent with names we are currently using and that will pop up in migrations:

```
DynamicRealmObject
RealmSchema
RealmObjectSchema
```

But perhaps the context is so clear that we only use the Realm prefix on top level classes (`Realm`, `RealmResults`, `RealmList`, `RealmObject`, `RealmQuery`) and try to keep it away from more level level classes. This would e.g. imply that `RealmSchema` -> `Schema` (which is almost too short) and `RealmObjectSchema` -> `ObjectSchema` (which is better).

But more opinions on the matter would be great.

kneth:
Yes, `FieldType` might be confusing as `Field` and `Type` are used in Java for reflection.

bmunkholm:
I feel quite strongly that "RealmType" is not acceptable. Read it like "which Type of Realm" and it get another meaning. A "Realm" is a file/database. The "Type" of "Realm" has another meaning than the "Field/Data" type.

realm/src/androidTest/java/io/realm/internal/test/TestHelper.java
```diff
@@ -20,24 +20,24 @@
      * @param o
      * @return
      */
-    public static ColumnType getColumnType(Object o){
+    public static RealmFieldType getColumnType(Object o){
```
bmunkholm: getColumnType -> getFieldType

realm/src/main/java/io/realm/RealmQuery.java
```diff
@@ -131,7 +131,7 @@ private boolean containsDot(String s) {
     }
 
     // TODO: consider another caching strategy so linked classes are included in the cache.
-    private long[] getColumnIndices(String fieldName, ColumnType fieldType) {
+    private long[] getColumnIndices(String fieldName, RealmFieldType fieldType) {
```
bmunkholm: getColumnIndices -> getFieldIndices

realm/src/main/java/io/realm/dynamic/RealmFieldType.java
```diff
@@ -0,0 +1,117 @@
...
+ * types through a {@link DynamicRealmObject}.
+ */
+public enum RealmFieldType {
+    // Make sure numbers match with <realm/column_type.hpp>
+    INTEGER(0),
+    BOOLEAN(1),
+    STRING(2),
+    BINARY(4),
+    TABLE(5),
+    MIXED(6),
```
bmunkholm: Maybe don't expose this - or rename it to UNSUPPORTED_MIXED

realm/src/main/java/io/realm/dynamic/RealmFieldType.java
```diff
@@ -0,0 +1,117 @@
...
+ * convert between normal Java types and the Realm types. However it is possible to access these
+ * types through a {@link DynamicRealmObject}.
+ */
+public enum RealmFieldType {
+    // Make sure numbers match with <realm/column_type.hpp>
+    INTEGER(0),
+    BOOLEAN(1),
+    STRING(2),
+    BINARY(4),
+    TABLE(5),
```
bmunkholm: Maybe don't expose this - or rename it to UNSUPPORTED_TABLE

realm/src/main/java/io/realm/dynamic/RealmFieldType.java
```diff
@@ -0,0 +1,117 @@
...
+    FLOAT(9),
+    DOUBLE(10),
+    LINK(12),
+    LINK_LIST(13),
+    BACKLINK(14);
+
+    // Primitive array for fast mapping between between native values and their Realm type.
+    private static RealmFieldType[] typeList = new RealmFieldType[15];
+    static {
+        RealmFieldType[] columnTypes = values();
```
bmunkholm: columnTypes -> fieldTypes

realm/src/main/java/io/realm/internal/Mixed.java
```diff
@@ -39,9 +41,9 @@ public Mixed(double value) {
         this.value = value;
     }
 
-    public Mixed(ColumnType columnType) {
+    public Mixed(RealmFieldType columnType) {
```
bmunkholm: columnType -> fieldType

realm/src/androidTest/java/io/realm/internal/test/TestHelper.java
cmelchior: Right now we still use the convention that the internal API talks about columns/tables etc. Changing this would break that convention. 

I know that renaming `ColumnType` to `RealmFieldType` is also a step in that direction, but that is motivated by it becoming public.

So I would rather keep the renaming of other internal API methods out of the scope of this PR as it might cascade into a much bigger renaming.

realm/src/main/java/io/realm/dynamic/RealmFieldType.java
cmelchior: Renaming for now as it looks like it will be used for primitive lists.

realm/src/main/java/io/realm/dynamic/RealmFieldType.java
cmelchior: Renaming for now as it looks like it will be used for dictionaries.

realm/src/main/java/io/realm/RealmQuery.java
cmelchior: Same argument as above

realm/src/main/java/io/realm/internal/Mixed.java
cmelchior: Same argument as above

cmelchior:
As pr. discussion elsewhere we are moving public API's back to `io.realm`. So `RealmFieldType` has moved there as well. MIXED and TABLE has been prefixed with UNSUPPORTED. Ideally I would like to comment them out, but that turned out to be painful as we have quite a lot of of functionality in unit tests and the internal API that depends on them. What do you think?

In any case. Ready for review @realm/java

realm/src/main/java/io/realm/RealmFieldType.java
```diff
@@ -0,0 +1,118 @@
+/*
+ * Copyright 2014 Realm Inc.
```
kneth: 2014 -> 2015

realm/src/main/java/io/realm/RealmFieldType.java
```diff
@@ -0,0 +1,118 @@
...
+    INTEGER(0),
+    BOOLEAN(1),
+    STRING(2),
+    BINARY(4),
+    UNSUPPORTED_TABLE(5),
+    UNSUPPORTED_MIXED(6),
+    DATE(7),
+    FLOAT(9),
+    DOUBLE(10),
+    LINK(12),
```
kneth: Proposal: rename to REALMOBJECT

realm/src/main/java/io/realm/RealmFieldType.java
```diff
@@ -0,0 +1,118 @@
...
+    LINK_LIST(13),
```
kneth: Proposal: rename to REALMLIST

realm/src/main/java/io/realm/RealmFieldType.java
```diff
@@ -0,0 +1,118 @@
...
+    BACKLINK(14);
```
kneth: Proposal: rename to REALMBACKLIST (sounds "funny")

realm/src/main/java/io/realm/RealmFieldType.java
cmelchior: Sounds fine with me.

realm/src/main/java/io/realm/RealmFieldType.java
cmelchior: Sounds fine with me.

realm/src/main/java/io/realm/RealmFieldType.java
beeender: Then we have `RealmFieldType.REALMOBJECT`. Don't you guys think there are too many unnecessary `Realm`s?

beeender:
:+1: But can we merge this after null? There would be some conflicts and easier to be fixed in this PR

cmelchior:
Sure @beeender I'll hold this

realm/src/main/java/io/realm/RealmFieldType.java
cmelchior: It was changed to `RealmFieldType.OBJECT` instead for those exact reasons :)

realm/src/main/java/io/realm/RealmFieldType.java
cmelchior: Changed to LIST instead

realm/src/androidTest/java/io/realm/internal/test/TestHelper.java
bmunkholm: good point.

cmelchior:
Closing this PR so a new can be made against #1494 instead.

PR #1447, opened 2015-09-02
Add RealmEncryptionNotSupportedException

beeender:
- [x] More comments
- [x] Core header file change

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -3,10 +3,11 @@
  * Fixed a bug which might cause failure when loading the native library.
  * Fixed a bug which might trigger a timeout in Context.finalize().
  * Fixed a bug which might cause RealmObject.isValid() to throw an exception if the object is deleted.
- * Updated Realm core to version 0.89.7
-   - fixed a potential stack overflow issue which might cause a crash when encryption was used.
-   - embedded crypto functions from OpenSSL into Realm dynamic lib to avoid random issues on some devices.
  * Fixed an issue preventing the call of listeners on refresh().
+ * Updated Realm core to version 0.89.8
+   - Fixed a potential stack overflow issue which might cause a crash when encryption was used.
+   - Embedded crypto functions into Realm dynamic lib to avoid random issues on some devices.
+   - Throw RealmEncryptionNotSupportedException if the device doesn't support Realm encryption. At least one device type (HTX One X) contains system bugs that prevents Realm's encryption from functioning properly. This is now detected, and an exception is thrown when trying to open/create an encrypted Realm file. It's up to the application to catch this and decide if it's OK to proceed without encryption instead.
 
 0.82.1
  * Fixed a bug where using the wrong encryption key first caused the right key to be seen as invalid.
--- a/realm/src/main/java/io/realm/Realm.java
+++ b/realm/src/main/java/io/realm/Realm.java
@@ -50,6 +50,7 @@ import java.util.concurrent.atomic.AtomicInteger;
 import io.realm.exceptions.RealmException;
 import io.realm.exceptions.RealmIOException;
 import io.realm.exceptions.RealmMigrationNeededException;
+import io.realm.exceptions.RealmEncryptionNotSupportedException;
 import io.realm.internal.ColumnIndices;
 import io.realm.internal.ColumnType;
 import io.realm.internal.ImplicitTransaction;
@@ -198,6 +199,7 @@ public final class Realm implements Closeable {
      * @param configuration Configuration used to open the Realm.
      * @param autoRefresh {@code true} if Realm should auto-refresh. {@code false} otherwise.
      * @throws IllegalArgumentException if trying to open an encrypted Realm with the wrong key.
+     * @throws RealmEncryptionNotSupportedException if the device doesn't support Realm encryption.
      */
     private Realm(RealmConfiguration configuration, boolean autoRefresh) {
         this.threadId = Thread.currentThread().getId();
@@ -375,6 +377,7 @@ public final class Realm implements Closeable {
      *
      * @throws RealmMigrationNeededException If no migration has been provided by the configuration and the
      * model classes or version has has changed so a migration is required.
+     * @throws RealmEncryptionNotSupportedException if the device doesn't support Realm encryption.
      * @see RealmConfiguration for details on how to configure a Realm.
      */
     public static Realm getInstance(RealmConfiguration configuration) {
--- a/realm/src/main/java/io/realm/RealmConfiguration.java
+++ b/realm/src/main/java/io/realm/RealmConfiguration.java
@@ -281,6 +281,10 @@ public class RealmConfiguration {
 
         /**
          * Sets the 64 bit key used to encrypt and decrypt the Realm file.
+         * <p>
+         * Note that a few older devices do not support the encryption used by Realm. These devices will instead throw a
+         * {@link io.realm.exceptions.RealmEncryptionNotSupportedException } when the Realm is opened. See
+         * {@link io.realm.exceptions.RealmEncryptionNotSupportedException } for further details.
          */
         public Builder encryptionKey(byte[] key) {
             if (key == null) {
--- a/realm/src/main/java/io/realm/exceptions/RealmEncryptionNotSupportedException.java
+++ b/realm/src/main/java/io/realm/exceptions/RealmEncryptionNotSupportedException.java
@@ -1,0 +1,31 @@
+/*
+ * Copyright 2015 Realm Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.realm.exceptions;
+
+/**
+ * On some devices (HTC One X for example), for some reason, the system doesn't pass the right
+ * parameter (siginfo_t.si_addr) to the segfault signal handler which our encryption mechanism
+ * relies on. Realm will try to detect if this problem exists on the device when an encrypted
+ * Realm is being created/opened. A RealmEncryptionNotSupportedException will be thrown if this
+ * problem exists which means that encryption cannot be used on this device.
+ */
+@SuppressWarnings("unused") // Thrown by JNI
+public class RealmEncryptionNotSupportedException extends RuntimeException {
+    public RealmEncryptionNotSupportedException(String message) {
+        super(message);
+    }
+}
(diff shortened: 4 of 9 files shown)


realm/src/main/java/io/realm/exceptions/RealmEncryptionNotSupportedException.java
```diff
@@ -0,0 +1,7 @@
+package io.realm.exceptions;
```
cmelchior: Missing License

beeender:
@realm/java Review, thanks. Core change is coming.

realm/src/main/java/io/realm/Realm.java
```diff
@@ -198,6 +198,8 @@ protected void checkIfValid() {
      * @param configuration Configuration used to open the Realm.
      * @param autoRefresh {@code true} if Realm should auto-refresh. {@code false} otherwise.
      * @throws IllegalArgumentException if trying to open an encrypted Realm with the wrong key.
+     * @throws io.realm.exceptions.RealmEncryptionNotSupportedException if the device doesn't
+     * support Realm encryption for some reasons.
```
cmelchior: Remove "for some reasons", it is just boilerplate :)

realm/src/main/java/io/realm/Realm.java
```diff
@@ -375,6 +377,8 @@ public static Realm getDefaultInstance() {
      *
      * @throws RealmMigrationNeededException If no migration has been provided by the configuration and the
      * model classes or version has has changed so a migration is required.
+     * @throws io.realm.exceptions.RealmEncryptionNotSupportedException if the device doesn't
+     * support Realm encryption for some reasons.
```
cmelchior: Same

cmelchior:
I think we should also add a line to `RealmConfiguration.Builder.encryptionKey()` about this as that is really the most obvious place people interact with encryption. Perhaps something along the lines of:

```
...

Note that some devices do not support the encryption used by Realm. These devices will instead throw a {@link RealmEncryptionNotSupportedException } when the Realm is opened. See {@link RealmEncryptionNotSupportedException } for further details.
```

realm/src/main/java/io/realm/exceptions/RealmEncryptionNotSupportedException.java
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+package io.realm.exceptions;
+
+/**
+ * On some devices (HTC One X for example), for some reason, the system doesn't pass the right
+ * parameter (siginfo_t.si_addr) to the segfault signal handler which our encryption mechanism
+ * relies on. Realm will try to detect if this problem exists on the device when an encrypted
+ * Realm is being created/opened. A RealmEncryptionNotSupportedException will be thrown if this
+ * problem exists which causes encryption cannot be done on this device.
```
cmelchior: which means that encryption cannot be used on this device.

changelog.txt
```diff
@@ -3,9 +3,10 @@
  * Fixed a bug which might cause failure when loading the native library.
  * Fixed a bug which might trigger a timeout in Context.finalize().
  * Fixed a bug which might cause RealmObject.isValid() to throw an exception if the object is deleted.
- * Updated Realm core to version 0.89.7
+ * Updated Realm core to version 0.89.8
    - fixed a potential stack overflow issue which might cause a crash when encryption was used.
    - embedded crypto functions from OpenSSL into Realm dynamic lib to avoid random issues on some devices.
+   - throw RealmEncryptionNotSupportedException if the device doesn't support Realm encryption.
```
bmunkholm: Add details here, like: 
At least one device type (HTX One X) contain system bugs that prevents Realms encryption to function properly. This is now detected, and an exception is thrown when trying to open/create an encrypted Realm file. It's up to the application to catch this and decide if it's ok to proceed without encryption instead.

realm/src/main/java/io/realm/Realm.java
```diff
@@ -375,6 +376,7 @@ public static Realm getDefaultInstance() {
      *
      * @throws RealmMigrationNeededException If no migration has been provided by the configuration and the
      * model classes or version has has changed so a migration is required.
+     * @throws io.realm.exceptions.RealmEncryptionNotSupportedException if the device doesn't support Realm encryption.
```
bmunkholm: For the other exceptions we don't include "io.realm.exceptions."

realm/src/main/java/io/realm/RealmConfiguration.java
```diff
@@ -281,6 +281,10 @@ public Builder name(String filename) {
 
         /**
          * Sets the 64 bit key used to encrypt and decrypt the Realm file.
+         * <p>
+         * Note that some devices do not support the encryption used by Realm. These devices will instead throw a
```
bmunkholm: "Note that some devices..." -> "Note that a few older devices.." 
Which ones have we identified? I think we should list them (somewhere - maybe not here) so people can decide if they want to support them.

realm/src/main/java/io/realm/exceptions/RealmEncryptionNotSupportedException.java
```diff
@@ -0,0 +1,31 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.realm.exceptions;
+
+/**
+ * On some devices (HTC One X for example), for some reason, the system doesn't pass the right
```
bmunkholm: Have we identified others than HTC One X? If so we just list that and mention we have only heard it for that device.

beeender: HTC One X+ maybe. I guess Alcatel crash reported by pinterest might be the same root cause. But I don't have evidence to prove that.

bmunkholm:
otherwise :+1:

changelog.txt
```diff
@@ -3,9 +3,10 @@
  * Fixed a bug which might cause failure when loading the native library.
  * Fixed a bug which might trigger a timeout in Context.finalize().
  * Fixed a bug which might cause RealmObject.isValid() to throw an exception if the object is deleted.
- * Updated Realm core to version 0.89.7
-   - fixed a potential stack overflow issue which might cause a crash when encryption was used.
-   - embedded crypto functions from OpenSSL into Realm dynamic lib to avoid random issues on some devices.
+ * Updated Realm core to version 0.89.8
+   - Fixed a potential stack overflow issue which might cause a crash when encryption was used.
+   - Embedded crypto functions from OpenSSL into Realm dynamic lib to avoid random issues on some devices.
```
kneth: Do we really wish to say something about implementation details? I think we should remove "from OpenSSL".

beeender: @bmunkholm Your opinion? For people who cares can easily find out we are using functions from OpenSSL i think...

beeender: And it is in the git history already :(

realm/src/main/java/io/realm/exceptions/RealmEncryptionNotSupportedException.java
kneth: Isn't the list of unsupported devices a good candidate for a FAQ?

beeender: In that case, we also need to put all the details like AOSP version, ROM version, all those information. For example, I believe HTC One X with another ROM (CM maybe) won't have this problem. This incompatibility is software issue which caused by the OEM but not the phone itself.

changelog.txt
kneth: I'm not against letting people know what we are using (in git log). But I find we should leave out implementation details in the change log.

bmunkholm: Yeah, don't think "from OpenSSL" adds any useful information/value to the sentence. So fine to remove.

realm/src/main/java/io/realm/exceptions/RealmEncryptionNotSupportedException.java
bmunkholm: My main point is that this seems very rare and therefore users needs to have an idea about how rare this is and that they should expect encryption to work in almost all cases. Do we have some stat's from the users involved? How many devices in % of their total were affected?

beeender:
retest this please

changelog.txt
```diff
@@ -3,10 +3,11 @@
...
  * Fixed a bug which might trigger a timeout in Context.finalize().
  * Fixed a bug which might cause RealmObject.isValid() to throw an exception if the object is deleted.
- * Updated Realm core to version 0.89.7
-   - fixed a potential stack overflow issue which might cause a crash when encryption was used.
-   - embedded crypto functions from OpenSSL into Realm dynamic lib to avoid random issues on some devices.
  * Fixed an issue preventing the call of listeners on refresh().
+ * Updated Realm core to version 0.89.8
+   - Fixed a potential stack overflow issue which might cause a crash when encryption was used.
+   - Embedded crypto functions into Realm dynamic lib to avoid random issues on some devices.
+   - Throw RealmEncryptionNotSupportedException if the device doesn't support Realm encryption. At least one device type (HTX One X) contains system bugs that prevents Realm's encryption to function properly. This is now detected, and an exception is thrown when trying to open/create an encrypted Realm file. It's up to the application to catch this and decide if it's OK to proceed without encryption instead.
```
cmelchior: from functioning properly

cmelchior:
Kinda massive blob to put in a changelog, but having it being really explicit is properly good. Apart from minor detail in changelog :+1:

bmunkholm:
Please add the exception here: https://github.com/realm/realm-java/blob/4ac64c87aaca494083bd76a3ec27c8a3fdee7ab6/realm/src/main/java/io/realm/internal/Util.java

And likely a comment in the jni code where excetions are defined, that we have to remember to add a test here.

beeender:
@bmunkholm Done.

bmunkholm:
:+1:

cotfas:
Hello

I just wanted to ask

If the phone model doesn`t support encryption (ex htconex+)

When I call 

 realm.writeEncryptedCopyTo(newFile, newKey);

Will RealmEncryptionNotSupportedException will be thrown?

Thanks

cotfas:
​@beeender 

I`ve tested the new jar

On getInstance the RealmEncryptionNotSupportedException  is thrown on my htconex

But when calling 
realm.writeEncryptedCopyTo(newFile, newKey);

No exception has been thrown

But when I`m going to files

The new encrypted file has 0kb


Here is a snapshot 

                        Log.e("REALM" ,"writeEncryptedCopyTo BEFORE realmNotEncrypted.writeEncryptedCopyTo(newFile, newKey) ");
                        realmNotEncrypted.writeEncryptedCopyTo(newFile, newKey);

                        Log.e("REALM", "writeEncryptedCopyTo AFTER realmNotEncrypted.writeEncryptedCopyTo(newFile, newKey) ");
                        realmNotEncrypted.close();

                        Log.e("REALM", "writeEncryptedCopyTo AFTER    realmNotEncrypted.close()");

                        Realm.deleteRealm(configNotEncrypted);

                        Log.e("REALM", "writeEncryptedCopyTo STOP");


Here is the log


09-03 18:46:11.591    2312-2312/com.cashdivider.app E/REALM﹕ writeEncryptedCopyTo BEFORE realmNotEncrypted.writeEncryptedCopyTo(newFile, newKey)
09-03 18:46:11.591    2312-2312/com.cashdivider.app A/libc﹕ Fatal signal 11 (SIGSEGV) at 0x00000000 (code=2), thread 2312 (cashdivider.app) - sent from sys



I`m not sure if I can detect if the htconex doesn`t support encryption because first I have to migrate the unencrypted to encrypted state and after that calling realm.getInstance to throw the RealmEncryptionNotSupportedException

PR #1449, opened 2015-09-03
writeEncryptedCopyTo might throw RENSE

beeender:
writeEncryptedCopyTo might throw RealmEncryptionNotSupportedException as well.

### DIFF
--- a/realm/src/main/java/io/realm/Realm.java
+++ b/realm/src/main/java/io/realm/Realm.java
@@ -914,6 +914,7 @@ public final class Realm implements Closeable {
      * <p>
      * @param destination File to save the Realm to
      * @throws java.io.IOException if any write operation fails
+     * @throws RealmEncryptionNotSupportedException if the device doesn't support Realm encryption.
      */
     public void writeEncryptedCopyTo(File destination, byte[] key) throws IOException {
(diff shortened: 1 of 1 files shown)


cmelchior:
:+1: It won't be true until the hotfix in core has been deployed right?

beeender:
@cmelchior It is true now if user call the `writeEncryptedCopyTo` first time before calling `getInstance` on an encrypted file. https://github.com/realm/realm-core/pull/1119 The problem before is the exception will only be throw once :smile:

realm/src/main/java/io/realm/Realm.java
```diff
@@ -914,6 +914,7 @@ public void writeCopyTo(File destination) throws IOException {
      * <p>
      * @param destination File to save the Realm to
      * @throws java.io.IOException if any write operation fails
+     * @throws RealmEncryptionNotSupportedException if the device doesn't support Realm encryption.
```
bmunkholm: want to add the full path to the exception here? As you did with the other? And use same text?

beeender: @bmunkholm I changed that :) full path is removed as you suggested (importing on the top instead)

PR #1469, opened 2015-09-11
fix for #1467

### LINKED ISSUE #1467: Different instances with same configuration & different thread return same handler
Different instances of Realm sharing the same configuration living in different threads should have different handlers

nhachicha:
@realm/java fix #1467


realm/src/main/java/io/realm/Realm.java
```diff
@@ -1405,13 +1406,7 @@ public void clear(Class<? extends RealmObject> clazz) {
...
     // Returns the Handler for this Realm on the calling thread
     Handler getHandler() {
-        String realmPath = configuration.getPath();
-        for (Map.Entry<Handler, String> entry : handlers.entrySet()) {
-            if (entry.getValue().equals(realmPath)) {
-                return entry.getKey();
-            }
-        }
-        return null;
+        return handler;
```
emanuelez: Might as well make handler package protected?

kneth:
Would it be useful to have a test of two instances opened on the same thread?

cmelchior:
This doesn't seem to fix the problem in `commitTransaction`. Right now the `handlers` HashMap maps between `Handler` and `canonicalPath`. This is wrong. We should have a Handler for each combination of `cannonicalPath + threadID` as `commitTransaction` will loop over all those handlers to notify them all.

cmelchior:
Otherwise, nice catch. This bug has been reported a few times without us being able to find the root cause :100:

cmelchior:
@kneth We already have unit tests for our caching mechanisms, so that should be covered there.

cmelchior:
Handler is now package protected + small refactoring of unit test + changelog.
@realm/java

realm/src/androidTest/java/io/realm/NotificationsTest.java
```diff
@@ -483,4 +483,60 @@ public void onChange() {
...
+        // Create Handler on Thread2 for the same Realm path and close the Realm instance again.
+        new Thread("thread2") {
+            @Override
+            public void run() {
+                Looper.prepare();
+                Realm realm = Realm.getInstance(realmConfiguration);
+                RealmChangeListener listener = new RealmChangeListener() {
+                    @Override
+                    public void onChange() {
+                        fail("This handler should not be notified");
```
nhachicha: you need to catch this inside try catch 
```java
                         try {
                                fail("This handler should not be notified");
                            } catch (AssertionFailedError e) {
                                threadAssertionError[0] = e;
                            }
```
then check at the end of the test
```java
        if (null != threadAssertionError[0]) {
            // throw any assertion errors happened in the background thread
            throw threadAssertionError[0];
        }
```

cmelchior: Ok that is to get a proper error message right, because the unit test fails as expected as far as I can tell?

cmelchior: The above pattern has been added. It does indeed make sure that the correct error is shown in the log.

emanuelez:
:+1:

changelog.txt
```diff
@@ -1,4 +1,5 @@
 0.83
+ * Fixed an issue where opening the same Realm file on two Looper threads could potentially lead to an IllegalStateException being thrown.
```
kneth: Can we have the "BREAKING CHANGE" entries at the top?

cmelchior: Sure, that probably makes sense.

kneth:
:+1:

beeender:
:+1:

PR #1480, opened 2015-09-19
enable filename globbing on clean task in realm-jni

zaki50:


emanuelez:
:+1:

cmelchior:
This will also clean the toolchains. Do we want that?

zaki50:
@cmelchior @emanuelez I'll close this PR because base code is removed by #1488 (d134030)

PR #1479, opened 2015-09-19
let clean task clean jni files

zaki50:

PR #1488, opened 2015-09-21
moved core files into realm-jni/build/ and keep archive file of core under root project. related to #1479

zaki50:
And introduced SHA-256 hash checking to archive file.
The archive file is downloaded only when specified hash differs from local archive.

When new core is comming, please update 'ext.coreSha256Hash' in realm-jni/build.gradle in addition to 'ext.coreVersion'.

### DIFF
--- a/realm-jni/build.gradle
+++ b/realm-jni/build.gradle
@@ -1,4 +1,9 @@
+import java.security.MessageDigest
+
 ext.coreVersion = '0.93.0'
+// empty or comment out this to disable hash checking
+ext.coreSha256Hash = '90044150ff6019a31c044306becccc513c7394f7486ffc8ba4880121c1ef46fa'
+ext.forceDownloadCore = false
 ext.clang = false // gcc is default for the NDK. It also produces smaller binaries
 
 def commonCflags = [ '-std=c++11', '-ffunction-sections', '-fdata-sections', '-flto' ]
@@ -59,19 +64,25 @@ def targets = [
 
 buildscript {
     repositories {
-        mavenCentral()
+        jcenter()
     }
     dependencies {
-        classpath 'de.undercouch:gradle-download-task:1.0'
+        classpath 'de.undercouch:gradle-download-task:2.0.0'
     }
 }
 
-apply plugin: 'download-task'
+apply plugin: 'de.undercouch.download'
 
 Properties localProperties = new Properties()
 localProperties.load(new FileInputStream("${projectDir}/../local.properties"))
 localProperties.entrySet().each() { entry ->
     project.ext[entry.getKey()] = localProperties.setProperty(entry.getKey(), entry.getValue())
+}
+if (project.clang instanceof String) {
+    ext.clang = Boolean.parseBoolean(project.clang)
+}
+if (project.forceDownloadCore instanceof String) {
+    ext.forceDownloadCore = Boolean.parseBoolean(project.forceDownloadCore)
 }
 
 task checkProperties(group: 'check', description: 'Check the user provided gradle.properties') << {
@@ -89,18 +100,91 @@ task checkProperties(group: 'check', description: 'Check the user provided gradl
     }
 }
 
-// TODO: Split this task into smaller subtasks
+ext.coreArchiveFile = project.file("../core-android-${project.coreVersion}.tar.gz")
+ext.coreDir = new File(buildDir, "core-${project.coreVersion}")
+
+def coreDownloaded = false
+
 task downloadCore(group: 'build setup', description: 'Download the latest version of realm core', dependsOn: checkProperties) {
-    outputs.dir "../core-${project.coreVersion}"
+    def isHashCheckingEnabled = {
+        return project.hasProperty('coreSha256Hash') && !project.coreSha256Hash.empty
+    }
+
+    def calcSha256Hash = {File targetFile ->
+        MessageDigest sha = MessageDigest.getInstance("SHA-256");
+        Formatter hexHash = new Formatter()
+        sha.digest(targetFile.bytes).each { b -> hexHash.format('%02x', b) }
+        return hexHash.toString()
+    }
+
+    def shouldDownloadCore = {
+        if (!project.coreArchiveFile.exists()) {
+            return true
+        }
+        if (project.forceDownloadCore) {
+            return true;
+        }
+        if (!isHashCheckingEnabled()) {
+            println "Skipping hash check(empty \'coreSha256Hash\')."
+            return false
+        }
+
+        def calculatedHash = calcSha256Hash(project.coreArchiveFile)
+        if (project.coreSha256Hash.equalsIgnoreCase(calculatedHash)) {
+            return false
+        }
+
+        println "Existing archive hash mismatch(Expected: ${project.coreSha256Hash.toLowerCase()}" +
+                " but got ${calculatedHash.toLowerCase()}). Download new version."
+        return true
+    }
+
     doLast {
-        download {
-            src "http://static.realm.io/downloads/core/realm-core-android-${project.coreVersion}.tar.gz"
-            dest new File(buildDir, "core-android-${project.coreVersion}.tar.gz")
-            onlyIfNewer true
+        if (shouldDownloadCore()) {
+            download {
+                src "http://static.realm.io/downloads/core/realm-core-android-${project.coreVersion}.tar.gz"
+                dest project.coreArchiveFile
+                onlyIfNewer false
+            }
+            coreDownloaded = true
+
+            if (isHashCheckingEnabled()) {
+                def calculatedHash = calcSha256Hash(project.coreArchiveFile)
+                if (!project.coreSha256Hash.equalsIgnoreCase(calculatedHash)) {
+                    throw new GradleException("Invalid checksum for file '" +
+                            "${project.coreArchiveFile.getName()}'. Expected " +
+                            "${project.coreSha256Hash.toLowerCase()} but got " +
+                            "${calculatedHash.toLowerCase()}.");
+                }
+            } else {
+                println 'Skipping hash check(empty \'coreSha256Hash\').'
+            }
+        }
+    }
+}
+
+task deployCore(group: 'build setup', description: 'Deploy the latest version of realm core') {
+    dependsOn { downloadCore }
+
+    outputs.upToDateWhen {
+        if (coreDownloaded) {
+            return false
+        }
+
+        return project.coreDir.exists()
+    }
+
+    doLast {
+        exec {
+            commandLine = [
+                    'rm',
+                    '-rf',
+                    project.coreDir.getAbsolutePath()
+            ]
         }
         copy {
-            from tarTree(new File(buildDir, "core-android-${project.coreVersion}.tar.gz"))
-            into "../core-${project.coreVersion}"
+            from tarTree(project.coreArchiveFile)
+            into project.coreDir
         }
         for (target in targets) {
             exec {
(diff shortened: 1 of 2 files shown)


zaki50:
This PR conflicts with #1480.

I'll update this PR after merge.

zaki50:
@emanuelez This change moves archive file from realm-jni/build/ to below root project.
Is it better to move the archive file under realm-jni/ ?

emanuelez:
Very nice! We need to be able to support to easily work with custom versions of core (built manually for example). The SHA256 is a bit in the way for that use-case. What do you guys think? @realm/java

cmelchior:
The version name is also in the way when doing custom builds, at least it was last week where we made multiple changes to the same version. So for custom versions I would probably suggest a override flag to force download. If that is the case, having the SHA256 only increase confidence that you have the right core version.

zaki50:
@cmelchior You mean like this?

* introduced 'ext.forceDownloadCore'. default is false. if true, always download archive and overwrite existing.
* if 'ext.coreSha256Hash' is empty or not defined, checking of hash is disabled.

```
diff --git a/realm-jni/build.gradle b/realm-jni/build.gradle
index b24156e..7954f68 100644
--- a/realm-jni/build.gradle
+++ b/realm-jni/build.gradle
@@ -1,7 +1,9 @@
 import java.security.MessageDigest
 
 ext.coreVersion = '0.93.0'
+// empty or comment out this to disable hash checking
 ext.coreSha256Hash = '90044150ff6019a31c044306becccc513c7394f7486ffc8ba4880121c1ef46fa'
+ext.forceDownloadCore = false
 ext.clang = false // gcc is default for the NDK. It also produces smaller binaries
 
 def commonCflags = [ '-std=c++11', '-ffunction-sections', '-fdata-sections', '-flto' ]
@@ -60,6 +62,10 @@ def targets = [
     new Target( name:'x86', jniFolder:'x86', toolchain:toolchains.find{it.name == 'x86'}, cflags:[] )
 ]
 
+def isHashCheckingEnabled() {
+    return project.hasProperty('coreSha256Hash') && !project.coreSha256Hash.empty
+}
+
 buildscript {
     repositories {
         jcenter()
@@ -97,10 +103,14 @@ ext.coreDir = new File(buildDir, "core-${project.coreVersion}")
 
 task downloadCore(group: 'build setup', description: 'Download the latest version of realm core', dependsOn: checkProperties) {
     outputs.upToDateWhen {
-        if (!project.coreArchiveFile.exists()) {
+        if (project.forceDownloadCore || !project.coreArchiveFile.exists()) {
             return false
         }
 
+        if (!isHashCheckingEnabled()) {
+            return true
+        }
+
         MessageDigest sha = MessageDigest.getInstance("SHA-256");
         Formatter hexHash = new Formatter()
         sha.digest(project.coreArchiveFile.bytes).each { b -> hexHash.format('%02x', b) }
@@ -111,7 +121,7 @@ task downloadCore(group: 'build setup', description: 'Download the latest versio
         download {
             src "http://static.realm.io/downloads/core/realm-core-android-${project.coreVersion}.tar.gz"
             dest project.coreArchiveFile
-            onlyIfNewer true
+            onlyIfNewer !project.forceDownloadCore
         }
     }
 }
@@ -119,10 +129,16 @@ task downloadCore(group: 'build setup', description: 'Download the latest versio
 task verifyCore(type: de.undercouch.gradle.tasks.download.Verify, dependsOn: downloadCore) {
     src       project.coreArchiveFile
     algorithm 'SHA-256'
-    checksum  project.coreSha256Hash
+    checksum  isHashCheckingEnabled() ? project.coreSha256Hash : ''
 }
 
-task deployCore(group: 'build setup', description: 'Deploy the latest version of realm core', dependsOn: verifyCore) {
+task deployCore(group: 'build setup', description: 'Deploy the latest version of realm core') {
+    if (isHashCheckingEnabled()) {
+        dependsOn { verifyCore }
+    } else {
+        dependsOn { downloadCore }
+    }
+
     doLast {
         exec {
             commandLine = [
```

zaki50:
I found one problem in this PR...

deployCore task is always executed and it cleans up object files in src/.
And that always leads to recompiling of C++ files...

zaki50:
I fixed above problem.

zaki50:
summary of current status.

* core-[version].tar.gz is moved to root project from realm-jni/build and is not removed on clean.
* extracted core files are moved to realm-jni/build/ and is removed on clean.
* core-[version].tar.gz is force downloaded if 'ext.forceDownloadCore' is true, 
* hash check is skipped if 'ext.coreSha256Hash' is empty or not defined.
* these variables can be set by local.properties

realm-jni/build.gradle
```diff
@@ -89,18 +104,75 @@ task checkProperties(group: 'check', description: 'Check the user provided gradl
...
+def coreDownloaded = false
+
 task downloadCore(group: 'build setup', description: 'Download the latest version of realm core', dependsOn: checkProperties) {
-    outputs.dir "../core-${project.coreVersion}"
+    outputs.upToDateWhen {
+        if (project.forceDownloadCore || !project.coreArchiveFile.exists()) {
+            return false
+        }
+        if (!isHashCheckingEnabled()) {
+            println 'skipped hash checking(empty \'coreSha256Hash\').'
```
cmelchior: skipped -> Skipping

realm-jni/build.gradle
```diff
@@ -89,18 +104,75 @@ task checkProperties(group: 'check', description: 'Check the user provided gradl
...
+
+        MessageDigest sha = MessageDigest.getInstance("SHA-256");
+        Formatter hexHash = new Formatter()
+        sha.digest(project.coreArchiveFile.bytes).each { b -> hexHash.format('%02x', b) }
+        return project.coreSha256Hash.equals(hexHash.toString())
+    }
+
     doLast {
+        if (project.forceDownloadCore) {
+            println 'forceDownloadCore is true.'
```
cmelchior: Perhaps just "Downloading core..."? It shouldn't come as an surprise that forceDownload is set so no need to print it in the log IMO

realm-jni/build.gradle
```diff
@@ -89,18 +104,75 @@ task checkProperties(group: 'check', description: 'Check the user provided gradl
...
+}
+
+task downloadAndVerifyCore(type: de.undercouch.gradle.tasks.download.Verify, dependsOn: downloadCore) {
+    src       project.coreArchiveFile
+    algorithm 'SHA-256'
+    checksum  isHashCheckingEnabled() ? project.coreSha256Hash : ''
+}
+
+task deployCore(group: 'build setup', description: 'Deploy the latest version of realm core') {
+    if (isHashCheckingEnabled()) {
```
cmelchior: Minor thing: It would feel more encapsulated if this if statement was part of just a `downloadCore` task, i.e. merge `downloadCore` and `downloadAndVerifyCore` instead of having having the hash being referenced across 3 functions.

cmelchior:
I like this, apart from some minor things it looks good.

realm-jni/build.gradle
zaki50: OK, I'll remove this

emanuelez:
retest this please

emanuelez:
retest this please

realm-jni/build.gradle
zaki50: I'll try to refactor this code.

zaki50:
@cmelchior @emanuelez done. please review again.

zaki50:
@cmelchior @emanuelez I revised last commit and force pushed. please review this.

realm-jni/build.gradle
```diff
@@ -89,18 +100,91 @@ task checkProperties(group: 'check', description: 'Check the user provided gradl
...
+
+    def shouldDownloadCore = {
+        if (!project.coreArchiveFile.exists()) {
+            return true
+        }
+        if (project.forceDownloadCore) {
+            return true;
+        }
+        if (!isHashCheckingEnabled()) {
+            println "Skipping hash checking(empty \'coreSha256Hash\')."
```
cmelchior: Skipping hash check

realm-jni/build.gradle
```diff
@@ -89,18 +100,91 @@ task checkProperties(group: 'check', description: 'Check the user provided gradl
...
+            return false
+        }
+
+        def calculatedHash = calcSha256Hash(project.coreArchiveFile)
+        if (project.coreSha256Hash.equalsIgnoreCase(calculatedHash)) {
+            return false
+        }
+
+        println "Existing archive hash mismatch(Expected: ${project.coreSha256Hash.toLowerCase()}" +
```
cmelchior: This sounds a bit "dangerous" also "force downloading" isn't really true, as it will also trigger if a new version is available. Rewriting the last sentence to something like "Download new version" would sound better imo

realm-jni/build.gradle
```diff
@@ -89,18 +100,91 @@ task checkProperties(group: 'check', description: 'Check the user provided gradl
...
+            if (isHashCheckingEnabled()) {
+                def calculatedHash = calcSha256Hash(project.coreArchiveFile)
+                if (!project.coreSha256Hash.equalsIgnoreCase(calculatedHash)) {
+                    throw new GradleException("Invalid checksum for file '" +
+                            "${project.coreArchiveFile.getName()}'. Expected " +
+                            "${project.coreSha256Hash.toLowerCase()} but got " +
+                            "${calculatedHash.toLowerCase()}.");
+                }
+            } else {
+                println 'Skipping hash checking(empty \'coreSha256Hash\').'
```
cmelchior: Skipping hash check

cmelchior:
Apart from some minor Log output this looks good :+1:

zaki50:
@cmelchior done. Should I squash commits in this PR?

cmelchior:
Yes please

zaki50:
@cmelchior after #1480 merged to master, I will rebase and squash this PR.

zaki50:
just squashed

zaki50:
@cmelchior I'll rebase #1480 after merged this PR.

zaki50:
@cmelchior @emanuelez I think there are more files to be removed on clean.

${rootProject}/realm/src/main/jniLibs/
${rootProject}/build/output/jniLibs-unstripped/

These directories contain shared libraries built by realm-jni module.

Can I add a code to remove these files?

cmelchior:
We moved the files from `${rootProject}/realm/src/main/jniLibs/` to the build dir recently, so the first looks like a bug if files are still in the `main` folder. The idea by putting them in `build` was that they should be cleaned automatically but I havn't verified that that is actually happening. If not, then yes, it makes perfectly good sense to add it.

zaki50:
@cmelchior on master branch, shared library files are copied under main.

```
    task "copyAndroidJni${target.name.capitalize()}"(dependsOn: "buildAndroidJni${target.name.capitalize()}") << {
        copy {
            from "${projectDir}/src/librealm-jni-${target.name}-stripped.so"
            into "${projectDir}/../realm/src/main/jniLibs/${target.jniFolder}"
            rename "librealm-jni-${target.name}-stripped.so", 'librealm-jni.so'
        }
      ...
}
```

zaki50:
How do you think about https://github.com/realm/realm-java/pull/1480#issuecomment-142197850 @realm/java

kneth:
We shouldn't clear the toolchain. When upgrading to a new toolchain, I think it should be done manually (that is, removing the toolchain).

cmelchior:
I tend to agree. We should probably document what to do in case of a toolchain upgrade though as I remember some bugs in past due to this.

cmelchior:
Discussing this with @emanuelez 
When we move to the experimental plugin this will change anyway, so for now it is fine
:+1:

PR #1494, opened 2015-09-24
[Feature] Dynamic + Migration API

cmelchior:
This PR is an umbrella for all PR's related to Dynamic/Migration API's. After agreement with @bmunkholm we don't want to merge to many breaking changes into master until Null has been released.

This means that all PRs in this area should be made against this branch instead of master.


cmelchior:
F*ck. Apparently i made a mistake somewhere and this PR got merged to master. It has been reverted. I will create a new one. Sorry the trouble :(

@realm/java

PR #1496, opened 2015-09-24
[Feature] Dynamic + Migration API

cmelchior:
This PR is an umbrella for all PR's related to Dynamic/Migration API's. After agreement with @bmunkholm we don't want to merge to many breaking changes into master until Null has been released.

This means that all PRs in this area should be made against this branch instead of master.

----
Following PR's are contained here:
- [x] #1495 ColumnType -> RealmFieldType
- [x] #1493 Constants to Enums in query parameters
- [x] #1498 Full dynamic API
- [x] Dynamic API w/ Null #1674  
- [x] Migration API #1677 
- [x] Update migration example #1677 
- [x] Internal API cleanup #1760 
- [x] Merge #1790 to this PR


cmelchior:
retest this please

changelog.txt
```diff
@@ -1,4 +1,11 @@
 0.85.0
+ * BREAKING CHANGE: The Migration API has been replaced with a new API.
+ * BREAKING CHANGE: RealmResults.SORT_ORDER_ASCENDING and RealmResults.SORT_ORDER_DESCENDING constants
+   has been replaced by Sort.ASCENDING and Sort.DESCENDING enums.
```
kneth: "has" -> "have"

changelog.txt
```diff
@@ -1,4 +1,11 @@
...
+ * BREAKING CHANGE: RealmQuery.CASE_SENSITIVE and RealmQuery.CASE_INSENSITIVE constants has been
```
kneth: "has" -> "have"

cmelchior:
Slight update to the DynamicRealm description @realm/java

realm/realm-library/src/main/java/io/realm/RealmMigration.java
```diff
@@ -17,49 +17,55 @@
...
  * {@link io.realm.Realm#deleteRealm(RealmConfiguration)}. This will delete the database file and eliminate the need for
  * any migrations.
  *
- * @see Realm#migrateRealm(RealmConfiguration)
- * @see Realm#migrateRealm(RealmConfiguration, RealmMigration)
+ * @see io.realm.RealmConfiguration.Builder#schemaVersion(long)
+ * @see io.realm.RealmConfiguration.Builder#migration(RealmMigration)
+ * @see io.realm.RealmConfiguration.Builder#deleteRealmIfMigrationNeeded()
  */
 public interface RealmMigration {
```
zaki50: How about adding `long getSchemaVersion()` here and remove(deprecate) `RealmConfiguration.Builder#schemaVersion(long)` since Migration logic and schema version are tightly coupled.

cmelchior: The schema version and migration code is only tightly coupled when you actually upgrade. New installs will only use the schema version. So IMO it makes sense to specify the version outside the migration code. This is also what e.g. the SQLite API's does, so it should be more familiar to people.

zaki50: If no upgrade needed, no need to specify schema version.

What does SQLite API mean? When using `SQLiteOpenHelper`, we make an sub class of `SQLiteOpenHelper` and specify db version in its constructor and provide migration logic in overriding method `onUpgrade()`.

cmelchior: You need the schema version for new installs, otherwise any future upgrades would break.
Yes, I was thinking about `SQLOpenHelper`. The version in the constructor is the one used for new installs and given as the `newVersion` in `onUpgrade(SQLiteDatabase db, int oldVersion, int newVersion)`

zaki50: In most cases, the schema version for new install can be `0` and that seem to work with later migration.

Do I overlook something?

cmelchior: Yes.

1) User A installs the app at version 0
2) User B installs the app at version 1
3) App gets upgraded to version 2:
- A will run through the migrations 0 -> 1 - > 2 correctly.
- B will have the schema for version 1 when installing a new app, so if 0 is used for the schema version he will run the wrong migrations.

zaki50: At 2), That app must have the migration code with schema version 1. It means schema version of User B never be `0`.

cmelchior: Yes, that is correct. The schema version for user B cannot be 0 when he install the app. I thought you argued for the opposite above?

zaki50: Oh, that was my bad.

I mean if no migration provided, the schema version can be `0`.
And if migration which coupled with schema version provided, created or migrated schema can both have that schema version.

realm/realm-library/src/androidTest/java/io/realm/RealmConfigurationTest.java
```diff
@@ -159,7 +159,7 @@ public void testVersionLessThanDiscVersionThrows() {
         int[] wrongVersions = new int[] { 0, 1, 41 };
         for (int version : wrongVersions) {
             try {
-                Realm.getInstance(new RealmConfiguration.Builder(getContext()).schemaVersion(version).build());
+                realm = Realm.getInstance(new RealmConfiguration.Builder(getContext()).schemaVersion(version).build());
```
beeender: Doesn't this generate unused lint warning? But no need to change, just a comments :)

cmelchior: No, because `realm` is using in the `tearDown` method

changelog.txt
```diff
@@ -13,7 +20,7 @@
  * Fixed a bug making it impossible to convert a field to become required during a migration (#1695).
  * Fixed a bug making it impossible to read Realms created using primary keys and created by iOS (#1703).
  * Fixed some memory leaks when an Exception is thrown (#1730).
- * Fixed a memory leak when using relationships (#1285).
+ * Fixed a memory leak when using relationships (#1285).s
```
beeender: remove s

beeender:
Except the extra char in changelog.txt,  :+1:

changelog.txt
```diff
@@ -1,5 +1,10 @@
 0.86.0
-* BREAKING CHANGE: `Realm.addChangeListener`, `RealmObject.addChangeListener` and `RealmResults.addChangeListener` hold a strong reference to the listener, you should unregister the listener to avoid memory leaks.
+ * BREAKING CHANGE: The Migration API has been replaced with a new API.
+ * BREAKING CHANGE: RealmResults.SORT_ORDER_ASCENDING and RealmResults.SORT_ORDER_DESCENDING constants have been replaced by Sort.ASCENDING and Sort.DESCENDING enums.
+ * BREAKING CHANGE: RealmQuery.CASE_SENSITIVE and RealmQuery.CASE_INSENSITIVE constants have been replaced by Case.SENSITIVE and Case.INSENSITIVE enums.
+ * BREAKING CHANGE: `Realm.addChangeListener`, `RealmObject.addChangeListener` and `RealmResults.addChangeListener` hold a strong reference to the listener, you should unregister the listener to avoid memory leaks.
```
kneth: Please don't use back-quotes in the change log.

examples/migrationExample/src/main/java/io/realm/examples/realmmigrationexample/model/Migration.java
```diff
@@ -16,24 +16,29 @@
...
+ * Example of migrating a Realm file from version 0 (initial version) to its last version (version 3).
+ */
 public class Migration implements RealmMigration {
+
     @Override
-    public long execute(Realm realm, long version) {
+    public void migrate(final DynamicRealm realm, long oldVersion, long newVersion) {
+        // During a migration, a DynamicRealm is exposed. A DynamicRealm is an untyped variant of a normal Realm, but
+        // with the same object creation and query capabilities.
+        // A DynamicRealm uses Strings instead of Class references which are needed as the Classes might not even
```
kneth: "which are needed as" -> "because"

realm/realm-library/src/androidTest/java/io/realm/RealmObjectSchemaTests.java
```diff
@@ -0,0 +1,586 @@
...
+        realm = DynamicRealm.getInstance(realmConfig);
+        realmSchema = realm.getSchema();
+        DOG_SCHEMA = realmSchema.get("Dog");
+        realm.beginTransaction();
+        schema = realmSchema.create("NewClass");
+    }
+
+    @Override
+    protected void tearDown() throws Exception {
+        super.tearDown();
```
kneth: It can probably no practical impact, but shouldn't `super.tearDown()` be at the end?

realm/realm-library/src/androidTest/java/io/realm/RealmSchemaTests.java
```diff
@@ -0,0 +1,152 @@
...
+            fail();
+        } catch (IllegalArgumentException ignored) {
+        }
+
+        try {
+            realmSchema.rename(CLASS_ALL_JAVA_TYPES, null);
+            fail();
+        } catch (IllegalArgumentException ignored) {
+        }
+    }
```
kneth: Renaming to `""` (empty string)?

realm/realm-library/src/androidTest/java/io/realm/RealmSchemaTests.java
```diff
@@ -0,0 +1,152 @@
...
+        try {
+            realmSchema.remove(null);
+            fail();
+        } catch (IllegalArgumentException ignored) {
+        }
+    }
+
+    // Test that it if { A -> B  && B -> A } you should remove the individual fields first before removing the entire
+    // class. This also include transitive dependencies :/
+    // Re-enable when this if fixed: https://github.com/realm/realm-core/pull/1267
```
kneth: Please create a github issue for it.

realm/realm-library/src/androidTest/java/io/realm/internal/test/TestHelper.java
```diff
@@ -0,0 +1,67 @@
+package io.realm.internal.test;
```
kneth: Header missing

realm/realm-library/src/androidTest/java/io/realm/RealmObjectSchemaTests.java
cmelchior: No. convention is it should be at the top. If you need to put it at the end, it is because your class is using some state from the superclass it probably shouldn't.

realm/realm-library/src/androidTest/java/io/realm/RealmSchemaTests.java
cmelchior: Done

realm/realm-library/src/androidTest/java/io/realm/internal/test/TestHelper.java
cmelchior: Done

PR #1497, opened 2015-09-24
ColumnIndices are now setup correctly for all Realms

cmelchior:
This fixes a recently introduced bug (caught by the async branch) where column indices are not configured correctly if the same Realm is opened on multiple threads causing queries to crash.

@realm/java

### DIFF
--- a/realm/src/main/java/io/realm/Realm.java
+++ b/realm/src/main/java/io/realm/Realm.java
@@ -308,8 +308,16 @@ public final class Realm extends BaseRealm {
                     throw e;
                 }
             }
+            setupColumnIndices(realm);
 
             return realm;
+        }
+    }
+
+    private static void setupColumnIndices(Realm realm) {
+        RealmProxyMediator mediator = realm.configuration.getSchemaMediator();
+        for (Class<? extends RealmObject> modelClass : mediator.getModelClasses()) {
+            realm.columnIndices.addClass(modelClass, mediator.getColumnIndices(modelClass));
         }
     }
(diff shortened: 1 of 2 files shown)


realm/src/main/java/io/realm/Realm.java
```diff
@@ -308,11 +308,19 @@ private static Realm createAndValidate(RealmConfiguration configuration, boolean
...
             }
+            setupColumnIndices(realm);
 
             return realm;
         }
     }
 
+    private static void setupColumnIndices(Realm realm) {
+        RealmProxyMediator mediator = realm.configuration.getSchemaMediator();
+        for (Class<? extends RealmObject> modelClass : mediator.getModelClasses()) {
```
emanuelez: You might want to consider to use a normal for loop as opposed to the iterator, just from a performance standpoint.

kneth:
:+1:

realm/src/main/java/io/realm/Realm.java
cmelchior: Good point. Fixed

cmelchior:
We got a report that this branch now potentially can result in "Handler sending message to Handler on a dead thread". Most likely the handler isn't cleared properly in all cases. Reverting to WIP until we find the root cause.

cmelchior:
Fixes to reported bugs have been put in #1501 and #1502. This PR is fine as it is, so back to Review

emanuelez:
:+1:

PR #1498, opened 2015-09-24
Full Dynamic API

cmelchior:
Copy from #1369 

--------
This introduces a full blown Dynamic API (apart from access to Schema which will be part of #880)

Only two new classes are required for this: `DynamicRealm` and `DynamicRealmObject`. `RealmQuery`, `RealmResults` and `RealmList` can be reused as they are because `DynamicRealmObject` is a specialized subtype of `RealmObject`.

* `DynamicRealm` makes it possible to query/create/delete using String arguments instead of Class references. See *Known Limitations* below.

* `DynamicRealmObject` makes it possible to interact with the `RealmObject` in a dynamic manner through getters/setters.

--- 
Other PR's
- [x] Handle nullable and boxed types fields ( #1548 )
- [ ] Add missing methods to DynamicRealm (if any)
- [ ] Add example using the dynamic API + Migration API  
- [ ] Migration API
- [ ] Add Async methods to DynamicRealm

Missing:
- [x] Generic getType() or get() in order to work with more generalised code ( #1394 ).
- [x] JavaDoc for new Dynamic/Base classes
- [x] Unit tests for the functionality that isn't shared with the typed counter parts
  * [x] DynamicRealm
  * [x] DynamicRealmObject
- [x] Discuss where the current booleans should be placed. Suggestion -> Move to enums in seperate classes instead
- [x] Split Dynamic API and typed API sharing the SharedGroup. The amount of book-keeping for this is insane and it introduces lots of weird edge cases. Writes across instances, setting auto refresh effects both instances and so on.  

----
Know limitations
- DynamicRealm is a stripped down version of a regular Realm. In the first iteration it will only provide simple query/create capabilities, i.e. no `json*` methods and no `createOrUpdate` and friends.
- `deleteRealm`, `compactRealm`, `migrateRealm` are not available in DynamicRealm. 
- All JSON* methods are not available as they will be moving away from Realm as well.
- copy* and copyOrUpdate* are not available.


cmelchior:
This PR is ready for review. Discussion on smaller changes are still ongoing in https://github.com/realm/realm-java/pull/1265 but it should still be possible to review at this stage.

It introduces the following breaking change:
- DynamicRealmObject have been moved from `io.realm.dynamic` to `io.realm`
- DynamicRealmList is being removed and replaced by `RealmList<DynamicRealmObject>`

@realm/java

cmelchior:
Review please @realm/java

realm/src/main/java/io/realm/DynamicRealm.java
```diff
@@ -0,0 +1,277 @@
...
+import java.util.Map;
+import java.util.concurrent.CopyOnWriteArrayList;
+
+import io.realm.exceptions.RealmException;
+import io.realm.internal.Table;
+import io.realm.internal.TableView;
+
+/**
+ * DynamicRealm is a dynamic variant of {@link io.realm.Realm}. This means that all access to data and/or queries are
+ * done using Strings instead of classes.
```
kneth: "Strings" -> "class names"?

realm/src/main/java/io/realm/DynamicRealm.java
```diff
@@ -0,0 +1,277 @@
...
+ *
+ * The same {@link io.realm.RealmConfiguration} can be used to open a Realm file in both dynamic and typed mode, but
+ * modifying the schema while doing so is highly discouraged and will most likely crash the typed Realm.
+ *
+ * Dynamic Realms do not enforce schemas and schema versions and doesn't trigger migrations even if they have been
```
kneth: "doesn't" -> "don't"

realm/src/main/java/io/realm/DynamicRealmObject.java
```diff
@@ -264,6 +290,78 @@ public boolean hasField(String fieldName) {
...
+                            "setter instead", fieldName, value));
+            }
+        }
+
+        if (value == null) {
+            row.nullifyLink(row.getColumnIndex(fieldName));
+            // TODO Add support for other types when Null is merged.
+        } else if (value instanceof Boolean) {
+            setBoolean(fieldName, (Boolean) value);
+        } else if (value instanceof Short) {
```
kneth: What about `Byte` as a 8 bit integer?

realm/src/main/java/io/realm/RealmQuery.java
```diff
@@ -49,23 +51,73 @@
...
+     * @param className  The type to query.
+     * @return {@link RealmQuery} object. After building the query call one of the {@code find*} methods
+     * to run it.
+     */
+    public static <E extends RealmObject> RealmQuery<E> createDynamicQuery(DynamicRealm realm, String className) {
+        return new RealmQuery<E>(realm, className);
+    }
+
+    /**
+     * Creates a sub-query from a existing {@link RealmResults}.
```
kneth: I'm not sure like to use "sub-query" as a term. Currently, we don't have projections and cannot return values from aggregate functions. In SQL, subqueries have a well-defined meaning, and I am afraid that users will be confused.

realm/src/main/java/io/realm/RealmQuery.java
```diff
@@ -49,23 +51,73 @@
...
+
+    /**
+     * Creates a sub-query from a existing {@link RealmList}.
+     *
+     * @param list   A existing @{link io.realm.RealmList} to query against.
+     * @return {@link RealmQuery} object. After building the query call one of the {@code find*} methods
+     * to run it.
+     */
+    @SuppressWarnings("unchecked")
+    public static <E extends RealmObject> RealmQuery<E> createSubQuery(RealmList<E> list) {
```
kneth: Can't we stick to `where()`?

realm/src/main/java/io/realm/RealmQuery.java
```diff
@@ -1283,9 +1374,80 @@ public long count() {
...
         long rowIndex = this.query.find();
         if (rowIndex >= 0) {
-            return realm.get(clazz, (view != null) ? view.getTargetRowIndex(rowIndex) : rowIndex);
+            return realm.get(clazz, className, (view != null) ? view.getTargetRowIndex(rowIndex) : rowIndex);
         } else {
             return null;
         }
     }
+
+    // TODO Replace with Schema when it is available
```
kneth: Please use `FIXME`

realm/src/main/java/io/realm/RealmQuery.java
cmelchior: We are still using `where` in the public API. However using `where` as a factory constructor is really `ambiguous` so I would rather not use that here.

realm/src/main/java/io/realm/RealmQuery.java
cmelchior: Fair enough concern. These methods should only be used by a very small subset of users, but they are still public. Do you have a better name for it?

kneth: I believe "query" is fine. The sentence will be "Creates a query for an existing RealmResults" and it is clear (at least in my mind).

realm/src/main/java/io/realm/DynamicRealmObject.java
cmelchior: Byte is handled by it's own setter. I would try to avoid doing these kinds of optimisations here as it really complicate the setters and the error message they might emit.

cmelchior:
retest this please

realm/src/main/java/io/realm/DynamicRealmObject.java
```diff
@@ -264,6 +290,78 @@ public boolean hasField(String fieldName) {
...
+                    throw new IllegalArgumentException(String.format("Field %s is not a String field, " +
+                            "and the provide value could not be automatically converted: %s. Use a typed" +
+                            "setter instead", fieldName, value));
+            }
+        }
+
+        if (value == null) {
+            row.nullifyLink(row.getColumnIndex(fieldName));
+            // TODO Add support for other types when Null is merged.
+        } else if (value instanceof Boolean) {
```
emanuelez: How about an `Enum` like:

```java
public Enum Action {
   Boolean {
      setField() {
         setBoolean(fieldName, (Boolean) value);
      }
   },
   Integer { // etc...  }

   abstract void setField (...);
}
```

That would allow for more elegant code IMO. I think to remember this pattern from Effective Java.

cmelchior: Wouldn't that just be another level of indirection? I need some way to convert a `Object` to the appropriate type. A `enum` could only do that with some static function, but internally that would have the same `if` statements.

emanuelez: It is be a way to encapsulate complexity, which has value IMO.
There is also a performance aspect of it. The performance of ìnstanceof` is more or less the same as `Object.getClass()` but we would need only one `getClass()` as opposed to many calls to `instanceof` for types down the chain of if statements.

cmelchior: Well, you are just replacing 

```
if (x instance y) {
} else if (x instance z) {
}
```

with 

```
Class clazz = x.getClass()
if (class == y.class) {
} else if (class == z.class) {
}
```

But you are probably right that those comparisons are faster than `instanceof`. Encapsulating it probably provides a lot cleaner code though. I'll look into it.

emanuelez: I would go further than that since lots of equals can be costly as well.

We could setup an `Enum` and mentioned above. Let's call it SupportedType.
Then we could have a `static Map<Class<?>, SupportedType>`
Then we could do:

```java
SupportedType type = Types.get(value.getClass());
if (type != null) {
   type.setField(fieldName, value);
} else {
   // blah blah...
}
```

cmelchior: I actually tested the performance of HashMap vs. just straight if/else, and at least for x > 20, straight if/else was actually faster in my tests

cmelchior: The methods of an enum cannot access the outer scope, even if it is a inner class (I learned something new today :) ). So we cannot really use an enum for this, but it should be possible to either refactor the if/else out and optimize it a bit `==` instead of `instanceof`.
```
Object obj = new Object();
Date date = new Date();

obj ==

**Comment by cmelchior** [comment]
Bump @realm/java

File: realm/src/androidTest/java/io/realm/DynamicRealmObjectTest.java [hunk 8]
```diff
@@ -106,7 +105,7 @@ private void callGetter(SupportedType type, List<String> arguments) {
         }
     }
 
-    // Test that all getters fail if given invalid field name
+    // Test that all getters fail if given a invalid field name
```
nhachicha: a -> an

realm/src/androidTest/java/io/realm/DynamicRealmObjectTest.java
```diff
@@ -232,7 +232,7 @@ public void testSetXXXNullValues() {
                     case BINARY:
                     case DATE:
                     default:
-                        continue; // Ignore other types for now
+                        continue; // Ignore other types for now until null support is merged.
```
nhachicha: update code/comment (null is merged)

realm/src/androidTest/java/io/realm/DynamicRealmObjectTest.java
```diff
@@ -255,11 +255,199 @@ public void testSetObjectWrongTypeThrows() {
...
+                    case SHORT:
+                    case INT:
+                    case LONG:
+                    case FLOAT:
+                    case DOUBLE:
+                    case STRING:
+                    case BINARY:
+                    case DATE:
+                    default:
+                        continue; // Ignore other types for now until Null support is merged.
```
nhachicha: update code since null is merged, add test for Boxed type

realm/src/androidTest/java/io/realm/DynamicRealmObjectTest.java
```diff
@@ -255,11 +255,199 @@ public void testSetObjectWrongTypeThrows() {
...
+                    case FLOAT:
+                        dObj.set(AllJavaTypes.FIELD_FLOAT, "1.23");
+                        assertEquals(1.23f, dObj.getFloat(AllJavaTypes.FIELD_FLOAT));
+                        break;
+                    case DOUBLE:
+                        dObj.set(AllJavaTypes.FIELD_DOUBLE, "1.234");
+                        assertEquals(1.234d, dObj.getDouble(AllJavaTypes.FIELD_DOUBLE));
+                        break;
+
+                    // These types doesn't have a string representation that should be parsed.
```
nhachicha: doesn't -> don't

realm/src/androidTest/java/io/realm/DynamicRealmObjectTest.java
```diff
@@ -255,11 +255,199 @@ public void testSetObjectWrongTypeThrows() {
...
+                    case OBJECT:
+                    case LIST:
+                    case STRING:
+                    case BINARY:
+                    case DATE:
```
nhachicha: how about a Date as timestamp or text using `JsonUtils.stringToDate`

realm/src/androidTest/java/io/realm/DynamicRealmObjectTest.java
```diff
@@ -255,11 +255,199 @@ public void testSetObjectWrongTypeThrows() {
...
+                            dObj.set(AllJavaTypes.FIELD_DOUBLE, "foo");
+                            break;
+
+                        // These types doesn't have a string representation that should be parsed.
+                        case BOOLEAN: // Boolean is special as it returns false for all strings != "true"
+                        case OBJECT:
+                        case LIST:
+                        case STRING:
+                        case BINARY:
+                        case DATE:
```
nhachicha: see previous comment

realm/src/androidTest/java/io/realm/DynamicRealmTest.java
```diff
@@ -0,0 +1,325 @@
...
+            fail();
+        } catch (IllegalStateException ignored) {
+        }
+
+        dynamicRealm.close();
+        assertTrue(Realm.deleteRealm(defaultConfig));
+    }
+
+    // Test that the closed Realm isn't kept in the Realm instance cache
+    public void testRealmCacheIsCleared() {
```
nhachicha: nice idea, we should have a similar test for `handlers`

realm/src/androidTest/java/io/realm/DynamicRealmTest.java
```diff
@@ -0,0 +1,325 @@
...
+        Realm typedRealm = Realm.getInstance(defaultConfig);
+        DynamicRealm dynamicRealm = DynamicRealm.getInstance(defaultConfig);
+
+        typedRealm.close(); // Still a instance open, but typed Realm cache must still be cleared.
```
nhachicha: a -> an 
or 
Although one instance is open Realm cache must still be cleared.

realm/src/androidTest/java/io/realm/DynamicRealmTest.java
```diff
@@ -0,0 +1,325 @@
...
+        }
+
+        dynamicRealm.close();
+        assertTrue(Realm.deleteRealm(defaultConfig));
+    }
+
+    // Test that the closed Realm isn't kept in the Realm instance cache
+    public void testRealmCacheIsCleared() {
+        Realm typedRealm = Realm.getInstance(defaultConfig);
+        DynamicRealm dynamicRealm = DynamicRealm.getInstance(defaultConfig);
```
nhachicha: could add a comment that this will return `realm` instance created in `setUp`

realm/src/androidTest/java/io/realm/DynamicRealmTest.java
```diff
@@ -0,0 +1,325 @@
...
+            typedRealm.close();
+        }
+    }
+
+    // Test that the closed DynamicRealms isn't kept in the DynamicRealm instance cache
+    public void testDynamicRealmCacheIsCleared() {
+        DynamicRealm dynamicRealm = DynamicRealm.getInstance(defaultConfig);
+        Realm typedRealm = Realm.getInstance(defaultConfig);
+
+        dynamicRealm.close(); // Still a instance open, but DynamicRealm cache must still be cleared.
```
nhachicha: a -> an

realm/src/androidTest/java/io/realm/DynamicRealmTest.java
```diff
@@ -0,0 +1,325 @@
...
+    }
+
+    public void testExecuteTransactionNull() {
+        realm.executeTransaction(null); // Nothing happens
+        assertFalse(realm.hasChanged());
+    }
+
+    public void testExecuteTransactionCommit() {
+        assertEquals(0, realm.allObjects(CLASS_OWNER).size());
+        realm.executeTransaction(new DynamicRealm.Transaction() {
```
nhachicha: why not using `Realm.Transaction`?

realm/src/androidTest/java/io/realm/DynamicRealmTest.java
```diff
@@ -0,0 +1,325 @@
...
+            @Override
+            public void execute(DynamicRealm realm) {
+                DynamicRealmObject owner = realm.createObject(CLASS_OWNER);
+                owner.setString("name", "Owner");
+            }
+        });
+        assertEquals(1, realm.allObjects(CLASS_OWNER).size());
```
nhachicha: you're not asserting `name==Owner`

realm/src/androidTest/java/io/realm/DynamicRealmTest.java
```diff
@@ -0,0 +1,325 @@
...
+
+        assertEquals("Brian", results2.get(1).getString(AllTypes.FIELD_STRING));
+        assertEquals(4, results2.get(1).getLong(AllTypes.FIELD_LONG));
+
+        assertEquals("Adam", results2.get(2).getString(AllTypes.FIELD_STRING));
+        assertEquals(5, results2.get(2).getLong(AllTypes.FIELD_LONG));
+    }
+
+    public void testSortMultiFailures() {
+
```
nhachicha: rm space

realm/src/androidTest/java/io/realm/internal/test/TestHelper.java
```diff
@@ -62,20 +64,27 @@ public static Table getTableWithAllColumnTypes(){
...
-
-        AllTypes object2 = testRealm.createObject(AllTypes.class);
-        object2.setColumnLong(4);
-        object2.setColumnString("Brian");
-
-        AllTypes object3 = testRealm.createObject(AllTypes.class);
-        object3.setColumnLong(4);
-        object3.setColumnString("Adam");
-        testRealm.commitTransaction();
+    public static void populateForMultiSort(Realm typedRealm) {
```
nhachicha: as per Async PR I merged our TestHelper classes, use the one under `io.realm`

realm/src/main/java/io/realm/BaseRealm.java
```diff
@@ -510,6 +579,35 @@ protected static synchronized void validateAgainstExistingConfigurations(RealmCo
...
+    }
+
+    // Used by RealmList/RealmResults
+    // Invariant: if dynamicClassName != null -> clazz == DynamicRealmObject
+    <E extends RealmObject> E get(Class<E> clazz, String dynamicClassName, long rowIndex) {
+        Table table;
+        E result;
+        if (dynamicClassName != null) {
+            table = getTable(dynamicClassName);
+            result = (E) new DynamicRealmObject();
```
nhachicha: you can use `clazz.cast(new DynamicRealmObject())` to avoid the unchecked cast

realm/src/main/java/io/realm/DynamicRealm.java
```diff
@@ -0,0 +1,277 @@
...
+import io.realm.exceptions.RealmException;
+import io.realm.internal.Table;
+import io.realm.internal.TableView;
+
+/**
+ * DynamicRealm is a dynamic variant of {@link io.realm.Realm}. This means that all access to data and/or queries are
+ * done using string based class names instead of class type references.
+ *
+ * The same {@link io.realm.RealmConfiguration} can be used to open a Realm file in both dynamic and typed mode, but
+ * modifying the schema while doing so is highly discouraged and will most likely crash the typed Realm.
```
nhachicha: this is confusing, we'll get a DynamicRealm for migration (this will modify the schema)?

realm/src/main/java/io/realm/DynamicRealm.java
```diff
@@ -0,0 +1,277 @@
...
+        return RealmQuery.createDynamicQuery(this, className);
+    }
+
+    /**
+     * Removes all objects of the specified class.
+     *
+     * @param className The class for which all objects should be removed.
+     */
+    public void clear(String className) {
+        getTable(className).clear();
```
nhachicha: shouldn't we `checkIfValid();` or another Thread can delete all. same for `Realm.checkIfValid();`

realm/src/main/java/io/realm/DynamicRealm.java
```diff
@@ -0,0 +1,277 @@
...
+    public RealmResults<DynamicRealmObject> allObjectsSorted(String className, String fieldName1,
+                                                                    Sort sortOrder1, String fieldName2,
+                                                                    Sort sortOrder2) {
+        return allObjectsSorted(className, new String[]{fieldName1, fieldName2}, new Sort[]{sortOrder1,
+                sortOrder2});
+    }
+
+    /**
+     * Get all objects of a specific class name sorted by multiple fields.  If no objects exist, the
+     * returned RealmResults will not be null. The RealmResults.size() to check the number of
```
nhachicha: > The RealmResults.size() to check the number of objects instead.

you mean: call RealmResults.size() to check the number of objects instead.

realm/src/main/java/io/realm/DynamicRealm.java
```diff
@@ -0,0 +1,277 @@
...
+            references = 0;
+        }
+        Map<RealmConfiguration, DynamicRealm> realms = realmsCache.get();
+        DynamicRealm realm = realms.get(configuration);
+        if (realm != null) {
+            localRefCount.put(configuration, references + 1);
+            return realm;
+        }
+
+        // Create new Realm and cache it. All exception code paths must close the Realm otherwise we risk serving
```
nhachicha: IIRC from test DynamicRealm are not cached, remove/update comment

realm/src/main/java/io/realm/DynamicRealmObject.java
```diff
@@ -34,31 +30,60 @@
...
      */
     public DynamicRealmObject(RealmObject obj) {
         if (obj == null) {
             throw new IllegalArgumentException("Non-null object must be provided.");
         }
-        Row row = RealmObject.getRow(obj);
+        Row row = obj.row;
         if (row == null) {
-            throw new IllegalArgumentException("A object managed by Realm must be provided. This is a standalone object.");
+            throw new IllegalArgumentException("A object managed by Realm must be provided. This " +
```
nhachicha: An object

realm/src/main/java/io/realm/DynamicRealmObject.java
```diff
@@ -34,31 +30,60 @@
...
     // Create a dynamic object. Only used internally
-    DynamicRealmObject(Realm realm, CheckedRow row) {
+    DynamicRealmObject() {
+
+    }
+
+    DynamicRealmObject(BaseRealm realm, Row row) {
         this.realm = realm;
-        this.row = row;
+        this.row = (row instanceof CheckedRow) ? (CheckedRow) row : ((UncheckedRow) row).convertToChecked();
```
nhachicha: you don't need to first cast `(CheckedRow) row ` & why we need to work with CheckedRow?

realm/src/main/java/io/realm/DynamicRealmObject.java
```diff
@@ -34,31 +30,60 @@
...
+    }
+
+    /**
+     * Returns the value for the given field. To avoid casting the value use one of the typed
+     * getters instead.
+     *
+     * @param fieldName Name of the field.
+     * @return The field value. Private types will be converted to their boxed variants.
+     */
+    public Object get(String fieldName) {
```
nhachicha: can't we use `public <E> E get(String fieldName)` to infer the returned type  (instead of returning Object)

realm/src/main/java/io/realm/DynamicRealmObject.java
```diff
@@ -264,6 +290,86 @@ public boolean hasField(String fieldName) {
...
+     * Using the typed setters will be faster than using this method.
+     *
+     * @throws IllegalArgumentException if field name doesn't exists or if the input value cannot be converted
+     * to the appropriate input type.
+     * @throws NumberFormatException if a String based number cannot be converted properly.
+     */
+    @SuppressWarnings("unchecked")
+    public void set(String fieldName, Object value) {
+        boolean isString = (value instanceof String);
+        String strValue = isString ? (String) value : null;
```
nhachicha: should we check if the user passed a Date as String? maybe using `JsonUtils.stringToDate`

realm/src/main/java/io/realm/DynamicRealmObject.java
```diff
@@ -264,6 +290,86 @@ public boolean hasField(String fieldName) {
...
+                default:
+                    throw new IllegalArgumentException(String.format("Field %s is not a String field, " +
+                            "and the provide value could not be automatically converted: %s. Use a typed" +
+                            "setter instead", fieldName, value));
+            }
+        }
+
+        if (value == null) {
+            row.nullifyLink(row.getColumnIndex(fieldName));
+            // FIXME Add support for other types when Null is merged.
```
nhachicha: now it's time to fix your FIXME 👮

realm/src/main/java/io/realm/DynamicRealmObject.java
```diff
@@ -264,6 +290,86 @@ public boolean hasField(String fieldName) {
...
+        } else if (valueClass == String.class) {
+            setString(fieldName, (String) value);
+        } else if (value instanceof Date) {
+            setDate(fieldName, (Date) value);
+        } else if (value instanceof byte[]) {
+            setBlob(fieldName, (byte[]) value);
+        } else if (valueClass == DynamicRealmObject.class) {
+            setObject(fieldName, (DynamicRealmObject) value);
+        } else if (valueClass == RealmList.class) {
+            // We need to verify that the list only contain DynamicRealmObjects.
```
nhachicha: will this work? http://stackoverflow.com/a/3403976

realm/src/main/java/io/realm/RealmList.java
```diff
@@ -183,6 +191,10 @@ public E set(int location, E object) {
 
     // Transparently copies a standalone object or managed object from another Realm to the Realm backing this RealmList.
     private E copyToRealmIfNeeded(E object) {
+        if (!(realm instanceof Realm))  {
+            throw new IllegalStateException("Not supported for RealmList containing DynamicRealmObjects");
+        }
+        Realm realm = (Realm) this.realm;
```
nhachicha: shouldn't you update the object Realm? 
`object.realm=this.realm` 
+ comment/name is confusing this is not a copy the implementation reads more like a `moveTo`

realm/src/main/java/io/realm/RealmQuery.java
```diff
@@ -1536,9 +1627,80 @@ public long count() {
...
-            return realm.get(clazz, (view != null) ? view.getTargetRowIndex(rowIndex) : rowIndex);
+            return realm.get(clazz, className, (view != null) ? view.getTargetRowIndex(rowIndex) : rowIndex);
         } else {
             return null;
         }
     }
+
+    // FIXME Replace with Schema when it is available
+    private static class DynamicColumnMap implements Map<String, Long> {
+
```
nhachicha: rm space

realm/src/main/java/io/realm/RealmResults.java
```diff
@@ -53,25 +53,49 @@
  */
 public class RealmResults<E extends RealmObject> extends AbstractList<E> {
 
-    private Class<E> classSpec;
-    private Realm realm;
+    BaseRealm realm;
+    Class<E> classSpec;   // Return type
+    String className;     // Class name for DynamicRealmObjects -> TODO This is not set properly
```
nhachicha: TODO -> FIXME ?

nhachicha:
LGTM beside some minor comments

realm/src/androidTest/java/io/realm/DynamicRealmObjectTest.java
cmelchior: This is fixed in https://github.com/realm/realm-java/pull/1548 instead. This PR does not contain null support.

realm/src/androidTest/java/io/realm/DynamicRealmObjectTest.java
cmelchior: This is fixed in #1548 instead. This PR does not contain null support.

realm/src/androidTest/java/io/realm/DynamicRealmObjectTest.java
cmelchior: Good idea. I'll add that.

realm/src/main/java/io/realm/DynamicRealmObject.java
cmelchior: Because `DynamicRealmObject extends RealmObject` it is actually possible to do ```new DynamicRealmObject(otherDynamicObject)``` I don't think there is as valid use case for it though, so I should probably just throw an exception in that case.

realm/src/main/java/io/realm/DynamicRealmObject.java
cmelchior: No, Java doesn't have overloaded return types, so you need the type information somehow. You could do `obj.get("foo", Integer.class)" that returned an Integer, but then you might as well just use `obj.getInt("foo")`.

realm/src/main/java/io/realm/DynamicRealmObject.java
cmelchior: Yes we should.

realm/src/main/java/io/realm/DynamicRealmObject.java
cmelchior: In this PR: #1548

realm/src/main/java/io/realm/DynamicRealmObject.java
cmelchior: Worth a shot, but no. It cannot tell you the type at Runtime, so all you get out is <E extends RealmObject> which is not useful. I could get the first element and ask for it's type, but then I actually prefer catching the exception as that will only be slow if it is actually wrong.

realm/src/main/java/io/realm/RealmList.java
cmelchior: No, this is the semantics we are already using when doing normal `realm.copyToRealm()`, it doesn't modify the existing object, but creates a new copy that is inserted. So this method actually does create a copy.

realm/src/main/java/io/realm/DynamicRealmObject.java
zaki50: `public <E> E get(String fieldName)` is a valid method declaration in Java and in most cases, E is decided by type inference.

```
Long value = realmObject.get("foo");
```

if `get(String)` returns for example `String` object, simply `ClassCastException` will be thrown.

This is not an overloaded return type because actual return type is `Object`.

cmelchior: I stand corrected, but it does require a LOT of "unsafe" casts inside the method. In this case it is fine though. So I will change it.

realm/src/main/java/io/realm/DynamicRealmObject.java
cmelchior: Now throwing an exception if trying to create a DynamicRealmObject from another DynamicRealmObject.

cmelchior:
All comments fixed.

cmelchior:
Bump @realm/java

realm/src/androidTest/java/io/realm/DynamicRealmObjectTest.java
```diff
@@ -255,11 +255,203 @@ public void testSetObjectWrongTypeThrows() {
...
+                        dObj.set(AllJavaTypes.FIELD_LONG, 42L);
+                        assertEquals(Long.parseLong("42"), dObj.get(AllJavaTypes.FIELD_LONG));
+                        break;
+                    case BYTE:
+                        dObj.set(AllJavaTypes.FIELD_BYTE, (byte) 4);
+                        assertEquals(Long.parseLong("4"), dObj.get(AllJavaTypes.FIELD_BYTE));
+                        break;
+                    case FLOAT:
+                        dObj.set(AllJavaTypes.FIELD_FLOAT, 1.23f);
+                        assertEquals(Float.parseFloat("1.23"), dObj.get(AllJavaTypes.FIELD_FLOAT));
```
beeender: Some one will ask you to use assertEquals(float, float, delta) instead :P

realm/src/androidTest/java/io/realm/DynamicRealmObjectTest.java
```diff
@@ -255,11 +255,203 @@ public void testSetObjectWrongTypeThrows() {
...
+                        break;
+                    case DOUBLE:
+                        dObj.set(AllJavaTypes.FIELD_DOUBLE, 1.234d);
+                        assertEquals(Double.parseDouble("1.234"), dObj.get(AllJavaTypes.FIELD_DOUBLE));
```
beeender: same

realm/src/androidTest/java/io/realm/DynamicRealmObjectTest.java
```diff
@@ -255,11 +255,203 @@ public void testSetObjectWrongTypeThrows() {
...
+                        dObj.set(AllJavaTypes.FIELD_INT, "42");
+                        assertEquals(42, dObj.getInt(AllJavaTypes.FIELD_INT));
+                        break;
+                    case LONG:
+                        dObj.set(AllJavaTypes.FIELD_LONG, "42");
+                        assertEquals((long) 42, dObj.getLong(AllJavaTypes.FIELD_LONG));
+                        break;
+                    case FLOAT:
+                        dObj.set(AllJavaTypes.FIELD_FLOAT, "1.23");
+                        assertEquals(1.23f, dObj.getFloat(AllJavaTypes.FIELD_FLOAT));
```
beeender: same

realm/src/androidTest/java/io/realm/DynamicRealmObjectTest.java
```diff
@@ -255,11 +255,203 @@ public void testSetObjectWrongTypeThrows() {
...
+                        break;
+                    case DOUBLE:
+                        dObj.set(AllJavaTypes.FIELD_DOUBLE, "1.234");
+                        assertEquals(1.234d, dObj.getDouble(AllJavaTypes.FIELD_DOUBLE));
```
beeender: same

realm/src/main/java/io/realm/DynamicRealm.java
```diff
@@ -0,0 +1,283 @@
...
+        if (configuration == null) {
+            throw new IllegalArgumentException("A non-null RealmConfiguration must be provided");
+        }
+        return create(configuration);
+    }
+
+    /**
+     * Instantiates and adds a new object to the Realm.
+     *
+     * @param className The class of the object to create.
```
beeender: 'The class` -> `the class name`

realm/src/main/java/io/realm/DynamicRealm.java
```diff
@@ -0,0 +1,283 @@
...
+     * @return The new object.
```
beeender: `The` -> `the`

beeender: And other capitalized first characters in this file. see example https://github.com/realm/realm-wiki/wiki/Java-Doc-guidelines

realm/src/main/java/io/realm/DynamicRealm.java
```diff
@@ -0,0 +1,283 @@
...
+     */
+    public RealmResults<DynamicRealmObject> allObjects(String className) {
+        return where(className).findAll();
+    }
+
+    /**
+     * Get all objects of a specific class name sorted by a field.  If no objects exist, the returned
+     * RealmResults will not be null. Use {@link RealmResults#size()} to check the number of objects instead.
+     *
+     * @param className The class to get all objects from.
```
beeender: `The` -> `the`

realm/src/main/java/io/realm/DynamicRealm.java
```diff
@@ -0,0 +1,283 @@
...
+     * @param fieldName the field name to sort by.
+     * @param sortOrder how to sort the results.
+     * @return A sorted RealmResults containing the objects.
```
beeender: {@link RealmResults} and others in this file

realm/src/main/java/io/realm/DynamicRealm.java
```diff
@@ -0,0 +1,283 @@
...
+        }
+
+        TableView tableView = table.getSortedView(columnIndex, sortOrder);
+        return RealmResults.createFromDynamicQuery(this, tableView, className);
+    }
+
+
+    /**
+     * Get all objects of a specific class name sorted by two specific field names.  If no objects exist,
+     * the returned RealmResults will not be null. Use {@link RealmResults#size()} to check the number of
```
beeender: {@code null}

realm/src/main/java/io/realm/DynamicRealm.java
cmelchior: Yup, this will have to be adjusted all over the place. The Javadoc was written before we decided on the change, and I forgot to update it :(

realm/src/main/java/io/realm/DynamicRealm.java
```diff
@@ -0,0 +1,283 @@
...
+     * objects instead.
+
+     you mean: call RealmResults.size() to check the number of objects instead.
```
beeender: ???

realm/src/main/java/io/realm/DynamicRealm.java
```diff
@@ -0,0 +1,283 @@
...
+
+        // Create new DynamicRealm and cache it. All exception code paths must close the DynamicRealm otherwise we risk
+        // serving faulty cache data.
+        validateAgainstExistingConfigurations(configuration);
+        boolean autoRefresh = Looper.myLooper() != null;
+        realm = new DynamicRealm(configuration, autoRefresh);
+        List<RealmConfiguration> pathConfigurationCache = globalPathConfigurationCache.get(canonicalPath);
+        if (pathConfigurationCache == null) {
+            pathConfigurationCache = new CopyOnWriteArrayList<RealmConfiguration>();
+            globalPathConfigurationCache.put(canonicalPath, pathConfigurationCache);
```
beeender: This would be a problem. `globalPathConfigurationCache` is a static member on `BaseRealm` and lock of this function is on `DynamicRealm`. 

It would be logically simpler that we just have lock on `BaseRealm.class` for all `BaseRealm`, `Realm` and `DynamicRealm`.

beeender: Uh.. `acquireFileReference` has a lock as well... then what about move the `globalPathConfigurationCache` related logic to a function in `BaseRealm`?

beeender: Just noticed all the static member of this class is `ThreadLocal`, then the the lock should be on `BaseRealm.class`

realm/src/main/java/io/realm/DynamicRealmObject.java
```diff
@@ -34,31 +32,65 @@
...
 
-     Realm realm;
-     Row row;
-
     /**
      * Creates a dynamic Realm object based on a existing object.
      *
      * @param obj Realm object to convert to a dynamic object. Only objects managed by Realm can be used.
-     * @throws IllegalArgumentException if object isn't managed by a Realm.
+     * @throws IllegalArgumentException if object isn't managed by Realm or is a {@link DynamicRealmObject} already.
```
beeender: {@link Realm}

realm/src/main/java/io/realm/DynamicRealmObject.java
```diff
@@ -34,31 +32,65 @@
...
      */
     public DynamicRealmObject(RealmObject obj) {
         if (obj == null) {
             throw new IllegalArgumentException("Non-null object must be provided.");
         }
-        Row row = RealmObject.getRow(obj);
+        if (obj instanceof DynamicRealmObject) {
+            throw new IllegalArgumentException("Object is already a DynamicRealmObject: " + obj);
```
beeender: "Object" -> This object / the object

realm/src/main/java/io/realm/DynamicRealmObject.java
```diff
@@ -34,31 +32,65 @@
...
     /**
      * Creates a dynamic Realm object based on a existing object.
      *
      * @param obj Realm object to convert to a dynamic object. Only objects managed by Realm can be used.
-     * @throws IllegalArgumentException if object isn't managed by a Realm.
+     * @throws IllegalArgumentException if object isn't managed by Realm or is a {@link DynamicRealmObject} already.
      */
     public DynamicRealmObject(RealmObject obj) {
         if (obj == null) {
             throw new IllegalArgumentException("Non-null object must be provided.");
```
beeender: "A non-null object must be provided."

realm/src/main/java/io/realm/Realm.java
```diff
@@ -1037,7 +1005,7 @@ public static void migrateRealm(RealmConfiguration configuration) {
      * @param migration {@link RealmMigration} to run on the Realm. This will override any migration set on the
      * configuration.
      */
-    public static void migrateRealm(RealmConfiguration configuration, RealmMigration migration) {
+    public synchronized static void migrateRealm(RealmConfiguration configuration, RealmMigration migration) {
```
beeender: Redundant `synchronized`?
`BaseRealm.migrateRealm` is already `synchronized`

realm/src/androidTest/java/io/realm/DynamicRealmTest.java
```diff
@@ -0,0 +1,328 @@
...
+        realm.createObject(CLASS_ALL_TYPES);
+        realm.commitTransaction();
+
+        RealmResults<DynamicRealmObject> results = realm.where(CLASS_ALL_TYPES).findAll();
+        assertEquals(1, results.size());
+    }
+
+    public void testClearInvalidName() {
+        realm.beginTransaction();
+        realm.clear("I don't exist");
```
beeender: This doesn't fail? Strange, `BaseRealm.getTable()` should return a `null` and calling `clear` on `Table null` should throw, right?

realm/src/androidTest/java/io/realm/DynamicRealmObjectTest.java
```diff
@@ -296,6 +488,17 @@ public void testHasFieldTrue() {
...
+    public void testGetFieldType() {
+        assertEquals(RealmFieldType.STRING, dObj.getFieldType(AllJavaTypes.FIELD_STRING));
+        assertEquals(RealmFieldType.BINARY, dObj.getFieldType(AllJavaTypes.FIELD_BINARY));
+        assertEquals(RealmFieldType.BOOLEAN, dObj.getFieldType(AllJavaTypes.FIELD_BOOLEAN));
+        assertEquals(RealmFieldType.DATE, dObj.getFieldType(AllJavaTypes.FIELD_DATE));
+        assertEquals(RealmFieldType.DOUBLE, dObj.getFieldType(AllJavaTypes.FIELD_DOUBLE));
+        assertEquals(RealmFieldType.FLOAT, dObj.getFieldType(AllJavaTypes.FIELD_FLOAT));
+        assertEquals(RealmFieldType.OBJECT, dObj.getFieldType(AllJavaTypes.FIELD_OBJECT));
+        assertEquals(RealmFieldType.LIST, dObj.getFieldType(AllJavaTypes.FIELD_LIST));
+    }
```
zaki50: missing FIELD_BYTE, FIELD_SHORT, FIELD_INT, FIELD_LONG?

realm/src/main/java/io/realm/DynamicRealmObject.java
```diff
@@ -34,31 +32,65 @@
...
+        Row row = obj.row;
         if (row == null) {
-            throw new IllegalArgumentException("A object managed by Realm must be provided. This is a standalone object.");
+            throw new IllegalArgumentException("An object managed by Realm must be provided. This " +
+                    "is a standalone object.");
         }
-        this.realm = RealmObject.getRealm(obj);
-        this.row = (row instanceof CheckedRow) ? (CheckedRow) row : ((UncheckedRow) row).convertToChecked();
+        this.realm = obj.realm;
+        this.row = ((UncheckedRow) row).convertToChecked();
```
zaki50: we should check if `row` is an instance of `InvalidRow` (== removed object)?

realm/src/androidTest/java/io/realm/DynamicRealmObjectTest.java
cmelchior: In this case we test `get(set(x)) = x`. Hopefully that won't introduce a delta error?

beeender: I bet it should be be exact `1.23`, otherwise it would be so strange... and i think `assertEquals(Float.parseFloat("1.23"), dObj.get(AllJavaTypes.FIELD_FLOAT));` actually means `assertEquals(Float.parseFloat("1.23"), dObj.get(AllJavaTypes.FIELD_FLOAT), 0);` ? Doc said the first function is deprecated, but i don't see warnings in AS.

realm/src/main/java/io/realm/DynamicRealm.java
cmelchior: Your right. We should use a lock on BaseRealm.class just like `Realm` is doing. It has been added.

cmelchior:
Updated with PR feedback @realm/java

cmelchior:
Bump @realm/java

realm/src/androidTest/java/io/realm/DynamicRealmObjectTest.java
```diff
@@ -255,11 +276,203 @@ public void testSetObjectWrongTypeThrows() {
...
+                    case SHORT:
+                    case INT:
+                    case LONG:
+                    case FLOAT:
+                    case DOUBLE:
+                    case STRING:
+                    case BINARY:
+                    case DATE:
+                    default:
+                        // Ignore other types for now until Null support is merged.
```
kneth: It is now possible to implement this.

cmelchior: It will be part of a separate PR. I try to avoid making the PR too big ;)

realm/src/androidTest/java/io/realm/DynamicRealmTest.java
```diff
@@ -0,0 +1,332 @@
...
+            assertEquals(1, typedRealm.where(AllTypes.class).count());
+            assertEquals(0, dynamicRealm.where(CLASS_ALL_TYPES).count());
+            typedRealm.cancelTransaction();
+        } finally {
+            typedRealm.close();
+            dynamicRealm.close();
+        }
+    }
+
+    // Test that Realms can only be deleted after all Typed and Dynamic instances are closed.
```
kneth: Is it worth a test to swap `Realm` and `DynamicRealm`? I mean, closing `dynamicRealm` first and `realm` second.

realm/src/main/java/io/realm/BaseRealm.java
```diff
@@ -494,8 +509,39 @@ static synchronized void releaseFileReference(RealmConfiguration configuration)
...
     }
 
+    // Public because of migrations
+    @Deprecated
+    public Table getTable(String className) {
+        className = Table.TABLE_PREFIX + className;
+        Table table = dynamicClassToTable.get(className);
+        if (table == null) {
+            if (!sharedGroupManager.hasTable(className)) {
+                throw new IllegalArgumentException("The type " + className + " doesn't exist in this Realm.");
```
kneth: "The type" -> "The class"

realm/src/main/java/io/realm/BaseRealm.java
```diff
@@ -505,6 +551,32 @@ protected void setVersion(long version) {
     }
 
     /**
+     * Sort a table using the given field names and sorting directions. If a field name doesn not
```
kneth: "doesn" -> "does"

realm/src/main/java/io/realm/RealmResults.java
```diff
@@ -70,7 +71,23 @@
     private Future<Long> pendingQuery;
     private boolean isCompleted = false;
 
-    RealmResults(Realm realm, Class<E> classSpec) {
+    static <E extends RealmObject> RealmResults<E> createFromClass(BaseRealm realm, Class<E> clazz) {
+        return new RealmResults<E>(realm, clazz);
+    }
+
+    static <E extends RealmObject> RealmResults<E> createFromQuery(BaseRealm realm, TableOrView table, Class<E> clazz) {
```
kneth: This method doesn't use a query to create a `RealmResults`. A better name might be createFromTableOrView`.

realm/src/main/java/io/realm/BaseRealm.java
cmelchior: This is for the Dynamic interface. I am not sure "class" is the correct word? I have a hard time finding a vocabulary for it. Right now I am using "className" as the variable name, which would suggest I should change the wording like you said, but should we find other words for it?

realm/src/androidTest/java/io/realm/DynamicRealmTest.java
cmelchior: Yes, i'll add it.

realm/src/main/java/io/realm/BaseRealm.java
```diff
@@ -77,6 +88,7 @@
     protected SharedGroupManager sharedGroupManager;
     protected boolean autoRefresh;
     Handler handler;
+    ColumnIndices columnIndices;
```
zaki50: `DynamicRealm` does not assign `ColumnIndices` instance to `columnIndices` and thus leave it `null`.

nhachicha:
LGTM

realm/src/main/java/io/realm/BaseRealm.java
zaki50: I read entire code again and I understood `columnIndices` is not used from Dynamic APIs.

It's better to move `columnIndices` to Realm.java in order not to accidentally access this field from Dynamic APIs.

realm/src/main/java/io/realm/RealmQuery.java
```diff
@@ -60,26 +63,76 @@
...
-    private final Table table;
-    private final LinkView view;
-    private final TableQuery query;
-    private final Map<String, Long> columns;
-    private final Class<E> clazz;
-
+    private BaseRealm realm;
+    private Class<E> clazz;
+    private String className;
+    private Map<String, Long> columns = new HashMap<String, Long>();
```
zaki50: seems no need to initialize here.

Every constructor initializes `columns` again.

beeender: Right, this doesn't seem to be necessary.

beeender:
Except one minor comment, :+1:

realm/src/main/java/io/realm/BaseRealm.java
cmelchior: It would yes, but right now RealmQuery, RealmList and RealmResults only know about `BaseRealm`. I have a feeling that the right answer is probably letting the Realm subclass convert all Class references to Strings right away, but it has of yet unknown performance/code clarity implications.

I would prefer to address this in another PR.

realm/src/main/java/io/realm/BaseRealm.java
cmelchior: I changed it to `class` for now. It is in line with the vocabulary used elsewhere currently

cmelchior:
Was that a thumbs up @nhachicha ? ;)

nhachicha:
:+1:

zaki50:
:+1:

realm/src/androidTest/java/io/realm/DynamicRealmObjectTest.java
```diff
@@ -66,6 +65,26 @@ protected void tearDown() throws Exception {
         BOOLEAN, SHORT, INT, LONG, BYTE, FLOAT, DOUBLE, STRING, BINARY, DATE, OBJECT, LIST;
     }
 
+    public void testIllegalInputObjectThrows() {
+        try {
+            new DynamicRealmObject(null);
+        } catch (IllegalArgumentException ignored) {
```
bmunkholm: You are missing `fail()` here and below. Otherwise this will pass even if an exception isn't thrown.

realm/src/androidTest/java/io/realm/DynamicRealmObjectTest.java
```diff
@@ -232,7 +253,7 @@ public void testSetXXXNullValues() {
                     case BINARY:
                     case DATE:
                     default:
-                        continue; // Ignore other types for now
+                        // Ignore other types for now until null support is merged.
```
bmunkholm: Should we have a TODO here?

cmelchior: Its being added here https://github.com/realm/realm-java/pull/1674

realm/src/androidTest/java/io/realm/DynamicRealmTest.java
```diff
@@ -0,0 +1,435 @@
...
+        }
+        realm.commitTransaction();
+    }
+
+    private void populateTestRealm() {
+        populateTestRealm(realm, TEST_DATA_SIZE);
+    }
+
+
+    // Test that the SharedGroupManager is not reused across Realm/DynamicRealm on the same thread.
```
bmunkholm: Great with descriptions above tests! :100:

realm/src/androidTest/java/io/realm/DynamicRealmTest.java
```diff
@@ -0,0 +1,435 @@
...
+    public void testWhere() {
+        realm.beginTransaction();
+        realm.createObject(CLASS_ALL_TYPES);
+        realm.commitTransaction();
+
+        RealmResults<DynamicRealmObject> results = realm.where(CLASS_ALL_TYPES).findAll();
+        assertEquals(1, results.size());
+    }
+
+    public void testClearInvalidNameThrows() {
```
bmunkholm: Consider to add a section comment: // *** Test clear()
Or make two extra newlines or something to group methods related to testing the same method.
Alternatively make it one method to test clear().

realm/src/androidTest/java/io/realm/DynamicRealmTest.java
```diff
@@ -0,0 +1,435 @@
...
+                obj.setString("indexString", withNull ? null :  "Test " + j);
+                obj.setBoolean("notIndexBoolean", j % 2 == 0);
+                obj.setLong("notIndexLong", j);
+                obj.setDate("notIndexDate", withNull ? null : new Date(1000 * j));
+                obj.setString("notIndexString", withNull ? null : "Test " + j);
+            }
+        }
+        realm.commitTransaction();
+    }
+
```
bmunkholm: remove empty line.

realm/src/androidTest/java/io/realm/entities/AllTypes.java
```diff
@@ -23,6 +23,18 @@
 import io.realm.annotations.Required;
 
 public class AllTypes extends RealmObject {
+
+    public static final String CLASS_NAME = "AllTypes";
+    public static final String FIELD_STRING = "columnString";
```
bmunkholm: In another PR, shouldn't we rename "column" to "field" for AllTypes?

cmelchior: That would probably make sense yes. At this point we could probably trim our test entities a bit in any case since we have a lot of overlap between them.

realm/src/main/java/io/realm/DynamicRealm.java
```diff
@@ -0,0 +1,337 @@
...
+import java.util.concurrent.CopyOnWriteArrayList;
+
+import io.realm.exceptions.RealmException;
+import io.realm.internal.Table;
+import io.realm.internal.TableView;
+
+/**
+ * DynamicRealm is a dynamic variant of {@link io.realm.Realm}. This means that all access to data and/or queries are
+ * done using string based class names instead of class type references.
+ *
```
bmunkholm: Maybe add to lines about what it's useful for.

realm/src/main/java/io/realm/DynamicRealm.java
```diff
@@ -0,0 +1,337 @@
...
+            };
+
+    private DynamicRealm(RealmConfiguration configuration, boolean autoRefresh) {
+        super(configuration, autoRefresh);
+    }
+
+    /**
+     * Realm static constructor that returns a dynamic variant of the Realm instance defined by provided
+     * {@link io.realm.RealmConfiguration}. Dynamic Realms do not care about schemaVersion and schemas, so opening a
+     * DynamicRealm will never trigger a migration
```
bmunkholm: missing "."

realm/src/main/java/io/realm/DynamicRealm.java
```diff
@@ -0,0 +1,337 @@
...
+     */
+    public RealmResults<DynamicRealmObject> allObjectsSorted(String className, String fieldName1,
+                                                                    Sort sortOrder1, String fieldName2,
+                                                                    Sort sortOrder2) {
+        return allObjectsSorted(className, new String[]{fieldName1, fieldName2}, new Sort[]{sortOrder1,
+                sortOrder2});
+    }
+
+    /**
+     * Get all objects of a specific class name sorted by multiple fields.  If no objects exist, the
```
bmunkholm: remove extra spaces in front of `If`

realm/src/main/java/io/realm/DynamicRealm.java
```diff
@@ -0,0 +1,337 @@
...
+        } catch (Error e) {
+            cancelTransaction();
+            throw e;
+        }
+    }
+
+    /**
+     * Get all objects of a specific class name.
+     *
+     * @param className the Class to get objects of
```
bmunkholm: missing `.` Also below.

realm/src/main/java/io/realm/DynamicRealm.java
```diff
@@ -0,0 +1,337 @@
...
+     * @param className the class to get all objects from.
+     * @param sortOrders sort ascending if SORT_ORDER_ASCENDING, sort descending if SORT_ORDER_DESCENDING.
+     * @param fieldNames an array of field names to sort objects by.
+     *        The objects are first sorted by fieldNames[0], then by fieldNames[1] and so forth.
+     * @return A sorted {@link RealmResults} containing the objects.
+     * @throws java.lang.IllegalArgumentException if a field name does not exist.
+     */
+    @SuppressWarnings("unchecked")
+    public RealmResults<DynamicRealmObject> allObjectsSorted(String className, String fieldNames[],
+                                                                    Sort sortOrders[]) {
```
bmunkholm: indent off...

realm/src/main/java/io/realm/DynamicRealm.java
```diff
@@ -0,0 +1,337 @@
...
+     *
+     * @param className the Class to get objects of.
+     * @param fieldName the field name.
+     * @return immediately an empty {@link RealmResults}. Users need to register a listener
+     * {@link io.realm.RealmResults#addChangeListener(RealmChangeListener)} to be notified
+     * when the query completes.
+     * @throws IllegalArgumentException if a field name does not exist or the field is not indexed.
+     */
+    public RealmResults<DynamicRealmObject> distinctAsync(String className, String fieldName) {
+        if (fieldName == null) {
```
bmunkholm: This code is repeated. Make helper. Actually I would prefer this validation code to be moved to C++ if at all possible.

realm/src/main/java/io/realm/DynamicRealmObject.java
```diff
@@ -28,45 +25,79 @@
...
+    /**
+     * Returns the value for the given field.
+     *
+     * @param fieldName name of the field.
+     * @return the field value. Private types will be converted to their boxed variants.
+     */
+    @SuppressWarnings("unchecked")
+    public <E> E get(String fieldName) {
+        long columnIndex = row.getColumnIndex(fieldName);
+        RealmFieldType type = row.getColumnType(columnIndex);
```
bmunkholm: What if the fieldName doesn't exist?

realm/src/main/java/io/realm/DynamicRealmObject.java
```diff
@@ -265,10 +299,93 @@ public boolean hasField(String fieldName) {
...
+        if (value == null) {
+            row.nullifyLink(row.getColumnIndex(fieldName));
+            // FIXME Add support for other types when Null is merged.
+        } else {
+            setValue(fieldName, value);
+        }
+    }
+
+    // Automatically finds the appropriate setter based on the objects type
+    private void setValue(String fieldName, Object value) {
```
bmunkholm: this looks horribly slow...

realm/src/main/java/io/realm/DynamicRealmObject.java
```diff
@@ -279,8 +396,8 @@ public void setBoolean(String fieldName, boolean value) {
     /**
      * Sets the {@code short} value of the given field.
      *
-     * @param fieldName Field name.
-     * @param value Value to insert.
+     * @param fieldName field name.
```
bmunkholm: Why add the many variants of integers as separate methods?
Did you consider just having one or having an setInteger() maybe with overloaded variants?

realm/src/main/java/io/realm/DynamicRealmObject.java
cmelchior: In the generic `set(fieldName, value)` method, and yes it is slower, but some use cases like reading from CSV files it makes client code much much more easy to write.

realm/src/main/java/io/realm/DynamicRealmObject.java
cmelchior: This was done in order to match the actual java types. It also has symmetry with the getters `getShort`/`getInt`/`getLong`. The idea of only one "integer" type is a core specific implementation detail I would try to keep away from the ObjectStore API if at all possible.

Also overloaded methods would still be the same method count in both autocomplete/JavaDoc, so imo. it would only reduce readability:

```
setInt(int)
setInt(short)
setInt(long)
```
vs.
```
setInt(int)
setShort(short)
setLong(long)
```

PR #1499, opened 2015-09-24
remove unused HashMap instance.

zaki50:


zaki50:
@realm/java please review this.

realm/src/main/java/io/realm/RealmQuery.java
```diff
@@ -73,6 +73,7 @@ public RealmQuery(Realm realm, Class<E> clazz) {
         this.realm = realm;
         this.clazz = clazz;
         this.table = realm.getTable(clazz);
+        view = null;
```
emanuelez: No need since it's the default value for fields extending `Object`. Being explicit about it might be a good idea and will probably be optimized away from the compiler, but still... I don't think this should be here. What does the rest of @realm/java think?

zaki50: There are two reasons to assign null here.

One is to clarify `view` is null as you mentioned.
Another is that it is mandatory in order to make `view` field final.

cmelchior: Personally I like being explicit about these things, so I would just keep it. It is not something I feel strongly about htough. Also It should be `this.view` to follow the naming convention used for the rest of the parameters.

But in general:
```
private boolean bool = false; // I like this better as it is faster to parse mentally.
private boolean bool;  
``

kneth: I tend to agree with @cmelchior

zaki50: Added `this.` to field access(bc80a08)

cmelchior:
Looks good to me :+1:

emanuelez:
:+1:

PR #1501, opened 2015-09-27
Fixed handler not being removed when using setAutoRefresh(false)

cmelchior:
This PR is the result of some bug hunting and turned out some minor issues:

1) If a Looper is explicitly quit (e.g. a IntentService). We risk triggering warnings in the Log with the following wording "Handler sending message to a handler on a dead thread". An additional log message is now added with a description on how to avoid this.

2) If using `realm.setAutoRefresh(false)` the internal handler is not properly cleared. This has now been fixed.

@realm/java

### DIFF
--- a/realm/src/androidTest/java/io/realm/NotificationsTest.java
+++ b/realm/src/androidTest/java/io/realm/NotificationsTest.java
@@ -18,6 +18,7 @@ package io.realm;
 import android.os.Handler;
 import android.os.Looper;
 import android.test.AndroidTestCase;
+import android.util.Log;
 
 import junit.framework.AssertionFailedError;
 
@@ -37,6 +38,8 @@ import java.util.concurrent.atomic.AtomicInteger;
 
 import io.realm.entities.AllTypes;
 import io.realm.entities.Dog;
+import io.realm.internal.log.Logger;
+import io.realm.internal.log.RealmLog;
 
 public class NotificationsTest extends AndroidTestCase {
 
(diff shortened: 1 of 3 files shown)


cmelchior:
Another potential race condition:

*Observations*
1) Handlers are currently stored in a `ConcurrentHashMap`
2) `.close()` or `.setAutoRefresh(false)` will remove a Handler from the Map
3) `.commitTransaction()` will loop all handlers and notify them.
4) Iterating Handlers will be done on a local copy which can be interleaved with add/remove operations.

*Race condition*

1) Thread1: `Realm realm = Realm.getDefaultInstance()`
2) Thread2: `Service.stop()` AND a `realm.commitTransaction()`
3) Thread1: realm.setAutoRefresh(false)

Step 2) Will still trigger the warning

*Solution*
Only solution (I can see) is to revert back to a normal HashMap and synchronise on some external lock. Performance characteristics are unknown: On the upside HashMap is faster than ConcurrentHashMap but on the downside we will lock while looping all handlers in `.commitTransaction`.

Downside of not fixing this is that users will occasionally see the "Handler sending message to a handler on a dead thread" with no way of fixing it, but apart from that app will work as normal.

realm/src/main/java/io/realm/BaseRealm.java
```diff
@@ -303,7 +304,10 @@ public void commitTransaction() {
                             && !handler.hasMessages(REALM_CHANGED)       // The right message
                             && handler.getLooper().getThread().isAlive() // The receiving thread is alive
                     ) {
-                handler.sendEmptyMessage(REALM_CHANGED);
+                if (!handler.sendEmptyMessage(REALM_CHANGED)) {
+                    RealmLog.w("Cannot update Looper threads when the Looper has quit. Use realm.setAutoRefresh(false)" +
```
emanuelez: There's not space between `(false)`and `to`

emanuelez:
Apart from comment :+1:

realm/src/androidTest/java/io/realm/NotificationsTest.java
```diff
@@ -557,4 +561,50 @@ public void onChange() {
...
+                Looper.prepare(); // Fake background thread with a looper, eg. a IntentService
+
+                Realm realm = Realm.getInstance(getContext());
+                realm.setAutoRefresh(false);
+                Looper.myLooper().quit();
+                backgroundLooperStartedAndStopped.countDown();
+                try {
+                    mainThreadCommitCompleted.await();
+                } catch (InterruptedException e) {
+                    fail();
```
nhachicha: this fail will not be reported by the test runner

realm/src/androidTest/java/io/realm/TestHelper.java
```diff
@@ -91,6 +97,73 @@ public static void prepareDatabaseFromAssets(Context context, String realmPath,
...
+     *
+     * @param failureLevel {@link Log} level from which the unit test will fail.
+     * @return Logger implementation
+     */
+    public static Logger getFailureLogger(final int failureLevel) {
+        return new Logger() {
+
+            private void failIfEqualOrAbove(int logLevel, int failureLevel) {
+                if (logLevel >= failureLevel) {
+                    fail("Message logged that was above valid level: " + logLevel + " >= " + failureLevel);
```
nhachicha: I think throwing exception is better at this level. One can call this method inside a thread that doesn't catch `AssertionError` & get nothing reported, whereas a checked exception force the caller to handle it.

cmelchior: Good point

cmelchior: Ah you mean CheckedException. That is really not possible from the Logger as that is the standard interface used by the RealmLog. In that case using `fail` is as good as a `RuntimeException`?

nhachicha:
:+1:

PR #1502, opened 2015-09-27
Guard globalPathConfigurationCache properly.

cmelchior:
We had a bug reported against current master:

```
88    09-26 22:15:09.497 E/ACRA ( 438): ACRA caught a ArrayIndexOutOfBoundsException for com.podfather.podfather
89    09-26 22:15:09.497 E/ACRA ( 438): java.lang.ArrayIndexOutOfBoundsException: length=0; index=0
90    09-26 22:15:09.497 E/ACRA ( 438): at java.util.concurrent.CopyOnWriteArrayList.get(CopyOnWriteArrayList.java:117)
91    09-26 22:15:09.497 E/ACRA ( 438): at io.realm.BaseRealm.validateAgainstExistingConfigurations(BaseRealm.java:482)
92    09-26 22:15:09.497 E/ACRA ( 438): at io.realm.Realm.createAndValidate(Realm.java:276)
93    09-26 22:15:09.497 E/ACRA ( 438): at io.realm.Realm.create(Realm.java:245)
94    09-26 22:15:09.497 E/ACRA ( 438): at io.realm.Realm.getDefaultInstance(Realm.java:200)
95    09-26 22:15:09.497 E/ACRA ( 438): at com.podfather.podfather.services.DataPostService.postPodImages(DataPostService.java:626)
96    09-26 22:15:09.497 E/ACRA ( 438): at com.podfather.podfather.services.DataPostService.onHandleIntent(DataPostService.java:86)
97    09-26 22:15:09.497 E/ACRA ( 438): at android.app.IntentService$ServiceHandler.handleMessage(IntentService.java:65)
98    09-26 22:15:09.497 E/ACRA ( 438): at android.os.Handler.dispatchMessage(Handler.java:102)
99    09-26 22:15:09.497 E/ACRA ( 438): at android.os.Looper.loop(Looper.java:135)
100    09-26 22:15:09.497 E/ACRA ( 438): at android.os.HandlerThread.run(HandlerThread.java:61)
```

It could happen because `releaseFileReference` was only synchronized against the instance of a Realm and not the `BaseRealm` class which other methods use when they access `globalPathConfigurationCache`. This can lead to a race condition where `validateAgainstExisitingConfiguration` will wrongly assume it has sole access which is wrong since `releaseFileReference` don't respect the lock on `BaseRealm`.

This PR changes the lock from the instance of Realm to the static `BaseRealm` instead.

@realm/java

### DIFF
--- a/realm/src/main/java/io/realm/BaseRealm.java
+++ b/realm/src/main/java/io/realm/BaseRealm.java
@@ -400,7 +400,7 @@ abstract class BaseRealm implements Closeable {
             lastLocalInstanceClosed();
             sharedGroupManager.close();
             sharedGroupManager = null;
-            releaseFileReference();
+            releaseFileReference(configuration);
         }
 
         int refCount = references - 1;
@@ -427,7 +427,7 @@ abstract class BaseRealm implements Closeable {
     /**
      * Acquire a reference to the given Realm file.
      */
-    protected synchronized void acquireFileReference(RealmConfiguration configuration) {
+    static synchronized void acquireFileReference(RealmConfiguration configuration) {
         String path = configuration.getPath();
         Integer refCount = globalRealmFileReferenceCounter.get(path);
         if (refCount == null) {
@@ -440,7 +440,7 @@ abstract class BaseRealm implements Closeable {
      * Releases a reference to the Realm file. If reference count reaches 0 any cached configurations
      * will be removed.
      */
-    protected synchronized void releaseFileReference() {
+    static synchronized void releaseFileReference(RealmConfiguration configuration) {
         String canonicalPath = configuration.getPath();
         List<RealmConfiguration> pathConfigurationCache = globalPathConfigurationCache.get(canonicalPath);
         pathConfigurationCache.remove(configuration);
@@ -471,7 +471,7 @@ abstract class BaseRealm implements Closeable {
      *
      * @throws IllegalArgumentException If the new configuration isn't valid.
      */
-    protected static void validateAgainstExistingConfigurations(RealmConfiguration newConfiguration) {
+    protected static synchronized void validateAgainstExistingConfigurations(RealmConfiguration newConfiguration) {
 
         String realmPath = newConfiguration.getPath();
         List<RealmConfiguration> pathConfigurationCache = globalPathConfigurationCache.get(realmPath);


kneth:
:+1:

emanuelez:
No unit test to exposes the problem?

realm/src/main/java/io/realm/BaseRealm.java
```diff
@@ -436,19 +436,21 @@ protected synchronized void acquireFileReference(RealmConfiguration configuratio
...
+            List<RealmConfiguration> pathConfigurationCache = globalPathConfigurationCache.get(canonicalPath);
+            pathConfigurationCache.remove(configuration);
+            if (pathConfigurationCache.isEmpty()) {
+                globalPathConfigurationCache.remove(canonicalPath);
+            }
 
-        Integer refCount = globalRealmFileReferenceCounter.get(canonicalPath);
-        if (refCount == null || refCount == 0) {
-            throw new IllegalStateException("Trying to release a Realm file that is already closed");
+            Integer refCount = globalRealmFileReferenceCounter.get(canonicalPath);
```
nhachicha: there is also this with `globalRealmFileReferenceCounter` since it's protected in `acquireFileReference` by the instance lock that access a static shared map

cmelchior: Actually `acquireFileReference` is only called from `createAndValidate` which has a lock on `BaseRealm.class` so right now it is safe. 

I could see why we want to have it inside this method as well though.

In general I think we should take a deep hard look at how we do locking currently. We have a mix of Concurrent* and locks that are very hard to reason about. So while it might be correct now, the chance of future bugs are very high, and it is impossible to write unit tests for.

I think it would probably to make the locking more coarse-grained vs. the ability to actually reason about it, but that should be another issue.

cmelchior:
@emanuelez It is not possible to create a unit test for this as it require a specific sequence of events that is not possible to unit test.

1) Thread 1 opens a Realm and calls `validateAgainstExistingConfigurations`
2) Thread 1: `validateAgainstExistingConfigurations` calls `globalPathConfigurationCache.get(realmPath)` and begins iterating
3) Thread 2 calls `.close()` and removes it's path from `globalPathConfigurationCache`
4) Thread 1 crashes when trying to do `pathConfigurationCache.get(0)` with a IndexOutOfBounds.

realm/src/main/java/io/realm/BaseRealm.java
nhachicha: But `deleteRealm` call `isFileOpen` that uses an instance lock not `BaseRealm.class`?

cmelchior: You are right. Sorry, misread your first comment

cmelchior: @nhachicha All static variables should now be guarded with a BaseRealm lock. Good catch.

cmelchior:
Bump @realm/java

beeender:
:+1:

dalinaum:
:+1:

Issue #1507, opened 2015-09-29
Date query doesn't work with date greater than 2038-01-19


nartex:
All date queries (lessThan, lessThanOrEqual, between,...) doesn't work with date greater than Tue Jan 19 2038 04:14:07 GMT+0100 (CET) (Timestamp : 2147483647000).

Many "/1000" in TableQuery causing Integer cast.

I will make a pull request in few time.

zaki50:
If `long / int` in Java code, `int` is promoted to `long`.

It seems that this problem comes from `static_cast<time_t>(value)` in jni code.

kneth:
Good old `time_t` :-( As `time_t` is a typedef of `long`, it will be a 32 bit integer on 32 bit architectures, and 64 bit integer on a 64 bit architecture. Realm's underlying storage engine is using `time_t` in its constructor of date objects, and we are force to use only 32 bit values in order to have cross-architecture compatibility.

kneth:
By the way, the limitation is documented in https://realm.io/docs/java/latest/#general (item 3).

bmunkholm:
We also expect to provide a better datetime format with enhanced range and accuracy.

zaki50:
@kneth @bmunkholm  That limitation is not described in cocoa/swift document.
Is that limitation really from underlying storage engine?

nartex:
@zaki50 Thank you, I have always believed the opposite...
@kneth My bad, I deserve to fully read the doc, again :-).

zaki50:
In https://github.com/realm/realm-cocoa/issues/2292#issuecomment-124780244
@segiddins showed that NSDate.distantPast() and NSDate.distantFuture() can be stored into Realm cocoa.

kneth:
We are able to expand the date range - see #1510 to track progress.

zaki50:
great news!

zaki50:
I made PR #1511

cmelchior:
As the issue is being tracked in #1510 and solved in #1511I will close this ticket here.

Issue #833, opened 2015-02-06
Date precisision not precise enough (dropps all milliseconds)


falkorichter:
Java Dates have a millisecond presicsion. Realm queries on dates should respect this precision.

See the example project:
https://github.com/falkorichter/realm-date-fail

The test class in question:
https://github.com/falkorichter/realm-date-fail/blob/master/app/src/androidTest/java/com/sensorberg/android/realmdatecomparison/ApplicationTest.java

The tests shows that with 999 milliseconds difference, realm fails to query correctly.

See the status of this test at: https://travis-ci.org/falkorichter/realm-date-fail.svg?branch=master

cmelchior:
Hi @falkorichter 
Thank you for the testcase. It is a known issue however, that is also documented in our current limitations: http://realm.io/docs/java/0.78.0/#general

The reason is that we want Realms to function cross-platform eg. NSDate uses floating points internally. This is why we currently only support second-precision (it is the same on iOS), but we definitely want to fix it going forward, it is just not a trivial thing to do.

For now, a workaround would be to store your dates as long and convert them back to Date when needed.

falkorichter:
@cmelchior why not drop the Date support completely on Android as it is crippled? Bring it back, when it is working.

virl:
Please fix this. Really annoying to write all the code to use Dates only to discover that precision is not enough and rewrite everything to use long.

esco:
:+1:

senorcris:
:+1:

PR #1508, opened 2015-09-29
Fix issue #1507

nartex:


cmelchior:
Hi @nartex
Thank you for this. It doesn't really fix the underlying problem though as you can see from the comments in your issue.. If you want to store dates further in the future I would suggest saving it as a long instead and do the conversion yourself.

That said we do have plans for providing Date fully in the future #833.

PR #1524, opened 2015-10-03
Synchronize access to Realm.defaultConfig field.

zaki50:
Problem:
access to `Realm.defaultConfig` field is not synchronized.

This PR adds a code to synchronize `Realm.defaultConfig` field access.

### DIFF
--- a/realm/src/main/java/io/realm/Realm.java
+++ b/realm/src/main/java/io/realm/Realm.java
@@ -194,10 +194,14 @@ public final class Realm extends BaseRealm {
      * model classes or version has has changed so a migration is required.
      */
     public static Realm getDefaultInstance() {
-        if (defaultConfiguration == null) {
+        final RealmConfiguration configuration;
+        synchronized (Realm.class) {
+            configuration = defaultConfiguration;
+        }
+        if (configuration == null) {
             throw new NullPointerException("No default RealmConfiguration was found. Call setDefaultConfiguration() first");
         }
-        return create(defaultConfiguration);
+        return create(configuration);
     }
 
     /**
@@ -227,7 +231,9 @@ public final class Realm extends BaseRealm {
         if (configuration == null) {
             throw new IllegalArgumentException("A non-null RealmConfiguration must be provided");
         }
-        defaultConfiguration = configuration;
+        synchronized (Realm.class) {
+            defaultConfiguration = configuration;
+        }
     }
 
     /**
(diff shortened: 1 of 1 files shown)


zaki50:
@realm/java review this please.

cmelchior:
While technically you are correct. I am not sure if it is worth adding the complexity. At worst you will get a NullPointerException if there is a race condition between `setDefaultConfiguration` and `getDefaultInstance` which really shouldn't happen.

But if the others think it is fine I won't object.

zaki50:

dalinaum:
It looks good to me.

realm/src/main/java/io/realm/Realm.java
```diff
@@ -194,10 +194,14 @@ public static Realm getInstance(Context context) {
      * model classes or version has has changed so a migration is required.
      */
     public static Realm getDefaultInstance() {
-        if (defaultConfiguration == null) {
+        final RealmConfiguration configuration;
+        synchronized (Realm.class) {
+            configuration = defaultConfiguration;
+        }
```
beeender: If the `create(configuration)` is not in the `synchronized` block, i don't really think this fix has too much difference compared with before. Since before calling create, there is still a chance for `setDefaultConfiguration` to change the default configuration. And a reference to the old config will be passed to the `create`. And this would be the exact same behaviour as before.

We might need to think about what do we want to achieve here? Do we want to allow user to call setDefaultConfiguration twice?

cmelchior: I would say yes. Calling `Realm.setDefaultConfiguration()` should only be called from very few places: either in the `Application` on startup or some RealmHelper wrapper. Prevent calling `setDefaultConfigration()` twice would imply we need to add a `removeDefaultConfiguration()` as well, and TBH I cannot see any scenario where you would have a race condition on that method.

zaki50: @cmelchior Hm... As you said, I don't have any real scenario.

Is it enough just to introduce local variable to avoid accidental NPE in `create()`?

```
    public static Realm getDefaultInstance() {
        final RealmConfiguration configuration = defaultConfiguration;
        if (configuration == null) {
            throw new NullPointerException("No default RealmConfiguration was found. Call setDefaultConfiguration() first");
        }
        return create(configuration);
    }
```

cmelchior: To be honest, I don't see this scenario happen either. Currently, in the worst case, we end up calling create twice (which is synchronized, so second time would hit the cache), but this is extremely unlikely. So in this case I don't think it is worth adding the code complexity.

zaki50: @cmelchior you are right.
I cancel this PR.

PR #1529, opened 2015-10-05
Add unit test for compatibility with iOS NSDate

cmelchior:
It is currently failing for some reason. 

For some reason the Realm-Cocoa date seems to be Epoch when DBL_MAX and -DBL_MAX is used to construct the NSDate. Possible hypothesis is either a mistake in my unit test (most likely) or Realm-Cocoa just saving epoch if a unsupported date is added. I haven't been able to find the source to verify that.

In any case. We need to figure out what the common subset of supported dates are. Add tests for it, and throw exceptions on all other dates.


stk1m1:
Quick search reveals that **DBL_MIN** and **DBL_MAX** are quite the large numbers.
(You can take a peek at [here](http://www.opensource.apple.com/source/Libc/Libc-498/include/float.h) and [here](http://www.opensource.apple.com/source/gcc/gcc-934.3/float.h). They are quite old, but I *think* constants like those would not change much over the time.)

I believe more suitable range would be [`[NSDate distantFuture]`](https://developer.apple.com/library/mac/documentation/Cocoa/Reference/Foundation/Classes/NSDate_Class/#//apple_ref/occ/clm/NSDate/distantFuture) and [`[NSDate distantPast]`](https://developer.apple.com/library/mac/documentation/Cocoa/Reference/Foundation/Classes/NSDate_Class/#//apple_ref/occ/clm/NSDate/distantPast), which would be `4000-01-01` and `0000-00-00` respectively.

stk1m1:
**`NSDate` <-> `Date`**

- What's happening.  
   1. **NSDate** 
   [`NSDate`](https://developer.apple.com/library/mac/documentation/Cocoa/Reference/Foundation/Classes/NSDate_Class/#//apple_ref/occ/clm/NSDate/dateWithTimeIntervalSinceReferenceDate:) is a [`Toll-Free Bridged`](https://developer.apple.com/library/ios/documentation/General/Conceptual/CocoaEncyclopedia/Toll-FreeBridgin/Toll-FreeBridgin.html) of [`CFDate`](https://developer.apple.com/library/mac/documentation/CoreFoundation/Reference/CFDateRef/index.html#//apple_ref/c/tdef/CFDateRef), and has [`CFAbsoluteTime _time`](https://github.com/opensource-apple/CF/blob/master/CFDate.c#L103) field in `__CFDate` where a time value gets stored. (`CFAbsoluteTime` -> `CFTimeInterval` -> `double`).  

   2. **From `double` to `long long`** 
   When [`RLMDateTimeForNSDate`](https://github.com/realm/realm-cocoa/blob/master/Realm/RLMUtil.hpp#L164) calls `-[NSDate dateWithTimeIntervalSince1970:]`, [`CFDateGetTimeIntervalSinceDate`](https://github.com/opensource-apple/CF/blob/master/CFDate.c#L186)(most likely) returns the aformentioned `_time`, compensated with a reference point. The returned value then gets converted to `long long` in `static_cast<long long>`. There is no range check or anything in `CFDateGetTimeIntervalSinceDate`.  

   3. **`static_cast<long long>`** 
	To my best knowledge, `static_cast<long long>` eventually reduces to [`CVTTSD2SI`](http://www.jaist.ac.jp/iscenter-new/mpc/altix/altixdata/opt/intel/vtune/doc/users_guide/mergedProjects/analyzer_ec/mergedProjects/reference_olh/mergedProjects/instructions/instruct32_hh/vc67.htm) in x86, and this is what `CVTTSD2SI` does. (I wasn't able to generate ARM version assembly with XCode like good'ol days. I can use some help.)  
_"If a converted result is larger than the maximum signed doubleword integer, the floating-point invalid exception is raised, and if this exception is masked, the indefinite integer value (80000000H) is returned."_  
	It is therefore accurate observation that realm-cocoa returns Epoch after a certain limit. Once `double` is statically converted, it is hard to recover the original value, and difficult to judge whether a certin value such as `DBL_MAX` is over such limit across heterogeneous-platforms.  
Here is a little experiment.  
```
Original Value of DBL_MAX : 1.798e+308

Statically casted 64bit int : 0x8000000000000000
Mem-copied 64bit int : 0x7fefffffffffffff
Reinterpret casted 64bit int : 0x7fefffffffffffff

Statically restored DBL_MAX : -9.223e+18
Mem-copied-back DBL_MAX : 1.798e+308
Reinterpret_cast DateValue : 1.798e+308
```


- How we can provide inter-operability.  
	For the good of Realm, I strongly this issue need to be solved in an inter-platform manner. Here are three options I could muster up.  
    1. **Use `memcpy` or `reinterpret_cast` to copy bit-by-bit between `double` and `long long`**. This doesn't convert anything but literally copies values bit-by-bit to store in `.realm`. This could provide, in the best scenario, a chance to gaurantee an end-to-end inter-operability between platforms. i.e. whatever goes to `NSDate` comes out in `Date` and vice versa. The implementation cost _is_ horrible because we need to check endianness, NaN, and etc.

    2. **Set REALM specific common date & time range across platform**, (e.g. REALM_DATE_MAX, REALM_DATE_MIN), to ensure users does not input values larger than what can be transferred across platforms. With the range, date conversion should check input values such that DBL_MAX should be rejected from whatever platform.
    
    3. **Utilize date-time data type constructor** which is already defined. (You know where to look into.) There is a constructor that asks us to break down a time-stamp to year, month, date, hour, miniutes, and seconds to build a date-time instance. Combined with the opt. 2, this will ensure the best compatilibilty and interoperability in my opinion. We can still retain the current mothod of spawing a platform specific date-time instance with `long long`, but I think we strongly need to respect the aformentioned constructor that can provide interoperability without too much implementation and maintainance cost.

stk1m1:
**What is done**
We can now check `NSDate`'s maximum value and minimum value. 

When `DBL_MAX` from `-[NSDate timeIntervalSince1970]` is saved, it is truncated as explained above, and `0x8000000000000000L` is saved as a date. When the date is reinstated in Java, The value gets multiplied by 1000 as `Date.getTime()` represent datetime in [millisecond](https://docs.oracle.com/javase/7/docs/api/java/util/Date.html#getTime()). `0x8000000000000000L` is pretty large number that would eventually become zero if you multiply 1000.

Contrary to `DBL_MAX`, `-DBL_MAX` gets truncated to 0x3a4fc880L and it becomes 0x3a4fc880L * 1000 as  `Date.getTime()` represent time in milliseconds.

stk1m1:
retest this please.

stk1m1:
retest this please.

stk1m1:
retest this please.

realm/realm-library/src/androidTest/assets/ios/README.md
```diff
@@ -105,46 +145,53 @@ RLM_ARRAY_TYPE(AllTypes)
...
+    obj.longLongCol = std::numeric_limits<int64_t>::min();
+    obj.floatCol = -FLT_MAX;
+    obj.doubleCol = -DBL_MAX;
     obj.byteCol = [NSData dataWithBytes:no_bytes length:sizeof(no_bytes)];
     obj.stringCol = @"";
-    obj.dateCol = [NSDate dateWithTimeIntervalSince1970: 0];
+    obj.dateCol = [NSDate dateWithTimeIntervalSinceReferenceDate:-DBL_MAX];
     [realm addObject:obj];
     [realm commitWriteTransaction];
+    NSLog(@"%@, obj.longLongCol : 0x%llx\n",[NSString stringWithFormat:@"%@-alltypes-min.realm", version], obj.longLongCol);
```
cmelchior: Debug info that could be removed?

realm/realm-library/src/androidTest/assets/ios/README.md
```diff
@@ -105,46 +145,53 @@ RLM_ARRAY_TYPE(AllTypes)
...
+    obj.longLongCol = std::numeric_limits<int64_t>::max();
     obj.floatCol = FLT_MAX;
     obj.doubleCol = DBL_MAX;
     obj.byteCol = [NSData dataWithBytes:no_bytes length:sizeof(no_bytes)];
     obj.stringCol = @"";
-    obj.dateCol = [NSDate dateWithTimeIntervalSince1970: LONG_MAX];
+    obj.dateCol = [NSDate dateWithTimeIntervalSinceReferenceDate:DBL_MAX];
     [realm addObject:obj];
     [realm commitWriteTransaction];
+    NSLog(@"%@, obj.longLongCol : 0x%llx\n",[NSString stringWithFormat:@"%@-alltypes-max.realm", version], obj.longLongCol);
```
cmelchior: Same ?

realm/realm-library/src/androidTest/assets/ios/README.md
```diff
@@ -105,46 +145,53 @@ RLM_ARRAY_TYPE(AllTypes)
...
-    [AppDelegate setDefaultRealm: [NSString stringWithFormat:@"%@-alltypes-min.realm", version]];
-    realm = [RLMRealm defaultRealm];
+    realm = [AppDelegate appDefaultRealm:[NSString stringWithFormat:@"%@-alltypes-min.realm", version]];
     [realm beginWriteTransaction];
     obj = [[IOSAllTypes alloc] init];
     obj.boolCol = FALSE;
     obj.shortCol = SHRT_MIN;
     obj.intCol = INT_MIN;
     obj.longCol = LONG_MIN;
-    obj.floatCol = FLT_MIN;
```
cmelchior: Ups, nice catch

realm/realm-library/src/androidTest/java/io/realm/IOSRealmTests.java
```diff
@@ -61,6 +61,12 @@ public void setUp() {
         RealmConfiguration defaultConfiguration = configFactory.createConfigurationBuilder()
                 .name(REALM_NAME)
                 .schema(IOSAllTypes.class, IOSChild.class)
+                .schemaVersion(1)
```
cmelchior: just use `deleteRealmIfMigrationNeeded()` instead.

cmelchior:
Apart from a few comments :+1:

realm/realm-library/src/androidTest/assets/ios/README.md
stk1m1: I'd suggest to leave debug output until the new datetime is released. Without one, it is very natural to expect to get the exactly the same value we put in. _We_ know that's not the case but still...

cmelchior: :+1:

stk1m1:
@realm/java review please.

realm/realm-library/src/androidTest/assets/ios/README.md
```diff
@@ -1,11 +1,33 @@
 This folder contains various Realm databases created on iOS and can be used to test interop with
-Realm-Android.
+Realm-Android. The databases are generated using the below iOS code.
```
beeender: `are` -> `is`

realm/realm-library/src/androidTest/assets/ios/README.md
```diff
@@ -1,11 +1,33 @@
...
 
-The databases are generated using the following iOS code:
+(02/10-2015) Note that debugging must be disabled as it is currently not possible to debug encrypted
```
beeender: Uh, I think we don't have this limitation anymore after remove signal handler for encryption? Worth to check with cocoa team.

stk1m1: I'll go check it out.

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@
 * Fixed bug when multiple calls of RealmResults.distinct() causes to return wrong results (#2198).
 * RealmResults.size() now returns Integer.MAX_VALUE when actual size is greater than Integer.MAX_VALUE (#2129).
 * Added RealmQuery.distinctAsync() and RealmResults.distinctAsync() (#2118).
+* Added Unit tests for compatibility with iOS NSDate (#1529).
```
cmelchior: No need for a changelog update. This is an internal change only

realm/realm-library/src/androidTest/assets/ios/README.md
cmelchior: `are` is accurate here, since it is multiple databases.

cmelchior:
Apart from comments :+1:

realm/realm-library/src/androidTest/java/io/realm/entities/IOSAllTypes.java
```diff
@@ -31,14 +31,12 @@
...
-    private long longCol;
+    private int longCol;
+    private long longLongCol;
     private float floatCol;
     private double doubleCol;
-    @Required
     private byte[] byteCol = new byte[0];
-    @Required
     private String stringCol = "";
-    @Required
```
beeender: Why do we remove `@Require`? Shall we test `null` compatibility as well?

stk1m1: @beeender I just realized that there is no `null` test for blob and string. (... aren't they same anyway?) I'll add those test first thing in the morning.

dalinaum: I just read all changes. I will review again after you will add null tests for blog and string.

stk1m1:
retest this please.

stk1m1:
`null` value test is added.

@realm/java

beeender:
:+1:

stk1m1:
retest this please

PR #1530, opened 2015-10-05
Updated Javadoc for getInstance(context)

cmelchior:
In order to make more clear the purpose of this method.

@realm/java @bmunkholm @timanglade

### DIFF
--- a/realm/src/main/java/io/realm/Realm.java
+++ b/realm/src/main/java/io/realm/Realm.java
@@ -72,7 +72,7 @@ import io.realm.internal.log.RealmLog;
  * Realm instances cannot be used across different threads. This means that you have to open an
  * instance on each thread you want to use Realm. Realm instances are cached automatically per
  * thread using reference counting, so as long as the reference count doesn't reach zero, calling
- * {@link #getInstance(android.content.Context)} will just return the cached Realm and should be
+ * {@link #getInstance(RealmConfiguration)} will just return the cached Realm and should be
  * considered a lightweight operation.
  * <p>
  * For the UI thread this means that opening and closing Realms should occur in either
@@ -163,19 +163,20 @@ public final class Realm extends BaseRealm {
     }
 
     /**
-     * Realm static constructor for the default Realm "default.realm".
-     * {@link #close()} must be called when you are done using the Realm instance.
-     * <p>
-     * It sets auto-refresh on if the current thread has a Looper, off otherwise.
-     *
+     * Realm static constructor for the default Realm file {@value io.realm.RealmConfiguration#DEFAULT_REALM_NAME}.
      * This is equivalent to calling {@code Realm.getInstance(new RealmConfiguration(getContext()).build()) }.
+     *
+     * This constructor is only provided for convenience. It is recommended to use
+     * {@link #getInstance(RealmConfiguration)} or {@link #getDefaultInstance()}.
+     *
+     * @param context a non-null Android {@link android.content.Context}
+     * @return an instance of the Realm class.
 
-     * @param context an Android {@link android.content.Context}
-     * @return an instance of the Realm class
-     * @throws RealmMigrationNeededException The model classes have been changed and the Realm
-     *                                       must be migrated
-     * @throws RealmIOException              Error when accessing underlying file
-     * @throws RealmException                Other errors
+     * @throws java.lang.IllegalArgumentException if no {@link Context} is provided.
+     * @throws RealmMigrationNeededException if the model classes no longer match the underlying Realm
+     *                                       and it must be migrated.
+     * @throws RealmIOException              if an error happened when accessing the underlying Realm
+     *                                       file.
      */
     public static Realm getInstance(Context context) {
         return Realm.getInstance(new RealmConfiguration.Builder(context)
@@ -187,7 +188,7 @@ public final class Realm extends BaseRealm {
      * Realm static constructor that returns the Realm instance defined by the {@link io.realm.RealmConfiguration} set
      * by {@link #setDefaultConfiguration(RealmConfiguration)}
      *
-     * @return an instance of the Realm class
+     * @return an instance of the Realm class.
      *
      * @throws java.lang.NullPointerException If no default configuration has been defined.
      * @throws RealmMigrationNeededException If no migration has been provided by the default configuration and the


realm/src/main/java/io/realm/Realm.java
```diff
@@ -163,19 +163,18 @@ protected void finalize() throws Throwable {
     }
 
     /**
-     * Realm static constructor for the default Realm "default.realm".
-     * {@link #close()} must be called when you are done using the Realm instance.
-     * <p>
-     * It sets auto-refresh on if the current thread has a Looper, off otherwise.
-     *
+     * Realm static constructor for the default Realm file "default.realm".
```
zaki50: "default.realm" can be {@value #DEFAULT_REALM_NAME}

emanuelez: Nice!

cmelchior: Didn't know that was possible. Nice

realm/src/main/java/io/realm/Realm.java
```diff
@@ -163,19 +163,18 @@ protected void finalize() throws Throwable {
...
+     *
+     * This constructor is only provided for convenience. It is recommend to use
+     * {@link #getInstance(RealmConfiguration)} or {@link #getDefaultInstance()} instead.
+     *
+     * @param context An Android {@link android.content.Context}
+     * @return An instance of the Realm class.
+     * @throws RealmMigrationNeededException If the model classes no longer match the underlying Realm
+     *                                       and it must be migrated.
+     * @throws RealmIOException              If an error happened when accessing the underlying Realm
+     *                                       file.
```
zaki50: Is there any criteria to mention runtime exception in Javadoc?
Some methods mention `IllegalArgumentException`/`NullPointerException`, others are not.

This method will throw `IllegalArgumentException` if `context` is `null`.

cmelchior: Not anything formalised. But you are right, in this case it should probably list `IllegalArgumentException` as well.

In general it would be nice to be able to move to using `@NotNull`, `@Nullable` annotations, but unfortunately it is not easy to add those if we at some point want the same codebase for both Android and JVM Java

realm/src/main/java/io/realm/Realm.java
cmelchior: Unfortunately It doesn't seem to work with the Gradle javadoc generator for some reason :(

zaki50: My build environment(of course Gradle) can generate Javadoc as expected.
What's the differences...

I have no idea because `[@value foo}` is not a new notation...

Can I see the error message?

zaki50:
:+1:

realm/src/main/java/io/realm/Realm.java
dalinaum: I can generate Javadoc correctly too.

cmelchior:
@zaki50 @dalinaum I put in the wrong path.  `{@value RealmConfiguration#REALM_DEFAULT_NAME}` didn't work while `{@value io.relam.RealmConfiguration#REALM_DEFAULT_NAME}` did

zaki50:
@cmelchior How about `{@value #REALM_DEFAULT_NAME}`?
Realm class has `REALM_DEFAULT_NAME` as its private constant.

But both `{@value #REALM_DEFAULT_NAME}` and `{@value io.relam.RealmConfiguration#REALM_DEFAULT_NAME}` must be same result.

cmelchior:
Realm.DEFAULT_REALM_NAME will be deprecated with the Dynamic API as the same default is used for both DynamicRealm and Realm, so `RealmConfiguration.DEFAULT_REALM_NAME` is actually the right location

realm/src/main/java/io/realm/Realm.java
emanuelez: Gradle has nothing to do with it. It simply calls the JDK facilities to generate Javadocs.

realm/src/main/java/io/realm/Realm.java
```diff
@@ -163,19 +163,20 @@ protected void finalize() throws Throwable {
...
-     * @return an instance of the Realm class
-     * @throws RealmMigrationNeededException The model classes have been changed and the Realm
-     *                                       must be migrated
-     * @throws RealmIOException              Error when accessing underlying file
-     * @throws RealmException                Other errors
+     *
+     * This constructor is only provided for convenience. It is recommend to use
+     * {@link #getInstance(RealmConfiguration)} or {@link #getDefaultInstance()} instead.
+     *
+     * @param context An non-null Android {@link android.content.Context}
```
beeender: "An" -> "an"

realm/src/main/java/io/realm/Realm.java
```diff
@@ -163,19 +163,20 @@ protected void finalize() throws Throwable {
...
      * This is equivalent to calling {@code Realm.getInstance(new RealmConfiguration(getContext()).build()) }.
-
-     * @param context an Android {@link android.content.Context}
-     * @return an instance of the Realm class
-     * @throws RealmMigrationNeededException The model classes have been changed and the Realm
-     *                                       must be migrated
-     * @throws RealmIOException              Error when accessing underlying file
-     * @throws RealmException                Other errors
+     *
+     * This constructor is only provided for convenience. It is recommend to use
```
beeender: "recommend" -> "recommended"

realm/src/main/java/io/realm/Realm.java
```diff
@@ -163,19 +163,20 @@ protected void finalize() throws Throwable {
...
+     * {@link #getInstance(RealmConfiguration)} or {@link #getDefaultInstance()} instead.
+     *
+     * @param context An non-null Android {@link android.content.Context}
+     * @return An instance of the Realm class.
```
beeender: "An" -> "an"

realm/src/main/java/io/realm/Realm.java
```diff
@@ -163,19 +163,20 @@ protected void finalize() throws Throwable {
...
+
+     * @throws java.lang.IllegalArgumentException If no {@link Context} is provided.
```
beeender: "If" -> "if"

realm/src/main/java/io/realm/Realm.java
```diff
@@ -163,19 +163,20 @@ protected void finalize() throws Throwable {
...
+     * @throws RealmMigrationNeededException If the model classes no longer match the underlying Realm
```
beeender: "If" -> "if"

realm/src/main/java/io/realm/Realm.java
```diff
@@ -163,19 +163,20 @@ protected void finalize() throws Throwable {
...
+     *                                       and it must be migrated.
+     * @throws RealmIOException              If an error happened when accessing the underlying Realm
```
beeender: "If" -> "if"
And not sure, should "an error happened" -> "an error happens"?

realm/src/main/java/io/realm/Realm.java
```diff
@@ -187,7 +188,7 @@ public static Realm getInstance(Context context) {
      * Realm static constructor that returns the Realm instance defined by the {@link io.realm.RealmConfiguration} set
      * by {@link #setDefaultConfiguration(RealmConfiguration)}
      *
-     * @return an instance of the Realm class
+     * @return An instance of the Realm class.
```
beeender: "An" -> "an"

realm/src/main/java/io/realm/Realm.java
```diff
@@ -72,7 +72,7 @@
  * Realm instances cannot be used across different threads. This means that you have to open an
  * instance on each thread you want to use Realm. Realm instances are cached automatically per
  * thread using reference counting, so as long as the reference count doesn't reach zero, calling
- * {@link #getInstance(android.content.Context)} will just return the cached Realm and should be
+ * {@link #getInstance(RealmConfiguration)}get will just return the cached Realm and should be
```
beeender: delete "get"?

beeender:
:+1:

realm/src/main/java/io/realm/Realm.java
```diff
@@ -163,19 +163,20 @@ protected void finalize() throws Throwable {
...
-     * @return an instance of the Realm class
-     * @throws RealmMigrationNeededException The model classes have been changed and the Realm
-     *                                       must be migrated
-     * @throws RealmIOException              Error when accessing underlying file
-     * @throws RealmException                Other errors
+     *
+     * This constructor is only provided for convenience. It is recommended to use
+     * {@link #getInstance(RealmConfiguration)} or {@link #getDefaultInstance()} instead.
+     *
+     * @param context an non-null Android {@link android.content.Context}
```
kneth: "an" -> "a"

realm/src/main/java/io/realm/Realm.java
```diff
@@ -163,19 +163,20 @@ protected void finalize() throws Throwable {
...
-
-     * @param context an Android {@link android.content.Context}
-     * @return an instance of the Realm class
-     * @throws RealmMigrationNeededException The model classes have been changed and the Realm
-     *                                       must be migrated
-     * @throws RealmIOException              Error when accessing underlying file
-     * @throws RealmException                Other errors
+     *
+     * This constructor is only provided for convenience. It is recommended to use
+     * {@link #getInstance(RealmConfiguration)} or {@link #getDefaultInstance()} instead.
```
kneth: Remove "instead"

kneth:
:+1: besides my minor comments.

PR #1396, opened 2015-08-18
Null support for all types

beeender:
Fully null support for all types
#1098 is merged to this one as well.
#1376 is merged for core integration. And #1376 should be MERGED FIRST.

See https://github.com/realm/realm-wiki/wiki/NULL-in-Realm-Bindings for more information on rules and etc..

TODO checks from #1098 nullable string and bytes.
 - [x] Basic support for strings (insert, retrieve, query)
 - [x] Basic support for binary data (insert, retrieve)
 - [x] Basic support for booleans
 - [x] Basic support for integers
 - [x] Basic support for dates
 - [x] Migration for String
 - [x] JSON
 - [x] Sorting on nullable String columns doesn't work

Checks for all types null support
 - [x] Migration testing
 - [x] Dynamic object testing (Will be handled by CM's PR)
 - [x] Handling null for linklist (JSON and setter)
 - [x] Fix JSON stream parser after handling linklist
 - [x] `@Required` for link 
 - [x] Check `copyToOrUpdate` method
 - [x] Handle NaN value for float/double
 - [x] `isNull` for link query
 - [x] 'max' 'min' 'avg' return Boxed type (Number), clean up interface in `RealmQuery`.
 - [x] Align behavior of nullable bytes with cocoa team
 - [x] `io.realm.internal.Table.convertToNotNullableColumn()`

PR #1543, opened 2015-10-07
Add Nullable suport

beeender:
* Add boxed type support.
* Add @Required annotation.
* isNull and isNotNull support all nullable types now.
* equalTo query can take null as input param for nullable fields.
* Add functions for nullable related migration.
* JSON converter behavior changes for null value.
* Update test cases.


beeender:
Review is done in #1396

PR #1550, opened 2015-10-08
Add Realm.isClose() method.

### LINKED ISSUE #1431: Realm isClosed method
Hi, i noticed there is no method to check my Realm instance is still open/valid. There is no problem when i have RealmObject somewhere because it have "isValid" method so i can call that. But if i have RealmResults or RealmList this validation is not available.
Possible associate with
https://github.com/realm/realm-java/issues/697

zaki50:
This change fixes #1431

### DIFF
--- a/realm/src/main/java/io/realm/BaseRealm.java
+++ b/realm/src/main/java/io/realm/BaseRealm.java
@@ -415,6 +415,19 @@ abstract class BaseRealm implements Closeable {
     }
 
     /**
+     * Check if the {@link io.realm.Realm} instance has already been closed.
+     *
+     * @return {@code true} if closed, {@code false} otherwise.
+     */
+    public boolean isClosed() {
+        if (this.threadId != Thread.currentThread().getId()) {
+            throw new IllegalStateException(INCORRECT_THREAD_MESSAGE);
+        }
+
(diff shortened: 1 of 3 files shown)


zaki50:
@realm/java review this please

realm/src/main/java/io/realm/BaseRealm.java
```diff
@@ -415,6 +415,19 @@ public void close() {
     }
 
     /**
+     * Check if the {@link io.realm.Realm} instance has already been closed.
```
beeender: I assume this also works for `DynamicRealm` ? So better to say {@link BaseRealm} here?

cmelchior: We shouldn't mention BaseRealm it is package protected and only really used as a placeholder for Code. So it is not even part of our JavaDoc

cmelchior:
It needs a unit test for 

```
realm.close()
realm.<randomOperation>
```

But otherwise it looks good.

zaki50:
@cmelchior sure

cmelchior:
Since the unit test is part of #1551.  Dynamic Realm hasn't been merged yet. So the doc update should be part of that PR.

:+1:

beeender:
:+1:

PR #1551, opened 2015-10-08
Fix BaseRealm#checkIfValid() which does not throw Exception after close().

zaki50:
@realm/java review this please.

### DIFF
--- a/realm/src/androidTest/java/io/realm/RealmTest.java
+++ b/realm/src/androidTest/java/io/realm/RealmTest.java
@@ -186,6 +186,18 @@ public class RealmTest extends AndroidTestCase {
         }
     }
 
+    public void testCheckValid() {
+        // checkIfValid() must not throw any Exception against valid Realm instance.
+        testRealm.checkIfValid();
+
+        testRealm.close();
+        try {
+            testRealm.checkIfValid();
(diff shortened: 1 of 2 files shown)


cmelchior:
:+1:

realm/src/androidTest/java/io/realm/RealmTest.java
```diff
@@ -186,6 +186,18 @@ public void testGetInstanceClearsCacheWhenFailed() {
...
 
+    public void testCheckValid() {
+        // checkIfValid() must not throw any Exception against valid Realm instance.
+        testRealm.checkIfValid();
+
+        testRealm.close();
+        try {
+            testRealm.checkIfValid();
+            fail("closed Realm instance must throw IllegalStateException.");
+        } catch (IllegalStateException e) {
```
beeender: `e` -> `ignored`

zaki50:
@beeender fixed

realm/src/androidTest/java/io/realm/RealmTest.java
zaki50: @beeender fixed

beeender:
:+1:  Better squash and then push :smile:

PR #1565, opened 2015-10-09
Fix warning caused by non-ascii chars

beeender:


beeender:
@realm/java

cmelchior:
Yay, someone was more annoyed than me by the log warnings :)

:+1:

PR #1566, opened 2015-10-09
Adding Realm.distinct() method.

### LINKED ISSUE #960: DISTINCT
In my usage, there are various times where I only want one out of a set of objects that share a value of a field - much like the 'DISTINCT' clause in SQL. For exampe, with the items

name | logInTime
-----|-------
Bill | 1426270572
Mark | 1426272004
Mark | 1426273505
Mark | 1426274666
Sally | 1426268301
Sally | 1426271123

in the table 'User', `realm.where(User.class).distinct("name").findAll()` would return 3 items (one for 'Mark', one for 'Bill', and one for 'Sally'). With it like this, which 'Mark' and which 'Sally' is undefined - presumably the first one Realm finds.

A potential advanced aspect of this would be to select which of the rows to return. `realm.where(User.class).distinct("name", "logInTime", false).findAll()` would return `(Bill, 1426270572), (Mark, 1426274666), (Sally,1426271123)`, or only having the first discovered in an sorted RealmResult appear.

kneth:
Issue #960 requests a `distinct()` method. As Realm's underlying storage engine added support for the method recently, it is time to add such a method to Realm Java.

Closes #960 

@realm/java

### DIFF
--- a/realm-jni/src/io_realm_internal_table.cpp
+++ b/realm-jni/src/io_realm_internal_table.cpp
@@ -1329,7 +1329,7 @@ JNIEXPORT jlong JNICALL Java_io_realm_internal_Table_nativeGetDistinctView(
             } CATCH_STD()
             break;
         default:
-            ThrowException(env, IllegalArgument, "Invalid type - only bool, integer, datatime, string are supported.");
+            ThrowException(env, IllegalArgument, "Invalid type - Only String, Date, boolean, short, int, long and their boxed variants are supported.");
             return 0;
         break;
     }
--- a/realm-jni/src/io_realm_internal_table.cpp
+++ b/realm-jni/src/io_realm_internal_table.cpp
@@ -1315,17 +1315,24 @@ JNIEXPORT jlong JNICALL Java_io_realm_internal_Table_nativeGetDistinctView(
     if (!TBL_AND_COL_INDEX_VALID(env, pTable, columnIndex))
         return 0;
     if (!pTable->has_search_index(S(columnIndex))) {
-        ThrowException(env, UnsupportedOperation, "The column must be indexed before distinct() can be used.");
+        ThrowException(env, UnsupportedOperation, "The field must be indexed before distinct() can be used.");
         return 0;
     }
-    if (pTable->get_column_type(S(columnIndex)) != type_String) {
-        ThrowException(env, IllegalArgument, "Invalid columntype - only string columns are supported.");
-        return 0;
+    switch (pTable->get_column_type(S(columnIndex))) {
+        case type_Bool:
+        case type_Int:
+        case type_DateTime:
+        case type_String:
+            try {
+                TableView* pTableView = new TableView( pTable->get_distinct_view(S(columnIndex)) );
+                return reinterpret_cast<jlong>(pTableView);
+            } CATCH_STD()
+            break;
+        default:
+            ThrowException(env, IllegalArgument, "Invalid type - only bool, integer, datatime, string are supported.");
+            return 0;
+        break;
     }
-    try {
-        TableView* pTableView = new TableView( pTable->get_distinct_view(S(columnIndex)) );
-        return reinterpret_cast<jlong>(pTableView);
-    } CATCH_STD()
     return 0;
 }
 
--- a/realm/src/androidTest/java/io/realm/RealmTest.java
+++ b/realm/src/androidTest/java/io/realm/RealmTest.java
@@ -27,7 +27,6 @@ import org.json.JSONObject;
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
-import java.lang.Thread;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Date;
@@ -47,6 +46,7 @@ import java.util.concurrent.atomic.AtomicBoolean;
 import io.realm.dynamic.DynamicRealmObject;
 import io.realm.entities.AllTypes;
 import io.realm.entities.AllTypesPrimaryKey;
+import io.realm.entities.AnnotationIndexTypes;
 import io.realm.entities.Cat;
 import io.realm.entities.CyclicType;
 import io.realm.entities.CyclicTypePrimaryKey;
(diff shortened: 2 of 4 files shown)


realm-jni/src/io_realm_internal_table.cpp
```diff
@@ -1315,17 +1315,24 @@ JNIEXPORT jlong JNICALL Java_io_realm_internal_Table_nativeGetDistinctView(
...
+        case type_Int:
+        case type_DateTime:
+        case type_String:
+            try {
+                TableView* pTableView = new TableView( pTable->get_distinct_view(S(columnIndex)) );
+                return reinterpret_cast<jlong>(pTableView);
+            } CATCH_STD()
+            break;
+        default:
+            ThrowException(env, IllegalArgument, "Invalid type - only bool, integer, datatime, string are supported.");
```
emanuelez: I think you should list the Java types here, in order not to confuse the users. Or maybe "float, Float, double and Double are not supported"

realm/src/main/java/io/realm/Realm.java
```diff
@@ -946,6 +946,30 @@ boolean contains(Class<? extends RealmObject> clazz) {
     }
 
     /**
+     * Return a distinct set of objects of a specific class. If no objects exist, the
```
cmelchior: Also describe the policy for which element is returned for the distinct set (first, last, undefined)?

cmelchior:
Great to get this in. I don't agree it completely resolves #960 though as `DISTINCT` on query results don't work with this, so either edit #960 to reflect that or create a new issue for tracking `distinct()` on `RealmResults`

realm/src/androidTest/java/io/realm/RealmTest.java
```diff
@@ -2058,4 +2058,47 @@ private void awaitOrFail(CountDownLatch latch) {
...
+
+        RealmResults<AnnotationIndexTypes> distinctBool = testRealm.distinct(AnnotationIndexTypes.class, "indexBoolean");
+        assertEquals(2, distinctBool.size());
+
+        for (String fieldName : new String[]{"Long", "Date", "String"}) {
+            RealmResults<AnnotationIndexTypes> distinct = testRealm.distinct(AnnotationIndexTypes.class, "index" + fieldName);
+            assertEquals("index" + fieldName, numberOfBlocks, distinct.size());
+        }
+
+        for (String fieldName : new String[]{"Boolean", "Long", "Date", "String"}) {
```
cmelchior: This should be a separate unit test.

realm/src/androidTest/java/io/realm/RealmTest.java
```diff
@@ -2058,4 +2058,47 @@ private void awaitOrFail(CountDownLatch latch) {
...
+            try {
+                testRealm.distinct(AnnotationIndexTypes.class, "notIndex" + fieldName);
+                fail("notIndex" + fieldName);
+            } catch (UnsupportedOperationException ignore) {
+            }
+        }
+
+        try {
```
cmelchior: Separate unit test

cmelchior: There also seems to be missing a unit test for fields not supported like e.g. object or lists.

realm/src/main/java/io/realm/Realm.java
kneth: IMHO, as a Realm doesn't have order, it doesn't really matter.

cmelchior: Lets put it this way: I should have known better and I wondered. Having a line like "As a Realm is unordered, it is undefined which element is returned for values with multiple candidates"  (could probably be rephrased)

It makes the contract of the method totally clear and might save a support ticket :)

kneth: That's a valid point.

realm/src/androidTest/java/io/realm/RealmTest.java
```diff
@@ -2058,4 +2058,78 @@ private void awaitOrFail(CountDownLatch latch) {
...
+                obj.setIndexDate(new Date(1000 * j));
+                obj.setIndexString("Test " + j);
+                obj.setNotIndexBoolean(j % 2 == 0);
+                obj.setNotIndexLong(j);
+                obj.setNotIndexDate(new Date(1000 * j));
+                obj.setNotIndexString("Test " + j);
+            }
+        }
+        realm.commitTransaction();
+    }
```
beeender: One more test case for `null`? Does the `distinct` work on same rows with `null` values as well?

cmelchior:
Ready for review @realm/java

beeender:
:+1:

cmelchior:
:+1:

realm-jni/src/io_realm_internal_table.cpp
```diff
@@ -1315,17 +1315,24 @@ JNIEXPORT jlong JNICALL Java_io_realm_internal_Table_nativeGetDistinctView(
...
+        case type_Int:
+        case type_DateTime:
+        case type_String:
+            try {
+                TableView* pTableView = new TableView( pTable->get_distinct_view(S(columnIndex)) );
+                return reinterpret_cast<jlong>(pTableView);
+            } CATCH_STD()
+            break;
+        default:
+            ThrowException(env, IllegalArgument, "Invalid type - Only String, Date, boolean, short, int, long and their boxed variants are supported.");
```
nhachicha: in async I use this message instead 
`Distinct is not supported on binary data, object references and RealmList`

nhachicha:
:+1:

realm-jni/src/io_realm_internal_table.cpp
kneth: `float` and `double` ain't supported too.

realm/src/androidTest/java/io/realm/RealmTest.java
```diff
@@ -2058,4 +2058,90 @@ private void awaitOrFail(CountDownLatch latch) {
...
+        final long numberOfBlocks = 25;
+        final long numberOfObjects = 10; // must be greater than 1
+
+        populateForDistinct(testRealm, numberOfBlocks, numberOfObjects, false);
+
+        for (String fieldName : new String[]{"Boolean", "Long", "Date", "String"}) {
+            try {
+                testRealm.distinct(AnnotationIndexTypes.class, "notIndex" + fieldName);
+                fail("notIndex" + fieldName);
+            } catch (UnsupportedOperationException ignore) {
```
bmunkholm: Don't we call it "expected" instead of "ignore" now?

Issue #1554, opened 2015-10-08
Realm is null (SIGSEGV) after updating to 0.83.0


bendywalker:
After removing a RealmObject using `.removeFromRealm()`, on the next(?) query of the Realm I'm getting a crash:

```
java.lang.NullPointerException: Attempt to invoke virtual method 'io.realm.RealmQuery io.realm.Realm.where(java.lang.Class)' on a null object reference
```
This is also sometimes accompanied with the following message: `Fatal signal 11 (SIGSEGV)`, so it looks like there is a crash in Realm's internals after performing the delete, that causes the Realm to be nulled.

I'm not doing anything especially fancy - just deleting the object. I do have a fair number of Realm change listeners going while this is happening, but I'm checking if objects are valid in those before using them. I'm using a single realm per activity, so at the point of the crash I'll have at most 2 open.

Not certain if this is related to 0.83.0 - with the upgrade I am now using `isValid()` in more places as it now [behaves as expected](https://github.com/realm/realm-java/issues/1393).

beeender:
@bendywalker Is it possible for you to make a simple project to reproduce this problem? 
Or would it be a similar problem with https://github.com/realm/realm-java/issues/1561 ?

thanks !

bendywalker:
@beeender It would be quite hard to pull the code out for a simple project; it's a big app and is for a client.

It doesn't seem to be related to #1561, but I'm not certain.

I just got the crash again, this time with a larger native crash log:

```
10-09 16:06:23.940 20465-20465/com.testing.development A/libc: Fatal signal 11 (SIGSEGV), code 1, fault addr 0x428 in tid 20465 (ing.development)
10-09 16:06:24.070 948-948/? I/DEBUG: *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
10-09 16:06:24.070 948-948/? I/DEBUG: Build fingerprint: 'generic_x86_64/sdk_phone_x86_64/generic_x86_64:5.1/LKY45/1737576:eng/test-keys'
10-09 16:06:24.070 948-948/? I/DEBUG: Revision: '0'
10-09 16:06:24.070 948-948/? I/DEBUG: ABI: 'x86_64'
10-09 16:06:24.070 948-948/? I/DEBUG: pid: 20465, tid: 20465, name: ing.development  >>> com.testing.development <<<
[...]

PR #1584, opened 2015-10-12
Missing realm check in some functions

beeender:
* Generate realm check for getter and setter of RealmObject and RealmList.
* Add realm check in RealmObject.removeFromRealm
* Add realm check in RealmResults's remove, removeLast, clear methods
* Add test cases
* Minor warning fix

Maybe related with #1554 ?

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -11,6 +11,7 @@
  * Following methods in managed RealmList now throw IllegalStateException instead of ArrayIndexOutOfBoundsException when RealmList.isValid() returns false: set(int,RealmObject), move(int,int), remove(int), get(int)
  * Following methods in managed RealmList now throw IllegalStateException instead of returning 0/null when RealmList.isValid() returns false: clear(), removeAll(Collection), remove(RealmObject), first(), last(), size(), where()
  * RealmPrimaryKeyConstraintException is now thrown instead of RealmException if two objects with same primary key are inserted.
+ * IllegalStateException is now thrown when calling Realm's clear(), RealmResults's remove(), removeLast(), clear() or RealmObject's removeFromRealm() from an incorrect thread.
  * Fixed a bug affecting RealmConfiguration.equals().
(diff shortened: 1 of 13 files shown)


beeender:
retest this please

beeender:
@realm/java

kneth:
What is the performance penalty for these checks?

beeender:
`checkIfValid` is relatively fast, most time consuming part is the `Thread.currentThread()` in there which leads to a jni call into bionic.

For now, since we are checking everywhere to make sure user won't use them in different thread, there is no reason to not check them in those places.

For the future, checking the thread in the jni code or even in the core would be a good idea. That could reduce the jni calls to accelerate a lot.

zaki50:
@beeender Also `RealmResults.size()` should have this check.

nhachicha:
this is off topic but we may consider in the future to expose the checking or not as a RealmConfiguration parameter. Of course we're not a this level of optimisations but it's worth considering post 1.0

zaki50:
Hmm.. calling `Realm.checkIfValid()` in `RealmResults.size()` is overkill?

nhachicha:
@beeender `Realm.clear` should also check IMO I add a comment on the respective Dynamic API call as well

beeender:
@zaki50 I think calling `Realm.checkIfValid()` in `RealmResults.size()` is a bit overkill... User cannot really do anything with a returned size.

beeender:
@realm/java Added check in `Realm.clear()`

cmelchior:
I think we are missing `checkIfValid` from `createObject` as well?

cmelchior:
Added some more thread checks + unit tests. 

Ready for review @realm/java

dalinaum:
LGTM for 390345f :+1:

beeender:
:+1:

zaki50:
:+1:

PR #1587, opened 2015-10-13
Fix #1490

### LINKED ISSUE #1490: Two configurations with different schemas are trying to open the same Realm file.
Hi!

I'm running into a random issue with captioned error when I use two Realms with different configurations (without overlapping modules for each configuration). Any idea for the cause of the error?

This error is not always reproducible and it has appeared under some random occasions on Android 4.3 and 5.1 with 0.82.2.

Thanks for your help!

zaki50:
change the return type of RealmProxyMediator#getModelClasses() to Set in order to guarantee its contents unordered and unique.

TODO

- [x] fix implementation (includes API changes)
- [x] add tests
- [x] update changelog

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -5,6 +5,7 @@
  * Following methods in managed RealmList now throw IllegalStateException instead of ArrayIndexOutOfBoundsException when RealmList.isValid() returns false: set(int,RealmObject), move(int,int), remove(int), get(int)
  * Following methods in managed RealmList now throw IllegalStateException instead of returning 0/null when RealmList.isValid() returns false: clear(), removeAll(Collection), remove(RealmObject), first(), last(), size(), where()
  * RealmPrimaryKeyConstraintException is now thrown instead of RealmException if two objects with same primary key are inserted.
+ * Fixed a bug affecting RealmConfiguration.equals().
 
 0.83
  * BREAKING CHANGE: Database file format update. The Realm file created by this version cannot be used by previous versions of Realm.
--- a/realm/src/androidTest/java/io/realm/MediatorTest.java
+++ b/realm/src/androidTest/java/io/realm/MediatorTest.java
@@ -1,0 +1,84 @@
+/*
+ * Copyright 2015 Realm Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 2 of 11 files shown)


realm/src/main/java/io/realm/Realm.java
```diff
@@ -318,10 +318,8 @@ private static Realm createAndValidate(RealmConfiguration configuration, boolean
 
     private static void setupColumnIndices(Realm realm) {
         RealmProxyMediator mediator = realm.configuration.getSchemaMediator();
-        List<Class<? extends RealmObject>> modelClasses = mediator.getModelClasses();
-        int size = modelClasses.size();
-        for (int i = 0; i < size; i++) {
-            Class<? extends RealmObject> modelClass = modelClasses.get(i);
+        Set<Class<? extends RealmObject>> modelClasses = mediator.getModelClasses();
+        for (Class<? extends RealmObject> modelClass : modelClasses) {
```
emanuelez: Too bad we lose the opportunity to iterate with indices, but it's a one-off operation

zaki50: We don't have any other choices...

zaki50:
@realm/java please review this

changelog.txt
```diff
@@ -5,6 +5,7 @@
  * Following methods in managed RealmList now throw IllegalStateException instead of ArrayIndexOutOfBoundsException when RealmList.isValid() returns false: set(int,RealmObject), move(int,int), remove(int), get(int)
  * Following methods in managed RealmList now throw IllegalStateException instead of returning 0/null when RealmList.isValid() returns false: clear(), removeAll(Collection), remove(RealmObject), first(), last(), size(), where()
  * RealmPrimaryKeyConstraintException is now thrown instead of RealmException if two objects with same primary key are inserted.
+ * Changed return type of RealmProxyMediator.getModelClasses() to Set<Class<? extends RealmObject>>.
```
emanuelez: In the changelog we only list changes visible to users. So I think you should just state the problem that this PR fixes.

emanuelez:
:+1: apart from the changelog comment

realm/src/androidTest/java/io/realm/MediatorTest.java
```diff
@@ -0,0 +1,69 @@
+package io.realm;
```
cmelchior: Missing license header

cmelchior:
Apart from comments this looks good :+1:

changelog.txt
zaki50: Oh, I didn't know that RealmProxyMediator is in an internal package.

zaki50:
@emanuelez I updated changelog.txt.

realm/src/androidTest/java/io/realm/MediatorTest.java
```diff
@@ -0,0 +1,84 @@
+/*
+ * Copyright 2014 Realm Inc.
```
beeender: 2015 Realm Inc.

zaki50:
@emanuelez sure

zaki50:
Oh, 2015...

zaki50:
thanks @beeender

dalinaum:
refs #1490

PR #1533, opened 2015-10-06
Fix for JsonUtils stringToDate to support Microsoft Dates with TimeZones

LateralKevin:
We are currently using realm in an Android project which receives JSON from an MVC project.

The current codebase only supports the following Date format

/Date(long)

The MVC web.config is configures with the following option

                    logger.Info("Enabling Json in to MicrosoftDateFormat. Timezone format: Local");
                    GlobalConfiguration.Configuration.Formatters.JsonFormatter.SerializerSettings =
                         new JsonSerializerSettings
                         {
                             DateTimeZoneHandling = DateTimeZoneHandling.Local,
                             DateFormatHandling = DateFormatHandling.MicrosoftDateFormat
                         };

Which add the timezone information to the long delivered to the client

/Date(1443854733376+0800)/

This causes a NumberFormatException, the code changes within this pull request support the addition of the +Zone in the long format.

I've also updated the RealmJsonTest.java to add a test case for this format of the Date object.

PR #1588, opened 2015-10-13
Fix for JsonUtils stringToDate to support Microsoft Dates with TimeZones

cmelchior:
Replaces #1533. This PR fixes the last couple of issues so it is ready to be merged.

@realm/java


realm/src/androidTest/java/io/realm/RealmJsonTest.java
```diff
@@ -121,6 +124,26 @@ public void testCreateObjectFromJson_dateAsString() throws JSONException {
         assertEquals(new Date(1000), obj.getColumnDate());
     }
 
+    public void testCreateObjectFromJson_dateAsStringTimeZone() throws JSONException {
+        JSONObject json = new JSONObject();
+
+        // Oct 03 2015 14:15
```
zaki50: 14:45?

realm/src/main/java/io/realm/internal/android/JsonUtils.java
```diff
@@ -17,18 +17,21 @@
...
+import java.util.Calendar;
 import java.util.Date;
+import java.util.GregorianCalendar;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 public class JsonUtils {
 
-    private static Pattern jsonDate = Pattern.compile("/Date\\((\\d*)\\)/");
+    private static Pattern jsonDate = Pattern.compile("Date\\((\\d*)([+-]\\d*)?\\)");
```
zaki50: `([+-]\\d*)` can be `(?:[+-]\\d*)` since we don't need to capture it.

kneth:
Look good to me.

zaki50:
:+1:

PR #1590, opened 2015-10-13
Upgrade to core 0.94.1

cmelchior:


beeender:
Need to check with @rrrlasse.

`is_null` on links behaviour changed in the latest release. before if the link is null, `is_null(something.link.string)` will return false. Now it returns true.

bdash:
@beeender, that behavior was intentionally changed as part of realm/realm-core#1161. It looks like I neglected to mention it in core's release notes, but it is called out in the commit message. Without the behavior change you end up with the unintuitive situation where a query like "link.foo == null && link.foo != null" doesn't include the full result set.

beeender:
@bdash Thanks a lot for the clarification! I am changing the test cases right now!

cmelchior:
retest this please

cmelchior:
Review please @realm/java

realm/src/androidTest/java/io/realm/RealmQueryTest.java
```diff
@@ -1072,35 +1072,36 @@ public void testSumValueForAllNullColumns() {
     public void testIsNullOnLinkField() {
         TestHelper.populateTestRealmForNullTests(testRealm);
 
+        // For the link with null value, query isNull on it's fields should return true.
         // 1 String
-        assertEquals(1, testRealm.where(NullTypes.class).isNull(
+        assertEquals(2, testRealm.where(NullTypes.class).isNull(
```
kneth: Do we have to update the documentation? It seems that `isNull()` is behaving differently now (breaking change?).

realm-jni/src/tablequery.hpp
```diff
@@ -28,7 +28,7 @@ class TableQuery : public realm::Query {
     std::vector<size_t> subtables;  // holds subtable column indices
 
 public:
-    TableQuery(const Query& copy) : realm::Query(copy, realm::Query::TCopyExpressionTag{}) {};
```
kneth: What is the implication? Is the `realm:Query` object deep-copied or not?

realm/src/androidTest/java/io/realm/RealmQueryTest.java
beeender: This detail is not mentioned before... So we can update the document in another PR?

realm/src/androidTest/java/io/realm/RealmTest.java
```diff
@@ -2038,6 +2039,8 @@ public void onChange() {
                     fail();
                 } finally {
                     bgRealm.close();
+                    bgClosedLatch.countDown();
+
```
zaki50: unnecessary empty line

cmelchior:
Updated with comments @realm/java

changelog.txt
```diff
@@ -1,4 +1,7 @@
-0.83.1
+0.84.0
```
emanuelez: Should't this stay as 0.83.1?

kneth:
:+1:

realm/src/androidTest/java/io/realm/RealmQueryTest.java
```diff
@@ -1072,35 +1072,36 @@ public void testSumValueForAllNullColumns() {
     public void testIsNullOnLinkField() {
         TestHelper.populateTestRealmForNullTests(testRealm);
 
+        // For the link with null value, query isNull on it's fields should return true.
```
emanuelez: it's -> its

realm/src/androidTest/java/io/realm/RealmTest.java
```diff
@@ -2013,12 +2013,13 @@ public void run() {
         }
     }
 
-    public void testProcessLocalListenersAfterRefresh() {
+    public void testProcessLocalListenersAfterRefresh() throws InterruptedException {
```
emanuelez: I guess we could just catch Exception here

zaki50:
:+1:

changelog.txt
cmelchior: Yes, but we need to rewrite the changelog before a release anyway. As this PR must be rebased on top of 0.83.0. For merging to master I think 0.84.0 is the correct version?

realm/src/androidTest/java/io/realm/RealmTest.java
cmelchior: You mean throws? Any exception not caught will fail the unit test anyway, so we need only to declare any checked exceptions?

emanuelez:
:+1: apart from `it's -> its`

realm/src/androidTest/java/io/realm/RealmTest.java
emanuelez: Not important. I should even have mentioned this. :)

changelog.txt
```diff
@@ -1,4 +1,7 @@
...
+ * Updated Realm core to version 0.94.1.
+   - Fixed a bug when using Realm.compactRealm().
```
bmunkholm: Expand description: How could one notice that bug?

PR #1597, opened 2015-10-14
Parallelize make jobs

danielpovlsen:
@realm/java

### DIFF
--- a/realm-jni/build.gradle
+++ b/realm-jni/build.gradle
@@ -254,7 +254,8 @@ targets.each { target ->
         environment REALM_ANDROID: '1'
         commandLine = [
             'make',
-            '-j', "${Runtime.getRuntime().availableProcessors()}", // logical cores (i.e. includes hyper-threading)
+            "-j${Runtime.getRuntime().availableProcessors() * 2}",
+            "-l${Runtime.getRuntime().availableProcessors()}",
             '-C', "${projectDir}/src",
             "CC_IS=${clang?'clang':'gcc'}",
             "REALM_CFLAGS=-Wno-variadic-macros -DREALM_HAVE_CONFIG -DPIC -I${project.coreDir}/include",
--- a/realm-jni/build.gradle
+++ b/realm-jni/build.gradle
@@ -254,6 +254,7 @@ targets.each { target ->
         environment REALM_ANDROID: '1'
         commandLine = [
             'make',
+            '-j',
             '-C', "${projectDir}/src",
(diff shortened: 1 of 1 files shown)


cmelchior:
Failing unit test is not related to this PR. It seems we have a unit test not behaving correctly elsewhere. 

This is "YES, PLEASE" :+1:

emanuelez:
Hmmm `-j` will kill the load and IO. Supposing most of us (and the CI servers) have 4 cores I would suggest something like `-j8 -l4` so we enforce a limit based on the load.

emanuelez:
retest this please

danielpovlsen:
Optimally a cradle expert *wink, wink* will find a way to portably detect number of CPU cores. ;-)

emanuelez:
I'm not aware of a way to do it in make (and that's why it's done in build.sh for core), but once we port our build system to pure Gradle it will be doable

danielpovlsen:
These jobs are fairly small, so unless we are very strapped for memory I do not think it will be slower with just `-j` compared to `-j8`. And Infinitely™ faster compared with the current, implicit `-j1`.

emanuelez:
Faster yes, but also making your machine hardly usable. At least that's my experience. @kspangsege Any experience on this?

emanuelez:
Actually, since we instrument make from Gradle we can inject a `MAKE_FLAGS` variable and we can get the number of cores from the JVM with something like `Runtime.getRuntime().availableProcessors()`

danielpovlsen:
w00t! :-D

zaki50:
I tried 

            "-j${Runtime.getRuntime().availableProcessors() * 2}",
            "-l${Runtime.getRuntime().availableProcessors()}",

and it works good for me

danielpovlsen:
@emanuelez Good to go?

danielpovlsen:
@zaki50 Thanks!

emanuelez:
Yes I agree with @zaki50 's suggestion

danielpovlsen:
@emanuelez Well, actually, `availableProcessors()` already report logical cores, so I doubt that we would want to 2x that? Check out the last commit.

emanuelez:
I know that. The 2x factor accounts for IO and it's pretty empirical but the `-l` assures we don't clog the CPU anyway

nhachicha:
:+1:

emanuelez:
:+1:

dalinaum:
:+1:

zaki50:
:+1:

PR #1606, opened 2015-10-16
Premature query validation

kneth:
Fixed a bug in `RealmQuery.isNull()` and `RealmQuery.isNotNull()` which validated the query prematurely.

@cmelchior @beeender

### DIFF
--- a/realm/src/main/java/io/realm/internal/TableQuery.java
+++ b/realm/src/main/java/io/realm/internal/TableQuery.java
@@ -595,14 +595,14 @@ public class TableQuery implements Closeable {
 
     // isNull and isNotNull
     public TableQuery isNull(long columnIndices[]) {
-        queryValidated = false;
         nativeIsNull(nativePtr, columnIndices);
+        queryValidated = false;
         return this;
     }
 
     public TableQuery isNotNull(long columnIndices[]) {
-        queryValidated = false;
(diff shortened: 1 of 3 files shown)


realm/src/main/java/io/realm/internal/TableQuery.java
```diff
@@ -589,16 +592,16 @@ public Date minimumDate(long columnIndex) {
         }
         return null;
     }
-    
+
     // isNull and isNotNull
     public TableQuery isNull(long columnIndices[]) {
-        validateQuery();
+        queryValidated = false;
```
cmelchior: In the other methods this is below the native method call. It probably doesn't matter much, but would be nice to have it moved for consistency.

realm/src/main/java/io/realm/internal/TableQuery.java
```diff
@@ -589,16 +592,16 @@ public Date minimumDate(long columnIndex) {
...
         nativeIsNull(nativePtr, columnIndices);
         return this;
     }
 
     public TableQuery isNotNull(long columnIndices[]) {
-        validateQuery();
+        queryValidated = false;
```
cmelchior: Same

cmelchior:
Apart from small cosmetic change :+1:

PR #1607, opened 2015-10-16
Add RealmQuery.isEmpty()

### LINKED ISSUE #1601: RealmQuery.isEmpty()
After introducing Null support, we redefined the semantics of `isNull`. This unfortunately meant removing the option of doing `isNull("list")` to check for 0 items.

This means that people no longer can check for the empty list. A proposed fix for this is adding `isEmpty()` which also has uses for other datatypes.

```
realm.where(Foo.class).isEmpty("list").findAll(); // Find all lists with 0 items
realm.where(Foo.class).isEmpty("str").findAll(); // Find all empty but non-null strings ""
realm.where(Foo.class).isEmpty("byteArray").findAll(); // Find all non-null, 0-length blobs
realm.where(Foo.class).isEmpty("integer").findAll(); // Throw IllegalArgumentException
```

cmelchior:
Fixes #1601 

Adds a method to RealmQuery called `isEmpty()` it can be used to check if Strings are empty (non-null), binary arrays have length 0 (non-null) and if RealmLists have size 0. 

That last was the primary reason as this was no longer possible after adding Null support.

@realm/java

### DIFF
--- a/realm-jni/src/io_realm_internal_TableQuery.cpp
+++ b/realm-jni/src/io_realm_internal_TableQuery.cpp
@@ -70,7 +70,7 @@ JNIEXPORT jstring JNICALL Java_io_realm_internal_TableQuery_nativeValidateQuery
 // Return TableRef used for build link queries
 TableRef getTableForLinkQuery(jlong nativeQueryPtr, jlong *arr, jsize arr_len) {
     TableRef table_ref = Q(nativeQueryPtr)->get_table();
-    for (int i=0; i<arr_len-1; i++) {
+    for (int i = 0; i < arr_len - 1; i++) {
         table_ref->link(size_t(arr[i]));
     }
     return table_ref;
@@ -79,7 +79,7 @@ TableRef getTableForLinkQuery(jlong nativeQueryPtr, jlong *arr, jsize arr_len) {
 // Return TableRef point to original table or the link table
 TableRef getTableByArray(jlong nativeQueryPtr, jlong *arr, jsize arr_len) {
     TableRef table_ref = Q(nativeQueryPtr)->get_table();
-    for (int i=0; i<arr_len-1; i++) {
+    for (int i = 0; i < arr_len - 1; i++) {
         table_ref = table_ref->get_link_target(size_t(arr[i]));
     }
     return table_ref;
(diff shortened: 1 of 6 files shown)


realm/src/androidTest/java/io/realm/RealmQueryTest.java
```diff
@@ -1446,4 +1443,116 @@ public void testIsValidOfRemovedParent() {
...
     }
+
+
+    public enum SupportedIsEmptyTypes {
+        STRING, BINARY, LIST
+    }
+
+    public enum NotSupportedIsEmptyTypes {
+        INTEGER, FLOAT, DOUBLE, BOOLEAN, OBJECT, DATE
+    }
```
zaki50: How about using ColumnType to ensure all types are tested.

for example:
```
    static List<ColumnType> SUPPORTED_IS_EMPTY_TYPES = Arrays.asList(
            ColumnType.STRING,
            ColumnType.BINARY,
            ColumnType.LINK_LIST
            );

    static List<ColumnType> NOT_SUPPORTED_IS_EMPTY_TYPES;
    static {
        final ArrayList<ColumnType> list = new ArrayList<ColumnType>(Arrays.asList(ColumnType.values()));
        list.removeAll(SUPPORTED_IS_EMPTY_TYPES);
        // remove unsupported types
        list.remove(ColumnType.MIXED);
        list.remove(ColumnType.TABLE);
    }
```

realm/src/androidTest/java/io/realm/RealmQueryTest.java
```diff
@@ -31,8 +32,9 @@
 import io.realm.entities.Owner;
 import io.realm.entities.StringOnly;
 import io.realm.exceptions.RealmError;
+import io.realm.internal.ColumnType;
```
zaki50: unused

realm-jni/src/io_realm_internal_TableQuery.cpp
```diff
@@ -1347,3 +1347,67 @@ JNIEXPORT void JNICALL Java_io_realm_internal_TableQuery_nativeIsNotNull
...
+                    break;
+                case type_Link:
+                case type_Bool:
+                case type_Int:
+                case type_Float:
+                case type_Double:
+                case type_DateTime:
+                default:
+                    ThrowException(env, IllegalArgument, "isEmpty() only works on String, byte[] and RealmList.");
+                    return;
```
zaki50: should not `return` to reach to `RELEASE_ARRAY()`.

realm-jni/src/io_realm_internal_TableQuery.cpp
```diff
@@ -1347,3 +1347,67 @@ JNIEXPORT void JNICALL Java_io_realm_internal_TableQuery_nativeIsNotNull
...
+                    break;
+                case type_Link:
+                case type_Bool:
+                case type_Int:
+                case type_Float:
+                case type_Double:
+                case type_DateTime:
+                default:
+                    ThrowException(env, IllegalArgument, "isEmpty() only works on String, byte[] and RealmList across links.");
+                    return;
```
zaki50: should not `return` to reach to `RELEASE_ARRAY()`.

realm/src/main/java/io/realm/RealmQuery.java
```diff
@@ -133,13 +134,27 @@ private boolean containsDot(String s) {
...
+     * Returns the column indices for the given field name. If a linked field is defined, the column index for
+     * each
+     *
+     * @param fieldDescription fieldName or link path to a field name.
+     * @param validColumnTypes Legal field type for the last field a
+     * @return
+     */
     // TODO: consider another caching strategy so linked classes are included in the cache.
-    private long[] getColumnIndices(String fieldName, ColumnType fieldType) {
+    private long[] getColumnIndices(String fieldDescription, ColumnType... validColumnTypes) {
```
zaki50: call of this method with second argument `null` should be fixed to `getColumnIndices(fieldName, (Object[]) null)` to avoid warning.

realm/src/main/java/io/realm/RealmQuery.java
```diff
@@ -152,24 +167,39 @@ private boolean containsDot(String s) {
...
+        }
+    }
+
+    private boolean isValidType(ColumnType columnType, ColumnType[] validColumnTypes) {
+        for (int i = 0; i < validColumnTypes.length; i++) {
+            if (validColumnTypes[i] == columnType) {
+                return true;
             }
-            return new long[] {columns.get(fieldName)};
+
```
zaki50: needless empty line

realm/src/main/java/io/realm/RealmQuery.java
cmelchior: Done

realm/src/androidTest/java/io/realm/RealmQueryTest.java
cmelchior: It's a bit more cumbersome to write, but this will detect if we add new types and forget to update this unit test. I updated them with this.

realm/src/androidTest/java/io/realm/RealmQueryTest.java
cmelchior: Not anymore :)

kneth:
:+1:

zaki50:
:+1:

dalinaum:
:+1:

Issue #1581, opened 2015-10-12
Illegal Argument Exception encountered when compressing a Realm file created by version 0.83.0


creativepsyco:
Hi peeps,

I am currently running into an `IllegalArgumentException` indicating that the format of the Realm file is invalid when I am trying to call `Realm.compactRealm(realmConfiguration...)` before I use the realm file.

I have a sample project to demonstrate the crash here: https://github.com/creativepsyco/sample-realm-crash/

Essentially I am trying to compact the realm file before it is ever used in the app.

Steps to reproduce:
1. Clone the project from https://github.com/creativepsyco/sample-realm-crash/
2. Compile and run. It will not crash because realm version is `0.82.2`
3. Now in `build.gradle` edit the version to `0.83.0`
4. Compile and run the app. 
5. It crashes with the stack trace:
```
 FATAL EXCEPTION: main
 Process: mohitkanwal.com.examplerealm, PID: 28687
 java.lang.RuntimeException: Unable to start activity
ComponentInfo{mohitkanwal.com.examplerealm/mohitkanwal.com.examplerealm.MainActivity}:\
java.lang.IllegalArgumentException: Illegal Argument: Invalid format of Realm file.
     at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:3119)
     at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:3218)
     at android.app.ActivityThread.access$1000(ActivityThread.java:198)
     at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1676)
     at android.os.Handler.dispatchMessage(Handler.java:102)
     at android.os.Looper.loop(Looper.java:145)
     at android.app.ActivityThread.main(ActivityThread.java:6837)
     at java.lang.reflect.Method.invoke(Native Method)
     at java.lang.reflect.Method.invoke(Method.java:372)
     at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:1404)
     at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1199)
  Caused by: java.lang.IllegalArgumentException: Illegal Argument: Invalid format of Realm file.
     at io.realm.internal.SharedGroup.createNativeWithImplicitTransactions(Native Method)
     at io.realm.internal.SharedGroup.<init>(SharedGroup.java:68)
     at io.realm.internal.SharedGroupManager.<init>(SharedGroupManager.java:49)
[...]

PR #1613, opened 2015-10-18
Add test to open Realm database after compactRealm

zaki50:
test for #1581

I'm not sure that we should have this test in realm-java since it seems to be just testing the core.


cmelchior:
Makes sense to me since :+1:

dalinaum:
:+1:

LiuDeng:
this still have problem！！！invalid realm format

cmelchior:
@LiuDeng Please create a new issue with all relevant information if you are still having problems with this. If you can provide a sample project that shows it, it would be even better.

Issue #1610, opened 2015-10-18
Field 'xxx': type mismatch. Was INTEGER, expected STRING


diegomontoya:
Related to https://github.com/realm/realm-java/issues/1319

We are still seeing this issue with 10-12 snapshot. 

```
java.lang.IllegalArgumentException: Field 'status': type mismatch. Was INTEGER, expected STRING.
       at io.realm.RealmQuery.getColumnIndices(RealmQuery.java:167)
       at io.realm.RealmQuery.equalTo(RealmQuery.java:305)
```

Crashed point with query 

```
realm.where(MyRealm.class).equalTo("status", new Integer(1)).findAll()
```

The column in question, 'status' is an "int". 

```
    @PrimaryKey
    private long a;

    @Index
    private int b;

    @Index
    private int c;

    @Required
    private String d;

    @Index
    private int status;
```

Can a db corruption cause column def corruption? If so, it may be time to move db def to a separate file to ensure such things cannot happen.

cmelchior:
It is very unlikely to be DB corruption, it is more likely our translation of your fields to our database types . We cache a mapping between your field names and database types when opening a Realm, this cache is used when creating the queries, so if something changed after we created the cache and we didn't clear it properly you would get an error like the one you are seeing. You can see the field just before `status` is a String.

Do you run any migrations on your database? Also are you able to reproduce this?

diegomontoya:
There is no migration defined for this realm instance/db. We are using separate realm db for each unique table so this db only has 1 table with no migration. We cannot reproduce this bug on our devices but the occurrence of the bug is relatively high at around 0.3% of users.

diegomontoya:
Sidebar

RealmQuery.java Line 161

columns.get(fieldName) called potentially 3 times. 

```java
            if (columns.get(fieldName) == null) {
                throw new IllegalArgumentException(String.format("Field '%s' does not exist.", fieldName));
            }

            ColumnType tableColumnType = table.getColumnType(columns.get(fieldName));
            if (fieldType != null && fieldType != tableColumnType) {
                throw new IllegalArgumentException(String.format("Field '%s': type mismatch. Was %s, expected %s.",
                        fieldName, fieldType, tableColumnType
                ));
            }
            return new long[] {columns.get(fieldName)};
```

Also, why does getColumnType need to call native jni to get column type every single time and not cached? Strange this has to be done every time since column index is cached but column data type is not. 

```java
    @Override
[...]

PR #1624, opened 2015-10-20
Prevent migration of open Realms

cmelchior:
Found while investigating #1610 

It should not be possible to migrate open Realms as that can potentially invalidate the schema. As the schema is only validated when opening the Realm this can cause all kinds of strange behavior in querying/reading and writing data.

@realm/java


beeender:
:+1:

emanuelez:
:+1:

dalinaum:
:+1:

PR #1636, opened 2015-10-21
Added dates for all releases to changelog

cmelchior:
Changelog has been cleaned up, so now all releases contain the release data instead of just the first 4.

@realm/java

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -22,7 +22,7 @@
    - Fixed a bug when using Realm.compactRealm().
    - Fixed a bug, so link queries now always return true if any part is null.s
 
-0.83
+0.83 (8 Oct 2015)
  * BREAKING CHANGE: Database file format update. The Realm file created by this version cannot be used by previous versions of Realm.
  * BREAKING CHANGE: Removed deprecated methods and constructors from the Realm class.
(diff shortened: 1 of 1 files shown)


emanuelez:
:-1:  we lose the ability to automate releases. :(

cmelchior:
Seems relatively easy to create a search/replace regexp, especially from Groovy?

changelog.txt
```diff
@@ -89,14 +89,14 @@
  * Solved ConcurrentModificationException thrown when addChangeListener/removeChangeListener got called in the onChange. (thanks @beeender)
  * Fixed duplicated listeners in the same realm instance. Trying to add duplicated listeners is ignored now. (thanks @beeender)
 
-0.80.2
+0.80.2 (4 Maj 2015)
```
kneth: May

kneth:
I do like the idea so you get :+1: from me.

dalinaum:
:+1:

emanuelez:
Not with the release process of the new build system, unless we decide to fork the plugin we are using.

cmelchior:
I'm fine with closing this. I think it is better to have the dates there, but it is a minor itch. I'll find some other way to scratch it :)

bmunkholm:
I've been missing the dates as well. I think we should include them. It's totally fine to not have the date there automatically, we can add that manually even after the release (just like it's been done now).

PR #1637, opened 2015-10-21
add async distinct

nhachicha:
- add `distinctAsync` + uni tests
- add a refactor to include all logic about Handler signal in `HandlerController`
- Making `RealmResults` final as well as some recently add methods from `RealmObject` (related to async)


nhachicha:
@realm/java

realm/src/main/java/io/realm/Realm.java
```diff
@@ -961,7 +961,38 @@ boolean contains(Class<? extends RealmObject> clazz) {
...
+        }
+
+        Table table = this.getTable(clazz);
+        long columnIndex = table.getColumnIndex(fieldName);
+        if (columnIndex == -1) {
+            throw new IllegalArgumentException(String.format("Field name '%s' does not exist.", fieldName));
+        }
+
+        // check if the field is indexed
+        if (!table.hasSearchIndex(columnIndex)) {
```
kneth: Why not do the check in C++? The `Realm.distinct()` method does that and then our code is consistent.

realm/src/main/java/io/realm/RealmQuery.java
```diff
@@ -1472,6 +1472,72 @@ public long count() {
         return this.query.count();
     }
 
+    RealmResults<E> distinctAsync(final long columnIndex) {
```
kneth: I would prefer that `distinctAsync()` is a method in `Realm` and `io.realm.internal.Table`. When a user reads the API doc, it is a bit confusing that `realm.where(Person.class).equalTo("name", "John").distinctAsync(5)` is not allowed (5 = field/column number for the `age` field).

realm-jni/src/io_realm_internal_TableQuery.cpp
```diff
@@ -966,6 +966,40 @@ JNIEXPORT jlong JNICALL Java_io_realm_internal_TableQuery_nativeFindAllWithHando
...
+                    TableView tableView(table->get_distinct_view(S(columnIndex)) );
+
+                    // handover the result
+                    std::unique_ptr<SharedGroup::Handover<TableView>> handover = SG(
+                            bgSharedGroupPtr)->export_for_handover(tableView, MutableSourcePayload::Move);
+                    return reinterpret_cast<jlong>(handover.release());
+                } CATCH_STD()
+                break;
+            default:
+                ThrowException(env, IllegalArgument, "Invalid type - Only String, Date, boolean, short, int, long and their boxed variants are supported.");
```
cmelchior: Shouldn't this just be "Invalid type - Distinct only work on fields that are indexed"?

realm/src/androidTest/java/io/realm/RealmAsyncQueryTests.java
```diff
@@ -2157,6 +2158,67 @@ public void run() {
...
+                    }
+                    if (realm != null) {
+                        realm.close();
+                    }
+                }
+            }
+        });
+
+        // wait until the callback of our async query proceed
+        TestHelper.awaitOrFail(signalCallbackFinished, 6000);
```
cmelchior: Any reason this is 6000 instead of just using the default 7 (by removing the last argument)?

realm/src/androidTest/java/io/realm/RealmAsyncQueryTests.java
```diff
@@ -2157,6 +2158,67 @@ public void run() {
...
+        executorService.shutdownNow();
+        if (null != threadAssertionError[0]) {
+            // throw any assertion errors happened in the background thread
+            throw threadAssertionError[0];
+        }
+        if (backgroundLooper[0] != null) {
+            // failing to quit the looper will not execute the finally block responsible
+            // of closing the Realm
+            backgroundLooper[0].quit();
```
cmelchior: On slow devices, the finally block might not be executed until after the next unit test has started. This will cause it to fail because it cannot delete the Realm. It is generally better to have a latch at the end of the background thread to make sure that the unit test doesn't exit until all background tasks are done

realm/src/androidTest/java/io/realm/RealmAsyncQueryTests.java
```diff
@@ -2254,6 +2316,228 @@ public void run() {
...
+
+        executorService.shutdownNow();
+        if (null != threadAssertionError[0]) {
+            // throw any assertion errors happened in the background thread
+            throw threadAssertionError[0];
+        }
+        if (backgroundLooper[0] != null) {
+            // failing to quit the looper will not execute the finally block responsible
+            // of closing the Realm
+            backgroundLooper[0].quit();
```
cmelchior: Same

realm/src/androidTest/java/io/realm/RealmAsyncQueryTests.java
```diff
@@ -2254,6 +2316,228 @@ public void run() {
...
```
cmelchior: Same

realm/src/androidTest/java/io/realm/RealmAsyncQueryTests.java
```diff
@@ -2254,6 +2316,228 @@ public void run() {
...
```
cmelchior: same

realm/src/main/java/io/realm/Realm.java
```diff
@@ -961,7 +961,38 @@ boolean contains(Class<? extends RealmObject> clazz) {
         }
 
         TableView tableView = table.getDistinctView(columnIndex);
-        return new RealmResults(this, tableView, clazz);
+        return new RealmResults<E>(this, tableView, clazz);
+    }
+
+    /**
+     * Return a distinct set of objects of a specific class. As a Realm is unordered, it is undefined which objects are
+     * returned in cases of multiple occurrences.
```
cmelchior: in case

realm/src/main/java/io/realm/RealmQuery.java
cmelchior: Agree. It should be moved to Realm

realm/src/main/java/io/realm/Realm.java
nhachicha: we discussed this yesterday in #java the point is to fail fast, I know the check is done in C++ but this will fail in a background thread, whereas in this approach we make sure all the arguments are valid

realm/src/main/java/io/realm/RealmQuery.java
nhachicha: You can't do `realm.where(Dot.class).equalTo("name", "John").distinctAsync(5);` distinctAsync is package private (you can only access it UniTest) + this use the same async mechanism to deliver & update the TableView, moving it to Realm will change completely how we deliver/update async TableView & I don't see why

realm/src/androidTest/java/io/realm/RealmAsyncQueryTests.java
nhachicha: oops I was debugging, hence increasing the timeout

realm/src/main/java/io/realm/RealmQuery.java
cmelchior: Ups, missed that little detail. This if fine :+1:

realm/src/androidTest/java/io/realm/RealmAsyncQueryTests.java
nhachicha: I'm not sure about the behaviour you're describing.. will update maybe in a separate PR

cmelchior: This error e.g. showed up when the Nexus S was test device.

1) unit test 1 calls looper.exit()
2) BG looper exist, but hasn't reached `realm.close()` yet on the BG thread.
3) Unit test exists
4) Unit test framework call setup for next unit test
5) Setup method fails because it tries to delete a open Realm (the BG one)
6) Old unit test BG thread closes the Realm (to late)
7) Result is "Cannot delete open Realm" in the log

realm/src/main/java/io/realm/Realm.java
kneth: That's a good point. As a side note, we should reconsider the "check at JNI level" strategy in general.

kneth:
:+1:

cmelchior:
If unit tests are fixed in another PR it looks fine except two minor nitpicks (awaitOrFail + small language question in https://github.com/realm/realm-java/pull/1637#discussion-diff-42658717R969

nhachicha:
@realm/java

cmelchior:
Please create an issue for fixing the unit tests. They will bite us at some point if not fixed. Otherwise :+1:

PR #1648, opened 2015-10-23
fix possible null assignment to Realm.columnIndices.

### LINKED ISSUE #1646: NPE: ColumnIndices.getColumnInfo(java.lang.Class) after upgrading to 0.84.0
Hi team,

After upgrading from 0.82.2 to 0.84.0 got this exception during the first app start:
```
java.lang.NullPointerException: Attempt to invoke virtual method 'io.realm.internal.ColumnInfo io.realm.internal.ColumnIndices.getColumnInfo(java.lang.Class)' on a null object reference
```

Didn't happen during the next starts.
Same happened when I deployed the app on another device with the same 0.82.2 realm version.

Here is the piece of code which triggered the crash on the first device (called on the worker thread):
```java
realm.where(PinRealm.class)
                    .equalTo("token", pinToken)
                    .findFirst();
```

And here is the piece of code which triggered the crash on another device (called on the main thread):
```java
realm.where(PinRealm.class)
                    .equalTo("deleted", false)
                    .equalTo("inboxState", NEW.ordinal())
                    .findAll();
```

zaki50:
Problem:
  `Realm#lastLocalInstanceClosed()` which is called by `Realm#close()` can remove entries from `validatedRealmFiles` while other thread is executing `Realm#createAndValidate()`.
  This might lead to `null` assignment to `Realm.columnIndices`.

Solution:
  Modified to deal with `validatedRealmFiles` in synchronized block of `Realm.class`.

this PR fixes #1646

### DIFF
--- a/realm/src/androidTest/java/io/realm/RealmTest.java
+++ b/realm/src/androidTest/java/io/realm/RealmTest.java
@@ -16,6 +16,7 @@
 package io.realm;
 
 import android.content.Context;
+import android.os.SystemClock;
 import android.test.AndroidTestCase;
 
 import junit.framework.AssertionFailedError;
@@ -44,7 +45,6 @@ import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicBoolean;
 
 import io.realm.dynamic.DynamicRealmObject;
-import io.realm.entities.AllJavaTypes;
 import io.realm.entities.AllTypes;
 import io.realm.entities.AllTypesPrimaryKey;
 import io.realm.entities.AnnotationIndexTypes;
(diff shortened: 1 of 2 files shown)


zaki50:

beeender:
:+1:

dalinaum:
It's not about PR. But it bothers me a little :)
```
`Realm#close()' shoulde be  `Realm#close()`.
```

zaki50:
@dalinaum thank you. fixed.

dalinaum:
:+1:

bmunkholm:
@dalinaum @beeender : When you review, remember to review for the missing things - usually the changelog update :-)

dalinaum:
@bmunkholm I see. I will pay attention to missing things later. :)

Issue #1208, opened 2015-06-11
Thread handoff for RealmObject and RealmResults / Frozen state


cmelchior:
It should be possible to move objects and query results between threads seamlessly. The underlying concepts introduced in #896 should be useable for this as well.

sawcioo:
Any progress?

cmelchior:
@sawcioo Yes, are working on #896 first which will introduce async query capabilities. Then this will follow afterwards.

cmelchior:
Following the discussion on last weeks offsite-meeting. We gravitated towards the following pattern:

```
Foo obj = realm.where(Foo.class).findFirst();
Foo immutableCopy = obj.freeze(); Create a copy that is frozen to the current version. Any modification will throw.
new Thread(new Runnable() {
  @Override
  public void run() {
    immutableCopy.getBar(); // Will work
    obj.getBar(); // Will throw RealmException
  }
}).start();
```

Adding a `.freeze()` to `RealmObject/RealmResults/RealmList` will make it possible to break the normal Realm contract of always being up date. Which can make sense. Any RxJava support would then also only return frozen objects:

```
Observable<RealmResults<Foo>> foo = realm.where(Foo.class).findAll().observable():
foo.subscribeOn(Schedulers.io())
  .observeOn(AndroidSchedulers.mainThread()
  .subscribe(new Action1<RealmResults<Foo>>() {
            @Override
            public void call(RealmResults<Foo> results) {
                results.get(0); // Would also work without throwing Thread exception
        }
);
```

osama-lionheart:
Thanks for the updates @cmelchior!

but what exactly would be "frozen" when we call `freeze`? only that object fields, or any linked objects as well??

cmelchior:
It would give you a completely consistent view of all linked data as well. This is actually what is already happening today across threads. The `freeze()` method just creates the necessary plumbing internally for  moving stuff across threads in a safe manner as well as having the nice benefit that it is really explicit that your data are no longer "are up to date".

jpiche:
`freeze` makes a lot of sense, but will there be a corresponding method to easily re-query the `RealmResults` for an unfrozen version? Also, would there be a method on `RealmResults` like `isFrozen` to check if a list is frozen or not?

cmelchior:
[...]

Issue #865, opened 2015-02-18
RxJava support


jenzz:
I'm not sure if the Realm internals allow for RxJava support, but it would be nice to be able to do something like this:

```
// Get observable of all users
Realm.getInstance(context)
    .where(User.class)
    .findAll()
    .observable()
    .subscribe(users -> {
        // do stuff with users
    });

// Get observable of a single user
Realm.getInstance(context)
    .where(User.class)
    .findFirst()
    .observable()
    .subscribe(user -> {
        // do stuff with user
    });

// Get observable of users table count
Realm.getInstance(context)
    .where(User.class)
    .findAll()
    .size()
    .observable()
    .subscribe(count -> {
        // do stuff with count
    });
```

kakkoyun:
+1

cmelchior:
We definitely have RxJava on the radar, as it quite a cool framework. A first step for us is probably documenting what works and what doesn't when combining Realm and RxJava, but going forward adding some sort of support is not out of the question.

That said we are currently focusing on other features right now, so I cannot give you any timeframe just yet.

DHuckaby:
+1

cmelchior:
Some resources:
- https://github.com/ReactiveX/RxJava/wiki
- http://blog.danlew.net/2014/09/15/grokking-rxjava-part-1/
- https://www.youtube.com/watch?v=_t06LRX0DV0

kboyarshinov:
+1

Internal RxJava support would be great. For now I'm using it with Realm by following several restrictions:  

- One observable for one transaction.
- Mapping Realm objects to non-Realm (e.g. AutoValue) when passing between observer/subscriber threads.

cmelchior:
Hi @kboyarshinov 
Can you elaborate a bit on one observable pr. transaction?

I have been evaluating RxJava support as well, and the biggest blockers I can see so far is that we need fine-grained notifications to observe on queries and that it is not possible to move RealmObjects between threads which conflict with ```Observable.subscribeOn``` and ```Observerable.oberserveOn()```.

The second part is probably the biggest issue.

kboyarshinov:
[...]

PR #1651, opened 2015-10-23
[Design] RxJava support

cmelchior:
This PR contains a document describing how we could go about adding RxJava support to Realm.

See #865 for the issue and https://github.com/realm/realm-java/tree/cm-lab-rxjava for some earlier prototype code.

**Update 26/10-2015**
After some discussion, we are narrowing down to Option B:

- First class `.observable()` support for all Realm classes.
- RxJava as an optional dependency.
- Support RxJava 1.*, RxJava 2.* and possible RxMobile using factory classes for constructing the observables for the correct version of Rx*. Should be pluggable. Depending on the differences between the factory classes we might want to release this as a separate dependency. If we are adding custom schedulers making it a separate plugin also makes a lot more sense.
- Investigate if we can use a custom schedulers to avoid Realms thread restrictions.

**Update 07/12-2015**

Basic API (v1)

- Add the following API's that work on top of RealmChangeListener
  * `Realm.asObservable()`
  * `RealmObject.asObservable()`
  * `RealmResults.asObservable()`
  * `DynamicRealm.asObservable()`
  * `DynamicRealmObject.asObservable()`
  * `RealmList.asObservable()` (not yet supported)

Using `asObservable()` instead of `observable()`. It will be consistent with the wording used by e.g Subjects and will fit well when we add `asCursor()` ( #1090 )

- Add the following to support different user configured ways of creating Observables.
 * New interface: `RxObservableFactory`
 * New class: `RealmObservabableFactory` for automatically converting RealmChangeListener to Observables (default behavior).
 * Add `RealmConfiguration.Builder.rxFactory(RxObservableFactory)`

Advanced API (v2):

- Add the following to RealmQuery, so people can execute queries on other threads. It will require exposing query handover in some shape.
  * `RealmQuery.asObservable()`

 * New class: `DetachedCopyObservableFactory` variant that automatically uses `Realm.copyFromRealm` to copy RealmObject/RealmResults to memory. Ideally it should be possible to control which thread does the copying which require exposing object/RealmResult handover.

Implementation notes:
- Realm objects are still thread confined and will be automatically updated. This will not change before #1208 is implemented. It means that some operators in RxJava have to be used with care. `Realm.copyFromRealm` can provide a temporary work-around, but is not without cost itself. 

**Examples**:
```
  .rxFactory(new MySpecialRxJavaObservableFactory())
  .build();

// Async queries done in their own worker thread, all other operations can be done on the main thread
// beware of operators that work on special schedulers.
realm.where(Person.class).findAllAsync().asObservable()
  .filter(results.isLoaded()) // Only proceed once results are available
  .subscribe(showResultsOnScreen(results));

// Execute query on background/copy data to memory to work around thread confinement/live-updates at the cost of zero copy/consistency.
realm.where(AllPerson.class).asObservable()
  .flatmap(query.findAll().asObservable())
  .map(realm.copyFromRealm(results)) 
  .subscribeOn(Schedulers.computation())
  .subscribe(showResultsOnScreen(results)); 
```

@realm/java

### DIFF
--- a/RxJava.md
+++ b/RxJava.md
@@ -2,6 +2,60 @@ Proposal for RxJava API
 
 Starting to think about an API for this:
 
+
+## API Proposal
+
+###  PUBLIC API
+
+All objects with RealmChangeListeners should also expose an Observable
+```
+Realm.observable()
+RealmObject.observable()
+RealmResults.observable()
+RealmList.observable()
+```
+
+Calling any of these methods should detect if RxJava is actually on the classpath and throw a proper exception if not.
+Note that just doing `realm.observable()` would not even compile in IntelliJ because it would not be able to import
+the `rx.Observable` class.
+
+
+Note, we don't expose an `observable()` on RealmQuery like e.g. SQLBrite does, this is not needed as your async
+RealmResults actually have the same function while being semantically clearer. Listining to changes to a query is a bit 
+odd, while changes to a query result makes a lot more sense.
+
+It has the slight downside that you cannot defer work when creating the observable. You actually have to trigger the
+query first. If this is deemed undesirable, we can expose the following methods on RealmQuery which will mimic our 
+current `find*` methods.
+
+```
+RealmQuery.observeAll(); // = RealmQuery.findAllAsync();
+RealmQuery.observeAllSorted(String name); 
+RealmQuery.observeAllSorted(String name, boolean sortOrder); 
+RealmQuery.observeAllSorted(String name, boolean sortOrder, String name2, boolean sortOrder2); 
+RealmQuery.observeFirst(); 
+```
+
+### PRIVATE API
+**RxObservableFactory**: Interface for all Rx implementations. Will make it easy to swap between RxJava1, RxJava2 if
+  needed.
+
+### CONFIGURING REALM
+Realm supports both libraries and app code using Realm independantly. However due to class loading conflicts we should
+not worry about a library wanting to use RxJava 1 while the app wanted to use RxJava 2. 
+
+This means we should be able to detect automatically which version of RxJava is being used and instantiate the
+appropriate factory without user involvement.
+
+### Examples 
+
+Examples of using this API can be found here:
+https://github.com/realm/realm-java/pull/1710
+
+
+-------------
+Update 01/11-2015: Until something 
+
 ## How to include RxJava?
 
 First concern is how to actually include RxJava. I see 3 solutions
@@ -155,3 +209,6 @@ Summary: We can add RxJava support today, but need to list a number of caveats.
 These caveats will automatically be reduced in future releases without effecting 
 the Observable API. When both #1208 and #989 are implemented we will have full 
 RxJava support.
+
+
+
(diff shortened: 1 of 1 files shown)


fuwaneko:
I'm certainly against C option to include RxJava. While it might be easiest way to do it, libraries like Retrofit have moved towards hybrid A/B approach exactly because it's the worst way to do it too. I see it similar to Retrofit, there's an "adapter" addon which turns findAll() and similar methods to return Observable and it has to be registered on configuration step. And given the fact that Realm instances can be configured when calling getInstance(), programmer can temporarily enable/disable it too by simply providing/not providing adapter.

Also, it would be awesome for Observable<RealmResults> to emit onNext with updated results every time Realm changes. Or even better, when specific RealmResults change.

Should these Observables also be deferred? I.e. wait until at least one subscriber?

cmelchior:
The Adapter approach is really cool. Unfortunately it works for Retrofit because the user defines the interfaces, unlike Realm where we provide them.

Regarding onNext, then that is exactly what we intend to do. This is already happening today because RealmObjects auto-refresh themselves, so it is just a matter of putting RxJava semantics on top of our own ChangeListeners.

As I mention in the docs, then we can only defer work if you observe on a RealmQuery (which brings a few more headaches regarding our different query options). Observables created from e.g. RealmResults would already have done the work as it is similar to observing on a List.

fuwaneko:
@cmelchior Ah, I see, so unless RealmQuery is used, it's practically just async query wrapped in Observable?

As for adapter approach, I was just telling my thoughts. I kinda figured it's not directly applicable. In any case C approach would introduce more coupling and I'm on "opt-in" side instead of "opt-out" even though I always use RxJava on Android and it's not a big deal for me personally.

cmelchior:
@fuwaneko Yes, that is correct.

I totally agree with your point about case C. It is pretty much a choice between "less coupling and slightly worse API" or "stronger coupling and better API".

fuwaneko:
@cmelchior haha, in that case I vote for less coupling and B version. In any case, thanks for working on adding this.

lpessoa:
Would go for B option as well. Better keep it open to allow for adding support for future libs that may come up.

FranciscoE-Reuters:
Why not go with simple Futures, so RxJava is not baked into the codebase yet still usable.

cmelchior:
We had some talks internally about this, and by the looks of it we will probably go with some variant of proposal B. It has been summarized in the top post.

kboyarshinov:
I like the option B as well. Will be pretty straightforward to use. But I do not think that RxJava support should be added today even if it can be. Threading and notification issues are preferred to be resolved before imo. This won't confuse users and lead them to inappropriate usage scenarios. Would be good if RxJava support will be in 1.0 release. And if it do not it will be still better than unfinished RxJava compatible version. 

Anyway :+1: for making this.

cmelchior:
Support has been released as part of 0.87. Closing

tonyshkurenko:
What about RxJava 2 support?

kneth:
@tonyshkurenko Please create a new issue for RxJava 2 support. Thanks in advance.

PR #1661, opened 2015-10-24
Removed update() method from generated proxy class corresponding to un-updatable model.

### LINKED ISSUE #1660: Remove update() method from generated proxy class if model class has no primary key.

zaki50:
If model class has no primary key, we don't need to generate `update()` method in generated Proxy class.

fixes #1660

### DIFF
--- a/realm-annotations-processor/src/main/java/io/realm/processor/RealmProxyClassGenerator.java
+++ b/realm-annotations-processor/src/main/java/io/realm/processor/RealmProxyClassGenerator.java
@@ -689,6 +689,9 @@ public class RealmProxyClassGenerator {
     }
 
     private void emitUpdateMethod(JavaWriter writer) throws IOException {
+        if (!metadata.hasPrimaryKey()) {
+            return;
+        }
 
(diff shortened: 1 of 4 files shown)


zaki50:

dalinaum:
3 update method in tests are deleted. Aren't they necessary?

zaki50:
@dalinaum those are necessary change since test cases compare actual generated code to those files.
If there are any differences, the test case fails.

We must update those files when we have modification to generator which affects generated code.

changelog.txt
```diff
@@ -1,5 +1,6 @@
 0.84.1
  * Fixed a bug where simultaneous opening and closing a Realm on different threads might result in a NullPointerException (#1646).
+ * Removed update() method from generated proxy class corresponding to un-updatable model (#1660).
```
cmelchior: I don't think we need a changelog update for this. It is an internal change, and invisible to the user.

cmelchior:
Good catch :+1:

dalinaum:
:+1:

Issue #1647, opened 2015-10-23
Kotlin @Required annotation not being detected?


corbt:
I have a Realm database I'm updating from 0.81-SNAPSHOT to 0.84. I use Kotlin.

When I open the database, I get the following error:

```
java.lang.RuntimeException: Unable to create application com.emberall.kiosk.TheApp: io.realm.exceptions.RealmMigrationNeededException: Field 'updatedAt' is required. Either set @Required to field 'updatedAt' or migrate using io.realm.internal.Table.convertColumnToNullable().
...
```

My model is as follows:

```kotlin
public open @RealmClass class Recording(
    @PrimaryKey
    @Expose public open var id: String = UUID.randomUUID().toString(),
    @Expose @Required public open var createdAt: Date = Date(),
    @Expose @Required public open var updatedAt: Date = Date(),

    @Expose @Required public open var name: String = "",
    @Expose @Required public open var videoFileStatus: String = "none", // "none", "device", "uploaded", "queued", "processed"
    @Expose public open var videoFileSize: Long? = null
) : RealmObject()
```

So, it looks like Realm isn't recognizing my `@Required` annotation -- any idea why?

zaki50:
@corbt are you importing `io.realm.annotations.Required` correctly?

cmelchior:
Hi @corbt Do you get that error when upgrading an old database or starting from a fresh one as well?

corbt:
@zaki50 yes, I think I'm importing the annotations correctly

```kotlin
import io.realm.annotations.PrimaryKey
import io.realm.annotations.RealmClass
import io.realm.annotations.Required
```

@cmelchior the error appears when upgrading an old database. A fresh one appears to be working fine, although I don't know whether the field is really @Required or not -- I suspect it's still ignoring that annotation (the .realm file isn't opening in the Realm Browser on my computer for some reason, it claims it's in a format that can't be read).

cmelchior:
Yes, sorry our Realm browser is not updated yet to support the new Null values. We are getting a new version improved by Apple currently, but you can download it from here: https://github.com/realm/realm-browser-osx/releases

corbt:
[...]

PR #1662, opened 2015-10-24
Updated Kotlin example to 1.0.0 beta

cmelchior:
Related to #1647 

Updates our Kotlin example to be compatible with the latest Kotlin 1.0.0-beta 

@realm/java

### DIFF
--- a/examples/kotlinExample/src/main/kotlin/io/realm/examples/kotlin/KotlinExampleActivity.kt
+++ b/examples/kotlinExample/src/main/kotlin/io/realm/examples/kotlin/KotlinExampleActivity.kt
@@ -17,12 +17,12 @@
 package io.realm.examples.kotlin
 
 import android.app.Activity
-import android.os.AsyncTask
 import android.os.Bundle
 import android.util.Log
 import android.widget.LinearLayout
 import android.widget.TextView
 import io.realm.Realm
+import io.realm.RealmConfiguration
 import io.realm.examples.kotlin.model.Cat
 import io.realm.examples.kotlin.model.Dog
 import io.realm.examples.kotlin.model.Person
@@ -34,7 +34,7 @@ import kotlin.properties.Delegates
 public class KotlinExampleActivity : Activity() {
 
     companion object {
-        public val TAG: String = javaClass<KotlinExampleActivity>().getName()
+        public val TAG: String = KotlinExampleActivity::class.qualifiedName as String
     }
 
     private var rootLayout: LinearLayout by Delegates.notNull()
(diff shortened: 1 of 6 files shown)


zaki50:
:+1:

dalinaum:
:+1:

zaki50:
@cmelchior one more thing 
He said following configuration needed in build.gradle.

I'm confirming now.

```
kapt {
    generateStubs = true
}
```

cmelchior:
It doesn't seem to be a problem in our case. It is described here: http://blog.jetbrains.com/kotlin/2015/06/better-annotation-processing-supporting-stubs-in-kapt/

As far as I understand it you only need that if you have Kotlin code that references classes generated by the annotation processor.

dalinaum:
device-2015-10-25-030047

`Number of persons` of `basic Query operation` and `basic link Query operation` are `0`. It seems meaningless tests.

dalinaum:
Should this PR update `changelog.txt`?

cmelchior:
It probably should, yes. Didn't notice the output was wrong, will take another look

dalinaum:
It looks to me like we can remove `field:` after merging #1663

cmelchior:
Yes

cmelchior:
even before, because I don't use @Required in the example project

cmelchior:
I updated the example so it no longer uses `@field:`. Also rearranged some code so the output to the screen makes more sense.

kneth:
It looks fine to me (never really tried Kotlin).

examples/kotlinExample/src/main/kotlin/io/realm/examples/kotlin/KotlinExampleActivity.kt
```diff
@@ -49,6 +49,10 @@ public class KotlinExampleActivity : Activity() {
         // These operations are small enough that
         // we can generally safely run them on the UI thread.
 
+        // Create configuration and reset Realm.
+        val config = RealmConfiguration.Builder(this).build();
```
nhachicha: No need for semicolons in Kotlin 👍

examples/kotlinExample/src/main/kotlin/io/realm/examples/kotlin/KotlinExampleActivity.kt
```diff
@@ -49,6 +49,10 @@ public class KotlinExampleActivity : Activity() {
...
+        Realm.deleteRealm(config);
```
nhachicha: same

PR #1663, opened 2015-10-24
Add annotations to @Required for better Kotlin compatibility

corbt:
Currently (as documented in #1647) the `@Required` annotation, when used from a constructor in Kotlin, must be applied to a field explicitly using Kotlin's `@field:` syntax.

This change brings the annotations for the `@Required` interface in line with other Realm annotation interfaces to improve Kotlin compatibility. It follows the same format found in the `@Ignore`, `@Index` and `@PrimaryKey` annotations.

cc @zaki50

### DIFF
--- a/realm-annotations/src/main/java/io/realm/annotations/Required.java
+++ b/realm-annotations/src/main/java/io/realm/annotations/Required.java
@@ -15,6 +15,10 @@
  */
 
 package io.realm.annotations;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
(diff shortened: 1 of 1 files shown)


cmelchior:
Ah, so Kotlin automatically applies the annotation correctly if it is restricted to only fields, nice.
In any case, adding this makes the annotation more precise, and I see you already signed out CLA, so big :+1:

dalinaum:
:+1:

zaki50:
@corbt Thanks!

corbt:
Awesome thanks guys.

PR #1668, opened 2015-10-25
Add format check to latest version string obtained from our server.

### LINKED ISSUE #1666: Glibberish message on every project build
So Realm has suddenly started showing me this message everytime I run my project:

Realm gibberish message

Update: Removing annotations like `@Nullable`, `@NonNull` from my RealmObject getters fixes the issue. But this is weird considering I'm already on the latest version (assuming the message is an update-notification for the developer)

zaki50:
If a user is in the separate network and he/she needs http proxy setting for accessing the Internet,
latest version string may not com from our server.

To deal with that situation, this change adds the check that obtained latest version string has expected format.

@realm/java 

fixes #1666

### DIFF
--- a/realm-annotations-processor/src/main/java/io/realm/processor/RealmVersionChecker.java
+++ b/realm-annotations-processor/src/main/java/io/realm/processor/RealmVersionChecker.java
@@ -30,6 +30,7 @@ public class RealmVersionChecker {
 
     private static final String VERSION_URL = "http://static.realm.io/update/java?";
     private static final String REALM_VERSION = Version.VERSION;
+    private static final String REALM_VERSION_PATTERN = "\\d+\\.\\d+\\.\\d+";
     private static final int READ_TIMEOUT = 2000;
     private static final int CONNECT_TIMEOUT = 4000;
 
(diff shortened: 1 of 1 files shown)


zaki50:
related to #1666

dalinaum:
:+1:

realm-annotations-processor/src/main/java/io/realm/processor/RealmVersionChecker.java
```diff
@@ -80,7 +81,11 @@ private String checkLatestVersion() {
             conn.setConnectTimeout(CONNECT_TIMEOUT);
             conn.setReadTimeout(READ_TIMEOUT);
             BufferedReader rd = new BufferedReader(new InputStreamReader(conn.getInputStream()));
-            result = rd.readLine();
+            String latestVersion = rd.readLine();
+            // if obtained string does not match the pattern, we are in the separate network.
```
cmelchior: if *the* obtained string does not match the pattern, we are in *a* separate network.

Also, i'm not entirely sure what a "separate network" mean? You mean through a proxy or a isolated/private network that might change the response?

cmelchior:
Apart from a little nitpick about the comment :+1:

kneth:
:+1:

realm-annotations-processor/src/main/java/io/realm/processor/RealmVersionChecker.java
```diff
@@ -30,6 +30,7 @@
 
     private static final String VERSION_URL = "http://static.realm.io/update/java?";
     private static final String REALM_VERSION = Version.VERSION;
+    private static final String REALM_VERSION_PATTERN = "\\d+(?:\\.\\d+)*";
```
emanuelez: The pattern should be `\\d+\\.\\d+\\.\\d+` IMO

zaki50: fixed

emanuelez:
:+1:

realm-annotations-processor/src/main/java/io/realm/processor/RealmVersionChecker.java
zaki50: > Also, i'm not entirely sure what a "separate network" mean? You mean through a proxy or a isolated/private network that might change the response?

right

PR #1672, opened 2015-10-26
Check that double commit throws.

### LINKED ISSUE #1417: Calling commitTransaction twice should throw an exception
```java
realm.beginTransaction()
realm.commitTransaction()
realm.commitTransaction()
```

Above code should generate an exception.

In the latest core, a native exception will be triggered

cmelchior:
Fixes #1417 

With core upgraded to 0.93+, this now throws an error in Core. Since we still have an `immutable` boolean in group it makes sense to keep the check there. So this PR just adds the test to verify that the behaviour doesn't change.

@realm/java

### DIFF
--- a/realm/src/androidTest/java/io/realm/RealmTest.java
+++ b/realm/src/androidTest/java/io/realm/RealmTest.java
@@ -713,6 +713,16 @@ public class RealmTest extends AndroidTestCase {
         }
     }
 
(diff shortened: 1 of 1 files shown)


kneth:
:+1:

nhachicha:
:+1:

PR #1674, opened 2015-10-26
Null support for DynamicRealmObject

cmelchior:
This adds null-support to DynamicRealmObject.

@realm/java


realm/src/androidTest/java/io/realm/DynamicRealmObjectTest.java
```diff
@@ -95,6 +99,9 @@ public void testGetXXXIllegalFieldNameThrows() {
 
         // Test all getters
         for (SupportedType type : SupportedType.values()) {
+
+            // We cannot modularize everything, so STRING is a special case with it's own set
```
kneth: "it's" -> "its"

realm/src/androidTest/java/io/realm/DynamicRealmObjectTest.java
```diff
@@ -231,29 +238,121 @@ public void testTypedGetterSettersXXX() {
...
                 switch (type) {
                     case OBJECT:
-                        dObj.setObject(AllJavaTypes.FIELD_OBJECT, null);
-                        assertNull(dObj.getObject(AllJavaTypes.FIELD_OBJECT));
+                        NullTypes childObj = new NullTypes();
+                        childObj.setId(1);
+                        DynamicRealmObject dynamicChildObject = new DynamicRealmObject(realm.copyToRealm(childObj));
+
+                        dObj.setObject(NullTypes.FIELD_OBJECT_NULL, dynamicChildObject);
+                        dObj.setObject(NullTypes.FIELD_OBJECT_NULL, null);
```
beeender: Is this covered by `testTypedSetNull`?

realm/src/androidTest/java/io/realm/DynamicRealmObjectTest.java
```diff
@@ -274,6 +373,44 @@ public void testSetObjectWrongTypeThrows() {
         }
     }
 
+    public void testSetListWrongTypeThrows() {
+        realm.beginTransaction();
```
kneth: don't you miss `realm.cancelTransaction()`?

realm/src/androidTest/java/io/realm/DynamicRealmObjectTest.java
```diff
@@ -414,13 +521,16 @@ public void testUntypedSetUsingStringConversion() {
                         dObj.set(AllJavaTypes.FIELD_DATE, "1000");
                         assertEquals(new Date(1000), dObj.getDate(AllJavaTypes.FIELD_DATE));
 
-                    // These types don't have a string representation that should be parsed.
+                    // These types doesn't have a string representation that could be parsed.
```
kneth: "doesn't" -> "don't"

realm/src/main/java/io/realm/DynamicRealmObject.java
```diff
@@ -177,10 +199,13 @@ public double getDouble(String fieldName) {
 
     /**
      * Returns the {@code byte[]} value for a given field.
+     * If the field is nullable use {@link #isNull(String)} to check for {@code null} instead of using
+     * this method.
```
beeender: Why don't just return `null` when the `byte[]` field is null?

realm/src/main/java/io/realm/DynamicRealmObject.java
```diff
@@ -95,10 +96,13 @@ public DynamicRealmObject(RealmObject obj) {
 
     /**
      * Returns the {@code boolean} value for a given field.
+     * If the field is nullable use {@link #isNull(String)} to check for {@code null} instead of using
```
kneth: "nullable" -> "not required"?

beeender:
Test cases are missing for test setters/getters on `@Required`/not-nullable fields.

And It looks a bit strange that `getter` returns primitive type, but `setter` takes boxed type as param. But i am fine with it.

realm/src/main/java/io/realm/DynamicRealmObject.java
```diff
@@ -177,10 +199,13 @@ public double getDouble(String fieldName) {
 
     /**
      * Returns the {@code byte[]} value for a given field.
+     * If the field is nullable use {@link #isNull(String)} to check for {@code null} instead of using
```
kneth: Can't we do like we do for `Date`? That is, return `null` if the value is `null`?

realm/src/main/java/io/realm/DynamicRealmObject.java
```diff
@@ -535,25 +560,59 @@ public void setObject(String fieldName, DynamicRealmObject value) {
...
-    public void setList(String fieldName, RealmList<DynamicRealmObject> list) {
+    public void setList(String fieldName, RealmList<? extends RealmObject> list) {
+        if (list == null) {
+            throw new IllegalArgumentException("Null values not allowed for lists");
+        }
+
+        String tableType = row.getTable().getName();
+        boolean typeValidated;
+        if (list.className == null && list.clazz == null) {
+            // Standalone lists doesn't know anything about the types they contain. They might even hold objects of
```
kneth: "doesn't" -> "don't"

realm/src/androidTest/java/io/realm/DynamicRealmObjectTest.java
cmelchior: Yes, that part was. Good catch.

realm/src/androidTest/java/io/realm/DynamicRealmObjectTest.java
cmelchior: Yes, that is on purpose. The Realm is closed on `tearDown` so that will effectively cancel the transaction at the same time.

realm/src/main/java/io/realm/DynamicRealmObject.java
cmelchior: Yes, we should do that. It has been fixed.

realm/src/main/java/io/realm/DynamicRealmObject.java
cmelchior: Yes, Date/String/byte[] should be allowed to return null.

realm/src/main/java/io/realm/DynamicRealmObject.java
cmelchior: Nullable isn't great, but "not required" has even less meaning IMO as it require people actually knowing about the @Required annotation and fully understanding what it does. At least nullability as a concept is hopefully familiar to most.

cmelchior:
Addressed all comments @realm/java 
Also added new unit tests to test `@Required` columns.

beeender:
:+1:

realm/src/main/java/io/realm/DynamicRealmObject.java
kneth: I fully understand that. My point is that we wish to minimise the number of concepts used in our documentation.

realm/src/androidTest/java/io/realm/DynamicRealmObjectTest.java
```diff
@@ -414,13 +552,16 @@ public void testUntypedSetUsingStringConversion() {
                         dObj.set(AllJavaTypes.FIELD_DATE, "1000");
                         assertEquals(new Date(1000), dObj.getDate(AllJavaTypes.FIELD_DATE));
 
-                    // These types don't have a string representation that should be parsed.
+                    // These types don't have a string representation that could be parsed.
```
kneth: "could" -> "can"

realm/src/androidTest/java/io/realm/DynamicRealmObjectTest.java
```diff
@@ -455,14 +596,17 @@ public void testUntypedSetIllegalImplicitConversionThrows() {
                             dObj.set(AllJavaTypes.FIELD_DATE, "foo");
                             break;
 
-                        // These types doesn't have a string representation that should be parsed.
+                        // These types don't have a string representation that should be parsed.
```
kneth: "should" -> "can"

realm/src/androidTest/java/io/realm/TestHelper.java
```diff
@@ -225,6 +225,15 @@ public void e(String message, Throwable t) {
         };
     }
 
+    public static String getRandomString(int length) {
+        Random r = new Random();
+        StringBuilder sb = new StringBuilder(length);
+        for (int i = 0; i < length; i++) {
+            sb.append((char) r.nextInt(128));
```
kneth: why 128?

realm/src/main/java/io/realm/DynamicRealm.java
```diff
@@ -32,6 +32,8 @@
  * DynamicRealm is a dynamic variant of {@link io.realm.Realm}. This means that all access to data and/or queries are
  * done using string based class names instead of class type references.
  *
+ * This is useful during migrations or when when working with string-based data like from CSV or XML files.
```
kneth: "when when" -> "when"

realm/src/main/java/io/realm/DynamicRealmObject.java
```diff
@@ -535,25 +555,59 @@ public void setObject(String fieldName, DynamicRealmObject value) {
...
+     * @throws IllegalArgumentException if field name doesn't exists, it is not a list field or the type
      * of the object represented by the DynamicRealmObject doesn't match.
      */
-    public void setList(String fieldName, RealmList<DynamicRealmObject> list) {
+    public void setList(String fieldName, RealmList<? extends RealmObject> list) {
+        if (list == null) {
+            throw new IllegalArgumentException("Null values not allowed for lists");
+        }
+
+        String tableType = row.getTable().getName();
```
kneth: `tableType` -> `tableName`

realm/src/main/java/io/realm/DynamicRealmObject.java
cmelchior: Yes, that definitely also has a lot of merit. However the only reason we went with `@Required` was because the `Nullable/NonNull/<insert other>` was overloaded by too many other libraries/frameworks. I think we all agreed that `Nullable/NotNull` was a lot better in term of expressing intent? That still leaves the question what we should use when describing what a method actually does.

Ping rest of @realm/java. What do you think?

realm/src/main/java/io/realm/DynamicRealmObject.java
```diff
@@ -535,25 +555,59 @@ public void setObject(String fieldName, DynamicRealmObject value) {
...
         long columnIndex = row.getColumnIndex(fieldName);
         LinkView links = row.getLinkList(columnIndex);
         links.clear();
-        for (DynamicRealmObject obj : list) {
+        for (int i = 0; i < list.size(); i++) {
+            RealmObject obj = list.get(i);
+            if (!typeValidated) {
+                String elementType = obj.row.getTable().getName();
+                if (!tableType.equals(elementType)) {
+                    throw new IllegalArgumentException(String.format("Element at index %s is not the proper type. " +
```
kneth: `index %s` -> `index %d`

kneth:
:+1:

PR #1678, opened 2015-10-26
Make security copies of the key in RealmConfiguration

emanuelez:
Saving and providing a mutable reference of the key is both a security threat and possibly the origin of subtle bugs.
We now make security copies of the byte arrays to solve the issue.

### DIFF
--- a/realm/src/androidTest/java/io/realm/RealmConfigurationTest.java
+++ b/realm/src/androidTest/java/io/realm/RealmConfigurationTest.java
@@ -17,6 +17,7 @@
 package io.realm;
 
 import android.test.AndroidTestCase;
+import android.test.MoreAsserts;
 
 import java.io.File;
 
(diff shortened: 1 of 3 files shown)


emanuelez:
@realm/java

emanuelez:
Changelog: Fixed a bug which allowed to externally modify the encryption key in a RealmConfiguration

beeender:
:+1:

kneth:
:+1:

cmelchior:
Shouldn't you do the copy when calling `encryptionKey(key)` as well?

emanuelez:
@cm That's the second one (line 299)

cmelchior:
Missing a unit test for it, but otherwise :+1:

emanuelez:
retest this please

realm/src/androidTest/java/io/realm/RealmConfigurationTest.java
```diff
@@ -416,4 +417,20 @@ public void testDifferentDirSameName() {
...
+    public void testKeyStorage() throws Exception {
+        // Generate a key and use it in a RealmConfiguration
+        byte[] oldKey = TestHelper.getRandomKey(12345);
+        byte[] key = oldKey;
+        RealmConfiguration config = new RealmConfiguration.Builder(getContext()).encryptionKey(key).build();
+
+        // Generate a different key and assign it to the same variable
+        byte[] newKey = TestHelper.getRandomKey(67890);
+        MoreAsserts.assertNotEqual(key, newKey);
+        key = newKey;
```
cmelchior: isn't the problem more if you do `key[0] = 42`, i.e changing the value inside the array?

PR #1687, opened 2015-10-28
Deep link query crashes cores.

cmelchior:
Exposing regression from Realm `0.83.0` to `0.83.1`


cmelchior:
Fastmerge /cc @realm/java

PR #1695, opened 2015-10-29
convertColumnToNotNullable now works correctly.

cmelchior:
`Table.convertColumnToNotNullable` did not create a non-nullable column, but a nullable one.

@realm/java


beeender:
:+1: It was a nice feature!

zaki50:
:+1:

Issue #1676, opened 2015-10-26
Change listener not being triggered on realm updates


Wattos:
According to the documentation:

*"If you want to be notified when the query completes and the RealmResults is updated, you can register a RealmChangeListener. This listener will be called every time the RealmResults is updated to reflect the latest changes in the Realm (usually after a commit)"*

I add a change listener to either a RealmObject or RealmResult, but that only gets triggered once (after the data is loaded). The documentation states that this listener should also be triggered when the real result is updated. That does not seem to be the case. 

Is the documentation incorrect or do I need to do something special besides commiting a transaction so that the listener gets triggered?

cmelchior:
Hi @Wattos 
The ChangeListener should be triggered on each update, not only when the data is loaded. If this isn't happening it sounds like a bug.

nhachicha:
Hi @Wattos 
It would be helpful if you can send us some code to reproduce this (either here or privately to help@realm.io)
Cheers,

Wattos:
@nhachicha Do you think that the code is really necessary? I did a quick search in the realm-java repository and the notifyChangeListener method is only called 3 times:

vikas-fueled:
I have encountered similar issue with when at times change listener is not being triggered even once for the same query, but is triggered when I query it again. `isLoaded` returns false in both cases on checking on same thread after registering listener.
```
public void fetchEpisodes(long seasonId) {
        logger.debug("fetch episodes for Season id :" + seasonId);
        final Realm realm = Realm.getDefaultInstance();
        RealmQuery<Content> query = realm.where(Content.class)
                .equalTo("season.id", seasonId);
        final RealmResults<Content> contents = query
                .findAllAsync();
        contents.addChangeListener(new RealmChangeListener() {
            @Override
            public void onChange() {
                logger.debug("Episodes size :" + contents.size());
                if (adapter != null) {
                    adapter.setEpisodes(contents, coverHandlerProvider);
                }
            }
        });
        logger.debug("Sync Episodes size :" + contents.size() + ", is Loaded :" + contents.isLoaded());
    }
```

vikas-fueled:
From [this discussion](https://github.com/realm/realm-java/issues/1675) seems like anonymous listeners should be fine for `RealmResults`. However when I changed my listener to a concrete one, the above issue was solved. So, this might help in investigating further.

cmelchior:
[...]

PR #1696, opened 2015-10-29
Verifying that RealmResults listeners are getting called on Realm updates.

cmelchior:
Verifying the behavior reported in https://github.com/realm/realm-java/issues/1676.
It looks like Realm is behaving correctly.

@nhachicha I wasn't entirely sure this use case was covered by unit tests. Can you confirm that? If not we should add it, otherwise I will close the PR again


cmelchior:
Closing as this unit test has been ported to another PR (no reference yet).

Issue #1699, opened 2015-10-29
As an library developer I want to be able to read default configuration


bryanjswift:
Inside my library when getting a `Realm` reference I would like to read the default `RealmConfiguration` held by `Realm` in order to determine the durability setting it is using. By reading the default `RealmConfiguration` a developer using my library can rely on setting the default configuration to `inMemory()` in their tests and the library code can respect the developer's choice.

cmelchior:
Not that I am inherently against the idea, but I am not sure I understand the use case? As a library developer you don't have any hooks into Realm except exposing a RealmModule the user can use if they choose? What exactly in the library code depends on the Users choice of persistence?

bryanjswift:
The library (SDK) I'm working on will hide the implementation details of the storage engine used within the library. So the user of the library / SDK says 'persist this object' and the library handles the persistence and returns an object. My use case is if the application developer is _also_ using Realm for persistence and sets the default `RealmConfiguration` to be in memory I want the library to respect that choice even if the developer doesn't tell the library elsewhere to run in memory.

PR #1700, opened 2015-10-29
Add a method to read the default configuration

bryanjswift:
Refs realm/realm-java#1699

A `RealmConfiguration` instance is read-only so it is safe to expose the
default configuration used to create `Realm` instances.

### DIFF
--- a/realm/realm-library/src/main/java/io/realm/Realm.java
+++ b/realm/realm-library/src/main/java/io/realm/Realm.java
@@ -233,6 +233,17 @@ public final class Realm extends BaseRealm {
     }
 
     /**
+     * Returns the {@link io.realm.RealmConfiguration} instance defined by
+     * {@link #setDefaultConfiguration(RealmConfiguration)}.
+     *
+     * @return the configuration set in {@link #setDefaultConfiguration(RealmConfiguration)}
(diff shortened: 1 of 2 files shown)


realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -233,6 +233,18 @@ public static void setDefaultConfiguration(RealmConfiguration configuration) {
     }
 
     /**
+     * Returns the {@link io.realm.RealmConfiguration} instance defined by
+     * {@link #setDefaultConfiguration(RealmConfiguration)}.
+     *
+     * @return the configuration set in {@link #setDefaultConfiguration(RealmConfiguration)}
+     *
```
cmelchior: Remove space.

cmelchior:
Hi @bryanjswift. As I wrote in #1699 I am not entirely sure I understand the use case for this. That said I can see why it might be relevant to access the default configuration so I wouldn't be against adding it.

Please add a unit test that verifies the behaviour and update the `changelog.txt` as well. After that it looks pretty good.

zaki50:
retest this please

dalinaum:
retest this please

cmelchior:
Hi @bryanjswift 
Looks like you forgot the unit test?

Issue #1693, opened 2015-10-29
nested Link Object isNull condition crashed


chouex:
Getting a crash reports when i try nested Linked isNull condition.

```
io.realm.exceptions.RealmError: Unrecoverable error. Combining link() and is_null() is currently not supported in io_realm_internal_TableQuery.cpp line 1497
    at io.realm.internal.TableQuery.nativeIsNull(Native Method)
    at io.realm.internal.TableQuery.isNull(TableQuery.java:677)
    at io.realm.RealmQuery.isNull(RealmQuery.java:249)
    at ***(SomeClass.java:80)
```

```
RealmQuery<ComplexSightseeing> query = realm.where(Level0.class);
query.isNotNull("Level1");
query.isNull("Level1.Level2");  //line 80
```

where Level0,Level1,Level2 class is

```
class Level0 extends RealmObject{
    @PrimaryKey
    private int id;


    private Level1 Level1;
...
}


class Level1 extends RealmObject{
    @PrimaryKey
    private int id;

    private Level2 Level2;
...
}


class Level2 extends RealmObject{
    @PrimaryKey
    private int id;

...
}
```

```
compile 'io.realm:realm-android:0.84.1' 
```

android 4.4.4

zaki50:
Hi @chouex 
Thank you for reporting.

I'll check that.

zaki50:
I confirmed this issue.
And it's a limitation of current underlying storage engine.

So, we should document this behavior as a current limitation.

chouex:
temporary solution is to create another column define wether link is null

cmelchior:
`isNull` for linked fields are now working, see https://github.com/realm/realm-java/issues/1367

andriy-svirskyy:
> `isNull` for linked fields are now working, see #1367

Still getting exception `Illegal Argument: isNotNull() by nested query for link field is not supported.`

Realm-android 10.5.0

PR #1702, opened 2015-10-30
Change to is_not_null for links

beeender:
Since it is supported by core now.

https://github.com/realm/realm-core/pull/1185


zaki50:
@beeender only available `isNotNull()`, and not `isNull()` yet?

beeender:
@zaki50 It is available before. Before this change, we are using `not().is_null()` for `isNotNull` query since there is no native support for it in core at that time. This change is just to adapt the new core interface.

zaki50:
So you don't mean this issue fixes #1693

beeender:
No, it doesn't. It doesn't change anything user can feel. Just to use the proper core interface.

zaki50:
@beeender I understand. thanks

zaki50:
:+1:

cmelchior:
:+1:

PR #1710, opened 2015-11-01
Add RxJava support 

### LINKED ISSUE #865: RxJava support

cmelchior:
This PR fixes #865 

This PR is a feature branch for collecting all RxJava sub-branches before merging. Right now the roadmap is this:

- [x] 1) This branch adds simple RxJava support by adding RealmConfiguration methods / Factory interface and a simple implementation supporting RealmObject/RealmResults/Realm.

- [x] 2) Sample project showing the current caveats and work-arounds to be mindful off (#1940)

- [x] 3) Add support for RealmQuery observable. We need to expose handover of queries to other threads. Being tracked in #1955 .

- [x] 4) Custom `DetachedCopyObservableFactory` class for making all RealmData in-memory data. We need to expose handover for RealmObject/RealmResults. -> Probably not possible  to do transparently before #1208 at which it most likely becomes moot anyway. 

- [x] 5) Add Support for RealmList (not relevant before fine-grained notification and changelisteners have been added). Being tracked in #1956  .

1 + 2 is required before merging to main branch, 3 would be nice to have, 4 + 5 will probably take a little longer and will tie into handover mechanisms already considered elsewhere.

-----
API: (for step 1)
```
Realm.asObservable();
RealmObject.asObservable();
RealmResults.asObservable();
DynamicRealm.asObservable();
DynamicRealmObject.asObservable();
RealmConfiguration.Builder.rxFactory(RxObservableFactory factory);
```

Note that all Rx classes are in a new package `io.realm.rx`. Should we move these into `io.realm` as that currently holds all our public API? It doesn't conflict with the proposal of adding `io.realm.android` (as RxJava is not Android only), but something to consider when adding future classes/functionality.

--------
- [x] Finalize API proposal in #1651. Especially if RxObservableFactory should move to its own dependency. Conclusion: Keep factory classes in repo for now. Other questions: 
   1) We cannot have rx classes in a `io.realm.rx` without making BaseRealm public, is having them in `io.realm` OK? Otherwise should we make BaseRealm public but perhaps exclude it from JavaDoc? Answer: Added methods explicitly naming Realm/DynamicRealm to work around this.
   2) Should we expose observable methods on RealmQuery? YES: See other PR
   3) RealmObject is now `RealmObject<E extends RealmObject<E>>` which is slightly painful if you want type safe observables for RealmObjects. You need to do `public class Person extends RealmObject<Person>`. It is not mandatory though. Answer: YES, IntelliJ will give you warning if no generic parameter is given, but otherwise it works fine in all other cases, i.e. only RxJava users will notice it.
- [x] Fix changelistener bug #1884 
- [x] Fix changeListeners not triggering for synchronous queries on RealmResults/RealmObjects ( #1676). 
- [x] Use `copyFromRealm` for `InMemoryObservableFactory` ( #1849 )
- [x] Investigate the possibility to use a custom RealmScheduler to facilitate moving across threads. Just like our current async API does now. Conclusion: Not possible as the scheduler does not have access to the parameters being parsed around. With auto-closable Realms it will be possible, or wait for freeze() semantics.
- [x] More examples of Observables with 'gotchas' have to be added.
- [x] Throw proper error message if calling `observable()` and RxJava is not on the classpath


realm/realm-library/src/main/java/io/realm/RealmObject.java
```diff
@@ -81,7 +82,7 @@
  */
 
 @RealmClass
-public abstract class RealmObject {
+public abstract class RealmObject<E extends RealmObject> {
```
emanuelez: Ouch :( That's not pretty...

cmelchior: Not it is horrible. Fortunately it is only required for RealmObjects as both RealmResults and RealmList has the type information. 

I tried different versions of generics black magic, but AFAIK it isn't possible for an abstract class to return subclass type without it being explicitly defined somewhere (either as a generic like here, or as input parameter to a method)

emanuelez: It kind of makes me reconsider `RealmObservable.from()` and allow users to throw anything at it, a bit like http://reactivex.io/RxJava/javadoc/rx/Observable.html#from(java.util.concurrent.Future)

cmelchior: Yes, the pattern is not unknown from RxJava itself: `Observable.from()`, `Observable.just()`, but it really really break up the flow when you are coding. I tried it when working on the examples.

Upside is that you don't actually *need* the generic, but if you don't add it you just get Object as the return type in the Observable chain.

I have been contemplating if we should add an explicit ListenerInterface, but that just seems like unnecessary boilerplate

realm/realm-library/src/main/java/io/realm/rx/RealmObservableFactory.java
```diff
@@ -0,0 +1,250 @@
...
+    @Override
+    public Observable<Realm> from(final Realm realm) {
+        checkRxJavaAvailable();
+        return Observable.create(new Observable.OnSubscribe<Realm>() {
+            @Override
+            public void call(final Subscriber<? super Realm> subscriber) {
+                final RealmChangeListener listener = new RealmChangeListener() {
+                    @Override
+                    public void onChange() {
+                        subscriber.onNext(realm);
```
austynmahoney: I believe this should be checking `subscriber.isUnsubscribed()` before calling `onNext` to follow the Rx contract. Same with all the other `from` methods.

More info:
http://ryanharter.com/blog/2015/07/07/wrapping-existing-libraries-with-rxjava/
https://www.reddit.com/r/androiddev/comments/3vebo3/resubmission_respect_the_observable_contract_when/

cmelchior: Hey @austynmahoney 
Thanks for the pointers. However I added the `subscriber.add(...)` below to prevent that exact scenario. That should unregister the listener when unsubscribing which means it shouldn't be needed to check for `subscriber.isUnsubscribed` in the callback. I won't rule out I made a mistake somewhere though, so i'll go over it again :)

Source: http://stackoverflow.com/questions/26695125/how-to-get-notified-of-a-observers-unsubscribe-action-in-a-custom-observable-in

austynmahoney: I'd still vote for following the contract and adding the check anyways. If some code changes down the road, or the unsubscribe happens on another scheduler, it could end up breaking the contract.

cmelchior:
Retest this please.

cmelchior:
retest this please

cmelchior:
Ready for review 
@realm/java  
@donnfelker

realm/realm-library/src/androidTest/java/io/realm/RxJavaTests.java
```diff
@@ -0,0 +1,288 @@
+package io.realm;
```
kneth: Standard header?

cmelchior: Added

kneth:
I see no tests for `RealmQuery` and `ReamList`. We should check if the correct exception is thrown - just to ensure that the JavaDoc is consistent with the behaviour.

cmelchior:
We don't have to write any tests for them yet, as the API is not exposed, so there is nothing that can crash. I added `from(RealmQuery)` and `from(RealmList)` to make the interface future-proof and avoid having to introduce breaking changes to it later.

cmelchior:
retest this please

realm/realm-library/src/main/java/io/realm/rx/RxObservableFactory.java
```diff
@@ -0,0 +1,134 @@
...
+/**
+ * Factory interface for creating Rx Observables for Realm classes.
+ */
+public interface RxObservableFactory {
+
+    /**
+     * Creates an Observable for a {@link Realm}. It should emit the initial state of the Realm when subscribed to and
+     * on each subsequent update of the Realm.
+     *
+     * Realm observables are hot observables as Realms automatically are kept up to date.
```
nhachicha: > Realm observables are hot observables as Realms automatically are kept up to date.

as Realms are automatically kept up to date.

realm/realm-library/src/androidTest/java/io/realm/RxJavaTests.java
```diff
@@ -0,0 +1,288 @@
...
+        assertEquals(2, subscriberCalled.get());
+    }
+
+    public void testAsyncObjectEmittedOnSubscribe() {
+        realm.beginTransaction();
+        realm.createObject(AllTypes.class);
+        realm.commitTransaction();
+
+        final AtomicBoolean subscribedNotified = new AtomicBoolean(false);
+        final AllTypes asyncObj = realm.where(AllTypes.class).findFirst();
```
nhachicha: `findFirstAsync()` instead

realm/realm-library/src/androidTest/java/io/realm/RxJavaTests.java
```diff
@@ -0,0 +1,288 @@
...
+        });
+        assertTrue(subscribedNotified.get());
+    }
+
+    public void testAsyncObjectEmittedOnUpdate() {
+        final AtomicInteger subscriberCalled = new AtomicInteger(0);
+        realm.beginTransaction();
+        AllTypes obj = realm.createObject(AllTypes.class);
+        realm.commitTransaction();
+        realm.where(AllTypes.class).findFirst().<AllTypes>asObservable().subscribe(new Action1<AllTypes>() {
```
nhachicha: `findFirstAsync()` instead

realm/realm-library/src/androidTest/java/io/realm/RxJavaTests.java
```diff
@@ -0,0 +1,288 @@
...
+                assertTrue(rxResults == results);
+                subscribedNotified.set(true);
+            }
+        });
+        assertTrue(subscribedNotified.get());
+    }
+
+    public void testAsyncResultsEmittedOnUpdate() {
+        final AtomicInteger subscriberCalled = new AtomicInteger(0);
+        realm.allObjects(AllTypes.class).asObservable().subscribe(new Action1<RealmResults<AllTypes>>() {
```
nhachicha: `realm.allObjects` is not creating async RealmResults

realm/realm-library/src/androidTest/java/io/realm/RxJavaTests.java
```diff
@@ -0,0 +1,288 @@
...
+        assertEquals(1, realm.handlerController.changeListeners.size());
+        new Thread(new Runnable() {
+            @Override
+            public void run() {
+                try {
+                    subscription.unsubscribe();
+                    fail();
+                } catch (IllegalStateException ignored) {
+                }
+                unsubscribeCompleted.countDown();
```
nhachicha: shouldn't this be in a `finally` block?

realm/realm-library/src/main/java/io/realm/RealmObject.java
```diff
@@ -263,6 +264,34 @@ public final void removeChangeListeners() {
     }
 
     /**
+     * Returns an Rx Observable that monitors changes to this RealmObject. It will emit the current object when
+     * subscribed to.
+     *
+     * If chaining a RealmObject observable use {@code obj.<MyRealmObjectClass>asObservable()} to parse on
```
nhachicha: > parse

pass ?

cmelchior:
Issues fixed @realm/java

kneth:
:+1: (you have responded to my comments)

realm/realm-library/src/androidTest/java/io/realm/RxJavaTests.java
```diff
@@ -0,0 +1,305 @@
...
+
+    public void testAsyncObjectEmittedOnUpdate() {
+        final AtomicInteger subscriberCalled = new AtomicInteger(0);
+        realm.beginTransaction();
+        AllTypes obj = realm.createObject(AllTypes.class);
+        realm.commitTransaction();
+        realm.where(AllTypes.class).findFirstAsync().<AllTypes>asObservable().subscribe(new Action1<AllTypes>() {
+            @Override
+            public void call(AllTypes rxObject) {
+                subscriberCalled.addAndGet(1);
```
emanuelez: This could be `subscriberCalled.incrementAndGet()`

realm/realm-library/src/androidTest/java/io/realm/RxJavaTests.java
```diff
@@ -0,0 +1,305 @@
...
+
+        Observable<CyclicType> obs = obj.asObservable();
+        obs.subscribe(new Action1<CyclicType>() {
+            @Override
+            public void call(CyclicType cyclicType) {
+                fail();
+            }
+        }, new Action1<Throwable>() {
+            @Override
+            public void call(Throwable throwable) {
```
emanuelez: Can we call this `ignored`?

realm/realm-library/src/main/java/io/realm/RealmResults.java
```diff
@@ -821,6 +822,29 @@ public void removeChangeListeners() {
...
+     */
+    @SuppressWarnings("unchecked")
+    public Observable<RealmResults<E>> asObservable() {
+        if (realm instanceof Realm) {
+            return realm.configuration.getRxFactory().from((Realm) realm, this);
+        } else if (realm instanceof DynamicRealm) {
+            DynamicRealm dynamicRealm = (DynamicRealm) realm;
+            RealmResults<DynamicRealmObject> dynamicResults = (RealmResults<DynamicRealmObject>) this;
+            @SuppressWarnings("UnnecessaryLocalVariable")
+            Observable results = realm.configuration.getRxFactory().from(dynamicRealm, dynamicResults);
```
emanuelez: Why introducing this explicit variable and suppressing the warning?

realm/realm-library/src/main/java/io/realm/rx/RealmObservableFactory.java
```diff
@@ -0,0 +1,225 @@
...
+                };
+                realm.addChangeListener(listener);
+                subscriber.add(Subscriptions.create(new Action0() {
+                    @Override
+                    public void call() {
+                        realm.removeChangeListener(listener);
+                    }
+                }));
+
+                // Immediately call onNext with the current value as due to Realms auto-update it will be the latest
```
emanuelez: Hard to read. I would add commas around `due to Realm's auto-update`

realm/realm-library/src/main/java/io/realm/rx/RealmObservableFactory.java
```diff
@@ -0,0 +1,225 @@
...
+                };
+                object.addChangeListener(listener);
+                subscriber.add(Subscriptions.create(new Action0() {
+                    @Override
+                    public void call() {
+                        object.removeChangeListener(listener);
+                    }
+                }));
+
+                // Immediately call onNext with the current value as due to Realms auto-update it will be the latest
```
emanuelez: Same

emanuelez:
:+1: apart from very minor comments

realm/realm-library/src/main/java/io/realm/RealmResults.java
cmelchior: It is to satisfy the generic type checks. We are allowed to return `Observable` but not `Observable<RealmResults<DynamicRealmObject>>`. Blame the generics system  :(

cmelchior:
retest this please

realm/realm-library/src/main/java/io/realm/rx/RealmObservableFactory.java
```diff
@@ -0,0 +1,225 @@
...
+                subscriber.add(Subscriptions.create(new Action0() {
+                    @Override
+                    public void call() {
+                        results.removeChangeListener(listener);
+                    }
+                }));
+
+                // Immediately call onNext with the current value, as due to Realms auto-update, it will be the latest
+                // value.
+                subscriber.onNext(results);
```
akarnokd: You can also check `subscriber.isUnsubscribed()` here as well or drop the `if` on line 128 depending on how eagerly you want to handle the small removeChangeListener/unsubscription window.

PR #1713, opened 2015-11-02
Added Realm.isEmpty()

dalinaum:

### DIFF
--- a/realm/realm-jni/src/io_realm_internal_Group.cpp
+++ b/realm/realm-jni/src/io_realm_internal_Group.cpp
@@ -289,3 +289,20 @@ JNIEXPORT jboolean JNICALL Java_io_realm_internal_Group_nativeEquals(
     } CATCH_STD()
     return false;
 }
+
+JNIEXPORT jboolean JNICALL Java_io_realm_internal_Group_nativeIsEmpty(
+    JNIEnv*, jobject, jlong nativeGroupPtr)
+{
+    Group* grp = G(nativeGroupPtr);
+    const string table_prefix(TABLE_PREFIX);
+    const size_t table_prefix_length = table_prefix.length();
+
+    for (size_t i = 0; i < grp->size(); ++i) {
+        ConstTableRef table = grp->get_table(i);
+        const string table_name = table->get_name();
+        if (table_name.compare(0, table_prefix_length, table_prefix) == 0 && !table->is_empty()) {
+            return false;
+        }
+    }
+    return true;
+}
--- a/realm/realm-jni/src/io_realm_internal_Util.cpp
+++ b/realm/realm-jni/src/io_realm_internal_Util.cpp
@@ -29,7 +29,9 @@ using std::string;
 
 // used by logging
 int trace_level = 0;
-const char *log_tag = "REALM";
+const char* log_tag = "REALM";
+
+const char* const TABLE_PREFIX = "class_";
 
 JNIEXPORT jint JNICALL JNI_OnLoad(JavaVM* vm, void*)
 {
@@ -71,6 +73,12 @@ JNIEXPORT void JNICALL Java_io_realm_internal_Util_nativeSetDebugLevel(JNIEnv*, 
 JNIEXPORT jlong JNICALL Java_io_realm_internal_Util_nativeGetMemUsage(JNIEnv*, jclass)
 {
     return GetMemUsage();
+}
+
+JNIEXPORT jstring JNICALL Java_io_realm_internal_Util_nativeGetTablePrefix(
+    JNIEnv* env, jclass)
+{
+    return to_jstring(env, string(TABLE_PREFIX));
 }
 
 // -------------------------- Testcases for exception handling
--- a/realm/realm-jni/src/util.hpp
+++ b/realm/realm-jni/src/util.hpp
@@ -136,7 +136,7 @@ jclass GetClass(JNIEnv* env, const char* classStr);
 
 // Debug trace
 extern int trace_level;
-extern const char *log_tag;
+extern const char* log_tag;
 
 #if TRACE
   #if defined(ANDROID)
@@ -583,5 +583,7 @@ inline jobject NewFloat(JNIEnv* env, float value)
 {
     return env->NewObject(java_lang_float, java_lang_float_init, value);
 }
+
+extern const char* const TABLE_PREFIX;
 
 #endif // REALM_JAVA_UTIL_HPP
--- a/realm/realm-library/src/androidTest/java/io/realm/RealmTest.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/RealmTest.java
@@ -2278,4 +2278,33 @@ public class RealmTest extends AndroidTestCase {
             throw exception.get(0);
         }
     }
+
+    public void testIsEmpty() {
+      RealmConfiguration realmConfig = TestHelper.createConfiguration(getContext(), "empty_test.realm");
+      Realm.deleteRealm(realmConfig);
+      Realm emptyRealm = Realm.getInstance(realmConfig);
+
+      assertTrue(emptyRealm.isEmpty());
+
+      emptyRealm.beginTransaction();
+      PrimaryKeyAsLong obj = new PrimaryKeyAsLong();
+      obj.setId(1);
+      obj.setName("Foo");
+      emptyRealm.copyToRealm(obj);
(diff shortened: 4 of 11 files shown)


realm/realm-jni/src/io_realm_internal_Group.cpp
```diff
@@ -289,3 +289,23 @@ JNIEXPORT jboolean JNICALL Java_io_realm_internal_Group_nativeEquals(
     } CATCH_STD()
     return false;
 }
+
+JNIEXPORT jboolean JNICALL Java_io_realm_internal_Group_nativeIsEmpty(
+    JNIEnv*, jobject, jlong nativeGroupPtr)
+{
+    Group* grp = G(nativeGroupPtr);
+    const std::string table_prefix = "class_";
```
beeender: Maybe this one should be in a common header? Like `util.hpp`?

realm/realm-library/src/main/java/io/realm/internal/Group.java
```diff
@@ -226,6 +226,15 @@ public void writeToFile(File file, byte[] key) throws IOException {
         return nativeWriteToMem(nativePtr);
     }
 
+    /*
+     * Check if the Group contains any objects.
+     *
+     * @return {@code true} if empty, @{code false} otherwise.
+     */
+    public boolean isEmptyTables() {
```
beeender: It seems `Group.size()` and `Group.isEmpty()` are useless now? Shall we just refactor them instead of creating a new method?

realm/realm-jni/src/io_realm_internal_Group.cpp
cmelchior: Currently Table.java also has a constant with this: `Table.TABLE_PREFIX`. Having it as an accessible constant from C++ would probably be nice, so we only have it defined 1 place.

realm/realm-library/src/androidTest/java/io/realm/RealmTest.java
```diff
@@ -2278,4 +2278,34 @@ public void run() {
...
+      PrimaryKeyAsLong obj = new PrimaryKeyAsLong();
+      obj.setId(1);
+      obj.setName("Foo");
+      emptyRealm.copyToRealm(obj);
+      assertFalse(emptyRealm.isEmpty());
+      emptyRealm.cancelTransaction();
+
+      assertTrue(emptyRealm.isEmpty());
+
+      emptyRealm.beginTransaction();
```
cmelchior: This section looks like a copy of the above. Just remove it?

realm/realm-library/src/androidTest/java/io/realm/RealmTest.java
```diff
@@ -2278,4 +2278,34 @@ public void run() {
...
+      obj = new PrimaryKeyAsLong();
+      obj.setId(1);
+      obj.setName("Foo");
+      emptyRealm.copyToRealm(obj);
+      emptyRealm.commitTransaction();
+
+      assertFalse(emptyRealm.isEmpty());
+
+      emptyRealm.close();
+      Realm.deleteRealm(realmConfig);
```
cmelchior: No need to delete the file. We don't do this in other unit tests. Closing the Realm is enough.

realm/realm-library/src/androidTest/java/io/realm/RealmTest.java
dalinaum: I think we need this section to check `assertFalse(emptyRealm.isEmpty())` on line 2306.

realm/realm-library/src/main/java/io/realm/internal/Group.java
```diff
@@ -226,6 +226,15 @@ public void writeToFile(File file, byte[] key) throws IOException {
         return nativeWriteToMem(nativePtr);
     }
 
+    /*
+     * Check if the Group contains any objects.
```
cmelchior: Also add that this only checked for `class_` tables, e.g. this would return true even if the `pk` table contained information.

realm/realm-library/src/androidTest/java/io/realm/RealmTest.java
dalinaum: I see. I should delete this.

realm/realm-library/src/main/java/io/realm/internal/Group.java
dalinaum: I don't know whether Group.size() and Group.isEmpty() are useless. Is it useless? Should I mark them as deprecated or just remove them?

cmelchior: I think there is a distinction between an empty group (no tables) and a group with tables but no objects. I would keep `Group.isEmpty()` and `Group.size()` for now as we might need them for other things.

I am not a big fan of this method name as it seems a bit ambiguous to me. Maybe `isRealmObjectTablesEmpty` instead?

beeender: :+1:  ^^

realm/realm-jni/src/io_realm_internal_Group.cpp
kneth: Yes, it is declared in `io.realm.internal.Table` but apparently `javah` does not generate macros for string constants.

realm/realm-jni/src/io_realm_internal_Group.cpp
```diff
@@ -289,3 +289,23 @@ JNIEXPORT jboolean JNICALL Java_io_realm_internal_Group_nativeEquals(
...
+    const size_t table_prefix_length = table_prefix.length();
+
+    for (size_t i = 0; i < grp->size(); ++i) {
+        ConstTableRef table = grp->get_table(i);
+        const std::string table_name = table->get_name();
```
kneth: No need for `std::`.

realm/realm-jni/src/io_realm_internal_Group.cpp
```diff
@@ -289,3 +289,23 @@ JNIEXPORT jboolean JNICALL Java_io_realm_internal_Group_nativeEquals(
...
+        if (table_name.size() < table_prefix_length || table_name.compare(0, table_prefix_length, table_prefix) != 0) {
```
emanuelez: How about something like:

```c++
if (table_name.compare(0, table_prefix_length, table_prefix) == 0 && !table.is_empty()) {
   return false;
}
```

realm/realm-jni/src/io_realm_internal_Group.cpp
dalinaum: I tried the following code:

dalinaum:
Please review this again @realm/java 

PS: @cmelchior I wonder whether I understand your proposal https://github.com/realm/realm-java/pull/1716#issue-114590642 correctly.

realm/realm-jni/src/util.hpp
```diff
@@ -584,4 +584,6 @@ inline jobject NewFloat(JNIEnv* env, float value)
     return env->NewObject(java_lang_float, java_lang_float_init, value);
 }
 
+extern const char * const TABLE_PREFIX;
```
beeender: -> `extern const char*`, remove the space.

realm/realm-jni/src/io_realm_internal_Util.cpp
```diff
@@ -31,6 +31,8 @@ using std::string;
 int trace_level = 0;
 const char *log_tag = "REALM";
 
+const char * const TABLE_PREFIX = "class_";
```
beeender: -> `extern const char*`, remove the space.

beeender:
Except the formatting issue, :+1:

realm/realm-library/src/androidTest/java/io/realm/RealmTest.java
cmelchior: Ah yes, sorry.

realm/realm-library/src/main/java/io/realm/internal/Group.java
```diff
@@ -226,6 +226,17 @@ public void writeToFile(File file, byte[] key) throws IOException {
         return nativeWriteToMem(nativePtr);
     }
 
+    /*
+     * Check if the Group contains any objects. It only checked for "class_"
```
cmelchior: checked -> checks

cmelchior:
Looks good. Except for the comments :+1:

Issue #766, opened 2015-01-19
Realm.isEmpty() method


cmelchior:
Feature request from this thread: https://groups.google.com/forum/#!topic/realm-java/u2DUhFbgMAc

Add a ```realm.isEmpty()``` method that returns true if there is no objects in the Realm. It is just a utility method, but fits nicely with the Realm object store abstraction.

bmunkholm:
I'm not convince this will be useful to many. Most of the time you would know which models you want to check? We should only add very useful or needed features to keep the API lean. So unless people really +1 this feature, I would close it.

cmelchior:
I agree, but the issue came up during a help request. This way we can also track it, even if we decide against it.

kristofer:
+1 - My situation is that I'm backing up the realm into cloudkit and when I switch devices, I need to know if the realm that was created is empty and I should "restore" from the "last backup"

cauli:
+1. Use-case: I want to bundle a prepopulated database with the app but also maintain a updatable one when possible. I need to check if the updatable is empty. If true, I'll use the prepulated one.

PR #1716, opened 2015-11-02
Primary keys now compatible with primary keys from the Object Store.

### LINKED ISSUE #1703: realm file incompatible if model has PrimaryKey
from slack.

Realm Swift reports an error when opening the realm file generated on Android.

model definition in Java:

```
public class DataItem extends RealmObject {
    @PrimaryKey
    private int id;

    // seter and getter here
}
```

model definition in Swift:

```
class DataItem: Object {
    dynamic var id = 0

    override static func primaryKey() -> String? {
        return "id"
    }
}
```

reported error:

```
fatal error: 'try!' expression unexpectedly raised an error: Error Domain=io.realm Code=0 "Migration is required for object type 'DataItem' due to the following errors:
- Property 'id' has been made a primary key." UserInfo={NSLocalizedDescription=Migration is required for object type 'DataItem' due to the following errors:
- Property 'id' has been made a primary key.}: file /Library/Caches/com.apple.xbs/Sources/swiftlang_PONDEROSA/swiftlang_PONDEROSA-700.1.101.6/src/swift/stdlib/public/core/ErrorType.swift, line 50
```

cmelchior:
Fixes #1703 

@realm/java 

This PR doesn't try to unify where `Table.TABLE_PREFIX` is set (see also #766) . Would adding a constant to e.g `Utils.hpp` be an idea. Then we could load statically when the Table class is loaded the first time through some `getTablePrefix()` method?


zaki50:
retest this please

examples/introExample/src/main/java/io/realm/examples/intro/model/AnnotationTypes.java
```diff
@@ -0,0 +1,69 @@
+/*
+ * Copyright 2014 Realm Inc.
```
kneth: "2014" -> "2015"

kneth:
:+1:

realm/realm-jni/src/io_realm_internal_table.cpp
```diff
@@ -1588,7 +1588,7 @@ JNIEXPORT jlong JNICALL Java_io_realm_internal_Table_nativeSetPrimaryKey(
     try {
         Table* table = TBL(nativeTablePtr);
         Table* pk_table = TBL(nativePrivateKeyTablePtr);
-        const char* table_name = table->get_name().data();
+        const char* table_name = table->get_name().substr(6).data(); // Remove "class_" prefix
```
zaki50: It's no longer table name.
Should be "model_name"?

cmelchior: I believe we still uses Cores terminology in the internal API's, e.g. Table/Column/Rows.

dalinaum:
:+1:

changelog.txt
```diff
@@ -8,7 +9,7 @@
  * Updated ProGuard configuration. See documentation https://realm.io/docs/java/latest/#proguard for more details.
  * Updated Kotlin example to use 1.0.0-beta.
  * Fixed warnings reported by `lint -Xlint:all` (#1644).
- * Fixed a bug where simultaneous opening and closing a Realm on different threads might result in a NullPointerException (#1646).
+ * Fixed a bug where simultaneous opening and closing a Realm on diffesrent threads might result in a NullPointerException (#1646).
```
jpsim: you introduced a typo here

PR #1707, opened 2015-10-31
Increase XX:MaxPermSize to avoid PermGen space problem.

dalinaum:
I encountered the following error.

```
Execution failed for task ':assembleExamples.'.
> Failed to notify build listener.
    > PermGen space
```

So I increased the default MaxPermSize to 512m.

Please review this, @realm/java.

PR #1718, opened 2015-11-02
Install SNAPSHOTs to mavenLocal() instead of ANDROID_HOME

cmelchior:
This PR install the java library + gradle plugin in mavenLocal() instead of ANDROID_HOME.
For some reason `uploadArchives` doesn't work for me, while `install` does. Which is a bit odd since @emanuelez said the opposite was true for him.

I changed the caching strategy to never cache snapshots for examples which makes it easier to update them during development.  

I also added an extra task for deploying both plugin + library `installRealmJava`.

@realm/java 
@emanuelez


emanuelez:
At a first quick glance if this works you can remove the maven-publish plugin.

cmelchior:
Isn't the `maven-publish` plugin responsible for handling the `publishing` block in the `realm` folder?

emanuelez:
You are absolutely right.
I would really love to deploy to a folder in the repo. That would make sure that a git clean will really reset the environment

cmelchior:
There already is a `cleanLocalMavenRepos` task that does exactly that?

emanuelez:
Yes, but that's a horrible hack IMO

cmelchior:
Could you use a `post-checkout` git hook and call that method? I doesn't look like there is a clean hook?

emanuelez:
Adding more horrible hacks? :)

cmelchior:
You're right :) In any case that is something that is not done in the current solution, so shouldn't really impact this PR?

realm/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -4,3 +4,4 @@ distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
 distributionUrl=https\://services.gradle.org/distributions/gradle-2.7-bin.zip
+set JAVA_OPTS="-Xmx512m"
```
zaki50: In #1707 we use gradle.properties.

Should this vm args also go to that file?

cmelchior: I must confess I am not entirely sure about the difference (if any) between the two files, but if it doesn't make difference, then yes, it should probably be moved to the `gradle.properties` file instead

zaki50: This file only affects gradle wraper(gradlew).

`gradle.properties` affects any gradle execution which is not limited to gradle wrapper.

And I realize that `set FOO` is not a syntax for properties file.
I suspect that this line has no effect at all.

cmelchior: I'll remove it again then. I was trying to fix a Java Heap space bug when compiling

nhachicha: have you tried the following inside `gradle.properties`
```groovy
org.gradle.jvmargs=-Xms256m -Xmx1024m
```
more here https://docs.gradle.org/current/userguide/build_environment.html

build.gradle
```diff
@@ -75,17 +75,24 @@ task connectedCheck(type:GradleBuild) {
 }
 
 task installGradlePlugin(type:GradleBuild) {
-    description = 'Install the Realm Gradle plugin to the Maven Repo in ANDROID_HOME'
+    description = 'Install the Realm Gradle plugin to mavenLocal()'
```
nhachicha: to-> into

build.gradle
```diff
@@ -75,17 +75,24 @@ task connectedCheck(type:GradleBuild) {
...
     group = 'Install'
     buildFile = file('gradle-plugin/build.gradle')
-    tasks = ['uploadArchives']
+    tasks = ['install']
 }
 
 task installRealm(type:GradleBuild) {
     group = 'Install'
-    description = 'Install the artifacts of Realm project to the Maven Repo in ANDROID_HOME'
+    description = 'Install the artifacts of Realm project to mavenLocal()'
```
nhachicha: to-> into

build.gradle
```diff
@@ -75,17 +75,24 @@ task connectedCheck(type:GradleBuild) {
...
     buildFile = file('realm/build.gradle')
-    tasks = ['uploadArchives']
+    tasks = ['install']
+}
+
+task installRealmJava(type:Task) {
+    dependsOn installGradlePlugin
+    dependsOn installRealm
+    group = 'Install'
+    description = 'Install the Realm library and Gradle plugin to mavenLocal()'
```
nhachicha: to->into

realm/gradle/wrapper/gradle-wrapper.properties
cmelchior: I just got another Java Heap space error. I'll try to add the above to `gradle.properties`

cmelchior:
Updated with comments.

zaki50:
:+1:

cmelchior:
Bump @realm/java

kneth:
:+1:

PR #1093, opened 2015-05-06
Moved RealmBaseAdapter to io.realm.android

cmelchior:
Don't merge this until https://github.com/realm/realm-java/pull/1090 are ready to be merged as well.

With RealmCursor soon done #1090 and RecyclerView.Adapter #972 at some point, we will start to have quite a few classes for integration with existing Android framework classes.

In order to not clutter the global io.realm package with support classes for Android, this PR moves them to ```io.realm.android```. This will give a better separation as ```io.realm``` now only contains pure framework classes.

Note it is a breaking change as people will have to fix their imports.

@emanuelez @kneth @bmunkholm 

Note: Remember to update findbugs-filter.xml before merging

PR #1725, opened 2015-11-04
fix #1675 RealmChangeListener are strong reference now. 

### LINKED ISSUE #1675: Be consistent with change listeners and weak referencing
According to the documentation:

*"Since 0.80.3, the RealmChangeListener is managed as a weak reference internally to avoid potential leaks. So DO NOT use anonymous RealmChangeListener when calling addListener. Instead, always maintain the reference to your listeners until you don’t need them anymore."*

This is not the case for the new API introduced in 0.84 (result.addChangeListener). addChangeListener for realmObject and realmresults are strong references:

*"Remember to unregister any listeners when exiting an Activity or Fragment to avoid memory leaks."*

I would personally prefer not to use weak references as that:
   * Creates warnings in the ide (Field can be local variable)
   * Disables the use of lambda expressions once these are available

In my opinion, it is best to return a Closeable (or destroyable) object which encapsulates all required context information to deallocate the resource. E.g.

```
List<Destroyable> destroyables = new List<>();
destroyables.add(realm.addChangeListener(...))
```
Then I can do

```
for (d in destroyables) {
   d.destroy();
}
```
then I can loop over all of them to destroy them. In either case, it would be nice if both

nhachicha:
**Breaking change:**
`realm.addChangeListener` hold a string reference 
```java
realm.addChangeListener(listener); //keep a strong reference to listener
realm.removeChangeListener(listener); // to remove a non-anonymous listener
realm.removeAllChangeListeners(); // to remove all listeners (including anonymous one)

realm. addChangeListenerAsWeakReference(listener); // add a WeakReference listener (this is package-private) for internal use only (RealmBaseAdapter)
```
- [x] update Javadoc as `RealmChangeListener` is used for both async & general notification, explain the triggering difference

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,3 +1,6 @@
+0.86.0
+* BREAKING CHANGE: `Realm.addChangeListener`, `RealmObject.addChangeListener` and `RealmResults.addChangeListener` hold a strong reference to the listener, you should unregister the listener to avoid memory leaks.
+
 0.85.0
  * BREAKING CHANGE: Removed RealmEncryptionNotSupportedException since the encryption implementation changed in Realm's underlying storage engine. Encryption is now supported on all devices.
  * BREAKING CHANGE: Realm.executeTransaction() now directly throws any RuntimeException instead of wrapping it in a RealmException (#1682).
--- a/realm/realm-library/src/androidTest/java/io/realm/NotificationsTest.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/NotificationsTest.java
@@ -16,6 +16,7 @@
 package io.realm;
 
 import android.os.Handler;
+import android.os.HandlerThread;
 import android.os.Looper;
 import android.test.AndroidTestCase;
 import android.util.Log;
@@ -473,7 +474,7 @@ public class NotificationsTest extends AndroidTestCase {
                 counter.incrementAndGet();
             }
         };
-        realm.addChangeListener(listener);
+        realm.addChangeListenerAsWeakReference(listener);
 
         // There is no guaranteed way to release the WeakReference,
         // just clear it.
@@ -605,6 +606,87 @@ public class NotificationsTest extends AndroidTestCase {
 
         realm.close();
         RealmLog.remove(logger);
+    }
+
+    public void testHandlerThreadShouldReceiveNotification() throws ExecutionException, InterruptedException {
+        final AssertionFailedError[] assertionFailedErrors = new AssertionFailedError[1];
+        final CountDownLatch backgroundThreadReady = new CountDownLatch(1);
+        final CountDownLatch numberOfInvocation = new CountDownLatch(1);
+
+        HandlerThread handlerThread = new HandlerThread("handlerThread");
+        handlerThread.start();
+        Handler handler = new Handler(handlerThread.getLooper());
+        handler.post(new Runnable() {
+            @Override
+            public void run() {
+                try {
+                    assertEquals("handlerThread", Thread.currentThread().getName());
+                } catch (AssertionFailedError e) {
+                    assertionFailedErrors[0] = e;
+                }
+                final Realm backgroundRealm = Realm.getInstance(getContext());
+                backgroundRealm.addChangeListener(new RealmChangeListener() {
+                    @Override
+                    public void onChange() {
+                        backgroundRealm.close();
+                        numberOfInvocation.countDown();
+                    }
+                });
+                backgroundThreadReady.countDown();
+            }
+        });
+        awaitOrThrow(backgroundThreadReady);
+        // At this point the background thread started & registered the listener
+
+        Realm realm = Realm.getInstance(getContext());
+        realm.beginTransaction();
+        realm.createObject(AllTypes.class);
+        realm.commitTransaction();
+
+        awaitOrThrow(numberOfInvocation);
+        realm.close();
+        handlerThread.quit();
+        if (assertionFailedErrors[0] != null) {
+            throw assertionFailedErrors[0];
+        }
+    }
+
+    public void testNonLooperThreadShouldNotifyLooperThreadAboutCommit() throws Throwable {
+        final CountDownLatch mainThreadReady = new CountDownLatch(1);
+        final CountDownLatch numberOfInvocation = new CountDownLatch(1);
+        Thread thread = new Thread() {
+            @Override
+            public void run() {
+                TestHelper.awaitOrFail(mainThreadReady);
+                Realm realm = Realm.getInstance(getContext());
+                realm.beginTransaction();
+                realm.createObject(AllTypes.class);
+                realm.commitTransaction();
+                realm.close();
+            }
+        };
+        thread.start();
+
+        HandlerThread mainThread = new HandlerThread("mainThread");
+        mainThread.start();
+        Handler handler = new Handler(mainThread.getLooper());
+        handler.post(new Runnable() {
+            @Override
+            public void run() {
+                final Realm mainRealm = Realm.getInstance(getContext());
+                mainRealm.addChangeListener(new RealmChangeListener() {
+                    @Override
+                    public void onChange() {
+                        mainRealm.close();
+                        numberOfInvocation.countDown();
+                    }
+                });
+                mainThreadReady.countDown();
+            }
+        });
+
+        awaitOrThrow(numberOfInvocation);
+        mainThread.quit();
     }
 
     private void awaitOrThrow(CountDownLatch latch) {
--- a/realm/realm-library/src/main/java/io/realm/BaseRealm.java
+++ b/realm/realm-library/src/main/java/io/realm/BaseRealm.java
@@ -30,8 +30,10 @@ import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.concurrent.CopyOnWriteArraySet;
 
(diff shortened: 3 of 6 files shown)


nhachicha:
@realm/java  fix #1725

realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -70,7 +72,10 @@
     // thread pool for all async operations (Query & transaction)
     static final RealmThreadPoolExecutor asyncQueryExecutor = RealmThreadPoolExecutor.getInstance();
 
-    protected final List<WeakReference<RealmChangeListener>> changeListeners =
+    protected final Set<RealmChangeListener> changeListeners =
```
cmelchior: This will make it undefined in which order the listeners are called, but prevent people registering the same listener twice. Would it make more sense to store them in a list instead, then the callback sequence is a bit more predictable (nice for debugging) at the expense of a little less security.

realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -70,7 +72,10 @@
...
+            new CopyOnWriteArraySet<RealmChangeListener>();
+
+    protected final List<WeakReference<RealmChangeListener>> weakChangeListeners =
```
cmelchior: I would add some small comments above both changeListener/weakChangeListener to explain why it is needed.

realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -207,14 +206,20 @@ protected void removeHandler() {
     }
 
     protected void sendNotifications() {
-        Iterator<WeakReference<RealmChangeListener>> iterator = changeListeners.iterator();
+        // notify strong reference listener
+        for (Iterator<RealmChangeListener> iterator = changeListeners.iterator(); iterator.hasNext(); ) {
```
cmelchior: Haven't seen this syntax used before :)
Would:
```
Iterator<RealmChangeListener> iterator = changeListeners.iterator();
while (iterator.hasNext(); ) {
```
be more idiomatic?

realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -148,14 +153,18 @@ public boolean isInTransaction() {
      */
     public void addChangeListener(RealmChangeListener listener) {
         checkIfValid();
-        for (WeakReference<RealmChangeListener> ref : changeListeners) {
+        changeListeners.add(listener);
+    }
+
+    void addChangeListenerAsWeakReference(RealmChangeListener listener) {
```
cmelchior: Add a small javadoc saying it is for internal use/UI components.

cmelchior:
What about `removeAllListeners` should it remove all WeakListeners as well?

realm/realm-library/src/main/java/io/realm/BaseRealm.java
nhachicha: User shouldn't worry about the order. 
I'll revert to `CopyOnWriteArrayList` though since `CopyOnWriteArraySet` is backed by `CopyOnWriteArrayList` anyway, then we can call `addIfAbsent` to avoid iterating all elements in `addChangeListener`

nhachicha:
@realm/java ready for review, I updated the Javadoc for `RealmChangeListener` 
@cmelchior `removeAllListeners` is public API, it should only impact other public API (i.e remove strong reference)

cmelchior:
Good point

realm/realm-library/src/androidTest/java/io/realm/NotificationsTest.java
```diff
@@ -607,6 +608,95 @@ public void run() {
...
+                        numberOfInvocation.countDown();
+                    }
+                });
+                mainThreadReady.countDown();
+            }
+        });
+
+        awaitOrThrow(numberOfInvocation);
+        mainThread.quit();
+        if (backgroundErrors[0] != null) {
```
cmelchior: Not very likely to happen, but you are not waiting for the background thread to close it's background Realm which might result in other unit tests not being able to delete the file.

realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -152,14 +161,25 @@ public boolean isInTransaction() {
...
+    }
+
+    /**
+     * For internal use only.
+     * Sometimes we don't know when to unregister listeners (ex: {@link RealmBaseAdapter}), using
+     * a WeakReference the listener.
+     *
+     * @param listener the change listener.
+     */
+    void addChangeListenerAsWeakReference(RealmChangeListener listener) {
```
cmelchior: This is a problem as we have a PR for moving all android components to `io.realm.android`. Some alternatives:

1) Make this public...eeew
2) Keep everything in `io.realm`. Also eew, but less so
3) Some other hack, like moving it to `io.realm.internal` and let `BaseRealm` and UI components talk to some API there. If it is considered internal it might make sense? Although having the changelisteners in BaseRealm.java is also nice.

realm/realm-library/src/main/java/io/realm/RealmChangeListener.java
```diff
@@ -17,7 +17,13 @@
 package io.realm;
 
 /**
- * Using RealmChangeListener, it is possible to be notified when a Realm instance has been updated.
+ * RealmChangeListener could be registered with a {@link Realm}, {@link RealmResults} or {@link RealmObject}
```
cmelchior: could -> can

cmelchior:
Missing changelog update, otherwise it looks fine except for a potential conflict with a current PR.

realm/realm-library/src/main/java/io/realm/BaseRealm.java
nhachicha: You're already making `getRealm` public in https://github.com/realm/realm-java/pull/1093 
this should work + `addChangeListenerAsWeakReference` is not Android specific.
It's just a behaviour we don't want to expose as public API.

realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -152,14 +161,25 @@ public boolean isInTransaction() {
      */
     public void addChangeListener(RealmChangeListener listener) {
```
beeender: We need to highlight somewhere to say that user has to call `removeChangeListener` to avoid memory leak.

realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -146,19 +155,35 @@ public boolean isInTransaction() {
      *     <li>On every call to {@link io.realm.Realm#refresh()}</li>
      * </ul>
      *
+     * Listeners are stored as strong reference, you need to remove the add listeners using {@link #removeChangeListener(RealmChangeListener)}
```
beeender: add listeners -> added.

beeender:
Except the one minor issue in the comment, :+1:

realm/realm-library/src/androidTest/java/io/realm/NotificationsTest.java
```diff
@@ -607,6 +608,87 @@ public void run() {
...
+                }
+                final Realm backgroundRealm = Realm.getInstance(getContext());
+                backgroundRealm.addChangeListener(new RealmChangeListener() {
+                    @Override
+                    public void onChange() {
+                        backgroundRealm.close();
+                        numberOfInvocation.countDown();
+                    }
+                });
+                backgroundThreadReady.countDown();
```
cmelchior: Nothing is awaiting for this?

realm/realm-library/src/main/java/io/realm/BaseRealm.java
cmelchior: Good point. We should probably add that to "best practices" on the website as well.

realm/realm-library/src/main/java/io/realm/RealmChangeListener.java
```diff
@@ -17,7 +17,13 @@
 package io.realm;
 
 /**
- * Using RealmChangeListener, it is possible to be notified when a Realm instance has been updated.
+ * RealmChangeListener can be registered with a {@link Realm}, {@link RealmResults} or {@link RealmObject}
+ * to receive a notification about updates.
+ * <p>
+ * When registered against a {@code Realm} you'll get notified when a Realm instance has been updated.
+ * When registered against a {@code RealmResults} or {@code RealmObject} you'll get notified only when the
```
cmelchior: That is not entirely true (yet). Perhaps we should add a caveat at the end about that?

cmelchior:
Apart from some minor doc issues: :+1:

dalinaum:
Beside some comments LGTM :+1:

changelog.txt
```diff
@@ -1,3 +1,6 @@
+0.86.0
+* BREAKING CHANGE: `Realm.addChangeListener`, `RealmObject.addChangeListener` and `RealmResults.addChangeListener` hold a strong reference to the listener, you should unregister the listener to avoid memory leak.
```
emanuelez: leak -> leaks

realm/realm-library/src/main/java/io/realm/BaseRealm.java
emanuelez: as *a* strong reference

realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -146,19 +155,35 @@ public boolean isInTransaction() {
...
+     * or {@link #removeAllChangeListeners()} which removes all listeners including the one added via anonymous classes.
```
emanuelez: the one**s**

PR #1729, opened 2015-11-05
throws IllegalArgumentException if isNull/isNotNull is called for unsupported field.

zaki50:
Currently, `isNull()`/`isNotNull()` for linked field whose last element is a link is not supported.
We should throw `IllegalArgumentException` instead of `RealmError` if that happened.

refs #1693

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -2,6 +2,7 @@
  * Added Realm.isEmpty().
  * Fixed a bug making it impossible to convert a field to become required during a migration (#1695).
  * Fixed a bug making it impossible to read Realms created using primary keys and created by iOS (#1703).
+ * RealmQuery.isNull() and RealmQuery.isNotNull() now throw IllegalArgumentException instead of RealmError if the fieldname is a linked field and the last element is a link (#1693).
 
 0.84.1
  * Updated Realm Core to 0.94.4
--- a/realm/realm-jni/src/io_realm_internal_TableQuery.cpp
+++ b/realm/realm-jni/src/io_realm_internal_TableQuery.cpp
@@ -1461,7 +1461,7 @@ JNIEXPORT void JNICALL Java_io_realm_internal_TableQuery_nativeIsNull(
         } else {
             switch (col_type) {
                 case type_Link:
-                    pQuery->and_query(src_table_ref->column<Link>(S(column_idx)).is_null());
+                    ThrowException(env, IllegalArgument, "isNull() by nested query for link field is not supported.");
                     break;
                 case type_LinkList:
                     // Cannot get here. Exception will be thrown in TBL_AND_COL_NULLABLE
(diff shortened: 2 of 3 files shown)


zaki50:

changelog.txt
```diff
@@ -2,6 +2,7 @@
  * Fixed a bug making it impossible to read Realms created by iOS if a class contained primary keys (#1703).
  * Added Realm.isEmpty().
  * Fixed bug making it impossible to convert a field to being required during a migration.
+ * RealmQuery.isNull() and RealmQuery.isNotNull() now throw IllegalArgumentException instead of RealmError if the query target is a nested link field.
```
cmelchior: if the fieldname is a linked field.

realm/realm-jni/src/io_realm_internal_TableQuery.cpp
```diff
@@ -1461,7 +1461,7 @@ JNIEXPORT void JNICALL Java_io_realm_internal_TableQuery_nativeIsNull(
         } else {
             switch (col_type) {
                 case type_Link:
-                    pQuery->and_query(src_table_ref->column<Link>(S(column_idx)).is_null());
+                    ThrowException(env, IllegalArgument, "isNull() by nested query for link field is not supported.");
```
cmelchior: Instead of doing it like this. Wouldn't it be better to catch the exception from core in CATCH_STD (if it is an explicit exception type) and transform it there. That way it will also be caught when when core actually adds support for it.
See util.cpp (ConvertException)

changelog.txt
zaki50: Does `linked field` mean both a) fieldname contains `.` and b) the last element is a field of RealmObject?

cmelchior: linked field means any field name with `.` , no matter what the field contains. At least in my mind.

realm/realm-jni/src/io_realm_internal_TableQuery.cpp
zaki50: @cmelchior Yes, but the thrown exception is `std::runtime_error` and we should not convert it `IllegalArgumentException`.
Should we ask to change the exception something like `not_implemented_exception`?

cmelchior: I wouldn't bother, because we couldn't be sure what C++ not_implemented should map to in Java, it could be a number of things. Just keep it as it is now then.

changelog.txt
zaki50: @cmelchior How about `if the fieldname is a linked field and the last element is a link.`

cmelchior: :+1:

zaki50:
retest this please

cmelchior:
:+1:

beeender:
:+1:

PR #1732, opened 2015-11-05
Add counter for Realm instance for all threads.

### LINKED ISSUE #1728: Trying to catch a dead lock
Hi there,

I'm trying to locate a dead-lock which is happening somewhere inside Realm.
By logging I found that I get an infinite ANR in ```onCreate()``` of the MainActivity on the line ```Realm.getDefaultInstance()```. Another place where it hangs is a worker thread on the line ```Realm.close()```.

Here is how the worker thread body looks like (without the domain details):
```java
Realm realm = null;
try {
    realm = Realm.getDefaultInstance();
    RealmResults<A> results = realm.where(A.class).findAll();
    realm.beginTransaction();
    for (int i = results.size() - 1; i >= 0; i--) {
        final A item = results.get(i);
        String value = item.getSomeStringValue();
        callSomeMethod(value); // this method throws an exception not related to realm when i == 1
    }
    realm.commitTransaction();
} finally {
    if (realm != null) {
        realm.close(); // hangs on this line based on the logs
    }
}
```

I was able to narrow down where exactly Realm.close() hangs. It is the line ```synchronized (BaseRealm.class)```:
```java
protected void lastLocalInstanceClosed() {
        // validatedRealmFiles must not modified while other thread is executing createAndValidate()
        synchronized (BaseRealm.class) {
            validatedRealmFiles.remove(configuration.getPath());
        }
        realmsCache.get().remove(configuration);
    }
}
```

Here is the sequence of actions accordingly to logs:

1. Application.onCreate() called and starts the worker thread
2. Worker thread is going through the items
3. MainActivity.onCreate() called and it hangs on Realm.getDefaultInstance()
4. Exception thrown in the worker thread, and it hangs on realm.close()
5. Infinite ANR in MainActivity

It's very hard to locate the problem because it's not 100% reproducible and time-dependent. If I remove transaction, or there is no exception then the dead-lock doesn't happen (could be just due to timing). Also, I have another running worker thread in background which uses realm. But it successfully finishes and doesn't hang on ```realm.close()```. 

I tried to reproduce it in a small project, but it didn't work out.

I hope this description could lead you to some ideas what could go wrong, or what I can check to locate the exact problem.

beeender:
To close #1728
The only usage of the Realm instance counter for now is to clean up
validatedRealmFiles.
We do need to refactor the whole counter/cache part.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -3,7 +3,8 @@
  * Fixed a bug making it impossible to convert a field to become required during a migration (#1695).
  * Fixed a bug making it impossible to read Realms created using primary keys and created by iOS (#1703).
  * Fixed some memory leaks when an Exception is thrown (#1730).
- * Fixed a memory leak when using relationships. (#1285)
+ * Fixed a memory leak when using relationships (#1285).
+ * Fixed a bug causing cached column indices to be cleared too soon (#1732).
  * RealmQuery.isNull() and RealmQuery.isNotNull() now throw IllegalArgumentException instead of RealmError if the fieldname is a linked field and the last element is a link (#1693).
 
 0.84.1
--- a/realm/realm-library/src/androidTest/java/io/realm/RealmTest.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/RealmTest.java
@@ -2281,6 +2281,64 @@ public class RealmTest extends AndroidTestCase {
         }
     }
 
+    // Bug reported https://github.com/realm/realm-java/issues/1728.
+    // Root cause is validatedRealmFiles will be cleaned when any thread's Realm ref counter reach 0.
+    public void testOpenRealmWhileTransactionInAnotherThread() throws Exception {
+        final CountDownLatch realmOpenedInBgLatch = new CountDownLatch(1);
+        final CountDownLatch realmClosedInFgLatch = new CountDownLatch(1);
+        final CountDownLatch transBeganInBgLatch = new CountDownLatch(1);
+        final CountDownLatch fgFinishedLatch = new CountDownLatch(1);
+        final CountDownLatch bgFinishedLatch = new CountDownLatch(1);
+        final List<Exception> exception = new ArrayList<Exception>();
+
+        // Step 1: testRealm is opened already.
+
+        Thread thread = new Thread(new Runnable() {
+            @Override
+            public void run() {
+                // Step 2: Open realm in background thread.
+                Realm realm = Realm.getInstance(testConfig);
+                realmOpenedInBgLatch.countDown();
+                try {
+                    realmClosedInFgLatch.await();
+                } catch (InterruptedException e) {
+                    exception.add(e);
+                    realm.close();
+                    return;
+                }
+
+                // Step 4: Start transaction in background
+                realm.beginTransaction();
+                transBeganInBgLatch.countDown();
+                try {
+                    fgFinishedLatch.await();
+                } catch (InterruptedException e) {
+                    exception.add(e);
+                }
+                // Step 6: Cancel Transaction and close realm in background
+                realm.cancelTransaction();
+                realm.close();
+                bgFinishedLatch.countDown();
+            }
+        });
+        thread.start();
+
+        realmOpenedInBgLatch.await();
+        // Step 3: Close all realm instances in foreground thread.
+        testRealm.close();
+        realmClosedInFgLatch.countDown();
+        transBeganInBgLatch.await();
+
+        // Step 5: Get a new Realm instance in foreground
+        testRealm = Realm.getInstance(testConfig);
+        fgFinishedLatch.countDown();
+        bgFinishedLatch.await();
+
+        if (!exception.isEmpty()) {
+            throw exception.get(0);
+        }
+    }
+
     public void testIsEmpty() {
         RealmConfiguration realmConfig = TestHelper.createConfiguration(getContext(), "empty_test.realm");
         Realm.deleteRealm(realmConfig);
--- a/realm/realm-library/src/main/java/io/realm/Realm.java
+++ b/realm/realm-library/src/main/java/io/realm/Realm.java
@@ -135,6 +135,10 @@ public final class Realm extends BaseRealm {
     private final Map<Class<? extends RealmObject>, Table> classToTable =
             new HashMap<Class<? extends RealmObject>, Table>();
 
+    // Reference count on currently opened Realm instances.
+    // We need to know if all typed Realm instance of all threads are closed in order to clean up validatedRealmFiles.
+    private static final Map<String, Integer> typedRealmFileReferenceCounter = new HashMap<String, Integer>();
+
     private static RealmConfiguration defaultConfiguration;
     protected ColumnIndices columnIndices;
     /**
@@ -285,6 +289,8 @@ public final class Realm extends BaseRealm {
 
             // Increment global reference counter
             realm.acquireFileReference(configuration);
+            // Increment Realm file reference counter
+            acquireRealmFileReference(configuration);
 
             // Check versions of Realm
             long currentVersion = realm.getVersion();
@@ -1127,7 +1133,10 @@ public final class Realm extends BaseRealm {
     protected void lastLocalInstanceClosed() {
         // validatedRealmFiles must not modified while other thread is executing createAndValidate()
         synchronized (BaseRealm.class) {
-            validatedRealmFiles.remove(configuration.getPath());
+            // All Realm instances with this file have been closed. Decrease the counter.
+            if (releaseRealmFileReference(configuration) == 0) {
+                validatedRealmFiles.remove(configuration.getPath());
+            }
         }
         realmsCache.get().remove(configuration);
     }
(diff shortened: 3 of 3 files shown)


beeender:
@realm/java

changelog.txt
```diff
@@ -2,6 +2,7 @@
  * Added Realm.isEmpty().
  * Fixed a bug making it impossible to convert a field to become required during a migration (#1695).
  * Fixed a bug making it impossible to read Realms created using primary keys and created by iOS (#1703).
+ * Fixed a bug cached column indexes were cleared improperly.
```
cmelchior: "Fixed a bug causing cached field indexes to be cleared to soon"
Also add a reference to the issue or PR

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -1269,6 +1279,37 @@ public static Object getDefaultModule() {
     }
 
     /**
+     * Acquire a reference to the given Realm file.
+     * This function call should be called in synchronized block on BaseRealm.class
```
cmelchior: in *a*

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -1269,6 +1279,37 @@ public static Object getDefaultModule() {
...
+     */
+    private static void acquireRealmFileReference(RealmConfiguration configuration) {
+        String path = configuration.getPath();
+        Integer refCount = realmFileReferenceCounter.get(path);
+        if (refCount == null) {
+            refCount = 0;
+        }
+        realmFileReferenceCounter.put(path, refCount + 1);
```
cmelchior: you should still call through to the parent method in BaseRealm, because it's counters also need to be accurate.

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -1269,6 +1279,37 @@ public static Object getDefaultModule() {
...
+        String path = configuration.getPath();
+        Integer refCount = realmFileReferenceCounter.get(path);
+        if (refCount == null || refCount == 0) {
+            throw new IllegalStateException("Trying to release a Realm file that is already closed");
+        }
+        if (refCount == 1) {
+            realmFileReferenceCounter.remove(path);
+        } else {
+            realmFileReferenceCounter.put(path, refCount - 1);
+        }
```

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -137,6 +137,10 @@
     private final Map<Class<? extends RealmObject>, Table> classToTable =
             new HashMap<Class<? extends RealmObject>, Table>();
 
+    // Reference count on currently opened Realm instances.
+    // We need to know if all Realm instance of all threads are closed in order to clean up validatedRealmFiles.
+    private static final Map<String, Integer> realmFileReferenceCounter = new HashMap<String, Integer>();
```
cmelchior: Perhaps use the term "typedRealm" and "dynamicRealm" instead of "realm" and "dynamicRealm"?.....otherwise it gets really hard to think about the variables IMO

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -291,6 +295,8 @@ private static Realm createAndValidate(RealmConfiguration configuration, Boolean
 
             // Increment global reference counter
             realm.acquireFileReference(configuration);
```
cmelchior: The code would be cleaner if this was moved to acquireRealmFileReference i think?

beeender:
@realm/java  Please help to review again

realm/realm-library/src/main/java/io/realm/Realm.java
beeender: agreed with cm. keep it like this.

cmelchior:
:+1:

realm/realm-library/src/androidTest/java/io/realm/RealmTest.java
```diff
@@ -2279,6 +2279,64 @@ public void run() {
...
+        final CountDownLatch fgFinishedLatch = new CountDownLatch(1);
+        final CountDownLatch bgFinishedLatch = new CountDownLatch(1);
+        final List<Exception> exception = new ArrayList<Exception>();
+
+        // Step 1: testRealm is opened already.
+
+        Thread thread = new Thread(new Runnable() {
+            @Override
+            public void run() {
+                // Step 2: Open realm is background thread.
```
zaki50: is -> in?

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -1154,7 +1160,11 @@ private void checkHasPrimaryKey(Class<? extends RealmObject> clazz) {
     protected void lastLocalInstanceClosed() {
         // validatedRealmFiles must not modified while other thread is executing createAndValidate()
         synchronized (BaseRealm.class) {
-            validatedRealmFiles.remove(configuration.getPath());
+            // All Realm instances with this file have been closed. Decrease the counter.
+            releaseRealmFileReference(configuration);
+            if (typedRealmFileReferenceCounter.get(configuration.getPath()) == null) {
```
zaki50: How about changing `releaseRealmFileReference()` to return counter number.

If so, we can write these two lines into `if (releaseRealmFileReference(configuration) == 0) {`
and reduce one hash map search.

changelog.txt
```diff
@@ -3,6 +3,7 @@
  * Fixed a bug making it impossible to convert a field to become required during a migration (#1695).
  * Fixed a bug making it impossible to read Realms created using primary keys and created by iOS (#1703).
  * RealmQuery.isNull() and RealmQuery.isNotNull() now throw IllegalArgumentException instead of RealmError if the fieldname is a linked field and the last element is a link (#1693).
+ * Fixed a bug causing cached column indexes to be cleared too soon. (#1732)
```
kneth: "indexes" -> "indices"

beeender:
@realm/java Please review again.

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -1269,6 +1278,45 @@ public static Object getDefaultModule() {
...
+     * Releases a reference to the Realm file. If reference count reaches 0 any cached configurations
+     * will be removed.
+     * This function call should be called in a synchronized block on BaseRealm.class
+     *
+     * @return the reference count after releasing.
+     */
+    private static int releaseRealmFileReference(RealmConfiguration configuration) {
+        String path = configuration.getPath();
+        Integer refCount = typedRealmFileReferenceCounter.get(path);
+        if (refCount == null || refCount <= 0) {
```
cmelchior: Is this scenario possible? It is called from `lastLocalInstanceClosed()` which has it's own counter check in BaseRealm?

emanuelez:
How about using PhantomReference for everything, Realm instances included?

beeender:
@emanuelez That is my next plan :grinning:

cmelchior:
That still require thread safe destructors, right?

beeender:
i think so... it would still be hard to get rid of the `Context`

emanuelez:
No need for thread safe destructors if we keep Context alive, which I think we should.

beeender:
#1752 would be the final solution for this. But we can merge this one as a hotfix before the "DynamicRealm` merged.

cmelchior:
:+1:

changelog.txt
```diff
@@ -3,6 +3,7 @@
  * Fixed a bug making it impossible to convert a field to become required during a migration (#1695).
  * Fixed a bug making it impossible to read Realms created using primary keys and created by iOS (#1703).
  * RealmQuery.isNull() and RealmQuery.isNotNull() now throw IllegalArgumentException instead of RealmError if the fieldname is a linked field and the last element is a link (#1693).
+ * Fixed a bug causing cached column indices to be cleared too soon. (#1732)
```
zaki50: It's better to reorder this

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -1269,6 +1278,45 @@ public static Object getDefaultModule() {
     }
 
     /**
+     * Acquire a reference to the given Realm file.
+     * This function call should be called in a synchronized block on BaseRealm.class
```
zaki50: BaseRealm.class should be {@code BaseRealm.class}
and . should be at the end of line.

changelog.txt
beeender: done

realm/realm-library/src/main/java/io/realm/Realm.java
beeender: done

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -1243,6 +1252,45 @@ public static Object getDefaultModule() {
     }
 
     /**
+     * Acquire a reference to the given Realm file.
+     * This function call should be called in a synchronized block on {@code BaseRealm.class}
```
zaki50: `.` is missing at the end of line.

zaki50:
:+1: apart from missing `.` in Javadoc comment.

PR #1760, opened 2015-11-12
Cleanup after adding RealmSchema

cmelchior:
This cleans our internal API's up a bit after introducing RealmSchema. All methods relevant to the Schema have been moved into RealmSchema and RealmObjectSchema


cmelchior:
Last PR to complete migrations @realm/java

realm/realm-library/src/main/java/io/realm/RealmObjectSchema.java
```diff
@@ -521,6 +521,99 @@ private void checkEmpty(String str) {
...
+     * @return list of column indices.
+     */
+    // TODO: consider another caching strategy so linked classes are included in the cache.
+    long[] getColumnIndices(String fieldDescription, RealmFieldType... validColumnTypes) {
+        if (fieldDescription == null || fieldDescription.equals("")) {
+            throw new IllegalArgumentException("Non-empty fieldname must be provided");
+        }
+        if (fieldDescription.endsWith(".")) {
+            throw new IllegalArgumentException("Illegal field name. It cannot end with a '.'");
+        }
```
kneth: What if `fieldDescription` begins with `.`?

cmelchior: It work work anyway. I added this because we had a unit test catching it, but you are right, we should also check for a starting `.`

kneth:
:+1: with a few comments ;-)

realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -511,33 +500,6 @@ static synchronized void releaseFileReference(RealmConfiguration configuration)
         globalRealmFileReferenceCounter.put(canonicalPath, refCount - 1);
     }
 
-    // Public because of migrations
-    @Deprecated
-    public Table getTable(String className) {
```
emanuelez: Yeah!!! :)

realm/realm-library/src/main/java/io/realm/RealmObjectSchema.java
```diff
@@ -407,7 +411,7 @@ public boolean hasPrimaryKey() {
      */
     public Set<String> getFieldNames() {
         int columnCount = (int) table.getColumnCount();
-        Set<String> columnNames = new TreeSet<String>();
+        Set<String> columnNames = new LinkedHashSet<>(columnCount);
```
emanuelez: I don't think we ever decided to use Java7's diamond operator

cmelchior: Ups, will fix

emanuelez: Why the switch to `LinkedHashSet`? What operations are we trying to optimize for?

realm/realm-library/src/main/java/io/realm/RealmObjectSchema.java
```diff
@@ -521,6 +521,99 @@ private void checkEmpty(String str) {
...
     /**
+     * Returns the column indices for the given field name. If a linked field is defined, the column index for
+     * each field is returned
+     *
+     * @param fieldDescription fieldName or link path to a field name.
+     * @param validColumnTypes Legal field type for the last field a
+     * @return list of column indices.
+     */
+    // TODO: consider another caching strategy so linked classes are included in the cache.
+    long[] getColumnIndices(String fieldDescription, RealmFieldType... validColumnTypes) {
```
emanuelez: Varargs... it's probably better to change the signature to (String s, RealmFieldType first, RealmFieldType... others)

http://jtechies.blogspot.dk/2012/07/item-42-use-varargs-judiciously.html

realm/realm-library/src/main/java/io/realm/RealmObjectSchema.java
cmelchior: Insertion order, so people get the order of fields in core when they iterator over the set.

realm/realm-library/src/main/java/io/realm/RealmObjectSchema.java
```diff
@@ -521,6 +521,99 @@ private void checkEmpty(String str) {
...
+            RealmFieldType tableColumnType = table.getColumnType(getFieldIndex(fieldDescription));
+            if (checkColumnType && !isValidType(tableColumnType, validColumnTypes)) {
+                throw new IllegalArgumentException(String.format("Field '%s': type mismatch. Was %s, expected %s.",
+                        fieldDescription, tableColumnType, Arrays.toString(validColumnTypes)));
+            }
+            return new long[] {getFieldIndex(fieldDescription)};
+        }
+    }
+
+    private boolean containsDot(String s) {
```
emanuelez: This is how the JVM implements contains: http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/6-b14/java/lang/String.java#String.contains%28java.lang.CharSequence%29

So here you're just replicating what the JVM does. We should just use `.contains(".")`

realm/realm-library/src/main/java/io/realm/RealmObjectSchema.java
```diff
@@ -521,6 +521,99 @@ private void checkEmpty(String str) {
...
+        }
+        return false;
+    }
+
+    /**
+     * Returns the underlying table backing this schema.
+     *
+     * @return the underlying {@link Table}.
+     */
+    Table getTable() {
```
emanuelez: Or we could make `table` package protected and avoid the getter. It's already final so we don't have to worry about setting the value to something else.

realm/realm-library/src/main/java/io/realm/RealmObjectSchema.java
```diff
@@ -539,4 +632,72 @@ public FieldMetaData(RealmFieldType realmType, boolean defaultNullable) {
             this.defaultNullable = defaultNullable;
         }
     }
+
+    static class DynamicColumnMap implements Map<String, Long> {
```
emanuelez: Why implement `Map` when we disable most of its functionality?

realm/realm-library/src/main/java/io/realm/RealmSchema.java
```diff
@@ -156,4 +177,64 @@ private void checkHasTable(String className, String errorMsg) {
...
     }
+
+    ColumnInfo getColumnInfo(Class<? extends RealmObject> clazz) {
+        final ColumnInfo columnInfo = columnIndices.getColumnInfo(clazz);
+        if (columnInfo == null) {
+            throw new IllegalStateException("No validated schema information found for " + realm.configuration.getSchemaMediator().getTableName(clazz));
+        }
+        return columnInfo;
+    }
+    Table getTable(String className) {
```
emanuelez: add an empty line

realm/realm-library/src/main/java/io/realm/RealmSchema.java
```diff
@@ -156,4 +177,64 @@ private void checkHasTable(String className, String errorMsg) {
             throw new IllegalArgumentException(errorMsg);
         }
     }
+
+    ColumnInfo getColumnInfo(Class<? extends RealmObject> clazz) {
+        final ColumnInfo columnInfo = columnIndices.getColumnInfo(clazz);
```
emanuelez: What if `clazz` is null?

realm/realm-library/src/main/java/io/realm/RealmObjectSchema.java
emanuelez: :+1:

realm/realm-library/src/main/java/io/realm/RealmObjectSchema.java
cmelchior: Because this needs to be compatible with the ColumnIndices class. I looked into a rewrite, but it turned out to be a bigger thing so would rather do that outside the scope of this PR

realm/realm-library/src/main/java/io/realm/RealmSchema.java
cmelchior: Then we crash, I just moved the method to this class. AFAIK the callers of this method guarantee that it is non-null

realm/realm-library/src/main/java/io/realm/RealmObjectSchema.java
emanuelez: I see :+1:

cmelchior:
@ez Updated with comments

Saketme:
What's the reason behind not using Java's diamond operator?

cmelchior:
That it doesn't work if we wish to support Java 6 on the server side. Java 6 has been end-of-life for a while now, but we have not made that decision yet, wether or not to drop support for it.

emanuelez:
:+1:

PR #1771, opened 2015-11-13
Improving .so loading by using the ReLinker library

kneth:
We have seen a number of reports on problems with loading `.so` files. The origins can be that the `.so` loader in Android is known to be unreliable and some end-users do not use Google's Play Store (often the case in China).

TODO:
 - [x] Proposed entry in `changelog.txt`: "Improving .so loading by using the ReLinker library (https://github.com/KeepSafe/ReLinker)"

@realm/java


emanuelez:
We should add ReLinker as a dependency in Gradle and not just copy pasting the file

cmelchior:
Agreeded. We can can shadow it the same way we do JavaWriter to avoid versioning conflicts from other libraries

realm/realm-library/src/main/java/io/realm/RealmConfiguration.java
```diff
@@ -244,6 +246,8 @@ public Builder(Context context) {
                 throw new IllegalArgumentException("A non-null Context must be provided");
             }
             initializeBuilder(context.getFilesDir());
+            // FIXME: Using ReLinker to load .so file adds a string Android dependency
```
cmelchior: Remove FIXME?

kneth:
@cmelchior You had the opposite opinion yesterday but I'll be happy to change it back.

realm/realm-library/src/main/java/io/realm/RealmConfiguration.java
```diff
@@ -244,6 +246,8 @@ public Builder(Context context) {
...
+            ReLinker.loadLibrary(context, "realm-jni");
```
cmelchior: What happens if the context is null?

cmelchior:
Alternative suggestion to the changelog: "Improving .so loading by using ReLinker (link)"

cmelchior:
@kneth Yes, I believe I might accidentally have voiced that opinion, sorry. Shadowing does the same thing, except we don't have to add the file to the code base.

realm/realm-library/src/main/java/io/realm/RealmConfiguration.java
kneth: You will get a NPE in line 248. But we can check for it.

realm/realm-library/build.gradle
```diff
@@ -17,11 +17,18 @@ android {
...
+repositories {
+    jcenter()
+    maven { url "https://jitpack.io" }
+}
 
 dependencies {
     compile 'com.intellij:annotations:+@jar'
     compile project(':realm-annotations')
 
+    // https://github.com/KeepSafe/ReLinker
```
bmunkholm: remove comment. Doesn't add any info.

realm/realm-library/build.gradle
```diff
@@ -17,11 +17,18 @@ android {
         project.archivesBaseName = "realm-android"
     }
 }
+repositories {
```
emanuelez: I would add this section to the `build.gradle` file in the `realm` folder

realm/realm-library/src/main/java/io/realm/RealmConfiguration.java
```diff
@@ -244,6 +246,7 @@ public Builder(Context context) {
                 throw new IllegalArgumentException("A non-null Context must be provided");
             }
             initializeBuilder(context.getFilesDir());
+            ReLinker.loadLibrary(context, "realm-jni");
```
beeender: Doesn't this cause problem with our JNI testing? If the case is only about Table, the `Builder` will not be called. And also, I think put this in a static block would be a good idea since it only needs to be ran once.

beeender:
The failed monkeyTest is because of the `ReLinker` class is not packed in the final apk file. Gradle experts @emanuelez @zaki50 ?

zaki50:
I think we need add `ReLInker` to https://github.com/realm/realm-java/blob/master/gradle-plugin/src/main/groovy/io/realm/gradle/Realm.groovy#L14 or repackage `ReLinker` into realm-library

emanuelez:
@zaki50 That should not be needed as it should be handled as a transitive dependency, just like JavaWriter

zaki50:
@emanuelez make sense. We need to add `"https://jitpack.io"` as maven repository?

emanuelez:
Ohhh... yes! That's probably it!

realm/realm-library/src/main/java/io/realm/RealmConfiguration.java
kneth: The problem with the static block is that it doesn't have access to a `Context`.

zaki50:
But should we avoid implicitly to add an maven repository?

zaki50:
RealmConfiguration.Builder has another constructor `Builder(File folder)` and it does not load the native library.

kneth:
Bump @realm/java

realm/realm-library/src/main/java/io/realm/internal/RealmCore.java
```diff
@@ -121,6 +125,14 @@ public static synchronized void loadLibrary() {
         Version.coreLibVersionCompatible(true);
     }
 
+    public static synchronized void loadLibrary(Context context) {
```
cmelchior: Missing Javadoc for this method and `loadLibrary()` so the difference is clear.

cmelchior:
Except for the missing Javadoc :+1:

realm/realm-library/src/main/java/io/realm/internal/RealmCore.java
```diff
@@ -121,6 +129,20 @@ public static synchronized void loadLibrary() {
         Version.coreLibVersionCompatible(true);
     }
 
+    /**
+     * Loads the .so file. Typically, the .so file is installed and can be found by System.loadLibrary() but
+     * can be damaged or missing. This happens for the Android installer, especially when apps are installed
+     * through other means than the official Play store. In this case, the .so file can be found in the .apk.
+     * In other to access the .apk, a Context much be provided.
```
nhachicha: much -> must

realm/realm-library/src/main/java/io/realm/internal/RealmCore.java
```diff
@@ -95,6 +99,10 @@ public static void print(String caption, AbstractCursor<?> cursor) {
 
     // Although loadLibrary is synchronized internally from AOSP 4.3, for the compatibility reason,
     // KEEP synchronized here for the old devices!
+
+    /**
+     * Loads the .so file. This method is useful for static blocks as it does not rely on access to a Context.
```
nhachicha: you can specify Android Context as opposed to Realm Context :) *same below*

nhachicha:
LGTM beside small comment :+1:

cmelchior:
Some comments about the Relinker library from the Internet: 

1) Upgrades
If the app requires the workaround and extracts the .so file and then you get an update of the APK which would also require the workaround, it won’t extract the library from the new APK

2) Files are placed in `lib` which is the system folder. You risk a name-conflict, better put it somewhere else.

3) `getWorkaroundLibFile()` should add the current version as a suffix

4) loadLibrary() might do File I/O on the UI thread. Is that acceptable?

kneth:
We are trying to solve `.so` loading issues for a small minority of users, and @cmelchior points out that ReLinker might not be the answer.

cmelchior:
No no, Relinker is still a good idea, it probably just needs a few modifications

dalinaum:
:+1: This branch has conflicts. It should be fixed.

kneth:
@realm/java I have resolved the conflicts but I am still hesitating to include it for the reasons @cmelchior points out.

kneth:
@beeender Did you have a chance to test ReLinker with devices not using Play store?

beeender:
Not yet.
Some Chinese stores support binary delta updates, but i assume that is the binary delta of APK file but not the content of dex and libs. Which means finally it will go through the `PackageManager` of the system which is the same process of the installation from g store.

I will test it.

Does it conflict if user has Relinker enabled already for other libs?

kneth:
I'll take a look at if multiple includes of ReLinker will work.

kneth:
@emanuelez I have added ReLinker as a dependency - please review.

kneth:
@realm/java We need to get this in - or find another solution. Please review and test.

cmelchior:
From what I can tell, ReLinker 1.1 solves must of the issue above:
- Files are now saved in the apps private file area
- Any old files are deleted before unpacking a new one 

We might still risk doing file I/O on the main thread, but since the chance of that is so little compared how we had to rewrite our entire API to be async, I think that is acceptable.

gradle-plugin/src/main/groovy/io/realm/gradle/Realm.groovy
```diff
@@ -9,10 +9,10 @@ class Realm implements Plugin<Project> {
...
     void apply(Project project) {
         project.plugins.apply(AndroidAptPlugin)
-        project.repositories.add(project.getRepositories().jcenter());
-        project.dependencies.add("compile", "io.realm:realm-android-library:${Version.VERSION}");
-        project.dependencies.add("compile", "io.realm:realm-annotations:${Version.VERSION}");
-        project.dependencies.add("apt", "io.realm:realm-annotations-processor:${Version.VERSION}");
-        project.dependencies.add("apt", "io.realm:realm-annotations:${Version.VERSION}");
+        project.repositories.add(project.getRepositories().jcenter())
+        project.repositories.add(project.repositories.maven { url "https://jitpack.io" })
+        project.dependencies.add("compile", "io.realm:realm-android:${Version.VERSION}")
```
cmelchior: This should be realm-android-library

realm/realm-library/build.gradle
```diff
@@ -19,12 +19,18 @@ android {
...
+}
+
 dependencies {
     compile 'com.intellij:annotations:+@jar'
     compile project(':realm-annotations')
 
     androidTestApt project(':realm-annotations-processor')
     androidTestApt project(':realm-annotations')
+
+    compile 'com.github.KeepSafe:ReLinker:1.1'
```
cmelchior: This should be moved up under the other `compile` dependencies`

cmelchior:
Apart from a few comments I think this looks good :+1:, note that some of @nhachicha comments are still  open

changelog.txt
```diff
@@ -8,6 +8,7 @@
  * Added new Dynamic API using DynamicRealm and DynamicRealmObject.
  * Added Realm.getSchema() and DynamicRealm.getSchema().
  * Realm.createOrUpdateObjectFromJson() now works correctly if the RealmObject class contains a primary key (#1777).
+ * Improved .so loading by using ReLinker (https://github.com/KeepSafe/ReLinker).
```
dalinaum: Should we move this to 0.87.0?

kneth: Good idea!

kneth:
@beeender Can you test this PR with devices not using the Play Store?

diegomontoya:
I just ran into the dreaded NoClassDefNotFound for io/realm/internal/SharedGroup bug on my dev machine building debug apk on a Huawei 4.0.4 device. 

Somehow, the package installer/adb install process was incomplete causing an inconsistent state. 

I manually added Relinker package to trigger if I catch the NoClassDefNotFound in Realm.getInstance(config).  It did not work. The new builds/installs was still failing with same exception. Repeated installs 5-6 times and still crashed. 

Strange thing is, the library .so appears to be loaded correctly with System.load() not triggering any errors. 

The only way to fix the problem is to unistall the apk from device. "Updating" it, after the "PackageManager bug", regardless of work-around does not work, at least for my instance.

beeender:
@diegomontoya When the problem happens, can you please check in the `/data/data/<package-name/lib` directory to see if the so file is the same with the newly installed version? If it is not, is it the previous version?

Can you please test if the Relinker solve this problem by calling Relinker to load the `librealm-jni.so` before calling any Realm functions

I really want to try this by myself but none of my phones can reproduce this issue...

diegomontoya:
@beeender The "NoClassDefNotFound" for io/realm/internal/SharedGroup for my case is not caused as result of the system unable to load the .so file aka System.loadLibrary() throwing errors.  That was very strange.

Unfortunately I have use builds/installed hundreds of times on the device and yesterday was the first time ever I could reproduce this type of bug on any device. Next time I will make sure to freeze the /data/data path for future inspection. Too bad I wasn't thinking clearly at the time. Opportunity wasted on catching the bigfoot.

>Strange thing is, the library .so appears to be loaded correctly with System.loadLibrary() not triggering any errors.

kneth:
@diegomontoya Do you suggest that corrupted `.so` is the result of a bug in the package manager? It's going to be hard to fix that :smile:

diegomontoya:
@kneth Yes, I am suggesting there are two bugs here, one of which relink can fix. The first bug is that the install completely bonks the .so copy so there are no .so file to be found. Relink can fix this. But the issue I encountered appears to be the .so loading was completed without errors but still triggering the cryptic io/ream/internal/SharedGroup not found error. The second bug may be hopeless since it may mean the .so is actually corrupted but loadable.

cmelchior:
Fixed the last bug in the unit tests, so should be ready now @realm/java

cmelchior:
:+1: from me, but I made the last change, so ... ;)

beeender:
retest this please

beeender:
:+1:

gradle-plugin/src/main/groovy/io/realm/gradle/Realm.groovy
```diff
@@ -9,10 +9,10 @@ class Realm implements Plugin<Project> {
...
-        project.repositories.add(project.getRepositories().jcenter());
-        project.dependencies.add("compile", "io.realm:realm-android-library:${Version.VERSION}");
-        project.dependencies.add("compile", "io.realm:realm-annotations:${Version.VERSION}");
-        project.dependencies.add("apt", "io.realm:realm-annotations-processor:${Version.VERSION}");
-        project.dependencies.add("apt", "io.realm:realm-annotations:${Version.VERSION}");
+        project.repositories.add(project.getRepositories().jcenter())
+        project.repositories.add(project.repositories.maven { url "https://jitpack.io" })
+        project.dependencies.add("compile", "io.realm:realm-android-library:${Version.VERSION}")
+        project.dependencies.add("compile", "io.realm:realm-annotations:${Version.VERSION}")
+        project.dependencies.add("apt", "io.realm:realm-annotations-processor:${Version.VERSION}")
```
emanuelez: Are you sure the annotations don't need to be added as `apt`?

realm/realm-library/build.gradle
```diff
@@ -24,9 +24,14 @@ android {
...
 }
 
+repositories {
+    maven { url "https://jitpack.io" }
+}
+
 dependencies {
     provided 'io.reactivex:rxjava:1.1.0'
     compile project(':realm-annotations')
+    compile 'com.github.KeepSafe:ReLinker:1.1'
```
emanuelez: Can we find a way to centralise the version of ReLinker? It's used both here and in `bintrayAar.gradle` so it's easy to forget one of the two

realm/realm-library/src/androidTest/java/io/realm/internal/PivotTest.java
```diff
@@ -23,61 +23,62 @@
 
 public class PivotTest extends TestCase {
 
-    Table t = new Table();
+    Table t;
```
emanuelez: Are the changes to this file related to the PR?

kneth: Yes, when we move the loading from a static block to a method, it is not possible to allocate a `Table` statically.

Issue #1594, opened 2015-10-13
API review & cleanup for 1.0


nhachicha:
The following things should be implemented:

- [x] JSON methods currently wrap all runtime exceptions in a RealmException. Consider removing this as it makes it hard to know exactly what failed -> Remove RealmException where possible. ( #2618 )

- [x] Cleanup sorted* methods. Remove all sort methods accepting 3 fields. Multiarg should be used instead. ( #2619  )

- [x] Remove `allObjects` and friends from Realm so all query options are in RealmQuery. ( #2620 )

- [x] RealmConfiguration `setModules()` rename to `modules()` ( #2621 )

- [X] All lists methods should automatically do `copyToRealm` or `copyToRealmOrUpdate` (mc Checked, all methods which could add/change object in the list, will call `copyToRealmIfNeeded`)

- [ ] RealmChangeListener should provide the changes object/realm/collection as well

- [x] realm.createObject(class, primaryKey) should be public ( #2622 )

- [x] All methods on RealmObject should be final. All public Realm classes should be final (#2675).

- [x] BaseRealm should be package protected (will most likely require moving some things to `io.realm`) (#2673)

- [x] HandlerController should be package protected. ( #2632 )

- [x] Constructor of `RealmAsyncTask` should be package protected (#2676).

-----------

This issue help track any suggested refactoring, cleanup of the current API to prepare for 1.0 release.
this is not a call for feature, but a take on the existing API to rename/refactor.

please share your suggestions, it would be helpful to include motivation & code snippet/prototype so people can have a quick idea to up/down vote.

cmelchior:
Some things on my list:

- Get rid of the difference between `copyToRealm` and `copyOrUpdateToRealm`. If your model class has primary keys just using `copyToRealm` doesn't make any sense. 

- Consider renaming `where` to `query` to get away from the SQL terminology.

astigsen:
> Consider renaming `where` to `query` to get away from the SQL terminology.

In Swift the method is called `filter`, so it might be an idea to standardize on that?

nhachicha:
As part of the 1.0 cleanup https://github.com/realm/realm-java-private/issues/21

cmelchior:
This should also include formalising our Javadoc style and do a cleanup based on that.

beeender:
* argument for `RealmChangeListener.onChange`
   * `RealmChangeListener.onChange(Realm)`
   * `RealmChangeListener.onChange(RealmObject)`
[...]

Issue #682, opened 2014-12-18
JSON API v2


cmelchior:
Design and implement a more complete solution for our JSON API.
Current design doc is here: 
https://github.com/cmelchior/realm-java/blob/29ab93bf7aeaa5cd048cbf468a228046496ecb13/json_api.md

Other features:
**Strict mode:** We should probably consider adding something like STRICT mode to our JSON parser which could throw an exception when encountering JSON properties it didn't recognise. At least for debugging it could save a lot of time.

**Json export** It should also be possible to convert any RealmObjects to JSON.

erichkleung:
@cmelchior Do you have a timeline for the custom constructors for RealmObjects (to use GSON)? Thanks!

cmelchior:
It is next on my list, right after we finish up merging outstanding PR's. So baring anything unexpected I will begin working on it today or tomorrow.

TepesLucian:
Hi! Are you guys planning to add a createOrReplaceFromJson? Right now we get a lot of duplicate objects just with the createFromJson and the populate method in RealmObjects is package protected.

cmelchior:
Yes, we definitely want to add that, but it require that we have primary keys working first https://github.com/realm/realm-java/pull/565, and they depend on standalone objects https://github.com/realm/realm-java/pull/728. Currently the plan is that the next release (0.77) will contain standalone objects, and I am pretty confident that we can have primary keys ready for 0.78.

TepesLucian:
Awesome :+1:

hareshsandeep:
@cmelchior  Is there any progress on storing dates with different formats ? for instance, I am trying to store date format EEE, dd MMM yyyy hh:mm:ss Z.

bmunkholm:
@hareshsandeep Could we please ask you to create a new issue with more details so we can better help? Thanks!

RustamG:
Are there any update on this? Is it hard to implement #1470 as one of the first steps?

cmelchior:
Hi @RustamG 
Sorry no, still update on this. For `@SerializedName` most of the existing JSON libraries can do this for you. However it is a feature we are looking into outside the contet of JSON as well. It is being tracked here: https://github.com/realm/realm-java/issues/451

TR4Android:
Ok, I'd be willing to work on this, but wanted to first discuss naming and similiar design choices. The following is on my checklist:

- [ ] Improve import API with annotations
- [ ] Add export API

[...]

PR #1780, opened 2015-11-16
copyObjectFromJson now works correctly with PrimaryKey

### LINKED ISSUE #1777: createOrUpdateObjectFromJson bugs with default value for primary key
example code:
```java
JSONObject object = new JSONObject("{\"id\":1,\"bar1\":{\"id\":0,\"value\":3},\"bar2\":{\"id\":1,\"value\":2}}");
        realm.beginTransaction();
        realm.createOrUpdateObjectFromJson(Foo.class, object);
        realm.commitTransaction();
```
```java
//class Foo
    @PrimaryKey
    private long id;

    private Bar bar1;
    private Bar bar2;

    public long getId() {
        return id;
    }

    public void setId(long id) {
        this.id = id;
    }

    public Bar getBar1() {
        return bar1;
    }

    public void setBar1(Bar bar1) {
        this.bar1 = bar1;
    }

    public Bar getBar2() {
        return bar2;
    }

    public void setBar2(Bar bar2) {
        this.bar2 = bar2;
    }

}
```
```java
//class Bar
public class Bar extends RealmObject {
    @PrimaryKey
    private int id;
    private int value;

    public int getValue() {
        return value;
    }

    public void setValue(int value) {
        this.value = value;
    }

    public int getId() {
        return id;
    }

    public void setId(int id) {
        this.id = id;
    }
}
```
The problem is that when it begins to create bar2, it finds that no bar with id=2 exists, so it will create a default bar, but that will conflict with bar1, since bar1 has id 0 and default bar has id 0 too.

cmelchior:
Fixes #1777 

Note, I also added a more sensible error message when using `copyObjectFromJson`. We need to rethink our JSON support anyway ( #1594 , #682 ), so this was the most non-intrusive change for now.

@realm/java


changelog.txt
```diff
@@ -2,7 +2,8 @@
  * Added Realm.isEmpty().
  * RealmQuery.isNull() and RealmQuery.isNotNull() now throw IllegalArgumentException instead of RealmError if the fieldname is a linked field and the last element is a link (#1693).
  * Setters in managed object for RealmObject and RealmList now throw IllegalArgumentException if the value contains an invalid (standalone, removed, closed, from different Realm) object (#1749).
- * Attempting to refresh a Realm while a transaction is in process will now throw an IllegalStateException (#1712) 
+ * Attempting to refresh a Realm while a transaction is in process will now throw an IllegalStateException (#1712)
+ * createOrUpdateObjectFromJson now works correctly if the RealmObject class contains a primary key (#1777)
```
kneth: `()` missing (indicates a method name)

kneth:
:+1:

realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmJsonTypeHelper.java
```diff
@@ -79,6 +79,11 @@ public void emitStreamTypeConversion(String setter, String fieldName, String fie
                         .emitStatement("obj.%s(JsonUtils.stringToDate(reader.nextString()))", setter)
                     .endControlFlow();
             }
+
+            @Override
+            public void emitGetObjectWithPrimaryKeyValue(String qualifiedRealmObjectClass, String fieldName, JavaWriter writer) throws IOException {
+                throw new IllegalArgumentException("Date not allowed as a primary key value");
```
beeender: Date is not allowed as a primary key value.

realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmJsonTypeHelper.java
```diff
@@ -105,9 +110,22 @@ public void emitStreamTypeConversion(String setter, String fieldName, String fie
                         .emitStatement("obj.%s(JsonUtils.stringToBytes(reader.nextString()))", setter)
                     .endControlFlow();
             }
+
+            @Override
+            public void emitGetObjectWithPrimaryKeyValue(String qualifiedRealmObjectClass, String fieldName, JavaWriter writer) throws IOException {
+                throw new IllegalArgumentException("byte[] not allowed as a primary key value");
```
beeender: byte[] is not allowed as a primary key value.    full stop at the end

realm/realm-annotations-processor/src/test/resources/io/realm/AllTypesRealmProxy.java
```diff
@@ -422,7 +422,15 @@ public static AllTypes createOrUpdateUsingJsonObject(Realm realm, JSONObject jso
...
+                if (json.isNull("columnString")) {
+                    obj = realm.createObject(AllTypes.class, null);
+                } else {
+                    obj = realm.createObject(AllTypes.class, json.getString("columnString"));
+                }
+            } else {
+                obj = realm.createObject(AllTypes.class);
+            }
         }
         if (json.has("columnString")) {
```
beeender: this if is not needed anymore, i think?

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -625,6 +625,11 @@ private static void initializeRealm(Realm realm) {
         if (clazz == null || inputStream == null) {
             return null;
         }
+        Table table = getTable(clazz);
+        if (table.hasPrimaryKey()) {
+            throw new UnsupportedOperationException("Cannot create RealmObjects with primary keys using Streams. " +
+                    "Use createOrUpdateObjectFromJson() instead.");
```
beeender: I am not sure how many people are using our JSONStream interface, but this would create a really bad user experience :( What about we use standalone object to set all values first?

cmelchior: Hmm, that might be an okay work-around

beeender: Or loop with the stream two rounds, find out the primary key in the first round, also document to suggest always put the primary key as the first one if user wants to use JSONStream.... silly solution...

realm/realm-annotations-processor/src/test/resources/io/realm/AllTypesRealmProxy.java
cmelchior: It is for objects without primary keys. I could probably be optimised a bit, but I would rather not do it  in this PR.

beeender: OK~

cmelchior:
Updated @realm/java

changelog.txt
```diff
@@ -1,3 +1,6 @@
+0.86.0
+ * Realm.createOrUpdateObjectFromJson() now works correctly if the RealmObject class contains a primary key (#1777)
```
kneth: Missing `.` at the end.

kneth:
Minor comment, otherwise :+1:

beeender:
:+1:

dalinaum:
:+1:

PR #1781, opened 2015-11-16
Fix the publishing to OJO and Bintray

emanuelez:


emanuelez:
Instamerge @realm/java

PR #1784, opened 2015-11-16
Add consumer ProGuard configuration

### LINKED ISSUE #1767: Add ProGuard to AAR distribution
When moving to shipping an AAR instead of a JAR we should also bundle the ProGuard configuration so people do not have to do that themselves:

http://stackoverflow.com/questions/30201420/how-to-include-a-proguard-configuration-in-my-android-library-aar

skyisle:
Fixes #1767


realm/realm-library/build.gradle
```diff
@@ -15,6 +15,7 @@ android {
         minSdkVersion 9
         targetSdkVersion 23
         project.archivesBaseName = "realm-android"
+        consumerProguardFiles 'consumer-proguard-rules.pro'
```
cmelchior: I would just call this `proguard-rules.pro`

cmelchior:
Hi @skyisle 
Thanks for adding this. Can you also add a changelog item about this. Something along the lines of `The Realm AAR now also contains the ProGuard configuration.`

cmelchior:
Looks good to me :+1: 

@realm/java

dalinaum:
:+1:

zaki50:
I also confirmed! :+1:

PR #1786, opened 2015-11-16
Adding ISO8601 date parsing support

### LINKED ISSUE #1213: Add JSON support for ISO8601
Is there a way to Realm.createObjectFromJson() with date attributes in format ISO8601? This is the current code that decodes date from JSON for the Android Version:

```
public static Date stringToDate(String date) {
        if (date == null || date.length() == 0) return null;
        Matcher matcher = jsonDate.matcher(date);
        if (matcher.matches()) {
            return new Date(Long.parseLong(matcher.group(1)));
        } else {
            return new Date(Long.parseLong(date));
       }
}
```

An strategy consist of use regex with the ISO 8601 format and perform the conversion. Is it possible?

donnfelker:
Fixes #1213 

Also Adding json util tests for existing code to ensure no regressions


realm/realm-library/src/androidTest/java/io/realm/internal/android/JsonUtilsTest.java
```diff
@@ -0,0 +1,90 @@
+package io.realm.internal.android;
```
emanuelez: Missing license header

donnfelker: Added

realm/realm-library/src/androidTest/java/io/realm/internal/android/JsonUtilsTest.java
```diff
@@ -0,0 +1,106 @@
...
+        Date date = cal.getTime();
+        cal.set(Calendar.MILLISECOND, 0);
+        Date dateZeroMillis = cal.getTime();
+        cal.set(Calendar.SECOND, 0);
+
+        Date d = ISO8601Utils.parse("2007-08-13T195123.789Z", new ParsePosition(0));
+
+        assertEquals(date, d);
+
+        d = JsonUtils.stringToDate("2007-08-13T195123Z");
```
cmelchior: Add a comment about what is being tested above each of these. Mentally parsing them is kinda hard :)

realm/realm-library/src/main/java/io/realm/internal/android/ISO8601Utils.java
```diff
@@ -0,0 +1,272 @@
...
+import java.text.ParsePosition;
+import java.util.*;
+
+/**
+ * Utilities methods for manipulating dates in iso8601 format. This is much much faster and GC friendly than using SimpleDateFormat so
+ * highly suitable if you (un)serialize lots of date objects.
+ *
+ * Supported parse format: [yyyy-MM-dd|yyyyMMdd][T(hh:mm[:ss[.sss]]|hhmm[ss[.sss]])]?[Z|[+-]hh[:]mm]]
+ *
+ * @see <a href="http://www.w3.org/TR/NOTE-datetime">this specification</a>
```
cmelchior: I would also add a link to the original source code

realm/realm-library/src/main/java/io/realm/internal/android/JsonUtils.java
```diff
@@ -41,15 +43,18 @@ public static Date stringToDate(String date) {
         if (matcher.find()) {
             String dateMatch = matcher.group(1);
             return new Date(Long.parseLong(dateMatch));
-        } else {
-            return new Date(Long.parseLong(date));
         }
+        try {
+            return ISO8601Utils.parse(date, new ParsePosition(0));
```
cmelchior: I could be tempted to say we could remove the `ParsePosition` class as it doesn't make sense in our context, but then it would be hard to keep our version of `ISO8601Utils` up to date. At least saving `ParsePosition` in a static variable will reduce objects allocated when parsing large amounts of Json.

realm/realm-library/src/main/java/io/realm/internal/android/JsonUtils.java
```diff
@@ -41,15 +43,18 @@ public static Date stringToDate(String date) {
...
+        } catch (ParseException ignored) {
+        }
+        return new Date(Long.parseLong(date));
     }
 
     /**
      * Converts a Json string to byte[]. String must be Base64 encoded.
      *
-     * @param str the base 64 encoded bytes.
+     * @param str the base 64 encoded bytevs.
```
cmelchior: bytevs -> bytes

realm/realm-library/src/main/java/io/realm/internal/android/JsonUtils.java
```diff
@@ -41,15 +43,18 @@ public static Date stringToDate(String date) {
...
             String dateMatch = matcher.group(1);
             return new Date(Long.parseLong(dateMatch));
-        } else {
-            return new Date(Long.parseLong(date));
         }
+        try {
+            return ISO8601Utils.parse(date, new ParsePosition(0));
+        } catch (ParseException ignored) {
+        }
+        return new Date(Long.parseLong(date));
```
cmelchior: I'm a bit worried about having this last. It means that if you actually have this format, you have to deal with a `ParseException` for each object, which is very expensive. Is there a lightweight way of detecting this scenario? A simple regexp would do I assume? `/-?\d*/`. Perhaps `string.contains("-")` or some other char that is always present (not sure there is any though judging from the unit tests).

cmelchior:
Missing changelog update as well

realm/realm-library/src/main/java/io/realm/internal/android/JsonUtils.java
```diff
@@ -37,13 +40,25 @@
...
+        }
+
+        // Check for millisecond based date
+        if(numericOnly.matcher(date).matches()) {
             return new Date(Long.parseLong(date));
         }
+
+        // Try for ISO8601 date
+        try {
+            return ISO8601Utils.parse(date, new ParsePosition(0));
```
cmelchior: `new ParsePosition(0)` should be a static variable. No need to allocate it each time.

realm/realm-library/src/androidTest/java/io/realm/internal/android/ISO8601UtilsTest.java
```diff
@@ -0,0 +1,198 @@
+/*
+ * Copyright 2015 FasterXML
+ * Copyright 2015 Realm Inc.
```
bmunkholm: Two copyrights? (same in other files). How does that work?

cmelchior: That is the standard way of dealing with changes to code that is Apache 2 licensed. Basically it just says that FasterXML did the original implementation and Realm added some changes to it.

donnfelker:
@cmelchior Please re-review when you get a chance. :)

realm/realm-library/src/androidTest/java/io/realm/internal/android/JsonUtilsTest.java
```diff
@@ -0,0 +1,112 @@
...
+        assertEquals(date, d);
+
+        // Short time without decimal second
+        d = JsonUtils.stringToDate("2007-08-13T195123Z");
+        assertEquals(dateZeroMillis, d);
+
+        // GMT+2 with decimal second
+        d = JsonUtils.stringToDate("2007-08-13T215123.789+02:00");
+        assertEquals(date, d);
+
```
cmelchior: One whitespace to many

realm/realm-library/src/main/java/io/realm/internal/android/JsonUtils.java
```diff
@@ -37,13 +41,26 @@
...
+        if(numericOnly.matcher(date).matches()) {
             return new Date(Long.parseLong(date));
         }
+
+        // Try for ISO8601 date
+        try {
+            parsePosition.setIndex(0); // reset the position each time
+            return ISO8601Utils.parse(date, parsePosition);
+        } catch (ParseException ignored) {
+            return null;
```
cmelchior: This looks like a change from the current behaviour where we crash if the date isn't in a format we support. The old behaviour seems more safe IMO. This implementation could cause objects to be imported successfully but with wrong dates. In the old implementation we end up throwing a ParseNumberException, so just throwing the parse exception here would probably be fine.

cmelchior:
Unit test failure is:
```
junit.framework.AssertionFailedError: DATE failed
at io.realm.DynamicRealmObjectTest.testUntypedSetIllegalImplicitConversionThrows(DynamicRealmObjectTest.java:612)
at android.test.AndroidTestRunner.runTest(AndroidTestRunner.java:191)
at android.test.AndroidTestRunner.runTest(AndroidTestRunner.java:176)
at android.test.InstrumentationTestRunner.onStart(InstrumentationTestRunner.java:555)
at android.app.Instrumentation$InstrumentationThread.run(Instrumentation.java:1837)
```

realm/realm-library/src/main/java/io/realm/internal/android/JsonUtils.java
donnfelker: @cmelchior The previous version threw a `NumberFormatException`. In this case, that could still happen and the docs still have that in the throws documentation. If we are to throw a `ParseException` then we will have to add the throws declaration to the method signature. That could cause a ripple of changes because `ParseException` is a typical `Exception` while `NumberFormatException` is a `RunTimeException`. Alternatively, we can catch the parse exception and throw a different `RuntimeException` (one of our own, etc). What would you like to do here? Thoughts?

beeender: Catch and throw a different `RuntimeException` +1

cmelchior: Ah didn't notice that `ParseException` was a checked exception. Yes, catching and throwing a runtime exception is a lot better.

I am not entirely sure what would be best. `RealmException` or just `RuntimeException`?  I am probably leaning towards just `RuntimeException` as we have been removing use of `RealmException` elsewhere, but either would be fine I think.

If we used `RealmException` we should probably convert the NumberFormatException to that as well though.

zilimojo:
When will this change be released?

cmelchior:
@zilimojo As part of 0.87.0, most likely some time next week.

cmelchior:
Looks good to me :+1: 

Ready for review @realm/java

changelog.txt
```diff
@@ -18,6 +18,7 @@
  * Updated Realm Core to 0.95.3
    - Fixed a bug where RealmQuery.average(String) returned a wrong value for a nullable Long/Integer/Short/Byte field (#1803).
    - Fixed a bug where RealmQuery.average(String) wrongly counted the null value for average calculation (#1854).
+ * Adding support for ISO8601 based dates for JSON import. If JSON dates are invalid a RealmException will be thrown. (#1213)
```
cmelchior: Before merging this should be moved to 0.87.0

kneth:
Are we required to add a section in `README.md` about FasterXML?

Besides that, it looks fine to me.

cmelchior:
As far as I can tell from https://tldrlegal.com/license/apache-license-2.0-(apache-2.0) we are compliant. The only point I wasn't sure about was the `NOTICE` file, but Jackson doesn't appear to have one, so it should be good.

kneth:
Yes, it was `NOTICE` I was concerned about. We can ask @timanglade and add it later.

timanglade:
If there is no `NOTICE` file in Jackson, then we don’t need to add anything else, I think.

PR #1790, opened 2015-11-17
RealmCache to decouple some logic from BaseRealm

### LINKED ISSUE #1735: Refactor Realm related caches
Now we have too many maps/ThreadLocals to cache Realm/DynamicRealm/RealmConfiguration/ColumnIndice/Ref counters/Handlers which make the code really hard to read and easy to create a threading problem.

We need a proper way to wrap those things and make it neat.

beeender:
* Add RealmCache class to cache Realm instances and related resources.
* Move cache related logic from BaseRealm, Realm, DynamicRealm to
  RealmCache.
* Move some test cases to RealmCacheTest.
* Fix test cases.

Close #1735

### DIFF
--- a/realm/realm-library/src/androidTest/java/io/realm/RealmCacheTest.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/RealmCacheTest.java
@@ -1,0 +1,282 @@
+/*
+ * Copyright 2015 Realm Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.realm;
+
+import android.test.AndroidTestCase;
+
+import java.io.File;
+import java.io.IOException;
+import java.util.concurrent.Callable;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+import io.realm.entities.AllTypes;
+import io.realm.entities.StringOnly;
+
+public class RealmCacheTest extends AndroidTestCase {
+
+    private RealmConfiguration defaultConfig;
+
+    @Override
+    protected void setUp() throws Exception {
+        super.setUp();
+        defaultConfig = TestHelper.createConfiguration(getContext());
+        Realm.deleteRealm(defaultConfig);
+    }
+
+    // Test that the closed Realm isn't kept in the Realm instance cache
+    public void testRealmCacheIsCleared() {
+        Realm typedRealm = Realm.getInstance(defaultConfig);
+        DynamicRealm dynamicRealm = DynamicRealm.getInstance(defaultConfig);
+
+        typedRealm.close(); // Still a dynamic instance open, but the typed Realm cache must still be cleared.
+        dynamicRealm.close();
+
+        Realm typedRealm1 = Realm.getInstance(defaultConfig);
+        try {
+            assertFalse(typedRealm == typedRealm1); // Must be different instance
+            // If cache isn't cleared this would crash because of a closed shared group.
+            assertEquals(0, typedRealm1.where(AllTypes.class).count());
+        } finally {
+            typedRealm1.close();
+        }
+    }
+
+    // Test that the closed DynamicRealms isn't kept in the DynamicRealm instance cache
+    public void testDynamicRealmCacheIsCleared() {
+        DynamicRealm dynamicRealm = DynamicRealm.getInstance(defaultConfig);
+        Realm typedRealm = Realm.getInstance(defaultConfig);
+
+        dynamicRealm.close(); // Still an instance open, but DynamicRealm cache must still be cleared.
+        typedRealm.close();
+
+        DynamicRealm dynamicRealm1 = DynamicRealm.getInstance(defaultConfig);
+        try {
+            // If cache isn't cleared this would crash because of a closed shared group.
+            assertFalse(dynamicRealm == dynamicRealm1);
+            assertEquals(0, dynamicRealm1.getVersion());
+        } finally {
+            dynamicRealm1.close();
+        }
+    }
+
+    public void testGetInstanceClearsCacheWhenFailed() {
+        String REALM_NAME = "invalid_cache.realm";
+        RealmConfiguration configA = TestHelper.createConfiguration(getContext(), REALM_NAME, TestHelper.getRandomKey(42));
+        RealmConfiguration configB = TestHelper.createConfiguration(getContext(), REALM_NAME, TestHelper.getRandomKey(43));
+
+        Realm.deleteRealm(configA);
+        Realm realm = Realm.getInstance(configA); // Create starting Realm with key1
+        realm.close();
+        try {
+            Realm.getInstance(configB); // Try to open with key 2
+        } catch (IllegalArgumentException ignored) {
+            // Delete Realm so key 2 works. This should work as a Realm shouldn't be cached
+            // if initialization failed.
+            assertTrue(Realm.deleteRealm(configA));
+            realm = Realm.getInstance(configB);
+            realm.close();
+        }
+    }
+
+    public void testRealmCache() {
+        Realm realm = Realm.getInstance(defaultConfig);
+        Realm newRealm = Realm.getInstance(defaultConfig);
+        try {
+            assertEquals(realm, newRealm);
+        } finally {
+            realm.close();
+            newRealm.close();
+        }
+    }
+
+    // We should not cache wrong configurations
+    public void testDontCacheWrongConfigurations() throws IOException {
+        Realm testRealm;
+        String REALM_NAME = "encrypted.realm";
+        TestHelper.copyRealmFromAssets(getContext(), REALM_NAME, REALM_NAME);
+        RealmMigration realmMigration = TestHelper.prepareMigrationToNullSupportStep();
+
+        RealmConfiguration wrongConfig = new RealmConfiguration.Builder(getContext())
+                .name(REALM_NAME)
+                .encryptionKey(TestHelper.SHA512("foo"))
+                .migration(realmMigration)
+                .schema(StringOnly.class)
+                .build();
+
+        RealmConfiguration rightConfig = new RealmConfiguration.Builder(getContext())
+                .name(REALM_NAME)
+                .encryptionKey(TestHelper.SHA512("realm"))
+                .migration(realmMigration)
+                .schema(StringOnly.class)
+                .build();
+
+        // Open Realm with wrong key
+        try {
+            Realm.getInstance(wrongConfig);
+            fail();
+        } catch (IllegalArgumentException ignored) {
+        }
+
+        // Try again with proper key
+        testRealm = Realm.getInstance(rightConfig);
+        assertNotNull(testRealm);
+        testRealm.close();
+    }
+
+    public void testDeletingRealmAlsoClearsConfigurationCache() throws IOException {
+        String REALM_NAME = "encrypted.realm";
+        byte[] oldPassword = TestHelper.SHA512("realm");
+        byte[] newPassword = TestHelper.SHA512("realm-copy");
+
+        TestHelper.copyRealmFromAssets(getContext(), REALM_NAME, REALM_NAME);
+        RealmMigration realmMigration = TestHelper.prepareMigrationToNullSupportStep();
+
+        RealmConfiguration config = new RealmConfiguration.Builder(getContext())
+                .name(REALM_NAME)
+                .encryptionKey(oldPassword)
+                .migration(realmMigration)
+                .schema(StringOnly.class)
+                .build();
+
+        // 1. Write a copy of the encrypted Realm to a new file
+        Realm testRealm = Realm.getInstance(config);
+        File copiedRealm = new File(config.getRealmFolder(), "encrypted-copy.realm");
+        if (copiedRealm.exists()) {
+            assertTrue(copiedRealm.delete());
+        }
+        testRealm.writeEncryptedCopyTo(copiedRealm, newPassword);
+        testRealm.close();
+
+        // 2. Delete the old Realm.
+        Realm.deleteRealm(config);
+
+        // 3. Rename the new file to the old file name.
+        assertTrue(copiedRealm.renameTo(new File(config.getRealmFolder(), REALM_NAME)));
+
+        // 4. Try to open the file again with the new password
+        // If the configuration cache wasn't cleared this would fail as we would detect two
+        // configurations with 2 different passwords pointing to the same file.
+        RealmConfiguration newConfig = new RealmConfiguration.Builder(getContext())
+                .name(REALM_NAME)
+                .encryptionKey(newPassword)
+                .migration(realmMigration)
+                .schema(StringOnly.class)
+                .build();
+
+        testRealm = Realm.getInstance(newConfig);
+        assertNotNull(testRealm);
+        testRealm.close();
+    }
+
+    // Tests that if the same Realm file is opened on multiple threads, we only need to validate the schema on the first thread.
+    public void testValidateSchemasOverThreads() throws InterruptedException, TimeoutException, ExecutionException {
+        final RealmConfiguration config = TestHelper.createConfiguration(getContext(), "foo");
+        Realm.deleteRealm(config);
+
+        final CountDownLatch bgThreadLocked = new CountDownLatch(1);
+        final CountDownLatch mainThreadDone = new CountDownLatch(1);
+
+        new Thread(new Runnable() {
+            @Override
+            public void run() {
+                Realm realm = Realm.getInstance(config);
+                realm.beginTransaction();
+                bgThreadLocked.countDown();
+                try {
+                    mainThreadDone.await(5, TimeUnit.SECONDS);
+                } catch (InterruptedException ignored) {
+                }
+                realm.close();
+            }
+        }).start();
+
+        ExecutorService executorService = Executors.newSingleThreadExecutor();
+        Future<Boolean> future = executorService.submit(new Callable<Boolean>() {
+            @Override
+            public Boolean call() throws Exception {
+                Realm realm = Realm.getInstance(config);
+                realm.close();
+                mainThreadDone.countDown();
+                return true;
+            }
+        });
+
+        bgThreadLocked.await(2, TimeUnit.SECONDS);
+        assertTrue(future.get(10, TimeUnit.SECONDS));
+    }
+
+    public void testDifferentThreadsDifferentInstance() throws InterruptedException {
+        final CountDownLatch closeLatch = new CountDownLatch(1);
+
+        final Realm realmA = Realm.getInstance(defaultConfig);
+
+        Thread thread = new Thread(new Runnable() {
+            @Override
+            public void run() {
+                Realm realmB = Realm.getInstance(defaultConfig);
+                assertFalse(realmA == realmB);
+                RealmCache.invokeWithGlobalRefCount(defaultConfig,
+                        new TestHelper.ExpectedCountCallback(RealmCacheTest.this, 2));
+                realmB.close();
+                closeLatch.countDown();
+            }
+        });
+        thread.start();
+
+        closeLatch.await();
+        RealmCache.invokeWithGlobalRefCount(defaultConfig, new TestHelper.ExpectedCountCallback(this, 1));
+        realmA.close();
+        RealmCache.invokeWithGlobalRefCount(defaultConfig, new TestHelper.ExpectedCountCallback(this, 0));
+    }
+
+    public void testReleaseCacheInOneThread() {
+        // Test release typed Realm instance
+        Realm realmA = RealmCache.createRealmOrGetFromCache(defaultConfig, Realm.class);
+        Realm realmB = RealmCache.createRealmOrGetFromCache(defaultConfig, Realm.class);
+        RealmCache.release(realmA);
+        assertNotNull(realmA.sharedGroupManager);
+        RealmCache.release(realmB);
+        assertNull(realmB.sharedGroupManager);
+        // No crash but warning in the log
+        RealmCache.release(realmB);
+
+        // Test release dynamic Realm instance
+        DynamicRealm dynamicRealmA = RealmCache.createRealmOrGetFromCache(defaultConfig,
+                DynamicRealm.class);
+        DynamicRealm dynamicRealmB = RealmCache.createRealmOrGetFromCache(defaultConfig,
+                DynamicRealm.class);
+        RealmCache.release(dynamicRealmA);
+        assertNotNull(dynamicRealmA.sharedGroupManager);
+        RealmCache.release(dynamicRealmB);
+        assertNull(dynamicRealmB.sharedGroupManager);
+        // No crash but warning in the log
+        RealmCache.release(dynamicRealmB);
+
+        // Test both typed Realm and dynamic Realm in same thread
+        realmA = RealmCache.createRealmOrGetFromCache(defaultConfig, Realm.class);
+        dynamicRealmA = RealmCache.createRealmOrGetFromCache(defaultConfig, DynamicRealm.class);
+        RealmCache.release(realmA);
+        assertNull(realmA.sharedGroupManager);
+        RealmCache.release(dynamicRealmA);
+        assertNull(realmA.sharedGroupManager);
+    }
+}
--- a/realm/realm-library/src/androidTest/java/io/realm/RealmTest.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/RealmTest.java
@@ -160,25 +160,6 @@ public class RealmTest extends AndroidTestCase {
         }
     }
 
-    public void testGetInstanceClearsCacheWhenFailed() {
-        String REALM_NAME = "invalid_cache.realm";
-        RealmConfiguration configA = TestHelper.createConfiguration(getContext(), REALM_NAME, TestHelper.getRandomKey(42));
-        RealmConfiguration configB = TestHelper.createConfiguration(getContext(), REALM_NAME, TestHelper.getRandomKey(43));
-
-        Realm.deleteRealm(configA);
-        Realm realm = Realm.getInstance(configA); // Create starting Realm with key1
-        realm.close();
-        try {
-            Realm.getInstance(configB); // Try to open with key 2
-        } catch (IllegalArgumentException expected) {
-            // Delete Realm so key 2 works. This should work as a Realm shouldn't be cached
-            // if initialization failed.
-            assertTrue(Realm.deleteRealm(configA));
-            realm = Realm.getInstance(configB);
-            realm.close();
-        }
-    }
-
     public void testCheckValid() {
         // checkIfValid() must not throw any Exception against valid Realm instance.
         testRealm.checkIfValid();
@@ -189,15 +170,7 @@ public class RealmTest extends AndroidTestCase {
             fail("closed Realm instance must throw IllegalStateException.");
         } catch (IllegalStateException ignored) {
         }
-    }
-
-    public void testRealmCache() {
-        Realm newRealm = Realm.getInstance(getContext());
-        try {
-          assertEquals(testRealm, newRealm);
-        } finally {
-          newRealm.close();
-        }
+        testRealm = null;
     }
 
     public void testInternalRealmChangedHandlersRemoved() {
@@ -211,6 +184,7 @@ public class RealmTest extends AndroidTestCase {
         Realm realm = null;
         try {
             realm = Realm.getInstance(realmConfig);
+            assertFalse(testRealm == realm);
             assertEquals(1, Realm.getHandlers().size());
             realm.close();
 
@@ -1911,6 +1885,7 @@ public class RealmTest extends AndroidTestCase {
         // After exception thrown in another thread, nothing should be changed to the realm in this thread.
         testRealm.checkIfValid();
         testRealm.close();
+        testRealm = null;
     }
 
     public void testRealmIsClosed() {
@@ -1949,120 +1924,6 @@ public class RealmTest extends AndroidTestCase {
         testRealm.close();
     }
 
-    // We should not cache wrong configurations
-    public void testDontCacheWrongConfigurations() throws IOException {
-        testRealm.close();
-        String REALM_NAME = "encrypted.realm";
-        TestHelper.copyRealmFromAssets(getContext(), REALM_NAME, REALM_NAME);
-        RealmMigration realmMigration = TestHelper.prepareMigrationToNullSupportStep();
-
-        RealmConfiguration wrongConfig = new RealmConfiguration.Builder(getContext())
-                .name(REALM_NAME)
-                .encryptionKey(TestHelper.SHA512("foo"))
-                .migration(realmMigration)
-                .schema(StringOnly.class)
-                .build();
-
-        RealmConfiguration rightConfig = new RealmConfiguration.Builder(getContext())
-                .name(REALM_NAME)
-                .encryptionKey(TestHelper.SHA512("realm"))
-                .migration(realmMigration)
-                .schema(StringOnly.class)
-                .build();
-
-        // Open Realm with wrong key
-        try {
-            testRealm = Realm.getInstance(wrongConfig);
-            fail();
-        } catch (IllegalArgumentException ignored) {
-        }
-
-        // Try again with proper key
-        testRealm = Realm.getInstance(rightConfig);
-        assertNotNull(testRealm);
-    }
-
-    public void testDeletingRealmAlsoClearsConfigurationCache() throws IOException {
-        testRealm.close();
-        String REALM_NAME = "encrypted.realm";
-        byte[] oldPassword = TestHelper.SHA512("realm");
-        byte[] newPassword = TestHelper.SHA512("realm-copy");
-
-        TestHelper.copyRealmFromAssets(getContext(), REALM_NAME, REALM_NAME);
-        RealmMigration realmMigration = TestHelper.prepareMigrationToNullSupportStep();
-
-        RealmConfiguration config = new RealmConfiguration.Builder(getContext())
-                .name(REALM_NAME)
-                .encryptionKey(oldPassword)
-                .migration(realmMigration)
-                .schema(StringOnly.class)
-                .build();
-
-        // 1. Write a copy of the encrypted Realm to a new file
-        testRealm = Realm.getInstance(config);
-        File copiedRealm = new File(config.getRealmFolder(), "encrypted-copy.realm");
-        copiedRealm.delete();
-        testRealm.writeEncryptedCopyTo(copiedRealm, newPassword);
-        testRealm.close();
-        testRealm = null;
-
-        // 2. Delete the old Realm.
-        Realm.deleteRealm(config);
-
-        // 3. Rename the new file to the old file name.
-        copiedRealm.renameTo(new File(config.getRealmFolder(), REALM_NAME));
-
-        // 4. Try to open the file again with the new password
-        // If the configuration cache wasn't cleared this would fail as we would detect two
-        // configurations with 2 different passwords pointing to the same file.
-        RealmConfiguration newConfig = new RealmConfiguration.Builder(getContext())
-                .name(REALM_NAME)
-                .encryptionKey(newPassword)
-                .migration(realmMigration)
-                .schema(StringOnly.class)
-                .build();
-
-        testRealm = Realm.getInstance(newConfig);
-        assertNotNull(testRealm);
-    }
-
-    // Tests that if the same Realm file is opened on multiple threads, we only need to validate the schema on the first thread.
-    public void testValidateSchemasOverThreads() throws InterruptedException, TimeoutException, ExecutionException {
-        final RealmConfiguration config = TestHelper.createConfiguration(getContext(), "foo");
-        Realm.deleteRealm(config);
-
-        final CountDownLatch bgThreadLocked = new CountDownLatch(1);
-        final CountDownLatch mainThreadDone = new CountDownLatch(1);
-
-        new Thread(new Runnable() {
-            @Override
-            public void run() {
-                Realm realm = Realm.getInstance(config);
-                realm.beginTransaction();
-                bgThreadLocked.countDown();
-                try {
-                    mainThreadDone.await(5, TimeUnit.SECONDS);
-                } catch (InterruptedException ignored) {
-                }
-                realm.close();
-            }
-        }).start();
-
-        ExecutorService executorService = Executors.newSingleThreadExecutor();
-        Future<Boolean> future = executorService.submit(new Callable<Boolean>() {
-            @Override
-            public Boolean call() throws Exception {
-                Realm realm = Realm.getInstance(config);
-                realm.close();
-                mainThreadDone.countDown();
-                return true;
-            }
-        });
-
-        bgThreadLocked.await(2, TimeUnit.SECONDS);
-        assertTrue(future.get(10, TimeUnit.SECONDS));
-    }
-
     // Realm validation & initialization is done once, still ColumnIndices
     // should be populated for the subsequent Realm sharing the same configuration
     // even if we skip initialization & validation
@@ -2086,7 +1947,9 @@ public class RealmTest extends AndroidTestCase {
                 } catch (Exception e) {
                     threadError[0] = e;
                 } finally {
-                    realm.close();
+                    if (!realm.isClosed()) {
+                        realm.close();
+                    }
                 }
             }
         }).start();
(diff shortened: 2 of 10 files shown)


realm/realm-library/src/androidTest/java/io/realm/RealmCacheTest.java
```diff
@@ -0,0 +1,283 @@
...
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+import io.realm.entities.AllTypes;
+import io.realm.entities.StringOnly;
+
+public class RealmCacheTest extends AndroidTestCase {
```
cmelchior: Any reason this class is public? It is inside `io.realm` so making it public means it is part of the public API?
Also a small class description would be nice.

realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -51,7 +51,7 @@
  *
  * @see io.realm.Realm
  */
-abstract class BaseRealm implements Closeable {
+class BaseRealm implements Closeable {
```
cmelchior: I would prefer to keep this class abstract. It shouldn't be used on it's own?

realm/realm-library/src/main/java/io/realm/DynamicRealm.java
```diff
@@ -241,42 +220,14 @@ public void executeTransaction(Transaction transaction) {
...
-            realm.acquireFileReference(configuration);
-
-            return realm;
-        }
+    /**
+     * Creates a {@link DynamicRealm} instance without checking the existence in the {@link RealmCache}.
+     *
+     * @return a {@link DynamicRealm} instance.
+     */
+    static DynamicRealm createInstance(RealmConfiguration configuration) {
```
cmelchior: What is the use case for this?

beeender: this will be called in RealmCache when there is no instance in current thread, call this method and put it into the cache. The ideal way could be 1. Ask from cache 2. no finding in cache. 3. Create the instance in `DynamicRealm.java` 4. put the new instance into the cache. Maybe this is better? maybe i can split the `createRealmOrGetFromCache` into two functions, and always create the instance here. But in that way, the problem is, the lock needs to be in this function as well :(

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -0,0 +1,269 @@
...
+ */
+class RealmCache {
+
+    private class RefAndCount {
+        // The Realm instance in this thread.
+        private final ThreadLocal<BaseRealm> localRealm = new ThreadLocal<>();
+        // How many references to this Realm instance in this thread.
+        private final ThreadLocal<Integer> localCount = new ThreadLocal<>();
+        // How many threads have instances refer to this configuration.
+        private int globalCount = 0;
```
cmelchior: Mixing ThreadLocal and class variables in this class seems really dangerous to me? It would only be safe if you wrap any access to globalCount in a lock (granted that seems to be happening in the code below, but still). It would feel much safer if you could only access the globalCount through a setter in the class which could guarantee that no race conditions happened? Converting it to an AtomicInt would also solve it?

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -0,0 +1,269 @@
...
+                return TYPED_REALM;
+            } else if (cls == DynamicRealm.class) {
+                return DYNAMIC_REALM;
+            }
+
+            throw new IllegalArgumentException("The type of Realm class must be Realm or DynamicRealm.");
+        }
+    }
+    // Separated references and counters for typed Realm and dynamic Realm.
+    private final EnumMap<RealmCacheType, RefAndCount> refAndCountMap;
```
cmelchior: Doesn't this mean you can only have 1 Realm on each thread as the RefAndCount get overridden by another Realm with a different configuration?

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -0,0 +1,269 @@
...
+import io.realm.internal.RealmProxyMediator;
+import io.realm.internal.SharedGroup;
+import io.realm.internal.log.RealmLog;
+
+/**
+ * To cache {@link Realm}, {@link DynamicRealm} instances and related resources.
+ * Every thread will share the same {@link Realm} and {@link DynamicRealm} instances which are referred to the same
+ * {@link RealmConfiguration}.
+ */
+class RealmCache {
```
cmelchior: Perhaps also add a RealmCache is created for each Configuration

realm/realm-library/src/main/java/io/realm/RealmCache.java
beeender: Why do you think it is dangerous? `RefAndCount` is for per cache and all operations in RealmCache is `synchronized` on `RealmCache.class`. AtomicInt doesn't seem to be necessary?

realm/realm-library/src/main/java/io/realm/RealmCache.java
beeender: Ah, yes, `RealmCache` is for per configuration.

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -214,7 +193,7 @@ public static Realm getInstance(RealmConfiguration configuration) {
         if (configuration == null) {
             throw new IllegalArgumentException("A non-null RealmConfiguration must be provided");
         }
-        return create(configuration);
+        return (Realm) RealmCache.createRealmOrGetFromCache(configuration, Realm.class);
```
cmelchior: Can't you avoid this cast by using generics?

```
    static synchronized <E extends BaseRealm> E createRealmOrGetFromCache(RealmConfiguration configuration, Class<E> realmClass) {
```

beeender: Thanks! I tried to figure out the right way to use generics here but failed!

realm/realm-library/src/main/java/io/realm/RealmCache.java
cmelchior: Yes, i missed that RealmCache is created pr. RealmConfiguration. I guess it is okay...

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -0,0 +1,269 @@
...
+    // Column indices are cached to speed up opening typed Realm. If a Realm instance is created in one thread, creating
+    // Realm instances in other threads doesn't have to initialize the column indices again.
+    private ColumnIndices typedColumnIndices;
+
+    // Different Realm configuration with same path are not allowed.
+    private static Map<String, RealmCache> cachesMap = new HashMap<>();
+
+    private static final String DIFFERENT_KEY_MESSAGE = "Wrong key used to decrypt Realm.";
+
+
```
cmelchior: Whitespace

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -0,0 +1,269 @@
...
+        }
+        if (refCount == null) {
+            refCount = 0;
+        }
+
+        if (refCount <= 0) {
+            RealmLog.w("Realm " + canonicalPath + " is not in the cache anymore.");
+            return -1;
+        }
+
```
cmelchior: Whitespace

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -0,0 +1,269 @@
...
+        if (cache != null) {
+            refAndCount = cache.refAndCountMap.get(RealmCacheType.valueOf(realm.getClass()));
+            refCount = refAndCount.localCount.get();
+        }
+        if (refCount == null) {
+            refCount = 0;
+        }
+
+        if (refCount <= 0) {
+            RealmLog.w("Realm " + canonicalPath + " is not in the cache anymore.");
```
cmelchior: More user friendly warning needed. Isn't this replacing the old "Calling Realm.close() too many times..." message?

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -0,0 +1,269 @@
...
+            // The last instance in this thread.
+            // Clear local ref & counter
+            refAndCount.localCount.set(null);
+            refAndCount.localRealm.set(null);
+
+            // Clear global counter
+            refAndCount.globalCount--;
+            if (refAndCount.globalCount < 0) {
+                // Should never happen.
+                throw new IllegalStateException("Global reference counter of Realm" + canonicalPath +
```
cmelchior: gets -> got.

realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -633,47 +540,51 @@ protected static synchronized void validateAgainstExistingConfigurations(RealmCo
     /**
      * Deletes the Realm file defined by the given configuration.
      */
-    protected static synchronized boolean deleteRealm(RealmConfiguration configuration) {
-        if (isFileOpen(configuration)) {
-            throw new IllegalStateException("It's not allowed to delete the file associated with an open Realm. " +
-                    "Remember to close() all the instances of the Realm before deleting its file.");
-        }
+    protected static boolean deleteRealm(RealmConfiguration configuration) {
+        synchronized (RealmCache.class) {
```
cmelchior: I was really hoping we could avoid something like this, as these kind of rules is what caused the mess in the first place. The only reason for this, seems to be because `isFileOpen` is calling down into the RealmCache?

On way to avoid having the lock here would be to provide a callback class to RealmCache.

RealmCache.isFileOpen(configuration, new Callback() {
  // Callback called from inside RealmCache when locking is guarenteed
});

What do you think?

beeender: Looks good! I will change it like that.

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -0,0 +1,269 @@
...
+        return refCount;
+    }
+
+    /**
+     * Make sure that the new configuration doesn't clash with any existing configurations for the
+     * Realm.
+     *
+     * @throws IllegalArgumentException If the new configuration isn't valid.
+     */
+    private static void validateAgainstExistingConfigurations(RealmConfiguration newConfiguration,
```
cmelchior: In the previous incarnation you could have multiple configurations pointing to the same file. Now there only seem to be one allowed. I am not sure we had a use case yet for the other implementation so it is probably fine, but at least renaming this to `validateAgainstExistingConfiguration`

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -0,0 +1,269 @@
...
+                        " has a different hash code with a cached configuration. " +
+                        "The new configuration's hashcode: " + newConfiguration.hashCode() +
+                        "The cached configuration's hashcode: " + cachedConfiguration.hashCode() + ".");
+    }
+
+    /**
+     * Returns the total reference count of {@link Realm} and {@link DynamicRealm} refer to the given
+     * {@link RealmConfiguration}.
+     * <p>
+     * This function is not thread safe. Consider to call this function and related logic in a {@code synchronized}
```
cmelchior: Yes! or see my other suggestion

realm/realm-library/src/main/java/io/realm/RealmCache.java
beeender: done

realm/realm-library/src/main/java/io/realm/RealmCache.java
beeender: done

realm/realm-library/src/main/java/io/realm/RealmCache.java
beeender: done

realm/realm-library/src/main/java/io/realm/RealmCache.java
beeender: done

realm/realm-library/src/main/java/io/realm/RealmCache.java
beeender: done

realm/realm-library/src/main/java/io/realm/RealmCache.java
beeender: done

realm/realm-library/src/main/java/io/realm/Realm.java
cmelchior: I might be wrong, so nevermind :)

realm/realm-library/src/main/java/io/realm/DynamicRealm.java
cmelchior: Ok, just I would just keep it as it is then. We definitely want to avoid having locks outside RealmCache. I would probably describe the use case a bit better in the Javadoc though as it sounds like this should only be called from RealmCache?

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -0,0 +1,271 @@
...
+    // Separated references and counters for typed Realm and dynamic Realm.
+    private final EnumMap<RealmCacheType, RefAndCount> refAndCountMap;
+
+    private RealmConfiguration configuration;
+    // Column indices are cached to speed up opening typed Realm. If a Realm instance is created in one thread, creating
+    // Realm instances in other threads doesn't have to initialize the column indices again.
+    private ColumnIndices typedColumnIndices;
+
+    // Different Realm configuration with same path are not allowed.
+    private static Map<String, RealmCache> cachesMap = new HashMap<>();
```
cmelchior: Java 7 `<>`

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -0,0 +1,271 @@
...
+
+    private static final String DIFFERENT_KEY_MESSAGE = "Wrong key used to decrypt Realm.";
+
+    private RealmCache(RealmConfiguration config) {
+        configuration = config;
+        refAndCountMap = new EnumMap<>(RealmCacheType.class);
```
cmelchior: Java 7 `<>`

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -0,0 +1,271 @@
...
+        for (RealmCacheType type : RealmCacheType.values()) {
+            refAndCountMap.put(type, new RefAndCount());
+        }
+    }
+
+    /**
+     * Creates a new Realm instance or get a existing instance for current thread.
```
cmelchior: a -> an

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -0,0 +1,271 @@
...
+        boolean isCacheInMap = true;
+        RealmCache cache = cachesMap.get(configuration.getPath());
+        if (cache == null) {
+            // Create a new cache
+            cache = new RealmCache(configuration);
+            // The new cache should be added to the map later
+            isCacheInMap = false;
+        } else {
+            // != takes no time, check it first
+            if (cache.configuration != configuration && cache.configuration.hashCode() != configuration.hashCode()) {
```
cmelchior: Couldn't this block be inlined in `RealmCache` as this would make the code here more readable?

```
cache.validateConfiguration(configuration)
```

or something?

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -0,0 +1,271 @@
...
+        RefAndCount refAndCount = cache.refAndCountMap.get(RealmCacheType.valueOf(realmClass));
+
+        if (refAndCount.localRealm.get() == null) {
+            // Create a new local Realm instance
+            BaseRealm realm;
+
+            if (realmClass == Realm.class) {
+                // RealmMigrationNeededException might be thrown here
+                realm = Realm.createInstance(configuration, cache.typedColumnIndices);
+            } else {
```
cmelchior: The method declaration uses `<E extends BaseRealm>`, so just using else here is a bit dangerous. I would replace with:

```
if (realmClass == Realm.class) {

} else if (realmClass == DynamicRealm.class) {

} else {
  throw
}
```

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -0,0 +1,271 @@
...
+ * {@link RealmConfiguration}.
+ * One {@link RealmCache} is created for each {@link RealmConfiguration}.
+ */
+class RealmCache {
+
+    interface GlobalReferenceCountCallback {
+        void onResult(int count);
+    }
+
+    private class RefAndCount {
```
zaki50: should be static?

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -0,0 +1,271 @@
...
+
+    /**
+     * Runs the callback function with the total reference count of {@link Realm} and {@link DynamicRealm} refer to the
+     * given {@link RealmConfiguration}.
+     *
+     * @param configuration the {@link RealmConfiguration} of {@link Realm} and {@link DynamicRealm}.
+     * @param callback the callback will be executed with the global reference count.
+     *
+     */
+    static synchronized void getGlobalRefCount(RealmConfiguration configuration, GlobalReferenceCountCallback callback) {
```
cmelchior: Something about the name of this method and the callback class just feels wrong to me.
Isn't the purpose of this method to run some code using the RealmCache, and getting the ref counter is just a side-effect of that?

Not sure I have a good alternative at the moment though

realm/realm-library/src/main/java/io/realm/RealmCache.java
beeender: Exception will be thrown if it is BaseRealm in line 108. But i will add the `if else` case to make it clearer.

beeender:
@realm/java Ready for review :)

realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -637,47 +533,61 @@ protected static synchronized void validateAgainstExistingConfigurations(RealmCo
     /**
      * Deletes the Realm file defined by the given configuration.
      */
-    protected static synchronized boolean deleteRealm(RealmConfiguration configuration) {
-        if (isFileOpen(configuration)) {
-            throw new IllegalStateException("It's not allowed to delete the file associated with an open Realm. " +
-                    "Remember to close() all the instances of the Realm before deleting its file.");
-        }
+    protected static boolean deleteRealm(final RealmConfiguration configuration) {
+        final boolean[] realmDeleted = {true};
```
cmelchior: I would replace this with an AtomicBoolean instead. It expresses the intent better IMO.

realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -637,47 +533,61 @@ protected static synchronized void validateAgainstExistingConfigurations(RealmCo
...
-    public static synchronized boolean compactRealm(RealmConfiguration configuration) {
+    public static boolean compactRealm(final RealmConfiguration configuration) {
         if (configuration.getEncryptionKey() != null) {
             throw new IllegalArgumentException("Cannot currently compact an encrypted Realm.");
         }
 
-        if (isFileOpen(configuration)) {
-            throw new IllegalStateException("Cannot compact an open Realm");
-        }
+        final boolean[] result = {false};
```
cmelchior: I would replace this with an AtomicBoolean instead. It expresses the intent better IMO.

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -0,0 +1,279 @@
...
+
+/**
+ * To cache {@link Realm}, {@link DynamicRealm} instances and related resources.
+ * Every thread will share the same {@link Realm} and {@link DynamicRealm} instances which are referred to the same
+ * {@link RealmConfiguration}.
+ * One {@link RealmCache} is created for each {@link RealmConfiguration}.
+ */
+class RealmCache {
+
+    interface RealmCacheCallback {
```
cmelchior: I would rename this to just `Callback` `RealmCache.RealmCacheCallback` seems a bit redundant.

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -0,0 +1,279 @@
...
+    }
+    // Separated references and counters for typed Realm and dynamic Realm.
+    private final EnumMap<RealmCacheType, RefAndCount> refAndCountMap;
+
+    private RealmConfiguration configuration;
+    // Column indices are cached to speed up opening typed Realm. If a Realm instance is created in one thread, creating
+    // Realm instances in other threads doesn't have to initialize the column indices again.
+    private ColumnIndices typedColumnIndices;
+
+    // Different Realm configuration with same path are not allowed.
```
cmelchior: The comment seems a bit odd, since it conflicts with the message in the class header. Given that description I would assume that this constraint is already enforced?

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -0,0 +1,279 @@
...
+                        "The cached configuration's hashcode: " + configuration.hashCode() + ".");
+    }
+
+    /**
+     * Runs the callback function with the total reference count of {@link Realm} and {@link DynamicRealm} refer to the
+     * given {@link RealmConfiguration}.
+     *
+     * @param configuration the {@link RealmConfiguration} of {@link Realm} and {@link DynamicRealm}.
+     * @param callback the callback will be executed with the global reference count.
+     *
```
cmelchior: Whitespace

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -0,0 +1,279 @@
...
+        // Should never get here
+        throw new IllegalArgumentException(
+                "The new configuration pointing to " + newConfiguration.getPath() +
+                        " has a different hash code with a cached configuration. " +
+                        "The new configuration's hashcode: " + newConfiguration.hashCode() +
+                        "The cached configuration's hashcode: " + configuration.hashCode() + ".");
+    }
+
+    /**
+     * Runs the callback function with the total reference count of {@link Realm} and {@link DynamicRealm} refer to the
```
cmelchior: who refer

cmelchior:
Looks good :+1: 
We might want to make sure to get extra :+1: s for this one as changing how caching/locks works have bitten us before.

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -0,0 +1,279 @@
...
+        // How many references to this Realm instance in this thread.
+        private final ThreadLocal<Integer> localCount = new ThreadLocal<Integer>();
+        // How many threads have instances refer to this configuration.
+        private int globalCount = 0;
+    }
+    private enum RealmCacheType {
+        TYPED_REALM,
+        DYNAMIC_REALM;
+
+        static RealmCacheType valueOf(Class<? extends BaseRealm> cls) {
```
nhachicha: cls -> clazz 
to keep the tradition 🐘

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -0,0 +1,279 @@
...
+                return DYNAMIC_REALM;
+            }
+
+            throw new IllegalArgumentException(WRONG_REALM_CLASS_MESSAGE);
+        }
+    }
+    // Separated references and counters for typed Realm and dynamic Realm.
+    private final EnumMap<RealmCacheType, RefAndCount> refAndCountMap;
+
+    private RealmConfiguration configuration;
```
nhachicha: final

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -0,0 +1,279 @@
...
+     *
+     * @param configuration {@link RealmConfiguration} will be used to create or get the instance.
+     * @param realmClass class of {@link Realm} or {@link DynamicRealm} to be created in or gotten from the cache.
+     * @return the {@link Realm} or {@link DynamicRealm} instance.
+     */
+    @SuppressWarnings("unchecked")
+    static synchronized <E extends BaseRealm> E createRealmOrGetFromCache(RealmConfiguration configuration,
+                                                        Class<E> realmClass) {
+        boolean isCacheInMap = true;
+        RealmCache cache = cachesMap.get(configuration.getPath());
```
nhachicha: we're not checking the thread associated with the `RealmCache`  we got from the map.
different thread should open a new Realm + RealmCache, & cachesMap is not thread local?

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -0,0 +1,279 @@
...
+            // This is the first instance in current thread, increase the global count.
+            refAndCount.globalCount++;
+        }
+        refAndCount.localCount.set(refCount + 1);
+
+        return (E) refAndCount.localRealm.get();
+    }
+
+    /**
+     * Releases a given {@link Realm} or {@link DynamicRealm} from cache. The instance will be closed by this function
```
nhachicha: function -> method

realm/realm-library/src/main/java/io/realm/RealmCache.java
beeender: RealmCache is shared between threads. different thread should open a new Realm + the same RealmCache.

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -0,0 +1,279 @@
...
+                        " has a different hash code with a cached configuration. " +
+                        "The new configuration's hashcode: " + newConfiguration.hashCode() +
+                        "The cached configuration's hashcode: " + configuration.hashCode() + ".");
+    }
+
+    /**
+     * Runs the callback function with the total reference count of {@link Realm} and {@link DynamicRealm} who refer to
+     * the given {@link RealmConfiguration}.
+     *
+     * @param configuration the {@link RealmConfiguration} of {@link Realm} and {@link DynamicRealm}.
```
nhachicha: and -> or

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -0,0 +1,279 @@
...
+
+    /**
+     * Make sure that the new configuration doesn't clash with any existing configurations for the
+     * Realm.
+     *
+     * @throws IllegalArgumentException If the new configuration isn't valid.
+     */
+    private void validateConfiguration(RealmConfiguration newConfiguration) {
+        // != takes no time, check it first
+        if (newConfiguration == configuration || newConfiguration.hashCode() == configuration.hashCode()) {
```
zaki50: same hashCode does not mean same configuration.

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -0,0 +1,279 @@
...
+        }
+
+        // No more local reference to this Realm in current thread.
+        if (refCount == 0) {
+            realm.doClose();
+        }
+    }
+
+    /**
+     * Make sure that the new configuration doesn't clash with any existing configurations for the
```
zaki50: any existing configurations -> cached configuration ?

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -0,0 +1,279 @@
...
+     * Realm.
+     *
+     * @throws IllegalArgumentException If the new configuration isn't valid.
```
zaki50: If -> if

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -0,0 +1,279 @@
...
+        if (!cachedDurability.equals(newDurability)) {
+            throw new IllegalArgumentException("A Realm cannot be both in-memory and persisted. Two conflicting " +
+                    "configurations pointing to " + newConfiguration.getPath() + " are being used.");
+        }
+
+        // Should never get here
+        throw new IllegalArgumentException(
+                "The new configuration pointing to " + newConfiguration.getPath() +
+                        " has a different hash code with a cached configuration. " +
+                        "The new configuration's hashcode: " + newConfiguration.hashCode() +
```
zaki50: hashcode -> hash code?

realm/realm-library/src/main/java/io/realm/RealmCache.java
nhachicha: you need to override/use equals 
You're already checking for the `EncryptionKey()` equality

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -250,73 +236,37 @@ private static synchronized Realm create(RealmConfiguration configuration) {
                 migrateRealm(configuration);
             }
 
-            return createAndValidate(configuration, true, autoRefresh);
+            return createAndValidate(configuration, columnIndices);
```
zaki50: Is there any case that `columnIndices` is valid after migrating?

Is `columnIndices` always `null` if reach here?

beeender: Before entering migration block, there should be no opened typed Realm, which means, columnIndices should always be null before migration.

realm/realm-library/src/main/java/io/realm/RealmCache.java
nhachicha: maybe you should update the Javadoc to clear any confusion?

> To cache {@link Realm}, {@link DynamicRealm} instances and related resources.

sounds like it's caching all Realm & DynamicRealm per conf, whereas the caching should be per thread, per configuration?

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -0,0 +1,279 @@
...
+    }
+
+    /**
+     * Runs the callback function with the total reference count of {@link Realm} and {@link DynamicRealm} who refer to
+     * the given {@link RealmConfiguration}.
+     *
+     * @param configuration the {@link RealmConfiguration} of {@link Realm} and {@link DynamicRealm}.
+     * @param callback the callback will be executed with the global reference count.
+     */
+    static synchronized void invokeWithGlobalRefCount(RealmConfiguration configuration, Callback callback) {
```
nhachicha: If it's used only for test purpose why not use Reflection? + you're adding a new public interface Callback?

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -0,0 +1,279 @@
...
+        if (refCount == 0) {
+            if (realmClass == Realm.class && refAndCount.globalCount == 0) {
+                cache.typedColumnIndices = refAndCount.localRealm.get().schema.columnIndices;
+            }
+            // This is the first instance in current thread, increase the global count.
+            refAndCount.globalCount++;
+        }
+        refAndCount.localCount.set(refCount + 1);
+
+        return (E) refAndCount.localRealm.get();
```
zaki50: we can narrow the scope of `@SuppressWarnings("unchecked")` by applying to variable declaration instead of applying to method declaration.

```
        E realm = (E) refAndCount.localRealm.get();
        return realm;
```

realm/realm-library/src/androidTest/java/io/realm/RealmCacheTest.java
```diff
@@ -0,0 +1,291 @@
...
+
+    @Override
+    protected void setUp() throws Exception {
+        super.setUp();
+        defaultConfig = TestHelper.createConfiguration(getContext());
+        Realm.deleteRealm(defaultConfig);
+    }
+
+    @Override
+    protected void tearDown() throws Exception {
```
nhachicha: remove, you're not doing any logic

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -0,0 +1,279 @@
...
+            // No more instance of typed Realm and dynamic Realm. Remove the configuration from cache.
+            if (totalRefCount == 0) {
+                cachesMap.remove(canonicalPath);
+            }
+        } else {
+            refAndCount.localCount.set(refCount);
+        }
+
+        // No more local reference to this Realm in current thread.
+        if (refCount == 0) {
```
zaki50: Why do we check `refCount == 0` twice?

realm/realm-library/src/androidTest/java/io/realm/RealmCacheTest.java
```diff
@@ -0,0 +1,291 @@
...
+        String REALM_NAME = "invalid_cache.realm";
+        RealmConfiguration configA = TestHelper.createConfiguration(getContext(), REALM_NAME, TestHelper.getRandomKey(42));
+        RealmConfiguration configB = TestHelper.createConfiguration(getContext(), REALM_NAME, TestHelper.getRandomKey(43));
+
+        Realm.deleteRealm(configA);
+        Realm realm = Realm.getInstance(configA); // Create starting Realm with key1
+        realm.close();
+        try {
+            Realm.getInstance(configB); // Try to open with key 2
+        } catch (IllegalArgumentException expected) {
```
nhachicha: expected -> ignored

zaki50:
@beeender I added some comments. It's great work and much easier to read!

realm/realm-library/src/androidTest/java/io/realm/RealmCacheTest.java
```diff
@@ -0,0 +1,291 @@
...
+                RealmCache.invokeWithGlobalRefCount(defaultConfig,
+                        new TestHelper.ExpectedCountCallback(RealmCacheTest.this, 2));
+                openLatch.countDown();
+                realmB.close();
+                closeLatch.countDown();
+            }
+        });
+        thread.start();
+
+        openLatch.await();
```
nhachicha: what's the purpose of `openLatch`? you can just use `closeLatch` to make the `InstrumentationThread` block,
you can even use `thread.join()` without any `CountDownLatch`

realm/realm-library/src/androidTest/java/io/realm/RealmCacheTest.java
```diff
@@ -0,0 +1,291 @@
...
+        assertNotNull(dynamicRealmA.sharedGroupManager);
+        RealmCache.release(dynamicRealmB);
+        assertNull(dynamicRealmB.sharedGroupManager);
+        // No crash but warning in the log
+        RealmCache.release(dynamicRealmB);
+
+        // Test both typed Realm and dynamic Realm in same thread
+        realmA = RealmCache.createRealmOrGetFromCache(defaultConfig, Realm.class);
+        dynamicRealmA = RealmCache.createRealmOrGetFromCache(defaultConfig,
+                DynamicRealm.class);
```
nhachicha: could be in same line?

realm/realm-library/src/androidTest/java/io/realm/RealmTest.java
```diff
@@ -2138,7 +2001,8 @@ public void onChange() {
         testRealm.createObject(Dog.class);
         testRealm.commitTransaction();
         bgThreadLatch.countDown();
-        awaitOrFail(bgClosedLatch);
+        //awaitOrFail(bgClosedLatch);
```
nhachicha: remove 💀 code

realm/realm-library/src/androidTest/java/io/realm/TestHelper.java
```diff
@@ -46,6 +47,23 @@
 
 public class TestHelper {
 
+    public static class ExpectedCountCallback implements RealmCache.Callback {
```
nhachicha: Like I mentioned at the bottom I don't like the fact we're using callback to communicate results (its a blocking call anyway so `invokeWithGlobalRefCount` should just return the value) this is starting to look like Javascript 😱

realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -637,47 +534,61 @@ protected static synchronized void validateAgainstExistingConfigurations(RealmCo
...
-        if (isFileOpen(configuration)) {
-            throw new IllegalStateException("It's not allowed to delete the file associated with an open Realm. " +
-                    "Remember to close() all the instances of the Realm before deleting its file.");
-        }
+    protected static boolean deleteRealm(final RealmConfiguration configuration) {
+        final AtomicBoolean realmDeleted = new AtomicBoolean(true);
+
+        RealmCache.invokeWithGlobalRefCount(configuration, new RealmCache.Callback() {
+            @Override
+            public void onResult(int count) {
```
nhachicha: see my previous comments about the callback, you can just call 
```java
int count = RealmCache.invokeWithGlobalRefCount(configuration);
```
and proceed with the rest of the logic

realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -637,47 +534,61 @@ protected static synchronized void validateAgainstExistingConfigurations(RealmCo
...
         }
 
-        if (isFileOpen(configuration)) {
-            throw new IllegalStateException("Cannot compact an open Realm");
-        }
+        final AtomicBoolean result = new AtomicBoolean(false);
+
+        RealmCache.invokeWithGlobalRefCount(configuration, new RealmCache.Callback() {
+            @Override
+            public void onResult(int count) {
```
nhachicha: if you need the following code to run in a synchronized block, just use an intrinsic or Reentrant lock

realm/realm-library/src/androidTest/java/io/realm/TestHelper.java
cmelchior: I don't agree. This change is good because then we can guarantee that the code runs within the proper locks. Otherwise you require the calling code to implement those locks, and then we end up where we are now.

It doesn't look particular nice though, that we can agree on :)

bmunkholm:
ObjectStore also handles caching of Realms, so you want to consider that as we start to integrate ObjectStore.

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -0,0 +1,279 @@
...
+            }
+
+            throw new IllegalArgumentException(WRONG_REALM_CLASS_MESSAGE);
+        }
+    }
+    // Separated references and counters for typed Realm and dynamic Realm.
+    private final EnumMap<RealmCacheType, RefAndCount> refAndCountMap;
+
+    private RealmConfiguration configuration;
+    // Column indices are cached to speed up opening typed Realm. If a Realm instance is created in one thread, creating
```
zaki50: add empty line before comment

realm/realm-library/src/main/java/io/realm/RealmCache.java
beeender: No, this one is called in `BaseRealm.compactRealm()` `BaseRealm.deleteRealm()` and `BaseRealm.migrateRealm()`

realm/realm-library/src/androidTest/java/io/realm/RealmCacheTest.java
beeender: My bad, removed openLatch.

realm/realm-library/src/main/java/io/realm/BaseRealm.java
beeender: That code needs to be synchronized on `RealmCache.class`, and use by using the callback, the relevant logic can be put in the `RealmCache` which is easier to read. Like cm commented before, this doesn't look very good, but still clear enough to force the callback happen in a locked block.

realm/realm-library/src/main/java/io/realm/RealmCache.java
beeender: Changed to using `equals`

beeender:
@bmunkholm That would be one reason we make this ---- decoupling the related code to another file. I think integrate ObjectStore won't be difficult, the relevant changes should be in the RealCache only. I will check the ObjectStore implementation.

beeender:
@realm/java Please, please, please help to review again. THANKS!

realm/realm-library/src/main/java/io/realm/RealmCache.java
zaki50: Isn't the checks below is a subset of `equals`?
If yes, we can simply remove the check of `equals`.

zaki50: I mean `if (newConfiguration == configuration)` should be here for just a shortcut.

cmelchior:
:+1:

realm/realm-library/src/main/java/io/realm/RealmCache.java
beeender: `if (newConfiguration == configuration)` is checked in the `equals`. And the point of below is just about to send the right exception message :)

zaki50: Difference of `deleteRealmIfMigrationNeeded` and `migration` are not checked below.

We need to remove the comment `// Should never get here` or to add more checks for those differences.

zaki50: If we add some more configuration items to the `RealmConfiguration` in the future, we need to update this method.
That means we should move this method to `RealmConfiguration` class?

beeender: I just checked the possibility of moving the `validateConfiguration` to `RealmConfiguration`, the problem is the exception message is not general enough to be in the `RealmConfiguration`. I suggest to keep it here.

dalinaum:
:+1:

zaki50:
:+1:

beeender:
retest this please

beeender:
retest this please

Issue #1803, opened 2015-11-20
average returns wrong value for nullable field


zaki50:
Reported by @dev75py on twitter. https://twitter.com/dev75py/status/667469537986527232

On 0.85.0, `RealmQuery.average()` returns wrong value for nullable field.

```
    public void testAverage() {
        final NullTypes obj1 = new NullTypes();
        obj1.setId(1);
        obj1.setFieldIntegerNotNull(3);
        obj1.setFieldIntegerNull(3);

        final NullTypes obj2 = new NullTypes();
        obj2.setId(2);
        obj2.setFieldIntegerNotNull(4);
        obj2.setFieldIntegerNull(4);

        testRealm.beginTransaction();
        testRealm.where(NullTypes.class).findAll().clear();
        testRealm.copyToRealm(obj1);
        testRealm.copyToRealm(obj2);
        testRealm.commitTransaction();

        assertEquals(3.5, testRealm.where(NullTypes.class).average(NullTypes.FIELD_INTEGER_NOT_NULL));
        assertEquals(3.5, testRealm.where(NullTypes.class).average(NullTypes.FIELD_INTEGER_NULL));
    }
```

The last assertion fails with actual value `9.0`.

cmelchior:
@zaki50 Hasn't this been fixed?

zaki50:
@cmelchior I think it's fixed.
I'll confirm and close it

zaki50:
This issue was fixed in 0.86.0

PR #1805, opened 2015-11-20
Add tests of average for nullable column

zaki50:
Added tests for aggregation methods in `RealmQuery` for partial null rows .

`RealmQueryTest.testSumGivesCorrectValueForPartialNullRows()` and `RealmQueryTest.testAvgGivesCorrectValueForAllNonNullRows()` tests wrong value for now.

These tests must be updated after core's fix.

related to #1803

### DIFF
--- a/realm/realm-library/src/androidTest/java/io/realm/RealmQueryTest.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/RealmQueryTest.java
@@ -1011,6 +1011,27 @@ public class RealmQueryTest extends AndroidTestCase {
         assertNull(query.minimumDate(NullTypes.FIELD_DATE_NULL));
     }
 
+    // Test min on columns with partial null rows
+    public void testMinGivesCorrectValueForAllNonNullRows() {
+        TestHelper.populateAllNonNullRowsForNumericTesting(testRealm);
+        RealmQuery<NullTypes> query = testRealm.where(NullTypes.class);
+
+        assertEquals(-1, query.min(NullTypes.FIELD_INTEGER_NULL).intValue());
+        assertEquals(-2f, query.min(NullTypes.FIELD_FLOAT_NULL).floatValue(), 0f);
+        assertEquals(-3d, query.min(NullTypes.FIELD_DOUBLE_NULL).doubleValue(), 0d);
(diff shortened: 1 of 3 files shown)


cmelchior:
Instead of letting the unit test fail, you could also let it detect the wrong value. That way we could merge these unit tests right away instead of letting them wait?

zaki50:
@cmelchior sure

zaki50:

realm/realm-library/src/androidTest/java/io/realm/RealmQueryTest.java
```diff
@@ -1011,6 +1011,27 @@ public void testMinValueForAllNullColumns() {
         assertNull(query.minimumDate(NullTypes.FIELD_DATE_NULL));
     }
 
+    // Test min on columns with partial null rows
+    public void testMinGivesCorrectValueForAllNonNullRows() {
```
cmelchior: I would remove "GivesCorrect" from these. Hopefully that should be implicit :) So `testMinForAllNonNullRows`, same for the others.

realm/realm-library/src/androidTest/java/io/realm/RealmQueryTest.java
```diff
@@ -1049,6 +1092,28 @@ public void testAverageValueForAllNullColumns() {
         assertEquals(0d, query.average(NullTypes.FIELD_DOUBLE_NULL), 0d);
     }
 
+    // Test average on columns with partial null rows
```
cmelchior: This description doesn't seem to match the unit test (seems like copy paste from below?)

cmelchior:
Apart from a few minor naming/doc issues it looks good :+1:

realm/realm-library/src/androidTest/java/io/realm/RealmQueryTest.java
zaki50: done

realm/realm-library/src/androidTest/java/io/realm/RealmQueryTest.java
zaki50: done

kneth:
:+1:

PR #1809, opened 2015-11-22
Enhancing RealmResults & RealmObject notifications

nhachicha:
- [X] Adding support for async queries for Dynamic Realm API
- [X] Fixing a bug where `findFirstAsync` was not rerunning the query if the initial value was empty  
- [X] Adding more tests for type-based notifications 
- [X] Callbacks (RealmChangeListeners) are now called on both synchronous & asynchronous `RealmResults` & `RealmObject`
- [X] Using [Table version](https://github.com/realm/realm-core/issues/1323) to reduce the number of callback invocations (this will filter out all commits non impacting the query used to obtain the RealmResults/RealmObject)
- [X] Some tests involving `RealmObject` callbacks are failing, we're waiting for https://github.com/realm/realm-core/issues/1323 implementation


nhachicha:
Hi @rgrinberg 
No since we'll be using internally a weak reference listeners for Adapter

nhachicha:
@realm/java

changelog.txt
```diff
@@ -1,3 +1,6 @@
+0.87.0
+* BREAKING CHANGE: Realm.addChangeListener, RealmObject.addChangeListener and RealmResults.addChangeListener will throw an IllegalStateException when invoked on a non-Looper thread. This is to prevent registering listeners that will never be invoked (since we rely on Handler/Looper).
```
cmelchior: Not sure about the comment in `()`...but not a big issue

realm/realm-jni/src/io_realm_internal_table.cpp
```diff
@@ -1693,3 +1693,20 @@ JNIEXPORT jboolean JNICALL Java_io_realm_internal_Table_nativeHasSameSchema
...
 }
+
+
+JNIEXPORT jlong JNICALL Java_io_realm_internal_Table_nativeVersion(
+        JNIEnv* env, jobject, jlong nativeTablePtr)
+{
+    bool valid = (nativeTablePtr != 0);
+    if (valid) {
+        if (!TBL(nativeTablePtr)->is_attached()) {
+            ThrowException(env, TableInvalid, "Table is closed, and no longer valid to operate on.");
```
cmelchior: This should be rephrased if it hits user space.

realm/realm-jni/src/io_realm_internal_table.cpp
```diff
@@ -1693,3 +1693,20 @@ JNIEXPORT jboolean JNICALL Java_io_realm_internal_Table_nativeHasSameSchema
...
+            return 0;
+        }
+    }
+    try {
```
cmelchior: Any reason this try/catch is outside the `if` statement. It would always fail if not valid?

realm/realm-jni/src/io_realm_internal_table.cpp
```diff
@@ -1693,3 +1693,20 @@ JNIEXPORT jboolean JNICALL Java_io_realm_internal_Table_nativeHasSameSchema
 {
     return *TBL(thisTablePtr)->get_descriptor() == *TBL(otherTablePtr)->get_descriptor();
 }
+
+
+JNIEXPORT jlong JNICALL Java_io_realm_internal_Table_nativeVersion(
+        JNIEnv* env, jobject, jlong nativeTablePtr)
+{
+    bool valid = (nativeTablePtr != 0);
```
kneth: It might be easier to read if you cast `nativeTablePtr` to a `Table*`. Then you can compare against `NULL`.

realm/realm-jni/src/io_realm_internal_table.cpp
kneth: Yes, if `nativeTablePtr == 0`, line 1709 will lead to a segmentation fault and the `try`/`catch` will probably give the user a strange error message. When is `nativeTablePtr == 0`?

realm/realm-library/src/androidTest/java/io/realm/NotificationsTest.java
```diff
@@ -690,17 +690,173 @@ public void onChange() {
...
-        awaitOrThrow(numberOfInvocation);
+        TestHelper.awaitOrFail(numberOfInvocation);
         mainThread.quit();
     }
 
-    private void awaitOrThrow(CountDownLatch latch) {
-        try {
-            if (!latch.await(5, TimeUnit.SECONDS)) {
-                fail("Latch timed out " + latch);
+    public void testAsyncRealmResultsShouldNotBlockBackgroundCommitNotification() throws Throwable {
```
cmelchior: I could use some more comments in this test. The title is good enough, but the flow is really hard to follow

realm/realm-library/src/androidTest/java/io/realm/NotificationsTest.java
```diff
@@ -690,17 +690,173 @@ public void onChange() {
...
+                                    });
+
+                                    new Thread() {
+                                        @Override
+                                        public void run() {
+                                            Realm realm = Realm.getInstance(getContext());
+                                            realm.beginTransaction();
+                                            realm.createObject(Dog.class);
+                                            realm.commitTransaction();
+                                            realm.close();
```
cmelchior: You risk leaking this realm before the next test starts as far as I can tell?

realm/realm-library/src/androidTest/java/io/realm/NotificationsTest.java
```diff
@@ -690,17 +690,173 @@ public void onChange() {
...
             }
-        } catch (InterruptedException e) {
-            fail("Latch was interrupted " + e);
-        }
+        });
+
+        TestHelper.exitOrThrow(executorService, signalTestFinished, signalClosedRealm, backgroundLooper, threadAssertionError);
+    }
+
+    public void testAsyncRealmObjectShouldNotBlockBackgroundCommitNotification() throws Throwable {
```
cmelchior: Same as above, missing comments about flow

realm/realm-library/src/androidTest/java/io/realm/TypeBasedNotificationsTest.java
```diff
@@ -0,0 +1,1367 @@
...
+
+    // ********************************************************************************* //
+    // UC 1.
+    // Callback should be invoked after a relevant commit (one that should impact the
+    // query from which we obtained our RealmObject or RealmResults)
+    // ********************************************************************************* //
+    //FIXME failing because sync RealmObject doesn't have the initial currentVersion, so any first commit
+    // (even if unrelated) will trigger the notification since the row.getTable().sync() will be != -1
+    // UC 1 for Sync RealmObject
+    public void test_callback_with_relevant_commit_realmobject_sync() {
```
cmelchior: Different style than normally used for test method names, I is more readable for those long method names though.

realm/realm-library/src/androidTest/java/io/realm/TypeBasedNotificationsTest.java
```diff
@@ -0,0 +1,1367 @@
...
+            TestHelper.awaitOrFail(cleanup);
+        }
+    }
+
+    // ********************************************************************************* //
+    // UC 1.
+    // Callback should be invoked after a relevant commit (one that should impact the
+    // query from which we obtained our RealmObject or RealmResults)
+    // ********************************************************************************* //
+    //FIXME failing because sync RealmObject doesn't have the initial currentVersion, so any first commit
```
cmelchior: The unit test isn't failing. So isn't this more like a TODO that hopefully will be fixed with proper fine-grained support?

realm/realm-library/src/androidTest/java/io/realm/TypeBasedNotificationsTest.java
```diff
@@ -0,0 +1,1367 @@
...
+
+import io.realm.entities.Dog;
+import io.realm.proxy.HandlerProxy;
+
+public class TypeBasedNotificationsTest extends AndroidTestCase {
+    private HandlerThread handlerThread;
+    private Handler handler;
+    private CountDownLatch signalTestFinished;
+    private AtomicInteger globalCommitInvocations;
+    private AtomicInteger fineGrainedCommitInvocations;
```
cmelchior: perhaps `typebasedCommitInvocations` instead?

realm/realm-library/src/androidTest/java/io/realm/TypeBasedNotificationsTest.java
```diff
@@ -0,0 +1,1367 @@
...
+        });
+        TestHelper.awaitOrFail(signalTestFinished);
+        assertEquals(2, fineGrainedCommitInvocations.get());
+    }
+
+    // ********************************************************************************* //
+    // UC 2.
+    // Multiple callbacks should be invoked after a relevant commit
+    // ********************************************************************************* //
+    //FIXME no initial value of currentVersion (getting double callback inovcation)
```
cmelchior: Is this true?

realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -302,6 +304,16 @@ public void refresh() {
         }
         sharedGroupManager.advanceRead();
         sendNotifications();
+        // notify RealmResults & RealmObject callbacks
+        if (handlerController != null) {
+            handlerController.notifyAsyncRealmResultsCallbacks();
```
cmelchior: Any reason all this logic isn't behind one method in handlerController?

realm/realm-library/src/main/java/io/realm/HandlerController.java
```diff
@@ -54,20 +58,92 @@
...
+    public boolean handleMessage(Message message) {
+        // Due to how a ConcurrentHashMap iterator is created we cannot be sure that other threads are
+        // aware when this threads handler is removed before they send messages to it. We don't wish to synchronize
+        // access to the handlers as they are the prime mean of notifying about updates. Instead we make sure
+        // that if a message does slip though (however unlikely), it will not try to update a SharedGroup that no
+        // longer exists. `sharedGroupManager` will only be null if a Realm is really closed.
+        if (realm.sharedGroupManager != null) {
+            switch (message.what) {
+                case REALM_CHANGED: {
+                    deleteWeakReferences();
```
cmelchior: I would move all this logic into it's own method just like the other switch statements

realm/realm-library/src/main/java/io/realm/HandlerController.java
```diff
@@ -54,20 +58,92 @@
...
+                        // empty async RealmObject shouldn' block the realm to advance
+                        // they're empty so no risk on running into a corrupt state
+                        // where the pointer (Row) is using one version of a Realm, whereas the
+                        // current Realm is advancing to a newer version (they're empty anyway)
+                        // as opposed to TableView, where even empty they still attached to a specific
+                        // version of a Realm (7 therefor should block advancing the Realm until they're updated)
+                        if (threadContainsAsyncEmptyRealmObject()) {
+                            // fire another QueryUpdateTask to re-run any pending empty
+                            // Async RealmObject (findFirstAsync returning an empty ROW)
+                            // we need to keep updating until
```
cmelchior: Sentence not finished

realm/realm-library/src/main/java/io/realm/HandlerController.java
```diff
@@ -54,20 +58,92 @@
...
+                        realm.sendNotifications();
+                        // notify RealmResults & RealmObject callbacks (type based notifications)
+                        notifySyncRealmResultsCallbacks();
+                        notifyRealmObjectCallbacks();
+
+                        // empty async RealmObject shouldn' block the realm to advance
+                        // they're empty so no risk on running into a corrupt state
+                        // where the pointer (Row) is using one version of a Realm, whereas the
+                        // current Realm is advancing to a newer version (they're empty anyway)
+                        // as opposed to TableView, where even empty they still attached to a specific
```
cmelchior: rephrase this

realm/realm-library/src/main/java/io/realm/HandlerController.java
```diff
@@ -345,16 +408,78 @@ private boolean threadContainsAsyncQueries () {
...
     }
 
+    boolean threadContainsAsyncEmptyRealmObject() {
+        boolean isEmpty = true;
+        Iterator<Map.Entry<WeakReference<RealmObject>, RealmQuery<?>>> iterator = emptyAsyncRealmObject.entrySet().iterator();
+        while (iterator.hasNext()) {
+            Map.Entry<WeakReference<RealmObject>, RealmQuery<?>> next = iterator.next();
+            if (next.getKey().get() == null) {
+                // clean the GC'ed instances
+                // we could've avoided this if we had a 'WeakIdentityHashmap' data structure. miss Guava :(
```
cmelchior: _pat_ :)

realm/realm-library/src/main/java/io/realm/HandlerController.java
```diff
@@ -345,16 +408,78 @@ private boolean threadContainsAsyncQueries () {
...
+                if (realmObject.row.isAttached()) {
+                    realmObject.notifyChangeListeners();
+                } else if (realmObject.row != Row.EMPTY_ROW) {
+                    iterator.remove();
+                }
+            }
+        }
+    }
+
+    private void  notifyRealmResults(Iterator<WeakReference<RealmResults<? extends RealmObject>>> iterator) {
```
cmelchior: double whitespace

realm/realm-library/src/main/java/io/realm/HandlerController.java
```diff
@@ -345,16 +408,78 @@ private boolean threadContainsAsyncQueries () {
...
+
+    void notifyAsyncRealmResultsCallbacks() {
+        notifyRealmResults(asyncRealmResults.keySet().iterator());
+    }
+
+    void notifySyncRealmResultsCallbacks() {
+        notifyRealmResults(syncRealmResults.keySet().iterator());
+    }
+
+    void notifyRealmObjectCallbacks() {
```
cmelchior: Shouldn't this either be `notifyRealmObjects` or change `notifyRealmResults` to `notifyRealmResultCallbacks` for consistent naming?

realm/realm-library/src/main/java/io/realm/RealmObject.java
```diff
@@ -119,6 +119,10 @@ public final boolean isValid() {
         return row != null && row.isAttached();
     }
 
+    protected Table getTable () {
+        return realm.schema.getTable(getClass());
```
cmelchior: We could probably cache this in the RealmObject?

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -1239,9 +1236,17 @@ public long count() {
...
         // then update it once the query completes in the background.
-        RealmResults<E> realmResults = new RealmResults<E>(realm, query, clazz);
-        final WeakReference<RealmResults<?>> weakRealmResults = new WeakReference<RealmResults<?>>(realmResults, realm.getReferenceQueue());
-        realm.addAsyncRealmResults(weakRealmResults, this);
+        RealmResults<E> realmResults;
+        if (isDynamicQuery()) {
+            //noinspection unchecked
+            realmResults = (RealmResults<E>) RealmResults.createFromDynamicClass(realm, query, className);
+        } else {
+            realmResults = new RealmResults<E>(realm, query, clazz);
```
cmelchior: All constructors should be private and the factory methods should be used instead

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -1329,9 +1334,17 @@ public Long call() throws Exception {
...
```

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -1329,9 +1334,17 @@ public Long call() throws Exception {
...
+        }
+
+        final WeakReference<RealmResults<?>> weakRealmResults = new WeakReference<RealmResults<?>>(realmResults,
```
cmelchior: RealmResults<? extends RealmObject

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -1435,9 +1448,17 @@ public Long call() throws Exception {
...
 
-        RealmResults<E> realmResults = new RealmResults<E>(realm, query, clazz);
-        final WeakReference<RealmResults<?>> weakRealmResults = new WeakReference<RealmResults<?>>(realmResults, realm.getReferenceQueue());
-        realm.addAsyncRealmResults(weakRealmResults, this);
+        RealmResults<E> realmResults;
+        if (isDynamicQuery()) {
+            //noinspection unchecked
+            realmResults = (RealmResults<E>) RealmResults.createFromDynamicClass(realm, query, className);
+        } else {
+            realmResults = new RealmResults<E>(realm, query, clazz);
```

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -1593,9 +1614,17 @@ private boolean isDynamicQuery() {
...
             // prepare the promise result
-            RealmResults<E> realmResults = new RealmResults<E>(realm, query, clazz);
-            final WeakReference<RealmResults<?>> weakRealmResults = new WeakReference<RealmResults<?>>(realmResults, realm.getReferenceQueue());
-            realm.addAsyncRealmResults(weakRealmResults, this);
+            RealmResults<E> realmResults;
+            if (isDynamicQuery()) {
+                //noinspection unchecked
+                realmResults = (RealmResults<E>) RealmResults.createFromDynamicClass(realm, query, className);
+            } else {
+                realmResults = new RealmResults<E>(realm, query, clazz);
```

realm/realm-library/src/main/java/io/realm/RealmResults.java
```diff
@@ -64,45 +65,51 @@
...
+    static RealmResults<DynamicRealmObject> createFromDynamicClass(BaseRealm realm, TableQuery query, String className) {
+        return new RealmResults<DynamicRealmObject>(realm, query, className);
     }
 
     static RealmResults<DynamicRealmObject> createFromDynamicTableOrView(BaseRealm realm, TableOrView table, String className) {
         return new RealmResults<DynamicRealmObject>(realm, table, className);
     }
 
-    private RealmResults(BaseRealm realm, Class<E> classSpec) {
+    RealmResults(BaseRealm realm, TableQuery query, Class<E> clazz) {
```
cmelchior: Use factory constructor instead. They where introduced because the no-name constructor was far to magical given the number of parameters provided. A named factory constructor make it it a lot easier to read the code elsewhere

realm/realm-library/src/main/java/io/realm/RealmResults.java
```diff
@@ -826,9 +828,16 @@ public void removeChangeListeners() {
      * Notifies all registered listeners.
      */
     void notifyChangeListeners() {
-        realm.checkIfValid();
```
cmelchior: This is checked before calling this method?

cmelchior:
Mind-melting to read 2000 lines of unit tests, but love the thoroughness of them. Generally looks good, mostly comments about style issues. I am a bit concerned about the complexity of the implementation though, but in general there is a lot of comments so that is good. Would it make sense to add a high-level description on how async work in general on the HandlerController (since that is the central class) akin to what was done for https://github.com/realm/realm-java/blob/master/realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmProcessor.java ?

realm/realm-jni/src/io_realm_internal_table.cpp
nhachicha: same in `Java_io_realm_internal_TableView_nativeSync` &  `view_valid_and_in_sync` what do you suggest ?

realm/realm-jni/src/io_realm_internal_table.cpp
nhachicha: Hmm, this is the same impl in `Java_io_realm_internal_TableView_nativeSync` 

`Table#createNative` i.e `Java_io_realm_internal_Table_createNative` also return 0 as a fallback, hence the check for 0 for the `nativeTablePtr`

realm/realm-library/src/androidTest/java/io/realm/TypeBasedNotificationsTest.java
nhachicha: using Camel case doesn't distinguish well sometimes between segments;
Ex `RealmResultsCommit` could be  understood  as
1. realm_resultsCommit
2. realmresults_Commit

realm/realm-library/src/androidTest/java/io/realm/TypeBasedNotificationsTest.java
nhachicha: yes it's dead comment 💀

realm/realm-jni/src/io_realm_internal_table.cpp
cmelchior: Hmmm, not sure.... `is_attached` returning false might also happen if the class was removed from the schema I would assume, not only closed, although that would probably be very rare".

Perhaps "The Realm has been closed and is no longer accessible"?

realm/realm-library/src/main/java/io/realm/RealmObject.java
nhachicha: this will be called only once inside `onCompleted(Long handoverRowPointer)` to create a valid Row

realm/realm-library/src/main/java/io/realm/RealmResults.java
nhachicha: `notifyChangeListeners` is always in a safe path (called internally) or preceded by a check  `realm.checkIfValid()`

cmelchior: :+1:

realm/realm-jni/src/io_realm_internal_table.cpp
```diff
@@ -1693,3 +1693,20 @@ JNIEXPORT jboolean JNICALL Java_io_realm_internal_Table_nativeHasSameSchema
...
 }
+
+
+JNIEXPORT jlong JNICALL Java_io_realm_internal_Table_nativeVersion(
+        JNIEnv* env, jobject, jlong nativeTablePtr)
+{
+    bool valid = (TBL(nativeTablePtr) != NULL);
+    if (valid) {
+        if (!TBL(nativeTablePtr)->is_attached()) {
+            ThrowException(env, TableInvalid, "The Realm has been closed and is no longer accessible");
```
zaki50: `.` at the end of message.

realm/realm-jni/src/io_realm_internal_tableview.cpp
```diff
@@ -28,10 +28,10 @@ using namespace realm;
 #define VIEW_VALID_AND_IN_SYNC(env, ptr) view_valid_and_in_sync(env, ptr)
 
 inline bool view_valid_and_in_sync(JNIEnv* env, jlong nativeViewPtr) {
-    bool valid = (nativeViewPtr != 0);
+    bool valid = (TV(nativeViewPtr) != NULL);
     if (valid) {
         if (!TV(nativeViewPtr)->is_attached()) {
-            ThrowException(env, TableInvalid, "Table is closed, and no longer valid to operate on.");
+            ThrowException(env, TableInvalid, "The Realm has been closed and is no longer accessible");
```
zaki50: `.` at the end of message.

realm/realm-jni/src/io_realm_internal_tableview.cpp
```diff
@@ -1003,10 +1003,10 @@ JNIEXPORT jlong JNICALL Java_io_realm_internal_TableView_nativeWhere(
 JNIEXPORT jlong JNICALL Java_io_realm_internal_TableView_nativeSync(
     JNIEnv* env, jobject, jlong nativeViewPtr)
 {
-    bool valid = (nativeViewPtr != 0);
+    bool valid = (TV(nativeViewPtr) != NULL);
     if (valid) {
         if (!TV(nativeViewPtr)->is_attached()) {
-            ThrowException(env, TableInvalid, "Table is closed, and no longer valid to operate on.");
+            ThrowException(env, TableInvalid, "The Realm has been closed and is no longer accessible");
```
zaki50: `.` at the end of message.

realm/realm-library/src/main/java/io/realm/RealmObject.java
cmelchior: :+1:

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -1837,20 +1810,24 @@ public Long call() throws Exception {
...
+                        sendMessageToHandler(weakHandler, HandlerController.COMPLETED_ASYNC_REALM_OBJECT, result);
 
-                        return handoverTableViewPointer;
+                        return handoverRowPointer;
 
                     } catch (Exception e) {
                         RealmLog.e(e.getMessage());
                         // handler can't throw a checked exception need to wrap it into unchecked Exception
-                        sendMessageToHandler(weakHandler, HandlerController.REALM_COMPLETED_ASYNC_FIND_FIRST, new Error(e));
+                        sendMessageToHandler(weakHandler, HandlerController.COMPLETED_ASYNC_REALM_OBJECT, new Error(e));
```
zaki50: Is that OK not to use `HandlerController.REALM_ASYNC_BACKGROUND_EXCEPTION`?

cmelchior:
Really like the refactoring of HandlerController. HandlerController could probably use an extended description of what is happening, but it doesn't have to be part of this PR. :+1:

realm/realm-library/src/main/java/io/realm/DynamicRealmObject.java
```diff
@@ -64,6 +65,11 @@ public DynamicRealmObject(RealmObject obj) {
         this.row = (row instanceof CheckedRow) ? (CheckedRow) row : ((UncheckedRow) row).convertToChecked();
     }
 
+    DynamicRealmObject(BaseRealm realm, String className) {
```
zaki50: we can remove the argument of `realm` since it will be assigned by RealmQuery.

realm/realm-library/src/main/java/io/realm/RealmObject.java
```diff
@@ -176,11 +180,16 @@ public final boolean load() {
     boolean onCompleted() {
         try {
             Long handoverResult = pendingQuery.get();// make the query blocking
-            // this may fail with BadVersionException if the caller and/or the worker thread
-            // are not in sync (same shared_group version).
-            // REALM_COMPLETED_ASYNC_FIND_FIRST will be fired by the worker thread
-            // this should handle more complex use cases like retry, ignore etc
-            onCompleted(handoverResult);
+            if (handoverResult != 0) { // no row found
```
zaki50: Is this comment right?

zaki50: In `HandlerController.completedAsyncRealmObject()`, `onCompleted(long)` and `notifyChangeListeners()` are always called even if `pointer == 0`.
Why the codes are different?

realm/realm-library/src/main/java/io/realm/DynamicRealmObject.java
```diff
@@ -711,4 +717,9 @@ public String toString() {
         sb.append("]");
         return sb.toString();
     }
+
+    @Override
+    protected Table getTable() {
+        return realm.schema.getTable(className);
```
zaki50: `getTable()` of `DynamicRealmObject` seems to work only if the instance is created by the constructor of `DynamicRealmObject(BaseRealm, String)`.

We should document this limitation or update to work with any constructor.

zaki50:
If a `DynamicRealmObject` is accessed before async query is complete, `getList()` seems to throw  NPE.
This should be fixed to return empty value instead.

zaki50:
Should we document that the dynamic async query and changing schema don't work at the same time?

realm/realm-library/src/androidTest/java/io/realm/DynamicRealmTest.java
```diff
@@ -415,4 +418,354 @@ public void testDistinctInvalidTypes() {
...
+                                bgRealm.close();
+                            }
+                        }.start();
+                    }
+                });
+            }
+        });
+
+        TestHelper.awaitOrFail(signalTestFinished);
+    }
```
zaki50: Shouldn't we call `handlerThread.quitSafely(); handlerThread.join()` to stop the thread before returning the test?

realm/realm-library/src/androidTest/java/io/realm/DynamicRealmTest.java
```diff
@@ -415,4 +418,354 @@ public void testDistinctInvalidTypes() {
...
+                                bgRealm.commitTransaction();
+                                bgRealm.close();
+                            }
+                        }.start();
+                    }
+                });
+            }
+        });
+
+        TestHelper.awaitOrFail(signalTestFinished);
```
zaki50: If timeout occur, the Realm in `handlerThread ` seems to be left open.
How about posting cleanup `Runnable` after `await` instead of closing Realm in the change listener.

```
try {
    TestHelper.awaitOrFail(signalTestFinished);
} finally {
    handler.post(new Runnable() {
        @Override
        public void run() {
            realmOnHandlerThread.close();
        }
    }
}
```

zaki50: Or, how about introducing following `DynamicRealmHandlerThread`

```
public class DynamicRealmHandlerThread extends HandlerThread {
    private RealmConfiguration config;
    private DynamicRealm realm;

    public DynamicRealmHandlerThread(String name, RealmConfiguration config) {
        super(name);
        this.config = config;
    }

    public DynamicRealmHandlerThread(String name, int priority, RealmConfiguration config) {
        super(name, priority);
        this.config = config;
    }

    @Override
    protected void onLooperPrepared() {
        super.onLooperPrepared();
        realm = DynamicRealm.getInstance(config);
    }

    protected DynamicRealm getRealm() {
        return realm;
    }

    @Override
    public void run() {
        super.run();
        realm.close();
    }
}
```

zaki50: Or, how about introducing following `DynamicRealmHandlerThread`

```
public static class DynamicRealmHandlerThread extends HandlerThread {
    private RealmConfiguration config;
    private DynamicRealm realm;

    public DynamicRealmHandlerThread(String name, RealmConfiguration config) {
        super(name);
        this.config = config;
    }

    public DynamicRealmHandlerThread(String name, int priority, RealmConfiguration config) {
        super(name, priority);
        this.config = config;
    }

    @Override
    protected void onLooperPrepared() {
        super.onLooperPrepared();
        realm = DynamicRealm.getInstance(config);
    }

    public DynamicRealm getRealm() {
        return realm;
    }

    @Override
    public void run() {
        super.run();

        realm.close();
    }
}
```

realm/realm-library/src/androidTest/java/io/realm/TypeBasedNotificationsTest.java
```diff
@@ -0,0 +1,1364 @@
...
+                    }
+                });
+            }
+        });
+        try {
+            TestHelper.awaitOrFail(signalTestFinished);
+            assertEquals(3, globalCommitInvocations.get());
+            assertEquals(2, typebasedCommitInvocations.get());
+        } catch (Throwable e) {
+            throw e;
```
zaki50: We don't need this catch clause.

realm/realm-library/src/main/java/io/realm/internal/IdentitySet.java
```diff
@@ -0,0 +1,30 @@
...
+import java.util.IdentityHashMap;
+
+/**
+ * Identity based Set, that guarantee store & retrieve in O(1)
+ * without a huge overhead in space complexity.
+ */
+public class IdentitySet<K> extends IdentityHashMap<K,Integer> {
+    private final static Integer PLACE_HOLDER = 0;
+
+    public void put(K key)  {
```
zaki50: use `add` instead of `put` for `Set`?

zaki50:
@nhachicha I reviewed and put some questions.

realm/realm-library/src/main/java/io/realm/RealmQuery.java
nhachicha: nice catch 👍

realm/realm-library/src/main/java/io/realm/RealmObject.java
nhachicha: we need to set the internal flag `isCompleted` to `true` `onCompleted` & `notifyChangeListeners` will handle internally the state where the pointer `== 0`

realm/realm-library/src/androidTest/java/io/realm/DynamicRealmTest.java
nhachicha: the `handlerThread` is not shared/reused between tests it will be GC'ed

realm/realm-library/src/androidTest/java/io/realm/DynamicRealmTest.java
nhachicha: nice pattern will refactor in another PR

nhachicha:
@realm/java

cmelchior:
:+1:

zaki50:
:+1:

nhachicha:
squashed commits

PR #1818, opened 2015-11-23
Fix some test files 

dalinaum:
* Add omitted copyright headers.
* createAndTestFilename public -> private
* remove an unused variable, get_data
* [Use assertEquals(expected, actual, delta) to compare floating-point numbers. assertEquals(expected, actual) for floating-pointer numbers is deprecated in JUnit.](http://junit.org/apidocs/org/junit/Assert.html#assertEquals(double, double))


dalinaum:
It's a small patch. Please review this @realm/java

dalinaum:
retest this please

realm/realm-library/src/androidTest/java/io/realm/RealmTest.java
```diff
@@ -71,6 +71,7 @@
 public class RealmTest extends AndroidTestCase {
 
     protected final static int TEST_DATA_SIZE = 10;
+    protected final static double DELTA = 1e-15;
```
cmelchior: why not 0? Since I assume we want to test `get(set(x)) = x`?

dalinaum: I thought 1e-15 is enough big, but I agree that 0 is appropriate.

beeender:
:+1:

cmelchior:
:+1:

kneth:
:+1:

Issue #1618, opened 2015-10-19
Example using split APK


nhachicha:
#1585 introduce AAR format, it would be nice to add an example that showcase how we can use [split APK](http://tools.android.com/tech-docs/new-build-system/user-guide/apk-splits) to help reduce APK size

emanuelez:
We could just convert one of the existing examples to use the feature

nhachicha:
good idea

dalinaum:
#1834 resolved this. I'll close this.

PR #1834, opened 2015-11-25
Simple, naive split apks on gridview example

donnfelker:
Relates to #1618. 

I'm not sure how in depth you wanted to go with the split apks. We can consider this the jump off point. 

@cmelchior @emanuelez @nhachicha What would you like to see in detail?

### DIFF
--- a/examples/gridViewExample/build.gradle
+++ b/examples/gridViewExample/build.gradle
@@ -26,6 +26,16 @@ android {
     command {
         events 2000
     }
+
+    splits {
+        // Split apks on build target ABI, view all options for the splits here:
+        // http://tools.android.com/tech-docs/new-build-system/user-guide/apk-splits
+        abi {
+            enable true
+            reset()
+            include 'arm', 'arm-v7a', 'arm64', 'mips', 'x86', 'x86_64'
+        }
+    }
 }
 
 dependencies {
--- a/changelog.txt
+++ b/changelog.txt
@@ -3,6 +3,7 @@
  * Added RealmObjectSchema.getFieldType() (#1883).
  * Added unitTestExample to showcase unit and instrumentation tests. Examples include jUnit3, jUnit4, Espresso, Robolectric, and MPowermock usage with Realm (#1440).
  * Adding support for ISO8601 based dates for JSON import. If JSON dates are invalid a RealmException will be thrown (#1213).
+ * Adding APK splits to gridViewExample (#1834).
 
 0.86.1
  * Fixed a bug where RealmQuery.findFirst() returned a wrong result if the RealmQuery had been created from a RealmResults.where() (#1905).


cmelchior:
It don't have to be terrible complex, but should still give a sense of the possibilities. The example from the  tools website seems fairly good? Except we don't want the universal package. http://tools.android.com/tech-docs/new-build-system/user-guide/apk-splits

It would probably be enough just to demonstrate how to do separate packages in the example and then add a small section about the Splits to our website where we also enumerate all the architectures we support so people know what string options they have available.

donnfelker:
@cmelchior I updated the sample to use a simple ABI split and then issued another PR for the docs. Let me know what you think.

cmelchior:
restest this please

examples/gridViewExample/build.gradle
```diff
@@ -26,6 +26,14 @@ android {
     command {
         events 2000
     }
+
+    splits {
+        // Split apks on build target version
+        abi {
+            enable true
+            include 'arm', 'arm-v7a', 'arm64', 'mips', 'x86', 'x86_64'
```
zaki50: `reset()` needed before `include` since include list contains all ABIs by default.

examples/gridViewExample/build.gradle
```diff
@@ -26,6 +26,14 @@ android {
...
+        }
```
zaki50: how about adding `universalApk true`.
The default id `false`.

examples/gridViewExample/build.gradle
```diff
@@ -26,6 +26,14 @@ android {
     command {
         events 2000
     }
+
+    splits {
+        // Split apks on build target version
```
zaki50: "target version" -> "target ABI"

examples/gridViewExample/build.gradle
donnfelker: @cmelchior Stated above everything but the universal package. Thoughts?

examples/gridViewExample/build.gradle
donnfelker: Good catch. I read that wrong in the docs. Thanks!

donnfelker:
@zaki50 Check this out when you get a chance for an additional review. :)

cmelchior:
Looks good to me :+1:

examples/gridViewExample/build.gradle
cmelchior: The `universalApk` setting is a bit orthogonal to the idea of splitting the APK to reduce the size of the file. So I think it is fair to leave it out in the example. Maybe we should add a link to the API docs in the sample, as it is documented there?

zaki50:
Looks good!
:+1:

donnfelker:
@cmelchior added the comment! Merge at will.

PR #1841, opened 2015-11-26
Remove duplicated gradle wrapper.

dalinaum:
If we need this duplicated gradle wrapper, we can use `../gradlew` instead.


dalinaum:

emanuelez:
The realm folder is a standalone gradle project so it needs those files

dalinaum:
I didn't delete gradle files of realm. They are all gradle wrapper. Is `../gradlew` enough to use realm folder? Is there corner cases? I don't think every gradle projects need its own gradle wrapper.

zaki50:
We import the `realm` folder to Android Studio as a project root, and Android Studio should be able to find the gradle wrapper on the project root.

So we should keep files of gradle wrapper in the `realm` folder.

dalinaum:
I see.I will close this.

PR #1842, opened 2015-11-26
Fixed unit tests for avg

cmelchior:
@realm/java


realm/realm-library/src/androidTest/java/io/realm/RealmQueryTest.java
```diff
@@ -1106,7 +1105,7 @@ public void testAvgForPartialNullRows() {
         RealmQuery<NullTypes> query = testRealm.where(NullTypes.class);
 
         // TODO update expected values after core's fix
```
emanuelez: Remove the TODO

cmelchior: No, it is still the wrong value, it is 2.333 but should be 3.5

PR #1844, opened 2015-11-27
Deprecate RealmConfiguration.getSchemaMediator() from public API and add getRealmObjectClasses() instead

### LINKED ISSUE #1797: RealmConfiguration.getSchemaMediator is public
It seems we by mistake left `RealmConfiguration.getSchemaMediator` public.
We should revert that to package private.

That said, there might still be use-cases where people want to know the classes part of the RealmConfiguration, so we should expose something like

```
// Not sure what is best?
RealmConfiguration.getClasses()
RealmConfiguration.getSchemaClasses()
RealmConfiguration.getRealmClasses()
```

One example of use case here: 
https://github.com/realm/realm-java/issues/1796

zaki50:
* Removed `RealmConfiguration.getSchemaMediator()` which was left public by mistake
* Added `RealmConfiguration.getRealmClasses()` which returns the unmodifiable set of model classes that belong to the configuration.

I refactored for all implementations of `RealmProxyMediator.getModelClassed()` to return unmodifiable `Set`.

closes #1797

### DIFF
--- a/realm/realm-library/src/main/java/io/realm/RealmConfiguration.java
+++ b/realm/realm-library/src/main/java/io/realm/RealmConfiguration.java
@@ -118,8 +118,25 @@ public class RealmConfiguration {
         return durability;
     }
 
+    /**
+     * Returns the mediator instance of schema which is defined by this configuration.
+     * This method is left public by mistake and will be removed.
+     *
+     * @return the mediator of the schema.
+     * @deprecated use {@link #getRealmObjectClasses()} instead if you need to access to the set of model classes.
+     */
+    @Deprecated
     public RealmProxyMediator getSchemaMediator() {
         return schemaMediator;
+    }
+
+    /**
+     * Returns the unmodifiable {@link Set} of model classes that make up the schema for this Realm.
+     *
+     * @return unmodifiable {@link Set} of model classes.
+     */
+    public Set<Class<? extends RealmObject>> getRealmObjectClasses() {
+        return schemaMediator.getModelClasses();
     }
 
     public String getPath() {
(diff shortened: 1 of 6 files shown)


realm/realm-library/src/main/java/io/realm/RealmConfiguration.java
```diff
@@ -118,10 +118,19 @@ public boolean shouldDeleteRealmIfMigrationNeeded() {
         return durability;
     }
 
-    public RealmProxyMediator getSchemaMediator() {
+    RealmProxyMediator getSchemaMediator() {
```
cmelchior: To follow our normal deprecation policies this should be marked as deprecated first before removing the public modifier.

realm/realm-library/src/main/java/io/realm/RealmConfiguration.java
```diff
@@ -118,10 +118,19 @@ public boolean shouldDeleteRealmIfMigrationNeeded() {
...
         return schemaMediator;
     }
 
+    /**
+     * Returns the unmodifiable {@link Set} of model classes that belong to this configuration.
```
cmelchior: Perhaps "Returns the unmodifiable Set of model classes that make up the schema for this Realm?

realm/realm-library/src/main/java/io/realm/RealmConfiguration.java
```diff
@@ -118,10 +118,19 @@ public boolean shouldDeleteRealmIfMigrationNeeded() {
...
+     *
+     * @return unmodifiable {@link Set} of model classes.
+     */
+    public Set<Class<? extends RealmObject>> getRealmClasses() {
```
cmelchior: Given the rest of our terminology this should perhaps be `getRealmObjectClasses()` instead. What do you think?

realm/realm-library/src/main/java/io/realm/RealmConfiguration.java
zaki50: fixed.
I added a javadoc comment for this method. please review that. @cmelchior

realm/realm-library/src/main/java/io/realm/RealmConfiguration.java
zaki50: fixed

realm/realm-library/src/main/java/io/realm/RealmConfiguration.java
zaki50: Agree. `getRealmObjectClasses()` is a bit long but much clearer.

zaki50: done

cmelchior:
:+1:

beeender:
:+1:

zaki50:
squashed and merged.

PR #1849, opened 2015-11-27
Adds Realm.copyFromRealm()

### LINKED ISSUE #931: Feature request. Create standalone copy from a RealmObject / copyFromRealm
If I get object reference from realm, sometimes I want to have `raw` object (with all fields filled), not a reference, since reference may become invalid. It would be great if we will have 'out of the box' method like `getRawObject`.

E.g.
```java
Dog dog = realmDatabase.getDogById(1000).getRawObject();
```

cmelchior:
Fixes #931 

Adds 4 new methods that makes it possible to make detached deep copies of managed RealmObjects:
```
RealmObject objectCopy = Realm.copyFromRealm(RealmObject object);
RealmObject objectCopy = Realm.copyFromRealm(RealmObject object, int maxDepth);
List<RealmObject> listCopy = Realm.copyFromRealm(Iterable<RealmObject> objects);
List<RealmObject> listCopy = Realm.copyFromRealm(Iterable<RealmObject> objects, int maxDepth);
```

As discussed this breaks Realms zero-copy and consistency guarantees, but enables architectural patterns currently not supported by Realm.


cmelchior:
@realm/java

Saketme:
Interesting! How does maxDepth work here?

realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmProxyClassGenerator.java
```diff
@@ -31,8 +31,11 @@
 import javax.lang.model.element.VariableElement;
 import javax.lang.model.type.DeclaredType;
 import javax.lang.model.util.Types;
+import javax.rmi.CORBA.Util;
```
zaki50: seems unused

cmelchior:
It is there to prevent copying to much data. Realm is an Object database so you might construct your data model in such a way that you can access the entire graph. This could be potentially very memory intensive to copy, so being able to break the graph makes this scenario to control so you don't run out of memory.

See e.g this unit test: https://github.com/realm/realm-java/pull/1849/files#diff-b11f4583c757aa2c5f52d4fb9e4ac750R2348

realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmProxyClassGenerator.java
cmelchior: Ups, will fix

Saketme:
Looks good. Thank you! I've been cloning into raw-objects manually in my code.

zaki50:
Current implementation searches and duplicates the object graph by depth-first search.
This causes reconstruction of sub graph if the same object appears more than once and depth is smaller than before.

To avoid this reconstruction, we can use the breadth-first search since the depth monotonically increases.

This is just an optimization and might not be needed at first release.

cmelchior:
Yes, it is depth first currently. Mostly because it was easier to implement and also because I think this feature will mostly be used on leaf nodes, which means we won't have to repopulate objects often (pure speculation though). 

So as far as I can see we have the following trade-offs:

Breath-first
* + Don't have to repopulate cached objects that are encountered at lower depths
* - Will always have to maintain an extra queue that consists of a tuple of `(realmObject,  standalone)` for all references at each depth.

Depth-first:
* + No need for extra datastrutucures = less memory pressure.
* - Will have to repopulate objects encountered at lower depths.

Given that I wouldn't anticipate this being used on either many or deep objects i opted for for simplicity, but I'm happy to hear arguments for changing it?

changelog.txt
```diff
@@ -1,3 +1,6 @@
+0.87.0
+ * Added Realm.copyFromRealm() for creating detached copies of Realm data.
```
kneth: "data" -> "objects"

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -746,6 +742,81 @@ void remove(Class<? extends RealmObject> clazz, long objectIndex) {
         return realmObjects;
     }
 
+    /**
+     * Makes a standalone in-memory copy of already persisted RealmObjects. This is a deep copy that will copy all
+     * referenced objects.
+     *
+     * The copied objects are all detached from Realm so they will no longer be automatically updated.
```
kneth: I think a warning of potential inconsistencies should be stated too.

cmelchior: Good idea, i'll add it

realm/realm-annotations-processor/src/test/resources/io/realm/AllTypesRealmProxy.java
```diff
@@ -668,6 +668,52 @@ public static AllTypes copy(Realm realm, AllTypes newObject, boolean update, Map
...
+        standaloneObject.setColumnBoolean(realmObject.isColumnBoolean());
+        standaloneObject.setColumnDate(realmObject.getColumnDate());
+        standaloneObject.setColumnBinary(realmObject.getColumnBinary());
+
+        // Deep copy of columnObject
+        standaloneObject.setColumnObject(AllTypesRealmProxy.createDetachedCopy(realmObject.getColumnObject(), currentDepth + 1, maxDepth, cache));
+
+        // Deep copy of columnRealmList
+        if (currentDepth == maxDepth) {
+            standaloneObject.setColumnRealmList(null);
```
beeender: This will be a little bit inconsistency since the `RealmList` of `RealmObjectProxy` cannot be null. But I don't know if there is anything better we can do except making the `RealmList` nullable.

cmelchior: Yes, but it would be more inconsistent to let it have the length 0 right? Because that might also be the _real_ value. By letting it be null is much more explicit that we reached max depth IMO.

beeender: Right, :+1:

realm/realm-annotations-processor/src/test/resources/io/realm/NullTypesRealmProxy.java
```diff
@@ -1167,6 +1167,50 @@ public static NullTypes copy(Realm realm, NullTypes newObject, boolean update, M
         return realmObject;
     }
 
+    public static NullTypes createDetachedCopy(NullTypes realmObject, int currentDepth, int maxDepth, Map<RealmObject, CacheData<RealmObject>> cache) {
+        if (currentDepth > maxDepth || realmObject == null) {
```
beeender: Integer overflow might happen if the `maxDapth == Integer.MAX_VALUE` . But not in the real life :)

beeender:
Nice! :+1:

realm/realm-annotations-processor/src/test/resources/io/realm/NullTypesRealmProxy.java
cmelchior: I'm pretty sure you would run out of Java heap space first :)

zaki50:
Depth-first is OK as first release. We should keep simple if possible.
And if we will find the use-case which requires Breath-first, we can introduce Breath-first implementation at any time.

:+1:

kneth:
Even though I do have reservations about diverging from the zero-copy principle I think it looks good.

Saketme:
Is it possible to start using 0.87 so that I can use this feature?

cmelchior:
Yes, you can use our SNAPSHOT releases : https://github.com/realm/realm-java#using-snapshots

Saketme:
Cool, thanks!

CDRussell:
Hi, I'm specifying my Realm version as `0.87.0-SNAPSHOT` but no sign of `Realm.copyFromRealm()`... any thoughts?

intrigua:
I wonder why this functionality maintained only in Java. Just run into the situation where it would be nice to have it in Swift.

cmelchior:
The Android and Swift communities have different paradigms, patterns and libraries that are commonly used. So while we strive to have feature parity between Java / Swift / Objective-C there will always be small differences.

If it is a something that is important to you I would suggest raising it as an issue for Realm Swift. They already have an issue tracking it here: https://github.com/realm/realm-cocoa/issues/3381

intrigua:
Thanks, I think I will find a simple workaround this time.

zhihuitang:
thanks @cmelchior, nice job

othreecodes:
Awesome!!!! thank you for this!

PR #1850, opened 2015-11-27
Delete an unnecessay target `get-inst-libraries`

dalinaum:


dalinaum:

realm/realm-jni/src/Makefile
```diff
@@ -3,7 +3,7 @@ lib_LIBRARIES = librealm-jni.a
 JNI_SOURCES := $(wildcard *.cpp)
 librealm_jni_a_SOURCES = $(JNI_SOURCES)
 
-# Used by ../../build.sh
+# Used by ../build.gradle
 get-inst-libraries:
```
zaki50: Is this target really used? It seems to be unused.
What do you think? @beeender

dalinaum: I cannot find any dependency with target `get-inst-libraries` too. Should we delete this target? @realm/java

dalinaum: I found out `calling get-inst-libraries` in delete `build.sh` file. I think it should be deleted too.

dalinaum:
@zaki50 I think you are right. I deleted target  `get-inst-libraries` instead just changing an outdated comment.

emanuelez:
Is this makefile used at all?

zaki50:
@emanuelez I think it is used in `realm-jni/build.gradle`. That gradle file executes following external command.

```
make -C "${projectDir}/src" ...
```

dalinaum:
@emanuelez I am not sure I quite understood your question. This Makefile include `../generic.mk`. `../build.gradle` uses targets defined in generic.mk.

zaki50:
I think deleting `get-inst-libraries` is :+1: 
And also think that we should get another :+1: by @beeender

beeender:
:+1:

emanuelez:
Please be aware that both generik.mk and the Makefile were copies from core so I'm not sure it's worth to modify them.

kneth:
I don't mind that we clean up the build files in general, but I agree with @emanuelez that it might not be worth it.

dalinaum:
@emanuelez @kneth Although `generic.mk` file is imported from core, `Makefile` is not similar to any Makefiles of core. I think that at least our Makefile is for only Realm-java now and `get-inst-libraries` is orinallay added for Realm-java.

beeender:
Makefile is copied from core as well, src/realm/Makefile. But that file changes a lot in core, so they don't look similar anymore :)  I think it is not harmful to remove this useless line :)

kneth:
:+1:

PR #1866, opened 2015-12-01
compact returns false if the file is opened

beeender:
Core will guarantee the file is not opened while compacting.
Core also guarantee opening Realm will be blocked until compacting
finished.


beeender:
@realm/java

cmelchior:
:+1:

zaki50:
:+1:

PR #1870, opened 2015-12-02
RxJava example

cmelchior:
Example split into its own PR from #1710


cmelchior:
This needs to be merged against the RxJava feature branch instead. Closing

PR #1872, opened 2015-12-02
log files are not deleted by deleteRealm

beeender:


beeender:
@realm/java

realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -555,7 +555,7 @@ public RealmSchema getSchema() {
     /**
      * Deletes the Realm file defined by the given configuration.
      */
-    protected static boolean deleteRealm(final RealmConfiguration configuration) {
+    static boolean deleteRealm(final RealmConfiguration configuration) {
```
cmelchior: I would prefer this as protected. It should only be called from sub-classes.

cmelchior:
Apart from small comment :+1:

realm/realm-library/src/main/java/io/realm/BaseRealm.java
beeender: `no modifier` is more strict i think? Even user extend the class, it still cannot be called.

cmelchior: Ah yes, sorry. Ignore me :) https://docs.oracle.com/javase/tutorial/java/javaOO/accesscontrol.html

zaki50:
:+1:

Issue #1851, opened 2015-11-28
Native crash in TableQuery.cpp line 1543


raminbbt:
im currently using latest snapshot version. i also used latest release version but same error.
```
D/REALM﹕ REALM_CHANGED realm:io.realm.HandlerController@41f9e5a0 no async queries, advance_read
D/REALM﹕ REALM_CHANGED realm:io.realm.HandlerController@41bc2690 updating async queries, total: 3
D/REALM﹕ REALM_UPDATE_ASYNC_QUERIES realm:io.realm.HandlerController@41bc2690 caller is behind  advance_read
D/REALM﹕ REALM_UPDATE_ASYNC_QUERIES realm:io.realm.HandlerController@41bc2690 updating RealmResults java.lang.ref.WeakReference@41bc9e88
D/REALM﹕ jni: ThrowingException 11, std::exception in io_realm_internal_TableQuery.cpp line 1543, .
D/REALM﹕ Exception has been throw: Unrecoverable error. std::exception in io_realm_internal_TableQuery.cpp line 1543
D/AndroidRuntime﹕ Shutting down VM
W/dalvikvm﹕ threadid=1: thread exiting with uncaught exception (group=0x4168ace0)
W/System.err﹕ io.realm.exceptions.RealmError: Unrecoverable error. std::exception in io_realm_internal_TableQuery.cpp line 1543
W/System.err﹕ at io.realm.internal.TableQuery.nativeImportHandoverTableViewIntoSharedGroup(Native Method)
W/System.err﹕ at io.realm.internal.TableQuery.importHandoverTableView(TableQuery.java:508)
W/System.err﹕ at io.realm.RealmResults.swapTableViewPointer(RealmResults.java:708)
W/System.err﹕ at io.realm.HandlerController.completedAsyncQueriesUpdate(HandlerController.java:222)
W/System.err﹕ at io.realm.HandlerController.handleMessage(HandlerController.java:312)
W/System.err﹕ at android.os.Handler.dispatchMessage(Handler.java:98)
W/System.err﹕ at android.os.Looper.loop(Looper.java:136)
W/System.err﹕ at android.app.ActivityThread.main(ActivityThread.java:5158)
W/System.err﹕ at java.lang.reflect.Method.invokeNative(Native Method)
W/System.err﹕ at java.lang.reflect.Method.invoke(Method.java:515)
[...]

PR #1878, opened 2015-12-03
Test case to reproduce #1851

beeender:
And throw the right BadVersion exception.

Working on identify the root cause. Push the PR to make it visible to others.


cmelchior:
Fixed by https://github.com/realm/realm-java/pull/1895

PR #1915, opened 2015-12-09
Call TableView::clear() with RemoveMode::unordered

zaki50:
Call TableView::clear() and TableView::remove() with RemoveMode::unordered.

This change requires  a new core release as unordered removal is broken without realm/realm-core#1384.

So this PR should be merged after #1913

### DIFF
--- a/realm/realm-jni/src/io_realm_internal_tableview.cpp
+++ b/realm/realm-jni/src/io_realm_internal_tableview.cpp
@@ -470,7 +470,7 @@ JNIEXPORT void JNICALL Java_io_realm_internal_TableView_nativeClear(
     try {
         if (!VIEW_VALID_AND_IN_SYNC(env, nativeViewPtr))
             return;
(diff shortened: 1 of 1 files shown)


zaki50:
I have no idea whether this change requires changelog entry.

zaki50:
related change in cocoa is realm/realm-cocoa#2954

cmelchior:
It is a purely internal change, so no changelog required IMO

cmelchior:
Was it intentional that this PR is against master while #1913 is against `release/0.86`? Otherwise :+1:

kneth:
I think that an entry in the change log would be nice: "Improved the performance of removing objects (RealmResults.clear() and RealmResults.remove()).

kneth:
Beside my comment above :+1:

zaki50:
Oh, not intentional..

I created #1917 and close this PR

PR #1917, opened 2015-12-09
Call TableView::clear() with RemoveMode::unordered

zaki50:
Call TableView::clear() and TableView::remove() with RemoveMode::unordered.

This change requires a new core release as unordered removal is broken without realm/realm-core#1384.

So this PR should be merged after #1913

I'll merge this since review has been done in #1915


zaki50:
Oh... I mistakenly merged into master...

cmelchior:
Not a big problem I would think. The change in core has been there since 0.93, but the changelog is now wrong. So either move the log entry to 0.87 or merge this into 0.86 as well.

zaki50:
I merged this to release/0.86

PR #1928, opened 2015-12-11
Unit test: Comparing with negative numbers triggered ARM bug

### LINKED ISSUE #1773: lessThan with negative id's not working as expected
I am working with RealmObjects where i use negative id's (as long as they are not synced to the server).
So, when i synchronize with the server, i can get all negative id's, synchronize those objects and when synchronized, update the id.

But, when i perform a lessThan query on a RealmResult, there are less objects returned than expected.
When i use between (with f.e. values -10000, 0), it is working and i am getting all the objects returned that i expect.

So it looks like the behaviour of lessThan with negative number is not working the way i would expect it to be, Is this a bug or is there a good explanation for this?

I created a small android studio project, with a TestApplication where this test is setup, which fails (only 2 of the 10 objects are retreived). Project is on github: https://github.com/Sein-IT/RealmLessThan.

I am testing with a Samsung Galaxy S6, Android version 5.1.1

kneth:
Fixes #1773.

This was fixed by a recent upgrade of the storage engine. I have tested `master` branch with the following devices: OnePlus One (Krait 400, ARMv7-a), Nexus 6 (Krait 450, ARMv7-a).

@realm/java


cmelchior:
:+1:

dalinaum:
I tested this on my Nexus 5X correctly. :+1:

Issue #1503, opened 2015-09-28
Document Android Framework class behavior


cmelchior:
We should document how certain Android framework classes behave in the context of Realm:

* IntentService: `onHandleIntent` is running in a separate Thread. Disable autoRefresh.
* AsyncTask: `doInBackground` is running in a separate Thread. 
* Others?

slidenerd:
IntentService, AsyncTask, JobSchedulerService, BroadcastReceiver [both local and remote]

kneth:
Thanks for the suggestions.

donnfelker:
@cmelchior chan this be closed? See the merge above.

cmelchior:
Yup. Thanks

PR #1946, opened 2015-12-16
Adding example that shows how to pass objects via ids

donnfelker:
Examples that help illustrate common uses cases that relate to  #1503

### DIFF
--- a/examples/threadExample/src/main/java/io/realm/examples/threads/PassingObjectsFragment.java
+++ b/examples/threadExample/src/main/java/io/realm/examples/threads/PassingObjectsFragment.java
@@ -1,0 +1,118 @@
+/*
+ * Copyright 2015 Realm Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.realm.examples.threads;
+
+
+import android.content.Intent;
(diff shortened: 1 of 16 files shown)


donnfelker:
@cmelchior Can you take a look when you get a chance?

examples/threadExample/src/main/java/io/realm/examples/threads/PassingObjectsFragment.java
```diff
@@ -0,0 +1,99 @@
...
+
+    @Nullable
+    @Override
+    public View onCreateView(LayoutInflater inflater, ViewGroup container, Bundle savedInstanceState) {
+        View view = inflater.inflate(R.layout.fragment_passing_objects, container, false);
+        textContent = (TextView) view.findViewById(R.id.text_content);
+        view.findViewById(R.id.start_activity).setOnClickListener(new View.OnClickListener() {
+            @Override
+            public void onClick(View v) {
+                Intent intent = new Intent(getActivity(), ReceivingActivity.class);
```
cmelchior: I would put a comment above this to explain the pattern. One time is probably enough instead of duplicating it across 3 click listeners. Alternative it could be a class comment?

cmelchior:
I think this looks good. A few comments in the code about the pattern used would make it more clear, otherwise :+1:

cmelchior:
@realm/java

cmelchior:
PS: Failed unit test is CI borking on a download checksum, rerunning the tests should fix it.

bmunkholm:
In example code it's very useful to be generous with comments explaining what we are really trying to show.

kneth:
:+1:

examples/threadExample/src/main/java/io/realm/examples/threads/PassingObjectsFragment.java
donnfelker: Extracted them and placed the comment onto the method. Let me know what you think.

donnfelker:
I extracted the click listeners into their own method and then annotated the method with proper docs as to why the code is being formed the way that it is.

donnfelker:
@cmelchior @kneth Changes were made, let me know what you guys think

cmelchior:
:+1:

kneth:
:+1:

PR #1958, opened 2015-12-17
Unit test to match #1581

kneth:


LiuDeng:
@kneth 
i still get this problem in io.realm:realm-android:0.87.4 if i do not use compactRealm the realm file increase fast
when will you fix this！

kneth:
Fast increasing Realm files is often seen in situations where you have an open Realm instance in one thread, and a long write transaction in another thread.

PR #1962, opened 2015-12-18
Throw exception when try to remove a changeListener from a non-looper thread.

hohnamkung:
Removing changeListener from realm that is created form non-looper thread occurs NPE.

    java.lang.NullPointerException: Attempt to invoke virtual method 'void io.realm.HandlerController.removeAllChangeListeners()' on a null object reference
        at io.realm.BaseRealm.removeAllChangeListeners(BaseRealm.java:171)
        at io.realm.Realm.removeAllChangeListeners(Realm.java:107)

So as like a `addChangeListener`, it's better to throw `IllegalStateException` when called from non-looper thread.

### DIFF
--- a/realm/realm-library/src/androidTest/java/io/realm/RealmTest.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/RealmTest.java
@@ -2697,16 +2697,20 @@ public class RealmTest extends AndroidTestCase {
         }
     }
 
-    public void testHandlerControllerThrowNullPointExceptionOnNonLooperThread() {
+    public void testRemoveChangeListenerThrowExceptionOnNonLooperThread() {
         final CountDownLatch signalTestFinished = new CountDownLatch(1);
         Thread thread = new Thread(new Runnable() {
             @Override
             public void run() {
                 Realm realm = Realm.getInstance(testConfig);
                 try {
-                    realm.handlerController.removeAllChangeListeners();
+                    realm.removeChangeListener(new RealmChangeListener() {
+                        @Override
+                        public void onChange() {
+                        }
+                    });
                     fail("");
-                } catch (NullPointerException ignored) {
+                } catch (IllegalStateException e) {
                     signalTestFinished.countDown();
                 } finally {
                     realm.close();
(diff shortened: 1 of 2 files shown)


kneth:
@hohnamkung Can you create a unit test which exposed the NPE issue?

hohnamkung:
Did you mean unit test for NPE issue or unit test for fixing?

kneth:
@hohnamkung A unit test to show under which conditions you can get NPE. Thanks in advance.

hohnamkung:
I add test cases for NPE condition and fixing one.

realm/realm-library/src/androidTest/java/io/realm/RealmTest.java
```diff
@@ -2696,4 +2696,54 @@ public void execute(Realm realm) {
...
+    public void testHandlerControllerThrowNullPointExceptionOnNonLooperThread() {
+        final CountDownLatch signalTestFinished = new CountDownLatch(1);
+        Thread thread = new Thread(new Runnable() {
+            @Override
+            public void run() {
+                Realm realm = Realm.getInstance(testConfig);
+                try {
+                    realm.handlerController.removeAllChangeListeners();
+                    fail("");
+                } catch (NullPointerException ignored) {
```
beeender: Should it be IllegalStateException？

beeender: Uh. my bad. This case can be simpler by just checking if `realm.handlerController` is null.

realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -175,11 +179,15 @@ public void removeChangeListener(RealmChangeListener listener) {
...
      * Removes all user-defined change listeners.
      *
+     * @throws IllegalStateException if you try to remove listeners from a non-Looper Thread.
      * @see io.realm.RealmChangeListener
      * @see #addChangeListener(RealmChangeListener)
      */
     public void removeAllChangeListeners() {
         checkIfValid();
+        if (handler == null) {
+            throw new IllegalStateException("You can't register listeners from a non-Looper thread ");
```
nhachicha: You can't remove

realm/realm-library/src/androidTest/java/io/realm/RealmTest.java
nhachicha: I think this UT is already covered by the test below. The path we need to test is calling `removeAllChangeListeners` on a non-looper thread should throw. Maybe refactor this test to cover  call to `removeChangeListener` instead?

hohnamkung:
Thanks for comments. I've updated UT and wrong exception description.

nhachicha:
squashed & merged in #1989 thanks @hohnamkung

PR #1989, opened 2015-12-22
Hohnamkung/fix remove listener

nhachicha:
taking ownership of @hohnamkung [PR](https://github.com/realm/realm-java/pull/1962) to fix NDK & target `release/0.87`


nhachicha:
@realm/java

realm/realm-library/src/androidTest/java/io/realm/RealmTest.java
```diff
@@ -2696,4 +2696,58 @@ public void execute(Realm realm) {
...
+            @Override
+            public void run() {
+                Realm realm = Realm.getInstance(testConfig);
+                try {
+                    realm.removeChangeListener(new RealmChangeListener() {
+                        @Override
+                        public void onChange() {
+                        }
+                    });
+                    fail("");
```
kneth: No need for empty string.

realm/realm-library/src/androidTest/java/io/realm/RealmTest.java
```diff
@@ -2696,4 +2696,58 @@ public void execute(Realm realm) {
...
+
+    public void testRemoveAllChangeListenersThrowExceptionOnNonLooperThread() {
+        final CountDownLatch signalTestFinished = new CountDownLatch(1);
+        Thread thread = new Thread(new Runnable() {
+            @Override
+            public void run() {
+                Realm realm = Realm.getInstance(testConfig);
+                try {
+                    realm.removeAllChangeListeners();
+                    fail("");
```
kneth: No need for empty string.

realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -154,11 +154,15 @@ public void addChangeListener(RealmChangeListener listener) {
      * Removes the specified change listener.
      *
      * @param listener the change listener to be removed.
+     * @throws IllegalStateException if you try to remove a listener from a non-Looper Thread.
      * @see io.realm.RealmChangeListener
      * @see #addChangeListener(RealmChangeListener)
      */
     public void removeChangeListener(RealmChangeListener listener) {
         checkIfValid();
+        if (handler == null) {
```
cmelchior: I am still not sure this is really needed. At worst it would be a no-op since you aren't allowed to add it in the first place as far as I can tell?

nhachicha: Sure you can't add listeners, but you can still invoke `removeChangeListener` & `removeAllChangeListeners` on a non Looper thread. I think it's elegant to crash with an _IllegalStateException_ rather than a _NPE_

cmelchior: Ah, yes, I got confused by this checking `handler` but below `handlerController` is used. It looks like both variables are guarded by the same `if`-statement. It would be more readable if this checked `handlerController` instead.

Otherwise, I can see why it makes sense to guard against it.

nhachicha: good point, updated using `handlerController` for more consistency

realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -154,11 +154,15 @@ public void addChangeListener(RealmChangeListener listener) {
...
      *
      * @param listener the change listener to be removed.
+     * @throws IllegalStateException if you try to remove a listener from a non-Looper Thread.
      * @see io.realm.RealmChangeListener
      * @see #addChangeListener(RealmChangeListener)
      */
     public void removeChangeListener(RealmChangeListener listener) {
         checkIfValid();
+        if (handlerController == null) {
+            throw new IllegalStateException("You can't remove a listener from a non-Looper thread ");
```
beeender: Is `can't` formal English?

beeender:
:+1:

dalinaum:
:+1:

nhachicha:
squashed

PR #1994, opened 2015-12-23
Always close shared group first before async msg

beeender:


beeender:
@realm/java Quite difficult to make a test case of this ...

beeender:
back to wip. My mind got corrupted today. This doesn't seem to be necessary.

beeender:
Stupid me. This is not necessary.

beeender:
Well, i found a good reason to merge this.

We support in-memory Realm which means if all of the Realm instances are closed, the next time, same name Realm opened should be a totally new Realm.

But if there is an shared group opened in other thread, core won't treat it as a brand new Realm. Instead, all of the old data will be there.

beeender:
@realm/java

nhachicha:
:+1:

dalinaum:
:+1:

kneth:
:+1:

PR #2004, opened 2015-12-25
Removed explicit GC call when commit transaction

### LINKED ISSUE #1925: Realm is probably triggering off a lot of blocking GCs
I am seeing a lot of explicitly triggered blocking GCs in Logcat in my app when inserting new items to the database. 

    12-10 11:33:25.874 7555-7555/com.tattva.app I/art: Starting a blocking GC Explicit
    12-10 11:33:25.955 7555-7561/com.tattva.app W/art: Suspending all threads took: 47.502ms
    12-10 11:33:25.961 7555-7555/com.tattva.app I/art: Explicit concurrent mark sweep GC freed 54692(2MB) AllocSpace objects, 30(1512KB) LOS objects, 40% free, 18MB/30MB, paused 1.205ms total 79.991ms

Is this normal? I can clearly see my app stuttering whenever they're triggered.

Blocking GCs are incredibly bad. What's the reason for triggering them?

beeender:
Fix #1925
GC should not be forced to run

### DIFF
--- a/realm/realm-library/src/androidTest/java/io/realm/TypeBasedNotificationsTest.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/TypeBasedNotificationsTest.java
@@ -25,14 +25,18 @@ import org.json.JSONObject;
 
 import java.io.IOException;
 import java.io.InputStream;
+import java.lang.ref.WeakReference;
 import java.util.Date;
+import java.util.WeakHashMap;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 
 import io.realm.entities.AllTypes;
 import io.realm.entities.AllTypesPrimaryKey;
+import io.realm.entities.Cat;
 import io.realm.entities.Dog;
+import io.realm.entities.Owner;
 import io.realm.entities.PrimaryKeyAsLong;
 import io.realm.proxy.HandlerProxy;
 
@@ -1160,6 +1164,9 @@ public class TypeBasedNotificationsTest extends AndroidTestCase {
     // ********************************************************************************* //
 
     // UC 3 for Sync RealmObject
+    // 1. Add listener to RealmObject which is queried synchronized.
+    // 2. Commit transaction in another non-looper thread
+    // 3. Listener on the RealmObject gets triggered.
     public void test_non_looper_thread_commit_realmobject_sync() {
         handler.post(new Runnable() {
             @Override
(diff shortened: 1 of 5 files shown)


beeender:
@realm/java

dalinaum:
:+1:

cmelchior:
Isn't that GC the while point of that method? /cc @nhachicha

emanuelez:
Yes, that was the point of the method

dalinaum:
I see. I should cancel :+1: then.

Saketme:
It makes the GC run everytime a transaction is executed asynchronously. I see lags in my app because the pause period is big in my phone. It always interferes with update animations in the app. :(

beeender:
I don't think put gc here is reasonable. It means every handler thread will call gc when Realm updated. And also, if we rely on the gc, most of times we are not doing the right thing ....

realm/realm-library/src/androidTest/java/io/realm/TypeBasedNotificationsTest.java
```diff
@@ -1168,7 +1171,7 @@ public void run() {
                 realm.addChangeListener(new RealmChangeListener() {
                     @Override
                     public void onChange() {
-                        if (globalCommitInvocations.incrementAndGet() == 2) {
+                        if (realm.where(Dog.class).count() == 2) {
```
beeender: The notification behaviour will be changed when multi-processes support merged. The notification won't be guaranteed to be triggered on every time the transaction commited, instead, it will be guaranteed to be triggered when the latest transaction committed.

Saketme:
@beeender Any updates on this? :)

diegomontoya:
Having GC() is just a signal to Android to GC() but never guaranteed. On Android 4 it actually GC, most of the time, but for for Android 5/6, ART will ignore most if not all of the GC() calls and do GC on it's on timetable. 

If realm has a real need to manually call GC() then a timed frequency cap can be used so sitatuation like this can be avoided. For example, at most trigger a GC once every 30 seconds, instead of 5000 GC() calls because someone just did 5000 realm updates within 1 second.

beeender:
@diegomontoya Thanks a lot for the comments! The `GC` call is majorly because of the async queries.

If references of `RealmResults` from async queries are kept, the async query will be re-run when the Realm changes to keep it updated. But if the all of the references to the `RealmResults` are freed, Realm has no way to know until it gets GCed. In that case, the async query will still be re-run which is wasting.

We are discussing about add some interface like `RealmResults.unload` to indicate that this async query is not needed anymore which will be abandoned and no longer updated.

IMO,  "trigger a GC once every 30 seconds" still has some problems since user loose the control of the timing. There is still chance that user could met the GC blocking the UI thread and can do nothing about that since GC is called in the lib.

cmelchior:
Following internal discussion: :+1:

dalinaum:
:+1:

PR #2037, opened 2016-01-06
Update Javadoc for Rx Observables to be more specific.

cmelchior:
Based on external feedback this updates the `asObservable()` Javadoc to be more specific about it's behavior.

@realm/java


kneth:
:+1:

dalinaum:
:+1:

beeender:
:+1: But when `first` called, the `onComplete` will called, right?

cmelchior:
Yes, `first()` will transform the observable, so it only emits the first item and then calls `onComplete`

PR #2048, opened 2016-01-08
RealmList.set() now correctly returns the old item.

### LINKED ISSUE #2044: RealmList.set(int location, E object) returns current object instead of previous.
Implementation of List interface requires from `set(int location, E object)` operation to return object that was previously set at given index. 
https://docs.oracle.com/javase/7/docs/api/java/util/List.html#set(int,%20E)

The implementation of `RealmList.set(int location, E object)` has a bug - it returns current object, instead of the previous one. 
It can be easily observed by runing `Collections.swap(list, index1, index2);` (it uses set operation internally). Instead of objects being swapped, the same object is at index1 and index2 .

cmelchior:
Fixes #2044 

Also migrated RealmList unit tests to Junit4.

@realm/java


dalinaum:
:+1:

donnfelker:
LGTM

PR #2043, opened 2016-01-07
Fix deploying source + javadoc as part of SNAPSHOT releases

cmelchior:
Fixes #1971

PR #2049, opened 2016-01-08
Fixed source and javadoc jars not being deployed correctly.

### LINKED ISSUE #1971: Snapshots: Source Packaging
Since Oct 29 v0.84.2 snapshot release, the source package is no longer part of the snapshot release process. It would be great if this condition can be reverted if not actually intended.

cmelchior:
Fixes #1971 

@realm/java


cmelchior:
Fast merge see #2043 for review comments

PR #2059, opened 2016-01-11
Fix a couple of Findbugs issues

emanuelez:
1) When implementing Comparable one should implement `equals`
2) A field seems to be unused, but it's actually used by the AP

### DIFF
--- a/realm/realm-library/src/main/java/io/realm/internal/SharedGroup.java
+++ b/realm/realm-library/src/main/java/io/realm/internal/SharedGroup.java
@@ -277,6 +277,28 @@ public class SharedGroup implements Closeable {
                     ", index=" + index +
                     '}';
         }
+
+        @Override
+        public boolean equals(Object object) {
+            if (this == object) return true;
+            if (object == null || getClass() != object.getClass()) return false;
+            if (!super.equals(object)) return false;
+
+            VersionID versionID = (VersionID) object;
+
+            if (version != versionID.version) return false;
+            if (index != versionID.index) return false;
+
+            return true;
+        }
+
+        @Override
+        public int hashCode() {
+            int result = super.hashCode();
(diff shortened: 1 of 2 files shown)


emanuelez:
@realm/java

realm/realm-library/src/main/java/io/realm/internal/SharedGroup.java
```diff
@@ -277,6 +277,28 @@ public String toString() {
...
+            if (this == object) return true;
+            if (object == null || getClass() != object.getClass()) return false;
+            if (!super.equals(object)) return false;
+
+            VersionID versionID = (VersionID) object;
+
+            if (version != versionID.version) return false;
+            if (index != versionID.index) return false;
+
+            return true;
```
cmelchior: I think that SharedGroup should also compare `getPath()` otherwise you risk having the same versionID across two files (highly unlikely, but still).

realm/realm-library/src/main/java/io/realm/internal/SharedGroup.java
```diff
@@ -277,6 +277,28 @@ public String toString() {
...
+        }
+
+        @Override
+        public int hashCode() {
+            int result = super.hashCode();
+            result = 31 * result + (int) (version ^ (version >>> 32));
+            result = 31 * result + (int) (index ^ (index >>> 32));
+            return result;
```
cmelchior: Same, use path info as well

realm/realm-library/src/main/java/io/realm/internal/SharedGroup.java
emanuelez: This is Sharedgroup$VersionID, not SharedGroup

cmelchior: Ah, then :+1:

PR #2072, opened 2016-01-13
Not throw always in TestRealmConfigurationFactory

beeender:
Only throw the exception caused by deleting the opened Realm if the test
case itself doesn't throw. Otherwise the test case cannot fail by the
original exception.

### DIFF
--- a/realm/realm-library/src/androidTest/java/io/realm/rule/TestRealmConfigurationFactory.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/rule/TestRealmConfigurationFactory.java
@@ -17,6 +17,8 @@
 package io.realm.rule;
 
 import org.junit.rules.TemporaryFolder;
+import org.junit.runner.Description;
+import org.junit.runners.model.Statement;
 
 import java.util.Collections;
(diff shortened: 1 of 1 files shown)


beeender:
@realm/java

beeender:
retest this please

realm/realm-library/src/androidTest/java/io/realm/rule/TestRealmConfigurationFactory.java
```diff
@@ -35,6 +37,25 @@
 public class TestRealmConfigurationFactory extends TemporaryFolder {
     private Map<RealmConfiguration, Boolean> map = new ConcurrentHashMap<RealmConfiguration, Boolean>();
     private Set<RealmConfiguration> configurations = Collections.newSetFromMap(map);
+    private boolean thrownBefore = false;
```
cmelchior: perhaps a better name for this would be `unitTestFailed` ?

cmelchior:
Except for a small comment :+1:

dalinaum:
:+1:

Issue #1568, opened 2015-10-09
Adding `distinct()` to RealmList


kneth:
For consistency, `RealmQuery.distinct()` and  `RealmResults.distinct()` should be added.

dalinaum:
https://github.com/realm/realm-core/issues/1064 is still open. Should we add the `Waiting-Core` label again? @kneth

kneth:
I believe it was added to core release 0.95.4 but I will try to get it confirmed.

kneth:
@dalinaum Confirmed that the core team forgot to close the issue. We can proceed and implement `distinct()` for `RealmResults`.

stk1m1:
I'll start looking into it.

kneth:
@stk1m1 Great. It will require some work on the JNI level to get it done. Don't hesitate to ask if you run into trouble.

stk1m1:
@kneth Thanks for head up!

cmelchior:
Right now `sort()` is part of the `OrderedRealmCollection` interface, we should consider adding this here as well. This would also mean it would work on `RealmList`

cmelchior:
@kneth Actually looking over our API we already have `RealmQuery.distinct()` and `RealmResults.distinct()`. So no entirely sure what https://github.com/realm/realm-java/issues/1568#issue-110661663 is about. It is pretty old, so perhaps the title/description just need an update?

Zhuinden:
@cmelchior this was added in 0.88.0 

```
RealmQuery.distinct() and RealmResults.distinct() (#1568).
RealmQuery.distinctAsync() and RealmResults.distinctAsync() (#2118).
```

beeender:
`RealmResults.distinct()` has been deprecated, since the API is quite ambiguous -- `distinct()` will always create a new `RealmResults` instead of distinct on the original one.

This means, if you want to do distinct on a `RealmResults`/`RealmList`, you should:
```java
realmResults.where().distinct();
realmList.where().distinct();
```

PR #2076, opened 2016-01-13
Adding `distinct()` to RealmQuery & RealmResults 

stk1m1:
[#1568](https://github.com/realm/realm-java/issues/1568)

- Only covers RealQuery.  
- JNI level code review required to double check whether usage of `TableView` is appropriate to generate `Query` object.  
- Fails tests with unexpected Boolean types.


stk1m1:
@cmelchior please review. I've put a comment on a specific line I'm wondering.

realm/realm-jni/src/io_realm_internal_TableQuery.cpp
```diff
@@ -1865,3 +1865,40 @@ JNIEXPORT void JNICALL Java_io_realm_internal_TableQuery_nativeIsEmpty
...
+    (JNIEnv *env, jobject, jlong nativeQueryPtr, jlongArray columnIndexes) {
+
+    JniLongArray arr(env, columnIndexes);
+    jlong column_idx = arr[arr.len() - 1];
+    TableRef table_ref = getTableForLinkQuery(nativeQueryPtr, arr);
+
+    Query* pQuery = Q(nativeQueryPtr);
+
+    try {
+
```
cmelchior: Remove whitespace

realm/realm-jni/src/io_realm_internal_TableQuery.cpp
```diff
@@ -1865,3 +1865,40 @@ JNIEXPORT void JNICALL Java_io_realm_internal_TableQuery_nativeIsEmpty
...
+        if (!QUERY_VALID(env, pQuery) || !TBL_AND_COL_INDEX_VALID(env, table_ref.get(), column_idx))
+            return;
+
```
cmelchior: remove whitespace

realm/realm-jni/src/io_realm_internal_TableQuery.cpp
```diff
@@ -1865,3 +1865,40 @@ JNIEXPORT void JNICALL Java_io_realm_internal_TableQuery_nativeIsEmpty
         }
     } CATCH_STD()
 }
+
+JNIEXPORT void JNICALL Java_io_realm_internal_TableQuery_nativeDistinct
+    (JNIEnv *env, jobject, jlong nativeQueryPtr, jlongArray columnIndexes) {
+
+    JniLongArray arr(env, columnIndexes);
+    jlong column_idx = arr[arr.len() - 1];
+    TableRef table_ref = getTableForLinkQuery(nativeQueryPtr, arr);
```
cmelchior: Do distinct() support link queries?

realm/realm-library/src/androidTest/java/io/realm/RealmQueryTest.java
```diff
@@ -1802,4 +1803,88 @@ public void execute(Realm realm) {
...
+
+        for (String field : new String[]{"columnRealmObject", "columnRealmList", "columnDouble", "columnFloat"}) {
+            try {
+                testRealm.where(AnnotationIndexTypes.class).distinct(field);
+                fail(field);
+            } catch (UnsupportedOperationException ignored) {
+            }
+        }
+    }
+
```
cmelchior: Missing test for link queries, i.e "columnRealmObject.columnString"

realm/realm-library/src/androidTest/java/io/realm/RealmQueryTest.java
```diff
@@ -1802,4 +1803,88 @@ public void execute(Realm realm) {
...
+    public void testDistinct() {
+        final long numberOfBlocks = 25;
+        final long numberOfObjects = 10; // must be greater than 1
+
+        populateForDistinct(testRealm, numberOfBlocks, numberOfObjects, false);
+
+        assertEquals(2, testRealm.where(AnnotationIndexTypes.class).distinct("indexBoolean").count());
+
```
cmelchior: Remove this whitespace, so all asserts are grouped together.

realm/realm-library/src/androidTest/java/io/realm/RealmQueryTest.java
```diff
@@ -1802,4 +1803,88 @@ public void execute(Realm realm) {
...
+            }
+        }
+        realm.commitTransaction();
+    }
+
+    public void testDistinct() {
+        final long numberOfBlocks = 25;
+        final long numberOfObjects = 10; // must be greater than 1
+
+        populateForDistinct(testRealm, numberOfBlocks, numberOfObjects, false);
```
cmelchior: I would remove this whitespace so all initializing is grouped together.

realm/realm-library/src/androidTest/java/io/realm/RealmQueryTest.java
```diff
@@ -1802,4 +1803,88 @@ public void execute(Realm realm) {
...
+
+        for (String fieldName : new String[]{"Long", "Date", "String"}) {
+            assertEquals(numberOfBlocks, testRealm.where(AnnotationIndexTypes.class).distinct("index" + fieldName).count());
+        }
+    }
+
+    public void testDistinctWithNull() {
+        final long numberOfBlocks = 25;
+        final long numberOfObjects = 10; // must be greater than 1
+
```
cmelchior: Same

realm/realm-library/src/androidTest/java/io/realm/RealmQueryTest.java
```diff
@@ -1802,4 +1803,88 @@ public void execute(Realm realm) {
...
+        for (String fieldName : new String[]{"Date", "String"}) {
+            assertEquals(1, testRealm.where(AnnotationIndexTypes.class).distinct("index" + fieldName).count());
+        }
+    }
+
+    public void testDistinctNotIndexedFields() {
+        final long numberOfBlocks = 25;
+        final long numberOfObjects = 10; // must be greater than 1
+
```
cmelchior: Same

realm/realm-library/src/androidTest/java/io/realm/RealmQueryTest.java
```diff
@@ -1802,4 +1803,88 @@ public void execute(Realm realm) {
...
+                fail("notIndex" + fieldName);
+            } catch (UnsupportedOperationException ignored) {
+            }
+        }
+    }
+
+    public void testDistinctDoesNotExist() {
+        final long numberOfBlocks = 25;
+        final long numberOfObjects = 10; // must be greater than 1
+
```
cmelchior: Same

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -1092,6 +1092,20 @@ public boolean isValid() {
...
+    /**
+     * Condition that finds a distinct set of objects of a specific class. As a Realm is unordered,
+     * it is undefined which objects are returned in case of multiple occurrences.
+     *
+     * @param fieldName the field name.
+     * @return the query object.
+     * @throws IllegalArgumentException if a field name does not exist or the field is not indexed.
+     */
+    public RealmQuery<E> distinct(String fieldName) {
+        long columnIndices[] = schema.getColumnIndices(fieldName, RealmFieldType.STRING);
```
cmelchior: This would only allow String fields, other types are allowed as well:  Boolean, Long, Date

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -1092,6 +1092,20 @@ public boolean isValid() {
...
+        this.query.distinct(columnIndices);
+        return this;
```
cmelchior: I am not sure this will work. I think of `distinct()` as a selector just like `findAll` and `findFirst` it is not something you can append to.

realm/realm-jni/src/io_realm_internal_TableQuery.cpp
kneth: No, `distinct()` takes a vector as argument, and it is the combination of columns which will have unique values.

realm/realm-library/src/main/java/io/realm/RealmQuery.java
cmelchior: I think this can only be a shorthand for `realm.where(Foo.class).findAll().distinct("field")`, but I am not totally sure.

cmelchior:
Great start, but I think you are missing `distinctAsync()` as well.

stk1m1:
Thanks for the thorough review. I've spent too much time on figuring out `distinct` query. It was a great learning though.

Could you elaborate a bit more about `distinctAsync()`? it is already implemented on RealmQuery. I believe. Do you want me to test it as well?

realm/realm-jni/src/io_realm_internal_TableQuery.cpp
```diff
@@ -1865,3 +1865,40 @@ JNIEXPORT void JNICALL Java_io_realm_internal_TableQuery_nativeIsEmpty
...
+
+        // Field queries
+        int col_type = table_ref->get_column_type(S(column_idx));
+        switch (col_type) {
+            case type_Bool:
+            case type_Int:
+            case type_DateTime:
+            case type_String:
+                try {
+                    pQuery->and_query(Query(table_ref->get_distinct_view(S(column_idx)).get_query()));
```
cmelchior: I am not sure this will work correctly, as the distinct_view quite possibly still have the same backing query. See my comment on the Java API. I think this just needs to run the query then restrict it to the distinct view

realm/realm-library/src/main/java/io/realm/RealmQuery.java
stk1m1: That's what I was wondering as well. But for example, can you go with something like `realm.where(Foo.class).lessThanOrEqualTo("something", new Date()).distinct("something")`?

The semantic of how and where to put `distinct` confused me a bit.

cmelchior: `distinct()` should only be allowed as the final "execute" method, which is why it should return a `RealmResults` and not a RealmQuery. You can still create a query on a RealmResults, but it is a  query that is restricted by the TableView created by the previous query, not something you append to a already existing query.

cmelchior: i.e. the following is allowed:
```
RealmResults<Foo> results = realm.where(Foo.class).findAll().distinct("fieldName");
RealmResults<Foo> results = realm.where(Foo.class).distinct("fieldName");
```

but the below should not be:
```
RealmResults<Foo> results = realm.where(Foo.class).distinct("fieldName").equalTo("fieldAge", 10).findAll();
```

cmelchior:
RealmQuery.distinctAsync() is a package protected method, so currently it is not part of the public API. I'm guessing it is to avoid having to much query logic inside the Realm class that uses that method, but It was @nhachicha that implemented it, so perhaps he can fill in the details?

realm/realm-library/src/main/java/io/realm/RealmQuery.java
stk1m1: I really appreciate your examples. They clear my view so much.

roughboy: How about below API?

RealmQuery:
//equivalent of findAllSorted(String fieldName)
`findAllDistinct(String fieldName)`
//equivalent of findAllSortedAsync(String fieldName)
`findAllDistinctAsync(String fieldName)`

RealResults:
//equivalent of sort(String fieldName)
`distinct(String fieldName)`

Realm:
//equivalent of allObjectsSorted(Class<E> clazz, String fieldName)
//requires renaming of already existing distinct() method
`allObjectsDistinct(Class<E> clazz, String fieldName)`

//requires renaming of already existing distincAsync() method
`allObjectsDistinctAsync(Class<E> clazz, String fieldName)`

As a side effect below sort method should also be added:
`allObjectsSortedAsync(Class<E> clazz, String fieldName)`

stk1m1: @roughboy thanks for your input. We'll consider thoroughly.

realm/realm-library/src/androidTest/java/io/realm/RealmQueryTest.java
```diff
@@ -1802,4 +1803,88 @@ public void execute(Realm realm) {
...
+        for (String fieldName : new String[]{"Boolean", "Long", "Date", "String"}) {
+            try {
+                testRealm.where(AnnotationIndexTypes.class).distinct("notIndex" + fieldName);
+                fail("notIndex" + fieldName);
+            } catch (UnsupportedOperationException ignored) {
+            }
+        }
+    }
+
+    public void testDistinctDoesNotExist() {
```
bmunkholm: In my book it's preferred to merge these "negative" testcases which tests invalid use cases and has identical setup.

cmelchior: I don't agree with that. We end up with massive test methods where the logic ends up being split into separate sections anyway to be able to handle the complexity. Especially for these that already test multiple things

realm/realm-library/src/main/java/io/realm/RealmQuery.java
bmunkholm: allDistinctObjects reads better to me.

cmelchior: On RealmQuery I think `distinct` is much preferable as the context is very clear IMO. `allDistinctObject` seems very verbose to me. On `Realm` I could see `allDistinctObjects` having a better case (Right now it is called `Realm.distinct()`) as it might not be it might not be as clear what you mean by calling it on the Realm object itself.

stk1m1:
**What is done**
1. `TableView @ JNI` -> `TableView @ Java` -> `RealQuery` is the modification path I built for this commit. 
2. The previous JNI code for `TableQuery` are all reverted.   

**Questions**
1. The diff between `TableView::sync_distinct_view()` and `TableView::distict()` isn't very clear to me. `distinct()` seems to be a wrapper for the former func().  
2. `TableView::distict()` can take a vector of indexes. I guess I can build a derivative of `RealmQuery.distinct()` that can take multiple columns. 
e.g. `testRealm.where(Foo.class).distinct('name', 'phone', 'address')`   
3. `TableView` seems to be proxy of `Table` that it really doesn't have values. I haven't look into core code yet but I suppose `TableView` is just a collection of pointers to the real values in `Table`. In that way, I can play with `TableView` however I want it without modifying the values in `Table`. Correct me if I'm dreaming.  
4. `TV(nativeViewPtr)->distinct(S(columnIndex));`  is what's happening down @ JNI. Does it look fine? I don't copy anything. Just removing duplicates. 
5. `public TableView getDistinctView(long columnIndex)' on `TableView @ JAVA`. I feel some kind of itch with the name and what it does.

realm/realm-library/src/androidTest/java/io/realm/RealmQueryTest.java
cmelchior: Testing for linked fields still seem to be missing?

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -1092,6 +1093,28 @@ public boolean isValid() {
...
+     */
+    public RealmResults<E> distinct(String fieldName) {
+        Table table = this.table.getTable();
+        long columnIndex = table.getColumnIndex(fieldName);
+        if (columnIndex == -1) {
+            throw new IllegalArgumentException(String.format("Field name '%s' does not exist.", fieldName));
+        }
+        TableView tableView = this.query.findAll().getDistinctView(columnIndex);
+        RealmResults<E> realmResults = RealmResults.createFromTableOrView(this.realm, tableView, this.clazz);
+        if (realm.handlerController != null) {
```
cmelchior: I think this should be moved to the RealmResults constructor. It seems way to easy to forget to add it.

realm/realm-library/src/main/java/io/realm/internal/TableView.java
```diff
@@ -789,6 +789,13 @@ public Table pivot(long stringCol, long intCol, PivotType pivotType){
         return result;
    }
 
+    public TableView getDistinctView(long columnIndex) {
```
cmelchior: Need Javadoc. All public methods should have javadoc even if they are in the internal API.

kneth:
Question 1: The `sync_distinct_view()` can be used to synchronise a `TableView` when changes to a `Table` occurs (maybe in another thread).

Question 2: Yes, the vector of indices is used as you explain. I think it is a recent enhancement, and we haven't implemented it for `Realm.distinct()`.

Question 3: A `TableView` is a kind of proxy of a `Table`. Every change to a `TableView` will change the underlying `Table`.

Question 4: `TV(nativeViewPtr)` is just a cast of a Java long to a `TableView *`, and `S()` casts a Java long to `std::size_t` (the bit width of `std::size_t` depends on architecture a while Java long is always 64 bits). The vector discussed in Question 2 should probably be used in the call here, so you need something like `std::vector<std::size_t> column_indices = ... ; TV(nativeViewPtr)->distinct(column_indices);`.

Question 5: I don't mind if you change the name.

realm/realm-jni/src/io_realm_internal_tableview.cpp
```diff
@@ -49,6 +49,37 @@ JNIEXPORT jlong JNICALL Java_io_realm_internal_TableView_createNativeTableView(
...
+        if (!TV(nativeViewPtr)->get_parent().has_search_index(S(columnIndex))) {
+            ThrowException(env, UnsupportedOperation, "The field must be indexed before distinct() can be used.");
+            return;
+        }
+        switch (TV(nativeViewPtr)->get_column_type(S(columnIndex))) {
+            case type_Bool:
+            case type_Int:
+            case type_DateTime:
+            case type_String:
+                try {
```
kneth: You are already in a `try`/`catch` block so it is not needed.

realm/realm-jni/src/io_realm_internal_tableview.cpp
```diff
@@ -49,6 +49,37 @@ JNIEXPORT jlong JNICALL Java_io_realm_internal_TableView_createNativeTableView(
     return 0;
 }
 
+JNIEXPORT void JNICALL Java_io_realm_internal_TableView_nativeGetDistinctView(
+    JNIEnv* env, jobject, jlong nativeViewPtr, jlong columnIndex)
+{
+    try {
```
kneth: Please move `try` to after the three validation steps - just before `switch`.

cmelchior:
Looks great. Only had a minor comment about a missing unit test :+1: 

Only question unanswered seems to be about how many parameters to have for distinct(). We normally avoid just adding methods to the public API unless there is a good reason for it. 

In this case however, we can just change the the API to `public RealmResults<Foo> distinct(String distinctField1, String... distinctFields)` though which can be used in both of these cases:

```
query.distinct("field1");
query.distinct("field1", "field2");
```
So we should probably do that. That should also be done for `Realm.distinct()`, but can be in another PR.

kneth:
Look very nice :+1:

dalinaum:
Besides a missing unit test, :+1:

stk1m1:
@dalinaum Thank you very much for taking a look into it.  :wink:
Would you kind enough to hint me what I should add in unit test area?
Thanks in advance!

dalinaum:
@stk1m1 Did you check https://github.com/realm/realm-java/pull/2076/files#r49596156 ?

stk1m1:
@dalinaum Thanks for pointing out. I'll look into it right away!  :thumbsup:

realm/realm-jni/src/io_realm_internal_tableview.cpp
```diff
@@ -49,6 +49,36 @@ JNIEXPORT jlong JNICALL Java_io_realm_internal_TableView_createNativeTableView(
...
+                break;
+            }
+            default:
+                ThrowException(env, IllegalArgument, "Invalid type - Only String, Date, boolean, short, int, long and their boxed variants are supported.");
+                return;
+            break;
+        }
+        return;
+    } CATCH_STD()
+    return;
```
beeender: redundant `return`

realm/realm-jni/src/io_realm_internal_tableview.cpp
```diff
@@ -49,6 +49,36 @@ JNIEXPORT jlong JNICALL Java_io_realm_internal_TableView_createNativeTableView(
...
+            case type_String:{
+                TV(nativeViewPtr)->distinct(S(columnIndex));
+                break;
+            }
+            default:
+                ThrowException(env, IllegalArgument, "Invalid type - Only String, Date, boolean, short, int, long and their boxed variants are supported.");
+                return;
+            break;
+        }
+        return;
```
beeender: redundant `return`

realm/realm-jni/src/io_realm_internal_tableview.cpp
```diff
@@ -49,6 +49,36 @@ JNIEXPORT jlong JNICALL Java_io_realm_internal_TableView_createNativeTableView(
...
+            case type_Int:
+            case type_DateTime:
+            case type_String:{
+                TV(nativeViewPtr)->distinct(S(columnIndex));
+                break;
+            }
+            default:
+                ThrowException(env, IllegalArgument, "Invalid type - Only String, Date, boolean, short, int, long and their boxed variants are supported.");
+                return;
+            break;
```
beeender: redundant `return` and fix the indent for break.

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -1092,6 +1093,29 @@ public boolean isValid() {
         return this;
     }
 
+    /**
+     * Returns a distinct set of objects of a specific class. As a Realm is unordered, it is
+     * undefined which objects are returned in case of multiple occurrences.
```
beeender: From the core doc, this behaviour is defined i think? From doc of `TableView.distinct`: 
> If two rows are indentical (for the given set of distinct-columns), then the last row is removed.

beeender: Will null value exist in the returned results?

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -1092,6 +1093,29 @@ public boolean isValid() {
...
+     *
+     * @param fieldName the field name.
+     * @return a non-null {@link RealmResults} containing the distinct objects.
+     * @throws IllegalArgumentException if a field name does not exist, the field is not indexed, or
+     * the field's type is not supported. UnsupportedOperationException if the field is not indexed.
```
beeender: -> `@throws UnsupportedOperationException if the field is not indexed.`

realm/realm-library/src/main/java/io/realm/internal/TableView.java
```diff
@@ -789,6 +789,23 @@ public Table pivot(long stringCol, long intCol, PivotType pivotType){
...
+     * Returns a {@link TableView} with distinct set of objects of a specific class.
+     * As a Realm is unordered, it is undefined which objects are returned in case of multiple
+     * occurrences.
+     *
+     * @param columnIndex the column index.
+     * @return a non-null {@link TableView} containing the distinct objects.
+     * @throws IllegalArgumentException if the type of the column is unsupported.
+     * UnsupportedOperationException if a column is not indexed.
+     */
+    public TableView getDistinctView(long columnIndex) {
```
beeender: why call this `getDistinctView`? Isn't `distinct` enough since this won't create any new `TableView`?

realm/realm-library/src/main/java/io/realm/internal/TableView.java
```diff
@@ -789,6 +789,23 @@ public Table pivot(long stringCol, long intCol, PivotType pivotType){
...
 
+    /**
+     * Returns a {@link TableView} with distinct set of objects of a specific class.
+     * As a Realm is unordered, it is undefined which objects are returned in case of multiple
+     * occurrences.
+     *
+     * @param columnIndex the column index.
+     * @return a non-null {@link TableView} containing the distinct objects.
+     * @throws IllegalArgumentException if the type of the column is unsupported.
+     * UnsupportedOperationException if a column is not indexed.
```
beeender: -> `@throws UnsupportedOperationException if a column is not indexed.`

realm/realm-library/src/main/java/io/realm/internal/TableView.java
```diff
@@ -789,6 +789,23 @@ public Table pivot(long stringCol, long intCol, PivotType pivotType){
         return result;
    }
 
+    /**
+     * Returns a {@link TableView} with distinct set of objects of a specific class.
+     * As a Realm is unordered, it is undefined which objects are returned in case of multiple
+     * occurrences.
```
beeender: I think it is defined behaviour...

realm/realm-library/src/main/java/io/realm/internal/TableView.java
stk1m1: @beeender That's exactly what I thought. ~~On the other hand, it would be nice to let users know they are getting `distincted` tableview.~~ NVM. I'm on my way to fix it.

realm/realm-library/src/main/java/io/realm/RealmQuery.java
stk1m1: @beeender I don't think so. I get the sense from your comment that Java side classes should ensemble their counterpart in core side. i.e. `void distinct()` -> `void distinct()`. Not only this makes my job easier but also we'd have one unified idea around core and whoever will follow would have short amount of time to get under the skin. 

I'll dearly follow your pointers around javadoc and coding style.

beeender: :+1:

stk1m1:
**What is done**
1. Miscellaneous minor fixes were applied (javadoc, indentation etc...) 
2. `RealmResults.distinct()` is ready for review.
3. `RealmResults.createFromTableOrView` & `RealmResults.createFromDynamicTableOrView` are refactored as it is mentioned [here](https://github.com/realm/realm-java/pull/2076#discussion_r49717879). 
As a consequence, all the `handlerController` are removed from `RealmQuery`. The side effects further expand their limbs to `DynamicRealm`. I tried to see what would happen with removing `handlerController` handling from `DynamicRealm` and found out it only generated 4 more failures. It will later be discussed in detail. 
4. `TableView.distinct()` on Java follows the convention of `TableView::distinct()` on core. 
5. Got a great feedback from @beeender and a sudden epiphany on the relation between `core` <-> `Java` with followings.
> @stk1m1 : I get the sense from your comment that Java side classes should ensemble their counterpart in core side. i.e. void distinct() -> void distinct(). Not only this makes my job easier but also we'd have one unified idea around core and whoever will follow would have short amount of time to get under the skin.    

**Questions**
1. Does `distinct` need to have linked test b/c it is not Supported?

realm/realm-jni/src/io_realm_internal_tableview.cpp
```diff
@@ -49,6 +49,34 @@ JNIEXPORT jlong JNICALL Java_io_realm_internal_TableView_createNativeTableView(
...
+        switch (TV(nativeViewPtr)->get_column_type(S(columnIndex))) {
+            case type_Bool:
+            case type_Int:
+            case type_DateTime:
+            case type_String:{
+                TV(nativeViewPtr)->distinct(S(columnIndex));
+                break;
+            }
+            default:{
+                ThrowException(env, IllegalArgument, "Invalid type - Only String, Date, boolean, short, int, long and their boxed variants are supported.");
```
beeender: `byte, short, int, long` --- `byte` is missing :)

realm/realm-library/src/androidTest/java/io/realm/RealmResultsTest.java
```diff
@@ -1219,4 +1220,85 @@ public void testComparisons() {
         assertEquals(10, testRealm.where(AllTypes.class).lessThanOrEqualTo(AllTypes.FIELD_LONG, -1).findAll().size());
         assertEquals(10, testRealm.where(AllTypes.class).lessThan(AllTypes.FIELD_LONG, 0).findAll().size());
     }
+
+    // RealmResults.distinct(): requires indexing, and type = boolean, integer, date, string
+    private void populateForDistinct(Realm realm, long numberOfBlocks, long numberOfObjects, boolean withNull) {
```
beeender: Maybe move this to `TestHelper`? But i am fine that do this later when cleaning up `RealmResultsTest`.

realm/realm-library/src/main/java/io/realm/RealmQuery.java
cmelchior: @beeender but the original order is undefined unless you sorted it, so I think this is accurate? Perhaps adding "unless sorted, in which case the first object is returned."

beeender: "unless sorted, in which case the first object is returned." looks good. Users who read the doc will get benefits from that :P

realm/realm-library/src/main/java/io/realm/RealmResults.java
```diff
@@ -538,6 +546,19 @@ public double average(String fieldName) {
...
+     * undefined which objects are returned in case of multiple occurrences.
+     *
+     * @param fieldName the field name.
+     * @return a non-null {@link RealmResults} containing the distinct objects.
+     * @throws IllegalArgumentException if a field name does not exist.
+     * @throws IllegalArgumentException if a field's type is not supported.
+     * @throws UnsupportedOperationException if a field is not indexed.
+     */
+    public RealmResults<E> distinct(String fieldName) {
+        return this.where().distinct(fieldName);
```
cmelchior: This ins extra work. Instead you can just call `distinct` on the TableView backing this RealmResults directly. The current implementation will call  `findAll` first which will create an unnecessary TableView.

cmelchior: Edit: Actually this implementation might be okay since it is returning the distinct view. If it modified in place it would make more sense to return `void`. Just leave this as it is.

stk1m1:
1. @beeender aforementioned item `byte` is appended.
2. `RealmQuery.distinct()` linked field tests for indexed, unindexed, and unsupported types are added. 
3. If the test methods look fine, I will propagate them on `RealmResults` side.

realm/realm-library/src/androidTest/java/io/realm/entities/AnnotationIndexTypes.java
```diff
@@ -165,4 +168,10 @@ public Date getNotIndexDate() {
     public void setNotIndexDate(Date notIndexDate) {
         this.notIndexDate = notIndexDate;
     }
+
+    public void setFieldObject(AnnotationIndexTypes aFieldObject) {
```
cmelchior: aFieldObject -> `fieldObject` or just `object`

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -1092,6 +1093,27 @@ public boolean isValid() {
...
+     * Returns a distinct set of objects of a specific class. As a Realm is unordered, it is
+     * undefined which objects are returned in case of multiple occurrences.
+     *
+     * @param fieldName the field name.
+     * @return a non-null {@link RealmResults} containing the distinct objects.
+     * @throws IllegalArgumentException if a field name does not exist.
+     * @throws IllegalArgumentException if a field's type is not supported.
+     * @throws UnsupportedOperationException if a field is not indexed.
+     */
+    public RealmResults<E> distinct(String fieldName) {
```
cmelchior: Did you decide to add multiple parameters later?

cmelchior:
Apart from a few small comments I think this looks good.

stk1m1:
@cmelchior  I'll move on to `RealmResults' test part then. Thanks for the review!  :satisfied:

realm/realm-jni/src/io_realm_internal_tableview.cpp
```diff
@@ -49,6 +49,34 @@ JNIEXPORT jlong JNICALL Java_io_realm_internal_TableView_createNativeTableView(
...
+    if (!TV(nativeViewPtr)->get_parent().has_search_index(S(columnIndex))) {
+        ThrowException(env, UnsupportedOperation, "The field must be indexed before distinct() can be used.");
+        return;
+    }
+    try {
+        switch (TV(nativeViewPtr)->get_column_type(S(columnIndex))) {
+            case type_Bool:
+            case type_Int:
+            case type_DateTime:
+            case type_String:{
```
kneth: You don't need the block scope (`{ ... }`) here.

realm/realm-jni/src/io_realm_internal_tableview.cpp
```diff
@@ -49,6 +49,34 @@ JNIEXPORT jlong JNICALL Java_io_realm_internal_TableView_createNativeTableView(
...
+                TV(nativeViewPtr)->distinct(S(columnIndex));
+                break;
+            }
+            default:{
```
kneth: You don't need the block scope (`{ ... }`) here.

realm/realm-library/src/main/java/io/realm/RealmQuery.java
stk1m1: I setup a tentative plan like below.

1. Finish `RealmResults` testing.
2. Implement `distinctAsync()` and testing.
3. Implement `Realm.distinct(...)`, `RealmQuery.distinct(...)`, and then finally `RealmResults.distinct(...)`. and finish testing.

I try to implement `em all in incremental manner, and I hope this sequence works with team's practice. If you see a need for change, please let me know. I will follow accordingly.

cmelchior: That is fine. 2 + 3 should probably be in their own PR though instead of keeping this alive for to long.

stk1m1: @cmelchior As soon as I finish up and fix all the minors, I will merge this. I'm planning to start right away on item 2 then.

stk1m1:
1. @kneth unnecessary (`{ ... }`) are removed.
2. `RealmResults.distinct()` linked field tests for indexed, unindexed, and unsupported types are added. 
3. `* Added RealmQuery.distinct() and RealmResults.distinct() (#1568).` added to *changelog.txt*
4. Instrument test results are [here](http://stk1m1.github.io/test-results/io.realm.html)

stk1m1:
retest this please.

stk1m1:
please review @realm/java. Jenkins says 
```
* What went wrong:
Task 'pmd' not found in root project 'realm'.
```

cmelchior:
You need to merge master to your branch

stk1m1:
retest this please.

stk1m1:
1. #2096 needs to be merged first. I'm getting following error.
```
realm-java/realm/realm-library/src/main/java/io/realm/RealmResults.java 	176 	These nested if statements could be combined
```
2. I've tried new `RealmResults.distinct()` method. As it is [mentioned](https://github.com/realm/realm-java/pull/2076#discussion_r49828243), the previous method takes extra work and I with the new one works better. In short, new one takes the backing table to get new distinct TableView. 
3. It still puzzles me if multiple `distinct` are acceptable. `testRealm.where(Foo.class).distinct("something").distinct("something")`

beeender:
:+1:

dalinaum:
retest this please

beeender:
This needs a rebase i think.

cmelchior:
@stk1m1 `testRealm.where(Foo.class).findAll().distinct("something").distinct("something")` should be allowed as it is equivalent to  
```
RealmResults<Foo> results = realm.where(Foo.class).findAll().distinct();
RealmResults<Foo> result2 = results.distinct()
```
The nicer API would be:

```
testRealm.where(Foo.class).findAll().distinct("field1", "field2) 
// or
testRealm.where(Foo.class).distinct("field1", "field2) 
```

stk1m1:
retest this please.

stk1m1:
retest this please.

realm/realm-library/src/main/java/io/realm/RealmQuery.java
cmelchior: This still needs to be addressed

realm/realm-library/src/main/java/io/realm/RealmResults.java
```diff
@@ -536,6 +545,37 @@ public double average(String fieldName) {
         }
     }
 
+    /**
+     * Returns a distinct set of objects of a specific class. As a Realm is unordered, it is
+     * undefined which objects are returned in case of multiple occurrences.
+     *
```
cmelchior: This JavaDoc should have the same change as the other.

cmelchior:
Except from a small javadoc change needed. This looks ready to merge :+1:  Remember to rebase and squash before merging :)

Also please create new issues for
- `distinctAsync`
- `distinct(String field1, String... otherFields)/distinctAsync(String field1, String... otherFields)`

So we can track them after this has been merged.

dalinaum:
retest this please

dalinaum:
Is this PR closed without merging?

stk1m1:
@dalinaum apparently so. Do I need to restore and merge?

dalinaum:
If you just want to rebase and squash your commits, you can use `git push -f origin stk1m1:stk/fix-issues` instead of creating another PR.

PR #2080, opened 2016-01-13
JUnit4 style convention

cmelchior:
As discussed last week, this PR exemplifies how our current unit tests could be converted to JUnit4. It should be seen as starting point for discussion. 

The current thinking is this:

Migrating our unit tests give us a lot more flexibility and control. 
Changes compared to previous:

1) Unit test method names no longer starts with "test". Use `@Test` instead.

2) Naming: camelCase + `_` to separate sections, e.g. `<methodName>_<param>_<description>`, `<methodName>_<description>`.

3) Failing unit tests no longer ends with "Throws". Use `@Test(expected = xxx.class)` or `ExpectedException` instead.

4) TestRealmConfigurationFactory _must_ be used to create RealmConfigurations.

5) Use @RunInLooperThread to massively simplify Handler tests.

6) remove `test` from variables, like `testRealm` and `testConfig`.

7) Tests should as much as possible be grouped in files mirroring the name of the class they belong to: `Realm` methods should be in `RealmTests`

Hopefully this will reduce the length of unit test method names + make them more readable.
Using `TestRealmConfigurationFactory` + `RunInLooperThread` will ensure that Realm files are properly deleted between each unit test.

@realm/java


realm/realm-library/src/androidTest/java/io/realm/DynamicRealmTests.java
```diff
@@ -0,0 +1,751 @@
+/*
+ * Copyright 2015 Realm Inc.
```
beeender: 2016

cmelchior: Damn you new year :confetti_ball:

realm/realm-library/src/androidTest/java/io/realm/rule/RunInLooperThread.java
```diff
@@ -83,13 +82,11 @@ public void run() {
...
                             try {
                                 realm = Realm.getInstance(realmConfiguration);
-
                                 base.evaluate();
-
                                 Looper.loop();
                             } catch (Throwable e) {
                                 threadAssertionError[0] = e;
-
+                                unitTestFailed = true;
```
beeender: I am not sure if `threadAssertionError[0]` is still necessary. JUNI4 will just stop when any thread throws.

realm/realm-library/src/androidTest/java/io/realm/rule/TestRealmConfigurationFactory.java
```diff
@@ -66,7 +68,9 @@ protected void before() throws Throwable {
     protected void after() {
         try {
             for (RealmConfiguration configuration : configurations) {
-                Realm.deleteRealm(configuration);
+                if (!Realm.deleteRealm(configuration)) {
```
beeender: Uh... we talked about this, the only case returns `false` right now is the file doesn't exist. `deleteRealm` itself will throw ISE when the Realm is still opened. right?

beeender:
Create a wiki page after this PR merged :angel:

realm/realm-library/src/androidTest/java/io/realm/rule/TestRealmConfigurationFactory.java
cmelchior: Yes, I think this is some debug code I accidentally left in :( I'll remove it.

realm/realm-library/src/androidTest/java/io/realm/rule/RunInLooperThread.java
cmelchior: Hmm, not sure either. I'll double check.

cmelchior:
Wouldn't it be better in the `CONTRIBUTING.md` file?

realm/realm-library/src/androidTest/java/io/realm/rule/RunInLooperThread.java
cmelchior: The following cleanup logic requires that we catch the exception and handle it outside the thread it was caught on, so in this case I think it is okay the way it is.

beeender: :+1:

beeender:
Yeah, `CONTRIBUTING.md` is better.

:+1:  And it worth to clean up the lint warning and findbugs issues if they are not checked :camel:

bmunkholm:
It's really hard to see the diff of this DynamicRealmTest.java.

Add the guideline in CONTRIBUTING.md so it can be reviewed.

Missing guidelines about how to organize the test methods and comment them.
It would visually make a LOT of sense to me to just group the related testcases visually. A single //---- line for each section of related tests for a method would make wonders. And a comment about what the test does instead of trying to fit that into unreadable method names. Also add comments to sections of the test code so it's fast to see what's it supposed to do (if it's not trivially small and obvious).

beeender:
Adjust test case orders and group them may cause more chaos in the diff ... so reorder and group should in another PR ---- with reorder and group only.

kneth:
I like the simplification of the async tests!

cmelchior:
Grouping them visually have very little benefit and is impossible to enforce IMO, which means it will drift (as it is very easily missed in GitHub review). Requiring that unit tests start with the method being tested e.g. `createObject_xxx` makes it a lot easier to use IDE tools to get an overview of all unit tests for a given method.

Priorities should be IMO 
1) Readable method name. That is the point of using `_` in the name. It allow for longer method names that are easier to parse than pure camelCase.
2) If that isn't enough, add a comment above the unit test .
3) Comments inside unit tests should be library used especially if it involve concurrency where the flow is hard to follow.

kneth:
I agree with @cmelchior's three priorities. Item 1 will in most cases be enough to find the unit test (using Android Studio's "Structure" pane).

stk1m1:
:thumbsup:

cmelchior:
Fixed issues discovered by FindBugs and added description to `CONTRIBUTING.md` @realm/java

beeender:
:+1: except a small issue.

CONTRIBUTING.md
```diff
@@ -27,8 +27,90 @@ Realm welcomes all contributions! The only requirement we have is that, like man
 
 ## Unit tests
 
-All PR's must be accompanied by related unit tests.
-
+All PR's must be accompanied by related unit tests. All bug fixes must have a unit test proving that the bug is fixed.
+You can use `./realm/gradlew connectedCheck createDebugCoverageReport` to generate a coverage report to check for 
+missing unit test coverage. The aim is 100% code coverage.
```
bmunkholm: line or branch coverage?

CONTRIBUTING.md
```diff
@@ -27,8 +27,90 @@ Realm welcomes all contributions! The only requirement we have is that, like man
...
+When writing unit tests, use the following guide lines:
+
+1) Unit tests must be written using JUnit4.
+
+2) All tests for a class should be grouped in a class called `<className>Tests`, unless the functionality is cross-
+   cutting like [`RxJavaTests`](https://github.com/realm/realm-java/blob/master/realm/realm-library/src/androidTest/java/io/realm/RxJavaTests.java) 
+   or [`RealmAsyncQueryTests`](https://github.com/realm/realm-java/blob/master/realm/realm-library/src/androidTest/java/io/realm/RealmAsyncQueryTests.java).
+
+3) Test methods should use camelCase and underscore `_` between logical sections to increase method name readability. 
+   Methods should ideally start with the name of the method being tested. Patterns like: `<methodName>_<description>`,
```
bmunkholm: "should ideally" -> "must"

CONTRIBUTING.md
cmelchior: I would assume branch coverage, since that is more "complete"

CONTRIBUTING.md
```diff
@@ -27,8 +27,90 @@ Realm welcomes all contributions! The only requirement we have is that, like man
...
+   or [`RunInLooperThread`](https://github.com/realm/realm-java/blob/master/realm/realm-library/src/androidTest/java/io/realm/rule/RunInLooperThread.java) 
+   test rules. This ensures that all Realms are properly closed and deleted between each test.
+
+5) Use the `@RunInLooperThread` rule for any test that depends on Realms notification system. 
+
+6) Input-parameters should be boundary tested. Especially `Null/NotNull`, but also the state of Realm objects like
+   standalone objects, deleted objects, objects from other threads.
+
+7) Unit tests are not required to only have 1 test. It is acceptable to combine multiple tests into one unit test, but
+   if it fails, it should be clear why it failed.
```
bmunkholm: Add "E.g. you can group related tests with the same setup like e.g. negative tests. If you do so, make sure to seprate each "subtest" with a comment stating what you test."

CONTRIBUTING.md
```diff
@@ -27,8 +27,90 @@ Realm welcomes all contributions! The only requirement we have is that, like man
...
+
+8) Use `@Test(expected = xxx.class)` or the `ExceptedException` rule to detect exceptions.
+
+9) Use comments to make the intent of the unit test clear if not possible in the method name. Comments inside unit tests
+   are encouraged, especially for unit tests that involve concurrency.
```
bmunkholm: "especially for unit tests that involve concurrency" -> "for anything not is not obvious at a glance."

CONTRIBUTING.md
```diff
@@ -27,8 +27,90 @@ Realm welcomes all contributions! The only requirement we have is that, like man
...
+
+6) Input-parameters should be boundary tested. Especially `Null/NotNull`, but also the state of Realm objects like
+   standalone objects, deleted objects, objects from other threads.
+
+7) Unit tests are not required to only have 1 test. It is acceptable to combine multiple tests into one unit test, but
+   if it fails, it should be clear why it failed.
+
+8) Use `@Test(expected = xxx.class)` or the `ExceptedException` rule to detect exceptions.
+
+9) Use comments to make the intent of the unit test clear if not possible in the method name. Comments inside unit tests
```
bmunkholm: I would suggest:
9) Use comments to make the intent of the unit test easily understandable at a glance. A simple one line comment is often easier to read thanALongCamelCasedSentenceThatAttemptsToDescribeWhatHappens. Describe the test steps inside the method, if it's not glaringly obvious.

CONTRIBUTING.md
cmelchior: Sometimes you are not just testing one method, but a sequence of events, in which case starting with a methodName would just be weird.

CONTRIBUTING.md
```diff
@@ -27,8 +27,90 @@ Realm welcomes all contributions! The only requirement we have is that, like man
...
+
+5) Use the `@RunInLooperThread` rule for any test that depends on Realms notification system. 
+
+6) Input-parameters should be boundary tested. Especially `Null/NotNull`, but also the state of Realm objects like
+   standalone objects, deleted objects, objects from other threads.
+
+7) Unit tests are not required to only have 1 test. It is acceptable to combine multiple tests into one unit test, but
+   if it fails, it should be clear why it failed.
+
+8) Use `@Test(expected = xxx.class)` or the `ExceptedException` rule to detect exceptions.
```
bmunkholm: I would strongly recommend against using this. It only works as intended for one line methods. Otherwise you have no control over when the exception is supposed to be thrown. We used to use it back in the days, but went away from it for that reason.

CONTRIBUTING.md
```diff
@@ -27,8 +27,90 @@ Realm welcomes all contributions! The only requirement we have is that, like man
...
+        }
+    });
+
+    realm.beginTransaction();
+    realm.createObject(Foo.class);
+    realm.commitTransaction();
+  }
+}
+```  
+
```
bmunkholm: I suggest to add a multi-test with comments :-)

CONTRIBUTING.md
cmelchior: Yes, that is a pretty annoying limitation I agree, but it does reduce some test cases from 4 lines to 1. If we switched to Kotlin for our unit tests we could get the best of both worlds :)

donnfelker: I like the Expected exception annotation values, but if clarity is what we're after and if the annotation has introduced problems before then I'd stay with the longer form to ensure stability. @bmunkholm Can you recall how often you ran into the `expected` version to have caused problems?

donnfelker:
I'm not a fan of this naming: `2) Naming: camelCase + _ to separate sections, e.g. <methodName>_<param>_<description>, <methodName>_<description>.` the reason is because when I write tests I like to read the sentence as if it were a behaviour (I'm quite BDD in that regard). I want to read what the test does. Such as: `public void shouldBeAbleToCheckIfARealmListContainsAnItem". Sure, its a really long method name, but just from reading it you can tell what it does. 

If consistency from the community is what is sought after then I can see why this proposed change is needed. If this proposed change is accepted there should be a good number 5-10 examples of naming methods for particular testing outcomes so contributors can visualize how they should name their method.

donnfelker:
I misread the proposal, deleted my old comment. I agree with the priorities. Tooling will help. :+1:

dalinaum:
:+1:

PR #2084, opened 2016-01-14
Comply with the most basic PMD ruleset

emanuelez:


emanuelez:
@realm/java

realm/realm-library/src/main/java/io/realm/DynamicRealmObject.java
```diff
@@ -62,7 +62,7 @@ public DynamicRealmObject(RealmObject obj) {
 
     DynamicRealmObject(BaseRealm realm, Row row) {
         this.realm = realm;
-        this.row = (row instanceof CheckedRow) ? (CheckedRow) row : ((UncheckedRow) row).convertToChecked();
+        this.row = row instanceof CheckedRow ? (CheckedRow) row : ((UncheckedRow) row).convertToChecked();
```
cmelchior: I'm not really a fan of enforcing this rule. In a lot of cases I find it increases readability IMO.
E.g here. I'll highlight a few more places where below

realm/realm-library/src/main/java/io/realm/RealmBaseAdapter.java
```diff
@@ -45,7 +45,7 @@ public RealmBaseAdapter(Context context, RealmResults<T> realmResults, boolean a
         this.context = context;
         this.realmResults = realmResults;
         this.inflater = LayoutInflater.from(context);
-        this.listener = (!automaticUpdate) ? null : new RealmChangeListener() {
+        this.listener = !automaticUpdate ? null : new RealmChangeListener() {
```
cmelchior: I would keep the `()` here

realm/realm-library/src/main/java/io/realm/RealmFieldType.java
```diff
@@ -76,22 +76,22 @@ public int getNativeValue() {
      */
     public boolean isValid(Object obj) {
         switch (nativeValue) {
-            case 0: return (obj instanceof Long || obj instanceof Integer || obj instanceof Short || obj instanceof Byte);
-            case 1: return (obj instanceof Boolean);
-            case 2: return (obj instanceof String);
-            case 4: return (obj instanceof byte[] || obj instanceof ByteBuffer);
-            case 5: return (obj == null || obj instanceof Object[][]);
-            case 6: return (obj instanceof Mixed ||
+            case 0: return obj instanceof Long || obj instanceof Integer || obj instanceof Short || obj instanceof Byte;
```
cmelchior: Especially here I find it becomes a lot less readable

realm/realm-library/src/main/java/io/realm/internal/Mixed.java
```diff
@@ -54,22 +54,22 @@ public Mixed(boolean value) {
     }
 
     public Mixed(Date value) {
-        assert (value != null);
+        assert value != null;
```
cmelchior: Same here, it becomes less readable IMO

realm/realm-library/src/main/java/io/realm/internal/RealmCore.java
```diff
@@ -57,7 +57,7 @@ private static String getJniFileName()
     public static boolean osIsWindows()
     {
         String os = System.getProperty("os.name").toLowerCase(Locale.getDefault());
-        return (os.contains("win"));
+        return os.contains("win");
```
cmelchior: This is one of the few places where I think it makes sense to remove the `()`

cmelchior:
I am not a fan of enforcing the rule about parentheses as a lot of code is more readable with them IMO. It is not a major objection though if the rest of @realm/java think its fine. The rest seems great :+1:

emanuelez:
I thought the same but the alternative is ignoring such role but then having to accept things like `do((x))`, which IMO is worse.

cmelchior:
I don't agree. I cannot remember we ever had to comment on such a style issue in any PR, while I think that 90% of the parentheses removed in this PR actually decrease readability.

realm/realm-library/build.gradle
```diff
@@ -62,8 +63,8 @@ task javadocJar(type: Jar, dependsOn: javadoc) {
 preBuild.dependsOn ':realm-jni:buildAndroidJni'
 
 task findbugs(type: FindBugs) {
-    dependsOn assemble
-    group = 'Test'
+    dependsOn assembleDebug
```
cmelchior: Thank you :)

realm/realm-library/src/main/java/io/realm/DynamicRealmObject.java
kneth: Using `()` in `?:` looks C like to me - and I wish us to move on ;-)

realm/realm-library/src/main/java/io/realm/RealmBaseAdapter.java
kneth: They don't increase the readability in my view.

emanuelez:
After a democratic vote this gets merged :)

realm/realm-library/src/main/java/io/realm/RealmFieldType.java
kneth: If you argued that `(obj instanceof Long) || (obj instanceof Integer) || (obj instanceof Short) || (obj instanceof Byte)` is easier to read, I might agree. But a single parenthesis doesn't.

cmelchior: Code styles, the best form of bike shedding there exists :)

kneth: Oh no, what I write is true in any sense :-)

realm/realm-library/src/main/java/io/realm/internal/RealmCore.java
```diff
@@ -122,15 +122,14 @@ public static synchronized void loadLibrary() {
...
     private static String loadLibraryWindows() {
-///*
         try {
             addNativeLibraryPath(BINARIES_PATH);
             resetLibraryPath();
-        }
-        catch (Throwable e) {
+        } catch (Throwable e) {
             // Above can't be used on Android.
+            System.out.println("Failed to add native library path\n");
```
bmunkholm: System.err ?? (same below)

realm/realm-library/src/main/java/io/realm/DynamicRealmObject.java
donnfelker: I'm with @cmelchior. In certain instances I also find additional parens increase readability. Readability > minimalist code. The less cognitive load a developer has to perform the less likely they are to make an error.

realm/realm-library/src/main/java/io/realm/RealmFieldType.java
donnfelker: I also find it a lot less readable.

PR #2090, opened 2016-01-15
Fix flaky tests related with async transaction

beeender:
* Background realm needs to be closed before notifying other threads
  in async transaction.
* Latch needs to be called after Realm closed.

### DIFF
--- a/realm/realm-library/src/main/java/io/realm/BaseRealm.java
+++ b/realm/realm-library/src/main/java/io/realm/BaseRealm.java
@@ -291,8 +291,23 @@ abstract class BaseRealm implements Closeable {
      * changes from this commit.
      */
     public void commitTransaction() {
+        commitTransaction(null);
+    }
+
+    /**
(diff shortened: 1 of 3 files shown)


cmelchior:
Concurrency is fun :+1:

realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -292,9 +292,24 @@ public void beginTransaction() {
      * changes from this commit.
      */
     public void commitTransaction() {
+        commitTransaction(null);
+    }
+
+    /**
+     * Commit transaction, run the given runnable and then send notifications. The runnable is useful to meet some
```
dalinaum: Commits transaction, runs the given runnable and then sends notifications

dalinaum:
Besides a minor comment, LGTM :+1:

beeender:
retest this please

PR #2091, opened 2016-01-15
Introduce basic PMD rules

emanuelez:
* Moves config files up one folder (so they can be used by the annotation processor in the future)
 * Introduces PMD basic PMD checks and fixes the issues found


emanuelez:
@realm/java

realm/realm-library/src/main/java/io/realm/rx/RealmObservableFactory.java
```diff
@@ -222,4 +222,9 @@ public boolean equals(Object o) {
         return o instanceof RealmObservableFactory;
     }
 
+    @Override
+    public int hashCode() {
+        return 37;
```
kneth: Really?

kneth:
Nice catches!

realm/realm-library/src/main/java/io/realm/rx/RealmObservableFactory.java
cmelchior: The class has no state that can be set from the outside, so any instance of this class is equal to any other instance of this class. A hardcoded hashCode is fine in that case :)

emanuelez:
retest this please

emanuelez:
retest this please

emanuelez:
retest this please

cmelchior:
:+1:

PR #2095, opened 2016-01-15
Added support for adding micro benchmarks

cmelchior:
This PR adds support for micro benchmarks of the Realm API:

* It uses [Spanner](https://github.com/cmelchior/spanner). Spanner is my fork of Caliper that is more Android friendly and allows us to compare results against baselines.
* All benchmarks are in a separate `src/benchmarks` folder but are treated as instrumentation tests.
* Two new gradle targets have been added `connectedBenchmarks` and `connectedUnitTests` that replace `connectedCheck`. Ideally I wanted to just create a `runBenchmarks` task, but it didn't seem possible while still reusing the instrumentation test setup already provided by Google.
* The benchmarks will produce JSON/CSV files and save them on the SD card.
* Json files is the internal Spanner storage format. CSV is for external use and can be configured as we like.
* Subsequent runs can then use the Json files as a baseline.
* New benchmarks are really easy to add (just add a new method + annotation).
* Runs as separate JUnit tests, which means that failures automatically fail builds.
* Can be run from within the IDE as well.

TODO
- [x] Release a non-SNAPSHOT version of Spanner we can depend upon.
- [ ] Figure out how Jenkins can support pulling baseline JSON files and store them.
- [ ] Figure out how Jenkins can put stored baseline files back on the device before running benchmarks.
- [ ] Determine exact format of CSV files.
- [x] Merge all unit tests from https://github.com/realm/realm-java/blob/bp-performance-test/realm/src/androidTest/java/io/realm/RealmPerformanceTest.java into this PR as a starting point. 
- [ ] When merged update CI to use the updated tasks.

Example of CSV format:
```
"Method name","Trial","Measurements","Min.","Max.","Mean","25pct.","50pct.","75pct."
"readDouble","1","9","151.80","240.00","169.75","152.89","155.14","182.79"
"readLong","1","9","130.75","149.53","137.29","133.41","137.23","138.85"
"readString","1","9","512.65","590.50","548.60","534.12","544.03","569.26"
```


cmelchior:
Put on hold. Awaiting test device: https://github.com/realm/realm-java-private/issues/39

cmelchior:
Ready for review @realm/java 

Note that all CI related tasks are being punted a bit, but at least this will provide easier tools for benchmarking locally when assessing API changes.

realm/realm-library/src/benchmarks/java/io/realm/benchmarks/RealmObjectWriteBenchmarks.java
```diff
@@ -0,0 +1,78 @@
...
+        RealmConfiguration config = new RealmConfiguration.Builder(InstrumentationRegistry.getTargetContext()).build();
+        Realm.deleteRealm(config);
+        realm = Realm.getInstance(config);
+        realm.beginTransaction();
+        writeObject = realm.createObject(AllTypes.class);
+    }
+
+    @AfterExperiment
+    public void after() {
+        realm.cancelTransaction();
```
kneth: Not sure what you're measuring if you cancel the transaction. The name (`RealmObjectWriteBenchmarks`) indicates that objects are written to disk. The commit part of a transaction is probably the expensive part, so why cancel?

realm/realm-library/src/benchmarks/java/io/realm/benchmarks/RealmResultsBenchmarks.java
```diff
@@ -0,0 +1,109 @@
...
+        realm.beginTransaction();
+        for (int i = 0; i < DATA_SIZE; i++) {
+            AllTypes obj = realm.createObject(AllTypes.class);
+            obj.setColumnLong(i);
+            obj.setColumnBoolean(i % 2 == 0);
+            obj.setColumnString("Foo " + i);
+            obj.setColumnDouble(i + 1.234D);
+        }
+        realm.commitTransaction();
+        results = realm.allObjects(AllTypes.class);
```
kneth: `allObjects()` -> `where().findAll()`

realm/realm-library/src/benchmarks/java/io/realm/benchmarks/config/BenchmarkConfig.java
```diff
@@ -0,0 +1,60 @@
+package io.realm.benchmarks.config;
```
kneth: Standard header is missing.

realm/realm-library/src/benchmarks/java/io/realm/benchmarks/config/CSVResultProcessor.java
```diff
@@ -0,0 +1,90 @@
+package io.realm.benchmarks.config;
```
kneth: Standard header is missing.

realm/realm-library/src/benchmarks/java/io/realm/benchmarks/RealmQueryBenchmarks.java
```diff
@@ -0,0 +1,94 @@
...
+
+    @AfterExperiment
+    public void after() {
+        realm.close();
+    }
+
+    @Benchmark
+    public void containsQuery(long reps) {
+        for (long i = 0; i < reps; i++) {
+            RealmResults<AllTypes> realmResults = realm.where(AllTypes.class).contains(AllTypes.FIELD_STRING, "Foo 1").findAll();
```
stk1m1: Wouldn't it be a good idea to query other than identical `"Foo 1"`, for example `"Foo " + (i % DATA_SIZE)`?

realm/realm-library/src/benchmarks/java/io/realm/benchmarks/RealmResultsBenchmarks.java
```diff
@@ -0,0 +1,109 @@
...
+
+    @AfterExperiment
+    public void after() {
+        realm.close();
+    }
+
+    @Benchmark
+    public void get(long reps) {
+        for (long i = 0; i < reps; i++) {
+            AllTypes item = results.get(0);
```
stk1m1: maybe `results.get(i % DATA_SIZE)` instead of `results.get(0)`?

realm/realm-library/src/benchmarks/java/io/realm/benchmarks/RealmObjectWriteBenchmarks.java
cmelchior: You are probably right, my rationale was that having `beginTransaction/commitTransaction` as part of the benchmark itself means we would be benchmarking persisting to disk rather than actually writing a value.

My intention with this was more to measure the overhead of the Java/JNI when writing different types which I think this does better.

That is not to say there is value in benchmarking a full `begin/commit` cycle. That is currently being done in RealmBenchmarks#emptyTransaction

realm/realm-library/src/benchmarks/java/io/realm/benchmarks/RealmQueryBenchmarks.java
cmelchior: I would rather avoid that because that means you are not potentially are not running the same codepath on each iteration.

Why do you think it would be a good idea?

realm/realm-library/src/benchmarks/java/io/realm/benchmarks/RealmResultsBenchmarks.java
cmelchior: I rather not, see above.

realm/realm-library/src/benchmarks/java/io/realm/benchmarks/RealmObjectWriteBenchmarks.java
kneth: Maybe a rename to `RealmObjectCreateBenchmarks` is appropriate ;-)

kneth: We should definitely also have a benchmark for creating and persisting objects.

kneth:
I think we should merge it as it is. And then create issues for the CI related tasks and create issues for benchmarks we wish to add.

realm/realm-library/src/benchmarks/java/io/realm/benchmarks/RealmQueryBenchmarks.java
stk1m1: What I thought was to potentially fault any sort of cache within the code path and possibly distribute equal chances to each element for being queried. If the test is to be executed in the same path, please ignore my comment.

cmelchior: Ok. While it does make sense to avoid hitting a cache mechanism, that is really up to the developer of the benchmark I think. The reason we want to hit the same codepath is that all the calculations depend on that.

There are other pitfalls to running the same codepath like the VM running doing optimisations midloop, but right now on Android we are in a very limited position to do anything about that.

cmelchior:
All comments addressed.

stk1m1:
👍

dalinaum:
:+1:

PR #2106, opened 2016-01-18
Converted RealmResultsTests to JUnit4

cmelchior:
Note some tests seemed out of place. I added TODO's for those so they can be fixed later instead of making noise to this PR.

Also prefixing all tests with the method being tested seemed to catch a couple of redundant tests :smile: 

@realm/java

### DIFF
--- a/realm/realm-library/src/androidTest/java/io/realm/RealmResultsTests.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/RealmResultsTests.java
@@ -16,7 +16,14 @@
 
 package io.realm;
 
-import android.test.AndroidTestCase;
+import android.support.test.runner.AndroidJUnit4;
+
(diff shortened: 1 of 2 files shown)


cmelchior:
retest this please

dalinaum:
I finally read all modifications. :+1:

beeender:
Stupid github review, doesn't show the diff at all.

> TODO Check if should returns null. Should removeLast work on the empty list?
`TableView.removeLast` is checking `isEmpty()` before removing, so it will work without doing anything. Note here: `TableView.removeLast` can just call core's `TableView.remove_last` directly. But it is something we can change in the object store adaption.

kneth:
:+1:

cmelchior:
retest this please

cmelchior:
retest this please

PR #2118, opened 2016-01-20
RealmQuery.distinctAsync(), RealmResults.distinctAsync() support

stk1m1:
`distinctAsync` on `RealmQuery` and `RealmResults`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 * BREAKING CHANGE: DynamicRealm.executeTransaction() now directly throws any RuntimeException instead of wrapping it in a RealmException (#1682).
 * BREAKING CHANGE: DynamicRealm.executeTransaction() now throws IllegalArgumentException instead of silently accepting a null Transaction object.
 * BREAKING CHANGE: String setters now throws IllegalArgumentException instead of RealmError for invalid surrogates. 
+* BREAKING CHANGE: DynamicRealm.distinct()/distinctAsync() and Realm.distinct()/distinctAsync() now throw IllegalArgumentException instead of UnsupportedOperationException for invalid type or unindexed field.
 * Fixed an error occurring during test and connectedCheck of unit test example (#1934).
 * Fixed bug in jsonExample (#2092).
 * Added RealmQuery.isNotEmpty() (#2025). (Thank you @stk1m1)
@@ -11,6 +12,7 @@
 * Added RealmQuery.distinct() and RealmResults.distinct() (#1568).
 * Fixed bug when multiple calls of RealmResults.distinct() causes to return wrong results (#2198).
 * RealmResults.size() now returns Integer.MAX_VALUE when actual size is greater than Integer.MAX_VALUE (#2129).
+* Added RealmQuery.distinctAsync() and RealmResults.distinctAsync() (#2118).
 
 ## 0.87.4
 * Updated Realm Core to 0.96.0
--- a/realm/realm-library/src/androidTest/java/io/realm/RealmResultsTests.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/RealmResultsTests.java
@@ -34,6 +34,7 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
 
 import io.realm.entities.AllJavaTypes;
 import io.realm.entities.AllTypes;
@@ -45,6 +46,8 @@ import io.realm.entities.NullTypes;
 import io.realm.entities.Owner;
 import io.realm.internal.Table;
 import io.realm.rule.TestRealmConfigurationFactory;
+import io.realm.rule.RunInLooperThread;
+import io.realm.rule.RunTestInLooperThread;
 
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
@@ -60,11 +63,14 @@ public class RealmResultsTests {
     private final static int TEST_DATA_FIRST_HALF = 2 * (TEST_DATA_SIZE / 4) - 1;
     private final static int TEST_DATA_LAST_HALF = 2 * (TEST_DATA_SIZE / 4) + 1;
     private final static long YEAR_MILLIS = TimeUnit.DAYS.toMillis(365);
+    private final static long DECADE_MILLIS = 10 * TimeUnit.DAYS.toMillis(365);
 
     @Rule
     public final TestRealmConfigurationFactory configFactory = new TestRealmConfigurationFactory();
     @Rule
     public final ExpectedException thrown = ExpectedException.none();
+    @Rule
+    public final RunInLooperThread looperThread = new RunInLooperThread();
 
     private  Realm realm;
 
@@ -1317,11 +1323,11 @@ public class RealmResultsTests {
         final long numberOfObjects = 10; // must be greater than 1
         populateForDistinct(realm, numberOfBlocks, numberOfObjects, false);
 
-        RealmResults<AnnotationIndexTypes> distinctBool = realm.where(AnnotationIndexTypes.class).findAll().distinct("indexBoolean");
+        RealmResults<AnnotationIndexTypes> distinctBool = realm.where(AnnotationIndexTypes.class).findAll().distinct(AnnotationIndexTypes.FIELD_INDEX_BOOL);
         assertEquals(2, distinctBool.size());
-        for (String fieldName : new String[]{"Long", "Date", "String"}) {
-            RealmResults<AnnotationIndexTypes> distinct = realm.where(AnnotationIndexTypes.class).findAll().distinct("index" + fieldName);
-            assertEquals("index" + fieldName, numberOfBlocks, distinct.size());
+        for (String field : new String[]{AnnotationIndexTypes.FIELD_INDEX_LONG, AnnotationIndexTypes.FIELD_INDEX_DATE, AnnotationIndexTypes.FIELD_INDEX_STRING}) {
+            RealmResults<AnnotationIndexTypes> distinct = realm.where(AnnotationIndexTypes.class).findAll().distinct(field);
+            assertEquals(field, numberOfBlocks, distinct.size());
         }
     }
 
@@ -1335,10 +1341,10 @@ public class RealmResultsTests {
         RealmResults<AnnotationIndexTypes> allResults = realm.where(AnnotationIndexTypes.class).findAll();
         assertEquals("All Objects Count", numberOfBlocks * numberOfBlocks * numberOfObjects, allResults.size());
         // distinctive dates
-        RealmResults<AnnotationIndexTypes> distinctDates = allResults.distinct("indexDate");
+        RealmResults<AnnotationIndexTypes> distinctDates = allResults.distinct(AnnotationIndexTypes.FIELD_INDEX_DATE);
         assertEquals("Distinctive Dates", numberOfBlocks, distinctDates.size());
         // distinctive Booleans
-        RealmResults<AnnotationIndexTypes> distinctBooleans = distinctDates.distinct("indexBoolean");
+        RealmResults<AnnotationIndexTypes> distinctBooleans = distinctDates.distinct(AnnotationIndexTypes.FIELD_INDEX_BOOL);
         assertEquals("Distinctive Booleans", 2, distinctBooleans.size());
         // all three results are the same object
         assertTrue(allResults == distinctDates);
@@ -1347,28 +1353,28 @@ public class RealmResultsTests {
     }
 
     @Test
-    public void distinct_withNull() {
+    public void distinct_withNullValues() {
         final long numberOfBlocks = 25;
         final long numberOfObjects = 10; // must be greater than 1
         populateForDistinct(realm, numberOfBlocks, numberOfObjects, true);
 
-        for (String fieldName : new String[]{"Date", "String"}) {
-            RealmResults<AnnotationIndexTypes> distinct = realm.where(AnnotationIndexTypes.class).findAll().distinct("index" + fieldName);
-            assertEquals("index" + fieldName, 1, distinct.size());
+        for (String field : new String[]{AnnotationIndexTypes.FIELD_INDEX_DATE, AnnotationIndexTypes.FIELD_INDEX_STRING}) {
+            RealmResults<AnnotationIndexTypes> distinct = realm.where(AnnotationIndexTypes.class).findAll().distinct(field);
+            assertEquals(field, 1, distinct.size());
         }
     }
 
     @Test
-    public void distinct_noneIndexedFields() {
+    public void distinct_notIndexedFields() {
         final long numberOfBlocks = 25;
         final long numberOfObjects = 10; // must be greater than 1
         populateForDistinct(realm, numberOfBlocks, numberOfObjects, false);
 
-        for (String fieldName : new String[]{"Boolean", "Long", "Date", "String"}) {
+        for (String field : AnnotationIndexTypes.NOT_INDEX_FIELDS) {
             try {
-                realm.where(AnnotationIndexTypes.class).findAll().distinct("notIndex" + fieldName);
-                fail("notIndex" + fieldName);
-            } catch (UnsupportedOperationException ignored) {
+                realm.where(AnnotationIndexTypes.class).findAll().distinct(field);
+                fail(field);
+            } catch (IllegalArgumentException ignored) {
             }
         }
     }
@@ -1387,14 +1393,14 @@ public class RealmResultsTests {
     }
 
     @Test
-    public void testDistinctInvalidTypes() {
+    public void distinct_invalidTypes() {
         populateTestRealm();
 
-        for (String field : new String[]{"columnRealmObject", "columnRealmList", "columnDouble", "columnFloat"}) {
+        for (String field : new String[]{AllTypes.FIELD_REALMOBJECT, AllTypes.FIELD_REALMLIST, AllTypes.FIELD_DOUBLE, AllTypes.FIELD_FLOAT}) {
             try {
                 realm.where(AllTypes.class).findAll().distinct(field);
                 fail(field);
-            } catch (UnsupportedOperationException ignored) {
+            } catch (IllegalArgumentException ignored) {
             }
         }
     }
@@ -1405,10 +1411,10 @@ public class RealmResultsTests {
         final long numberOfObjects = 10; // must be greater than 1
         populateForDistinct(realm, numberOfBlocks, numberOfObjects, true);
 
-        for (String fieldName : new String[]{"Boolean", "Long", "Date", "String"}) {
+        for (String field : AnnotationIndexTypes.INDEX_FIELDS) {
             try {
-                realm.where(AnnotationIndexTypes.class).findAll().distinct(AnnotationIndexTypes.FIELD_OBJECT + ".index" + fieldName);
-                fail("Unsupported Index" + fieldName + " linked field");
+                realm.where(AnnotationIndexTypes.class).findAll().distinct(AnnotationIndexTypes.FIELD_OBJECT + "." + field);
+                fail("Unsupported Index" + field + " linked field");
             } catch (IllegalArgumentException ignored) {
             }
         }
@@ -1420,10 +1426,10 @@ public class RealmResultsTests {
         final long numberOfObjects = 10; // must be greater than 1
         populateForDistinct(realm, numberOfBlocks, numberOfObjects, true);
 
-        for (String fieldName : new String[]{"Boolean", "Long", "Date", "String"}) {
+        for (String field : AnnotationIndexTypes.NOT_INDEX_FIELDS) {
             try {
-                realm.where(AnnotationIndexTypes.class).findAll().distinct(AnnotationIndexTypes.FIELD_OBJECT + ".notIndex" + fieldName);
-                fail("Unsupported notIndex" + fieldName + " linked field");
+                realm.where(AnnotationIndexTypes.class).findAll().distinct(AnnotationIndexTypes.FIELD_OBJECT + "." + field);
+                fail("Unsupported notIndex" + field + " linked field");
             } catch (IllegalArgumentException ignored) {
             }
         }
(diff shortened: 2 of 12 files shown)


realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -1127,6 +1127,23 @@ public boolean isValid() {
...
+     * undefined which object is returned.
+     *
+     * @param fieldName the field name.
+     * @return a non-null {@link RealmResults} containing the distinct objects.
+     * @throws IllegalArgumentException if a field name does not exist.
+     * @throws IllegalArgumentException if a field's type is not supported.
+     * @throws IllegalArgumentException if a field points linked properties.
+     * @throws UnsupportedOperationException if a field is not indexed.
+     */
+    public RealmResult<E> distinctAsync(String fieldName) {
```
dalinaum: RealmResult -> RealmResults

realm/realm-library/src/main/java/io/realm/RealmResults.java
```diff
@@ -578,6 +578,23 @@ public double average(String fieldName) {
...
+     * undefined which object is returned.
+     *
+     * @param fieldName the field name.
+     * @return a non-null {@link RealmResults} containing the distinct objects.
+     * @throws IllegalArgumentException if a field name does not exist.
+     * @throws IllegalArgumentException if a field's type is not supported.
+     * @throws IllegalArgumentException if a field points linked properties.
+     * @throws UnsupportedOperationException if a field is not indexed.
+     */
+    public RealmResult<E> distinctAsync(String fieldName) {
```
dalinaum: Same.

stk1m1:
retest this please

stk1m1:
retest this please

stk1m1:
**What is done**
1. `RealmQuery#distinctAsync(String fieldName)` is a wrapper on `RealmQuery#distinctAsync(long columnIndex)` with some sanity checks.
2. [Positive and negative tests](https://github.com/realm/realm-java/commit/43b389655ef92c6041bc61ac213322be15c2b922#diff-9a34a561cebc828bda5a4f8664d6a1f1R2078) were done against `RealmQuery#distinctAsync(String fieldName)` just as they are done on `RealmQuery#distinct(String fieldName)`. 

**Questions**
1. I'm a big fan of running very frequent instrument tests, and just recently realize that I would get lots of errors about unclosed `Realm` instances on devices whose API are under 23. Is this an expected behavior?
2. I believe we have few ways to dealing with exceptions in JUnit4 style. They are
    - Postfix of `*throws() {` in test function definition.
    - `@Test(expected =`
    - Put `throws Throwable` at the end of a definition.
I'd love to have some contextual clarity on the usage of each.  
3. As I'm implementing `RealmResults#distinctAsync()`, I've realized that there are basically two approaches I can take. The first one being

cmelchior:
1. If you are getting warnings about unclosed Realm instances, that should be fixed as that should not happen. All unit tests should be selfcontained which means they should close all resources before exiting.

2. We just added the description to CONTRIBUTING.md, https://github.com/realm/realm-java/blob/master/CONTRIBUTING.md. If that description is not clear there, we need to change it so it is :)

3. The current implementation of `distinctAsync` in `RealmQuery` is package private and operate on a Table, not a TableView, so you cannot use that directly for `RealmQuery.distinctAsync` as that needs to work on a TableView. I don't think there is any way around doing it the hard way.

4. I cannot exactly remember how the other async queries work in that regard, but the behavior you be similar, e..g `realm.where(Foo.class).findAllAsync().where().equalTo("field", true).findFirstAsync()`. Try it out :)

realm/realm-library/src/androidTest/java/io/realm/RealmQueryTest.java
```diff
@@ -55,15 +67,20 @@
     public final TestRealmConfigurationFactory configFactory = new TestRealmConfigurationFactory();
     @Rule
     public final ExpectedException thrown = ExpectedException.none();
+    @Rule
+    public final RunInLooperThread workerThread = new RunInLooperThread();
```
cmelchior: A better name for this is `looperThread`

realm/realm-library/src/androidTest/java/io/realm/RealmQueryTest.java
```diff
@@ -55,15 +67,20 @@
...
+    private final static long DECADE_MILLIS = 10 * TimeUnit.DAYS.toMillis(365);
+
     protected Realm testRealm;
 
-    private final static long DECADE_MILLIS = 10 * TimeUnit.DAYS.toMillis(365);
+    private Context context;
 
     @Before
     public void setUp() throws Exception {
+        context = InstrumentationRegistry.getInstrumentation().getContext();
```
cmelchior: You shouldn't have any need for the context. Everything should go through the configFactory

realm/realm-library/src/androidTest/java/io/realm/RealmQueryTest.java
```diff
@@ -88,23 +105,23 @@ private void populateTestRealm(int objects) {
...
-        populateTestRealm(TEST_DATA_SIZE);
+    private void populateTestRealm(Realm realm) {
+        populateTestRealm(realm, TEST_DATA_SIZE);
     }
 
     @Test
     public void between() {
         final int TEST_OBJECTS_COUNT = 200;
-        populateTestRealm(TEST_OBJECTS_COUNT);
+        populateTestRealm(testRealm, TEST_OBJECTS_COUNT);
```
cmelchior: `testRealm` should be renamed to `realm`

realm/realm-library/src/androidTest/java/io/realm/RealmQueryTest.java
```diff
@@ -2058,4 +2075,290 @@ public void distinct_invalidTypesLinkedFieldsThrows() {
         } catch (IllegalArgumentException ignored) {
         }
     }
+
+    // distinctAsync
+    private Realm openRealmInstance(String name) {
+        RealmConfiguration realmConfiguration = new RealmConfiguration.Builder(context)
```
cmelchior: Use `configFactory` instead, then you don't have to delete the Realm either.

realm/realm-library/src/androidTest/java/io/realm/RealmQueryTest.java
```diff
@@ -2058,4 +2075,290 @@ public void distinct_invalidTypesLinkedFieldsThrows() {
...
+                    signalClosedRealm.countDown();
+                }
+            }
+        });
+
+        TestHelper.exitOrThrow(executorService, signalCallbackFinished, signalClosedRealm, backgroundLooper, threadAssertionError);
+    }
+
+    @Test
+    public void distinctAsync_withNull () throws Throwable {
```
cmelchior: This and a lot of the other tests can be refactored to use `@RunTestInLooperThread` instead. It will allow you to remove a lot of the boilerplate and make the test safer. You can look in `RealmTests` for examples on how to use it.

stk1m1:
retest this please

stk1m1:
**What is done**
1. `RealmQuery#distinctAsync(String fieldName)` is a series of execution on `TableQuery` that firstly executes `findAll`, gets a `TableView` and, makes the view distinctive. JNI details are [here](https://github.com/realm/realm-java/commit/2eef61092fb6d3faa71d3f3c32b20cfb9151c6f7#diff-9d2a8c0c4b818471db501d6251042a38L145).
2. `ArgumentHolder` has new type `ArgumentsHolder.TYPE_FIND_AND_GET_DISTINCT_VIEW` for the new operation.

**Questions**
1. I think `RealmQuery` is supposed to work on its backing `TableQuery`. Then `RealmResults` is supposed to work on its backing `Table` or `TableView` to get a distinctive `RealmQuery`.
2. `HandlerController` only accepts `RealmQuery` to its **asyncRealmResults** map. What would be the least frictionless way to make it accept `RealmResults` as well?

realm/realm-library/src/main/java/io/realm/RealmResults.java
```diff
@@ -578,6 +589,139 @@ public double average(String fieldName) {
...
+     *
+     * @param fieldName the field name.
+     * @return a non-null {@link RealmResults} containing the distinct objects.
+     * @throws IllegalArgumentException if a field name does not exist.
+     * @throws IllegalArgumentException if a field's type is not supported.
+     * @throws IllegalArgumentException if a field points linked properties.
+     * @throws UnsupportedOperationException if a field is not indexed.
+     */
+    public RealmResults<E> distinctAsync(String fieldName) {
+        realm.checkIfValid();
```
cmelchior: This seems like massively overkill? Just doing `where().distinctAsync()` should be a lot easier?

stk1m1:
**What is done and tried**
1. [`RealmQuery#createQueryFromResultView(RealmResults<E> queryResults, TableOrView table)`](https://github.com/realm/realm-java/commit/1f5521fc081af35600382af29a9608b7cdab0a28#diff-547a7819b14951fe50f7a50fdfa07725R116) is a package private static function that takes `RealmResults` and `TableOrView`. As `RealmQuery` can then take `TableView`, it is now possible to pass `TableView`, rather than `Table`, of `RealmResults`.  
2. In [`RealmQuery#distinctAsync(String fieldName)`](https://github.com/realm/realm-java/blob/1f5521fc081af35600382af29a9608b7cdab0a28/realm/realm-library/src/main/java/io/realm/RealmQuery.java#L1169), one can pass `TableQuery` or `TableOrView` to a newly created empty `RealmResults`. I've tried with `TableOrView` then eventually stuck where `TableQuery` creates a new `TableView` with *HandOver* In *JNI* level.  
What happens, interestingly, then is that `TableQuery` of `TableView` or `Table` doesn't seem to import a newly handed-over `TableView` to the **right** thread. Tests complains that they waits more than 7 secs and fails. I went a step further to copy [the handover importing function](https://github.com/realm/realm-java/blob/master/realm/realm-library/src/main/java/io/realm/internal/TableQuery.java#L506) of `TableQuery` on `TableView` and `Table` and faced same type of failure.  
3. Eventually I went back to passing `TableQuery` to an empty `RealmResults` and everything went right back to where it was. i.e. `RealmResults.distinctAsync()` from `realm.where(Foo.class).findAll().distinctAsync("name")` starts return indistinctive elements.  
4. My conjecture narrows down this. Suppose we pass a `TableQuery` of `TableView` in `RealmQuery#distinctAsync(String fieldName)`. The query represents whatever previous search condition we set to. I don't think that it is necessary to call `query->find_all()` as [it is done](https://github.com/realm/realm-java/commit/3fe0c7fa46b2f8e0eb2d49a06352262d4f1e3678#diff-9d2a8c0c4b818471db501d6251042a38R145) in the current PR.

stk1m1:
@nhachicha I think I need your help.

The goal of this PR is to implement **`realm.where(Foo.class).findAll().distinctAsync("name")`**.
Bear with me about the messy implementation on Java side. It will be straighten up.

**Run Path**
1.[ `RealmResults.distinctAsync(String fieldName)`](https://github.com/realm/realm-java/blob/a1514b221a875fca3790515cd48564d6df5d576b/realm/realm-library/src/main/java/io/realm/RealmResults.java#L609) hands its `TableOrView` to a `RealmQuery` instance and runs [`RealmQuery.distinctAsyncQuery()`](https://github.com/realm/realm-java/blob/a1514b221a875fca3790515cd48564d6df5d576b/realm/realm-library/src/main/java/io/realm/RealmQuery.java#L116). 
2. the `TableQuery` of the handed `TableView` is [saved to `query` field](https://github.com/realm/realm-java/blob/a1514b221a875fca3790515cd48564d6df5d576b/realm/realm-library/src/main/java/io/realm/RealmResults.java#L609).
3. The `TableQuery` of the `TableOrView` then gets a `TableView` from `findAll`, makes the view distinctive, and export the `TableView` via handover. [JNI detail] (https://github.com/realm/realm-java/blob/2eef61092fb6d3faa71d3f3c32b20cfb9151c6f7/realm/realm-jni/src/io_realm_internal_TableQuery.cpp#L145).
4. `ArgumentHolder` has new type `ArgumentsHolder.TYPE_FIND_AND_GET_DISTINCT_VIEW` for the new operation.
5. Then `RealmResults` from step 1, becomes notified from `HandlerController` and imports the distinctive `TableView` via `swapTableViewPointer()`. 

- The caveat to watch out for is to preserve the backing `TableView` of a `RealmResults` like the synchronous cousin [`distinct()`](https://github.com/realm/realm-java/blob/master/realm/realm-library/src/main/java/io/realm/RealmResults.java#L576). 
- The tests associated to `RealmResults.distinctAsync()` are [here](https://github.com/realm/realm-java/blob/a1514b221a875fca3790515cd48564d6df5d576b/realm/realm-library/src/androidTest/java/io/realm/RealmResultsTests.java#L1445).


**Observations**
1. Instrument tests fails with following message. 
```
[COMPLETED_ASYNC_REALM_RESULTS java.lang.ref.WeakReference@67ad72c] , realm:io.realm.HandlerController@25e99fb same versions, using results (RealmResults is not loaded)
[COMPLETED_ASYNC_REALM_RESULTS java.lang.ref.WeakReference@d00c976] , realm:io.realm.HandlerController@dfbeb77 same versions, using results (RealmResults is not loaded)
```
```
io.realm.RealmResultsTests > distinctAsync[PREVIEW - Google Nexus 5X - 6.0.0 - API 23 - 1080x1920 - 6.0] FAILED
	junit.framework.AssertionFailedError: Test took longer than 30 seconds
	at junit.framework.Assert.fail(Assert.java:50)
``` 
AFAIK, a `TableView` resides only in one thread unless it is handed over by `TableQuery`. I think recycling `TableQuery` of a `TableView` might be the source of the problem.     

- If I am to run [`RealmQuery.distinctAsync()`](https://github.com/realm/realm-java/blob/a1514b221a875fca3790515cd48564d6df5d576b/realm/realm-library/src/main/java/io/realm/RealmQuery.java#L1171) with a new instance of `RealmResults` backed by `TableQuery`, I get notified for sure, but don't get distinctive elements at all. The results include the whole set.

Could you take a look at the code and point me what I'm missing?

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -1127,6 +1156,175 @@ public boolean isValid() {
         return realmResults;
     }
 
+    /**
+     * Asynchronously returns a distinct set of objects of a specific class. If the result is
+     * sorted, the first object will be returend in case of multiple occurences, otherwise it is
```
nhachicha: `s/returend/returned`

realm/realm-library/src/main/java/io/realm/RealmResults.java
```diff
@@ -586,6 +586,29 @@ public double average(String fieldName) {
         return this;
     }
 
+    /**
+     * Asynchronously returns a distinct set of objects of a specific class. If the result is
+     * sorted, the first object will be returend in case of multiple occurences, otherwise it is
```
nhachicha: `s/returend/returned/`

nhachicha:
@stk1m1 
I didn't dig deep into the implementation but I think there is a problem with 
`RealmResults#currentTableViewVersion` causing the notification to be skipped even after your query completes (whether it completes with the correct results is another pb you'll have)

This line 
`RealmResults<AnnotationIndexTypes> allBool = asyncRealm.where(AnnotationIndexTypes.class).findAll();`
will call the constructor `RealmResults(BaseRealm realm, TableOrView table, Class<E> classSpec)`
that will init the value of the `currentTableViewVersion` with `table.sync()`

since you didn't modify the table, once your distinct query completes it will invoke `notifyChangeListeners` which will skip 
delivering the notification since `table.sync()` is still the same.

If the callback is not notified then the CountDownLatch is never decremented

You can easily reproduce this by setting break points

stk1m1:
**What's done?**
1. `TableView` [instance copy](https://github.com/realm/realm-java/commit/0e27303f9a84e8c9befcfa49500674ac277ae62c#diff-9d2a8c0c4b818471db501d6251042a38L145) no more.
2. [`RealmResults.distinctAsync()`](https://github.com/realm/realm-java/commit/0e27303f9a84e8c9befcfa49500674ac277ae62c#diff-3a71747f3d5c0ad4ef27cfaeeee36dc3L1466) unit tests clean up. 
3. `RealmResults.distinctAsync()` now generates a new instance of `RealmResults`, getting rid of auxiliary [`RealmQuery.distinctAsync*`](https://github.com/realm/realm-java/commit/a1514b221a875fca3790515cd48564d6df5d576b#diff-547a7819b14951fe50f7a50fdfa07725L1252) families. 
This leads `RealmResults.distinctAsync()` to have different behavior from `RealmResults.distinct()`. 

**What took you so long?**
1. I took some time to test if passing a tableview via HO works fine at around JNI call level & core. (built some core test funcs, monitored JNI level calls, & etc). Core works as expected.
2. I've tracked down what happens is this. When a handed over (imported) tableview gets spawned as POJO in the caller thread, its size changes.   
As it is illustrated in a captured log below, you can see a HO Tableview w/ ptr (ffffffffdf186000) emerges from a worker thread, gets imported to the caller thread with preserved size, and then finally weirdly expends to the size of 6250 as POJO. (It's calling `nativeSize` but you get the idea).
```
TestRunner: started: distinctAsync(io.realm.RealmResultsTests)
findAndGetDistinctViewWithHandover: [NATIVE] Exporting Distinctive TableView size 2
REALM   : [RealmQuery.distinctAsync] : handoverTableViewPointer [ffffffffdf186000] RealmResults [java.lang.ref.WeakReference@35cd24a]
REALM   : [HandlerController.completedAsyncRealmResults] : handoverTableViewPointer [ffffffffdf186000] RealmResults [java.lang.ref.WeakReference@35cd24a]
REALM   : [COMPLETED_ASYNC_REALM_RESULTS java.lang.ref.WeakReference@35cd24a] , realm:io.realm.HandlerController@efd7c6d same versions, using results (RealmResults is not loaded)
Java_io_realm_internal_TableQuery_nativeImportHandoverTableViewIntoSharedGroup: [NATIVE] HandedOver TableView HO[0xffffffffdf186000] size 2
REALM   : [TableView.importHandoverTableView] TableView HO[ffffffffdf186000] size() 6250
REALM   : RealmResults.swapTableViewPointer() table size 6250
REALM   : RealmResults.notifyChangeListeners() listeners is not empty : 1
REALM   : RealmResults.notifyChangeListeners() pendingQuery completed.
REALM   : RealmResults.notifyChangeListeners() currentTableViewVersion : -1 new Table Version : 56484
TestRunner: failed: distinctAsync(io.realm.RealmResultsTests)
```

realm/realm-jni/src/io_realm_internal_TableQuery.cpp
```diff
@@ -127,6 +127,35 @@ static jlong getDistinctViewWithHandover
...
+    if(!ROW_INDEXES_VALID(env, table.get(), start, end, limit))
+        return 0;
+    switch (table->get_column_type(S(columnIndex))) {
+        case type_Bool:
+        case type_Int:
+        case type_DateTime:
+        case type_String: {
+            TableView tableView = query->find_all(S(start), S(end), S(limit));
+            tableView.distinct(S(columnIndex));
+__android_log_print(ANDROID_LOG_DEBUG, __FUNCTION__, "[NATIVE] Exporting Distinctive TableView size %zu\n",tableView.size());
```
kneth: Please use the `TR()` macro.

realm/realm-jni/src/io_realm_internal_TableQuery.cpp
```diff
@@ -127,6 +127,35 @@ static jlong getDistinctViewWithHandover
     return 0;
 }
 
+static jlong findAndGetDistinctViewWithHandover
+        (JNIEnv *env, jlong bgSharedGroupPtr, std::unique_ptr<Query> query, jlong start, jlong end, jlong limit, jlong columnIndex)
+{
+    TableRef table = query->get_table();
+    if (!QUERY_VALID(env, query.get()))
+        return 0;
+    if(!TBL_AND_COL_INDEX_VALID(env, table.get(), columnIndex))
```
kneth: "if(" -> "if ("

realm/realm-jni/src/io_realm_internal_TableQuery.cpp
```diff
@@ -127,6 +127,35 @@ static jlong getDistinctViewWithHandover
...
+        return 0;
+    if(!ROW_INDEXES_VALID(env, table.get(), start, end, limit))
```
kneth: "if(" -> "if ("

realm/realm-jni/src/io_realm_internal_TableQuery.cpp
```diff
@@ -1675,6 +1727,7 @@ JNIEXPORT jlong JNICALL Java_io_realm_internal_TableQuery_nativeImportHandoverTa
         if (SG(callerSharedGrpPtr)->is_attached()) {
             std::unique_ptr<TableView> tableView = SG(callerSharedGrpPtr)->import_from_handover(
                     std::move(handoverTableView));
+__android_log_print(ANDROID_LOG_DEBUG, __FUNCTION__, "[NATIVE] HandedOver TableView HO[0x%llx] size %zu\n", handoverPtr, tableView->size());
```
kneth: Please use the `TR()` macro.

stk1m1:
Two changes to legacy behavior: `DynamicRealm.distinct()/distinctAsync()` and `Realm.distinct()/distinctAsync()` now throws `IllegalArgumentException` instead of `UnsupportedOperationException`.

@realm-java please review.

kneth:
Looks fine with me :+1:

realm/realm-library/src/androidTest/java/io/realm/RealmResultsTests.java
```diff
@@ -16,6 +16,8 @@
 
 package io.realm;
 
+import android.os.Handler;
+import android.os.Looper;
```
zaki50: these 2 imports seem to be unused.

realm/realm-library/src/androidTest/java/io/realm/RealmResultsTests.java
```diff
@@ -29,12 +31,15 @@
 import java.util.Date;
 import java.util.List;
 import java.util.concurrent.Callable;
+import java.util.concurrent.CountDownLatch;
```
zaki50: same

realm/realm-library/src/androidTest/java/io/realm/RealmResultsTests.java
```diff
@@ -29,12 +31,15 @@
...
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
 
+import dalvik.annotation.TestTarget;
```
zaki50: same

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -1127,6 +1122,118 @@ public boolean isValid() {
...
+        });
+
+        realmResults.setPendingQuery(pendingQuery);
+        return realmResults;
+    }
+
+    // Find and validate the column index for the field name used to create a distinctive TableView.
+    static long getAndValidateDistinctColumnIndex(String fieldName, Table table) {
+        // Check empty field name
+        if (fieldName == null || fieldName.equals("")) {
```
zaki50: I prefer `fieldName.isEmpty()` to `fieldName.equals("")`

realm/realm-library/src/androidTest/java/io/realm/RealmResultsTests.java
```diff
@@ -1440,6 +1450,211 @@ public void distinct_invalidTypesLinkedFields() {
...
+        populateForDistinct(realm, numberOfBlocks, numberOfObjects, false);
+
+        realm.where(AnnotationIndexTypes.class).findAll().distinctAsync("doesNotExist");
+    }
+
+    @Test
+    public void distinctAsync_invalidTypes() {
+        populateTestRealm(realm, TEST_DATA_SIZE);
+
+        for (String field : new String[]{"columnRealmObject", "columnRealmList", "columnDouble", "columnFloat"}) {
```
zaki50: It's better to use `AllTypes.FIELD_*` constants.

realm/realm-library/src/androidTest/java/io/realm/RealmResultsTests.java
```diff
@@ -1440,6 +1450,211 @@ public void distinct_invalidTypesLinkedFields() {
...
+        });
+    }
+
+    @Test
+    public void distinctAsync_notIndexedFields() {
+        final long numberOfBlocks = 25;
+        final long numberOfObjects = 10;
+        populateForDistinct(realm, numberOfBlocks, numberOfObjects, false);
+
+        for (String fieldName : new String[]{"Boolean", "Long", "Date", "String"}) {
```
zaki50: It's better to define field name constants in `AnnotationIndexTypes`.

And, how about defining `public static final String[] INDEXED_FIELDS = {INDEX_STRING, INDEX_INT, ...}` and  `public static final String[] NON_INDEXED_FIELDS = {NON_INDEX_STRING, NON_INDEX_INT, ...}` in `AnnotationIndexTypes`.

zaki50:
@stk1m1 Reviewed.
I put some comments.

stk1m1:
@zaki50 Thanks for the pointers! I'll get on them!

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
 * Added RealmQuery.distinct() and RealmResults.distinct() (#1568).
 * Fixed bug when multiple calls of RealmResults.distinct() causes to return wrong results (#2198).
 * RealmResults.size() now returns Integer.MAX_VALUE when actual size is greater than Integer.MAX_VALUE (#2129).
+* Added RealmQuery.distinctAsync() and RealmResults.distinctAsync() (#2118).
```
cmelchior: We need to add a BREAKING CHANGE section, outlining that we now throw IllegalArgumentException instead of UnsupportedOperationException.

realm/realm-library/src/androidTest/java/io/realm/RealmResultsTests.java
```diff
@@ -1440,6 +1450,211 @@ public void distinct_invalidTypesLinkedFields() {
...
+    @Test
+    @RunTestInLooperThread
+    public void distinctAsync() throws Throwable {
+        final AtomicInteger changeListenerCalled = new AtomicInteger(4);
+        final Realm realm = looperThread.realm;
+        final long numberOfBlocks = 25;
+        final long numberOfObjects = 10; // must be greater than 1
+        populateForDistinct(realm, numberOfBlocks, numberOfObjects, false);
+
+        final RealmResults<AnnotationIndexTypes> distinctBool = realm.where(AnnotationIndexTypes.class).findAll().distinctAsync("indexBoolean");
```
cmelchior: Use `AnnotationIndexTypes.FIELD_INDEXEDBOOLEAN` or similar. All fields should ideally be referenced through constants.

realm/realm-library/src/androidTest/java/io/realm/RealmResultsTests.java
```diff
@@ -1440,6 +1450,211 @@ public void distinct_invalidTypesLinkedFields() {
...
+        for (String fieldName : new String[]{"Boolean", "Long", "Date", "String"}) {
+            try {
+                realm.where(AnnotationIndexTypes.class).findAll().distinctAsync("notIndex" + fieldName);
+                fail("notIndex" + fieldName);
+            } catch (IllegalArgumentException ignored) {
+            }
+        }
+    }
+
+    @Test(expected = IllegalArgumentException.class)
```
cmelchior: Only use this for 1-line tests. Either use `ExceptedException` or `try/catch` instead as outlined in CONTRIBUTING.md.

realm/realm-library/src/androidTest/java/io/realm/RealmResultsTests.java
```diff
@@ -1440,6 +1450,211 @@ public void distinct_invalidTypesLinkedFields() {
...
+        for (String fieldName : new String[]{"Boolean", "Long", "Date", "String"}) {
+            try {
+                realm.where(AnnotationIndexTypes.class).findAll().distinctAsync(AnnotationIndexTypes.FIELD_OBJECT + ".index" + fieldName);
+                fail("Unsupported Index" + fieldName + " linked field");
+            } catch (IllegalArgumentException ignored) {
+            }
+        }
+    }
+
+    @Test(expected = IllegalArgumentException.class)
```
cmelchior: Same

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -995,17 +995,16 @@ boolean contains(Class<? extends RealmObject> clazz) {
      * @param clazz the Class to get objects of.
      * @param fieldName the field name.
      * @return a non-null {@link RealmResults} containing the distinct objects.
-     * @throws IllegalArgumentException if a field name does not exist or the field is not indexed.
+     * @throws IllegalArgumentException if a field name is an empty string.
```
cmelchior: All these seems kinda excessive. In other methods we usually only have 1 throws for each type. Perhaps something like "@throws IllegalArgumentException if a field is `null`, doesn't exists, is not indexed or the field points to a linked field."

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -1020,22 +1019,15 @@ boolean contains(Class<? extends RealmObject> clazz) {
      * @return immediately an empty {@link RealmResults}. Users need to register a listener
      *      {@link io.realm.RealmResults#addChangeListener(RealmChangeListener)} to be notified when the query
      *      completes.
-     * @throws IllegalArgumentException if a field name does not exist or the field is not indexed.
+     * @throws IllegalArgumentException if a field name is an empty string.
```

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -1127,6 +1122,118 @@ public boolean isValid() {
...
+    /**
+     * Asynchronously returns a distinct set of objects of a specific class. If the result is
+     * sorted, the first object will be returned in case of multiple occurrences, otherwise it is
+     * undefined which object is to return.
+     *
+     * @param fieldName the field name.
+     * @return immediately a {@link RealmResults}. Users need to register a listener
+     *          {@link io.realm.RealmResults#addChangeListener(RealmChangeListener)} to be notified
+     *          when the query completes.
+     * @throws IllegalArgumentException if a field name is an empty string.
```

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -1127,6 +1122,118 @@ public boolean isValid() {
         return realmResults;
     }
 
+    /**
+     * Asynchronously returns a distinct set of objects of a specific class. If the result is
+     * sorted, the first object will be returned in case of multiple occurrences, otherwise it is
+     * undefined which object is to return.
```
cmelchior: is to return -> is returned.

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -1127,6 +1122,118 @@ public boolean isValid() {
...
+     *
+     * @param fieldName the field name.
+     * @return immediately a {@link RealmResults}. Users need to register a listener
+     *          {@link io.realm.RealmResults#addChangeListener(RealmChangeListener)} to be notified
```
cmelchior: Spacing seems wrong here. It should match the first line?

realm/realm-library/src/main/java/io/realm/RealmResults.java
```diff
@@ -567,19 +567,20 @@ public double average(String fieldName) {
...
+     * object is to return.
      *
      * @param fieldName the field name.
      * @return a non-null {@link RealmResults} containing the distinct objects.
+     * @throws IllegalArgumentException if a field name is an empty string.
      * @throws IllegalArgumentException if a field name does not exist.
      * @throws IllegalArgumentException if a field's type is not supported.
-     * @throws IllegalArgumentException if a field points linked properties.
-     * @throws UnsupportedOperationException if a field is not indexed.
+     * @throws IllegalArgumentException if a field name points to linked fields.
```
cmelchior: Same, combine all IllegalArgument descriptions

realm/realm-library/src/main/java/io/realm/RealmResults.java
```diff
@@ -590,6 +591,25 @@ public double average(String fieldName) {
...
+    /**
+     * Asynchronously returns a distinct set of objects of a specific class. If the result is
+     * sorted, the first object will be returned in case of multiple occurrences, otherwise it is
+     * undefined which object is to return.
+     *
+     * @param fieldName the field name.
+     * @return immediately a {@link RealmResults}. Users need to register a listener
+     *          {@link io.realm.RealmResults#addChangeListener(RealmChangeListener)} to be notified
+     *          when the query completes.
+     * @throws IllegalArgumentException if a field name is an empty string.
```
cmelchior: Same

realm/realm-library/src/main/java/io/realm/RealmResults.java
```diff
@@ -590,6 +591,25 @@ public double average(String fieldName) {
...
     }
 
+    /**
+     * Asynchronously returns a distinct set of objects of a specific class. If the result is
+     * sorted, the first object will be returned in case of multiple occurrences, otherwise it is
+     * undefined which object is to return.
+     *
+     * @param fieldName the field name.
+     * @return immediately a {@link RealmResults}. Users need to register a listener
+     *          {@link io.realm.RealmResults#addChangeListener(RealmChangeListener)} to be notified
```
cmelchior: Spacing

cmelchior:
Apart from comments regarding javadoc/unit tests I think this looks good.

stk1m1:
@zaki50 @cmelchior All the reviewed points are applied. ;)

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 * BREAKING CHANGE: DynamicRealm.executeTransaction() now directly throws any RuntimeException instead of wrapping it in a RealmException (#1682).
 * BREAKING CHANGE: DynamicRealm.executeTransaction() now throws IllegalArgumentException instead of silently accepting a null Transaction object.
 * BREAKING CHANGE: String setters now throws IllegalArgumentException instead of RealmError for invalid surrogates. 
+* BREAKING CHANGE: DynamicRealm.distinct()/distinctAsync() and Realm.distinct()/distinctAsync() throw IllegalArgumentException for invalid type or unindexed field.
```
cmelchior: now throw IllegalArgumentException instead of UnsupportedOperationException

realm/realm-library/src/main/java/io/realm/DynamicRealm.java
```diff
@@ -271,26 +267,15 @@ static DynamicRealm createInstance(RealmConfiguration configuration) {
...
      * @param className the Class to get objects of.
      * @param fieldName the field name.
-     * @return immediately an empty {@link RealmResults}. Users need to register a listener
-     * {@link io.realm.RealmResults#addChangeListener(RealmChangeListener)} to be notified
-     * when the query completes.
-     * @throws IllegalArgumentException if a field name does not exist or the field is not indexed.
+     * @return immediately a {@link RealmResults}. Users need to register a listener
+     * {@link io.realm.RealmResults#addChangeListener(RealmChangeListener)} to be notified when the
+     * query completes.
+     * @throws IllegalArgumentException if a field is null, does not exist, is in unsupported type,
```
cmelchior: in -> an

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -995,17 +995,13 @@ boolean contains(Class<? extends RealmObject> clazz) {
      * @param clazz the Class to get objects of.
      * @param fieldName the field name.
      * @return a non-null {@link RealmResults} containing the distinct objects.
-     * @throws IllegalArgumentException if a field name does not exist or the field is not indexed.
+     * @throws IllegalArgumentException if a field is null, does not exist, is in unsupported type,
```
cmelchior: in -> an

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -1018,24 +1014,14 @@ boolean contains(Class<? extends RealmObject> clazz) {
      * @param clazz the Class to get objects of.
      * @param fieldName the field name.
      * @return immediately an empty {@link RealmResults}. Users need to register a listener
-     *      {@link io.realm.RealmResults#addChangeListener(RealmChangeListener)} to be notified when the query
-     *      completes.
-     * @throws IllegalArgumentException if a field name does not exist or the field is not indexed.
+     * {@link io.realm.RealmResults#addChangeListener(RealmChangeListener)} to be notified when the
+     * query completes.
+     * @throws IllegalArgumentException if a field is null, does not exist, is in unsupported type,
```
cmelchior: in -> an

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -1095,26 +1096,17 @@ public boolean isValid() {
...
+     * object will be returned in case of multiple occurrences, otherwise it is undefined which
      * object is returned.
      *
      * @param fieldName the field name.
      * @return a non-null {@link RealmResults} containing the distinct objects.
-     * @throws IllegalArgumentException if a field name does not exist.
-     * @throws IllegalArgumentException if a field's type is not supported.
-     * @throws IllegalArgumentException if a field points linked properties.
-     * @throws UnsupportedOperationException if a field is not indexed.
+     * @throws IllegalArgumentException if a field is null, does not exist, is in unsupported type,
```
cmelchior: in -> an

realm/realm-library/src/main/java/io/realm/DynamicRealm.java
```diff
@@ -249,17 +249,13 @@ static DynamicRealm createInstance(RealmConfiguration configuration) {
      * @param className the Class to get objects of.
      * @param fieldName the field name.
      * @return A non-null {@link RealmResults} containing the distinct objects.
-     * @throws IllegalArgumentException if a field name does not exist or the field is not indexed.
+     * @throws IllegalArgumentException if a field is null, does not exist, is in unsupported type,
```
cmelchior: in -> an

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -1127,6 +1119,115 @@ public boolean isValid() {
...
+    /**
+     * Asynchronously returns a distinct set of objects of a specific class. If the result is
+     * sorted, the first object will be returned in case of multiple occurrences, otherwise it is
+     * undefined which object is returned.
+     *
+     * @param fieldName the field name.
+     * @return immediately a {@link RealmResults}. Users need to register a listener
+     * {@link io.realm.RealmResults#addChangeListener(RealmChangeListener)} to be notified when the
+     * query completes.
+     * @throws IllegalArgumentException if a field is null, does not exist, is in unsupported type,
```
cmelchior: in -> an

realm/realm-library/src/main/java/io/realm/RealmResults.java
```diff
@@ -590,6 +588,22 @@ public double average(String fieldName) {
...
+    /**
+     * Asynchronously returns a distinct set of objects of a specific class. If the result is
+     * sorted, the first object will be returned in case of multiple occurrences, otherwise it is
+     * undefined which object is returned.
+     *
+     * @param fieldName the field name.
+     * @return immediately a {@link RealmResults}. Users need to register a listener
+     * {@link io.realm.RealmResults#addChangeListener(RealmChangeListener)} to be notified when the
+     * query completes.
+     * @throws IllegalArgumentException if a field is null, does not exist, is in unsupported type,
```
cmelchior: in -> an

cmelchior:
Apart from some minor spelling in the Javadoc, this looks :+1:

stk1m1:
retest this please

PR #2121, opened 2016-01-20
Removing LTO from debugging builds

kneth:
Using a locally build core, it is possible to single step into C++. You will have to add the following two lines to `realm.properties`:

    debugBuild=true
    stripSymbols=false

@cmelchior @emanuelez @beeender

### DIFF
--- a/realm/realm-jni/build.gradle
+++ b/realm/realm-jni/build.gradle
@@ -2,7 +2,7 @@ import java.security.MessageDigest
 
 ext.coreVersion = '0.95.7'
 // empty or comment out this to disable hash checking
-//ext.coreSha256Hash = '11c82f5da9b43b62fbf55fbd3aa61dc6956bc3483bbaa5f69262018144c8cc3c'
+ext.coreSha256Hash = '11c82f5da9b43b62fbf55fbd3aa61dc6956bc3483bbaa5f69262018144c8cc3c'
 ext.forceDownloadCore =
         project.hasProperty('forceDownloadCore') ? project.getProperty('forceDownloadCore').toBoolean() : false
 // gcc is default for the NDK. It also produces smaller binaries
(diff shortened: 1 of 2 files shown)


realm/realm-jni/build.gradle
```diff
@@ -2,7 +2,7 @@ import java.security.MessageDigest
 
 ext.coreVersion = '0.95.7'
 // empty or comment out this to disable hash checking
-ext.coreSha256Hash = '11c82f5da9b43b62fbf55fbd3aa61dc6956bc3483bbaa5f69262018144c8cc3c'
+//ext.coreSha256Hash = '11c82f5da9b43b62fbf55fbd3aa61dc6956bc3483bbaa5f69262018144c8cc3c'
```
cmelchior: Is this a mistake?

realm/realm-jni/build.gradle
```diff
@@ -15,7 +15,11 @@ ext.stripSymbols = project.hasProperty('stripSymbols') ? project.getProperty('st
 // core source code.
 ext.coreSourcePath = project.hasProperty('coreSourcePath') ? project.getProperty('coreSourcePath') : null
 
-def commonCflags = [ '-Os', '-std=c++11', '-ffunction-sections', '-fdata-sections', '-flto' ]
+def commonCflags = [ '-Os', '-std=c++11' ]
+// LTO and debugging don't play well together
+if (!ext.debugBuild) {
+    commonCflags += [ '-ffunction-sections', '-fdata-sections', '-flto' ]
```
emanuelez: should visibility hidden be here?

kneth:
@emanuelez @cmelchior Thanks for the comments. It is work-in-progress, and I merely wished to show you how to do native debugging.

kneth:
Since migrating to the new experimental plugin seems to be too hard (other plugins won't work together with it), this is what we can do to get native debugging work.

@realm/java

dalinaum:
:+1:

beeender:
:+1:

emanuelez:
Can we make it so that if a build a debug one then symbols are kept by default?

kneth:
@emanuelez You mean, that `buildDebug = true` implies `stripSymbols = false`?

emanuelez:
Yes

beeender:
@emanuelez @kneth It has already been like that.

```
 if (ext.debugBuild) {
     // Debug build should never strip symbols
     ext.stripSymbols = false
 }
```

PR #2124, opened 2016-01-20
Enable proper iterator behavior

### LINKED ISSUE #602: Fatal crash in for each loop
Hi, I have issue similar to https://github.com/realm/realm-java/issues/484 and to https://github.com/realm/realm-java/issues/589 . I trying to run query in cycle, check if this record not exists, then create it. Before ```for/each cycle``` at the same thread and the same realm instance query works fine. But when we entered to cycle it crashed. I use ```realm library version 0.74.0, Android studio 1.0rc```.
```
Song model:
```
```
public class Song extends RealmObject {
    private String id;
    private String ownerId;
    private String artist;
    private String title;
    private int duration;
    private String url;
    private String lyrics_id;
    private int genreId;
    private String albumId;
    private int index;
    private boolean needToRemove;
**** and their getters and setters ****
}
```

```
Part of the code where it crash:
```
```
{
 List<Song> songs = lists[0];
 Realm workerRealm = getRealmInstanceForWorkerThread();
 String myId = PreferenceHelper.getUserId();
 // Mark all songs need to remove true
 RealmResults<Song> songRealmList = workerRealm.where(Song.class)
          .equalTo("ownerId", myId)
          .findAll();                                         // Works fine
 workerRealm.beginTransaction();
 for (Song song : songRealmList) {
    song.setNeedToRemove(true);         // Works fine
 }
 for (Song song : songs) {                    // Entering to cycle
   // Check if the record exist
  Song querySong = workerRealm.where(Song.class)
         .beginGroup()
         .equalTo("id", song.getId())
         .equalTo("ownerId", song.getOwnerId())
         .endGroup()
         .findFirst();                                 // And here it crash              

 if (querySong == null) {
   querySong = workerRealm.createObject(Song.class);
 }
 querySong.setOwnerId(song.getOwnerId());
 querySong.setUrl(song.getUrl());
 querySong.setLyrics_id(song.getLyrics_id());
 querySong.setGenreId(song.getGenreId());
 querySong.setDuration(song.getDuration());
 querySong.setTitle(song.getTitle());
 querySong.setId(song.getId());
 querySong.setAlbumId(song.getAlbumId());
 querySong.setArtist(song.getArtist());
 querySong.setNeedToRemove(song.isNeedToRemove());
 querySong.setIndex(song.getIndex());
}
```
Error message: ``` Method threw 'java.lang.RuntimeException' exception. Cannot evaluate io.realm.SongRealmProxy.toString() ```
  ```A/libc﹕ Fatal signal 11 (SIGSEGV), code 1, fault addr 0xa1e25ffc in tid -2049854000 (UpdateMultipleO)```

```UpdateMultipleO``` - it is my method where I do this( full name ```updateMultipleOwnSongsAndAlbumsInDB```).

### LINKED ISSUE #640: ConcurrentModificationException when using iterators methods
Hi, I have such exception even I using iterators methods:
Realm version 0.75.1, Android studio 1.0rc2
```
Caused by: java.util.ConcurrentModificationException: No outside changes to a Realm is allowed while iterating a RealmResults. Use iterators methods instead.
            at io.realm.RealmResults.assertRealmIsStable(RealmResults.java:405)
            at io.realm.RealmResults.access$200(RealmResults.java:44)
            at io.realm.RealmResults$RealmResultsIterator.hasNext(RealmResults.java:422)
            at com.coachesdirectory.coach.db.RealmHelper$5.doInBackground(RealmHelper.java:360)
            at com.coachesdirectory.coach.db.RealmHelper$5.doInBackground(RealmHelper.java:347)
            at android.os.AsyncTask$2.call(AsyncTask.java:288)
            at java.util.concurrent.FutureTask.run(FutureTask.java:237)
            at android.os.AsyncTask$SerialExecutor$1.run(AsyncTask.java:231)
            at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1112)
            at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:587)
            at java.lang.Thread.run(Thread.java:818)
```

Here is part of my code:
```
****AsyncTask*****
List<SchoolResult> favoriteSchools = params[0];
Realm realm = Realm.getInstance(context);
realm.beginTransaction();

// Set all schools to remove
RealmResults<RealmSchoolResult> markToRemove = realm.where(RealmSchoolResult.class)
                        .equalTo("userAccountId", userId)
                        .findAll();
Iterator<RealmSchoolResult> schoolResultIterator = markToRemove.iterator();
while (schoolResultIterator.hasNext()) { // Crash at this line
  RealmSchoolResult realmSchoolResult = schoolResultIterator.next();
  realmSchoolResult.setNeedRemove(true);
}
```

### LINKED ISSUE #2051: No notifications triggered after sequence of disabling and enabling auto-refresh
I have following scenario: I create some realm results by runing a query
`mRealmItems = mRealm.where(RealmItem.class).findAll();`
and then add to it listener
`mRealmItems.addChangeListener(this);`
Listener makes notification when new item is added. After adding some items i disable auto-refresh:
`mRealm.setAutoRefresh(false)` and do some stuff that I dont want to be notified about.
Next i enable auto-notifications again:
`mRealm.setAutoRefresh(true);`, however after adding new item no notifications are triggered.

My assumption was that as long as I did not removed listener and auto-refresh is enabled the notifications will be trigerred. Unfortunately in this use case they are not. 

Is it intended behavior? It seems like bug to me.

### PR DESCRIPTION
**Pull request by cmelchior**
Closes #602 
Closes #640 
Closes #2051 

This PR changes the semantics of when RealmResults auto-refresh. In order to enable standard iterator behaviour we have to prevent the `size` of RealmResults to change automatically. This PR disables `sync_if_needed` for RealmResults in all cases, except when manually calling `Realm.refresh()` or receiving a change notification from another thread through the looper. It is specifically disabled when committing on the same thread.

This changes the semantics of RealmResults so there now is a chance it might contain deleted objects if one was deleted in the same looper event, but it will ensure that size() is stable for that duration as well.

This change is motivated by the fact that the first behaviour is much more common than relying on RealmResults _always_ being up to date.

The HandlerController has been refactored in the process as it was already keeping track of all RealmResults. Is is now also responsible for refreshing RealmResults when appropriate. 

TODO
- [x] Refactor HandlerController to support explicit refresh.
- [x] Create Iterator for RealmResults + unit tests.
- [x] Create ListIterator for RealmResults + unit tests
- [x] Create Iterator for RealmList + unit tests
- [x] Create ListIterator for RealmList + unit tests
- [x] Unit tests for the behaviour or refresh, especially with async results. 
- [x] Update website documentation
- [x] Test aggregate methods on a RealmResult with removed results.
- [x] https://github.com/realm/realm-core/issues/1509
- [x] Blocked pending RealmCollection so the behaviour of `remove` is the same.

### CODE REVIEW

**Comment by Saketme** [comment]
Does this mean that RealmResults will now not auto-refresh itself when the items under the original query change?

**Comment by cmelchior** [comment]
Not within the same loop event, no. It will still be refreshed at the beginning of the run loop, if you call `refresh` or another thread makes any changes.

**Comment by Saketme** [comment]
Interesting :+1:

**Comment by renam** [comment]
is there a timeline for this PR to be merged in? I have a few bugs in my project that seem to be related with this. I can implement workarounds but I would prefer not to.

**Comment by cmelchior** [comment]
It is under active development, but there is no deadline. You should be able to use the work-around describe here though: https://github.com/realm/realm-java/issues/640#issuecomment-103798064

**Comment by cmelchior** [comment]
This is blocked until we have merged the RealmCollection interface as we need to have uniform behaviour of `remove()` for iterators to have a sane behaviour.

**Comment by cmelchior** [comment]
retest this please

**Comment by cmelchior** [comment]
Ready for review @realm/java

File: realm/realm-jni/src/io_realm_internal_tableview.cpp [hunk 1]
```diff
@@ -34,7 +34,12 @@ inline bool view_valid_and_in_sync(JNIEnv* env, jlong nativeViewPtr) {
             ThrowException(env, TableInvalid, "The Realm has been closed and is no longer accessible.");
             return false;
         }
-        TV(nativeViewPtr)->sync_if_needed();
+        if (!TV(nativeViewPtr)->is_in_sync() && TV(nativeViewPtr)->depends_on_deleted_linklist()) {
```
kneth: If a table view doesn't have any links list, will `depends_on_delete_linklist()` return true unconditionally? @rrrlasse, can you us a hint?

realm/realm-jni/src/io_realm_internal_tableview.cpp
```diff
@@ -165,9 +170,11 @@ JNIEXPORT jlong JNICALL Java_io_realm_internal_TableView_nativeGetSourceRowIndex
...
         if (!VIEW_VALID_AND_IN_SYNC(env, nativeViewPtr))
-            return npos;
+            return -1;
         if (!ROW_INDEX_VALID(env, TV(nativeViewPtr), rowIndex))
-            return npos;
+            return -1;
+        if (!TV(nativeViewPtr)->is_row_attached(rowIndex))
+            return -1;
     } CATCH_STD()
     return TV(nativeViewPtr)->get_source_ndx(S(rowIndex));   // noexcept
```
kneth: The checks above should be outside the `try`/`catch` block, and the call of `get_source_ndx()` should be guarded by the `try`/`catch` block.

realm/realm-library/src/androidTest/java/io/realm/CollectionTests.java
```diff
@@ -84,20 +84,22 @@ protected void populateRealm(Realm realm, int objects) {
...
-            nonLatinFieldNames.setΔέλτα(i);
-        }
-
-        // Add all items to the RealmList on the first object
-        AllJavaTypes firstObj = realm.where(AllJavaTypes.class).equalTo(AllJavaTypes.FIELD_LONG, 0).findFirst();
-        RealmResults<AllJavaTypes> listData = realm.where(AllJavaTypes.class).findAllSorted(AllJavaTypes.FIELD_LONG, Sort.ASCENDING);
-        RealmList<AllJavaTypes> list = firstObj.getFieldList();
-        for (int i = 0; i < listData.size(); i++) {
-            list.add(listData.get(i));
+        if (objects > 0) {
```
kneth: Why is this needed? If `objects <= 0`, no iterations are executed, right?

realm/realm-library/src/androidTest/java/io/realm/OrderedRealmCollectionIteratorTests.java
```diff
@@ -0,0 +1,964 @@
...
+        collection = realm.where(AllJavaTypes.class).lessThan(AllJavaTypes.FIELD_LONG, TEST_SIZE).findAll();
+        assertEquals(TEST_SIZE, collection.size());
+        realm.beginTransaction();
+        for (int i = 0; i < collection.size(); i++) {
+            AllJavaTypes obj = collection.get(i);
+            obj.setFieldLong(obj.getFieldLong() + TEST_SIZE);
+        }
+        realm.commitTransaction();
+
+        // Verify that all elements where modified
```
kneth: "where" -> "were"

realm/realm-library/src/androidTest/java/io/realm/OrderedRealmCollectionIteratorTests.java
```diff
@@ -0,0 +1,964 @@
...
+
+        collection = realm.where(AllJavaTypes.class).lessThan(AllJavaTypes.FIELD_LONG, TEST_SIZE).findAll();
+        assertEquals(TEST_SIZE, collection.size());
+        for (AllJavaTypes obj : collection) {
+            realm.beginTransaction();
+            obj.setFieldLong(obj.getFieldLong() + TEST_SIZE);
+            realm.commitTransaction();
+        }
+
+        // Verify that all elements where modified
```
kneth: "where" -> "were"

realm/realm-library/src/androidTest/java/io/realm/OrderedRealmCollectionIteratorTests.java
```diff
@@ -0,0 +1,964 @@
...
+        collection = realm.where(AllJavaTypes.class).lessThan(AllJavaTypes.FIELD_LONG, TEST_SIZE).findAll();
+        assertEquals(TEST_SIZE, collection.size());
+        for (int i = 0; i < collection.size(); i++) {
+            realm.beginTransaction();
+            AllJavaTypes obj = collection.get(i);
+            obj.setFieldLong(obj.getFieldLong() + TEST_SIZE);
+            realm.commitTransaction();
+        }
+
+        // Verify that all elements where modified
```
kneth: search and replace? :-)

realm/realm-library/src/androidTest/java/io/realm/RealmTests.java
```diff
@@ -2165,26 +2165,33 @@ public void run() {
         }
     }
 
-    // FIXME HandlerThread
+    // This test assure that calling refresh will not trigger local listeners until after the Looper receives a
```
kneth: "assure" -> "assures"

realm/realm-library/src/main/java/io/realm/RealmList.java
```diff
@@ -724,4 +760,174 @@ public String toString() {
...
+                lastRet = -1;
+                expectedModCount = modCount;
+            } catch (IndexOutOfBoundsException e) {
+                throw new ConcurrentModificationException();
+            }
+        }
+
+        final void checkConcurrentModification() {
+            // A Realm ListView is backed by the original Table and not a TableView, this means
+            // that all changes are reflected immediately. It is therefor not possible to use
```
kneth: "therefor" -> "therefore"

realm/realm-jni/src/io_realm_internal_tableview.cpp
cmelchior: This is the pattern we use elsewhere, also `get_source_ndx` is marked as noexcept.
I don't mind changing it, I just assumed there was a reason our `*_VALID` methods are inside a `try/catch`?

realm/realm-library/src/main/java/io/realm/RealmList.java
```diff
@@ -724,4 +760,174 @@ public String toString() {
...
+            // the same version pinning trick we use for RealmResults (avoiding calling sync-if-needed)
```
kneth: "sync-if-needed" -> "sync_if_needed"

realm/realm-library/src/main/java/io/realm/RealmList.java
```diff
@@ -724,4 +760,174 @@ public String toString() {
...
+            // Fortunately a LinkView does not change unless manually altered (unlike RealmResults)
+            // So therefor it should be acceptable to use the same heuristic as a normal AbstractList
```
kneth: "therefor" -> "therefore"

realm/realm-library/src/androidTest/java/io/realm/CollectionTests.java
cmelchior: It is because the empty collection have to be handled in a special way in order to create an empty RealmList.

realm/realm-library/src/main/java/io/realm/RealmList.java
```diff
@@ -724,4 +760,174 @@ public String toString() {
...
+        }
+    }
+
+    private class RealmListItr extends RealmItr implements ListIterator<E> {
+
+        RealmListItr(int index) {
+            if (index >= 0 && index <= size()) {
+                cursor = index;
+            } else {
+                throw new IndexOutOfBoundsException("Starting location must be a valid index: [0, " + (size() - 1) + "]. Yours was " + index);
```
kneth: "Yours" -> "Index"

realm/realm-library/src/main/java/io/realm/RealmResults.java
```diff
@@ -78,12 +77,15 @@
...
     private static final long TABLE_VIEW_VERSION_NONE = -1;
-    private long currentTableViewVersion = TABLE_VIEW_VERSION_NONE;
 
+    private long currentTableViewVersion = TABLE_VIEW_VERSION_NONE;
     private final TableQuery query;
     private final List<RealmChangeListener> listeners = new CopyOnWriteArrayList<RealmChangeListener>();
     private Future<Long> pendingQuery;
-    private boolean isCompleted = false;
+    private boolean asyncQueryCompleted = false;
+    // Keep track of changes to the RealmResult. Is updated after a call `syncIfNeeded. Calling notifyListeners will
```
kneth: "\`syncIfNeeded" -> "\`syncIfNeeded()\`"

realm/realm-library/src/main/java/io/realm/internal/TableOrView.java
```diff
@@ -372,10 +382,13 @@
...
 
     /**
-     * Syncs the tableview with the underlying table data. It is not required to call this explicitly, all other API
-     * methods will automatically sync the view as well.
+     * Syncs the tableview with the underlying table data. This is effectively the same as rerunning the query, so it
+     * should not be called on TableViews created by an async query.
+     *
+     * This method gets automatically called when calling {@link Realm#refresh()} or when another threads updates
+     * the Realm, but it will _not_ be called the same thread commits a transaction.
      *
```
stk1m1: `another threads` ->  `another thread`
`called the same thread commits` -> `called on the same thread committing`

realm/realm-library/src/main/java/io/realm/RealmResults.java
```diff
@@ -723,54 +727,54 @@ public boolean addAll(Collection<? extends E> collection) {
...
+        }
 
-        protected void assertRealmIsStable() {
-            long version = table.sync();
-            if (tableViewVersion > -1 && version != tableViewVersion) {
-                throw new ConcurrentModificationException("No outside changes to a Realm is allowed while iterating a RealmResults. Use iterators methods instead.");
+        protected void checkRealmIsStable() {
+            long version = table.getVersion();
+            // All changes inside a write transaction will continuously update the table version, and we can
+            // thus not depend on the tableVersion heuristic in that case .
```
kneth: "case ." -> "case."

realm/realm-library/src/main/java/io/realm/RealmResults.java
```diff
@@ -723,54 +727,54 @@ public boolean addAll(Collection<? extends E> collection) {
...
+            throw new UnsupportedOperationException("remove() is not supported by RealmResults iterators.");
+        }
 
-        protected void assertRealmIsStable() {
-            long version = table.sync();
-            if (tableViewVersion > -1 && version != tableViewVersion) {
-                throw new ConcurrentModificationException("No outside changes to a Realm is allowed while iterating a RealmResults. Use iterators methods instead.");
+        protected void checkRealmIsStable() {
+            long version = table.getVersion();
+            // All changes inside a write transaction will continuously update the table version, and we can
```
kneth: "Any changes within a write transaction will immediately update the table version"

realm/realm-library/src/main/java/io/realm/RealmResults.java
```diff
@@ -781,26 +785,40 @@ protected void assertRealmIsStable() {
...
+        /**
+         * Unsupported by RealmResults iterators.
+         *
+         * @throws UnsupportedOperationException, always.
+         */
         @Override
+        @Deprecated
         public void add(E object) {
-            throw new RealmException("Adding elements not supported. Use Realm.createObject() instead.");
+            throw new UnsupportedOperationException("Adding elements is not supported. Use Realm.createObject() instead.");
```
kneth: "is" -> "are"

kneth: "Adding an element is"

realm/realm-library/src/main/java/io/realm/RealmResults.java
```diff
@@ -810,25 +828,24 @@ public E previous() {
...
-            throw new RealmException("Replacing elements not supported.");
-        }
-
         /**
-         * Removes the RealmObject at the current position from both the list and the underlying Realm.
+         * Unsupported by RealmResults iterators.
          *
-         * WARNING: This method is currently disabled and will always throw an
-         * {@link io.realm.exceptions.RealmException}
+         * @throws UnsupportedOperationException, always.
```
kneth: "always" -> "unconditionally"

realm/realm-library/src/main/java/io/realm/RealmResults.java
```diff
@@ -810,25 +828,24 @@ public E previous() {
...
          */
         @Override
-        public void remove() { throw new RealmException("Removing elements not supported."); }
+        @Deprecated
+        public void set(E object) {
+            throw new UnsupportedOperationException("Replacing elements not supported.");
```
kneth: "Replacing an element is not supported."

realm/realm-library/src/main/java/io/realm/RealmResults.java
```diff
@@ -859,7 +876,7 @@ void setPendingQuery(Future<Long> pendingQuery) {
             // the query completed before RealmQuery
             // had a chance to call setPendingQuery to register the pendingQuery (used btw
```
kneth: remove "btw"

realm/realm-library/src/main/java/io/realm/RealmResults.java
```diff
@@ -891,7 +908,7 @@ public boolean load() {
         } else {
             // doesn't guarantee to import correctly the result (because the user may have advanced)
```
kneth: "import correctly" -> "correctly import"

realm/realm-library/src/main/java/io/realm/internal/TableOrView.java
```diff
@@ -372,10 +382,13 @@
     Table pivot(long stringCol, long intCol, PivotType pivotType);
 
     /**
-     * Syncs the tableview with the underlying table data. It is not required to call this explicitly, all other API
-     * methods will automatically sync the view as well.
+     * Syncs the tableview with the underlying table data. This is effectively the same as rerunning the query, so it
```
kneth: "tableview" -> either "table view" or "TableView"

kneth:
I have completed a first round focusing on two languages: C++ and English.

realm/realm-library/src/main/java/io/realm/RealmResults.java
cmelchior: Isn't the former better for non-native english speakers?

realm/realm-jni/src/io_realm_internal_tableview.cpp
stk1m1: I think it was @beeender who explained to me why `*_VALID` should be called outside `try/catch`. When `CATCH_STD()` catches a throw via `ThrowException()` from `*_VALID`, it _again_ calls `ThrowException` to throw the same exception. So, there is no need to call `*_VALID` from `try/catch`.

realm/realm-jni/src/io_realm_internal_tableview.cpp
```diff
@@ -165,9 +170,11 @@ JNIEXPORT jlong JNICALL Java_io_realm_internal_TableView_nativeGetSourceRowIndex
 {
     try {
         if (!VIEW_VALID_AND_IN_SYNC(env, nativeViewPtr))
-            return npos;
+            return -1;
```
stk1m1: It would be nice to use `jlong(-1)`.

realm/realm-jni/src/io_realm_internal_tableview.cpp
stk1m1: if the last line is `noexcept`, then `to_jlong_or_not_found(TV(nativeViewPtr)->get_source_ndx(S(rowIndex)))` would be nice.

stk1m1:
Nice PR, especially refactored `handlerController` to support explicit refresh. ;)

cmelchior:

stk1m1:
:+1:

realm/realm-library/src/main/java/io/realm/RealmResults.java
kneth: Maybe just leave out "always".

realm/realm-library/src/androidTest/java/io/realm/CollectionTests.java
kneth: Yes, I understand that. But consider the snippet:

    int N = 0;
    int m = 0;
    for(int i = 0; i < N; i++) {
        m++;
    }

My question is: what is `m` at the end?

kneth:
@cmelchior My comment about `depends_on_deleted_linklist()` hasn't been address yet. 

| in_sync | depends | not in_sync | not in_sync && depends |
|------------|-------------|-----------------|----------------------------------|
| true       | true         | false            | false                               |
| true       | false       | false             | false                               |
| false      | true        | true              | true                                 |
| false      | false      | true              | false                                |
|------------|------------|------------------|----------------------------------|

As you can see, in order to call `sync_if_needed()`, `depends_on_deleted_linklist()` must return `true`.

realm/realm-library/src/androidTest/java/io/realm/OrderedRealmCollectionIteratorTests.java
```diff
@@ -0,0 +1,964 @@
...
+
+        assertFalse(collectionClass == CollectionClass.MANAGED_REALMLIST ? !types.isValid() : types.isValid());
+    }
+
+    @Test
+    public void listIterator_remove_doesNotDeleteObject() {
+        ListIterator<AllJavaTypes> it = collection.listIterator();
+        AllJavaTypes obj = it.next();
+        assertEquals("test data 0", obj.getFieldString());
+        realm.beginTransaction();
```
bmunkholm: Why is this transaction not being closed/cancelled anywhere?
Same goes for other methods below.

cmelchior: The transaction is automatically canceled when closing the Realm, so there shouldn't be any need to commit/cancel unless needed in the test.

realm/realm-library/src/androidTest/java/io/realm/CollectionTests.java
cmelchior: m is 0, The reason i added that if statement is because of the code below the loop: https://github.com/realm/realm-java/pull/2124/files#diff-af07352af4c866c231b76d8b9a522827R96 It would blow up for the special case where `objects = 0`

cmelchior:
Added comment to tableview.cpp to address @kneth question

realm/realm-library/src/main/java/io/realm/RealmResults.java
```diff
@@ -996,17 +1013,22 @@ public void removeChangeListeners() {
     /**
      * Notifies all registered listeners.
      */
-    void notifyChangeListeners() {
+    void notifyChangeListeners(boolean syncBeforeNotifying) {
+        notifyChangeListeners(syncBeforeNotifying, false);
+    }
+
+    void notifyChangeListeners(boolean syncBeforeNotifying, boolean forceNotify) {
```
beeender: No need to expose this method to outside? mark it as `private`?

realm/realm-library/src/main/java/io/realm/RealmResults.java
```diff
@@ -996,17 +1013,22 @@ public void removeChangeListeners() {
     /**
      * Notifies all registered listeners.
      */
-    void notifyChangeListeners() {
+    void notifyChangeListeners(boolean syncBeforeNotifying) {
```
beeender: When called outside, the `syncBeforeNotifying` is always true? worth to only expose as `notifyChangeListeners()` without any param?

realm/realm-library/src/androidTest/java/io/realm/RealmObjectTests.java
```diff
@@ -263,48 +261,31 @@ public void deleteFromRealm_removedFromResults() {
...
-            catch (IllegalStateException ignored) {}
-
-            realm.commitTransaction();
-
-            // and removed from realm and remaining objects are place correctly
-            RealmResults<Dog> remainingDogs = realm.allObjects(Dog.class);
-            assertEquals(TEST_SIZE - i - 1, remainingDogs.size());
-            for (Dog dog : remainingDogs) {
-                assertTrue(ages.contains(dog.getAge()));
+                dogs.get(i).deleteFromRealm();
```
beeender: Does this function do anything when `atFirst == false`?

realm/realm-library/src/main/java/io/realm/RealmList.java
```diff
@@ -724,4 +760,174 @@ public String toString() {
...
+         * to remove.
+         */
+        int lastRet = -1;
+
+        /**
+         * The modCount value that the iterator believes that the backing
+         * List should have.  If this expectation is violated, the iterator
+         * has detected concurrent modification.
+         */
+        int expectedModCount = modCount;
```
beeender: Do you ever increase the value of `modCount`? Also, will `ConcurrentModificationException` ever be thrown since the realm checking is always ran first?

realm/realm-library/src/main/java/io/realm/RealmResults.java
```diff
@@ -723,54 +727,54 @@ public boolean addAll(Collection<? extends E> collection) {
...
-                throw new ConcurrentModificationException("No outside changes to a Realm is allowed while iterating a RealmResults. Use iterators methods instead.");
+        protected void checkRealmIsStable() {
+            long version = table.getVersion();
+            // Any change within a write transaction will immediately update the table version. This means that we
+            // cannot depend on the tableVersion heuristic in that case.
+            // You could argue that in that case it is not really a "ConcurrentModification", but this interpretation
+            // is still more lax than what the standard Java Collection API gives.
+            // TODO: Try to come up with a better scheme
+            if (!realm.isInTransaction() && tableViewVersion > -1 && version != tableViewVersion) {
+                throw new ConcurrentModificationException("No outside changes to a Realm is allowed while iterating a RealmResults. Don't call Realm.refresh() while iterating.");
```
beeender: Cannot find a test case for this. something like:
```java
        for (AllTypes allTypes : results) {
            realm.beginTransaction();
            realm.commitTransaction();
            realm.refresh();
        }
```

realm/realm-library/src/androidTest/java/io/realm/OrderedRealmCollectionIteratorTests.java
```diff
@@ -0,0 +1,964 @@
...
+            fail();
+        }
+        assertEquals(0, collection.size());
+    }
+
+    @Test
+    public void iterator_oneElement() {
+        collection = createCollection(realm, collectionClass, 1);
+        Iterator<AllJavaTypes> it = collection.iterator();
+        while (it.hasNext()) {
```
beeender: //noinspection WhileLoopReplaceableByForEach

realm/realm-library/src/androidTest/java/io/realm/OrderedRealmCollectionIteratorTests.java
```diff
@@ -0,0 +1,964 @@
...
+        realm.commitTransaction();
+    }
+
+    /**
+     * Helper method for checking if the unit test isn't supported for the current collectionClass.
+     * @param unsupportedTypes list of unsupported test types
+     * @return {@code true} if the unit test should be aborted, {@code false} if it should continue.
+     */
+    private boolean skipTest(CollectionClass... unsupportedTypes) {
+        for (int i = 0; i < unsupportedTypes.length; i++) {
```
beeender: `//noinspection ForLoopReplaceableByForEach`

realm/realm-library/src/androidTest/java/io/realm/OrderedRealmCollectionIteratorTests.java
```diff
@@ -0,0 +1,964 @@
...
+                assertEquals(TEST_SIZE, collection.size());
+                break;
+
+            default:
+                fail();
+        }
+    }
+
+    @Test
+    public void iterator_removedCalledTwice() {
```
beeender: `iterator_removedCalledTwice` -> `iterator_removeCalledTwice` no **d**

realm/realm-library/src/androidTest/java/io/realm/OrderedRealmCollectionIteratorTests.java
```diff
@@ -0,0 +1,964 @@
...
+            case UNMANAGED_REALMLIST:
+            case REALMRESULTS:
+                assertEquals(TEST_SIZE, collection.size());
+                break;
+        }
+        it = collection.listIterator();
+        it.next();
+        AllJavaTypes types = it.next(); // Iterator can still access the deleted object
+
+        assertFalse(collectionClass == CollectionClass.MANAGED_REALMLIST ? !types.isValid() : types.isValid());
```
beeender: `assertTrue(collectionClass == CollectionClass.MANAGED_REALMLIST ? types.isValid() : !types.isValid());`
for easier reading

realm/realm-library/src/androidTest/java/io/realm/TypeBasedNotificationsTests.java
```diff
@@ -1402,25 +1399,12 @@ public void onChange() {
 
     // ****************************************************************************************** //
     // UC 5.
-    // Callback should be notified if we call refresh (even without getting the REALM_CHANGE yet)
+    // Callback should be notified if we call refresh
```
beeender: Is this still a valuable test case? Even without calling `refresh`, test won't fail i think?

realm/realm-library/src/androidTest/java/io/realm/TypeBasedNotificationsTests.java
```diff
@@ -1462,28 +1446,17 @@ public void run() {
     @RunTestInLooperThread
     public void refresh_should_notify_callbacks_realmobject_async() {
```
beeender: same

realm/realm-library/src/androidTest/java/io/realm/TypeBasedNotificationsTests.java
```diff
@@ -1513,20 +1486,6 @@ public void run() {
     @RunTestInLooperThread
     public void refresh_should_notify_callbacks_realmresults_sync() {
```
beeender: same

realm/realm-library/src/androidTest/java/io/realm/TypeBasedNotificationsTests.java
```diff
@@ -1563,19 +1522,6 @@ public void run() {
     @RunTestInLooperThread
     public void refresh_should_notify_callbacks_realmresults_async() {
```
beeender: same

beeender:
Done the 1st round

realm/realm-library/src/androidTest/java/io/realm/RealmObjectTests.java
cmelchior: No, this test was a pretty bad refactor. I changed it to make more sense now.

realm/realm-library/src/main/java/io/realm/RealmList.java
cmelchior: Your right, there seems to be a big hole in the test coverage regarding ConcurrentModificationExceptions. I'll look into it

realm/realm-library/src/androidTest/java/io/realm/TypeBasedNotificationsTests.java
cmelchior: Great catch 💯

cmelchior:
All comments addressed @beender / @realm/java

Final commit message:
-----
Title: RealmCollection iterators are now stable 

This commit changes the semantics of how live RealmResults really are. Before this commit
RealmResults where live _all the time_, which meant that code like the below didn't work as expected (only half the elements would be deleted):

```
for (int i = 0; i < results.size(); i++) {
   results.get(i).deleteFromRealm();
}

```

The rather unintuitive work-around was counting backwards:

```
for (int i = results.size() -1; i >=0; i--) {
   results.get(i).deleteFromRealm();
}
```

This commit changes the behaviour of RealmResults, so they no longer are automatically up to date. They are instead refreshed using Looper events just like changes from other threads.

RealmList iterators are not live in the same way that RealmResults are but act more like standard ArrayLists, so they don't suffer from the same liveness issues.

The high-level arguments for doing this change was:

Pros:
* All iterators now work as you would normally expect (ease of use).
* The liveness of Realm objects are now all controlled through the Looper (simpler internal code / easier to document Realm behaviour).

Cons:
* There is a small chance that people can accidentally reference a deleted RealmObject in a RealmResult. (Can be checked through use of `RealmObject.isValid()`)

realm/realm-library/src/main/java/io/realm/RealmResults.java
cmelchior: Good catch, I added a unit test for it.

realm/realm-library/src/main/java/io/realm/RealmList.java
cmelchior: Unit tests added

realm/realm-library/src/main/java/io/realm/RealmList.java
```diff
@@ -319,14 +324,16 @@ public void clear() {
...
             checkValidView();
-            E removedItem = get(location);
+            removedItem = get(location);
             view.remove(location);
-            return removedItem;
         } else {
-            return nonManagedList.remove(location);
+            removedItem = nonManagedList.remove(location);
         }
+        modCount++;
```
beeender: Doesn't seem to be big issue, but `ArrayList.remove()` will do `modCount++`. Same with others.

realm/realm-library/src/main/java/io/realm/RealmList.java
```diff
@@ -351,7 +358,11 @@ public boolean remove(Object object) {
         if (managedMode && !realm.isInTransaction()) {
             throw new IllegalStateException(REMOVE_OUTSIDE_TRANSACTION_ERROR);
         }
-        return super.remove(object);
+        boolean success = super.remove(object);
+        if (success) {
+            modCount++;
```
beeender: Is this necessary? `super.remove(object)` will call `iterator.remove()` which should do `modCount` ?

beeender:
Except some details about `modCount++`, :+1:

realm/realm-jni/src/io_realm_internal_tableview.cpp
```diff
@@ -165,9 +171,11 @@ JNIEXPORT jlong JNICALL Java_io_realm_internal_TableView_nativeGetSourceRowIndex
...
         if (!VIEW_VALID_AND_IN_SYNC(env, nativeViewPtr))
-            return npos;
+            return jlong(-1);
         if (!ROW_INDEX_VALID(env, TV(nativeViewPtr), rowIndex))
-            return npos;
+            return jlong(-1);
+        if (!TV(nativeViewPtr)->is_row_attached(rowIndex))
+            return jlong(-1);
     } CATCH_STD()
     return TV(nativeViewPtr)->get_source_ndx(S(rowIndex));   // noexcept
```
kneth: `to_jlong_or_not_found()`?

realm/realm-library/src/androidTest/java/io/realm/OrderedRealmCollectionIteratorTests.java
```diff
@@ -0,0 +1,1102 @@
...
+    @Test
+    public void iterator() {
+        Iterator<AllJavaTypes> it = collection.iterator();
+        int i = 0;
+        while (it.hasNext()) {
+            AllJavaTypes item = it.next();
+            assertEquals("Failed at index: " + i, i, item.getFieldLong());
+            i++;
+        }
+        assertEquals(TEST_SIZE, collection.size());
```
kneth: You could add a check of `i` having the excepted value (`TEST_SIZE`).

realm/realm-library/src/androidTest/java/io/realm/OrderedRealmCollectionIteratorTests.java
```diff
@@ -0,0 +1,1102 @@
...
+    }
+
+    @Test
+    public void iterator_empty() {
+        collection = createCollection(realm, collectionClass, 0);
+        Iterator<AllJavaTypes> it = collection.iterator();
+        while (it.hasNext()) {
```
kneth: `assertFalse(it.hasNext())` maybe

realm/realm-library/src/androidTest/java/io/realm/OrderedRealmCollectionIteratorTests.java
```diff
@@ -0,0 +1,1102 @@
...
+    @Test
+    public void iterator_oneElement() {
+        collection = createCollection(realm, collectionClass, 1);
+        Iterator<AllJavaTypes> it = collection.iterator();
+        //noinspection WhileLoopReplaceableByForEach
+        while (it.hasNext()) {
+            AllJavaTypes item = it.next();
+            assertEquals(0, item.getFieldLong());
+        }
+        assertEquals(1, collection.size());
```
kneth: Adding a counter to check if the number of iterations is as excepted?

realm/realm-library/src/androidTest/java/io/realm/OrderedRealmCollectionIteratorTests.java
```diff
@@ -0,0 +1,1102 @@
...
+    }
+
+    @Test
+    public void listIterator_deleteManagedObjectIndirectly() {
+        realm.beginTransaction();
+        Iterator<AllJavaTypes> it = collection.iterator();
+        it.next();
+        it.next().deleteFromRealm();
+        realm.commitTransaction();
+
```
kneth: Many empty lines

realm/realm-library/src/main/java/io/realm/RealmList.java
```diff
@@ -724,4 +779,174 @@ public String toString() {
...
+
+        /**
+         * {@inheritDoc}
+         */
+        public void set(E e) {
+            realm.checkIfValid();
+            if (lastRet < 0) {
+                throw new IllegalStateException();
+            }
+            checkConcurrentModification();
```
kneth: Should this check go just after `checkIfValid()`?

realm/realm-library/src/main/java/io/realm/RealmResults.java
```diff
@@ -997,16 +1014,21 @@ public void removeChangeListeners() {
...
             // but we have already registered listeners for it)
-            if (pendingQuery != null && !isCompleted) return;
-            long version = table.sync();
-            if (currentTableViewVersion != version) {
-                currentTableViewVersion = version;
-                for (RealmChangeListener listener : listeners) {
-                    listener.onChange();
-                }
+            if (pendingQuery != null && !asyncQueryCompleted) return;
+            if (!viewUpdated && !forceNotify) return;
```
kneth: `forceNotify` implies to me that notification is done no matter what. But that's not accurate, right? SInce it's a private method, it isn't that important.

kneth:
:+1:

realm/realm-library/src/main/java/io/realm/RealmList.java
cmelchior: It could, but this order is the same as in ArrayList, so I would prefer keeping it.

realm/realm-library/src/androidTest/java/io/realm/OrderedRealmCollectionIteratorTests.java
cmelchior: Fixed

realm/realm-library/src/androidTest/java/io/realm/OrderedRealmCollectionIteratorTests.java
cmelchior: Done

realm/realm-library/src/main/java/io/realm/RealmList.java
cmelchior: True, but it also has to work for managed lists that don't call the ArrayList methods

beeender:
:+1:

realm/realm-library/src/main/java/io/realm/HandlerController.java
```diff
@@ -42,29 +43,34 @@
 /**
  * Centralises all Handler callbacks, including updating async queries and refreshing the Realm.
  */
-public class HandlerController implements Handler.Callback {
+public final class HandlerController implements Handler.Callback {
+
+    public static final int REALM_CHANGED = 14930352; // Hopefully it won't clash with other message IDs.
```
nhachicha: why public?

realm/realm-library/src/main/java/io/realm/HandlerController.java
```diff
@@ -443,10 +450,8 @@ private void completedAsyncQueriesUpdate(QueryUpdateTask.Result result) {
         SharedGroup.VersionID callerVersionID = realm.sharedGroupManager.getVersion();
         int compare = callerVersionID.compareTo(result.versionID);
         if (compare > 0) {
-            RealmLog.d("COMPLETED_UPDATE_ASYNC_QUERIES realm:" + HandlerController.this + " caller is more advanced, rerun updates");
-            // The caller is more advance than the updated queries ==>
-            // need to refresh them again (if there is still async queries)
-            realm.handler.sendEmptyMessage(REALM_CHANGED);
```
nhachicha: why not rerunning queries ?

realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -271,13 +271,13 @@ public void refresh() {
...
-            handlerController.notifyAllListeners();
-            // if we have empty async RealmObject then rerun
-            if (handlerController.threadContainsAsyncEmptyRealmObject()) {
-                handlerController.updateAsyncEmptyRealmObject();
-            }
+        if (!handlerController.isAutoRefreshEnabled()) {
+            // non Looper Thread, just advance the Realm
+            // registering listeners is not allowed, hence nothing to notify
+            sharedGroupManager.advanceRead();
+            handlerController.refreshTableViews();
```
nhachicha: you're not refreshing `RealmResults` obtained asynchronously ?

cmelchior: This code is only triggered if there is no auto-refresh, so there isn't any async queries either

cmelchior: You mean beside the one above this statement? `refreshTableViews` are called in both cases?

realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -76,6 +76,7 @@ protected BaseRealm(RealmConfiguration configuration, boolean autoRefresh) {
         this.configuration = configuration;
         this.sharedGroupManager = new SharedGroupManager(configuration);
         this.schema = new RealmSchema(this, sharedGroupManager.getTransaction());
+        this.handlerController = new HandlerController(this);
```
nhachicha: what's the point of creating a HandlerController for non Looper thread?

realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -271,13 +271,13 @@ public void refresh() {
...
-            if (handlerController.threadContainsAsyncEmptyRealmObject()) {
-                handlerController.updateAsyncEmptyRealmObject();
-            }
+        if (!handlerController.isAutoRefreshEnabled()) {
+            // non Looper Thread, just advance the Realm
+            // registering listeners is not allowed, hence nothing to notify
+            sharedGroupManager.advanceRead();
+            handlerController.refreshTableViews();
+        } else {
+            handlerController.notifyCurrentThreadRealmChanged();
```
nhachicha: calling the `advance_read` will have the same effect as `refreshTableViews` I suppose?

realm/realm-library/src/main/java/io/realm/RealmResults.java
```diff
@@ -998,16 +1015,21 @@ public void removeChangeListeners() {
      * Notifies all registered listeners.
      */
     void notifyChangeListeners() {
+        notifyChangeListeners(true, false);
+    }
+
+    private void notifyChangeListeners(boolean syncBeforeNotifying, boolean forceNotify) {
+        if (syncBeforeNotifying) {
+            syncIfNeeded();
```
nhachicha: will this rerun the async `RealmResults` query?

realm/realm-library/src/main/java/io/realm/RealmResults.java
```diff
@@ -998,16 +1015,21 @@ public void removeChangeListeners() {
      * Notifies all registered listeners.
      */
     void notifyChangeListeners() {
```
nhachicha: in the `HandlerController` you need to call the overload one with `syncBeforeNotifiying == false` here example: https://github.com/realm/realm-java/blob/cm/bug/iterators/realm/realm-library/src/main/java/io/realm/HandlerController.java#L403

realm/realm-library/src/main/java/io/realm/BaseRealm.java
cmelchior: None, but it possible for a Looper thread to enable/disable autorefresh without loosing all registered listeners. We had a feature request for it #2051

realm/realm-library/src/main/java/io/realm/HandlerController.java
cmelchior: Setting to package private

realm/realm-library/src/androidTest/java/io/realm/TypeBasedNotificationsTests.java
```diff
@@ -1401,286 +1398,6 @@ public void onChange() {
         TestHelper.awaitOrFail(signalTestFinished);
     }
 
-    // ****************************************************************************************** //
```
nhachicha: what happens to my babies 🤔

cmelchior: They died in the big fire of "refresh() is triggering on next looper event" refactoring

realm/realm-library/src/main/java/io/realm/BaseRealm.java
nhachicha: No just a comment to avoid calling `refreshTableViews` from a Looper thread, and expecting it to refresh async `RealmResults`

cmelchior: Ok

realm/realm-library/src/main/java/io/realm/BaseRealm.java
cmelchior: This is slightly different. This abstraction is there so it is only decided in one place if we trigger advance_read/notifications straight away or use the looper.

realm/realm-library/src/main/java/io/realm/RealmResults.java
cmelchior: Yes

nhachicha:
👍

PR #2128, opened 2016-01-21
Access to RealmResults based on deleted RealmList

beeender:
* When the original RealmList is deleted, for most methods of
  RealmResults should just work without crash by just treat it like an
  empty RealmResults.
* RealmResults.where() return null in this case.

Close 1945

### DIFF
--- a/realm/realm-library/src/main/java/io/realm/RealmResults.java
+++ b/realm/realm-library/src/main/java/io/realm/RealmResults.java
@@ -29,6 +29,7 @@ import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.Future;
 
 import io.realm.exceptions.RealmException;
+import io.realm.internal.DeleteRealmListException;
 import io.realm.internal.TableOrView;
 import io.realm.internal.TableQuery;
 import io.realm.internal.TableView;
@@ -51,6 +52,10 @@ import rx.Observable;
  * <p>
  * Notice that a RealmResults is never {@code null} not even in the case where it contains no objects. You should always
  * use the size() method to check if a RealmResults is empty or not.
+ * <p>
+ * If a RealmResults is built on RealmList through {@link RealmList#where()}, it will become invalid when the source
+ * RealmList gets deleted. When that happens, the RealmResults will behave like a empty RealmResults, but calling
+ * {@link #where()} will throw an {@link IllegalStateException}. Use {@link #isValid} to detect this situation.
  *
  * @param <E> The class of objects in this list.
  * @see RealmQuery#findAll()
@@ -65,7 +70,9 @@ public final class RealmResults<E extends RealmObject> extends AbstractList<E> {
     private TableOrView table = null;
 
     private static final String TYPE_MISMATCH = "Field '%s': type mismatch - %s expected.";
-    private long currentTableViewVersion = -1;
+    private static final long TABLE_VIEW_VERSION_NONE = -1;
+    private static final long TABLE_VIEW_VERSION_REALM_LIST_DELETED = -2;
+    private long currentTableViewVersion = TABLE_VIEW_VERSION_NONE;
 
     private final TableQuery query;
     private final List<RealmChangeListener> listeners = new CopyOnWriteArrayList<RealmChangeListener>();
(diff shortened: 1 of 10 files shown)


cmelchior:
Shouldn't this PR be against the release branch?

realm/realm-library/src/main/java/io/realm/RealmResults.java
```diff
@@ -856,13 +866,24 @@ void notifyChangeListeners() {
...
             //       https://github.com/realm/realm-core/pull/1390
-            long version = table.sync();
+            long version;
+            try {
+                version = table.sync();
+            } catch (DeleteRealmListException e) {
+                // Although this RealmResults won't be updated anymore, it is good to give user a chance to do update.
+                version = TABLE_VIEW_VERSION_REALM_LIST_DELTED;
+                RealmLog.d("The parent RealmList has been deleted already.");
+            }
```
beeender: Now the behaviour is the listener will be only triggered once after the parent `RealmList` deleted. What do you guys think about this? I think user deserve a chance to update UI in this case? But it would be weird that `isValid` still return true at this time.

realm/realm-library/src/main/java/io/realm/RealmResults.java
```diff
@@ -149,6 +152,13 @@ public boolean isValid() {
      */
     public RealmQuery<E> where() {
         realm.checkIfValid();
+        try {
```
beeender: * Tricky part. Now we just return a `null` to user. And in the real world I can predict there will be some user met the `NullPointerException` because of this.
* Another option is we throw a checked exception to user to force user to catch it. But we don't use any checked exception in the lib, it might look like strange we have one here.
* IllegalStateException is something i don't like in this case, since there are some other case could also throw ISE here. eg.: a closed Realm ----- which is a user's bug should be fixed.

beeender:
@realm/java This is not finished, but i do need some discussion on the interface behaviours. See my comments in the code. thanks!

realm/realm-library/src/main/java/io/realm/RealmResults.java
cmelchior: Why not throw `IllegalStateException` and say in the message that you are querying on data that was deleted? I don't see the problem with multiple ISE? 
Returning null is not a good option IMO.

realm/realm-library/src/main/java/io/realm/RealmResults.java
cmelchior: I would also just throw an IllegalStateException in this case I think. Right now we are just preventing people from getting native crashes that are totally obfuscated. At least with a proper exception we can give them a work-around until this problem is fixed?

beeender: Throw ISE here is absolutely a mess for user in my opinion. That means:
```
final RealmResults<Dog> dogs =                       allTypes.getColumnRealmList().where().equalTo(Dog.NAME, "name_0").findAll();
                dogs.addChangeListener(listener);

                // Trigger the listener at the first time.
                realm.beginTransaction();
                allTypes.removeFromRealm();
                realm.commitTransaction();
```

the above code will crash. But it is totally legal code IMO. We cannot force user to remove change listener in above transaction block, that will be not necessary strict.

stk1m1: Since I'm still getting my feet wet, I'll just put my opinion as an user. As a user, I would be very convinced if `dogs`' `ChangedListener` returns the last value just before `allTypes` gets deleted. 

I would accept to have ISEs from `dogs` whenever I try to modify or read it after that.

cmelchior: I agree it is messy, but it is a work-around that ensures that people knows _why_ it is crashing until  it get properly fixed. IMO I would add a descriptive error message + point to the GitHub issue. Using to much energy on a work-around that hopefully don't have to be in place for long also seems like a bit overkill to me?

beeender: @cmelchior 
We have discussion with @jpsim in https://github.com/realm/realm-core/pull/1434#issuecomment-171008559 . It tends to be the most convenient way for user is just give user a empty results when this happens. @rrrlasse will make core behaviour like that in the later core release. Currently i just catch the exception and make the behaviour the same as what we could have in core in the future.

@stk1m1 Think about you might delete the Parent object when you get response from server side in another thread, and that will be super difficult for user to remove the change listener in the UI thread, with all kind of timing headaches.

BTW, logically, it is not pretend the results is empty ---- it is literally empty, because the query condition is empty (deleted).

cmelchior: I agree that the empty result is the "correct" behavior. Couldn't we just check for TABLE_VIEW_VERSION_REALM_LIST_DELTED in `isValid` ?

beeender: Good idea, will do that.

beeender:
Wrong branch. See #2133 instead.

PR #2133, opened 2016-01-22
Access to RealmResults based on deleted RealmList

### LINKED ISSUE #1945: Detect invalid RealmResults
The following code might lead to a native crash:

    realm.beginTransaction();
    RealmResults<Owner> owners = realm.allObjects(Owner.class);
    RealmResults<Dog> dogs = owners.first().where().equalTo("dogs.name", "Fido").findAll();
    owners.clear();
    long nDogs = dogs.size(); // crash here
    realm.commitTransaction();

The reason is that Realm's storage engine doesn't invalidate `dogs` after `owners.clear()`. Once it does, an exception should be thrown at `dogs.size()`.

See also https://github.com/realm/realm-core/issues/1413

beeender:
* When the original RealmList is deleted, for most methods of
  RealmResults should just work without crash by just treat it like an
  empty RealmResults.
* RealmResults.where() throws IllegalStateExecption in this case.
* RealmResults.isValid() return false in this case.

This is a temp fix, check https://github.com/realm/realm-core/pull/1434
for more details.

Close #1945

### DIFF
--- a/realm/realm-jni/src/io_realm_internal_tableview.cpp
+++ b/realm/realm-jni/src/io_realm_internal_tableview.cpp
@@ -24,21 +24,35 @@
 
 using namespace realm;
 
-// if you disable the validation, please remember to call sync_in_needed() 
-#define VIEW_VALID_AND_IN_SYNC(env, ptr) view_valid_and_in_sync(env, ptr)
+// The validation will try to sync the table view as well.
+// if you disable the validation, please remember to call sync_if_needed() 
+#define VIEW_VALID_AND_IN_SYNC(env, ptr) (is_view_valid(env, ptr) && sync_table_view(env, ptr))
 
-inline bool view_valid_and_in_sync(JNIEnv* env, jlong nativeViewPtr) {
-    bool valid = (TV(nativeViewPtr) != NULL);
-    if (valid) {
-        if (!TV(nativeViewPtr)->is_attached()) {
-            ThrowException(env, TableInvalid, "The Realm has been closed and is no longer accessible.");
-            return false;
-        }
-        TV(nativeViewPtr)->sync_if_needed();
+inline bool is_view_valid(JNIEnv* env, jlong nativeViewPtr) {
+    if (TV(nativeViewPtr) == NULL) {
+        // Should never get here
+        ThrowException(env, FatalError, "Null pointer of table view.");
+        return false;
     }
-    return valid;
+    if (!TV(nativeViewPtr)->is_attached()) {
+        ThrowException(env, TableInvalid, "The Realm has been closed and is no longer accessible.");
+        return false;
+    }
+    return true;
 }
 
+// Sync the TableView and return false if sync failed.
+inline bool sync_table_view(JNIEnv* /*env*/, jlong nativeViewPtr) {
+    try {
+        TV(nativeViewPtr)->sync_if_needed();
+    } catch (realm::DeletedLinkView&) {
+        // FIXME: Temp fix for https://github.com/realm/realm-core/pull/1434
+        // Better solution would be core only throw the exception when really necessary, methods like size
+        // should just return 0.
+        return false;
+    }
+    return true;
+}
 
 JNIEXPORT jlong JNICALL Java_io_realm_internal_TableView_createNativeTableView(
     JNIEnv* env, jobject, jobject, jlong)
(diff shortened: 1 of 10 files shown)


beeender:
@realm/java Ready for review. CI will pass after new core integrated.

realm/realm-library/src/androidTest/java/io/realm/RealmResultsTest.java
```diff
@@ -1146,4 +1146,100 @@ public void testComparisons() {
...
+        RealmResults<Dog> dogs = owner.getDogs().where().equalTo(Dog.FIELD_NAME, "name_0").findAll();
+        //dogs = dogs.where().findFirst().getOwner().getDogs().where().equalTo(Dog.FIELD_NAME, "name_0").findAll();
+
+        testRealm.beginTransaction();
+        owner.removeFromRealm();
+        testRealm.commitTransaction();
+        return dogs;
+    }
+
+    public void testisValidResultsBuiltOnDeletedLinkView() {
```
cmelchior: We will have to remember to convert all these to JUnit4 when merging with master :(

realm/realm-library/src/androidTest/java/io/realm/TypeBasedNotificationsTest.java
```diff
@@ -1827,8 +1826,8 @@ public void onChange() {
                         // Check if cat has been added to the realmObjects in case of the behaviour of getCat changes
                         for (WeakReference<RealmObject> weakReference : realm.handlerController.realmObjects.keySet()) {
                             if (weakReference.get() == cat) {
-                               foundKey = true;
-                               break;
+                                foundKey = true;
```
cmelchior: Isn't this indentation wrong?

realm/realm-library/src/main/java/io/realm/RealmResults.java
```diff
@@ -856,13 +872,42 @@ void notifyChangeListeners() {
...
+                RealmLog.d("The parent RealmList has been deleted already.");
+            }
             if (currentTableViewVersion != version) {
                 currentTableViewVersion = version;
                 for (RealmChangeListener listener : listeners) {
                     listener.onChange();
                 }
             }
+            // No point to keep those listeners, just clear them.
+            if (version == TABLE_VIEW_VERSION_REALM_LIST_DELETED) {
```
cmelchior: Is this a good idea? As the user are not aware they get removed? I would probably log it is happening.

realm/realm-library/src/main/java/io/realm/internal/DeleteRealmListException.java
```diff
@@ -0,0 +1,28 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.realm.internal;
+
+import io.realm.exceptions.RealmException;
+
+/**
+ *  Triggered from JNI level when access a RealmResults whose parent RealmList has been deleted already.
```
cmelchior: access -> accessing

cmelchior:
Apart from a few comments. It looks good :+1: 
 I'm most concerned about removing the changelisteners without warning. Can you elaborate on the reasoning for doing it?

realm/realm-jni/src/io_realm_internal_tableview.cpp
```diff
@@ -25,20 +25,32 @@
...
+        ThrowException(env, TableInvalid, "The Realm has been closed and is no longer accessible.");
+        return false;
     }
-    return valid;
+    return true;
 }
 
+inline bool is_view_in_sync(JNIEnv* /*env*/, jlong nativeViewPtr) {
+    try {
+        TV(nativeViewPtr)->sync_if_needed();
```
kneth: Can `sync_if_needed()` throw other exceptions? If so, please can them too.

realm/realm-jni/src/io_realm_internal_tableview.cpp
```diff
@@ -25,20 +25,32 @@
...
+    } catch (realm::DeletedLinkView&) {
+        // FIXME: Temp fix for https://github.com/realm/realm-core/pull/1434
```
kneth: Yes, using exceptions for flow control is a bad habit :-)

realm/realm-library/src/androidTest/java/io/realm/TypeBasedNotificationsTest.java
cmelchior: Nvm. This edit is correct

realm/realm-jni/src/io_realm_internal_tableview.cpp
```diff
@@ -25,20 +25,32 @@
 using namespace realm;
 
 // if you disable the validation, please remember to call sync_in_needed()
```
kneth: `sync_in_needed` -> `sync_is_needed`

realm/realm-jni/src/io_realm_internal_tableview.cpp
```diff
@@ -25,20 +25,32 @@
...
+    }
+    if (!TV(nativeViewPtr)->is_attached()) {
+        ThrowException(env, TableInvalid, "The Realm has been closed and is no longer accessible.");
+        return false;
     }
-    return valid;
+    return true;
 }
 
+inline bool is_view_in_sync(JNIEnv* /*env*/, jlong nativeViewPtr) {
```
kneth: The name suggests that it's a predicate but the side effect is that the table view is sync'ed.

realm/realm-library/src/androidTest/java/io/realm/RealmResultsTest.java
beeender: Yes, sure!

beeender:
@cmelchior User can check with `isValid` in the listener. It would also be OK that we keep the change listener, but actually the `RealmResults` in that case won't be changed any more.

cmelchior:
@beeender Good point. Removing them sounds fair in that case as it will make other change notifications be faster and explicitly calling remove(listener) would just be a no-op.

I would change the comment above the section to mention that no further changes can happen though as the current wording did confuse me a bit.

Apart from that, it looks ready to merge.

dalinaum:
:+1:

PR #2143, opened 2016-01-24
RealmList.get().getFloat() crashes core

cmelchior:
Unit test that exposes the problem.


cmelchior:
Replaced by #2144, which is against the release branch

PR #2144, opened 2016-01-24
RealmList not returning DynamicRealmObjects of the proper type

cmelchior:
RealmLists containing DynamicRealmObjects wrongly sat the type for them as the parent objects types instead of the linkview type. This caused segmentation faults if the wrong fields was accessed as the underlying type didn't match the getter used.

@realm/java

### DIFF
--- a/realm/realm-library/src/androidTest/java/io/realm/RealmListTest.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/RealmListTest.java
@@ -162,7 +162,7 @@ public class RealmListTest {
     public void testAddManagedObject_nonManagedMode() {
         RealmList<AllTypes> list = new RealmList<AllTypes>();
         testRealm.beginTransaction();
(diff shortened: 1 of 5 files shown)


realm/realm-library/src/androidTest/java/io/realm/RealmListTest.java
```diff
@@ -901,4 +901,21 @@ public void testSettingListClearsOldItems() {
 
         assertEquals(1, two.getObjects().size());
     }
-}
+
+    // Tests that a RealmList containing DynamicRealmObject create DynamicRealmObjects of the proper underlying type.
+    @Test
+    public void get_dynamicRealmObject_properType() {
```
beeender: move this to `DynamicRealmObjectTests` and rename to `getList` when merging?

cmelchior:
Done @beeender / @realm/java

beeender:
:+1:

zaki50:
:+1:

PR #2147, opened 2016-01-25
NotificationTests converted to use TestRealmConfigurationFactory

cmelchior:
Converted NotificationTests to use TestRealmConfigurationFactory so one unit test failure doesn't cause all the other tests to fail as well.

@realm/java


stk1m1:
:+1:

zaki50:
:+1:

PR #2148, opened 2016-01-25
Fix wrong usages of StringData.data().

zaki50:
`StringData.data()` does not guarantee the null termination of returned char array.

We must use `size()` information to properly handle the char array in the StringData like `std::string(str.data(), str.size())`.

Added test case does not fail on my local machine even if this fix was not applied since the next byte of char array happens to be null character.

### DIFF
--- a/realm/realm-jni/src/io_realm_internal_table.cpp
+++ b/realm/realm-jni/src/io_realm_internal_table.cpp
@@ -1551,7 +1551,7 @@ bool check_valid_primary_key_column(JNIEnv* env, Table* table, StringData column
                     int64_t next_val = results.get_int(column_index, i);
                     if (val == next_val) {
                         std::ostringstream error_msg;
-                        error_msg << "Field \"" << table->get_column_name(column_index).data() << "\" cannot be a primary key, ";
+                        error_msg << "Field \"" << column_name << "\" cannot be a primary key, ";
                         error_msg << "it already contains duplicate values: " << val;
                         ThrowException(env, IllegalArgument, error_msg.str());
                         return false;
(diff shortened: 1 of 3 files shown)


realm/realm-library/src/androidTest/java/io/realm/RealmObjectSchemaTests.java
```diff
@@ -342,6 +342,26 @@ public void testAddPrimaryKeyFieldModifier_illegalFieldTypeThrows() {
         }
     }
 
+    public void testAddPrimaryKeyFieldModifier_duplicateValues() {
```
cmelchior: If this unit test doesn't trigger it, I think we should just remove it.

cmelchior:
I would just remove the unit test, otherwise it looks :+1:

realm/realm-library/src/androidTest/java/io/realm/RealmObjectSchemaTests.java
zaki50: @cmelchior I think we should keep this since there are no tests for `addPrimaryKey()` for duplicate values.

I overlooked?

realm/realm-jni/src/io_realm_internal_table.cpp
```diff
@@ -1551,7 +1551,7 @@ bool check_valid_primary_key_column(JNIEnv* env, Table* table, StringData column
                     int64_t next_val = results.get_int(column_index, i);
                     if (val == next_val) {
                         std::ostringstream error_msg;
-                        error_msg << "Field \"" << table->get_column_name(column_index).data() << "\" cannot be a primary key, ";
```
kneth: I guess that `std::basic_ostream<C,T>& operator<<(std::basic_ostream<C,T>& out, const StringData& d)` is called, and then each character is added to the `ostringstream` one by one?

zaki50: @kneth you mean that it's better to just removing the `.data()`?

If so, I agree with that.

kneth: Oh yes, `column_name` must be enough.

realm/realm-library/src/androidTest/java/io/realm/RealmObjectSchemaTests.java
cmelchior: Ah okay. Just keep it then.

realm/realm-jni/src/io_realm_internal_table.cpp
zaki50: sure, I'll update the code.

zaki50: @kneth done

kneth:
:+1:

PR #2151, opened 2016-01-25
Convert RealmSchemaTests to JUnit4 + reenable fixed unit test.

### LINKED ISSUE #1829: Re-enable migration unit test
Re-enable this unit test once https://github.com/realm/realm-core/pull/1267 has been resolved.
https://github.com/realm/realm-java/pull/1496/files#diff-a1183a67c2d3874da2a022a68ada7088R136

cmelchior:
Fixes #1829 + re-enables the previously failing unit test (which is now fixed by the new core)

This was actually fixed in 0.87.3, so this PR adds the changelog to that version.

@realm/java

### DIFF
--- a/realm/realm-library/src/androidTest/java/io/realm/RealmSchemaTests.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/RealmSchemaTests.java
@@ -16,8 +16,15 @@
 
 package io.realm;
 
-import android.test.AndroidTestCase;
+import android.support.test.runner.AndroidJUnit4;
 
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Set;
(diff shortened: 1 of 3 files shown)


dalinaum:
:+1:

realm/realm-library/src/androidTest/java/io/realm/RealmSchemaTests.java
```diff
@@ -25,38 +31,47 @@
...
     }
 
-    public void testGetAllClasses() {
+    @Test
+    public void getAll() {
         Set<RealmObjectSchema> objectSchemas = realmSchema.getAll();
         assertEquals(5, objectSchemas.size());
 
-        List<String> expectedTables = Arrays.asList(CLASS_ALL_JAVA_TYPES, "Owner", "Cat", "Dog", "DogPrimaryKey");
+        List<String> expectedTables = Arrays.asList(AllJavaTypes.CLASS_NAME, "Owner", "Cat", "Dog", "DogPrimaryKey");
```
zaki50: `objectSchemas ` does not seem to contain the classes other than "AllJavaTypes" and "Owner";
How about adding the assertion that checks all expected classes are contained like:

```
        List<String> expectedTables = new ArrayList<String>(Arrays.asList(AllJavaTypes.CLASS_NAME, "Owner"));
        for (RealmObjectSchema objectSchema : objectSchemas) {
            if (!expectedTables.remove(objectSchema.getClassName())) {
                fail(objectSchema.getClassName() + " is unexpected.");
            }
        }
        assertTrue(expectedTables.toString() + " was not found.", expectedTables.isEmpty());
```

realm/realm-library/src/androidTest/java/io/realm/RealmSchemaTests.java
```diff
@@ -81,29 +98,33 @@ public void testCreateClassInvalidNameThrows() {
...
         assertFalse(realmSchema.contains("Owner"));
         assertTrue(realmSchema.contains("Owner2"));
     }
 
-    public void testRenameClassInvalidArgumentsThrows() {
+    @Test
+    public void rename_invalidArgumentThrows() {
         String[] illegalNames = new String[] { null, "" };
 
         // Test as first parameters
```
zaki50: last?

realm/realm-library/src/androidTest/java/io/realm/RealmSchemaTests.java
```diff
@@ -112,19 +133,21 @@ public void testRenameClassInvalidArgumentsThrows() {
         // Test as last parameter
```
zaki50: first?

zaki50:
Changes in this PR is :+1: 
I added comments for original test code.

realm/realm-library/src/androidTest/java/io/realm/RealmSchemaTests.java
cmelchior: If a class is added to the schema, all it's dependencies are added as well. This is why we have 5 classes, so the unit test is accurate. You can see it also checks for `size` before iterating.

You can make the argument that silently adding more classes is a bad idea, but we should probably take that in another issue/PR. I am probably leaning towards the current implementation being easier for users though :)

realm/realm-library/src/androidTest/java/io/realm/RealmSchemaTests.java
cmelchior: fixed

realm/realm-library/src/androidTest/java/io/realm/RealmSchemaTests.java
cmelchior: fixed

Issue #1893, opened 2015-12-04
onSuccess might not be called with the right Realm version for async transaction


beeender:
Now the `onSuccess` relies on the `REALM_CHANGED` triggers the `advance_read` first. If it is running with a unfinished async query, the `REALM_CHANGED` might be swallowed by `updateAsyncQueries` and the Realm doesn't make a `andvance_read`.

In above case, the next message in the handler queue is to run `onSuccess`, but it the function will be called in a wrong earlier Realm version which the changes in the async transaction is not in.

--------------------- Update Jan. 25th 2016 -------------------------
http://stackoverflow.com/questions/34938152/callback-never-be-invoked-when-i-excute-a-asynchronous-transaction-with-realm

See above question on SO. We need to discuss of the user case of `onSuccess`. Currently if the Realm instance is closed, then the `onSuccess` won't be called depends on the timing. But it might not be what user expect.

Shall we just let `onSuccess` be delivered whenever the transaction is done, and document it clearly that in the `onSuccess`, the Realm might not be the latest version?

dalinaum:
@beeender Are you looking into it? Otherwise, I'll start looking into it.

cmelchior:
Good example in that SO post.
I am not sure we are able to deliver the `onSuccess` handler if the Realm gets closed, as if it gets fully closed, we close the Handler and remove all messages. Not to mention the risk of everything getting GC'ed depending on how the user code is.

The pattern is also a bit strange IMO, as putting async behaviour inside a try-with-resources looks like a very error-prone pattern to me (not just with Realm). I can see why it is easy to do though.

IMO, I think we just need to document it clearly. Alternatively, we either log a warning or throw an exception if you close a Realm while a async transaction is still executing. Thoughts, @beeender ?

beeender:
> throw an exception if you close a Realm while a async transaction is still executing 
This would be too strict I think.

And you are right, because of the `HandlerManager` is closed as well, we cannot deliver the `onSuccess` anymore. 

Let's just doc the behaviour for `onSuccess` and closed Realm.

cmelchior:
Would logging a warning make sense? When closing a Realm we could check for any pending async transactions and if there is a callback registered?

beeender:
Yes, I think logging a warning would help. I guess @dalinaum can add the warning when fix this issue.

dalinaum:
I am reading related code, I will try to add logging a warning.

zaki50:
[...]

PR #2152, opened 2016-01-25
Using separate interfaces for async transaction also fixes #2130

### LINKED ISSUE #1657: Use interface instead of class for async transaction callback
We have 2 option:
1. Use one interface, users not interested in overriding `onError` or `onSuccess` will create an Adapter 
2. Use 2 interfaces, one for each callback + overload `executeTransaction` to take onSuccess/onError or both this will enable nice lambda syntax

### LINKED ISSUE #2130: Exceptions during async transactions are failing silently
So I've ran into a very frustrating error. Exceptions during asynchronous transactions aren't getting reported. No signs of stacktrace. 

This made me spend a couple of hours trying to figure out why my logic isn't working only to realize that Realm has stopped printing Error stacktraces.

##### Steps to reproduce
Simply throw an `Error` inside a transaction.

##### Example code

````java
private void testRealm() {

    final Realm.Transaction.Callback callback = new Realm.Transaction.Callback() {

        @Override
        public void onSuccess() {
            super.onSuccess();
        }

        @Override
            super.onError(e);
            e.printStackTrace();
        }

    };

    getRealm().executeTransaction(realm -> {
        throw new IllegalAccessError("Poop");
    }, callback);

}
````
The stacktrace never gets printed.

**Realm version**: 0.87.2

Any ideas how do i fix this?

### PR DESCRIPTION
**Pull request by nhachicha**
Fixes #1657 
Fixes #2130

### DIFF
--- a/realm/realm-library/src/androidTest/java/io/realm/RealmAsyncQueryTests.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/RealmAsyncQueryTests.java
@@ -18,7 +18,6 @@ package io.realm;
 
 import android.content.Context;
 import android.os.Handler;
-import android.os.HandlerThread;
 import android.os.Looper;
 import android.os.MessageQueue;
 import android.os.SystemClock;
@@ -56,6 +55,7 @@ import io.realm.rule.RunTestInLooperThread;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 
@@ -78,34 +78,96 @@ public class RealmAsyncQueryTests {
     // start asynchronously a transaction to insert one element
     @Test
     @RunTestInLooperThread
-    public void executeTransaction_async() throws Throwable {
+    public void executeTransactionAsync() throws Throwable {
         assertEquals(0, looperThread.realm.allObjects(Owner.class).size());
 
-        looperThread.realm.executeTransaction(new Realm.Transaction() {
+        looperThread.realm.executeTransactionAsync(new Realm.Transaction() {
             @Override
             public void execute(Realm realm) {
                 Owner owner = realm.createObject(Owner.class);
                 owner.setName("Owner");
             }
-        }, new Realm.Transaction.Callback() {
+        }, new Realm.Transaction.OnSuccess() {
             @Override
             public void onSuccess() {
                 assertEquals(1, looperThread.realm.allObjects(Owner.class).size());
                 assertEquals("Owner", looperThread.realm.where(Owner.class).findFirst().getName());
                 looperThread.testComplete();
             }
+        }, new Realm.Transaction.OnError() {
 
             @Override
-            public void onError(Exception e) {
-                looperThread.testComplete();;
-                fail(e.getMessage());
+            public void onError(Throwable error) {
+                fail(error.getMessage());
+            }
+        });
+    }
+
+    @Test
+    @RunTestInLooperThread
+    public void executeTransactionAsync_onSuccess() throws Throwable {
+        assertEquals(0, looperThread.realm.allObjects(Owner.class).size());
+
+        looperThread.realm.executeTransactionAsync(new Realm.Transaction() {
+            @Override
+            public void execute(Realm realm) {
+                Owner owner = realm.createObject(Owner.class);
+                owner.setName("Owner");
+            }
+        }, new Realm.Transaction.OnSuccess() {
+            @Override
+            public void onSuccess() {
+                assertEquals(1, looperThread.realm.allObjects(Owner.class).size());
+                assertEquals("Owner", looperThread.realm.where(Owner.class).findFirst().getName());
+                looperThread.testComplete();
+            }
+        });
+    }
+
+    @Test
+    @RunTestInLooperThread
+    public void executeTransactionAsync_onError() throws Throwable {
+        assertEquals(0, looperThread.realm.allObjects(Owner.class).size());
+
+        looperThread.realm.executeTransactionAsync(new Realm.Transaction() {
+            @Override
+            public void execute(Realm realm) {
+                throw new RuntimeException("Oh! What a Terrible Failure");
+            }
+        }, new Realm.Transaction.OnError() {
+            @Override
+            public void onError(Throwable error) {
+                assertEquals(0, looperThread.realm.allObjects(Owner.class).size());
+                assertNull(looperThread.realm.where(Owner.class).findFirst());
+                looperThread.testComplete();
+            }
+        });
+    }
+
+    @Test
+    @RunTestInLooperThread
+    public void executeTransactionAsync_NoCallbacks() throws Throwable {
+        assertEquals(0, looperThread.realm.allObjects(Owner.class).size());
+
+        looperThread.realm.executeTransactionAsync(new Realm.Transaction() {
+            @Override
+            public void execute(Realm realm) {
+                Owner owner = realm.createObject(Owner.class);
+                owner.setName("Owner");
+            }
+        });
+        looperThread.realm.addChangeListener(new RealmChangeListener() {
+            @Override
+            public void onChange() {
+                assertEquals("Owner", looperThread.realm.where(Owner.class).findFirst().getName());
+                looperThread.testComplete();
             }
         });
     }
 
     // Test that an async transaction that throws an exception propagate it properly to the user.
     @Test
-    public void executeTransaction_async_exceptionHandling() throws Throwable {
+    public void executeTransactionAsync_exceptionHandling() throws Throwable {
         final TestHelper.TestLogger testLogger = new TestHelper.TestLogger();
         final CountDownLatch signalCallbackFinished = new CountDownLatch(1);
         final CountDownLatch signalClosedRealm = new CountDownLatch(1);
@@ -125,7 +187,7 @@ public class RealmAsyncQueryTests {
 
                     assertEquals(0, realm[0].allObjects(Owner.class).size());
 
-                    realm[0].executeTransaction(new Realm.Transaction() {
+                    realm[0].executeTransactionAsync(new Realm.Transaction() {
                         @Override
                         public void execute(Realm realm) {
                             Owner owner = realm.createObject(Owner.class);
@@ -133,7 +195,7 @@ public class RealmAsyncQueryTests {
                             realm.cancelTransaction(); // Cancel the transaction then throw
                             throw new RuntimeException("Boom");
                         }
-                    }, new Realm.Transaction.Callback() {
+                    }, new Realm.Transaction.OnSuccess() {
                         @Override
                         public void onSuccess() {
                             try {
@@ -144,9 +206,9 @@ public class RealmAsyncQueryTests {
                                 signalCallbackFinished.countDown();
                             }
                         }
-
+                    }, new Realm.Transaction.OnError() {
                         @Override
-                        public void onError(Exception e) {
+                        public void onError(Throwable error) {
                             try {
                                 // Ensure we are giving developers quality messages in the logs.
                                 assertEquals(testLogger.message, "Could not cancel transaction, not currently in a transaction.");
@@ -182,9 +244,9 @@ public class RealmAsyncQueryTests {
     // Test if the background Realm is closed when transaction success returned.
     @Test
     @RunTestInLooperThread
-    public void executeTransaction_async_realmClosedOnSuccess() {
+    public void executeTransactionAsync_realmClosedOnSuccess() {
         final AtomicInteger counter = new AtomicInteger(100);
-        final Realm realm = Realm.getInstance(looperThread.createConfiguration("testClosedBeforeAsyncTransactionSuccess"));
+        final Realm realm = looperThread.realm;
         final RealmCache.Callback cacheCallback = new RealmCache.Callback() {
             @Override
             public void onResult(int count) {
@@ -195,7 +257,7 @@ public class RealmAsyncQueryTests {
                 }
             }
         };
-        final Realm.Transaction.Callback transactionCallback = new Realm.Transaction.Callback() {
+        final Realm.Transaction.OnSuccess transactionCallback = new Realm.Transaction.OnSuccess() {
             @Override
             public void onSuccess() {
                 RealmCache.invokeWithGlobalRefCount(realm.getConfiguration(), cacheCallback);
@@ -203,7 +265,7 @@ public class RealmAsyncQueryTests {
                     // Finish testing
                     return;
                 }
-                realm.executeTransaction(new Realm.Transaction() {
+                realm.executeTransactionAsync(new Realm.Transaction() {
                     @Override
                     public void execute(Realm realm) {
                     }
@@ -211,7 +273,7 @@ public class RealmAsyncQueryTests {
             }
         };
 
-        realm.executeTransaction(new Realm.Transaction() {
+        realm.executeTransactionAsync(new Realm.Transaction() {
             @Override
             public void execute(Realm realm) {
             }
@@ -223,7 +285,7 @@ public class RealmAsyncQueryTests {
     @RunTestInLooperThread
     public void executeTransaction_async_realmClosedOnError() {
         final AtomicInteger counter = new AtomicInteger(100);
-        final Realm realm = Realm.getInstance(looperThread.createConfiguration("testClosedBeforeAsyncTransactionSuccess"));
+        final Realm realm = looperThread.realm;
         final RealmCache.Callback cacheCallback = new RealmCache.Callback() {
             @Override
             public void onResult(int count) {
@@ -234,15 +296,15 @@ public class RealmAsyncQueryTests {
                 }
             }
         };
-        final Realm.Transaction.Callback transactionCallback = new Realm.Transaction.Callback() {
+        final Realm.Transaction.OnError transactionCallback = new Realm.Transaction.OnError() {
             @Override
-            public void onError(Exception e) {
+            public void onError(Throwable error) {
                 RealmCache.invokeWithGlobalRefCount(realm.getConfiguration(), cacheCallback);
                 if (counter.get() == 0) {
                     // Finish testing
                     return;
                 }
-                realm.executeTransaction(new Realm.Transaction() {
+                realm.executeTransactionAsync(new Realm.Transaction() {
                     @Override
                     public void execute(Realm realm) {
                         throw new RuntimeException("Dummy exception");
@@ -251,7 +313,7 @@ public class RealmAsyncQueryTests {
             }
         };
 
-        realm.executeTransaction(new Realm.Transaction() {
+        realm.executeTransactionAsync(new Realm.Transaction() {
             @Override
             public void execute(Realm realm) {
                 throw new RuntimeException("Dummy exception");
--- a/realm/realm-library/src/main/java/io/realm/Realm.java
+++ b/realm/realm-library/src/main/java/io/realm/Realm.java
@@ -19,7 +19,6 @@ package io.realm;
 import android.annotation.TargetApi;
 import android.content.Context;
 import android.os.Build;
-import android.os.Handler;
 import android.os.Looper;
 import android.util.JsonReader;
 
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Comment by nhachicha** [comment]
@realm/java

File: realm/realm-library/src/main/java/io/realm/Realm.java [hunk 1]
```diff
@@ -1070,17 +1069,59 @@ public void executeTransaction(Transaction transaction) {
      * Similar to {@link #executeTransaction(Transaction)} but runs asynchronously from a worker thread.
      *
      * @param transaction {@link io.realm.Realm.Transaction} to execute.
-     * @param callback optional, to receive the result of this query.
      * @return a {@link RealmAsyncTask} representing a cancellable task.
      * @throws IllegalArgumentException if the {@code transaction} is {@code null}, or if the realm is opened from another thread.
      */
-    public RealmAsyncTask executeTransaction(final Transaction transaction, final Transaction.Callback callback) {
```
dalinaum: Should we delete `executeTransaction`? Is the fallback interface not needed?

nhachicha: I'm not following you `executeTransaction` is the sync call

cmelchior: To follow our normal deprecation policies we would need to deprecate this for 2 releases instead of removing it outright.

realm/realm-library/src/androidTest/java/io/realm/RealmAsyncQueryTests.java
```diff
@@ -81,24 +81,89 @@ public void setUp() throws Exception {
...
+    @RunTestInLooperThread
+    public void executeTransaction_async_onError() throws Throwable {
+        assertEquals(0, looperThread.realm.allObjects(Owner.class).size());
+
+        looperThread.realm.executeTransactionAsync(new Realm.Transaction() {
+            @Override
+            public void execute(Realm realm) {
+                Owner owner = realm.createObject(Owner.class);
+                owner.setName("Owner");
+                realm.beginTransaction(); //Nested transaction should throw IllegalStateException
```
cmelchior: Maybe more clear to just throw a RuntimeException or similar?

cmelchior:
Looks good :+1:  but we need to mark the old approach as deprecated instead of deleting it outright. At least if we want to follow normal deprecation policies (which seems possible in this case).
But talk to @dalinaum about #1893 because I think you might be stepping on each other toes.

zaki50:
need an entry in changelog

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -1070,17 +1069,59 @@ public void executeTransaction(Transaction transaction) {
...
+    public RealmAsyncTask executeTransactionAsync(final Transaction transaction) {
+        return executeTransactionAsync(transaction, null, null);
+    }
+
+    /**
+     * Similar to {@link #executeTransactionAsync(Transaction)}.
+     *
+     * @param transaction {@link io.realm.Realm.Transaction} to execute.
+     * @param onSuccess callback invoked when the transaction succeed.
```
beeender: succeed -> succeeds

realm/realm-library/src/androidTest/java/io/realm/RealmAsyncQueryTests.java
```diff
@@ -81,24 +81,89 @@ public void setUp() throws Exception {
...
                 looperThread.testComplete();
             }
+        }, new Realm.Transaction.OnError() {
 
             @Override
-            public void onError(Exception e) {
-                looperThread.testComplete();;
-                fail(e.getMessage());
+            public void onError(Throwable error) {
+                looperThread.testComplete();
```
beeender: `looperThread.testComplete();` should be removed. `fail` will just trigger a throw.

realm/realm-library/src/androidTest/java/io/realm/RealmAsyncQueryTests.java
```diff
@@ -81,24 +81,89 @@ public void setUp() throws Exception {
...
+            @Override
+            public void execute(Realm realm) {
+                Owner owner = realm.createObject(Owner.class);
+                owner.setName("Owner");
+                looperThread.realm.handler.postDelayed(new Runnable() {
+                    @Override
+                    public void run() {
+                        looperThread.testComplete();
+                    }
+                }, 16);
```
beeender: Hope 16 ms is long enough XD

realm/realm-library/src/androidTest/java/io/realm/RealmAsyncQueryTests.java
```diff
@@ -223,7 +288,7 @@ public void execute(Realm realm) {
     @RunTestInLooperThread
     public void executeTransaction_async_realmClosedOnError() {
         final AtomicInteger counter = new AtomicInteger(100);
-        final Realm realm = Realm.getInstance(looperThread.createConfiguration("testClosedBeforeAsyncTransactionSuccess"));
+        final Realm realm = Realm.getInstance(looperThread.createConfiguration("executeTransaction_async_realmClosedOnError"));
```
beeender: No need to create a specific named Realm anymore, just use `looperThread.realm`?

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -1070,17 +1069,59 @@ public void executeTransaction(Transaction transaction) {
...
+     */
+    public RealmAsyncTask executeTransactionAsync(final Transaction transaction, final Realm.Transaction.OnSuccess onSuccess, final Realm.Transaction.OnError onError) {
         if (transaction == null)
             throw new IllegalArgumentException("Transaction should not be null");
 
         // If the user provided a Callback then we make sure, the current Realm has a Handler
         // we can use to deliver the result
-        if (callback != null && handler == null) {
+        if ((onSuccess != null || onError != null)  && handler == null) {
             throw new IllegalStateException("Your Realm is opened from a thread without a Looper" +
```
beeender: "Your Realm is opened from a thread without a Looper" - > "Your Realm is opened from a thread without a Looper or it has already been closed"

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -1128,23 +1169,22 @@ public void run() {
...
                                 @Override
                                 public void run() {
-                                    callback.onSuccess();
+                                    onSuccess.onSuccess();
                                 }
                             });
-                        } else if (exception[0] != null) {
-                            // transaction has not been canceled by there is a exception during transaction.
+                        }
+                        if (onError != null && exception[0] != null) {
```
beeender: If there is an exception, and user doesn't supply a `onError`, I think we should just throw.

cmelchior: Good catch. Yeah, throwing is :+1:

nhachicha:
@realm/java

kneth:
Looks fine!

realm/realm-library/src/androidTest/java/io/realm/RealmAsyncQueryTests.java
cmelchior: This hasn't been addressed?

realm/realm-library/src/androidTest/java/io/realm/RealmAsyncQueryTests.java
```diff
@@ -81,24 +81,86 @@ public void setUp() throws Exception {
...
+
+    @RunTestInLooperThread
+    public void executeTransaction_async_onError() throws Throwable {
+        assertEquals(0, looperThread.realm.allObjects(Owner.class).size());
+
+        looperThread.realm.executeTransactionAsync(new Realm.Transaction() {
+            @Override
+            public void execute(Realm realm) {
+                throw new RuntimeException("Oh! What a Terrible Failure");
+
```
cmelchior: Remove whitespace

realm/realm-library/src/androidTest/java/io/realm/RealmAsyncQueryTests.java
cmelchior: This hasn't been addressed.

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -1082,7 +1083,7 @@ public RealmAsyncTask executeTransaction(final Transaction transaction, final Tr
         // we can use to deliver the result
         if (callback != null && handler == null) {
             throw new IllegalStateException("Your Realm is opened from a thread without a Looper" +
-                    " and you provided a callback, we need a Handler to invoke your callback");
+                    " (or it has already been closed) and you provided a callback, we need a Handler to invoke your callback");
```
cmelchior: Any reason or changing this when it is deprecated?
Also wouldn't it make more sense to check for `realm.isValid()` instead of doing this?

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -1157,6 +1158,147 @@ public void run() {
     }
 
     /**
+     * Similar to {@link #executeTransaction(Transaction)} but runs asynchronously from a worker thread.
```
cmelchior: from -> on?

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -1157,6 +1158,147 @@ public void run() {
...
+    /**
+     * Similar to {@link #executeTransactionAsync(Transaction)}.
+     *
+     * @param transaction {@link io.realm.Realm.Transaction} to execute.
+     * @param onSuccess callback invoked when the transaction succeeds.
+     * @return a {@link RealmAsyncTask} representing a cancellable task.
+     * @throws IllegalArgumentException if the {@code transaction} is {@code null}, or if the realm is opened from another thread.
+     */
+    public RealmAsyncTask executeTransactionAsync(final Transaction transaction, final Realm.Transaction.OnSuccess onSuccess) {
+        if (onSuccess == null)
```
cmelchior: Missing `{}`

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -1157,6 +1158,147 @@ public void run() {
...
+    /**
+     * Similar to {@link #executeTransactionAsync(Transaction)}.
+     *
+     * @param transaction {@link io.realm.Realm.Transaction} to execute.
+     * @param onError callback invoked when the transaction failed.
+     * @return a {@link RealmAsyncTask} representing a cancellable task.
+     * @throws IllegalArgumentException if the {@code transaction} is {@code null}, or if the realm is opened from another thread.
+     */
+    public RealmAsyncTask executeTransactionAsync(final Transaction transaction, final Realm.Transaction.OnError onError) {
+        if (onError == null)
```

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -1157,6 +1158,147 @@ public void run() {
...
+     */
+    public RealmAsyncTask executeTransactionAsync(final Transaction transaction, final Realm.Transaction.OnSuccess onSuccess) {
+        if (onSuccess == null)
+            throw new IllegalArgumentException("onSuccess callback can't be null");
+
+        return executeTransactionAsync(transaction, onSuccess, null);
+    }
+
+    /**
+     * Similar to {@link #executeTransactionAsync(Transaction)}.
```
cmelchior: This description is a bit funny to me, either add `, but also accepts an OnError handler` or perhaps just write the full description? It does lead to documentation being copied, but that is IMO OK as you want the documentation as local as possible.

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -1157,6 +1158,147 @@ public void run() {
...
+     */
+    public RealmAsyncTask executeTransactionAsync(final Transaction transaction, final Realm.Transaction.OnError onError) {
+        if (onError == null)
+            throw new IllegalArgumentException("onError callback can't be null");
+
+        return executeTransactionAsync(transaction, null, onError);
+    }
+
+    /**
+     * Similar to {@link #executeTransactionAsync(Transaction)}.
```

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -1157,6 +1158,147 @@ public void run() {
...
+     *
+     * @param transaction {@link io.realm.Realm.Transaction} to execute.
+     * @param onSuccess callback invoked when the transaction succeeds.
+     * @param onError callback invoked when the transaction failed.
+     * @return a {@link RealmAsyncTask} representing a cancellable task.
+     * @throws IllegalArgumentException if the {@code transaction} is {@code null}, or if the realm is opened from another thread.
+     */
+    public RealmAsyncTask executeTransactionAsync(final Transaction transaction, final Realm.Transaction.OnSuccess onSuccess, final Realm.Transaction.OnError onError) {
+        if (transaction == null)
```
cmelchior: Bracers missing

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -1157,6 +1158,147 @@ public void run() {
...
+            throw new IllegalArgumentException("Transaction should not be null");
+
+        // If the user provided a Callback then we make sure, the current Realm has a Handler
+        // we can use to deliver the result
+        if ((onSuccess != null || onError != null)  && handler == null) {
+            throw new IllegalStateException("Your Realm is opened from a thread without a Looper" +
+                    " (or it has already been closed) and you provided a callback, we need a Handler to invoke your callback");
```
cmelchior: Why not check for `realm.isValid` instead of adding the description in `()`?

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -1157,6 +1158,147 @@ public void run() {
...
+                            && handler.getLooper().getThread().isAlive()) {
+                        if (onSuccess != null && transactionCommitted) {
+                            handler.post(new Runnable() {
+                                @Override
+                                public void run() {
+                                    onSuccess.onSuccess();
+                                }
+                            });
+                        }
+                        if (exception[0] != null) {
```
cmelchior: This could be pulled into its own variable. Will make the below code more readable.

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -1326,5 +1468,21 @@ public static Object getDefaultModule() {
...
+        /**
+         * Callback invoked to notify the caller thread about the success of the transaction.
+         */
+        interface OnSuccess {
+            void onSuccess();
+        }
+
+        /**
+         * Callback invoked to notify the caller thread about error during the transaction.
+         * The transaction will be rolled back and the background Real will be closed before
```
cmelchior: Real -> Realm

cmelchior:
Done, mostly minor stuff, but there where also a few comments left unaddressed from last round.

nhachicha:
updated PR

realm/realm-library/src/androidTest/java/io/realm/RealmAsyncQueryTests.java
```diff
@@ -81,24 +81,84 @@ public void setUp() throws Exception {
...
+            @Override
+            public void onSuccess() {
+                assertEquals(1, looperThread.realm.allObjects(Owner.class).size());
+                assertEquals("Owner", looperThread.realm.where(Owner.class).findFirst().getName());
+                looperThread.testComplete();
+            }
+        });
+    }
+
+    @RunTestInLooperThread
```
beeender: `@Test` is missing.

realm/realm-library/src/androidTest/java/io/realm/RealmAsyncQueryTests.java
```diff
@@ -81,24 +81,84 @@ public void setUp() throws Exception {
...
-                fail(e.getMessage());
+            public void onError(Throwable error) {
+                fail(error.getMessage());
+            }
+        });
+    }
+
+    @Test
+    @RunTestInLooperThread
+    public void executeTransaction_async_onSuccess() throws Throwable {
```
beeender: better to name it as `executeTransactionAsync_onSuccess` ?

realm/realm-library/src/androidTest/java/io/realm/RealmAsyncQueryTests.java
```diff
@@ -81,24 +81,84 @@ public void setUp() throws Exception {
...
+            public void onSuccess() {
+                assertEquals(1, looperThread.realm.allObjects(Owner.class).size());
+                assertEquals("Owner", looperThread.realm.where(Owner.class).findFirst().getName());
+                looperThread.testComplete();
+            }
+        });
+    }
+
+    @RunTestInLooperThread
+    public void executeTransaction_async_onError() throws Throwable {
```
beeender: `executeTransactionAsync_onSuccess`

realm/realm-library/src/androidTest/java/io/realm/RealmAsyncQueryTests.java
```diff
@@ -81,24 +81,84 @@ public void setUp() throws Exception {
...
+                assertEquals(0, looperThread.realm.allObjects(Owner.class).size());
+                assertNull(looperThread.realm.where(Owner.class).findFirst());
+                looperThread.testComplete();
+            }
+        });
+    }
+
+    @Test
+    @RunTestInLooperThread
+    public void executeTransaction_async_NoCallbacks() throws Throwable {
```
beeender: `executeTransactionAsync_NoCallbacks`

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -1157,6 +1160,153 @@ public void run() {
...
+    /**
+     * Similar to {@link #executeTransactionAsync(Transaction)}, but also accepts an OnSuccess callback.
+     *
+     * @param transaction {@link io.realm.Realm.Transaction} to execute.
+     * @param onSuccess callback invoked when the transaction succeeds.
+     * @return a {@link RealmAsyncTask} representing a cancellable task.
+     * @throws IllegalArgumentException if the {@code transaction} is {@code null}, or if the realm is opened from another thread.
+     */
+    public RealmAsyncTask executeTransactionAsync(final Transaction transaction, final Realm.Transaction.OnSuccess onSuccess) {
+        if (onSuccess == null){
```
beeender: space between `){`

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -1157,6 +1160,153 @@ public void run() {
...
+                        if (backgroundException != null) {
+                            if (onError != null) {
+                                handler.post(new Runnable() {
+                                    @Override
+                                    public void run() {
+                                        onError.onError(backgroundException);
+                                    }
+                                });
+                            } else {
+                                handler.post(new Runnable() {
```
beeender: Should it just throw in the worker thread? And need a test case for this.

realm/realm-library/src/androidTest/java/io/realm/RealmAsyncQueryTests.java
```diff
@@ -78,34 +78,95 @@ public void setUp() throws Exception {
...
+            @Override
+            public void execute(Realm realm) {
+                Owner owner = realm.createObject(Owner.class);
+                owner.setName("Owner");
+            }
+        });
+        looperThread.realm.addChangeListener(new RealmChangeListener() {
+            @Override
+            public void onChange() {
+                looperThread.testComplete();
```
cmelchior: Should you verify that 1 owner object exists here?

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -1157,6 +1160,160 @@ public void run() {
...
+                                    @Override
+                                    public void run() {
+                                        throw new RuntimeException(backgroundException);
```
cmelchior: We shouldn't wrap exceptions in `RuntimeException` you might convert `Error` to a `RuntimeException`?

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -1157,6 +1160,160 @@ public void run() {
...
+                                    }
+                                });
+                            }
+                        }
+                    } else {
+                        // Throw exception in the worker thread if the caller thread terminated
+                        if (backgroundException != null) {
+                            //noinspection ThrowFromFinallyBlock
+                            throw new RuntimeException(backgroundException);
```
cmelchior: Same. Why convert to RuntimeException?

realm/realm-library/src/main/java/io/realm/Realm.java
cmelchior: There isn't any difference between this and triggering the error callback above right? I guess that technically you can quit a looper thread before it finished receiving all events, but by that assumption we couldn't propagate any errors. I would probably say that chance of this happening is so little, and it would be against the contract you agreed to when doing it async, so basically "it is your own fault"?

cmelchior:
Just discovered that we convert exceptions to RuntimeExceptions which looks like a bad idea? Apart from that :+1:

nhachicha:
@realm/java

kneth:
That's a lot of conditions to keep track of - but it looks just fine to me :smile:

cmelchior:
:+1: Once master has been merged into this :)

nhachicha:
squashed

PR #2045, opened 2016-01-08
Test async transaction when async query existing

beeender:
To reproduce #1893

PR #2164, opened 2016-01-26
Fix async transaction problem when async query already exist.

### LINKED ISSUE #1893: onSuccess might not be called with the right Realm version for async transaction

dalinaum:
Fix #1893 It imports #2045 for test.

### DIFF
--- a/realm/realm-library/src/androidTest/java/io/realm/RealmAsyncQueryTests.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/RealmAsyncQueryTests.java
@@ -31,6 +31,7 @@ import java.util.Date;
 import java.util.Iterator;
 import java.util.Map;
 import java.util.Random;
+import java.util.concurrent.Callable;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.atomic.AtomicInteger;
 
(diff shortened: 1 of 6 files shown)


dalinaum:

realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -66,6 +67,7 @@
     RealmSchema schema;
     Handler handler;
     HandlerController handlerController;
+    List<Runnable> asyncTransactionCallbacks = new ArrayList<>();
```
zaki50: we should avoid diamond operator

realm/realm-library/src/main/java/io/realm/HandlerController.java
```diff
@@ -509,6 +509,13 @@ private void completedAsyncQueriesUpdate(QueryUpdateTask.Result result) {
 
             updateAsyncQueriesTask = null;
         }
+
+        if (realm.asyncTransactionCallbacks.size() > 0) {
```
zaki50: Use `!isEmpty()` instead of `size() > 0`

dalinaum: Updated. Thanks.

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -1138,87 +1138,19 @@ public void executeTransaction(Transaction transaction) {
      */
     @Deprecated
     public RealmAsyncTask executeTransaction(final Transaction transaction, final Transaction.Callback callback) {
```
cmelchior: Perhaps just delete this instead? It has been deprecated for a while now?

realm/realm-library/src/androidTest/java/io/realm/RealmAsyncQueryTests.java
```diff
@@ -1947,6 +1947,36 @@ public void onChange() {
...
     }
 
+    @Test
+    @RunTestInLooperThread
+    public void testAsyncTransactionWorksWithAsyncQuery() {
+        RealmResults<AllTypes> results = looperThread.realm.where(AllTypes.class).findAllAsync();
+        results.addChangeListener(new RealmChangeListener() {
+            @Override
+            public void onChange() {
+                //assertEquals(looperThread.realm.where(AllTypes.class).count(), 1);
```
cmelchior: If we don't use this listener, perhaps just remove it?

realm/realm-library/src/androidTest/java/io/realm/RealmAsyncQueryTests.java
```diff
@@ -1947,6 +1947,36 @@ public void onChange() {
         result.load();
     }
 
+    @Test
+    @RunTestInLooperThread
+    public void testAsyncTransactionWorksWithAsyncQuery() {
+        RealmResults<AllTypes> results = looperThread.realm.where(AllTypes.class).findAllAsync();
```
cmelchior: A pattern we have used elsewhere is:
```
Realm realm = looperThread.realm;
```

It makes it slightly more readable, and it is easier to replace the Realm instance if we need to.

realm/realm-library/src/androidTest/java/io/realm/RealmAsyncQueryTests.java
```diff
@@ -1947,6 +1947,36 @@ public void onChange() {
         result.load();
     }
 
+    @Test
+    @RunTestInLooperThread
+    public void testAsyncTransactionWorksWithAsyncQuery() {
```
cmelchior: I could use a comment about what this unit test is for. You are doing nothing with the `findAllAsync` result as far as I can tell?

cmelchior:
This PR only seem to solve half the problem? e.g the problem described here http://stackoverflow.com/questions/34938152/callback-never-be-invoked-when-i-excute-a-asynchronous-transaction-with-realm is not addressed

Looking at it more closely it looks to me like #1893 and #2045 is two different issues? I'm fine with this PR just solving #2045 but we shouldn't close #1893 in that case. Thoughts @beeender ?

dalinaum:
Sorry. I almost forgot that #1893 are actually two issues. I will add logging a warning tonight.

dalinaum:
Use #2780 instead.

PR #2174, opened 2016-01-28
Convert migration and cache tests to JUNIT4

beeender:


realm/realm-library/src/androidTest/java/io/realm/RealmMigrationTests.java
```diff
@@ -30,24 +38,38 @@
...
-    public Realm realm;
+    private Realm realm;
+    private Context context;
 
-    @Override
-    protected void tearDown() throws Exception {
-        super.tearDown();
+    @Before
+    public void setup() {
+        context = InstrumentationRegistry.getInstrumentation().getContext();
```
cmelchior: This context doesn't look like it is getting used?

realm/realm-library/src/androidTest/java/io/realm/rule/TestRealmConfigurationFactory.java
```diff
@@ -109,4 +116,20 @@ public RealmConfiguration createConfiguration(String name, byte[] key) {
     public RealmConfiguration.Builder createConfigurationBuilder() {
         return new RealmConfiguration.Builder(getRoot());
     }
+
+    // Copies a Realm file from assets to temp dir
+    public void copyRealmFromAssets(Context context, String realmPath, String newName)
```
cmelchior: This is also in `TestHelper`, but I agree it makes more sense here. Can we delete it outright from `TestHelper` or do we need to deprecate it first until all classes have moved to use `TestRealmConfigurationFactory`?

cmelchior: This is also in `TestHelper`, but I agree it makes more sense here. Can we delete it outright from `TestHelper` or do we need to deprecate it first until all classes have moved to use `TestRealmConfigurationFactory`?

cmelchior:
Apart from a small comment :+1:

realm/realm-library/src/androidTest/java/io/realm/rule/TestRealmConfigurationFactory.java
beeender: The one in TestHelper is being used IOSRealmTest. No need to bother mark it as deprecated. I will just convert that on later today, just a few lines of code.

cmelchior: :+1:

cmelchior: :+1:

stk1m1:
:+1: I've found few tricks I could use from the conversion. Thanks!

zaki50:
:+1:

PR #2176, opened 2016-01-28
Convert IOSRealmTests and JNITransaction tests

beeender:


realm/realm-library/src/androidTest/java/io/realm/rule/TestRealmConfigurationFactory.java
```diff
@@ -120,6 +120,12 @@ public RealmConfiguration createConfiguration(String name, byte[] key) {
     // Copies a Realm file from assets to temp dir
     public void copyRealmFromAssets(Context context, String realmPath, String newName)
             throws IOException {
+        // Delete the existing file before copy
+        RealmConfiguration configToDelete = new RealmConfiguration.Builder(getRoot())
```
cmelchior: Is this needed. Won't it just be overwritten if it exists?

cmelchior:
Apart from comment :+1:

realm/realm-library/src/androidTest/java/io/realm/rule/TestRealmConfigurationFactory.java
beeender: It is. The copying is done by opening the file and opening the output stream to write, there won't be a new file to be created, it is just the old file being written. If you have an opened Realm instance, the db file will be overwritten and you will see all kinds of strange crashes.

So, just try to delete the Realm and it will throw if any instance is not closed.

cmelchior: :+1:

stk1m1:
:+1:

PR #2107, opened 2016-01-18
RealmQuery, RealmResults distinct() pt1

stk1m1:
A new PR for #1568

PR #2198, opened 2016-01-29
RealmQuery 'distinct()' TableView Patch

stk1m1:
`RealmResults` gets `distinct()` results from its backing `TableView`.

fixes a bug introduced in #2107

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -9,6 +9,7 @@
  * Improved .so loading by using ReLinker (https://github.com/KeepSafe/ReLinker).
  * Improved performance of RealmList#contains() (#897).
  * Added RealmQuery.distinct() and RealmResults.distinct() (#1568).
+ * Fixed bug when multiple calls of RealmResults.distinct() causes to return wrong results (#2198).
 
 0.87.4
  * Updated Realm Core to 0.96.0
--- a/realm/realm-library/src/androidTest/java/io/realm/RealmResultsTests.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/RealmResultsTests.java
@@ -1311,6 +1311,27 @@ public class RealmResultsTests {
     }
 
     @Test
+    public void distinct_restrictedByPreviousDistinct() {
+        final long numberOfBlocks = 25;
+        final long numberOfObjects = 10;
+        populateForDistinct(realm, numberOfBlocks, numberOfObjects, false);
+
+        // all objects
+        RealmResults<AnnotationIndexTypes> allResults = realm.where(AnnotationIndexTypes.class).findAll();
+        assertEquals("All Objects Count", numberOfBlocks * numberOfBlocks * numberOfObjects, allResults.size());
+        // distinctive dates
+        RealmResults<AnnotationIndexTypes> distinctDates = allResults.distinct("indexDate");
+        assertEquals("Distinctive Dates", numberOfBlocks, distinctDates.size());
+        // distinctive Booleans
+        RealmResults<AnnotationIndexTypes> distinctBooleans = distinctDates.distinct("indexBoolean");
+        assertEquals("Distinctive Booleans", 2, distinctBooleans.size());
+        // all three results are the same object
+        assertTrue(allResults == distinctDates);
+        assertTrue(allResults == distinctBooleans);
+        assertTrue(distinctDates == distinctBooleans);
+    }
+
+    @Test
     public void distinct_withNull() {
         final long numberOfBlocks = 25;
         final long numberOfObjects = 10; // must be greater than 1
--- a/realm/realm-library/src/main/java/io/realm/RealmResults.java
+++ b/realm/realm-library/src/main/java/io/realm/RealmResults.java
@@ -576,22 +576,14 @@ public final class RealmResults<E extends RealmObject> extends AbstractList<E> {
     public RealmResults<E> distinct(String fieldName) {
         realm.checkIfValid();
         long columnIndex = getColumnIndex(fieldName);
+
         TableOrView tableOrView = getTable();
-
-        TableView tableView;
         if (tableOrView instanceof Table) {
-            tableView = ((Table) tableOrView).getDistinctView(columnIndex);
+            this.table = ((Table) tableOrView).getDistinctView(columnIndex);
         } else {
-            tableView = ((TableView) tableOrView).getTable().getDistinctView(columnIndex);
+            ((TableView) tableOrView).distinct(columnIndex);
         }
-
-        RealmResults<E> realmResults;
-        if (realm instanceof DynamicRealm) {
-            realmResults =  (RealmResults<E>) RealmResults.createFromDynamicTableOrView(realm, tableView, className);
-        } else {
-            realmResults = RealmResults.createFromTableOrView(realm, tableView, classSpec);
-        }
-        return realmResults;
+        return this;
     }
 
     // Deleting


stk1m1:
@realm/java please review.

realm/realm-library/src/main/java/io/realm/internal/TableView.java
```diff
@@ -68,6 +68,19 @@ protected TableView(Context context, Table parent, long nativePtr, TableQuery qu
...
+     * Creates a copy of already created Java TableView Object.
+     * The method is not supposed to be called by the user of the db. The method is for internal use only.
+     *
+     * @param tableView An existing TableView
+     */
+    protected TableView(TableView tableView) {
+        this.context = tableView.context;
+        this.parent = tableView.parent;
+        this.nativePtr = nativeGetCopy(tableView.nativePtr);
+        this.query = tableView.query;
```
cmelchior: This seems a bit dangerous, as that query reference is now used in two tableviews that might be used independently. 
This should probably be refactored to something like `nativeGetQuery()` or something like that (not sure if it already exists).

realm/realm-jni/src/io_realm_internal_tableview.cpp
```diff
@@ -131,6 +131,18 @@ JNIEXPORT void JNICALL Java_io_realm_internal_TableView_nativeClose(
     delete TV(nativeViewPtr);
 }
 
+JNIEXPORT jlong JNICALL Java_io_realm_internal_TableView_nativeGetCopy(
+    JNIEnv* env, jobject, jlong nativeViewPtr)
+{
+    if (!VIEW_VALID_AND_IN_SYNC(env, nativeViewPtr))
+        return 0;
+    try {
+        TableView* copiedTableView = new TableView(*TV(nativeViewPtr));
```
kneth: Does the `TableView` constructor do a deep copy? If not, `TavleView copied = TV(nativeViewPtr)->get_parent().where(TV(nativeViewPtr)).find_all()` be safer.

stk1m1: @kneth I think it does. There is a inliner. But I'll check with the team to make sure of it.

stk1m1:
retest this please

stk1m1:
retest this please

realm/realm-jni/src/io_realm_internal_tableview.cpp
cmelchior: I doubt that doing a new `find_all` would be accurate. The TableView might be restricted by some query. I am pretty sure that copying a TableView will also copy any parent views/queries, but @rrrlasse / @finnschiermer can probably confirm that.

realm/realm-library/src/androidTest/java/io/realm/RealmResultsTests.java
```diff
@@ -1311,6 +1311,21 @@ public void distinct() {
     }
 
     @Test
+    public void distinct_resultsWithTableView() {
```
cmelchior: The name of this unit test could be clearer IMO. `distinct_restrictedByPreviousDistinct` or something like that would be clearer.

realm/realm-library/src/androidTest/java/io/realm/RealmResultsTests.java
```diff
@@ -1311,6 +1311,21 @@ public void distinct() {
...
+        final long numberOfBlocks = 25;
+        final long numberOfObjects = 10; // must be greater than 1
+        populateForDistinct(realm, numberOfBlocks, numberOfObjects, false);
+
+        RealmResults<AnnotationIndexTypes> rootResults = realm.where(AnnotationIndexTypes.class).findAll();
+        assertEquals("Total Count", numberOfBlocks * numberOfBlocks * numberOfObjects, rootResults.size());
+        RealmResults<AnnotationIndexTypes> parentResults = rootResults.distinct("indexDate");
```
cmelchior: I would add some more comments in this code, preferably above each `RealmResults` to indicate what is happening. Right now the code is very dense and hard to read at a glance.

realm/realm-library/src/main/java/io/realm/RealmResults.java
```diff
@@ -582,7 +582,8 @@ public double average(String fieldName) {
         if (tableOrView instanceof Table) {
             tableView = ((Table) tableOrView).getDistinctView(columnIndex);
         } else {
-            tableView = ((TableView) tableOrView).getTable().getDistinctView(columnIndex);
+            tableView = ((TableView) tableOrView).getCopy();
```
cmelchior: I am really surprised that we need to copy a tableview to apply multiple distinct parameters. It sounds like a bug in Core to me or at least a future enhancement. We should verify what the intended behaviour of multiple distinct calls is.

realm/realm-jni/src/io_realm_internal_tableview.cpp
bmunkholm: If it's not clear from the documentation in the headerfile of core, make sure to make a PR/Issue in core to properly document the behaviour.

realm/realm-library/src/main/java/io/realm/internal/TableView.java
```diff
@@ -68,6 +68,28 @@ protected TableView(Context context, Table parent, long nativePtr, TableQuery qu
...
     }
 
+    /**
+     * Creates a copy of already created Java TableView Object.
+     * The method is not supposed to be called by the user of the db. The method is for internal use only.
+     *
+     * @param tableView An existing TableView
+     * @param nativeViewPtr A pointer to a clone of the existing TableView.
+     */
+    protected TableView(TableView tableView, long nativeViewPtr) {
```
cmelchior: Why `protected` and not `private` ?
I would rename the variables to be more telling: `tableView` -> `originalTableView`  and `nativeViewPtr` -> `newTableViewPtr` or something like that.

realm/realm-library/src/main/java/io/realm/internal/TableView.java
```diff
@@ -68,6 +68,28 @@ protected TableView(Context context, Table parent, long nativePtr, TableQuery qu
...
+        long nativeQueryPtr = nativeWhere(nativeViewPtr);
+        TableQuery q;
+        try {
+            q = new TableQuery(tableView.context, tableView.parent, nativeQueryPtr, this);
+        } catch (RuntimeException e) {
```
cmelchior: What errors is this catching? If the native code threw any exception we should close the resource there.

realm/realm-library/src/main/java/io/realm/internal/TableView.java
```diff
@@ -804,12 +826,30 @@ public void distinct(long columnIndex) {
...
+     *
+     * @return A copy of the existing TableView Object.
+     */
+    public TableView getCopy() {
+        // Execute the disposal of abandoned realm objects each time a new realm object is created
+        this.context.executeDelayedDisposal();
+        long nativeViewPtr = nativeGetCopy(this.nativePtr);
+        try {
+            return new TableView(this, nativeViewPtr);
+        } catch (RuntimeException e) {
```
cmelchior: Same

cmelchior:
We also need a changelog update.

stk1m1:
**What is done**
1. `RealmResults.distinct()` does **NOT** spawn a new instance. `RealmResults` simply selects distinctive elements on its backing `TableView` and returns itself.  
2. In case a `RealmResults` is backed up by a `Table`, a new distinctive `TableView` replaces `table` field of `RealmResults`.

@realm/java review please.

realm/realm-jni/src/io_realm_internal_tableview.cpp
rrrlasse: @kneth yes, it performs a copy which gives you a new valid independent tableview, while keeping the original tableview valid and independent as well. So kindof a deep copy.

However, just note that the two tableviews can only be used from the same thread.

changelog.txt
```diff
@@ -9,6 +9,7 @@
  * Improved .so loading by using ReLinker (https://github.com/KeepSafe/ReLinker).
  * Improved performance of RealmList#contains() (#897).
  * Added RealmQuery.distinct() and RealmResults.distinct() (#1568).
+ * Fixed bug when multiple calls of RealmResults.distinct() causes to return wrong results (#2198)
```
kneth: Please add `.` at the end.

kneth:
I do find this a bit counter-intuitive. If I do 

    RealmResults<Person> allPersons = realm.allObjects(Person.class);
    long nPerons = allPersons.size();
    RealmResults<Person> uniqPersons = allPersons.distinct("name");

then `nPersons == allPersons.size()` will be false.

stk1m1:
@kneth That's the main thought behind the all previous commits. i.e. let every `RealmResults` reside in its own world view (`TableView`). In order to do that, we should give it a separate `TableView` and a `TableQuery`. It consequently requires a copy of the previous `TableView`. 

It is reverted now. Could you be kind enough to let me have your :+1: btw?

realm/realm-library/src/main/java/io/realm/RealmResults.java
```diff
@@ -576,22 +576,15 @@ public double average(String fieldName) {
...
-            tableView = ((Table) tableOrView).getDistinctView(columnIndex);
-        } else {
-            tableView = ((TableView) tableOrView).getTable().getDistinctView(columnIndex);
-        }
-
-        RealmResults<E> realmResults;
-        if (realm instanceof DynamicRealm) {
-            realmResults =  (RealmResults<E>) RealmResults.createFromDynamicTableOrView(realm, tableView, className);
+            TableView tableView = ((Table) tableOrView).getDistinctView(columnIndex);
+            this.table = tableView;
```
cmelchior: Why not just set the `this.table` directly?

stk1m1: @cmelchior I left it for the compiler to work on it ;) 
It's just that I'm not a big fan of assigning a new instance to a field right away.

cmelchior: It doesn't matter much in this case, but why not?

cmelchior:
Looks good to me :+1:

kneth:
:+1:

beeender:
:+1:

PR #2201, opened 2016-01-29
Adding javadoc to explain that items will be continually be emitted u…

### LINKED ISSUE #2135: Update `asObservable` JavaDoc to mention that they keep emitting and never call onComplete
I request only once to my method that uses realm query with asObservable.
(I saw Log for sure that I requested really once, and it's true.)

But subscribe event from the query which is observable came multiple, twice sometimes triple.

Besides, one of the subscribe events gives me wrong results as 0, even if table has some datas.

How can I get subscribe event only once when I request once query with observable?

donnfelker:
…nless other wise instructed with the first() operator.

Closes #2135

### DIFF
--- a/realm/realm-library/src/main/java/io/realm/BaseRealm.java
+++ b/realm/realm-library/src/main/java/io/realm/BaseRealm.java
@@ -167,8 +167,18 @@ abstract class BaseRealm implements Closeable {
     }
 
     /**
-     * Returns an Rx Observable that monitors changes to this Realm. It will emit the current state when subscribed
-     * to.
+     * Returns an RxJava Observable that monitors changes to this Realm. It will emit the current state
+     * when subscribed to. Items will continually be emitted as the Realm is updated -
+     * {@code onComplete} will never be called.
+     *
+     * If you would like the {@code asObservable()} to stop emitting items you can instruct RxJava to
+     * only emit only the first item by using the {@code first()} operator:
+     *
+     * <pre>
+     * {@code
+     * realm.asObservable().first().subscribe( ... ) // You only get the results once
+     * }
+     * </pre>
      *
      * @return RxJava Observable that only calls {@code onNext}. It will never call {@code onComplete} or {@code OnError}.
      * @throws UnsupportedOperationException if the required RxJava framework is not on the classpath.
--- a/realm/realm-library/src/main/java/io/realm/RealmObject.java
+++ b/realm/realm-library/src/main/java/io/realm/RealmObject.java
@@ -264,11 +264,24 @@ public abstract class RealmObject {
     }
 
     /**
-     * Returns an Rx Observable that monitors changes to this RealmObject. It will emit the current object when
-     * subscribed to.
+     * Returns an RxJava Observable that monitors changes to this RealmObject. It will emit the current object when
+     * subscribed to. Object updates will continually be emitted as the RealmObject is updated -
+     * {@code onComplete} will never be called.
      *
      * If chaining a RealmObject observable use {@code obj.<MyRealmObjectClass>asObservable()} to pass on
      * type information, otherwise the type of the following observables will be {@code RealmObject}.
+     *
+     * If you would like the {@code asObservable()} to stop emitting items you can instruct RxJava to
+     * only emit only the first item by using the {@code first()} operator:
+     *
+     * <pre>
+     * {@code
+     * obj.asObservable()
+     *      .filter(obj -> obj.isLoaded())
+     *      .first()
+     *      .subscribe( ... ) // You only get the object once
+     * }
+     * </pre>
      *
      * @param <E> RealmObject class that is being observed. Must be this class or its super types.
      * @return RxJava Observable that only calls {@code onNext}. It will never call {@code onComplete} or {@code OnError}.
--- a/realm/realm-library/src/main/java/io/realm/RealmResults.java
+++ b/realm/realm-library/src/main/java/io/realm/RealmResults.java
@@ -900,7 +900,20 @@ public final class RealmResults<E extends RealmObject> extends AbstractList<E> {
 
     /**
      * Returns an Rx Observable that monitors changes to this RealmResults. It will emit the current RealmResults when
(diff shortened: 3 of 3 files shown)


realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -167,8 +167,21 @@ public void removeChangeListener(RealmChangeListener listener) {
     }
 
     /**
-     * Returns an Rx Observable that monitors changes to this Realm. It will emit the current state when subscribed
-     * to.
+     * Returns an Rx Observable that monitors changes to this Realm. It will emit the current state
+     * when subscribed to. Items will continually be emitted as the Realm is updated -
+     * {@code onComplete} will never be called.
+     *
+     * If you would like the {@code asObservable()} to stop emitting items you can instruct Rx to
```
cmelchior: Is it RxJava or just Rx ?

realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -167,8 +167,21 @@ public void removeChangeListener(RealmChangeListener listener) {
...
+     * only emit only the first item by using the {@code first()} operator:
+     *
+     * <pre>
```
cmelchior: Do you need both `<pre>` and `{@code` I thought it was the same thing?

cmelchior:
I like it, and great with the example as well. We need to update the other `asObservable` methods as well though:  `RealmResults`, `RealmObject`

realm/realm-library/src/main/java/io/realm/BaseRealm.java
donnfelker: I should make it RxJava. Will fix.

realm/realm-library/src/main/java/io/realm/BaseRealm.java
donnfelker: Which do you prefer to use?

donnfelker:
Ahh, will do. I'll update the other classes.

donnfelker:
@cmelchior Updated the RealmObject and RealmResults. Let me know about the code formatting <pre> and {@code} preference and I'll update it.

cmelchior:
Looks like I was wrong, It seems that using both `<pre>` and `{@code` is correct: http://stackoverflow.com/questions/541920/multiple-line-code-example-in-javadoc-comment.

realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -167,8 +167,21 @@ public void removeChangeListener(RealmChangeListener listener) {
...
+     * Returns an RxJava Observable that monitors changes to this Realm. It will emit the current state
+     * when subscribed to. Items will continually be emitted as the Realm is updated -
+     * {@code onComplete} will never be called.
+     *
+     * If you would like the {@code asObservable()} to stop emitting items you can instruct RxJava to
+     * only emit only the first item by using the {@code first()} operator:
+     *
+     * <pre>
+     * {@code
+     * realm.where(Foo.class).findAllAsync().asObservable()
```
cmelchior: I would adjust the example to fit the context. In this one, we are observing on a Realm:

```
realm.asObservable().first()
```

realm/realm-library/src/main/java/io/realm/RealmObject.java
```diff
@@ -264,12 +264,25 @@ public final void removeChangeListeners() {
...
      * If chaining a RealmObject observable use {@code obj.<MyRealmObjectClass>asObservable()} to pass on
      * type information, otherwise the type of the following observables will be {@code RealmObject}.
      *
+     * If you would like the {@code asObservable()} to stop emitting items you can instruct RxJava to
+     * only emit only the first item by using the {@code first()} operator:
+     *
+     * <pre>
+     * {@code
+     * obj.asObservable()
+     *      .filter(obj->isLoaded())
```
cmelchior: Isn't the proper syntax `obj::isLoaded` or `obj -> obj.isLoaded()` ?

realm/realm-library/src/main/java/io/realm/RealmResults.java
```diff
@@ -900,7 +900,20 @@ public void removeChangeListeners() {
...
-     * subscribed to.
+     * subscribed to. RealmResults will continually be emitted as the RealmResults are updated -
+     * {@code onComplete} will never be called.
+     *
+     * If you would like the {@code asObservable()} to stop emitting items you can instruct RxJava to
+     * only emit only the first item by using the {@code first()} operator:
+     *
+     * <pre>
+     * {@code
+     * results.asObservable()
```
cmelchior: I would use the async example from BaseRealm here instead.

realm/realm-library/src/main/java/io/realm/RealmResults.java
```diff
@@ -900,7 +900,20 @@ public void removeChangeListeners() {
...
+     *      .filter(results->isLoaded())
```
cmelchior: Same, correct syntax?

realm/realm-library/src/main/java/io/realm/RealmObject.java
donnfelker: You're right, it was the second one. :)

donnfelker:
@cmelchior All requested changes have been made. :)

cmelchior:
Looks good to me. :+1: once squashed.
@realm/java

beeender:
:+1:

donnfelker:
@cmelchior Squashed.

cmelchior:
Hmm, looks like a regression on our part somewhere: io.realm.RealmSchemaTests.remove_classWithReferencesThrows (from io.realm.DynamicRealmObjectTests)

cmelchior:
Retest this please

donnfelker:
Do you need me to do anything to help push this along?

cmelchior:
No, this PR shouldn't cause any build failure. I'll look into it.

PR #2205, opened 2016-02-01
Converted TypedBasedNotificationTests to JUnit4

cmelchior:
@realm/java


beeender:
https://github.com/realm/realm-java/blob/cm/junit4-typedbased/realm/realm-library/src/androidTest/java/io/realm/TypeBasedNotificationsTests.java#L651
No need of `try` block i think? `looperThread.testComplete();` should be called after other 3 looper thread quit. Otherwise it is a risk that test is finished, but the other 3 looper threads are not.

beeender:
Except one comments, :+1:

cmelchior:
Fixed

beeender:
:+1:

zaki50:
:+1:

PR #2225, opened 2016-02-03
Bugfix to catch RuntimeException during native resource allocation in TableView

stk1m1:
Native function calls `native*` returning a Java instance in `TableView` are now inside _try-catch_ block.

This is to catch `RuntimeException` that might happen during native resource allocation.

### DIFF
--- a/realm/realm-library/src/main/java/io/realm/internal/TableView.java
+++ b/realm/realm-library/src/main/java/io/realm/internal/TableView.java
@@ -510,8 +510,8 @@ public class TableView implements TableOrView, Closeable {
     public TableView findAllLong(long columnIndex, long value){
         // Execute the disposal of abandoned realm objects each time a new realm object is created
         context.executeDelayedDisposal();
-        long nativeViewPtr = nativeFindAllInt(nativePtr, columnIndex, value);
-        try { 
+        try {
+            long nativeViewPtr = nativeFindAllInt(nativePtr, columnIndex, value);
             return new TableView(this.context, this.parent, nativeViewPtr);
         } catch (RuntimeException e) {
             TableView.nativeClose(nativeViewPtr);
@@ -523,8 +523,8 @@ public class TableView implements TableOrView, Closeable {
     public TableView findAllBoolean(long columnIndex, boolean value) {
         // Execute the disposal of abandoned realm objects each time a new realm object is created
         context.executeDelayedDisposal();
-        long nativeViewPtr = nativeFindAllBool(nativePtr, columnIndex, value);
-        try { 
+        try {
+            long nativeViewPtr = nativeFindAllBool(nativePtr, columnIndex, value);
             return new TableView(this.context, this.parent, nativeViewPtr);
         } catch (RuntimeException e) {
             TableView.nativeClose(nativeViewPtr);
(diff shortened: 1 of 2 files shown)


cmelchior:
Isn't the only failure from native resource allocation OutOfMemoryError? Which A) this won't catch and B) we shouldn't catch it as it is a fatal state to get into? /cc @kneth

stk1m1:
@realm/java please review.

realm/realm-library/src/main/java/io/realm/internal/TableView.java
```diff
@@ -510,8 +510,9 @@ public long upperBoundLong(long columnIndex, long value) {
     public TableView findAllLong(long columnIndex, long value){
         // Execute the disposal of abandoned realm objects each time a new realm object is created
         context.executeDelayedDisposal();
-        long nativeViewPtr = nativeFindAllInt(nativePtr, columnIndex, value);
-        try { 
+        long nativeViewPtr = 0;
+        try {
+            nativeViewPtr = nativeFindAllInt(nativePtr, columnIndex, value);
```
beeender: Is this necessary? When this happens, the `new` on `TableView` will fail, which means `nativeViewPtr` is still `0`. `TableView.nativeClose(0)` basically is doing nothing. This actually won't change anything i think?

stk1m1: @beeender A `nativeViewPtr` before `try-catch` block is just a placeholder, initialized to have a null-value for 1) steering clear of compiler warning and 2) providing a measure to check if native resource indeed is allocated even if `new TableView`fails.

 ( the `native*` series in this PR allocates native resource within them. we can further check with if `nativeViewPtr` is really null.)

cmelchior: Hmm, not sure I understand that because `long` can never be `null`, did you mean `0`?

stk1m1: Yes. I'll update the comment.

beeender:
```
JNIEXPORT jlong JNICALL Java_io_realm_internal_TableView_nativeFindAllInt(
    JNIEnv* env, jobject, jlong nativeViewPtr, jlong columnIndex, jlong value)
{
    try {
        if (!VIEW_VALID_AND_IN_SYNC(env, nativeViewPtr) ||
            !COL_INDEX_AND_TYPE_VALID(env, TV(nativeViewPtr), columnIndex, type_Int))
            return 0;
        TableView* pResultView = new TableView( TV(nativeViewPtr)->find_all_int( S(columnIndex), value) );
        return reinterpret_cast<jlong>(pResultView);
    } CATCH_STD()
    return 0;
}
```
For the above code, only `new TableView` will throw. and when that throws, `nativeClose` doesn't have to be called and the `RealmException` will be thrown anyway.

Or do I misunderstand the purpose of this change?

cmelchior: I agree, I am not sure that these changes will prevent anything. 
Also, if a change isn't accompanied by unit tests, please add a description explaining the rationale as we have to be extra careful reviewing those changes.

stk1m1: @beeender You're right. I just looked into `CATCH_STD()` and realized that this change isn't actually covering anything. The `native*` calls can live outside as is now. 

Thanks for the pointer!

PR #2233, opened 2016-02-05
All changelisteners are now delayed until next looper event

cmelchior:
Prerequisite for https://github.com/realm/realm-java/pull/2124

This PR changes how change listeners work. Currently we will trigger all synchronous change listeners on the same thread when doing a `commitTransaction` on that thread.

This is unfortunate when #2124 is implemented as that will pause the autoupdate of RealmResults for the event loop (or until refresh is called) which means that if the change listener was triggered you could potentially access deleted objects when reacting to the change.

Triggering all change listeners from the Looper is also more consistent, but does introduce the slight weirdness that change listeners added after a commit will also be triggered. However, change listeners are inherently asynchronous so the change in this behaviour shouldn't cause regressions. The biggest impact is potentially on unit tests that depended on the exact time a listener was triggered.


cmelchior:
Ready for review @nhachicha @realm/java 
Note that most changes happened in our type based unit tests that depended on the knowledge that local listeners where triggered immediately.

changelog.txt
```diff
@@ -1,7 +1,8 @@
 0.88.0
  * BREAKING CHANGE: DynamicRealm.executeTransaction() now directly throws any RuntimeException instead of wrapping it in a RealmException (#1682).
  * BREAKING CHANGE: DynamicRealm.executeTransaction() now throws IllegalArgumentException instead of silently accepting a null Transaction object.
- * BREAKING CHANGE: String setters now throws IllegalArgumentException instead of RealmError for invalid surrogates. 
+ * BREAKING CHANGE: String setters now throws IllegalArgumentException instead of RealmError for invalid surrogates.
```
kneth: "throws" -> "throw"

cmelchior:
Reverting to WIP, some unit tests are failing spectacular on CI but not on my device.

cmelchior:
Ready for review @realm/java

realm/realm-library/src/main/java/io/realm/HandlerController.java
```diff
@@ -112,6 +112,7 @@ public boolean handleMessage(Message message) {
                 }
                 case COMPLETED_UPDATE_ASYNC_QUERIES: {
                     // this is called once the background thread completed the update of the async queries
+                    notifyAllListeners();
```
beeender: Is this correct? `completedAsyncQueriesUpdate` will emit `REALM_CHANGED` again, and `notifyAllListeners()` will be called anyway.

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -1093,7 +1093,7 @@ public void run() {
                     transaction.execute(bgRealm);
 
                     if (!Thread.currentThread().isInterrupted()) {
-                        bgRealm.commitTransaction(new Runnable() {
+                        bgRealm.commitTransaction(true, new Runnable() {
```
beeender: `true` here is actually redundant. Here, the realm is supposed to be closed before sending notifications, which means, even true is passed here, the local realm won't be notified because of it is supposed to be closed before.

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -18,6 +18,7 @@
 
 
 import android.os.Handler;
+import android.util.Log;
```
beeender: Log is not used.

realm/realm-library/src/main/java/io/realm/internal/async/RealmThreadPoolExecutor.java
```diff
@@ -81,6 +81,9 @@ protected void beforeExecute(Thread t, Runnable r) {
         }
     }
 
+    /**
+     * Pause the executor. Pausing means the executor will stop starting new tasks (but complete current ones).
```
beeender: `Pause` -> `Pauses`

realm/realm-library/src/androidTest/java/io/realm/NotificationsTest.java
```diff
@@ -551,7 +582,9 @@ public void onChange() {
         RealmChangeListener listenerB = new RealmChangeListener() {
             @Override
             public void onChange() {
-                counter.incrementAndGet();
+                assertEquals(0, counter.get());
```
beeender: Now we rely on the order of listener invocations. Worth to create another test case to ensure that ListenerA is called first.

realm/realm-library/src/androidTest/java/io/realm/TypeBasedNotificationsTests.java
```diff
@@ -501,95 +482,108 @@ public void run() {
         dog.addChangeListener(new RealmChangeListener() {
             @Override
             public void onChange() {
-                assertEquals("Akamaru", dog.getName());
+                // Step 4: Respond to relevant change
                 typebasedCommitInvocations.incrementAndGet();
+                assertEquals("Akamaru", dog.getName());
+                assertEquals(17, dog.getAge());
```
beeender: call `typebasedCommitInvocations.incrementAndGet();` after assertions, to avoid flaky test.

beeender: Uh, never mind, they are in the same thread.

realm/realm-library/src/androidTest/java/io/realm/NotificationsTest.java
cmelchior: Good idea. Added.

realm/realm-library/src/main/java/io/realm/Realm.java
cmelchior: Good catch

realm/realm-library/src/main/java/io/realm/HandlerController.java
cmelchior: I _think_ it is OK since REALM_CHANGED will only be triggered if the caller thread has advanced since it started its async query, but I hope @nhachicha will review this throughly.

cmelchior:
Feedback fixed @realm/java

beeender:
`commitTransaction` in `BaseRealm.migrateRealm()` doesn't seem to be necessary to trigger notifications. Maybe change that as well?

realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -201,7 +201,8 @@ public void removeAllChangeListeners() {
         handlerController.removeAllChangeListeners();
     }
 
-    void setHandler (Handler handler) {
+    // WARNING: If this method is used after calling a any async method, the old handler will still be used.
```
nhachicha: `s/calling a any/calling any`
also we can highlight that this is for test only

realm/realm-library/src/main/java/io/realm/HandlerController.java
```diff
@@ -331,23 +332,27 @@ private void realmChanged() {
...
+        }
+        if (!realm.isClosed()) {
+            notifyRealmObjectCallbacks();
+        }
+
+        // empty async RealmObject shouldn't block the realm to advance
+        // they're empty so no risk on running into a corrupt state
+        // where the pointer (Row) is using one version of a Realm, whereas the
+        // current Realm is advancing to a newer version (they're empty anyway)
+        if (!realm.isClosed() && threadContainsAsyncEmptyRealmObject()) {
```
nhachicha: I would keep this outside of this method, the method's responsibility is to only notify listeners

realm/realm-library/src/main/java/io/realm/HandlerController.java
nhachicha: I'm not sure about this either.
The current implementation of `completedAsyncQueriesUpdate` will notify the listeners if we advance the Realm.
```java
// notify listeners only when we advanced
            if (compare != 0) {
                notifyGlobalListeners();
                // notify RealmResults & RealmObject callbacks (type based notifications)
                notifySyncRealmResultsCallbacks();
                notifyRealmObjectCallbacks();
            }
```
maybe you can refactor it to 
```java
 if (compare != 0) {
    notifyAllListeners();
  }
```

nhachicha:
@cmelchior I refactored some of the code related to notifications to make it easy to reason about listeners notifications. I also fixed some failing tests.

The idea is to use `notifyAllListeners` when possible this will simply notify `notifyGlobalListeners` & `notifyTypeBasedListeners` this way it's easy to follow which action trigger which notification

realm/realm-library/src/androidTest/java/io/realm/NotificationsTest.java
```diff
@@ -567,11 +601,33 @@ public void onChange() {
...
+
+            @Override
+            public void onChange() {
+                firstListenerCalled.countDown();
+            }
+        });
+        realm.addChangeListener(new RealmChangeListener() {
+            @Override
+            public void onChange() {
+                TestHelper.awaitOrFail(firstListenerCalled);
```
beeender: They are in the same thread, no need to wait. just assert.

realm/realm-library/src/main/java/io/realm/HandlerController.java
```diff
@@ -469,13 +466,11 @@ private void completedAsyncQueriesUpdate(QueryUpdateTask.Result result) {
...
 
-            // notify listeners only when we advanced
-            if (compare != 0) {
-                notifyGlobalListeners();
-                // notify RealmResults & RealmObject callbacks (type based notifications)
-                notifySyncRealmResultsCallbacks();
-                notifyRealmObjectCallbacks();
-            }
+            // We need to notify the rest of listeners, since the original REALM_CHANGE
+            // was delayed/swallowed in order to be able to update async queries
```
beeender: @nhachicha Why did you remove the `if (compare != 0) {`? if `compare == 0`, `advanceRead` is not called, and no changes in this `SharedGroup` instance.

cmelchior:
Updated PR @realm/java

realm/realm-library/src/androidTest/java/io/realm/NotificationsTest.java
```diff
@@ -567,11 +601,32 @@ public void onChange() {
...
+        realm.addChangeListener(new RealmChangeListener() {
+            @Override
+            public void onChange() {
+                assertTrue(firstListenerCalled.get());
+                looperThread.testComplete();
+            }
+        });
+
+        realm.beginTransaction();
+        realm.commitTransaction();
```
beeender: oops, just realized this is not enough. need to test the other order as well in case the container changes to eg. `Set`.

cmelchior: Not sure what you mean?

beeender: It could be accidentally pass the test if we only test `addListener(listener1)` `addListener(listener2)`, to avoid that we have to also test the other way around with same `listener1` and `listener2` : `addListener(listener2)` `addListener(listener1)`

cmelchior:
Updated the unit test @beeender / @realm/java

beeender:
https://github.com/realm/realm-java/pull/2233/files#r53263681 Still have one pending comments? @nhachicha

realm/realm-library/src/main/java/io/realm/HandlerController.java
nhachicha: Like I mentioned in the comment I've added. This `updateAsyncEmptyRealmObject` is a consequence of a `REALM_CHANGE` we delayed/swallowed, so even if we don't advance the Realm, we still need to update sync RealmResults & RealmObject (because we originally delayed their notif  until this completes)

beeender: OK. But what about `notifyGlobalListeners()`, shouldn't it like:
```
if (compare != 0) {
    notifyGlobalListeners();
}
notifySyncRealmResultsCallbacks();
notifyRealmObjectCallbacks();
```

beeender:
I have a flaky test `callback_with_relevant_commit_from_different_looper_realmobject_async` with this change now :disappointed:

cmelchior:
It was already flaky before. I actually convinced myself that this PR fixed it, but guess not :(

beeender:
This still have some flaky tests... I am trying to fix now.

realm/realm-library/src/main/java/io/realm/HandlerController.java
beeender: :+1:  Just figured out this could happen when the transaction is committed in the same thread which has async queries.

beeender:
:+1:  and @realm/java please take a look at my flaky testing in this PR.

cmelchior:
Changes to unit tests looks good. Just one more :+1: @realm/java

dalinaum:
:+1:

cmelchior:
@dalinaum This was merged before being squashed :/

PR #2234, opened 2016-02-05
Update Kotlin used in example to 1.0-RC

zaki50:
We should migrate to realm gradle plugin. I'll do that after merging of better objects PR since that PR includes changes for Kotlin.

### DIFF
--- a/examples/kotlinExample/build.gradle
+++ b/examples/kotlinExample/build.gradle
@@ -1,5 +1,5 @@
 buildscript {
-    ext.kotlin_version = '1.0.0-beta-4589'
+    ext.kotlin_version = '1.0.0-rc-1036'
     repositories {
         jcenter()
     }
@@ -45,11 +45,16 @@ android {
     }
 }
 
+repositories {
+    // for ReLinker
+    maven { url 'https://jitpack.io/'}
+}
+
 dependencies {
     compile "org.jetbrains.kotlin:kotlin-stdlib:${kotlin_version}"
     compile "org.jetbrains.kotlin:kotlin-reflect:${kotlin_version}"
-    compile 'org.jetbrains.anko:anko-sdk15:0.7.2'
-    compile "io.realm:realm-android-library:${version}@aar"
+    compile 'org.jetbrains.anko:anko-sdk15:0.8.2'
+    compile "io.realm:realm-android-library:${version}"
     compile "io.realm:realm-annotations:${version}"
     kapt "io.realm:realm-annotations:${version}"
     kapt "io.realm:realm-annotations-processor:${version}"


examples/kotlinExample/build.gradle
```diff
@@ -45,11 +45,18 @@ android {
...
 dependencies {
     compile "org.jetbrains.kotlin:kotlin-stdlib:${kotlin_version}"
     compile "org.jetbrains.kotlin:kotlin-reflect:${kotlin_version}"
-    compile 'org.jetbrains.anko:anko-sdk15:0.7.2'
-    compile "io.realm:realm-android-library:${version}@aar"
+    compile 'org.jetbrains.anko:anko-sdk15:0.8.2'
+    compile ("io.realm:realm-android-library:${version}@aar") {
+        transitive=true
+    }
     compile "io.realm:realm-annotations:${version}"
```
cmelchior: If `transitive = true` cant we remove this?

examples/kotlinExample/build.gradle
```diff
@@ -45,11 +45,18 @@ android {
...
+}
+
 dependencies {
     compile "org.jetbrains.kotlin:kotlin-stdlib:${kotlin_version}"
     compile "org.jetbrains.kotlin:kotlin-reflect:${kotlin_version}"
-    compile 'org.jetbrains.anko:anko-sdk15:0.7.2'
-    compile "io.realm:realm-android-library:${version}@aar"
+    compile 'org.jetbrains.anko:anko-sdk15:0.8.2'
+    compile ("io.realm:realm-android-library:${version}@aar") {
+        transitive=true
```
cmelchior: space between ` = `

cmelchior:
Apart from comments :+1: 
But yes, we should move to the Gradle plugin as soon as we can, but I am not sure the Gradle plugin support Kotlin currently. That will only happen with Better Objects right?

examples/kotlinExample/build.gradle
zaki50: We can't remove this since we specify `aar` dependency instead of default `pom` dependency.

If remove this, we got `java.lang.NoClassDefFoundError: Failed resolution of: Lcom/getkeepsafe/relinker/ReLinker;`

examples/kotlinExample/build.gradle
zaki50: done

examples/kotlinExample/build.gradle
zaki50: Ah, yes we don't need to specify `aar`.
I'll update this.

zaki50: done

zaki50:
@cmelchior Maybe we can move to gradle plugin now, but we need to update the project again after merging Better Objects since Better Objects contains the change which adds `kapt` configuration.

cmelchior:
I guess it would be nice to have all example projects using the plugin

zaki50:
I tried to move to gradle plugin, and got an error `Please do not use `com.neenbedankt.android-apt` with kapt.`

Should we pick the change of plugin in Better Objects and merge it into master now?

cmelchior:
It isn't hugely important so I would just wait IMO

zaki50:
retest this please

cmelchior:
:+1:

dalinaum:
:+1:

PR #2236, opened 2016-02-05
corrects Javadoc comment of RealmResults#removeLast()

### LINKED ISSUE #2112: RealmResults.removeLast JavaDoc is wrong
`Removes and returns the last object in the list. This also deletes the object from the underlying Realm.`

It does not return any value, nor can it since the element has been deleted.

zaki50:
fixes #2112 

@realm/java


cmelchior:
:+1:

dalinaum:
:+1:

PR #2238, opened 2016-02-05
now RealmResults.size() honors the contract of the Collection interface

### LINKED ISSUE #2129: RealmResults.size() returns invalid result for large sizes()
Current implementation is this:

```
    @Override
    public int size() {
        if (!isLoaded()) {
            return 0;
        } else {
            return ((Long)getTable().size()).intValue();
        }
    }
```

It should be the below as the current implementation will overflow for large sizes, which does not honor the contract in the Collection interface : https://docs.oracle.com/javase/7/docs/api/java/util/Collection.html#size()

```
    @Override
    public int size() {
        if (!isLoaded()) {
            return 0;
        } else {
            long size = getTable().size();
            return (size > Integer.MAX_VALUE) ? Integer.MAX_VALUE : (int) size;
        }
    }
```

Will be fun writing a unit test for that :/

zaki50:
closes #2129 

@realm/java

### DIFF
--- a/realm/realm-library/build.gradle
+++ b/realm/realm-library/build.gradle
@@ -44,6 +44,8 @@ dependencies {
     androidTestCompile 'com.android.support:support-annotations:23.1.1'
     androidTestCompile 'com.android.support.test:runner:0.4.1'
     androidTestCompile 'com.android.support.test:rules:0.4.1'
+    androidTestCompile 'com.google.dexmaker:dexmaker:1.2'
+    androidTestCompile 'com.google.dexmaker:dexmaker-mockito:1.2'
 
     androidTestApt project(':realm-annotations-processor')
     androidTestApt project(':realm-annotations')
--- a/realm/realm-library/src/androidTest/java/io/realm/RealmResultsTests.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/RealmResultsTests.java
@@ -24,6 +24,7 @@ import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.ExpectedException;
 import org.junit.runner.RunWith;
+import org.mockito.Mockito;
 
 import java.util.Date;
 import java.util.List;
@@ -42,6 +43,7 @@ import io.realm.entities.Dog;
 import io.realm.entities.NonLatinFieldNames;
 import io.realm.entities.NullTypes;
 import io.realm.entities.Owner;
+import io.realm.internal.Table;
 import io.realm.rule.TestRealmConfigurationFactory;
 
 import static org.junit.Assert.assertEquals;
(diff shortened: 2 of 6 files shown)


realm/realm-library/src/androidTest/java/io/realm/TestHelper.java
```diff
@@ -695,4 +695,23 @@ public static void quitLooperOrFail() {
             Assert.fail();
         }
     }
+
+    public static <T extends RealmObject> RealmResults<T> newRealmResults(Realm realm, TableOrView table, Class<T> tableClass) {
```
cmelchior: Can you add some javadoc to why this is useful?

realm/realm-library/src/androidTest/java/io/realm/RealmResultsTests.java
```diff
@@ -996,6 +998,19 @@ public void where_findAll_size() {
     }
 
     @Test
+    public void size_returns_Integer_MAX_VALUE_for_huge_results() {
```
cmelchior: This test is highly dependant on the implementation of RealmResults, but there is probably no way around that.

cmelchior:
Apart from missing Javadoc :+1:

realm/realm-library/src/androidTest/java/io/realm/TestHelper.java
zaki50: @cmelchior added. Could you review the javadoc comment?

realm/realm-library/build.gradle
```diff
@@ -44,6 +44,8 @@ dependencies {
     androidTestCompile 'com.android.support:support-annotations:23.1.1'
     androidTestCompile 'com.android.support.test:runner:0.4.1'
     androidTestCompile 'com.android.support.test:rules:0.4.1'
+    androidTestCompile 'com.google.dexmaker:dexmaker:1.2'
+    androidTestCompile 'com.google.dexmaker:dexmaker-mockito:1.2'
```
nhachicha: any chance this will interfere with the upcoming javassist bytecde manipulation?

nhachicha:
:+1:  we need to document the limitation about how many elements a `RealmResults` should contains, since we can't do any useful operations beyond `Integer.MAX_VALUE` limit.
Ex: even if your table_view contains more element than `Integer.MAX_VALUE` you can't `get(int location)`, let alone `size()`

kneth:
I am a bit worried. Consider the follow case: I add `2 * Interger.MAX_VALUE` objects with an long field like:

    for(long i = 1; i <= 2*Integer.MAX_VALUE; i++) {
        MyObject o = realm.createObject(MyObject.class);
        o.field = i;
    }

Now I do `RealmResults<MyObject> allObjects = realm.allObjects(MyObject.class)`. Of course `allObjects.average("field")` will return the correct result. But `allObjects.sum("field")/allObjects.size()` will not. Moreover, the follow will also produce a different result:

    double sum = 0.0;
    for(int i = 0; i < allObjects.size(); i++) {
        sum += allObjects.get(i).getField();
    }

For most practical cases, it is more important to follow the Collection interface, We could resolve this by providing an extra method `proper_size()` which returns a `long`.

cmelchior:
IMO we are purely in the hypothetical Realm (pun intended) at this stage. The behaviour for large int's is documented in the collection API, and I would avoid polluting our public API with methods that nobody is going to use, because by all accounts they probably ran out of disk space first.

kneth:
Oh yes, for most (all) practical cases, we will never run into this issue. But as @nhachicha says, it is wise to document the limitation.

zaki50:
How about adding following to class comment of `RealmResults` and `RealmList`

```
 * <p>
 * {@link RealmResults} can contains more elements than {@code Integer.MAX_VALUE}.
 * In that case, you only can access first {@code Integer.MAX_VALUE} elements in it.
```

realm/realm-library/build.gradle
zaki50: I don't think so.
What do you think? @emanuelez

cmelchior:
Sounds fine to me

zaki50:
```
you only can access first {@code Integer.MAX_VALUE} elements
```

should be

```
you can access only first {@code Integer.MAX_VALUE} elements
```
?

zaki50:
please review added comments

realm/realm-library/build.gradle
cmelchior: AFAIK mockito just overrides the ClassLoader and does the bytecode manipulation at runtime.

emanuelez: What Christian said. Well... not really manipulation, that's not possible to do on Android at runtime, but it can create new classes.

dalinaum:
:+1:

cmelchior:
:+1:

PR #2242, opened 2016-02-06
make a test of RealmObject.hashCode() path independent.

zaki50:
@realm/java

### DIFF
--- a/realm/realm-library/src/androidTest/java/io/realm/RealmObjectTests.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/RealmObjectTests.java
@@ -16,7 +16,6 @@
 
 package io.realm;
 
-import android.support.test.InstrumentationRegistry;
 import android.support.test.runner.AndroidJUnit4;
 
 import org.junit.After;
@@ -26,6 +25,7 @@ import org.junit.Test;
 import org.junit.rules.ExpectedException;
 import org.junit.runner.RunWith;
 
+import java.io.File;
 import java.util.Calendar;
 import java.util.Date;
 import java.util.HashSet;
@@ -51,6 +51,7 @@ import io.realm.rule.TestRealmConfigurationFactory;
 import static io.realm.internal.test.ExtraTests.assertArrayEquals;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotEquals;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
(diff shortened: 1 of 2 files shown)


zaki50:
hmm
```
org.junit.ComparisonFailure: expected:</data/[data]/io.realm.test/files...> but was:</data/[user/11]/io.realm.test/files...>
at org.junit.Assert.assertEquals(Assert.java:115)
at org.junit.Assert.assertEquals(Assert.java:144)
```

cmelchior:
That is expected. You are not guaranteed that path to the sdcard, so I only think this PR will add more sporadic failures. We have one unit test that depends on the path (hashCode - something), and you see that fail sometimes as well.

zaki50:
@cmelchior How about updating this test to be path independent?

The point of this test is not an actual value, but the fact that the value is the same between same object and the fact that the value is different with different object.

zaki50:
@cmelchior I updated the test. please review again.

realm/realm-library/src/androidTest/java/io/realm/RealmObjectTests.java
```diff
@@ -411,18 +412,66 @@ public void toString_cyclicObject() {
...
+            @Override
+            public void execute(Realm realm) {
+                foo.setName(foo.getName() + "1234");
+            }
+        });
+        // Check that Updating the value of its field does not affect the hash code.
+        assertEquals(originalHashCode, foo.hashCode());
+
+        // Check the hash code of the object from a Realm in different file name.
+        RealmConfiguration realmConfig_differentName = new RealmConfiguration.Builder(realmConfig.getRealmFolder())
```
cmelchior: Use TestRealmConfigurationFactory instead?

realm/realm-library/src/androidTest/java/io/realm/RealmObjectTests.java
```diff
@@ -411,18 +412,66 @@ public void toString_cyclicObject() {
...
+                .name("another_" + realmConfig.getRealmFileName())
+                .build();
+        Realm.deleteRealm(realmConfig_differentName);
```
cmelchior: This shouldn't bee needed if using TestRealmConfigurationFactory

realm/realm-library/src/androidTest/java/io/realm/RealmObjectTests.java
```diff
@@ -411,18 +412,66 @@ public void toString_cyclicObject() {
...
+        }
+
+        // Check the hash code of the object from a Realm in different folder.
+        final File anotherDir = new File(realmConfig.getRealmFolder(), "anotherDir");
+        //noinspection ResultOfMethodCallIgnored
+        anotherDir.mkdirs();
+        RealmConfiguration realmConfig_differentPath = new RealmConfiguration.Builder(anotherDir)
+                .name(realmConfig.getRealmFileName())
+                .build();
+        Realm.deleteRealm(realmConfig_differentPath);
```
cmelchior: Same

realm/realm-library/src/androidTest/java/io/realm/RealmObjectTests.java
```diff
@@ -411,18 +412,66 @@ public void toString_cyclicObject() {
...
+            realm_differentPath.commitTransaction();
+
+            assertNotEquals(fooFromDifferentPath.hashCode(), foo.hashCode());
+        } finally {
+            realm_differentPath.close();
+        }
     }
 
-    private CyclicType createCyclicData() {
+    private static CyclicType createCyclicData(Realm realm) {
```
cmelchior: Any reason it is static?

realm/realm-library/src/androidTest/java/io/realm/RealmObjectTests.java
zaki50: ah, right

realm/realm-library/src/androidTest/java/io/realm/RealmObjectTests.java
zaki50: Because this method does not depend on fields in this class, but it's ok to make it non-static.

realm/realm-library/src/androidTest/java/io/realm/RealmObjectTests.java
zaki50: removed

realm/realm-library/src/androidTest/java/io/realm/RealmObjectTests.java
zaki50: removed

realm/realm-library/src/androidTest/java/io/realm/rule/TestRealmConfigurationFactory.java
```diff
@@ -94,6 +94,18 @@ public RealmConfiguration createConfiguration() {
         return configuration;
     }
 
+    public RealmConfiguration createConfiguration(String subDir, String name) {
+        final File folder = new File(getRoot(), subDir);
+        //noinspection ResultOfMethodCallIgnored
+        folder.mkdirs();
```
cmelchior: Fore the exceptional case where it might fail, perhaps `assertTrue(folder.mkDirs())`. You could also remove the noinspection then

cmelchior:
Apart from small comment :+1:

kneth:
:+1:

PR #2260, opened 2016-02-11
Chained distinct/distintAsync calls does not yield expected results

stk1m1:
The idea is to interweave different values in 2's multiplier and 3's multiplier in a way that
the outcome is different if the order of distinct\* operations alternates.

An example is illustrated below.

```
Object        : O1| O2| O3| O4| O5| O6
indexedString : A | A | B | B | A | A
indexedLong   : 1 | 1 | 1 | 2 | 2 | 2
```

Suppose we are to execute `realm.where(Foo.class).findAll().distinct("indexedString").distinct("indexedLong")`, the outcome might be O1(A, 1) and the total count should be one (1) since distinct takes the first element when two elements are identical.

In case of `realm.where(Foo.class).findAll().distinct("indexedLong").distinct("indexedString")`, the outcome would be O1(A,1) and O2(B,2) and the total count should two (2). Similar outcome is expected to chained `distinctAsync()`.


realm/realm-library/src/androidTest/java/io/realm/RealmResultsTests.java
```diff
@@ -1657,6 +1657,91 @@ public void distinctAsync_notIndexedLinkedFields() {
...
+    }
+
+    @Test
+    public void distinct_checkChainedOpertionValidity() {
+        final long numberOfBlocks = 1; // must be greater than 1
+        populateForChainedOperation(realm, numberOfBlocks);
+
+        final RealmResults<AnnotationIndexTypes> distinctStringAndLong = realm.where(AnnotationIndexTypes.class).findAll().distinct("indexString").distinct("indexLong");
+        final RealmResults<AnnotationIndexTypes> distinctLongAndString = realm.where(AnnotationIndexTypes.class).findAll().distinct("indexLong").distinct("indexString");
+        assertEquals("distinctStringAndLong", 1, distinctStringAndLong.size());
```
stk1m1: In terms of API representation, these two lines flow as
`Table.where().findAll()` ->  `TableView.distinct()` ->  `TableView.distinct()` (two `TableView` instances are the same).

realm/realm-library/src/androidTest/java/io/realm/RealmResultsTests.java
```diff
@@ -1657,6 +1657,91 @@ public void distinctAsync_notIndexedLinkedFields() {
...
+    @RunTestInLooperThread
+    public void distinctAsync_checkChainedOpertionValidity() throws Throwable {
+        final AtomicInteger changeListenerCalled = new AtomicInteger(2);
+        final Realm realm = looperThread.realm;
+        final long numberOfBlocks = 1; // must be greater than 1
+        populateForChainedOperation(realm, numberOfBlocks);
+
+        final RealmResults<AnnotationIndexTypes> distinctStringAndLong = realm.where(AnnotationIndexTypes.class).findAll().distinctAsync("indexString").distinctAsync("indexLong");
+        final RealmResults<AnnotationIndexTypes> distinctLongAndString = realm.where(AnnotationIndexTypes.class).findAll().distinctAsync("indexLong").distinctAsync("indexString");
+
```
stk1m1: In terms of API representation, two two lines work as 
`Table.where().findAll()` -> `TableView.where().getTable().getDistinctView()` -> `TableView.where().getTable().getDistinctView()` (The last two `TableView` are different ones).

cmelchior:
I think this is the same as https://github.com/realm/realm-java/issues/2263 and the solution should also be the same.

stk1m1:
A related issue #2282 has created.

stk1m1:
Tested with Core commit 581d546 & Java commit 927a39b.

cmelchior:
Outdated. Closing

PR #2279, opened 2016-02-15
Adapt core fix for the delete RealmList

beeender:
Update core to 0.96.1.

No more exception will be thrown when access RealmResults if the base
RealmList has be deleted. Instead, it will be treated as an empty
RealmResults forever.


dalinaum:
@beeender Is this PR closed without creating new PR?

PR #2294, opened 2016-02-17
fixes #2115 IllegalStateException Caller thread behind the worker thread

### LINKED ISSUE #2115: [0.88.0-SNAPSHOT] IllegalStateException: Caller thread behind the worker thread
http://crashes.to/s/d9d24a52572
This exception is thrown after I migrated my codebase from synchronous to asynchronous realm calls using observables. The stacktrace doesn't show where it went wrong from my code.
```
messagesSubscription = realm.where(Message.class)
        .equalTo("conversationId", conversationId)
        .findAllSortedAsync("updated")
        .asObservable()
        .filter(RealmResults::isLoaded)
        .subscribe((messages) -> {
          if (adapter != null) {
            adapter.call(messages);
          }
        });
```
The above is the only asynchronous call in that problematic activity. Any ideas why the exception is thrown?

nhachicha:
Fixes #2115

### DIFF
--- a/realm/realm-jni/src/io_realm_internal_TableQuery.cpp
+++ b/realm/realm-jni/src/io_realm_internal_TableQuery.cpp
@@ -1037,7 +1037,7 @@ JNIEXPORT jlong JNICALL Java_io_realm_internal_TableQuery_nativeFind(
     return -1;
 }
 
-std::unique_ptr<Query> getHandoverQuery (jlong bgSharedGroupPtr, jlong replicationPtr, jlong queryPtr)
+std::unique_ptr<Query> getHandoverQuery (jlong bgSharedGroupPtr, jlong replicationPtr, jlong queryPtr, bool advanceToLatestVersion)
 {
     SharedGroup::Handover<Query> *handoverQueryPtr = HO(Query, queryPtr);
     std::unique_ptr<SharedGroup::Handover<Query>> handoverQuery(handoverQueryPtr);
@@ -1053,7 +1053,7 @@ std::unique_ptr<Query> getHandoverQuery (jlong bgSharedGroupPtr, jlong replicati
     }
 
     std::unique_ptr<Query> query = SG(bgSharedGroupPtr)->import_from_handover(std::move(handoverQuery));
-    if (isDifferentVersions) {
+    if (advanceToLatestVersion && isDifferentVersions) {
         LangBindHelper::advance_read(*SG(bgSharedGroupPtr), *CH(replicationPtr));
     }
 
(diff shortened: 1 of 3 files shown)


nhachicha:
@realm/java

realm/realm-jni/src/io_realm_internal_TableQuery.cpp
```diff
@@ -1066,7 +1066,7 @@ JNIEXPORT jlong JNICALL Java_io_realm_internal_TableQuery_nativeFindWithHandover
 {
     TR_ENTER()
     try {
-        std::unique_ptr<Query> query = getHandoverQuery(bgSharedGroupPtr, replicationPtr, queryPtr);
+        std::unique_ptr<Query> query = getHandoverQuery(bgSharedGroupPtr, replicationPtr, queryPtr, false);
```
cmelchior: Remind me again: What is the reason that findFirst have to behave differently than findAll(). This is also evident in HandlerController where the flow is hard to follow because AsyncRealmObjects behave differently than AsyncRealmResults?

stk1m1: I'm just curious to know if this indicates that it is safe to return an older version of object with `RealmQuery.findFirstAsync()`.

stk1m1:
:+1:

cmelchior:
I am not entirely sure this is a correct way of fixing it, but maybe I am just missing something?

To me it looks like a problem in us handling asyncRealmResults and asyncRealmObjects differently, e.g `realmChanged()` in HandlerController only check for asyncRealmResults. Why doesn't it check for asyncRealmObjects as well.

nhachicha:
The main difference between Async `RealmObject` & Async `RealmResults` is the update/rerun.

Async `RealmObject` is just queried, once you've got a valid pointer we don't rerun the query (the overhead is bigger than calling _sync_if_needed_ when advancing a Realm to a newer version).

On the other hand `nativeFindAll*WithHandover` is used for both querying & updating (rerun) the `table_view` (once we receive a `REALM_CHANGE`). This is why after you import the handover of the query, you need to run the data against the latest Realm (hence the boolean to control this behaviour).

In the case of a `findFirstAsync` you only have 2 UC.
1- Your caller Realm didn't change so the version you're importing is good
2- The caller Realm is ahead of the worker thread, in this case we need to rerun the query against the newest Realm. By running the query against the same version of the handover query (`advanceToLatestVersion=false`) you have the guarantee upon the previous UC.

cmelchior:
Okay it is starting to make sense, since `findFirst` returns a reference to a object and not a wrapper around `results.get(0)`. 

I wonder if that is the behavior that people except though, but it is a discussion we can take outside this PR.

Looks :+1: to me.

beeender:
I still don't understand how could this change fix the issue. How could the caller's version be greater than workers? We do intercept all the `REALM_CHANGE` already, the `advance_read` should not be called before this.

cmelchior:
People can do a local commit on the thread starting the async query. We cannot really prevent that (well, we can, but don't currently).

beeender:
Right. :+1:

dalinaum:
:+1:

nhachicha:
@realm/java

nhachicha:
squashed

cmelchior:
:shipit:

PR #2295, opened 2016-02-17
Convert test case to JUNIT4

beeender:
RealmAnnotationTest and RealmConfigurationTest.


beeender:
retest this please

cmelchior:
IMO we should just disable the failing unit test for now and await https://github.com/realm/realm-core/pull/1475

cmelchior:
It looks like we are missing at least 1 test that actually uses a proper context now and not `getRoot()`, otherwise :+1:

stk1m1:
:+1:

PR #2196, opened 2016-01-29
Allow custom methods and public fields

emanuelez:
This PR introduces the support for custom methods and public fields. This is achieved with bytecode manipulation which will perform two tasks:

 * For every field add accessors named `<fieldName>$realmGetter()` and `<fieldName>$realmSetter()`
 * Substitute all field accesses (apart from those in the new accessors) with the invocation of the new accessors

Since the bytecode manipulation step need to happen after the annotation processor execution, the new accessors have been added to the model classes of the unit tests of `realm/realm-library`.

Also, the annotation processor has been modified to generate accessors with the new name, instead of having to guess what the user did.

Fixes #909
Fixes #502

Issue #1036, opened 2015-04-15
Remove RealmObject class? Decide.


cmelchior:
Having to extend RealmObject seem to irk some people (me among them) as annotations are seen as being more natural for this kind of thing. Due to support for RealmModules incoming ( #938 ) and that we only have 2 methods on RealmObject, the only real use a more strict API has is that it catches more errors on compile time. However due to RealmModules allowing custom Schemas pr. Realm, this type check will only be partial anyway.

There is no technical limitation that prevents us from removing this, except that a lot of API methods will now require Object instead of RealmObject. This is not ideal and is more error prone, but the upside is that Entity classes are now more clean.

This would change the current API to this:
```
@RealmClass
public class Foo {
  ....
}

Realm realm = Realm.getInstance(getContext());
Foo foo = realm.createObject(Foo.class):
realm.isValid(foo);
realm.removeFromRealm(foo);
```

Note this still doesn't allow subclassing which is a different issue : #761

cmelchior:
Copying this comment from #440 (as it really belongs here):

One possible solution for achieving the best of both worlds could perhaps be to to develop a custom Realm IntelliJ plugin that did static analysis. This would obviously only work inside IntelliJ, but Google are doing the same with their support annotations: http://tools.android.com/tech-docs/support-annotations and they even managed to get it to work with the Gradle plugin.

Something like this:

```
// 1. We remove the RealmObject class
// 2. annotate all methods that require an RealmObject like so
public <E> E createObject(@RealmClass Class<E> class) {
 // ...
}

// 3. When calling this method we do static analysis and let IntelliJ mark it as an error
realm.createObject(MyRealmObject.class); // This would work
realm.createObject(NonRealmObject.class); // This would be marked as an error
```

cmelchior:
Some further thoughts on this. We will probably soon introduce a lot of features where having a base class will make the API much nicer:

* KVO 
* ChangeListeners
* Freeze

With a base class it could look like this:
```
Foo foo = realm.where(Foo.class).findFirst();

// KVO
foo.addFieldChangeListener("bar", new RealmChangeListener() {
  public void onChange() {
    // Field bar changed
  }
}

// ChangeListener
foo.addChangeListener(new RealmChangeListener() {
  public void onChange() {
    // Object was updated
  }
}

// RxJava
Observable<Foo> fooObservable = foo.observe();

// Immutability
Foo frozenFoo = foo.freeze(); // Creates an immutable copy
```

[...]

PR #2296, opened 2016-02-17
[Discussion] RealmModel interface as a supplement to RealmObject

cmelchior:
This is an attempt at figuring out a solution for #1036 

Since the release of of Realm there as been objections against our requirements of forcing the `RealmObject` base class. See #1036 arguments on both sides. 

After discussion both internally and externally, it seems the biggest dealbreaker was the base class, implementing Interfaces was seen as OK. Having just the @RealmClass annotation  was also OK.

## Public API change
This proposal adds a secondary way of specifying RealmObjects, namely that people can just implement a `RealmModel` interface. This removes the primary concern from our side, which is type safety while still allowing classes to be more decoupled from Realm. 

```
// New marker interface that RealmObject extends
@RealmClass
public interface RealmModel {
}

// Classes can just implement the interface
public class Foo implements RealmModel {
   // ...
}
RealmObject.isValid(foo);

// Or continue to use the base class.
   // ...
}
foo.isValid();
```

This will be a none-breaking change, but will give people more freedom + remove problems with wanting methods like `isValid()` and `isLoaded()` in your own model class.

## Review
- Most important: Is this approach a good solution?
- The project should compile, but doesn't run. Interesting classes is `RealmModel`, `RealmObject`, `DynamicRealmObject` and `RealmObjectProxy`.
- All public methods on `RealmObject` will also have a static variant on RealmObject. Location of these could be debated.
- All package protected methods in `RealmObject` is moved to the `RealmObjectProxy` interface`. 
- How should the RealmModel class be named?
       
## Consequences:
- DynamicRealmObject becomes a bit of a bastard object since we need to copy implementation between our proxy classes and the DynamicRealmObject. A possible solution could be to contain state in a `RealmObjectProxyState` class and use composition in proxies and DynamicRealmObject.
- Proxy now all suddenly contain a lot more code, ie. all current package protected classes needs to be re-implemented in the annotation processor.
- We get a nicer seperation of concerns in the internal API. It is now a lot more visible which objects are actually Proxy objects and which are just "standalone" objects.

## RealmModel naming
- Keeping RealmObject/RealmClass is nice to avoid breaking to many things
- Other interface names could be `RealmEntity` or `RealmType`. Other suggestions are welcome.

## Other
This PR is not to be merged without further work but will  stay in "Review" until a solution is agreeded upon. Then it will either be reworked or closed in favour of another. Note #2196 is a requirement before any real work is done as it significantly changes how the proxies work as well.


cmelchior:
Ready for review @realm/java

realm/realm-library/src/main/java/io/realm/DynamicRealmObject.java
```diff
@@ -29,16 +33,27 @@
  * Class that wraps a normal RealmObject in order to allow dynamic access instead of a typed interface.
  * Using a DynamicRealmObject is slower than using the regular RealmObject class.
  */
-public final class DynamicRealmObject extends RealmObject {
+public final class DynamicRealmObject extends RealmObject implements RealmObjectProxy {
     private String className;
 
+    // TODO This fields is duplicated between DynamicRealmObject and Proxy classes....eeew
+    // Brainstorm idea: Composition over inheritance. Perhaps encapsulate in a RealmObjectProxyState class ?
```
donnfelker: Agree here. Composition > Inheritance. I'd propose the name `ProxyState` because the alternative just seems too long.

realm/realm-library/src/main/java/io/realm/RealmModel.java
```diff
@@ -0,0 +1,28 @@
...
+ * Interface for marking classes as RealmObjects, it can be used instead of extending {@link RealmObject}.
+ *
+ * All helper methods available to classes that extend RealmObject are instead available as static methods:
+ *
+ * <pre>
+ * {@code
+ *   Person p = realm.createObject(Person.class);
+ *
+ *   // With the RealmModel interface
+ *   RealmObject.isValid(p);
```
donnfelker: Not sure how I feel about this yet. I understand that this is the penalty for not extending a base class. Still works though ... the developer using the API needs to determine the tradeoffs.

realm/realm-library/src/main/java/io/realm/RealmObject.java
```diff
@@ -75,41 +70,58 @@
...
+public abstract class RealmObject implements RealmModel {
 
-    protected Row row;
-    protected BaseRealm realm;
 
-    private final List<RealmChangeListener> listeners = new CopyOnWriteArrayList<RealmChangeListener>();
-    private Future<Long> pendingQuery;
-    private boolean isCompleted = false;
-    protected long currentTableVersion = -1;
+// TODO Move this to the proxy classes and expose methods in the RealmObjectProxy interface
```
donnfelker: Is this still a TODO or an old comment? I saw the fields in the `RealmObjectProxy` below.

realm/realm-library/src/main/java/io/realm/RealmObject.java
```diff
@@ -75,41 +70,58 @@
...
+     * @param object RealmObject to remove from the underlying Realm.
+     * @throws IllegalArgumentException if the object is not managed by Realm.
+     * @throws IllegalStateException if the Realm is closed or accessed from the wrong thread.
      */
-    public void removeFromRealm() {
-        if (row == null) {
-            throw new IllegalStateException("Object malformed: missing object in Realm. Make sure to instantiate RealmObjects with Realm.createObject()");
+    public static <E extends RealmModel> void removeFromRealm(E object) {
+        if (!(object instanceof RealmObjectProxy)) {
+            // TODO What type of exception IllegalArgument/IllegalState?
```
donnfelker: What about `IllegalOperationException`?

donnfelker:
Whoa changes galore! :smile: 

That said, my gut says this feels like this is a move in the right direction. A lot of folk's do not like inheriting from a concrete object, regardless of what it gives you. Going to stew on this a bit and then provide feedback on the review items above in the AM.

AKiniyalocts:
:+1: This looks great guys, thanks for all of your hard work!

version.txt
```diff
@@ -1 +1 @@
-0.88.0-SNAPSHOT
+0.90.0-SNAPSHOT
```
beeender: Very accurate prediction

emanuelez: :crystal_ball:

realm/realm-library/src/main/java/io/realm/RealmObject.java
beeender: IllegalArgumentException +1

beeender:
My only concern is the naming, then we have `RealmObject`, `RealmModel` and `@RealmClasss` which are quite confusing for beginners. Is it possible to merge `RealmModel` and `@RealmClasss` into one and give it a better name?

realm/realm-library/src/main/java/io/realm/RealmModel.java
cmelchior: No way around this, as that is the trade-off you get with no baseclass. We can always debate where to put the methods though. I see 4 options:

1) Static methods on Realm
2) instance methods on Realm
3) Static methods on RealmObject
4) Somewhere else?

I went with 3) as it avoids polluting the Realm class with to many unrelated methods + it emphasiss the context more to use RealmObject IMO. Happy to hear arguments for other locations though.

realm/realm-library/src/main/java/io/realm/RealmObject.java
cmelchior: It is still a TODO. The methods are there, but they are no-ops currently.

cmelchior:
Yes, that is my worry as well. IMO we should still use `RealmObject` as the _default_ mode of operation, and then just have a small section about how to use the interface instead.

There is no way around having 3 different files: A base class, an interface and an annotation. How they are named is very much up for discussion, e.g. you could have both an interface and an annotation named `RealmClass` but then you need to to fully qualify the annotation when used (which is only happening in our interface + if the compiler doesn't support inherited annotations (:wave: Eclipse, which we deprecate anyway)).

Alternatively we move the `RealmClass` annotation out of the public API? People should never use it directly anyway.

realm/realm-library/src/main/java/io/realm/RealmObject.java
stk1m1: :+1:

cmelchior:
The more I think about it, the more I like the idea that we move `io.realm.annotation.RealmClass` out of the public API. We only had two reasons to have it in the public API: Eclipse and Kotlin. Eclipse is about to be deprecated and Kotlin now supports inherited annotations.

This leaves us with only two names: `RealmObject` and `Realm<Model/Class/Type/Entity/Something>` for the interface. We can then decide for a free name for the annotation afterwards (just to avoid people picking the wrong import from the autocomplete menu)

nhachicha:
Can the user define the following model?
```java
abstract class  Animal implements RealmModel {
    abstract boolean isMammal();
}

class Dog extends Animal {

    @Override
    boolean isMammal() {
        return true;
    }
}

class Canary extends Animal {
    @Override
    boolean isMammal() {
        return false;
    }
}
```
- Are we going to prevent inheritance/abstract class at the  annotation processor level?
- What if `Animal` was not `abstract` does the schema include an `Animal` table now? will this impact migration?

cmelchior:
No, with this proposal we still wouldn't support inheritance from abstract classes, since we still don't have the query support we need.

cmelchior:
Since we seem to agree on the general approach, I am closing this PR now. Further work is being tracked in #2347

Issue #2303, opened 2016-02-19
`./gradlew test` is not running on CI, and mockito cannot mock final Realm class


beeender:
Two issues:

1) `./gradlew test` will fail `unitTestExample` because of
```
when(Realm.getInstance(any(RealmConfiguration.class))).thenReturn(mockRealm);
```
   will fail. Realm is an static class which cannot be mocked by Mockito. We might have to use PowerMock instead

2) Enable `./gradlew test` on CI.

cmelchior:
We should also make all examples work when running `gw check`. This also include fixing lint warnings or ignore them

beeender:
@emanuelez #2320 has been merged. Can you please help to enable the `./gradlew test` for `examples/unitTestExample/src/test` on CI and close this issue? Thanks!

emanuelez:
This is now done both for PRs and normal builds

PR #2304, opened 2016-02-19
Deprecate Realm.getInstance(Context)

beeender:
* Deprecate Realm.getInstance(Context).
 * Replease `Realm.getInstance(Context)` in examples.

 See #2303 , `UnitTestExample` doesn't work even before.
 Realm doc needs to be updated as well.

### DIFF
--- a/examples/introExample/src/main/java/io/realm/examples/intro/IntroExampleActivity.java
+++ b/examples/introExample/src/main/java/io/realm/examples/intro/IntroExampleActivity.java
@@ -182,22 +182,11 @@ public class IntroExampleActivity extends Activity {
                 dogName = pers.getDog().getName();
             }
             status += "\n" + pers.getName() + ":" + pers.getAge() + " : " + dogName + " : " + pers.getCats().size();
-
-            // The field tempReference is annotated with @Ignore
-            // Though we initially set its value to 42, it has
-            // not been saved as part of the Person RealmObject:
-            if (BuildConfig.DEBUG && pers.getTempReference() != 0) {
-                throw new AssertionError();
-            }
         }
 
         // Sorting
         RealmResults<Person> sortedPersons = realm.allObjects(Person.class);
         sortedPersons.sort("age", Sort.DESCENDING);
-        if (BuildConfig.DEBUG &&
-                !realm.allObjects(Person.class).last().getName().equals(sortedPersons.first().getName())) {
-            throw new AssertionError();
-        }
         status += "\nSorting " + sortedPersons.last().getName() + " == "
                 + realm.allObjects(Person.class).first().getName();
 
--- a/examples/introExample/src/main/java/io/realm/examples/intro/IntroExampleActivity.java
+++ b/examples/introExample/src/main/java/io/realm/examples/intro/IntroExampleActivity.java
@@ -24,6 +24,7 @@ import android.widget.LinearLayout;
 import android.widget.TextView;
 
 import io.realm.Realm;
+import io.realm.RealmConfiguration;
 import io.realm.RealmResults;
 import io.realm.Sort;
 import io.realm.examples.intro.model.Cat;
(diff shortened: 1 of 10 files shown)


examples/introExample/src/main/java/io/realm/examples/intro/IntroExampleActivity.java
```diff
@@ -182,14 +186,20 @@ private String complexReadWrite() {
...
+                throw new AssertionError();
+            }
         }
 
         // Sorting
         RealmResults<Person> sortedPersons = realm.allObjects(Person.class);
         sortedPersons.sort("age", Sort.DESCENDING);
-        assert(realm.allObjects(Person.class).last().getName() == sortedPersons.first().getName());
-        status += "\nSorting " + sortedPersons.last().getName() + " == " + realm.allObjects(Person.class).first().getName();
+        if (BuildConfig.DEBUG &&
```
cmelchior: I would probably remove these asserts all together from an intro example. They don't provide any value for how Realm works IMO?

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -79,6 +79,19 @@
  * A standard pattern for working with Realm in Android activities can be seen below:
  * <p>
  * <pre>
+ * public class RealmApplication extends Application {
+ *
+ *     private static RealmConfiguration defaultRealmConfig;
```
cmelchior: Any reason to keep it in a static variable?

examples/introExample/src/main/java/io/realm/examples/intro/IntroExampleActivity.java
beeender: I agree, will do.

realm/realm-library/src/main/java/io/realm/exceptions/RealmMigrationNeededException.java
```diff
@@ -39,10 +39,10 @@ public RealmMigrationNeededException(String canonicalRealmPath, String detailMes
      * Returns the canonical path to the Realm file that needs to be migrated.
      *
      * This can be used for easy reference during a migration:
-     *
+     * FIXME: This example doesn't make too much sense anymore.
```
cmelchior: Just delete it then?

realm/realm-library/src/main/java/io/realm/Realm.java
beeender: uh... i used to have 
```
if (defaultRealmConfig == null) {
    defaultRealmConfig = new RealmConfiguration.Builder(this).build();
}
```
in `onCreate`. But then i think it might confuse user, so it is removed. Forgot to change the static.

What do you think? Does init only once would create some value for beginners?

cmelchior:
Done with round 1. Good idea ... people are constantly misunderstanding `Realm.getInstance(context)`.

realm/realm-library/src/main/java/io/realm/Realm.java
cmelchior: The Application class is only created when starting the app. The only case you need to reference the configuration (instead of using `getDefaultInstance`), is if you want to manually delete or migrate the Realm, but for the Javadoc I think we can go with the minimal setup. I would remove the static var.

beeender: @cmelchior Do you think it would be even better we change the `defaultRealmConfig` to a local var?

cmelchior: Not really. It just complicates the example IMO:

```
public void onCreate() {
  RealmConfiguration realmConfig = new RealmConfiguration.Builder(this).build();
  Realm.setDefaultConfiguration(realmConfig);
}  
```

Would be enough IMO (at least for a basic intro, which I think this should be).

beeender: Yeah, that is what i mean. a local `realmConfig`. Will do.

cmelchior:
:+1:

beeender:
https://github.com/realm/realm.io/pull/1813 for the doc part update.

stk1m1:
:+1:

Issue #2227, opened 2016-02-03
Library shouldn't include android:allowBackup="true"


intrications:
I don't think this should be set in the library. Is there a reason it is there?

Including android:allowBackup="true" means it is a hassle to override this to false in my app.

https://github.com/realm/realm-java/blob/master/realm/realm-library/src/main/AndroidManifest.xml#L4

cmelchior:
Hi @intrications 
I believe the idea is that as Realm create database files these should be backed up by default. As far as I know `allowBackup` default is `true` (http://developer.android.com/guide/topics/manifest/application-element.html) but Android studio some time ago enabled Lint warnings that complained if it wasn't explicitly set as it is a potentially security issue. 

I was under the impression that this setting could be overridden in the apps manifest, but there seem to be some outstanding issues: https://code.google.com/p/android/issues/detail?id=70073 which I agree is highly unfortunate.

We will have to look into any potential negative interaction with Lint, but otherwise I am not opposed to removing it.

mikescamell:
This is causing me some problems also as having data backup makes no sense for the application i am developing. The data i store becomes irrelevant quickly. 

Without this i cannot update to 0.87.4 and see if some issues i'm having have been fixed regarding the clearing of data.

cmelchior:
@intrications @mikescamell Just to be sure. Adding the below to your manifest doesn't work?

```
<application android:allowBackup="false" tools:replace="allowBackup" />
```

mikescamell:
Cheers @cmelchior , i just remembered i could do that so edited my comment!

cmelchior:
Closed by https://github.com/realm/realm-java/pull/2307

PR #2307, opened 2016-02-19
Removing allow back up

donnfelker:
References: #2227

The consequence of this is that the developer will have ignore the lint warning in their project because we are not setting it. This can be accomplished by adding a `lint.xml` file to the application with the following contents: 

```
<?xml version="1.0" encoding="UTF-8"?>
<lint>
    <issue id="AllowBackup" severity="ignore" />
</lint>
```

Then, in the applications build.gradle add the following lint config in the android block.

```
lintOptions {
    lintConfig file('lint.xml')
}
```

At that point, when `./gradlew lint` is run, the `AllowBackup` warning will be ignored.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 * Fixed bug when multiple calls of RealmResults.distinct() causes to return wrong results (#2198).
 * RealmResults.size() now returns Integer.MAX_VALUE when actual size is greater than Integer.MAX_VALUE (#2129).
 * Added RealmQuery.distinctAsync() and RealmResults.distinctAsync() (#2118).
+* Removed allowBackup from AndroidManifest (#2307).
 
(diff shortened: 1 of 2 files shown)


nhachicha:
:+1:  anyway in Android 6.0 the data is saved by default https://developer.android.com/training/backup/autosyncapi.html#configuring

emanuelez:
:+1:

zaki50:
@donnfelker the following configuration is probably the same.

```
lintOptions {
    ignore 'AllowBackup'
}
```

donnfelker:
You're correct, it is. ;)

>

cmelchior:
We should also update the changelog, but apart from that :+1:

donnfelker:
Changelog updated. Will wait for build success before merging.

PR #2320, opened 2016-02-23
Fix unit test example mocks

beeender:
#2303 
Also we need to enable `./gradlew check` on CI.


beeender:
retest this please

cmelchior:
It is unfortunate that this causes this exposes so many gotchas, but :+1:

dalinaum:
:+1:

PR #2327, opened 2016-02-24
migrateRealm throws when db doesn't exsit

### LINKED ISSUE #2316: Add a check if database file exists when Realm.migrateRealm() is called.
if `Realm.migrateRealm()` is called when Realm file does not exist, it may create wrong Realm file(no schema contained) or may cause the exception in Migration code.

This issue is reported by @tiro105 on slack chat. Thanks!

We have 2 choices in the case of missing Realm file.

* throw exception if Realm file does not exist
* create Realm file and execute normal initialization of schema instead of executing migration

beeender:
Close #2316 .

### DIFF
--- a/realm/realm-library/src/androidTest/java/io/realm/RealmMigrationTests.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/RealmMigrationTests.java
@@ -24,10 +24,13 @@ import org.junit.After;
 import org.junit.Before;
 import org.junit.Rule;
 import org.junit.Test;
+import org.junit.rules.ExpectedException;
 import org.junit.runner.RunWith;
 
 import java.io.File;
+import java.io.FileNotFoundException;
 import java.io.IOException;
+import java.util.Date;
 
 import io.realm.entities.AllTypes;
 import io.realm.entities.AnnotationTypes;
@@ -49,6 +52,8 @@ public class RealmMigrationTests {
 
     @Rule
     public final TestRealmConfigurationFactory configFactory = new TestRealmConfigurationFactory();
+    @Rule
+    public final ExpectedException thrown = ExpectedException.none();
 
     private Realm realm;
     private Context context;
@@ -165,7 +170,7 @@ public class RealmMigrationTests {
         try {
             realm = Realm.getInstance(realmConfig);
             fail();
-        } catch (RealmMigrationNeededException expected) {
+        } catch (RealmMigrationNeededException ignored) {
         } finally {
             if (realm != null) {
                 realm.close();
@@ -307,7 +312,7 @@ public class RealmMigrationTests {
         try {
             realm = Realm.getInstance(configFactory.createConfiguration());
             fail();
-        } catch (RealmMigrationNeededException expected) {
+        } catch (RealmMigrationNeededException ignored) {
         }
     }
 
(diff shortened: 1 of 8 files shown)


realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -1415,8 +1421,9 @@ private void checkMaxDepth(int maxDepth) {
      * version, nothing will happen.
      *
      * @param configuration {@link RealmConfiguration}
+     * @throws FileNotFoundException if the Realm file doesn't exist.
      */
-    public static void migrateRealm(RealmConfiguration configuration) {
+    public static void migrateRealm(RealmConfiguration configuration) throws FileNotFoundException {
```
cmelchior: Do we really want to make this a checked exception? I would say it isn't an expected error, but a logical error in the code path?

realm/realm-library/src/androidTest/java/io/realm/RealmMigrationTests.java
```diff
@@ -526,7 +540,7 @@ public void migrate(DynamicRealm realm, long oldVersion, long newVersion) {
     }
 
     @Test
-    public void realmOpenBeforeMigrationThrows() {
+    public void realmOpenBeforeMigrationThrows() throws FileNotFoundException {
```
cmelchior: See comment below

realm/realm-library/src/androidTest/java/io/realm/RealmMigrationTests.java
```diff
@@ -544,6 +558,29 @@ public void migrate(DynamicRealm realm, long oldVersion, long newVersion) {
...
+        RealmConfiguration config = configFactory.createConfigurationBuilder().migration(new RealmMigration() {
+            @Override
+            public void migrate(DynamicRealm realm, long oldVersion, long newVersion) {
+
+            }
+        }).build();
+        Realm.migrateRealm(config);
+    }
+
+    @Test(expected = FileNotFoundException.class)
```
cmelchior: Don't use this annotation if the test case contains multiple lines of code. The wrong code might throw instead: https://github.com/realm/realm-java/blob/master/CONTRIBUTING.md#unit-tests ;)

realm/realm-library/src/main/java/io/realm/Realm.java
cmelchior: Thinking about this a bit more. FileNotFoundException is a lot better than `IllegalArgument/IllegalState`, and perhaps it is good to force people to consider that the file they try to migrate doesn't exist. I think I could be convinced this is OK :+1:

cmelchior:
Looks good, note that the migration example fails to compile ;)
Apart from small comment about a test case  :+1:

realm/realm-library/src/main/java/io/realm/Realm.java
beeender: Yeah, I got to the same conclusion after found `FileNotFoundException` is a checked exception. It does make sense to be checked.

cmelchior:
Looks good :+1:

stk1m1:
:+1:

zaki50:
:+1:

realm/realm-library/src/androidTest/java/io/realm/RealmMigrationTests.java
```diff
@@ -544,6 +560,31 @@ public void migrate(DynamicRealm realm, long oldVersion, long newVersion) {
...
+        thrown.expect(FileNotFoundException.class);
+        Realm.migrateRealm(config);
+    }
+
+    @Test
+    public void migrateRealm_configMigration_nonExistingRealmFile() throws FileNotFoundException {
+        RealmMigration migration = new RealmMigration() {
+            @Override
+            public void migrate(DynamicRealm realm, long oldVersion, long newVersion) {
+
```
studamits: zxcv

PR #2337, opened 2016-02-25
allow 'realm' and 'row' as a field name of model class. This fixes #2255

### LINKED ISSUE #2255: Naming restriction of attributes in RealmObjects
Goal: create a RealmObject class with an attribute named "realm".
Expected results: No compilation errors
Actual results: "Error: realm has private accessor" compiler error, due to internal realm.io variable also called "realm", thus the whole proxy class is getting confused as to which realm to use for the proxy functions.
Steps to reproduce: Create a class extending RealmObject, add an attribute named realm, compile

Realm Version: 0.87.0

Well, the problem is pretty obvious. The proxy class is extending the actual class, tries to override the attribute named realm, because apparently you use a variable called "realm" for the proxy implementation. The proxy class then finds that this variable has a private accessor and throws an exception. Either disallow attributes with the name "realm" (bad idea) or rename the internal attribute to "proxyRealm" or something like that.

zaki50:
fixes #2255

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@
 * RealmResults.size() now returns Integer.MAX_VALUE when actual size is greater than Integer.MAX_VALUE (#2129).
 * Removed allowBackup from AndroidManifest (#2307).
 * Added multi-arguments distinct(...) for Realm, DynamicRealm, RealmQuery, and RealmResults (#2284).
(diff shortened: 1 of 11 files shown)


nhachicha:
:+1:

cmelchior:
:+1:

beeender:
:+1:

Issue #575, opened 2014-11-18
RealmList<Integer / String / ...>


an-k:
Dear realm-Team, 

I'm really impressed by Realm and I do like the way you're providing support and solutions on github, SO and gGroups! I like working with Realm - however there are three really big drawbacks for me:

1) No model-instantiation by `= new ...` - I know you're on it. I can't wait for this! Do we have to wait for the "whole" JSON-Upgrade or will there be an earlier release providing this feature?

2) No model-passing across threads. That's hindering all the great benefits of EventBus and Android Priority Job Queue. But I read that it will be possible in the further versions and I'm looking forward on this.

3) **The reason for this issue:**
In my model class I'm allowed to have primitive fields as well as object fields of objects that extend `RealmObject`.
But I'm not allowed to have a field type like `RealmList<Integer>` or `RealmList<String>`.
However it is possible to work around this by

``` java
RealmList<RealmString> names;
```

and

``` java
public class RealmString extends RealmObject {
private String name;
// getter & setter
```

It works but it's _really_ ugly and should be unnecessary in my opinion.
Is there any improvement planned regarding this?

Thanks again for your great work - I'm glad Realm helps me developing my app!

--- 

Related PRs:
 - [ ] API design (#4165)
 - [ ] Implementation (#5031)

jenzz:
Can I ask what the current state and/or priority of this feature request is?
Or: What is the recommended workaround for now?

mpost:
I agree that this is a worthwhile feature. +1.

PeterKomoot:
Yes, please

kboyarshinov:
+1

kaushikmukunda:
Would help solve issues like http://stackoverflow.com/questions/28733024/gson-deserialization-of-liststring-into-realmlistrealmstring 
+1

FHellmann:
+1

theyann:
+1 for the List of primitive and also +1 for the Enums.
I'm actually kinda stuck now as I have a JSON with a list of strings that needs to go to the DB, not quite sure yet how to deal with that.

RobertoArtiles:
+1

kbpontius:
+1

jmachete:
+1

marianmoldovan:
+1

theostanton:
+1

eygraber:
+1

harish1312:
A list of Strings is a very common use case for JSON responses. When you are given an array of Strings such as 

[...]

PR #2345, opened 2016-02-25
RealmCollection API

### LINKED ISSUE #1363: RealmCollection interface
We need a `RealmCollection` to be able to parse different lists into UI components.

- [ ] Figure out exactly which methods should be in the interface 
- [ ] Make `RealmResults` and `RealmList` implement the interface
- [ ] Document the interface on the website (#2258).

v1 Needs to support `RealmResults` and `RealmList` /cc #972  
v2 Support for RealmSet when it becomes available #759 
v3 Support for RealmMap. Will most likely require its own interface. #759

cmelchior:
Fixes #1363 

# TODO

- [x] Split existing unit tests so they work as parameterised tests.
- [x] Implement new methods in RealmList.
- [x] Implement new methods in RealmResults.
- [x] Determine exactly what should happen with `remove`/`clear` for RealmResults. Core doesn't currently support mutating a TableView. RealmResults will have immutable behaviour until https://github.com/realm/realm-core/issues/1530 is fixed.
- [x] Ensure that everybody is onboard with the naming of interfaces.
- [ ] Should we shorten `delete*FromRealm` to just `delete*` ?
- [x] `Realm.clear()` -> `Realm.delete/deleteAll`
- [x] `RealmObject.removeFromRealm()` -> `RealmObject.deleteFromRealm()`
- [x] Change BaseRealmAdapter to work with `RealmOrderedCollection`.

**Behavior changes**
- `first()`/`last()` now returns null instead of throwing an Exception if the collection is empty (RealmList did that / RealmResults threw an exception before).
- `remove`/`removeAll`/`retainAll`/`clear` now all throws an UnsupportedException for RealmResults. All methods have been replaced by `delete*` instead.
- `indexOf` no longer throws UnsupportedOperation on RealmResults.
- `sort()` have been added to RealmList and both now returns a sorted copy instead of sorting in-place.
- BaseRealm is now public. Should it be renamed to something else? `AbstractRealm`, `BaseRealm`?

**Test coverage**
Coverage (Instructions / Branches)
RealmList: 
new: 96% / 88%
old: 89% / 78%
RealmResults: 
new: 92% / 88%
old: 77% / 76%

------
# Discussion / Details
The primary driver for this is the following design goal

1) Expose interfaces to UI components so they don't have to know the specific implementation.
   This is currently a problem in `BaseRealmAdapter` that only accepts `RealmResults` but not 
   `RealmList`.

2) Support future collection types.

3) Mirror RealmCollectionType in Swift: https://realm.io/docs/swift/latest/api/Protocols/RealmCollectionType.html 

## The current situation

Right now Realm exposes two collections: `RealmList` and `RealmResults`. Both extend from 
`AbstractList` but have different public interface behavior. Most noticeably around 
`List.remove()` which in `RealmList` only removes objects from the list, while in `RealmResult` 
it also removes the object from the underlying Realm.

This proposal also fixes these inconsistencies.

## What collections do we need? 

```
// Current collections
RealmList
RealmResults

// Future collections
RealmSet
RealmMap

// Even further in the future
TreeRealmMap
CustomRealmList
...
```

## The proposal

The driving design principle was to mirror the structure in the Java Collection API to increase 
flexibility in regards to the number of collection types we implement later.

```
// New interfaces

public interface RealmCollection<E extends RealmObject> extends Collection {
    // Query methods
    RealmQuery<E> where();
    Number min(String field);
    Number max(String field);
    Number sum(String field);
    Number average(String field);
    Date maxDate(String field);
    Date minDate(String field);

    // Removal from the underlying realm. `clear` will not
    // do that anymore, only remove objects from the Collection.
    boolean deleteAllFromRealm();

    // Async methods ()
    boolean isLoaded();
    boolean load();
    
    boolean isValid();
}

public interface OrderedRealmCollection<E extends RealmObject> extends List, RealmCollection {

    // Removal from the underlying realm. `remove(location)` will not
    // do that anymore, only remove objects from the Collection.
    // delete* methods returns a boolean where possible to mimic `remove` behaviour.
    void deleteFromRealm(int location);
    boolean deleteFirstFromRealm();
    boolean deleteLastFromRealm();

    E first();
    E last();

    // These have been changed from sort-in-place to return a sorted copy. Only way to support      RealmList.sort() which isn't ready in core yet, but behaviour is also more flexible.
    RealmResults<E> sort(String field);
    RealmResults<E> sort(String field, Sort sortOrder);
    RealmResults<E> sort(String field1, Sort sortOrder1, String field2, Sort sortOrder2);
    RealmResults<E> sort(String[] fields, Sort[] sortOrders);
}

public interface RealmUnorderedCollection extends Set, RealmCollection {
    // Not used right now, so don't implement right away.
}
```

Methods like `clear` and `remove` have origins the Collection API which is also why we have have method names like `RealmObject.removeFromRealm()` and `Realm.clear()` as we wanted to make that conncection stronger. However if we need to reserve the use of these methods to only operate on the collection, we should also strongly consider  renaming these methods to make them more apart from normal Collection methods. Swift uses `delete`/`deleteAll` and I propose we change to that terminology as well.

* `Realm.clear()` -> `Realm.deleteAll()`
* `Realm.clear(Class)` -> `Realm.delete(Class)`
* `RealmObject.removeFromRealm()` -> `RealmObject.deleteFromRealm()`
* `RealmList.removeAllFromRealm()` -> `RealmList.deleteAllFromRealm()`

## Consequences

- All collections now only modify the collection with the Collection API methods, i.e. data is not 
  removed from the underlying Realm using these methods. This is not a guarentee that objects that
  the object is excluded always. In the case of RealmResults, it might appear again after a refresh
  but since this proposal build on top of Delayed Notifactions, it will be gone until a ChangeListener
  is triggered. This makes it possible to manipulate RealmResults much more freely than today, where
  RealmResults are "live" at all times.

- Since we are departing from using the Collection API to cover actual deletes from Realm this is
  also a good change to unify the naming. This is a depature from previous decissions where we
  wanted to stay as close as possible to the Collection API (which are now giving us a lot of 
  headache).

- This proposal means that all collections will work like RealmList does today. It will introduce
  a breaking change for RealmResults that need to move from `results.remove(index)` to 
  `results.deleteFromRealm(index)`.

- Going forward this proposal means that all Collections no have the same semantics for the methods
  implemented as well as enable the support of more advanced collection types later. 

### Observations and reasoning

* List / Set / Map already used by the Java framework API

* We cannot differentiate between Mutable and Immutable lists. We have to throw in RealmResults for 
  mutator methods  :(

* CollectionType in Swift is not the same as Collection in Java. 

* Our current collections are very list-centric or "ordered".

* Take-aways from the guys designing the offical Collection interface: 
  http://docs.oracle.com/javase/1.5.0/docs/guide/collections/designfaq.html#1

* Java Collection API overview: http://ordinarygeek.me/wp-content/uploads/2009/12/collections-2.png

* Even though the most accurate name for a shared class between `RealmResults` and `RealmList` would 
  be something like `AbstractRealmList` it would be a quite "ugly" name as input argument to a method.
  RealmList would be much more accurate (and fit the List interface), but is unfortunately already
  taken. This proposal suggests "OrderedRealmCollection" and "UnorderedRealmCollection".

* RealmList does not support ChangeListeners, so change listeners will either be added later or split into their own interfaces. Once we have fine-grained notifications they will most likely influence the design of that.

### Alternative names considered

* RealmOrderedCollection:
    - OrderedRealmCollection
    - RealmEnumerable
    - RealmAbstractList
    - AbstractRealmList
    - RealmCollectionType
    - RealmList -> RealmArrayList. Raise RealmList to interface status.

## Feedback requested

* Is the proposal in general acceptable?
* Naming: Do you like OrderedRealmCollection?
* Naming: What do you think about renaming from `remove*` to `delete*`?
* Should `remove`/`clear` be allowed for RealmResults or should it just throw?
* Should we move all `delete*` methods to Realm instead of having them on `RealmObject`, `RealmResults` and `RealmList`?

----
EDIT
Regarding sort methods. If we want to support sort on all `RealmOrderedCollection`s we need to use `RealmResults<Foo> sort(field, sort)` as a RealmList cannot be sorted in-place and needs to be converted to a RealmResults in order to so.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,19 @@
 ## 0.89.0
 
-### Enhancements
+### Breaking changes
 
+* RealmResults.clear() now throws UnsupportedOperationException. Use RealmResults.deleteAllFromRealm() instead.
+* RealmResults.remove(int) now throws UnsupportedOperationException. Use RealmResults.deleteFromRealm() instead.
+
+### Deprecated
+
+* RealmObject.removeFromRealm() in place of RealmObject.deleteFromRealm()
+* Realm.clear(Class) in favour of Realm.delete(Class).
+* DynamicRealm.clear(Class) in place of DynamicRealm.delete(Class).
+
+### Enhancements
+* RealmCollection and OrderedRealmCollection have been added. RealmList and RealmResults both implement these interfaces.
+* RealmBaseAdapter now accept an OrderedRealmCollection instead of only RealmResults.
 * RealmObjectSchema.isPrimaryKey(String) (#2440)
 
 ### Bug fixes
--- a/realm/realm-library/src/androidTest/java/io/realm/CollectionTests.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/CollectionTests.java
@@ -1,0 +1,217 @@
+/*
+ * Copyright 2016 Realm Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.realm;
+
+import java.util.Date;
+import java.util.concurrent.TimeUnit;
+
+import io.realm.entities.AllJavaTypes;
+import io.realm.entities.CyclicType;
+import io.realm.entities.NonLatinFieldNames;
+import io.realm.entities.NullTypes;
+
+/**
+ * Super class for all RealmCollection related tests.
+ * This class only contains configuration and helper methods.
+ */
+public abstract class CollectionTests {
+
+    protected final static long YEAR_MILLIS = TimeUnit.DAYS.toMillis(365);
+
+    // Enumerate all known collection classes from the Realm API.
+    protected enum CollectionClass {
+        MANAGED_REALMLIST, UNMANAGED_REALMLIST, REALMRESULTS
+    }
+
+    // Enumerate all current supported collections that can be in un-managed mode.
+    protected enum UnManagedCollection {
+        UNMANAGED_REALMLIST
+    }
+
+    // Enumerate all current supported collections that can be managed by Realm.
+    protected enum ManagedCollection {
+        MANAGED_REALMLIST, REALMRESULTS
+    }
+
+    // Enumerate all methods from the RealmCollection interface that depend on Realm API's.
+    protected enum RealmCollectionMethod {
+        WHERE, MIN, MAX, SUM, AVERAGE, MIN_DATE, MAX_DATE, DELETE_ALL_FROM_REALM, IS_VALID
+    }
+
+    // Enumerate all methods from the Collection interface
+    protected enum CollectionMethod {
+        ADD_OBJECT, ADD_ALL_OBJECTS, CLEAR, CONTAINS, CONTAINS_ALL, EQUALS, HASHCODE, IS_EMPTY, ITERATOR, REMOVE_OBJECT,
+        REMOVE_ALL, RETAIN_ALL, SIZE, TO_ARRAY, TO_ARRAY_INPUT
+    }
+
+    // Enumerate all methods on the List interface and OrderedRealmCollection interface that doesn't depend on Realm
+    // API's.
+    protected enum ListMethod {
+        FIRST, LAST, ADD_INDEX, ADD_ALL_INDEX, GET_INDEX, INDEX_OF, LAST_INDEX_OF, LIST_ITERATOR, LIST_ITERATOR_INDEX, REMOVE_INDEX,
+        SET, SUBLIST
+    }
+
+    // Enumerate all methods from the OrderedRealmCollection interface that depend on Realm API's.
+    protected enum OrderedRealmCollectionMethod {
+        DELETE_INDEX, DELETE_FIRST, DELETE_LAST, SORT, SORT_FIELD, SORT_2FIELDS, SORT_MULTI
+    }
+
+    // Enumerate all methods that can mutate a RealmCollection
+    protected enum CollectionMutatorMethod {
+        DELETE_ALL, ADD_OBJECT, ADD_ALL_OBJECTS, CLEAR, REMOVE_OBJECT, REMOVE_ALL, RETAIN_ALL
+    }
+
+    // Enumerate all methods that can mutate a RealmOrderedCollection
+    protected enum OrderedCollectionMutatorMethod {
+        DELETE_INDEX, DELETE_FIRST, DELETE_LAST, ADD_INDEX, ADD_ALL_INDEX, SET, REMOVE_INDEX
+    }
+
+    protected void populateRealm(Realm realm, int objects) {
+        realm.beginTransaction();
+        realm.delete(AllJavaTypes.class);
+        realm.delete(NonLatinFieldNames.class);
+        for (int i = 0; i < objects; i++) {
+            AllJavaTypes obj = realm.createObject(AllJavaTypes.class, i);
+            fillObject(i, objects, obj);
+            NonLatinFieldNames nonLatinFieldNames = realm.createObject(NonLatinFieldNames.class);
+            nonLatinFieldNames.set델타(i);
+            nonLatinFieldNames.setΔέλτα(i);
+        }
+
+        // Add all items to the RealmList on the first object
+        AllJavaTypes firstObj = realm.where(AllJavaTypes.class).equalTo(AllJavaTypes.FIELD_LONG, 0).findFirst();
+        RealmResults<AllJavaTypes> listData = realm.where(AllJavaTypes.class).findAllSorted(AllJavaTypes.FIELD_LONG, Sort.ASCENDING);
+        RealmList<AllJavaTypes> list = firstObj.getFieldList();
+        for (int i = 0; i < listData.size(); i++) {
+            list.add(listData.get(i));
+        }
+        realm.commitTransaction();
+    }
+
+    protected RealmList<AllJavaTypes> populateInMemoryList(int objects) {
+        RealmList<AllJavaTypes> list = new RealmList<AllJavaTypes>();
+        for (int i = 0; i < objects; i++) {
+            AllJavaTypes obj = new AllJavaTypes();
+            fillObject(i, objects, obj);
+            list.add(obj);
+        }
+        return list;
+    }
+
+    private void fillObject(int index, int totalObjects, AllJavaTypes obj) {
+        obj.setFieldLong(index);
+        obj.setFieldInt(index);
+        obj.setFieldBoolean(((index % 2) == 0));
+        obj.setFieldBinary(new byte[]{1, 2, 3});
+        obj.setFieldDate(new Date(YEAR_MILLIS * 20 * (index - totalObjects / 2)));
+        obj.setFieldDouble(3.1415 + index);
+        obj.setFieldFloat(1.234567f + index);
+        obj.setFieldString("test data " + index);
+    }
+
+    // Creates a collection that is based on a RealmList that was deleted.
+    protected OrderedRealmCollection<CyclicType> populateCollectionOnDeletedLinkView(Realm realm, ManagedCollection collectionClass) {
+        realm.beginTransaction();
+        CyclicType parent = realm.createObject(CyclicType.class);
+        for (int i = 0; i < 10; i++) {
+            CyclicType child = new CyclicType();
+            child.setName("name_" + i);
+            child.setObject(parent);
+            parent.getObjects().add(child);
+        }
+        realm.commitTransaction();
+
+        OrderedRealmCollection<CyclicType> result;
+        switch(collectionClass) {
+            case MANAGED_REALMLIST:
+                result = parent.getObjects();
+                break;
+            case REALMRESULTS:
+                result = parent.getObjects().where().equalTo(CyclicType.FIELD_NAME, "name_0").findAll();
+                break;
+            default:
+                throw new AssertionError("Unknown collection: " + collectionClass);
+        }
+
+        realm.beginTransaction();
+        parent.deleteFromRealm();
+        realm.commitTransaction();
+        return result;
+    }
+
+    // Create a number of objects that mix null and real values for number type fields.
+    protected void populatePartialNullRowsForNumericTesting(Realm realm) {
+        NullTypes nullTypes1 = new NullTypes();
+        nullTypes1.setId(1);
+        nullTypes1.setFieldIntegerNull(1);
+        nullTypes1.setFieldFloatNull(2F);
+        nullTypes1.setFieldDoubleNull(3D);
+        nullTypes1.setFieldBooleanNull(true);
+        nullTypes1.setFieldStringNull("4");
+        nullTypes1.setFieldDateNull(new Date(12345));
+
+        NullTypes nullTypes2 = new NullTypes();
+        nullTypes2.setId(2);
+
+        NullTypes nullTypes3 = new NullTypes();
+        nullTypes3.setId(3);
+        nullTypes3.setFieldIntegerNull(0);
+        nullTypes3.setFieldFloatNull(0F);
+        nullTypes3.setFieldDoubleNull(0D);
+        nullTypes3.setFieldBooleanNull(false);
+        nullTypes3.setFieldStringNull("0");
+        nullTypes3.setFieldDateNull(new Date(0));
+
+        realm.beginTransaction();
+        realm.copyToRealm(nullTypes1);
+        realm.copyToRealm(nullTypes2);
+        realm.copyToRealm(nullTypes3);
+        realm.commitTransaction();
+    }
+
+    // Create a list of AllJavaTypes with its `fieldString` field set to a given value.
+    protected OrderedRealmCollection<AllJavaTypes> createStringCollection(Realm realm, ManagedCollection collectionClass, String... args) {
+        realm.beginTransaction();
+        realm.deleteAll();
+        switch (collectionClass) {
+            case MANAGED_REALMLIST:
+                int id = 0;
+                for (String arg : args) {
+                    AllJavaTypes obj = realm.createObject(AllJavaTypes.class, id++);
+                    obj.setFieldString(arg);
+                }
+                realm.commitTransaction();
+                return realm.allObjects(AllJavaTypes.class);
+
+            case REALMRESULTS:
+                AllJavaTypes first = realm.createObject(AllJavaTypes.class);
+                first.setFieldString(args[0]);
+                first.getFieldList().add(first);
+                for (int i = 1; i < args.length; i++) {
+                    AllJavaTypes obj = realm.createObject(AllJavaTypes.class, i);
+                    obj.setFieldString(args[i]);
+                    first.getFieldList().add(obj);
+                }
+                realm.commitTransaction();
+                return first.getFieldList();
+
+            default:
+                throw new AssertionError("Unknown collection: " + collectionClass);
+        }
+    }
+
+}
--- a/realm/realm-library/src/androidTest/java/io/realm/ManagedOrderedRealmCollectionTests.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/ManagedOrderedRealmCollectionTests.java
@@ -1,0 +1,762 @@
+/*
+ * Copyright 2016 Realm Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.realm;
+
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+
+import io.realm.entities.AllJavaTypes;
+import io.realm.entities.Cat;
+import io.realm.entities.Dog;
+import io.realm.entities.NullTypes;
+import io.realm.entities.Owner;
+import io.realm.rule.TestRealmConfigurationFactory;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+
+/**
+ * Test class for all methods specific to OrderedRealmCollections that are not implementation specific.
+ *
+ * Methods tested in this class:
+ *
+ * # OrderedRealmCollection
+ *
+ * - E first()
+ * - E last()
+ * + void sort(String field)
+ * + void sort(String field, Sort sortOrder)
+ * + void sort(String field1, Sort sortOrder1, String field2, Sort sortOrder2)
+ * + void sort(String[] fields, Sort[] sortOrders)
+ * + void deleteFromRealm(int location)
+ * + void deleteFirstFromRealm()
+ * + void deleteLastFromRealm();
+ *
+ * # List
+ *
+ *  - void add(int location, E object);
+ *  - boolean addAll(int location, Collection<? extends E> collection);
+ *  - E get(int location);
+ *  - int indexOf(Object object);
+ *  - int lastIndexOf(Object object);
+ *  - ListIterator<E> listIterator();
+ *  - ListIterator<E> listIterator(int location);
+ *  - E remove(int location);
+ *  - E set(int location, E object);
+ *  - List<E> subList(int start, int end);
+ *
+ * # RealmCollection
+ *
+ * - RealmQuery<E> where();
+ * - Number min(String fieldName);
+ * - Number max(String fieldName);
+ * - Number sum(String fieldName);
+ * - double average(String fieldName);
+ * - Date maxDate(String fieldName);
+ * - Date minDate(String fieldName);
+ * - void deleteAllFromRealm();
+ * - boolean isLoaded();
+ * - boolean load();
+ * - boolean isValid();
+ * - BaseRealm getRealm();
+ *
+ * # Collection
+ *
+ * - public boolean add(E object);
+ * - public boolean addAll(Collection<? extends E> collection);
+ * - public void deleteAll();
+ * - public boolean contains(Object object);
+ * - public boolean containsAll(Collection<?> collection);
+ * - public boolean equals(Object object);
+ * - public int hashCode();
+ * - public boolean isEmpty();
+ * - public Iterator<E> iterator();
+ * - public boolean remove(Object object);
+ * - public boolean removeAll(Collection<?> collection);
+ * - public boolean retainAll(Collection<?> collection);
+ * - public int size();
+ * - public Object[] toArray();
+ * - public <T> T[] toArray(T[] array);
+ *
+ * @see RealmCollectionTests
+ * @see ManagedRealmCollectionTests
+ * @see UnManagedRealmCollectionTests
+ */
+
+@RunWith(Parameterized.class)
+public class ManagedOrderedRealmCollectionTests extends CollectionTests {
+
+    private static final int TEST_SIZE = 10;
+    private final static int TEST_DATA_FIRST_HALF = (int) ((TEST_SIZE / 2.0D) - 1);
+    private final static int TEST_DATA_LAST_HALF = (int) ((TEST_SIZE / 2.0D) + 1);
+
+    @Rule
+    public final TestRealmConfigurationFactory configFactory = new TestRealmConfigurationFactory();
+    @Rule
+    public final ExpectedException thrown = ExpectedException.none();
+
+    private final ManagedCollection collectionClass;
+    private Realm realm;
+    private OrderedRealmCollection<AllJavaTypes> collection;
+
+    @Parameterized.Parameters(name = "{0}")
+    public static List<ManagedCollection> data() {
+        return Arrays.asList(ManagedCollection.values());
+    }
+
+    public ManagedOrderedRealmCollectionTests(ManagedCollection collectionType) {
+        this.collectionClass = collectionType;
+    }
+
+    @Before
+    public void setup() {
+        realm = Realm.getInstance(configFactory.createConfiguration());
+        populateRealm(realm, TEST_SIZE);
+        collection = createCollection(collectionClass);
+    }
+
+    OrderedRealmCollection<AllJavaTypes> createCollection(ManagedCollection collectionClass) {
+        switch (collectionClass) {
+            case MANAGED_REALMLIST:
+                return realm.where(AllJavaTypes.class)
+                        .equalTo(AllJavaTypes.FIELD_LONG, 0)
+                        .findFirst()
+                        .getFieldList();
+
+            case REALMRESULTS:
+                return realm.allObjects(AllJavaTypes.class);
+
+            default:
+                throw new AssertionError("Unsupported class: " + collectionClass);
+        }
+    }
+
+    private OrderedRealmCollection<NullTypes> createEmptyCollection(Realm realm, ManagedCollection collectionClass) {
+        switch (collectionClass) {
+            case MANAGED_REALMLIST:
+                realm.beginTransaction();
+                NullTypes obj = realm.createObject(NullTypes.class);
+                realm.commitTransaction();
+                return obj.getFieldListNull();
+
+            case REALMRESULTS:
+                return realm.where(NullTypes.class).findAll();
+        }
+
+        throw new AssertionError("Unknown collection: " + collectionClass);
+    }
+
+    @After
+    public void tearDown() {
+        realm.close();
+    }
+
+    @Test
+    public void sort_twoFields() {
+        OrderedRealmCollection<AllJavaTypes> sortedList = collection.sort(AllJavaTypes.FIELD_BOOLEAN, Sort.ASCENDING, AllJavaTypes.FIELD_LONG, Sort.DESCENDING);
+        AllJavaTypes obj = sortedList.first();
+        assertFalse(obj.isFieldBoolean());
+        assertEquals(TEST_SIZE - 1, obj.getFieldLong());
+    }
+
+    @Test
+    public void sort_boolean() {
+        OrderedRealmCollection<AllJavaTypes> sortedList = collection.sort(AllJavaTypes.FIELD_BOOLEAN, Sort.DESCENDING);
+        assertEquals(TEST_SIZE, sortedList.size());
+        assertEquals(false, sortedList.last().isFieldBoolean());
+        assertEquals(true, sortedList.first().isFieldBoolean());
+        assertEquals(true, sortedList.get(TEST_DATA_FIRST_HALF).isFieldBoolean());
+        assertEquals(false, sortedList.get(TEST_DATA_LAST_HALF).isFieldBoolean());
+
+        RealmResults<AllJavaTypes> reverseList = sortedList.sort(AllJavaTypes.FIELD_BOOLEAN, Sort.ASCENDING);
+        assertEquals(TEST_SIZE, reverseList.size());
+        assertEquals(true, reverseList.last().isFieldBoolean());
+        assertEquals(false, reverseList.first().isFieldBoolean());
+        assertEquals(false, reverseList.get(TEST_DATA_FIRST_HALF).isFieldBoolean());
+        assertEquals(true, reverseList.get(TEST_DATA_LAST_HALF).isFieldBoolean());
+
+        RealmResults<AllJavaTypes> reserveSortedList = reverseList.sort(AllJavaTypes.FIELD_BOOLEAN, Sort.DESCENDING);
+        assertEquals(TEST_SIZE, reserveSortedList.size());
+        assertEquals(reserveSortedList.first(), sortedList.first());
+    }
+
+    @Test
+    public void sort_string() {
+        OrderedRealmCollection<AllJavaTypes> resultList = collection;
+        OrderedRealmCollection<AllJavaTypes> sortedList = createCollection(collectionClass);
+        sortedList = sortedList.sort(AllJavaTypes.FIELD_STRING, Sort.DESCENDING);
+
+        assertEquals(resultList.size(), sortedList.size());
+        assertEquals(TEST_SIZE, sortedList.size());
+        assertEquals(resultList.first().getFieldString(), sortedList.last().getFieldString());
+
+        RealmResults<AllJavaTypes> reverseList = sortedList.sort(AllJavaTypes.FIELD_STRING, Sort.ASCENDING);
+        assertEquals(TEST_SIZE, reverseList.size());
+        assertEquals(resultList.first().getFieldString(), reverseList.first().getFieldString());
+
+        int numberOfDigits = 1 + ((int) Math.log10(TEST_SIZE));
+        int largestNumber = 1;
+        largestNumber = (int) (largestNumber * Math.pow(10, numberOfDigits - 1));
+        largestNumber = largestNumber - 1;
+        assertEquals(resultList.get(largestNumber).getFieldString(), reverseList.last().getFieldString());
+        RealmResults<AllJavaTypes> reverseSortedList = reverseList.sort(AllJavaTypes.FIELD_STRING, Sort.DESCENDING);
+        assertEquals(TEST_SIZE, reverseSortedList.size());
+    }
+
+    @Test
+    public void sort_double() {
+        OrderedRealmCollection<AllJavaTypes> resultList = collection;
+        OrderedRealmCollection<AllJavaTypes> sortedList = createCollection(collectionClass);
+        sortedList = sortedList.sort(AllJavaTypes.FIELD_DOUBLE, Sort.DESCENDING);
+        assertEquals(resultList.size(), sortedList.size());
+        assertEquals(TEST_SIZE, sortedList.size());
+            assertEquals(resultList.first().getFieldDouble(), sortedList.last().getFieldDouble(), 0D);
+
+        RealmResults<AllJavaTypes> reverseList = sortedList.sort(AllJavaTypes.FIELD_DOUBLE, Sort.ASCENDING);
+        assertEquals(TEST_SIZE, reverseList.size());
+        assertEquals(resultList.first().getFieldDouble(), reverseList.first().getFieldDouble(), 0D);
+        assertEquals(resultList.last().getFieldDouble(), reverseList.last().getFieldDouble(), 0D);
+
+        RealmResults<AllJavaTypes> reverseSortedList = reverseList.sort(AllJavaTypes.FIELD_DOUBLE, Sort.DESCENDING);
+        assertEquals(TEST_SIZE, reverseSortedList.size());
+    }
+
+    @Test
+    public void sort_float() {
+        OrderedRealmCollection<AllJavaTypes> resultList = collection;
+        OrderedRealmCollection<AllJavaTypes> sortedList = createCollection(collectionClass);
+        sortedList = sortedList.sort(AllJavaTypes.FIELD_FLOAT, Sort.DESCENDING);
+        assertEquals(resultList.size(), sortedList.size());
+        assertEquals(TEST_SIZE, sortedList.size());
+        assertEquals(resultList.first().getFieldFloat(), sortedList.last().getFieldFloat(), 0D);
+
+        RealmResults<AllJavaTypes> reverseList = sortedList.sort(AllJavaTypes.FIELD_FLOAT, Sort.ASCENDING);
+        assertEquals(TEST_SIZE, reverseList.size());
+        assertEquals(resultList.first().getFieldFloat(), reverseList.first().getFieldFloat(), 0D);
+        assertEquals(resultList.last().getFieldFloat(), reverseList.last().getFieldFloat(), 0D);
+
+        RealmResults<AllJavaTypes> reverseSortedList = reverseList.sort(AllJavaTypes.FIELD_FLOAT, Sort.DESCENDING);
+        assertEquals(TEST_SIZE, reverseSortedList.size());
+    }
+
+    private void doTestSortOnColumnWithPartialNullValues(String fieldName,
+                                                         OrderedRealmCollection<NullTypes> original,
+                                                         OrderedRealmCollection<NullTypes> copy) {
+
+        RealmResults<NullTypes> sortedList = copy.sort(fieldName, Sort.ASCENDING);
+        assertEquals("Should have same size", original.size(), sortedList.size());
+        // Null should always be the first one in the ascending sorted list
+        assertEquals(2, sortedList.first().getId());
+        assertEquals(1, sortedList.last().getId());
+
+        // Descending
+        sortedList = sortedList.sort(fieldName, Sort.DESCENDING);
+        assertEquals("Should have same size", original.size(), sortedList.size());
+        assertEquals(1, sortedList.first().getId());
+        // Null should always be the last one in the descending sorted list
+        assertEquals(2, sortedList.last().getId());
+    }
+
+    // Test sort on nullable fields with null values partially
+    @Test
+    public void sort_rowsWithPartialNullValues() {
+        populatePartialNullRowsForNumericTesting(realm);
+        OrderedRealmCollection<NullTypes> original;
+        OrderedRealmCollection<NullTypes> copy;
+        switch (collectionClass) {
+            case MANAGED_REALMLIST:
+                realm.beginTransaction();
+                RealmResults<NullTypes> objects = realm.allObjects(NullTypes.class);
+                NullTypes parent = realm.createObject(NullTypes.class, 0);
+                for (int i = 0; i < objects.size(); i++) {
+                    NullTypes object = objects.get(i);
+                    if (object.getId() != 0) {
+                        parent.getFieldListNull().add(object);
+                    }
+                }
+                realm.commitTransaction();
+                original = parent.getFieldListNull().where().findAll();
+                copy = parent.getFieldListNull();
+                break;
+
+            case REALMRESULTS:
+                original = realm.allObjects(NullTypes.class);
+                copy = realm.allObjects(NullTypes.class);
+                break;
+
+            default:
+                throw new AssertionError("Unknown collection class: " + collectionClass);
+        }
+
+        // 1 String
+        doTestSortOnColumnWithPartialNullValues(NullTypes.FIELD_STRING_NULL, original, copy);
+
+        // 3 Boolean
+        doTestSortOnColumnWithPartialNullValues(NullTypes.FIELD_BOOLEAN_NULL, original, copy);
+
+        // 6 Integer
+        doTestSortOnColumnWithPartialNullValues(NullTypes.FIELD_INTEGER_NULL, original, copy);
+
+        // 7 Float
+        doTestSortOnColumnWithPartialNullValues(NullTypes.FIELD_FLOAT_NULL, original, copy);
+
+        // 8 Double
+        doTestSortOnColumnWithPartialNullValues(NullTypes.FIELD_DOUBLE_NULL, original, copy);
+
+        // 10 Date
+        doTestSortOnColumnWithPartialNullValues(NullTypes.FIELD_DATE_NULL, original, copy);
+    }
+
+    @Test
+    public void sort_nonExistingColumn() {
+        RealmResults<AllJavaTypes> resultList = realm.where(AllJavaTypes.class).findAll();
+        thrown.expect(IllegalArgumentException.class);
+        resultList.sort("Non-existing");
+    }
+
+    @Test
+    public void sort_danishCharacters() {
+        OrderedRealmCollection<AllJavaTypes> collection = createStringCollection(realm, collectionClass,
+                "Æble",
+                "Øl",
+                "Århus"
+        );
+
+        collection = collection.sort(AllJavaTypes.FIELD_STRING);
+
+        assertEquals(3, collection.size());
+        assertEquals("Æble", collection.get(0).getFieldString());
+        assertEquals("Øl", collection.get(1).getFieldString());
+        assertEquals("Århus", collection.get(2).getFieldString());
+
+        collection = collection.sort(AllJavaTypes.FIELD_STRING, Sort.DESCENDING);
+        assertEquals(3, collection.size());
+        assertEquals("Århus", collection.get(0).getFieldString());
+        assertEquals("Øl", collection.get(1).getFieldString());
+        assertEquals("Æble", collection.get(2).getFieldString());
+    }
+
+    @Test
+    public void sort_russianCharacters() {
+        OrderedRealmCollection<AllJavaTypes> collection = createStringCollection(realm, collectionClass,
+                "Санкт-Петербург",
+                "Москва",
+                "Новороссийск"
+        );
+
+        collection = collection.sort(AllJavaTypes.FIELD_STRING);
+
+        assertEquals(3, collection.size());
+        assertEquals("Москва", collection.get(0).getFieldString());
+        assertEquals("Новороссийск", collection.get(1).getFieldString());
+        assertEquals("Санкт-Петербург", collection.get(2).getFieldString());
+
+        collection = collection.sort(AllJavaTypes.FIELD_STRING, Sort.DESCENDING);
+        assertEquals(3, collection.size());
+        assertEquals("Санкт-Петербург", collection.get(0).getFieldString());
+        assertEquals("Новороссийск", collection.get(1).getFieldString());
+        assertEquals("Москва", collection.get(2).getFieldString());
+    }
+
+    @Test
+    public void sort_greekCharacters() {
+        OrderedRealmCollection<AllJavaTypes> collection = createStringCollection(realm, collectionClass,
+                "αύριο",
+                "ημέρες",
+                "δοκιμές"
+        );
+
+        collection = collection.sort(AllJavaTypes.FIELD_STRING);
+
+        assertEquals(3, collection.size());
+        assertEquals("αύριο", collection.get(0).getFieldString());
+        assertEquals("δοκιμές", collection.get(1).getFieldString());
+        assertEquals("ημέρες", collection.get(2).getFieldString());
+
+        collection = collection.sort(AllJavaTypes.FIELD_STRING, Sort.DESCENDING);
+        assertEquals(3, collection.size());
+        assertEquals("ημέρες", collection.get(0).getFieldString());
+        assertEquals("δοκιμές", collection.get(1).getFieldString());
+        assertEquals("αύριο", collection.get(2).getFieldString());
+    }
+
+    //No sorting order defined. There are Korean, Arabic and Chinese characters.
+    @Test
+    public void sort_manyDifferentCharacters() {
+        OrderedRealmCollection<AllJavaTypes> collection = createStringCollection(realm, collectionClass,
+                "단위",
+                "테스트",
+                "وحدة",
+                "اختبار",
+                "单位",
+                "试验",
+                "單位",
+                "測試"
+        );
+
+        collection.sort(AllJavaTypes.FIELD_STRING);
+        assertEquals(8, collection.size());
+
+        collection.sort(AllJavaTypes.FIELD_STRING, Sort.DESCENDING);
+        assertEquals(8, collection.size());
+    }
+
+    @Test
+    public void sort_twoLanguages() {
+        OrderedRealmCollection<AllJavaTypes> collection = createStringCollection(realm, collectionClass,
+                "test",
+                "αύριο",
+                "work"
+        );
+
+        try {
+            collection.sort(AllJavaTypes.FIELD_STRING);
+        } catch (IllegalArgumentException e) {
+            fail("Failed to sort with two kinds of alphabets");
+        }
+    }
+
+    @Test
+    public void sort_usingChildObject() {
+        realm.beginTransaction();
+        Owner owner = realm.createObject(Owner.class);
+        owner.setName("owner");
+        Cat cat = realm.createObject(Cat.class);
+        cat.setName("cat");
+        owner.setCat(cat);
+        realm.commitTransaction();
+
+        RealmQuery<Owner> query = realm.where(Owner.class);
+        RealmResults<Owner> owners = query.findAll();
+
+        try {
+            owners.sort("cat.name");
+            fail("Sorting by child object properties should result in a IllegalArgumentException");
+        } catch (IllegalArgumentException ignore) {
+        }
+    }
+
+    @Test
+    public void sort_nullArguments() {
+        OrderedRealmCollection<AllJavaTypes> result = collection;
+        try {
+            result.sort(null);
+            fail("Sorting with a null field name should throw an IllegalArgumentException");
+        } catch (IllegalArgumentException ignored) {
+        }
+        try {
+            result.sort((String) null, null);
+            fail();
+        } catch (IllegalArgumentException ignored) {
+        }
+    }
+
+    @Test
+    public void sort_emptyResults() {
+        OrderedRealmCollection<NullTypes> collection = createEmptyCollection(realm, collectionClass);
+        assertEquals(0, collection.size());
+        collection.sort(NullTypes.FIELD_STRING_NULL);
+        assertEquals(0, collection.size());
+    }
+
+    @Test
+    public void sort_singleField() {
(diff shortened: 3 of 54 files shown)


stk1m1:
- **Is the proposal in general acceptable?** I think this is great since FP flavor suits well in data manipulation and I see this as the first step toward in that direction.  
- **Naming: Do you like RealmOrderedCollection?** Question: Do `map`, `set` and `results` fall in _`RealmCollection`_ and `list` on _`RealmOrderedCollection`_? It seems _`RealmCollection`_ and _`RealmOrderedCollection`_ would sit on top of collection hierarchy eventually.
- **Naming: What do you think about renaming from remove* to delete*?** With the curtesy of Oxford Dictionary, [remove](http://www.oxforddictionaries.com/definition/english/remove) is _"Take (something) away or off from the position occupied"_ and [delete](http://www.oxforddictionaries.com/definition/english/delete) is _"Remove or obliterate (written or printed matter), especially by drawing a line through it"_.   
Former naming most likely implies in cognitive sense an object still remains in memory, and later does not I guess?
- **Should remove/clear be allowed for RealmResults or should it just throw?** I think it would be a great idea to have only two interfaces. In Scala, they have these `mutable` and `immutable` interfaces all over the places that it eventually bogs down to 1) performance issue due to memory pressure or 2) ugly implementation on mutable collections. I don't know if `RealmResults` is immutable. I would suggest we setup the 1st principle and follow it all the way through. 
- **Should we move all delete* methods to Realm instead of having them on RealmObject, RealmResults and RealmList?** I am not able to suggest a clear cut opinion here. I'd have `delete` on all of them just like `distinct`. (I guess we'd eventually have them in few months down the road?)

1. Have you also thought about having `map`, `filter`, and `reduce` in these interfaces? Even python has them. 
2. Any thought on how we convert one to anther. i.e. (toSet, toMap, and/or toList)? I think people would ask the feature. ( I, for one, for sure. )

beeender:
* "All collections now only modify the collection with the Collection API methods, i.e. data is not
removed from the underlying Realm using these methods. " This feels really really strange to me. If there is user case that user needs to hide some elements temporary, maybe a `RealmFilteredCollection` would be more accurate, the filter can applied automatically when Realm changes. Ask user to do this after refresh sounds like a broken API.
* `isValid` is also needed for `RealmCollection`
* change listeners related API for `RealmCollection`? The collection needs to be considered to support fine grained notification.
* `RealmUnorderedCollection` should implement `RealmCollection` , right?
* `RealmOrderedCollection` is against my instinct since `Set`/`Map` is **sorted**. Maybe something like `RealmSortableCollection`?
* Why `first()`/`last()` is not a part of `RealmCollection`?

cmelchior:
> All collections now only modify the collection with the Collection API methods, i.e. data is not removed from the underlying Realm using these methods. " This feels really really strange to me. If there is user case that user needs to hide some elements temporary, maybe a RealmFilteredCollection would be more accurate, the filter can applied automatically when Realm changes. Ask user to do this after refresh sounds like a broken API.

I agree it is slightly weird for `RealmResults`. `RealmList` works this way today. The underlying problem is that with RealmList we _don't_ delete items from Realm, while for `RealmResults` we do. We have to pick on or the other in order to unify them. IMO the first approach is closer to standard Java behaviour (an Object isn't GC'ed because your remove it), so it is preferable IMO, but it still raises the question if we should just throw in the `remove`/`clear` methods for `RealmResults`.

You are right that a FilterableCollection would perhaps be a more appropriate abstraction for this, e.g.: 
1) We throw in `remove/clear`
2) Add a `results.filter(new RealmResults.Filter() {
      public boolean boolean apply(Foo obj) {
        // return true to keep object, false to remove
      }
    }); 

I am afraid it would be really slow though and it is ideally something we should have in core. Allowing people to manually `remove/clear` would be manual filter though so the above is really just syntactical sugar. But would love more opinions on it as this is actually one of the core problems with this refactor.

> isValid is also needed for RealmCollection

Yes I will add it.

> change listeners related API for RealmCollection? The collection needs to be considered to support fine grained notification.

Yes, i kept them out of this as we still aren't entirely sure how it will look. Alternatively changelisteners will get their own interface that collections just extend at some point. 

> RealmUnorderedCollection should implement RealmCollection , right?

Yes, will fix it.

> RealmOrderedCollection is against my instinct since Set/Map is sorted. Maybe something like RealmSortableCollection?

HashMap/HashSet is not ordered, TreeMap/TreeSet is. I would be a bit hesitant to call it RealmSortableCollection if we don't have any sort methods on it :) but sorted/ordered is semantically the same thing right?

> Why first()/last() is not a part of RealmCollection?

It doesn't make sense to talk about "first" in an unordered collection like Set.

beeender:
>> Why first()/last() is not a part of RealmCollection?

> It doesn't make sense to talk about "first" in an unordered collection like Set.

A bit confused, but there is `RealmCollection.get(0)`, right? won't be that the same with `RealmCollection.first()`?

cmelchior:
> Naming: Do you like RealmOrderedCollection? Question: Do map, set and results fall in RealmCollection and list on RealmOrderedCollection? It seems RealmCollection and RealmOrderedCollection would sit on top of collection hierarchy eventually.

`RealmList`, `RealmResults ` are ordered. `RealmSet` is unordered. Map is not covered by this as they have their own interface in the Java Collection API http://ordinarygeek.me/wp-content/uploads/2009/12/collections-2.png

> Naming: What do you think about renaming from remove* to delete*? With the curtesy of Oxford Dictionary, remove is "Take (something) away or off from the position occupied" and delete is "Remove or obliterate (written or printed matter), especially by drawing a line through it". 
Former most likely implies in cognitive sense an object still remains in memory, and later does not I guess?

Yes, that + it aligns us with Cocoa is why I think it could be a good idea to rename them.


> Should remove/clear be allowed for RealmResults or should it just throw? I think it would be a great idea to have only two interfaces. In Scala, they have these mutable and immutable interfaces all over the places that it eventually bogs down to 1) performance issue due to memory pressure or 2) ugly implementation on mutable collections. I don't know if RealmResults is immutable. I would suggest we setup the 1st principle and follow it all the way through.

RealmResults are not immutable, but you cannot add items to it.  I am not entirely sure what you mean by "the `1st principle"?

> Have you also thought about having map, filter, and reduce in these interfaces? Even python has them.
Any thought on how we convert one to anther. i.e. (toSet, toMap, and/or toList)? I think people would ask the feature. ( I, for one, for sure. )

I thought about it, but feel that it is perhaps a bit out of scope for this issue as the implications on both the API and performance are really unclear, especially since we wouldn't have core support for any of it.

cmelchior:
>>> Why first()/last() is not a part of RealmCollection?

>> It doesn't make sense to talk about "first" in an unordered collection like Set.

> A bit confused, but there is `RealmCollection.get(0)`, right? won't be that the same with `RealmCollection.first()`?

No, RealmCollection only extends the standard Collection: https://docs.oracle.com/javase/7/docs/api/java/util/Collection.html which doesn't have `get(index)`. 

That is only available for the `List` interface https://docs.oracle.com/javase/7/docs/api/java/util/List.html, which is what `RealmOrderedCollection` extends. `RealmUnorderedCollection` extends `Set` which doesn't have `get(index)` either. http://docs.oracle.com/javase/7/docs/api/java/util/Set.html

stk1m1:
>>  I don't know if RealmResults is immutable. I would suggest we setup the 1st principle and follow it all the way through. 

> RealmResults are not immutable, but you cannot add items to it. I am not entirely sure what you mean by "the `1st principle"?

`the 1st principle` refers a solid goal of what this proposal want to accomplish. e.g. _we want a set of new **mutable** collections defined by these interfaces_. I guess that's what we are trying to draw from discussion, and I'd suggest to make it somehow clear up there. (The link was super great to read though. :+1:)

stk1m1:
>>> Why first()/last() is not a part of RealmCollection?

>> It doesn't make sense to talk about "first" in an unordered collection like Set.

> A bit confused, but there is `RealmCollection.get(0)`, right? won't be that the same with `RealmCollection.first()`?

Although it should be more than feasible, I, too, don't think that it make sense to put `first` and `last` to unordered collections in semantic sense. I can see it would result in very much unified behavior though.

cmelchior:
Another slightly annoying problem is that we probably end up having to change to  `RealmList<Object>` in order to support #575. This have implications for this interface as well as we need to refactor both RealmQuery and RealmResults to `RealmQuery<Object>` and `RealmResults<Object>` :confused: 

As this is just relaxing the types, it can be done without breaking anything, so maybe it is something se should just punt until we actually begin tackling #575 (that isn't far away in any case though).

stk1m1:
Is there anyway we can remove `remove*` family with better object? It would be much better for users to know there isn't such a thing in _compile time_ rather than in _runtime_.

cmelchior:
This is ready for review, but we should probably do a small session before people started reviewing it as there is quite a few changes.

Any review should just focus on `RealmCollection, `OrderedRealmCollection`, `RealmList` and `RealmResults`. The changes to all the unit tests are less interesting (and are 80% of this PR).

@realm/java

realm/realm-library/src/main/java/io/realm/DynamicRealm.java
```diff
@@ -120,9 +120,20 @@ public DynamicRealmObject createObject(String className, Object primaryKeyValue)
     /**
      * Removes all objects of the specified class.
      *
+     * DEPRECATED: Use {@link #delete(String)} instead.
+     *
      * @param className the class for which all objects should be removed.
      */
     public void clear(String className) {
```
zaki50: We should have `Deprecated` annotation here.

realm/realm-library/src/main/java/io/realm/RealmObjectSchema.java
```diff
@@ -593,16 +593,30 @@ private boolean isValidType(RealmFieldType columnType, RealmFieldType[] validCol
...
+     * Returns the column index in the underlying table for the given field name.
+     *
+     * @param fieldName field name to find index for.
+     * @return column index.
+     * @throws IllegalArgumentException if the field does not exists.
+     */
+    long getAndCheckFieldIndex(String fieldName) {
+        Long index =  columnIndices.get(fieldName);
+        if (index == null) {
+            throw new IllegalArgumentException("Field does not exist:" + fieldName);
```
zaki50: It's better to have angle space after `:`.

realm/realm-library/src/main/java/io/realm/RealmObjectSchema.java
```diff
@@ -593,16 +593,30 @@ private boolean isValidType(RealmFieldType columnType, RealmFieldType[] validCol
...
 
     /**
+     * Returns the column index in the underlying table for the given field name.
+     *
+     * @param fieldName field name to find index for.
+     * @return column index.
+     * @throws IllegalArgumentException if the field does not exists.
+     */
+    long getAndCheckFieldIndex(String fieldName) {
+        Long index =  columnIndices.get(fieldName);
```
stk1m1: Single space bar after `Long index =`?

realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -555,8 +555,11 @@ public RealmSchema getSchema() {
     /**
      * Removes all objects from this Realm.
      *
+     * DEPRECATED: Use {@link #deleteAll()}
+     *
      * @throws IllegalStateException if the corresponding Realm is closed or on an incorrect thread.
      */
+    @Deprecated
     public void clear() {
         checkIfValid();
```
zaki50: It's better to replace this method body with just `deleteAll();`.

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -1381,14 +1381,28 @@ public void run() {
...
+        delete(clazz);
+    }
+
+    /**
+     * Deletes all objects of the specified class from the Realm.
+     *
+     * @param clazz the class which objects should be removed.
+     * @throws IllegalStateException if the corresponding Realm is closed or called from an incorrect thread.
+     */
+    public void delete(Class<? extends RealmObject> clazz) {
```
zaki50: I prefer `deleteAll(Class)` instead of `delete(Class)` since this method deletes multiple objects.

What do you think @realm/java?

cmelchior: Swift uses `delete(Class)`, but I was debating the above myself as there are precedence for both, e.g. `Collection.removeAll(Collection)`. Personally I lean slightly towards using `delete(X)` for deleting any subset and `deleteAll()` for deleting the entire thing on either Realm/RealmList/RealmResults.

realm/realm-library/src/main/java/io/realm/RealmCollection.java
```diff
@@ -0,0 +1,157 @@
...
+import java.util.Date;
+
+import io.realm.RealmObject;
+import io.realm.RealmQuery;
+
+/**
+ * {@code RealmCollection} is the root of the collection hierarchy that Realm supports. It defines operations on data
+ * collections and the behavior that they will have in all implementations of {@code RealmCollection}s.
+ *
+ * Methods marked with (optional) can throw an
```
zaki50: No methods marked with (optional).

realm/realm-library/src/main/java/io/realm/Realm.java
zaki50: I can accept that. Please leave as it is.

realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -45,7 +45,7 @@
  * @see io.realm.Realm
  * @see io.realm.DynamicRealm
  */
-abstract class BaseRealm implements Closeable {
+public abstract class BaseRealm implements Closeable {
```
beeender: Why public?

realm/realm-library/src/main/java/io/realm/OrderedRealmCollection.java
```diff
@@ -0,0 +1,122 @@
...
+ * index, with the first index being zero. Normally, {@code OrderedRealmCollection}s allow duplicate
+ * elements, as compared to Sets, where elements have to be unique.
+ */
+public interface OrderedRealmCollection<E extends RealmObject> extends List<E>, RealmCollection<E> {
+
+    /**
+     * Gets the first object from the list or {@code null} if the collection is empty.
+     *
+     * @return the first object or {@code null} if the collection is empty.
+     */
```
beeender: As we discussed in the meeting, throw `IndexOutOfBoundsException` if the size is 0.

realm/realm-library/src/androidTest/java/io/realm/CollectionTests.java
```diff
@@ -0,0 +1,219 @@
...
+ * Super class for all RealmCollection related tests.
+ * This class only contains configuration and helper methods.
+ */
+public abstract class CollectionTests {
+
+    protected final static long YEAR_MILLIS = TimeUnit.DAYS.toMillis(365);
+
+    // Enumerate all known collection classes from the Realm API.
+    protected enum CollectionClass {
+        MANAGED_REALMLIST, UNMANAGED_REALMLIST, REALMRESULTS;
```
beeender: Remove the semicolons here and below to solve the lint warnings. Also, AS thinks `UNMANAGED_REALMLIST` is a typo which makes me quite uncomfortable :mask:

realm/realm-library/src/main/java/io/realm/BaseRealm.java
cmelchior: It is an unfortunate side-effect of having to make `getRealm` part of the interface, which is needed for weak change listeners in RealmBaseAdapter :(

cmelchior: If you have any ideas for improving it I am all ears though. Exposing the configuration so adapters can open their own instance is not really possible either as there is no lifecycle callback on an Adapter when the view is no longer attached :(

beeender: Yeah... i am thinking hard... making it public make this https://github.com/realm/realm-java/issues/1594#issuecomment-194175373 be a problem .... although i was wrong before (`getRealm` won't be a problem if `BaseRealm` is not public).

beeender: Maybe we do the same as `https://github.com/realm/realm-java/issues/2401#issuecomment-194192660` ? Let `RealmBaseAdapter` take a param of <E extends Realm>? It is kind of reminding user that the life cycle of Realm should be taken care out of the adapter.

cmelchior: It feels a bit hacky though as that generic parameter should ideally be on RealmResults .... we could maybe do that though.

realm/realm-library/src/androidTest/java/io/realm/CollectionTests.java
cmelchior: There is a quick fix to add them to your dictionary. I would much prefer `REALMLIST` instead of `REALM_LIST`

CHANGELOG.md
```diff
@@ -1,3 +1,12 @@
+## 0.89.0
```
kneth: Please change to the format/sections.

realm/realm-jni/build.gradle
```diff
@@ -71,12 +71,12 @@ def toolchains = [
 ]
 
 def targets = [
-    new Target( name:'arm', jniFolder:'armeabi', toolchain:toolchains.find{it.name == 'arm'}, cflags:[ '-mthumb' ] ),
-    new Target( name:'arm-v7a', jniFolder:'armeabi-v7a', toolchain:toolchains.find{it.name == 'arm'}, cflags:[ '-mthumb', '-march=armv7-a', '-mfloat-abi=softfp', '-mfpu=vfpv3-d16' ] ),
-    new Target( name:'arm64', jniFolder:'arm64-v8a', toolchain:toolchains.find{it.name == 'arm64'}, cflags:[] ),
-    new Target( name:'mips', jniFolder:'mips', toolchain:toolchains.find{it.name == 'mips'}, cflags:[] ),
+//    new Target( name:'arm', jniFolder:'armeabi', toolchain:toolchains.find{it.name == 'arm'}, cflags:[ '-mthumb' ] ),
```
kneth: Please change this before merging.

cmelchior:
Bump for review

cmelchior:
retest this please

realm/realm-library/src/main/java/io/realm/RealmCollection.java
```diff
@@ -0,0 +1,153 @@
...
+ * limitations under the License.
+ */
+
+package io.realm;
+
+import java.util.Collection;
+import java.util.Date;
+
+import io.realm.RealmObject;
+import io.realm.RealmQuery;
```
beeender: Above two line, unused import.

realm/realm-library/src/main/java/io/realm/RealmCollection.java
```diff
@@ -0,0 +1,153 @@
...
+     * @throws IllegalStateException if the corresponding Realm is closed or in an incorrect thread.
+     * @return {@code true} if objects was deleted, {@code false} otherwise.
+     * @throws java.lang.IllegalStateException if the Realm has been closed or called from an incorrect thread.
+     */
+    boolean deleteAllFromRealm();
+
+    /**
+     * Checks if a collection has finished loading its data yet.
+     *
+     * @return {@code true} if data has loaded and is available {@code false} if they are still loading.
```
beeender: `has loaded` -> `has been loaded` `are still loading`-> `are still being loaded`

realm/realm-library/src/main/java/io/realm/OrderedRealmCollection.java
```diff
@@ -0,0 +1,124 @@
...
+ * index, with the first index being zero. Normally, {@code OrderedRealmCollection}s allow duplicate
+ * elements, as compared to Sets, where elements have to be unique.
+ */
+public interface OrderedRealmCollection<E extends RealmObject> extends List<E>, RealmCollection<E> {
+
+    /**
+     * Gets the first object from the collection.
+     *
+     * @return the first object or {@code null} if the collection is empty.
+     * @throws IndexOutOfBoundsException if the collection is empty.
```
beeender: > or {@code null} if the collection is empty
> throws IndexOutOfBoundsException if the collection is empty

These two lines are conflicted.

realm/realm-library/src/main/java/io/realm/OrderedRealmCollection.java
```diff
@@ -0,0 +1,124 @@
...
+    /**
+     * Gets the last object from the collection.
+     *
+     * @return the last object or {@code null} if the collection is empty.
+     * @throws IndexOutOfBoundsException if the collection is empty.
+     */
+    E last();
+
+    /**
+     * Sorts (ascending) a collection..
```
beeender: An extra `.` at the end. Also can make it more formal like `Sorts a collection in ascending order.`

realm/realm-library/src/main/java/io/realm/OrderedRealmCollection.java
```diff
@@ -0,0 +1,124 @@
...
+    /**
+     * Sorts existing collection in-place.
+     *
+     * @param fieldName the field name to sort by. Only fields of type boolean, short, int, long, float, double, Date,
+     *                  and String are supported.
+     * @param sortOrder the direction to sort by.
+     * @throws java.lang.IllegalArgumentException if field name does not exist or has an invalid type.
+     * @throws java.lang.IllegalStateException if the Realm is closed, called on the wrong thread or the collection is
+     *                                         an un-managed collection.
+     * @return A sorted copy of the collection.
```
beeender: `A sorted copy`?

realm/realm-library/src/main/java/io/realm/OrderedRealmCollection.java
```diff
@@ -0,0 +1,124 @@
...
+     */
+    RealmResults<E> sort(String fieldName, Sort sortOrder);
+
+    /**
+     * Sorts existing {@link io.realm.RealmResults} in-place using two fields.
+     *
+     * @param fieldName1 first field name.
```
beeender: missing `Only fields of type boolean, short, int, long, float, double, Date, and String are supported.` same for fieldName2

realm/realm-library/src/main/java/io/realm/OrderedRealmCollection.java
```diff
@@ -0,0 +1,124 @@
...
+
+    /**
+     * Deletes the object at the given index from the Realm. This also removes it from the collection.
+     *
+     * @param location the array index identifying the object to be removed.
+     * @throws IndexOutOfBoundsException if {@code location < 0 || location >= size()}.
+     * @throws java.lang.IllegalStateException if the Realm is closed, called on the wrong thread or the collection is
+     *                                         an un-managed collection.
+     */
+    void deleteFromRealm(int location);
```
beeender: Shouldn't this return `boolean` as well? since all the deletion methods return `boolean` right now.

realm/realm-library/src/main/java/io/realm/RealmList.java
```diff
@@ -18,10 +18,11 @@
...
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Collections;
+import java.util.Date;
 import java.util.List;
 
-import io.realm.exceptions.RealmException;
 import io.realm.internal.InvalidRow;
 import io.realm.internal.LinkView;
 import io.realm.internal.Table;
```
beeender: Remove ` import io.realm.internal.Table;`, useless now.

realm/realm-library/src/main/java/io/realm/RealmList.java
```diff
@@ -118,10 +119,20 @@ public boolean isValid() {
         if (!managedMode) {
             return false;
         }
-
+        if (realm == null || realm.isClosed()) {
```
beeender: `managedMode` is actually redundant, checking null for `realm` is enough. But i am fine to solve it in another PR.

realm/realm-library/src/main/java/io/realm/RealmList.java
```diff
@@ -323,22 +337,86 @@ public E remove(int location) {
     }
 
     /**
-     * Removes all elements from this list and delete them from the corresponding Realm. This method can be called on a
-     * non-managed {@link RealmList} if all of the RealmObjects in the list are managed by Realm.
+     * Removes one instance of the specified object from this {@code Collection} if one
+     * is contained . This implementation iterates over this
+     * {@code Collection} and tests for each element {@code e} returned by the iterator,
```
beeender: remove `{@code e}`

realm/realm-library/src/main/java/io/realm/RealmList.java
```diff
@@ -323,22 +337,86 @@ public E remove(int location) {
...
+     * @param object the object to remove.
+     * @return {@code true} if this {@code Collection} is modified, {@code false} otherwise.
+     * @throws ClassCastException if the object passed is not of the correct type.
+     * @throws NullPointerException  if {@code object} is {@code null}.
+     */
+    @Override
+    public boolean remove(Object object) {
+        if (managedMode && !realm.isInTransaction()) {
+            throw new IllegalStateException("Objects can only be removed from inside a write transaction");
+        }
```
beeender: create a private method to do throwing?

realm/realm-library/src/main/java/io/realm/RealmList.java
```diff
@@ -323,22 +337,86 @@ public E remove(int location) {
...
+     * remove} method is called on the iterator.
+     *
+     * @param collection the collection of objects to remove.
+     * @return {@code true} if this {@code Collection} is modified, {@code false} otherwise.
+     * @throws ClassCastException if one or more elements of {@code collection} isn't of the correct type.
+     * @throws NullPointerException if {@code collection} contains at least one {@code null} element.
+     * @throws NullPointerException if {@code collection} is {@code null}.
+     */
+    @Override
+    public boolean removeAll(Collection<?> collection) {
```
beeender: A test case to throw `ClassCastException`

realm/realm-library/src/main/java/io/realm/RealmList.java
```diff
@@ -323,22 +337,86 @@ public E remove(int location) {
...
+        if (managedMode && !realm.isInTransaction()) {
+            throw new IllegalStateException("Objects can only be removed from inside a write transaction");
+        }
+        return super.removeAll(collection);
+    }
+
+    /**
+     * {@inheritDoc}
```
beeender: Possible to add one more @throws UnsupportedOperationException to `inheritDoc`? Or just add throws to the parent's doc.

realm/realm-library/src/main/java/io/realm/RealmList.java
```diff
@@ -386,11 +466,66 @@ public E first() {
...
+     */
+    @Override
+    public void deleteFromRealm(int location) {
+        if (managedMode) {
+            checkValidView();
+            view.removeTargetRow(location);
+        } else {
+            throw new UnsupportedOperationException("deleteFromRealm() not supported by un-managed RealmLists.");
```
beeender: just use `ONLY_IN_MANAGED_MODE_MESSAGE` everywhere? and aboves

realm/realm-library/src/main/java/io/realm/RealmResults.java
```diff
@@ -213,33 +219,53 @@ public E get(int location) {
...
-     * @throws ArrayIndexOutOfBoundsException if RealmResults is empty.
+     * {@inheritDoc}
      */
-    public E last() {
-        return get(size()-1);
+    @Override
+    public void deleteFromRealm(int location) {
+        realm.checkIfValid();
+        TableOrView table = getTable();
+        table.remove(location);
```
beeender: `IndexOutOfBoundsException` can be thrown from JNI?

realm/realm-library/src/main/java/io/realm/internal/LinkView.java
```diff
@@ -153,6 +154,14 @@ public void removeAllTargetRows() {
         nativeRemoveAllTargetRows(nativePointer);
     }
 
+    /**
+     * Remove target row from both the Realm and the LinkView.
```
beeender: Removes

beeender:
Just went through the code (except test part), looks really nice! Will go through the test cases ASAP.

realm/realm-library/src/androidTest/java/io/realm/RealmListTests.java
```diff
@@ -137,11 +121,6 @@ public void isValid_nonManagedMode() {
         assertFalse(list.isValid());
     }
 
-    @Test (expected = RealmException.class)
```
beeender: Why remove this test?

realm/realm-library/src/androidTest/java/io/realm/RealmListTests.java
```diff
@@ -697,261 +634,133 @@ public void where_afterContainerObjectRemoved() {
...
-                = new RealmList<Dog>(managedDogs.toArray(new Dog[managedDogs.size()]));
-        Dog dog = managedDogs.get(0);
-        assertTrue("Should contain a particular dog", nonManagedDogs.contains(dog));
-    }
+    public void setList_clearsOldItems() {
+        realm.beginTransaction();
+        CyclicType one = realm.copyToRealm(new CyclicType());
+        CyclicType two = realm.copyToRealm(new CyclicType());
+        two.setObjects(new RealmList<CyclicType>(one));
+        two.setObjects(new RealmList<CyclicType>(one));
```
beeender: This test is not good enough. `two.setObjects(new RealmList<CyclicType>(one)); ` twice could work even when there are bugs. maybe set a size=1 list first then size=2 list

realm/realm-library/src/androidTest/java/io/realm/RealmListTests.java
```diff
@@ -39,97 +36,84 @@
...
-
     /*********************************************************
-     * Non-Managed mode tests                                *
+     * Un-managed mode tests                                *
      *********************************************************/
 
+    @Test(expected = IllegalArgumentException.class)
+    public void constructor_nonManaged_null() {
+        AllTypes[] args = null;
+        new RealmList<AllTypes>(args);
```
beeender: just `new RealmList<AllTypes>(null);` to fix the lint warning.

realm/realm-library/src/androidTest/java/io/realm/RealmListTests.java
```diff
@@ -697,261 +634,133 @@ public void where_afterContainerObjectRemoved() {
...
-            assertTrue(method.toString(), methodWrongThread(method));
         }
     }
 
     @Test
     public void setList_ClearsOldItems() {
-        testRealm.beginTransaction();
-        CyclicType one = testRealm.copyToRealm(new CyclicType());
-        CyclicType two = testRealm.copyToRealm(new CyclicType());
+        realm.beginTransaction();
```
beeender: This is exact the same test with line 683.

realm/realm-library/src/androidTest/java/io/realm/RealmListTests.java
```diff
@@ -962,58 +771,31 @@ public void removeAllFromRealm_outsideTransaction() {
     }
 
     @Test
-    public void removeAllFromRealm_listWithStandaloneObjectShouldThrow() {
```
beeender: I cannot find this test case anymore, is it deleted?

realm/realm-library/src/androidTest/java/io/realm/RealmListTests.java
```diff
@@ -697,261 +634,133 @@ public void where_afterContainerObjectRemoved() {
...
+                    case IS_VALID: continue; // Does not throw
+                }
+                fail(method + " should have thrown an Exception.");
+            } catch (IllegalStateException ignored) {
             }
         }
-    }
-
-    @Test
-    public void contains_shouldNotContainDeletedRealmObject() {
```
beeender: Is this deleted?

realm/realm-library/src/androidTest/java/io/realm/RealmListTests.java
```diff
@@ -697,261 +634,133 @@ public void where_afterContainerObjectRemoved() {
...
-        assertFalse("Should not contain a null item.", owner.getDogs().contains(null));
+        assertEquals(1, two.getObjects().size());
     }
 
-    /**
-     * Test that the {@link Realm#contains(Class)} method of one Realm will not contain a
-     * {@link RealmObject} from another Realm.
-     */
     @Test
-    public void contains_doesNotContainAnItem() {
```
beeender: Is this deleted?

realm/realm-library/src/androidTest/java/io/realm/RealmListTests.java
```diff
@@ -697,261 +634,133 @@ public void where_afterContainerObjectRemoved() {
...
+        assertEquals(TEST_SIZE, realm.allObjects(Dog.class).size());
+        realm.commitTransaction();
     }
 
-    /**
-     * Test to see if a particular item that does exist in the same Realm does not
-     * exist in a query that excludes said item.
-     */
     @Test
-    public void contains_sameRealmNotContained() {
```
beeender: This test case is missing

realm/realm-library/src/androidTest/java/io/realm/RealmListTests.java
```diff
@@ -639,18 +568,26 @@ public void remove_atAfterContainerObjectRemoved() {
     public void remove_bbjectAfterContainerObjectRemoved() {
```
beeender: typo `bbject`

realm/realm-library/src/main/java/io/realm/OrderedRealmCollection.java
cmelchior: I was thinking the same, but there is the same difference between `remove(Object)` and `remove(int)`. This pattern fits better the current behaviour in the generic Collection API.

beeender:
Suppress lint warings in `RealmResultsTests.unsupportedMethods()`

beeender:
I cannot remember if we have this discussion before, do we need

```java
boolean RealmCollection.deleteFromRealm(E object);
```

realm/realm-library/src/androidTest/java/io/realm/ManagedOrderedRealmCollectionTests.java
```diff
@@ -0,0 +1,730 @@
...
+        collection.deleteFirstFromRealm();
+        realm.commitTransaction();
+        assertEquals(0, collection.size());
+    }
+
+    @Test
+    public void deleteLastFromRealm() {
+        assertEquals(TEST_SIZE - 1, collection.last().getFieldLong());
+        realm.beginTransaction();
+        collection.deleteLastFromRealm();
```
beeender: assertTrue(collection.deleteLastFromRealm());

realm/realm-library/src/androidTest/java/io/realm/ManagedOrderedRealmCollectionTests.java
```diff
@@ -0,0 +1,730 @@
...
+        realm.commitTransaction();
+        assertEquals(TEST_SIZE - 1, collection.size());
+        assertEquals(TEST_SIZE - 2, collection.last().getFieldLong());
+    }
+
+    @Test
+    public void deleteLastFromRealm_emptyCollection() {
+        OrderedRealmCollection<NullTypes> collection = createEmptyCollection(realm, collectionClass);
+        realm.beginTransaction();
+        collection.deleteLastFromRealm();
```
beeender: assertFalse(collection.deleteLastFromRealm());

realm/realm-library/src/androidTest/java/io/realm/ManagedOrderedRealmCollectionTests.java
```diff
@@ -0,0 +1,730 @@
...
+                throw new AssertionError("Unknown collection class: " + collectionClass);
+        }
+
+    }
+
+    @Test
+    public void deleteFirstFromRealm_emptyCollection() {
+        OrderedRealmCollection<NullTypes> collection = createEmptyCollection(realm, collectionClass);
+        realm.beginTransaction();
+        collection.deleteFirstFromRealm();
```
beeender: assertFalse(collection.deleteFirstFromRealm());

realm/realm-library/src/androidTest/java/io/realm/ManagedOrderedRealmCollectionTests.java
```diff
@@ -0,0 +1,730 @@
...
+        sortedList = sortedList.sort(AllJavaTypes.FIELD_LONG, Sort.DESCENDING);
+        assertEquals(TEST_SIZE, sortedList.size());
+    }
+
+    @Test
+    public void deleteFromRealm() {
+        OrderedRealmCollection<Dog> collection = createNonCyclicCollection(realm, collectionClass);
+        assertEquals(1, collection.get(1).getAge());
+        realm.beginTransaction();
+        collection.deleteFromRealm(0);
```
beeender: assertTrue(collection.deleteFromRealm());

realm/realm-library/src/androidTest/java/io/realm/ManagedOrderedRealmCollectionTests.java
```diff
@@ -0,0 +1,730 @@
...
+                    case DELETE_INDEX: collection.deleteFromRealm(0); break;
+                    case DELETE_FIRST: collection.deleteFirstFromRealm(); break;
+                    case DELETE_LAST: collection.deleteLastFromRealm(); break;
+                    case ADD_INDEX: collection.add(0, new AllJavaTypes()); break;
+                    case ADD_ALL_INDEX: collection.addAll(0, Arrays.asList(new AllJavaTypes())); break;
+                    case SET: collection.set(0, new AllJavaTypes()); break;
+                    case REMOVE_INDEX: collection.remove(0); break;
+                }
+                fail("Unknown method or it failed to throw: " + method);
+            } catch (IllegalStateException ignored) {
```
beeender: better separate `IllegalStateException` and `unsupportedOperationException`. And aboves are the same ...

realm/realm-library/src/main/java/io/realm/BaseRealm.java
beeender: :+1:

beeender:
Finally done!! next time we should keep the old test cases first, then remove them in another PR which will make review much faster :wine_glass:

realm/realm-library/src/main/java/io/realm/RealmList.java
cmelchior: As long as we don't support inheritance, it should be impossible to call this method with the wrong types (since `RealmObject` is an abstract class) ?

realm/realm-library/src/main/java/io/realm/RealmList.java
cmelchior: Added to parent docs that un-managed collections throw UnsupportedOperations.

realm/realm-library/src/main/java/io/realm/RealmList.java
beeender: `Collection<?>` means `dogs.removeAll(Collections.<Cat>emptyList());` can be compiled, right?

cmelchior: Ups, you right of course. I'll add tests

cmelchior: It seems that throwing that exception is optional and ArrayList actually doesn't, but just return false.  IMO that behaviour is a bit more sane, so is it okay if I just override `contains` in RealmCollection and specify it returns false instead? Still needs a test for it though.

realm/realm-library/src/main/java/io/realm/RealmResults.java
cmelchior: Yes

realm/realm-library/src/androidTest/java/io/realm/RealmListTests.java
cmelchior: Changed

realm/realm-library/src/androidTest/java/io/realm/RealmListTests.java
cmelchior: Actually doing that will cause javac to complain. That was what lead to this. I'll suppress the warning in IntelliJ instead.

realm/realm-library/src/androidTest/java/io/realm/RealmListTests.java
cmelchior: Nice catch. Removed

realm/realm-library/src/androidTest/java/io/realm/RealmListTests.java
cmelchior: Yes, since we now split into `remove` and `delete` methods. This behaviour have now changed making the test no longer needed.

`deleteAllFromRealm` will throw UnsupportedOperation on the un-managed list.
`removeAll` doesn't care about managed/unmanaged, it just checks object equality.

realm/realm-library/src/androidTest/java/io/realm/RealmListTests.java
cmelchior: Yes, this test didn't make much sense, since it more about how RealmQuery works than the Collection

realm/realm-library/src/androidTest/java/io/realm/ManagedOrderedRealmCollectionTests.java
cmelchior: This returns void or throws an exception

realm/realm-library/src/androidTest/java/io/realm/ManagedOrderedRealmCollectionTests.java
cmelchior: These unit tests are haunting me a bit as well, but the above is not a bad idea. I'll see how easy the can be refactored.

cmelchior:

> boolean RealmCollection.deleteFromRealm(E object);

@beeender Regarding above. I considered it, but it would be a lot easier to just do `object.deleteFromRealm()`, so I kinda left it out for now. Do you think it will be used?

beeender:
@cmelchior Yeah, never mind. `object.deleteFromRealm()` would be good enough.

realm/realm-library/src/main/java/io/realm/OrderedRealmCollection.java
```diff
@@ -47,10 +47,10 @@
      *
      * @param fieldName the field name to sort by. Only fields of type boolean, short, int, long, float, double, Date,
      *                  and String are supported.
+     * @return A sorted view of the collection.
```
beeender: `A` -> `a`

realm/realm-library/src/main/java/io/realm/OrderedRealmCollection.java
```diff
@@ -60,10 +60,10 @@
      * @param fieldName the field name to sort by. Only fields of type boolean, short, int, long, float, double, Date,
      *                  and String are supported.
      * @param sortOrder the direction to sort by.
+     * @return A sorted view of the collection.
```
beeender: `A` -> `a`

beeender: `A` -> `a`

realm/realm-library/src/main/java/io/realm/RealmCollection.java
```diff
@@ -0,0 +1,162 @@
...
+
+    /**
+     * Tests whether this {@code Collection} contains the specified object. Returns
+     * {@code true} if and only if at least one element {@code elem} in this
+     * {@code Collection} meets following requirement:
+     * {@code (object==null ? elem==null : object.equals(elem))}.
+     *
+     * @param object the object to search for.
+     * @return {@code true} if object is an element of this {@code Collection}, {@code false} otherwise.
+     * @throws NullPointerException if the object to look for is {@code null} and this {@code Collection} doesn't
```
beeender: Maybe `IllegalStateException` is more accurate than the NPE here?

realm/realm-library/src/main/java/io/realm/OrderedRealmCollection.java
```diff
@@ -0,0 +1,126 @@
...
+     * @throws IndexOutOfBoundsException if the collection is empty.
+     */
+    E last();
+
+    /**
+     * Sorts a collection based on the provided field in ascending order.
+     *
+     * @param fieldName the field name to sort by. Only fields of type boolean, short, int, long, float, double, Date,
+     *                  and String are supported.
+     * @return A sorted view of the collection.
```
beeender: `A` -> `a`

realm/realm-library/src/main/java/io/realm/OrderedRealmCollection.java
```diff
@@ -0,0 +1,126 @@
...
+     */
+    RealmResults<E> sort(String fieldName);
+
+    /**
+     * Sorts a collection based on the provided field and sort order.
+     *
+     * @param fieldName the field name to sort by. Only fields of type boolean, short, int, long, float, double, Date,
+     *                  and String are supported.
+     * @param sortOrder the direction to sort by.
+     * @return A sorted view of the collection.
```
beeender: `A` -> `a`

realm/realm-library/src/androidTest/java/io/realm/ManagedOrderedRealmCollectionTests.java
```diff
@@ -0,0 +1,749 @@
...
+                        case DELETE_INDEX: collection.deleteFromRealm(0); break;
+                        case DELETE_FIRST: collection.deleteFirstFromRealm(); break;
+                        case DELETE_LAST: collection.deleteLastFromRealm(); break;
+                        case SORT: collection.sort(AllJavaTypes.FIELD_STRING); break;
+                        case SORT_FIELD: collection.sort(AllJavaTypes.FIELD_STRING, Sort.ASCENDING); break;
+                        case SORT_2FIELDS: collection.sort(AllJavaTypes.FIELD_STRING, Sort.ASCENDING, AllJavaTypes.FIELD_LONG, Sort.DESCENDING); break;
+                        case SORT_MULTI: collection.sort(new String[] { AllJavaTypes.FIELD_STRING }, new Sort[] { Sort.ASCENDING }); break;
+                    }
+                    return false;
+                } catch (IllegalStateException ignored) {
```
beeender: Check the exception type like `mutableMethodsOutsideTransactions`

realm/realm-library/src/androidTest/java/io/realm/ManagedOrderedRealmCollectionTests.java
```diff
@@ -0,0 +1,749 @@
...
+                        case INDEX_OF: collection.indexOf(new AllJavaTypes()); break;
+                        case LAST_INDEX_OF: collection.lastIndexOf(new AllJavaTypes()); break;
+                        case LIST_ITERATOR: collection.listIterator(); break;
+                        case LIST_ITERATOR_INDEX: collection.listIterator(0); break;
+                        case REMOVE_INDEX: collection.remove(0); break;
+                        case SET: collection.set(0, new AllJavaTypes()); break;
+                        case SUBLIST: collection.subList(0, 1); break;
+                    }
+                    return false;
+                } catch (IllegalStateException ignored) {
```
beeender: Check the exception type like `mutableMethodsOutsideTransactions`

realm/realm-library/src/androidTest/java/io/realm/ManagedRealmCollectionTests.java
```diff
@@ -0,0 +1,759 @@
...
+                    case DELETE_ALL: collection.deleteAllFromRealm(); break;
+                    case ADD_OBJECT: collection.add(new AllJavaTypes());
+                    case ADD_ALL_OBJECTS: collection.addAll(Arrays.asList(new AllJavaTypes())); break;
+                    case CLEAR: collection.clear(); break;
+                    case REMOVE_OBJECT: collection.remove(new AllJavaTypes()); break;
+                    case REMOVE_ALL: collection.removeAll(Arrays.asList(new AllJavaTypes())); break;
+                    case RETAIN_ALL: collection.retainAll(Arrays.asList(new AllJavaTypes())); break;
+                }
+                fail("Unknown method or it failed to throw: " + method);
+            } catch (IllegalStateException ignored) {
```
beeender: Check the exception type like `mutableMethodsOutsideTransactions`

realm/realm-library/src/androidTest/java/io/realm/RealmListTests.java
beeender: :+1:

realm/realm-library/src/main/java/io/realm/RealmList.java
beeender: :+1:

realm/realm-library/src/androidTest/java/io/realm/RealmListTests.java
beeender: :+1:

beeender:
Except some minor comments on the test cases, :+1:

CHANGELOG.md
```diff
@@ -1,7 +1,19 @@
...
+### Breaking changes
 
+* RealmResults.clear() now throws UnsupportedOperationException. Use RealmResults.deleteAllFromRealm() instead.
+* RealmResults.remove(int) now throws UnsupportedOperationException. Use RealmResults.deleteFromRealm() instead.
+
+### Deprecated
+
+* RealmObject.removeFromRealm() in place of RealmObject.deleteFromRealm()
+* Realm.clear(Class) in place of Realm.delete(Class).
+* DynamicRealm.clear(Class) in place of DynamicRealm.delete(Class).
```
nhachicha: > in place of

use instead ...

realm/realm-library/src/main/java/io/realm/OrderedRealmCollection.java
```diff
@@ -0,0 +1,126 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.realm;
+
+import java.util.List;
+
+/**
+ * A {@code OrderedRealmCollection} is a collection which maintains an ordering for its elements. Every
```
nhachicha: A -> An

realm/realm-library/src/main/java/io/realm/OrderedRealmCollection.java
```diff
@@ -0,0 +1,126 @@
...
+     * Sorts a collection based on the provided field in ascending order.
+     *
+     * @param fieldName the field name to sort by. Only fields of type boolean, short, int, long, float, double, Date,
+     *                  and String are supported.
+     * @return A sorted view of the collection.
+     * @throws java.lang.IllegalArgumentException if field name does not exist or it has an invalid type.
+     * @throws java.lang.IllegalStateException if the Realm is closed, called on the wrong thread or the collection is
+     *                                         an un-managed collection.
+     */
+    RealmResults<E> sort(String fieldName);
```
nhachicha: The sort is not inplace (returning a copy I believe) so why not returning the super type `OrderedRealmCollection<E>`? leaking abstraction?

realm/realm-library/src/main/java/io/realm/RealmList.java
```diff
@@ -150,6 +158,9 @@ public void add(int location, E object) {
         checkValidObject(object);
         if (managedMode) {
             checkValidView();
+            if (location < 0 || location > size()) {
```
nhachicha: aren't those check already done in JNI?

realm/realm-library/src/main/java/io/realm/RealmResults.java
```diff
@@ -213,33 +219,53 @@ public E get(int location) {
...
-     *
-     * @return the last object.
-     * @throws ArrayIndexOutOfBoundsException if RealmResults is empty.
+     * {@inheritDoc}
      */
-    public E last() {
-        return get(size()-1);
+    @Override
+    public void deleteFromRealm(int location) {
+        realm.checkIfValid();
```
nhachicha: we should check for `realm.isInTransaction()`  like in `RealmList#delete*` same for other modifications methods

realm/realm-library/src/main/java/io/realm/RealmList.java
```diff
@@ -386,11 +462,66 @@ public E first() {
...
+        } else {
+            throw new UnsupportedOperationException(ONLY_IN_MANAGED_MODE_MESSAGE);
+        }
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void deleteFromRealm(int location) {
```
nhachicha: we should check we're in write transaction

realm/realm-library/src/main/java/io/realm/RealmList.java
```diff
@@ -422,11 +553,117 @@ public int size() {
...
+            throw new UnsupportedOperationException(ONLY_IN_MANAGED_MODE_MESSAGE);
         }
     }
 
     /**
+     * {@inheritDoc}
+     */
+    @Override
+    public boolean deleteAllFromRealm() {
+        if (managedMode) {
```
nhachicha: check for write transaction

realm/realm-library/src/main/java/io/realm/RealmCollection.java
```diff
@@ -0,0 +1,162 @@
...
+
+package io.realm;
+
+import java.util.Collection;
+import java.util.Date;
+
+/**
+ * {@code RealmCollection} is the root of the collection hierarchy that Realm supports. It defines operations on data
+ * collections and the behavior that they will have in all implementations of {@code RealmCollection}s.
+ *
```
nhachicha: [Collections.unmodifiableCollection](http://developer.android.com/reference/java/util/Collections.html#unmodifiableCollection(java.util.Collection<? extends E>)) will not guarantee an immutable collection (since the underlying `RealmResults`/`RealmList` could change/advance_read) , should we warn the user about it here?

realm/realm-library/src/main/java/io/realm/BaseRealm.java
nhachicha: Tests & adapterExample are working without `BaseRealm` being `public`

realm/realm-library/build.gradle
```diff
@@ -54,6 +54,7 @@ dependencies {
     androidTestCompile 'com.android.support.test:rules:0.4.1'
     androidTestCompile 'com.google.dexmaker:dexmaker:1.2'
     androidTestCompile 'com.google.dexmaker:dexmaker-mockito:1.2'
+    androidTestCompile 'junit:junit:4.12'
```
nhachicha: not needed, `com.android.support:support-annotations:23.0.1` should pull it
```
androidTestCompile - Classpath for compiling the androidTest sources.
+--- io.reactivex:rxjava:1.1.0
+--- com.android.support:support-annotations:23.1.1
+--- com.android.support.test:runner:0.4.1
|    +--- com.android.support.test:exposed-instrumentation-api-publish:0.4.1
|    +--- com.android.support:support-annotations:23.0.1 -> 23.1.1
|    \--- junit:junit:4.12
|         \--- org.hamcrest:hamcrest-core:1.3
+--- com.android.support.test:rules:0.4.1
|    \--- com.android.support.test:runner:0.4.1 (*)
+--- com.google.dexmaker:dexmaker:1.2
\--- com.google.dexmaker:dexmaker-mockito:1.2
     +--- com.google.dexmaker:dexmaker:1.2
     \--- org.mockito:mockito-core:1.9.5
          +--- org.hamcrest:hamcrest-core:1.1 -> 1.3
          \--- org.objenesis:objenesis:1.0
```

realm/realm-library/src/androidTest/java/io/realm/ManagedOrderedRealmCollectionTests.java
```diff
@@ -0,0 +1,749 @@
...
+ * @see RealmCollectionTests
+ * @see ManagedRealmCollectionTests
+ * @see UnManagedRealmCollectionTests
+ */
+
+@RunWith(Parameterized.class)
+public class ManagedOrderedRealmCollectionTests extends CollectionTests {
+
+    private static final int TEST_SIZE = 10;
+    private final static int TEST_DATA_FIRST_HALF = (int) (2 * (TEST_SIZE / 4.0D) - 1);
```
nhachicha: why / 4.0? is this instead first_quarter & last_quarter of data ?

realm/realm-library/src/androidTest/java/io/realm/ManagedOrderedRealmCollectionTests.java
```diff
@@ -0,0 +1,749 @@
...
+        assertEquals(resultList.first().getFieldString(), sortedList.last().getFieldString());
+
+        RealmResults<AllJavaTypes> reverseList = sortedList.sort(AllJavaTypes.FIELD_STRING, Sort.ASCENDING);
+        assertEquals(TEST_SIZE, reverseList.size());
+        assertEquals(resultList.first().getFieldString(), reverseList.first().getFieldString());
+
+        int numberOfDigits = 1 + ((int) Math.log10(TEST_SIZE));
+        int largestNumber = 1;
+        for (int i = 1; i < numberOfDigits; i++)
+            largestNumber *= 10;  // 10*10* ... *10
```
nhachicha: can't you use [Math.pow](http://developer.android.com/reference/java/lang/Math.html#pow(double, double)) ?

nhachicha:
Amazing work with the test coverage :+1: 
Questions:

1- Why can't we make `findAll*` returns `OrderedRealmCollection`
```java
OrderedRealmCollection<TimeStamp> timeStamps = realm.where(TimeStamp.class).findAll();
```
2- You should update the examples as well.
`adapterExample` is throwing an exception (when deleting elements)
```
E/AndroidRuntime( 6137): Process: io.realm.examples.realmadapters, PID: 6137
E/AndroidRuntime( 6137): java.lang.UnsupportedOperationException: This method is not supported by RealmResults.
E/AndroidRuntime( 6137): 	at io.realm.RealmResults.clear(RealmResults.java:681)
E/AndroidRuntime( 6137): 	at io.realm.examples.realmadapters.WorkerHandler.handleMessage(WorkerHandler.java:54)
E/AndroidRuntime( 6137): 	at android.os.Handler.dispatchMessage(Handler.java:102)
E/AndroidRuntime( 6137): 	at android.os.Looper.loop(Looper.java:135)
E/AndroidRuntime( 6137): 	at io.realm.examples.realmadapters.WorkerThread.run(WorkerThread.java:34)
```
in `WorkerHandler` you should replace 
```java
realm.where(TimeStamp.class).equalTo("timeStamp", timestamp).findAll().clear();
```
by 
```java 
realm.where(TimeStamp.class).equalTo("timeStamp", timestamp).findAll().deleteAllFromRealm();
```
in `AdapterExampleActivity`
we can already use `OrderedRealmCollection`

```java
OrderedRealmCollection<TimeStamp> timeStamps = realm.where(TimeStamp.class).findAll();
```

Also in `MyAdapter`, use 
```java 
public MyAdapter(Context context, int resId, OrderedRealmCollection<TimeStamp> realmResults, boolean automaticUpdate) {
```

3- What about async methods (sortAsync etc.) do we plan to add them?

realm/realm-library/src/main/java/io/realm/RealmCollection.java
cmelchior: This is the standard behaviour in the standard Collection interface, so I would rather keep this.

beeender: :+1:

realm/realm-library/src/main/java/io/realm/OrderedRealmCollection.java
cmelchior: I had the same thoughts, but the problem is that the underlying type _is_ RealmResults and RealmResults are quite restricted in terms of what you can do with it. So exposing RealmResults as the returned type has two benefits:

1) It is really clear that we return a view of the data and not some modified version of the original.
2) The methods not supported on the returned collection is well-defined.

realm/realm-library/src/main/java/io/realm/RealmList.java
cmelchior: Yes, I think so, but the code just below this will copy the object to Realm before checking the index, so if you are catching the IndexOutOfBounds, you suddenly have an orphaned object. By checking first we can avoid that.

realm/realm-library/src/main/java/io/realm/RealmResults.java
cmelchior: Ups, yes. Added the check to a lot of places in both RealmList and RealmResults

cmelchior: This is actually being done by TableView/Table through `isImmutable`. I modified the exception message to be more user friendly.

realm/realm-library/src/main/java/io/realm/RealmList.java
cmelchior: It is done through `checkImmutable` in Table/TableView

realm/realm-library/src/main/java/io/realm/RealmList.java
cmelchior: Done inside the Internal API

realm/realm-library/src/main/java/io/realm/RealmCollection.java
cmelchior: Yes, good idea. I'll add it.

cmelchior:
@nhachicha 

1) IMO Since RealmResults are our wrapper around any query results, I think keeping it as a return type is appropriate (it also preserves backwards compatibility). The only reason I can see for changing it to OrderedRealmCollection would be to have the flexibility to change the actual representation which seems very unlikely at this point? Also since RealmResults are general immutable, making the type explicit is better as it increase readability (as all mutator methods are marked as deprecated on RealmResults). What would your argument be for returning OrderedRealmCollection?

2) Ups, all examples fixed. Thanks

3) Currently no. Right now the `*Async` methods are only on RealmQuery (also in 0.88.x), and keeping them there seems like a good idea to me for two reasons:

a) It keeps the number of methods down while still providing some utility. IMO I think it is fair to think of async behaviour as something you decide "outside" the 
b) It is consistent with the Swift API

Happy to hear other arguments though.

cmelchior:
All comments addressed @beeender @nhachicha @realm/java

realm/realm-library/src/main/java/io/realm/BaseRealm.java
cmelchior: Yep, but since `getRealm` in `RealmCollection` now returns a BaseRealm, it unfortunately have to be public.

cmelchior: After talking to @beeender We decided we can pull BaseRealm from the public API after all. Only downside is adding a little bit of complexity to BaseRealmAdapter.

beeender:
:+1:

nhachicha:
:+1:

cmelchior:
@nhachicha :heart:

PR #2349, opened 2016-02-26
Add a check whether transformer was applied or not.

zaki50:
This commit adds default implementation of `transformerApplied()` to `RealmProxyMediator` class which returns `false`.
And then Realm transformer adds overriding methods to its subclasses and those methods returns `true`.

If Realm transformer was applied, `RealmProxyMediator.transformerApplied()` returns `true` and
if not, `RealmProxyMediator.transformerApplied()` returns `false`.
That enable us to detect Realm transformer was applied or not at runtime.

review this please @realm/java @emanuelez

### DIFF
--- a/realm/realm-library/src/main/java/io/realm/RealmConfiguration.java
+++ b/realm/realm-library/src/main/java/io/realm/RealmConfiguration.java
@@ -65,7 +65,13 @@ public class RealmConfiguration {
     static {
         DEFAULT_MODULE = Realm.getDefaultModule();
         if (DEFAULT_MODULE != null) {
-            DEFAULT_MODULE_MEDIATOR = getModuleMediator(DEFAULT_MODULE.getClass().getCanonicalName());
+            final RealmProxyMediator mediator = getModuleMediator(DEFAULT_MODULE.getClass().getCanonicalName());
+            if (!mediator.transformerApplied()) {
+                throw new ExceptionInInitializerError("RealmTransformer doesn't seem to be applied." +
+                        " Please update the project configuration to use the Realm Gradle plugin." +
+                        " See https://realm.io/news/android-installation-change/");
+            }
+            DEFAULT_MODULE_MEDIATOR = mediator;
         } else {
             DEFAULT_MODULE_MEDIATOR = null;
         }
--- a/realm/realm-library/src/main/java/io/realm/internal/RealmProxyMediator.java
+++ b/realm/realm-library/src/main/java/io/realm/internal/RealmProxyMediator.java
@@ -138,6 +138,17 @@ public abstract class RealmProxyMediator {
      */
     public abstract <E extends RealmObject> E createDetachedCopy(E realmObject, int maxDepth, Map<RealmObject, RealmObjectProxy.CacheData<RealmObject>> cache);
 
+    /**
+     * Returns whether Realm transformer has been applied or not. Subclasses of this class are
+     * created by the annotation processor and the Realm transformer will add an override of
+     * this method that always return {@code true} if the transform was successful.
+     *
+     * @return {@code true} if Realm transformer was applied, {@code false} otherwise.
+     */
+    public boolean transformerApplied() {
+        return false;
+    }
+
     @Override
     public boolean equals(Object o) {
         if (!(o instanceof RealmProxyMediator)) {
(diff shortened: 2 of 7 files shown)


realm/realm-library/src/main/java/io/realm/RealmConfiguration.java
```diff
@@ -65,7 +65,13 @@
     static {
         DEFAULT_MODULE = Realm.getDefaultModule();
         if (DEFAULT_MODULE != null) {
-            DEFAULT_MODULE_MEDIATOR = getModuleMediator(DEFAULT_MODULE.getClass().getCanonicalName());
+            final RealmProxyMediator mediator = getModuleMediator(DEFAULT_MODULE.getClass().getCanonicalName());
+            if (!mediator.transformerApplied()) {
+                throw new ExceptionInInitializerError("RealmTransformer doesn't seem to be applied." +
+                        " Please update the project configuration to use Realm gradle plugin." +
```
cmelchior: to use the Realm Gradle plugin

realm/realm-library/src/main/java/io/realm/internal/RealmProxyMediator.java
```diff
@@ -138,6 +138,17 @@
      */
     public abstract <E extends RealmObject> E createDetachedCopy(E realmObject, int maxDepth, Map<RealmObject, RealmObjectProxy.CacheData<RealmObject>> cache);
 
+    /**
+     * Returns whether Realm transformer has been applied or not. Realm transformer adds overriding method
```
cmelchior: 2nd sentance, perhaps this instead: "Subclasses of this class are created by the annotation processor and the Realm transformer will add an overirde of this method that always return {@code true} if the transform was successfull."

realm/realm-library/src/main/java/io/realm/internal/modules/CompositeMediator.java
```diff
@@ -115,6 +115,14 @@ public String getTableName(Class<? extends RealmObject> clazz) {
         return mediator.createDetachedCopy(realmObject, maxDepth, cache);
     }
 
+    @Override
+    public boolean transformerApplied() {
+        if (mediators.isEmpty()) {
```
cmelchior: Is this accurate? I mean, this will crash with a "Transform not running", but wouldn't the cause more likely be that the annotation processor didn't ran?

realm/realm-library/src/main/java/io/realm/internal/modules/CompositeMediator.java
```diff
@@ -115,6 +115,14 @@ public String getTableName(Class<? extends RealmObject> clazz) {
...
+            return super.transformerApplied();
+        }
+        return mediators.get(0).transformerApplied();
```
cmelchior: This should loop all mediators and only return true if all returned true

realm/realm-library/src/main/java/io/realm/internal/modules/FilterableMediator.java
```diff
@@ -131,6 +131,14 @@ public String getTableName(Class<? extends RealmObject> clazz) {
         return originalMediator.createDetachedCopy(realmObject, maxDepth, cache);
     }
 
+    @Override
+    public boolean transformerApplied() {
+        if (originalMediator == null) {
```
cmelchior: Same as the above. Will this not result in a wrong exception message?

cmelchior:
Really good idea :100:
I had a few comments to some of the checks, but otherwise it looks good

realm/realm-library/src/main/java/io/realm/RealmConfiguration.java
zaki50: done

realm/realm-library/src/main/java/io/realm/internal/RealmProxyMediator.java
zaki50: thanks. fixed.

realm/realm-library/src/main/java/io/realm/internal/modules/CompositeMediator.java
zaki50: Right, this is not accurate. But there is no way to detect whether transformed or not if `mediators` is empty.

Now we only checks the return value of `DefaultRealmModuleMediator` class in `RealmConfiguration`.
That means the implementation of `CompositeMediator` and `FilterableMediator`
does not affect the result of the check.

Is returning `true` better than current code?

cmelchior:
Looks good to me :+1:

stk1m1:
:+1: Hats off!

emanuelez:
:+1:

PR #2351, opened 2016-02-26
Better publishing

emanuelez:
Improve the publishing process by moving the logic to each respective project


emanuelez:
@realm/java

cmelchior:
Nice to get rid of all those files :+1:

PR #2352, opened 2016-02-26
Update README.md for OS X users.

zaki50:
@realm/java

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ Prerequisites:
     export NDK_HOME=/usr/local/Cellar/android-ndk/r10e
     ```
 
- * If you are using OS X, you'd be better to add following lines to `~/.profile` in order for Android Studio to see those environment variables.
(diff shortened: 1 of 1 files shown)


stk1m1:
:+1:

README.md
```diff
@@ -87,6 +87,12 @@ Prerequisites:
    export REALM_CORE_DOWNLOAD_DIR=~/.realmCore
    ```
 
+   OS X users should also add following line to `~/.profile` in order for Android Studio to see this environment variable..
```
cmelchior: I think this is only true if you use the default terminal right? I think @emanuelez and @beeender have been complaining about it not working.

zaki50: right. `.zprofile` instead of `.profile` must be used for `zsh`.

zaki50: @cmelchior updated

dalinaum:
:+1:

emanuelez:
:+1:

PR #2355, opened 2016-02-26
removed runtime dependency from 'realm-transformer' to 'com.android.tools.build:transform-api:1.5.0'.

### LINKED ISSUE #2348: [0.88-SNAPSHOT] RealmTransformer should support both 1.5 and 2.0.0-beta
Using latest snapshot i get following error during gradle sync:
`Error:Unable to load class 'io.realm.transformer.RealmTransformer'.`

	buildscript {
		repositories {
			jcenter()
			maven {
				url 'http://oss.jfrog.org/artifactory/oss-snapshot-local'
			}
			maven {
				url "https://jitpack.io"
			}
		}
		dependencies {
			classpath 'io.realm:realm-gradle-plugin:0.88.0-SNAPSHOT'
		}
	}

	apply plugin: 'realm-android'

zaki50:
We don't need to have runtime dependency to  `com.android.tools.build:transform-api` since android gradle plugin has that dependency.

Removing this explicit dependency from `realm-transformer` solves the compatibility issue against android gradle plugin 2.0.0-beta.

This fixes #2348

### DIFF
--- a/gradle-plugin/src/main/groovy/io/realm/gradle/Realm.groovy
+++ b/gradle-plugin/src/main/groovy/io/realm/gradle/Realm.groovy
@@ -35,6 +35,10 @@ class Realm implements Plugin<Project> {
             throw new GradleException("'com.android.application' or 'com.android.library' plugin required.")
         }
 
+        if (!isTransformAvailable()) {
+            throw new GradleException('Realm gradle plugin only supports android gradle plugin 1.5.0 or later.')
+        }
+
         def isKotlinProject = project.plugins.find {
             it.getClass().name == 'org.jetbrains.kotlin.gradle.plugin.KotlinAndroidPluginWrapper'
         }
@@ -54,6 +58,15 @@ class Realm implements Plugin<Project> {
         } else {
             project.dependencies.add("apt", "io.realm:realm-annotations:${Version.VERSION}")
             project.dependencies.add("apt", "io.realm:realm-annotations-processor:${Version.VERSION}")
+        }
+    }
+
+    private static boolean isTransformAvailable() {
+        try {
+            Class.forName('com.android.build.api.transform.Transform')
+            return true
+        } catch (Exception ignored) {
+            return false
         }
     }
 }
--- a/realm-transformer/build.gradle
+++ b/realm-transformer/build.gradle
@@ -30,7 +30,6 @@ dependencies {
     compile gradleApi()
     compile "io.realm:realm-annotations:${version}"
     compile 'com.android.tools.build:gradle:1.5.0'
-    compile 'com.android.tools.build:transform-api:1.5.0'
     compile 'org.javassist:javassist:3.20.0-GA'
 
     testCompile('org.spockframework:spock-core:1.0-groovy-2.4') {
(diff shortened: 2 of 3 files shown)


realm-transformer/build.gradle
```diff
@@ -25,12 +25,23 @@ repositories {
...
+    }
+}
+
 dependencies {
     compile localGroovy()
     compile gradleApi()
     compile "io.realm:realm-annotations:${version}"
     compile 'com.android.tools.build:gradle:1.5.0'
-    compile 'com.android.tools.build:transform-api:1.5.0'
+    provided 'com.android.tools.build:transform-api:1.5.0'
```
cmelchior: Isn't the transform API a dependency for the build tools already? Shouldn't we be able to just remove this? This would also mean it should be the build tools that should be a provided dependency?

zaki50: Right! I just removed this dependency.

kneth:
Nice (re)move.

beeender:
:+1:

realm-transformer/build.gradle
```diff
@@ -30,7 +30,6 @@ dependencies {
     compile gradleApi()
     compile "io.realm:realm-annotations:${version}"
     compile 'com.android.tools.build:gradle:1.5.0'
```
cmelchior: Should this be provided instead? 

Build tools version might matter quite a lot to people so if we force some minimum version onto them, other subtle things might break.

I would probably be more comfortable with A) having this as provided B) Document that 1.5 is the minimum version we support and C) Just let user projects crash if they use below 1.5

What do you think?

zaki50: good point. I'll try to implement it.

zaki50:
I also removed a dependency to 'com.android.tools.build:gradle:1.5.0' and add a version check of android gradle plugin instead.

please revirew this @realm/java

realm-transformer/build.gradle
zaki50: done.

zaki50:
I got following error when I use android gradle plugin 1.4.0-beta6 which is the last version of 1.4.0-beta.

```
* What went wrong:
A problem occurred evaluating project ':app'.
> Failed to apply plugin [id 'realm-android']
   > Realm gradle plugin only supports android gradle plugin 1.5.0 or later.

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output.

BUILD FAILED
```

gradle-plugin/src/main/groovy/io/realm/gradle/Realm.groovy
```diff
@@ -35,6 +35,10 @@ class Realm implements Plugin<Project> {
             throw new GradleException("'com.android.application' or 'com.android.library' plugin required.")
         }
 
+        if (!isTransformAvailable()) {
+            throw new GradleException('Realm gradle plugin only supports android gradle plugin 1.5.0 or later.')
```
cmelchior: Great idea

cmelchior:
:+1:

emanuelez:
Very nice! :+1:

PR #2358, opened 2016-02-27
Copy resource files when transforming byte code.

### LINKED ISSUE #2356: Kotlin: Package level functions cannot be imported from library project with new gradle plugin
This is a pretty elaborate case so hopefully I can explain it well enough.

We have a project consisting of a main application and a library project.  The library project contains Realm models that are shared.  Using the new plugin on both the main application and the library project causes the main project to not see the package level functions in the library.

I've attached a small sample project to show what I mean.

Given this file in the library project 

~~~kotlin
package io.vokal.realmtest.model

import io.realm.*
import io.realm.annotations.*

@RealmClass
open class Dog(
    @PrimaryKey
    public open var name: String = "spot"
): RealmObject() { }

@RealmModule(library = true, allClasses = true)
public class LibraryModule {
}

public fun Realm.findSpot(): Dog {
    return where(Dog::class.java).equalTo("name", "spot").findFirst()
}
~~~

I should be able to import `findSpot` in the main activity.  

Instead I receive `e: /home/wmbest2/realm-library-test/app/src/main/kotlin/io/vokal/realmtest/MainActivity.kt: (37, 23): Unresolved reference: findSpot`

[realm-library-test.zip](https://github.com/realm/realm-java/files/149012/realm-library-test.zip)

zaki50:
Realm Transformer must copy non-class files to output dir since Kotlin compiler generates `*.kotlin_module` file in `META-INF`.

This fixes #2356

### DIFF
--- a/realm-transformer/src/main/groovy/io/realm/transformer/RealmTransformer.groovy
+++ b/realm-transformer/src/main/groovy/io/realm/transformer/RealmTransformer.groovy
@@ -19,6 +19,7 @@ import com.android.SdkConstants
 import com.android.build.api.transform.*
 import com.google.common.collect.ImmutableSet
 import com.google.common.collect.Sets
+import com.google.common.io.Files
 import groovy.io.FileType
 import io.realm.annotations.Ignore
 import javassist.ClassPool
@@ -119,9 +120,10 @@ class RealmTransformer extends Transform {
             logger.info "  Modifying class ${it}"
             def ctClass = classPool.getCtClass(it)
             BytecodeModifier.useRealmAccessors(ctClass, managedFields, modelClasses)
-            ctClass.writeFile(outputProvider.getContentLocation(
-                    'realm', getInputTypes(), getScopes(), Format.DIRECTORY).canonicalPath)
+            ctClass.writeFile(getOutputFile(outputProvider).canonicalPath)
         }
+
+        copyResourceFiles(inputs, outputProvider)
 
         def toc = System.currentTimeMillis()
         logger.info "Realm Transform time: ${toc-tic} milliseconds"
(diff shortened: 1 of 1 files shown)


cmelchior:
I am very surprised we need to do this and it smells like a bug in either Kotlin or the Transform API, but apart from that :+1:

realm-transformer/src/main/groovy/io/realm/transformer/RealmTransformer.groovy
```diff
@@ -197,4 +199,26 @@ class RealmTransformer extends Transform {
...
+    private copyResourceFiles(Collection<TransformInput> inputs, TransformOutputProvider outputProvider) {
+        inputs.each {
+            it.directoryInputs.each {
+                def dirPath = it.file.absolutePath
+                it.file.eachFileRecurse(FileType.FILES) {
+                    if (!it.absolutePath.endsWith(SdkConstants.DOT_CLASS)) {
+                        logger.info "  Copying resource ${it}"
+                        def dest = new File(getOutputFile(outputProvider),
+                                it.absolutePath.substring(dirPath.length()))
+                        FileUtils.copyFile(it, dest);
```
emanuelez: I wonder which library includes `FileUtils`. I generally prefer to use Guava's `Files`, but I don't have a strong opinion about it.

emanuelez:
This is quite confusing. We specify the scope to be only classes so how do other resource appear in the inputs?

kneth:
Please add a line in `CHANGELOG.md` as it closes an issue.

cmelchior:
This is still unreleased code, so IMO it doesn't need an changelog (it was a bug found in our SNAPSHOT).

beeender:
:+1:

emanuelez:
:+1:

kneth:
@cmelchior Hmm, maybe an acknowledgement to "snapshot testers"?

cmelchior:
I don't have any strong opinions either way on that.

cmelchior:
@zaki50, @kneth will look at the above outside this PR, so I think this is safe to merge?

realm-transformer/src/main/groovy/io/realm/transformer/RealmTransformer.groovy
zaki50: done

zaki50:
I tested this issue with attached project. That project includes app project which uses
1. library project
2. library project in library project
3. binary aar

[realm-library-test_issue2356.zip](https://github.com/realm/realm-java/files/165368/realm-library-test_issue2356.zip)

cmelchior:
@zaki50 I just finished verifying the project. Looks :+1:

cmelchior:
:+1

PR #2361, opened 2016-02-29
Only load RealmObservableFactory when RxJava exist

### LINKED ISSUE #1990: Info message from art: Rejecting re-init
Just as an information, when I start the app with the most current version of Realm I get the following log messages:

```
        I/art: Rejecting re-init on previously-failed class java.lang.Class<io.realm.rx.RealmObservableFactory$4>
        I/art: Rejecting re-init on previously-failed class java.lang.Class<io.realm.rx.RealmObservableFactory$4>
        I/art: Rejecting re-init on previously-failed class java.lang.Class<io.realm.rx.RealmObservableFactory$3>
        I/art: Rejecting re-init on previously-failed class java.lang.Class<io.realm.rx.RealmObservableFactory$3>
        I/art: Rejecting re-init on previously-failed class java.lang.Class<io.realm.rx.RealmObservableFactory$2>
        I/art: Rejecting re-init on previously-failed class java.lang.Class<io.realm.rx.RealmObservableFactory$2>
        I/art: Rejecting re-init on previously-failed class java.lang.Class<io.realm.rx.RealmObservableFactory$1>
        I/art: Rejecting re-init on previously-failed class java.lang.Class<io.realm.rx.RealmObservableFactory$1>
```

We are currently not using the new `rx`feature, but just wanted to see if this is something to be concerned about.

The device was a Nexus 5 running Android 6.

beeender:
Close #1990.

* Check RxJava existence only once.
* Don't load RealmObservableFactory if RxJava doesn't exist.
* Correct exceptions in the java doc.

NOTE: When reflection called on the RealmObject/RealmResults, crash will
still happen because of Observable doesn't exist in the class path. In
such a case, a dummy rx.Observable is still needed as a workaround.

### DIFF
--- a/realm/realm-library/src/main/java/io/realm/RealmConfiguration.java
+++ b/realm/realm-library/src/main/java/io/realm/RealmConfiguration.java
@@ -61,6 +61,7 @@ public class RealmConfiguration {
 
     private static final Object DEFAULT_MODULE;
     private static final RealmProxyMediator DEFAULT_MODULE_MEDIATOR;
+    private static Boolean rxJavaAvailable;
 
     static {
         DEFAULT_MODULE = Realm.getDefaultModule();
@@ -154,9 +155,16 @@ public class RealmConfiguration {
     /**
      * Returns the {@link RxObservableFactory} that is used to create Rx Observables from Realm objects.
      *
+     * @throws UnsupportedOperationException if the required RxJava framework is not on the classpath.
      * @return the factory instance used to create Rx Observables.
      */
     public RxObservableFactory getRxFactory() {
+        // Since RxJava doesn't exist, rxObservableFactory is not initialized.
+        if (rxObservableFactory == null) {
+            throw new UnsupportedOperationException("RxJava seems to be missing from the classpath. " +
+                    "Remember to add it as a compile dependency." +
+                    " See https://realm.io/docs/java/latest/#rxjava for more details.");
+        }
         return rxObservableFactory;
     }
 
@@ -269,6 +277,23 @@ public class RealmConfiguration {
     }
 
     /**
+     * Check if RxJava is can be loaded.
+     *
+     * @return true if RxJava dependency exist.
+     */
+    private static synchronized boolean isRxJavaAvailable() {
+        if (rxJavaAvailable == null) {
+            try {
+                Class.forName("rx.Observable");
+                rxJavaAvailable = true;
+            } catch (ClassNotFoundException ignore) {
+                rxJavaAvailable = false;
+            }
+        }
+        return rxJavaAvailable;
+    }
+
+    /**
      * RealmConfiguration.Builder used to construct instances of a RealmConfiguration in a fluent manner.
      */
     public static class Builder {
@@ -281,7 +306,7 @@ public class RealmConfiguration {
         private SharedGroup.Durability durability;
         private HashSet<Object> modules = new HashSet<Object>();
         private HashSet<Class<? extends RealmObject>> debugSchema = new HashSet<Class<? extends RealmObject>>();
-        private RxObservableFactory rxFactory = new RealmObservableFactory();
+        private RxObservableFactory rxFactory;
 
         /**
          * Creates an instance of the Builder for the RealmConfiguration.
@@ -486,6 +511,9 @@ public class RealmConfiguration {
          * @return the created {@link RealmConfiguration}.
          */
         public RealmConfiguration build() {
+            if (rxFactory == null && isRxJavaAvailable()) {
+                rxFactory = new RealmObservableFactory();
+            }
             return new RealmConfiguration(this);
         }
 
--- a/realm/realm-library/src/main/java/io/realm/RealmObject.java
+++ b/realm/realm-library/src/main/java/io/realm/RealmObject.java
@@ -275,7 +275,8 @@ public abstract class RealmObject {
      *
      * @param <E> RealmObject class that is being observed. Must be this class or its super types.
      * @return RxJava Observable that only calls {@code onNext}. It will never call {@code onComplete} or {@code OnError}.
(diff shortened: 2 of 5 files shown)


beeender:
@realm/java

realm/realm-library/src/main/java/io/realm/RealmConfiguration.java
```diff
@@ -269,6 +270,23 @@ public String toString() {
     }
 
     /**
+     * Check if the RxJava is can be loaded.
+     *
+     * @return true if RxJava dependency exist.
+     */
+    public static synchronized boolean isRxJavaAvailable() {
```
cmelchior: Any reason this is public?

realm/realm-library/src/main/java/io/realm/RealmConfiguration.java
```diff
@@ -269,6 +270,23 @@ public String toString() {
...
+        if (rxJavaAvailable  == null) {
```
cmelchior: Too much whitespace before ==

realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -182,7 +182,7 @@ public void removeChangeListener(RealmChangeListener listener) {
      * </pre>
      *
      * @return RxJava Observable that only calls {@code onNext}. It will never call {@code onComplete} or {@code OnError}.
-     * @throws UnsupportedOperationException if the required RxJava framework is not on the classpath.
+     * @throws NoClassDefFoundError if the required RxJava framework is not on the classpath.
```
cmelchior: It looks a bit odd that the other methods can throw both UnsupportedOperation and NoClassDef, while this only throw the latter?

realm/realm-library/src/main/java/io/realm/RealmObject.java
```diff
@@ -290,7 +291,7 @@ public final void removeChangeListeners() {
             Observable<E> observable = (Observable<E>) realm.configuration.getRxFactory().from(dynamicRealm, dynamicObject);
             return observable;
         } else {
-            throw new UnsupportedOperationException(realm.getClass() + " not supported");
+            throw new UnsupportedOperationException(realm.getClass() + " does not support RxJava.");
```
cmelchior: The exception in the RealmObservableFactory throws an message with a pointer to our docs. IMO we should do the same in these now as people new to Realm might not have read the docs and accidentally use this method?

cmelchior:
Nice correction. Good idea to move the check to RealmConfiguration

realm/realm-library/src/main/java/io/realm/RealmConfiguration.java
beeender: `io/realm/RealmConfiguration.java` and `io/realm/rx/RealmObservableFactory.java` are in different package.

realm/realm-library/src/main/java/io/realm/BaseRealm.java
beeender: Whenever jvm needs the information of `Observable`, `NoClassDefFoundError` will be thrown. `UnsupportedOperationException` will only be thrown when user inherit the `BaseRealm` and create `Observable` from there. I am not sure why did you throw that at the first place. If you don't recall the reason, maybe we could just remove that?

realm/realm-library/src/main/java/io/realm/RealmConfiguration.java
cmelchior: But the method is only used in RealmConfiguration right, so couldn't it just be private?

beeender: Uh, you are right. will fix it.

cmelchior:
:+1:

realm/realm-library/src/main/java/io/realm/RealmConfiguration.java
```diff
@@ -61,6 +61,7 @@
 
     private static final Object DEFAULT_MODULE;
     private static final RealmProxyMediator DEFAULT_MODULE_MEDIATOR;
+    private static Boolean rxJavaAvailable;
```
emanuelez: Why not a primitive here?

realm/realm-library/src/main/java/io/realm/RealmConfiguration.java
```diff
@@ -269,6 +277,23 @@ public String toString() {
     }
 
     /**
+     * Check if the RxJava is can be loaded.
```
emanuelez: -> Check if RxJava can be loaded.

realm/realm-library/src/main/java/io/realm/rx/RealmObservableFactory.java
```diff
@@ -206,17 +189,11 @@ public void call() {
     }
 
     @Override
-    public Observable<RealmQuery<DynamicRealmObject>> from(final DynamicRealm realm, final RealmQuery<DynamicRealmObject> query) {
+    public Observable<RealmQuery<DynamicRealmObject>> from(final DynamicRealm realm,
+                                                           final RealmQuery<DynamicRealmObject> query) {
```
emanuelez: What's with the indenting?

kneth:
:+1:

realm/realm-library/src/main/java/io/realm/RealmConfiguration.java
beeender: `rxJavaAvailable` can be `null`/`false`/`true` . `null` means it is not checked yet.

realm/realm-library/src/main/java/io/realm/rx/RealmObservableFactory.java
beeender: It is longer than 120 chars

PR #2364, opened 2016-02-29
Realm Observables now holds a Realm instance reference

### LINKED ISSUE #2357: Realm, RxJava, asObservable() and doOnUnsubscribe()
I posted a question on [stackoverflow](http://stackoverflow.com/questions/35661450/realm-rxjava-asobservable-and-doonunsubscribe), describing my issue. But long story short, I can't find a clean way to `close()` a realm instance when I unsubscribe from an `Observable` exposed through the Realm API.

Because `RealmObservableFactory.getRealmResultsObservable()` adds a `Subscription` to `removeChangeListener` I can't do something like this:

    public Observable<City> getHomeTown() {
        final Realm realm = Realm.getDefaultInstance();
        return realm.where(City.class).equalTo("name", "Cluj-Napoca").findAllAsync().asObservable()
                .doOnUnsubscribe(new Action0() {
                    @Override
                    public void call() {
                        realm.close();
                    }
                })
                .compose(new NullIfNoRealmObject<City>());
    }

Something like `Realm.asObservable(boolean closeAfterUnsubscribe)` would be really helpful for me, but someone else might have a good solution that does not involve changing the API.

cmelchior:
Fixes #2357 

Due to how RxJava executes the chain of Observables, currently it isn't possible to use `doOnUnsubscribe()` to close Realm instances as the side-effect is triggered before the Observable is truly unsubscribed to (crashing our 'removeChangeListeners() method).

This PR changes the current implementation so each Observable also holds onto a Realm instance prevent the Realm from being fully closed until the Observable is unsubscribed to. This means we can now guarantee that the Realm is alive for as long as the Observable is subscribed to. The only downside is that you risk leaking Realm instances if you forget to unsubscribe, but that would be a memory leak in any case.

@realm/java 
/cc @lucamtudor

### DIFF
--- a/realm/realm-library/src/androidTest/java/io/realm/RxJavaTests.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/RxJavaTests.java
@@ -37,6 +37,7 @@ import io.realm.rule.RunTestInLooperThread;
 import io.realm.rule.TestRealmConfigurationFactory;
 import rx.Observable;
 import rx.Subscription;
+import rx.functions.Action0;
 import rx.functions.Action1;
 
 import static org.junit.Assert.assertEquals;
@@ -49,20 +50,29 @@ public class RxJavaTests {
     @Rule
     public final UiThreadTestRule uiThreadTestRule = new UiThreadTestRule();
     @Rule
-    public final RunInLooperThread looperThread = new RunInLooperThread();
+    public final RunInLooperThread looperThread = new RunInLooperThread() {
+        @Override
+        public void looperTearDown() {
+            if (subscription != null && !subscription.isUnsubscribed()) {
+                subscription.unsubscribe();
+            }
+        }
+    };
     @Rule
     public final TestRealmConfigurationFactory configFactory = new TestRealmConfigurationFactory();
 
     private Realm realm;
+    private Subscription subscription;
 
     @Before
     public void setUp() throws Exception {
-        RealmConfiguration realmConfig = configFactory.createConfiguration();
-        realm = Realm.getInstance(realmConfig);
+        // For non-LooperThread tests
+        realm = Realm.getInstance(configFactory.createConfiguration());
     }
 
     @After
     public void tearDown() throws Exception {
+        // For non-LooperThread tests
         if (realm != null) {
             realm.close();
         }
@@ -76,7 +86,7 @@ public class RxJavaTests {
         realm.commitTransaction();
 
         final AtomicBoolean subscribedNotified = new AtomicBoolean(false);
-        obj.<AllTypes>asObservable().subscribe(new Action1<AllTypes>() {
+        subscription = obj.<AllTypes>asObservable().subscribe(new Action1<AllTypes>() {
             @Override
             public void call(AllTypes rxObject) {
                 assertTrue(rxObject == obj);
@@ -84,6 +94,7 @@ public class RxJavaTests {
             }
         });
         assertTrue(subscribedNotified.get());
+        subscription.unsubscribe();
     }
 
     @Test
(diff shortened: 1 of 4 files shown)


realm/realm-library/src/androidTest/java/io/realm/RxJavaTests.java
```diff
@@ -311,7 +336,7 @@ public void call(DynamicRealm rxRealm) {
     @UiThreadTest
     public void unsubscribe() {
```
beeender: change name to `unsubscribe_onWrongThread` ?

realm/realm-library/src/androidTest/java/io/realm/rule/RunInLooperThread.java
```diff
@@ -103,6 +103,12 @@ public void run() {
                                 threadAssertionError[0] = e;
                                 unitTestFailed = true;
                             } finally {
+                                try {
+                                    looperTearDown();
+                                } catch (Throwable t) {
+                                    threadAssertionError[0] = t;
```
beeender: Check if `threadAssertionError[0]` is `null` first. The first exception is more important.

realm/realm-library/src/main/java/io/realm/rx/RealmObservableFactory.java
```diff
@@ -80,48 +86,81 @@ public void call() {
...
                 // value.
-                subscriber.onNext(realm);
+                subscriber.onNext(observableRealm);
             }
         });
     }
 
     @Override
-    public <E extends RealmObject> Observable<RealmResults<E>> from(Realm realm, RealmResults<E> results) {
+    public <E extends RealmObject> Observable<RealmResults<E>> from(final Realm realm, final RealmResults<E> results) {
```
beeender: I think those params should be final in `RxObservableFactory` as well.

beeender:
Besides some minor comments, :+1:

realm/realm-library/src/main/java/io/realm/rx/RealmObservableFactory.java
cmelchior: I don't think it is custom to have final parameters in interfaces? Mostly because the only reason for doing so is because of implementation specific details which the interface does not care about.

beeender:
:+1:

realm/realm-library/src/main/java/io/realm/rx/RealmObservableFactory.java
```diff
@@ -54,25 +55,30 @@ public RealmObservableFactory() {
     @Override
     public Observable<Realm> from(final Realm realm) {
         checkRxJavaAvailable();
+        final RealmConfiguration realmConfig = realm.getConfiguration();
         return Observable.create(new Observable.OnSubscribe<Realm>() {
             @Override
             public void call(final Subscriber<? super Realm> subscriber) {
+                // Get instance to make sure that the Realm is open for as long as the
+                // Observable is subscribed to.
```
emanuelez: to -> to it

realm/realm-library/src/main/java/io/realm/rx/RealmObservableFactory.java
```diff
@@ -80,48 +86,81 @@ public void call() {
     @Override
     public Observable<DynamicRealm> from(final DynamicRealm realm) {
         checkRxJavaAvailable();
+        final RealmConfiguration realmConfig = realm.getConfiguration();
         return Observable.create(new Observable.OnSubscribe<DynamicRealm>() {
             @Override
             public void call(final Subscriber<? super DynamicRealm> subscriber) {
+                // Get instance to make sure that the Realm is open for as long as the
+                // Observable is subscribed to.
```
emanuelez: same

emanuelez:
just minor comments issues, but they happen in a bunch of places. Otherwise this looks good to me :+1:

realm/realm-library/src/main/java/io/realm/rx/RealmObservableFactory.java
```diff
@@ -162,21 +202,49 @@ public void call() {
...
+                object.addChangeListener(listener);
+                subscriber.add(Subscriptions.create(new Action0() {
+                    @Override
+                    public void call() {
+                        object.removeChangeListener(listener);
+                        observableRealm.close();
+                    }
+                }));
+
+                // Immediately call onNext with the current value, as due to Realms auto-update, it will be the latest
```
kneth: "Realm's"

PR #2372, opened 2016-03-01
RealmConfiguration: Integrating object store

### LINKED ISSUE #1445: Migrate to ObjectStore
* [ ] `Realm`
* [x] `RealmConfiguration` - see #2372 
* [ ] `RealmResults`
* [ ] `RealmObject`

kneth:
Integrating object store

*Notice: it will fail when building*

Closes #1445


bmunkholm:
I suggest you move the above subtasks to #1445 and make this a smal PR like it currently is.

bmunkholm:
Seems to miss all the object store files.
Is this RealmConfiguration required for the Object Store to work? It doesn't seem to help with anything at all by itself?

kneth:
The object store has a configuration concept which is the same as we have in the Java binding. In order to open a Realm, you must supply a configuration.

Yes, this PR cannot build since it doesn't include the object store files. We need to figure out a way to distribute the object store - copying files from a random branch isn't really a sane approach.

kneth:
Replaced by #3148

PR #2379, opened 2016-03-02
Update issue template

bmunkholm:
@cmelchior


emanuelez:
The gradle version would be useful as well

cmelchior:
Looks fine to me :+1: 
I am not sure the Gradle version is that useful? Sometimes it can be tricky to find and I don't think we ever had a bug related to the Gradle version? IMO Gradle version fits into the same category as buildToolsVersion i.e. we can ask for them if we think it matter, but to keep this fairly succinct we can probably leave it out.

beeender:
:+1:

PR #2390, opened 2016-03-04
More checkings when modify RealmList

### LINKED ISSUE #2381: Missing checking when call `RealmList.add()`/`set` with `DynamicObject`
The below test should pass as well as other `RealmList`'s methods `add(int location, E object)`/`set(int location, E object)`

```java
    @Test
    public void getList_addWrongTypeTo() {
        DynamicRealm dynamicRealm = DynamicRealm.getInstance(realm.configuration);
        dynamicRealm.beginTransaction();

        DynamicRealmObject allTypes = dynamicRealm.createObject(AllTypes.CLASS_NAME);
        allTypes.setString(AllTypes.FIELD_STRING, "bender");

        DynamicRealmObject cat = dynamicRealm.createObject(Cat.CLASS_NAME);
        cat.setString("name", "nibbler");

        RealmList<DynamicRealmObject> list = allTypes.getList(AllTypes.FIELD_REALMLIST);
        list.add(cat);
        thrown.expect(IllegalArgumentException.class);

        dynamicRealm.commitTransaction();
    }
```

beeender:
Fix #2381

* New method LinkView.getTargetTable().
* Proper checking to add DynamicRealmObject to RealmList.
* Disallow modifying with RealmObject belongs to another Realm instance.

### DIFF
--- a/realm/realm-jni/src/io_realm_internal_LinkView.cpp
+++ b/realm/realm-jni/src/io_realm_internal_LinkView.cpp
@@ -193,11 +193,24 @@ JNIEXPORT jlong JNICALL Java_io_realm_internal_LinkView_nativeFind
 }
 
 JNIEXPORT void JNICALL Java_io_realm_internal_LinkView_nativeRemoveAllTargetRows
-  (JNIEnv *env, jobject, jlong nativeLinkViewPtr) {
+  (JNIEnv *env, jobject, jlong nativeLinkViewPtr)
+{
     TR_ENTER_PTR(nativeLinkViewPtr)
     try {
         LinkView *lv = LV(nativeLinkViewPtr);
         lv->remove_all_target_rows();
     } CATCH_STD()
- }
+}
+
+JNIEXPORT jlong JNICALL Java_io_realm_internal_LinkView_nativeGetTargetTable
+  (JNIEnv*, jobject, jlong nativeLinkViewPtr)
+{
+    TR_ENTER_PTR(nativeLinkViewPtr)
+
+    LinkView* lv = LV(nativeLinkViewPtr);
+    Table* pTable = &(lv->get_target_table());
+    LangBindHelper::bind_table_ptr(pTable);
+
+    return reinterpret_cast<jlong>(pTable);
+}
 
--- a/realm/realm-library/src/androidTest/java/io/realm/RealmListTests.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/RealmListTests.java
@@ -27,12 +27,14 @@ import org.junit.runner.RunWith;
 
 import java.util.Collections;
 import java.util.concurrent.Callable;
+import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 
 import io.realm.entities.AllTypes;
+import io.realm.entities.Cat;
 import io.realm.entities.CyclicType;
 import io.realm.entities.CyclicTypePrimaryKey;
 import io.realm.entities.Dog;
@@ -103,6 +105,7 @@ public class RealmListTests {
     private void checkMethodsOnEmptyList(Realm realm, RealmList<Dog> list) {
         realm.beginTransaction();
         for (int i = 0; i < 4; i++) {
+            //noinspection TryWithIdenticalCatches
             try {
                 switch (i) {
                     case 0: list.get(0); break;
@@ -915,7 +918,7 @@ public class RealmListTests {
     }
 
     @Test
-    public void testRemoveAllFromRealm() {
+    public void removeAllFromRealm() {
         Owner owner = testRealm.where(Owner.class).findFirst();
         RealmList<Dog> dogs = owner.getDogs();
         assertEquals(TEST_OBJECTS, dogs.size());
@@ -928,7 +931,7 @@ public class RealmListTests {
     }
 
     @Test
-    public void testRealmRemoveAllNotManagedList() {
+    public void removeAllFromRealm_notManagedList() {
         Owner owner = testRealm.where(Owner.class).findFirst();
         RealmList<Dog> dogs = owner.getDogs();
         assertEquals(TEST_OBJECTS, dogs.size());
@@ -947,7 +950,7 @@ public class RealmListTests {
     }
 
     @Test
-    public void testRealmRemoveAllOutsideTransaction() {
+    public void removeAllFromRealm_outsideTransaction() {
         Owner owner = testRealm.where(Owner.class).findFirst();
         RealmList<Dog> dogs = owner.getDogs();
         try {
@@ -959,7 +962,7 @@ public class RealmListTests {
     }
 
     @Test
-    public void testRemoveAllFromListStandaloneObjectShouldThrow() {
+    public void removeAllFromRealm_listWithStandaloneObjectShouldThrow() {
         final RealmList<Dog> list = new RealmList<Dog>();
 
         testRealm.beginTransaction();
@@ -986,7 +989,7 @@ public class RealmListTests {
     }
 
     @Test
-    public void testRemoveAllFromRealmEmptyList() {
+    public void removeAllFromRealm_emptyList() {
         RealmList<Dog> dogs = testRealm.where(Owner.class).findFirst().getDogs();
         assertEquals(TEST_OBJECTS, dogs.size());
 
@@ -1006,7 +1009,7 @@ public class RealmListTests {
     }
 
     @Test
-    public void testRemoveAllFromRealmInvalidListShouldThrow() {
+    public void removeAllFromRealm_invalidListShouldThrow() {
         RealmList<Dog> dogs = testRealm.where(Owner.class).findFirst().getDogs();
         assertEquals(TEST_OBJECTS, dogs.size());
         testRealm.close();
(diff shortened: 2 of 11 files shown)


realm/realm-library/src/main/java/io/realm/RealmList.java
```diff
@@ -218,15 +218,30 @@ public E set(int location, E object) {
...
-        // We don't support moving DynamicRealmObjects across Realms automatically. The overhead is too big as you
-        // have to run a full schema validation for each object.
         if (object instanceof DynamicRealmObject) {
-            throw new IllegalArgumentException("Automatically copying DynamicRealmObjects from other Realms are not supported");
+            String listClassName = view.getTargetTable().getName();
+            String objectClassName = ((DynamicRealmObject) object).row.getTable().getName();
+            if (object.realm == realm && listClassName.equals(objectClassName)) {
+                // Same Realm instance and same target table
+                return object;
+            } else if (object.realm == realm) {
```
cmelchior: It looks a bit funny with 3 these if's at the same level, I would probably inline this in the first if like this:
```
if (object.realm == realm) {
  if (listClassName.equals(objectClassName) {
   //...
   } else {
   //...
   }
} else {

}
```

realm/realm-library/src/main/java/io/realm/RealmList.java
```diff
@@ -218,15 +218,30 @@ public E set(int location, E object) {
...
-        if (object.row != null && object.realm.getPath().equals(realm.getPath())) {
-            return object;
-        }
-
-        // We don't support moving DynamicRealmObjects across Realms automatically. The overhead is too big as you
-        // have to run a full schema validation for each object.
         if (object instanceof DynamicRealmObject) {
-            throw new IllegalArgumentException("Automatically copying DynamicRealmObjects from other Realms are not supported");
+            String listClassName = view.getTargetTable().getName();
+            String objectClassName = ((DynamicRealmObject) object).row.getTable().getName();
```
cmelchior: DynamicRealmObject has a `getType` method for this.

cmelchior: DynamicRealmObject has a `getType` method for this.

realm/realm-library/src/androidTest/java/io/realm/RealmListTests.java
```diff
@@ -1019,4 +1022,213 @@ public void testRemoveAllFromRealmInvalidListShouldThrow() {
             assertEquals("This Realm instance has already been closed, making it unusable.", e.getMessage());
         }
     }
+
+    private interface ListModifier <T extends RealmObject> {
```
cmelchior: Please add a small description for this interface. At a glance the name is not that telling.

realm/realm-library/src/androidTest/java/io/realm/RealmListTests.java
```diff
@@ -1019,4 +1022,213 @@ public void testRemoveAllFromRealmInvalidListShouldThrow() {
...
+        void modify(RealmList<T> list, T element);
+    }
+
+    private void doTestModifyListFromOtherThread(final ListModifier<Dog> modifier) {
```
cmelchior: Also, small description for this method

realm/realm-library/src/androidTest/java/io/realm/RealmListTests.java
```diff
@@ -1019,4 +1022,213 @@ public void testRemoveAllFromRealmInvalidListShouldThrow() {
...
+            @Override
+            public void modify(RealmList<Dog> list, Dog element) {
+                list.add(element);
+            }
+        });
+    }
+
+    @Test
+    public void add_withLocationObjectFromOtherThread() {
+        doTestModifyListFromOtherThread(new ListModifier<Dog>() {
```
cmelchior: In e.g. RealmResults all these tests are combined in one method: `runMethodOnWrongThread`. It takes up a lot less space than this? Not sure what is better, but I probably lean slightly towards what RealmResultsTests are doing?

cmelchior:
Mostly a few comments about the test cases and a few improvements to checking, otherwise it looks good.

beeender:
@realm/java Updated

realm/realm-library/src/main/java/io/realm/internal/LinkView.java
```diff
@@ -153,6 +153,12 @@ public void removeAllTargetRows() {
         nativeRemoveAllTargetRows(nativePointer);
     }
 
+    public Table getTargetTable() {
```
kneth: Can't you use `Table::get_table_ref()`?

beeender: @kneth Yep, updated the commit. Please help to check.

realm/realm-library/src/androidTest/java/io/realm/RealmListTests.java
```diff
@@ -1019,4 +1022,167 @@ public void testRemoveAllFromRealmInvalidListShouldThrow() {
             assertEquals("This Realm instance has already been closed, making it unusable.", e.getMessage());
         }
     }
+
+    @Test
+    public void add_set_objectFromOtherThread() {
+        final CountDownLatch finishedLatch = new CountDownLatch(1);
+        final Dog dog = testRealm.where(Dog.class).findFirst();
+        final String expectedMsg = "Cannot copy an object from another Realm instance.";
```
cmelchior: Shouldn't this throw an IllegalState because of a wrong thread first? Then check for this later?

realm/realm-library/src/androidTest/java/io/realm/RealmListTests.java
```diff
@@ -1019,4 +1022,167 @@ public void testRemoveAllFromRealmInvalidListShouldThrow() {
...
+        }).start();
+        TestHelper.awaitOrFail(finishedLatch);
+    }
+
+    @Test
+    public void add_set_dynamicObjectFromOtherThread() {
+        final CountDownLatch finishedLatch = new CountDownLatch(1);
+        DynamicRealm dynamicRealm = DynamicRealm.getInstance(testRealm.getConfiguration());
+        final DynamicRealmObject dynDog = dynamicRealm.where(Dog.CLASS_NAME).findFirst();
+        final String expectedMsg = "Cannot copy an object from another Realm instance.";
```
cmelchior: Same

realm/realm-library/src/main/java/io/realm/internal/LinkView.java
cmelchior: Should we cache this? Right now you create a new Table object this method is called.

realm/realm-library/src/androidTest/java/io/realm/RealmListTests.java
beeender:
```java
} else {
                // We don't support moving DynamicRealmObjects across Realms automatically. The overhead is too big as
                // you have to run a full schema validation for each object.
                // And copying from another Realm instance pointed to the same Realm file is not supported as well.
                throw new IllegalArgumentException("Cannot copy an object from another Realm instance.");
            }
```

change to 
```
            } else if (realm.threadId ==object.realm.threadId) {
                // We don't support moving DynamicRealmObjects across Realms automatically. The overhead is too big as
                // you have to run a full schema validation for each object.
                // And copying from another Realm instance pointed to the same Realm file is not supported as well.
                throw new IllegalArgumentException("Cannot copy an object from another Realm instance.");
            } else {
                throw new IllegalStateException("Cannot copy an object from another Realm instance in anther thread.");
            }
```

Something like above is better?

cmelchior: I was more thinking about the standard `realm.checkIfValid()` check.

Would `object.realm.checkIfValid()` not be enough?

beeender: No, for the case to check `DynamicRealm` and `Realm` in the same thread.

cmelchior: Do, you are right. Just ignore my comment then. The original wording is fine. The newer one is technically more correct, but I am not sure the complexity is worth it.

realm/realm-library/src/main/java/io/realm/internal/LinkView.java
beeender: It is only a new java `Table` object which points to the same `Table` pointer in the core. No need to cache like other tables we created in java.

realm/realm-library/src/androidTest/java/io/realm/RealmListTests.java
```diff
@@ -1019,4 +1022,167 @@ public void testRemoveAllFromRealmInvalidListShouldThrow() {
...
+
+        new Thread(new Runnable() {
+            @Override
+            public void run() {
+                Realm realm = Realm.getInstance(testRealm.getConfiguration());
+                realm.beginTransaction();
+                RealmList<Dog> list = realm.createObject(Owner.class).getDogs();
+                list.add(realm.createObject(Dog.class));
+                try {
```
bmunkholm: Refactor the common code out of this method and the following.

beeender: @bmunkholm That was what i did before -- use common code https://github.com/realm/realm-java/pull/2390#discussion_r55023084 . But we found it is not easy to read.

realm/realm-library/src/androidTest/java/io/realm/RealmListTests.java
```diff
@@ -1019,4 +1022,167 @@ public void testRemoveAllFromRealmInvalidListShouldThrow() {
...
+        try {
+            list.set(0, dynCat);
+            fail();
+        } catch (IllegalArgumentException expected) {
+            assertEquals(expectedMsg, expected.getMessage());
+
+        }
+
+        dynamicRealm.cancelTransaction();
+
```
stk1m1: White space removal? I noticed line 1185 doesn't have one.

realm/realm-library/src/main/java/io/realm/RealmList.java
```diff
@@ -218,15 +219,32 @@ public E set(int location, E object) {
...
-        // Object is already in this realm
-        if (object.row != null && object.realm.getPath().equals(realm.getPath())) {
-            return object;
-        }
-
-        // We don't support moving DynamicRealmObjects across Realms automatically. The overhead is too big as you
-        // have to run a full schema validation for each object.
         if (object instanceof DynamicRealmObject) {
-            throw new IllegalArgumentException("Automatically copying DynamicRealmObjects from other Realms are not supported");
+            String listClassName = view.getTargetTable().getName().substring(Table.TABLE_PREFIX.length());
```
cmelchior: I would prefer if this information was accessed through RealmSchema somehow. All these rules about naming should all be enforced there instead of being spread around the codebase (IMO). It probably means adding something like `getSchemaForTable` as I am not sure the functionality exists yet.

realm/realm-jni/src/io_realm_internal_LinkView.cpp
```diff
@@ -193,11 +193,24 @@ JNIEXPORT jlong JNICALL Java_io_realm_internal_LinkView_nativeFind
...
+JNIEXPORT jlong JNICALL Java_io_realm_internal_LinkView_nativeGetTargetTable
+  (JNIEnv *, jobject, jlong nativeLinkViewPtr)
+{
+    TR_ENTER_PTR(nativeLinkViewPtr)
+
+    LinkView *lv = LV(nativeLinkViewPtr);
+    Table* pTable = &(lv->get_target_table());
+    LangBindHelper::bind_table_ptr(pTable);
+
+    return (jlong)pTable;
```
stk1m1: I'm trying be cautious, but wouldn't it be better to use `return reinterpret_cast<jlong>(pTable);` since the returned value would be re-casted to `Table*` again?

realm/realm-library/src/androidTest/java/io/realm/RealmListTests.java
stk1m1: > } else if (realm.threadId ==object.realm.threadId) {

:+1: for checking thread.

beeender: Done

realm/realm-library/src/androidTest/java/io/realm/RealmListTests.java
beeender: Done

realm/realm-library/src/main/java/io/realm/internal/LinkView.java
```diff
@@ -153,6 +153,20 @@ public void removeAllTargetRows() {
...
+        // Execute the disposal of abandoned realm objects each time a new realm object is created
+        context.executeDelayedDisposal();
+        long nativeTablePointer = nativeGetTargetTable(nativePointer);
+        try {
+            // Copy context reference from parent
+            return new Table(context, this.parent, nativeTablePointer);
+        }
+        catch (RuntimeException e) {
```
zaki50: please remove new line before `catch`

realm/realm-jni/src/io_realm_internal_LinkView.cpp
beeender: done

cmelchior:
:+1:

realm/realm-jni/src/io_realm_internal_LinkView.cpp
```diff
@@ -193,11 +193,24 @@ JNIEXPORT jlong JNICALL Java_io_realm_internal_LinkView_nativeFind
...
- }
+}
+
+JNIEXPORT jlong JNICALL Java_io_realm_internal_LinkView_nativeGetTargetTable
+  (JNIEnv *, jobject, jlong nativeLinkViewPtr)
+{
+    TR_ENTER_PTR(nativeLinkViewPtr)
+
+    LinkView *lv = LV(nativeLinkViewPtr);
+    Table* pTable = &(lv->get_target_table());
```
kneth: Where to put `*`?

beeender: opps... copied lines from different places...

kneth:
:+1:

PR #2397, opened 2016-03-08
Upgrading to Realm Core 0.97.0

kneth:
@realm/java

### DIFF
--- a/realm/realm-jni/src/util.hpp
+++ b/realm/realm-jni/src/util.hpp
@@ -101,7 +101,7 @@ std::string num_to_string(T pNumber)
 #define G(x)    reinterpret_cast<realm::Group*>(x)
 #define ROW(x)  reinterpret_cast<realm::Row*>(x)
 #define SG(ptr) reinterpret_cast<realm::SharedGroup*>(ptr)
-#define CH(ptr) reinterpret_cast<realm::ClientHistory*>(ptr)
+#define CH(ptr) reinterpret_cast<realm::Replication*>(ptr)
 #define HO(T, ptr) reinterpret_cast<realm::SharedGroup::Handover <T>* >(ptr)
 
 // Exception handling
--- a/realm/realm-library/src/main/java/io/realm/internal/SharedGroupManager.java
+++ b/realm/realm-library/src/main/java/io/realm/internal/SharedGroupManager.java
@@ -21,6 +21,7 @@ import java.io.File;
 import java.io.IOException;
 
 import io.realm.RealmConfiguration;
+import io.realm.internal.log.RealmLog;
 
 /**
  * This class wraps access to a given Realm file on a single thread including its {@link SharedGroup} and
@@ -162,6 +163,7 @@ public class SharedGroupManager implements Closeable {
 
     /**
      * Compacts a Realm file. It cannot be open when calling this method.
+     * Returns true if compaction succeeded, false otherwise.
      */
     public static boolean compact(RealmConfiguration configuration) {
         SharedGroup sharedGroup = null;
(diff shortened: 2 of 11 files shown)


beeender:
Not sure if https://github.com/realm/realm-object-store/pull/52/files#r55271290 this change has impact on us as well. in `completedAsyncQueriesUpdate`

kneth:
@beeender The error in the failing test seems to indicate it :-)

cmelchior:
Should this PR also modify our `deleteRealm` method? since (all?) the auxillary files are now inlined?

kneth:
@cmelchior We can remove the aux. files from the list but then removing old database will leave these files untouched. Adding a note to the change log that you must manually remove aux. files is an option.

cmelchior:
I guess leaving it in for now isn't a big issue.

realm/realm-library/src/main/java/io/realm/internal/SharedGroupManager.java
```diff
@@ -169,10 +169,12 @@ public static boolean compact(RealmConfiguration configuration) {
         try {
             sharedGroup = new SharedGroup(
                     configuration.getPath(),
-                    SharedGroup.EXPLICIT_TRANSACTION,
+                    SharedGroup.IMPLICIT_TRANSACTION,
```
nhachicha: Why?

realm/realm-library/src/main/java/io/realm/internal/SharedGroupManager.java
```diff
@@ -169,10 +169,12 @@ public static boolean compact(RealmConfiguration configuration) {
...
                     SharedGroup.Durability.FULL,
                     configuration.getEncryptionKey());
             result = sharedGroup.compact();
+        } catch (Exception e) {
+            return false;
```
nhachicha: Shouldn't we at least log the error _RealmLog_

realm/realm-jni/src/util.hpp
```diff
@@ -101,7 +101,7 @@ std::string num_to_string(T pNumber)
 #define G(x)    reinterpret_cast<realm::Group*>(x)
 #define ROW(x)  reinterpret_cast<realm::Row*>(x)
 #define SG(ptr) reinterpret_cast<realm::SharedGroup*>(ptr)
-#define CH(ptr) reinterpret_cast<realm::ClientHistory*>(ptr)
+#define CH(ptr) reinterpret_cast<realm::Replication*>(ptr)
```
nhachicha: I don't think this macro is used anymore. remove it?

nhachicha:
LGTM besides small comments

beeender:
The `TableView`'s version could be `std::numeric_limits<uint64_t>::max()` for the default of the source `LinkView` gets deleted. See https://github.com/realm/realm-core/pull/1510 . After the max `uint64_t` convert to jlong, it give `-1` which is accidentally the same with our `RealmResults.TABLE_VIEW_VERSION_NONE`. I checked the code, and it seems to be no problem with current implementation. However, @nhachicha would you please help to check again? This can be hard to debug if there is a problem in the future.

cmelchior:
I think that `std::numeric_limits<uint64_t>::max()` have always been the value used internally by Core, but now it just got more visible.

realm/realm-library/src/main/java/io/realm/internal/SharedGroupManager.java
kneth: We get a `LogicError` exception when you try to open the Realm with explicit transaction if it is already opened with implicit transaction. I haven't talked with the core team so I don't know if they are aware of it.

realm/realm-jni/src/util.hpp
kneth: It is used three places in `io_realm_internal_SharedGroup.cpp`.

cmelchior:
:+1:

nhachicha:
:+1:

beeender:
:+1:

stk1m1:
:+1:

PR #2399, opened 2016-03-08
Rename clear and removeAllFromRealm

cmelchior:
0.88 will be released before the RealmCollection API is done with review. Since we are adding two new methods in 0.88 (`Realm.clear()` and `RealmList.removeAllFromRealm()` ) and these are not compatible with the refactoring done by the RealmCollection API. This PR changes these methods so they don't need to be deprecated right after they are released.

It means that API gets slightly confusing since we will have `Realm.clear(Class)` and `Realm.deleteAll()`, at least until the next release. I think that is acceptable.

@realm/java


beeender:
:+1:

stk1m1:
:+1:

kneth:
:+1:

PR #2400, opened 2016-03-08
Remove optional APIs if dependencies don't meet

beeender:
Bytecode manipulate is quite fun :)

What are we trying to solve in this PR:

We are going to have more and more APIs, lots of them might have external dependencies like `RealmObject.asObeservable()` which needs RxJava. Not all users need it, but it is supplied as the first class Realm API for convenience. 

1) This increases the methods count in the final apk which help users to hit the DEX methods limitation quickly.
2) This will cause crash if the apk doesn't meet the dependency and some other lib try to use reflection on those classes. See https://realm.io/docs/java/latest/#jackson-databind

To solve this, this PR created a new internal annotation `OptionalAPI` which can be set on methods with specific dependencies like:

```java
public abstract class RealmObject {
    //...
    @OptionalAPI(dependencies = {"rx.Observable"})
    public <E extends RealmObject> Observable<E> asObservable() {
    //...
    }
    //...
}
```

During the apk compiling time, `RealmOptionalAPITransformer` will try to figure out if user's project has all the dependencies the optional API needs, if no, the method will just be removed from the class.

Also, we might need another way to let user choose if they want to have those APIs even the dependencies are there.


beeender:
@realm/java The code here is still a mess. So please just take a quick look of the PR message to see if the idea is OK.

realm-transformer/src/main/groovy/io/realm/transformer/RealmOptionalAPITransformer.groovy
```diff
@@ -0,0 +1,183 @@
...
+    }
+
+    @Override
+    Set<QualifiedContent.Scope> getScopes() {
+        return Sets.immutableEnumSet(Scope.EXTERNAL_LIBRARIES)
+    }
+
+    @Override
+    Set<Scope> getReferencedScopes() {
+        return Sets.immutableEnumSet(Scope.PROJECT, Scope.PROJECT_LOCAL_DEPS, Scope.SUB_PROJECTS,
```
cmelchior: Shouldn't the scope only be the PROJECT?

stk1m1:
> During the apk compiling time, `RealmOptionalAPITransformer` will try to figure out if user's project has all the dependencies the optional API needs, if no, the method will just be removed from the class.

:100:

cmelchior:
Great idea :dancers: 

Ideally we shouldn't have any optional API's, but since we don't have extension functions this looks like a great way to work around libraries that uses reflection like Mockito and Jackson-databind.

I haven't dug that deep into the Transform API yet to comment on the specifics, but your scopes seems a bit large? Also can we please write these Transforms in a typesafe language like Kotlin/Java? :)

gradle-plugin/src/main/groovy/io/realm/gradle/Realm.groovy
```diff
@@ -48,6 +49,9 @@ class Realm implements Plugin<Project> {
         }
 
         project.android.registerTransform(new RealmTransformer())
+        if (!isAndroidLib) {
```
emanuelez: Why can't a library use an optional API?

realm-transformer/src/main/groovy/io/realm/transformer/RealmOptionalAPITransformer.groovy
```diff
@@ -0,0 +1,183 @@
...
+        // Find all the class names
+        def classNames = getClassNames(inputs)
+        def refClassNames = getClassNames(referencedInputs)
+        // Create and populate the Javassist class pool
+        def classPool = createClassPool(inputs)
+
+        def pkgNameQuote = Pattern.quote("io.realm")
+        def pattern = Pattern.compile("^$pkgNameQuote\\.[^\\.]*\$")
+        logger.info "$pattern"
+        classNames.findAll {
```
emanuelez: use ´filter´ instead

realm-transformer/src/main/groovy/io/realm/transformer/RealmOptionalAPITransformer.groovy
```diff
@@ -0,0 +1,183 @@
...
+                                ).replace(File.separatorChar, '.' as char)
+                        classNames.add(className)
+                    }
+                }
+            }
+
+            it.jarInputs.each {
+                def jarFile = new JarFile(it.file)
+                logger.info "jarFile " + jarFile.name
+                jarFile.entries().each {
```
emanuelez: instead of `each` and `if` I would probably go with `filter` and `each` just to make the code flatter and have separation of concerns

realm-transformer/src/main/groovy/io/realm/transformer/RealmOptionalAPITransformer.groovy
```diff
@@ -0,0 +1,183 @@
...
+    }
+
+    private static Set<CtMethod> getPackageMethods(ClassPool classPool, Set<String> classNames, String packageName) {
+        Logger logger = LoggerFactory.getLogger('realm-logger')
+        logger.info "getPackageMethods started."
+        Set<CtMethod> methods = new HashSet<CtMethod>();
+        def pkgNameQuote = Pattern.quote(packageName)
+        def pattern = Pattern.compile("^$pkgNameQuote\\.[^\\.]*\$")
+        logger.info "$pattern"
+        classNames.findAll {
```
emanuelez: use `filter` instead

realm-transformer/src/main/groovy/io/realm/transformer/RealmOptionalAPITransformer.groovy
```diff
@@ -0,0 +1,183 @@
...
+            return false
+        }.each {
+            classPool.get(it).getDeclaredMethods().each {
+                def optionalAPIAnnotation = (OptionalAPI)it.getAnnotation(OptionalAPI.class)
+                if (optionalAPIAnnotation == null) {
+                    logger.info "$it.declaringClass.name $it.name doesn't have @OptionalAPI annotation"
+                } else if (optionalAPIAnnotation.dependencies().size() == 0) {
+                    throw new RuntimeException("$it.name doesn't have proper dependencies.")
+                } else if (!refClassNames.containsAll(optionalAPIAnnotation.dependencies())) {
+                    // Doesn't have enough dependencies, remove the API
```
kneth: Do you mean that nothing depends on it?

realm/realm-library/build.gradle
```diff
@@ -33,8 +34,10 @@ android {
 }
 
 import io.realm.transformer.RealmTransformer
+//import io.realm.transformer.RealmOptionalAPITransformer
```
kneth: Should it be removed?

realm/realm-library/build.gradle
```diff
@@ -8,6 +8,7 @@ apply plugin: 'com.jfrog.bintray'
 apply plugin: 'findbugs'
 apply plugin: 'pmd'
 apply plugin: 'checkstyle'
+//apply plugin: 'maven-publish'
```
kneth: Should it be removed?

gradle-plugin/src/main/groovy/io/realm/gradle/Realm.groovy
beeender: This just to remove the optional API as the "almost last" step before packing apk. In an android lib project this won't be needed as it will be used in a app project and that may have a required dependency.

zaki50:
I think it's a good idea! :+1:

realm-transformer/src/main/groovy/io/realm/transformer/RealmOptionalAPITransformer.groovy
beeender: That is the reference scope is used to identify if the rx.observable exists. So i think it should contain all of them?

realm-transformer/src/main/groovy/io/realm/transformer/RealmOptionalAPITransformer.groovy
beeender: It means for example, the `rx.observable` doesn't exist.

beeender:
Clean and rewrite in kotlin which is much easier (at least for me).

 Add @OptionalAPI annotation
    
    * Add @OptionalAPI to mark a method as optional with dependencies.
    * New bytecode transformer to remove an optional API if it doesn't
      have enough dependencies.
    * Modify gradle plugin.
    * Mark the rx.Observable related APIs as optional.
    
    This is not the perfect solution since the RxObservableFactory is still
    in the final result. But it is good enough for us to work around the
    issue with Jackson and other lib which needs to do reflection on the
    RealmObject.

@realm/java

beeender:
oops, kotlin files are missing ....

beeender:
retest this please

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@
 * RealmCollection and OrderedRealmCollection have been added. RealmList and RealmResults both implement these interfaces.
 * RealmBaseAdapter now accept an OrderedRealmCollection instead of only RealmResults.
 * RealmObjectSchema.isPrimaryKey(String) (#2440)
+* Removes rxjava related APIs during bytecode transforming to make RealmObject plays well with reflection when rx.Observable doesn't exist.
```
cmelchior: rxjava -> RxJava

realm-annotations/src/main/java/io/realm/annotations/internal/OptionalAPI.java
```diff
@@ -0,0 +1,16 @@
+package io.realm.annotations.internal;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * This annotation is used to mark an API as optional with one or more class {@link #dependencies()}. The bytecode
+ * transformer will decide on build time to remove the corresponding API if it doesn't full fill the dependencies.
```
cmelchior: fulfil

realm-transformer/build.gradle
```diff
@@ -1,14 +1,18 @@
 buildscript {
+    ext.kotlin_version = '1.0.1-2'
```
cmelchior: Sneaking in Kotlin through the backdoor? :)
Personally I don't have any problems with it, but we probably shouldn't mix Kotlin/Groovy if we can avoid it. We should at least have team consensus if that is the direction we want to go in for the Transform API's.

realm-transformer/src/main/kotlin/io/realm/transformer/RealmOptionalAPITransformer.kt
```diff
@@ -0,0 +1,88 @@
...
+            Scope.SUB_PROJECTS, Scope.SUB_PROJECTS_LOCAL_DEPS, Scope.EXTERNAL_LIBRARIES)
+
+    override fun transform(context: Context?,
+                           inputs: MutableCollection<TransformInput>?,
+                           referencedInputs: MutableCollection<TransformInput>?,
+                           outputProvider: TransformOutputProvider?,
+                           isIncremental: Boolean) {
+
+        val classNames = HashSet<String>()
+        inputs!!.appendAllToClassNames(classNames)
```
cmelchior: Why this? `classNames` is empty?

realm-transformer/src/main/kotlin/io/realm/transformer/RealmOptionalAPITransformer.kt
```diff
@@ -0,0 +1,88 @@
...
+        val refClassNames = HashSet<String>()
+        referencedInputs!!.appendAllToClassNames(refClassNames)
```
cmelchior: Same, `refClassNames` is empty

realm-transformer/build.gradle
beeender: I would love to change the groovy as well. Those two transformers share some common functions which can be refactored. Anyone loves groovy more? @emanuelez ?

realm-transformer/src/main/kotlin/io/realm/transformer/RealmOptionalAPITransformer.kt
```diff
@@ -0,0 +1,88 @@
...
+        val classPool = createClassPool()
+        inputs.appendAllToClassPool(classPool)
+
+        classNames.filter { it.startsWith("io.realm.") }.forEach {
+            classPool.get(it).declaredMethods.forEach {
+                val optionalAPIAnnotation = it.getAnnotation(OptionalAPI::class.java) as? OptionalAPI
+                val dependenciesList = optionalAPIAnnotation?.dependencies?.toList()
+
+                if (optionalAPIAnnotation == null) {
+                    logger.info("${it.declaringClass.name} ${it.name} doesn't have @OptionalAPI annotation.")
```
cmelchior: Isn't this more debug info? It seems a bit harsh to log to info all the classes that doesn't have the annotation?

realm-transformer/src/main/kotlin/io/realm/transformer/RealmOptionalAPITransformer.kt
beeender: `inputs` is from java method which declared as optional, but it will never be `null`. Safe to cast by `!!`

realm-transformer/src/main/kotlin/io/realm/transformer/RealmOptionalAPITransformer.kt
```diff
@@ -0,0 +1,88 @@
...
+                } else if (dependenciesList == null || dependenciesList.size == 0) {
+                    throw IllegalArgumentException("${it.name} doesn't have proper dependencies: " +
```
cmelchior: Why throw the error if the dependency list is empty? Doesn't that just imply we should remove the method?

realm-transformer/src/main/kotlin/io/realm/transformer/RealmOptionalAPITransformer.kt
```diff
@@ -0,0 +1,88 @@
...
+                            "${optionalAPIAnnotation.dependencies}.")
+                } else if (!refClassNames.containsAll(dependenciesList)) {
+                    // Doesn't have enough dependencies, remove the API
+                    logger.info("${it.declaringClass.name} ${it.name} will be removed since some of the dependencies " +
+                            "in $dependenciesList don't exist.")
+                    it.declaringClass.removeMethod(it)
+                } else {
+                    logger.info("${it.declaringClass.name} ${it.name} has all dependencies in $dependenciesList.")
```
cmelchior: debug instead of info?

realm-transformer/src/main/kotlin/io/realm/transformer/RealmOptionalAPITransformer.kt
beeender: If we have something like `@OptionalAPI(dependencies = null)`, it throws. I don't think there is a reason we should mark an API like that.

realm-transformer/src/main/kotlin/io/realm/transformer/RealmOptionalAPITransformer.kt
```diff
@@ -0,0 +1,88 @@
...
+                }
+            }
+        }
+
+        // Create outputs
+        classNames.forEach {
+            val ctClass = classPool.getCtClass(it)
+            ctClass.writeFile(outputProvider!!.getContentLocation(
+                    "realm-optional-api", inputTypes, scopes, Format.DIRECTORY).canonicalPath)
```
cmelchior: Isn't "realm-optional-api" a constant? So perhaps define it at the top of the file. Kotlin doesn't really support static fields but perhaps just `val NAME = "realm-optional-api"` ? Not entirely sure what would be idiomatic there.

realm-transformer/src/main/kotlin/io/realm/transformer/Utils.kt
```diff
@@ -0,0 +1,73 @@
...
+
+import com.android.SdkConstants
+import com.android.build.api.transform.TransformInput
+import javassist.ClassPool
+import javassist.LoaderClassPath
+import java.io.File
+import java.util.jar.JarFile
+
+// Find all classes and append their names to the given String collection
+fun MutableCollection<TransformInput>.appendAllToClassNames(classNames: MutableCollection<String>) {
```
cmelchior: Minor thing: I am not entirely sure if a best practise for extension methods have emerged yet, but putting them in a class called `Util` does seems a bit weird to me especially if you add a lot for many different classes.. Would a `MutableCollectionExt` class make more sense?

realm-transformer/src/main/kotlin/io/realm/transformer/Utils.kt
```diff
@@ -0,0 +1,73 @@
...
+    for (item in this) {
```
cmelchior: I know that we have type inference for this, but when reviewing code it is still nice if the type is part of the declaration.

realm-transformer/src/main/kotlin/io/realm/transformer/Utils.kt
beeender: Yup, I did struggle with that as well. Will put it to `MutableCollectionExt` and let's make it as rule for our kotlin code-style.

realm-transformer/src/main/kotlin/io/realm/transformer/Utils.kt
```diff
@@ -0,0 +1,73 @@
...
+        for (input in item.directoryInputs) {
+            val dirPath = input.file.absolutePath
+            input.file.walkTopDown().filter { it.isFile && it.endsWith(SdkConstants.DOT_CLASS) }.forEach {
+                val className = it.absolutePath.substring(
+                        dirPath.length + 1,
+                        input.file.absolutePath.length - SdkConstants.DOT_CLASS.length)
+                        .replace(File.separatorChar, '.')
```
cmelchior: Cant this fit on the above line? Breaking like this looks a bit funny

cmelchior: Also didn't we just fix a windows bug about this?

realm-transformer/src/main/kotlin/io/realm/transformer/Utils.kt
```diff
@@ -0,0 +1,73 @@
...
+import java.io.File
+import java.util.jar.JarFile
+
+// Find all classes and append their names to the given String collection
+fun MutableCollection<TransformInput>.appendAllToClassNames(classNames: MutableCollection<String>) {
+    for (item in this) {
+        for (input in item.directoryInputs) {
+            val dirPath = input.file.absolutePath
+            input.file.walkTopDown().filter { it.isFile && it.endsWith(SdkConstants.DOT_CLASS) }.forEach {
+                val className = it.absolutePath.substring(
```
cmelchior: I would probably use `qualifiedClassName` instead?

realm-transformer/src/main/kotlin/io/realm/transformer/Utils.kt
beeender: That is only in jar, in `dirpath` it think it's still safe to use `File.separatorChar`.

realm-transformer/src/main/kotlin/io/realm/transformer/RealmOptionalAPITransformer.kt
cmelchior: Ah yes sorry, I think the method name confused me. `appendAll` would normally apply to `inputs`, but in this case it actually the other way around. Would it make more sense to put the extension method on the HashSet? Or find a better name ( I don't have anything better though).

cmelchior:
Love the idea, but we need consensus on the Kotlin approach and code style before merging :+1:

emanuelez:
retest this please

emanuelez:
retest this please

realm-transformer/src/main/kotlin/io/realm/transformer/Utils.kt
cmelchior: 👍

realm-transformer/src/main/kotlin/io/realm/transformer/RealmOptionalAPITransformer.kt
cmelchior: 👍

beeender:
@realm/java

cmelchior:
👍

realm-transformer/src/main/kotlin/io/realm/transformer/RealmOptionalAPITransformer.kt
```diff
@@ -0,0 +1,92 @@
...
+                val dependenciesList = optionalAPIAnnotation?.dependencies?.toList()
+
+                if (optionalAPIAnnotation == null) {
+                    logger.debug("${it.declaringClass.name} ${it.name} doesn't have @OptionalAPI annotation.")
+                } else if (dependenciesList == null || dependenciesList.size == 0) {
+                    throw IllegalArgumentException("${it.name} doesn't have proper dependencies: " +
+                            "${optionalAPIAnnotation.dependencies}.")
+                } else if (!refClassNames.containsAll(dependenciesList)) {
+                    // Doesn't have enough dependencies, remove the API
+                    logger.debug("${it.declaringClass.name} ${it.name} will be removed since some of the dependencies " +
```
kneth: "don't" -> "doesn't"

kneth:
:+1:

zaki50:
copying resource files seems to be missing.

see https://github.com/realm/realm-java/pull/2358/files

beeender:
@zaki50 I did some testing and I don't think this transformer needs to copy resource files. #2358 seems to be useful when it is a library project. But this transfer will only run for the app project.

beeender:
@emanuelez Shall I add a task in the root `build.gradle` to run `connectedCheck` for `integration-tests`?

zaki50:
@beeender As long as kotlin, that's true since generated resource is only used in compile time.
But there is a chance to exist other gradle plugin which generates resource files used at runtime.

zaki50:
@beeender But I don't know the plugin which behave like that.

I think we can ignore that possibility for now.

cmelchior:
I like the idea of integration tests, but they sure do pull in a lot of files and project setup. 

Some comments:

- Couldn't we just solve this through unit tests? e.g. verifying that the transformer works if used? Do we really need a full integration test suite for it?

- If yes: Do we want it as a root folder or under `realm`. I could see arguments both ways?
- If yes: Please clean out any files not needed: Logos / resource files.

beeender:
> Couldn't we just solve this through unit tests? e.g. verifying that the transformer works if used? Do we really need a full integration test suite for it?

Unit tests will be super difficult to do like check if the dependency exist, or tests needed for lib project. I think we could also benefit from those tests on issues like https://github.com/realm/realm-java/pull/2358

> If yes: Do we want it as a root folder or under realm. I could see arguments both ways?

Any suggestions? I think it would also work as a sub-module of `realm` project. i will try

> If yes: Please clean out any files not needed: Logos / resource files.

Will do it anyway.

beeender:
Use #2727 instead, to the `releases` branch.

PR #2406, opened 2016-03-09
transform accesses to model fields in library projects

zaki50:
before: modify the access to fields defined in models only in target project.
after: modify the access to fields defined in all models including ones in dependencies.

Suppose `Cow` is defined in target(app) project, and `Cat` is defined in dependency project.

```
        Cow cow = new Cow();
        cow.setName("mooo");
        realm.beginTransaction();
        cow = realm.copyToRealm(cow);
        realm.commitTransaction();

        Cat cat = new Cat();
        cat.setName("doraemon");
        realm.beginTransaction();
        cat = realm.copyToRealm(cat);
        realm.commitTransaction();

        Log.d("FOO", "cow.getName(): " + cow.getName());
        Log.d("FOO", "cow.name: " + cow.name);
        Log.d("FOO", "cat.getName(): " + cat.getName());
        Log.d("FOO", "cat.name: " + cat.name);
```

before: 
```
FOO     : cow.getName(): mooo
FOO     : cow.name: mooo
FOO     : cat.getName(): doraemon
FOO     : cat.name: null
```

after:
```
FOO     : cow.getName(): mooo
FOO     : cow.name: mooo
FOO     : cat.getName(): doraemon
FOO     : cat.name: doraemon
```

please review this @realm/java @emanuelez

### DIFF
--- a/realm-transformer/src/main/groovy/io/realm/transformer/RealmTransformer.groovy
+++ b/realm-transformer/src/main/groovy/io/realm/transformer/RealmTransformer.groovy
@@ -15,6 +15,7 @@
  */
 
 package io.realm.transformer
+
 import com.android.SdkConstants
 import com.android.build.api.transform.*
 import com.google.common.collect.ImmutableSet
@@ -28,9 +29,11 @@ import org.slf4j.Logger
 import org.slf4j.LoggerFactory
 
 import java.lang.reflect.Modifier
+import java.util.jar.JarFile
 import java.util.regex.Pattern
 
 import static com.android.build.api.transform.QualifiedContent.*
+
 /**
  * This class implements the Transform API provided by the Android Gradle plugin.
  */
@@ -56,7 +59,8 @@ class RealmTransformer extends Transform {
 
     @Override
     Set<Scope> getReferencedScopes() {
-        return Sets.immutableEnumSet(Scope.EXTERNAL_LIBRARIES, Scope.TESTED_CODE)
+        return Sets.immutableEnumSet(Scope.EXTERNAL_LIBRARIES, Scope.PROJECT_LOCAL_DEPS,
+                Scope.SUB_PROJECTS, Scope.SUB_PROJECTS_LOCAL_DEPS, Scope.TESTED_CODE)
     }
 
     @Override
@@ -72,7 +76,9 @@ class RealmTransformer extends Transform {
         def tic = System.currentTimeMillis()
 
         // Find all the class names
-        def classNames = getClassNames(inputs)
+        def inputClassNames = getClassNames(inputs)
+        def referencedClassNames = getClassNames(referencedInputs)
+        def allClassNames = merge(inputClassNames, referencedClassNames);
 
         // Create and populate the Javassist class pool
         ClassPool classPool = createClassPool(inputs, referencedInputs)
@@ -82,7 +88,7 @@ class RealmTransformer extends Transform {
         // mark as transformed
         def baseProxyMediator = classPool.get('io.realm.internal.RealmProxyMediator')
         def mediatorPattern = Pattern.compile('^io\\.realm\\.[^.]+Mediator$')
-        def proxyMediatorClasses = classNames
+        def proxyMediatorClasses = inputClassNames
                 .findAll { it.matches(mediatorPattern) }
                 .collect { classPool.getCtClass(it) }
                 .findAll { it.superclass?.equals(baseProxyMediator) }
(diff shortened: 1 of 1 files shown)


realm-transformer/src/main/groovy/io/realm/transformer/RealmTransformer.groovy
```diff
@@ -93,33 +99,36 @@ class RealmTransformer extends Transform {
...
                 it.getAnnotation(Ignore.class) == null && !Modifier.isStatic(it.getModifiers())
             })
         }
-        logger.info "Managed Fields: ${managedFields*.name}"
+        logger.info "Managed Fields: ${allManagedFields*.name}"
 
         // Add accessors to the model classes
-        modelClasses.each {
+        allModelClasses.each {
+            if (!inputClassNames.contains(it.name)) {
```
cmelchior: Any reason we are not just looping over inputClassNames instead?

cmelchior:
Apart from the comment I think it looks good :+1:

realm-transformer/src/main/groovy/io/realm/transformer/RealmTransformer.groovy
zaki50: `inputClassNames` contains all classes not only model classes but normal Java classes such as `MainActivity`.
We need to iterate only model classes defined in target project.

Of course we can prepare `inputModelClasses`, but it does not improve anything.

dalinaum:
LGTM :+1:

kneth:
Nice catch! :+1:

realm-transformer/src/main/groovy/io/realm/transformer/RealmTransformer.groovy
```diff
@@ -93,33 +99,36 @@ class RealmTransformer extends Transform {
...
                 it.getAnnotations()
                 it.getAnnotation(Ignore.class) == null && !Modifier.isStatic(it.getModifiers())
             })
         }
-        logger.info "Managed Fields: ${managedFields*.name}"
+        logger.info "Managed Fields: ${allManagedFields*.name}"
 
         // Add accessors to the model classes
-        modelClasses.each {
+        allModelClasses.each {
```
beeender: Wouldn't iterating `inputClassNames` be enough? I think even we manipulate the reference classes, the results won't take any effect.

realm-transformer/src/main/groovy/io/realm/transformer/RealmTransformer.groovy
```diff
@@ -56,7 +59,8 @@ class RealmTransformer extends Transform {
 
     @Override
     Set<Scope> getReferencedScopes() {
-        return Sets.immutableEnumSet(Scope.EXTERNAL_LIBRARIES, Scope.TESTED_CODE)
+        return Sets.immutableEnumSet(Scope.EXTERNAL_LIBRARIES, Scope.PROJECT_LOCAL_DEPS,
```
emanuelez: Are you sure you have to modify `getReferencedScopes()`? I would think we would have to modify `getScopes()` instead.

zaki50: I think we don't need to modify `getScope()` since we modify only classes in target project.

By modifying `getReferencedScope()`, we can collect all model classes in whole application.

realm-transformer/src/main/groovy/io/realm/transformer/RealmTransformer.groovy
zaki50: @beeender Is that comment the same with the one below from Christian?

beeender: Uh, my bad. `ClassPool classPool = createClassPool(inputs, referencedInputs)` actually contains both `input` and `inputReference`, and we actually don't have to modify classes in the `inputReference`. Would it be better we split the input and reference input to different `ClassPool`? But i can accept current implementation.

zaki50: I don't think we can split the `ClassPool` since it is not just a `Set` of classes. `ClassPool` must contain all class paths to load classes in application project.

realm-transformer/src/main/groovy/io/realm/transformer/RealmTransformer.groovy
zaki50: And we need all model classes and fields in them later since we must detect the accesses to all fields.

realm-transformer/src/main/groovy/io/realm/transformer/RealmTransformer.groovy
beeender: :+1:

realm-transformer/src/main/groovy/io/realm/transformer/RealmTransformer.groovy
zaki50: @cmelchior @beeender I introduced `inputModelClasses` to improve readability.

realm-transformer/src/main/groovy/io/realm/transformer/RealmTransformer.groovy
emanuelez: True

beeender:
:+1:

realm-transformer/src/main/groovy/io/realm/transformer/RealmTransformer.groovy
```diff
@@ -183,8 +192,25 @@ class RealmTransformer extends Transform {
                     }
                 }
             }
-        }
 
+            it.jarInputs.each {
+                def jarFile = new JarFile(it.file)
+                def entries = jarFile.entries()
+                entries.each {
```
emanuelez: This could be more compact and readable using `filter` IMO:

```java
entries
   .filter {
      !it.directory && it.path.endsWith(SdkConstants.DOT_CLASS)
   }
   .each {
      def className = path.substring(0, path.length() - SdkConstants.DOT_CLASS.length()).replace(File.separatorChar, '.' as char)
      classNames.add(className)
   }
```

emanuelez:
Minor comment, otherwise :+1:

realm-transformer/src/main/groovy/io/realm/transformer/RealmTransformer.groovy
zaki50: `filter` method does not seem to exist.
```
Error:Execution failed for task ':moduleExample:library:transformClassesWithRealmTransformerForRelease'.
> No signature of method: java.util.jar.JarFile$JarEntryIterator.filter() is applicable for argument types: (io.realm.transformer.RealmTransformer$_getClassNames_closure14$_closure22$_closure24) values: [io.realm.transformer.RealmTransformer$_getClassNames_closure14$_closure22$_closure24@64ebbf79]
```

zaki50: I updated the code. 

```
                jarFile.entries().findAll {
                    !it.directory && it.name.endsWith(SdkConstants.DOT_CLASS)
                }.each {
                    def path = it.name
                    def className = path.substring(0, path.length() - SdkConstants.DOT_CLASS.length())
                            .replace(File.separatorChar, '.' as char)
                    classNames.add(className)
                }
```

emanuelez:
:+1:

PR #2426, opened 2016-03-11
Fix potential race condition causing BadVersion

### LINKED ISSUE #2417: std::exception in io_realm_internal_TableQuery.cpp line 1682 when calling RealmResults.load() with multiple Realm instances
#### Steps to reproduce

- Obtain an instance of `Realm`
- Execute an async transaction on it
- Obtain a second instance of `Realm` with same configuration and on the same thread (it will return the same instance as the first `Realm`)
- Execute an async query
- Call `load()` to synchronously wait for the query to execute 

#### Issue

The `load()` method throws a native exception: `std::exception in io_realm_internal_TableQuery.cpp line 1682`. this exception is swallowed on the Java side, but I managed to debug it and I discovered it's actually a `BadVersionException`.
After some investigation I think the issue is in the invocation of `Realm.getInstance()` after submitting the first transaction. If that happens before (or if I reuse the same `Realm` variable) everything works correctly.
I already discussed with @cmelchior on Slack about the issue.

#### Code Sample
(in Kotlin :stuck_out_tongue: )
```kotlin
val realm = Realm.getInstance(realmConfiguration)
realm.executeTransactionAsync {
    it.deleteAll()
}
val realm2 = Realm.getInstance(realmConfiguration)
realm2.where(RealmDocument::class.java).findAllAsync().load()
```

Realm version(s): 0.88.0

Android Studio version: 2.0

Which Android version and device: Emulator with 6.0

cmelchior:
Fixes #2417


cmelchior:
I believe this PR should fix the race condition, but we are now getting a segmentation fault in core in `TableQuery.nativeImportHandoverTableViewIntoSharedGroup` which baffles me a bit :unamused:

realm/realm-jni/src/io_realm_internal_TableQuery.cpp
```diff
@@ -1037,24 +1037,26 @@ JNIEXPORT jlong JNICALL Java_io_realm_internal_TableQuery_nativeFind(
...
-        SG(bgSharedGroupPtr)->begin_read();
-    }
-
     std::unique_ptr<Query> query = SG(bgSharedGroupPtr)->import_from_handover(std::move(handoverQuery));
-    if (advanceToLatestVersion && isDifferentVersions) {
+    if (advanceToLatestVersion) {
         LangBindHelper::advance_read(*SG(bgSharedGroupPtr));
+    } else {
+        try {
+            LangBindHelper::advance_read(*SG(bgSharedGroupPtr), handoverQuery->version);
```
finnschiermer: At this point in time, it may not be safe to access the handoverQuery. It is moved() at line 1045, so the access to version here is not well-defined.

cmelchior: Ah yes, good point

cmelchior: Unfortunately that didn't seem to the root cause of the crash

nhachicha:
updated the PR please review @cmelchior @tgoyne @finnschiermer

realm/realm-jni/src/io_realm_internal_TableQuery.cpp
```diff
@@ -1037,23 +1037,23 @@ JNIEXPORT jlong JNICALL Java_io_realm_internal_TableQuery_nativeFind(
...
+    // if the SharedGroup is not in Read Transaction, we position it at the same version as the handover
+    if (SG(bgSharedGroupPtr)->get_transact_stage() != SharedGroup::transact_Reading) {
+        SG(bgSharedGroupPtr)->begin_read(handoverQuery->version);
 
-    SharedGroup::VersionID currentVersion = SG(bgSharedGroupPtr)->get_version_of_current_transaction();
-    bool isDifferentVersions = (currentVersion != handoverQuery->version);
-    if (isDifferentVersions) {
+    } else if (SG(bgSharedGroupPtr)->get_version_of_current_transaction() != handoverQuery->version) {
+        SG(bgSharedGroupPtr)->end_read();
         SG(bgSharedGroupPtr)->begin_read(handoverQuery->version);
```
cmelchior: This line can still cause a BadVersion which is currently not getting caught and will crash. 
Also, right now our executor service has more than 1 thread, by restricting it to 1 thread It seems we can control the concurrency enough for a stable unit tests.

I have an idea for fixing it + actually making a proper unit test for this. Update incoming

cmelchior:
@nhachicha I updated the PR. We didn't properly handle BadVersionException in our findAllAsync methods from what I could tell. In a lot of cases we treated that exception as fatal (even though a retry was imminent). I think I fixed it, but still trying to wrap my head around testing it in a sane way. So far I have done manual testing but stubbing out values.

Please review, especially around retrying and batch updates.

cmelchior:
Closing this PR as it should be merged into `releases` instead.

jcj52436999:
This message was created automatically by mail delivery software.

A message that you sent could not be delivered to one or more of its
recipients. This is a temporary error. The following address(es) deferred:

  jcjacksonengineering@gmail.com
    Domain joe-jackson.net has exceeded the max emails per hour (101/100 (101%)) allowed.  Message will be reattempted later

------- This is a copy of the message, including all the headers. ------
Received: from github-smtp2-ext1.iad.github.net ([192.30.252.192]:53611 helo=github-smtp2a-ext-cp1-prd.iad.github.net)
	by rs5.websitehostserver.net with esmtps (TLSv1.2:ECDHE-RSA-AES256-GCM-SHA384:256)
	(Exim 4.86_1)
	(envelope-from <noreply@github.com>)
	id 1afmIx-001w0H-88
	for jcjackson@joe-jackson.net; Tue, 15 Mar 2016 06:34:23 -0400
Date: Sun, 13 Mar 2016 15:14:53 -0700
DKIM-Signature: v=1; a=rsa-sha256; c=relaxed/relaxed; d=github.com;
	s=pf2014; t=1457907293;
	bh=B4E5IMH1d8hKnaUh3tuakjox6lbrGaZ+sz+h02bMWBk=;
	h=From:Reply-To:To:In-Reply-To:References:Subject:List-ID:
	 List-Archive:List-Post:List-Unsubscribe:From;
	b=W3CQVkLf3kd75gt3AdXGH4OIPG3pGsAl4DUY+GuaHRLVY+BKmu4Wm2GMFbxtm6Qcw
	 Sfvb22kd8UnA0EDjjWZ1k3INb9qUc/u5XlSpEhYz6uHKFDzaByOeyCbKcnu+hROxPo
	 5br02SI41HwdCeuPCdh+XZwE2vwC2zjo8jyy53K8=

PR #2441, opened 2016-03-14
Fix the String from DynamicRealmObject.toString().

### LINKED ISSUE #2439: DynamicRealmObject.toString() does not honor null values
#### Goal

* `DynamicRealmObject.toString()` shows null value as "null".
* Remove "class_" prefix from type name in that String.
* Align the format to the String from typed `RealmObject`.

#### Expected Results

```
Nullables = [{str:null},{bool:null},{num:null},{time:null},{bytes:null}]
```

#### Actual Results

```
class_Nullables = [{str: null}, {bool: false}, {num: 0}, {time: Thu Jan 01 09:00:00 GMT+09:00 1970}, {bytes: null}]
```

#### Code Sample

```java
public class Nullables extends RealmObject {

    private String str;
    private Boolean bool;
    private Integer num;
    private Date time;
    private byte[] bytes;

    // getters and setters
}
```

```java
Realm realm = Realm.getInstance(new RealmConfiguration.Builder(this).deleteRealmIfMigrationNeeded().build());
try {
    final Nullables nullables = new Nullables();
        @Override
            realm.deleteAll();
            realm.copyToRealm(nullables);
        }
    });
} finally {
    realm.close();
}

final DynamicRealm dynamicRealmBefore = DynamicRealm.getInstance(new RealmConfiguration.Builder(this).build());
try {
    System.out.println(dynamicRealmBefore.allObjects("Nullables").first().toString());
} finally {
    dynamicRealmBefore.close();
}
```

Realm version(s): 0.88.0

Android Studio version: 2.1 Preview1

Which Android version and device: N preview1

zaki50:
Now DynamicRealmObject.toString() shows null value as "null".
"class_" prefix is removed from type name in that String.
Align the format to the String from typed `RealmObject`.

This fixes #2439 

### DIFF
--- a/realm/realm-library/src/main/java/io/realm/DynamicRealmObject.java
+++ b/realm/realm-library/src/main/java/io/realm/DynamicRealmObject.java
@@ -683,35 +683,52 @@ public final class DynamicRealmObject extends RealmObject {
         if (row == null || !row.isAttached()) {
             return "Invalid object";
         }
-        StringBuilder sb = new StringBuilder(row.getTable().getName() + " = [");
+
+        final String className = Table.tableNameToClassName(row.getTable().getName());
+        StringBuilder sb = new StringBuilder(className + " = [");
         String[] fields = getFieldNames();
         for (String field : fields) {
             long columnIndex = row.getColumnIndex(field);
             RealmFieldType type = row.getColumnType(columnIndex);
             sb.append("{");
+            sb.append(field).append(":");
             switch (type) {
-                case BOOLEAN: sb.append(field).append(": ").append(row.getBoolean(columnIndex)); break;
-                case INTEGER: sb.append(field).append(": ").append(row.getLong(columnIndex)); break;
-                case FLOAT: sb.append(field).append(": ").append(row.getFloat(columnIndex)); break;
-                case DOUBLE: sb.append(field).append(": ").append(row.getDouble(columnIndex)); break;
-                case STRING: sb.append(field).append(": ").append(row.getString(columnIndex)); break;
-                case BINARY: sb.append(field).append(": ").append(Arrays.toString(row.getBinaryByteArray(columnIndex))); break;
-                case DATE: sb.append(field).append(": ").append(row.getDate(columnIndex)); break;
+                case BOOLEAN:
+                    sb.append(row.isNull(columnIndex) ? "null" : row.getBoolean(columnIndex));
+                    break;
+                case INTEGER:
+                    sb.append(row.isNull(columnIndex) ? "null" : row.getLong(columnIndex));
+                    break;
+                case FLOAT:
+                    sb.append(row.isNull(columnIndex) ? "null" : row.getFloat(columnIndex));
+                    break;
+                case DOUBLE:
+                    sb.append(row.isNull(columnIndex) ? "null" : row.getDouble(columnIndex));
+                    break;
+                case STRING:
+                    sb.append(row.getString(columnIndex));
+                    break;
+                case BINARY:
+                    sb.append(Arrays.toString(row.getBinaryByteArray(columnIndex)));
+                    break;
+                case DATE:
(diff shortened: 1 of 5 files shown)


realm/realm-library/src/main/java/io/realm/DynamicRealmObject.java
```diff
@@ -683,35 +683,35 @@ public String toString() {
...
             sb.append("{");
             switch (type) {
-                case BOOLEAN: sb.append(field).append(": ").append(row.getBoolean(columnIndex)); break;
-                case INTEGER: sb.append(field).append(": ").append(row.getLong(columnIndex)); break;
-                case FLOAT: sb.append(field).append(": ").append(row.getFloat(columnIndex)); break;
-                case DOUBLE: sb.append(field).append(": ").append(row.getDouble(columnIndex)); break;
-                case STRING: sb.append(field).append(": ").append(row.getString(columnIndex)); break;
-                case BINARY: sb.append(field).append(": ").append(Arrays.toString(row.getBinaryByteArray(columnIndex))); break;
-                case DATE: sb.append(field).append(": ").append(row.getDate(columnIndex)); break;
+                case BOOLEAN: sb.append(field).append(":").append(row.isNull(columnIndex) ? "null" : row.getBoolean(columnIndex)); break;
```
beeender: I think it is too long for one line... use the normal format of `switch ... case` block?

realm/realm-library/src/main/java/io/realm/internal/Table.java
```diff
@@ -579,11 +579,8 @@ public long getPrimaryKey() {
                 return NO_PRIMARY_KEY; // Free table = No primary key
             }
 
-            String tableName = getName();
-            if (tableName.startsWith(TABLE_PREFIX)) {
-                tableName = tableName.substring(TABLE_PREFIX.length());
-            }
-            long rowIndex = pkTable.findFirstString(PRIMARY_KEY_CLASS_COLUMN_INDEX, tableName);
+            String className = tableNameToClassName(getName());
+            long rowIndex = pkTable.findFirstString(PRIMARY_KEY_CLASS_COLUMN_INDEX, className);
```
beeender: @realm/java does anyone remember why do we have `if (tableName.startsWith(TABLE_PREFIX))` before?

dalinaum:
:+1:

realm/realm-library/src/main/java/io/realm/internal/Table.java
zaki50: @beeender related to https://github.com/realm/realm-java/issues/1703?

beeender: @zaki50 That only impacts on pkTable, right? So it is not necessary to be in `tableNameToClassName`

realm/realm-library/src/main/java/io/realm/DynamicRealmObject.java
zaki50: done

realm/realm-library/src/main/java/io/realm/internal/Table.java
zaki50: The name of the all tables created from model classes have "class_" prefix and the model name in pkTable does not have that prefix.

I don't think we can find correct `rowIndex` if we omit the `tableNameToClassName()` here.

beeender: :+1:

beeender:
Only thing is I don't quite understand why do we need `if (!tableName.startsWith(Table.TABLE_PREFIX))` checking in the `tableNameToClassName()` now and before. But it seems it is not harmful? 

:+1:

cmelchior:
:+1:

zaki50:
@beeender I'd like to leave this as it is since it's `releases` branch. It's better to fix that in `master` branch if needed.

beeender:
Yeah, no problem

stk1m1:
:+1:

PR #2479, opened 2016-03-16
changed to use 'advance_read()' instead of 'end_read()' and 'begin_read()' when possible.

### LINKED ISSUE #2470: We can reduce the chance of `BadVersionException` in some cases
https://github.com/realm/realm-java/blob/5444c96b8bec2a6537f30401cca3b452aab1a42a/realm/realm-jni/src/io_realm_internal_TableQuery.cpp#L1055

https://github.com/realm/realm-java/blob/5444c96b8bec2a6537f30401cca3b452aab1a42a/realm/realm-jni/src/io_realm_internal_TableQuery.cpp#L1166

```
    } else if (sg->get_version_of_current_transaction() != handoverQuery->version) {
        sg->end_read();
        sg->begin_read(handoverQuery->version);
    }
```

The code above can be:

```
    } else if (sg->get_version_of_current_transaction() < handoverQuery->version) {
        sg->advance_read(handoverQuery->version);
    } else if (handoverQuery->version < sg->get_version_of_current_transaction()) {
        sg->end_read();
        sg->begin_read(handoverQuery->version);
    }
```

By avoiding to call `end_read()`, we can reduce the chance of `BadVersionException`.

zaki50:
This fixes #2470

### DIFF
--- a/realm/realm-jni/src/io_realm_internal_TableQuery.cpp
+++ b/realm/realm-jni/src/io_realm_internal_TableQuery.cpp
@@ -1051,7 +1051,11 @@ std::unique_ptr<Query> handoverQueryToWorker(jlong bgSharedGroupPtr, jlong query
     if (sg->get_transact_stage() != SharedGroup::transact_Reading) {
         // if the SharedGroup is not in Read Transaction, we position it at the same version as the handover
         sg->begin_read(handoverQuery->version);
-    } else if (sg->get_version_of_current_transaction() != handoverQuery->version) {
+    } else if (sg->get_version_of_current_transaction() < handoverQuery->version) {
+        // going forward
+        LangBindHelper::advance_read(*sg, handoverQuery->version);
+    } else if (handoverQuery->version < sg->get_version_of_current_transaction()) {
+        // going backward
         sg->end_read();
         sg->begin_read(handoverQuery->version);
     }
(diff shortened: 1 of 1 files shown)


realm/realm-jni/src/io_realm_internal_TableQuery.cpp
```diff
@@ -1051,7 +1051,11 @@ std::unique_ptr<Query> handoverQueryToWorker(jlong bgSharedGroupPtr, jlong query
     if (sg->get_transact_stage() != SharedGroup::transact_Reading) {
         // if the SharedGroup is not in Read Transaction, we position it at the same version as the handover
         sg->begin_read(handoverQuery->version);
-    } else if (sg->get_version_of_current_transaction() != handoverQuery->version) {
+    } else if (sg->get_version_of_current_transaction() < handoverQuery->version) {
+        // going forward
+        LangBindHelper::advance_read(*sg, handoverQuery->version);
+    } else if (handoverQuery->version < sg->get_version_of_current_transaction()) {
+        // going backward
```
zaki50: I think we should not go backward and it's enough to leave the version as it is.

What do you think? @nhachicha @realm/java

cmelchior: I'll need to some time to think through the flow of this, especially regarding where the SharedGroup is actually created in the flow.

beeender:
Any test result to show how much it reduce the possibility of generating bad version exception?

zaki50:
@beeender I'll try to make the testcase.

realm/realm-jni/src/io_realm_internal_TableQuery.cpp
nhachicha: we do need to go backward, this is most of the time what happens, when you receive a `REALM_CHANGE` an you try to update your query, obviously the UI thread delayed the SharedGroup advance, until the async queries updates. so the handed over query will most likely be behind the latest worker thread shared group

cmelchior:
@zaki50 Given the discussion above. This can just be closed, right?

zaki50:
@cmelchior yes.

PR #2483, opened 2016-03-16
Ignored field should not have accessor

### LINKED ISSUE #2478: @Ignore'd members of certain types in RealmObject produce errors
Error reported on SO http://stackoverflow.com/questions/36011077/ignored-members-of-certain-types-in-realmobject-produce-errors

Consider the following code:

```java
    // ...
    import org.json.JSONObject;
    import android.graphics.Bitmap;
    // ...
    
    public class someObject extends RealmObject
    {
    	@PrimaryKey
    	private String token;
    	private byte[] data;
    	private long updated;
    
    	@Ignore
    	private jsonBuilder jbRoot;
    
    //	@Ignore
    //	private JSONObject JSON; // causes compiling error once uncommented
    //  @Ignore
    //  private Bitmap bmp; // also causes the error
```

Prior to 0.88.0 the code above compiled perfectly. It would not compile on both 0.88.0 & 0.88.1 if I uncommented the ignored JSONObject and/or Bitmap:

```
    Error:Execution failed for task     ':<package>:transformClassesWithRealmTransformerForDebug'.
    > javassist.CannotCompileException: cannot find org.json.JSONObject
```

and

```
    Error:Execution failed for task     ':<package>:transformClassesWithRealmTransformerForDebug'.
    > javassist.CannotCompileException: cannot find android.graphics.Bitmap
```

respectively.

But it accepts the ignored private member of one of my own types as shown with the `jsonBuilder` object.

Build script in case anyone needs it:

```
        buildscript {
            repositories {
                jcenter()
            }
            dependencies {
                classpath "io.realm:realm-gradle-plugin:0.88.1" // tried 0.88.0 as well
            }
        }
        apply plugin: 'realm-android'
```

Thank you. The error above is not actually a blocker but Im just curious why that happens. But I like the 0.88.x changes. :)

beeender:
Fix #2478


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 ### Bug fixes
 
 * Race condition causing BadVersionException when running multiple async writes and queries at the same time (#2021/#2391/#2417).
+* Field annotated with @Ignored should have accessor generated by transformer (#2478).
```
cmelchior: should _not_ ;)

cmelchior: This should also be under 0.88.3, no?

cmelchior:
Apart from comment about changelog :+1:

CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## 0.88.3
+
+### Bug fixes
+
+* Field annotated with @Ignored should not have accessor generated by transformer (#2478).
```
zaki50: `accessors`?

zaki50:
Apart from comment about changelog :+1:

dalinaum:
:+1:

PR #2493, opened 2016-03-17
RealmResults/Objects are no longer accidentially GC'ed

### LINKED ISSUE #2485: RealmResults::asObservable not always returns a value when subscription is not kept in memory
#### Goal

Do a lot of simple realm queries with `findAllAsync().asObservable()` and have them all return a value, even without explicitly holding a reference to its subscription.

(or add a note in the docs about the importance of always holding a reference to the subscriptions? for the noobs like me...)

#### Expected Results

Whenever we create an observable with `findAllAsync().asObservable()`, it would be nice to always get a `RealmResults`, even when we don't hold a reference to the subscription.

#### Actual Results

Some observables created with `findAllAsync().asObservable()` will never call the `onNext(...)` of the subscription.

I created a small Android app showing the problem I'm facing on a bigger application.

The important code is on this line:
https://github.com/AranHase/TestFindAllAsync/blob/13ba8e68d9b9061825125ff3475c6d7091aa93bf/app/src/main/java/com/example/hasegawa/testfindallasync/MainActivity.kt#L84

In this test, one thousand "Realm observables" are created. Every time one observable is created, a counter gets incremented. For each of those observable, one subscription is created (and may or may not be explicitly stored in memory). When this subscription returns any value, another counter is incremented.

When the subscription is stored in memory, then we get the expected behavior (one thousand observables created, and one thousand results). However, when we do not store the subscription in memory, the number of results may not be one thousand.

This line determines if the subscription should be kept in memory or not, and it will change the behavior of the program:
https://github.com/AranHase/TestFindAllAsync/blob/13ba8e68d9b9061825125ff3475c6d7091aa93bf/app/src/main/java/com/example/hasegawa/testfindallasync/MainActivity.kt#L106

I also tried a `flatMap` approach in here:
https://github.com/AranHase/TestFindAllAsync/blob/13ba8e68d9b9061825125ff3475c6d7091aa93bf/app/src/main/java/com/example/hasegawa/testfindallasync/MainActivity.kt#L114

However, `flatMap` fails with: 

                                                                                       java.util.concurrent.RejectedExecutionException: Task java.util.concurrent.FutureTask@b297227 rejected from io.realm.internal.async.RealmThreadPoolExecutor@873a166[Running, pool size = 5, active threads = 5, queued tasks = 100, completed tasks = 125]
                                                                                                 at java.util.concurrent.ThreadPoolExecutor$AbortPolicy.rejectedExecution(ThreadPoolExecutor.java:2014)
                                                                                                 at java.util.concurrent.ThreadPoolExecutor.reject(ThreadPoolExecutor.java:794)
                                                                                                 at java.util.concurrent.ThreadPoolExecutor.execute(ThreadPoolExecutor.java:1340)
                                                                                                 at java.util.concurrent.AbstractExecutorService.submit(AbstractExecutorService.java:104)
                                                                                                 at io.realm.internal.async.RealmThreadPoolExecutor.submit(RealmThreadPoolExecutor.java:68)
                                                                                                 at io.realm.RealmQuery.findAllAsync(RealmQuery.java:1470)
                                                                                                 at com.example.hasegawa.testfindallasync.MainActivity$onClick$1.call(MainActivity.kt:90)
                                                                                                 at com.example.hasegawa.testfindallasync.MainActivity$onClick$1.call(MainActivity.kt:19)
                                                                                                 at rx.Observable$29.onNext(Observable.java:8031)
                                                                                                 at rx.observers.SafeSubscriber.onNext(SafeSubscriber.java:139)
                                                                                                 at rx.internal.operators.OnSubscribeRange$RangeProducer.fastpath(OnSubscribeRange.java:126)
                                                                                                 at rx.internal.operators.OnSubscribeRange$RangeProducer.request(OnSubscribeRange.java:63)
                                                                                                 at rx.Subscriber.setProducer(Subscriber.java:209)
                                                                                                 at rx.Subscriber.setProducer(Subscriber.java:205)
                                                                                                 at rx.internal.operators.OnSubscribeRange.call(OnSubscribeRange.java:38)
                                                                                                 at rx.internal.operators.OnSubscribeRange.call(OnSubscribeRange.java:26)
                                                                                                 at rx.Observable.subscribe(Observable.java:8191)
                                                                                                 at rx.Observable.subscribe(Observable.java:8158)
                                                                                                 at rx.Observable.subscribe(Observable.java:8017)
                                                                                                 at com.example.hasegawa.testfindallasync.MainActivity.onClick(MainActivity.kt:85)
                                                                                                 at android.view.View.performClick(View.java:5198)
                                                                                                 at android.view.View$PerformClick.run(View.java:21147)
                                                                                                 at android.os.Handler.handleCallback(Handler.java:739)
                                                                                                 at android.os.Handler.dispatchMessage(Handler.java:95)
                                                                                                 at android.os.Looper.loop(Looper.java:148)
                                                                                                 at android.app.ActivityThread.main(ActivityThread.java:5417)
                                                                                                 at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:726)
                                                                                                 at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:616)

Realm version(s): 0.88.1

Android Studio version: 2.1 Preview 1

Which Android version and device: Nexus 5 API 23 (emulator)

cmelchior:
Fixes #2485 

This change means that RealmResults and RealmObjects are now strongly referenced from their Observable. This was not the case previous which could cause a RealmResult to be prematurely GC'ed, causing the Observable to stop emitting.

Users are now required to unsubscribe or terminate (success/error) and Observable to free this reference. They needed to do that anyway to free the Realm instance.

@realm/java


bmunkholm:
Maybe add a bit more of the good description in this PR to the changelog?

realm/realm-library/src/main/java/io/realm/rx/RealmObservableFactory.java
```diff
@@ -41,6 +44,22 @@
  */
 public class RealmObservableFactory implements RxObservableFactory {
 
+    // Maps for storing strong references to RealmResults while they are subscribed to.
+    // This is needed if users create Observables without manually maintaining a reference to them.
+    // In that case RealmObjects/RealmResults might be GC'ed to early.
```
stk1m1: `to early.` -> `too early.`

realm/realm-library/src/main/java/io/realm/rx/RealmObservableFactory.java
```diff
@@ -240,6 +272,7 @@ public void onChange() {
                     public void call() {
                         object.removeChangeListener(listener);
                         observableRealm.close();
+                        objectRefs.get().acquireReference(object);
```
stk1m1: `objectRefs.get().acquireReference(object)` -> `objectRefs.get().releaseReference(object)`

I think you want to decrease the reference count here?

stk1m1:
Does thread-local copy of `StrongReferenceCounter` variable automatically GC'ed when a `Thread` dies?

realm/realm-library/src/main/java/io/realm/rx/RealmObservableFactory.java
```diff
@@ -270,4 +303,30 @@ public int hashCode() {
         return 37;
     }
 
+
+    // Helper class for keeping track of strong references to objects.
+    private static class StrongReferenceCounter<K> {
+
+        private final Map<K, Integer> references = new IdentityHashMap<K, Integer>();
```
beeender: I am a bit uncertain about using `IdentityHashMap`. 
```
RealmXXObject obj1 = realm.where(RealmXXObject.class).findFirst();
RealmXXObject obj2 = realm.where(RealmXXObject.class).findFirst();

Observable<RealmXXObject> observable1 = obj1.asObservable();
Observable<RealmXXObject> observable2 = obj2.asObservable();
```

Since `observable2.equals(observable1) == true`, the `observable2` won't be added to the map which means it will still be GCed.

cmelchior: But that is the point of the IdentityHashMap?

It uses reference equality (`==`) instead of `equals` to avoid the exact scenario you are describing.

beeender: Oops... How could I understand it in a reverse way... :+1:

beeender:
:+1:

cmelchior:
Comments fixed @stk1m1 

It doesn't make a lot of sense to talk about threads living longer than the application, since the application is controlled by the lifecycle events which are orthogonal to threading. If the system needs the resources the entire process is killed. So there shouldn't be any memory leaks.

stk1m1:
:+1:

Issue #2496, opened 2016-03-18
io.realm.internal.Table.finalize() timed out after 10 seconds


cmelchior:
Copy from #1599 
People are still seeing finalizer timeouts on 0.88.2: http://crashes.to/s/360245fd72a
We should look into how hard it would be to get rid of finalizers completely IMO.

/cc @dalinaum

GaborPeto:
If there are no finalizers how can I still get this exception in v1.2?
` java.util.concurrent.TimeoutException: io.realm.internal.Table.finalize() timed out after 10 seconds
       at java.util.regex.Matcher.closeImpl(Matcher.java)
       at java.util.regex.Matcher.finalize(Matcher.java:652)
       at java.lang.Daemons$FinalizerDaemon.doFinalize(Daemons.java:187)
       at java.lang.Daemons$FinalizerDaemon.run(Daemons.java:170)
       at java.lang.Thread.run(Thread.java:841)`

beeender:
@GaborPeto the PR(#3144) to remove finalizer  has not been merged yet. Sorry for the exception you got.

lorenzopolidori:
Does anybody know in which Realm release this fix will be? I can still see this crash using realm-java 2.2.1 on android.
```
Fatal Exception: java.util.concurrent.TimeoutException: io.realm.internal.Table.finalize() timed out after 10 seconds
       at io.realm.internal.Table.finalize(SourceFile:88)
       at java.lang.Daemons$FinalizerDaemon.doFinalize(Daemons.java:206)
       at java.lang.Daemons$FinalizerDaemon.run(Daemons.java:189)
       at java.lang.Thread.run(Thread.java:818)
```

beeender:
@lorenzopolidori The fix has not been released yet. Currently only our snapshot release contains this fix. The fix will be released in `v2.2.2` soon.

PR #2525, opened 2016-03-29
Finalizer optimization: Replace hashmap

beeender:
Problems we had by using hashmap:
 1. Hashmap is not thread safe and we access the same hashmap in
 finalizer thread and caller thread of Row's constructor.
 2. Removing reference from the hashmap is a little bit slower when
 there are many references waited to be GCed.

Refactor the NativeObjectReference, make it as a final class and calling
native resource's deallocation directly inside the class. This is needed
for further optimization, like batch delete native pointers to avoid
multiple JNI calls.

* On Genymotion Nexus 6 5.1.0 this improves around 50%
* On Huawei Honor 7 Android M this only improves **-2**% - 9%
* On ARMv7 4.1.2 this improves around 30%

Some benchmark comparison using the array solution vs hashmap solution:
(Benchmark running on Genymotion Nexus 6 5.1.0)

```
(hashmap) Creating 200000 rows - in avg 10000 rows takes 65ms
(array) Creating 200000 rows - in avg 10000 rows takes 67ms
(hashmap) Deleting 200000 rows - in avg 10000 rows takes 8ms
(array) Deleting 200000 rows - in avg 10000 rows takes 4ms

(hashmap) Creating 2000000 rows - in avg 10000 rows takes 30ms
(array) Creating 2000000 rows - in avg 10000 rows takes 28ms
(hashmap) Deleting 200000 rows - in avg 10000 rows takes 95ms
(array) Deleting 200000 rows - in avg 10000 rows takes 46ms

(hashmap) Creating 10000000 rows - in avg 10000 rows takes 33ms
(array) Creating 10000000 rows - in avg 10000 rows takes 28ms
(hashmap) Deleting 1000000 rows - in avg 10000 rows takes 62ms
(array) Deleting 1000000 rows - in avg 10000 rows takes 23ms
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### Enhancements
 
 * Throws an IllegalArgumentExcpetion when calling Realm.copyToRealm()/Realm.copyToRealmOrUpdate() with a RealmObject which belongs to another Realm instance in a different thread.
+* Improves speed of cleaning up native resources (#2496).
 
 ### Bug fixes
 
--- a/realm/realm-library/src/main/java/io/realm/internal/Context.java
+++ b/realm/realm-library/src/main/java/io/realm/internal/Context.java
@@ -16,13 +16,41 @@
 
 package io.realm.internal;
 
-import java.lang.ref.Reference;
 import java.lang.ref.ReferenceQueue;
 import java.util.ArrayList;
-import java.util.HashMap;
 import java.util.List;
 
 public class Context {
+
+    // Pool to hold the phantom references.
+    // The size of array for storing phantom references will never decrease. Instead, we use another array to hold the
+    // index of the free slot. When adding the reference, pick the last index from freeIndexList and put the reference
+    // to the corresponding slot. When removing the reference, simply add the index to the end of freeIndexList without
+    // setting the corresponding slot to null for efficiency reasons. The reference will be freed finally when the slot
+    // gets overwritten or the whole context gets freed.
+    private static class ReferencesPool {
+        ArrayList<NativeObjectReference> pool = new ArrayList<NativeObjectReference>();
+        ArrayList<Integer> freeIndexList = new ArrayList<Integer>();
+
+        void add(NativeObjectReference ref) {
+            if (pool.size() <= ref.refIndex) {
+                pool.add(ref);
+            } else {
+                pool.set(ref.refIndex, ref);
+            }
+        }
+
+        Integer getFreeIndex() {
+            Integer index;
+            int freeIndexListSize = freeIndexList.size();
+            if (freeIndexListSize == 0) {
+                index = pool.size();
+            } else {
+                index = freeIndexList.remove(freeIndexListSize - 1);
+            }
+            return index;
+        }
+    }
 
     // Each group of related Realm objects will have a Context object in the root.
     // The root can be a table, a group, or a shared group.
@@ -30,47 +58,50 @@ public class Context {
     // whose disposal need to be handed over from the garbage 
     // collection thread to the users thread.
 
-    // Reserved to be used only as a placeholder by native references Map to avoid autoboxing allocations
-    static final Integer NATIVE_REFERENCES_VALUE = 0;
-
     private List<Long> abandonedTables = new ArrayList<Long>();
     private List<Long> abandonedTableViews = new ArrayList<Long>();
     private List<Long> abandonedQueries = new ArrayList<Long>();
 
-    HashMap<Reference<?>, Integer> rowReferences = new HashMap<Reference<?>, Integer>();
-    ReferenceQueue<NativeObject> referenceQueue = new ReferenceQueue<NativeObject>();
+    private ReferencesPool referencesPool = new ReferencesPool();
+    private ReferenceQueue<NativeObject> referenceQueue = new ReferenceQueue<NativeObject>();
 
     private boolean isFinalized = false;
 
-    public void executeDelayedDisposal() {
-        synchronized (this) {
-            for (int i = 0; i < abandonedTables.size(); i++) {
-                long nativePointer = abandonedTables.get(i);
-                Table.nativeClose(nativePointer);
-            }
-            abandonedTables.clear();
-
-            for (int i = 0; i < abandonedTableViews.size(); i++) {
-                long nativePointer = abandonedTableViews.get(i);
-                TableView.nativeClose(nativePointer);
-            }
-            abandonedTableViews.clear();
-
-            for (int i = 0; i < abandonedQueries.size(); i++) {
-                long nativePointer = abandonedQueries.get(i);
-                TableQuery.nativeClose(nativePointer);
-            }
-            abandonedQueries.clear();
-
-            cleanNativeReferences();
-        }
+    public synchronized void addReference(int type, NativeObject referent) {
+        referencesPool.add(new NativeObjectReference(type, referent, referenceQueue, referencesPool.getFreeIndex()));
     }
 
-    public void cleanNativeReferences() {
+    public synchronized void executeDelayedDisposal() {
+        for (int i = 0; i < abandonedTables.size(); i++) {
+            long nativePointer = abandonedTables.get(i);
+            Table.nativeClose(nativePointer);
+        }
+        abandonedTables.clear();
+
+        for (int i = 0; i < abandonedTableViews.size(); i++) {
+            long nativePointer = abandonedTableViews.get(i);
+            TableView.nativeClose(nativePointer);
+        }
+        abandonedTableViews.clear();
+
+        for (int i = 0; i < abandonedQueries.size(); i++) {
+            long nativePointer = abandonedQueries.get(i);
+            TableQuery.nativeClose(nativePointer);
+        }
+        abandonedQueries.clear();
+
+        cleanNativeReferences();
+    }
+
+    private void cleanNativeReferences() {
         NativeObjectReference reference = (NativeObjectReference) referenceQueue.poll();
         while (reference != null) {
-            reference.clear();
-            rowReferences.remove(reference);
+            // Dealloc the native resources
+            reference.cleanup();
+            // Inline referencesPool.remove() to make it faster.
+            // referencesPool.pool.set(index, null); is not really needed. Make it faster by not
+            // setting the slot to null.
+            referencesPool.freeIndexList.add(reference.refIndex);
             reference = (NativeObjectReference) referenceQueue.poll();
         }
     }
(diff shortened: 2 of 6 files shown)


beeender:
Partial fix of #2496 @realm/java

realm/realm-library/src/main/java/io/realm/internal/Context.java
```diff
@@ -66,11 +94,16 @@ public void executeDelayedDisposal() {
...
-    public void cleanNativeReferences() {
+    private void cleanNativeReferences() {
         NativeObjectReference reference = (NativeObjectReference) referenceQueue.poll();
         while (reference != null) {
+            // Dealloc the native resources
+            reference.cleanup();
             reference.clear();
-            rowReferences.remove(reference);
+            // Inline referencesPool.freeIndexList.remove() to make it faster.
+            // referencesPool.freeIndexList.arrayList.set(index, null); is not really needed. Make it faster by not
```
stk1m1: I think this would indeed referring `referencesPool.freeIndexList.set(index, null);` rather than `referencesPool.freeIndexList.arrayList.set(index, null);`.

realm/realm-library/src/main/java/io/realm/internal/Context.java
```diff
@@ -16,32 +16,60 @@
...
+    private ReferenceQueue<NativeObject> referenceQueue = new ReferenceQueue<NativeObject>();
 
     private boolean isFinalized = false;
 
+    public synchronized void addReference(int type, NativeObject referent) {
+        referencesPool.add(new NativeObjectReference(type, referent, referenceQueue, referencesPool.getFreeIndex()));
+    }
+
     public void executeDelayedDisposal() {
         synchronized (this) {
```
stk1m1: This is merely a suggestion. Wouldn't it be in alignment with `public synchronized void addReference()` above if this synchronized statement turns into a synchronized method?

stk1m1: I am not as versed as I should be about `executeDelayedDisposal()` but it involves with checking the sizes of three unprotected `ArrayList`; `abandonedTables`, `abandonedTableViews`, and `abandonedQueries`. 

Those three are handled without protection in `asyncDisposeTable()`, `asyncDisposeTableView()`, `asyncDisposeQuery()`. Although it is clear that they are not your former work, I would highly appreciate if you could  look into the methods and apply appropriate measures to protect the lists.

realm/realm-library/src/main/java/io/realm/internal/Context.java
beeender: right, will change it!

realm/realm-library/src/main/java/io/realm/internal/NativeObjectReference.java
```diff
@@ -23,25 +23,43 @@
...
-    protected abstract void cleanup();
-
-    @Override
-    public void clear() {
-        cleanup();
-        super.clear();
+    void cleanup() {
+        switch (type) {
+            case TYPE_LINK_VIEW:
+                LinkView.nativeClose(nativePointer);
```
stk1m1: Another suggestion is to setup an interface that could be applied toward classes having `nativeClose()`. Currently we have `Table.nativeClose()`,  `TableView.nativeClose()` `TableQuery.nativeClose()` and `LinkView.nativeClose()` exposed in the files in PR. 

I think It would be really nice to have a common interface for such native cleanup ops that future work would not  need to evolve around loosening up accessibility and exposing native method. Your expertise in designing a structure could highly be appreciated here.

realm/realm-library/src/main/java/io/realm/internal/Context.java
```diff
@@ -66,11 +94,16 @@ public void executeDelayedDisposal() {
...
         NativeObjectReference reference = (NativeObjectReference) referenceQueue.poll();
         while (reference != null) {
+            // Dealloc the native resources
+            reference.cleanup();
             reference.clear();
-            rowReferences.remove(reference);
+            // Inline referencesPool.freeIndexList.remove() to make it faster.
+            // referencesPool.freeIndexList.arrayList.set(index, null); is not really needed. Make it faster by not
+            // setting the slot to null.
+            referencesPool.freeIndexList.add(reference.refIndex);
```
stk1m1: By no means I am against the idea of having in-liners for speed up. ( I love them so much in fact! Don't tell CM! ), but for the sake of maintainability, I would request to pack them into a method if that does not affect overall performance. It seems this part should be very tight since it is in protected area. But... you know...

stk1m1:
I really love the way native references gets handled in `ReferencesPool`, when you invoke `addReference()` and `executeDelayedDisposal()`. Nice work! :+1:

realm/realm-library/src/main/java/io/realm/internal/Context.java
beeender: Yeah, i tried to pack them in one method, but according to the test result, it slows down thing ~10% on Davik .... Java is kind of unpredictable language ...

realm/realm-library/src/main/java/io/realm/internal/Context.java
```diff
@@ -16,32 +16,60 @@
...
+        }
+
+        Integer getFreeIndex() {
+            Integer index;
+            if (freeIndexList.isEmpty()) {
+                index = arrayList.size();
+            } else {
+                int freeIndexToRemove = freeIndexList.size() - 1;
+                index = freeIndexList.get(freeIndexToRemove);
+                freeIndexList.remove(freeIndexToRemove);
```
zaki50: these three lines can be

```
index = freeIndexList.remove(freeIndexList.size() - 1);
```

and if we use `ArrayDeque` instead of `ArrayList`, we can write it as

```
index = freeIndexList.removeLast()
```

realm/realm-library/src/main/java/io/realm/internal/Context.java
```diff
@@ -16,32 +16,60 @@
...
 import java.util.List;
 
 public class Context {
 
+    // Pool to hold the phantom references.
+    // The size of array for storing phantom references will never decrease. Instead, we use another array to hold the
+    // index of the free slot. When adding the reference, pick the last index from freeIndexList and put the reference
+    // to the corresponding slot.
+    private static class ReferencesPool {
+        ArrayList<NativeObjectReference> arrayList = new ArrayList<>();
```
zaki50: `arrayList` -> `pool`?

realm/realm-library/src/main/java/io/realm/internal/Context.java
```diff
@@ -16,32 +16,60 @@
...
+        ArrayList<Integer> freeIndexList = new ArrayList<Integer>();
```
cmelchior: Something inside me really dislike using ArrayList/Integer for this as we are wasting quite a lot of memory compared to a `int[]` list. However it would also increase the complexity of this quite a bit.

cmelchior: Perhaps we should reimplement our own ArrayDeque that is based on a primitive int array?

beeender: Tried with int[] as well, the thing is, it won't speed up and saves very little memory compares to ArrayList.

beeender: One more thing, i think @zaki50 or someone mentioned `ArrayDeque` somewhere, that could optimize this even more. I will try to benchmark that.

cmelchior: According to this, you would be spending 4x the amount of memory: https://speakerdeck.com/romainguy/android-memories?slide=36. You are probably right it will not provide any measurable speed difference, but increasing the number of objects will have a negative impact on the GC, which means that cost might be outside the benchmark.

But benchmarking with ArrayDeque sounds fine to me.

realm/realm-library/src/main/java/io/realm/internal/Context.java
beeender: It is a bit against instinct, but `ArrayList` is still faster. +6% for adding to the `freeIndexList`. `getFreeIndex` will be faster, but not too much. So I'd like to keep the `ArrayList` here.

beeender: Put them in one line is a great idea, will do it.

realm/realm-library/src/main/java/io/realm/internal/Context.java
beeender: Will `synchronized` for `executeDelayedDisposal()`

The other three are bit complex. Maybe we change them when refactoring everything into PhantomReference.

realm/realm-library/src/main/java/io/realm/internal/NativeObjectReference.java
beeender: The problem is, here we won't have an specific native object anymore, so interface doesn't work :(

realm/realm-library/src/main/java/io/realm/internal/Context.java
```diff
@@ -16,61 +16,90 @@
...
-import java.lang.ref.Reference;
 import java.lang.ref.ReferenceQueue;
 import java.util.ArrayList;
-import java.util.HashMap;
 import java.util.List;
 
 public class Context {
 
+    // Pool to hold the phantom references.
+    // The size of array for storing phantom references will never decrease. Instead, we use another array to hold the
```
cmelchior: Perhaps also add a comment about the pool holding old values that are not deleted for efficiency reasons: https://github.com/realm/realm-java/pull/2525/files#diff-a5b83c6e9041e4c99a95e860f0c4dd54R100

cmelchior:
Apart from a small comment, this looks :+1: 
Should we add something to the changelog? Even though it is not really a user facing change, it was still triggered by a user bug?

beeender:
@realm/java change log & comments updated

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ### Enhancements
 
 * Throws an IllegalArgumentExcpetion when calling Realm.copyToRealm()/Realm.copyToRealmOrUpdate() with a RealmObject which belongs to another Realm instance in a different thread.
+* Uses ArrayList to store references to make finalizer faster (#2496).
```
cmelchior: This sounds really internal to me. Perhaps just "Improved speed when cleaning up native resources" or something like that. At least something that the user better can relate to?

cmelchior:
Lightning @beeender :+1:

zaki50:
:+1:

stk1m1:
:+1:

beeender:
retest this please

beeender:
retest this please

Issue #2545, opened 2016-04-02
toString


alfaro28:
#### Goal

override toString

#### Expected Results

toString to return my custom string

#### Actual Results

my custom method is not being called and the default realm toString implementation is shown

#### Steps & Code to Reproduce

create custom object, override toString

#### Version of Realm and tooling
Realm version(s): 0.88.2

Android Studio version: 1.5.1

Which Android version and device: Icecream,  Samsung S3

fobid:
I resolved this issue with custom method. But it's bothering me to create custom, it has too many fields..

dalinaum:
Hello @alfaro28 and @fobid. I just reproduced this issue. It is a bug. We will investigate this issue.

alfaro28:
@fobid I'm using my object in an ArrayAdapter so I can't use that workaround, @dalinaum thanks

dalinaum:
I just made a simple patch #2554. It will resolve this issue.

dalinaum:
#2554 just merged into 0.88.3 and it will be released soon.

PR #2550, opened 2016-04-04
Enable custom equals(), toString() and hashCode()

dalinaum:
Custom methods such as equals(), toString() and hashCode() are not being called correctly. Fix it. (#2545)


dalinaum:
I'll close this issue. I will merge this patch into releases branch instead of master.

PR #2551, opened 2016-04-04
Remove unused imports in generated proxy class

MGaetan89:
It seems that those three imports are never used in the generated proxy class.

### DIFF
--- a/realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmProxyClassGenerator.java
+++ b/realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmProxyClassGenerator.java
@@ -68,17 +68,14 @@ public class RealmProxyClassGenerator {
         imports.add("io.realm.internal.TableOrView");
         imports.add("io.realm.internal.ImplicitTransaction");
         imports.add("io.realm.internal.LinkView");
-        imports.add("io.realm.internal.android.JsonUtils");
         imports.add("java.io.IOException");
         imports.add("java.util.ArrayList");
         imports.add("java.util.Collections");
         imports.add("java.util.List");
-        imports.add("java.util.Date");
         imports.add("java.util.Map");
         imports.add("java.util.HashMap");
         imports.add("org.json.JSONObject");
         imports.add("org.json.JSONException");
-        imports.add("org.json.JSONArray");
         imports.add(metadata.getFullyQualifiedClassName());
 
         for (VariableElement field : metadata.getFields()) {


dalinaum:
Thanks for your contribution, @MGaetan89.
LGTM :+1:

beeender:
retest this please

realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmProxyClassGenerator.java
```diff
@@ -68,17 +68,14 @@ public void generate() throws IOException, UnsupportedOperationException {
         imports.add("io.realm.internal.TableOrView");
         imports.add("io.realm.internal.ImplicitTransaction");
         imports.add("io.realm.internal.LinkView");
-        imports.add("io.realm.internal.android.JsonUtils");
         imports.add("java.io.IOException");
         imports.add("java.util.ArrayList");
         imports.add("java.util.Collections");
         imports.add("java.util.List");
-        imports.add("java.util.Date");
```
cmelchior: I am pretty sure this is needed if you Model class contains date objects. Will have to look into the others. Unfortunately JavaWriter doesn't allow you to only import the Classes that is actually used :(

MGaetan89: I didn't find any direct usage of the `Date` class. If I missed something, I'll obviously add it back.

MGaetan89:
Any idea why the build fails? I do not have access to the result.

emanuelez:
```
/home/jenkins/jenkins/workspace/java_pr/examples/introExample/build/generated/source/apt/debug/io/realm/PersonRealmProxy.java:296: error: cannot find symbol
                JSONArray array = json.getJSONArray("cats");
                ^
  symbol:   class JSONArray
  location: class PersonRealmProxy
1 error
```

It looks like JSONArray is needed after all. I'm pretty sure the other two are needed as well.

cmelchior:
```
Note: Processing class Cat
Note: Processing class Person
Note: Processing class Dog
Note: Creating DefaultRealmModule
/home/jenkins/jenkins/workspace/java_pr/examples/introExample/build/generated/source/apt/debug/io/realm/PersonRealmProxy.java:296: error: cannot find symbol
                JSONArray array = json.getJSONArray("cats");
                ^
  symbol:   class JSONArray
  location: class PersonRealmProxy
```

Note that the content of a Proxy class depend on the fields  present, e.g. all the above imports might be used by the following class:

```
  private Date birthday;
  private RealmList<Foo> foos;
}
```

That is a limitation of JavaWriter really. From my understanding JavaPoet is able to only include the relevant imports and we also wants to migrate at some point, but it hasn't really been pressing so far.

cmelchior:
Hi @MGaetan89 
Thank you for this, but I just went through all our dependencies and all 3 imports are used in some corner cases. Ideally we should dynamically detect when that is, but as long as we use JavaWriter and not JavaPoet that will not be possible.

So for now we will have to close this PR.

PR #2587, opened 2016-04-11
Move the analytics collection from the AP to the transformer.

emanuelez:
In the process the tracking of Kotlin usage has been added.


emanuelez:
@realm/java

realm-transformer/src/main/groovy/io/realm/transformer/RealmTransformer.groovy
```diff
@@ -135,6 +136,41 @@ class RealmTransformer extends Transform {
 
         def toc = System.currentTimeMillis()
         logger.info "Realm Transform time: ${toc-tic} milliseconds"
+
+        sendAnalytics(inputs, inputModelClasses)
+    }
+
+    /**
+     * Sends the analytics
+     * @param inputs the inputs provided by the Transform API
```
cmelchior: Missing whitespace between

realm-transformer/src/main/groovy/io/realm/transformer/RealmTransformer.groovy
```diff
@@ -135,6 +136,41 @@ class RealmTransformer extends Transform {
...
+     * @param inputModelClasses a list of ctClasses describing the Realm models
+     */
+    private static sendAnalytics(Collection<TransformInput> inputs, List<CtClass> inputModelClasses) {
+        def containsKotlin = false
+        inputs.each {
```
cmelchior: Do we really need to loop through files to do this? Can't we just check to see if the kotlin plugin is applied just like when we select APT/KAPT ?

realm-transformer/src/main/java/io/realm/transformer/RealmAnalytics.java
```diff
@@ -68,6 +68,7 @@
             + "      \"Anonymized MAC Address\": \"%USER_ID%\",\n"
             + "      \"Anonymized Bundle ID\": \"%APP_ID%\",\n"
             + "      \"Binding\": \"java\",\n"
+            + "      \"Uses Kotlin\": \"%USES_KOTLIN%\",\n"
```
cmelchior: I think we need to check with @timanglade about the exact wording for this, especially as we might want to track other JVM languages like Groovy and potentially Scala.

timanglade: - It’s worth checking this will work as intended via a quick test but I think it might be better to return this under the `Language` key as a [MixPanel list value](https://mixpanel.com/docs/properties-or-segments/property-data-types).
- Values should be lowercase abbreviated forms of the language name if there are any. The current values other bindings return are `objc`, `swift`, `js` and `c#`. Unless there are any abbreviation I’m not aware of, just report `kotlin`, `java`, `groovy`, etc. ?
- That means we should also include `Java` as a value in that list if we have any reason to suspect Java is being used. (Right now since we technically do not know, all Java runs are reported with the `Language` property left undefined, which is correct since we do not know for sure)
- Note `Language` singular instead of plural, because other bindings already return that property key singular and we should be consistent to simplify reporting. Ideally that will let us do queries in Mixpanel like `Language is one of “java”, “groovy” or “js”.
- When in doubt, feel free to [go in MixPanel and check if language are reported as you expect](https://mixpanel.com/report/686603/segmentation/#action:segment,arb_event:Run,bool_op:and,chart_analysis_type:linear,chart_type:line,from_date:-56,ms_checked:(c%23:!t,js:!t,objc:!t,swift:!t,undefined:!t),ms_values:!(undefined,swift,objc,js,c%23),segfilter:!((dropdown_tab_index:0,property:(name:Language,source:properties,type:string),selected_property_type:string,type:string)),segment_type:string,to_date:-8,type:unique,unit:week)

zaki50:
Moving `RealmAnalytics.java` breaks the link at the end of https://realm.io/docs/java/latest/#i-see-a-network-call-to-mixpanel-when-i-run-my-app-what-is-that

Please do not forget tp update the documents when merging this PR.

realm-transformer/src/main/java/io/realm/transformer/RealmAnalytics.java
emanuelez: @timanglade Yeah, I thought about that before starting to code. The problem is that it's very possible (and common) to have hybrid projects, that contain both Java and Kotlin code, so the singular `Language` is a bit of a problem here. That's why I opted for the boolean flag. Checking whether a project if purely a Kotlin one would require the traversal of the whole project tree, which can be time consuming.

timanglade: Yeah, don’t mind the key name `Language` as opposed to `Languages` as long as we send a List of values I think it’s fine. No need to check if the project uses exclusively Kotlin or Java, as long as we can add either language to the list if we can detect it uses one of them at least once.

realm-transformer/src/main/groovy/io/realm/transformer/RealmTransformer.groovy
emanuelez: Not really. At this point in time we don't have access to Gradle's internals

realm-transformer/src/main/java/io/realm/transformer/RealmAnalytics.java
emanuelez: To detect the usage of Java requires traversing all the project folders, which can be quite slow for big projects. So that leaves us at being able to easily detect pure Java or possibly mixed project. but it would be expensive to compute whether a project is surely mixed or pure Kotlin.

realm-transformer/src/main/groovy/io/realm/transformer/RealmTransformer.groovy
cmelchior: :+1:

realm-transformer/src/main/java/io/realm/transformer/RealmAnalytics.java
timanglade: OK, so assuming the code works as it does now, do you see a syntax that’s cleaner than what I suggested above? I’m still happy with what I suggested, even if it means most of the time it will return nothing (undefined), and sometimes will report only Kotlin even if other languages are being used. Maybe over time we’ll have better ways to detect Java & Groovy I mean it will require some context to interpret the data but it’s cleaner than a bunch of top-level boolean properties for every possible JVM language if our detection ability improves in the future.

PR #2593, opened 2016-04-11
Fix javadoc

dalinaum:
@realm/java

### DIFF
--- a/realm/realm-annotations-processor/src/main/java/io/realm/processor/ModuleMetaData.java
+++ b/realm/realm-annotations-processor/src/main/java/io/realm/processor/ModuleMetaData.java
@@ -44,7 +44,7 @@ public class ModuleMetaData {
     }
 
     /**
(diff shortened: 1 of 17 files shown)


realm/realm-annotations-processor/src/main/java/io/realm/processor/ModuleMetaData.java
```diff
@@ -111,7 +111,7 @@ public boolean generate(ProcessingEnvironment processingEnv) {
         return true;
     }
 
-    // Detour needed to access the class elements in the array
+    // Detours needed to access the class elements in the array
```
cmelchior: Detour is correct here as there is only 1

cmelchior:
Apart from small comment :+1:

dalinaum:
Done.

zaki50:
:+1:

PR #2601, opened 2016-04-13
Add a test to verify that instance methods can be used in constructors

emanuelez:
@realm/java


kneth:
I guess it relates to http://stackoverflow.com/questions/36583951/realm-error-transformclasseswithrealmtransformerfordebug-failed-to-resolve-typ

And your test look good to me.

Issue #2254, opened 2016-02-10
Support creating an initial dataset for a Realm


cmelchior:
From this question: http://stackoverflow.com/questions/35304125/how-can-i-create-a-realm-database-with-initial-data-for-my-android-app

It would make sense to allow people to specify a closure for creating a initial dataset when a Realm file is created for the first time.

Proposal for API:

```
RealmConfiguration config = new RealmConfiguration.Builder(context)
  .initialData(new RealmTransaction() {
     @Override
      public void excute(Realm realm) {
        // Just reuse our existing transaction concept.
        // Right now users have to keep track of this in SharedPreferences or similar, which is something
        // that Realm could track just as easily.
      }
  })
  .build();

// The first time the file is created or reset due to a RealmMigrationNeeded, it will execute the initialData transaction.
Realm realm = Realm.getInstance(config);
```

Thoughts @realm/java ?

kneth:
It's not a bad idea. We should extend `initialData()` with a method which takes a file name as argument. It can then copy a Realm from the assets to the file name set by the `Builder()`.

PR #2602, opened 2016-04-13
Initial dataset for a Realm

thesurix:
Fixes [#2254](https://github.com/realm/realm-java/issues/2254)


realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -316,6 +316,11 @@ private static void initializeRealm(Realm realm) {
                 columnInfoMap.put(modelClass, mediator.validateTable(modelClass, realm.sharedGroupManager.getTransaction()));
             }
             realm.schema.columnIndices = new ColumnIndices(columnInfoMap);
+
+            final Transaction transaction = realm.getConfiguration().getInitialDataTransaction();
```
cmelchior: I don't think this is enough to ensure that we only trigger this when the file is first created, e.g. I am pretty sure this will get triggered in the following case:

```
RealmConfiguration config = new RealmConfiguration.Builder(context).build();
Realm.getInstance(config).close();
Realm.getInstance(config).close(); // This will trigger it again since the schema has been cleared
```

Also, the file might be copied from assets instead. I think the only way of doing this is to have a method that checks for the existence of the Realm file and setting a flag. A good place would probably be just before calling the BaseRealm constructor.

Also, unit tests are needed to verify the behaviour of this.

realm/realm-library/src/main/java/io/realm/RealmConfiguration.java
```diff
@@ -481,6 +495,17 @@ public Builder rxFactory(RxObservableFactory factory) {
             return this;
         }
 
+        /**
+         * Sets the initial data in {@link io.realm.Realm}. This transaction will be executed only for the first time
+         * when database file is created or while migrating the data when {@link Builder#deleteRealmIfMigrationNeeded()} is set.
+         *
+         * @param transaction transaction to execute.
+         */
+        public Builder initialData(Realm.Transaction transaction) {
```
cmelchior: Note that a RealmConfiguration can be used to create both an `DynamicRealm` and a typed `Realm`. The issue didn't really describe this, but it means we probably need to support both 

```
public Builder initialData(Realm.Transaction transaction);
public Builder initialData(DynamicRealm.Transaction transaction);
```

It is slightly annoying though as the semantics gets a bit muddled (e.g. opening a DynamicRealm will not trigger the typed transaction and vice versa), but it might be the best of multiple evils.

What do you think @thesurix @realm/java ?

cmelchior:
Hi @thesurix 
Thanks a lot for this. I really like the direction of this PR, but I am not sure it works correctly in all corner cases, so I commented on some missing unit tests that could impact the implementation.

realm/realm-library/src/main/java/io/realm/Realm.java
beeender: by checking `version == UNVERSIONED` would probably do the trick

realm/realm-library/src/main/java/io/realm/RealmConfiguration.java
kneth: My opinion is that supporting `Realm` only is fine. Most apps will use model classes and thereby `Realm`. Of course, it is important that we state it clearly in the documentation.

cmelchior: I tend to agree. We can err on the side of caution and just add support for `Realm.Transaction` now, then reevaluate if needed.

beeender: I don't really think user really need `public Builder initialData(DynamicRealm.Transaction transaction);`, by supplying that, it also coupled the typed Realm and dynamic Realm even more. We also have to consider a right place to trigger the initial dynamic Realm transaction when user only uses dynamic Realm. So i prefer:
1. Only initial transaction for typed Realm
2. It will only be triggered when first time create Realm (version == unversioned)

thesurix: Yes, I agree with @beeender. The idea behind this initial data transaction is to use `version == unversioned` condition (really simple solution). In `DynamicRealm` it is not easy to spot place where we can use this initial dataset transaction.

thesurix:
Thanks for review. I will fix missing test soon. 

I know there is an idea about asset file, but as you mentioned we can create another issue and do it in this way :+1:

realm/realm-library/src/main/java/io/realm/Realm.java
thesurix: @cmelchior Yes, you are right. It will be executed second time but without `realm.commitTransaction`. I think simple `version` check will do the job.

kneth:
@thesurix Thanks in advance :-) Just notify us when you're ready.

thesurix:
@kneth @cmelchior I fixed this PR and it is ready :)

kneth:
Looks fine to me

dalinaum:
:+1: Looks good to me.

cmelchior:
Looks good to me :+1: I'll add the changelog entry in a separate commit. Thanks for adding this @thesurix

dalinaum:
#2612 includes the changelog entry.

Issue #2415, opened 2016-03-10
Gradle: Cannot find Relinker


mithrann:
Hi,
I am getting error when i try to use 0.88.0 instead of 0.87.5

`Error:Failed to resolve: com.github.KeepSafe:ReLinker:1.1`

cmelchior:
Are you using the new gradle plugin as described here: https://realm.io/news/realm-java-0.88.0/ ?

mithrann:
Yes,  i am using new gradle plugin in my datahandler library of my app
 `buildscript {
 repositories {
    jcenter()
 }
 dependencies {
    classpath 'io.realm:realm-gradle-plugin:0.88.0'
 }
}
apply plugin: 'realm-android'`  and it gives that error.

zoonooz:
@mithrann ran into same problem, quick fix is add these lines in the app build.gradle

```
repositories {
    maven { url 'https://jitpack.io' }
}
```

emanuelez:
That's unexpected since the Realm Gradle plugin should be adding the jitpack repository as you can see here:

https://github.com/realm/realm-java/blob/master/gradle-plugin/src/main/groovy/io/realm/gradle/Realm.groovy#L52

Did you try cleaning your repo? Sometime the Gradle cache interferes with dependencies.

jjhesk:
+1

zoonooz:
@emanuelez I think it's working only in normal android project but not with the library module. As it will compile to AAR without external library inside so the app module have to add the repository again when using it.

mithrann:
Yup, still not working in lib project.

zoonooz:
@mithrann @jjhesk It's working for me in lib project with this setup

**app/build.gradle**
```
buildscript {
    ....
}
repositories {
    maven { url 'https://jitpack.io' }
}
```

**lib/build.gradle**
```
buildscript {
    repositories {
        jcenter()
    }
    dependencies {
        classpath "io.realm:realm-gradle-plugin:0.88.0"
    }
}
apply plugin: 'com.android.library'
apply plugin: 'realm-android'
```

mithrann:
@zoonooz that is also not working for me.

zoonooz:
:disappointed_relieved:

mithrann:
@cmelchior  any update on the issue?

Thanks.

emanuelez:
A fix has been merged and should soon be available in our SNAPSHOT release.

jcj52436999:
[...]

PR #2443, opened 2016-03-14
Use a local copy of ReLinker

emanuelez:
The problem is that libraries that use Realm were forced to tell users to add the jitpack repo manually, which is inconvenient.

Fixes #2415

PR #2608, opened 2016-04-14
Use ReLinker of JCenter instead of including the dependency ourselves.

### LINKED ISSUE #2596: Use ReLinker of JCenter
#### Goal

> @cmelchior:"ReLinker is now on JCenter. We should use that instead of including the dependency ourselves." #2495

#### Expected Results

Use ReLinker of JCenter instead of including the dependency ourselves.

#### Actual Results

Includes the dependency ourselves.

dalinaum:
There was the problem that se Realm were forced to tell users to add the jitpack repo manually, so we have decided to include the dependency ourselves (#2415).

ReLinker is on JCenter now. We can use Relinker of JCenter instead. It will resolve #2596.


cmelchior:
If CI is happy I am happy :+1:

emanuelez:
We need to modify the gradle plugin as well so that it doesn't add jitpack as a repo

dalinaum:
@emanuelez Was it already deleted? https://github.com/realm/realm-java/pull/2443/files#diff-fbc37097ca151e18a6db623e59ebdba8L52

emanuelez:
true :+1:

PR #2610, opened 2016-04-14
Changed made as per realm java 0.88.3

shadabunique:
Updated code to remove all the data items as of realm java 0.88.3


cmelchior:
Hi @shadabunique 
Thank you for this, but our master reflect the latest changes, not the current release. In 0.89 which is the next release, the example is correct.

PR #2615, opened 2016-04-18
Fix unexistent method call

josemigallas:
The method deleteAllFromRealm does not exist in class RealmResults. Instead, **clear** must be used. (Already verified with the [API documentation](https://realm.io/docs/java/latest/api/io/realm/RealmResults.html))


cmelchior:
Hi @josemigallas 
Thank you very much for making sure our examples are correct, but in this case we cannot accept your PR.
It is an upcoming change in 0.89, and our master branch include all changes for that release (See e.g. the changelog here https://github.com/realm/realm-java/blob/master/CHANGELOG.md)

Our previous releases are all on the `/releases/` branch here: https://github.com/realm/realm-java/tree/releases, but if you find any mistakes on that branch we would be more than glad to merge any fixes.

PR #2618, opened 2016-04-19
Throw the most accurate exception possible in JSON methods

cmelchior:
Part of #1594. All bullets should be merged in one release. 

@realm/java

Final commit message:
---
Title: Only throw `RealmException` if absolutely necessary

We should avoid wrapping lower level exceptions in RealmExceptions unless there is a good reason for it. We had multiple support issues where people were confused about the RealmException and didn't understand they had to dig through the stack trace for the original exception. 

This change alters the behaviour of all our JSON methods so they now only convert JSONException to RealmException in order to prevent checked exceptions to reach the user. All other exceptions should be thrown directly.


stk1m1:
👍

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -350,7 +350,7 @@ private static void initializeRealm(Realm realm) {
         for (int i = 0; i < json.length(); i++) {
             try {
                 configuration.getSchemaMediator().createOrUpdateUsingJsonObject(clazz, this, json.getJSONObject(i), false);
-            } catch (Exception e) {
+            } catch (JSONException e) {
                 throw new RealmException("Could not map Json", e);
```
kneth: "Json" -> "JSON"
(do a search-n-replace)

kneth:
:+1: with only minor comment

beeender:
Would it make more sense to add `throws JSONException` to those methods? It is an checked exception, IMO it still should be a checked exception when we calling those methods to convert JSON.

cmelchior:
@mc I really loathe JSONException being checked. I can see arguments why though, so instead of trying to hide it, maybe you are right that we should just use the checked exception. At least in the places where people send in JSONArray/JSONObject. Having it as a checked exception for e.g. `createObjectFromJson(String)` feels like leaking implementation details. What do you think?

mc:
i have no idea what youre talking about.

beeender:
@cmelchior 
> Having it as a checked exception for e.g. createObjectFromJson(String) feels like leaking implementation details.

OK. I agree we can keep it like this until we implement the JSON converter extension.

:+1:

cmelchior:
@mc Sorry about that. I really need that Slack -> GitHub username converter 😄

PR #2515, opened 2016-03-24
PrimaryKey value can be null for String, Byte, Short, Integer, and Long types.

stk1m1:
#### `@PrimaryKey` in `String`, `Byte`, `Short`, `Integer`, and `Long` types can be `null`.
_(The first item for #1766)_

**Few observations on realm-cocoa**. (thanks to @jpsim)
1. `+[RLMObject primaryKey]` guarantees the uniqueness of the key value, and there will exist only one, single `nil` key for a model. `nil` is treated as a unique state for the set of primary key values if and only if the primary key property type is optional.
2. Primary keys themselves cannot be updated, even if they are `nil`. The object with `nil` primary key can be, however, updated.
3. Since ObjectStore has been deployed already, most of the primary key logic lives in there.

**This PR Covers**
- [x] A `@PrimaryKey` annotated field in `java.lang.String`/`Byte`/`Short`/`Integer`/`Long` type can be `null`.
- [x] The object with `nil` primary key can be updated.
- [x] Allow nullable type `@PrimaryKey` to be nullable in schema. This is loosening what has applied in #2003
- [x] Migration checks if existing nullable type `@PrimaryKey` is set to be nullable in existing file. If not, throws `RealmMigrationNeeded`.
- [x] Unit tests for covering `findFirstNull`, and nullable PK types.  
  * `null` values for Boxed integer type PK. (copy, copyOrUpdate, createOrUpdateObjectFromJson).  
  * Mix `null` value and non-null value to see if object with `null` PK updates properly (String, Integer).  
  * `RealmQuery` on nullable PK to see if it brings expected results.  
  * Migration of Boxed integer type PK. (existing realm to new realm, migration to/from primitive type PK).

PR #2622, opened 2016-04-19
Make Realm.createObject(Class,PrimaryKey) public

cmelchior:
Part of #1594. Release all changes in one release.

@realm/java 

Final commit message
-----
Title: Make Realm.createObject(Class,PrimaryKey) public

This method is a way to solve the problem that people might have data with a primary key that is is the default value for that type, e.g a long primary key with the value 0.

Once this object is saved in Realm any calls to the normal `Realm.createObject()` would throw an exception as it would try to assign the default value (zero) to the object before the user could set it to something else. With this method that is no longer a problem.

Internally we already use this method and it is already public in the Dynamic API. The reason for not making this public earlier was due to fear of mis-use since there is no type-safe guarantee at compile time. However since it is already public in the Dynamic API and we haven't seen any indication of that being misused we feel it should be safe to make this public. Also our Query API is also semi-threadsafe so there is precedence there as well.


CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## 0.90.0
+
+### Enhancements
+
+* `Realm.createObject(Class, Object)` for creating objects with a primary key directly.
```
kneth: Do we really back quote in the change log?

kneth:
Looks fine to me

beeender:
#2515 needs to have some more test case for `createObject` after this merged.

:+1:  for this one.

stk1m1:
👍

CHANGELOG.md
cmelchior: Since we moved to MarkDown, I think it makes it more readable?

zaki50:
`io.realm.RealmTests > createObjectWithPrimaryKey_valueAlreadyExists` and `io.realm.RealmTests > createObjectWithPrimaryKey_valueAlreadyExists` is failed

Dr-Emann:
If we only support integers (`short`, `int`, `long`) or Strings as primary keys, should we use overloads to define 

```java
Realm.createObject(Class, short)
Realm.createObject(Class, int)
Realm.createObject(Class, long)
Realm.createObject(Class, String)
```

They could all delegate to `Realm.createObject(Class, Object)` internally

beeender:
@Dr-Emann Yes, I thing you can do that. The only reason we don't supply the primitive overload versions of `Realm.createObject()` is to avoid increasing the methods count of the Realm lib (the dex 64k methods limit :( ).

cmelchior:
Also, it wouldn't really increase type safety as you could still accidentally provide a String primary key for a Integer field

beeender:
We won't have type checking for this method which would cause below problems. Shall we doc this?

```java
    @Test
    public void createObjectWithPrimaryKey_integerOverflow() {
        realm.beginTransaction();

        // Byte
        realm.createObject(PrimaryKeyAsBoxedByte.class, 0x00);
        realm.createObject(PrimaryKeyAsBoxedByte.class, 0xff00);
        realm.createObject(PrimaryKeyAsBoxedByte.class, 0xff0000);

        realm.commitTransaction();

        RealmResults<PrimaryKeyAsBoxedByte> results = realm.where(PrimaryKeyAsBoxedByte.class).findAll();
        assertEquals(3, results.size());
        for (PrimaryKeyAsBoxedByte obj : results) {
            assertEquals(0, obj.getId().byteValue());
        }
    }
```

cmelchior:
@beeender Yes, I guess overflow could happen with this method and we should probably make it clear in the Javadoc that there are edge cases when using it.

PR #2626, opened 2016-04-20
Add the Jenkins logic to the repo

emanuelez:


emanuelez:
@cmelchior @teotwaki

cmelchior:
Is the name of the file the only allowed one? E.g. Travis uses hidden files ".travis.yml", which at a glance looks like a better idea, but apart from that 👍

emanuelez:
Any name is allowed, but Jenkinsfile is the default one

cmelchior:
In that case I would probably prefer ".jenkins" or something like that, but it is a minor thing.

orrc:
I believe a regular pipeline job allows you to specify a filename other than `Jenkinsfile`, but other plugins, such as multibranch pipeline seem to rely on this filename.

cmelchior:
Thanks @orrc, if there is a chance we might break other plugins down the line we should probably just stick to the default name then.

emanuelez:
Nicely spotted @orrc. I guess that deviating from default values if often a dangerous practice

emanuelez:
retest this please

emanuelez:
Unfortunately this is currently blocked by https://issues.jenkins-ci.org/browse/JENKINS-33719.
The workaround proposed in the issue is not available in our case.

emanuelez:
retest this please

emanuelez:
retest this please

emanuelez:
Ready for review again @cmelchior @teotwaki . Also in conjunction with the test PR job: https://ci.realm.io/job/android-pr-jenkinsfile/configure

Jenkinsfile
```diff
@@ -0,0 +1,155 @@
...
+        sendABIMetric('armeabi')
+        sendABIMetric('armeabi-v7a')
+        sendABIMetric('arm64-v8a')
+        sendABIMetric('mips')
+        sendABIMetric('x86')
+        sendABIMetric('x86_64')
+    }
+}
+
+{ ->
```
teotwaki: From what I understand, this actually forces the main job (coordinator) to run on the Jenkins master. If instead, you replace this closure by a `node()` call, some slave will be the coordinator. If we start using this a lot, we might overload master, so scheduling it on a node might make sense.

Maybe we should have a coordinator tag, and give it to a bunch of slaves that have good uptime/availability?

Jenkinsfile
```diff
@@ -0,0 +1,155 @@
...
+                    doGenerateSubmoduleConfigurations: false,
+                    extensions: [[$class: 'CleanCheckout']],
+                    gitTool: 'native git',
+                    submoduleCfg: [],
+                    userRemoteConfigs: [[url: 'https://github.com/realm/realm-java.git']]
+                ])                
+            }
+
+           stage 'JVM tests'
+           sh 'chmod +x gradlew && ./gradlew assemble check javadoc --stacktrace'
```
teotwaki: Might make more sense to just invoke `bash gradlew` rather than chmod'ing all over the place.

```
def gradle(command) {
  sh "bash ./gradlew ${command}"
}
```

Jenkinsfile
```diff
@@ -0,0 +1,155 @@
...
+                  stash name: 'test-apk', includes: 'realm-android-library-debug-androidTest-unaligned.apk'
+               }
+           }
+        }
+
+        node('android-hub') {
+            stage 'run instrumented tests'
+            unstash 'test-apk'
+            sh 'adb devices'
+            sh 'adb devices | grep -v List | grep -v ^$ | awk \'{print $1}\' | parallel \'adb -s {} uninstall io.realm.test; adb -s {} install realm-android-library-debug-androidTest-unaligned.apk; adb -s {} shell am instrument -w -r io.realm.test/android.support.test.runner.AndroidJUnitRunner > test_result_{}.txt; java -jar /opt/log-converter.jar test_result_{}.txt\''
```
teotwaki: Could be simplified to

```bash
sh "adb devices | grep -vE '^(List|$)' | cut -f1 ...
```

Jenkinsfile
```diff
@@ -0,0 +1,155 @@
...
+        testResults: path
+    ])
+}
+
+def collectAarMetrics() {
+    dir('realm/realm-library/build/outputs/aar') {
+        sh 'unzip realm-android-library-release.aar -d unzipped'
+        sh 'find $ANDROID_HOME -name dx | sort -r | head -n 1 > dx'
+        sh '$(cat dx) --dex --output=temp.dex unzipped/classes.jar'
+        sh 'cat temp.dex | head -c 92 | tail -c 4 | hexdump -e \'1/4 "%d"\' > methods'
```
teotwaki: Consider squashing all of these into a single `sh` invocation, to be more efficient.

Jenkinsfile
```diff
@@ -0,0 +1,155 @@
...
+                abi=`basename $path`
+                echo -n $abi_size > abi_$abi
+              done
+        '''
+        sendABIMetric('armeabi')
+        sendABIMetric('armeabi-v7a')
+        sendABIMetric('arm64-v8a')
+        sendABIMetric('mips')
+        sendABIMetric('x86')
+        sendABIMetric('x86_64')
```
teotwaki: Can this be done in a `parallel()` step?

Jenkinsfile
```diff
@@ -0,0 +1,155 @@
...
+               dir('realm-library/build/outputs/apk') {
+                  stash name: 'test-apk', includes: 'realm-android-library-debug-androidTest-unaligned.apk'
+               }
+           }
+        }
+
+        node('android-hub') {
+            stage 'run instrumented tests'
+            unstash 'test-apk'
+            sh 'adb devices'
```
teotwaki: What is the purpose of this line?

Jenkinsfile
teotwaki: Also, this should probably be more robust against the case where there is no device ID returned by `adb devices`.

Jenkinsfile
emanuelez: Master is always the coordinator in pipeline jobs. But it's not the main job. See the job I linked in the description. The main job will be the one doing the initial checkout to retrieve the Jenkinsfile

Jenkinsfile
```diff
@@ -34,27 +33,50 @@ def storeJunitResults(String path) {
...
+    // when this method is called, not when we pass it to parallel.
+    // To do this, you need to wrap the code below in { }, and either return
+    // that explicitly, or use { -> } syntax.
+    return {
+        sh "adb -s ${device} shell getprop ro.product.model | tee model-name.txt"
+        def modelName = readFile('model-name.txt').trim().replaceAll(' ', '_')
+
+        sh "adb -s ${device} uninstall io.realm.test"
+        sh "adb -s ${device} install realm-android-library-debug-androidTest-unaligned.apk"
+        sh "adb -s ${device} shell am instrument -w -r io.realm.test/android.support.test.runner.AndroidJUnitRunner > test_result_${modelName}_${device}.txt"
```
cmelchior: This was recently modified to account for the micro-benchmarks.

cmelchior:
Apart from a small change to how the unit tests are run this looks 👍 (I won't pretend to understand I 100% understand everything though)

PR #2627, opened 2016-04-20
introduce `buildTargetABIs` property to realm-jni project in order to allow to build only specified ABI(s)

zaki50:
example(only build jni files for arm64 and x86_64):
```
./gradlew assemble -PbuildTargetABIs=arm64,x86_64
```

example(build jni files for connected device and execute connectedCheck):
```
./gradlew connectedCheck -PbuildTargetABIs=$(adb shell getprop ro.product.cpu.abi)
```

@realm/java

### DIFF
--- a/realm/realm-jni/build.gradle
+++ b/realm/realm-jni/build.gradle
@@ -16,6 +16,9 @@ ext.stripSymbols = project.hasProperty('stripSymbols') ? project.getProperty('st
 ext.coreSourcePath = project.hasProperty('coreSourcePath') ? project.getProperty('coreSourcePath') : null
 // The location of core archive.
 ext.coreArchiveDir = System.getenv("REALM_CORE_DOWNLOAD_DIR")
+// target ABIs to build(null means all).
+// To obtain the ABI of the connected device, execute "adb shell getprop ro.product.cpu.abi"
+ext.buildTargetAbis = project.hasProperty('buildTargetABIs') ? project.getProperty('buildTargetABIs').split(',').collect {it.trim()} : null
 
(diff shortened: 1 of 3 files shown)


cmelchior:
Nice idea 👍 
Perhaps we should add a section at the bottom of the README with the various configuration options we have no. There is starting to be quite a bit, and it would be nice for new hires.

emanuelez:
Nice! 👍

stk1m1:
Super nice! 👍 💯

zaki50:
added a check of ABI names from `buildTargetABIs` property and small example to `README.md`.

please review last 2 commits. @realm/java

realm/realm-jni/build.gradle
```diff
@@ -83,7 +83,21 @@ def allTargets = [
     new Target( name:'x86_64',  abi:'x86_64',      toolchain:toolchains.find {it.name == 'x86_64'}, cflags:[] )
 ]
 
-def targets = ext.buildTargetAbis == null ? allTargets : allTargets.findAll {ext.buildTargetAbis.contains(it.abi)}
+def targets
+if (ext.buildTargetAbis == null) {
+    targets = allTargets;
+} else {
+    targets = ext.buildTargetAbis.collect {
+        def targetAbi = it
```
emanuelez: why introducing this variable? if you want `it` to be renamed you can just use `collect { targetAbi -> ... }`

zaki50: Right!

fixed.

emanuelez:
:shipit:

cmelchior:
:+1:

PR #2634, opened 2016-04-21
PrimaryKey value can be null for String, Byte, Short, Integer, and Long types.

stk1m1:
### Final commit note

**Title: `@PrimaryKey` in `String`, `Byte`, `Short`, `Integer`, and `Long` types can be `null`.**

@PrimaryKey is allowed to be null for String and Boxed primitive classes.

Features part of this commit:

1. A `@PrimaryKey` annotated field in `java.lang.String`/`Byte`/`Short`/`Integer`/`Long` type can be `null`.
2. The object with `nil` primary key can be updated.
4. Migration checks if existing nullable type `@PrimaryKey` is set to be nullable in existing file. If not, throws `RealmMigrationNeeded`.
4. Duplicated `null` as PrimaryKey throws `RealmPrimaryKeyConstraintException `.
5. When a new empty row is added, the default value checked for duplicated is still the same empty string (`""`) at [`Table.addEmptyRow()`](https://github.com/realm/realm-java/blob/master/realm/realm-library/src/main/java/io/realm/internal/Table.java#L358) 
6. Since there isn't equivalent term for `Optional` in Java 7 yet, nullable PK types are configured as following in this PR.

|           |  String PK  | Number PK | Primitive PK     |
|---------|--------------------|---------------------|------------------|
| `Table` |   *Nullable*    |    *Nullable*     | Not Nullable |


cmelchior:
:shipit:

PR #2648, opened 2016-04-22
Add getPrimaryKey() method to RealmObjectSchema

### LINKED ISSUE #2636: Add a "getPrimaryKeyType()" method to RealmObjectSchema
#### Goal

`RealmObjectSchema` already has the `isPrimaryKey(String)`, `hasPrimaryKey()`, `addPrimaryKey(String)`, and `removePrimaryKey()` methods, but it lacks an easy way to check the `RealmFieldType` of an existing primary key field (if there is such a field, of course).

#### Expected Results

I would expect there to be some method with a signature like `RealmFieldType getPrimaryKeyType()`. Alternatively, there could be a `String getPrimaryKeyName()` method, which might be even more useful, and would allow something simple like `getFieldType(getPrimaryKeyName());` to be done.

#### Actual Results

Once I know there's a primary key, I have to do something like the following to find out what its `RealmFieldType` is:  
```java
// Assuming I already have a RealmObjectSchema variable called "schema" somewhere.
Set<String> fields = schema.getFieldNames();
RealmFieldType pkType;
for (String field : fields) {
    if (schema.isPrimaryKey(field)) {
        pkType = schema.getFieldType(field);
        break;
    }
}
```

bkromhout:
Resolves #2636: this PR adds a `getPrimaryKey()` method to `RealmObjectSchema`.

`getPrimaryKey()` returns the name of the primary key field if the model has a primary key, or throws an `IllegalStateException` if the model doesn't have a primary key field (using the same message that `removePrimaryKey()` uses in such a situation).

I've also modified/added tests in `RealmObjectSchemaTests.java` in order to attain 100% coverage for `getPrimaryKey()`.

I'm new to GitHub pull requests, so if I've done something wrong here please don't hesitate to tell me!

### DIFF
--- a/realm/realm-library/src/androidTest/java/io/realm/RealmObjectSchemaTests.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/RealmObjectSchemaTests.java
@@ -29,7 +29,10 @@ import org.junit.runner.RunWith;
 import java.util.Date;
 import java.util.Set;
 
-import static org.junit.Assert.*;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
 
 @RunWith(AndroidJUnit4.class)
 public class RealmObjectSchemaTests {
--- a/realm/realm-library/src/androidTest/java/io/realm/RealmObjectSchemaTests.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/RealmObjectSchemaTests.java
@@ -17,7 +17,8 @@
 package io.realm;
 
 import android.support.test.runner.AndroidJUnit4;
-
+import io.realm.entities.AllJavaTypes;
+import io.realm.rule.TestRealmConfigurationFactory;
 import org.junit.After;
 import org.junit.Before;
 import org.junit.Rule;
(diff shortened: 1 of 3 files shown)


realm/realm-library/src/androidTest/java/io/realm/RealmObjectSchemaTests.java
```diff
@@ -28,13 +29,7 @@
...
 import java.util.Set;
 
-import io.realm.entities.AllJavaTypes;
-import io.realm.rule.TestRealmConfigurationFactory;
-
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertFalse;
-import static org.junit.Assert.assertTrue;
-import static org.junit.Assert.fail;
+import static org.junit.Assert.*;
```
cmelchior: We generally try to avoid '*' star imports.

realm/realm-library/src/main/java/io/realm/RealmObjectSchema.java
```diff
@@ -16,19 +16,13 @@
...
-import java.util.LinkedHashSet;
-import java.util.Map;
-import java.util.Set;
-
 import io.realm.annotations.Required;
 import io.realm.internal.ImplicitTransaction;
 import io.realm.internal.Table;
 import io.realm.internal.TableOrView;
 
+import java.util.*;
```
cmelchior: Same

cmelchior:
Hi @bkromhout 
Thank you for this. It looks great, also the tests. I only had a few comments about the star imports which we try to avoid and you also need to update the 0.90 changelog under enhancements. Besides that this is 👍 

Please review as well @realm/java

bkromhout:
Hi @cmelchior, all comments addressed. Sorry about the imports, I've set Android Studio to not do that in case I contribute anything else in the future.

cmelchior:
Looks good to me. Thanks 👍

emanuelez:
👍

PR #2673, opened 2016-04-26
BaseRealm should be package protected

beeender:


cmelchior:
If this is enough 👍 , but you should probably get @nhachicha to verify.

nhachicha:
LGTM, when I refactored the `ProxyState` to be returned inside the generated Proxy class, I didn't noticed that we end up with `DynamicRealmObject` and the generated classes being in the same package `io.realm` so having `BaseRealm` as package private makes sense now 👍

beeender:
CI: retest this please

beeender:
retest this please

beeender:
CI passed, but cannot send result back. See https://ci.realm.io/job/android-pr/239/

PR #2675, opened 2016-04-26
Make all methods on RealmObject and all classes in public API final

zaki50:
Part of #1594. Merge all together in one release.

@realm/java

Final commit note:
-----

Title: Make all methods on RealmObject and all classes in public API final

We should mark all methods that should not be overridden final.
We should mark all classes that should not be inherited final.


cmelchior:
👍

beeender:
:+1:

PR #2676, opened 2016-04-26
mark constructor of RealmAsyncTask package protected

zaki50:
Part of #1594. Merge all together in one release.

@realm/java

Final commit note:
----

Title: mark constructor of RealmAsyncTask package protected

Removed constructor from the public API as the method is not useful for public use and we want to keep the public API as clean as possible


cmelchior:
👍 

Regarding the commit message. I would probably describe the reasoning the behind the change instead of the actual action if that makes sense, .e.g "Removed constructor from the public API as the method is not useful for public use and we want to keep the public API as clean as possible".

That will provide better information for anyone in the future wondering.

zaki50:
@cmelchior thanks. I've updated the message.

beeender:
:+1:

PR #2678, opened 2016-04-26
add proguardFiles configuration to modulesExample/app project

zaki50:
ProGuard example should have `proguardFiles` configuration which enables
app to have application specific configurations for ProGurad since
real app needs to have configuration for it's library in most cases.

@realm/java


nhachicha:
👍

cmelchior:
👍

nhachicha:
Reminder: we should also remove the ProGuard section from the doc (but that's a different PR against the doc)

PR #2679, opened 2016-04-26
Expanding time resolution into millisecs

### LINKED ISSUE #833: Date precisision not precise enough (dropps all milliseconds)

kneth:
Closes #833 

Current status: ready to be merged


realm/realm-jni/src/io_realm_internal_TableQuery.cpp
```diff
@@ -111,7 +111,6 @@ static jlong getDistinctViewWithHandover
         switch (table->get_column_type(S(columnIndex))) {
             case type_Bool:
             case type_Int:
-            case type_DateTime:
```
bmunkholm: Does timestamp not support distinct?

realm/realm-jni/src/util.hpp
```diff
@@ -669,6 +671,31 @@ inline jobject NewFloat(JNIEnv* env, float value)
...
+}
+
+inline realm::Timestamp from_milliseconds(jlong t)
+{
+    // FIXME: find more efficient way to convert msecs -> (secs, nsecs)
+    // FIXME: guard against overflows
+    int64_t seconds;
+    uint32_t nanoseconds;
+    if (t >= 0) {
+        seconds = int64_t(double(t)*0.001);
```
zaki50: why converted to `double`?

`seconds = int64_t(t / 1000);` does not work?

realm/realm-jni/src/util.hpp
```diff
@@ -669,6 +671,31 @@ inline jobject NewFloat(JNIEnv* env, float value)
...
+        nanoseconds = 1000000*(int64_t(t)-1000*seconds);
+    }
+    else {
+        seconds = int64_t(double(t)*0.001) - 1;
```
zaki50: same

realm/realm-jni/src/io_realm_internal_CheckedRow.cpp
```diff
@@ -103,13 +103,13 @@ JNIEXPORT jdouble JNICALL Java_io_realm_internal_CheckedRow_nativeGetDouble
     return Java_io_realm_internal_UncheckedRow_nativeGetDouble(env, obj, nativeRowPtr, columnIndex);
 }
 
-JNIEXPORT jlong JNICALL Java_io_realm_internal_CheckedRow_nativeGetDateTime
+JNIEXPORT jlong JNICALL Java_io_realm_internal_CheckedRow_nativeGetTimestamp
   (JNIEnv* env, jobject obj, jlong nativeRowPtr, jlong columnIndex)
 {
-    if (!ROW_AND_COL_INDEX_AND_TYPE_VALID(env, ROW(nativeRowPtr), columnIndex, type_DateTime))
+    if (!ROW_AND_COL_INDEX_AND_TYPE_VALID(env, ROW(nativeRowPtr), columnIndex, type_Timestamp))
         return 0;
```
stk1m1: extra line.

realm/realm-jni/src/io_realm_internal_table.cpp
```diff
@@ -1384,16 +1398,16 @@ JNIEXPORT jlong JNICALL Java_io_realm_internal_Table_nativeGetSortedView(
...
         case type_Double:
         case type_Float:
+        case type_Timestamp:
             try {
                 TableView* pTableView = new TableView( pTable->get_sorted_view(S(columnIndex), ascending != 0 ? true : false) );
                 return reinterpret_cast<jlong>(pTableView);
             } CATCH_STD()
         default:
-            ThrowException(env, IllegalArgument, "Sort is currently only supported on integer, boolean, double, float, String, and Date columns.");
+            ThrowException(env, IllegalArgument, "Invalid type - Only String, Date, boolean, short, int, long and their boxed variants are supported.");
```
stk1m1: boolean, _`byte`_, short, int, long

realm/realm-jni/src/io_realm_internal_tableview.cpp
```diff
@@ -268,15 +268,15 @@ JNIEXPORT jdouble JNICALL Java_io_realm_internal_TableView_nativeGetDouble(
     return TV(nativeViewPtr)->get_double( S(columnIndex), S(rowIndex));  // noexcept
 }
 
-JNIEXPORT jlong JNICALL Java_io_realm_internal_TableView_nativeGetDateTimeValue(
+JNIEXPORT jlong JNICALL Java_io_realm_internal_TableView_nativeGetTimestamp(
     JNIEnv* env, jobject, jlong nativeViewPtr, jlong columnIndex, jlong rowIndex)
 {
     try {
```
stk1m1: Can we remove `try`/`catch` block?

realm/realm-jni/src/io_realm_internal_table.cpp
```diff
@@ -783,13 +793,16 @@ JNIEXPORT void JNICALL Java_io_realm_internal_Table_nativeSetString(
...
+    JNIEnv* env, jobject, jlong nativeTablePtr, jlong columnIndex, jlong rowIndex, jlong timestampValue)
 {
-    if (!TBL_AND_INDEX_AND_TYPE_VALID(env, TBL(nativeTablePtr), columnIndex, rowIndex, type_DateTime))
+    if (!TBL_AND_INDEX_AND_TYPE_VALID(env, TBL(nativeTablePtr), columnIndex, rowIndex, type_Timestamp))
         return;
     try {
-        TBL(nativeTablePtr)->set_datetime( S(columnIndex), S(rowIndex), dateTimeValue);
+        // FIXME: find more accurate and efficient way to convert msec to (sec, nsec)
+        int64_t seconds = int64_t(timestampValue/1000);
+        uint32_t nanoseconds = uint32_t(1000000*(timestampValue-1000*seconds));
```
zaki50: why convert logic is different from one in `from_milliseconds`?

realm/realm-jni/src/io_realm_internal_tableview.cpp
```diff
@@ -36,7 +36,7 @@ inline bool view_valid_and_in_sync(JNIEnv* env, jlong nativeViewPtr) {
         }
         // depends_on_deleted_linklist() will return true if and only if the current TableView was created from a
         // query on a RealmList and that RealmList was then deleted (as a result of the object being deleted).
-        if (!TV(nativeViewPtr)->is_in_sync() && TV(nativeViewPtr)->depends_on_deleted_linklist()) {
+        if (!TV(nativeViewPtr)->is_in_sync() && TV(nativeViewPtr)->depends_on_deleted_object()) {
```
bmunkholm: Is this change related to timestamp?

realm/realm-library/src/main/java/io/realm/internal/Table.java
```diff
@@ -1237,7 +1237,7 @@ public long findFirstDate(long columnIndex, Date date) {
         if (date == null) {
             throw new IllegalArgumentException("null is not supported");
         }
-        return nativeFindFirstDate(nativePtr, columnIndex, date.getTime() / 1000);
+        return nativeFindFirstTimestamp(nativePtr, columnIndex, date.getTime() / 1000);
```
bmunkholm: Shouldn't "/1000" here I assume?

realm/realm-jni/src/io_realm_internal_tableview.cpp
cmelchior: Did the core API change this method?

realm/realm-jni/src/util.hpp
```diff
@@ -654,6 +655,7 @@ extern jmethodID java_lang_float_init;
 extern jclass java_lang_double;
 extern jmethodID java_lang_double_init;
 
+
```
cmelchior: Remove newline

realm/realm-library/src/androidTest/java/io/realm/RealmObjectTests.java
```diff
@@ -560,10 +560,8 @@ private CyclicType createCyclicData() {
 
     @Test
     public void dateType() {
-        long testDatesValid[] = {-1000, 0, 1000};
-        long testDatesLoosePrecision[] = {Long.MIN_VALUE, 1, 1001, Long.MAX_VALUE};
+        long testDatesValid[] = {-1001, -1000, -1, 0, 1, 1000, 1001};
```
cmelchior: Why remove the boundary checks? Long.MIN/MAX ?

realm/realm-library/src/androidTest/java/io/realm/RealmQueryTests.java
```diff
@@ -1827,6 +1827,7 @@ public void isValid_removedParent() {
         list.removeAll(SUPPORTED_IS_EMPTY_TYPES);
         list.remove(RealmFieldType.UNSUPPORTED_MIXED);
         list.remove(RealmFieldType.UNSUPPORTED_TABLE);
+        list.remove(RealmFieldType.UNSUPPORTED_DATE);
```
cmelchior: If we are cleaning up the API, it might make sense to remove these from RealmFieldType. Could be done in another PR though.

realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -575,6 +575,7 @@ public void deleteAll() {
      * Deletes the Realm file defined by the given configuration.
      */
     static boolean deleteRealm(final RealmConfiguration configuration) {
+        final String management = ".management";
```
cmelchior: Not related to datetime, but the core upgrade in general?

realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -602,6 +603,28 @@ public void onResult(int count) {
...
+                    for (File fileToDelete : filesToDelete) {
+                        if (fileToDelete.exists()) {
+                            boolean deleteResult = fileToDelete.delete();
+                            if (!deleteResult) {
+                                realmDeleted.set(false);
+                                RealmLog.w("Could not delete the file " + fileToDelete);
+                            }
+                        }
+                    }
+                    boolean deleteResult = managementFolder.delete();
```
cmelchior: This would break apps with multiple Realm files?

realm/realm-library/src/main/java/io/realm/RealmFieldType.java
```diff
@@ -37,7 +37,8 @@
     BINARY(4),
     UNSUPPORTED_TABLE(5),
     UNSUPPORTED_MIXED(6),
-    DATE(7),
+    UNSUPPORTED_DATE(7),
```
cmelchior: I would probably just remove this together with other unsupported types. No need to expose them in the public API I think?

realm/realm-library/src/main/java/io/realm/internal/Table.java
cmelchior: This hasn't been addressed?

cmelchior: And more importantly. Why isn't it getting caught in a unit test?

realm/realm-library/src/main/java/io/realm/internal/TableView.java
```diff
@@ -370,7 +370,7 @@ public void setDouble(long columnIndex, long rowIndex, double value){
     @Override
     public void setDate(long columnIndex, long rowIndex, Date value){
         if (parent.isImmutable()) throwImmutable();
-        nativeSetDateTimeValue(nativePtr, columnIndex, rowIndex, value.getTime()/1000);
+        nativeSetTimestampValue(nativePtr, columnIndex, rowIndex, value.getTime()/1000);
```
cmelchior: We shouldn't divide by 1000 here. No unit test catching this?

realm/realm-library/src/main/java/io/realm/internal/TableView.java
```diff
@@ -493,7 +493,9 @@ public long findFirstDouble(long columnIndex, double value) {
 
     @Override
     public long findFirstDate(long columnIndex, Date date) {
-        return nativeFindFirstDate(nativePtr, columnIndex, date.getTime()/1000);
+        // FIXME: waiting for implementation
+        return NO_MATCH;
```
cmelchior: Nitpick: Throw UnsupportedOperation instead?

cmelchior:
Done with a quick pass. Generally looks great. Like the internal renaming from Date to Timestamp. I am a bit concerned that some dividing by 1000 hasn't been caught by unit tests, which indicate our test coverage could be better?

realm/realm-jni/src/io_realm_internal_tableview.cpp
kneth: No, but we skipped a few core releases. In principle I agree that it should be in a separate PR but I simply had to push this forward.

realm/realm-jni/src/io_realm_internal_TableQuery.cpp
kneth: I had to disable it as I got seg. faults. It has been fixed in core.

realm/realm-jni/src/util.hpp
kneth: I have changed the conversion as the `Timestamp` constructor now supports negative nanoseconds.

realm/realm-jni/src/io_realm_internal_table.cpp
kneth: Indeed so

realm/realm-jni/src/io_realm_internal_tableview.cpp
kneth: Yes, I believe we can. Also for other getters.

realm/realm-jni/src/io_realm_internal_table.cpp
kneth: I have consolidated it by calling `from_milliseconds`.

realm/realm-jni/src/io_realm_internal_tableview.cpp
kneth: @cmelchior Yes, the method has been renamed recently.

realm/realm-library/src/main/java/io/realm/internal/Table.java
kneth: We have a test but I disabled it initially and apparently forgot to reenable it. Nice catch!

realm/realm-library/src/androidTest/java/io/realm/RealmObjectTests.java
kneth: I readd the boundaries.

realm/realm-library/src/androidTest/java/io/realm/RealmQueryTests.java
kneth: `RealmFieldType` is supposed to be in sync with a similar enum in core, so I'm a bit reluctant to remove them.

realm/realm-library/src/main/java/io/realm/BaseRealm.java
kneth: Yes, we skipped a few core releases. Of course I should have fixed with in another PR.

realm/realm-library/src/androidTest/java/io/realm/RealmQueryTests.java
cmelchior: But it is also a public API. Having exposed types that are not used at all in the binding is a bit weird, no?

realm/realm-library/src/main/java/io/realm/BaseRealm.java
kneth: Oh yes, that's a valid point. But the management folder's name is derived from the Realm file's name and you will only have a collision if two Realms have the same name - and then the two Realms are the same and we only have one :-)

realm/realm-library/src/main/java/io/realm/RealmFieldType.java
kneth: As noted above, this enum is kept in sync with a similar enum in core. I rather not removed the three unused/unsupported types.

realm/realm-library/src/main/java/io/realm/internal/TableView.java
kneth: Because of inheritance

realm/realm-library/src/main/java/io/realm/internal/TableView.java
kneth: The `TableView` setters are not exposed in our public API, and I'm not sure if they can be called easily.

realm/realm-library/src/main/java/io/realm/BaseRealm.java
cmelchior: Ah yes. Missed that 👍

kneth:
@cmelchior We should definitely investigate our coverage to see if important methods are not tested. But in the case of the `TableView` setters, the conclusion might be that we should remove them.

cmelchior:
Removing unused setters / getters seems like a good idea. We really shouldn't have unused code. Probably in another PR though.

realm/realm-jni/src/io_realm_internal_TableQuery.cpp
```diff
@@ -1523,42 +1521,44 @@ JNIEXPORT jdouble JNICALL Java_io_realm_internal_TableQuery_nativeAverageDouble(
 
 
 // date aggregates
-
-JNIEXPORT jobject JNICALL Java_io_realm_internal_TableQuery_nativeMaximumDate(
+// FIXME: This is a rough workaround while waiting for https://github.com/realm/realm-core/issues/1720 to be solved
```
nhachicha: the issue you're pointing to is closed, are we still waiting for Core?

realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -602,6 +603,28 @@ public void onResult(int count) {
...
                     }
                 }
+                File managementFolder = new File(realmFolder, realmFileName + management);
+                if (managementFolder.exists()) {
+                    filesToDelete = Arrays.asList(
+                            new File(managementFolder, realmFileName + ".lock"),
+                            new File(managementFolder, realmFileName + ".log_a"),
+                            new File(managementFolder, realmFileName + ".log_b"),
+                            new File(managementFolder, realmFileName + ".log"));
+                    for (File fileToDelete : filesToDelete) {
```
nhachicha: this logic is duplicated (with the legacy _Realm files_ above) can we have a 
```java
boolean deletes (File rootFolder, String realmFileName) {
// ...
}
```
then we can call it with 
```java
deletes(realmFolder, realmFileName);
```
and 
```java
deletes(managementFolder, realmFileName);
```

realm/realm-jni/src/io_realm_internal_TableQuery.cpp
kneth: I'll update the comment to be a link to https://github.com/realm/realm-core/issues/1745

realm/realm-jni/build.gradle
```diff
@@ -201,7 +201,7 @@ task downloadCore(group: 'build setup', description: 'Download the latest versio
                     throw new GradleException("Invalid checksum for file '" +
                             "${project.coreArchiveFile.getName()}'. Expected " +
                             "${project.coreSha256Hash.toLowerCase()} but got " +
-                            "${calculatedHash.toLowerCase()}.");
+                                "${calculatedHash.toLowerCase()}.");
```
cmelchior: Wrong indent?

realm/realm-jni/src/io_realm_internal_TableQuery.cpp
cmelchior: Please update the link in the comment ☝️ 😄

realm/realm-jni/src/io_realm_internal_TableQuery.cpp
```diff
@@ -1523,42 +1523,44 @@ JNIEXPORT jdouble JNICALL Java_io_realm_internal_TableQuery_nativeAverageDouble(
...
-        !COL_INDEX_AND_TYPE_VALID(env, pTable, columnIndex, type_DateTime) ||
+        !COL_INDEX_AND_TYPE_VALID(env, pTable, columnIndex, type_Timestamp) ||
         !ROW_INDEXES_VALID(env, pTable, start, end, limit))
         return NULL;
     try {
         size_t return_ndx;
-        DateTime result = pQuery->minimum_int(S(columnIndex), NULL, S(start), S(end), S(limit), &return_ndx);
-        if (return_ndx != npos) {
-            return NewLong(env, result.get_datetime());
+        // FIXME: use Query::minimum_timestamp()
```
cmelchior: Remove fixme

realm/realm-jni/src/io_realm_internal_Util.cpp
```diff
@@ -78,7 +80,9 @@ JNIEXPORT jlong JNICALL Java_io_realm_internal_Util_nativeGetMemUsage(JNIEnv*, j
 JNIEXPORT jstring JNICALL Java_io_realm_internal_Util_nativeGetTablePrefix(
     JNIEnv* env, jclass)
 {
-    return to_jstring(env, string(TABLE_PREFIX));
+    string table_prefix(TABLE_PREFIX);
```
cmelchior: std:string is used in "internal_version" below. I am not sure we have a consensus on this? But at least being consistent in the same PR would be nice :)

I prefer std::string above string though.

realm/realm-jni/src/io_realm_internal_Version.cpp
```diff
@@ -34,7 +35,9 @@ JNIEXPORT jint JNICALL Java_io_realm_internal_Version_nativeGetAPIVersion(JNIEnv
 JNIEXPORT jstring JNICALL Java_io_realm_internal_Version_nativeGetVersion(JNIEnv *env, jclass)
 {
     try {
-        return to_jstring(env, Version::get_version());
+        std::string str(Version::get_version());
```
cmelchior: See above

realm/realm-jni/src/io_realm_internal_table.cpp
```diff
@@ -254,6 +259,9 @@ JNIEXPORT void JNICALL Java_io_realm_internal_Table_nativeConvertColumnToNullabl
                 case type_Table:
                     // checked previously
                     break;
+                case type_OldDateTime:
+                    // not used
```
cmelchior: Throw exception instead? In case someone accidentally uses it?

realm/realm-jni/src/io_realm_internal_table.cpp
```diff
@@ -374,6 +384,9 @@ JNIEXPORT void JNICALL Java_io_realm_internal_Table_nativeConvertColumnToNotNull
...
```
cmelchior: Same

realm/realm-jni/src/io_realm_internal_table.cpp
```diff
@@ -1285,18 +1299,20 @@ JNIEXPORT jlong JNICALL Java_io_realm_internal_Table_nativeFindAllBool(
     return reinterpret_cast<jlong>(pTableView);
 }
 
-JNIEXPORT jlong JNICALL Java_io_realm_internal_Table_nativeFindAllDate(
+// FIXME: reenable when find_first_timestamp() is implemented
```
cmelchior: This method is still out-commented. Is this not supported yet?

realm/realm-jni/src/io_realm_internal_table.cpp
```diff
@@ -1682,6 +1698,8 @@ JNIEXPORT void JNICALL Java_io_realm_internal_Table_nativeMigratePrimaryKeyTable
     const size_t CLASS_COLUMN_INDEX = io_realm_internal_Table_PRIMARY_KEY_CLASS_COLUMN_INDEX;
     const size_t FIELD_COLUMN_INDEX = io_realm_internal_Table_PRIMARY_KEY_FIELD_COLUMN_INDEX;
 
+    const string table_prefix(TABLE_PREFIX);
```
cmelchior: std::string

realm/realm-jni/src/io_realm_internal_tableview.cpp
```diff
@@ -204,10 +204,9 @@ JNIEXPORT jstring JNICALL Java_io_realm_internal_TableView_nativeGetColumnName
 JNIEXPORT jlong JNICALL Java_io_realm_internal_TableView_nativeGetColumnIndex
    (JNIEnv *env, jobject, jlong nativeViewPtr, jstring columnName)
 {
-    try {
-        if (!VIEW_VALID_AND_IN_SYNC(env, nativeViewPtr))
+    if (!VIEW_VALID_AND_IN_SYNC(env, nativeViewPtr))
```
cmelchior: This change seems at odds with what you wrote below?

realm/realm-jni/src/io_realm_internal_tableview.cpp
```diff
@@ -603,6 +602,8 @@ JNIEXPORT jlong JNICALL Java_io_realm_internal_TableView_nativeFindFirstDouble(
     return 0;
 }
 
+// FIXME: find_first_timestamp() isn't implemented
```
cmelchior: Not implemented yet? Does it have any impact on the binding?

realm/realm-jni/src/io_realm_internal_tableview.cpp
```diff
@@ -684,6 +686,8 @@ JNIEXPORT jlong JNICALL Java_io_realm_internal_TableView_nativeFindAllDouble(
     return 0;
 }
 
+// FIXME: find_all_timestamp() isn't implemented
```
cmelchior: Same?

realm/realm-jni/src/util.hpp
```diff
@@ -669,6 +670,24 @@ inline jobject NewFloat(JNIEnv* env, float value)
     return env->NewObject(java_lang_float, java_lang_float_init, value);
 }
 
+inline jlong to_milliseconds(const realm::Timestamp& ts)
+{
+    // From core's reference implementation aka unit test
+    // FIXME: check for overflow
```
cmelchior: FIXME not addressed

realm/realm-library/src/androidTest/java/io/realm/RealmObjectTests.java
```diff
@@ -560,10 +560,8 @@ private CyclicType createCyclicData() {
 
     @Test
     public void dateType() {
-        long testDatesValid[] = {-1000, 0, 1000};
-        long testDatesLoosePrecision[] = {Long.MIN_VALUE, 1, 1001, Long.MAX_VALUE};
+        long testDatesValid[] = {Long.MIN_VALUE, -1001, -1000, -1, 0, 1, 1000, 1001, Long.MAX_VALUE};
```
cmelchior: 👍

realm/realm-library/src/androidTest/java/io/realm/ManagedRealmCollectionTests.java
```diff
@@ -491,12 +490,14 @@ public void avg_partialNullRows() {
 
     @Test
     public void maxDate() {
-        assertEquals(new Date(YEAR_MILLIS * 20 * 4), collection.maxDate(AllJavaTypes.FIELD_DATE));
+        assertEquals(TEST_SIZE, collection.size());
+        assertEquals(new Date(YEAR_MILLIS * 20 * (TEST_SIZE/2 - 1)), collection.maxDate(AllJavaTypes.FIELD_DATE));
```
cmelchior: Why did max/min date suddenly change?

realm/realm-library/src/androidTest/java/io/realm/RealmQueryTests.java
cmelchior: Any comment to this? Maybe just have a list of *all* types inside `RealmFieldType` in a comment, but only expose the supported types?

realm/realm-library/src/main/java/io/realm/RealmFieldType.java
cmelchior: But having them in user facing code serves no purpose. I can see some merit in mirroring the values, but we can just as well do that in Javadoc instead of actual code?

realm/realm-library/src/main/java/io/realm/internal/TableOrView.java
```diff
@@ -328,8 +328,6 @@
 
     TableView findAllDouble(long columnIndex, double value);
 
-    TableView findAllDate(long columnIndex, Date value);
```
cmelchior: Is this because of it not being supported by Core yet or is there some other reason? Perhaps just add a FIXME if it is the first?

realm/realm-library/src/main/java/io/realm/BaseRealm.java
kneth: Good idea!

cmelchior:
Done with review. Mostly minor stuff, but there seems to be a bunch of FIXME's around `find_all_date/find_first_date`. I am not sure we even use those in the public API, so maybe something we should consider removing all together? Don't have to be in this PR though.

realm/realm-jni/src/io_realm_internal_table.cpp
kneth: It should not be possible. All old Realm files are converted, and new Realm will only use `Timestamp`.

realm/realm-jni/src/io_realm_internal_Util.cpp
kneth: Removing the `std` name space will probably required that we write `std::size_t` many places.

realm/realm-jni/src/io_realm_internal_table.cpp
cmelchior: Then throwing an exception should be safe :) IMO I would much rather we threw exceptions in these cases as it will catch anyone accidentally using them in the future.

realm/realm-jni/src/io_realm_internal_table.cpp
kneth: No, and `find_all_datetime()` was actually a hack.

realm/realm-jni/src/io_realm_internal_tableview.cpp
kneth: Yes, indeed so.

realm/realm-jni/src/io_realm_internal_tableview.cpp
kneth: The impact is that `findFirstDate()` cannot be implemented.

realm/realm-jni/src/io_realm_internal_tableview.cpp
kneth: Yes. but I don't think our public API ever had a `FindAllDate()` on `RealmResults`.

realm/realm-jni/src/util.hpp
kneth: True, it's for dates roughly 400 years in the future.

realm/realm-library/src/androidTest/java/io/realm/ManagedRealmCollectionTests.java
kneth: It didn't but `4` is assuming a value of `TEST_SIZE` and the test would fail if `TEST_SIZE` was changed. I just made it a bit for robust.

realm/realm-library/src/main/java/io/realm/internal/TableOrView.java
kneth: Yes, it has to be removed as Core doesn't support it. Again, core has never supported it but a clever hack made it possible to trick a non-checking core to perform the operation.

realm/realm-jni/src/io_realm_internal_UncheckedRow.cpp
```diff
@@ -217,7 +217,7 @@ JNIEXPORT jlong JNICALL Java_io_realm_internal_UncheckedRow_nativeGetLinkView
     if (!ROW_VALID(env, ROW(nativeRowPtr)))
         return 0;
 
-    LinkView* link_view_ptr = LangBindHelper::get_linklist_ptr( *ROW( nativeRowPtr ), S( columnIndex) );
+    LinkViewRef* link_view_ptr = const_cast<LinkViewRef*>(&(LangBindHelper::get_linklist_ptr(*ROW(nativeRowPtr), S(columnIndex))));
     return reinterpret_cast<jlong>(link_view_ptr);
```
AndyDentFree: I don't understand how this ends up as a string and there's a lot of casting here. Unless we have a string buried as some kind of opaque type, this looks like a bug, which predates the change to Timestamp.

realm/realm-jni/src/io_realm_internal_table.cpp
```diff
@@ -1384,16 +1400,16 @@ JNIEXPORT jlong JNICALL Java_io_realm_internal_Table_nativeGetSortedView(
...
         case type_Double:
         case type_Float:
+        case type_Timestamp:
             try {
                 TableView* pTableView = new TableView( pTable->get_sorted_view(S(columnIndex), ascending != 0 ? true : false) );
                 return reinterpret_cast<jlong>(pTableView);
             } CATCH_STD()
         default:
-            ThrowException(env, IllegalArgument, "Sort is currently only supported on integer, boolean, double, float, String, and Date columns.");
+            ThrowException(env, IllegalArgument, "Invalid type - Only String, Date, boolean, byte, short, int, long and their boxed variants are supported.");
```
AndyDentFree: Was this deliberately re-worded to be the same as the `FindAllString` exception above or is it a copy-and-paste error? It loses the original's mention of **Sort**.

realm/realm-jni/src/io_realm_internal_table.cpp
```diff
@@ -1433,15 +1449,15 @@ JNIEXPORT jlong JNICALL Java_io_realm_internal_Table_nativeGetSortedViewMulti(
...
             case type_String:
             case type_Double:
             case type_Float:
+            case type_Timestamp:
                 indices[i] = S(long_arr[i]);
                 ascendings[i] = S(bool_arr[i]);
                 break;
             default:
-                ThrowException(env, IllegalArgument, "Sort is currently only supported on integer, boolean, double, float, String, and Date columns.");
+                ThrowException(env, IllegalArgument, "Invalid type - Only String, Date, byte, boolean, short, int, long and their boxed variants are supported.");
```
AndyDentFree: Same question as on line 1412 above about losing Sort in the message.

realm/realm-jni/src/io_realm_internal_table.cpp
```diff
@@ -1710,9 +1728,11 @@ JNIEXPORT void JNICALL Java_io_realm_internal_Table_nativeMigratePrimaryKeyTable
     size_t number_of_rows = pk_table->size();
     for (size_t row_ndx = 0; row_ndx < number_of_rows; row_ndx++) {
         StringData table_name = pk_table->get_string(CLASS_COLUMN_INDEX, row_ndx);
-        if (table_name.begins_with(TABLE_PREFIX)) {
+        if (table_name.begins_with(table_prefix)) {
             // New string copy is needed, since the original memory will be changed.
-            pk_table->set_string(CLASS_COLUMN_INDEX, row_ndx, std::string(table_name.substr(strlen(TABLE_PREFIX))));
+            std::string str(table_name.substr(strlen(TABLE_PREFIX)));
```
AndyDentFree: As we have already copied this into `table_prefix` can use `table_prefix.length()` instead of recalculating `strlen`

realm/realm-jni/src/io_realm_internal_Util.cpp
AndyDentFree: There's a common pattern of creating a stack `std::string table_prefix` to wrap the compile-time constant `TABLE_PREFIX`. Could this be a file-level `static`, or at least locally a `static`, or would you prefer to leave the compiler to optimise that (which I'm a bit doubtful about)?

realm/realm-jni/src/io_realm_internal_table.cpp
bmunkholm: If it's not possible - it should assert.

realm/realm-library/src/androidTest/java/io/realm/RealmQueryTests.java
kneth: The trick is that when you add a new type, your IDE will be unhappy in all the `switch` constructors so you remember to update them all.

realm/realm-jni/src/io_realm_internal_UncheckedRow.cpp
kneth: I am not quite sure what you mean: no strings are involved there. With core 0.100.0, `LangBindHelper::get_linklist_ptr()` has been changed to return a `const LinkViewRef` (a `shared_ptr<LinkView*>`. As we cannot use a `shared_ptr` in Java, I have to store a pointer to `shared_ptr` (returned as a `jlong`). And I need to get rid of the constness too.

That gives us a double indirection which you can see in all the `LinkView` methods.

realm/realm-jni/src/io_realm_internal_table.cpp
kneth: It's not copy-and-paste error but copy-and-paste laziness :-)

realm/realm-jni/src/io_realm_internal_table.cpp
kneth: Agree. Using `char*` is probably too much C style these days.

realm/realm-jni/src/io_realm_internal_UncheckedRow.cpp
AndyDentFree: Sorry, scratch my comment - I was misinterpreting the github diff as the return type being a jstring - hadn't noticed the line number jump that it was a different method!

AndyDentFree:
👍  preferably with those exception strings fixed to restore the word _sorting_

cmelchior:
👍 , but please create an issue with the overflow problem with dates coming from other platforms. As a minimum we need to document it somewhere instead of hiding it in a FIXME in the code.

kneth:
@cmelchior An exception is thrown (https://github.com/realm/realm-java/pull/2679#discussion_r61776462)

And I have created #2722.

cmelchior:
👍

Issue #2680, opened 2016-04-26
Updating realm to 0.89.0 RealmList#first() throw IndexOutOfBoundsException at RealmList size is 0.


keiji:
Empty RealmList#first() throw IndexOutOfBoundsException instead of return null.

This issue is very similar to below issue.

https://github.com/realm/realm-java/issues/2659

#### Expected Results

RealmList#first() result is null;

    /**
     * Finds the first object.
     *
     * @return the first object or {@code null} if the list is empty.
     * @throws IllegalStateException if Realm instance has been closed or parent object has been removed.
     */
    public E first() {

#### Actual Results

    Process: io.keiji.kehai.debug, PID: 13565
    java.lang.IndexOutOfBoundsException: The list is empty.
     at io.realm.RealmList.first(RealmList.java:463)

#### Code Sample

```java

// setting.getList().size() is 0.
final WiFiApSetting wifiSetting = setting.getList().first();

if (wifiSetting == null) {
    return;
}

```

#### Version of Realm and tooling
Realm version(s): 0.89.1

Android Studio version: 2.0 (Stable)
gradle 2.0.0
buildToolsVersion 23.0.2

Which Android version and device: Nexus 5X (Android 6.0.1)

zaki50:
@keiji Thank you for reporting this.

This is an intended change related to introducing `OrderedRealmCollection`.
We should clearly state that in `CHANGELOG.md`.

If you are using `first()` against possible empty `RealmList `, please check that with `isEmpty()`.

Zhuinden:
Well the docs do say

     * @return the first object or {@code null} if the list is empty.

cmelchior:
@Zhuinden Ups, look like mistake. That Javadoc should have been removed. You can see the correct behavior described in `OrderedRealmCollection`

tbsandee:
I got tripped up on this too recently. Javadocs for RealmList are still wrong for first() and last().

PR #2656, opened 2016-04-24
Add bullet to changelog about new sorting behaviour on collections.

cmelchior:
Fixes https://github.com/realm/realm-java/issues/2655

@realm/java

PR #2682, opened 2016-04-26
Add bullet to changelog about behavior change of RealmList.first() and RealmList.last()

zaki50:
#2680 

plus, this PR includes back-port of #2656 (Add bullet to changelog about new sorting behaviour on collections.)


zaki50:
this PR should target `releases` branch.

PR #2692, opened 2016-04-27
Asset file as an initial dataset

thesurix:
Support for Realm asset file as an initial dataset.


kneth:
Please use our template when creating issues.

You can take a look at https://github.com/realm/realm-java/blob/master/examples/migrationExample/src/main/java/io/realm/examples/realmmigrationexample/MigrationExampleActivity.java for inspiration (method `copyBundledRealmFile`).

cmelchior:
Hi @thesurix. Thanks for adding this. I'll review more throughly later, but it seems that Findbugs complains about something in the PR:

```
* What went wrong:
Execution failed for task ':realm-library:findbugs'.
> FindBugs rule violations were found. See the report at: file:///home/jenkins/jenkins/workspace/android-pr/realm/realm-library/build/findbugs/findbugs-output.html
```

You can run that locally as well.

emanuelez:
```
io.realm.RealmCache.copyAssetFile(RealmConfiguration) may fail to clean up java.io.OutputStream on checked exception
Bug type OBL_UNSATISFIED_OBLIGATION_EXCEPTION_EDGE (click for details) 
In class io.realm.RealmCache
In method io.realm.RealmCache.copyAssetFile(RealmConfiguration)
Reference type java.io.OutputStream
1 instances of obligation remaining
Obligation to clean up resource created at RealmCache.java:[line 281] is not discharged
Path continues at RealmCache.java:[line 282]
Path continues at RealmCache.java:[line 284]
Remaining obligations: {OutputStream x 1}
```

thesurix:
@cmelchior @emanuelez 
Thanks for hints, I will check this.

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -248,4 +265,28 @@ static synchronized void invokeWithGlobalRefCount(RealmConfiguration configurati
         }
         callback.onResult(totalRefCount);
     }
+
+    /**
+     * Copies Realm database file from Android asset directory to the directory given in the {@link RealmConfiguration}.
+     * Copy is performed only at the first time when there is no Realm database file.
+     *
+     * @param configuration configuration object for Realm instance.
+     * @throws IOException if the asset file is not valid.
```
kneth: It sounds like you validate the Realm file. Maybe "if copying the file fails." instead.

realm/realm-library/src/main/java/io/realm/RealmConfiguration.java
```diff
@@ -151,6 +159,23 @@ RealmProxyMediator getSchemaMediator() {
     }
 
     /**
+     * Returns input stream object to the Realm asset file.
+     *
+     * @return input stream to the asset file.
+     * @throws IOException if the asset file is not valid.
```
kneth: Similar to my comment above.

realm/realm-library/src/main/java/io/realm/RealmConfiguration.java
```diff
@@ -353,6 +380,27 @@ public Builder(Context context) {
             initializeBuilder(context.getFilesDir());
         }
 
+        /**
+         * Creates an instance of the Builder for the RealmConfiguration.
+         *
+         * This will use the apps own internal directory for storing the Realm file. This does not require any
```
kneth: "apps" -> "app's"

realm/realm-library/src/androidTest/java/io/realm/RealmConfigurationTests.java
```diff
@@ -764,19 +766,77 @@ public void initialDataTransactionAssetFile() throws IOException {
...
+    @Test
+    public void assetFileValidFile() throws IOException {
+        // Remove default instance
+        Realm.deleteRealm(defaultConfig);
+
+        Context context = InstrumentationRegistry.getInstrumentation().getContext();
+        RealmConfiguration configuration = new RealmConfiguration.Builder(context, "asset_file.realm")
+                .build();
+        Realm.deleteRealm(configuration);
+
```
kneth: Asserting that the file doesn't exist might be a good idea.

kneth:
@thesurix Nice work. I've added a few minor comments.

realm/realm-library/src/main/java/io/realm/RealmConfiguration.java
```diff
@@ -353,6 +380,27 @@ public Builder(Context context) {
...
+        /**
+         * Creates an instance of the Builder for the RealmConfiguration.
+         *
+         * This will use the app's own internal directory for storing the Realm file. This does not require any
+         * additional permissions. The default location is {@code /data/data/<packagename>/files}, but can
+         * change depending on vendor implementations of Android.
+         *
+         * Initial dataset will be imported from the provided Realm asset file.
+         *
+         * @param context Android context.
```
cmelchior: I would call this "Android application context", just so it is more clear that you shouldn't use the Activity context.

realm/realm-library/src/main/java/io/realm/RealmConfiguration.java
```diff
@@ -353,6 +380,27 @@ public Builder(Context context) {
...
         }
 
+        /**
+         * Creates an instance of the Builder for the RealmConfiguration.
+         *
+         * This will use the app's own internal directory for storing the Realm file. This does not require any
+         * additional permissions. The default location is {@code /data/data/<packagename>/files}, but can
+         * change depending on vendor implementations of Android.
+         *
+         * Initial dataset will be imported from the provided Realm asset file.
```
cmelchior: I would rephrase this to: 

When opening the Realm for the first time, instead of creating an empty file, the Realm file will be copied from the provided assets file and used instead. WARNING: This could potentially be a lengthy operation so should ideally be done on a background thread.

realm/realm-library/src/main/java/io/realm/RealmConfiguration.java
```diff
@@ -151,6 +159,23 @@ RealmProxyMediator getSchemaMediator() {
...
     /**
+     * Returns input stream object to the Realm asset file.
+     *
+     * @return input stream to the asset file.
+     * @throws IOException if copying the file fails.
+     */
+    InputStream getAssetFile() throws IOException {
+        if (contextWeakRef != null) {
+            Context context = contextWeakRef.get();
+            if (context != null) {
```
cmelchior: The only reason this should be null is if the Activity context was used instead, so in that case we should throw an IllegalStateException or RealmException saying that the application context should be used, otherwise all the import logic will fail silently as far as I can tell.

realm/realm-library/src/main/java/io/realm/RealmConfiguration.java
```diff
@@ -151,6 +159,23 @@ RealmProxyMediator getSchemaMediator() {
...
+                return context.getAssets().open(assetFilePath);
+            }
+        }
+
+        return null;
```
cmelchior: See above. Throw exception instead of failing silently.

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -109,7 +118,11 @@ private RealmCache(RealmConfiguration config) {
             // Create a new local Realm instance
             BaseRealm realm;
 
+
             if (realmClass == Realm.class) {
+                if (configuration.getDurability() != SharedGroup.Durability.MEM_ONLY) {
```
cmelchior: I would move this check to `inMemory()` in the builder instead and throw an exception there instead.

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -248,4 +261,49 @@ static synchronized void invokeWithGlobalRefCount(RealmConfiguration configurati
...
+    /**
+     * Copies Realm database file from Android asset directory to the directory given in the {@link RealmConfiguration}.
+     * Copy is performed only at the first time when there is no Realm database file.
+     *
+     * @param configuration configuration object for Realm instance.
+     * @throws IOException if copying the file fails.
+     */
+    private static void copyAssetFile(RealmConfiguration configuration) {
+        File realmFile = new File(configuration.getRealmFolder(), configuration.getRealmFileName());
+        if (!realmFile.exists()) {
```
cmelchior: This precondition should be checked in the constructor of the builder instead and throw an exception if it fails.

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -248,4 +261,49 @@ static synchronized void invokeWithGlobalRefCount(RealmConfiguration configurati
...
+            InputStream inputStream = null;
+            FileOutputStream outputStream = null;
+            try {
+                inputStream = configuration.getAssetFile();
+                if (inputStream != null) {
```
cmelchior: throw exception is `inpuStream == null`

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -109,7 +118,11 @@ private RealmCache(RealmConfiguration config) {
             // Create a new local Realm instance
             BaseRealm realm;
 
+
             if (realmClass == Realm.class) {
+                if (configuration.getDurability() != SharedGroup.Durability.MEM_ONLY) {
+                    copyAssetFile(configuration);
```
beeender: I think line 111 is a better place to call `copyAssetFile(configuration);`. Line 111 will run when the first `Realm`/`DynamicRealm` created across all threads. Here is the first typed `Realm.getInstance()` called in the current thread.

cmelchior:
Hi @thesurix Done with round 1 of reviews. Generally looks good, but I had a few comments about throwing exceptions instead of failing silently.

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -248,4 +261,49 @@ static synchronized void invokeWithGlobalRefCount(RealmConfiguration configurati
...
+                    byte[] buf = new byte[1024];
+                    int bytesRead;
+                    while ((bytesRead = inputStream.read(buf)) > -1) {
+                        outputStream.write(buf, 0, bytesRead);
+                    }
+                    outputStream.close();
+                    inputStream.close();
+                }
+            } catch (IOException e) {
+                throw new RealmIOException("Could not resolve the path to the Realm asset file.");
```
beeender: Better to add the original `IOException` as cause as well.

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -248,4 +261,49 @@ static synchronized void invokeWithGlobalRefCount(RealmConfiguration configurati
...
+            } finally {
+                if (inputStream != null) {
+                    try {
+                        inputStream.close();
+                    } catch (IOException e) {
+                        // Ignore this exception because any significant errors should already have been handled
```
beeender: I think we should throw, `IOException` is something we can never get recovered from.

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -248,4 +261,49 @@ static synchronized void invokeWithGlobalRefCount(RealmConfiguration configurati
...
+    private static void copyAssetFile(RealmConfiguration configuration) {
+        File realmFile = new File(configuration.getRealmFolder(), configuration.getRealmFileName());
+        if (!realmFile.exists()) {
+            InputStream inputStream = null;
+            FileOutputStream outputStream = null;
+            try {
+                inputStream = configuration.getAssetFile();
+                if (inputStream != null) {
+                    outputStream = new FileOutputStream(realmFile);
+                    byte[] buf = new byte[1024];
```
beeender: 1024 might not be the most optimized length, i doubt 4096 would be faster and bigger buffer should be acceptable even on very low-end devices.

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -248,4 +261,49 @@ static synchronized void invokeWithGlobalRefCount(RealmConfiguration configurati
...
     }
+
+    /**
+     * Copies Realm database file from Android asset directory to the directory given in the {@link RealmConfiguration}.
+     * Copy is performed only at the first time when there is no Realm database file.
+     *
+     * @param configuration configuration object for Realm instance.
+     * @throws IOException if copying the file fails.
+     */
+    private static void copyAssetFile(RealmConfiguration configuration) {
```
beeender: -> `copyAssetFileIfNeeded`?

realm/realm-library/src/main/java/io/realm/RealmCache.java
thesurix: Yes we can throw but then we will mask original exception, I think it is better to throw original than this.

thesurix:
@cmelchior @beeender 
PR is ready, please review.

kneth:
Looks fine to me!

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -15,12 +15,21 @@
  */
 package io.realm;
 
+import android.util.Log;
```
cmelchior: This doesn't seem to be used

realm/realm-library/src/main/java/io/realm/RealmCache.java
cmelchior: Ignoring IOExceptions on close seems fine to me or what is your concern @beeender ?

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -248,4 +260,49 @@ static synchronized void invokeWithGlobalRefCount(RealmConfiguration configurati
...
+    /**
+     * Copies Realm database file from Android asset directory to the directory given in the {@link RealmConfiguration}.
+     * Copy is performed only at the first time when there is no Realm database file.
+     *
+     * @param configuration configuration object for Realm instance.
+     * @throws IOException if copying the file fails.
+     */
+    private static void copyAssetFileIfNeeded(RealmConfiguration configuration) {
+        if (configuration.hasAssetFile()) {
+            File realmFile = new File(configuration.getRealmFolder(), configuration.getRealmFileName());
```
cmelchior: I know we have checks elsewhere where we verify the file doesn't exist, but all those situations look racy to me, i.e you might accidentally create the file between the configuration is created and this method is called, e.g. the Backup/Restore API might reintroduce the file I think. So IMO we should have a safety check here as well and just ignore the assetFile in that case.


```
File realmFile = new File(configuration.getRealmFolder(), configuration.getRealmFileName());
if (realmFile.exists()) {
  return;
}
```

realm/realm-library/src/main/java/io/realm/RealmConfiguration.java
```diff
@@ -151,6 +159,30 @@ RealmProxyMediator getSchemaMediator() {
...
+     *
+     * @return input stream to the asset file.
+     * @throws IOException if copying the file fails.
+     */
+    InputStream getAssetFile() throws IOException {
+        Context context = contextWeakRef.get();
+        if (context != null) {
+            return context.getAssets().open(assetFilePath);
+        } else {
+            throw new IllegalArgumentException("Context can not be null. Use application context instead.");
```
cmelchior: I would specify this a bit more. Something like 
"Context should not be null. Use Application Context instead of Activity Context."

cmelchior:
@thesurix Done with review. Generally looks good. Only minor issues left.

realm/realm-library/src/main/java/io/realm/RealmCache.java
beeender: Although close on InputStream here should never throw, closing outputSteam surely could throw `IOException` as it says in the javadoc `if an I/O error occurs.` . And when that happens (probably a disk failure), nothing would be reliable for the following operations. Generally, the process should stop before there some other exceptions happen with no clue.

realm/realm-library/src/main/java/io/realm/RealmConfiguration.java
```diff
@@ -353,6 +387,33 @@ public Builder(Context context) {
...
+         * change depending on vendor implementations of Android.
+         *
+         * When opening the Realm for the first time, instead of creating an empty file,
+         * the Realm file will be copied from the provided assets file and used instead.
+         * WARNING: This could potentially be a lengthy operation so should ideally be done on a background thread.
+         *
+         * @param context Android application context.
+         * @param assetFile path to the asset database file.
+         */
+        public Builder(Context context, final String assetFile) {
```
beeender: Some other thoughts about the API, would it be better that we just add a method `Build.assetFile()` instead? That gives user more freedom to choose the folder to store the file.

realm/realm-library/src/main/java/io/realm/RealmCache.java
thesurix: @beeender So as I understand, your suggestion is to put `throw new RealmIOException()` where we close `outputStream`?

realm/realm-library/src/main/java/io/realm/RealmConfiguration.java
thesurix: @beeender  From my perspective providing asset file without context is strange because we can get asset file only with App context. On the other hand we can provide `assetFile(context, assetFile)` or simply throw an exception in `assetFile()` if there is no provided context. What do you think?

cmelchior: I was thinking the same, but providing the context twice is also a bit weird.
Having `builder.assetFile(context, assetFile)` is probably a bit more in line with the rest of the class though, but I don't have a strong opinion on it. In either case it will be a bit weird.

Having only a `assetFile(path)` is a bad idea I think though as the interaction with the constructor would be really strange.

beeender: `builder.assetFile(context, assetFile)` +1

thesurix:
@cmelchior @beeender 
PR is ready

realm/realm-library/src/main/java/io/realm/RealmCache.java
beeender: @thesurix Yes, I would like to throw it after closing `outputStream`. @cmelchior what do you think?

beeender:
Except the IOException discussion, LGTM! :+1:

realm/realm-library/src/main/java/io/realm/RealmCache.java
cmelchior: Yes, throwing if outputStream fail makes a lot of sense. As @beeender says, if the Realm file is not in a proper state, all behavior is undefined.

realm/realm-library/src/main/java/io/realm/RealmConfiguration.java
```diff
@@ -515,6 +554,36 @@ public Builder initialData(Realm.Transaction transaction) {
...
+         * @param assetFile path to the asset database file.
+         */
+        public Builder assetFile(Context context, final String assetFile) {
+            if (context == null) {
+                throw new IllegalArgumentException("A non-null Context must be provided");
+            }
+            if (TextUtils.isEmpty(assetFile)) {
+                throw new IllegalArgumentException("A non-empty asset file path must be provided");
+            }
+            if (new File(folder, fileName).exists()) {
```
cmelchior: This check here seems kinda harsh. It prevents the use case where you might delete the file _after_ the configuration was created. I would just remove it, since this method is designed to "fail" silently anyway. There is also the check in `copyRealmFileIfNeeded` that prevents the copy anyway.

cmelchior:
Just needs to fix the IOException + remove a check in the the `assetFile()` method. Otherwise this looks 👍

beeender:
:+1: :tada:

cmelchior:
Looks good to me 👍 
Thanks for this @thesurix

PR #2693, opened 2016-04-28
APT is now correctly applied for unit tests in library projects.

cmelchior:
Apparently just having "apt" is not enough if you have a library project with unit tests that uses Realm.
Explicitly adding the annotation processor to `androidTestApt` fixes that.

I am guessing we also need to do the same for JVM tests with `testApt`, but since we don't support that yet. I didn't do that in this PR.

@realm/java

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,7 +23,7 @@
 
 * `RealmObjectSchema.getPrimaryKey()`. (#2636)
 * `Realm.createObject(Class, Object)` for creating objects with a primary key directly.
-* Unit tests in Android library projects now detects Realm model classes.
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -23,6 +23,7 @@
 
 * `RealmObjectSchema.getPrimaryKey()`. (#2636)
 * `Realm.createObject(Class, Object)` for creating objects with a primary key directly.
+* Unit tests in Android library projects now detects Realm model classes.
```
emanuelez: detects -> detect

emanuelez:
👍 apart from typo

zaki50:
:+1:

PR #2712, opened 2016-05-02
Use the motherboard identifier instead of the BIOS one for Windows machines

emanuelez:
The solution relying on the BIOS turned out to be not too solid.


emanuelez:
@realm/java

cmelchior:
👍

stk1m1:
👍

PR #2715, opened 2016-05-02
Better exception message when using custom migrations

cmelchior:
People not overriding `equals` in their custom Migration class is a common source of errors if people are creating multiple RealmConfiguration instances (which they shouldn't in the first place), but at least this gives better hints to what is wrong.

@realm/java

### DIFF
--- a/realm/realm-library/src/androidTest/java/io/realm/RealmConfigurationTests.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/RealmConfigurationTests.java
@@ -779,4 +779,27 @@ public class RealmConfigurationTests {
         realm.close();
         verify(transaction, never()).execute(realm);
     }
+
+    private static class MigrationWithNoEquals implements RealmMigration {
+        @Override
+        public void migrate(DynamicRealm realm, long oldVersion, long newVersion) {
+            // Do nothing
+        }
+    }
+
+    @Test
+    public void detectMissingEqualsInCustomMigration() {
+        RealmConfiguration config1 = configFactory.createConfigurationBuilder().migration(new MigrationWithNoEquals()).build();
+        RealmConfiguration config2 = configFactory.createConfigurationBuilder().migration(new MigrationWithNoEquals()).build();
+
+        Realm realm = Realm.getInstance(config1);
+        try {
+            Realm.getInstance(config2);
+            fail();
+        } catch (IllegalArgumentException e) {
+            assertTrue(e.getMessage().contains("The most likely cause are that equals() and hashCode() are not overridden"));
+        } finally {
+            realm.close();
+        }
+    }
 }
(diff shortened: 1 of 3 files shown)


dalinaum:
:+1:

realm/realm-library/src/androidTest/java/io/realm/RealmConfigurationTests.java
```diff
@@ -779,4 +779,27 @@ public void initialDataTransactionAssetFile() throws IOException {
...
+    public void detectMissingEqualsInCustomMigration() {
+        RealmConfiguration config1 = configFactory.createConfigurationBuilder().migration(new MigrationWithNoEquals()).build();
+        RealmConfiguration config2 = configFactory.createConfigurationBuilder().migration(new MigrationWithNoEquals()).build();
+
+        Realm realm = Realm.getInstance(config1);
+        try {
+            Realm.getInstance(config2);
+            fail();
+        } catch (IllegalArgumentException e) {
+            assertTrue(e.getMessage().contains("The most likely cause are that equals() and hashCode() are not overridden"));
```
zaki50: `cause are` -> `cause is` or `causes are`?

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -223,6 +223,16 @@ private void validateConfiguration(RealmConfiguration newConfiguration) {
         if (!Arrays.equals(configuration.getEncryptionKey(), newConfiguration.getEncryptionKey())) {
             throw new IllegalArgumentException(DIFFERENT_KEY_MESSAGE);
         } else {
+            // A common problem is that people are forgetting to override `equals` in their custom migration class.
+            // Try to detect this problem specifically so we can throw a better error message.
+            RealmMigration newMigration = newConfiguration.getMigration();
+            RealmMigration oldMigration = configuration.getMigration();
+            if (oldMigration != null && newMigration != null && oldMigration.getClass().equals(newMigration.getClass())) {
```
zaki50: We should have another condition `!newMigration.equals(oldMigration )`.

zaki50:
:+1: besides small comments.

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -223,6 +223,16 @@ private void validateConfiguration(RealmConfiguration newConfiguration) {
...
+                throw new IllegalArgumentException("Configurations cannot be different if used to open the same file. " +
+                        "The most likely cause are that equals() and hashCode() are not overridden in the " +
```
zaki50: `cause are` -> `cause is`?

beeender:
:+1:

kneth:
:+1:

Issue #2686, opened 2016-04-27
Memory Leak leads to Crash


Honghe:
Sorry, I am not very sure whether this Crash is caused by Realm, but Android Studio `Allocation Tracking` shows relevant information with Realm, so I past the main code.
Thanks!

#### Goal

> Fix a Memory Leak bug

#### Expected Results

> Fix

#### Actual Results

When the App is on the `MainActivity` (as bellow code) page, the memory is increased slowly, and then crashed.

![screenshot from 2016-04-27 18-32-00](https://cloud.githubusercontent.com/assets/1092722/14850628/efd5858c-0cac-11e6-81ac-13973a60faf0.png)

#### Steps & Code to Reproduce

> Describe your current debugging efforts.

#### Code Sample

MainActivity.java
```java

package top.honhe.server.activity;
import android.content.SharedPreferences;
import android.os.Bundle;
import android.os.Handler;
import android.os.Message;
import android.support.v7.app.AppCompatActivity;
import android.util.Log;
import android.widget.ListView;
import org.greenrobot.eventbus.EventBus;
import org.greenrobot.eventbus.Subscribe;
import org.greenrobot.eventbus.ThreadMode;
import java.util.Map;
import java.util.Timer;
import java.util.TimerTask;
import java.util.concurrent.ExecutorService;
import java.util.concurrent.Executors;
import io.realm.Realm;
import io.realm.RealmConfiguration;
import io.realm.RealmResults;
import top.honhe.server.R;
import top.honhe.server.adapter.BedAdapter;
import top.honhe.server.event.PingBedResultEvent;
import top.honhe.server.model.Bed;
import top.honhe.server.utils.Ping;
import top.honhe.server.utils.SharePreferenceUtil;

public class MainActivity extends AppCompatActivity{
    private static final int PING_BEDTASK_PERIOD = 8000;
    private static final int PING_BEDTASK_DELAY = 5000;
    private ExecutorService fixedThreadPool;
    private EventBus eventBus;
    private Realm realm;
    private SharedPreferences.OnSharedPreferenceChangeListener prefListener;
    private TimerTask timerTask;
    private  Timer timer;
    private static final int MSG_MONITOR_BED = 1;
    private final int nThreads = 10;
    public RealmConfiguration realmConf;
    private Handler handler = new Handler() {
        @Override
        public void handleMessage(Message msg) {
            super.handleMessage(msg);
            switch (msg.what) {
                case MSG_MONITOR_BED:
[...]

PR #2723, opened 2016-05-03
Add to realmObjects when addChangeListener called

### LINKED ISSUE #2569: RealmChangeListener onChange() method is not triggered
#### Goal

In Android, I am trying to update different views - such as `ViewHolder` in `RecyclerView` and a bunch of other custom views - that contain different instances of the logically-same object using a `RealmChangeListener`.

This is the object class:
```java
public class Post extends RealmObject {
    ...
}
```
I am currently using the [FastAdapter](https://github.com/mikepenz/FastAdapter) library so I have this `PostItem` class which is basically the item that is going to be stored inside the adapter and provides the logic to bind to a `ViewHolder`:
```java
public class PostItem extends AbstractItem<PostItem, PostItem.ViewHolder> {
    private Post mPost;
    private RealmChangeListener mRealmChangeListener;
    ...
    public PostItem withPost(Post post) {
        mPost = post;
        mRealmChangeListener = new RealmChangeListener() {
            @Override
            public void onChange() {
                // here I call a custom listener which should then update the Recycler 
                // of which it is responsible - something like:
                // myCustomListener.onRealmChange(PostItem.this, mPost)
            }
        };
        mPost.removeChangeListeners();
        mPost.addChangeListener(mRealmChangeListener);
    }
    ...
}
```
The user can then make some action such as liking or saving the post. Each time I am making a request to the server and upon success I take the object coming back in the responde and update it in the database in this way:
```java
public class LikePostTask extends AsyncTask<Void, Void, Post> {
    ...
    @Override
    protected void onPostExecute(Post post) {
        DomainController.getInstance().save(post);
    }
}

public class DomainController {
    ...
    public <T extends RealmObject> T save(final T object) {
        realm.beginTransaction();
        T realmObject = realm.copyToRealmOrUpdate(object);
        realm.commitTransaction();
        return realmObject;
    }
}
```

#### Expected Results

If I understand it correctly after the `save()` method is executed the `onChange()` method of the `RealmChangeListener` should be called upon the next iteration of the run loop on the UI thread.

#### Actual Results

However the `onChange()` method is not triggered. I tried putting a breakpoint inside it and it's never reached. 

The thing is the objects instances inside the `PostItem` objects are updated. After scrolling to force a rebind of the `ViewHolder` the UI is updated with the right information but again whatever I put inside the `onChange()` method is never executed.

Realm version(s): 0.88.2

Android Studio version: 2.1 preview 5

Which Android version and device: 6.0.1 on a Galaxy S6

beeender:
* Only add the RealmObject to HandlerController.realmObjects when
  addChangeListener called. This will avoid we have too many objects in
  the map.
* addToRealmObjects missed reference queue which would lead to the map
  grows always without cleaning.

Close #2569, potentially a fix to #2686 .

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,14 +25,14 @@
 
 ### Enhancements
 
-* `RealmObjectSchema.getPrimaryKey()`. (#2636)
+* `RealmObjectSchema.getPrimaryKey()` (#2636).
 * `Realm.createObject(Class, Object)` for creating objects with a primary key directly.
 * Unit tests in Android library projects now detect Realm model classes.
 * `Realm.waitForChange()`/`stopWaitForChange()` and `DynamicRealm.waitForChange()`/`stopWaitForChange()` (#2386).
 
 ### Bug fixes
 
-* `RealmChangeListener` on `RealmObject` is not triggered when adding listener on returned `RealmObject` of `copyToRealmOrUpdate()`. (#2569)
+* `RealmChangeListener` on `RealmObject` is not triggered when adding listener on returned `RealmObject` of `copyToRealmOrUpdate()` (#2569).
 
 ### Credits
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,10 @@
 * Unit tests in Android library projects now detect Realm model classes.
 * `Realm.waitForChange()`/`stopWaitForChange()` and `DynamicRealm.waitForChange()`/`stopWaitForChange()` (#2386).
 
+### Bug fixes
+
+* `RealmChangeListener` on `RealmObject` is not triggered when adding listener on returned `RealmObject` of `copyToRealmOrUpdate()`. (#2569)
+
 ### Credits
 
 * Thanks to Brenden Kromhout (@bkromhout) for adding `RealmObjectSchema.getPrimaryKey()`.
--- a/realm/realm-library/src/androidTest/java/io/realm/RealmObjectTests.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/RealmObjectTests.java
@@ -16,6 +16,8 @@
 
 package io.realm;
 
+import android.support.test.annotation.UiThreadTest;
+import android.support.test.rule.UiThreadTestRule;
 import android.support.test.runner.AndroidJUnit4;
 
 import org.junit.After;
@@ -29,6 +31,7 @@ import java.io.FileNotFoundException;
 import java.lang.ref.WeakReference;
 import java.util.Calendar;
 import java.util.Date;
+import java.util.Map;
 import java.util.concurrent.Callable;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutionException;
@@ -75,6 +78,8 @@ public class RealmObjectTests {
     public final ExpectedException thrown = ExpectedException.none();
     @Rule
     public final RunInLooperThread looperThread = new RunInLooperThread();
+    @Rule
+    public final UiThreadTestRule uiThreadTestRule = new UiThreadTestRule();
 
     private Realm realm;
     private RealmConfiguration realmConfig;
@@ -1675,14 +1680,15 @@ public class RealmObjectTests {
     @Test
     @RunTestInLooperThread
     public void addChangeListener_returnedObjectOfCopyToRealmOrUpdate() {
-        looperThread.realm.beginTransaction();
-        looperThread.realm.createObject(AllTypesPrimaryKey.class, 1);
+        Realm realm = looperThread.realm;
+        realm.beginTransaction();
+        realm.createObject(AllTypesPrimaryKey.class, 1);
 
         AllTypesPrimaryKey allTypesPrimaryKey = new AllTypesPrimaryKey();
         allTypesPrimaryKey.setColumnLong(1);
         allTypesPrimaryKey.setColumnFloat(42f);
-        allTypesPrimaryKey = looperThread.realm.copyToRealmOrUpdate(allTypesPrimaryKey);
-        looperThread.realm.commitTransaction();
+        allTypesPrimaryKey = realm.copyToRealmOrUpdate(allTypesPrimaryKey);
+        realm.commitTransaction();
 
         allTypesPrimaryKey.addChangeListener(new RealmChangeListener<AllTypesPrimaryKey>() {
             @Override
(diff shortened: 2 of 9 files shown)


beeender:
@realm/java

realm/realm-library/src/androidTest/java/io/realm/RealmObjectTests.java
```diff
@@ -1668,4 +1670,107 @@ public void removeChangeListeners_throwOnUnmanagedObject() {
             looperThread.testComplete();
         }
     }
+
+    // Bug https://github.com/realm/realm-java/issues/2569
+    @Test
+    @RunTestInLooperThread
+    public void addChangeListener_returnedObjectOfCopyToRealmOrUpdate() {
+        looperThread.realm.beginTransaction();
```
cmelchior: In other tests we use `Realm realm = looperThread.realm;` it makes the test easier to read.

realm/realm-library/src/androidTest/java/io/realm/RealmObjectTests.java
```diff
@@ -1668,4 +1670,107 @@ public void removeChangeListeners_throwOnUnmanagedObject() {
...
+    }
+
+    // The object should be added to HandlerController.realmObjects only when the first time addListener called.
+    @Test
+    @RunTestInLooperThread
+    public void addChangeListener_shouldAddTheObjectToHandlerRealmObjects() {
+        looperThread.realm.beginTransaction();
+        AllTypesPrimaryKey allTypesPrimaryKey = looperThread.realm.createObject(AllTypesPrimaryKey.class, 1);
+        looperThread.realm.commitTransaction();
+        assertTrue(looperThread.realm.handlerController.realmObjects.isEmpty());
```
cmelchior: pull `realm.handlerController.realmObjects` into it's own variable since it is used a number of times

realm/realm-library/src/androidTest/java/io/realm/RealmObjectTests.java
```diff
@@ -1668,4 +1670,107 @@ public void removeChangeListeners_throwOnUnmanagedObject() {
...
+            public void onChange(AllTypesPrimaryKey element) {
+                assertEquals(42.0f, element.getColumnFloat(), 0f);
+                looperThread.testComplete();
+            }
+        });
+    }
+
+    // The object should be added to HandlerController.realmObjects only when the first time addListener called.
+    @Test
+    @RunTestInLooperThread
```
cmelchior: In this case where you don't need the actual looper behaviour, you can just use @UiThreadTest from the standard Android test rules.

realm/realm-library/src/main/java/io/realm/RealmObject.java
```diff
@@ -244,6 +244,12 @@ public final boolean load() {
             if (!listeners.contains(listener)) {
                 listeners.add(listener);
             }
+            if (isLoaded(object)) {
```
cmelchior: Use proxy instead?

realm/realm-library/src/main/java/io/realm/RealmObject.java
```diff
@@ -244,6 +244,12 @@ public final boolean load() {
...
+                // Try to add this object to the realmObjects if it has already been loaded.
+                // For newly created async objects, it will be handled in RealmQuery.findFirstAsync &
+                // HandlerController.completedAsyncRealmObject.
+                realm.handlerController.addToRealmObjects((RealmObjectProxy)object);
```
cmelchior: You already cast object to proxy, just use that instead?

CHANGELOG.md
```diff
@@ -30,6 +30,10 @@
 * Unit tests in Android library projects now detect Realm model classes.
 * `Realm.waitForChange()`/`stopWaitForChange()` and `DynamicRealm.waitForChange()`/`stopWaitForChange()` (#2386).
 
+### Bug fixes
+
+* `RealmChangeListener` on `RealmObject` is not triggered when adding listener on returned `RealmObject` of `copyToRealmOrUpdate()`. (#2569)
```
kneth: ". (#2569)" -> "(#2569)."

cmelchior:
Apart from a few comments 👍

realm/realm-library/src/androidTest/java/io/realm/RealmObjectTests.java
```diff
@@ -1668,4 +1670,107 @@ public void removeChangeListeners_throwOnUnmanagedObject() {
...
+    public void addChangeListener_shouldAddTheObjectToHandlerRealmObjects() {
+        looperThread.realm.beginTransaction();
+        AllTypesPrimaryKey allTypesPrimaryKey = looperThread.realm.createObject(AllTypesPrimaryKey.class, 1);
+        looperThread.realm.commitTransaction();
+        assertTrue(looperThread.realm.handlerController.realmObjects.isEmpty());
+
+        allTypesPrimaryKey.addChangeListener(new RealmChangeListener<RealmModel>() {
```
nhachicha: please use `AllTypesPrimaryKey` instead of `RealmModel`. We can't infer the type correctly for `realmObject.addChangeListener` this is why we use `<T extends RealmModel>` in the  method definition

realm/realm-library/src/androidTest/java/io/realm/RealmObjectTests.java
```diff
@@ -1668,4 +1670,107 @@ public void removeChangeListeners_throwOnUnmanagedObject() {
...
+    public void addChangeListener_shouldNotAddDupEntriesToHandlerRealmObjects() {
+        looperThread.realm.beginTransaction();
+        AllTypesPrimaryKey allTypesPrimaryKey = looperThread.realm.createObject(AllTypesPrimaryKey.class, 1);
+        looperThread.realm.commitTransaction();
+        for (WeakReference<RealmObjectProxy> ref : looperThread.realm.handlerController.realmObjects.keySet()) {
+            assertFalse(ref.get() == allTypesPrimaryKey);
+        }
+
+        // Add different listeners twice
+        allTypesPrimaryKey.addChangeListener(new RealmChangeListener<RealmModel>() {
```
nhachicha: replace `RealmModel` by `AllTypesPrimaryKey`

realm/realm-library/src/androidTest/java/io/realm/RealmObjectTests.java
```diff
@@ -1668,4 +1670,107 @@ public void removeChangeListeners_throwOnUnmanagedObject() {
...
+            @Override
+            public void onChange(RealmModel element) {
+            }
+        });
+        allTypesPrimaryKey.addChangeListener(new RealmChangeListener<RealmModel>() {
```
nhachicha: replace `RealmModel` by `AllTypesPrimaryKey`

realm/realm-library/src/androidTest/java/io/realm/RealmObjectTests.java
```diff
@@ -1668,4 +1670,107 @@ public void removeChangeListeners_throwOnUnmanagedObject() {
...
+    // The object should not be added to HandlerController again after the async query loaded.
+    @Test
+    @RunTestInLooperThread
+    public void addChangeListener_checkHandlerRealmObjectsWhenCallingOnAsyncObject() {
+        looperThread.realm.beginTransaction();
+        looperThread.realm.createObject(AllTypesPrimaryKey.class, 1);
+        looperThread.realm.commitTransaction();
+
+        final AllTypesPrimaryKey allTypesPrimaryKey = looperThread.realm.where(AllTypesPrimaryKey.class).findFirstAsync();
+        allTypesPrimaryKey.addChangeListener(new RealmChangeListener<RealmModel>() {
```
nhachicha: replace `RealmModel` by `AllTypesPrimaryKey`

realm/realm-library/src/androidTest/java/io/realm/RealmObjectTests.java
```diff
@@ -1668,4 +1670,107 @@ public void removeChangeListeners_throwOnUnmanagedObject() {
...
+            @Override
+            public void onChange(RealmModel element) {
+                allTypesPrimaryKey.addChangeListener(new RealmChangeListener<RealmModel>() {
```
nhachicha: replace `RealmModel` by `AllTypesPrimaryKey`

kneth:
:+1:

realm/realm-library/src/main/java/io/realm/HandlerController.java
```diff
@@ -637,9 +637,17 @@ void addToRealmResults(RealmResults<? extends RealmModel> realmResults) {
         syncRealmResults.add(realmResultsWeakReference);
     }
 
-    // add to the list of RealmObject to be notified after a commit
-    <E extends RealmModel> void addToRealmObjects(E realmobject) {
-        realmObjects.put(new WeakReference<RealmObjectProxy>((RealmObjectProxy) realmobject), null);
+    // Add to the list of RealmObject to be notified after a commit.
+    // This method will check if the object exists in the list. It won't add the same object multiple times
+    <E extends RealmObjectProxy> void addToRealmObjects(E realmObject) {
+        for (WeakReference<RealmObjectProxy> ref : realmObjects.keySet()) {
```
nhachicha: this is `O(n)` which will traverse each time all the `RealmObject` just to avoid duplicate, I suggest to use an alternative data structure (ex: check our `IdentitySet`, that guarantees `O(1)` on retrieve)

beeender: The problem is `WeakReference` doesn't override `equals` and `hashcode`, so we need to extend the `WeakReference` to avoid traversing.

And also, realmObjects is not thread safe, we need to fix that. Created https://github.com/realm/realm-java/issues/2724 to track that. And this won't be fixed in this PR.

cmelchior:
👍

PR #2726, opened 2016-05-03
Nh/fix listener generics

nhachicha:
this is to avoid having the IDE auto-complete with `<? extends BaseRealm>` so instead we'll have
```java
realm.addChangeListener(new RealmChangeListener<Realm>() {
            @Override
            public void onChange(Realm element) {
                
            }
        });
```


cmelchior:
👍

stk1m1:
:+1:

PR #1867, opened 2015-12-01
Add support for Roborectric

dalinaum:
#904

* [ ] - Add the gradle script
* [ ] - Add the Robolectric test sample.
* [ ] - Apply static core lib.
* [ ] - Support Win
* [ ] - Support Linux

PR #2734, opened 2016-05-04
Support unit tests with the plugin

droidpl:
This change will inject also the dependency for jUnit plain tests in Android

### DIFF
--- a/gradle-plugin/src/main/groovy/io/realm/gradle/Realm.groovy
+++ b/gradle-plugin/src/main/groovy/io/realm/gradle/Realm.groovy
@@ -57,6 +57,8 @@ class Realm implements Plugin<Project> {
         } else {
             project.dependencies.add("apt", "io.realm:realm-annotations:${Version.VERSION}")
             project.dependencies.add("apt", "io.realm:realm-annotations-processor:${Version.VERSION}")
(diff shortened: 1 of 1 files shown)


cmelchior:
@droidpl I am not against adding this, but since you cannot run Realm on the JVM yet, you could not really use it, but perhaps you want to mock Realm in JVM tests?

👍 

Please review @realm/java

zaki50:
And we need to apply our transformer to jvm tests

cmelchior:
Ohh, good point. Forgot about that. That isn't supported by yet, right?

droidpl:
As far as I know with robolectric there is some way to make it work instead of mocking everything, that is why I am proposing it. What do you think?

cmelchior:
We have an issue here for tracking Robolectric support: https://github.com/realm/realm-java/issues/1867 , but as @zaki50 points out, right now the Transforms are not applied to JVM tests. Until that is fixed I don't think we should merge this PR. Sorry.

droidpl:
Sure, just keep it there and lets wait it to be fixed. Thanks :)

bmunkholm:
.

cmelchior:
Outdated. Closing

Issue #2323, opened 2016-02-23
HashCode not working correctly


cmelchior:
Right now we use the following for `hashCode`: 

```
        String realmName = realm.getPath();
        String tableName = row.getTable().getName();
        long rowIndex = row.getIndex();

        int result = 17;
        result = 31 * result + ((realmName != null) ? realmName.hashCode() : 0);
        result = 31 * result + ((tableName != null) ? tableName.hashCode() : 0);
        result = 31 * result + (int) (rowIndex ^ (rowIndex >>> 32));
        return result;
```

The problem is `rowIndex` might suddenly change through the use of `moveLastOver`. This will change the hashCode unexpectedly and break things like Set: https://docs.oracle.com/javase/7/docs/api/java/util/Set.html

Cocoa currently uses the PrimaryKey value (which is not allowed to change) if possible, otherwise fallback to the default hashcode implementation. This has the downside that `result.get(0).equals(result.get(0)) == false`.

Without persistent identifiers we have to accept one of the short-comings :cry: 

I would be tempted to say that our current approach is OK and we should just document that `hashCode` is not stable and can change. With #2196 soon ready, people have more tools to work around it anyway.

What do you think @realm/java ?

beeender:
Why don't just use row's raw pointer to compute the hashcode? then we don't even need to has the realm name and table name.

cmelchior:
The raw pointer can also change between TableViews as far as I know:
```
realm.allObjects(Foo.class).get(0).equals(realm.allObjects(Foo.class).get(0)) == false
```

beeender:
Then we should use the `TableView::get_source_ndx()` which is the index in the table

cmelchior:
Yes, but the source index changes with `moveLastOver`, which is why the current proposal is problematic.

cmelchior:
@stk1m1 I think the fix for this is to just document the behaviour in both Javadoc for RealmObject and on the website.

stk1m1:
Until we have an immutable `RealmObject` identifier, the behavior should remain documented. The issue is closed for now.

PR #2750, opened 2016-05-09
hashCode behavior javadoc

stk1m1:
Title : hashCode behavior Javadoc

The behavior of hashCode is documented as it is pointed in issue #2323. Until realm-java has an immutable identifier, the behavior of hashCode should be documented.

@realm/java please review.

### DIFF
--- a/realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmProxyClassGenerator.java
+++ b/realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmProxyClassGenerator.java
@@ -929,8 +929,8 @@ public class RealmProxyClassGenerator {
 
     /**
      * Currently, the hash value emitted from this could suddenly change as an object's index might
-     * alternate along with the changes of the Realm. It is therefore advisable not to solely
-     * rely on the hash value to identify an instance of RealmObject.
+     * alternate due to Realm Java using {@code Table#moveLastOver()}. Hash codes should therefore not
+     * be considered stable, i.e. don't save them in a HashSet or use them as a key in a HashMap.
      */
     private void emitHashcodeMethod(JavaWriter writer) throws IOException {
         if (metadata.containsHashCode()) {
--- a/realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmProxyClassGenerator.java
+++ b/realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmProxyClassGenerator.java
@@ -927,6 +927,11 @@ public class RealmProxyClassGenerator {
         writer.emitEmptyLine();
     }
 
+    /**
+     * Currently, the hash value emitted from this could suddenly change as an object's index might
+     * alternate along with the changes of the Realm. It is therefore advisable not to solely
+     * rely on the hash value to identify an instance of RealmObject.
+     */
     private void emitHashcodeMethod(JavaWriter writer) throws IOException {
         if (metadata.containsHashCode()) {
             return;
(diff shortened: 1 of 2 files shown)


realm/realm-library/src/main/java/io/realm/DynamicRealmObject.java
```diff
@@ -641,6 +641,13 @@ public RealmFieldType getFieldType(String fieldName) {
         return proxyState.getRow$realm().getColumnType(columnIndex);
     }
 
+    /**
+     * Currently, the hash value returned from this could suddenly change as the object's index might
+     * become different along with the changes of the Realm. Thus, it is  advisable not to solely rely
+     * on the hash value to identify an instance of {@link DynamicRealmObject}.
```
beeender: `it is  advisable` -> `it is advisable`

cmelchior: Instead of completely replacing the default javadoc for hashcode I would reuse the standard one and then add our caveats. Something like:

```
     * Returns an integer hash code for this object. By contract, any two
     * objects for which {@link #equals} returns {@code true} must return
     * the same hash code value. 
     *
     * <p>Note that because RealmObjects are "live" objects, they might be updated by writes from other threads. 
     * This means that the hash code is not stable and RealmObjects should not be used as keys in HashMap saved in a HashSet.
     *
     * @return this object's hash code.
     * @see #equals

Comment (stk1m1) [inline realm/realm-library/src/main/java/io/realm/DynamicRealmObject.java]: Will do. :+1:

File: realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmProxyClassGenerator.java [hunk 2]
```diff
@@ -927,6 +927,11 @@ private void emitToStringMethod(JavaWriter writer) throws IOException {
         writer.emitEmptyLine();
     }
 
+    /**
+     * Currently, the hash value emitted from this could suddenly change as an object's index might
+     * alternate along with the changes of the Realm. It is therefore advisable not to solely
```
cmelchior: The internal Javadoc don't need to document the public behaviour IMO, but more describe what is actually happening.
Something like "Currently, the hash value emitted from this could suddenly change as an object's index might change due to Realm Java using `move_last_over`. Hash codes should therefor not be considered stable, i.e. don't save them in a HashSet or use them as a key in a HashMap.


alternate -> " change due to other changes in the Realm

cmelchior:
We should also document this somewhere people can actually see it. IMO the Class description for `RealmObject` and on the website documentation.

stk1m1:
@cmelchior I'll do that as well.

stk1m1:
All comments are addressed.

beeender:
:+1: Unstable hashCode feels quite wrong, but we don't have a better option right now :( Doc is best thing can be done now

realm/realm-library/src/main/java/io/realm/DynamicRealmObject.java
```diff
@@ -641,6 +641,19 @@ public RealmFieldType getFieldType(String fieldName) {
         return proxyState.getRow$realm().getColumnType(columnIndex);
     }
 
+    /**
+     * Returns a hash code value for the {@link DynamicRealmObject} object.
+     * <p>
+     * By the general contract of {@link Object#hashCode()}, any two objects for which {@link #equals}
+     * returns {@code true} must return the same hash code value.
+     * <p>
+     * Note that a {@link RealmObject} is <em>live</em> object, and it might be updated by changes from
```
cmelchior: is a live object

cmelchior:
Apart from small comment 👍

kneth:
:+1:

PR #2751, opened 2016-05-09
Give async related vars better names

beeender:
asyncQueryExecutor and pendingQuery seem to be very confusing since they
are used for async transaction as well.


beeender:
wrong branch

PR #2752, opened 2016-05-09
Give async related vars better names

beeender:
asyncQueryExecutor and pendingQuery seem to be very confusing since they
are used for async transaction as well.

### DIFF
--- a/realm/realm-library/src/main/java/io/realm/Realm.java
+++ b/realm/realm-library/src/main/java/io/realm/Realm.java
@@ -1128,7 +1128,7 @@ public final class Realm extends BaseRealm {
         // to perform the transaction
         final RealmConfiguration realmConfiguration = getConfiguration();
 
-        final Future<?> pendingQuery = asyncQueryExecutor.submit(new Runnable() {
+        final Future<?> pendingTransaction = asyncTaskExecutor.submit(new Runnable() {
             @Override
             public void run() {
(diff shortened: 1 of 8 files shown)


beeender:
@realm/java

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -1128,7 +1128,7 @@ public RealmAsyncTask executeTransaction(final Transaction transaction, final Tr
         // to perform the transaction
         final RealmConfiguration realmConfiguration = getConfiguration();
 
-        final Future<?> pendingQuery = asyncQueryExecutor.submit(new Runnable() {
+        final Future<?> pendingTransaction = asyncTaskExecutor.submit(new Runnable() {
```
cmelchior: `pendingTask` ? A transaction would imply a write, no?

cmelchior:
Apart from small comment 👍

realm/realm-library/src/main/java/io/realm/Realm.java
beeender: That is a write, no? the function is `public RealmAsyncTask executeTransaction`

cmelchior: Do, yeah 👍

kneth:
:+1:

PR #2754, opened 2016-05-09
Java doc for RealmResults.sort

### LINKED ISSUE #2710: Regression: New RealmResults.sort() behavior doesn't carry over RealmChangeListeners to returned RealmResults
#### Behavior Regression/Possible Bug

In Realm 0.89.0, the `RealmResults.sort()` methods changed from sorting "in place" to returning a new, sorted, instance of `RealmResults`. 

In previous versions of Realm, if an instance of `RealmResults` had any `RealmChangeListener`s attached to it, those listeners would still be attached after a call to one of its `sort()` methods.

#### Expected Results

In Realm 0.89.0 an above, any `RealmChangeListener`s attached to a `RealmResults` instance whose `sort()` method is called should be "transferred" to the returned (sorted) instance of `RealmResults`, thus preserving the original behavior.

#### Actual Results

In Realm 0.89.0 and above, any `RealmChangeListener`s attached to an instance of `RealmResults` are not "transferred" to the new instance of `RealmResults` which is returned by the various `RealmResults.sort()` methods.

#### Other Comments

Since this is a fairly self-explanatory issue, I don't have a specific set of steps or code sample prepared.

I wouldn't be surprised if many haven't actually run into this regression yet–depending on what people are using their `RealmChangeListener`s to trigger, the effects of losing them due to sorting can range from subtle to blatant.

Even if it's decided that this is an acceptable regression in behavior, I think it might be worth documenting somewhere that `RealmChangeListener`s need to be reattached following a call to one of the `RealmResults.sort()` methods.

Realm version(s): 0.89.0 and above

Android Studio version: N/A

beeender:
Close #2710.

### DIFF
--- a/realm/realm-library/src/main/java/io/realm/RealmResults.java
+++ b/realm/realm-library/src/main/java/io/realm/RealmResults.java
@@ -329,6 +329,8 @@ public final class RealmResults<E extends RealmModel> extends AbstractList<E> im
 
     /**
      * {@inheritDoc}
+     *
+     * @return a new sorted {@link RealmResults}. The original {@link RealmResults} stays unchanged.
      */
     @Override
     public RealmResults<E> sort(String fieldName) {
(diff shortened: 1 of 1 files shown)


beeender:
wrong branch ....

PR #2769, opened 2016-05-10
Upgrading to core 0.100.2

kneth:
The storage engine (aka core) has been fixed a number of bugs.

@realm/java


cmelchior:
Shouldn't this be against `release` and `0.90.1`? Apart from that 👍

kneth:
@cmelchior Forgot we are doing a minor release only :-(

beeender:
Use #2771 instead.

Issue #2158, opened 2016-01-26
JavaDoc: Describe how to use `isValid` when using findFirstAsync to detect null


rfrost77:
First, premise, I use async methods with asObservable.

'findAllAsync' returns RealmResults with 'isLoaded' true whether it is empty or not.
So, I put 'filter(RealmResult::isLoaded)' and check empty in subscribe event.

'findFirstAsync' returns RealmObject whether it is null or not, but the object loaded instance is false when it is null, so I cannot know it is null or just not loaded.

If I put 'filter(RealmObject::isLoaded)' when it is null, I cannot get subscribe event. If I don't put the filter, I must get double event by design and have to execute logic for null object despite it is not null. I put code below for explaining this.

```
realm.where(BannerRealm.class).equalTo("type", type).findFirstAsync().asObservable()
                .filter(RealmObject::isLoaded).first()
                .subscribe(realmObject -> {
                    // It cannot get subscribe event when it is null.
                })
```

```
realm.where(BannerRealm.class).equalTo("type", type).findFirstAsync().asObservable()
                .subscribe(realmObject -> {
                    if(!realmObject.isLoaded()) {
                        // Do null logic
                    } else {
                        // Do not null logic
                    }
                    // I must get double event by design and execute unnecessary logic for null when it is not null.
                })
```

All of this above is same for 'isValid'. How can I null check for 'findFirstAsync' clear? or I think you should fix this.

cmelchior:
You can use `isValid` to check if the object is a valid RealmObject, so that can effectively work as a null-check:

```
realm.where(BannerRealm.class).equalTo("type", type).findFirstAsync().asObservable()
                .filter(result.isLoaded() && result.isValid())
                .first()
```

rfrost77:
If it filter like you mentioned, I cannot get event when it is null. I want get subscribe event both of it is null and not null situation. So, if I don't filter it, I should get event twice when it is null and not null, by realm design as I asked before. It causes unnecessary executing null logic when it is null.

How can I do 'findFirstAsync' exactly same with 'findAllAsync'?

```
realm.where(BannerRealm.class).equalTo("type", type).findFirstAsync().asObservable()
                .filter(result.isLoaded() && result.isValid()).first()
                .subscribe(realmObject -> {
                    // It cannot get subscribe event when it is null.
                })
```

```
realm.where(BannerRealm.class).equalTo("type", type).findAllAsync().asObservable()
                .filter(result.isLoaded() && result.isValid()).first()
[...]

PR #2775, opened 2016-05-11
A note on the behavior of RealmObject.isValid() added

stk1m1:
Title : A note on the behavior of RealmObject.isValid() 

When an invalid `RealmObject` is subscribed, no proper notification will be triggered for `rx.Observer`.

This PR addresses the issue #2158. 
@realm/java

### DIFF
--- a/realm/realm-library/src/main/java/io/realm/RealmObject.java
+++ b/realm/realm-library/src/main/java/io/realm/RealmObject.java
@@ -127,11 +127,18 @@ public abstract class RealmObject implements RealmModel {
      * Checks if the RealmObject is still valid to use i.e. the RealmObject hasn't been deleted nor has the
      * {@link io.realm.Realm} been closed. It will always return false for stand alone objects.
      *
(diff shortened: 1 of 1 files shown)


dalinaum:
👍

realm/realm-library/src/main/java/io/realm/RealmObject.java
```diff
@@ -127,6 +127,10 @@ public final void deleteFromRealm() {
      * Checks if the RealmObject is still valid to use i.e. the RealmObject hasn't been deleted nor has the
      * {@link io.realm.Realm} been closed. It will always return false for stand alone objects.
      *
+     * <p>Note that a notification for Observer will <em>not</em> be triggered when a {@link RealmQuery}
```
cmelchior: This sentence doesn't seem to match what is described here? https://github.com/realm/realm-java/issues/2158#issuecomment-174847313

We other methods we have embedded code samples so people can see how this can be used. Perhaps this was an idea here as well.

stk1m1:
review points are addressed.

kneth:
:+1:

PR #2780, opened 2016-05-11
Fix async transaction problem when async query already exist.

### LINKED ISSUE #1893: onSuccess might not be called with the right Realm version for async transaction

dalinaum:
Fixes #1893


CHANGELOG.md
```diff
@@ -2,10 +2,15 @@
 
 * Updated Realm Core to 0.100.2.
 
+### Breaking changes
+
+* Removed `Realm.executeTransaction(transaction, callback)`. Use `Realm.executeAsyncTransaction()` instead.
```
zaki50: executeAsyncTransaction -> executeTransactionAsync

realm/realm-library/src/androidTest/java/io/realm/RealmResultsTests.java
```diff
@@ -348,12 +348,12 @@ public void changeListener_syncIfNeeded_updatedFromOtherThread() {
         assertEquals(10, results.size());
 
         // 1. Delete first object from another thread.
-        realm.executeTransaction(new Realm.Transaction() {
+        realm.executeTransactionAsync(new Realm.Transaction() {
             @Override
             public void execute(Realm realm) {
-               realm.where(AllTypes.class).equalTo(AllTypes.FIELD_LONG, 0).findFirst().removeFromRealm();
+                realm.where(AllTypes.class).equalTo(AllTypes.FIELD_LONG, 0).findFirst().removeFromRealm();
```
zaki50: revert this

dalinaum: Should it be another PR?

dalinaum: Created #2781 for formatting.

realm/realm-library/src/androidTest/java/io/realm/RealmAsyncQueryTests.java
```diff
@@ -279,6 +280,117 @@ public void execute(Realm realm) {
         }, transactionCallback);
     }
 
+    // Test case for https://github.com/realm/realm-java/issues/1893
+    // Ensure that onSuccess might not be called with the right Realm version for async transaction.
```
beeender: `might not` -> `is`

CHANGELOG.md
```diff
@@ -2,10 +2,15 @@
 
 * Updated Realm Core to 0.100.2.
 
+### Breaking changes
+
+* Removed `Realm.executeTransaction(transaction, callback)`. Use `Realm.executeTransactionAsync()` instead.
```
beeender: remove deprecate methods should be in the next minor release

realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -66,6 +69,8 @@
     RealmSchema schema;
     Handler handler;
     HandlerController handlerController;
+    List<Runnable> asyncTransactionCallbacks = new ArrayList<Runnable>();
+    List<Future<?>> asyncTransactions = new ArrayList<Future<?>>();
```
beeender: `ArrayList` is not thread safe, but it is accessed in caller thread and the async transaction thread.

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -1344,6 +1248,13 @@ public void run() {
                                 });
                             }
                         }
+
+                        if (!handlerController.threadContainsAsyncQueries() && !asyncTransactionCallbacks.isEmpty()) {
+                            for (Runnable transactionCallback : asyncTransactionCallbacks) {
+                                handler.post(transactionCallback);
```
beeender: I am thinking about just call those callbacks in the next `HandlerController.realmChanged()` to make the logic simpler. What do you think?

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -172,6 +172,11 @@ static synchronized void release(BaseRealm realm) {
             return;
         }
 
+        realm.tidyPerformedTransactions();
+        if (!realm.asyncTransactions.isEmpty() || !realm.asyncTransactionCallbacks.isEmpty()) {
```
beeender: 1. I am not quite sure if it worth to add a list `asyncTransactions` just for printing some warnings
2. If it does, better to have this logic in `BaseRealm.doClose()`
@realm/java

realm/realm-library/src/main/java/io/realm/BaseRealm.java
beeender: I also would like to have those list in `HandlerController`

CHANGELOG.md
cmelchior: Yes, we shouldn't remove methods in patch releases, even if they have been deprecated.

realm/realm-library/src/androidTest/java/io/realm/RealmAsyncQueryTests.java
```diff
@@ -279,6 +280,117 @@ public void execute(Realm realm) {
...
+    @Test
+    @RunTestInLooperThread
+    public void executeTransactionAsync_asyncQuery() {
+        final Realm realm = looperThread.realm;
+        RealmResults<AllTypes> results = realm.where(AllTypes.class).findAllAsync();
+        looperThread.keepStrongReference.add(results);
```
cmelchior: I don't think this is needed

realm/realm-library/src/androidTest/java/io/realm/RealmAsyncQueryTests.java
```diff
@@ -279,6 +280,117 @@ public void execute(Realm realm) {
         }, transactionCallback);
     }
 
+    // Test case for https://github.com/realm/realm-java/issues/1893
+    // Ensure that onSuccess might not be called with the right Realm version for async transaction.
+    @Test
+    @RunTestInLooperThread
+    public void executeTransactionAsync_asyncQuery() {
+        final Realm realm = looperThread.realm;
+        RealmResults<AllTypes> results = realm.where(AllTypes.class).findAllAsync();
```
cmelchior: assert that size is 0 before beginning the async transaction

realm/realm-library/src/androidTest/java/io/realm/RealmAsyncQueryTests.java
```diff
@@ -279,6 +280,117 @@ public void execute(Realm realm) {
...
+        }, new Realm.Transaction.OnError() {
+            @Override
+            public void onError(Throwable error) {
+                fail();
+            }
+        });
+    }
+
+    // Test case for https://github.com/realm/realm-java/issues/1893
+    // Ensure that Realm.close() logs a warning if you close a Realm while an async transaction is pending.
```
cmelchior: if you close the Realm on the async worker thread.

realm/realm-library/src/androidTest/java/io/realm/RealmAsyncQueryTests.java
```diff
@@ -279,6 +280,117 @@ public void execute(Realm realm) {
...
+    @Test
+    @RunTestInLooperThread
+    public void executeTransactionAsync_closeWithPendingTransaction() {
+        final CountDownLatch queryLatch = new CountDownLatch(1);
+        final CountDownLatch closeLatch = new CountDownLatch(1);
+
+        Realm realm = looperThread.realm;
+        Realm.asyncTaskExecutor.submit(new Callable<Long>() {
```
cmelchior: I am not sure I understand why this needs to be run from within the asyncTaskExecutor?

realm/realm-library/src/androidTest/java/io/realm/RealmAsyncQueryTests.java
```diff
@@ -279,6 +280,117 @@ public void execute(Realm realm) {
...
+        RealmLog.add(testLogger);
+        realm.close();
+        RealmLog.remove(testLogger);
+        queryLatch.countDown();
+
+        String[] logs = testLogger.message.split("will ");
+        assertEquals("be closed with pending async transactions or callbacks.", logs[1]);
+
+        TestHelper.awaitOrFail(closeLatch);
+        looperThread.testComplete();
```
cmelchior: If this is the last statement it is usually an indication you can replace `@RunTestInLooperThread` with the standard `@UiThreadTest`

realm/realm-library/src/androidTest/java/io/realm/RealmAsyncQueryTests.java
```diff
@@ -279,6 +280,117 @@ public void execute(Realm realm) {
...
+            public void onSuccess() {
+            }
+        });
+
+        TestHelper.awaitOrFail(inTransactionLatch);
+
+        Realm.asyncTaskExecutor.submit(new Callable<Long>() {
+            @Override
+            public Long call() throws Exception {
+                TestHelper.awaitOrFail(asyncLatch);
```
cmelchior: Same as the other test. I don't think this is needed?

realm/realm-library/src/main/java/io/realm/RealmCache.java
cmelchior: I tend to agree, but this arose from a user having problems because it wasn't clear that closing the Realm would also kill any notifications.

Instead of creating new lists keeping track of this. Couldn't we just extend the RealmThreadPoolExecutor to keep track of this?

CHANGELOG.md
```diff
@@ -2,10 +2,15 @@
...
+
 ### Bug fixes
 
 * Opening a Realm while closing a Realm in another thread could lead to a race condition.
 * Automatic migration to the new file format could in rare circumstances lead to a crash.
+* `OnSuccess.OnSuccess()` might not be called with the right Realm version for async transaction (#1893).
```
kneth: "right" -> "correct"

realm/realm-library/src/androidTest/java/io/realm/RealmAsyncQueryTests.java
dalinaum: My intent is as following:
* `realm.executeTransactionAsync` is not finished.
* close the Realm instance on UI thread.
* After checking warning log, close the Realm instance on the async worker thread.

I modified the code and added the comment.

realm/realm-library/src/androidTest/java/io/realm/RealmAsyncQueryTests.java
dalinaum: Added.

realm/realm-library/src/main/java/io/realm/Realm.java
dalinaum: `HandlerController.realmChanged()` is invoked by `BaseRealm.commitTransaction` and `BaseRealm.refresh`. If we move this logic to `HandlerController.realmChanged()` and there is no other transactions and refreshes, callbacks won't be invoked.

I made `HandlerController.executeAsyncTransactionCallbacks` for `HandlerController.completedAsyncQueriesUpdate` and `executeTransactionAsync` instead. How about this?

realm/realm-library/src/main/java/io/realm/RealmCache.java
dalinaum: I moved the list to RealmThreadPoolExecutor and the checking routine to `BaseRealm.doClose()`.

dalinaum:
Updated. Please review this @realm/java

realm/realm-library/src/main/java/io/realm/internal/async/RealmThreadPoolExecutor.java
```diff
@@ -59,6 +64,12 @@ private RealmThreadPoolExecutor(int corePoolSize, int maxPoolSize) {
...
     }
 
+    public Future<?> submitTransaction(Runnable task) {
+        Future<?> future = this.submit(task);
+        transactions.add(future);
+        return future;
+    }
+
     @Override
     public Future<?> submit(Runnable task) {
```
cmelchior: This should be renamed `submitQuery` for clarity IMO

realm/realm-library/src/main/java/io/realm/internal/async/RealmThreadPoolExecutor.java
```diff
@@ -82,6 +93,27 @@ protected void beforeExecute(Thread t, Runnable r) {
         }
     }
 
+    @Override
+    protected void afterExecute(Runnable r, Throwable t) {
+        super.afterExecute(r, t);
+        tidyTransactions();
+    }
+
+    public boolean hasPendingTransactions() {
```
cmelchior: Please add JavaDoc to public methods, even though it the internal API it should still be well documented.

realm/realm-library/src/androidTest/java/io/realm/TestHelper.java
```diff
@@ -246,60 +246,77 @@ public static String getRandomString(int length) {
      */
     public static class TestLogger implements Logger {
 
+        public String previousMessage;
         public String message;
         public Throwable throwable;
 
+        private void backupMessage() {
+            if (this.message != null) {
```
cmelchior: Shouldn't it backup the previous message, no matter if it is null or not?

realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -66,6 +67,7 @@
     RealmSchema schema;
     Handler handler;
     HandlerController handlerController;
+    List<Runnable> asyncTransactionCallbacks = new ArrayList<Runnable>();
```
cmelchior: This shouldn't be needed anymore?

realm/realm-library/src/androidTest/java/io/realm/RealmAsyncQueryTests.java
```diff
@@ -194,7 +194,7 @@ public void onSuccess() {
             @Override
             public void onError(Throwable error) {
                 // Ensure we are giving developers quality messages in the logs.
-                assertEquals("Could not cancel transaction, not currently in a transaction.", testLogger.message);
+                assertEquals("Could not cancel transaction, not currently in a transaction.", testLogger.previousMessage);
```
cmelchior: Why do we need the previous message here? Instead of the current one?

realm/realm-library/src/main/java/io/realm/internal/async/RealmThreadPoolExecutor.java
```diff
@@ -82,6 +93,27 @@ protected void beforeExecute(Thread t, Runnable r) {
...
+        tidyTransactions();
+        return !transactions.isEmpty();
+    }
+
+    private void tidyTransactions() {
```
cmelchior: Perhaps `cleanupTransactions` instead of `tidyTransactions`

realm/realm-library/src/androidTest/java/io/realm/RealmAsyncQueryTests.java
dalinaum: There are two warning messages. The previous one is for `Could not cancel transaction, not currently in a transaction.` and the current one is for `closed with pending async transactions or callbacks.`.

realm/realm-library/src/main/java/io/realm/BaseRealm.java
dalinaum: `RealmThreadPoolExecutor.transactions` is for warnings and `BaseRealm.asyncTransactionCallbacks ` is for pending transactions. Although they can be moved into other classes such as `Realm` or `HandlerController`, this kind of list is still needed.

realm/realm-library/src/androidTest/java/io/realm/RealmAsyncQueryTests.java
```diff
@@ -281,6 +281,88 @@ public void execute(Realm realm) {
...
+            }
+        });
+
+        final TestHelper.TestLogger testLogger = new TestHelper.TestLogger();
+        RealmLog.add(testLogger);
+        realm.close();
+        RealmLog.remove(testLogger);
+
+        String[] logs = testLogger.message.split("will ");
+        assertEquals("be closed with pending async transactions or callbacks.", logs[1]);
```
cmelchior: Minor, but just use `contains` instead?

realm/realm-library/src/androidTest/java/io/realm/RealmAsyncQueryTests.java
```diff
@@ -281,6 +281,88 @@ public void execute(Realm realm) {
...
+                realm.close(); // It is for avoiding java.lang.IllegalStateException: It's not allowed to delete the file associated with an open Realm.
+                waitAsyncTransaction.countDown();
+            }
+        });
+
+        final TestHelper.TestLogger testLogger = new TestHelper.TestLogger();
+        RealmLog.add(testLogger);
+        realm.close();
+        RealmLog.remove(testLogger);
+        String[] logs = testLogger.message.split("will ");
```
cmelchior: minor, use contains?

realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -458,6 +458,11 @@ public void close() {
      * Closes the Realm instances and all its resources without checking the {@link RealmCache}.
      */
     void doClose() {
+        if (asyncTaskExecutor.hasPendingTransactions() || !handlerController.asyncTransactionCallbacks.isEmpty()) {
```
kneth: Since it is possible to deduce if we have pending transactions or callback (or both), we could log more precisely. It might help our user.

realm/realm-library/src/main/java/io/realm/internal/async/RealmThreadPoolExecutor.java
```diff
@@ -59,19 +64,46 @@ private RealmThreadPoolExecutor(int corePoolSize, int maxPoolSize) {
                 new ArrayBlockingQueue<Runnable>(QUEUE_SIZE));
     }
 
-    @Override
-    public Future<?> submit(Runnable task) {
+    /**
+     * Submits a runnable for execution a transaction.
```
kneth: "execution" -> "executing"

realm/realm-library/src/main/java/io/realm/internal/async/RealmThreadPoolExecutor.java
```diff
@@ -59,19 +64,46 @@ private RealmThreadPoolExecutor(int corePoolSize, int maxPoolSize) {
...
+     * @return a future representing pending completion of the task
+     */
+    public Future<?> submitQueryUpdate(Runnable task) {
         return super.submit(new BgPriorityRunnable(task));
     }
 
-    @Override
-    public <T> Future<T> submit(Callable<T> task) {
+    /**
+     * Submits a runnable for execution a query.
```
kneth: "execution" -> "executing"

realm/realm-library/src/main/java/io/realm/internal/async/RealmThreadPoolExecutor.java
```diff
@@ -38,6 +42,7 @@
     private boolean isPaused;
     private ReentrantLock pauseLock = new ReentrantLock();
     private Condition unpaused = pauseLock.newCondition();
+    private List<Future<?>> transactions = Collections.synchronizedList(new ArrayList<Future<?>>());
```
beeender: I still don't think use a list `transactions` is a good idea. The major problems right now:
1. It costs memory which is not needed
2. There is a leak until `Realm.close()` gets called. This list will grows without being cleaned up.

And also i am totally fine we remove the warning part out of the scope of this PR, instead of focusing on the shared group version part.

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -1323,14 +1255,14 @@ public void run() {
 
                         if (backgroundException != null) {
                             if (onError != null) {
-                                handler.post(new Runnable() {
+                                handlerController.asyncTransactionCallbacks.add(new Runnable() {
```
beeender: If there is an error, no need to wait for REALM_CHANGED, it can be posted to the handler right away.

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -1323,14 +1255,14 @@ public void run() {
...
                                     @Override
                                     public void run() {
                                         onError.onError(backgroundException);
                                     }
                                 });
                             } else {
-                                handler.post(new Runnable() {
+                                handlerController.asyncTransactionCallbacks.add(new Runnable() {
```
beeender: Same, post it right away.

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -1344,6 +1276,10 @@ public void run() {
                                 });
                             }
                         }
+
+                        if (!handlerController.threadContainsAsyncQueries()) {
+                            handlerController.executeAsyncTransactionCallbacks();
```
beeender: I would like to simplify the logic by just calling `executeAsyncTransactionCallbacks` in `HandlerController.realmChanged()` the else clause there.

realm/realm-library/src/main/java/io/realm/HandlerController.java
```diff
@@ -515,6 +517,17 @@ private void completedAsyncQueriesUpdate(QueryUpdateTask.Result result) {
 
             updateAsyncQueriesTask = null;
         }
+
+        executeAsyncTransactionCallbacks();
```
beeender: I would like to simplify the logic by just calling executeAsyncTransactionCallbacks in HandlerController.realmChanged() the else clause there.

realm/realm-library/src/main/java/io/realm/HandlerController.java
```diff
@@ -584,7 +597,7 @@ private void completedAsyncRealmObject(QueryUpdateTask.Result result) {
      * @return {@code true} if there is at least one (non GC'ed) instance of {@link RealmResults} {@code false}
      * otherwise.
      */
-    private boolean threadContainsAsyncQueries() {
+    boolean threadContainsAsyncQueries() {
```
beeender: According to my other comments, this could still be `private` 😁

realm/realm-library/src/main/java/io/realm/HandlerController.java
```diff
@@ -97,6 +97,8 @@
     final ConcurrentHashMap<WeakReference<RealmObjectProxy>, Object> realmObjects =
             new ConcurrentHashMap<WeakReference<RealmObjectProxy>, Object>();
 
+    final List<Runnable> asyncTransactionCallbacks = new ArrayList<Runnable>();
```
beeender: This has to be thread safe

realm/realm-library/src/main/java/io/realm/HandlerController.java
```diff
@@ -515,6 +517,17 @@ private void completedAsyncQueriesUpdate(QueryUpdateTask.Result result) {
...
+
+        executeAsyncTransactionCallbacks();
+    }
+
+    void executeAsyncTransactionCallbacks() {
+        if (!asyncTransactionCallbacks.isEmpty()) {
+            for (Runnable transactionCallback : asyncTransactionCallbacks) {
+                realm.handler.post(transactionCallback);
+            }
+            asyncTransactionCallbacks.clear();
```
beeender: Be careful about the `clear()` here. There is a race condition that the executor task just added something without running and gets cleared here.

realm/realm-library/src/androidTest/java/io/realm/RealmAsyncQueryTests.java
dalinaum: I applied `assertThat(X, containsString(Y))` instead.

cmelchior:
Updated for review @realm/java . Note this PR has changed scope a bit. Logging a warning if you close a Realm while still having pending async transactions will be handled in another PR (as it is not as critical).

This PR should now fix the following
- #1893 
- #2978 
- #3037 

I have rewritten how async transactions works. Before they did this:

1) commitOnBackgroundThread() -> post REALM_CHANGED -> post onSuccess runnable

The problem with that approach was that the REALM_CHANGED would be swallowed if async queries existed which means that the async transaction would call complete on an old version of the Realm (making it look like it didn't work).

Instead we now do this:

1) commitOnBackgroundThread -> post runnable that calls `HandlerController.handleAsyncTransactionCompleted(runnable)`

The special runnable is treated as a combined `REALM_CHANGED + Callback`, which makes it possible for us to queue up callbacks until we are finally able to trigger all of them. Unfortunately the Handler has poor support for this so it means that we no longer can detect if such a message is in the queue.

The downside of this is that it introduces a slight chance of such a event plus a real REALM_CHANGED event to be in the message queue at the same time. I considered that acceptable (since it can already happen today), and since preventing this will introduce more complexity to something that is already entirely to complex.

realm/realm-library/src/main/java/io/realm/internal/async/RealmThreadPoolExecutor.java
```diff
@@ -59,19 +61,46 @@ private RealmThreadPoolExecutor(int corePoolSize, int maxPoolSize) {
...
 
-    @Override
-    public Future<?> submit(Runnable task) {
+    /**
+     * Submits a runnable for executing a transaction.
+     *
+     * @param task the task to submit
+     * @return a future representing pending completion of the task
+     */
+    public Future<?> submitTransaction(Realm realm, Runnable task) {
```
beeender: ??? `realm` is unused param

cmelchior: Ups, was an experiment I forgot to remove. Will fix.

beeender: BTW, if they take the same param `submitTransaction`/`submitQuery` and do the exact same thing, no need to split into two functions ....

cmelchior: Yup. We will need it for https://github.com/realm/realm-java/issues/3040 though, and I kinda liked the readability of it. I can revert it until we need it if it is a big problem though?

beeender: Then it's fine, just keep it ~~

realm/realm-library/src/androidTest/java/io/realm/RealmAsyncQueryTests.java
```diff
@@ -281,6 +281,34 @@ public void execute(Realm realm) {
...
+
+        realm.executeTransactionAsync(new Realm.Transaction() {
+            @Override
+            public void execute(Realm realm) {
+                realm.createObject(AllTypes.class);
+            }
+        }, new Realm.Transaction.OnSuccess() {
+            @Override
+            public void onSuccess() {
+                assertEquals(realm.where(AllTypes.class).count(), 1);
```
beeender: Since we had the results sync problem before, worth to check if the results get updated here as well.

cmelchior: Added

realm/realm-library/src/main/java/io/realm/HandlerController.java
```diff
@@ -146,6 +151,23 @@ public boolean handleMessage(Message message) {
         return true;
     }
 
+    /**
+     * Properly handles when a async transaction completes. This will be treated as a REALM_CHANGED event when
+     * determining which queries to re-run and when to notify listeners.
+     *
+     * NOTE: This is needed as it is not possible to combine a `Message.what` value and a callback runnable. So instead
```
kneth: Please add `<p>`

realm/realm-library/src/main/java/io/realm/HandlerController.java
```diff
@@ -146,6 +151,23 @@ public boolean handleMessage(Message message) {
         return true;
     }
 
+    /**
+     * Properly handles when a async transaction completes. This will be treated as a REALM_CHANGED event when
```
kneth: "a" -> "an"

realm/realm-library/src/main/java/io/realm/HandlerController.java
```diff
@@ -265,7 +287,7 @@ private void updateAsyncEmptyRealmObject() {
     }
 
     /**
-     * This method calls all registered listeners on Realm, RealmResults and RealmObjects.
+     * This method calls all registered listeners on Realm, RealmResults and RealmObjects and async transactions.
```
kneth: Listeners for Realm, RealmResults and RealmObjects, and callbacks for async transactions.

realm/realm-library/src/main/java/io/realm/HandlerController.java
```diff
@@ -557,6 +582,19 @@ private void completedAsyncQueriesUpdate(QueryUpdateTask.Result result) {
         }
     }
 
+    /**
+     * Trigger onSuccess for all completed async transaction.
+     * NOTE: Should only be called from {@link #notifyAllListeners(List)}.
```
kneth: Please add `<p>`

realm/realm-library/src/main/java/io/realm/HandlerController.java
```diff
@@ -557,6 +582,19 @@ private void completedAsyncQueriesUpdate(QueryUpdateTask.Result result) {
...
+     */
+    private void notifyAsyncTransactionCallbacks() {
+        if (!pendingOnSuccessAsyncTransactionCallbacks.isEmpty()) {
+            for (Runnable callback : pendingOnSuccessAsyncTransactionCallbacks) {
+                realm.handler.post(callback);
```
beeender: I feel something will go wrong by posting the runnable instead of calling it directly. Consider by posting it, other threads will have chance to post a change event in front of it. What is the reason do posting here?

cmelchior: Ups, you very much right. I planned to call them directly here 👍

realm/realm-library/src/main/java/io/realm/HandlerController.java
```diff
@@ -146,6 +151,23 @@ public boolean handleMessage(Message message) {
...
+     * of posting two messages, we post a runnable that runs this method. This means it is possible to interpret
+     * `REALM_CHANGED + Runnable` as one atomic message.
+     *
+     * @param onSuccess onSuccess callback to run for the async transaction that completed.
+     */
+    public void handleAsyncTransactionCompleted(Runnable onSuccess) {
+        if (onSuccess != null) {
+            pendingOnSuccessAsyncTransactionCallbacks.add(onSuccess);
+        }
+        realmChanged(false);
```
beeender: What is the reason that not using some thread safe container for callbacks and still trigger changes by `REALM_CHANGED`? 

The good thing about `REALM_CHANGED` is we will have 2 events in the queue for the worst case. But posting a callback mean below code will be a nightmare:

```
loop( 500 times) {
    executeAsyncTransaction(...)
}
```

cmelchior: Mostly it was to keep the idea that an asyncTransactionCompleted event could function as an atomic event in the message queue which helped avoiding introducing yet another thread safe list, but I agree that because Handler does not support Runnables very well, the code gets a bit ugly. 

The above example you have is kinda likely though and might cause a lot of un-needed change events to happen, so perhaps going back to a CopyOnWriteArrayList is better.

Anyone know roughly where a ReadWriteLock is better than a CopyOnWriteList? Because in this case we would probably write more to it than we would iterate.

beeender: More thoughts about this, thread safe container is not enough. We need to lock before `advance_read` happen, or say, we should have a copy of the callback list before `advance_read` happen. Any callbacks posted later after this `advance_read` should be called after next `advance_read`.

cmelchior: Good point, and that is actually an argument against trying to add callbacks from outside the message queue. Trying to reason about where to lock will increase the complexity even further. Not to mention that the current HandlerController currently separate change notifications from advance_read. That would also have be even more tightly coupled. 

I am starting to think that the implementation in this PR is actually better as it separates the concerns in a nice way. Also, the example you gave is not entirely accurate. It will put a lot of tasks in the RealmThreadExecutor for sure, but events will still only trickle through as each async job is completed, so I don't think we need to worry about pushing to many events to the message queue.

What do you think @beeender ?

beeender: I kind of agree. except i am thinking about `pendingOnSuccessAsyncTransactionCallbacks`, right now, the size of it will never be greater than 1?

beeender: nope, I was wrong. There will be more than one element if async query exists.

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -1092,15 +1091,23 @@ public void run() {
...
                             handler.post(new Runnable() {
                                 @Override
                                 public void run() {
-                                    onSuccess.onSuccess();
+                                    handlerController.handleAsyncTransactionCompleted(onSuccess != null ? new Runnable() {
+                                        @Override
+                                        public void run() {
+                                            onSuccess.onSuccess();
+                                        }
+                                    } : null);
```
beeender: How about just go for the normal routine `REALM_CHANGED` if the `onSuccess` is `null`? I am fine with this as well.

cmelchior: Using the REALM_CHANGED if `onSuccess` is null will actually bloat this code quite a bit as we then need to send a empty message to the handler instead. By doing it this way we hide that complexity in `handleAsyncTransactionCompleted`

beeender:
:+1: Maybe it makes sense to do a `HandlerQueue.removeMessage(REALM_CHNAGED)` at the beginning of `realmChange()`. but it is beyond the scope of this PR.

cmelchior:
Hmm, there have now been two native crashes in different places for this PR, but not reproducible on my side. Reverting to In progress while investigating.

cmelchior:
Hmm, just a flaky device it seems. Not related to this PR. Last review missing

realm/realm-library/src/main/java/io/realm/HandlerController.java
```diff
@@ -99,6 +99,11 @@
     final ConcurrentHashMap<WeakReference<RealmObjectProxy>, Object> realmObjects =
             new ConcurrentHashMap<WeakReference<RealmObjectProxy>, Object>();
 
+    // List of onSuccess callbacks from async transactions. We need to track all callbacks as notifying listeners might
+    // be delayed due to the presence of async queries. This can mean that multiple async transactions can complete
+    // before we are ready to notify all of them.
+    private final List<Runnable> pendingOnSuccessAsyncTransactionCallbacks = new ArrayList<Runnable>();
```
nhachicha: what happen when the user cancel the transaction using the provided `RealmAsyncTask`?
Also is there a risk of memory leak?

cmelchior: Callbacks are added through the looper queue, so if a job is canceled they will never be sent in the first place. This queue is part of the HandlerController (which is GC'ed if the Realm is closed) and will be emptied on any other REALM_CHANGED event as well. So there shouldn't be any risk of leaking.

Issue #2607, opened 2016-04-14
RealmObject.asObservable() throws IllegalStateException when use Schedulers.io()


ramanbranavitski:
I tried to add RealmObject item to Realm and wrap this operation into Observable. This is my code

```
return Observable.defer(new Func0<Observable<T>>() {
            @Override
            public Observable<T> call() {
                Log.d("TEST", Thread.currentThread().getName());
                final Realm mRealm = Realm.getDefaultInstance();
                mRealm.beginTransaction();
                Observable<T> result = mRealm.copyToRealmOrUpdate(item)
                        .<T>asObservable()
                        .map(new Func1<RealmObject, T>() {
                            @Override
                            public T call(RealmObject realmObject) {
                                Log.d("TEST", Thread.currentThread().getName());
                                return (T) mRealm.copyFromRealm(realmObject);
                            }
                        })
                        .observeOn(AndroidSchedulers.mainThread());
                mRealm.commitTransaction();
                return result;
            }
        });
```

And use this as

```
rep.addOrUpdate(mUser)
                .subscribeOn(Schedulers.io())
                .subscribe(new Action1<User>() {
                    @Override
                    public void call(final User user) {
                        Log.d("TEST", Thread.currentThread().getName());
                    }
                });
```

Tried all available schedulers.

When I try to run it I receive IllegalStateException: You can't register a listener from a non-Looper thread .

cmelchior:
Hi @ramanbranavitski 
That is an expected error as your RealmResults is created on the `Schedulers.io()` thread and then accessed on the UI thread. It is also described here: https://github.com/realm/realm-java/blob/master/examples/rxJavaExample/src/main/java/io/realm/examples/rxjava/gotchas/GotchasActivity.java#L94-Lundefined

I was just going over documentation, and it could be better explained there, so we will fix that.

ramanbranavitski:
I call ```
 .map(new Func1<RealmObject, T>() {
                            @Override
                            public T call(RealmObject realmObject) {
                                Log.d("TEST", Thread.currentThread().getName());
                                return (T) mRealm.copyFromRealm(realmObject);
                            }
                        })
```
According to documentation `Realm.copyFromRealm() which makes it possible to copy Realm data into normal Java objects, effectively detaching them from Realm.` And this also perform on `Schedulers.io()`. 

This exception is throw by your `addChangeListener method ````
if (realm.handler == null) {
            throw new IllegalStateException("You can't register a listener from a non-Looper thread ");
[...]

PR #2787, opened 2016-05-12
Javadoc for using Realm with SubscribedOn() 

stk1m1:
It is recommended to use Realm.where().find*Async() instead of SubscribedOn() to avoid Realm being accessed from different threads.

This PR addresses issue #2607.
@realm/java

### DIFF
--- a/realm/realm-library/src/main/java/io/realm/BaseRealm.java
+++ b/realm/realm-library/src/main/java/io/realm/BaseRealm.java
@@ -163,7 +163,7 @@ abstract class BaseRealm implements Closeable {
      * </pre>
      *
      * <p>Note that when the Realm is accessed on threads other than where it was created, {@link IllegalStateException}
-     * will be thrown. Please avoid using {@code subscribeOn()}, and use {@code Realm.where().findAllAsync*()} methods instead.
+     * will be thrown. Please avoid using {@code subscribeOn()}, and use {@code Realm.where().*Async()} methods instead.
      *
      * @return RxJava Observable that only calls {@code onNext}. It will never call {@code onComplete} or {@code OnError}.
      * @throws UnsupportedOperationException if the required RxJava framework is not on the classpath.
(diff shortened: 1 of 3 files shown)


kneth:
:+1:

realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -162,6 +162,9 @@ public void removeChangeListener(RealmChangeListener<? extends BaseRealm> listen
      * }
      * </pre>
      *
+     * <p>Note that when the Realm is accessed from threads other than where it was created, {@link IllegalStateException}
```
cmelchior: This issue is about RealmObject not Realm, so this should ideally be moved to RealmResults/RealmObject instead.

realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -162,6 +162,9 @@ public void removeChangeListener(RealmChangeListener<? extends BaseRealm> listen
...
+     * will be thrown. Please avoid using {@code subscribeOn()}, and use {@code Realm.where().*Async()} methods instead.
```
cmelchior: Actually it is the combination of observeOn/subscribeOn that is problematic, since that will cause the thread changes, so I would probably rephrase this to:

"Care should  be taken when using different schedulers with {@code subscribeOn()} and {@code observeOn}. Consider using {@code Realm.where().find*Async()} instead."

stk1m1:
The review points are addressed.

realm/realm-library/src/main/java/io/realm/RealmObject.java
```diff
@@ -327,6 +327,10 @@ public final void removeChangeListeners() {
      * }
      * </pre>
      *
+     * <p>Note that when the {@link Realm} is accessed from threads other than where it was created,
+     * {@link IllegalStateException} will be thrown. Please avoid using {@code subscribeOn()}, and use
```
cmelchior: Imo this still doesn't capture it, pr. my previous review. It is the combination of subscribeOn/observeOn that is problematic.

Also, DynamicRealmObject also have a asObservable method :)

stk1m1:
I don't think `DynamicRealmObject`has overridden `asObservable()` method. Review points are addressed.

cmelchior:
👍

PR #2799, opened 2016-05-13
One RealmModule per file

### LINKED ISSUE #2566: Only one RealmModule per file
http://stackoverflow.com/questions/36317366/realm-release-build-with-proguard-fails-could-not-find-io-realm-realmhelperma indicates that it is not possible to have multiple `RealmModule`s in one file.

Of course, it can be a bug but at least we should document it.

kneth:
Closes #2566 

The general documentation has been updated.

@realm/java


realm-annotations/src/main/java/io/realm/annotations/RealmModule.java
```diff
@@ -52,7 +56,7 @@
  * App developers are not required to specify any modules, as they implicitly use the {@code DefaultRealmModule}, but
  * they now has the option of adding the library project classes to their schema using
  * {@code RealmConfiguration.addModule()}.
- *
+ *
```
cmelchior: Nitpick: Looks like unwanted whitespace?

cmelchior:
👍

PR #2801, opened 2016-05-13
Upload native lib with symbols to s3

beeender:
And remove some useless lines

Unfortunately crashlytics's symbols uploading is not what we want. It would be useful if user is using ndk and 3rd party libs exist in the `lib` dir. I couldn't make it work.

Having debug so on s3 would be good enough for us right now.

TODO:
- [ ] Update wiki about decode symbols 

**PS. Haven't successfully tested the uploading part since it always fails in my network :(**

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -172,10 +172,6 @@ task distributionPackage(type:Zip) {
     from('changelog.txt')
     from('LICENSE')
     from('version.txt')
-    from("${buildDir}/outputs/gradle") {
-        include "realm-android-${currentVersion}.jar"
-        into 'gradle'
-    }
     from('realm/realm-library/build/libs') {
         include 'realm-android-${currentVersion}-javadoc.jar'
         into 'docs'
@@ -189,6 +185,19 @@ task distributionPackage(type:Zip) {
         exclude '**/.gradle'
         exclude '**/build'
         into 'examples'
+    }
+}
+
+task distributionJniUnstrippedPackage(type:Zip) {
+    description = 'Generate native libs package with debug symbols'
+    dependsOn assembleRealm
+
+    group = 'Artifact'
+    archiveName = "realm-java-jni-libs-unstripped-${currentVersion}.zip"
+    destinationDir = file("${buildDir}/outputs/distribution")
+
+    from("realm/realm-jni/build/outputs/jniLibs-unstripped") {
+        include '**/*.so'
     }
 }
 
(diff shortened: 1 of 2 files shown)


realm/realm-jni/build.gradle
```diff
@@ -327,7 +327,7 @@ targets.each { target ->
         // Store the unstripped version
         copy {
             from "${projectDir}/src/librealm-jni-${target.name}${getDebugExt()}.so"
-            into "${projectDir}/../build/output/jniLibs-unstripped/${target.abi}"
+            into "${buildDir}/outputs/jniLibs-unstripped/${target.abi}"
```
beeender: Change it to `realm/realm-jni/build/outputs` since it makes more sense and not coupled with other projects.

cmelchior:
👍

build.gradle
```diff
@@ -192,6 +188,19 @@ task distributionPackage(type:Zip) {
...
 
+task distributionJniUnstrippedPackage(type:Zip) {
+    description = 'Generate native libs package with debug symbols'
+    dependsOn assembleRealm
+
+    group = 'Artifact'
+    archiveName = "realm-java-jni-libs-unstripped-${currentVersion}.zip"
+    destinationDir = file("${buildDir}/outputs/distribution")
+
+    from("realm/realm-jni/build/outputs/jniLibs-unstripped") {
```
nhachicha: can't you use `${buildDir}` as above?

realm/realm-jni/build.gradle
```diff
@@ -347,7 +347,6 @@ task clean(type: Delete) {
     delete project.buildDir
 
     delete fileTree(dir: "${projectDir}/../realm-library/src/main/jniLibs/", include: '**/librealm-jni*.so')
-    delete fileTree(dir: "${projectDir}/../build/output/jniLibs-unstripped/", include: '**/librealm-jni*.so')
```
nhachicha: how about _outputs/jniLib**s**-unstripped_

build.gradle
beeender: no, not the same project. `${buildDir}` points to `root/build`

realm/realm-jni/build.gradle
beeender: ?? This line is deleted ....

nhachicha:
:+1:

thorbenprimke:
Thanks @beeender! Crashlytics test logs look good.

awesomesauce

cmelchior:
@thorbenprimke Does that mean you actual got the symbols working with Crashlytics. I got the impression from @beeender that he couldn't make it work. So if you have a working setup we would love to document it on the website?

thorbenprimke:
Yes, I did. However Crashlytics doesn't support Android 6 at the moment. [They are working on it.](https://twittercommunity.com/t/android-native-crash-parsing-failure-android-6-m/67638/2)

@cmelchior Is this what you expect the stacktrace to look like? It shows the method that I'm calling that triggers the crash. 

screen shot 2016-05-26 at 10 37 22 am

cmelchior:
Hmm, both yes and no. Ideally you would also have the "(missing)" in librealm-jni.so as well, but @beeender know the details.

beeender:
The symbol for the bt 3 is not stripped (Java_io_realm_internal_xxx) should always be visible. It is located in our JNI code.
the bt 2 is actually the problem we want to solve.
With the unstripped version `so`s, it is possible to decode it with `addr2line` or `ndk-stack`. But it is sort of meaningless to you since that code is in our underlying storage engine which is not opensourced yet.

@thorbenprimke Can you share the crashlytics link with us? I haven't seen this kind of crash before. And what is the Realm version when this crash happened?

cmelchior:
It's an intentional crash I gave him by using our internal API inproperly, it should not happen when using the normal API's ☺️

PR #2803, opened 2016-05-13
Realm Core updated to 0.100.4.

### LINKED ISSUE #2749: Openssl vulnerabilities
According to announcement from Google

> Beginning July 11, 2016, Google Play will block publishing of any new apps or updates that use older versions of OpenSSL. If you’re using a 3rd party library that bundles OpenSSL, you’ll need to upgrade it to a version that bundles OpenSSL 1.02f/1.01r or higher.

>The vulnerabilities were addressed in OpenSSL 1.02f/1.01r. The latest versions of OpenSSL can be downloaded here. To confirm your OpenSSL version, you can do a grep search for ($ unzip -p YourApp.apk | strings | grep "OpenSSL").

I have created a small application to test which OpenSSL version is currently using in Realm and the result is OpenSSL 1.0.1p as below
screen shot 2016-05-09 at 8 56 51 am

Is this safe ? If not,have any plans to update new version to fix this issue ?

kneth:
Closes #2749

@realm/java

### DIFF
--- a/realm/realm-library/src/androidTest/java/io/realm/RealmResultsTests.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/RealmResultsTests.java
@@ -1005,9 +1005,7 @@ public class RealmResultsTests extends CollectionTests {
     }
 
     @Test
-    @RunTestInLooperThread
     public void deleteAndDeleteAll() {
-        final Realm realm = looperThread.realm;
         realm.beginTransaction();
         for (int i = 0; i < 10; i++) {
             StringOnly stringOnly = realm.createObject(StringOnly.class);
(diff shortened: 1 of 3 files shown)


realm/realm-library/src/androidTest/java/io/realm/RealmResultsTests.java
```diff
@@ -1002,4 +1003,39 @@ public void run() {
...
+
+        realm.beginTransaction();
+        // remove the rest
+        stringOnlies.deleteAllFromRealm();
+        realm.commitTransaction();
+
+        assertEquals(0, realm.where(StringOnly.class).findAll().size());
+
+        // The above commit should have put a REALM_CHANGED event on the Looper queue before this runnable.
+        looperThread.postRunnable(new Runnable() {
```
cmelchior: No need for this or the `@RunTestInLooperTread`. Just ending the test with the assert should be fine.

cmelchior:
Apart from comment 👍

dalinaum:
:+1:

zaki50:
Shouldn't this go to `releases` branch?

PR #2804, opened 2016-05-13
Merge remote-tracking branch 'origin/releases' into mc/merge/0.90_82e0e0d1

beeender:


dalinaum:
:+1: Because there is no change, the review is very easy.

cmelchior:
Perhaps we should just close it then :)

PR #2820, opened 2016-05-16
Changed to throw IllegalStateException when setAutoRefresh() is called on non-Looper thread

zaki50:
@realm/java

### DIFF
--- a/realm/realm-library/src/main/java/io/realm/BaseRealm.java
+++ b/realm/realm-library/src/main/java/io/realm/BaseRealm.java
@@ -77,23 +77,29 @@ abstract class BaseRealm implements Closeable {
         this.sharedGroupManager = new SharedGroupManager(configuration);
         this.schema = new RealmSchema(this, sharedGroupManager.getTransaction());
         this.handlerController = new HandlerController(this);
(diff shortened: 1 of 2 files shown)


realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -89,7 +89,7 @@ protected BaseRealm(RealmConfiguration configuration, boolean autoRefresh) {
      * thread.
      *
      * @param autoRefresh {@code true} will turn auto-refresh on, {@code false} will turn it off.
-     * @throws IllegalStateException if called from a non-Looper thread.
+     * @throws IllegalStateException if called from a non-Looper thread with {@code autoRefresh} {@code true}.
```
cmelchior: Wouldn't it make more sense to just throw an IllegalStateException always? This behavior seems a bit weird?

zaki50: @cmelchior done. review again please

zaki50:
updated the title

cmelchior:
👍

beeender:
:+1:

PR #2821, opened 2016-05-17
Improved detection when using interfaces in Realm classes

### LINKED ISSUE #2808: Generated RealmProxy class try to access wrong RealmProxy class for internal fields when using an interface hierarchy
#### Goal

I'm trying to take advantage of the new interface `RealmModel` and add some abstraction to my models. My goal is to define an interface for my Realm classes to hide details of the implementation.

Given the following structure (in Kotlin, but should be the same in Java):

```kotlin
interface Address : RealmModel
open class RealmAddress() : Address {
    [...]
}

interface User : RealmModel {
    val addresses: MutableList<Address>
}
open class RealmUser() : User {
    override var addresses: RealmList<Address> = RealmList()
}
```

The generated `RealmUserRealmProxy` try to access `RealmAddressRealmProxy` and its table name by relying on a convention on the type of the field `addresses`, but in this case it fails since it tries to look up `AddressRealmProxy` and `class_Address` instead of `RealmAddressRealmProxy` and `class_RealmAddress` since the type of `addresses` is `RealmList<Address>` instead of `RealmList<RealmAddress>`.

I think the code I'm writing should be correct, if not then probably I misunderstand the purpose of using the `RealmModel` interface instead of extending `RealmObject`...

Realm version(s): 0.90.1

Android Studio version: N/A

cmelchior:
Fixes #2808 

This PR adds additional checking to the annotation processor to prevent the error reported int #2808.
#2808 would not compile anyway, but this way we can report the real error to the user.

@realm/java


beeender:
:+1:

realm/realm-annotations-processor/src/main/java/io/realm/processor/ClassMetaData.java
```diff
@@ -156,15 +161,46 @@ private boolean checkForFinalFields() {
...
         }
+
         return true;
     }
 
+    private boolean checkReferenceTypes() {
+        for (VariableElement field : fields) {
+            if (Utils.isRealmModel(field)) {
+                // Check that the referenced type is a concrete class and not an interface
+                TypeElement typeElement = elements.getTypeElement(field.asType().toString());
```
kneth: Common code could be put in a small aux. method (making sure that error message is always the same).

kneth:
:+1: with a minor comment.

realm/realm-annotations-processor/src/main/java/io/realm/processor/ClassMetaData.java
cmelchior: Not sure I understand this? This particular statement is not used elsewhere and the error message is different: One talks about Lists, the other about Objects.

kneth: Sorry, my mistake.

kneth:
Correcting my mistake: :+1:

bmunkholm:
@emanuelez Can you put an extra pair of eyes on this?

realm/realm-annotations-processor/src/main/java/io/realm/processor/ClassMetaData.java
```diff
@@ -156,15 +161,46 @@ private boolean checkForFinalFields() {
...
                     return false;
                 }
+
+                // Check that the referenced type is a concrete class and not an interface
+                TypeMirror fieldType = field.asType();
+                List<? extends TypeMirror> typeArguments = ((DeclaredType) fieldType).getTypeArguments();
+                String genericCanonicalType = typeArguments.get(0).toString();
+                TypeElement typeElement = elements.getTypeElement(genericCanonicalType);
+                if (typeElement.getSuperclass().getKind() == TypeKind.NONE) {
+                    Utils.error("Only concrete Realm classes are allowed in Lists. Neither " +
```
nhachicha: minor: 
_are allowed in RealmCollection_?

realm/realm-annotations-processor/src/test/resources/some/test/InterfaceObjectReference.java
```diff
@@ -0,0 +1,27 @@
...
+
+import io.realm.RealmList;
+import io.realm.RealmModel;
+import io.realm.annotations.RealmClass;
+
+@RealmClass
+public class InterfaceObjectReference implements CustomInterface {
+    String name;
+    CustomInterface obj; // Polymorphism, not yet supported.
+}
```
nhachicha: new line for checkstyle

nhachicha:
👍 minor comments

realm/realm-annotations-processor/src/test/resources/some/test/InterfaceObjectReference.java
cmelchior: Done

CHANGELOG.md
```diff
@@ -21,6 +22,7 @@
 ### Credits
 
 * Paweł Surówka (@thesurix) for adding the `RealmConfiguration.Builder.assetFile()`.
+>>>>>>> master
```
kneth: Remove

emanuelez:
👍

PR #2832, opened 2016-05-18
addPrimaryKey() on RealmObjectSchema misses Index addition.

### LINKED ISSUE #2819: "Index not defined for existing Realm file" error while performing migration.
Hello,

I'm using Realm 0.90.1.

I'm trying to perform a migration in which I want to create two more tables. One of them has a @PrimaryKey.

I've tried the manual way like this:

And the migration looks like this:

```

		if (oldVersion == 0) {
			schema.create("Table1").addField("myid", long.class)
					.addField("test", boolean.class)
					.addPrimaryKey("myid");

			schema.create("Table2").addField("wID", int.class)
				.addField("uID", int.class)
				.addField("ts", long.class)
				.addField("type", String.class);

			oldVersion++;
		}
```

My applications crashes with the following exception:

`                                                                         
Process: com.vectorwatch.android, PID: 22182 java.lang.RuntimeException: Unable to create application com.android.tools.fd.runtime.BootstrapApplication: io.realm.exceptions.RealmMigrationNeededException: RealmMigration must be provided
                                                                             at android.app.ActivityThread.handleBindApplication(ActivityThread.java:4664)`

I've dived deeper in Realm code and I've saw that an exception is thrown with the following detailsMessage:

`Index not defined for field myid in existing Realm file. Either set @Index or migrate using io.realm.internal.Table.removeSearchIndex().
`

So I assumed that I should call also addIndex() in the migration. This did the trick, but this is was kind of annoying. From the Realm documentation, I see that "(the annotation @PrimaryKey implicitly sets the annotation @Index)". The javadoc of the addPrimaryKey method says:

`
     * Adds a primary key to a given field. This is the same as adding the {@link io.realm.annotations.PrimaryKey} annotation on the field.
`

So I assumed that addPrimaryKey() also adds an Index. It is nowhere mentioned that when performing a migration, you should call addIndex() when you have an addPrimaryKey().

Can I make a pull request to add this use case in your Migration example app? Also please tell if I can help anymore you with this issue.

Thank you very much for this great framework!

stk1m1:
`RealmObjectSchema.addPrimaryKey()` is supposed to add a search index to its primary key field but it is currently missing.  This is to fix the addition and removal of a search index related to primary key methods in `RealmObjectSchema`. Closes #2819.

@realm/java please review.

### DIFF
--- a/realm/realm-library/src/androidTest/java/io/realm/RealmMigrationTests.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/RealmMigrationTests.java
@@ -226,6 +226,41 @@ public class RealmMigrationTests {
         }
     }
 
+    @Test
+    public void settingPrimaryKeyWithObjectSchema() {
+        // Create v0 of the Realm
+        RealmConfiguration originalConfig = configFactory.createConfigurationBuilder()
+                .schema(AllTypes.class)
+                .build();
+        Realm.getInstance(originalConfig).close();
+
+        RealmMigration migration = new RealmMigration() {
+            @Override
+            public void migrate(DynamicRealm realm, long oldVersion, long newVersion) {
+                RealmSchema schema = realm.getSchema();
+                schema.create("AnnotationTypes")
+                        .addField("id", long.class)
+                        .addPrimaryKey("id")    // use addPrimaryKey() instead of adding FieldAttribute.PrimaryKey
+                        .addField("indexString", String.class)
+                        .addIndex("indexString") // use addIndex() instead of FieldAttribute.Index
+                        .addField("notIndexString", String.class);
+            }
+        };
+
+        // Create v1 of the Realm
+        RealmConfiguration realmConfig = configFactory.createConfigurationBuilder()
+                .schemaVersion(1)
+                .schema(AllTypes.class, AnnotationTypes.class)
+                .migration(migration)
+                .build();
+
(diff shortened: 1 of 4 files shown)


kneth:
:+1: when `CHANGELOG.md` is updated :-)

stk1m1:
@kneth I will add it in no time.

kneth:
@stk1m1 Me like

realm/realm-library/src/androidTest/java/io/realm/RealmMigrationTests.java
```diff
@@ -226,6 +226,41 @@ public void migrate(DynamicRealm realm, long oldVersion, long newVersion) {
...
+
+        // Create v1 of the Realm
+        RealmConfiguration realmConfig = configFactory.createConfigurationBuilder()
+                .schemaVersion(1)
+                .schema(AllTypes.class, AnnotationTypes.class)
+                .migration(migration)
+                .build();
+
+        realm = Realm.getInstance(realmConfig);
+        RealmObjectSchema schema = realm.getSchema().getSchemaForClass(AnnotationTypes.class);
```
cmelchior: `getSchemaForClass` is an internal API, the public method is just called `get`. We should use the public API if possible IMO

cmelchior:
Apart from small comment 👍

nhachicha:
+1

PR #2839, opened 2016-05-19
Remove log files in Realm.deleteRealm

### LINKED ISSUE #2834: Cannot delete realm
When starting up my app the first thing I am trying to delete the realm file.

The attached screenshot shows that if fails to delete the management directory.
/data/user/0/com.agrimap.mobile.android/files/default.realm.management

I checked the File.delete() method and it internally catches
android.system.ErrnoException: remove failed: ENOTEMPTY (Directory not empty)

The directory appears to have three files in it.
log_acces, log_a, log_b

screen shot 2016-05-19 at 12 09 12 pm

Realm version(s): 0.90.1
Android Studio version: 2.1.1
Which Android version and device: Emulator - Nexus 4

beeender:
Close #2834

### DIFF
--- a/realm/realm-library/src/androidTest/java/io/realm/RealmTests.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/RealmTests.java
@@ -1911,7 +1911,7 @@ public class RealmTests {
         realm.beginTransaction();
         realm.createObject(AllTypes.class);
         realm.commitTransaction();
-        // Move the directory so core won't be able to delete log files
+        // A core upgrade might change the location of the files
         assertTrue(tempDir.renameTo(tempDirRenamed));
         readyToCloseLatch.countDown();
 
--- a/realm/realm-library/src/androidTest/java/io/realm/RealmTests.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/RealmTests.java
@@ -1884,20 +1884,9 @@ public class RealmTests {
 
     @Test
     public void deleteRealm() throws InterruptedException {
-        File tempDir = new File(context.getFilesDir(), "delete_test_dir");
-        if (!tempDir.exists()) {
-            assertTrue(tempDir.mkdir());
-        }
-
-        assertTrue(tempDir.isDirectory());
-
-        // Delete all files in the directory
-        File[] files = tempDir.listFiles();
-        if (files != null) {
-            for (File file : files) {
-                assertTrue(file.delete());
-            }
-        }
+        File tempDir = new File(configFactory.getRoot(), "delete_test_dir");
+        File tempDirRenamed = new File(configFactory.getRoot(), "delete_test_dir_2");
+        assertTrue(tempDir.mkdir());
 
         final RealmConfiguration configuration = new RealmConfiguration.Builder(tempDir).build();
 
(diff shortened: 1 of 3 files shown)


bmunkholm:
@finnschiermer Please review.

kneth:
:+1:

@finnschiermer Please see if we have overlooked some aux. files.

kneth:
After a brief chat with @finnschiermer, it is safer to remove all files in the `.management` folder. See http://stackoverflow.com/questions/7768071/how-to-delete-directory-content-in-java how to do so.

stk1m1:
👍

realm/realm-library/src/androidTest/java/io/realm/RealmTests.java
```diff
@@ -1922,14 +1911,15 @@ public void run() {
         realm.beginTransaction();
         realm.createObject(AllTypes.class);
         realm.commitTransaction();
+        // Move the directory so core won't be able to delete log files
```
cmelchior: Instead describe why it could happen, i.e during an upgrade of Realm where core changed the location of log files.

cmelchior:
Apart from minor comment 👍

realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -593,9 +593,14 @@ static private boolean deletes(String canonicalPath, File rootFolder, String rea
         List<File> filesToDelete = Arrays.asList(
                 new File(rootFolder, realmFileName),
                 new File(rootFolder, realmFileName + ".lock"),
+                // Old core log file naming styles
                 new File(rootFolder, realmFileName + ".log_a"),
                 new File(rootFolder, realmFileName + ".log_b"),
                 new File(rootFolder, realmFileName + ".log"),
+                // New styles
```
beeender: Those four lines can be removed.

emanuelez:
👍

realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -633,9 +629,19 @@ public void onResult(int count) {
...
                 String realmFileName = configuration.getRealmFileName();
                 File managementFolder = new File(realmFolder, realmFileName + management);
-                realmDeleted.set(deletes(realmFolder.getPath()+ "/" + realmFileName + management, managementFolder, realmFileName));
-                realmDeleted.set(realmDeleted.get() && deletes(canonicalPath, realmFolder, realmFileName));
 
+                // delete files in management folder and the folder
+                // there is no subfolders in the management folder
+                File[] files = managementFolder.listFiles();
+                if (files != null) {
+                    for (File file : files) {
```
cmelchior: Nitpick, but use a simple for loop instead to save the allocation of an iterator.

realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -633,9 +629,19 @@ public void onResult(int count) {
...
+                        realmDeleted.set(realmDeleted.get() && file.delete());
+                    }
+                }
+                realmDeleted.set(realmDeleted.get() && managementFolder.delete());
+
+                // delete specific files in root folder
+                realmDeleted.set(realmDeleted.get() && deletes(canonicalPath, realmFolder, realmFileName));
```
cmelchior: This will override any failures set by the previous set?

zaki50:
👍

nhachicha:
👍

realm/realm-library/src/main/java/io/realm/BaseRealm.java
kneth: `realmDelete()` will return true if and only if all `delete` calls return true.

zaki50: I think all delete operations are skipped once `realmDeleted` is set to `false`.

cmelchior: `set` will override any previous value. You need `compareAndSet` if not.

cmelchior: In this case if the management folder isn't deleted, that failure will not be detected? Or am I missing something?

cmelchior: Oh wait, i'm stupid, nevermind 👍

PR #2881, opened 2016-05-26
Backlinks

### LINKED ISSUE #607: Reverse lookup (Backlinks)
Hi guys,

I looked at the documentation and found support for linked queries. Is reverse lookup currently possible with the java version of Realm? I'm also using Realm on Cocoa which just added support for reverse lookup and wondered if there is already support for this feature in Java.

Thanks a lot!

cmelchior:
Fixes #607 

---
### Design goals

Backlink API proposal

We need to support Backlinks the following places in the public API:
- Annotations
- RealmQuery

```
    private Dog favoriteDog;    
    private RealmList<Dog> dogs;
}

    @Backlink("dogs")
    RealmResults<Person> owners;    

    @Backlink("favoriteDog")
    RealmResults<Person> favoriteOwner; 

}

```
### Design
- [ ] Is `@Backlink` the correct name for the annotation?
- [x] Should backlinks be followed when using `copyFromRealm`? -> No
- [x] How should copyToRealm react? Ignore it or throw an exception? -> Ignore
- [x] Should backlinks be null or empty list when using `copyFromRealm` ? -> Null ensures people don't accidentally assume it is a managed object. This ensures that people don't accidentially assume it is a managed object. It unfortunately also means that unmanaged and managed objects are no longer as "equal" as we might like. 
- [x] How should the API in RealmObjectSchema look like? Do we want to introduce a RealmField wrapper class at this point? or continue with helper methods. Natural extensions to this would be Single reference Backlinks and or user defined views (think subqueries defined by the schema) -> See #2904 .
- [ ] How live is the RealmResults and how should changelisteners be triggered ?
### 
- [x] Add support in the annotation processor. Only allowed on RealmResults. RealmResults not allowed without `@Backlink`.
- [ ] Add support for backlinks in `RealmResults`.
- [ ] Ignore backlinks in the various JSON and copyToRealm methods.
- [ ] Extend createAndValidate() to check if @Backlink does point to a proper class/field
- [ ] Extend query predicates in RealmQuery (equalTo() etc.) to be able to query backlinks
- [ ] The byte code transformation must change the field accessor (cat.owners) to getting a RealmResults using Table::get_backlink_view() making something like cat.owners.where().equalTo("name", "John").findAll() possible
- [ ] Add support in RealmQuery
- [ ] Consider how CheckedRow should handle BackLinks. 
- [ ] Update website documentation


cmelchior:
It appears that .NET have called their feature "Backlinks" https://realm.io/docs/xamarin/latest/#inverse-relationships.

IMO there is no good reason for having different names for the same concept across bindings, so we should probably align on this.

So far it looks like:

* The feature is named "Inverse Relationship" in the docs
* Swift uses a special class "LinkingObjects(fromType: Person.self, property: "dogs")" to capture the intent: https://realm.io/docs/swift/latest/#inverse-relationships
* .NET uses a Backlink annotation: `[Backlink(nameof(Person.Dogs))]` : https://realm.io/docs/xamarin/latest/#inverse-relationships

cmelchior:
Replaced by #4200

PR #2884, opened 2016-05-26
Avoid Table.nativeToString/nativeRowToString

### LINKED ISSUE #2429: Received an error in Jni: Failure when converting long string to UTF-16
```java
JNI DETECTED ERROR IN APPLICATION: input is not valid Modified UTF-8: illegal continuation byte 0x2e
03-12 00:19:49.679 8033-10913/com.tattva.app A/art: art/runtime/java_vm_ext.cc:410]     string: 'Unrecoverable error. Failure when converting long string to UTF-16 error_code = 8; retcode = 0; StringData.size = 116289; StringData.data = ...
```

I've emailed the entire stacktrace to your email `(help@realm.io)`, from my email `saket@saket.me`. Could not post it here publicly since it contains data I do not wish to make public.

#### Goal

Nothing. I was debugging my app and this happened out of nowhere.

No idea. Couldn't reproduce. 

Realm version(s): 0.87

Android Studio version: 2.1

Which Android version and device: Android 5.1, Moto G3

beeender:
Table::row_to_string Table::to_string are for debugging purpose only, it
calls string::substr which might truncate an UTF-8 string at any char.
Thus a crash could happend in JNI when converting to UTF-16.

This will only happen when user attach a debugger and the debugger is
trying to call Table.toString.

Close #2429

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### Bug fixes
 
-* Fixed a crash when calling Table.toString in debugger (#2429).
+* Fixed a crash when calling Table.toString() in debugger (#2429).
 
 ## 1.0.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,10 @@
-# 1.0.0
+## 1.0.1
+
+### Bug fixes
+
+* Fixed a crash when calling Table.toString in debugger (#2429).
+
+## 1.0.0
 
 No changes since 0.91.1.
 
--- a/realm/realm-library/src/androidTest/java/io/realm/internal/JNITableTest.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/internal/JNITableTest.java
@@ -49,7 +49,7 @@ public class JNITableTest extends AndroidTestCase {
         t.add("s1", 1, true);
         t.add("s2", 2, false);
 
-        String expected = "Table contains 3 columns: stringCol, intCol, boolCol. And 2 rows.";
+        String expected = "The Table contains 3 columns: stringCol, intCol, boolCol. And 2 rows.";
         assertEquals(expected, t.toString());
     }
 
--- a/realm/realm-library/src/androidTest/java/io/realm/internal/JNITableTest.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/internal/JNITableTest.java
@@ -49,7 +49,7 @@ public class JNITableTest extends AndroidTestCase {
         t.add("s1", 1, true);
         t.add("s2", 2, false);
 
-        String expected = "Table  contains 3 columns: stringCol, intCol, boolCol. And 2 rows.";
+        String expected = "Table contains 3 columns: stringCol, intCol, boolCol. And 2 rows.";
         assertEquals(expected, t.toString());
     }
 
--- a/realm/realm-library/src/androidTest/java/io/realm/internal/JNITableTest.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/internal/JNITableTest.java
@@ -49,11 +49,7 @@ public class JNITableTest extends AndroidTestCase {
         t.add("s1", 1, true);
         t.add("s2", 2, false);
 
-        String expected =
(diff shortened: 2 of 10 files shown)


beeender:
@realm/java

CHANGELOG.md
```diff
@@ -1,4 +1,10 @@
-# 1.0.0
+## 1.0.1
+
+### Bug fixes
+
+* Fixed a crash when calling Table.toString in debugger (#2429).
```
cmelchior: `Table.toString()` in the debugger

realm/realm-library/src/main/java/io/realm/internal/TableOrView.java
```diff
@@ -334,8 +334,12 @@
 
     String toString();
 
+    // For testing purpose only. It might crash in UTF16 conversion.
+    @Deprecated
     String toString(long maxRows);
```
cmelchior: Any reason we don't just remove them outright?

realm/realm-library/src/main/java/io/realm/internal/TableOrView.java
```diff
@@ -334,8 +334,12 @@
...
 
+    // For testing purpose only. It might crash in UTF16 conversion.
+    @Deprecated
     String rowToString(long rowIndex);
```
cmelchior: Same

realm/realm-library/src/main/java/io/realm/internal/TableOrView.java
beeender: 1. We might want use for debugging internally
2. Object store is coming
3. I am lazy ....

realm/realm-library/src/androidTest/java/io/realm/internal/JNITableTest.java
```diff
@@ -49,11 +49,7 @@ public void testTableToString() {
         t.add("s1", 1, true);
         t.add("s2", 2, false);
 
-        String expected =
-"    stringCol  intCol  boolCol\n" +
-"0:  s1              1     true\n" +
-"1:  s2              2    false\n" ;
-
+        String expected = "Table  contains 3 columns: stringCol, intCol, boolCol. And 2 rows.";
```
cmelchior: Nitpick, when a table doesn't have a name there is an extra space here :)

cmelchior:
Apart from minor comments 👍

realm/realm-library/src/main/java/io/realm/internal/TableOrView.java
cmelchior: Hmm, it would probably remove them, but I don't feel strongly about it.

realm/realm-library/src/main/java/io/realm/internal/TableView.java
```diff
@@ -724,7 +724,25 @@ public String toJson() {
 
     @Override
     public String toString() {
-        return nativeToString(nativePtr, 500);
+        long columnCount = getColumnCount();
+        StringBuilder stringBuilder = new StringBuilder("This TableView ");
```
kneth: "This" -> "The"

realm/realm-library/src/main/java/io/realm/internal/Table.java
```diff
@@ -1380,7 +1380,30 @@ public String toJson() {
 
     @Override
     public String toString() {
-        return nativeToString(nativePtr, INFINITE);
+        long columnCount = getColumnCount();
+        String name = getName();
+        StringBuilder stringBuilder = new StringBuilder("Table ");
```
kneth: "Table" -> "The Table"

kneth:
I think we should remove the native methods since we are not using them.

cmelchior:
👍

stk1m1:
I'd ask @kneth's 👍  on this. Other than that, 💯

beeender:
@kneth cannot be more agree with you, no worries!

bmunkholm:
@beeender Please refactor to reuse code instead of duplicating it. There is only one work of difference.

PR #2904, opened 2016-05-30
RFC: Inverse relationship query syntax

cmelchior:
This is an RFC for the backlink query syntax. Part of #2881 and #607 

Please review
@realm/java @bdash

### DIFF
--- a/realm/rfc/0001-backlink-query-syntax.md
+++ b/realm/rfc/0001-backlink-query-syntax.md
@@ -1,7 +1,7 @@
 - Feature Name: Inverse relationship syntax
 - Start Date: 2016-05-30
 - RFC PR: 
-- Version: 3
+- Version: 4
 
 # Summary
 
@@ -21,22 +21,30 @@ This proposal introduces a new syntax for queries so it is possible to query
 across a backlink even though it hasn't been defined by a model class. This is
 called *unnamed backlinks*.
 
+Realm Objective C/Swift shipped with only support for named backlinks. This was mostly
+prompted by restrictions in NSPredicate, which do not have a syntax for unnamed backlinks.
+
+Realm Java does not have this constraint and at the same time the Java Migration/Dynamic API
+is more full-fledged and expected to have the same feature set as the static API due to how the
+query api works.
+
+Adding support for the unnamed variant now will also most likely save us time documenting the
+difference as well as reduce the amount of support we have to do.
+
 Any design should ideally solve all these constraints:
 
-- Should support a list of backlinks.
+- Should support a list of backlink objects.
 
-- Should support a single backlink, enforcing the constraint that only 1 object 
+- Should support a single backlink object, enforcing the constraint that only 1 object
   can point to it. This is not supported by Core, but it is a natural extension.
-
-- Should support backlinks from multiple fields.
 
 - Should be usable from the Dynamic API.
 
 - Should either support polymorphic objects or not prevent them from being 
   implemented.
 
-- Ideally it should work for both a pure text-based variant as well as the 
-  current mixed style found in Realm Java.
+- Ideally it should work for both the pure text-based variant of the query language in the
+  Object Store as well as the current mixed style found in Realm Java.
   
 
 # Detailed design
@@ -52,22 +60,17 @@ public class Person extends RealmObject {
 public class Dog extends RealmObject {
 	private String name
 
-	// Named backlinks below
-	@Backlink({"favoriteDog"})
+	// Named backlinks
+	@LinkingObjects("favoriteDog")
 	private RealmResults<Person> favoriteOwner;
-
-	@Backlink({"favoriteDog", "dogs"})
-	private RealmResults<Person> allOwners;
 }
 ```
 
 **Introduce the following query extension**
 
-- Use `"y[x.z]"` to search field Y on type X where X is a backlink through 
-  field Z.
-- Searching multiple fields should be done in individual query methods:
-  `equalTo("y[x.z]", "foo").equalTo("y[x.z1]", "foo")`
-- This is a supplement to defining the named backlinks.
+- Use `linkingObjects(x.z).y` to search field Y on type X where X is a backlink object through
+  field Z. This is a supplement to defining the named backlinks, which will continue to work as normal
+  link queries: `favoriteOwner.name`
 - No changes to the migration API.
 - No changes to the schema definition.
 
@@ -75,18 +78,14 @@ Example:
 
 ```
 // Named backlink queries
-realm.where(Dog.class).equalTo("allOwners.name", "John").findAll();
 realm.where(Dog.class).equalTo("favoriteOwner.name", "John").findAll();
 
 // Unnamed backlink queries
-realm.where(Dog.class).equalTo("name[Person.favoriteDog]", "John").findAll();
-realm.where(Dog.class)
-  .equalTo("name[Person.favoriteDog]", "John")
-  .equalTo("name[Person.dogs]", "John")
-  .findAll();
+realm.where(Dog.class).equalTo("linkingObjects(Person.dogs).name", "John").findAll();
+dynamicRealm.where("Dog").equalTo("linkingObjects(Person.dogs).name", "John").findAll();
 
 // Example in a purely text based variant
-"name[Person.favoriteDog] = 'John'"
+"ANY linkingObjects(Person.dogs).name = 'John'"
 ```
 
 **Implementation**
@@ -96,29 +95,44 @@ split a field name into a name and a type part.
 
 Validating that the type is an actual backlink with the given field does require 
 an `O(n)` lookup where `m: number of fields in Person`. Since number of fields 
-are generally low, this shouldn't be a problem in practise.
+are generally low, this should not be a problem in practise.
 
 **Advantages**
-- Using "[]" is not an allowed field characters in any of our supported 
+- Using "()" is not allowed in field names in any of our supported
   languages as far as I know. So parsing the above expressions should be 
   possible with no ambiguity.
 
-- The default case is easy to describe.
+- Using a method invocation syntax make it descriptive what is happening instead of relying
+  on people understanding some unfamiliar syntax.
 
-- Using [] as a modifier already has precedent in NSPredicate.
+- Using a method invocation syntax make it easy to add more functionality to the query
+  language.
 
+- Will feel like a natural way of doing it if coming from NSPredicate.
 
 **Disadvantages**
 
-- Unknown syntax. This concept is not used nor supported in neither NSPredicate, 
-  LINQ or SQL. It will be a "power feature", and might make the language feel 
-  complicated and hard to use.
+- Unknown syntax, with no help from the type system. The concept of unnamed backlinks is not used
+  nor supported in neither NSPredicate, LINQ or SQL. It will be a "power feature", and might make
+  the language feel complicated and hard to use.
 
 - Moves the Java query language further away from being type safe.
 
+- Nesting text-based method calls inside other Java method calls feels and reads a bit strange:
+  `equalTo("linkingObjects(foo.bar).baz", "42")`.
+
 - Will introduce two ways of querying backlinks.
 
-- Without wildcards for unnamed backlinks queries risk getting extremely verbose.
+
+## Alternative annotation name suggestions
+
+The name of the annotation should reflect how we otherwise document and talk about the feature.
+
+Possible names for the annotation so far:
+
+- `@Backlink` (Internal name)
+- `@InverseRelationship` (Name used on Cocoa/CoreData docs)
+- `@LinkingObjects` (API name used by Cocoa)
 
 
 ## Alternative syntax suggestions
@@ -127,26 +141,21 @@ As an alternative to the above, I have listed a some of variants that were
 considered:
 
 ```
-1) Type in front
+1) Use [] as seperator
 
-- "[Person]name = 'John'"
-- "[Person.favoriteDog]name = 'John'"
+- "name[Person.dogs] = 'John'"
 
+2) Type in front
 
-2) All fields explicit. Group fields using ()
-
-- "name[Person.(favoriteDog, dogs)]  = 'John'"
-
+- "[Person.dogs]name = 'John'"
 
 3) Other modifiers (^ up) , (<- back)
 
-- "^Person(favoriteDog, dogs).name"
-- "Person(favoriteDog, dogs)<-.name"
-
+- "^[Person.dogs].name"
+- "[Person.dogs]<-.name"
 
 4) Special grouping operator
 
-```
 realm.where(Dog.class)
    .backlinks("Dogs", "dogs")
       .equalTo("state", "CA")
@@ -155,8 +164,12 @@ realm.where(Dog.class)
    .findAll()
 ```
 
+The primary argument against 1-3 is that it will introduce a completely new syntax that is unknown
+to users, also the use of special chars will make it harder to google for help.
 
-```
+The primary argument against 4 is that we should try to avoid adding query methods only needed by
+the Dynamic API.
+
 
 ## Possible extensions
 
@@ -164,14 +177,12 @@ The proposed query syntax can also be extended to support the following features
 
 ```
 // Wild cards
-- "name[*] = 'John' " // All backlink types
-- "name[Person.*] = 'John' " // All fields are included from the backlink type.
-- "name[* extends Person] = 'John' " // Polymorphic backlinks
-- *[*] = 'John' // Any string based field on all types of backlinks
+- "linkingObjects(Person.*).name = 'John' // search all backlink fields on Person
+- "linkingObjects(*.*).name  = 'John' // Include all backlink types
+- "linkingObjects(Person.dogs, Person.favoriteDog).name = 'John' // Multiple fields at the same time.
+```
 
 However this is not part of this RFC.
-
-```
 
 
 # Drawbacks
@@ -229,4 +240,7 @@ None.
 2: Backlink fields must now be enumerated. Removed "[Person]" as a shortcut
    for enumerating all fields on Person. Added suggestion that introduces
    a special `backlinks()/backlinksEnd()` grouping method.
-3: Fixed Typos and mistakes.
+3: Fixed typos and mistakes.
+4: Annotation renamed to `@LinkingObjects`. Syntax is now`linkingObjects(x.y).z`. Added more details
+   to the motivation section. Only one field is now allowed in the annotation, multiple backlink
+   fields must be added or queried separately.
--- a/realm/rfc/0001-backlink-query-syntax.md
+++ b/realm/rfc/0001-backlink-query-syntax.md
@@ -53,13 +53,12 @@ Any design should ideally solve all these constraints:
 
 ```
 public class Person extends RealmObject {
+	private String name
 	private Dog favoriteDog;
 	private RealmList<Dog> dogs;
 }
 
 public class Dog extends RealmObject {
-	private String name
-
 	// Named backlinks
 	@LinkingObjects("favoriteDog")
 	private RealmResults<Person> favoriteOwner;
@@ -78,11 +77,11 @@ Example:
 
 ```
 // Named backlink queries
-realm.where(Dog.class).equalTo("favoriteOwner.name", "John").findAll();
+RealmResults<Dog> dogs = realm.where(Dog.class).equalTo("favoriteOwner.name", "John").findAll();
 
 // Unnamed backlink queries
-realm.where(Dog.class).equalTo("linkingObjects(Person.dogs).name", "John").findAll();
-dynamicRealm.where("Dog").equalTo("linkingObjects(Person.dogs).name", "John").findAll();
+RealmResults<Dog> dogs = realm.where(Dog.class).equalTo("linkingObjects(Person.dogs).name", "John").findAll();
+RealmResults<DynamicRealmObjcet> dogs = dynamicRealm.where("Dog").equalTo("linkingObjects(Person.dogs).name", "John").findAll();
 
 // Example in a purely text based variant
 "ANY linkingObjects(Person.dogs).name = 'John'"
--- a/realm/rfc/0001-backlink-query-syntax.md
+++ b/realm/rfc/0001-backlink-query-syntax.md
@@ -1,0 +1,204 @@
+- Feature Name: Backlink query syntax
+- Start Date: 2016-05-30
+- RFC PR: 
+
+# Summary
+
+It should be possible to define and query backlinks in both `Realm` and 
+`DynamicRealm`. As a `DynamicRealm` can work without _any_ Realm model classes,
+it should be possible to query backlinks without them being defined there.
+
+# Motivation
+
+Backlinks are special in Realm. They are not defined explicitly in the schema,
+but are automatically created when defining Links and LinkLists (forward links).
+
+This creates some conflicts with the current query system as it rely on all 
+fields being defined by the schema. This is called *named backlinks*.
+
+This proposal introduces a new syntax for queries so it is possible to query
+across a backlink even though it hasn't been defined by a model class. This is
+called *unnamed backlinks*.
+
+Any design should ideally solve all these constraints:
+
+- Should support a list of backlinks.
+
+- Should support a single backlink, enforcing the constraint that only 1 object 
+  can point to it. This is not supported by Core, but it is a natural extension.
+
+- Should support backlinks from multiple fields.
+
+- Should be usable from the Dynamic API.
+
+- Should either support polymorphic objects or not prevent them from being 
+  implemented.
+
+- Ideally it should work for both a pure text-based variant as well as the 
+  current mixed style found in Realm Java.
+  
+
+# Detailed design
+
+**Example model classes:**
+
+```
+public class Person extends RealmObject {
+	private Dog favoriteDog;
+	private RealmList<Dog> dogs;
+}
+
+public class Dog extends RealmObject {
+	private String name
+
+	// Named backlinks below
+	@Backlink
+	private RealmResults<Person> owner;
+
+	@Backlink("favoriteDog")
+	private RealmResults<Person> favoriteOwner;
+}
+```
+
+**Introduce the following query extension**
+
+- Use `"y[x]"`to search field Y on type X where X is the backlink type.
+- Use `"y[x.z]"` to search field Y on type X where X is a backlink through 
+  field Z.
+- This is a supplement to defining the named backlinks.
+- No changes to the migration API.
+- No changes to the schema definition.
+
+Example:
+
+```
+// Named backlink queries
+realm.where(Dog.class).equalTo("owners.name", "John").findAll();
+realm.where(Dog.class).equalTo("favoriteOwner.name")
+
+// Unnamed backlink queries
+realm.where(Dog.class).equalTo("name[Person]", "John").findAll();
+realm.where(Dog.class).equalTo("name[Person.favoriteDog]").findAll();
+
+// Example in a purely text based variant
+"name[Person] = 'John'"
+```
+
+**Implementation**
+Implementation is straight forward as we have all schema information available 
+when parsing queries. So simple String manipulation will make it possible to 
+split a field name into a name and a type part.
+
+Validating that the type is an actual backlink with the given field does require 
+an `O(n)` lookup where `m: number of fields in Person`. Since number of fields 
+are generally low, this shouldn't be a problem in practise.
+
+**Advantages**
+- Using "[]" is not an allowed field characters in any of our supported 
+  languages as far as I know. So parsing the above expressions should be 
+  possible with no ambiguity.
+
+- The default case is easy to describe.
+
+- Using [] as a modifier already has precedent in NSPredicate.
+
+
+**Disadvantages**
+
+- Unknown syntax. This concept is not used nor supported in neither NSPredicate, 
+  LINQ or SQL. It will be a "power feature", and might make the language feel 
+  complicated and hard to use.
+
+- Moves the Java query language further away from being type safe.
+
+- Will introduce two ways of querying backlinks.
+
+
+
+## Alternative syntax suggestions
+
+As an alternative to the above, I have listed a some of variants that were 
+considered:
+
+```
+1) Type in front
+
+- "[Person]name = 'John'"
+- "[Person.favoriteDog]name = 'John'"
+
+
+2) All fields explicit. Group fields using ()
+
+- "name[Person.(favoriteDog, dogs)]  = 'John'"
+
+
+3) Other modifiers (^ up) , (<- back)
+
+- "^Person(favoriteDog, dogs).name"
+- "Person(favoriteDog, dogs)<-.name"
+
+```
+
+## Possible extensions
+
+The proposed query syntax can also be extended to support the following features:
+
+```
+// Wild cards
+- "name[*] = 'John' " // All backlink types
+- "name[* extends Person] = 'John' " // Polymorphic backlinks
+- *[*] = 'John' // Any string based field on all types of backlinks
+
+However this is not part of this RFC.
+
+```
+
+
+# Drawbacks
+
+It is not clear that his approach is the correct way forward. I am not sure what 
+can be done to gain more confidence in this, expect asking for more input. Please 
+see the list of advantages and disadvantages in the last section as well as the 
+proposed alternatives.
+
+
+# Alternatives
+
+A number of alternatives exists:
+
+## 1. Only named backlinks
+
+Objective C / Swift only allow named backlinks that are defined as part
+of the Realm model class.
+
+**Advantages:**
+- Minimal effort required to implement backlink queries. Current query system is 
+  enough.
+- Already supported by Objective C / Swift
+
+**Disadvantages:**
+- Does not work _at all_ in the dynamic API.
+- User ambiguity regarding manual migrations: Some fields must be added in a 
+  migration, others not needed. This is already happening with ignored fields
+  though.
+
+## 2. Backlinks as first class citizens
+
+An extended version of #1, but require backlinks to be defined as part of the 
+schema.
+
+**Advantages:**
+- Backlinks are treated just like all other fields for the purpose of migrations.
+- Minimal effort required to implement backlink queries. Current query system is 
+  enough.
+
+**Disadvantages:**
+- Forcing users to do work that is really not needed. Document that "computed" 
+  fields do not require migrations might be enough though.
+- Has to be implemented in the Object Store.
+- Require changes to the manual Migration API.
+
+
+# Unresolved questions
+
+None.
--- a/realm/rfc/0001-backlink-query-syntax.md
+++ b/realm/rfc/0001-backlink-query-syntax.md
@@ -186,7 +186,7 @@ However this is not part of this RFC.
 
 # Drawbacks
 
-It is not clear that his approach is the correct way forward. I am not sure what 
+It is not clear that this approach is the correct way forward. I am not sure what
 can be done to gain more confidence in this, expect asking for more input. Please 
 see the list of advantages and disadvantages in the last section as well as the 
 proposed alternatives.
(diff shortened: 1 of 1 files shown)


cmelchior:
Feedback:
In order to prevent people accidentally getting wrong results after a migration, all backlinks should be defined fully with fields.

Example:

V1: Initial version
```
class Person extends RealmObject {
    String name
    Person bestBuddy;

    @Backlink
    RealmResults<Person> friends;
}

realm.where(Person.class).equalTo("friends.name", "Jane").findAll();
```

V2: After a migration
```
class Person extends RealmObject {
    String name
    Person bestBuddy;
    Person nemesis;

    @Backlink
    RealmResults<Person> friends;
}

realm.where(Person.class).equalTo("friends.name", "Jane").findAll(); // Will also find all nemesis'.
```

Requiring that all backlinks are fully defined eliminate that risk: 

```
@Backlink("bestBuddy")
realm.where(Person.class).equalTo("name[Person.bestBuddy]", "Jane").findAll();
```

Downside is that the common case becomes a little longer to write

kneth:
What about multiple layers? Say, you have three classes, A, B, and C. A field in A refers to a B object (A.a -> B), a field in B refers to C (B.c -> C), and C has at field one field (c). Currently you can query forward (`where(A).equalTo(a.b.c, 42)`).

If you declare back links using `@BackLink`, should it be support to do `where(C).equalTo(back_to_b.back_to_a.id, 42)`?

spl:
I'm not all that clear on how backlinks are expected to work. Given:

```java
class Person extends RealmObject {
    String name
    Person bestBuddy;

    @Backlink
    RealmResults<Person> friends;
}

public static void init(Realm realm) {
    realm.beginTransaction();
    Person a = realm.createObject(Person.class);
    a.name = "April";
    Person b = realm.createObject(Person.class);
    b.name = "Bob";
    // Fill in the remaining code here.
    realm.commitTransaction();
}
```

What exactly do you expect for `p1.friends` and `p2.friends` after putting each of the following in place of the comment above and running `init(realm)`?

1. `b.bestBuddy = a;`
2. `a.bestBuddy = a;`

If I were to guess based on looking at all of the given code, I would expect `friends` to be an empty `RealmResults<Person>`. Nothing there tells me that `friends` should be linked to anything. If there is an implicit link here between `friends` and `bestBuddy`, it's quite hard to find in the code. I think that would lead to subtle bugs.

cmelchior:
@spl Yes, there is a task in describing the feature in a way that make sense and so people can understand what is going in. Backlinks are also called "Inverse Relationship" in Realm Cocoa so that might be a better name for it.

Generally the idea is that it is just a way to go "go back" along any reference towards a single object. This has a lot of benefits especially in combination with queries.

So in your case the backlink is actually `@Backlink("bestBuddy")`:
``
Alice.bestBuddy = Alice
Bob.bestBuddy = Alice

// Then friends will be the following
Alice.friends = { Alice, Bob }
Bob.friends = { }
```

I hope that makes more sense?

**Comment by cmelchior** [comment]
@kneth I see no reason why we could not not support. For the named backlinks it is just a matter of recursively defining the backlinks in referencering classes. For the unnamed backlinks it would be a matter of introducing something like a wildcard operator. This would also help with @emanuelez concern about it being explicit what the backlink field was, e.g:

```
@Backlink("*") // all fields in parent type
realm.where("C").equalTo("name[B.*.A.foo]").findAll(); // Search all fields in B referencing C and only A.foo that reference B 
```

For the named API it will work fine I think, for the unnamed backlinks It does introduce a layer of complexity I am not entirely comfortable with, especially if you consider polymorphic types, but it would solve the problem.

File: realm/rfc/0001-backlink-query-syntax.md [hunk 1]
```diff
@@ -0,0 +1,204 @@
+- Feature Name: Backlink query syntax
```
bmunkholm: Use same terminology as Cocoa.

spl:
@cmelchior Thanks. I think I see the benefits. I guess it's the implicitness of the backlink that bothers me. What does it depend on? If I understand [your comment above](https://github.com/realm/realm-java/pull/2904#issuecomment-222488106), I'm guessing that the query implicitly looks at all `Person` fields. If this is correct, I would suggest requiring an explicit backlink reference (e.g. `@Backlink "bestBuddy"` as you mentioned) in the annotation and disallowing any implicit backlinks. Intuitively, it makes sense because you want to specify what reference the backlink is to. Would that cause any problems or reduce the expressiveness/usefulness?

realm/rfc/0001-backlink-query-syntax.md
```diff
@@ -0,0 +1,204 @@
...
+This proposal introduces a new syntax for queries so it is possible to query
+across a backlink even though it hasn't been defined by a model class. This is
+called *unnamed backlinks*.
+
+Any design should ideally solve all these constraints:
+
+- Should support a list of backlinks.
+
+- Should support a single backlink, enforcing the constraint that only 1 object 
+  can point to it. This is not supported by Core, but it is a natural extension.
```
astigsen: I think this will be problematic. From a high-level perspective there is nothing like this in object oriented programming. Objects can point to one or multiple other objects (based on the property types), but there is no way to limit who can point to any given object (it is just pointers).

That does not mean that we could not extend further than what is natural in oop (we already do that in several areas), but it will seem as a bit of a foreign concept.

The big question is how it should be enforced? Should it throw an exception if you try to point to an object that already has pointers to it? Or maybe null/remove the the previous link? Could end up seeming a bit "magic".

cmelchior: Throwing an exception seems acceptable to me, especially considering it is a modelling constraint.
The primary driver is just ease of use though, e.g. not having everybody go through a `owners.size() > 0 ? owners.first() : null`. All the proposed solutions in the RFC support adding this in the future though. It does not have to be part of V1

cmelchior:
@spl No, there is no problem requiring that all fields are enumerated. This was the same feedback in https://github.com/realm/realm-java/pull/2904#issuecomment-222488106.

In fact doing that in V1 might be preferable since it would be safer. Then we can always iterate later if needed.

astigsen:
Another alternative syntax would be to make it more explicit that we are talking about backlinks:

```java

// Find objects that nobody are linking to
realm.where("C").backlinks().count() == 0;

// count links from a specific class
realm.where("C").backlinks("B").count() > 0;

// count links from a specific property in a specific class
realm.where("C").backlinks("B", "my_prop").count() > 0;

// find all dogs owned by people in California
realm.where("Dog").backlinks("Person", "dogs").equalTo("state", "CA").findAll()
```

cmelchior:
For trivial queries that might work, but I would be very concerned about the readability if there is multiple query parameters:

```
// What does the last equalTo refer to?
realm.where("Dog").backlinks("Person", "dogs").equalTo("state", "CA").equalTo("name", "Fido").findAll()
```

Another option would be to create a static factory method, which would give you the same textual output as in my RFC, but with better type safety:

```
realm.where("Dog").equalTo(backlink("Person", "dogs"), "state", "CA").equalTo("name", "Fido").findAll()
```

emanuelez:
I quite like @astigsen proposal. One way to deal with more complex queries would be by setting begin and end methods, as we do for group. Ignore the naming, but:

```java
realm.where(Dog.class)
   .backlinks(Person.class, "dogs")
      .equalTo("state", "CA")
   .backlinksEnd()
   .equalTo("name", "Fido")
   .findAll()
```

If no end method is provided then it's considered to be open until the end of the query.

cmelchior:
Keep in mind that we don't need additional constructs for named backlinks, those can easily be handled by our current query methods. So I would be very hesitant to add query constructs only needed by the Dynamic API.

astigsen:
>  I would be very hesitant to add query constructs only needed by the Dynamic API.

I agree. I think the optimal would be if we could have a single interface that worked for both the dynamic and static API.

It is important to also think of how to express more open ended questions, like does this object have any incoming links at all, or any incoming links from this or that type of objects?

realm/rfc/0001-backlink-query-syntax.md
```diff
@@ -0,0 +1,204 @@
...
+**Advantages:**
+- Minimal effort required to implement backlink queries. Current query system is 
+  enough.
+- Already supported by Objective C / Swift
+
+**Disadvantages:**
+- Does not work _at all_ in the dynamic API.
+- User ambiguity regarding manual migrations: Some fields must be added in a 
+  migration, others not needed. This is already happening with ignored fields
+  though.
```
beeender: Also, this won't be naturally working well with async queries.

realm/rfc/0001-backlink-query-syntax.md
```diff
@@ -0,0 +1,204 @@
...
+realm.where(Dog.class).equalTo("owners.name", "John").findAll();
+realm.where(Dog.class).equalTo("favoriteOwner.name")
+
+// Unnamed backlink queries
+realm.where(Dog.class).equalTo("name[Person]", "John").findAll();
+realm.where(Dog.class).equalTo("name[Person.favoriteDog]").findAll();
+
+// Example in a purely text based variant
+"name[Person] = 'John'"
+```
```
beeender: what about
```
realm.where(Dog.class).backlink(Person).equalTo("name", "John);
```

and

```
realm.where(Dog.class).backlink(Person.class, "favouriteDog").equalTo("name", "John);
```

The query syntax makes me feel quite difficult to understand. Especially `name[Person.favoriteDog]`, my first feeling is `name` here is a field of `Person.fovoriteDog`.

beeender: Opps, saw the same proposal from @astigsen . So just ignore me.

realm/rfc/0001-backlink-query-syntax.md
cmelchior: No, it should be possible to do async queries on them as well. owners is not a TableView but a BacklinkView (I won't claim I fully understand the difference yet though). I am just re-using RealmResults instead of inventing a new class to hold them. Cocoa had to implement an entire new class due to other constraints. I am hoping we can avoid that since, for the end user, it could semantically be seen as a view into underlying data (RealmResults).

realm/rfc/0001-backlink-query-syntax.md
cmelchior: @beeender Do you have another suggestion for making it easier to parse? IMO the suggestion by @astigsen and @emanuelez doesn't really work well because you are adding query methods for the sole benefit of the Dynamic API, which is a deal breaker in my mind.

It is fair to consider the proposed syntax to weird, but would it be learnable? Keep in mind that 
that this is a real corner case, since most people will operate on named backlinks that are defined the model class. Do you have a better suggestion for a String based representation that could work for both the static and dynamic API?

Also, if we believe it will ruin the query language to much, I sketched an alternative in https://github.com/realm/realm-java/pull/2904/files#diff-b684aefe1e063faddcad9b319552f6e2R205 that means we will enforce named backlinks as part of the schema. Is that an acceptable solution instead?

cmelchior:
I have updated the RFC with feedback so far. It is now a requirement that all fields are enumerated.

realm/rfc/0001-backlink-query-syntax.md
```diff
@@ -52,19 +53,20 @@ public class Dog extends RealmObject {
...
 
 	// Named backlinks below
-	@Backlink
-	private RealmResults<Person> owner;
-
 	@Backlink("favoriteDog")
 	private RealmResults<Person> favoriteOwner;
+
+	@Backlink("favoriteDog", "dogs")
+	private RealmResults<Person> favoriteOwner;
```
spl: Typo? Duplicate fields: `favoriteOwner`

realm/rfc/0001-backlink-query-syntax.md
```diff
@@ -0,0 +1,231 @@
...
+  `equalTo("y[x.z]", "foo").equalTo("y[x.z1]", "foo")`
+- This is a supplement to defining the named backlinks.
+- No changes to the migration API.
+- No changes to the schema definition.
+
+Example:
+
+```
+// Named backlink queries
+realm.where(Dog.class).equalTo("owners.name", "John").findAll();
```
spl: Typo. I believe `owners` should be `owner`.

spl: Correction: Neither `owner` nor `owners` exist anymore (assuming I'm looking at the right diff).

cmelchior: Thanks. It has been fixed.

spl:
There are several examples in the text with only one argument to `equalTo()`. Is that intentional? (Apologies if I'm nitpicking too much. I'm just trying to understand. Feel free to tell me to go away. 😄)

cmelchior:
@spl No, it is great having someone catching all my stupid mistakes. They have been fixed.

realm/rfc/0001-backlink-query-syntax.md
```diff
@@ -0,0 +1,232 @@
...
+}
+
+public class Dog extends RealmObject {
+	private String name
+
+	// Named backlinks below
+	@Backlink("favoriteDog")
+	private RealmResults<Person> favoriteOwner;
+
+	@Backlink("favoriteDog", "dogs")
```
emanuelez: I'm not sure this kind of syntax is even possible. You can't have varargs as arguments for annotations. What you can do is something like:

```java
public @interface Backlink {
    String[] value();
}
```
which needs to be used like this:
```java
@Backlink({"favoriteDog", "dogs"})
```

realm/rfc/0001-backlink-query-syntax.md
```diff
@@ -0,0 +1,232 @@
...
+
+- "^Person(favoriteDog, dogs).name"
+- "Person(favoriteDog, dogs)<-.name"
+
+
+4) Special grouping operator
+
+```
+realm.where(Dog.class)
+   .backlinks("Dogs", "dogs")
```
emanuelez: Maybe we could even use the already existing `beginGroup()` and `endGroup()`

realm/rfc/0001-backlink-query-syntax.md
cmelchior: You are right. I'll update it.

realm/rfc/0001-backlink-query-syntax.md
```diff
@@ -0,0 +1,232 @@
...
+
+
+2) All fields explicit. Group fields using ()
+
+- "name[Person.(favoriteDog, dogs)]  = 'John'"
+
+
+3) Other modifiers (^ up) , (<- back)
+
+- "^Person(favoriteDog, dogs).name"
```
beeender: `^Person(favoriteDog, dogs).name` for the String based query, this one is my favourite one till now. or maybe use `[]`. Any way `^` makes it clearly a backlink query.

cmelchior: Yes, i kinda like like it as well. Only problem is that describing polymorphic fields will be kinda hard, but then again. It is not entirely pretty in the current incarnation either.
```
// Some random ideas
^* extends Person(favoriteDog, dogs).name
^[* extends Person](favoriteDog, dogs).name
^*(favoriteDog, dogs).name
```

realm/rfc/0001-backlink-query-syntax.md
cmelchior: Is there any particular reason you prefer grouping operators for backlinks? Keep in mind that for  named backlinks they are not needed as normal link syntax will work fine there.

If you find my current syntax proposal too complex or impossible to use, we also have the option of making backlinks part of the schema. Would that be better?

emanuelez: I would really prefer if backlinks were not part of the schema

realm/rfc/0001-backlink-query-syntax.md
```diff
@@ -0,0 +1,232 @@
...
+public class Person extends RealmObject {
+	private Dog favoriteDog;
+	private RealmList<Dog> dogs;
+}
+
+public class Dog extends RealmObject {
+	private String name
+
+	// Named backlinks below
+	@Backlink({"favoriteDog"})
```
bdash: Is `Backlink` the best term to use in the API for this? On the Cocoa side of the world we've completely avoided using this term in our API and documentation since its meaning is not obvious to users without further explanation. I feel like we should make an attempt to keep the terminology used for a given concept consistent between bindings.

realm/rfc/0001-backlink-query-syntax.md
```diff
@@ -0,0 +1,232 @@
...
+
+Backlinks are special in Realm. They are not defined explicitly in the schema,
+but are automatically created when defining Links and LinkLists (forward links).
+
+This creates some conflicts with the current query system as it rely on all 
+fields being defined by the schema. This is called *named backlinks*.
+
+This proposal introduces a new syntax for queries so it is possible to query
+across a backlink even though it hasn't been defined by a model class. This is
+called *unnamed backlinks*.
```
bdash: I'm having a hard time seeing why unnamed backlinks are important enough to warrant special query syntax which makes me feel like I'm missing some of the motivation behind this proposal. Can the "Motivation" section be expanded to discuss the motivation behind this?

realm/rfc/0001-backlink-query-syntax.md
```diff
@@ -0,0 +1,232 @@
...
+
+Any design should ideally solve all these constraints:
+
+- Should support a list of backlinks.
+
+- Should support a single backlink, enforcing the constraint that only 1 object 
+  can point to it. This is not supported by Core, but it is a natural extension.
+
+- Should support backlinks from multiple fields.
```
bdash: What do you mean by this?

realm/rfc/0001-backlink-query-syntax.md
```diff
@@ -0,0 +1,232 @@
...
+
+- Should be usable from the Dynamic API.
+
+- Should either support polymorphic objects or not prevent them from being 
+  implemented.
+
+- Ideally it should work for both a pure text-based variant as well as the
```
bdash: Is this referring to a text-based query language?

realm/rfc/0001-backlink-query-syntax.md
cmelchior: `Backlink` is just so much shorter and better sounding IMO :) ... `@ReverseRelationship` is awful long and it isn't a concept in Java the same way that it is in CoreData either :(.

That said, you are probably right that people can more easily guess the meaning of the word as `InverseRelationship`, and given that you already named it as such in Cocoa probably means that my "fluffy" reasons for liking Backlink better is not good enough to warrant a difference between the Cocoa and Java API's.

One thing though is that `Inverse Relationship` is not a well defined concept in Java compared to CoreData.

realm/rfc/0001-backlink-query-syntax.md
```diff
@@ -0,0 +1,232 @@
...
+are generally low, this shouldn't be a problem in practise.
+
+**Advantages**
+- Using "[]" is not an allowed field characters in any of our supported 
+  languages as far as I know. So parsing the above expressions should be 
+  possible with no ambiguity.
+
+- The default case is easy to describe.
+
+- Using [] as a modifier already has precedent in NSPredicate.
```
bdash: This use of `[]` doesn't feel at all similar to how `NSPredicate` uses it (modifying a comparison operator to indicate that it should be performed case- or diactrictic-insensitively). If anything this would be a disadvantage since it conflicts with how `NSPredicate` uses those symbols, adding potential confusion.

realm/rfc/0001-backlink-query-syntax.md
```diff
@@ -0,0 +1,232 @@
...
+- Unknown syntax. This concept is not used nor supported in neither NSPredicate, 
+  LINQ or SQL. It will be a "power feature", and might make the language feel 
+  complicated and hard to use.
+
+- Moves the Java query language further away from being type safe.
+
+- Will introduce two ways of querying backlinks.
+
+- Without wildcards for unnamed backlinks queries risk getting extremely verbose.
+
```
bdash: Disadvantage: no way to understand what the syntax means for a new user.
Disadvantage: syntax using only special characters is very hard to search for documentation on.

realm/rfc/0001-backlink-query-syntax.md
```diff
@@ -0,0 +1,232 @@
...
+}
+
+public class Dog extends RealmObject {
+	private String name
+
+	// Named backlinks below
+	@Backlink({"favoriteDog"})
+	private RealmResults<Person> favoriteOwner;
+
+	@Backlink({"favoriteDog", "dogs"})
```
bdash: How are the backlinks from two properties combined?

AFAICT this isn't something that is supported by core.

bdash:
It's not entirely clear to me from this RFC which aspects of it you intend to tackle immediately vs those that are future directions.

realm/rfc/0001-backlink-query-syntax.md
cmelchior: Oh, I kinda assumed that was the case. It will simply things a lot though and the use cases where it is relevant are not entirely clear either. I'll rewrite it so this part is moved into future extensions.

cmelchior:
The immediate parts, is the Java annotations (named backlinks) and the query constructs needed by the dynamic API as described here: https://github.com/realm/realm-java/pull/2904/files#diff-b684aefe1e063faddcad9b319552f6e2R64 Any hints on how to better structure it so that is more clear would be appreciated. 

Also, if you don't like the proposed query syntax. What do you feel about the proposed alternatives?

realm/rfc/0001-backlink-query-syntax.md
bdash: The term used in the Cocoa API is `LinkingObjects`, which is a collection type. "Inverse relationship" is how we talk about the concept in the documentation.

realm/rfc/0001-backlink-query-syntax.md
cmelchior: Yes

realm/rfc/0001-backlink-query-syntax.md
cmelchior: Fair enough. How do you feel about any of the alternatives or do you have a completely different suggestion?

realm/rfc/0001-backlink-query-syntax.md
cmelchior: Sure. Primary reason is Migrations where you don't have Realm model classes at all and in Java you can also access a Realm in "full dynamic mode", where no Realm classes needs to be defined either. This is e.g. used by browsers and other libraries that work with data in a largely dynamic fashion (CSV import ).

realm/rfc/0001-backlink-query-syntax.md
cmelchior: I got the impression that you only added "LinkingObjects" because you couldn't re-use Results? In our case that seems entirely doable, so I would rather avoid having to introduce a new base class.

realm/rfc/0001-backlink-query-syntax.md
cmelchior: `@Backlink({"foo", bar"})`, but as pointed out here https://github.com/realm/realm-java/pull/2904#discussion_r65192945 that might not be supported by Core, so it should be removed.

realm/rfc/0001-backlink-query-syntax.md
bdash: Prior to the introduction of that type we had a method that returned an array of objects. It also used the term "linking objects" to refer to the objects involved in the inverse relationship.

bdash: And note that I'm raising an issue with the term `Backlink` being used for the annotation, not suggesting that a new type be used in place of `RealmResults`.

realm/rfc/0001-backlink-query-syntax.md
bdash: To me those seem like cases where it would be _nice_ to support querying over unnamed backlinks, but they're much less important than supporting named backlinks. I'm curious why you feel they're important enough to warrant tackling in your initial implementation? On the Realm Cocoa side we concluded that they weren't necessary for our initial implementation, and we wanted to avoid inventing special query syntax to accommodate them at this point.

cmelchior: 1) Because our Migration API works differently than Cocoas. Java's is intended to provide the full functionality of Realm, just in a dynamic way.
2) Not having feature parity between the static and dynamic API is harder to document and support (= potentially more support).
3) We have time to actually implement it now. Punting just to avoid the discussion seems...odd.

bdash:
The sort of syntax that I have in mind for a `NSPredicate`-style query over unnamed backlinks would look like:

```swift
// Children with a parent named Mark.
ANY linkingObjects(Person.children).name == 'Mark'

// Children with no grandparent named Michael.
NONE linkingObjects(Person.children).linkingObjects(Person.children).name == 'Michael'
```

It's more explicit, but IMO that is the right tradeoff for syntax that's likely to be used infrequently. It also allows for other potential extensions to the query language without inventing new syntax.

realm/rfc/0001-backlink-query-syntax.md
bdash: Ok. These are all things that should be mentioned in the motivations section since without that context it's hard to see why it's worth going to the effort of inventing query support for the dynamic API.

realm/rfc/0001-backlink-query-syntax.md
bdash: Core exposes backlinks as a `TableView`.

cmelchior:
Yes, that sort of syntax has the advantage that it read like methods, which 1) hopefully makes it really explicit and 2) is easy to extend to other cases without introducing all sorts of weird characters.

cmelchior:
RFC updated: @bdash @realm/java

Major changes:
- Only one field is allowed pr. annotation
- Annotation name changed to `@LinkingObjects`
- Syntax changed to `linkingObjects(x.y).z`

cmelchior:
@bdash Can you elaborate on what led to you documenting this as `Inverse Relationship` while using `LinkingObjects` in the API? It confuses me a bit, especially since `LinkingObjects` seems to be an invented term. What was the reason for not just using Backlink in the API?

spl:
Minor comment with reasoning: I do like “linking object” over either “back link” or “inverse relationship.” The former is straightforward in its description of an object linking to others, but the latter two require you to determine what is meant by the direction “back” or “inverse.” This indirection adds unnecessary cognitive overhead, I think.

cmelchior:
@spl Thanks! It is nice with input from someone who are not tainted by all our internal discussions. These sort of naming decisions are better if looked at with fresh eyes.

realm/rfc/0001-backlink-query-syntax.md
```diff
@@ -0,0 +1,246 @@
...
+- "linkingObjects(*.*).name  = 'John' // Include all backlink types
+- "linkingObjects(Person.dogs, Person.favoriteDog).name = 'John' // Multiple fields at the same time.
+```
+
+However this is not part of this RFC.
+
+
+# Drawbacks
+
+It is not clear that his approach is the correct way forward. I am not sure what
```
stk1m1: `this approach` ?

realm/rfc/0001-backlink-query-syntax.md
```diff
@@ -0,0 +1,246 @@
...
+- Require changes to the manual Migration API.
+
+
+# Unresolved questions
+
+None.
+
+# Version history
+
+1: Initial version
```
stk1m1: double space at the end `  ` would be nice.

realm/rfc/0001-backlink-query-syntax.md
```diff
@@ -0,0 +1,246 @@
...
+**Example model classes:**
+
+```
+public class Person extends RealmObject {
+	private Dog favoriteDog;
+	private RealmList<Dog> dogs;
+}
+
+public class Dog extends RealmObject {
+	private String name
```
stk1m1: Shouldn't this `String name` field go to `Person`? 

Correct me if I am wrong. I am percieving that we are to look for `Person` rather than `Dog` when we're to search `realm.where(Dog.class).equalTo("favoriteOwner.name", "John").findAll();`.

spl: > I am perceiving that we are to look for `Person` rather than `Dog` when we're to search `realm.where(Dog.class).equalTo("favoriteOwner.name", "John").findAll();`.

I understood it to be a query for `Dog` since it has `where(Dog.class)`. That indicates some confusion on one or both of our sides. 😉

stk1m1:
**This is just 2 cents.** 
I'm definitely more leaning toward `@Backlink` as it is more intuitive and less fancy. I think it should be considered thoroughly to push aside _the intuitive name_ in lui of unified terminology across bindings such as `@LinkingObject`. 

`InverseRelation` is fine but one would ask then why not just say the obvious? Besides, it is taken in CoreData.

realm/rfc/0001-backlink-query-syntax.md
stk1m1: > > I am perceiving that we are to look for `Person` rather than `Dog` when we're to search `realm.where(Dog.class).equalTo("favoriteOwner.name", "John").findAll();`.

> I understood it to be a query for `Dog` since it has `where(Dog.class)`. That indicates some confusion on one or both of our sides. 😉 

So in a humane term, `we are looking for all dogs whose favorite owner's name is John`. That makes sense. Thanks for correcting me. Then isn't it more clear to put `String name` in `Person`?

realm/rfc/0001-backlink-query-syntax.md
```diff
@@ -0,0 +1,246 @@
...
+  field Z. This is a supplement to defining the named backlinks, which will continue to work as normal
+  link queries: `favoriteOwner.name`
+- No changes to the migration API.
+- No changes to the schema definition.
+
+Example:
+
+```
+// Named backlink queries
+realm.where(Dog.class).equalTo("favoriteOwner.name", "John").findAll();
```
stk1m1: It would be nice to put `RealmResults<Dogs> dogs = ` at front for dull users like me.   :blush:

realm/rfc/0001-backlink-query-syntax.md
cmelchior: Yes, it is a mistake it is not there. I'll fix it :)

bdash:
@cmelchior, I wasn't present when the initial decision was made to use the term "linking objects" in the Cocoa API so I can't speak to how it was settled on. The reason @spl mentions for liking the term is a strong one, though.

The use of the term "inverse relationship" in the documentation is because they are discussed as part of the [greater concept relationships in Realm](https://realm.io/docs/swift/latest/#relationships), with subsections of To-One Relationships, To-Many Relationships, and Inverse Relationships. I feel that these terms are relatively easy to understand in context, but they don't lend themselves well to naming functions or types in code due to their describing relationships between objects rather than the objects that make up one side of the relationship. "Backlinks" has a similar problem in this regard.

@stk1m1, IMO it's difficult to argue that "backlinks" is such a superior term that we should introduce a significant naming inconsistency between bindings. I also don't think any of the terms involved here are particularly intuitive. I suspect familiarity with core's use of the term "backlinks" is influencing people's opinions on that matter.

stk1m1:
@bdash Thanks for the pointers in your comments. I wasn't fully aware that "Inverse Relation" is a part of the great concept Realm aims to approach. I'll take another read.

My motivation for favoring `BackLink` is due to its bit more precise (yes, I for one) wording on describing the nature of relation between two entities. When an inverse relation is in use, it more often than not points a relation linking _up_ to a higher order entity in an object relation graph. The order could change from where one stands, but it usually implies to be top-down. In that sense, `BackLink` does not carry an extra baggage for recognizing its character.

If the object graph tends to be flat and level-less as depicted in the relationships in Realm, then what @spl points is pretty accurate. Plus, I'd like to clarify myself that I am to fully support the consistency across binding by all means and all the times possible.

<sup>*</sup>My hands indeed have longer physical, body memory in handing ObjC + CoreData than doing Java that I can attest I am not under the influence of core's exchange. 😃

realm/rfc/0001-backlink-query-syntax.md
```diff
@@ -0,0 +1,249 @@
...
+public class Dog extends RealmObject {
+	// Named backlinks
+	@LinkingObjects("favoriteDog")
+	private RealmResults<Person> favoriteOwner;
+}
+```
+
+**Introduce the following query extension**
+
+- Use `linkingObjects(x.z).y` to search field Y on type X where X is a backlink object through
```
stk1m1: > Use `linkingObjects(x.z).y` to search field Y on type X where X is a backlink object through field Z.

I think this is unnamed backlink, and it would be nice if `unnamed` is mentioned first. Further X is called type and object at the same time. From the look of examples, I would suggest drop the first `type` like below.

> Use `linkingObjects(x.z).y` to search field Y on X where X is an unnamed backlink type through field Z.

Alex293:
Hi, nice work here
What is the status of this ?
Is it ready for testing purposes ?

cmelchior:
It appears that .NET have called their feature "Backlinks" https://realm.io/docs/xamarin/latest/#inverse-relationships.

IMO there is no good reason for having different names for the same concept across bindings, so we should probably align on this.

So far it looks like:

* The feature is named "Inverse Relationship" in the docs
* Swift uses a special class "LinkingObjects(fromType: Person.self, property: "dogs")" to capture the intent: https://realm.io/docs/swift/latest/#inverse-relationships
* .NET uses a Backlink annotation: `[Backlink(nameof(Person.Dogs))]` : https://realm.io/docs/xamarin/latest/#inverse-relationships

Zhuinden:
LinkingObjects is so much less obvious. -_-

The realm-cocoa binding has some really strange naming schematics. Probably because it's what came first.

realm/rfc/0001-backlink-query-syntax.md
```diff
@@ -0,0 +1,249 @@
...
+```
+// Named backlink queries
+RealmResults<Dog> dogs = realm.where(Dog.class).equalTo("favoriteOwner.name", "John").findAll();
+
+// Unnamed backlink queries
+RealmResults<Dog> dogs = realm.where(Dog.class).equalTo("linkingObjects(Person.dogs).name", "John").findAll();
+RealmResults<DynamicRealmObjcet> dogs = dynamicRealm.where("Dog").equalTo("linkingObjects(Person.dogs).name", "John").findAll();
+
+// Example in a purely text based variant
+"ANY linkingObjects(Person.dogs).name = 'John'"
```
astigsen: The dynamic interface also opens the possibility of more expressive queries where you don't have to specify the specific properties the links come from:

```java
// Find all objects with no incoming links (useful for manual handling of cascading deletes)
RealmResults<Dog> dogs = realm.where(Dog.class).equalTo("linkingObjects().count()", 0).findAll();

// count links from a specific class (without caring about which properties are linked from)
RealmResults<Dog> dogs = realm.where(Dog.class).equalTo("linkingObjects(Person).count()", 0).findAll();
```

kneth: The cost is that we have to parse the query (the string in `equalTo()` in the snippet above). If we decide to go down that route, I think it is better to use the query language implemented in Object Store.

kneth:
Closing as #4704 has been merged

Zhuinden:
The newly added/merged linking object queries also support queries across unnamed backlinks, like

        private int age;
    }

    public class Owner extends RealmObject {
         private String name;
         private RealmList<Dog> dogs;
    }

    realm.where(Dog.class).equalTo("linkingObjects(Person.dogs).name", "Jack").findAll();

?

kneth:
@Zhuinden Correct, we haven't added the logic to deduce the inverse relationships. In some cases, it is - for a human - trivial.

marchy:
@kneth How does the current implementation handle cases where you want to find all objects which do NOT have the existing linked object expression? ie: the equivalent of RealmSwift's ```SUBQUERY(linkedThings, $linkedThing, $linkedThing.owner.id == '\(User.currentUserID)').@count == 0``` (ie: "return linked things which the current user DOESN'T own")

With the current Android implementation it seems possible to do a positive check (ie: query for linkedThings which have a specific owner (through ```query.equalTo( "linkedThings.owner.id", User.currentUserPersonID.stringValue )```, but how can you check for linkedThings that do NOT have the particular owner?

You can't just do ```notEqualTo``` since that could return false positives – ie: return linked things which have other owners, but those things could also have the given owner, which is what you're trying to exclude.

PR #2908, opened 2016-05-30
PrimaryKey index rearrangement in migration revisited

### LINKED ISSUE #2829: Primary key not defined for field 'id' in existing Realm file. Add @PrimaryKey.
Hello! I upgraded from 0.88.2 to 0.90.1 and refactored my RealmObject classes and renamed them. Everything works fine if I clear the app data and it's a new, clean Realm, but migrating from the old version doesn't work. 

Stacktrace:
>io.realm.exceptions.RealmMigrationNeededException: Primary key not defined for field 'id' in existing Realm file. Add @PrimaryKey.
at io.realm.ChannelRealmProxy.validateTable(ChannelRealmProxy.java:154)
at io.realm.DefaultRealmModuleMediator.validateTable(DefaultRealmModuleMediator.java:71)
at io.realm.Realm.initializeRealm(Realm.java:316)
at io.realm.Realm.createAndValidate(Realm.java:284)
at io.realm.Realm.createInstance(Realm.java:263)
at io.realm.RealmCache.createRealmOrGetFromCache(RealmCache.java:118)
at io.realm.Realm.getDefaultInstance(Realm.java:195)
...

Channel.java:
```java
public class Channel extends RealmObject {
    @PrimaryKey
    private long id;
    private String name;
    //...
```
DbMigration.java:
```java
//Don't want to create a Channel schema, if the previously named one exists
if(schema.get("TvObject") == null && schema.get("Channel") == null) {
    schema.create("Channel")
        .addField("id", long.class, FieldAttribute.PRIMARY_KEY)
        //redacted, just a bunch of String, int and boolean fields
}
//...
if(oldVersion == 7) { 
    schema.rename("TvObject", "Channel");
    schema.get("Channel")
        //redacted the changes, it's a bunch of field renaming and removing
    oldVersion++;
}
```
It also happens with two other classes that I renamed, and it's not always the same one that causes the crash. It changes when I redo the upgrade from the old app version to the new one.

I've tried using Long with ```@Required``` instead of long, adding the primary key in the migration (which causes an exception), nothing helps. Am I doing something wrong? 

P.S. Is there a better way to structure migrations with regards to creating the objects? If I rename the objects again, I'll have three ```== null``` checks in the if()s.

stk1m1:
This PR fixes followings. 

1. When a column with a smaller index than that of a primary key field gets removed, the primary key field index decrements, which causes a cached primary key index to point a wrong field. 

2. When a primary key field is renamed, primary key table is not updated accordingly.

Closes #2829.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * Fixed a crash when calling Table.toString() in debugger (#2429).
 * Revised `RealmResults.isLoaded()` description (#2895).
+* Fixed a bug when migration loses tracking of primary key field in `RealmObjectSchema.removeField()` and `RealmObjectSchema.renameField()` (#2829).
 
 ## 1.0.0
 
--- a/realm/realm-library/src/androidTest/java/io/realm/RealmMigrationTests.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/RealmMigrationTests.java
@@ -33,6 +33,7 @@ import java.io.IOException;
 import java.util.Date;
 import java.util.concurrent.atomic.AtomicBoolean;
 
+import io.realm.entities.AllJavaTypes;
 import io.realm.entities.AllTypes;
 import io.realm.entities.AnnotationTypes;
 import io.realm.entities.FieldOrder;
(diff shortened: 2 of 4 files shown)


stk1m1:
@realm/java please review.

cmelchior:
Since this is a patch, it should be against releases?

stk1m1:
@cmelchior 

I'll close this and create one against `releases`.

PR #2912, opened 2016-05-31
Add RealmInvalidDatabaseException

beeender:
This is necessary.
For some reasons, the db file might get corrupted (the most reasonable
thing is final user install an old version of app which tries to open a
Realm db created by newer version of Realm). The only way to solve that
is deleting the Realm file and start over.

Instead of asking user to catch IllegalStateException and checking the
message, throw a special exception for the purpose is more feasible.


beeender:
@realm/java

CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## 1.1.0
+
+### Breaking changes
+
+* A `RealmInvalidDatabaseException` is thrown instead of `IllegalStateException` when opening a corrupted Realm file, a non-Realm file, a Realm file created by newer version of Realm, or an encrypted Realm file with a wrong key.
```
cmelchior: Isn't this technically a 2.0 change then?

realm/realm-library/src/main/java/io/realm/exceptions/RealmInvalidDatabaseException.java
```diff
@@ -0,0 +1,14 @@
+package io.realm.exceptions;
+
+import io.realm.internal.Keep;
+
+/**
+ * {@code RealmInvalidDatabaseException} is thrown when opening a corrupted Realm file, a non-Realm file, a Realm file
+ * created by newer version of Realm, or an encrypted Realm file with a wrong key.
```
cmelchior: newer versions

kneth: or "a newer version"

beeender:
Hold on until we figure out which branch to go

cmelchior:
Outdated. Closing

PR #2915, opened 2016-06-01
Add object-store as subtree

beeender:
**DO NOT SQUASH THIS WHEN MERGING!!!**

To get a brief idea of "git subtree", see http://blogs.atlassian.com/2013/05/alternatives-to-git-submodule-git-subtree/

This add source code of `realm-object-store` repo as a subtree in `realm/realm-jni/object-store` folder.


cmelchior:
Can you make a quick list of advantages and disadvantages using subtrees vs. submodules?

beeender:
I think the best thin I like from subtree is actually we don't have to wait commit merged to the object-store before we use it in realm-java ---- We can always do it later. This freedom is not easy to get from submodule.

Also since we are using `git subtree` to create a copy of object-store repo in realm-java, then outside users don't need permission to object-store repo to compile realm-java. This is a big win from `git subtree` otherwise we need to setup some script to do it and make it part of the release process.

And @emanuelez will tell you, `submodule` is stupid for no reason.
====

Copied from http://blogs.atlassian.com/2013/05/alternatives-to-git-submodule-git-subtree/
Why use subtree instead of submodule?

There are several reasons why you might find subtree better to use:

* Management of a simple workflow is easy.
* Older version of git are supported (even before v1.5.2).
* The sub-project’s code is available right after the clone of the super project is done.
* subtree does not require users of your repository to learn anything new, they can ignore the fact that you are using subtree to manage dependencies.
* subtree does not add new metadata files like submodules doe (i.e. .gitmodule).
* Contents of the module can be modified without having a separate repository copy of the dependency somewhere else.

In my opinion the drawbacks are acceptable:

* You must learn about a new merge strategy (i.e. subtree).
* Contributing code back upstream for the sub-projects is slightly more complicated.
* The responsibility of not mixing super and sub-project code in commits lies with you.

bdash:
I don't think submodules are an option here as the object store repository is private.

stk1m1:
👍  Submodule has given me enough pain as a user. It seems commiters need to be extra careful though.

cmelchior:
Subtree sounds good to me 👍

PR #2916, opened 2016-06-01
Build Object-Store

beeender:
**DO NOT SQUASH WHEN MERGING!! #2915 SHOULD BE MERGED FIRST!!!**

All credits go to @kneth 

1. Fixed compile errors when build with NDK, and changes are intended to be in one separated commit which will be pushed to os repo as well.
2. Continue kg's work to build the object store source.

### DIFF
--- a/realm/realm-jni/build.gradle
+++ b/realm/realm-jni/build.gradle
@@ -20,7 +20,7 @@ ext.coreArchiveDir = System.getenv("REALM_CORE_DOWNLOAD_DIR")
 // To obtain the ABI of the connected device, execute "adb shell getprop ro.product.cpu.abi"
 ext.buildTargetAbis = project.hasProperty('buildTargetABIs') ? project.getProperty('buildTargetABIs').split(',').collect {it.trim()} : null
 
-def commonCflags = [ '-Os', '-std=c++11' ]
+def commonCflags = [ '-Os', '-std=c++14' ]
 // LTO and debugging don't play well together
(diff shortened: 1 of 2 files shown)


beeender:
CI fails because of https://code.google.com/p/android/issues/detail?id=62648
It fails on arm ARCH only because `exception_ptr` is not available .... :(

realm/realm-jni/build.gradle
```diff
@@ -75,7 +75,8 @@ def toolchains = [
 ]
 
 def allTargets = [
-    new Target( name:'arm',     abi:'armeabi',     toolchain:toolchains.find {it.name == 'arm'},    cflags:[ '-mthumb' ] ),
+    // gcc for armeabi has problem to compile std::excpetion_ptr. Disable this ABI for now. We might drop it or use the new NDK clang to compile it.
+    //new Target( name:'arm',     abi:'armeabi',     toolchain:toolchains.find {it.name == 'arm'},    cflags:[ '-mthumb' ] ),
```
stk1m1: Are we going to stop `armabi` support for sure?

kneth: clang seems to support `exception_ptr` but we are still investigating.

stk1m1: > clang seems to support `exception_ptr` but we are still investigating.

Never mind about my question. I almost forgot the point. Thanks for the reminder. 👍

PR #2920, opened 2016-06-01
 PrimaryKey index rearrangement in migration revisited 

### LINKED ISSUE #2829: Primary key not defined for field 'id' in existing Realm file. Add @PrimaryKey.

stk1m1:
This PR fixes followings. 

1. When a column with a smaller index than that of a primary key field gets removed, the primary key field index decrements, which causes a cached primary key index to point a wrong field. 

2. When a primary key field is renamed, primary key table is not updated accordingly.

Closes #2829. 

@realm/java please review.


realm/realm-library/src/androidTest/java/io/realm/RealmMigrationTests.java
```diff
@@ -227,6 +228,54 @@ public void migrate(DynamicRealm realm, long oldVersion, long newVersion) {
...
+            public void migrate(DynamicRealm realm, long oldVersion, long newVersion) {
+                realm.getSchema()
+                        .rename("AllJavaTypes", "PrimaryKeyAsLong")
+                        // There are three columns coming prior to the PK field 'fieldLong', and
+                        // removing "fieldString", "fieldShort" & "fieldInt" should not rearrange the PK
+                        .removeField("fieldString")
+                        .removeField("fieldShort")
+                        .removeField("fieldInt")
+
+                        // removal of rest fields should not affect PK attribute
```
cmelchior: IMO You cannot test all 3 conditions at the same time. This should be split into 3 tests. 1) Remove prior columns, remove next columns and rename

kneth:
:+1:

realm/realm-library/src/main/java/io/realm/RealmObjectSchema.java
```diff
@@ -204,6 +204,14 @@ public RealmObjectSchema removeField(String fieldName) {
         if (table.getPrimaryKey() == columnIndex) {
             table.setPrimaryKey(null);
         }
+        // When PK does not exist (-2) this cannot happen. Meanwhile a removal target has a smaller
```
cmelchior: Can you expand this comment because I don't understand it?
Primary keys should be saved by their field name, so removing a field should not impact the primary key stored in the pk table as far as I know?

realm/realm-library/src/main/java/io/realm/RealmObjectSchema.java
```diff
@@ -204,6 +204,14 @@ public RealmObjectSchema removeField(String fieldName) {
...
+        // index than the PK index, we can expect an rearrangement.
+        if (columnIndex < table.getPrimaryKey()) {
+            String pkField = getPrimaryKey();
+            table.removeColumn(columnIndex);
+            table.setPrimaryKey(pkField);
+            return this;
+        }
         table.removeColumn(columnIndex);
```
cmelchior: You are calling `table.removeColumn()` twice for the same columnIndex. Don't you risk deleting columns?

realm/realm-library/src/main/java/io/realm/internal/Table.java
```diff
@@ -204,6 +204,16 @@ public void removeColumn(long columnIndex) {
     @Override
     public void renameColumn(long columnIndex, String newName) {
         verifyColumnName(newName);
+        if (getPrimaryKey() == columnIndex) {
+            Table pkTable = getPrimaryKeyTable();
+            String className = tableNameToClassName(getName());
+            long rowIndex = pkTable.findFirstString(PRIMARY_KEY_CLASS_COLUMN_INDEX, className);
+            if (rowIndex != NO_MATCH) {
+                pkTable.getUncheckedRow(rowIndex).setString(PRIMARY_KEY_FIELD_COLUMN_INDEX, newName);
```
cmelchior: You can just call `setString` directly on the table. No need to create an unchecked row for it.

realm/realm-library/src/main/java/io/realm/RealmObjectSchema.java
```diff
@@ -204,6 +204,14 @@ public RealmObjectSchema removeField(String fieldName) {
         if (table.getPrimaryKey() == columnIndex) {
             table.setPrimaryKey(null);
         }
+        // When PK does not exist (-2) this cannot happen. Meanwhile a removal target has a smaller
+        // index than the PK index, we can expect an rearrangement.
+        if (columnIndex < table.getPrimaryKey()) {
+            String pkField = getPrimaryKey();
+            table.removeColumn(columnIndex);
+            table.setPrimaryKey(pkField);
```
cmelchior: Reading through the Table class.  This seems like a hack to recalculate the `cachedPrimaryKeyColumnIndex` in the `Table` class. Wouldn't it be easier to have a `invalidatePrimaryKeyIndex()` method instead that could set the new proper index?

realm/realm-library/src/main/java/io/realm/RealmObjectSchema.java
stk1m1: It is true that removing a field does not affect PK stored in PK table. The removal however could generates index re-ordering of PK column if a column prior to PK gets removed. For example

```java
table.addField("a"); // index of 0
table.addField("b"); // index of 1
table.addField("c"); // index of 2
table.setPrimaryKey("b");
```
Now let's remove field `a`. 
```java
table.removeField("a");
```
Then `b`'s index is 0, and `c`'s index is 1 while `cachedPrimaryKeyColumnIndex` still has the value of 1 pointing to `c`. 

Then let's remove field `c`. 
```java
table.removeField("c");
```
What happens is `if (table.getPrimaryKey() == columnIndex)` becomes true and PK gets removed from PK table. `cachedPrimaryKeyColumnIndex` gets invalidated as well. 

**I was only looking for removing non-PK field `c` though**.

Whenever we are to remove a column with smaller index than that of PK field, we should recalculate `cachedPrimaryKeyColumnIndex` to properly point the correct PK field.

- - - 
So this what I did. 

When execution reaches at the point of `if (columnIndex < table.getPrimaryKey())`, there should be only two cases; 
1. PK is removed or does not exists so `table.getPrimaryKey()` will return -2. `columnIndex` should have proper value (i.e. 0 or bigger) that the condition should be false.
2. PK exists and has a bigger index value than `columnIndex` and this is exactly what we are looking for. So we go into the path recalculate `cachedPrimaryKeyColumnIndex` again and return.

- - - 
After writing down the detail, I can see I should have made the code more readable. I'll revise them.

realm/realm-library/src/main/java/io/realm/RealmObjectSchema.java
stk1m1: Uh, not really. A column will only be removed once. You can see it in [my comment](https://github.com/realm/realm-java/pull/2920/files#r65501958)

realm/realm-library/src/main/java/io/realm/internal/Table.java
stk1m1: I'll fix it in no time.

realm/realm-library/src/androidTest/java/io/realm/RealmMigrationTests.java
stk1m1: I was looking into it. But then I would create two more not much useful test entities that I avoided it. I'll create them and build proper unit tests.

realm/realm-library/src/main/java/io/realm/RealmObjectSchema.java
stk1m1: `invalidatePrimaryKeyIndex()` is something I was really looked into it. Then I realize that this is a patch and 1) I do not want to create a public method and 2) OS is coming.

cmelchior: Table is in the internal API, so we can do anything we want there. Also we shouldn't make bad code just because it might be replaced in a few months :)

stk1m1:
**What is done for 2nd round**

1. Since this builds up from `Table`, few more tests are added to `RealmAnnotationTests`, `RealmObjectSchemaTests` and `RealmMigrationTests`. The ones that went to `RealmAnnotationTests` should be placed in `RealmObjectSchemaTests`, but the annotation tests has more tests related to removal and renaming of primary key field.
2. `RealmMigrationTests` now has 1) renaming of a schema, 2) removal of prior/posterior indexes to a PK, and 3) renaming of a PK field.
3. `RealmObjectSchema` only judges whether it is dealing with a PK column or not and routes to an appropriate method of `Table` for an action.
4. `Table` does all the heavy-lifting since it has direct access to PK table.

@realm/java please review.

realm/realm-library/src/main/java/io/realm/internal/Table.java
```diff
@@ -1025,6 +1025,39 @@ private Table getPrimaryKeyTable() {
...
+     * @throws {@link io.realm.exceptions.RealmException} if a primary key does not exist.
+     */
+    public void renamePrimaryKeyColumn(String newFieldName) {
+        if (newFieldName == null || newFieldName.length() == 0) {
+            throw new IllegalArgumentException("Primary key field cannot be renamed with an empty string.");
+        }
+        verifyColumnName(newFieldName);
+        long pkColumnIndex = getPrimaryKey();
+        if (pkColumnIndex == NO_PRIMARY_KEY) {
+            throw new RealmException("Non-existent primary key field cannot be renamed.");
```
beeender: Maybe an `IllegalStateException` is more suitable? @realm/java

realm/realm-library/src/main/java/io/realm/RealmObjectSchema.java
```diff
@@ -205,6 +205,11 @@ public RealmObjectSchema removeField(String fieldName) {
             table.setPrimaryKey(null);
         }
         table.removeColumn(columnIndex);
+        // When PK does not exist this cannot happen. But if you remove a column with a smaller index
+        // than that of PK column, you need to rearrange PK column index.
+        if (table.hasPrimaryKey() && columnIndex < table.getPrimaryKey()) {
+            table.invalidateCachedPrimaryKeyIndex();
+        }
```
beeender: What about calling `invalidateCachedPrimaryKeyIndex` in `Table.removeColumn` and make it private?

realm/realm-library/src/androidTest/java/io/realm/RealmAnnotationTests.java
```diff
@@ -227,6 +228,63 @@ public void primaryKey_isIndexed() {
         assertTrue(table.hasSearchIndex(table.getColumnIndex("id")));
     }
 
+    @Test
```
beeender: Those three tests should be removed to `JNITableTest` since they cannot be compiled after OS and they are more like unit testing for `renamePrimaryKeyColumn`. And it doesn't really require the dependencies from defined schema. I would like to create the shemas with `Table`'s API, and test them in `JNITableTest`.

realm/realm-library/src/androidTest/java/io/realm/entities/AllJavaTypes.java
```diff
@@ -44,6 +44,8 @@
     public static final String   INVALID_LINKED_BINARY_FIELD_FOR_DISTINCT = AllJavaTypes.FIELD_OBJECT + "." + AllJavaTypes.FIELD_BINARY;
     public static final String[] INVALID_LINKED_TYPES_FIELDS_FOR_DISTINCT = new String[]{FIELD_OBJECT + "." + FIELD_BINARY, FIELD_OBJECT + "." + FIELD_OBJECT, FIELD_OBJECT + "." + FIELD_LIST};
 
+    public static final long DEFAULT_PRIMARYKEY_INDEX = 3;
```
beeender: I would like to not assume pk index as a constant 3. Instead, it will be more helpful by checking if the PK index is correct (than checking the result decreased every time). I will decouple the test case from the `Table` implementation to make it still useful after we moved to OS implementation.

realm/realm-library/src/androidTest/java/io/realm/RealmMigrationTests.java
```diff
@@ -226,6 +231,132 @@ public void migrate(DynamicRealm realm, long oldVersion, long newVersion) {
         }
     }
 
+    // this builds a temporary schema that does not exist in io.realm.entities
+    private void buildBaseMigrationSchema(final String schemaName, final boolean createBase) {
+        RealmConfiguration originalConfig = configFactory.createConfigurationBuilder().build();
+        // Ensure that there is no data
+        Realm.deleteRealm(originalConfig);
```
beeender: `deleteRealm` is not needed. It is guaranteed by `configFactory`

realm/realm-library/src/androidTest/java/io/realm/RealmMigrationTests.java
```diff
@@ -226,6 +231,132 @@ public void migrate(DynamicRealm realm, long oldVersion, long newVersion) {
...
+        Realm realm = Realm.getInstance(originalConfig);
+        realm.executeTransaction(new Realm.Transaction() {
+            @Override
+            public void execute(Realm realm) {
+                // firstly remove an existing schema
+                realm.getSchema().remove(schemaName);
```
beeender: Is this necessary?

realm/realm-library/src/main/java/io/realm/RealmObjectSchema.java
stk1m1: Good point. I'll do that.

realm/realm-library/src/androidTest/java/io/realm/RealmAnnotationTests.java
stk1m1: Would it be fine to remove tests unrelated to migration then? I think that way, we can see how this PR affects migration process clearly.

beeender: I think so. As long as we covered the corner case related with `RealmObjectSchema.xxxPrimaryKey`, is good enough then IMO.

realm/realm-library/src/androidTest/java/io/realm/RealmMigrationTests.java
stk1m1: Yes. `configFactory.createConfigurationBuilder().build()` gives all schema including `entities.migration` package.

realm/realm-library/src/androidTest/java/io/realm/RealmMigrationTests.java
stk1m1: I'll remove the line.

realm/realm-library/src/androidTest/java/io/realm/entities/AllJavaTypes.java
stk1m1: I'll remove the line and use `RealmObjectSchema.getPrimaryKey()`. 

At this point, I think there probably is a merit to remove all tests unrelated to migration in this PR though.

realm/realm-library/src/main/java/io/realm/RealmObjectSchema.java
stk1m1: It turns out when a PK is removed by `setPrimaryKey()`, `cachedPrimaryKeyColumnIndex` is already set to becomes `NO_PRIMARY_KEY`. 😉 

I've set to `NO_PRIMARY_KEY` in `removeColumn()` but it soon became clear that the cache handling should be separated.

stk1m1:
@realm/java reviews are addressed.

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ### Bug fixes
 
 * Fixed a crash when calling Table.toString() in debugger (#2429).
+* Fixed a bug when migration loses tracking of primary key field in `RealmObjectSchema.removeField()` and `RealmObjectSchema.renameField()` (#2829).
```
cmelchior: Fixed a bug that could cause Realm to loose track of the correct primary key when using `RealmObjectSchema.removeField()` and `RealmObjectSchema.renameField()`

realm/realm-library/src/androidTest/java/io/realm/RealmMigrationTests.java
```diff
@@ -226,6 +231,129 @@ public void migrate(DynamicRealm realm, long oldVersion, long newVersion) {
         }
     }
 
+    // this builds a temporary schema that does not exist in io.realm.entities
+    private void buildBaseMigrationSchema(final String schemaName, final boolean createBase) {
```
cmelchior: schemaName -> className

realm/realm-library/src/androidTest/java/io/realm/RealmMigrationTests.java
```diff
@@ -226,6 +231,129 @@ public void migrate(DynamicRealm realm, long oldVersion, long newVersion) {
...
+                        .addField(MigrationPrimaryKey.FIELD_PRIMARY, String.class, FieldAttribute.PRIMARY_KEY)
+                        .addField(MigrationPrimaryKey.FIELD_FOURTH,  Integer.class)
+                        .addField(MigrationPrimaryKey.FIELD_FIFTH,   Long.class);
+            }
+        });
+        realm.close();
+    }
+
+    @Test
+    public void renameTransferPrimaryKey() {
```
cmelchior: Add a small comment about this, e..g : "Renaming the Primary Key field should also rename the field in the pk metadata table"

realm/realm-library/src/androidTest/java/io/realm/entities/migration/MigrationSchemaRenamed.java
```diff
@@ -0,0 +1,78 @@
...
+ * limitations under the License.
+ */
+
+package io.realm.entities.migration;
+
+import io.realm.RealmObject;
+import io.realm.annotations.PrimaryKey;
+import io.realm.migration.MigrationPrimaryKey;
+
+public class MigrationSchemaRenamed extends RealmObject implements MigrationPrimaryKey {
```
cmelchior: MigrationSchemaRenamed -> MigrationClassRenamed

realm/realm-library/src/androidTest/java/io/realm/RealmMigrationTests.java
cmelchior: renameClassTransfersPrimaryKey

realm/realm-library/src/androidTest/java/io/realm/RealmMigrationTests.java
```diff
@@ -226,6 +231,129 @@ public void migrate(DynamicRealm realm, long oldVersion, long newVersion) {
...
+                .schemaVersion(1)
+                .schema(MigrationSchemaRenamed.class)
+                .migration(migration)
+                .build();
+        Realm realm = Realm.getInstance(realmConfig);
+        Table table = realm.getSchema().getTable(MigrationSchemaRenamed.class);
+
+        assertEquals(MigrationSchemaRenamed.DEFAULT_FIELDS_COUNT, table.getColumnCount());
+        assertTrue(table.hasPrimaryKey());
+        assertEquals(MigrationSchemaRenamed.DEFAULT_PRIMARY_INDEX, table.getPrimaryKey());
```
cmelchior: Also assert that the old primary key is gone.

realm/realm-library/src/androidTest/java/io/realm/RealmMigrationTests.java
```diff
@@ -226,6 +231,129 @@ public void migrate(DynamicRealm realm, long oldVersion, long newVersion) {
...
+        assertEquals(MigrationSchemaRenamed.FIELD_PRIMARY, table.getColumnName(table.getPrimaryKey()));
+    }
+
+    @Test
+    public void removeIndexesPriorToPrimaryKey() {
```
cmelchior: removeFieldsBeforePrimaryKeyField

realm/realm-library/src/main/java/io/realm/internal/Table.java
```diff
@@ -196,6 +196,11 @@ public long addColumnLink (RealmFieldType type, String name, Table table) {
     @Override
     public void removeColumn(long columnIndex) {
         nativeRemoveColumn(nativePtr, columnIndex);
+        // When PK does not exist this cannot happen. But if you remove a column with a smaller index
+        // than that of PK column, you need to rearrange PK column index.
+        if (hasPrimaryKey() && columnIndex < getPrimaryKey()) {
```
beeender: `columnIndex < getPrimaryKey()` -> ``columnIndex <= getPrimaryKey()`? If the column is the primary key, the cached index should be cleared.

realm/realm-library/src/androidTest/java/io/realm/RealmMigrationTests.java
```diff
@@ -226,6 +231,129 @@ public void migrate(DynamicRealm realm, long oldVersion, long newVersion) {
...
+        Table table = realm.getSchema().getTable(MigrationPosteriorIndexOnly.class);
+
+        assertEquals(MigrationPosteriorIndexOnly.DEFAULT_FIELDS_COUNT, table.getColumnCount());
+        assertTrue(table.hasPrimaryKey());
+        assertEquals(MigrationPosteriorIndexOnly.DEFAULT_PRIMARY_INDEX, table.getPrimaryKey());
+        assertEquals(MigrationPosteriorIndexOnly.FIELD_PRIMARY, table.getColumnName(table.getPrimaryKey()));
+    }
+
+    @Test
+    public void removeIndexesPosteriorToPrimaryKey() {
```
cmelchior: removeFieldsAfterPrimaryKeyField

realm/realm-library/src/androidTest/java/io/realm/entities/migration/MigrationFieldRenamed.java
```diff
@@ -0,0 +1,81 @@
...
+ * limitations under the License.
+ */
+
+package io.realm.entities.migration;
+
+import io.realm.RealmObject;
+import io.realm.annotations.PrimaryKey;
+import io.realm.migration.MigrationPrimaryKey;
+
+// This class has a renamed PrimaryKey field and same class name as that of original.
```
cmelchior: Add a reference to the original (i.e. what is the original called).

realm/realm-library/src/androidTest/java/io/realm/migration/MigrationPrimaryKey.java
```diff
@@ -0,0 +1,31 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.realm.migration;
+
+// This class has should be created dynamically in the first version of realm. This does not inherit
+// RealmObject
```
cmelchior: Also describe the purpose of the class.

realm/realm-library/src/main/java/io/realm/internal/Table.java
```diff
@@ -1025,6 +1030,39 @@ private Table getPrimaryKeyTable() {
...
     }
 
+    /**
+     * Rename the primary key column for the table. This needs to be called manually before inserting data into the table.
+     *
+     * @param newFieldName the name of the field replacing old primary key field name.
+     * @throws {@link IllegalArgumentException} if {@code newFieldName} is an empty string, or exceeds field name length limit.
+     * @throws {@link io.realm.exceptions.RealmException} if a primary key does not exist.
+     */
+    public void renamePrimaryKeyColumn(String newFieldName) {
```
cmelchior: Instead of having a separate method for this, why not check for this special case in `renameColumn` ? Right now this is a bit inconsistent with addColumn and removeColumn which do not have a special `*PrimaryKeyColumn` method (nor do they need one IMO)

realm/realm-library/src/main/java/io/realm/internal/Table.java
cmelchior: I would fold this method into `renameColumn`. That would make this exception not needed.

stk1m1:
@realm/java reviews points are applied. ;)

realm/realm-library/src/androidTest/java/io/realm/RealmMigrationTests.java
```diff
@@ -226,6 +233,147 @@ public void migrate(DynamicRealm realm, long oldVersion, long newVersion) {
         }
     }
 
+    /**
+     * Builds a temporary schema to be modified later in migration. {@link MigrationPrimaryKey} is
```
cmelchior: in a migration

realm/realm-library/src/androidTest/java/io/realm/RealmMigrationTests.java
```diff
@@ -226,6 +233,147 @@ public void migrate(DynamicRealm realm, long oldVersion, long newVersion) {
...
+     * the base class when specified.
+     *
+     * <p>MigrationPrimaryKey is supposed to be a RealmObject, but that would hamper our steps toward
+     * testing migration as Realm looks for it in migration. It is thus set to be an interface.
```
cmelchior: in migrations

realm/realm-library/src/androidTest/java/io/realm/entities/migration/MigrationPosteriorIndexOnly.java
```diff
@@ -0,0 +1,57 @@
...
+ * limitations under the License.
+ */
+
+package io.realm.entities.migration;
+
+import io.realm.RealmObject;
+import io.realm.annotations.PrimaryKey;
+import io.realm.migration.MigrationPrimaryKey;
+
+public class MigrationPosteriorIndexOnly extends RealmObject implements MigrationPrimaryKey {
```
cmelchior: Add a comment about this class, e.g "Class used for testing what happens if you modify fields defined after the primary key field"

realm/realm-library/src/androidTest/java/io/realm/entities/migration/MigrationPriorIndexOnly.java
```diff
@@ -0,0 +1,57 @@
...
+ * limitations under the License.
+ */
+
+package io.realm.entities.migration;
+
+import io.realm.RealmObject;
+import io.realm.annotations.PrimaryKey;
+import io.realm.migration.MigrationPrimaryKey;
+
+public class MigrationPriorIndexOnly extends RealmObject implements MigrationPrimaryKey {
```
cmelchior: "Class used for testing what happens if you modify fields defined before the primary key field"

realm/realm-library/src/main/java/io/realm/internal/Table.java
```diff
@@ -16,6 +16,8 @@
 
 package io.realm.internal;
 
+import android.text.TextUtils;
```
cmelchior: Please don't include android dependencies in the internal API. It will make it harder to migrate to the JVM.

realm/realm-library/src/main/java/io/realm/internal/Table.java
```diff
@@ -196,14 +198,35 @@ public long addColumnLink (RealmFieldType type, String name, Table table) {
     @Override
     public void removeColumn(long columnIndex) {
         nativeRemoveColumn(nativePtr, columnIndex);
+        // When PK does not exist this cannot happen. But if you remove a column with a smaller index
+        // than that of PK column, you need to rearrange PK column index.
```
cmelchior: you need to recalculate the PK column index as core could have changed its column index.

realm/realm-library/src/main/java/io/realm/internal/Table.java
```diff
@@ -196,14 +198,35 @@ public long addColumnLink (RealmFieldType type, String name, Table table) {
...
+        if (hasPrimaryKey() && columnIndex <= getPrimaryKey()) {
+            invalidateCachedPrimaryKeyIndex();
+        }
     }
 
     /**
-     * Renames a column in the table.
+     * Renames a column in the table. This needs to be called manually before inserting data into the table.
```
cmelchior: Not sure I understand "This needs to be called manually before inserting data into the table." ?

realm/realm-library/src/main/java/io/realm/internal/Table.java
```diff
@@ -196,14 +198,35 @@ public long addColumnLink (RealmFieldType type, String name, Table table) {
...
+     * @param newName a new name replacing the old column name.
+     * @throws {@link IllegalArgumentException} if {@code newFieldName} is an empty string, or exceeds field name length limit.
      */
     @Override
     public void renameColumn(long columnIndex, String newName) {
         verifyColumnName(newName);
+        // a primary key renamed
+        if (getPrimaryKey() == columnIndex) {
+            if (TextUtils.isEmpty(newName)) {
+                throw new IllegalArgumentException("Primary key field cannot be renamed with an empty string.");
```
cmelchior: You cannot rename anything to the empty string? Not just PK fields.

realm/realm-library/src/main/java/io/realm/internal/Table.java
```diff
@@ -196,14 +198,35 @@ public long addColumnLink (RealmFieldType type, String name, Table table) {
...
+     *
+     * @param columnIndex the column index to be renamed.
+     * @param newName a new name replacing the old column name.
+     * @throws {@link IllegalArgumentException} if {@code newFieldName} is an empty string, or exceeds field name length limit.
      */
     @Override
     public void renameColumn(long columnIndex, String newName) {
         verifyColumnName(newName);
+        // a primary key renamed
+        if (getPrimaryKey() == columnIndex) {
```
cmelchior: Add this after the native call. Right now if the native call throws an exception, you can in some corner cases end up with a wrong pk-table.

stk1m1: That is a great great point! 👍

cmelchior:
Starting to look good. We need to add more safety checks when renaming the column though as it should behave as an atomic commit. You don't know if people might catch exceptions somewhere instead of canceling the transaction.

realm/realm-library/src/main/java/io/realm/internal/Table.java
stk1m1: I've modified `verifyColumnName()` itself to check an empty or long column name. It will affect
`addColumn()`, `addColumnLink()`, and `renameColumn()`.

realm/realm-library/src/main/java/io/realm/internal/Table.java
stk1m1: In stead of `try/finally`, which could indiscriminately roll back any changes, `Table`'s column name gets reverted in case of error on renaming the PK meta table.

stk1m1:
@realm/java review points are addressed.

stk1m1:
retest this please

realm/realm-library/src/main/java/io/realm/internal/Table.java
```diff
@@ -191,20 +191,82 @@ public long addColumnLink (RealmFieldType type, String name, Table table) {
...
+        // by old name.
+        final long oldPkColumnIndex = getPrimaryKey();
+
+        // then let's try to rename a column. If an error occurs for some reasons, we'll throw.
         nativeRenameColumn(nativePtr, columnIndex, newName);
+
+        // Rename a primary key. At this point, renaming the column name should have been fine.
+        if (oldPkColumnIndex == columnIndex) {
+            try {
+                String className = tableNameToClassName(getName());
```
beeender: use `newName` instead of `getName()` to avoid a JNI call.

realm/realm-library/src/main/java/io/realm/internal/Table.java
```diff
@@ -191,20 +191,82 @@ public long addColumnLink (RealmFieldType type, String name, Table table) {
...
+                Table pkTable = getPrimaryKeyTable();
+                long pkRowIndex = pkTable.findFirstString(PRIMARY_KEY_CLASS_COLUMN_INDEX, className);
+                if (pkRowIndex != NO_MATCH) {
+                    pkTable.setString(PRIMARY_KEY_FIELD_COLUMN_INDEX, pkRowIndex, newName);
+                } else {
+                    throw new IllegalStateException("Non-existent PrimaryKey column cannot be renamed");
+                }
+            } catch (Exception e) {
+                // we failed to rename the pk meta table. roll back the column name, not pk meta table
```
beeender: Is this necessary? when it throws, transaction should be canceled. The roolback doesn't have to be handled here.

stk1m1: You're exactly right on the point. When PK meta table gets rolled back due to a failure ( or whatever error happens to get the `pkTable.setString()`), we have an inconsistency between the meta table and actual column which will result in non-existent pk. 

So, [the idea](https://github.com/realm/realm-java/pull/2920#discussion_r66075994) is to synchronize the column name between the table and the meta table by reverting back the column name when an error occurs.

beeender: I think we don't have to do that? In that case, user only has two options:

1. Catch the exception and cancel the transaction which will rollback everything
2. Crash the process, and since the transaction is not committed, the next time db gets opened, it will stay at the version without those changes.

@cmelchior Do you have any special reason we need to this in this function? Otherwise, it seems we have to do reverting in every single function which might change the meta table.

beeender:
Except the comment on the exception part, :+1:

realm/realm-library/src/main/java/io/realm/internal/Table.java
cmelchior: If a user is catching exceptions inside a transaction and ignore it, you risk corrupting the PK table. Which is kinda critical. I agree it is a extremely unlikely use case, but we should still guard against it I think. Also in other codepaths touching it.

The easy solution would be to catch it and just convert any exception to Error, but that also seem kinda heavy.

beeender: OK. I am fine with this. :+1:

cmelchior:
Looks good to me 👍

PR #2926, opened 2016-06-02
RealmResults is not synced in global listener

### LINKED ISSUE #2408: ArrayIndexOutOfBoundsException: rowIndex is less than 0
## Goal (what do you want to achieve?)

I am changing status of a field in realm table continuously  

## Expected Results

The status should change in the database

## Actual Results (e.g. full stack trace with exception)

```
03-10 17:27:50.122 28586-28586/com.workoutguru.workoutcash E/AndroidRuntime: FATAL EXCEPTION: main
03-10 17:27:50.122 28586-28586/com.workoutguru.workoutcash E/AndroidRuntime: Process: com.workoutguru.workoutcash, PID: 28586
03-10 17:27:50.122 28586-28586/com.workoutguru.workoutcash E/AndroidRuntime: java.lang.ArrayIndexOutOfBoundsException: rowIndex is less than 0.
03-10 17:27:50.122 28586-28586/com.workoutguru.workoutcash E/AndroidRuntime:     at io.realm.internal.TableView.nativeGetSourceRowIndex(Native Method)
03-10 17:27:50.122 28586-28586/com.workoutguru.workoutcash E/AndroidRuntime:     at io.realm.internal.TableView.getSourceRowIndex(TableView.java:158)
03-10 17:27:50.122 28586-28586/com.workoutguru.workoutcash E/AndroidRuntime:     at io.realm.RealmResults.get(RealmResults.java:174)
03-10 17:27:50.122 28586-28586/com.workoutguru.workoutcash E/AndroidRuntime:     at com.workoutguru.workoutcash.CompanyGoals_AllGoals_Fragment$CompanyAllGoalsAdapter$CompanyAllGoalsAdapterViewHolder$1$1.onResponse(CompanyGoals_AllGoals_Fragment.java:668)
03-10 17:27:50.122 28586-28586/com.workoutguru.workoutcash E/AndroidRuntime:     at retrofit.ExecutorCallAdapterFactory$ExecutorCallback$1.run(ExecutorCallAdapterFactory.java:86)
03-10 17:27:50.122 28586-28586/com.workoutguru.workoutcash E/AndroidRuntime:     at android.os.Handler.handleCallback(Handler.java:739)
03-10 17:27:50.122 28586-28586/com.workoutguru.workoutcash E/AndroidRuntime:     at android.os.Handler.dispatchMessage(Handler.java:95)
03-10 17:27:50.122 28586-28586/com.workoutguru.workoutcash E/AndroidRuntime:     at android.os.Looper.loop(Looper.java:135)
03-10 17:27:50.122 28586-28586/com.workoutguru.workoutcash E/AndroidRuntime:     at android.app.ActivityThread.main(ActivityThread.java:5343)
03-10 17:27:50.122 28586-28586/com.workoutguru.workoutcash E/AndroidRuntime:     at java.lang.reflect.Method.invoke(Native Method)
03-10 17:27:50.122 28586-28586/com.workoutguru.workoutcash E/AndroidRuntime:     at java.lang.reflect.Method.invoke(Method.java:372)
03-10 17:27:50.122 28586-28586/com.workoutguru.workoutcash E/AndroidRuntime:     at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:905)
03-10 17:27:50.122 28586-28586/com.workoutguru.workoutcash E/AndroidRuntime:     at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:700)
```

## Version of Realm and tooling
Realm version(s): compile 'io.realm:realm-android:0.87.5'

Android Studio version:  android 1.4

Android version(s) on device/simulator: Device version 5.1,API:22

beeender:
Close #2408

RealmResults are synced when calling its listener, since we need to
check the table version before calling the listener. So sync it just
after advance read won't be an option - in that way, the result's
listener won't be triggered.

So we notify the global listeners as the last thing to do, at that
point, result will be synced already.

Also a test case is added to ensure the calling sequence of synced
listeners.

### DIFF
--- a/realm/realm-library/src/androidTest/java/io/realm/NotificationsTest.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/NotificationsTest.java
@@ -255,7 +255,8 @@ public class NotificationsTest {
         final AtomicBoolean isRealmOpen = new AtomicBoolean(true);
         final Map<Integer, Integer> results = new ConcurrentHashMap<Integer, Integer>();
         final Looper[] looper = new Looper[1];
-        final RealmChangeListener<Realm> listener[] = new RealmChangeListener[1];
+        //noinspection unchecked
+        final RealmChangeListener<Realm>[] listener = new RealmChangeListener[1];
 
         ExecutorService executorService = Executors.newSingleThreadExecutor();
         Future<Boolean> future = executorService.submit(new Callable<Boolean>() {
@@ -1043,5 +1044,124 @@ public class NotificationsTest {
                 looperThread.testComplete();
             }
         });
+    }
+
+    @Test
+    @RunTestInLooperThread
+    public void realmListener_realmResultShouldBeSynced() {
+        final AtomicBoolean changedOnce = new AtomicBoolean(false);
+        final Realm realm = looperThread.realm;
+        realm.executeTransaction(new Realm.Transaction() {
+            @Override
+            public void execute(Realm realm) {
+                realm.createObject(AllTypes.class);
+            }
+        });
+        final RealmResults<AllTypes> results = realm.where(AllTypes.class).findAll();
+        assertEquals(1, results.size());
+
+        realm.addChangeListener(new RealmChangeListener<Realm>() {
+            @Override
+            public void onChange(Realm element) {
+                if (!changedOnce.get()) {
+                    // Change event triggered by populating
+                    assertEquals(1, realm.where(AllTypes.class).count());
+                    assertEquals(1, results.size());
+
+                    realm.executeTransactionAsync(new Realm.Transaction() {
+                        @Override
+                        public void execute(Realm realm) {
+                            AllTypes allTypes = realm.where(AllTypes.class).findFirst();
+                            assertNotNull(allTypes);
+                            allTypes.deleteFromRealm();
+                            final RealmResults<AllTypes> results = realm.where(AllTypes.class).findAll();
+                            assertEquals(0, results.size());
+                        }
+                    });
+                    changedOnce.set(true);
+                } else {
+                    // Change event triggered by deletion in async transaction.
+                    assertEquals(0, realm.where(AllTypes.class).count());
+                    assertEquals(0, results.size());
+                    realm.close();
+                    looperThread.testComplete();
+                }
+            }
+        });
+    }
+
+    // We precisely depend on the order of triggering change listeners right now.
+    // So it should be:
+    // 1. Synced object listener
+    // 2. Synced results listener
+    // 3. Global listener
+    // If this case fails on your code, think more before changing the test!
+    // https://github.com/realm/realm-java/issues/2408 is related with this test!
+    @Test
+    @RunTestInLooperThread
+    public void callingOrdersOfListeners() {
+        final Realm realm = looperThread.realm;
+        final AtomicInteger count = new AtomicInteger(0);
+
+        final RealmChangeListener<RealmResults<AllTypes>> syncedResultsListener =
+                new RealmChangeListener<RealmResults<AllTypes>>() {
+                    @Override
+                    public void onChange(RealmResults<AllTypes> element) {
+                        // First called
+                        assertEquals(0, count.getAndAdd(1));
+                    }
+                };
+
+        final RealmChangeListener<AllTypes> syncedObjectListener = new RealmChangeListener<AllTypes>() {
+            @Override
+            public void onChange(AllTypes element) {
+                // Second called
+                assertEquals(1, count.getAndAdd(1));
+            }
+        };
+        final RealmChangeListener<Realm> globalListener = new RealmChangeListener<Realm>() {
+            @Override
+            public void onChange(Realm element) {
+                // third called
+                assertEquals(2, count.getAndAdd(1));
+                looperThread.testComplete();
+            }
+        };
+
+
+        realm.beginTransaction();
+        final AllTypes allTypes = realm.createObject(AllTypes.class);
+        realm.commitTransaction();
+
+        // We need to create one objects first and let the pass the first change event
+        final RealmChangeListener<Realm> initListener = new RealmChangeListener<Realm>() {
+            @Override
+            public void onChange(Realm element) {
+                looperThread.postRunnable(new Runnable() {
+                    @Override
+                    public void run() {
+                        // Clear the change listeners
+                        realm.removeAllChangeListeners();
+
+                        // Now we can start testing
+                        allTypes.addChangeListener(syncedObjectListener);
+                        RealmResults<AllTypes> results = realm.where(AllTypes.class).findAll();
+                        results.addChangeListener(syncedResultsListener);
+                        realm.addChangeListener(globalListener);
+
+                        // Now we trigger those listeners
+                        realm.executeTransactionAsync(new Realm.Transaction() {
+                            @Override
+                            public void execute(Realm realm) {
+                                AllTypes allTypes = realm.where(AllTypes.class).findFirst();
+                                assertNotNull(allTypes);
+                                allTypes.setColumnLong(42);
+                            }
+                        });
+                    }
+                });
+            }
+        };
+        realm.addChangeListener(initListener);
     }
 }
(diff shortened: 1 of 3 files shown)


realm/realm-library/src/main/java/io/realm/HandlerController.java
```diff
@@ -268,6 +267,12 @@ void notifyAllListeners() {
         if (!realm.isClosed() && threadContainsAsyncEmptyRealmObject()) {
             updateAsyncEmptyRealmObject();
         }
+        // It is very important to notify the global listeners at last.
```
spl: ~~at~~

realm/realm-library/src/main/java/io/realm/HandlerController.java
```diff
@@ -268,6 +267,12 @@ void notifyAllListeners() {
...
+        // We don't sync RealmResults in realmChanged, instead, they are synced in notifySyncRealmResultsCallbacks.
+        // This is because of we need to compare the table view version to decide if it changes. Thus, we cannot sync
+        // the RealmResults together with advance read - it will fail the result's listener to be called.
```
spl: ~~fail the result's listener to be called~~ fail to call the results listener?

realm/realm-library/src/androidTest/java/io/realm/NotificationsTest.java
```diff
@@ -1044,4 +1045,123 @@ public void onChange(AllTypes object) {
...
+                    assertEquals(1, realm.where(AllTypes.class).count());
+                    assertEquals(1, results.size());
+
+                    realm.executeTransactionAsync(new Realm.Transaction() {
+                        @Override
+                        public void execute(Realm realm) {
+                            AllTypes allTypes = realm.where(AllTypes.class).findFirst();
+                            assertNotNull(allTypes);
+                            allTypes.deleteFromRealm();
+                            final RealmResults<AllTypes> results = realm.where(AllTypes.class).findAll();
```
cmelchior: Perhaps just `realm.where(AllTypes.class).count()` instead?

realm/realm-library/src/androidTest/java/io/realm/NotificationsTest.java
```diff
@@ -1044,4 +1045,123 @@ public void onChange(AllTypes object) {
             }
         });
     }
+
+    @Test
+    @RunTestInLooperThread
+    public void realmListener_realmResultShouldBeSynced() {
+        final AtomicBoolean changedOnce = new AtomicBoolean(false);
```
cmelchior: Since you are actually using two states. It would probably read more easily if you used an AtomicInteger instead and did switches on the number

realm/realm-library/src/androidTest/java/io/realm/NotificationsTest.java
```diff
@@ -1044,4 +1045,123 @@ public void onChange(AllTypes object) {
...
+                    realm.close();
+                    looperThread.testComplete();
+                }
+            }
+        });
+    }
+
+    // We precisely depend on the order of triggering change listeners right now.
+    // So it should be:
+    // 1. Synced object listener
```
cmelchior: What about the async listeners?

realm/realm-library/src/androidTest/java/io/realm/NotificationsTest.java
```diff
@@ -1044,4 +1045,123 @@ public void onChange(AllTypes object) {
...
+    // 2. Synced results listener
+    // 3. Global listener
+    // If this case fails on your code, think more before changing the test!
```
cmelchior: If this test case fails in your code. Think twice before changing the test.

realm/realm-library/src/androidTest/java/io/realm/NotificationsTest.java
```diff
@@ -1044,4 +1045,123 @@ public void onChange(AllTypes object) {
...
+    // https://github.com/realm/realm-java/issues/2408 is related with this test!
```
cmelchior: is related to this test.

realm/realm-library/src/main/java/io/realm/HandlerController.java
```diff
@@ -268,6 +267,12 @@ void notifyAllListeners() {
         if (!realm.isClosed() && threadContainsAsyncEmptyRealmObject()) {
             updateAsyncEmptyRealmObject();
         }
+        // It is very important to notify the global listeners last.
+        // We don't sync RealmResults in realmChanged, instead, they are synced in notifySyncRealmResultsCallbacks.
+        // This is because of we need to compare the table view version to decide if it changes. Thus, we cannot sync
```
cmelchior: This is because we need to compare the TableView version in order to decide if it changed.

realm/realm-library/src/androidTest/java/io/realm/NotificationsTest.java
beeender: async listeners will swallow `REALM_CHANGE` events and triggered by other events `COMPLETED_ASYNC_REALM_RESULTS`/`COMPLETED_ASYNC_REALM_OBJECT`. They are not really interacting with global listeners.

cmelchior:
Good catch 💯 . Only had minor comments 👍 
I am slightly concerned about changing the order of callbacks though. I can see why there is no way around it, I am just concerned if we might break other apps in the process that accidentally relied on that order. On the other hand, we have never promised which order they where called in, and the most important thing is that they are consistent (which they will be after this). Thoughts?

realm/realm-library/src/androidTest/java/io/realm/NotificationsTest.java
beeender: "Think twice"!! that is the phrase! I did scratch my hair to recall it.

realm/realm-library/src/androidTest/java/io/realm/NotificationsTest.java
cmelchior: I know, my comment was mostly that they where missing from your 1.2.3 list. So maybe just add a comment about that?

beeender:
I think we should not promise the order of they get called ever... User's code logic should never rely on it. Even we might have a consistent order from now on, but that could be changed in the future.

We can doc something like the orders of calling those listeners are not guaranteed. So, please don't rely on it? or just don't doc it which user should know it is a "undocumented" behaviour.

cmelchior:
I would probably be fine with just calling it out in the changelog, because as you say. We never promised any ordering.

realm/realm-library/src/androidTest/java/io/realm/NotificationsTest.java
```diff
@@ -1044,4 +1045,128 @@ public void onChange(AllTypes object) {
...
+                realm.createObject(AllTypes.class);
+            }
+        });
+        final RealmResults<AllTypes> results = realm.where(AllTypes.class).findAll();
+        assertEquals(1, results.size());
+
+        realm.addChangeListener(new RealmChangeListener<Realm>() {
+            @Override
+            public void onChange(Realm element) {
+                switch (changeCounter.getAndAdd(1)) {
```
nhachicha: minor: similar to `getAndIncrement`

realm/realm-library/src/androidTest/java/io/realm/NotificationsTest.java
```diff
@@ -1044,4 +1045,128 @@ public void onChange(AllTypes object) {
...
+                        assertEquals(1, realm.where(AllTypes.class).count());
+                        assertEquals(1, results.size());
+
+                        realm.executeTransactionAsync(new Realm.Transaction() {
+                            @Override
+                            public void execute(Realm realm) {
+                                AllTypes allTypes = realm.where(AllTypes.class).findFirst();
+                                assertNotNull(allTypes);
+                                allTypes.deleteFromRealm();
+                                assertEquals(0, realm.where(AllTypes.class).count());
```
nhachicha: will this assert failure be reported inside the `RunTestInLooperThread`

nhachicha:
👍  minor things

realm/realm-library/src/androidTest/java/io/realm/NotificationsTest.java
beeender: Yes!

Issue #2901, opened 2016-05-30
ThrowingException 16, std::exception in io_realm_internal_TableQuery.cpp line 1288


hohnamkung:
After  `ThrowingException 16, std::exception in io_realm_internal_TableQuery.cpp line 1288` exception is logged, i can't
1. load data using Observable (Observable' s onNext not called, only `ThrowingException 16, std::exception in io_realm_internal_TableQuery.cpp line 1288` is logged in logcat)
2. load data synchronously (`Requested table is not in this Realm. Creating it requires a transaction` exception happens)
3. insert data. Whenever i insert data, `A RealmObject with no @PrimaryKey cannot be updated` error happened, although that class has a `@PrimaryKey` annotated field.
4. finally initialize Realm with `ArrayIndexOutOfBoundsException: columnIndex > available columns.`exception

**Case 1~3 dismiss after process is restarted (swipe it in recent tasks), but i could reproduce below steps. After case 4, i uninstalled application and no longer reproduce any of problems.**
#### Steps & Code to Reproduce

The class that have this problem query data and is listening changes (using observable) in main thread. At the same time every data about that class are removed (`realm.delete(*.class)`) and re-inserted with fresh one from server in worker thread.

I tried to find reproduce step for 1~2 hours, but i couldn't. 
#### Version of Realm and tooling

Realm version(s): 1.0.0

Android Studio version: 2.1.1

Which Android version and device: 6.0.1 (Galaxy S6)

stk1m1:
Hello @hohnamkung 

Thanks for the detailed description of the symptom. Is there anyway we could ask a simple sample to reproduce the problem you are going through? We could be more than happy to accelerate solving the issue if there is a tangible specimen. We will strongly look forward.

hohnamkung:
Already i made a effort to find a way to reproduce it, but i couldn't find any. 😞 
After case 1~3, i thought i can reproduce it so i uninstall application without backup realm file.

I'm not sure i can find a step to reproduce it or make a sample application.
But i and my team members have this issues time to time, so i'll share a realm file if i run into it.

hohnamkung:
Already i made a effort to find a way to reproduce it, but i couldn't find any. 😞 
After case 1~3, i thought i can reproduce it so i uninstall application without backup realm file.

I'm not sure i can find a step to reproduce it or make a sample application.
But i and my team members have this issues time to time, so i'll share a realm file if i run into it.

stk1m1:
@hohnamkung 

[...]

PR #2943, opened 2016-06-06
Listeners can call un-synchronized RealmResults (Async queries)

cmelchior:
When async queries are updated, they call listeners in a specific order. This currently has a very subtle bug, so accessing synchronous RealmResults from inside a async RealmResult listener might hit a detached row accessor.

Reason being is that the Realm is advanced, then all listeners called in order, but RealmResults are not synchronised until just before listener is called. This was discovered when investigating #2901, but I doubt it is the root cause.


cmelchior:
Wrong branch :/

PR #2951, opened 2016-06-07
Listeners can call un-synchronized RealmResults (Async queries)

cmelchior:
When async queries are updated, they call listeners in a specific order. This currently has a very subtle bug, so accessing synchronous RealmResults from inside a async RealmResult listener might hit a detached row accessor.

Reason being is that the Realm is advanced, then all listeners called in order, but RealmResults are not synchronised until just before listener is called. This was discovered when investigating #2901, but I doubt it is the root cause.


cmelchior:
Ready for review @realm/java 
I still don't fully understand https://github.com/realm/realm-java/pull/2951/files#diff-c6f79e9457082a73f91697a0d6ed683dR285 though. Perhaps @nhachicha has a good explanation?

realm/realm-library/src/main/java/io/realm/HandlerController.java
```diff
@@ -257,58 +261,66 @@ private void updateAsyncEmptyRealmObject() {
...
+     *
+     * @param realmResultsToBeNotified list of all RealmResults listeners that can be notified.
+     */
+    void notifyAllListeners(List<RealmResults<? extends RealmModel>> realmResultsToBeNotified) {
 
-        // empty async RealmObject shouldn't block the realm to advance
-        // they're empty so no risk for running into a corrupt state
-        // where the pointer (Row) is using one version of a Realm, whereas the
-        // current Realm is advancing to a newer version (they're empty anyway)
+        // Notify all Realmresults (async and synchronous).
```
kneth: "Realmresults" -> "RealmResults"

realm/realm-library/src/main/java/io/realm/HandlerController.java
```diff
@@ -462,7 +478,9 @@ private void completedAsyncQueriesUpdate(QueryUpdateTask.Result result) {
         SharedGroup.VersionID callerVersionID = realm.sharedGroupManager.getVersion();
         int compare = callerVersionID.compareTo(result.versionID);
         if (compare > 0) {
-            // if the caller thread is advanced i.e it already sent a REALM_CHANGE that will update the queries
+            // if the caller thread is more advanced than the worker thread, it means it did a local commit.
```
kneth: "if" -> "If"

realm/realm-library/src/main/java/io/realm/RealmResults.java
```diff
@@ -615,6 +615,14 @@ public boolean deleteLastFromRealm() {
         }
     }
 
+    /**
+     * Syncs this RealmResults, so it is update to date after `advance_read` has been called.
```
kneth: "update" -> "up"

realm/realm-library/src/main/java/io/realm/HandlerController.java
```diff
@@ -257,58 +261,66 @@ private void updateAsyncEmptyRealmObject() {
...
+     * @param realmResultsToBeNotified list of all RealmResults listeners that can be notified.
+     */
+    void notifyAllListeners(List<RealmResults<? extends RealmModel>> realmResultsToBeNotified) {
 
-        // empty async RealmObject shouldn't block the realm to advance
-        // they're empty so no risk for running into a corrupt state
-        // where the pointer (Row) is using one version of a Realm, whereas the
-        // current Realm is advancing to a newer version (they're empty anyway)
+        // Notify all Realmresults (async and synchronous).
+        for (Iterator<RealmResults<? extends RealmModel>> it = realmResultsToBeNotified.iterator(); it.hasNext() && !realm.isClosed(); ) {
```
emanuelez: The HandlerController is associated to a Realm instance, which means that it's local to a thread. So why do we need to keep checking if the realm has been closed?

cmelchior: I'm guessing because you can close the Realm inside a change listener, but I haven't really investigated it. Perhaps @nhachicha / @beeender knows?

beeender: Yes, to prevent the realm gets closed in the `onChange` from last iteration.

realm/realm-library/src/main/java/io/realm/HandlerController.java
```diff
@@ -412,7 +427,8 @@ private void completedAsyncRealmResults(QueryUpdateTask.Result result) {
                         // swap pointer
                         realmResults.swapTableViewPointer(result.updatedTableViews.get(weakRealmResults));
                         // notify callbacks
-                        realmResults.notifyChangeListeners();
+                        realmResults.syncIfNeeded();
```
beeender: Is this necessary? `callerVersionID.compareTo(result.versionID)` kind of ensure that they have the same version? @nhachicha

realm/realm-library/src/main/java/io/realm/HandlerController.java
```diff
@@ -499,25 +519,17 @@ private void completedAsyncQueriesUpdate(QueryUpdateTask.Result result) {
                 } else {
                     // update the instance with the new pointer
                     realmResults.swapTableViewPointer(query.getValue());
-
-                    // it's dangerous to notify the callback about new results before updating
-                    // the pointers, because the callback may use another RealmResults not updated yet
-                    // this is why we defer the notification until we're done updating all pointers
-                    callbacksToNotify.add(realmResults);
+                    realmResults.syncIfNeeded();
+                    resultsToBeNotified.add(realmResults);
```
beeender: I will double check this. It feels it will be added to the list in the following `collectSyncRealmResultsCallbacks(resultsToBeNotified);`

beeender:
Done the first round check... will check it again tomorrow

realm/realm-library/src/main/java/io/realm/HandlerController.java
beeender: OK, the following is notify sync results. :+1:

realm/realm-library/src/main/java/io/realm/HandlerController.java
kneth: Would it make sense to write `!realm.isClosed() && it.hasNext()` instead? I mean, is `realm` is closed, it sounds safer to terminate the loop.

realm/realm-library/src/main/java/io/realm/HandlerController.java
cmelchior: Yes, that call also update the cached version inside RealmResults. This is needed to detect if the listeners should be triggered at all.

realm/realm-library/src/main/java/io/realm/HandlerController.java
```diff
@@ -257,58 +261,66 @@ private void updateAsyncEmptyRealmObject() {
...
             WeakReference<RealmResults<? extends RealmModel>> weakRealmResults = iterator.next();
             RealmResults<? extends RealmModel> realmResults = weakRealmResults.get();
             if (realmResults == null) {
                 iterator.remove();
             } else {
-                // It should be legal to modify asyncRealmResults and syncRealmResults in the listener
+                // Sync the RealmResult so it is completely up to date.
+                // This is a prerequisite to calling the listener, so when the listener is finally triggered, all
+                // RealmResults will be up to date.
+                realmResults.syncIfNeeded();
```
stk1m1: Would it be possible to separate this step after collecting all listeners especially in `realmChanged()` that it would look like             
```java
collectAsyncRealmResultsCallbacks(resultsToBeNotified);
collectSyncRealmResultsCallbacks(resultsToBeNotified); 
// sync if needed
enforceSyncCollectedRealmResultsCallbaks(resultsToBeNotified);
```

It might take another loop, but separating this step might help others not to forget and reuse.

cmelchior: Good idea, but the problem is that the `collectAsyncRealmResultsCallbacks(resultsToBeNotified);` is only used in one place, in the other location it is replaced by a completely different implementation that also swaps pointers. The the amount of re-use is practically non-existent ( i tried :) )

cmelchior:
All issues addressed @realm/java

beeender:
:+1:

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 * Fixed a crash when calling Table.toString() in debugger (#2429).
 * Fixed a crash that RealmResults was not updated in Realm's change listener by adjusting the calling orders of listeners on Realm, RealmObject and RealmResults (#2926).
+* Fixed a potential crash when accessing other RealmResults from inside a RealmResults' RealmChangeListener (#2951).
```
beeender: Maybe merge these two items. And after this commit, the order of listeners doesn't matter too much anymore.

realm/realm-library/src/androidTest/java/io/realm/NotificationsTest.java
```diff
@@ -946,20 +946,20 @@ public void run() {
     public void asyncRealmObjectShouldNotBlockBackgroundCommitNotification() {
         final AtomicInteger numberOfRealmCallbackInvocation = new AtomicInteger(0);
         final CountDownLatch signalClosedRealm = new CountDownLatch(1);
-        looperThread.realm.addChangeListener(new RealmChangeListener<Realm>() {
+        final Realm realm = looperThread.realm;
+        realm.addChangeListener(new RealmChangeListener<Realm>() {
             @Override
             public void onChange(Realm object) {
```
kneth: Can we find a better name? I mean, when you read quickly, you can easily read it as a `RealmObject`.

realm/realm-library/src/main/java/io/realm/HandlerController.java
```diff
@@ -257,58 +261,66 @@ private void updateAsyncEmptyRealmObject() {
...
         }
-
-        for (Iterator<RealmResults<? extends RealmModel>> it = resultsToBeNotified.iterator(); it.hasNext() && !realm.isClosed(); ) {
-            RealmResults<? extends RealmModel> realmResults = it.next();
-            realmResults.notifyChangeListeners();
-        }
     }
 
+    /**
+     * INVARIANT: Should only be called from {@link #notifyAllListeners(List)}.
```
kneth: Nitpick: Is this really an invariant? Maybe it's just a NOTE?

CHANGELOG.md
```diff
@@ -3,8 +3,8 @@
 ### Bug fixes
 
 * Fixed a crash when calling Table.toString() in debugger (#2429).
-* Fixed a bug that could cause Realm to lose track of primary key when using `RealmObjectSchema.removeField()` and `RealmObjectSchema.renameField()` (#2829).
-* Fixed a crash that RealmResults was not updated in Realm's change listener by adjusting the calling orders of listeners on Realm, RealmObject and RealmResults (#2926).
+* Fixed a race condition which would cause some RealmResults to not be properly updated inside a RealmChangeListener. This could result in crashes when accessing items from those results (#2926/#2951).
```
kneth: Backquote RealmResults and RealmChangeListener.

kneth:
:+1: with a few pedantic comments.

stk1m1:
👍

PR #2952, opened 2016-06-07
Race condition between Realm change notifications and UI events

cmelchior:
Fixes https://github.com/realm/realm-android-adapters/issues/11
Fixes https://github.com/realm/realm-android-adapters/issues/48

This PR fixes a bug surfaced by our Realm Android Adapter example. Before this PR we posted all REALM_CHANGED events to the looper queue as normal. This meant that potential touch input events could get in front which in turn could cause a relayout of a ListView which would then crash because it detected a difference between it's cached adapter size and the actual one.

This PR now forces all commits to be placed at the front of the queue eliminating that risk.

I am not sure it is possible to test this from `realm-java` (hints appreciated if it is), but the bug fix was verified from the realm-android-adapters side.

Using `postAtFrontOfQueue` is highly discouraged by Google since you risk starving the UI thread and introduce lag. However commits are generally fairly infrequent and any loop (commits inside a change listener) will now be more easily detected. Before this, they would probably work, although making the UI sluggish. 

@realm/java-admin 

--- 
EDIT: This PR should have been against `releases`. I'll let the review finish here before moving it 😢


beeender:
The problem of this approach is `handler.hasMessages(HandlerController.REALM_CHANGED)` could return true when the other thread posted a `REALM_CHANGED` event before. In that case, the event it not at the front of the queue.

That is the reason i propose to add another event `REALM_RESULTS_CHAGED` to be posted at the font of current thread's handler only, and only sync the `RealmResults` without advance read.

Thoughts for better ideas?

cmelchior:
Not sure I follow? There is a Handler for each Looper thread, so If another thread posted to the UI handler, that event would have to be at the front of the queue, no?

beeender:
> If another thread posted to the UI handler, that event would have to be at the front of the queue, no?

If another thread post to UI handler, it can be at the front of the queue if the queue at that time is empty or it also can be at the end of the queue if there are already some `redraw` event existing in the queue.

```
+                } else {
+                    messageSent = handler.sendEmptyMessage(HandlerController.REALM_CHANGED);
+                }
```

The else case (not current thread) doesn't post at the front. But if it does, i can imagine it will be a disaster for the UI queue as google's doc said :(

beeender:
uh, my bad, it is checking the `MainLooper`. Let me check it again

beeender:
If all other threads try to post at the front of UI event queue, this will freeze a UI a lot.
Also it is not generic enough to solve the similar problems for other looper threads :(

so with this, a loop of transaction in a non-UI thread will freeze UI forever which is a non-necessary side effect.

cmelchior:
We could just post to the front of the queue on all Handlers. It would also simplify the logic.

A loop of transactions on a background thread would only freeze the UI thread if they spanned multiple event windows, e.g. as soon as the first event is on the queue, no more events would be added. So a freeze would only happen if you could say, let a background thread make a new commit every 10th millisecond. If that is happening you are most likely doing something wrong anyway? 

I do agree though that the performance implications of this is not entirely clearcut, and that it could get easier to starve the UI thread. So that aspect is probably worth investigating.

Another simple solution would be to completely disallow writes on the UI thread 💃

cmelchior:
After internal discussion with @beeender, this needs further work.

We are afraid that sending all REALM_CHANGED events to the front of the queue can risk event starvation in some hard to predict cases.

Instead the following is a lot less risky:

1) When doing a commit on a Looper thread, send REALM_CHANGED to the front of the local queue and REALM_CHANGED to the back of all other. threads
2) This means we only risk starving the UI thread if you are doing commits unconditionally inside a RealmChangeListener which is a bug anyway (cannot come up with a sound use case at least).
3) For 1.1 we introduce a `allowWritesOnUIThread(bool)` on the RealmConfiguration. That will allow people to catch all these cases. Doing writes on the UI thread is most likely something you want to avoid anyway.

cmelchior:
Ready for review again @realm/java @beeender

realm/realm-library/src/main/java/io/realm/HandlerController.java
```diff
@@ -112,6 +114,7 @@ public boolean handleMessage(Message message) {
             QueryUpdateTask.Result result;
             switch (message.what) {
 
+                case LOCAL_COMMIT:
```
beeender:
```
    private void realmChanged() {
        deleteWeakReferences();
        if (threadContainsAsyncQueries()) {
            updateAsyncQueries();

        } else {
            RealmLog.d("REALM_CHANGED realm:" + HandlerController.this + " no async queries, advance_read");
            realm.sharedGroupManager.advanceRead();
            notifyAllListeners();
        }
    }
```

`if (threadContainsAsyncQueries()) {` is not covered ... @#$!@#$@!$# This hurts my brain too much !!!

beeender:
See the test cases I added. The last two won't pass.

The other fundamental problem is, if user make transaction in the thread which has an async query submitted before, if the query is done already, we cannot swallow the change event and rerun the query in background thread anymore. It this case, the loaded async `RealmResults` has to be updated in THIS thread. Otherwise we risk a invalid object in the results.

cmelchior:
@beender Yes, you are right. After playing a bit with this I came to the same conclusion. It is impossible to guarantee a completely consistent view if you mix synchronous commits and async queries. The problem is already loaded async results which we cannot leave "alone" without risking detached rows.

I still think we should implement the LOCAL_CHANGE event change (so synchronous commits are pushed to the front of the queue), but apart from that we detect if you mix async queries + local commits and log a warning in the log. We already have all the metadata for this, so should be trivial.

We could add an opt in `allowWritesOnUIThread()` so people can opt into a more aggresive behaviour, where UI writes will crash the app as well. This will for now be `true` but we set it to `false` from 2.0 onwards. 

What do you think? Also @realm/java ?

beeender:
| Things need to do | REALM_CHANGED | LOCAL_COMMIT |
| ---               |  ---          |  ---         |
| Re-query finished async query | YES | NO, just sync them |
| Re-query unfinished async query | YES | YES |
| Sync synced `RealmResults` | YES | YES |
| advance read | YES | NOT necessary. But it doesn't seem to be harmful either. |
| Notifify global listeners | YES | YES |
| Notifify typed listeners | YES | YES |


I tried to list all the thing we need to do when event comes. It seems the only different is about how to handle the finished async queries. Also, I totally agree `LOCAL_COMMIT` should exist for those purposes.

realm/realm-library/src/androidTest/java/io/realm/NotificationsTest.java
```diff
@@ -1169,4 +1166,204 @@ public void execute(Realm realm) {
...
+    // Step 2: Post a runnable to caller thread.
+    //         Event Queue: |Posted Runnable| <- TOP
+    // Step 3: Delete object which will make the results contain a invalid object at this moment
+    //         Right Event Queue: |LOCAL_COMMIT   |   Wrong Event Queue: |Posted Runnable           |  <- TOP
+    //                            |Posted Runnable|                      |REALM_CHANGED/LOCAL_COMMIT|
+    // Step 4: Posted runnable called.
+    @Test
+    @RunTestInLooperThread(/*step1*/PopulateOneAllTypes.class)
+    public void realmListener_localChangeShouldBeSendAtFrontOfTheQueue() {
+        final AtomicInteger changeCounter = new AtomicInteger(0);
```
stk1m1: `changeCounter` is not used.

cmelchior:
As the PR is growing in scope. I think it is better to make it properly against `releases`. Closing this and moving relevant discussion.

PR #2961, opened 2016-06-08
gcc is overly paranoid

kneth:
@beeender


beeender:
Then we can enable `-Werror` as well?

kneth:
Yes, that will make everyone more aware of warnings :-)

beeender:
:+1:

PR #2979, opened 2016-06-10
Add a Dockerfile to the repo

emanuelez:
This is done to facilitate CI and build reproducibility

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,46 @@
+FROM ubuntu:16.04
+
+# Locales
+RUN locale-gen en_US.UTF-8
+ENV LANG "en_US.UTF-8"
+ENV LANGUAGE "en_US.UTF-8"
+ENV LC_ALL "en_US.UTF-8"
+
+# Set the environment variables
+ENV JAVA_HOME /usr/lib/jvm/java-8-openjdk-amd64
+ENV ANDROID_HOME /opt/android-sdk-linux
+ENV NDK_HOME /opt/android-ndk
+ENV PATH ${PATH}:${ANDROID_HOME}/tools:${ANDROID_HOME}/platform-tools
+ENV PATH ${PATH}:${NDK_HOME}
+
+# Install the JDK
+# We are going to need some 32 bit binaries because aapt requires it
+# file is need by the script that creates NDK toolchains
+RUN DEBIAN_FRONTEND=noninteractive dpkg --add-architecture i386 \
+    && apt-get update -qq \
+    && apt-get install -y file git curl wget zip unzip \
+                       build-essential \
+                       openjdk-8-jdk-headless \
+                       libc6:i386 libstdc++6:i386 libgcc1:i386 libncurses5:i386 libz1:i386 \
+    && apt-get clean
+
+# Install the Android SDK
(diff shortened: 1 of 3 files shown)


emanuelez:
retest this please

emanuelez:
@realm/java @teotwaki @alebsack

Dockerfile
```diff
@@ -0,0 +1,46 @@
...
+RUN echo y | android update sdk --no-ui --all --filter build-tools-23.0.0 | grep 'package installed'
+
+# Install the NDK
+RUN mkdir /opt/android-ndk-tmp
+RUN cd /opt/android-ndk-tmp && wget -q http://dl.google.com/android/ndk/android-ndk-r10e-linux-x86_64.bin -O android-ndk.bin
+RUN cd /opt/android-ndk-tmp && chmod a+x ./android-ndk.bin && ./android-ndk.bin
+RUN cd /opt/android-ndk-tmp && mv ./android-ndk-r10e /opt/android-ndk
+RUN rm -rf /opt/android-ndk-tmp
+
+ADD * /opt/realm-java/
```
alebsack: Why not use a VOLUME statement for /opt/realm-java and map it during `docker run`?  This should cut down on docker build cache misses and produce faster builds.

Jenkinsfile
```diff
@@ -1,5 +1,92 @@
...
+            stage 'Docker build'
+            def buildEnv = docker.build 'realm-java:snapshot'
+            buildEnv.inside("--privileged -v /dev/bus/usb:/dev/bus/usb -v ${env.HOME}/gradle-cache:/root/.gradle") {
+                stage 'JVM tests'
+                try {
+                    gradle 'assemble check javadoc'
+                } finally {
+                    storeJunitResults 'realm/realm-annotations-processor/build/test-results/TEST-*.xml'
+                }
+                if (env.BRANCH_NAME == 'master') {
```
cmelchior: Perhaps move this to the bottom alongside the OJO upload. No need to collect metrics if it fails some of the quality checks?

cmelchior:
I don't have any knowledge about Docker to give qualified feedback on this. From the little I did understand it seems OK though.

Dockerfile
emanuelez: Well spotted! Actually it's not even needed at all! Jenkins does the mounting and setting of the working directory automatically. Shaving a few seconds off! :)

dalinaum:
Dockerfile seems good.

I was trying to find `docker run` or `docker build`, but there are not such commands. Because I don't know about Jenkins, I don't know that a docker image would be distributed, reused or rebuilt each test. How will it work?

emanuelez:
@dalinaum the docker is built on every run, but it will most likely be already cached on the Jenkins slave, so the operation will take only a couple of seconds. then Jenkins runs the image mounting the workarea.
This means that now we have all the information on how to build the binding in our repo, which has great advantages. If, for example, somebody makes a PR that changes the NDK version, he will not need to ask anybody to install the new NDK on every Jenkins slave: he will simply have to modify the Dockerfile accordingly.

beeender:
Looks good although I have no clue about docker :P

PR #2983, opened 2016-06-11
Updated beginTransaction doc

cmelchior:
We recently had a number of questions around refreshing and modifying data.
I extended the `beginTransaction` Javadoc to be very explicit about what happens and what the best practise is around this.

@realm/java

### DIFF
--- a/realm/realm-library/src/main/java/io/realm/BaseRealm.java
+++ b/realm/realm-library/src/main/java/io/realm/BaseRealm.java
@@ -308,8 +308,8 @@ abstract class BaseRealm implements Closeable {
      * {@link io.realm.Realm#cancelTransaction()}. Transactions are used to atomically create, update and delete objects
      * within a Realm.
      * <p>
-     * Writing to a Realm always happen on the latest version of the data. So before beginning a transaction, this
-     * instance of Realm is updated to the latest version that include changes from all threads. This update does not
+     * Writing to a Realm always happens on the latest version of the data. So before beginning a transaction, this
+     * instance of Realm is updated to the latest version that includes changes from all threads. This update does not
      * trigger any registered {@link RealmChangeListener}.
      * <p>
      * For that reason, it is best practise to query for any data that should be modified from inside the
@@ -323,7 +323,7 @@ abstract class BaseRealm implements Closeable {
      * persons.first().setName("John");
      * realm.commitTransaction;
      *
-     * // But this
+     * // Do this instead
      * realm.beginTransaction();
      * RealResults<Person> persons = realm.where(Person.class).findAll();
      * persons.first().setName("John");
--- a/realm/realm-library/src/main/java/io/realm/BaseRealm.java
+++ b/realm/realm-library/src/main/java/io/realm/BaseRealm.java
@@ -308,11 +308,11 @@ abstract class BaseRealm implements Closeable {
      * {@link io.realm.Realm#cancelTransaction()}. Transactions are used to atomically create, update and delete objects
      * within a Realm.
      * <p>
-     * Before beginning a transaction the Realm instance is updated to the latest version, which included all the
-     * changes from the other threads. This update does not trigger any registered {@link RealmChangeListener}.
+     * Before beginning a transaction, the Realm instance is updated to the latest version in order to include all
+     * changes from other threads. This update does not trigger any registered {@link RealmChangeListener}.
      * <p>
-     * It is therefore recommended to query for the data that should be modified from inside the transaction, otherwise
-     * there is a risk that data have been deleted or modified when the transaction begins.
+     * It is therefore recommended to query for the items that should be modified from inside the transaction. Otherwise
+     * there is a risk that some of the results have been deleted or modified when the transaction begins.
      * <p>
      * <pre>
      * {@code
--- a/realm/realm-library/src/main/java/io/realm/BaseRealm.java
+++ b/realm/realm-library/src/main/java/io/realm/BaseRealm.java
@@ -308,12 +308,11 @@ abstract class BaseRealm implements Closeable {
      * {@link io.realm.Realm#cancelTransaction()}. Transactions are used to atomically create, update and delete objects
      * within a Realm.
      * <p>
-     * Writing to a Realm always happens on the latest version of the data. So before beginning a transaction, this
-     * instance of Realm is updated to the latest version that includes changes from all threads. This update does not
-     * trigger any registered {@link RealmChangeListener}.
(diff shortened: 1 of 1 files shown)


realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -308,8 +308,28 @@ public void onCall() {
      * {@link io.realm.Realm#cancelTransaction()}. Transactions are used to atomically create, update and delete objects
      * within a Realm.
      * <p>
-     * Before beginning the transaction, {@link io.realm.Realm#beginTransaction()} updates the Realm in the case of
-     * pending updates from other threads.
+     * Writing to a Realm always happen on the latest version of the data. So before beginning a transaction, this
```
stk1m1: `always happens on`

stk1m1:
👍

realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -308,8 +308,28 @@ public void onCall() {
...
+     * instance of Realm is updated to the latest version that include changes from all threads. This update does not
```
zaki50: that include -> that includes

realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -308,8 +308,28 @@ public void onCall() {
...
+     * <p>
+     * <pre>
+     * {@code
+     * // Don't do this
+     * RealResults<Person> persons = realm.where(Person.class).findAll();
+     * realm.beginTransaction();
+     * persons.first().setName("John");
+     * realm.commitTransaction;
+     *
+     * // But this
```
spl: ~~But this~~ Do this instead.

spl:
The example is great! It seems like it would be a good candidate for the [tutorial](https://realm.io/docs/java/latest/), too.

cmelchior:
@spl I am not sure there is a good fit for it in the tutorial since we are not modifying any queried data there at this point. However, we should update the documentation to reflect it though, e.g. in https://realm.io/docs/java/latest/#transaction-blocks

spl:
What about the [Iterations](https://realm.io/docs/java/latest/#iterations) and [Deletion](https://realm.io/docs/java/latest/#deletion) sections? Perhaps you could change the code in those sections and add a note in Deletions about querying within the transaction?

cmelchior:
Yes, those sections should be updated as well.

realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -308,8 +308,28 @@ public void onCall() {
...
      * within a Realm.
      * <p>
-     * Before beginning the transaction, {@link io.realm.Realm#beginTransaction()} updates the Realm in the case of
-     * pending updates from other threads.
+     * Writing to a Realm always happens on the latest version of the data. So before beginning a transaction, this
+     * instance of Realm is updated to the latest version that includes changes from all threads. This update does not
+     * trigger any registered {@link RealmChangeListener}.
+     * <p>
+     * For that reason, it is best practise to query for any data that should be modified from inside the
+     * transaction, otherwise there is a risk those data have been deleted when the transaction begins.
```
kneth: "deleted" -> "deleted or modified"

radu-tutueanu:
@cmelchior https://ci.realm.io/job/android-pr-jenkinsfile/500/console looks stale and it is keeping other builds from running. Is it ok if I kill it?

realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -308,8 +308,28 @@ public void onCall() {
      * {@link io.realm.Realm#cancelTransaction()}. Transactions are used to atomically create, update and delete objects
      * within a Realm.
      * <p>
-     * Before beginning the transaction, {@link io.realm.Realm#beginTransaction()} updates the Realm in the case of
-     * pending updates from other threads.
+     * Writing to a Realm always happens on the latest version of the data. So before beginning a transaction, this
```
emanuelez: Before beginning a transaction the Realm instance is updated to the latest version, which included all the changes from the other threads.

realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -308,8 +308,28 @@ public void onCall() {
...
+     * instance of Realm is updated to the latest version that includes changes from all threads. This update does not
+     * trigger any registered {@link RealmChangeListener}.
+     * <p>
+     * For that reason, it is best practise to query for any data that should be modified from inside the
```
emanuelez: It is therefore recommended to query for data from within the transaction that is going to modify it in order to avoid the risk of data mismatch.

emanuelez:
I left a couple of suggestions for improvements, but I think a native english speaker should chime in. @ironage do you have a second? :)

realm/realm-library/src/main/java/io/realm/BaseRealm.java
kneth: Can we assume that the term "version" is known at this point? Or should we add a paragraph above: "Realm is a multi-versioned concurrent data store, and threads might have different versions of the Realm instance at any time."

realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -308,8 +308,27 @@ public void onCall() {
...
      * within a Realm.
      * <p>
-     * Before beginning the transaction, {@link io.realm.Realm#beginTransaction()} updates the Realm in the case of
-     * pending updates from other threads.
+     * Before beginning a transaction the Realm instance is updated to the latest version, which included all the
+     * changes from the other threads. This update does not trigger any registered {@link RealmChangeListener}.
+     * <p>
+     * It is therefore recommended to query for the data that should be modified from inside the transaction, otherwise
+     * there is a risk that data have been deleted or modified when the transaction begins.
+     * <p>
```
ironage: Just to be picky, although you are correct that "data" is plural, "data have" reads awkward to me and makes me read it twice. Here is my suggestion: 
    * Before beginning a transaction, the Realm instance is updated to the latest version in order to include all
    * changes from other threads. This update does not trigger any registered {@link RealmChangeListener}.
    * <p>
    * It is therefore recommended to query for the items that should be modified from inside the transaction. Otherwise
    * there is a risk that some of the results have been deleted or modified when the transaction begins.

Zhuinden:
You wrote `RealResults` instead of `RealmResults` by accident

beeender:
:+1:

Issue #1684, opened 2015-10-27
Bulk insert


beeender:
This is feature requested by http://stackoverflow.com/questions/33358880/realm-java-fastest-bulk-write-methodology

Most SQL database support "batch insert" like `INSERT INTO tbl (col1, col2) VALUES ('val1', 'val2'), ('val3', 'val4'), ...`

Our current bottle neck would be when inserting lots of data, lots of duplicated checking are made, and lots or jni calls are made. We should find a way to pass a batch of different types of data to JNI and insert them through one JNI call.

cmelchior:
One of the huge performance penalties is also the time spend by the Garbage collector, because we allocate a lot of temporary objects. Some sort of object-pool pattern where users could "give back" an object would go a long way and could easily be added to our current API::

```
Foo foo = realm.createObject(Foo.class);
foo.setBar("bar");
for (int i = 1; i < 100000; i++ ) {
  foo = realm.createObject(Foo.class, foo); // Reuse object -> Manual cleanup of row-accessor, reuse all Java objects
  foo.setBar("newBar");
}
```

beeender:
+1 from http://stackoverflow.com/questions/33358880/realm-java-fastest-bulk-write-methodology?noredirect=1#comment54535556_33358880

cmelchior:
Another suggestion: Our 3 primary performance hogs are: GC, Thread-check and JNI boundary.

I think that we can address the first 2 relatively easy without big API changes, while the later is a lot more tricky as it means we should parse multiple arguments in one call.

If we could define a scope where within we disabled the thread-checks for RealmObject accessors we would also gain a lot:

```
// V1: Boolean in begin transaction / or maybe just as part of beginTransaction()?
realm.beginTransaction(boolean iPromiseOnly1Thread)
// do stuff
realm.commitTransaction()
```

beeender:
I do prefer the idea before that we offer user a API to disable the thread checking globally, not only inside the transaction. As long as users know what they are doing, thread checking is not necessary in the release build.

cmelchior:
If we can find find a nice way to do it, then that might work as well. I just think that the use cases where reading so much data are very small compared to the write-use cases.

It would actually be fairly easy to add these optimisations to our current `copyToRealm(collection)/bulk JSON` methods as we have a very well-defined scope there.

cmelchior:
Copy from #987 

----

Creating a new object in a Realm is often done like:

realm.beginTransaction();
[...]

PR #2987, opened 2016-06-12
Free object on nativeBatchUpdateQueries after each iteration.

ShikaSD:
Currently nativeBatchUpdateQueries method fails with overflow of jni table,
as there are more than 500 items passed to JNI if realm needs to update frequently.
That is why we need to free the objects after they are used (= after each iteration).

### DIFF
--- a/realm/realm-jni/src/io_realm_internal_TableQuery.cpp
+++ b/realm/realm-jni/src/io_realm_internal_TableQuery.cpp
@@ -1183,17 +1183,17 @@ JNIEXPORT jlongArray JNICALL Java_io_realm_internal_TableQuery_nativeBatchUpdate
 
         // Step3: Run & export the queries against the latest shared group
         for (size_t i = 0; i < number_of_queries; ++i) {
-            JniLongArray query_param_array(env, (jlongArray) env->GetObjectArrayElement(query_param_matrix, i));
-            switch (query_param_array[0]) { // 0, index of the type of query, the next indicies are parameters
+            JniLongArray* query_param_array = new JniLongArray(env, (jlongArray) env->GetObjectArrayElement(query_param_matrix, i));
+            switch ((*query_param_array)[0]) { // 0, index of the type of query, the next indicies are parameters
                 case QUERY_TYPE_FIND_ALL: {// nativeFindAllWithHandover
                     exported_handover_tableview_array[i] =
                             findAllWithHandover
                                     (env,
                                      bgSharedGroupPtr,
                                      std::move(queries[i]),
-                                     query_param_array[1]/*start*/,
-                                     query_param_array[2]/*end*/,
-                                     query_param_array[3]/*limit*/);
+                                     (*query_param_array)[1]/*start*/,
+                                     (*query_param_array)[2]/*end*/,
+                                     (*query_param_array)[3]/*limit*/);
                     break;
                 }
                 case QUERY_TYPE_DISTINCT: {// nativeGetDistinctViewWithHandover
@@ -1202,7 +1202,7 @@ JNIEXPORT jlongArray JNICALL Java_io_realm_internal_TableQuery_nativeBatchUpdate
                                     (env,
                                      bgSharedGroupPtr,
                                      std::move(queries[i]),
-                                     query_param_array[1]/*columnIndex*/);
+                                     (*query_param_array)[1]/*columnIndex*/);
                     break;
                 }
                 case QUERY_TYPE_FIND_ALL_SORTED: {// nativeFindAllSortedWithHandover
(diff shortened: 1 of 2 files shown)


stk1m1:
@ShikaSD 

Thanks for your contribution. We are looking into batch insert #1684, and I'd have to ask your attention to two points that stand out at this point. 

1. Could you provide a use case that demonstrate the overflow in JNI table; an unit test that passes more than 500 items? Without a tangible ground, it is hard to establish a discussion to proceed.
2. I'm seeing `query_param_array` being  allocated in heap which could degrade the performance of all batch operations. Have you tried compare the performance?
3. Lastly, this isn't something I'd ask you to consider, but what you're touching is actually a sensitive part in multi-threaded Realm operations that I'd be extra careful about it. 

@kneth @cmelchior @beeender please review.

ShikaSD:
Sorry, I was a bit straight-forward with that issue.
The main problem there is that `HandlerController` is not releasing objects from its inner maps and there is overflow when updates are frequent. Unfortunately, I was not able to recreate crashes in unit testing, as in application it is happening from time to time when certain state is achieved.
I've looked through the `HandlerController` source and have found that you are not actually adding the `WeakReference` of the object to queue. As I know there should be call to `WeakReference.enqueue()` to do that. As a consequence, the maps (`asyncRealmResults` in `HandlerController#addToAsyncRealmResults` in my case) were overflowed with old ones and JNI is dropping it.
I'm not sure if I am right, but hope it will help to solve the **actual** issue.

realm/realm-jni/src/io_realm_internal_TableQuery.cpp
```diff
@@ -1183,17 +1183,17 @@ JNIEXPORT jlongArray JNICALL Java_io_realm_internal_TableQuery_nativeBatchUpdate
 
         // Step3: Run & export the queries against the latest shared group
         for (size_t i = 0; i < number_of_queries; ++i) {
-            JniLongArray query_param_array(env, (jlongArray) env->GetObjectArrayElement(query_param_matrix, i));
-            switch (query_param_array[0]) { // 0, index of the type of query, the next indicies are parameters
+            JniLongArray* query_param_array = new JniLongArray(env, (jlongArray) env->GetObjectArrayElement(query_param_matrix, i));
```
kneth: Allocating on the heap instead of the stack might be a good idea.

realm/realm-jni/src/util.hpp
```diff
@@ -578,6 +578,7 @@ class JniLongArray {
     ~JniLongArray()
     {
         m_env->ReleaseLongArrayElements(m_javaArray, m_array, m_releaseMode);
+        m_env->DeleteLocalRef(m_javaArray);
```
kneth: We should add similar to all our array classes in `util.hpp` to prevent memory leaks.

kneth:
@ShikaSD I can see that allocating on the heap might be better in cases with many elements. But please observe that we are talking about an array of an array of longs. It sounds like you have 500 arrays of smaller long arrays, and you are working on the smaller long arrays (3-6 longs). So I'm not sure if your patch will help you.

Said that, your patch brings in something important: deleting the local reference. You might have found a memory leak which can explain your issues.

As @stk1m1 points out, we would like to see a test which demonstrate the issue.

kneth:
@ShikaSD Thanks for your observation of `HandlerController`.

nhachicha:
Hi @ShikaSD 

we do release objects from the `HandlerController ` when calling `deleteWeakReferences` after a new commit. This will clear GC'ed reference from the differents maps
nice catch for the `DeleteLocalRef` 

Cheers

ShikaSD:
Hello @nhachicha,

Actually, you don't. I just was debugging the method and found out that the reference queue is always empty. You need to call `WeakReference#enqueue` after initialization to force item to be in the queue.

realm/realm-jni/src/io_realm_internal_TableQuery.cpp
beeender: Use `unique_ptr` to avoid deletion especially for the `return NULL` case. Like:
`std::unique_ptr<JniLongArray> query_param_array(new JniLongArray(env, (jlongArray) env->GetObjectArrayElement(query_param_matrix, i));`

nhachicha:
Hi @ShikaSD 
putting something (_the `Reference` whether it's Weak, Soft or Phantom_ ) to the `ReferenceQueue` means that the `Referent` (the instance you're observing the lifecycle for) is ready to be GC'ed.

If you call `enqueue` explicitly it's like telling I know my `Referent` is GC'ed. Since we don't know when an instance/Referent goes out of scope to become eligible for GC, we rely on the GC to put our `Reference` into the `ReferenceQueue` as explained in the Java doc

```java
/**
     * ... when the garbage collector
     * enqueues references it does so directly, without invoking this method.
     *...
     */
    public boolean enqueue() {
        return this.queue.enqueue(this);
    }
```

when debugging try to invoke `System.gc()` to hint to the GC to start collecting references in the queue. Btw we have a unit test for this [here](https://github.com/realm/realm-java/blob/fe65795f0873beaead47358a6a529a20b2c4ab44/realm/realm-library/src/androidTest/java/io/realm/RealmAsyncQueryTests.java#L1380)

Cheers,

ShikaSD:
Hello @nhachicha,

Thanks for your explanation. The actual problem was on my side, as I was using `RealmResults.asObservable` inside some flatMap and it was generating new observables during updates. As you are keeping strong references to that observables in the factory, new ones were not GC'ed until unsubscribing(I have it in onDestroy).
That is why so many queries were kept in the memory and JNI table overflow happened.
Sorry for that confusion, I suppose I will change the code towards realm sync api.

realm/realm-jni/src/io_realm_internal_TableQuery.cpp
```diff
@@ -1239,6 +1240,7 @@ JNIEXPORT jlongArray JNICALL Java_io_realm_internal_TableQuery_nativeBatchUpdate
                     ThrowException(env, FatalError, "Unknown type of query.");
                     return NULL;
             }
+            delete query_param_array;
```
kneth: A `unique_ptr` will be deleted when it goes out of scope.

PR #2990, opened 2016-06-13
Race condition between Realm change notifications and UI events

cmelchior:
Fixes realm/realm-android-adapters#11
Fixes realm/realm-android-adapters#48

From #2952 

This PR fixes a bug surfaced by our Realm Android Adapter example. Before this PR we posted all REALM_CHANGED events to the looper queue as normal. This meant that potential touch input events could get in front which in turn could cause a relayout of a ListView which would then crash because it detected a difference between it's cached adapter size and the actual one.

This PR now forces all local commits to be placed at the front of the queue eliminating that risk.

Using postAtFrontOfQueue is highly discouraged by Google since you risk starving the UI thread and introduce lag. However commits are generally fairly infrequent and any loop (commits inside a change listener) will now be more easily detected. Before this, they would probably work, although making the UI sluggish.

Implementation is as follows:

| Things need to do | REALM_CHANGED | LOCAL_COMMIT |
| ---               |  ---          |  ---         |
| Finished async query | Swallow the event and schedule a requery | advance_read/sync them on the UI thread |
| Unfinished async query | Swallow the event and schedule a requery | Ignore |
| Sync synced `RealmResults` | YES | YES |
| advance read | YES | YES |
| Notify global listeners | YES | YES |
| Notify typed listeners | YES | YES (only loaded RealmResults/Objects) |

Implementation comment:
Ideally on LOCAL_COMMIT we want to stop any non-loaded queries as they will fail when reaching the Looper and will need to retry anyway. For async queries that are already loaded, we don't need to reschedule as they will be synced on the Looper thread. However currently it is not possible to make that distinction, and I would rather keep this out of scope for this PR, as it would be a bigger change for little benefit.

For that reason I am instead just letting any un-loaded queries run to completion including retry.


realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -359,13 +360,37 @@ void commitTransaction(boolean notifyLocalThread, Runnable runAfterCommit) {
...
+                    // prioritize local commits.
+                    //
+                    // If a user is doing commits inside a RealmChangeListener this can cause the Looper thread to get
+                    // event starved as it only starts handling Realm events instead. This is an acceptable risk as
+                    // that behaviour indicate a user bug. Previously this would be hidden as the UI would still
+                    // be responsive.
+                    Message msg = Message.obtain();
+                    msg.what = HandlerController.LOCAL_COMMIT;
+                    if (!handler.hasMessages(HandlerController.LOCAL_COMMIT)) {
+                        messageHandled = handler.sendMessageAtFrontOfQueue(msg);
```
beeender: It would be unharmful to do a `handler.removeMessage(HandlerController.REALM_CHANGED);` here. But it is quite minor. I am fine without it.

beeender:
:+1: looks good! and i cannot imagine another error case by now :P

realm/realm-library/src/androidTest/java/io/realm/NotificationsTest.java
```diff
@@ -938,9 +939,10 @@ public void run() {
         TestHelper.exitOrThrow(executorService, signalTestFinished, signalClosedRealm, backgroundLooper, threadAssertionError);
     }
 
-    // The presence of async RealmResults block any `REALM_CHANGE` notification causing historically the Realm
-    // to advance to the latest version. We make sure in this test that all Realm listeners will be notified
-    // regardless of the presence of an async RealmObject that will delay the `REALM_CHANGE` sometimes
+    // The presence of async RealmResults block any `REALM_CHANGE` notification . We make sure in this test that all
```
kneth: "block" -> "blocks"

realm/realm-library/src/androidTest/java/io/realm/NotificationsTest.java
```diff
@@ -938,9 +939,10 @@ public void run() {
...
+    // Realm listeners will be notified regardless of the presence of an async RealmObject. RealmObjects are special
+    // in the sense that once you got an row accessor to that object, it is automatically up to date as soon as you
```
kneth: "an" -> "a"

realm/realm-library/src/androidTest/java/io/realm/NotificationsTest.java
```diff
@@ -938,9 +939,10 @@ public void run() {
...
+    // advance_read.
```
kneth: "call advance_read()`?

realm/realm-library/src/androidTest/java/io/realm/NotificationsTest.java
```diff
@@ -1170,6 +1168,174 @@ public void execute(Realm realm) {
         realm.addChangeListener(initListener);
     }
 
+    // See https://github.com/realm/realm-android-adapters/issues/48
+    // Step 1: Populate the db
+    // Step 2: Post a runnable to caller thread.
+    //         Event Queue: |Posted Runnable| <- TOP
+    // Step 3: Delete object which will make the results contain a invalid object at this moment
```
kneth: "a" -> "an"

realm/realm-library/src/main/java/io/realm/BaseRealm.java
kneth: Since it goes against the recommendations, we could write a message to the log. It could ease debugging later.

cmelchior: I would much prefer not flooding the log with this, but we could add a debug log message just like we do for a lot of the other async callbacks. That would leave them disabled by default, but could be enabled if we suspected it.

stk1m1:
I can only say it makes lot of sense in unit test, but I'd say we could appreciate @nhachicha's review on this. LGTM 👍

realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -359,13 +360,38 @@ void commitTransaction(boolean notifyLocalThread, Runnable runAfterCommit) {
...
-            if (
-                    realmPath.equals(configuration.getPath())                           // It's the right realm
-                            && !handler.hasMessages(HandlerController.REALM_CHANGED)    // The right message
-                            && handler.getLooper().getThread().isAlive()                // The receiving thread is alive
-                            && !handler.sendEmptyMessage(HandlerController.REALM_CHANGED)) {
-                RealmLog.w("Cannot update Looper threads when the Looper has quit. Use realm.setAutoRefresh(false) " +
-                        "to prevent this.");
+            Looper looper = handler.getLooper();
+            if (realmPath.equals(configuration.getPath())          // It's the right realm
+                            && looper.getThread().isAlive()) {     // The receiving thread is alive
```
bmunkholm: wrong indent.

realm/realm-library/src/androidTest/java/io/realm/NotificationsTest.java
```diff
@@ -1170,6 +1168,173 @@ public void execute(Realm realm) {
...
+    // Step 3: Post a runnable to caller thread.
+    //         Event Queue: |Posted Runnable| <- TOP
+    // Step 4: Delete object which will make the results contain a invalid object at this moment
+    //         Right Event Queue: |LOCAL_COMMIT   |   Wrong Event Queue: |Posted Runnable           |  <- TOP
+    //                            |Posted Runnable|                      |REALM_CHANGED/LOCAL_COMMIT|
+    // Step 5: Posted runnable called.
+    //
+    @Test
+    @RunTestInLooperThread(/*step1*/PopulateOneAllTypes.class)
+    public void realmListener_localChangeShouldBeSendAtFrontOfTheQueueWithPausedAsync() {
```
nhachicha: How's this test different from the previous one?
I don't see where you pause the async query, same for the previous one what do you mean by LoadedAsync? is it calling `load`?

nhachicha:
LGTM small comment 👍

realm/realm-library/src/androidTest/java/io/realm/NotificationsTest.java
cmelchior: `Realm.asyncTaskExecutor.pause();` is pausing it.
The difference is that the first test loads the async query first, while the second one pauses the query so we hit two different codepaths for async RealmResults. The difference is step 2, but yes, I agree it is slightly hidden.

PR #2995, opened 2016-06-14
Wrong JNI function declaration caused test timeout

beeender:
The worst thing is async query won't throw those exceptions in the
background.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 * Fixed a crash when calling `Table.toString()` in debugger (#2429).
 * Fixed a race condition which would cause some `RealmResults` to not be properly updated inside a `RealmChangeListener`. This could result in crashes when accessing items from those results (#2926/#2951).
 * Fixed a bug that could cause Realm to lose track of primary key when using `RealmObjectSchema.removeField()` and `RealmObjectSchema.renameField()` (#2829).lts (#2926).
+* Fixed a bug that some async related JNI functions cannot be found on some devices due to wrong declaration.
 * Updated ProGuard configuration in order not to depend on Android's default configuration (#2972).
 
 ### Enhancements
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 * Fixed a crash when calling `Table.toString()` in debugger (#2429).
 * Fixed a race condition which would cause some `RealmResults` to not be properly updated inside a `RealmChangeListener`. This could result in crashes when accessing items from those results (#2926/#2951).
 * Fixed a bug that could cause Realm to lose track of primary key when using `RealmObjectSchema.removeField()` and `RealmObjectSchema.renameField()` (#2829).lts (#2926).
+* Fixed a bug that some async related JNI functions cannot be found on some devices due to wrong declaration.
```
cmelchior: Fixed a bug that prevented some devices from finding async related JNI methods correctly.

cmelchior:
Not sure I fully understand this error. What devices did this happen on? Also why is it only the async API's that needs to change. We are using `jobject` all over the place?

beeender:
The problem is `nativeFindAllMultiSortedWithHandover` is defined as a `static` method in java, in C++ code, the corresponding function should take a `jclass` as param not `jobject`.

This was found during the cmake testing on android emulator nexus 6. Not sure it is because of cmake changes some compile options or other reason.

Because of the wrong function declaration, java cannot find the jni function at runtime. A "method cannot be found" exception will be thrown at the runtime. And since we have something like:

```
final Future<Long> pendingQuery = Realm.asyncTaskExecutor.submit(new Callable<Long>() {
    @Override
    public Long call() throws Exception {
        try {
            //...
            query.findAllWithHandover()
        } catch (Exception e) {
            RealmLog.e(e.getMessage(), e);
            closeSharedGroupAndSendMessageToHandler(sharedGroup,
                                weakHandler, HandlerController.REALM_ASYNC_BACKGROUND_EXCEPTION, new Error(e));
        }
    }
}
```

Which doesn't throw, this issue might be there but no body thinks it ever happens.

stk1m1:
👍

cmelchior:
So for cmake, all static methods should take a `jclass` instead of `jobject`? I guess that makes sense.

👍

beeender:
even for make, all static methods should take `jclass` and it should report this error on every device ... Maybe I will spend some time to investigate why it doesn't on most devices.

PR #2960, opened 2016-06-08
Move JNI build to cmake

beeender:
This is exciting. We can finally get rid of the jni project and generate the JNI header files on the fly.

AND THE NATIVE DEBUGGING IS WORKING IN THE AS!! (2.2 preview 3)

To use this:
* `ANDROID_NDK_HOME` needs to be set
* The ` com.android.tools.build:gradle:2.2.0-alpha2` requires java8

TODO:

- [x] Make all ABIs work, currently only x86
- [x] Make all compiling options right
- [x] Make core downloading work
- [x] Strip the release version so files
- [x] Copy the final results with debugs symbols out
- [x] Fix hardcoded path to `javah` in `CMakeList.txt`
- [x] Make CI work with cmake
- [x] ~~Run CI test with released version so files.~~ This is not necessary. Debugging version is even better since lots of checks are disabled in release.
- [x] Remove the `realm-jni` project
- [x] Fix the disabled findbugs task
- [x] Verify armeabi
- [x] Verify armeabi-v7a
- [x] Verify arm64
- [x] Verify x86
- [x] Verify x86_64
- [ ] Verify mips (BUT, how???)

PR #3010, opened 2016-06-16
JNI clean up

beeender:
* Enable -Wmissing-declarations and -Werror to ensure all global
  functions are defined with a proper declaration. JNI function not
  found problem can only be seen at run time, thus we need to do so.
* Add static keyword for local functions.
* Fix wrong JNI function declaration.
* Remove useless functions.


realm/realm-jni/src/io_realm_internal_Group.cpp
```diff
@@ -300,17 +275,6 @@ JNIEXPORT jstring JNICALL Java_io_realm_internal_Group_nativeToString(
     return 0;
 }
 
-JNIEXPORT jboolean JNICALL Java_io_realm_internal_Group_nativeEquals(
```
cmelchior: Is this useless? Seems more like `Group.equals()` should actually use it instead?

beeender: Do we ever need to call `Group.equals()`?

realm/realm-jni/src/io_realm_internal_table.cpp
```diff
@@ -1512,17 +1512,6 @@ JNIEXPORT jstring JNICALL Java_io_realm_internal_Table_nativeToJson(
     return NULL;
 }
 
-JNIEXPORT jboolean JNICALL Java_io_realm_internal_Table_nativeEquals(
```
cmelchior: Same, seems like Table should just implement equals properly.

realm/realm-jni/src/io_realm_internal_Group.cpp
cmelchior: We don't right now at least, and Group might go away with the object store, so yeah, ¯\\\_(ツ)\_/¯

cmelchior:
I would think a lot more methods needs to be static, but maybe they already are? Except from that 👍

realm/realm-jni/src/io_realm_internal_table.cpp
stk1m1: If this function is to compare pointers only, it would be fine to do so in Java as we have native pointers there.

beeender:
Yeah, created an issue to track the `static` problem https://github.com/realm/realm-java/issues/3011

stk1m1:
Could you also rename following two filenames according to the naming convention? It always bugs me.

- io_realm_internal_tableview.cpp
- io_realm_internal_table.cpp

Other than that, 👍

realm/realm-jni/src/util.cpp
```diff
@@ -185,21 +185,6 @@ jclass GetClass(JNIEnv* env, const char* classStr)
     return myClass;
 }
 
-void jprint(JNIEnv *env, char *txt)
```
bmunkholm: This has been useful for debugging in the past. But I assume we have another way now?

beeender: no idea, didn't spend time to check. No one is calling them so instead of making them `static`, I just removed them. They would be easily back by checking git history if needed.

stk1m1: I also saw this function was removed, but thought it was ok as we'd be able to put a breakpoint after @beeender's #2960 completed. Relying printf for JNI debugg has been pain in the neck, tbh.

kneth: We are the `TR` macros.

kneth: We are using the `TR` macros for logging. It will make it easier to support non-Android.

kneth:
:+1:

PR #3013, opened 2016-06-16
Optimistic opening of a Realm file

cmelchior:
This PR is the result of investigating https://github.com/realm/realm-java/issues/2459

We have anecdotal evidence that multiple processes might be open for the same app. We have not been able to verify this, but the result matches the "IncompatibleLockFile" errors reported in #2459. This happens if the two processes have two versions of Realm (e.g. during app upgrades).

This PR is based on our assumption that such an overlap is not intentional and only happens because one process is started before the previous was completely shut down.

So this PR introduces an optimistic opening scheme where we retry for 1 second before crashing as before.

I also added logging so users can log if they run into this issue.
The settings for this are currently public in SharedGroup, which will allow users to modify the timeout and retry-policy without us needing to release a new version of Realm Java. 

@realm/java

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
 * Updated ProGuard configuration in order not to depend on Android's default configuration (#2972).
 * Fixed a race condition between Realms notifications and other UI events. This could e.g. cause ListView to crash (#2990).
 * Fixed a bug that allowed both `RealmConfiguration.Builder.assetFile()`/`deleteRealmIfMigrationNeeded()` to be configured at the same time, which leads to the asset file accidentally being deleted in migrations (#2933).
-* In case of two processes opening the same Realm file, Realm will now optimistically retry opening for 1 second before throwing an Error. Previously it crashed outright (#2459).
+* Realm crashed outright when the same Realm file was opened in two processes. Realm will now optimistically retry opening for 1 second before throwing an Error (#2459).
 
 ### Enhancements
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 * Updated ProGuard configuration in order not to depend on Android's default configuration (#2972).
 * Fixed a race condition between Realms notifications and other UI events. This could e.g. cause ListView to crash (#2990).
 * Fixed a bug that allowed both `RealmConfiguration.Builder.assetFile()`/`deleteRealmIfMigrationNeeded()` to be configured at the same time, which leads to the asset file accidentally being deleted in migrations (#2933).
+* In case of two processes opening the same Realm file, Realm will now optimistically retry opening for 1 second before throwing an Error. Previously it crashed outright (#2459).
 
 ### Enhancements
 
--- a/realm/realm-jni/src/util.hpp
+++ b/realm/realm-jni/src/util.hpp
@@ -74,8 +74,7 @@ JNIEXPORT jint JNICALL JNI_OnLoad(JavaVM *vm, void *reserved);
                 std::string(e.what()) + " path: " + e.get_path()); \
     } \
     catch (realm::IncompatibleLockFile& e) { \
-        ThrowException(env, LockFileError, string(fileName), \
-                std::string(e.what())); \
+        ThrowException(env, LockFileError, std::string(e.what())); \
     } \
 
 
--- a/realm/realm-jni/src/util.hpp
+++ b/realm/realm-jni/src/util.hpp
@@ -72,7 +72,13 @@ JNIEXPORT jint JNICALL JNI_OnLoad(JavaVM *vm, void *reserved);
     catch (util::File::AccessError& e) { \
         ThrowException(env, FileAccessError, string(fileName), \
                 std::string(e.what()) + " path: " + e.get_path()); \
-    }
+    } \
+    catch (realm::IncompatibleLockFile& e) { \
+        ThrowException(env, LockFileError, string(fileName), \
+                std::string(e.what())); \
+    } \
+
+
 
 #define CATCH_STD() \
     catch (...) { \
@@ -125,7 +131,8 @@ enum ExceptionKind {
     RuntimeError = 12,
     RowInvalid = 13,
     CrossTableLink = 15,
-    BadVersion = 16
+    BadVersion = 16,
+    LockFileError = 17
 // NOTE!!!!: Please also add test cases to Util.java when introducing a new exception kind.
 };
 
--- a/realm/realm-library/src/androidTest/java/io/realm/RealmObjectTests.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/RealmObjectTests.java
@@ -1186,7 +1186,7 @@ public class RealmObjectTests {
     public void getter_afterDeleteFromOtherThreadThrows() {
         final CountDownLatch bgRealmDone = new CountDownLatch(1);
         realm.beginTransaction();
-        final AllJavaTypes obj = realm.createObject(AllJavaTypes.class);
+        final AllTypes obj = realm.createObject(AllTypes.class);
         realm.commitTransaction();
 
         new Thread(new Runnable() {
@@ -1194,7 +1194,7 @@ public class RealmObjectTests {
             public void run() {
                 Realm bgRealm = Realm.getInstance(realm.getConfiguration());
                 bgRealm.beginTransaction();
-                bgRealm.delete(AllJavaTypes.class);
+                bgRealm.delete(AllTypes.class);
                 bgRealm.commitTransaction();
                 bgRealm.close();
                 bgRealmDone.countDown();
@@ -1206,7 +1206,7 @@ public class RealmObjectTests {
         // Object should no longer be available
         assertFalse(obj.isValid());
         try {
-            obj.getFieldDate();
+            obj.getColumnLong();
             fail();
         } catch (IllegalStateException ignored) {
         }
--- a/realm/realm-library/src/androidTest/java/io/realm/RealmObjectTests.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/RealmObjectTests.java
@@ -40,6 +40,7 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 import java.util.concurrent.atomic.AtomicReference;
 
+import io.realm.entities.AllJavaTypes;
 import io.realm.entities.AllTypes;
 import io.realm.entities.AllTypesPrimaryKey;
 import io.realm.entities.ConflictingFieldName;
@@ -1185,7 +1186,7 @@ public class RealmObjectTests {
     public void getter_afterDeleteFromOtherThreadThrows() {
         final CountDownLatch bgRealmDone = new CountDownLatch(1);
         realm.beginTransaction();
-        final AllTypes obj = realm.createObject(AllTypes.class);
+        final AllJavaTypes obj = realm.createObject(AllJavaTypes.class);
         realm.commitTransaction();
 
         new Thread(new Runnable() {
@@ -1193,7 +1194,7 @@ public class RealmObjectTests {
             public void run() {
                 Realm bgRealm = Realm.getInstance(realm.getConfiguration());
                 bgRealm.beginTransaction();
-                bgRealm.delete(AllTypes.class);
+                bgRealm.delete(AllJavaTypes.class);
                 bgRealm.commitTransaction();
                 bgRealm.close();
                 bgRealmDone.countDown();
@@ -1205,7 +1206,7 @@ public class RealmObjectTests {
         // Object should no longer be available
         assertFalse(obj.isValid());
         try {
-            obj.getColumnLong();
+            obj.getFieldDate();
             fail();
         } catch (IllegalStateException ignored) {
         }
--- a/realm/realm-library/src/androidTest/java/io/realm/RealmObjectTests.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/RealmObjectTests.java
@@ -40,7 +40,6 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 import java.util.concurrent.atomic.AtomicReference;
 
-import io.realm.entities.AllJavaTypes;
 import io.realm.entities.AllTypes;
 import io.realm.entities.AllTypesPrimaryKey;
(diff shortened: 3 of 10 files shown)


realm/realm-library/src/main/java/io/realm/internal/SharedGroup.java
```diff
@@ -72,6 +81,52 @@ public SharedGroup(String canonicalPath, boolean enableImplicitTransactions, Dur
...
+                    RealmLog.d("Waiting for Realm to open interrupted: " + path);
+                }
+            }
+        }
+
+        throw new RealmError("Could not open the Realm file: " + lastError.getMessage());
+    }
+
+    // Returns the time to sleep before retrying opening the SharedGroup.
+    private long getSleepTime(int tries) {
```
emanuelez: It looks like this could be static

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 * Updated ProGuard configuration in order not to depend on Android's default configuration (#2972).
 * Fixed a race condition between Realms notifications and other UI events. This could e.g. cause ListView to crash (#2990).
 * Fixed a bug that allowed both `RealmConfiguration.Builder.assetFile()`/`deleteRealmIfMigrationNeeded()` to be configured at the same time, which leads to the asset file accidentally being deleted in migrations (#2933).
+* In case of two processes opening the same Realm file, Realm will now optimistically retry opening for 1 second before throwing an Error. Previously it crashed outright (#2459).
```
stk1m1: I would put what this PR fixes first.

* Fixed a bug that caused Realm to crash outright when the same Realm file is opened in two processes.  Realm will now optimistically retry opening for 1 second before throwing an Error (#2459).

realm/realm-jni/src/util.hpp
```diff
@@ -72,7 +72,12 @@ JNIEXPORT jint JNICALL JNI_OnLoad(JavaVM *vm, void *reserved);
     catch (util::File::AccessError& e) { \
         ThrowException(env, FileAccessError, string(fileName), \
                 std::string(e.what()) + " path: " + e.get_path()); \
-    }
+    } \
```
beeender: Why not put this to `CATCH_STD()`

realm/realm-library/src/main/java/io/realm/internal/SharedGroup.java
```diff
@@ -18,12 +18,21 @@
...
+import io.realm.exceptions.RealmError;
 import io.realm.exceptions.RealmIOException;
 import io.realm.internal.async.BadVersionException;
+import io.realm.internal.log.RealmLog;
 
 public class SharedGroup implements Closeable {
 
+    // Keep these public so we can ask users to experiment with these values if needed.
+    // Should be locked down as soon as possible.
+    public static long[] INCREMENTAL_BACKOFF_MS = new long[] {1, 3, 10, 20}; // Will keep re-using last value until LIMIT is hit
```
kneth: Can it be `private`?

realm/realm-jni/src/util.hpp
beeender: And `CATCH_FILE` actually should be removed since we have the file info from core exception. I will do it later.

realm/realm-library/src/androidTest/java/io/realm/RealmObjectTests.java
```diff
@@ -40,6 +40,7 @@
 import java.util.concurrent.Future;
 import java.util.concurrent.atomic.AtomicReference;
 
+import io.realm.entities.AllJavaTypes;
```
stk1m1: Unnecessary?

realm/realm-library/src/androidTest/java/io/realm/RealmTests.java
```diff
@@ -3332,4 +3334,24 @@ public void run(Realm realm) {
...
+        assertTrue(lockFile.exists());
+        FileOutputStream fooStream = new FileOutputStream(lockFile, false);
+        fooStream.write("Boom".getBytes());
+        fooStream.close();
+
+        try {
+            // This will try to open a second SharedGroup which should fail when the .lock file is corrupt
+            DynamicRealm.getInstance(realm.getConfiguration());
+            fail();
+        } catch (RealmError e) {
```
stk1m1: e -> expected

CHANGELOG.md
emanuelez: Crashing outright is not really a bug. We just changed the behaviour. Maybe something in the lines of:

 * Changed the behaviour that caused Realm to crash outright when the same Realm file was opened in two processes. Realm will now optimistically retry to open for 1 second before throwing an Error (#2459).

realm/realm-library/src/main/java/io/realm/internal/SharedGroup.java
```diff
@@ -72,6 +81,52 @@ public SharedGroup(String canonicalPath, boolean enableImplicitTransactions, Dur
...
+        long start = System.nanoTime();
+        RuntimeException lastError = null;
+        while (TimeUnit.MILLISECONDS.convert(System.nanoTime() - start, TimeUnit.NANOSECONDS) < INCREMENTAL_BACKOFF_LIMIT_MS) {
+            try {
+                long nativePtr = createNativeWithImplicitTransactions(nativeReplicationPtr, durability.value, key);
+                if (i > 0) {
+                    RealmLog.w("IncompatibleLockFile was detected. Error was resolved after " + i + " retries");
+                }
+                return nativePtr;
+
```
kneth: remove blank line

realm/realm-jni/src/util.hpp
cmelchior: Mostly because it seemed more relevant here (it will only be thrown when opening a SharedGroup)

realm/realm-library/src/main/java/io/realm/internal/SharedGroup.java
```diff
@@ -72,6 +81,52 @@ public SharedGroup(String canonicalPath, boolean enableImplicitTransactions, Dur
...
     }
 
+    private long openSharedGroupOrFail(Durability durability, byte[] key) {
+        // We have anecdotal evidence that on some versions of Android it is possible for two versions of an app
+        // to exist in two processes during an app upgrade. This is problematic since the lock file might not be
+        // compatible across two versions of Android. See https://github.com/realm/realm-java/issues/2459. If this
+        // happens we assume the overlap is really small so instead of failing outright we retry using incremental
+        // backoff.
+        int i = 0;
+        long start = System.nanoTime();
```
stk1m1: I'd make it `final long start` just in case.

realm/realm-library/src/main/java/io/realm/internal/SharedGroup.java
cmelchior: Yes, but prefer to keep it public for now. It means that we can work closer with users about resolving this. Right now we are scrambling a bit to find the root cause. Having it public means users can change the values without us releasing a new version of Realm (it is still in the internal API)

realm/realm-library/src/main/java/io/realm/internal/SharedGroup.java
```diff
@@ -72,6 +81,52 @@ public SharedGroup(String canonicalPath, boolean enableImplicitTransactions, Dur
...
+            }
+        }
+
+        throw new RealmError("Could not open the Realm file: " + lastError.getMessage());
+    }
+
+    // Returns the time to sleep before retrying opening the SharedGroup.
+    private long getSleepTime(int tries) {
+        if (INCREMENTAL_BACKOFF_MS == null) {
+            return 0;
```
kneth: I think returning `INCREMENTAL_BACKOFF_LIMIT_MS` is better. Otherwise you will get many calls to `Thread.sleep()`.

realm/realm-library/src/main/java/io/realm/internal/SharedGroup.java
```diff
@@ -72,6 +81,52 @@ public SharedGroup(String canonicalPath, boolean enableImplicitTransactions, Dur
...
+                try {
+                    Thread.sleep(getSleepTime(i));
+                    RealmLog.d("Waiting for another process to release the Realm file: " + path);
+                } catch (InterruptedException ignored) {
+                    RealmLog.d("Waiting for Realm to open interrupted: " + path);
+                }
+            }
+        }
+
+        throw new RealmError("Could not open the Realm file: " + lastError.getMessage());
```
beeender: just `throw lastError` instead?

stk1m1: Wouldn't it be better to throw `IncompatibleLockFileException` instead of `RealmError`? I think it could give users and us more insight when issues occur.

cmelchior: That would technically be a breaking change, but yes, otherwise I agree.

cmelchior: Besides. The RealmError will contain the relevant error message, just like before.

stk1m1:
Awesome job! 👍 (It would be x100 awesome if there is a test, but oh well... 😢)

realm/realm-library/src/main/java/io/realm/internal/SharedGroup.java
```diff
@@ -18,12 +18,21 @@
...
 import io.realm.exceptions.RealmIOException;
 import io.realm.internal.async.BadVersionException;
+import io.realm.internal.log.RealmLog;
 
 public class SharedGroup implements Closeable {
 
+    // Keep these public so we can ask users to experiment with these values if needed.
+    // Should be locked down as soon as possible.
+    public static long[] INCREMENTAL_BACKOFF_MS = new long[] {1, 3, 10, 20}; // Will keep re-using last value until LIMIT is hit
+    public static long INCREMENTAL_BACKOFF_LIMIT_MS = 1000;
```
beeender: I would like make the default time out longer, like 3 seconds. The time before ANR to kill is 5 seconds. As long as we don't trigger ANR, we will always get the accurate results.

cmelchior: Yes, I could probably live with that. Retrying every 20 ms for 3 seconds is kinda massive though, so I'll probably cap at 50 ms instead then? So 1,3, 10, 20, 50 ?

beeender: `[1, 10, 100, 200, 400]` makes more sense IMO.  The magnitudes of retry times should be enough differentiated. And also, we should give some free cpu time for kernel to do scheduling.

realm/realm-library/src/main/java/io/realm/internal/SharedGroup.java
stk1m1: I'd put a bit more user friendly description on it with date so when users find out, they can see it can be changed, and has a limited lifespan for exposure.

e.g.
>  The timeout and retry-policy settings for opening Realm lock is currently public for now, which will allow users to modify without us needing to release a new version of Realm Java. (June 16, 2016).

realm/realm-library/src/main/java/io/realm/internal/SharedGroup.java
cmelchior: Adjust to {1, 10, 20, 50, 100, 200, 400} and 3 sec timeout. This is a good comprise between being responsive and allowing the scheduler to work.

beeender:
You made it!! :+1:

realm/realm-library/src/main/java/io/realm/internal/SharedGroup.java
cmelchior: It isn't really "public" as such, but if we are working with users to solve this. We can tell them that this option exists. I wouldn't advertise it beyond that I think.

realm/realm-library/src/androidTest/java/io/realm/RealmTests.java
```diff
@@ -3332,4 +3334,24 @@ public void run(Realm realm) {
...
     }
+
+    @Test
+    public void incompatibleLockFile() throws IOException {
+        // Replace .lock file with a corrupted one
+        File lockFile = new File(realmConfig.getPath() + ".lock");
+        assertTrue(lockFile.exists());
+        FileOutputStream fooStream = new FileOutputStream(lockFile, false);
+        fooStream.write("Boom".getBytes());
+        fooStream.close();
```
nhachicha: maybe add a `flush` to ensure the file is modified before proceeding with the rest of the test

realm/realm-library/src/main/java/io/realm/internal/SharedGroup.java
```diff
@@ -72,6 +81,51 @@ public SharedGroup(String canonicalPath, boolean enableImplicitTransactions, Dur
...
+        // to exist in two processes during an app upgrade. This is problematic since the lock file might not be
+        // compatible across two versions of Android. See https://github.com/realm/realm-java/issues/2459. If this
+        // happens we assume the overlap is really small so instead of failing outright we retry using incremental
+        // backoff.
+        int i = 0;
+        final long start = System.nanoTime();
+        RuntimeException lastError = null;
+        while (TimeUnit.MILLISECONDS.convert(System.nanoTime() - start, TimeUnit.NANOSECONDS) < INCREMENTAL_BACKOFF_LIMIT_MS) {
+            try {
+                long nativePtr = createNativeWithImplicitTransactions(nativeReplicationPtr, durability.value, key);
```
nhachicha: do we need to delete the corrupted lock file? or does Core override it ?

cmelchior: I believe core overrides it if it is safe to do so, otherwise we should have had way more reports on this since it would happen for everybody when they did an upgrade.

nhachicha:
LGTM 👍

realm/realm-library/src/androidTest/java/io/realm/RealmTests.java
cmelchior: `flush` is a no-op for FileOutputStream:
```
    /**
     * Flushes this stream. Implementations of this method should ensure that
     * any buffered data is written out. This implementation does nothing.
     *
     * @throws IOException
     *             if an error occurs while flushing this stream.
     */
    public void flush() throws IOException {
        /* empty */
    }
```

PR #3023, opened 2016-06-17
Start ADB before Gradle runs.

cmelchior:
Tries to prevent 'No connected Devices'


cmelchior:
Seems to be a hardware problem. This does not have any effect. Closing.

Issue #3022, opened 2016-06-17
Realm Plugin/Transformation 1.0.1 Suspicious "unexpectedly contains class" Warnings


diegomontoya:
We are getting massive amount of gradle warning logs during build with 1.0.1. 1.0.0 has no such issue. 

```java
Warning: class [folders/1/10/realm-optional-api/it/unimi/dsi/fastutil/chars/Char2ByteRBTreeMap$KeySet.class] unexpectedly contains class [it.unimi.dsi.fastutil.chars.Char2ByteRBTreeMap$KeySet]
Warning: class [folders/1/10/realm-optional-api/it/unimi/dsi/fastutil/chars/Char2ByteArrayMap$EntrySet$2.class] unexpectedly contains class [it.unimi.dsi.fastutil.chars.Char2ByteArrayMap$EntrySet$2]
api/it/unimi/dsi/fastutil/chars/Char2ByteRBTreeMap$KeySet.class] unexpectedly contains class [it.unimi.dsi.fastutil.chars.Char2ByteRBTreeMap$KeySet]
Warning: class [folders/1/10/realm-optional-api/it/unimi/dsi/fastutil/chars/Char2ByteArrayMap$EntrySet$2.class] unexpectedly contains class [it.unimi.dsi.fastutil.chars.Char2ByteArrayMap$EntrySet$2]
Warning: class [folders/1/10/realm-optional-api/it/unimi/dsi/fastutil/chars/Char2LongLinkedOpenHashMap.class] unexpectedly contains class [it.unimi.dsi.fastutil.chars.Char2LongLinkedOpenHashMap]
Warning: class [folders/1/10/realm-optional-api/it/unimi/dsi/fastutil/chars/CharBigArrays$1.class] unexpectedly contains class [it.unimi.dsi.fastutil.chars.CharBigArrays$1]
Warning: class [folders/1/10/realm-optional-api/it/unimi/dsi/fastutil/chars/AbstractChar2ObjectMap.class] unexpectedly contains class [it.unimi.dsi.fastutil.chars.AbstractChar2ObjectMap]
Warning: class [folders/1/10/realm-optional-api/it/unimi/dsi/fastutil/chars/Char2LongMaps$EmptyMap.class] unexpectedly contains class [it.unimi.dsi.fastutil.chars.Char2LongMaps$EmptyMap]
[...]

PR #3028, opened 2016-06-19
Disable optional api transformer

beeender:
Because of #3022, this transformer doesn't play well with DexGuard.
The idea way is we can supply an option to disable it by users. But
unfortunately it seems to be difficult:

* The realm gradle options block is evaluated after calling
android.registerTransformer()
* We can pass a closure to transformer later, but the problem
Transformer.getScope() cannot return null or an empty set, otherwise
an exception will be thrown.

Just disable it until we find a proper way.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 1.0.2
+
+### Bug fixes
+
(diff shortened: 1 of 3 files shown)


cmelchior:
👍

beeender:
This is pushed to releases already (by accident). close

Issue #2936, opened 2016-06-03
Realm Transformer not correctly applied to classes inside androidTest


bkromhout:
#### Bug Details
I've just recently started trying to add instrumented tests to my application project so that I can test the portions of my code which have to do with data manipulation and persistence (which is much of the code).

I tried to start out small by adding a test class, `RBookTest` which tests just one method I have on one of my model classes, `RBook`. However, when I run the `:app:assembleDebugAndroidTest` task to assemble the test APK, Realm's bytecode transformer is causing the build to fail when it gets to this line in `RBookTest`:

```java
book1.isNew = false;
```

I've attached a log with more content in the Actual Results section, but this is the top-level failure message from Gradle:

```
Execution failed for task ':app:transformClassesWithRealmTransformerForDebugAndroidTest'.
> javassist.CannotCompileException: [source error] realmSet$isNew(boolean) not found in com.bkromhout.minerva.realm.RBook
```

I should note that I'm having absolutely no issues assembling the regular debug APK (the `:app:assembleDebug` task), and am aware of a few places which use the same sort of line as the problematic one in the test class.

#### Expected Results
I should be able to compile the test APK (using the `:app:assembleDebugAndroidTest`) in order to run instrumented tests on an Android emulator or device.

#### Actual Results
Running the `:app:assembleDebugAndroidTest` task fails. I've attached a file which contains the log output which results from running that task with the `--info` and `--stacktrace` options, but have omitted the output of all but the `:app:transformClassesWithRealmTransformerForDebugAndroidTest` task for brevity: [Log.txt](https://github.com/realm/realm-java/files/298517/Log.txt)

#### Steps & Code to Reproduce
I think the steps to reproduce the issue are fairly self-explanatory. The app's code is open source, so if you want to clone it to try and reproduce the issue, [use this link to the repo (commit hash 3366874f7b)](https://github.com/bkromhout/Minerva/tree/3366874f7b553ee802b2d31501fe3b5da491a5ab). That's the commit where I pushed the affected test class, `RBookTest`. The transformer seems to be choking when it tries to manipulate the setter on line 48 (`book1.isNew = false`).

If you just want to quickly reference the relevant files, here are links to a couple of them as of that commit:

* [`RBook` class](https://github.com/bkromhout/Minerva/blob/3366874f7b553ee802b2d31501fe3b5da491a5ab/app/src/main/java/com/bkromhout/minerva/realm/RBook.java)
[...]

PR #3030, opened 2016-06-20
Classes are not properly transformed during androidTests

cmelchior:
Reproduces #2936 

Observations so far:
- All unit tests seems to work fine when run from inside the `realm-library` folder.
- Log says that all files should be transformed. 
- Decompiled .class files have not been transformed (which is also why the androidTest transformer crashes)
- Can only be reproduced if run from inside a proper app project (I modified unitTestExamples)

### DIFF
--- a/examples/unitTestExample/src/main/java/io/realm/examples/unittesting/ExampleActivity.java
+++ b/examples/unitTestExample/src/main/java/io/realm/examples/unittesting/ExampleActivity.java
@@ -115,7 +115,7 @@ public class ExampleActivity extends Activity {
             public void execute(Realm realm) {
                 // Add a person
                 Person person = realm.createObject(Person.class);
-                person.setId(1);
+                person.id = 1;
                 person.setName("John Young");
                 person.setAge(14);
             }
--- a/examples/unitTestExample/src/main/java/io/realm/examples/unittesting/model/Person.java
+++ b/examples/unitTestExample/src/main/java/io/realm/examples/unittesting/model/Person.java
@@ -38,7 +38,7 @@ public class Person extends RealmObject {
     @Ignore
     private int tempReference;
 
-    private long id;
+    public long id;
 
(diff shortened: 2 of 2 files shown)


emanuelez:
This seems to be a bug in Google's Transformer API. It looks like `TESTED_SCOPE` is not allowed as a scope but only as a referenced scope. https://github.com/JetBrains/adt-tools-base/blob/502506f2faf7acc13b5ebc909ea254fe32d02d33/build-system/gradle-core/src/main/groovy/com/android/build/gradle/internal/pipeline/TransformManager.java#L400

cmelchior:
@emanuelez It looks intended to me though, so perhaps something we should create an issue for on `b.android.com`? I am not entirely sure why they would disallow transformers to run on test code though. Perhaps some other use case I am missing. 

So actions right now should the following, right?

1) Create bug on `b.android.com`
2) Document the quirks for `androidTest` code.

chiuki:
Any news on this? I'm looking into using realm in my Android app and I have lots of tests and lots of direct field access. Thanks!

cmelchior:
Hey @chiuku. Sorry  not yet. So far our best guess is that it is a bug in the Transform API. So for now it looks like direct field access cannot be used in instrumentation tests, unfortunately 😢

chiuki:
That's unfortunate. I'll wait until that's fixed before looking into realm again.

Do you have the link to the bug on `b.android.com`? I'll watch that as well. Thanks!

cmelchior:
Hey @chiuki 
Sorry for the late reply. This issue unfortunately got a bit lost in a vacation frenzy. I have created the issue here on b.android.com: https://code.google.com/p/android/issues/detail?id=219639&thanks=219639&ts=1470759569

cmelchior:
I'll close this PR instead of keeping it open. It is still beeing referenced from https://github.com/realm/realm-java/issues/2936, so we can always dig it out if there is a change on Googles side

PR #3031, opened 2016-06-20
Integrate Object Store [PART2] - SharedRealm

beeender:
Target for this PR:

- [x] Remove `SharedGroupManager`
- [x] Remove `SharedGroup`
- [x] Remove `Group`
- [x] ~~Re-think about the refs ownership. Maybe it is better just use `TableRef`/etc.. to control the ownership of objects instead of having complex crossed ref raw pointer in java.~~
- [x] Still handle realm cache in RealmCache and fix possible problems with RealmCoordinator
- [x] Still handle notifications in Java and fix possible issues with RealmCoordinator
- [x] Clean up related JNI code
- [x] Exception conversion
- [x] Create basic test cases for `SharedRealm`.


realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
```diff
@@ -0,0 +1,153 @@
+#include <object-store/src/object_store.hpp>
+#include "io_realm_internal_SharedRealm.h"
+
+#include "shared_realm.hpp"
+#include "util.hpp"
+
+using namespace realm;
+
+static std::vector<char> jbytearray_to_vector(JNIEnv *env, jbyteArray bytes) {
```
kneth: Why don't we move it to `util.hpp`?

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
```diff
@@ -0,0 +1,153 @@
...
+    TR_ENTER_PTR(config_ptr)
+    delete RC(config_ptr);
+}
+
+JNIEXPORT jlong JNICALL
+Java_io_realm_internal_SharedRealm_nativeGetSharedRealm
+(JNIEnv *, jclass, jlong config_ptr) {
+    TR_ENTER_PTR(config_ptr)
+    SharedRealm shared_realm = Realm::get_shared_realm(*RC(config_ptr));
+    return reinterpret_cast<jlong>(new SharedRealm(shared_realm));
```
kneth: Why not `shared_realm->get()`?

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
```diff
@@ -0,0 +1,153 @@
...
+}
+
+JNIEXPORT void JNICALL
+Java_io_realm_internal_SharedRealm_nativeCloseSharedRealm
+(JNIEnv *, jclass, jlong shared_realm_ptr) {
+    TR_ENTER_PTR(shared_realm_ptr)
+    delete SR(shared_realm_ptr);
```
kneth: This will delete the instance created in line 48 but not the one in line 47?

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
beeender: yes, will do

realm/realm-library/src/main/cpp/object-store/src/shared_realm.cpp
```diff
@@ -437,7 +437,7 @@ void Realm::notify()
     }
 }
 
-bool Realm::refresh()
+bool Realm::refresh(SharedGroup::VersionID version)
```
kneth: Can't we set a default value for `version` so other bindings don't break?

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
beeender: `SharedRealm` is actually `std::shared_ptr<Realm`.
And we are not holding a pointer to `Realm`, instead, we are holding a pointer to `shared_ptr`.

So this will delete the `share_ptr` created in line 48. The one created/got in line 47 may/may not be deleted, since it depends one how many `shared_ptr` still point to the `Realm`.

realm/realm-library/src/main/cpp/object-store/src/shared_realm.cpp
beeender: yes, an empty one `{}` is set in the header file, see the header file :P

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
beeender: Same reason with https://github.com/realm/realm-java/pull/3031#discussion_r67657968

realm/realm-library/src/androidTest/java/io/realm/RealmTests.java
```diff
@@ -651,9 +651,8 @@ public void executeTransaction_null() {
             realm.executeTransaction(null);
             fail("null transaction should throw");
         } catch (IllegalArgumentException ignored) {
-
         }
-        assertFalse(realm.hasChanged());
+        //assertFalse(realm.hasChanged());
```
kneth: Remove or not remove?

realm/realm-library/src/main/cpp/CMakeLists.txt
```diff
@@ -15,14 +15,22 @@ include (UseJava)
 
 set(CMAKE_VERBOSE_MAKEFILE ON)
 
+if (CMAKE_BUILD_TYPE STREQUAL "Release")
```
kneth: Nice

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
```diff
@@ -0,0 +1,209 @@
...
+Java_io_realm_internal_SharedRealm_nativeGetVersion
+(JNIEnv *, jclass, jlong shared_realm_ptr) {
+    TR_ENTER_PTR(shared_realm_ptr)
+    return ObjectStore::get_schema_version((*SR(shared_realm_ptr))->read_group());
+}
+
+JNIEXPORT jboolean JNICALL
+Java_io_realm_internal_SharedRealm_nativeIsEmpty(JNIEnv *, jclass, jlong shared_realm_ptr) {
+    TR_ENTER_PTR(shared_realm_ptr)
+    return JB(ObjectStore::is_empty((*SR(shared_realm_ptr))->read_group()));
```
kneth: I think we should have a inlined function called `to_jboolean` instead of `JB`. We already have `to_jstring()` and the upper case macros are used for casting of pointers.

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
```diff
@@ -0,0 +1,209 @@
+#include <object-store/src/object_store.hpp>
```
kneth: A general note: Instead of doing `*SR(...)` in the code, you could do `SharedRealm sr = *SR(...)` at the beginning of the method. I believe it will be easier to read.

kneth:
I have a few comments above - primarily on code style. My only concern is that the term `SharedRealm` can easily be misinterpreted: don't try to shared it between threads.

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
beeender: or we use `TO_XX` for or macro helpers? Sometimes `CH(xx)` is a bit difficult to read.

kneth: If the macro helpers have too long names, we don't gain much. Will `auto *shared_realm = static_cast<SharedRealm*>(native_ptr)` work? Since we are migrating to C++14, these old macros might be obsolete.

beeender: `auto* shared_realm = static_cast<SharedRealm*>(native_ptr)` honestly I like this more than the macros, although it takes one more line, the readability is so much better.

kneth: Can you try to use it a few places? And I think we should have a rule like "cast all native pointers at the beginning of a function".

realm/realm-library/src/main/cpp/util.hpp
```diff
@@ -56,29 +56,6 @@ JNIEXPORT jint JNICALL JNI_OnLoad(JavaVM *vm, void *reserved);
 #define STRINGIZE(x) STRINGIZE_DETAIL(x)
 
 // Exception handling
-
-#define CATCH_FILE(fileName) \
```
kneth: Oh, this will simply our code base :-)

beeender:
@kneth `SharedRealm` would be an internal concept only and maps to the OS's `Realm::SharedRealm` . If you have a better name, I'd love to change it since I kind of feel strange about it as well.

realm/realm-library/src/main/java/io/realm/internal/SharedRealm.java
```diff
@@ -168,6 +168,16 @@ public boolean isClosed() {
         return nativePtr == 0 || nativeIsClosed(nativePtr);
     }
 
+    public void writeCopy(File file, byte[] key) {
+        if (file.isFile() && file.exists()) {
+            throw new IllegalArgumentException("The destination file must not exist");
+        }
+        if (key != null && key.length != 64) {
```
kneth: Isn't key length checked by object store?

beeender: It is `REALM_ASSERT` without clear message, and it doesn't throw in the release build I think?

kneth: Then we have to do the check. But I think it a bit strange that the object store doesn't check.

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
```diff
@@ -45,110 +47,156 @@ JNIEXPORT void JNICALL
...
 }
 
 JNIEXPORT void JNICALL
 Java_io_realm_internal_SharedRealm_nativeBeginTransaction
-(JNIEnv *, jclass, jlong shared_realm_ptr) {
+(JNIEnv *env, jclass, jlong shared_realm_ptr) {
     TR_ENTER_PTR(shared_realm_ptr)
-    (*SR(shared_realm_ptr))->begin_transaction();
+
+    SharedRealm shared_realm = *(reinterpret_cast<SharedRealm*>(shared_realm_ptr));
```
kneth: `auto`?

beeender: Uh! yes, i forgot!

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
```diff
@@ -0,0 +1,314 @@
...
+}
+
+JNIEXPORT jlong JNICALL
+Java_io_realm_internal_SharedRealm_nativeGetSharedRealm
+(JNIEnv *env, jclass, jlong config_ptr) {
+    TR_ENTER_PTR(config_ptr)
+
+    try {
+        SharedRealm shared_realm = Realm::get_shared_realm(*RC(config_ptr));
+        return reinterpret_cast<jlong>(new SharedRealm(shared_realm));
```
bdash: Using `std::move(shared_realm)` here would avoid the refcount churn that the copy currently causes.

realm/realm-library/src/main/cpp/object-store/src/shared_realm.hpp
```diff
@@ -148,6 +149,11 @@ namespace realm {
         void init(std::shared_ptr<_impl::RealmCoordinator> coordinator);
         Realm(Config config);
 
+        // FIXME: Java still needs the SharedGroup pointers for some places.
+        //        Remove this after the final clean up.
+        SharedGroup& get_shared_group() { return *m_shared_group; }
```
bdash: I'd suggest declaring this later in the class, alongside `read_group()`.

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -150,12 +150,14 @@
 
     @Override
     protected void finalize() throws Throwable {
+        /*
```
bdash: I'd guess this block wants to log if `Realm::is_closed` is false?

realm/realm-library/src/main/java/io/realm/internal/SharedRealm.java
bdash: https://github.com/realm/realm-object-store/blob/master/src/shared_realm.cpp#L127-L129

realm/realm-library/src/main/java/io/realm/Realm.java
beeender: Yes! But it seems this should be in the `BaseRealm` since currently we missed the log in `DynamicRealm`'s finalizer. Will fix it later.

realm/realm-library/src/main/cpp/object-store/src/shared_realm.hpp
beeender: Pushed some new commits which created a some helper functions in OS. Let me know if it looks better, or maybe it looks worse :P

realm/realm-library/src/main/java/io/realm/Realm.java
beeender: Fixed.

realm/realm-library/src/main/java/io/realm/internal/SharedRealm.java
beeender: Changed `REALM_ASSERT` to excpetion in OS.

realm/realm-library/src/main/cpp/object-store/src/shared_realm.hpp
```diff
@@ -118,8 +118,12 @@ namespace realm {
         void cancel_transaction();
         bool is_in_transaction() const noexcept;
         bool is_in_read_transaction() const { return !!m_group; }
+        SharedGroup::VersionID get_version_of_current_transaction() const {
+            return m_shared_group->get_version_of_current_transaction();
+        }
 
-        bool refresh();
+        // FIXME: The param should not be needed.
+        bool refresh(SharedGroup::VersionID version = {});
```
beeender: @bdash This might not be suitable. All java needs is advance to the latest/a specific version. And refresh to a specific version doesn't seem to be a generic requirement for others. So maybe I should add a new API somewhere? Something like `SharedRealm::advance(Version)` .

realm/realm-library/src/androidTest/java/io/realm/RealmTests.java
```diff
@@ -208,7 +208,9 @@ public void getInstance_writeProtectedFile() throws IOException {
         assertTrue(realmFile.createNewFile());
         assertTrue(realmFile.setWritable(false));
 
-        thrown.expect(RealmIOException.class);
+        // FIXME: Why do we throw RealmIOException for this case, but IAE for other cases when opening Realm?
```
beeender: @realm/java See above

realm/realm-library/src/main/cpp/object-store/src/object_store.hpp
```diff
@@ -75,6 +76,36 @@ namespace realm {
         static std::string table_name_for_object_type(StringData class_name);
         static StringData object_type_for_table_name(StringData table_name);
 
+        // Those are needed by java for the async query
```
beeender: @bdash Would those look OK? Or simply expose the `SharedGroup` pointer and keep those functions in JNI to make the Object Store API cleaner?

realm/realm-library/src/androidTest/java/io/realm/RealmTests.java
kneth: `FileWriter` throws `IOExcetion` when opening a read-only file while `File` will throw `IllegalArgumentException`. We have probably modelled it over `FileWriter`.

beeender: So I think we can just throw `IllegalArgumentException` when open Realm files now?

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
```diff
@@ -224,9 +232,9 @@ Java_io_realm_internal_SharedRealm_nativeHasTable(JNIEnv *env, jclass, jlong sha
     auto shared_realm = *(reinterpret_cast<SharedRealm*>(shared_realm_ptr));
     try {
         JStringAccessor name(env, table_name);
-        return JB(shared_realm->read_group()->has_table(name));
+        return static_cast<jboolean >(shared_realm->read_group()->has_table(name));
```
kneth: Remove space before `>`

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
```diff
@@ -224,9 +232,9 @@ Java_io_realm_internal_SharedRealm_nativeHasTable(JNIEnv *env, jclass, jlong sha
...
     } CATCH_STD()
-    return JB(false);
+    return static_cast<jboolean >(false);
```
kneth: Remove space before `>`

realm/realm-library/src/main/java/io/realm/internal/Table.java
```diff
@@ -49,7 +49,7 @@
     private static final long PRIMARY_KEY_FIELD_COLUMN_INDEX = 1;
     private static final long NO_PRIMARY_KEY = -2;
 
-    protected long nativePtr;
+    long nativePtr;
```
kneth: private?

realm/realm-library/src/androidTest/java/io/realm/internal/JNICloseTest.java
```diff
@@ -18,86 +18,14 @@
...
-import java.util.ArrayList;
-import java.util.List;
-
 import io.realm.RealmFieldType;
 import io.realm.TestHelper;
 
+
 // Tables get detached
 
 public class JNICloseTest extends AndroidTestCase {
```
beeender: Not need anymore since `Table.close()` is removed. (It is removed since no-one is using it and it should be handled by Phantom Ref in the quite near future.)

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
```diff
@@ -0,0 +1,351 @@
...
+                      static_cast<SchemaMode>(io_realm_internal_SharedRealm_SCHEMA_MODE_VALUE_RESET_FILE), "");
+static_assert(SchemaMode::Additive ==
+                      static_cast<SchemaMode>(io_realm_internal_SharedRealm_SCHEMA_MODE_VALUE_ADDITIVE), "");
+static_assert(SchemaMode::Manual ==
+              static_cast<SchemaMode>(io_realm_internal_SharedRealm_SCHEMA_MODE_VALUE_MANUAL), "");
+
+JNIEXPORT jlong JNICALL
+Java_io_realm_internal_SharedRealm_nativeCreateConfig
+(JNIEnv *env, jclass, jstring realm_path, jbyteArray key, jbyte schema_mode, jboolean in_memory,
+ jboolean cache, jboolean disable_format_upgrade, jboolean auto_change_notification) {
```
bmunkholm: Allover "{" must begin on it's own line for functions.

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
```diff
@@ -0,0 +1,351 @@
...
+    config->in_memory = in_memory;
+    config->cache = cache;
+    config->disable_format_upgrade = disable_format_upgrade;
+    config->automatic_change_notifications = auto_change_notification;
+    return reinterpret_cast<jlong>(config);
+}
+
+JNIEXPORT void JNICALL
+Java_io_realm_internal_SharedRealm_nativeCloseConfig
+(JNIEnv *, jclass, jlong config_ptr) {
```
bmunkholm: Placing parameters on its own line are deviating from the style used all other places. Please have it at the end of the function name.

realm/realm-library/src/androidTest/java/io/realm/internal/JNICloseTest.java
kneth: Cool

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
beeender: I am changing others to this as well, since it looks much better --- Almost all jni functions are quite long, normally they need to be broke into two lines. Shall we make the `Placing parameters on its own line` for JNI functions as a rule?

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
```diff
@@ -0,0 +1,351 @@
...
+    } CATCH_STD()
+}
+
+JNIEXPORT void JNICALL
+Java_io_realm_internal_SharedRealm_nativeRefresh__JJJ(JNIEnv *env, jclass, jlong shared_realm_ptr,
+                                                      jlong version, jlong index) {
+    TR_ENTER_PTR(shared_realm_ptr)
+
+    auto shared_realm = *(reinterpret_cast<SharedRealm*>(shared_realm_ptr));
+    SharedGroup::VersionID version_id(static_cast<SharedGroup::version_type>(version),
```
kneth: Move into `try` block?

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
```diff
@@ -0,0 +1,351 @@
...
+JNIEXPORT jlongArray JNICALL
+Java_io_realm_internal_SharedRealm_nativeGetVersionID(JNIEnv *env, jclass, jlong shared_realm_ptr) {
+    TR_ENTER_PTR(shared_realm_ptr)
+
+    auto shared_realm = *(reinterpret_cast<SharedRealm*>(shared_realm_ptr));
+    try {
+        using rf = realm::_impl::RealmFriend;
+        SharedGroup::VersionID version_id = rf::get_shared_group(*shared_realm).get_version_of_current_transaction();
+
+        jlong version_array [2];
```
kneth: Remove space before `[`

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
```diff
@@ -0,0 +1,351 @@
...
+    TR_ENTER_PTR(shared_realm_ptr)
+
+    try {
+        JStringAccessor name(env, table_name); // throws
+        auto shared_realm = *(reinterpret_cast<SharedRealm*>(shared_realm_ptr));
+        if (!shared_realm->read_group().has_table(name) && !shared_realm->is_in_transaction()) {
+            std::ostringstream ss;
+            ss << "Table " << name << " doesn't exist and the shared realm is not in transaction.";
+            ThrowException(env, IllegalState, ss.str());
+            return static_cast<jlong>(NULL);
```
kneth: Should we using `nullptr` instead?

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
```diff
@@ -0,0 +1,351 @@
...
+Java_io_realm_internal_SharedRealm_nativeGetTable(JNIEnv *env, jclass, jlong shared_realm_ptr,
+                                                  jstring table_name) {
+    TR_ENTER_PTR(shared_realm_ptr)
+
+    try {
+        JStringAccessor name(env, table_name); // throws
+        auto shared_realm = *(reinterpret_cast<SharedRealm*>(shared_realm_ptr));
+        if (!shared_realm->read_group().has_table(name) && !shared_realm->is_in_transaction()) {
+            std::ostringstream ss;
+            ss << "Table " << name << " doesn't exist and the shared realm is not in transaction.";
```
kneth: "realm -> "Realm"

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
beeender: This constructor should never throw.

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
```diff
@@ -0,0 +1,351 @@
...
+Java_io_realm_internal_SharedRealm_nativeHasTable(JNIEnv *env, jclass, jlong shared_realm_ptr,
+                                                  jstring table_name) {
+    TR_ENTER_PTR(shared_realm_ptr)
+
+    auto shared_realm = *(reinterpret_cast<SharedRealm*>(shared_realm_ptr));
+    try {
+        JStringAccessor name(env, table_name);
+        return static_cast<jboolean >(shared_realm->read_group().has_table(name));
+    } CATCH_STD()
+    return static_cast<jboolean >(false);
```
kneth: `return JNI_FALSE` is also a possibility.

realm/realm-library/src/main/cpp/io_realm_internal_TestUtil.cpp
```diff
@@ -0,0 +1,127 @@
+#include "io_realm_internal_TestUtil.h"
+#include "util.hpp"
+
+static jstring throwOrGetExpectedMessage(JNIEnv *env, jlong testcase, bool should_throw);
+
+JNIEXPORT jlong JNICALL
+Java_io_realm_internal_TestUtil_getMaxExceptionNumber
+(JNIEnv *, jclass) {
```
bmunkholm: Same in this file - please use consistent style with the rest of the file.

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
```diff
@@ -0,0 +1,351 @@
...
+
+JNIEXPORT jboolean JNICALL
+Java_io_realm_internal_SharedRealm_nativeHasTable(JNIEnv *env, jclass, jlong shared_realm_ptr,
+                                                  jstring table_name) {
+    TR_ENTER_PTR(shared_realm_ptr)
+
+    auto shared_realm = *(reinterpret_cast<SharedRealm*>(shared_realm_ptr));
+    try {
+        JStringAccessor name(env, table_name);
+        return static_cast<jboolean >(shared_realm->read_group().has_table(name));
```
kneth: Please remove space before `>`

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
beeender: > static_cast from 'nullptr_t' to 'jlong' (aka 'long long') is not allowed

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
```diff
@@ -0,0 +1,351 @@
...
+Java_io_realm_internal_SharedRealm_nativeRenameTable(JNIEnv *env, jclass, jlong shared_realm_ptr,
+                                                     jstring old_table_name, jstring new_table_name) {
+    TR_ENTER_PTR(shared_realm_ptr)
+
+    auto shared_realm = *(reinterpret_cast<SharedRealm*>(shared_realm_ptr));
+    try {
+        JStringAccessor old_name(env, old_table_name);
+        if (!shared_realm->is_in_transaction()) {
+            std::ostringstream ss;
+            ss << "Table " << old_name << " cannot be removed when the shared realm is not in transaction.";
```
kneth: I don't like that we are talking about "tables".

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
```diff
@@ -0,0 +1,351 @@
...
+
+    return 0;
+}
+
+JNIEXPORT void JNICALL
+Java_io_realm_internal_SharedRealm_nativeWriteCopy(JNIEnv *env, jclass,
+        jlong shared_realm_ptr, jstring path, jbyteArray key) {
+    TR_ENTER_PTR(shared_realm_ptr);
+
+    JStringAccessor path_str(env, path);
```
kneth: Move into `try` block since the constructor can throw.

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
```diff
@@ -0,0 +1,351 @@
...
+    JniByteArray key_buffer(env, key);
```
kneth: Move into `try` block since the constructor can throw.

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
```diff
@@ -0,0 +1,351 @@
...
+Java_io_realm_internal_SharedRealm_nativeWaitForChange(JNIEnv *env, jclass, jlong shared_realm_ptr) {
+    TR_ENTER_PTR(shared_realm_ptr);
+
+    auto shared_realm = *(reinterpret_cast<SharedRealm*>(shared_realm_ptr));
+    try {
+        using rf = realm::_impl::RealmFriend;
+        return static_cast<jboolean>(rf::get_shared_group(*shared_realm).wait_for_change());
+    } CATCH_STD()
+
+    return static_cast<jboolean>(false);
```
kneth: `JNI_FALSE`?

realm/realm-library/src/main/cpp/util.hpp
```diff
@@ -107,32 +82,33 @@ std::string num_to_string(T pNumber)
...
-    UnsupportedOperation = 9,
-    OutOfMemory = 10,
-    FatalError = 11,
-    RuntimeError = 12,
-    RowInvalid = 13,
-    CrossTableLink = 15,
-    BadVersion = 16,
-    LockFileError = 17
-// NOTE!!!!: Please also add test cases to Util.java when introducing a new exception kind.
+    NoSuchField,
```
bmunkholm: Why are the numbers removed? It's there to ensure we don't accidentally add new exceptions in the middle that don't match in the Java type.

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
beeender: It is more like a internal exception. We could change it to RealmObject, but what if we internally calling this method to remove a pk table? The the message will be quite confusing ... So, any better term than Table or RealmObject?

realm/realm-library/src/main/java/io/realm/internal/SharedRealm.java
```diff
@@ -0,0 +1,290 @@
...
+        public String toString() {
+            return "VersionID{" +
+                    "version=" + version +
+                    ", index=" + index +
+                    '}';
+        }
+
+        @Override
+        public boolean equals(Object object) {
+            if (this == object) return true;
```
bmunkholm: returns on separate line

realm/realm-library/src/main/cpp/util.hpp
beeender: The number doesn't matter at all, since java is not using it to match the exception. It is matched by switch/case in `util.cpp:ThrowException`. The number was using to create java test, now we are using `ExceptionKindMax` to ensure all exceptions are covered.

realm/realm-library/src/main/java/io/realm/internal/Table.java
```diff
@@ -593,48 +569,6 @@ private boolean isPrimaryKeyColumn(long columnIndex) {
     }
 
     /**
-     * Returns a view sorted by the specified column and order.
-     *
-     * @param columnIndex the column index.
-     * @param sortOrder the sort order.
-     * @return a sorted view.
-     */
-    public TableView getSortedView(long columnIndex, Sort sortOrder){
```
bmunkholm: Why remove this?

beeender: Since no one is using it and i don't see any reason it should be used in a near future. If you know any good reason it will be used in this year, please let me know.

realm/realm-library/src/main/java/io/realm/internal/Util.java
```diff
@@ -48,43 +48,6 @@ public static String getTablePrefix() {
     }
     static native String nativeGetTablePrefix();
 
-
-    // Testcases run in nativeCode
-    public enum Testcase {
```
bmunkholm: Ok - disregard the other comment in the corresponding cpp file - I see it's not transferred :-)

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
bmunkholm: Please don't - it doesn't follow our code-style guideline for C++. We are introducing clang-format in core which will ensure a consistent style automatically. And that will change this back - so you would waste time doing this :-)

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
bmunkholm: Maybe also remove "shared" from the text as you mentioned the concept should not be exposed to users :-)

realm/realm-library/src/main/java/io/realm/internal/Table.java
bmunkholm: It's not exposed at all to users?

bmunkholm:
Great work @beeender!
It would be great if the PR's could be much smaller - in particular to separate stylistic chaYou could work with smaller PR's against a feature-branch to facilitate easier reviews.

realm/realm-library/src/main/java/io/realm/internal/Table.java
beeender: Nope, user should always build `TableView` from `Query`.  Sure for some cases build `TableView` directly from the `Table` will be `faster`, but I don't think it will be fast enough for worthy doing that -- that means we also need to handle this special case for asyc query especially.

beeender:
commit message:
This simplified our code base a lot. Basically all APIs we need from
SharedGroup/Group are wrapped in the SharedRealm. So we can just remove
those classes.

But we do need a few APIs from SharedGroup which is not supplied by
SharedRealm because of async queries. Currently we expose those in a
friend class of ShareRealm, see https://github.com/realm/realm-object-store/pull/141

We are still managing Realm caches in Java although there are mechanism
in OS to do the same thing. The major reason is we have method like
deleteRealm needs information from cache to check if all Realm instances
are closed. The ShareRealm is actually a std::shared_ptr<Realm>. We hold
the pointer to the std::shared_ptr<Realm> in Java.
Another fundamental change is that we used to have a separated
SharedGroups for DynamicRealm and typed Realm in the same thread. But
now they are using different SharedRealm but actually the different
SharedRealms are point to the same SharedGroup.

And some other code cleanup.

@realm/java You need also help to check https://github.com/realm/realm-object-store/pull/141
@cmelchior See above for the DynamicRealm part

realm/realm-library/src/main/java/io/realm/internal/Table.java
Zhuinden: Does this influence my use-case here

    return resultPost.getResultList().where().findAll().sort(ResultListFields.RANK Sort.ASCENDING);

?

beeender: @Zhuinden , nope, the `where()` created a `Query` object which is building `RealmResults` from `Query` not the `Table`.

Jenkinsfile
```diff
@@ -11,6 +11,11 @@ try {
       checkout scm
       // Make sure not to delete the folder that Jenkins allocates to store scripts
       sh 'git clean -ffdx -e .????????'
+      // Use https url to work around the permission problem
+      sh 'git config --file=.gitmodules submodule.realm/realm-library/src/main/cpp/object-store.url https://github.com/realm/realm-object-store.git'
```
cmelchior: Will this just pull in master? Wouldn't it be safer if we specified a specific SHA or tag? Also for reproducibility.

beeender: Specific SHA1 is in the PR, but github doesn't show it. You can see it with gitk after pull to the local. This is only to solve the jenkins problem with `git@` protocol

realm/realm-library/src/androidTest/java/io/realm/RealmTests.java
```diff
@@ -1911,6 +1901,14 @@ public void writeEncryptedCopyTo() throws Exception {
     }
 
     @Test
+    public void writeEncryptedCopyTo_wrongKeyLength() {
+        byte[]  wrongLentKey = new byte[42];
```
cmelchior: wrongKeyLength

realm/realm-library/src/androidTest/java/io/realm/internal/JNICloseTest.java
```diff
@@ -18,86 +18,14 @@
...
 
-import java.io.Closeable;
-import java.io.File;
-import java.util.ArrayList;
-import java.util.List;
-
 import io.realm.RealmFieldType;
 import io.realm.TestHelper;
 
+
```
cmelchior: Space could be removed

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "realm/realm-library/src/main/cpp/object-store"]
+	path = realm/realm-library/src/main/cpp/object-store
+	url = git@github.com:realm/realm-object-store.git
```
bdash: FWIW, we've preferred to use the HTTPS URLs in our other projects as not all users have SSH keys configured on GitHub.

Jenkinsfile
bdash: You can make this work in Jenkins by having it do the SCM operations with an SSH agent active with the appropriate SSH keys loaded. It looks something like:
```groovy
sshagent(['realm-ci-ssh']) {
    // …
}
```

Though, as I mentioned above, I think it's a mistake to specify the `git@` URL in the .gitmodules file to begin with.

beeender: will just to change to use https in the .gitmodule

realm/realm-library/src/androidTest/java/io/realm/internal/SharedRealmTests.java
```diff
@@ -0,0 +1,135 @@
...
+        sharedRealm.beginTransaction();
+        sharedRealm.getTable("TableToRemove");
+        assertTrue(sharedRealm.hasTable("TableToRemove"));
+        sharedRealm.removeTable("TableToRemove");
+        assertFalse(sharedRealm.hasTable("TableToRemove"));
+        sharedRealm.commitTransaction();
+    }
+
+    @Test
+    public void removeTable_notInTransactionThrows() {
```
cmelchior: Test missing for removing a table that doesn't exists.

realm/realm-library/src/androidTest/java/io/realm/internal/SharedRealmTests.java
```diff
@@ -0,0 +1,135 @@
...
+        sharedRealm.getTable("OldTable");
+        assertTrue(sharedRealm.hasTable("OldTable"));
+        sharedRealm.renameTable("OldTable", "NewTable");
+        assertFalse(sharedRealm.hasTable("OldTable"));
+        assertTrue(sharedRealm.hasTable("NewTable"));
+        sharedRealm.commitTransaction();
+    }
+
+    @Test
+    public void renameTable_notInTransactionThrows() {
```
cmelchior: Test missing for renaming non-existent class.

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
```diff
@@ -0,0 +1,371 @@
...
+
+    return static_cast<jlong>(NULL);
+}
+
+JNIEXPORT void JNICALL
+Java_io_realm_internal_SharedRealm_nativeCloseConfig(JNIEnv *, jclass, jlong config_ptr)
+{
+    TR_ENTER_PTR(config_ptr)
+
+    auto config = reinterpret_cast<realm::Realm::Config*>(config_ptr);
```
cmelchior: Should we have `C(config_ptr)` macro for this?

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
```diff
@@ -0,0 +1,371 @@
...
+    } CATCH_STD()
+    return static_cast<jlong>(NULL);
+}
+
+JNIEXPORT void JNICALL
+Java_io_realm_internal_SharedRealm_nativeCloseSharedRealm(JNIEnv *, jclass, jlong shared_realm_ptr)
+{
+    TR_ENTER_PTR(shared_realm_ptr)
+
+    auto ptr = reinterpret_cast<SharedRealm*>(shared_realm_ptr);
```
cmelchior: `SR()` macro?

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
beeender: See https://github.com/realm/realm-java/pull/3031#discussion_r68214660

I really prefer to make it clear than the short macros -- that is quite hard to read (for me)

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
```diff
@@ -0,0 +1,371 @@
...
+Java_io_realm_internal_SharedRealm_nativeGetTable(JNIEnv *env, jclass, jlong shared_realm_ptr, jstring table_name)
+{
+    TR_ENTER_PTR(shared_realm_ptr)
+
+    try {
+        JStringAccessor name(env, table_name); // throws
+        auto shared_realm = *(reinterpret_cast<SharedRealm*>(shared_realm_ptr));
+        if (!shared_realm->read_group().has_table(name) && !shared_realm->is_in_transaction()) {
+            std::ostringstream ss;
+            ss << "Table " << name << " doesn't exist and the shared Realm is not in transaction.";
```
cmelchior: If this bubbles up to end user we should not use `Table` but `Class`

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
beeender: See above

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
cmelchior: I don't have a strong opinion. So fine with me. Having it explicit also means that using auto still keeps it readable. eg`auto config = C(config_ptr)`is probably a bit too much magic.

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
```diff
@@ -0,0 +1,371 @@
...
+        jstring old_table_name, jstring new_table_name)
+{
+    TR_ENTER_PTR(shared_realm_ptr)
+
+    auto shared_realm = *(reinterpret_cast<SharedRealm*>(shared_realm_ptr));
+    try {
+        JStringAccessor old_name(env, old_table_name);
+        if (!shared_realm->is_in_transaction()) {
+            std::ostringstream ss;
+            ss << "Table " << old_name << " cannot be removed when the realm is not in transaction.";
```
cmelchior: Same

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
```diff
@@ -0,0 +1,371 @@
...
+Java_io_realm_internal_SharedRealm_nativeRemoveTable(JNIEnv *env, jclass, jlong shared_realm_ptr, jstring table_name)
+{
+    TR_ENTER_PTR(shared_realm_ptr)
+
+    auto shared_realm = *(reinterpret_cast<SharedRealm*>(shared_realm_ptr));
+    try {
+        JStringAccessor name(env, table_name);
+        if (!shared_realm->is_in_transaction()) {
+            std::ostringstream ss;
+            ss << "Table " << name << " cannot be removed when the realm is not in transaction.";
```
cmelchior: Same

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
beeender: @cmelchior what about https://github.com/realm/realm-java/pull/3031#discussion_r76013582

realm/realm-library/src/main/cpp/util.hpp
```diff
@@ -107,32 +82,33 @@ std::string num_to_string(T pNumber)
...
+    FatalError,
+    RuntimeError,
+    RowInvalid,
+    CrossTableLink,
+    BadVersion,
+    LockFileError,
+    IllegalState,
+    // NOTE!!!!: Please also add test cases to io_realm_internal_TestUtil when introducing a
+    // new exception kind.
+    ExceptionKindMax
```
cmelchior: This always have to be last right? Perhaps add a comment about that

realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -226,9 +228,8 @@ protected void removeHandler() {
      * the last transaction was committed.
      *
      * @param destination file to save the Realm to.
-     * @throws java.io.IOException if any write operation fails.
      */
-    public void writeCopyTo(File destination) throws java.io.IOException {
+    public void writeCopyTo(File destination) {
```
cmelchior: Is this correct? I would assume this operation could fail in a number of ways.

realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -243,15 +244,14 @@ public void writeCopyTo(File destination) throws java.io.IOException {
      *
      * @param destination file to save the Realm to.
      * @param key a 64-byte encryption key.
-     * @throws java.io.IOException if any write operation fails.
      * @throws IllegalArgumentException if destination argument is null.
      */
-    public void writeEncryptedCopyTo(File destination, byte[] key) throws java.io.IOException {
+    public void writeEncryptedCopyTo(File destination, byte[] key) {
```
cmelchior: Same

realm/realm-library/src/main/java/io/realm/BaseRealm.java
beeender: It throws a `RealmFileException` in OS, and we converted it to a IAE.

cmelchior: But right now the Javadoc doesn't mention that.

cmelchior: Also technically it is a breaking change. I would think that in this case actually exposing  it as an IO Exception is probably the  most accurate?

beeender: Not really ... `RealmFileExcpetion` has different kinds, and it could throw even when open Realm, see `translate_file_exception` in OS code. At least it should not be a checked exception.

beeender: Shall we create `RealmFileException` in java and match it to the OS one. and remove the `RealmIOException` which is a checked exception?

beeender: Uh, no. `RealmIOException` is not checked, we can still use it and match it to the `RealmFileException` in OS, does this make sense?

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
cmelchior: I think they key point is where it ends up. If this is getting caught in java and converted to something else, then it is fine, but if it ends up in the end users log, i would rename it.

realm/realm-library/src/main/java/io/realm/BaseRealm.java
cmelchior: I would probably remove RealmIOException. I don't see it having any value over a normal IOException. I think this is a case where Java treats IO exceptions as checked, while C++ does not. IMO we should stick with Java conventions.

cmelchior:
Except some minor comments and a discussion on how `IOExceptions` should work, I think this looks great. Nice to get rid of so much code 👍

realm/realm-library/src/main/java/io/realm/BaseRealm.java
beeender: The the biggest change is `Realm.getInstance()` will throw a `IOExcpetion` and user has to catch it which doesn't seem to be necessary at all...

cmelchior: Your right, that doesn't make sense. I think the main reason I was concerned about IOEception for `writeCopyTo` is that the the `File` param could cause it, not opening the Realm file itself

cmelchior: Ideally, I would probably like to keep the IOException for problems with the destination + add a comment about IllegalArgumentException being thrown if invalid Realm file.

But I am not sure if the ObjectStore exposes that as two different exceptions, if not just adding a comment about IllegalArgumentException is probably the next best thing

beeender:
Since the IOException change is big, i will create another PR for it.

Zhuinden:
Are you sure you want the users to have to check for IOException at every Realm.getInstance() call? That's kinda ugly, although I guess it makes sense. :|

Although wrapping it outside of Realm isn't that much of a task, `RealmUtils.getInstance(realmConfig)` works pretty well too I guess :tongue:

cmelchior:
No, we ended up agreeing on a RealmFileException instead which a unchecked exception. So the only change will happen in `writeCopyTo()` which no longer throws a checked exception

Zhuinden:
Oh I see! Cool, I misunderstood then. :smile:

beeender:
The `IOException` related things are fixed in https://github.com/realm/realm-java/pull/3349

zaki50:
After introducing this, we need to execute `git submodule` command after `git clone`.

How about adding following instructions to `README.md` (after the `Prerequisites` section) ?

    Download sources:
    
    You can download the source code of Realm Java by using git.
    
    ```
    git clone git@github.com:realm/realm-java.git
    ```
    
    or
    
    ```
    git clone https://github.com/realm/realm-java.git
    ```
    
    And then you need to update submodules.
    
    ```
    cd realm-java
    git submodule init
    git submodule update
    ```

bdash:
Why not just pass the `--recursive` option to `git clone`?

zaki50:
@bdash Oh, I didn't know that!

README.md
```diff
@@ -101,6 +101,22 @@ Prerequisites:
    launchctl setenv REALM_CORE_DOWNLOAD_DIR "$REALM_CORE_DOWNLOAD_DIR"
    ```
 
+### Download sources
+
+You can download the source code of Realm Java by using git. Since realm-java has git submodules, use `--recursive` when clone.
```
cmelchior: when cloning the repository.

cmelchior:
Besides one tiny comment to the README :shipit:

PR #3035, opened 2016-06-21
RealmResults as empty list when its backing Table is removed

### LINKED ISSUE #2953: Realm is not Closed but Illegal State: The Realm has been closed and is no longer accessible.
#### Goal

my app needs to get configuration from server to create a Table/Schema after API call.
since the schema is a job specific, another job will have a slight different schema. so i need to delete Schema once user is done with a job.

e.g. Product class schema. (not all product have expiry, may not always appear)

``` json
[
  {
    "Id":"1",
    "Name":"ProductName"
  },
  { 
    "Id":"12",
    "Name":"Expiry"
  }
]
```
#### Expected Results

Able to dynamically create a table/schema, perform CRUD and delete Schema when user is done.
#### Actual Results

>  E.g. full stack trace with exception

```
com.example.realm I/tag: John
com.example.realm I/tag: isClosed = false
com.example.realm I/tag: schema removed, isClosed = false
com.example.realm D/OpenGLRenderer: Use EGL_SWAP_BEHAVIOR_PRESERVED: true

   [ 06-07 18:55:00.921 26948:26948 D/         ]
   HostConnection::get() New Host Connection established 0x7fe9ecc0e600, tid 26948
com.example.realm D/REALM: jni: ThrowingException 8, The Realm has been closed and is no longer accessible., .
com.example.realm D/REALM: Exception has been throw: Illegal State: The Realm has been closed and is no longer accessible.
com.example.realm D/AndroidRuntime: Shutting down VM

   --------- beginning of crash
06-07 18:55:00.961 26948-26948/com.example.realm E/AndroidRuntime: FATAL EXCEPTION: main
      Process: com.example.realm, PID: 26948
      java.lang.IllegalStateException: Illegal State: The Realm has been closed and is no longer accessible.
       at io.realm.internal.TableView.nativeSyncIfNeeded(Native Method)
       at io.realm.internal.TableView.syncIfNeeded(TableView.java:831)
       at io.realm.RealmResults.syncIfNeeded(RealmResults.java:618)
       at io.realm.RealmResults.notifyChangeListeners(RealmResults.java:1001)
       at io.realm.RealmResults.notifyChangeListeners(RealmResults.java:996)
       at io.realm.HandlerController.notifyRealmResultsCallbacks(HandlerController.java:303)
       at io.realm.HandlerController.notifySyncRealmResultsCallbacks(HandlerController.java:284)
       at io.realm.HandlerController.notifyTypeBasedListeners(HandlerController.java:275)
       at io.realm.HandlerController.notifyAllListeners(HandlerController.java:262)
       at io.realm.HandlerController.realmChanged(HandlerController.java:385)
       at io.realm.HandlerController.handleMessage(HandlerController.java:116)
       at android.os.Handler.dispatchMessage(Handler.java:98)
       at android.os.Looper.loop(Looper.java:148)
       at android.app.ActivityThread.main(ActivityThread.java:5417)
       at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:726)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:616)
```

Run the code, it crashes immediately. but it successfully log all action.
#### Code Sample

``` java

public class MainActivity extends AppCompatActivity {

    DynamicRealm realm;

    @Override

        RealmConfiguration realmConfig = new RealmConfiguration.Builder(this).build();
        realm = DynamicRealm.getInstance(realmConfig);

        realm.executeTransaction(new DynamicRealm.Transaction() {
            @Override
            public void execute(DynamicRealm realm) {
                // remove old schema, or we get exception for testing
                if (schema.contains("Person")) schema.remove("Person");

                // Create Person class with two fields: name and age
                schema.create("Person")
                        .addField("name", String.class)
                        .addField("age", int.class);

                // Fields with special properties
                schema.get("Person")
                        .addField("id", long.class, FieldAttribute.PRIMARY_KEY);

            }
        });

        realm.executeTransaction(new DynamicRealm.Transaction() {
            @Override
            public void execute(DynamicRealm realm) {
                // Creating new data during migrations
                DynamicRealmObject person = realm.createObject("Person");
                person.setString("name", "John");
            }
        });

        RealmResults<DynamicRealmObject> persons = realm.where("Person")
                .findAll();

        for (DynamicRealmObject ppl : persons) {
            Log.i("tag", ppl.getString("name"));
        }

        Log.i("tag", "isClosed = " + realm.isClosed());
        realm.executeTransaction(new DynamicRealm.Transaction() {
            @Override
            public void execute(DynamicRealm realm) {
                realm.getSchema().remove("Person");
                Log.i("tag", "schema removed, isClosed = " + realm.isClosed());
            }
        });
    }

    @Override
    protected void onDestroy() {
        super.onDestroy();
        realm.close();
    }
}

```

Realm version(s): 
classpath "io.realm:realm-gradle-plugin:1.0.0"

Android Studio version: 2.1.2

Which Android version and device: 
Android Emulator version 25.1.7, Version 6.0 - API 23 (Rev 1)

stk1m1:
`RealmResults` should be treated as an empty list when the underlying table is removed. Closes #2953.

The goal of this PR is to provide gracious treatments to `RealmResults` retrieved before their originating `Table` is removed.

``` java
// Give me some results
RealmObject<DynamicRealmObject> results1 = dynamicRealm.where("Person").findAll();

// Let's remove Table
dynamicRealm.executeTransaction(new DynamicRealm.Transaction() {
    @Override
    public void execute(DynamicRealm realm) {
        realm.getSchema().remove("Person");
    }
});

// Give me some results but this will fail
RealmObject<DynamicRealmObject> results2 = dynamicRealm.where("Person").findAll();
```
- The `results1` does not have connection/knowledge to the removal of its backing `Table` as `RealmSchema` does not notify. `RealmSchema` simply deletes a `Table` with `Group.removeTable()`
- `RealmResults` still has `TableView` points to the removed `Table` and that lack of update leads to ISE when any method is called on `TableView`. 
- Further, the caches in `RealmSchema` (`classToTable`, `dynamicClassToTable`) are not flushed after a `remove(className)` is called which leads to false sense of existance of `Table` in `RealmSchema`.

There are two approaches we can take.
1. When `Table` is removed, we track all the `TableView` linked and invalidate it. Unfortunately, I have not found a way to do this.
2. Make every methods in `RealmList` check existence of their backing `Table` and behave as an empty list if the table is not found.

For this PR, the 2nd approach is taken.


stk1m1:
**We should throw once we've found underlying table does not exist**

When `RealmSchema.remove(className)` is called, it removes underlying table. Then following happen.

1. Most of `RealmResults` & `RealmList` which belong to the removed  table do not work once backing `Table` is gone. 
2. Yet, subsequent query method such as `sort()` keeps running upon the removed table until it crashes somewhere down the path.

<sup>*</sup>At the same time, it is true users cannot _close_ `RealmResults` or `RealmList`.

stk1m1:
When `RealmSchema.remove(className)` is called, it removes underlying table. Then following happens.

- Most methods of `RealmResults` which belong to the removed  table do not work once backing `Table` is gone. 
- Yet, subsequent query method such as `sort()` keeps running upon the removed table until it crashes somewhere down the path.

Shouldn't we throw to prevent the 2nd scenario? At the same time, it is true that users cannot _close_ `RealmResults`.

Further, it might be costly to call `transaction.hasTable()` in every methods in `RealmResults` to make sure its originating `Table` exists. Would it be a good idea to check caches in `RealmSchema` and throw? Then it doesn't fail silently anymore.

This PR is overdue, but still in progress. Please share your view @realm/java

stk1m1:
retest this please

stk1m1:
retest this please

cmelchior:
Well you can call any method, not just sort?

We are already tracking all RealmResults in HandlerController. So we have a hook for iterating over all of them there. What about creating a no-op TableView that just works like a 0-size one, just like we do for Row when it is deleted? Would that work?

stk1m1:
retest this please

stk1m1:
retest this please

stk1m1:
This PR is ready as far as `RealmResults` is covered. The implementation itself isn't that much, but the semantic of it takes good amount of time to fiddle around to find the best candidate without sacrificing performance or compromising tested logic. With that said, here are few points I need your opinion.

1. The main focus is to have `RealmResults` behave as an empty list when its backing Class (`Table`, internally) is gone. When, for example, `RealmResults.size()` is called after its class eradicated, the result should return 0 instead of an exception. 
  - It is becuase there is no way users can `close` results,
  - and it provides a polished experience
2. However, it was challenging to maintain the proposition above when it is clearly required to fail. For example, when `RealmResults.get(int index)` is called, I think it is beneficial for users and us to throw. At this point, it simply return `null` whatever index is given.

  ```java
  // Give me some results
  RealmObject<DynamicRealmObject> results = dynamicRealm.where("Person").findAll();

  // Let's remove the class
  dynamicRealm.executeTransaction(new DynamicRealm.Transaction() {
      @Override
      public void execute(DynamicRealm realm) {
          realm.getSchema().remove("Person");
      }
  });

  // This will give us NPE, instead of ISE
  results2.get(0).getString("name")
  ```

3. The main appartus of an psuedo empty list is [`EMPTY_TABLEORVIEW`](https://github.com/realm/realm-java/pull/3035/files#diff-b910a6d8e7be35feac9789f4f3c7316dR390) instance. It provides a great deal of flexibility that could help us achieve the main proposition (#1). For example, when users have `RealmQuery` and `RealmResults` derived after the removal of a class, one can still perform the following, but he/she will not feel any pain other then getting void results.

  ```java
  RealmObject<DynamicRealmObject> results = dynamicRealm.where("Person").findAll();

  // Let's remove the class. This can happen somewhere in background thread as well.
  dynamicRealm.executeTransaction(new DynamicRealm.Transaction() {
      @Override
      public void execute(DynamicRealm realm) {
          realm.getSchema().remove("Person");
      }
  });  

  RealmObject<DynamicRealmObject> sorted = 
      results.where()
              .equalTo("name", "john")
              .findAllSorted("age");
  
  // should be 0
  sorted.size();              
  ```
Nevertheless, this scheme fails with an unexpected shortcoming where [`RealmFieldType getColumnType(long)`](https://github.com/realm/realm-java/pull/3035/files#diff-b910a6d8e7be35feac9789f4f3c7316dR435) return `null` and fails on [`switch (table.getColumnType(columnIndex))`](https://github.com/realm/realm-java/pull/3035/files#diff-cf75bf60257682afe04f9e8815af9a4fR438) with NPE. I've placed `isTablePresent()` everywhere to prevent NPE, and that's very untasteful.

4. Instead of `RealmSchema.contains(String className)`, [`TableView::is_attached()`](https://github.com/realm/realm-java/pull/3035/files#diff-676e5bdf1866092843787e07645d46eaR974) is added and used as the latter method provides a rather straight-forward check.
5. For multi-threaded class removal, i.e. when class is removed from other thread, it follows the path below.

  ```java
  RealmSchema.remove() ->    
  RealmCache.invalidateRemovedClassFromCachedRealm() ->
  BaseRealm.HandlerController.invalidateRemovedTableView() ->
  RealmResults.setTableEmpty()  
  ```
In building the path above, I've realized that there is no global `BaseRealm` instance thread across multiple threads. Hence, [`ArrayList<WeakReference<BaseRealm>> globalBaseRealmReference`](https://github.com/realm/realm-java/pull/3035/files#diff-79e6ca3beee0a83dd76772285b606694R105) is added.
  
6. All the Numerical value getters such as `sum`, `maxDate` returns `null` except `average` as the last one returns instance of `double` primitive.

*ToDo*
- ChangeListener behavior  
- `RealmQuery` as an empty query  
- RealmResults within ChangeListner and/or AsyncTransaction  
- RealmObject  

@realm/java please review

cmelchior:
More in-depth review following, but the key points:

- `results.get(0)` should throw a IndexOutOfBoundsException. That would be the proper semantics for a empty collection.

- IMO we should trigger the changelistener if we detect this. It should be considered a change like everything else. `isValid` should be false.

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ### Bug fixes
 
+* Fixed a crash that did not treat an existing `RealmResult` as an empty list when its originating `RealmObject` class was removed (#2953).
```
cmelchior: * Fixed a crash caused by RealmResults if the underlying schema definition is removed. It is now instead treated as a empty list instead.

realm/realm-library/src/androidTest/java/io/realm/RealmResultsTests.java
```diff
@@ -1027,4 +1029,459 @@ public void deleteAndDeleteAll() {
...
+        // make sure we contains an object in RealmResults
+        RealmResults<DynamicRealmObject> results = realm.where(CLASS_NAME).findAll();
+        DynamicRealmObject object = realm.where(CLASS_NAME).findFirst();
+        assertTrue(results.contains(object));
+
+        // remove class in background
+        new Thread(new Runnable() {
+            @Override
+            public void run() {
+                DynamicRealm bgRealm = DynamicRealm.getInstance(realmConfig);
```
cmelchior: I would use `realm.getConfiguration()` instead. Right now it isn't clear that these two Realms have the same config.

realm/realm-library/src/androidTest/java/io/realm/RealmResultsTests.java
```diff
@@ -1027,4 +1029,459 @@ public void deleteAndDeleteAll() {
...
+        // Make sure we have expected values
+        RealmResults<DynamicRealmObject> results = realm.where(CLASS_NAME).findAll();
+        assertEquals(results.get(0).getString("name"),  "John");
+        assertEquals(results.first().getString("name"), "John");
+        assertEquals(results.last().getString("name"),  "John");
+
+        // remove the backing class
+        removeClassFromDynamicRealm(realm, CLASS_NAME);
+        // indexed get
+        assertNull(results.get(0));
```
cmelchior: Should throw IndexOutOfBounds instead, just like `first()`

realm/realm-library/src/androidTest/java/io/realm/RealmResultsTests.java
```diff
@@ -1027,4 +1029,459 @@ public void deleteAndDeleteAll() {
...
+                DynamicRealm bgRealm = DynamicRealm.getInstance(realmConfig);
+                removeClassFromDynamicRealm(bgRealm, CLASS_NAME);
+                bgRealm.close();
+                bgRealmClosed.countDown();
+            }
+        }).start();
+
+        TestHelper.awaitOrFail(bgRealmClosed);
+        // indexed get
+        assertNull(results.get(0));
```
cmelchior: Same

realm/realm-library/src/main/java/io/realm/HandlerController.java
```diff
@@ -786,4 +786,23 @@ public boolean isAutoRefreshEnabled() {
...
     }
 
+    /**
+     * When a Class is removed while {@link RealmResults} derived from the class exist, it is required
+     * for the RealmResults to be reconciled with the removal by voiding their backing TableView.
+     * This should be called in a propagated event from {@link RealmCache#invalidateRemovedClassFromCachedRealm}.
+     *
+     * @param clazzName to invalidate the {@link RealmResults} from which are derived.
+     */
+    void invalidateRemovedTableView(String clazzName) {
```
cmelchior: className

realm/realm-library/src/main/java/io/realm/HandlerController.java
```diff
@@ -786,4 +786,23 @@ public boolean isAutoRefreshEnabled() {
         return autoRefresh;
     }
 
+    /**
+     * When a Class is removed while {@link RealmResults} derived from the class exist, it is required
```
cmelchior: When the schema defintion for a Class is removed while a RealmResults is still present, the underlying TableView should be replaced by an always empty tableview instead.

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -70,6 +73,16 @@ static RealmCacheType valueOf(Class<? extends BaseRealm> clazz) {
...
 
+    /**
+     * This is a global reference of all the Realm/DynamicRealm instances across thread. As {@link #refAndCountMap}
+     * allows its {@link #refAndCountMap#localRealm} to yield {@link BaseRealm} instance to a specific thread,
+     * it is necessary to have a global list of BaseRealm that could be approached from any thread.
+     *
+     * <p>As one can see there is no protection on this list from being accessed via multiple threads.
+     * Please take caution when this is exposed to un-synchronized methods.
+     */
+    private final ArrayList<WeakReference<BaseRealm>> globalBaseRealmReference;
```
cmelchior: I don't think we need this. See e.g. how `Realm.deleteRealm()` access the global count?

realm/realm-library/src/main/java/io/realm/RealmResults.java
```diff
@@ -154,6 +154,50 @@ TableOrView getTable() {
...
+        return tableExists;
+    }
+
+    /**
+     * This is to make {@link RealmResults#table} as an empty list with {@link TableOrView#EMPTY_TABLEORVIEW}
+     * The benefits of setting dummy TableOrView are that it absorbs operations as users proceed,
+     * saves travels to the native layer to see if the corresponding {@link Table} still exists, and
+     * propagates same effects on {@link RealmQuery} with ease.
+     */
+    void setTableEmpty() {
```
cmelchior: `set` sounds to simple I would probably use `convertToEmptyTableView()`.

realm/realm-library/src/main/java/io/realm/RealmResults.java
```diff
@@ -200,6 +244,9 @@ public boolean contains(Object object) {
     public E get(int location) {
         E obj;
         realm.checkIfValid();
+        if (!isTablePresent()) {
```
cmelchior: Remove this

realm/realm-library/src/main/java/io/realm/RealmResults.java
```diff
@@ -241,6 +288,9 @@ public E last() {
     @Override
     public void deleteFromRealm(int location) {
         realm.checkIfValid();
+        if (!isTablePresent()) {
```
cmelchior: Remove this. Should be caught by `table.remove()`

realm/realm-library/src/main/java/io/realm/RealmResults.java
```diff
@@ -154,6 +154,50 @@ TableOrView getTable() {
     }
 
     /**
+     * Check if the underlying {@link Table} exists.
+     *
+     * @return {@code true} if the backing Table is valid, {@code false} otherwise.
+     */
+    private boolean isTablePresent() {
```
cmelchior: I would somehow cache the result of this. Right now you are introducing an extra JNI call to every method in this class.

realm/realm-library/src/main/java/io/realm/internal/TableOrView.java
```diff
@@ -373,4 +373,322 @@
...
+     *
+     * @return {@code true} if the originating {@link Table} is alive, {@code false} otherwise.
+     */
+    boolean isAttached();
+
+    /**
+     * This is a void manifestation of {@link TableOrView} that takes calls and does nothing or returns
+     * {@code null} value to imitate an empty {@link TableView}.
+     */
+    TableOrView EMPTY_TABLEORVIEW = new TableOrView() {
```
cmelchior: Crazy idea, but before a class is removed save a snapshot of the column names/field types/field indices. It would mean that this class would still respond to basic queries like column name/index but otherwise behave as an empty tableview.

It would also mean you could remove all `tableExist()` calls in RealmResults.

Downside is that we need to instantiate a copy of this class for each type being removed. I think that is acceptable.

stk1m1: > Downside is that we need to instantiate a copy of this class for each type being removed. I think that is acceptable.

That's what I was looking for. Will do. 👍

realm/realm-library/src/main/java/io/realm/RealmResults.java
stk1m1: If this is to be removed, it goes to `BaseRealm.get()` and tries to get the removed `Table`. Is it fine?

stk1m1:
**Review points reflected**

- `EmptyTableView` captures fieldNames, fieldTypes, and columnCount to provide basic information of a removed Table.
- There is only one `EmptyTableView` instance in a process at any given time for a removed `Table`to avoid performance hit and discrepancy between Realms in different threads.
- `RealmResults.convertToEmptyTableView()` sets `.table` field to an `EmptyTableView` as well as `asyncQueryCompleted` with `true` to prevent  a `RealmResults` from being updated by `HandlerController`.
- With `isLoaded()` returning true, sync/async RealmResults will deliver a change event through change listener in a looper thread as an empty list.
- `isTablePresent()` is renamed `isTableRemoved()` and only checks if `.table` field is an instance of `EmptyRealmResults` which quickly provides an indication that a RealmResults is derived a deleted Table.
- With the faster `isTableRemoved()`, few methods, `distinct()`, `contains()`, `get()`, `addChangeListener()`, checks and behaves like an empty list with `OutOfIndexException`. 
- The major reason for those methods to have `isTableRemoved()` is that they either goes into `Realm` instance to have a direct access to `Table` or uses methods existing only on `Table`. It is simply awkward to get around for `EmptyRealmResults` to absorb all the operations.

stk1m1:
**Review points reflected**

1. `EmptyTableView` captures fieldNames, fieldTypes, and columnCount to provide basic information of a removed Table.
2. There is only one `EmptyTableView` instance in a process at any given time for a removed `Table` to avoid performance hit and discrepancy between Realms in different threads.
3. `RealmResults.convertToEmptyTableView()` sets `.table` field to an `EmptyTableView` as well as `asyncQueryCompleted` with `true` to prevent  a `RealmResults` from being updated by `HandlerController`.
4. With `isLoaded()` returning true, sync/async RealmResults will deliver a change event through change listener in a looper thread as an empty list.
5. `isTablePresent()` is renamed `isTableRemoved()` and only checks if `.table` field is an instance of `EmptyRealmResults` which quickly provides an indication that a RealmResults is derived from a deleted Table.
6. With the faster `isTableRemoved()`, few methods, `distinct()`, `contains()`, `get()`, `addChangeListener()`, uses it to check and behave like an empty list.
  - The major reason for those methods to have `isTableRemoved()` is that they either goes into `Realm` instance to have a direct access to `Table` or uses methods existing only on `Table`. It is simply awkward to get around for `EmptyRealmResults` to absorb all the operations.

@realm/java please review

stk1m1:
After some thoughts, I've realized that there is a flaw in converting RealmResults from HandlerController in Looper thread.

In the following line, `syncRealmResults` and `asyncRealmResults` are accessed and their member RealmResults are converted, bypassing (w/o sending a message to) `Handler`. This could cause an unexpected race condition as I wasn't able to get my head around the mechanism fully. 

```java
void invalidateRealmResults(final EmptyTableView emptyTableView) {
    convertResultToEmptyTableView(syncRealmResults.keySet().iterator(), emptyTableView);
    convertResultToEmptyTableView(asyncRealmResults.keySet().iterator(), emptyTableView);
}
```

stk1m1:
@beeender quotes,

> I don't like to check the local schema because of:
  1. It couples RealmResults with RealmSchema, which is not necessary
  2. It won't have complex timing sequence if we want to support multi-processes

> So I am thinking about:
  1. check the `TableView.is_valid()` when change event arrive
  2. depends on if we want to handle the case in the same loop, maybe check the local RealmResults list to see if it needs to be set as empty table view

> but anyway, for above 2), if user is doing that, he is doing wrong.

beeender:
I will finish this on behalf of @stk1m1 😄 After #3834 merged, this should be implemented in OS side and adding a few test cases in java side.

cmelchior:
Outdated and will be made possible through after #3834. Closing

PR #3039, opened 2016-06-21
Clean up unused sub-table

beeender:
sub-table is not used in java-binding right now, and if it is going to be, the relevant logic might be more suitable in the object store.

Since we are moving to the object store, the sub-table related code needs to be cleaned anyway. So we do the clean up directly to the master branch to make the future PR easier to be reviewed.


realm/realm-jni/src/io_realm_internal_TableView.cpp
```diff
@@ -1035,8 +999,7 @@ JNIEXPORT jlong JNICALL Java_io_realm_internal_TableView_nativeWhere(
         if (!VIEW_VALID_AND_IN_SYNC(env, nativeViewPtr))
             return 0;
 
-        Query query = TV(nativeViewPtr)->get_parent().where(TV(nativeViewPtr));
-        TableQuery* queryPtr = new TableQuery(query);
+        Query *queryPtr = new Query(std::move(TV(nativeViewPtr)->get_parent().where(TV(nativeViewPtr))));
```
cmelchior: Any reason for changing this from a TableQuery to Query?

realm/realm-jni/src/io_realm_internal_TableQuery.cpp
```diff
@@ -38,7 +37,7 @@ inline bool query_valid(JNIEnv* env, Query* pQuery)
 
 inline bool query_col_type_valid(JNIEnv* env, jlong nativeQueryPtr, jlong colIndex, DataType type)
 {
-    return TBL_AND_COL_INDEX_AND_TYPE_VALID(env, TQ(nativeQueryPtr)->get_current_table().get(), colIndex, type);
+    return TBL_AND_COL_INDEX_AND_TYPE_VALID(env, Q(nativeQueryPtr)->get_table().get(), colIndex, type);
```
cmelchior: Same, why this change?

realm/realm-jni/src/io_realm_internal_TableView.cpp
beeender: `TableQuery` is created only for JNI and to wrap around the subtable. And it is deleted in this PR as well.

cmelchior:
Looks 👍 
But why the change from TableQuery to Query?

realm/realm-jni/src/io_realm_internal_TableView.cpp
cmelchior: 👍

realm/realm-jni/src/io_realm_internal_TableQuery.cpp
stk1m1: I think it is b/c TableQuery is to be removed. In TQ, we have

```c++
realm::TableRef get_current_table() {
    realm::TableRef table = get_table();

    // Go through the stack of subtables to find current subtable (if any)
    size_t size = subtables.size(); 
    for (size_t i = 0; i < size; ++i) {
        size_t index = subtables[i];
        table = table->get_subtable(index, 0);
    }
    return table;
}
```

<sup>*</sup>There is _finding current table_ logic

kneth: Yes, when you query a subtable, you will need to get a reference to it. `get_current_table()` descents down to the subtable as the query only has access to the parent table.

realm/realm-jni/src/io_realm_internal_TableView.cpp
kneth: Well, `TableQuery` was also introduced so we could do a deep copy of the query. @beeender are you certain that the core's `Query` constructor does a deep copy?

kneth:
@cmelchior `TableQuery` was partly introduced to keep track of subtable queries (balancing begin and end markers).

realm/realm-jni/src/io_realm_internal_TableView.cpp
beeender: It is calling move constructor and confirmed with core, it should work :tada:

stk1m1:
Other than what @kneth pointed, 👍

bmunkholm:
@beeender Please include motivation in the description besides just the title.
It's not clear that it has been confirmed that we are not going to need subtables for any of the other future datatypes we want to support.

beeender:
@bmunkholm PR message updated.

realm/realm-jni/src/io_realm_internal_LinkView.cpp
```diff
@@ -178,8 +177,7 @@ JNIEXPORT jlong JNICALL Java_io_realm_internal_LinkView_nativeWhere
     try {
         LinkViewRef *lv = LV(nativeLinkViewPtr);
         LinkViewRef lvr = *lv;
-        Query query = lvr->get_target_table().where(LinkViewRef(lvr));
-        TableQuery* queryPtr = new TableQuery(query);
+        Query *queryPtr = new Query(std::move(lvr->get_target_table().where(LinkViewRef(lvr))));
```
bdash: The use of `std::move` here is unnecessary since `lvr->get_target_table().where(LinkViewRef(lvd))` is already an rvalue.

realm/realm-jni/src/io_realm_internal_Table.cpp
```diff
@@ -1159,8 +1065,7 @@ JNIEXPORT jlong JNICALL Java_io_realm_internal_Table_nativeWhere(
     if (!TABLE_VALID(env, TBL(nativeTablePtr)))
         return 0;
     try {
-        Query query = TBL(nativeTablePtr)->where();
-        TableQuery* queryPtr = new TableQuery(query);
+        Query *queryPtr = new Query(std::move(TBL(nativeTablePtr)->where()));
```
bdash: Ditto about `std::move` being unnecessary.

PR #3041, opened 2016-06-21
Remove unused mixed type

beeender:
The mixed type is not used by now and it is very unlikely going to be
used in the future.


cmelchior:
Some RealmProcessor unit tests seems to fail, but apart from that 👍

realm/realm-jni/src/tablebase_tpl.hpp
```diff
@@ -17,14 +17,15 @@
 #ifndef REALM_JNI_TABLEBASE_TPL_HPP
 #define REALM_JNI_TABLEBASE_TPL_HPP
 
+#include <realm.hpp>
 
 template <class T>
 jbyteArray tbl_GetByteArray(JNIEnv* env, jlong nativeTablePtr, jlong columnIndex, jlong rowIndex)
```
stk1m1: is it possible to inline this as this isn't used in many places?

stk1m1:
👍

cmelchior:
👍

kneth:
:+1:

PR #3044, opened 2016-06-22
Add cmake to docker

beeender:


stk1m1:
I don't think that I can bear the weight of the significance this particular request brings to us. I'd simply like to applause and enchant your achievement here 💯

PR #3049, opened 2016-06-22
Save LogCat output if unit tests fail.

cmelchior:
If unit tests crashes with native exceptions we currently have no way of knowing why.
This change means that we now save the logcat output from in case of unit test failures.

A local test on my phone indicates that a full logcat file is ~8MB in size and 400Kb when zipped.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -93,15 +93,14 @@
 }
 
 def startLogCatCollector() {
-    sh '''rm -f logcat.txt
-    adb logcat -c
+    sh '''adb logcat -c
     adb logcat > "logcat.txt" &
     '''
 }
 
 def stopLogCatCollector(boolean archiveLog) {
     sh '''jobs
-       kill %1'
+       kill %1
        '''
     if (archiveLog) {
         zip([
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -107,7 +107,7 @@ def stopLogCatCollector(String backgroundPid, boolean archiveLog) {
         zip([
             'zipFile': 'logcat.zip',
             'archive': true,
-            'glob' : 'logcat.txt'
+            'glob' : 'realm/logcat.txt'
         ])
     }
     sh 'rm logcat.txt '
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -102,7 +102,7 @@ def String startLogCatCollector() {
 }
 
 def stopLogCatCollector(String backgroundPid, boolean archiveLog) {
-    sh 'kill $backgroundPid'
+    sh "kill ${backgroundPid}"
     if (archiveLog) {
         zip([
             'zipFile': 'logcat.zip',
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -107,7 +107,7 @@ def stopLogCatCollector(String backgroundPid, boolean archiveLog) {
         zip([
             'zipFile': 'logcat.zip',
             'archive': true,
-            'glob' : 'realm/logcat.txt'
+            'glob' : '**/logcat.txt'
         ])
     }
     sh 'rm logcat.txt '
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -57,8 +57,9 @@
 
                 stage 'Run instrumented tests'
                 try {
+                    sh 'cd realm'
                     startLogCatCollector()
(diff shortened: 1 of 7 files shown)


cmelchior:
CI: Retest this please

Jenkinsfile
```diff
@@ -87,6 +92,27 @@
     }
 }
 
+def startLogCatCollector() {
+    sh '''rm -f logcat.txt
```
emanuelez: this is not needed since a `git clean -xfd` is ran after checkout

Jenkinsfile
```diff
@@ -87,6 +92,27 @@
...
+    adb logcat -c
+    adb logcat > "logcat.txt" &
+    '''
+}
+
+def stopLogCatCollector(String pid, boolean archiveLog) {
+    sh '''jobs
+       kill %1'
```
emanuelez: I suppose `jobs` is just for debugging purposes, right?
if you use double quotes you will be able to use the `pid` argument

Jenkinsfile
```diff
@@ -87,6 +92,27 @@
...
+       '''
+    if (archiveLog) {
+        zip([
+            'zipFile': 'logcat.zip',
+            'archive': true,
```
emanuelez: true -> archivelog

cmelchior: You mean `'archive' : 'archivelog',` ? 
This https://jenkins.io/doc/pipeline/steps/pipeline-utility-steps/ says it is a boolean?

beeender: if it take `archiveLog` then `if (archiveLog)` is not needed i think?

beeender:
can we get a download link to it :8ball:

Jenkinsfile
emanuelez: oops... missed that. sorry

Jenkinsfile
```diff
@@ -87,6 +92,26 @@
     }
 }
 
+def startLogCatCollector() {
+    sh '''adb logcat -c
+    adb logcat > "logcat.txt" &
+    '''
+}
+
+def stopLogCatCollector(boolean archiveLog) {
```
emanuelez: I think we should use a PID. this version is fragile as other poeple might be logged into the host and jobs might return unexpected results

Jenkinsfile
emanuelez: no need to zip if everything passes. We should keep it as it is

Jenkinsfile
cmelchior: For some reason I didn't get the PID in the previous version. No idea why?

Jenkinsfile
```diff
@@ -87,6 +93,25 @@
...
 
+def String startLogCatCollector() {
+    sh '''adb logcat -c
+    `adb logcat > "logcat.txt" &` & echo $1 > pid
+    '''
+    return readFile("pid").trim()
+}
+
+def stopLogCatCollector(String backgroundPid, boolean archiveLog) {
+    sh 'kill $backgroundPid'
```
emanuelez: I hope SIGTERM will be enough in all cases. But I would of course prefer not to use SIGKILL.

Jenkinsfile
```diff
@@ -87,6 +93,25 @@
     }
 }
 
+def String startLogCatCollector() {
```
emanuelez: it's not very Groovysh to specify the return type, but I guess it doesn't hurt

Jenkinsfile
```diff
@@ -87,6 +93,26 @@
...
+    return readFile("pid").trim()
+}
+
+def stopLogCatCollector(String backgroundPid, boolean archiveLog) {
+    sh "kill ${backgroundPid}"
+    if (archiveLog) {
+        zip([
+            'zipFile': 'logcat.zip',
+            'archive': true,
+            'glob' : '**/logcat.txt'
```
emanuelez: such a wildcard can potentially make looking for the file very slow. We know where the file is so let's use the proper path

cmelchior: Yeah, but it didn't find the file for some reason when using `realm/logcat.txt` ... 😢

emanuelez:
See? That was easy! :)

cmelchior:
Wrong branch. Moving this to `releases`

PR #3052, opened 2016-06-22
Remove redundant calls to std::move

bdash:
`std::move` casts an lvalue to an rvalue. It's redundant to call it with an rvalue as its argument.

/cc @beeender


beeender:
Thanks! :+1:

cmelchior:
👍

kneth:
:+1:

PR #3056, opened 2016-06-23
Fix the adb key for docker

beeender:


emanuelez:
Instamerge! Already tested :)

PR #3058, opened 2016-06-23
Save LogCat file in case of test failure

cmelchior:
Move from https://github.com/realm/realm-java/pull/3049

If unit tests crashes with native exceptions we currently have no way of knowing why.

With this PR we now save the LogCat file in case there is a test failure. For green builds nothing is saved.

A local test on my phone indicates that a full logcat file is ~8MB in size and 400Kb when zipped.

The CI log seems to be much smaller (~4Mb compressed to 130k) 
An example is here: https://ci.realm.io/job/android-pr-jenkinsfile/630/

We need to enable logging before starting tests as Android only have a 256Kb buffer for log entries and rotates between multiple files. This means that dumping the log afterwards are missing _a lot_ of entries.

@realm/java 
@emanuelez

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -93,15 +93,14 @@
 }
 
 def startLogCatCollector() {
-    sh '''rm -f logcat.txt
-    adb logcat -c
+    sh '''adb logcat -c
     adb logcat > "logcat.txt" &
     '''
 }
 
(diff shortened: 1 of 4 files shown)


emanuelez:
👍

stk1m1:
👍

Issue #3051, opened 2016-06-22
Build failure during initial build due to android-23 SDK not being installed


bdash:
#### Goal

Build realm-java by following instructions in README.

#### Expected Results

Success!

#### Actual Results

Java exception thrown about android-23 not being found in ~/Library/Android/sdks when running `./gradlew assemble`. It did not happen when I reran the build command, and I neglected to save the exception message 😞If I can figure out which local state I can clear out to give me a clean slate to reproduce this from I'll provide more detail.

android-23 is not mentioned as required in realm-java's README, and doesn't seem to really be required to build.

#### Version of Realm and tooling
Realm version(s): 76c6178c01ff62c69d95543b219a4c23f5a23873
Android Studio version: 2.1.2

stk1m1:
Hi @bdash 

Thanks for reporting the issue to us with the template filled! Nice of you! 

It is nevertheless our short-coming to overlook clearly stating the requirements, Android-sdk v23 is nevertheless required as it is specified in [here](https://github.com/realm/realm-java/blob/master/realm/realm-library/build.gradle#L14). 

Realm-java build system is constructed with [Gradle](http://gradle.org/) and it allows us to have this beautiful, compartmentalized, and hierarchical build requirement tree such that it is possible to construct each sub-project (module in Gradle) to have their own per-requisites. It might not be obvious from the first look though.

I'll escalate this issue and queued for proper update! Thanks!

zaki50:
README.md has been fixed in #3062 (releases branch) and #3074 (master branch).

Thanks @bdash

PR #3062, opened 2016-06-23
Update compileSdkVersion and build tools to 24

zaki50:
refs #3051

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -32,8 +32,8 @@ RUN cd /opt && rm -f android-sdk.tgz
 # Grab what's needed in the SDK
 RUN echo y | android update sdk --no-ui --all --filter platform-tools | grep 'package installed'
 RUN echo y | android update sdk --no-ui --all --filter extra-android-support | grep 'package installed'
-RUN echo y | android update sdk --no-ui --all --filter android-23 | grep 'package installed'
-RUN echo y | android update sdk --no-ui --all --filter build-tools-23.0.0 | grep 'package installed'
+RUN echo y | android update sdk --no-ui --all --filter android-24 | grep 'package installed'
+RUN echo y | android update sdk --no-ui --all --filter build-tools-24.0.0 | grep 'package installed'
 
 # Install the NDK
 RUN mkdir /opt/android-ndk-tmp
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,10 +30,11 @@ RUN cd /opt && tar -xvzf android-sdk.tgz
 RUN cd /opt && rm -f android-sdk.tgz
 
 # Grab what's needed in the SDK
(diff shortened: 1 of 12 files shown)


zaki50:
@realm/java Should we keep `targetSdkVersion` of `realm-library` to `23` in `releases` branch for better compatibility to `1.0.1`?

examples/build.gradle
```diff
@@ -1,5 +1,5 @@
-project.ext.sdkVersion = 23
-project.ext.buildTools = '23.0.0'
+project.ext.sdkVersion = 24
```
dalinaum: Because most devices cannot support SDK version 24 yet, I don't know that using SDK version 23 is a good idea.

cmelchior: The SDK version should not have any effect on what is actual running on the device. This is just used to compile everything with.

cmelchior:
@zaki50 Do you know of any problems between 23.0.0 and 24.0.0 that might effect us? Otherwise I would say it should be safe to upgrade.

kneth:
Yes, I think we should keep `23` in `releases`.

zaki50:
@cmelchior I don't think that effects for us. But it has some effect for the users.
Realm's `targetSdkVersion` is merged into app's AndroidManifest.xml and then app's behavior might change.

I'll keep `targetSdkVersion` to `23` on `releases` branch.

cmelchior:
Ah yes, good point 👍

Dockerfile
```diff
@@ -30,10 +30,11 @@ RUN cd /opt && tar -xvzf android-sdk.tgz
 RUN cd /opt && rm -f android-sdk.tgz
 
 # Grab what's needed in the SDK
+RUN echo y | android update sdk --no-ui --all --filter tools > /dev/null
```
emanuelez: can you explain why this different pattern to install `tools`? Maybe a comment will help not changing it to be like the others by mistake.

zaki50: Right.

The story is a bit weird.

Running `RUN echo y | android update sdk --no-ui --all --filter tools` produces following output.
```
...
Installing Archives:
  Preparing to install archives
  Skipping 'Android SDK Tools, revision 25.1.7'; it depends on 'Android SDK Platform-tools, revision 24' which was not installed.
  Done. Nothing was installed.
```

This says that install was skipped and nothing was installed. But, if I remove this line, the lint check does not pass.

@emanuelez Do you know why this happens?

cmelchior: They made some changes to the command line tool in the last versions, it might not even work anymore....

kneth:
LGTM

zaki50:

dalinaum:
LGTM

cmelchior:
👍

Issue #1625, opened 2015-10-20
Cache fieldType alongside fieldIndex


cmelchior:
Copy from #1610 

```
RealmQuery.java Line 161

columns.get(fieldName) called potentially 3 times.

            if (columns.get(fieldName) == null) {
                throw new IllegalArgumentException(String.format("Field '%s' does not exist.", fieldName));
            }

            ColumnType tableColumnType = table.getColumnType(columns.get(fieldName));
            if (fieldType != null && fieldType != tableColumnType) {
                throw new IllegalArgumentException(String.format("Field '%s': type mismatch. Was %s, expected %s.",
                        fieldName, fieldType, tableColumnType
                ));
            }
            return new long[] {columns.get(fieldName)};

```

Also, why does getColumnType need to call native jni to get column type every single time and not cached? Strange this has to be done every time since column index is cached but column data type is not.

cmelchior:
Folding this into #3066

Issue #1489, opened 2015-09-22
Format JNI source code


beeender:
Some rules below:
1) JNI headers should be generated with gradle
2) JNI cpp files name should be like `io_realm_internal_ClassName.cpp`
3) Vars should be in forms of `underscore_case`
4) Use terms `RealmObject` and `RealmList` instead of `link` and `link list`. In general avoid using words like column/table/link/linklist if the error will reach users. All these concepts are core concepts and not exposed to end users.
5) `catch` and `CATCH_STD` should be in the new line.
6) Lines should never exceed **118** characters (https://github.com/realm/realm-core/blob/master/doc/development/coding_style_guide.cpp#L6)
7) JNI function definition:

``` c++
JNIEXPORT jlong JNICALL // This 3 things at the first line
Java_io_realm_internal_LinkView_nativeGetFinalizer(JNIEnv *, jclass)
{
    TR_ENTER()
    return static_cast<jlong>(reinterpret_cast<uintptr_t>(&finalize_link_view));
}
```

8) Others follow core's style guide https://github.com/realm/realm-core/blob/master/doc/development/coding_style_guide.cpp
9) return NULL instead of 0 for invalid pointer
10) Define a const value for -1 to return for invalid index/version

nhachicha:
as commented [here](https://github.com/realm/realm-java/pull/1214#discussion_r40843188)
find a consistent way of using `TR_ENTER` & it's variants

cmelchior:
Also, 
- comments for methods
- general code style like e.g. bracers

nhachicha:
`return NULL` instead of 0 or -1 for pointer (in case of error) 
-1 should only be used for invalid indexes

kneth:
We are now using `clang-format` and header files are always generated by CMake/Gradle.

PR #3066, opened 2016-06-24
Integrate Object Store [PART3] - Property, ObjectSchema and Schema

kneth:
Depends on branch `mc/feature/os-shared-realm`

_This is very much work-in-progress_


realm/realm-library/src/main/cpp/io_realm_RealmObjectSchema.cpp
```diff
@@ -0,0 +1,122 @@
...
+
+JNIEXPORT void JNICALL
+Java_io_realm_internal_RealmObjectSchema_nativeRemovePropertyByName
+(JNIEnv *env, jclass, jlong native_ptr, jstring name) {
+    TR_ENTER_PTR(native_ptr)
+    auto* object_schema = reinterpret_cast<ObjectSchema*>(native_ptr);
+    try {
+        JStringAccessor str(env, name);
+        auto iterator = object_schema->persisted_properties.begin();
+        while (iterator != object_schema->persisted_properties.end()) {
```
bdash: This loop looks to be a reimplementation of `std::remove_if`. Use it instead?

realm/realm-library/src/main/cpp/io_realm_RealmSchema.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+JNIEXPORT void JNICALL
+io_realm_RealmSchema_nativeRemoveObjectSchemaByName(JNIEnv *env, jclass, jlong native_schema_ptr, jstring j_name)
+{
+    TR_ENTER_PTR(native_schema_ptr)
+    auto *schema = reinterpret_cast<Schema*>(native_schema_ptr);
+    try {
+        JStringAccessor name(env, j_name);
+        auto object_schema = schema->find(name);
+        schema->erase(object_schema);
```
bdash: If `name` is not found, `object_schema` will be the `end()` iterator. It's not valid to call `erase` with the end iterator. Some form of validation here could prevent subtle issues stemming from this.

realm/realm-library/src/main/cpp/io_realm_RealmSchema.cpp
```diff
@@ -0,0 +1,101 @@
...
+}
+
+JNIEXPORT jlong JNICALL
+io_realm_RealmSchema_nativeGetObjectSchemaByName(JNIEnv *env, jclass, jlong native_schema_ptr, jstring j_name)
+{
+    TR_ENTER_PTR(native_schema_ptr)
+    auto *schema = reinterpret_cast<Schema*>(native_schema_ptr);
+    try {
+        JStringAccessor name(env, j_name);
+        auto object_schema_iterator = schema->find(name);
```
bdash: It's somewhat conventional to use the name `it` for this purpose. Whichever name you settle on should likely also be used in the functions earlier in this file, rather than having `object_schema` sometimes be an iterator and other times an `ObjectSchema`.

realm/realm-library/src/main/cpp/io_realm_RealmSchema.cpp
```diff
@@ -0,0 +1,101 @@
...
+        if (object_schema_iterator == schema->end()) {
+            return nullptr;
+        }
+        else {
+            auto object_schema = *object_schema_iterator;
+            return reinterpret_cast<jlong>(&object_schema);
```
bdash: This code creates a _copy_ of the `ObjectSchema` on the stack, and returns the address of this stack copy. `auto&` is needed to have `object_schema` refer to the object in the vector.

realm/realm-library/src/main/cpp/object-store/src/schema.hpp
```diff
@@ -51,6 +51,8 @@ class Schema : private std::vector<ObjectSchema> {
     using base::end;
     using base::empty;
     using base::size;
+    using base::push_back;
```
bdash: I think the intent is for `Schema` to be immutable after it is created.

realm/realm-library/src/main/cpp/io_realm_RealmSchema.cpp
```diff
@@ -0,0 +1,101 @@
...
+    auto *schema = reinterpret_cast<Schema*>(native_ptr);
+    delete schema;
+}
+
+JNIEXPORT void JNICALL
+io_realm_RealmSchema_nativeAddObjectSchema(JNIEnv *env, jclass, jlong native_schema_ptr, jlong native_object_schema_ptr)
+{
+    TR_ENTER_PTR(native_schema_ptr)
+    auto *schema = reinterpret_cast<Schema*>(native_schema_ptr);
+    auto object_schema = *reinterpret_cast<ObjectSchema*>(native_object_schema_ptr);
```
bdash: This creates a copy of `ObjectSchema` which is then copied again when passed to `push_back` below. One of these copies should be eliminated (make the local a `const auto&`, or `std::move` it into `push_back`).

realm/realm-library/src/main/cpp/object-store/src/schema.hpp
tgoyne: `push_back()` in particular is not a sensible thing to expose directly as the ObjectSchema are stored in sorted order.

realm/realm-library/src/main/cpp/io_realm_RealmSchema.cpp
kneth: I might have assumed it was validated in Java but I agree that it might too easy to make mistakes.

realm/realm-library/src/main/cpp/io_realm_RealmSchema.cpp
kneth: Thanks for pointing out. We have https://github.com/realm/realm-java/issues/1489 and you are welcome to add a few comments :-)

realm/realm-library/src/main/cpp/object-store/src/schema.hpp
kneth: I'll revert it and change how the schema is processed in the binding instead.

realm/realm-library/src/main/cpp/io_realm_RealmSchema.cpp
kneth: Thanks for pointing out.

Zhuinden:
I'm guessing this one doesn't really have anything to do with Metadata tables, right?

kneth:
@Zhuinden Not really. I hope that I can return to this soon so we can get a full integration with object store.

cmelchior:
FYI: I folded https://github.com/realm/realm-java/issues/1625 into this

beeender:
#4887 merged

PR #3094, opened 2016-06-30
Fix lint checks in examples.

cmelchior:
Due to our gradle project structure, when one module pulls in Timber, the lint check is applied to all (looks like a bug in how Lint is run). This caused Lint to fail and have prevented releasing SNAPSHOTs from `master`

This PR fixes all lint checks across all the examples.

@realm/java


examples/moduleExample/app/src/main/res/values/strings.xml
```diff
@@ -2,5 +2,6 @@
 <resources>
```
stk1m1: Extra line.

stk1m1:
👍

kneth:
LGTM

PR #3100, opened 2016-07-01
Make findXxxWithHandover static

beeender:
I am not quite sure why they are like those at the first place. But
They don't seem to be necessary at all.

Those methods don't have to be member fuctions. Especially lock to
dispose resources might block caller thread.

### DIFF
--- a/realm/realm-library/src/main/java/io/realm/internal/TableQuery.java
+++ b/realm/realm-library/src/main/java/io/realm/internal/TableQuery.java
@@ -420,14 +420,10 @@ public class TableQuery implements Closeable {
      * Performs a find query then handover the resulted Row (ready to be imported by another thread/shared_group).
      *
      * @param bgSharedGroupPtr current shared_group from which to operate the query.
-     * @param nativeReplicationPtr replication pointer associated with the shared_group.
      * @param ptrQuery query to run the the find against.
      * @return pointer to the handover result (table_view).
      */
-    public long findWithHandover(long bgSharedGroupPtr, long nativeReplicationPtr, long ptrQuery) {
-        validateQuery();
-        // Execute the disposal of abandoned realm objects each time a new realm object is created
-        context.executeDelayedDisposal();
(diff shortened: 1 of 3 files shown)


realm/realm-library/src/main/java/io/realm/internal/TableQuery.java
```diff
@@ -462,31 +458,19 @@ public TableView findAll() {
     // handover find* methods
     // this will use a background SharedGroup to import the query (using the handover object)
     // run the query, and return the table view to the caller SharedGroup using the handover object.
-    public long findAllWithHandover(long bgSharedGroupPtr, long nativeReplicationPtr,  long ptrQuery) throws BadVersionException {
-        validateQuery();
```
stk1m1: Wouldn't it be necessary to check if this `TableQuery` is valid?

beeender: The problem this method is calling in a worker thread and validating the query created in the caller thread, which is quite strange.

stk1m1: Indeed! Shell we place `validateQuery()` in caller thread to see if it fails before going in Worker thread? You can call me a paranoid checker maniac but...

cmelchior: I agree validating the query on the worker thread is a bit odd, but isn't there a chance that the query might be invalid? e.g. You move to the worker thread, but the schema has changed?

Also with this we are not validating on the UI thread in e.g. `findAllAsync()`. I don't think we will catch errors like having `beginGroup()` without `endGroup()` without it.

beeender: Hmm, right. at least `validateQuery` should be called outside. Will change it.

beeender:
Those are going to be removed by #3834

PR #3102, opened 2016-07-01
Integrate Object Store [PART1] - Build

beeender:
This is the base PR for the big Object Store PR integration.

This won't be merged for now even it gets reviewed. But for making the whole process easier, we should review this one first. Then all others can be reviewed with a minimized diff.


beeender:
close this one, we can just get #3031 merged to the master.

PR #3110, opened 2016-07-01
improve generated javadoc

larsgrefer:
link against JavaSE, Android, and RxJava

see also: realm/realm-android-adapters#55

### DIFF
--- a/realm/realm-library/build.gradle
+++ b/realm/realm-library/build.gradle
@@ -89,6 +89,10 @@ task javadoc(type: Javadoc) {
         charSet = 'UTF-8'
         locale = 'en_US'
         overview = 'src/overview.html'
+
+        links "http://docs.oracle.com/javase/7/docs/api/"
+        links "http://reactivex.io/RxJava/javadoc/"
+        linksOffline "http://developer.android.com/reference/", "${project.android.sdkDirectory}/docs/reference"
     }
     exclude '**/internal/**'
     exclude '**/BuildConfig.java'


stk1m1:
@LarsFrid 

Thanks for your contribution. It would be easier if you could upload generated javadocs somewhere to compare. Javadoc generation is a part of release cycle and the end results affects everyone so we'd like to take some caution. Great work btw 👍

larsgrefer:
@stk1m1 Here you go: http://larsgrefer.de/realm-javadoc/

For direct comparsion:

---
Current: https://realm.io/docs/java/1.1.0/api/io/realm/Realm.html
New: http://larsgrefer.de/realm-javadoc/io/realm/Realm.html

The reference to `java.lang.Object` is clickable now and links directly to http://docs.oracle.com/javase/7/docs/api/java/lang/Object.html?is-external=true

---
Current: https://realm.io/docs/java/1.1.0/api/io/realm/RealmConfiguration.Builder.html#constructor.summary
New: http://larsgrefer.de/realm-javadoc/io/realm/RealmConfiguration.Builder.html#constructor.summary

`android.content.Context` now links to http://developer.android.com/reference/android/content/Context.html?is-external=true

Note that all links to http://developer.android.com/reference/ only work in the `NO FRAMES` view

beeender:
:+1: :clap: :tada:

cmelchior:
Very nice 👍 💯

Issue #3105, opened 2016-07-01
insertOrUpdate(T t) fails on RealmList<Persons> field


Zhuinden:
#### Goal

> What do you want to achieve?

Save a `RentalProgram` object that contains a `RealmList<Persons>` object

#### Expected Results

It should work just like when I used `copyToRealmOrUpdate()`, which works

#### Actual Results

Using `insertOrUpdate()`, the `rentalProgram.getCast()` is empty!

#### Steps & Code to Reproduce

> Describe your current debugging efforts.

Frickin' ProGuard eats my debug efforts, but it's quite clear that the `cast` list is empty if I use `insertOrUpdate()`, while it works as intended with `copyToRealmOrUpdate()`.

#### Code Sample

```java

public class RentalProgram extends RealmObject {
    @PrimaryKey
    private long id;
    private RealmList<RentalProgramVersion> versions;
    private String title;
    private String href;
    private String imageSrc;
    private String baseInfoText;
    private String originalTitle;
    private String trailerUrl;
    private String description;
    private String ageRating;
    private RealmList<Persons> cast;
    private String categoryHref;
    private String releaseYear;
    // getters setters
}

public class RentalProgramVersion extends RealmObject {
    @PrimaryKey
    private long id;
    private String releaseYear;
    private int ageRating;
    private String title;
    private long duration;
    private long price;
    private String vodType;
    private String packageName;
    private long packageExternalId;
    //getters setters
}

public class Persons
        extends RealmObject {
    @PrimaryKey
    private long id;
    private String role;
    private RealmList<Person> persons;
    //getters setters
}

public class Person
        extends RealmObject {
    @PrimaryKey
    private long id;
    private String name;
    //getters setters
}
```

and the code where persistence happens is

``` java
        RealmList<Persons> castList = new RealmList<>();
        if(rentalProgramAdditionalXML.getCast() != null && rentalProgramAdditionalXML.getCast()
                .getPersons() != null) {
            for(PersonsXML personsXML : rentalProgramAdditionalXML.getCast().getPersons()) {
                castList.add(personsMapper.toRealmObject(realm, personsXML, null));
            }
        }
        realm.beginTransaction();
        castList = personsRepository.persistOrUpdate(realm, castList);
        rentalProgram.setCast(castList);
        rentalProgram.setOriginalTitle(RealmUtils.nullToEmpty(rentalProgramAdditionalXML.getOriginalTitle()));
[...]

PR #3116, opened 2016-07-03
Nh/fixing #3105

nhachicha:
Fix a scenario where inserting the same managed `RealmModel` will cause the `RealmList` to be cleared  details in #3105 

- Now `insert` and `insertOrUpdate` will behave similarly as `copy` or `copyToRealm` when passed a managed `RealmModel`

The benchmark with the current fix still looks good 
benchmark_batchwrite


nhachicha:
@realm/java

kneth:
LGTM

realm/realm-annotations-processor/src/test/resources/io/realm/AllTypesRealmProxy.java
```diff
@@ -720,7 +723,12 @@ public static void insert(Realm realm, Iterator<? extends RealmModel> objects, M
         while (objects.hasNext()) {
             object = (some.test.AllTypes) objects.next();
             if(!cache.containsKey(object)) {
-                long rowIndex = Table.nativeAddEmptyRow(tableNativePtr, 1);
+                long rowIndex;
+                if (object instanceof RealmObjectProxy && ((RealmObjectProxy)object).realmGet$proxyState().getRealm$realm() != null && ((RealmObjectProxy)object).realmGet$proxyState().getRealm$realm().getPath().equals(realm.getPath())) {
+                    rowIndex = ((RealmObjectProxy)object).realmGet$proxyState().getRow$realm().getIndex();
```
beeender: If it is managed object already, shouldn't is just `continue` the loop?

realm/realm-annotations-processor/src/test/resources/io/realm/AllTypesRealmProxy.java
```diff
@@ -844,17 +855,22 @@ public static void insertOrUpdate(Realm realm, Iterator<? extends RealmModel> ob
         while (objects.hasNext()) {
             object = (some.test.AllTypes) objects.next();
             if(!cache.containsKey(object)) {
-                String primaryKeyValue = ((AllTypesRealmProxyInterface) object).realmGet$columnString();
                 long rowIndex = TableOrView.NO_MATCH;
-                if (primaryKeyValue == null) {
-                    rowIndex = table.findFirstNull(pkColumnIndex);
+                if (object instanceof RealmObjectProxy && ((RealmObjectProxy)object).realmGet$proxyState().getRealm$realm() != null && ((RealmObjectProxy)object).realmGet$proxyState().getRealm$realm().getPath().equals(realm.getPath())) {
+                    rowIndex = ((RealmObjectProxy)object).realmGet$proxyState().getRow$realm().getIndex();
```
beeender: Same, if it is managed already, no need to set value again?

realm/realm-annotations-processor/src/test/resources/io/realm/AllTypesRealmProxy.java
```diff
@@ -665,6 +665,9 @@ public static String getTableName() {
     }
 
     public static long insert(Realm realm, some.test.AllTypes object, Map<RealmModel,Long> cache) {
+        if (object instanceof RealmObjectProxy && ((RealmObjectProxy)object).realmGet$proxyState().getRealm$realm() != null && ((RealmObjectProxy)object).realmGet$proxyState().getRealm$realm().getPath().equals(realm.getPath())) {
```
cmelchior: Minor, but this could be rewritten to something like the below better readability:
```
if (object instanceof RealmObjectProxy) {
  BaseRealm proxyState = ((RealmObjectProxy) object).realmGet$proxyState();
  if (proxyState.getRealm$realm() != null && proxyState.getRealm$realm().getPath().equals(realm.getPath()) {
    proxyState.getRow$realm().getIndex();
  }
}
```

realm/realm-annotations-processor/src/test/resources/io/realm/AllTypesRealmProxy.java
cmelchior: I think caching it makes sense? E.g. if you have the same object as both a normal link and in a link list.

realm/realm-library/src/androidTest/java/io/realm/BulkInsertTests.java
```diff
@@ -634,4 +634,18 @@ public void linkingManagedToUnmanagedObject() {
         assertEquals(42, first.getFieldObject().getFieldLong());
         assertEquals(2, realm.where(AllJavaTypes.class).count());
     }
+
+    @Test
```
cmelchior: IMO we should also add a test for what actually triggered this issue, namely that lists are no longer cleared

realm/realm-annotations-processor/src/test/resources/io/realm/AllTypesRealmProxy.java
beeender: If it is managed, then it will never be really used again i think? just a `continue` makes the generator logic a bit simpler, save one entry in map and not harmful at all. I will just add a continue here.

realm/realm-library/src/androidTest/java/io/realm/BulkInsertTests.java
beeender: Done

realm/realm-annotations-processor/src/test/resources/io/realm/AllTypesRealmProxy.java
beeender: Then it add one more nested statement which is difficult to read in the following code... So I prefer this ...

beeender:
@realm/java Review please

nhachicha:
CI: retest this please

cmelchior:
Replaced by #3306 . Closing

PR #3126, opened 2016-07-05
Forward all throwables but not only excpetions - Fix the CI timeout

beeender:
For example the UnsatisfiedLinkError should cause the process terminated.
This is the reason of all the time out issues we found for CI.

### DIFF
--- a/realm/realm-library/src/main/java/io/realm/HandlerController.java
+++ b/realm/realm-library/src/main/java/io/realm/HandlerController.java
@@ -44,13 +44,15 @@ import io.realm.internal.log.RealmLog;
 /**
  * Centralises all Handler callbacks, including updating async queries and refreshing the Realm.
  */
-final class HandlerController implements Handler.Callback {
+public final class HandlerController implements Handler.Callback {
 
     static final int REALM_CHANGED = 14930352; // Hopefully it won't clash with other message IDs.
     static final int COMPLETED_UPDATE_ASYNC_QUERIES = 24157817;
     static final int COMPLETED_ASYNC_REALM_RESULTS = 39088169;
     static final int COMPLETED_ASYNC_REALM_OBJECT = 63245986;
-    static final int REALM_ASYNC_BACKGROUND_EXCEPTION = 102334155;
+    // TODO: public (and the public HandlerController) because of QueryUpdateTask. This needs some refactor, one idea is
+    // catching and forwarding throwable in BgPriorityRunnable for all background tasks.
+    public static final int REALM_ASYNC_BACKGROUND_EXCEPTION = 102334155;
     static final int LOCAL_COMMIT = 165580141;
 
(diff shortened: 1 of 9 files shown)


realm/realm-library/src/main/java/io/realm/HandlerController.java
```diff
@@ -44,13 +44,15 @@
 /**
  * Centralises all Handler callbacks, including updating async queries and refreshing the Realm.
  */
-final class HandlerController implements Handler.Callback {
+public final class HandlerController implements Handler.Callback {
```
cmelchior: Why public? This will make it part of our public API 😢

cmelchior:
Great catch 💯 , but we should really avoid making HandlerController public.

realm/realm-library/src/main/java/io/realm/HandlerController.java
beeender: uh, yes, forgot about the API ... let me find a better a way to do it.

zaki50:
nice!

realm/realm-library/src/main/java/io/realm/internal/async/QueryUpdateTask.java
```diff
@@ -100,8 +101,17 @@ public void run() {
                 handler.obtainMessage(message, result).sendToTarget();
             }
 
-        } catch (Exception e) {
+        } catch (BadVersionException e) {
+            // In some rare race conditions, this can happen. In that case, just ignore the error.
```
beeender: @cmelchior @nhachicha I believe we should cat it here as well, right?

zaki50:
:+1: Great!

realm/realm-library/src/main/java/io/realm/internal/HandlerControllerConstants.java
```diff
@@ -0,0 +1,26 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.realm.internal;
+
+public final class HandlerControllerConstants {
```
cmelchior: Add a small Javadoc stating the purpose of this class.

realm/realm-library/src/main/java/io/realm/internal/async/QueryUpdateTask.java
cmelchior: Good catch. Yes it should. Although with 1.2, the handover object should now prevent the version from being invalidated, so I think we can actually remove all our code that handles the BadVersionException, but I would do that in another PR as it needs testing.

cmelchior:
Apart from missing Javadoc 👍

PR #3131, opened 2016-07-05
GCed ref causes flaky test

beeender:


realm/realm-library/src/androidTest/java/io/realm/DynamicRealmTests.java
```diff
@@ -338,6 +338,7 @@ public void findFirstAsync() {
                 .between(AllTypes.FIELD_LONG, 4, 9)
                 .findFirstAsync();
         assertFalse(allTypes.isLoaded());
+        looperThread.keepStrongReference.add(allTypes);
```
cmelchior: If this indeed fixes its we should have a ton of bugs in all other unit tests using the `@RunTestInLooperThread`?
`allTypes` do go out of scope when exiting the method though 😢

stk1m1:
👍 as soon as Mr. Jenkins says ok.

zaki50:
:+1:

realm/realm-library/src/androidTest/java/io/realm/DynamicRealmTests.java
beeender: I think I fixed most of them last time. So there should be quite a few left.

PR #3132, opened 2016-07-05
Trying to reproduce #3130

### LINKED ISSUE #3130: "This Realm instance has already been closed, making it unusable." - Even when copying object?
#### Goal

To be able to access the data in an object after calling realm.close.

#### Expected Results

> To get an object out of the Realm data store and use it's data after closing realm, by copying the object first.

#### Actual Results

>  RuntimeException: This Realm instance has already been closed, making it unusable.

Code I am using:

```
        User user = realm.where(User.class)
                .equalTo(Constants.ID, 1)
                .findFirst();
        User copy = realm.copyFromRealm(user);
        realm.close();
        return copy;
```

I am currently under the impression that copying the object first gives me the ability to use that object freely, instead of it being just a reference to the realm backing store. I've got various helper methods to make queries to Realm for me, where I open a realm then close it in one go then return the data. Is this incorrect? By copying the object shouldn't there be no issue?

Interesting to hear your thoughts on the subject and what the correct procedure should be. I'm assuming I need to keep the realm open but that currently doesn't make sense to me.

Realm version(s): 1.1.0

Android Studio version: 2.1.2

Which Android version and device: Nexus 5 Marshmallow.

kneth:
Closes #3130

### DIFF
--- a/realm/realm-library/src/androidTest/java/io/realm/RealmTests.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/RealmTests.java
@@ -2759,6 +2759,17 @@ public class RealmTests {
         }
     }
 
+    @Test
+    public void copyFromRealm_useAfterClose() {
+        Realm realm = Realm.getInstance(configFactory.createConfiguration());
+        populateTestRealm(realm, 10);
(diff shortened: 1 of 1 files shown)


cmelchior:
Should this PR just be closed?

kneth:
Yes!

PR #3133, opened 2016-07-05
Add RealmQuery.in()

dalinaum:


realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -403,6 +403,191 @@ public boolean isValid() {
...
+
+    /**
+     * Equal-to-any comparison.
+     *
+     * @param fieldName the field to compare.
+     * @param values the array of values to compare with.
+     * @return the query object.
+     * @throws java.lang.IllegalArgumentException if one or more arguments do not match class or field type.
+     */
+    public RealmQuery<E> equalToAny(String fieldName, String[] values) {
```
cmelchior: use `in()` instead of `equalToAny()`

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -404,6 +404,191 @@ public boolean isValid() {
...
+    /**
+     * Equal-to-any comparison.
+     *
+     * @param fieldName the field to compare.
+     * @param values the array of values to compare with.
+     * @return the query object.
+     * @throws java.lang.IllegalArgumentException if one or more arguments do not match class or field type.
+     */
+    public RealmQuery<E> in(String fieldName, String[] values, Case casing) {
+        if (values == null || values.length == 0) {
```
zaki50: Shouldn't we treat `null` as an exception?

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -404,6 +404,191 @@ public boolean isValid() {
...
+            return this;
+        }
+        equalTo(fieldName, values[0], casing);
```
zaki50: We should use `beginGroup()` and `endGroup()` since `equalTo("a") and in("b", "c", "d")` does not equal to `equalTo("a") and equalTo("b") or equalTo("c") or equalTo("d")`.

cmelchior: a good test case for this would also be `where(Foo.class).not().in("foo", list)`

dalinaum: I see. I'm adding `beginGroup()`, `endGroup()` and additional tests. Thanks, @zaki50 and @cmelchior

dalinaum:

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -404,6 +404,191 @@ public boolean isValid() {
...
+
+    /**
+     * In comparison. This allow you to test if objects matches any value in an array of values.
+     *
+     * @param fieldName the field to compare.
+     * @param values the array of values to compare with.
+     * @return the query object.
+     * @throws java.lang.IllegalArgumentException if values is empty or do not match class or field type.
+     */
+    public RealmQuery<E> in(String fieldName, String[] values) {
```
zaki50: How about using vararg? e.g. `String... values` instead of `String[] values`

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -404,6 +404,191 @@ public boolean isValid() {
         return this;
     }
 
+    // EqualToAny
+
+    /**
+     * In comparison. This allow you to test if objects matches any value in an array of values.
+     *
+     * @param fieldName the field to compare.
+     * @param values the array of values to compare with.
```
zaki50: Shouldn't we state that all elements in this array must be non-null?

realm/realm-library/src/main/java/io/realm/RealmQuery.java
dalinaum: Since values are given by another method or an existing field, I prefer the array version more than the variant argument version. If we use variant arguments, how can we use another method or an existing array field to fill arguments?

realm/realm-library/src/main/java/io/realm/RealmQuery.java
dalinaum: I see. I will update Javadocs.

realm/realm-library/src/main/java/io/realm/RealmQuery.java
zaki50: We can just use `values` as vararg. i.e. no need to change the method body.

zaki50: Ah, but the vararg must be the last element of the arguments.
We need to move the `Case` argument.

dalinaum: @zaki50 I see. I had a misunderstanding about variable arguments. Thanks.

dalinaum:
retest this please

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -404,6 +404,192 @@ public boolean isValid() {
...
+     * @param values values to compare with and it cannot be null or empty.
+     * @return the query object.
+     * @throws java.lang.IllegalArgumentException if values is empty or do not match class or field type.
+     */
+    public RealmQuery<E> in(String fieldName, Case casing, String... values) {
+        if (values == null || values.length == 0) {
+            throw new IllegalArgumentException(("Non-empty values must be provided."));
+        }
+        beginGroup().equalTo(fieldName, values[0], casing);
+        for (int i = 1; i < values.length; ++i) {
```
zaki50: Please use `i++` instead of `++i` to align with existing code.

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -404,6 +404,192 @@ public boolean isValid() {
...
+        }
+        beginGroup().equalTo(fieldName, values[0], casing);
+        for (int i = 1; i < values.length; i++) {
+            or().equalTo(fieldName, values[i], casing);
+        }
+        return endGroup();
+    }
+
+    /**
+     * In comparison. This allow you to test if objects matches any value in an array of values.
```
kneth: "allow" -> "allows"
"matches" -> "match"

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -404,6 +404,192 @@ public boolean isValid() {
...
+     *
+     * @param fieldName the field to compare.
+     * @param casing how casing is handled. {@link Case#INSENSITIVE} works only for the Latin-1 characters.
+     * @param values values to compare with and it cannot be null or empty.
+     * @return the query object.
+     * @throws java.lang.IllegalArgumentException if values is empty or do not match class or field type.
+     */
+    public RealmQuery<E> in(String fieldName, Case casing, String... values) {
+        if (values == null || values.length == 0) {
+            throw new IllegalArgumentException(("Non-empty values must be provided."));
```
kneth: Please remove one set of `()`

kneth:
My comments apply to all the `in()` methods.

dalinaum:
Thanks. Just updated. Please review this again @realm/java

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -404,6 +404,192 @@ public boolean isValid() {
...
+
+    /**
+     * In comparison. This allows you to test if objects match any value in an array of values.
+     *
+     * @param fieldName the field to compare.
+     * @param values values to compare with and it cannot be null or empty.
+     * @return the query object.
+     * @throws java.lang.IllegalArgumentException if values is empty or do not match class or field type.
+     */
+    public RealmQuery<E> in(String fieldName, String... values) {
```
cmelchior: I am not a big fan of this API, it is very hard to decipher what is the field name and what is the parameter. 
```
in("foo", "bar", "baz") // hard to decipher 
in("foo", new String[] { "bar", baz" }), much easier to read, but a bit harder to type
```
IMO readability should always win.
Also moving the `Case` to the middle is not in line with the rest of the API.

IMO we should use `in(String fieldName, String[] values, Case case)`  instead of var args.

Thoughts @realm/java ?

cmelchior:
I really like the extended test cases. See my comment about the API.

realm/realm-library/src/main/java/io/realm/RealmQuery.java
dalinaum: The first version I made was the array version. @zaki50 suggested variant arguments version, I tested variant arguments several times and I adopted vargs instead of arrays.

In vargs version, we can invoke `in` operators by any fashions.

```
in("foo", "bar", "baz") // <- OK
in("foo", new String[] { "bar", baz" }) // <- OK
```

Although `Case case` parameter should not be the final parameter in the vargs versions, I thought it could be tolerable.

cmelchior: I know that the var arg variant can also be written with arrays, but a better solution for that would actually be `in(String field Name, Case case, String first field, String... additionalFields)` (see Effective Java item 42), since that would enforce at least one argument.

However moving the Case parameter feels awkward to me, especially since I don't think we should promote the non-array version. Just because something is possible doesn't mean we have to do it, especially if readability and consistency suffers 😄 

But would be nice to hear from @zaki50 why the var arg version is better.

zaki50: Hmm, I'm leaning toward @cmelchior's opinion. Putting the `Case` parameter before values is awkward...

I vote reverting varargs version of `in()` API.

dalinaum: I see. I will revert the variable arguments version.

dalinaum:
Just updated this PR, please review this again @realm/java

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -404,6 +404,192 @@ public boolean isValid() {
...
 
+    // In
+
+    /**
+     * In comparison. This allows you to test if objects match any value in an array of values.
+     *
+     * @param fieldName the field to compare.
+     * @param values values to compare with and it cannot be null or empty.
+     * @return the query object.
+     * @throws java.lang.IllegalArgumentException if values is empty or do not match class or field type.
```
beeender: `do` -> `does`. `class and field` -> `the class and the field` same as below.

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -404,6 +404,192 @@ public boolean isValid() {
...
+     */
+    public RealmQuery<E> in(String fieldName, String[] values) {
+        return in(fieldName, values, Case.SENSITIVE);
+    }
+
+    /**
+     * In comparison. This allows you to test if objects match any value in an array of values.
+     *
+     * @param fieldName the field to compare.
+     * @param values values to compare with and it cannot be null or empty.
```
beeender: `values to compare` -> `the values to compare`

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -404,6 +404,192 @@ public boolean isValid() {
...
+     * @param casing how casing is handled. {@link Case#INSENSITIVE} works only for the Latin-1 characters.
+     * @return the query object.
+     * @throws java.lang.IllegalArgumentException if values is empty or do not match class or field type.
+     */
+    public RealmQuery<E> in(String fieldName, String[] values, Case casing) {
+        if (values == null || values.length == 0) {
+            throw new IllegalArgumentException("Non-empty values must be provided.");
```
beeender: Make a const string field for the exception message.

beeender: also, quote the `values` like:
`"Non-empty `values` must be provided."`

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -404,6 +404,192 @@ public boolean isValid() {
...
+
+    /**
+     * In comparison. This allows you to test if objects match any value in an array of values.
+     *
+     * @param fieldName the field to compare.
+     * @param values values to compare with and it cannot be null or empty.
+     * @return the query object.
+     * @throws java.lang.IllegalArgumentException if values is empty or do not match class or field type.
+     */
+    public RealmQuery<E> in(String fieldName, byte[] values) {
```
beeender: We should only supply the boxed version for `byte[]`, `int[]` and etc. like other similar queries. This cannot handle the nullable cases. @realm/java What do you think?

realm/realm-library/src/androidTest/java/io/realm/RealmQueryTests.java
```diff
@@ -405,6 +408,198 @@ public void equalTo_nonLatinCharacters() {
...
+        }
+        RealmResults<AllTypes> resultList = realm.where(AllTypes.class).in(AllTypes.FIELD_STRING, new String[] {"test data 15"}).findAll();
+        assertEquals(1, resultList.size());
+        resultList = realm.where(AllTypes.class).in(AllTypes.FIELD_STRING, new String[] {"test data 15", "test data 117", "test data 30", "test data 199"}).findAll();
+        assertEquals(4, resultList.size());
+        resultList = realm.where(AllTypes.class).in(AllTypes.FIELD_STRING, new String[] {"TEST data 15", "TEST data 117", "test DATA 30", "test DATA 199"}, Case.INSENSITIVE).findAll();
+        assertEquals(4, resultList.size());
+        resultList = realm.where(AllTypes.class).not().in(AllTypes.FIELD_STRING, new String[] {"TEST data 15", "TEST data 117", "test DATA 30", "test DATA 199"}, Case.INSENSITIVE).findAll();
+        assertEquals(196, resultList.size());
+        resultList = realm.where(AllTypes.class).not().in(AllTypes.FIELD_STRING, new String[] { "TEST data 15", "TEST data 117", "test DATA 30", "test DATA 199"}, Case.INSENSITIVE).findAll();
```
beeender: need to test `null` in the list as well. also needs to test it throws when the given field is not nullable.

beeender: Same for others which has filed could be `null`

realm/realm-library/src/main/java/io/realm/RealmQuery.java
cmelchior: I am a bit torn. Allowing `null` here is going to be a pain to support since we need to call another method in core for it, and I am hard pressed to see it ever being used e.g `in("myNumber", new Integer[] { null, 1, 2 })` doesn't seem like code you would normally write?

Going over our other operators, we seem to use the boxed type where it make sense, e.g `equalTo()` and `notEqualTo()`, but not in operators like `between()`. These cases are a bit more clear though as using `between()` on `null` is a clear semantic error.

I would lean towards just keeping the primitive versions for now to be honest, and just have an issue for adding boxed support if need be. I could probably be convinced if you feel strongly about it though?

realm/realm-library/src/androidTest/java/io/realm/RealmQueryTests.java
cmelchior: Good catch 👍

realm/realm-library/src/main/java/io/realm/RealmQuery.java
beeender: > a pain to support since we need to call another method in core for it

No, no need to call another method (at least in java).
`RealmQuery.equalTo(field, null)` is supported well and all checks are done when implementing nullable. Simply by changing it to boxed type everything will just work.

`in` is more like a helper function if you need to call `equalTo`, if we support `equalTo` with `null` values, why not support it in `in` as well?

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -404,6 +404,192 @@ public boolean isValid() {
...
+     * In comparison. This allows you to test if objects match any value in an array of values.
+     *
+     * @param fieldName the field to compare.
+     * @param values values to compare with and it cannot be null or empty.
+     * @param casing how casing is handled. {@link Case#INSENSITIVE} works only for the Latin-1 characters.
+     * @return the query object.
+     * @throws java.lang.IllegalArgumentException if values is empty or do not match class or field type.
+     */
+    public RealmQuery<E> in(String fieldName, String[] values, Case casing) {
+        if (values == null || values.length == 0) {
```
cmelchior: Should we allow the 0-length `values` array? It would basically default to the query always returning the empty set, but I would imagine that this `values` array in most cases will be generated in a loop based on some other input, so I think that this array in a lot of cases will actually be empty, making it a hassle for the user to work around.  

Thoughts @realm/java?

realm/realm-library/src/main/java/io/realm/RealmQuery.java
cmelchior: Hmm, good point. For large lists there will be a lot of boxing happening, but I doubt that it is a big problem. I guess the boxed API is fine 👍

realm/realm-library/src/main/java/io/realm/RealmQuery.java
beeender: hmmm, right. But how can we support it? Hack it in `RealmQuery` to say this will return an empty results? Or ask some support from core?

cmelchior: Hmm, right. Supporting it would actually be a huge hassle. I still think it is a good idea, but we should probably wait for object-store/core support in that case.

cmelchior: Unless there is some easy way to add a query operator that we know will never be true.

bdash: In Cocoa we have [`TrueExpression` and `FalseExpression`](https://github.com/realm/realm-cocoa/blob/30947f35a115db9e62e208a2ee2c2b75a3009e9b/Realm/RLMQueryUtil.mm#L88-L114) that we use for cases like [handling an `or` group containing zero elements](https://github.com/realm/realm-cocoa/blob/30947f35a115db9e62e208a2ee2c2b75a3009e9b/Realm/RLMQueryUtil.mm#L727-L732).

dalinaum: Since it would complicate this PR, I just made an issue for it: #3238.

realm/realm-library/src/androidTest/java/io/realm/RealmQueryTests.java
dalinaum: Thanks. Some null tests were added now.

dalinaum:
Updated this PR, please review this again @realm/java :)

dalinaum:
Since CI has failed, this PR will be tagged as `S:In Progress` instead of `S:Review` and I will look into failed tests.

dalinaum:
Fix two tests, please review this again @realm/java

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -404,6 +406,192 @@ public boolean isValid() {
         return this;
     }
 
+    // In
```
beeender: Delete this line

realm/realm-library/src/androidTest/java/io/realm/RealmMigrationTests.java
```diff
@@ -694,9 +694,24 @@ public void migrate(DynamicRealm realm, long oldVersion, long newVersion) {
     public void openPreNullWithRequired() throws IOException {
         configFactory.copyRealmFromAssets(context,
                 "default-before-migration.realm", Realm.DEFAULT_REALM_NAME);
+        RealmMigration migration = new RealmMigration() {
```
beeender: It is a bit annoy ... but I think it is better to update the `default-before-migration.realm` instead. This might have side effect on the test case itself it make the test case gets confused ... YEAH, update the binary file is really annoy.

realm/realm-library/src/main/java/io/realm/RealmQuery.java
dalinaum: There are `// Equal` and `// Not Equal` in this file. Should we delete them too?

beeender: I think you can actually use `NullTypes` or `NoPrimaryKeyNullTypes` for testing. no need to change the `AllTypes`, then no need to update this binary file. 

Also, you can add test for nullable fields which is missed right now.

realm/realm-library/src/androidTest/java/io/realm/RealmMigrationTests.java
beeender: I think you can actually use `NullTypes` or `NoPrimaryKeyNullTypes` for testing. no need to change the `AllTypes`, then no need to update this binary file.

Also, you can add test for nullable fields which is missed right now.

realm/realm-library/src/main/java/io/realm/RealmQuery.java
beeender: Uh.. never mind, it was my fault. I thought it is a unfinished javadoc.
Sorry :grin: !!! Just leave it here.

beeender:
A few comments on the tests. others :+1:

realm/realm-library/src/androidTest/java/io/realm/RealmMigrationTests.java
dalinaum: I see. I will try to use one of them instead of `AllTypes`.

cmelchior: What does this unit test have to do with `in()`?

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -404,6 +406,192 @@ public boolean isValid() {
...
+
+    /**
+     * In comparison. This allows you to test if objects match any value in an array of values.
+     *
+     * @param fieldName the field to compare.
+     * @param values the values to compare with and it cannot be null or empty.
+     * @return the query object.
+     * @throws java.lang.IllegalArgumentException if values is empty or does not match the class and the field type.
```
cmelchior: If the field isn't a String field or {@code values} is empty.

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -404,6 +406,192 @@ public boolean isValid() {
...
+    }
+
+    /**
+     * In comparison. This allows you to test if objects match any value in an array of values.
+     *
+     * @param fieldName the field to compare.
+     * @param values the values to compare with and it cannot be null or empty.
+     * @param casing how casing is handled. {@link Case#INSENSITIVE} works only for the Latin-1 characters.
+     * @return the query object.
+     * @throws java.lang.IllegalArgumentException if values is empty or does not match the class and the field type.
```
cmelchior: If the field isn't a String field or {@code values} is empty.

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -404,6 +406,192 @@ public boolean isValid() {
...
+        return endGroup();
+    }
+
+    /**
+     * In comparison. This allows you to test if objects match any value in an array of values.
+     *
+     * @param fieldName the field to compare.
+     * @param values the values to compare with and it cannot be null or empty.
+     * @return the query object.
+     * @throws java.lang.IllegalArgumentException if values is empty or does not match the class and the field type.
```
cmelchior: If the field isn't a Byte field or {@code values} is {@code null} or empty.

Same for all the others.

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -404,6 +406,192 @@ public boolean isValid() {
...
```

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -404,6 +406,192 @@ public boolean isValid() {
...
```

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -404,6 +406,192 @@ public boolean isValid() {
...
```

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -404,6 +406,192 @@ public boolean isValid() {
...
```

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -404,6 +406,192 @@ public boolean isValid() {
...
```

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -404,6 +406,192 @@ public boolean isValid() {
...
```

cmelchior:
👍

dalinaum:
Updated this PR again. Please review this @realm/java

realm/realm-library/src/androidTest/java/io/realm/RealmQueryTests.java
```diff
@@ -405,6 +438,493 @@ public void equalTo_nonLatinCharacters() {
...
+        assertEquals(1, resultList.size());
+        resultList = realm.where(NoPrimaryKeyNullTypes.class).in(NoPrimaryKeyNullTypes.FIELD_STRING_NOT_NULL, new String[]{"test data 14", "test data 118", "test data 31", "test data 199"}).findAll();
+        assertEquals(4, resultList.size());
+        resultList = realm.where(NoPrimaryKeyNullTypes.class).in(NoPrimaryKeyNullTypes.FIELD_STRING_NOT_NULL, new String[]{"TEST data 14", "test data 118", "test data 31", "test DATA 199"}, Case.INSENSITIVE).findAll();
+        assertEquals(4, resultList.size());
+        resultList = realm.where(NoPrimaryKeyNullTypes.class).not().in(NoPrimaryKeyNullTypes.FIELD_STRING_NOT_NULL, new String[]{"TEST data 14", "test data 118", "test data 31", "test DATA 199"}, Case.INSENSITIVE).findAll();
+        assertEquals(196, resultList.size());
+        resultList = realm.where(NoPrimaryKeyNullTypes.class).not().in(NoPrimaryKeyNullTypes.FIELD_STRING_NOT_NULL, new String[]{"TEST data 14", "test data 118", "test data 31", "test DATA 199"}, Case.INSENSITIVE).findAll();
+        assertEquals(196, resultList.size(
+
```
kneth: Please remove empty lines

realm/realm-library/src/androidTest/java/io/realm/RealmQueryTests.java
```diff
@@ -405,6 +438,493 @@ public void equalTo_nonLatinCharacters() {
...
+        try {
+            realm.where(NoPrimaryKeyNullTypes.class).not().in(NoPrimaryKeyNullTypes.FIELD_STRING_NOT_NULL, new String[]{"TEST data 14", "test data 118", null, "test DATA 199"}, Case.INSENSITIVE).findAll();
+            fail();
+        } catch (IllegalArgumentException ignored) {
+        }
+    }
+
+    @Test
+    public void in_stringNull() {
```
kneth: `in_stringNotNull()` and `in_stringNull()` are almost identical, and it must be possible to have one common method which `in_stringNotNull()` and `in_stringNull()` can call with the field name as parameter.

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -404,6 +406,192 @@ public boolean isValid() {
         return this;
     }
 
+    // In
+
+    /**
+     * In comparison. This allows you to test if objects match any value in an array of values.
+     *
+     * @param fieldName the field to compare.
+     * @param values the values to compare with and it cannot be null or empty.
```
kneth: "the values" -> "array of values"

kneth:
:+1: with comments on generalising tests and a bit of language in javadoc.

dalinaum:
Just updated this PR, please review this again @realm/java

cmelchior:
👍

beeender:
:+1:

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -404,6 +406,192 @@ public boolean isValid() {
...
+
+    /**
+     * In comparison. This allows you to test if objects match any value in an array of values.
+     *
+     * @param fieldName the field to compare.
+     * @param values array of values to compare with and it cannot be null or empty.
+     * @return the query object.
+     * @throws java.lang.IllegalArgumentException if the field isn't a Byte field or {@code values} is {@code null} or empty.
+     */
+    public RealmQuery<E> in(String fieldName, Byte[] values) {
```
bmunkholm: Is it needed to support lists of both Byte, Short, Integer and Long? I would think Long would be sufficient.

realm/realm-library/src/androidTest/java/io/realm/RealmQueryTests.java
```diff
@@ -404,6 +437,359 @@ public void equalTo_nonLatinCharacters() {
...
+        doTestForInString(NoPrimaryKeyNullTypes.FIELD_STRING_NULL);
+        RealmResults<NoPrimaryKeyNullTypes> resultList = realm.where(NoPrimaryKeyNullTypes.class).not().in(NoPrimaryKeyNullTypes.FIELD_STRING_NULL, new String[]{"TEST data 14", "test data 118", null, "test DATA 199"}, Case.INSENSITIVE).findAll();
```
bmunkholm: These lines are too long.

bmunkholm:
@dalinaum I added a few comments after the merge.
I'm very doubtful that we need so many variants of the method, since I suspect people will have a need for this API where they construct the list of conditions programmatically and specifically for this method, and can therefore use whatever number type we require.
/cc @cmelchior

realm/realm-library/src/androidTest/java/io/realm/RealmQueryTests.java
dalinaum: Right. I will make a reformatting patch for this.

realm/realm-library/src/main/java/io/realm/RealmQuery.java
dalinaum: Althogh it could reduce methods count, I think it could be more inconvenient. One of Byte, Short and Integer cannot be converted to Long. Users should iterate an array, unbox data of arrays to primitives, and box them into Long array.

If we decide to remove other in operators, should we remove equalTo methods for numeric types?

cmelchior:
Right now our `equalTo()` methods have the same variants. In order to stay consistent with that I think the current implementation is fine.

It might make sense to clean it up, but we should do it everywhere then IMO

realm/realm-library/src/main/java/io/realm/RealmQuery.java
cmelchior: Right now our equalTo() methods have the same variants. In order to stay consistent with that I think the current implementation is fine.

It might make sense to clean it up, but we should do it everywhere then IMO

PR #3135, opened 2016-07-05
More GCed ref for flaky test

beeender:


emanuelez:
Even better if we could detect automatically this kind of thing with some static code analysis

zaki50:
:+1:

I executed `connectedCheck` task several times on this branch.
I feet our test is much more stable than before. 💯

beeender:
@emanuelez Create a issue #3142 to track that.

PR #3137, opened 2016-07-05
Clean and unify the Jenkinsfile

emanuelez:
It also adds Slack notifications for the `master` and `releases` branches and introduces lint checking.
It doesn't fail be build yet since we have 135 issues to solve (in our unit tests)


emanuelez:
@realm/java @teotwaki @alebsack

Jenkinsfile
```diff
@@ -1,169 +1,158 @@
...
-        def rootFolder = new File('unzipped/jni')
-        rootFolder.traverse (type: DIRECTORIES) { folder ->
-            def abiName = folder.name()
-            def libSize = new File(folder, 'librealm-jni.so').size() as String
-            sendTaggedMetric('abi_size', libSize, 'type', abiName)
-        }
+  dir('realm/realm-library/build/outputs/aar') {
+    sh '''set -xe
+    unzip realm-android-library-release.aar -d unzipped
+    find $ANDROID_HOME -name dx | sort -r | head -n 1 > dx
```
kneth: Why store in a file? An environment variable might be an alternative.

kneth:
It looks fine to me - but it's a bit out of my typical range ;-)

Jenkinsfile
emanuelez: because that's the only way to pass data from a `sh` section back into Groovy. Environment variables are ephemeral in this context.

zaki50: but it doesn't seem to be used in Groovy.

emanuelez: Ah right! That's is old code from Christian. I didn't touch that part (apart from the indentation). Agreed, a variable could do, but it's a bit out of the scope of the PR

Jenkinsfile
```diff
@@ -1,169 +1,158 @@
...
+
+def buildSuccess = false
+try {
+  node('docker') {
+    // Allocate a custom workspace to avoid having % in the path (it breaks ld)
+    ws('/tmp/realm-java') {
+      stage 'SCM'
+      checkout scm
+      // Make sure not to delete the folder that Jenkins allocates to store scripts
+      sh 'git clean -ffdx -e .????????'
```
cmelchior: Could we narrow the regular expression a bit? ie. to only alphanumerical characters or something?

Jenkinsfile
```diff
@@ -1,169 +1,158 @@
...
-            'archive': true,
-            'glob' : 'logcat.txt'
+        // TODO: add support for running monkey on the example apps
+
+        if (env.BRANCH_NAME == 'master') {
+          stage 'Collect metrics'
+          collectAarMetrics()
+
+          stage 'Publish to OJO'
+          gradle 'assemble ojoUpload'
```
cmelchior: should this be `gradle('assemble ojoUpload') instead ?

Jenkinsfile
emanuelez: I wish we had regexes! :( Only shell wildcards are allowed here

Jenkinsfile
emanuelez: In Groovy if you only have one argument you can omit the brackets and it's used all over the place in the Jenkinsfile DSL as you can see

alebsack:
:+1 Although my groovy isn't great yet...  I'll be sure to track this repo for examples. :)

Jenkinsfile
```diff
@@ -1,169 +1,158 @@
...
-            }
+import groovy.json.JsonOutput
+
+def buildSuccess = false
+try {
+  node('docker') {
+    // Allocate a custom workspace to avoid having % in the path (it breaks ld)
+    ws('/tmp/realm-java') {
+      stage 'SCM'
+      checkout scm
```
radu-tutueanu: Does `checkout scm` use the parameters defined in the Jenkins job SCM stage?

radu-tutueanu:
👍 Looks good. I will be using this as a reference for groovy scripts as well.

Jenkinsfile
emanuelez: Yes, and in this case it's completely handled by the Github Organisation Folder plugin

PR #3144, opened 2016-07-06
BYE BYE finalizer

### LINKED ISSUE #2496: io.realm.internal.Table.finalize() timed out after 10 seconds

beeender:
This PR is trying to move the native desctructions from finalizer to Phantom Deamon.
And some ideas are inspired by https://android-review.googlesource.com/#/c/175765/10
- [x] Use linked list for phantom reference queue which should be faster than current array implementation. (no more chunk allocation)
- [x] Use function pointer to generalize the deallocation for native objects instead of current `switch`/`case` based routine.
- [x] All native objects go with the phantom reference, (maybe even the `SharedGroup`/`Group`, needs to confirm). _This will be handled in other PR._
- [x] Use Phantom Deamon thread to do deallocation.
- [x] Check if we still need lock on `Context` if we can do deallocation only on one thread.
- [x] Some exciting benchmarks.

#
## Why?
* To close #2496 . Some UI freeze is caused by the Realm's finalizer.
* See [Finalization and Phantom References](https://dzone.com/articles/finalization-and-phantom) about the problem with with Finalization and why the phantom reference is preferred.
* It basically simplify our code for deallocating native resource.

# 
## Memory Usage Benchmarks (Mem usage in KB)

The memory usage benchmark is done by [this script](https://gist.github.com/beeender/730b058e3e44985444e4a8402bfb7335). It gets the java/native alloc heap every 100ms.

**From the results, it shows a slightly better memory usage for low end devices, but more memory used for high end devices. It doesn't necessarily mean that Phantom reference it better/worse than finalizer, but it tells we don't have a native memory pressure problem with it.**

Blow benchmark measure and compare the memory usage during running realm-java benchmark unit tests. 

| Device | Average Java Heap | Peak Java Heap | Average Native Heap | Peak Native Heap |
| --- | --- | --- | --- | --- |
| Nokia X2 4.3 512MB - Phantom | 7116 | 9328 | 2983 | 24409 |
| Nokia X2 4.3 512MB - Finalizer | 7198 | 9423 | 3104 | 24408 |
| Simulator 6.0 1.5GB - Phantom | 4174 | 23191 | 51624 | 271318 |
| Simulator 6.0 1.5GB - Finalizer | 3833 | 17921 | 35738 | 225334 |
| Huawei Honor 7 6.0 3GB - Phantom | 18409 | 24979 | 50426 | 101464 |
| Huawei Honor 7 6.0 3GB - Finalizer | 18436 | 25172 | 42681 | 86378 |
| Xoom 4.1.2 1GB - Phantom | 3155 | 6658 | 2373 | 14691 |
| Xoom 4.1.2 1GB - Finalizer | 3207 | 5708 | 4349 | 21698 |
## Performance Benchmarks

The left results are for Phantom deamon, the right ones are for finalizers. Tests are done with Xoom 4.1.2 device. No big changes in most cases except objects creations.

phantomvsfinalizer
# 
## Commit Message

Use phantom reference and daemon thread to replace finalizer for
destruct native objects.

Some key points of this commit:
- The phantom reference pool is implemented as a linked list which
  suppose to be fast insertion/removal.
- A daemon thread is created to monitor and free all phantom reference.
- Delayed disposal before native object creation is not needed any more.
- SharedGroup still gets freed in the caller thread with a lock on
  context.
- Native object needs to pass a destruction function pointer in addition
  to the native object pointer.


realm/realm-library/src/androidTest/java/io/realm/GCTests.java
```diff
@@ -0,0 +1,127 @@
...
+import io.realm.entities.AllTypes;
+import io.realm.entities.Dog;
+import io.realm.rule.TestRealmConfigurationFactory;
+
+import static junit.framework.TestCase.assertNotNull;
+
+// This test is for the fact we don't have locks for native objects creation that when finalizer/phantom thread free the
+// native object, the same Realm could have some native objects creation at the same time.
+@RunWith(AndroidJUnit4.class)
+public class GCTests {
```
beeender: This test takes some time. Maybe we should disable it by default. But it worth to run before release/core updates.

beeender:
@realm/java This is ready

beeender:
back to in progress because of native crash.

realm/realm-jni/src/io_realm_internal_LinkView.cpp
```diff
@@ -248,3 +250,15 @@ JNIEXPORT void JNICALL Java_io_realm_internal_LinkView_nativeRemoveTargetRow
...
+static void finalize_link_view(jlong ptr)
+{
+    TR_ENTER_PTR(ptr)
+    LangBindHelper::unbind_linklist_ptr(*LV(ptr));
+}
+
+JNIEXPORT jlong JNICALL Java_io_realm_internal_LinkView_nativeGetFinalizer
+  (JNIEnv *, jclass)
+{
+    return static_cast<jlong>(reinterpret_cast<uintptr_t>(&finalize_link_view));
```
stk1m1: it seems `static_cast<jlong>(reinterpret_cast<uintptr_t>())` gets repeated multiple times (6 times total). Any chance to use a MACRO?

beeender: I had a discussion with @kneth before... actually this more readable than macro. So I think it is better we keep this style.

stk1m1:
Great work 👍  Unstoppable @beeender!

beeender:
Tests still fail on my simulator with below crash log, checking now.

```
F/libc    ( 4257): Fatal signal 6 (SIGABRT), code -6 in tid 4276 (pool-1-thread-1)
I/DEBUG   ( 1139): *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
I/DEBUG   ( 1139): Build fingerprint: 'Android/sdk_google_phone_x86/generic_x86:5.1.1/LMY48X/3038874:userdebug/test-keys'
I/DEBUG   ( 1139): Revision: '0'
I/DEBUG   ( 1139): ABI: 'x86'
I/DEBUG   ( 1139): pid: 4257, tid: 4276, name: pool-1-thread-1  >>> io.realm.test <<<
I/DEBUG   ( 1139): signal 6 (SIGABRT), code -6 (SI_TKILL), fault addr --------
I/DEBUG   ( 1139):     eax 00000000  ebx 000010a1  ecx 000010b4  edx 00000006
I/DEBUG   ( 1139):     esi a2be9db8  edi 00000002
I/DEBUG   ( 1139):     xcs 00000073  xds 0000007b  xes 0000007b  xfs 00000077  xss 0000007b
I/DEBUG   ( 1139):     eip b74aa0f6  ebp 000010b4  esp a2be9070  flags 00200286
I/DEBUG   ( 1139):
I/DEBUG   ( 1139): backtrace:
I/DEBUG   ( 1139):     #00 pc 000750f6  /system/lib/libc.so (tgkill+22)
I/DEBUG   ( 1139):     #01 pc 0002217b  /system/lib/libc.so (pthread_kill+155)
I/DEBUG   ( 1139):     #02 pc 000239f4  /system/lib/libc.so (raise+36)
I/DEBUG   ( 1139):     #03 pc 0001bdf4  /system/lib/libc.so (abort+84)
I/DEBUG   ( 1139):     #04 pc 001566a4  /data/app/io.realm.test-1/lib/x86/librealm-jni.so (__gnu_cxx::__verbose_terminate_handler()+452)
I/DEBUG   ( 1139):     #05 pc 00124fc7  /data/app/io.realm.test-1/lib/x86/librealm-jni.so (__cxxabiv1::__terminate(void (*)())+23)
I/DEBUG   ( 1139):     #06 pc 00156145  /data/app/io.realm.test-1/lib/x86/librealm-jni.so (__cxa_call_terminate+69)
I/DEBUG   ( 1139):     #07 pc 00124ab1  /data/app/io.realm.test-1/lib/x86/librealm-jni.so (__gxx_personality_v0+321)
I/DEBUG   ( 1139):     #08 pc 0016ce56  /data/app/io.realm.test-1/lib/x86/librealm-jni.so (_Unwind_RaiseException_Phase2+140)
I/DEBUG   ( 1139):     #09 pc 0016d2a4  /data/app/io.realm.test-1/lib/x86/librealm-jni.so (_Unwind_Resume+92)
I/DEBUG   ( 1139):     #10 pc 00078a52  /data/app/io.realm.test-1/lib/x86/librealm-jni.so (realm::_impl::WriteLogCollector::map_header_if_needed() const+114)
I/DEBUG   ( 1139):     #11 pc 0007a8f0  /data/app/io.realm.test-1/lib/x86/librealm-jni.so (realm::_impl::WriteLogCollector::get_changesets(unsigned long long, unsigned long long, realm::BinaryData*) const+42)
I/DEBUG   ( 1139):     #12 pc 00023985  /data/app/io.realm.test-1/lib/x86/librealm-jni.so (realm::_impl::ChangesetInputStream::next_block(char const*&, char const*&)+127)
I/DEBUG   ( 1139):     #13 pc 00024e7d  /data/app/io.realm.test-1/lib/x86/librealm-jni.so (realm::_impl::TransactLogParser::next_input_buffer()+23)
I/DEBUG   ( 1139):     #14 pc 0007b9fc  /data/app/io.realm.test-1/lib/x86/librealm-jni.so (realm::Group::advance_transact(unsigned int, unsigned int, realm::_impl::NoCopyInputStream&)+182)
I/DEBUG   ( 1139):     #15 pc 00106aac  /data/app/io.realm.test-1/lib/x86/librealm-jni.so (_ZN5realm11SharedGroup15do_advance_readINS_5_impl23NullInstructionObserverEEEbPT_NS0_9VersionIDERNS2_7HistoryE.constprop.1544+316)
I/DEBUG   ( 1139):     #16 pc 00024a11  /data/app/io.realm.test-1/lib/x86/librealm-jni.so (realm::LangBindHelper::advance_read(realm::SharedGroup&, realm::SharedGroup::VersionID)+209)
I/DEBUG   ( 1139):     #17 pc 00033f8c  /data/app/io.realm.test-1/lib/x86/librealm-jni.so (handoverQueryToWorker(long long, long long, bool)+220)
I/DEBUG   ( 1139):     #18 pc 00039ff0  /data/app/io.realm.test-1/lib/x86/librealm-jni.so (Java_io_realm_internal_TableQuery_nativeFindAllWithHandover+135)
I/DEBUG   ( 1139):     #19 pc 0066f38b  /data/dalvik-cache/x86/data@app@io.realm.test-1@base.apk@classes.dex
W/ActivityManager( 1500): Error in app io.realm.test running instrumentation ComponentInfo{io.realm.test/android.support.test.runner.AndroidJUnitRunner}:
W/ActivityManager( 1500):   Native crash
```

realm/realm-jni/src/io_realm_internal_TableView.cpp
```diff
@@ -148,10 +150,7 @@ JNIEXPORT void JNICALL Java_io_realm_internal_TableView_nativePivot(
 JNIEXPORT void JNICALL Java_io_realm_internal_TableView_nativeClose(
     JNIEnv*, jclass, jlong nativeViewPtr)
 {
-    if (nativeViewPtr == 0)
```
bmunkholm: This code is not transferred to the new finalize_table_view(). Are you ensuring nativeViewPtr is not 0 somewhere else?

realm/realm-library/src/androidTest/java/io/realm/internal/JNICloseTest.java
```diff
@@ -1,183 +0,0 @@
-/*
```
bmunkholm: Where else are you checking behaviour of using realm after it's been closed?

realm/realm-jni/src/io_realm_internal_TableView.cpp
beeender: It should never be 0. If it is, then it is a bug in our side.

realm/realm-library/src/androidTest/java/io/realm/internal/JNICloseTest.java
beeender: GCTests

realm/realm-library/src/main/java/io/realm/internal/LinkView.java
```diff
@@ -21,20 +21,32 @@
 /**
  * The LinkView class represents a core {@link RealmFieldType#LIST}.
  */
-public class LinkView extends NativeObject {
+public class LinkView implements NativeObject {
 
     private final Context context;
     final Table parent;
     final long columnIndexInParent;
+    private final long nativePointer;
```
bmunkholm: Keep it consistent with other use and make it more explicit: nativeLinkViewPtr

realm/realm-library/src/main/java/io/realm/internal/LinkView.java
```diff
@@ -21,20 +21,32 @@
...
+    private static final long nativeFinalizerPointer = nativeGetFinalizer();
```
bmunkholm: -> nativeFinalizerPtr for consistency

realm/realm-library/src/main/java/io/realm/internal/NativeObject.java
```diff
@@ -19,8 +19,9 @@
...
  * This abstract class represents a native object from core.
  * It specifies the operations common to all such objects.
- * All Java classes wrapping a core class should extend NativeObject.
+ * All Java classes wrapping a core class should implement NativeObject.
  */
-public abstract class NativeObject {
-    long nativePointer;
+interface NativeObject {
+    long getNativePointer();
+    long getNativeFinalizer();
```
bmunkholm: ->getNativeFinalizerPtr()

realm/realm-library/src/main/java/io/realm/internal/NativeObject.java
```diff
@@ -19,8 +19,9 @@
 /**
  * This abstract class represents a native object from core.
  * It specifies the operations common to all such objects.
- * All Java classes wrapping a core class should extend NativeObject.
+ * All Java classes wrapping a core class should implement NativeObject.
  */
-public abstract class NativeObject {
-    long nativePointer;
+interface NativeObject {
+    long getNativePointer();
```
bmunkholm: -> getNativePtr()

beeender:
Back to in progress.
1. Found a crash with low end devices
2. I think it is possible to totally remove the `Context with a small modification in core side.

beeender:
@realm/java We can start review this. Removing the context will be another PR when I got time.

realm/realm-library/src/main/java/io/realm/internal/LinkView.java
beeender: will fix

realm/realm-jni/src/io_realm_internal_LinkView.cpp
kneth: Yeah, especially when we move to C++14 and start using `auto`.

realm/realm-jni/src/io_realm_internal_LinkView.cpp
```diff
@@ -248,3 +250,15 @@ JNIEXPORT void JNICALL Java_io_realm_internal_LinkView_nativeRemoveTargetRow
         return lvr->remove_target_row( S(pos) );
     } CATCH_STD()
 }
+
+static void finalize_link_view(jlong ptr)
+{
+    TR_ENTER_PTR(ptr)
```
kneth: I would prefer to have the `TR_ENTER` and `TR_ENTER_PTR` macros at the JNI border. The macros writes the function's name to the log, and `finalize_link_view` provide us with less information compare to, say, `Java_io_realm_internal_LinkView_nativeClose`.

realm/realm-library/src/main/java/io/realm/internal/Context.java
```diff
@@ -17,135 +17,22 @@
 package io.realm.internal;
 
 import java.lang.ref.ReferenceQueue;
-import java.util.ArrayList;
-import java.util.List;
 
+// Currently we free native objects in two threads, the SharedGroup is freed in the caller thread, others are freed in
```
kneth: @finnschiermer Please read this. Can core help us?

beeender: Related test cases have been added to core. https://github.com/realm/realm-core/blob/master/test/test_destructor_thread_safety.cpp

finnschiermer: Destruction of SharedGroup (and hence Group and Table) is currently not thread-safe with respect to destruction of other accessors, you have two ensure mutual exclusion. This is also illustrated by the use of locks in the testcases @beeender mentions above. Explicit call of SharedGroup::close() or Table::detach() is also not thread-safe with respect to destruction of other accessors.

realm/realm-jni/src/io_realm_internal_LinkView.cpp
beeender: @kneth This could be called through function pointer. And actually we don't need `nativeClose()` anymore since those thing are supposed to be freed by phantom thread. Shall I just remove those `nativeClose()`?

beeender: Hmm, after checking, the `LinkView.nativeClose` is called by bulk insertion for speed. We need to keep that.

realm/realm-library/src/main/java/io/realm/internal/LinkView.java
beeender: since the interface is `getNativePtr`, so use `nativePtr` for all looks better. will change in that way.

Zhuinden:
Is there an ETA on this merge? I've seen lots of `Context.finalize()` so I expect this to add a lot of additional stability.

realm/realm-library/src/androidTest/java/io/realm/GCTests.java
```diff
@@ -0,0 +1,128 @@
...
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+import io.realm.entities.AllTypes;
+import io.realm.entities.Dog;
+import io.realm.rule.TestRealmConfigurationFactory;
+
+import static junit.framework.TestCase.assertNotNull;
+
+// This test is for the fact we don't have locks for native objects creation that when finalizer/phantom thread free the
```
nhachicha: can we trigger/assert the collection of references (using the daemon thread) to ensure the assumption.
> the same Realm could have some native objects creation at the same time

realm/realm-library/src/main/java/io/realm/internal/Context.java
```diff
@@ -17,127 +17,24 @@
...
+// PhantomDaemon thread. And the destruction in both threads are locked by the corresponding context.
+// The purpose of locking on Context is:
+// Destruction of SharedGroup (and hence Group and Table) is currently not thread-safe with respect to destruction of
+// other accessors, you have two ensure mutual exclusion. This is also illustrated by the use of locks in the test
```
nhachicha: you have to

realm/realm-library/src/main/java/io/realm/internal/Context.java
```diff
@@ -17,127 +17,24 @@
...
+// test_destructor_thread_safety.cpp. Explicit call of SharedGroup::close() or Table::detach() is also not thread-safe
+// with respect to destruction of other accessors.
 public class Context {
+    private final static ReferenceQueue<NativeObject> referenceQueue = new ReferenceQueue<NativeObject>();
+    private final static Thread phantomThread = new Thread(new FinalizerRunnable(referenceQueue));
```
nhachicha: minor: scary name 👻 I know it's soon 🎃 but maybe `gcThread` is more appropriate?

realm/realm-library/src/main/java/io/realm/internal/Context.java
```diff
@@ -17,127 +17,24 @@
...
-
-    public void asyncDisposeQuery(long nativePointer) {
-        if (isFinalized) {
-            TableQuery.nativeClose(nativePointer);
-        }
-        else {
-            abandonedQueries.add(nativePointer);
-        }
+    static {
+        phantomThread.setName("RealmPhantomDaemon");
```
nhachicha: ditto

realm/realm-library/src/main/java/io/realm/internal/NativeObjectReference.java
```diff
@@ -23,43 +23,70 @@
...
+        this.nativePtr = referent.getNativePtr();
+        this.nativeFinalizerPtr = referent.getNativeFinalizerPtr();
+        this.context = context;
+        referencePool.add(this);
     }
 
     /**
      * To dealloc native resources.
      */
     void cleanup() {
```
nhachicha: For some implementation reasons, I think it's also advised to call https://developer.android.com/reference/java/lang/ref/Reference.html#clear()

> Unlike soft and weak references, phantom references are not automatically cleared by the garbage collector as they are enqueued
from http://docs.oracle.com/javase/6/docs/api/java/lang/ref/PhantomReference.html

realm/realm-library/src/main/java/io/realm/internal/LinkView.java
```diff
@@ -21,20 +21,32 @@
...
  * The LinkView class represents a core {@link RealmFieldType#LIST}.
  */
-public class LinkView extends NativeObject {
+public class LinkView implements NativeObject {
 
     private final Context context;
     final Table parent;
     final long columnIndexInParent;
+    private final long nativePtr;
+    private static final long nativeFinalizerPtr = nativeGetFinalizerPtr();
```
nhachicha: why static? is this shared between `LinkView`?

realm/realm-library/src/main/java/io/realm/internal/NativeObject.java
```diff
@@ -19,8 +19,9 @@
 /**
  * This abstract class represents a native object from core.
  * It specifies the operations common to all such objects.
- * All Java classes wrapping a core class should extend NativeObject.
+ * All Java classes wrapping a core class should implement NativeObject.
  */
-public abstract class NativeObject {
-    long nativePointer;
+interface NativeObject {
```
nhachicha: any Comment/Javadoc explaining the purpose of these methods?

realm/realm-library/src/main/java/io/realm/internal/NativeObjectReference.java
```diff
@@ -23,43 +23,70 @@
...
  * This is required as phantom references cannot access the original objects for this value.
  */
-public final class NativeObjectReference extends PhantomReference<NativeObject> {
+final class NativeObjectReference extends PhantomReference<NativeObject> {
 
-    // Using int here instead of enum to make it faster since the cleanup needs to be called
-    // in a loop to dealloc every native reference.
-    public static final int TYPE_LINK_VIEW = 0;
-    public static final int TYPE_ROW = 1;
+    // Linked list to keep the reference of the PhantomReference
```
nhachicha: why not use/extends `	java.util.LinkedList<NativeObjectReference>`?

realm/realm-library/src/main/java/io/realm/internal/Table.java
```diff
@@ -43,11 +43,16 @@
     private static final long PRIMARY_KEY_FIELD_COLUMN_INDEX = 1;
     private static final long NO_PRIMARY_KEY = -2;
 
-    long nativePtr;
+    protected long nativePtr;
+    private static final long nativeFinalizerPtr;
```
nhachicha: why static is the function pointer the same for all `Table`?

realm/realm-library/src/main/java/io/realm/internal/NativeObjectReference.java
beeender: I think i tried `java.util.LinkedList`, but that was much slower than this custom implementation, mainly because of it adds overhead by some extra features we don't need here.

realm/realm-library/src/main/java/io/realm/internal/Table.java
beeender: yes!

realm/realm-library/src/main/java/io/realm/internal/LinkView.java
beeender: Yes, it just returns an function pointer, all `LinkView` share the same one.

realm/realm-library/src/main/java/io/realm/internal/NativeObjectReference.java
beeender: > Clears this reference object. Invoking this method will not cause this object to be enqueued.
This method is invoked only by Java code; when the garbage collector clears references it does so directly, without invoking this method.

The `Reference.clear()` has a slightly different meaning than the `cleanup()` here. Although the `PhantomReference.clear()` is not called by the jvm. I prefer to still using a different method than it ...

realm/realm-library/src/main/cpp/io_realm_internal_LinkView.cpp
```diff
@@ -19,10 +19,13 @@
 
 using namespace realm;
 
+static void finalize_link_view(jlong ptr);
+
 JNIEXPORT void JNICALL Java_io_realm_internal_LinkView_nativeClose
```
beeender: @nhachicha This is only called by the bulk insert. I think we could remove the calling from there and just rely on GC to free it?

beeender: Also, it should make the insertion even faster.

realm/realm-library/src/main/java/io/realm/internal/Context.java
```diff
@@ -17,127 +17,24 @@
...
-
-    public void asyncDisposeQuery(long nativePointer) {
-        if (isFinalized) {
-            TableQuery.nativeClose(nativePointer);
-        }
-        else {
-            abandonedQueries.add(nativePointer);
-        }
+    static {
+        finalizingThread.setName("RealmFinalizingDaemon");
```
zaki50: Shouldn't we make `finalizingThread` as a daemon thread?

realm/realm-library/src/main/java/io/realm/internal/Context.java
```diff
@@ -17,127 +17,24 @@
 package io.realm.internal;
 
 import java.lang.ref.ReferenceQueue;
-import java.util.ArrayList;
-import java.util.List;
 
+// Currently we free native objects in two threads, the SharedGroup is freed in the caller thread, others are freed in
+// PhantomDaemon thread. And the destruction in both threads are locked by the corresponding context.
```
zaki50: `PhantomDaemon` -> `RealmFinalizingDaemon`

realm/realm-library/src/main/java/io/realm/internal/Context.java
beeender: I have thought about that, but i don't really know what is benefit of daemon thread in Android. https://groups.google.com/forum/#!topic/android-developers/E67KirHZd3I It seems no difference on Android?

zaki50: No actual difference on Android.
I just thought that if we treat a thread as a daemon thread, we should mark it as daemon.

zaki50:
@beeender Reviewed. My comments other than https://github.com/realm/realm-java/pull/3144#pullrequestreview-6231524 is all minor.

If https://github.com/realm/realm-java/pull/3144#pullrequestreview-6231524 is OK, I'll give :+1:

realm/realm-library/src/main/java/io/realm/internal/NativeObjectReference.java
nhachicha: why not call `clear` after your remove the PhantomReference from the queue?
```java
// Remove the PhantomReference from the pool to free it.
        referencePool.remove(this);
        clear();
```

realm/realm-library/src/main/cpp/io_realm_internal_LinkView.cpp
nhachicha: getting the LinkView uses dynamic allocation `new LinkViewRef` I'm not sure the GC will be able to close/delete the reference without an explicit call to close, If it's the case (using `finalize_link_view`) I'm ok with the refactor.

realm/realm-library/src/main/java/io/realm/internal/Table.java
```diff
@@ -61,7 +66,8 @@ public Table() {
         if (nativePtr == 0) {
             throw new java.lang.OutOfMemoryError("Out of native memory.");
         }
-        sharedRealm = null;
+        this.sharedRealm = null;
+        context.addReference(this);
```
beeender: The Context is created while the root component is created. This constructor is mainly to support the case that a Table can be created without any Group or SharedGroup. Not a real case but only for testing purpose. We can remove this constructor in the future. So, the rule is like when your table is created from SharedRealm, the context is created by the SharedRealm since that is the root component and everything branched out from that should be synced with that context when it get destructed. For the Table created from nowhere, like here, the Table itself it he root component.

zaki50: I saw this constructor is used somewhere in our library code(I'm now mobile).

zaki50: https://github.com/realm/realm-java/blob/master/realm/realm-library/src/main/java/io/realm/internal/TableView.java#L745

GaborPeto:
When is this expected to be merged and released?

realm/realm-library/src/main/cpp/io_realm_internal_LinkView.cpp
beeender: removed `LinkView.nativeClose()`.

realm/realm-library/src/main/java/io/realm/internal/NativeObjectReference.java
beeender:
```
public void clear() {
        this.referent = null;
}
```
The `Reference.clear()` implementation is like above which is not necessary since the ref will be GCed after `cleanup()` called. Avoid calling it to make the finalizer daemon loop faster.

realm/realm-library/src/main/java/io/realm/internal/NativeObjectReference.java
cmelchior: Perhaps add that as a small comment above the class?

realm/realm-library/src/main/java/io/realm/internal/NativeObjectReference.java
```diff
@@ -23,43 +23,70 @@
...
-            case TYPE_LINK_VIEW:
-                LinkView.nativeClose(nativePointer);
-                break;
-            case TYPE_ROW:
-                UncheckedRow.nativeClose(nativePointer);
-                break;
-            default:
-                // Cannot get here.
-                throw new IllegalStateException("Unknown native reference type " + type + ".");
+        synchronized (context) {
```
cmelchior: Is this required? There doesn't seem to be any other synchronizing on context going on?

realm/realm-library/src/main/java/io/realm/internal/Table.java
```diff
@@ -43,11 +43,16 @@
...
     private static final long NO_PRIMARY_KEY = -2;
 
-    long nativePtr;
+    protected long nativePtr;
+    private static final long nativeFinalizerPtr;
     private final Context context;
     private final SharedRealm sharedRealm;
     private long cachedPrimaryKeyColumnIndex = NO_MATCH;
 
+    static {
```
cmelchior: Why in a block instead of just setting it directly on the field?

beeender: There might be a reason but I cannot remember ... maybe my brain was shorted at that time ... I will change it to setting directly.

realm/realm-library/src/main/cpp/io_realm_internal_TableQuery.cpp
```diff
@@ -1912,3 +1909,17 @@ JNIEXPORT void JNICALL Java_io_realm_internal_TableQuery_nativeIsEmpty
...
+{
+    TR_ENTER_PTR(ptr)
+    delete Q(ptr);
+}
+
+JNIEXPORT jlong JNICALL Java_io_realm_internal_TableQuery_nativeGetFinalizerPtr
+  (JNIEnv *, jclass)
+{
+    TR_ENTER()
+    return static_cast<jlong>(reinterpret_cast<uintptr_t>(&finalize_table_query));
```
kneth: Can't we `reinterpret_cast<jlong>(...)` and avoid one cast?

beeender: nope, we cannot. See http://stackoverflow.com/questions/1096341/function-pointers-casting-in-c

kneth: Thanks for educating me :-)

beeender: Cannot get the `std::function` right. Changed to what you suggested.

Zhuinden:
https://github.com/realm/realm-java/pull/3840

PR #3146, opened 2016-07-06
Fix a bunch of lint warnings

emanuelez:


realm/realm-library/src/androidTest/java/io/realm/DynamicRealmTests.java
```diff
@@ -54,7 +54,7 @@
 public class DynamicRealmTests {
 
     @Rule
-    public final RunInLooperThread looperThread = new RunInLooperThread();
+    public static final RunInLooperThread looperThread = new RunInLooperThread();
```
zaki50: Why static?

PR #3149, opened 2016-07-06
RealmObjectSchema.setClassName() transfers primary key for new class name

### LINKED ISSUE #3118: RealmObjectSchema.setClassName() should amend pkTable entry
#### Goal

We have covered cases where `@PrimaryKey` field is renamed or `RealmSchema.rename()` is involved.
When [`RealmObjectSchema.setClassName()`](https://github.com/realm/realm-java/blob/master/realm/realm-library/src/main/java/io/realm/RealmObjectSchema.java#L108) is called, however, it has potential to lose track of `@PrimaryKey` of a `Table`. This is addressed by @beeender and should be fixed.

#### Expected Results

`RealmObjectSchema.setClassName()` is called, `pkTable` should be adjusted accordingly.

#### Actual Results

`pkTable` is not adjusted as of now [3ff5741 (July 4th, 2016)](https://github.com/realm/realm-java/commit/3ff5741582e8bb550f66b1b4adcc925027c0dce6).

N/A

Realm version(s): 1.1.0
Android Studio version: N/A

stk1m1:
`RealmObjectSchema.setClassName()` transfers a primary key for a new class name that the renamed class can maintain the old primary key. Closes #3118.

@realm/java


stk1m1:
retest this please

stk1m1:
retest this please

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 * Fixed a wrong JNI method declaration which might cause "method not found" crash on some devices.
 * Fixed a bug that `Error` in the background async thread is not forwared to the caller thread.
 * Fixed a crash when an empty `Collection` is passed to `insert()`/`insertOrUpdate()` (#3103).
+* Fixed a bug that does not transfer a primary key when `RealmSchemaObject.setClassName()` is called to rename a class (#3118).
```
cmelchior: a primary key -> the primary key

realm/realm-library/src/androidTest/java/io/realm/RealmMigrationTests.java
```diff
@@ -293,6 +293,35 @@ public void migrate(DynamicRealm realm, long oldVersion, long newVersion) {
...
+                .build();
+        Realm realm = Realm.getInstance(realmConfig);
+
+        Table table = realm.getSchema().getTable(MigrationClassRenamed.class);
+        assertTrue(table.hasPrimaryKey());
+        assertEquals(MigrationClassRenamed.DEFAULT_FIELDS_COUNT, table.getColumnCount());
+        assertEquals(MigrationClassRenamed.DEFAULT_PRIMARY_INDEX, table.getPrimaryKey());
+        assertEquals(MigrationClassRenamed.FIELD_PRIMARY, table.getColumnName(table.getPrimaryKey()));
+        //old schema does not exists
+        assertNull(realm.getSchema().get(MigrationPrimaryKey.CLASS_NAME));
```
cmelchior: Wouldn't the problem be that if you recreated MigrationPrimaryKey with a similar named field that wasn't a primary key, then it would still think that field was a primary key?

realm/realm-library/src/main/java/io/realm/RealmObjectSchema.java
```diff
@@ -111,7 +112,24 @@ public RealmObjectSchema setClassName(String className) {
         if (transaction.hasTable(internalTableName)) {
             throw new IllegalArgumentException("Class already exists: " + className);
         }
+        // in case if this table has a primary key, we need to transfer it after renaming the table.
```
cmelchior: in case this table

realm/realm-library/src/main/java/io/realm/RealmObjectSchema.java
```diff
@@ -111,7 +112,24 @@ public RealmObjectSchema setClassName(String className) {
...
+        String oldTableName = null;
+        String pkField = null;
+        if (table.hasPrimaryKey()) {
+            oldTableName = table.getName();
+            pkField = getPrimaryKey();
+        }
+        // this will throw when things go wrong
```
cmelchior: Either make a more detailed description exactly how this could throw or remove it.

realm/realm-library/src/main/java/io/realm/RealmObjectSchema.java
```diff
@@ -111,7 +112,24 @@ public RealmObjectSchema setClassName(String className) {
...
         transaction.renameTable(table.getName(), internalTableName);
+        if (pkField != null && !pkField.isEmpty()) {
+            try {
+                table.resetPrimaryKeyForRenamedClass(oldTableName, internalTableName);
+            } catch (Exception e) {
+                // revert the table name back when something goes wrong
+                transaction.renameTable(table.getName(), oldTableName);
```
cmelchior: Nice 👍

realm/realm-library/src/main/java/io/realm/internal/Table.java
```diff
@@ -263,6 +263,27 @@ public void renameColumn(long columnIndex, String newName) {
     }
 
     /**
+     * When a {@link Table} is renamed, it is necessary to set the {@link Table#PRIMARY_KEY_CLASS_COLUMN_INDEX}
+     * entry of its primary Key table with a new class name to transfer the primary key.
+     *
+     * @param oldTableName the old table name to search its primary key.
+     * @param newTableName the new table name to reset the primary key
+     */
+    public void resetPrimaryKeyForRenamedClass(String oldTableName, String newTableName) {
```
cmelchior: `reset` seems the wrong word here, perhaps "movePrimaryKeyToRenamedClass" instead?

realm/realm-library/src/main/java/io/realm/internal/Table.java
```diff
@@ -263,6 +263,27 @@ public void renameColumn(long columnIndex, String newName) {
...
+        // make sure the new table is suitable for column
+        verifyColumnName(newTableName);
```
cmelchior: IMO this checked as soon as possible, e.g. right after checking if the table already exists.

realm/realm-library/src/main/java/io/realm/internal/Table.java
```diff
@@ -263,6 +263,27 @@ public void renameColumn(long columnIndex, String newName) {
...
+        String newClassName = tableNameToClassName(newTableName);
+        String oldClassName = tableNameToClassName(oldTableName);
+        Table pkTable = getPrimaryKeyTable();
+        long pkRowIndex = pkTable.findFirstString(PRIMARY_KEY_CLASS_COLUMN_INDEX, oldClassName);
```
cmelchior: This is implementing logic that already exists. It would be much better if you used the already existing `setPrimaryKey()` function for removing/adding the keys. That way any extra rules will automatically be enforced. It probably means this method needs to be restructured/removed though.

realm/realm-library/src/androidTest/java/io/realm/RealmMigrationTests.java
stk1m1: I'd guide your attention to a case where there are two ways to rename a class, `RealmSchema.rename()` & `RealmSchemaObject.setClassName()`, and this is to address the latter one. 😉

realm/realm-library/src/main/java/io/realm/internal/Table.java
stk1m1: I'll go with `setPrimaryKey()`. 👍

realm/realm-library/src/androidTest/java/io/realm/RealmMigrationTests.java
cmelchior: I know, but that doesn't change the problem. Before this PR, if you did:
```
realmSchema.get("Dog").setClassName("Cat")
realmSchema.create("Dog").addField("id", int.class);
```

If `id` was the primary key in the old `Dog`, Realm would still falsely think that the new Dog also had a primary key for that field, right?

stk1m1: I think I have to blame my short-sighted understanding of this issue. Yes, I should cover that one with a test.

kneth:
Since @cmelchior did all the hard work of commenting, I can only say it looks good.

stk1m1:
retest this please

zaki50:
retest this please

stk1m1:
retest this please

stk1m1:
retest this please

stk1m1:
retest this please

realm/realm-library/src/main/java/io/realm/RealmObjectSchema.java
```diff
@@ -100,18 +100,40 @@ public String getClassName() {
...
+     * Sets a new name for this RealmObject class. This is equivalent to renaming it. When {@link RealmObjectSchema#table}
+     * has a primary key, this will transfer the primary key for the new class name.
      *
      * @param className the new name for this class.
+     * @throws IllegalArgumentException if {@param className} is {@code null} or an empty string, or its length exceeds 57 characters.
      * @see RealmSchema#rename(String, String)
      */
     public RealmObjectSchema setClassName(String className) {
         checkEmpty(className);
+        if (57 < className.length()) {
```
beeender: Doesn't it throw from core? I think we don't have to check it in java?

beeender:
except one comment. :+1:

realm/realm-library/src/main/java/io/realm/RealmObjectSchema.java
stk1m1: I'm afraid [core doesn't](https://github.com/realm/realm-core/blob/a38f2d69da4c300b116753cd5a8f7dafc0811ad0/src/realm/group.cpp#L630) 😒

beeender: Doesn't it? https://github.com/realm/realm-core/blob/a38f2d69da4c300b116753cd5a8f7dafc0811ad0/src/realm/replication.cpp#L530

But not sure if the macro is enabled in the realm mode.

stk1m1: You're right. It checks size. 😊

realm/realm-library/src/androidTest/java/io/realm/RealmMigrationTests.java
```diff
@@ -289,10 +289,106 @@ public void migrate(DynamicRealm realm, long oldVersion, long newVersion) {
...
+                .schema(MigrationClassRenamed.class)
+                .migration(migration)
+                .build();
+        Realm realm = Realm.getInstance(realmConfig);
+
+        assertTrue(realm.getSchema().get(MigrationClassRenamed.CLASS_NAME).hasPrimaryKey());
+        assertFalse(realm.getSchema().get(MigrationPrimaryKey.CLASS_NAME).hasPrimaryKey());
+    }
+
+    // Test to show renaming a class does not hinder its PK field's attribute
```
cmelchior: Not entirely sure what this means? Perhaps "Test to show that renaming a class does not effect the primary key"?

realm/realm-library/src/main/java/io/realm/RealmObjectSchema.java
cmelchior: Please avoid magic constants in code. Perhaps extract this into a named constant like perhaps `CLASSNAME_MAX_LENGTH = 64 - TABLE_PREFIX.length`
or something like that.

realm/realm-library/src/main/java/io/realm/RealmObjectSchema.java
```diff
@@ -100,18 +100,40 @@ public String getClassName() {
...
+            throw new IllegalArgumentException("Class name length exceeds 57 characters.");
```
cmelchior: For better errors also output the class name, for bonus points also the length of that 😄 
`Class name length exceeds 57 characters: foo (3)`

cmelchior:
Except a few minor comments, this looks 👍

realm/realm-library/src/main/java/io/realm/RealmObjectSchema.java
stk1m1: Actually, this has a deeper issue at Core level. As @beeender points out, Core is supposed to impose 64 chars length of hard-limit on table name including `class_` prefix. The char length is supposed to be checked and thrown when violated due to consistent behavior across bindings.

What happens now is that we can actually insert a class name longer than 57 chars and a table name longer than 64 chars as [_the checker_](https://github.com/realm/realm-core/blob/master/src/realm/replication.cpp#L530) is not working. As soon as the char length checker at Java binding is gone, we can actually do following, which means the bug exists as we speak.

@beeender is to look into this issue in detail. I'd ask we take some time.
screen shot 2016-07-29 at 6 50 43 pm

cmelchior: The core issue is unrelated to the PR, no? 
I mean ideally we would want Core to make the check, but since you are already implementing it in Java, we might as well do it now and just add a FIXME for changing it when core actually does a proper check?

Or the Object Store if we get around to adopting it first.

stk1m1:
@realm/java going for 2nd round. All the reviews are addressed.

cmelchior:
@stk1m1 FindBugs didn't like that :) Caused by: org.gradle.api.GradleException: FindBugs rule violations were found. See the report at: file:///tmp/realm-java/realm/realm-library/build/findbugs/findbugs-output.html

realm/realm-library/src/androidTest/java/io/realm/RealmMigrationTests.java
```diff
@@ -389,6 +389,43 @@ public void migrate(DynamicRealm realm, long oldVersion, long newVersion) {
...
+                .schemaVersion(1)
+                .migration(migration)
+                .build();
+
+        // create Realm instance fails
+        try {
+            Realm.getInstance(realmConfig);
+            fail();
+        } catch (IllegalArgumentException expected) {
+            assertEquals("Class name is to long. Limit is 57 characters: 'MigrationNameIsLongerThan56charThisShouldThrowAnException' (57)",
```
cmelchior: 💯 for that word being 57 chars long 😁 

That said, the sentence is a bit weird. If the limit is 57 I would normally read that as inclusive, so we should probably rephrase it to "Limit is 56 characters..."

cmelchior:
Apart from comment 👍

realm/realm-library/src/androidTest/java/io/realm/RealmMigrationTests.java
stk1m1: I think it should. 

... renamed. ;)

stk1m1:
retest this please

stk1m1:
CI:retest

stk1m1:
retest this please

stk1m1:
retest this please

stk1m1:
retest this please.

stk1m1:
retest this please

PR #3150, opened 2016-07-06
Don't rely on dir() in docker containers in the Jenkinsfile

emanuelez:
Also, we need to install `hexdump` on our docker images in order to perform the method count.


emanuelez:
@cmelchior @alebsack

cmelchior:
👍

PR #3153, opened 2016-07-07
Fix some issues regarding Jenkinsfile and master

emanuelez:
Handle Bintray credentials and the monitoring of metrics


emanuelez:
Instamerge. The changes can only be tested in master

PR #3158, opened 2016-07-07
Fix metrics collection

emanuelez:


emanuelez:
Instamerge because already tested extensively

PR #3204, opened 2016-07-22
Improve formatting of CHANGELOG.md

dalinaum:


bmunkholm:
The `### Internal` headline is not used correctly as it also lists the very public bugfixes as internal.
There is hardly much value in changing all the old release notes. I would suggest to just update the last one and continue with that style in the future.

dalinaum:
I see. I will close this PR.

PR #3221, opened 2016-07-28
Add quotes around terms in exceptions message

### LINKED ISSUE #3217: Add quotes around field names in RealmProxy exceptions
**Feature Request**

#### Goal

Read an exception regarding a field and determing which field is being referred to. The field's name is `code`.

#### Expected Results

Immediate recognition that the problematic field is `code`.

#### Actual Results

Temporary confusion about message.

It is arguable that `code` is a bad name for a field, but I'm sure there are others that are just as problematic.

I ask that you simply change the exception string from:

```java
throw new IllegalArgumentException("Trying to set non-nullable field code to null.");
```

to:

```java
throw new IllegalArgumentException("Trying to set non-nullable field 'code' to null.");
```

This is the error I came across, but you may want to go through other generated error messages to see if there is a similar issue.

beeender:
Close #3217


stk1m1:
Does this pass unit tests? 😜

cmelchior:
👍

PR #3222, opened 2016-07-29
Enable armv7a only to test CI

beeender:


cmelchior:
I assume this was a test? Close it?

PR #3224, opened 2016-07-29
Fix flaky test

beeender:


realm/realm-library/src/androidTest/java/io/realm/NotificationsTest.java
```diff
@@ -686,38 +687,30 @@ public void run() {
                 RealmChangeListener<Realm> listener = new RealmChangeListener<Realm>() {
                     @Override
                     public void onChange(Realm object) {
-                        try {
-                            fail("This handler should not be notified");
-                        } catch (AssertionFailedError e) {
-                            threadAssertionError[0] = e;
-                            handlerNotified.countDown(); // Make sure that that await() doesn't fail instead.
-                        }
+                        fail("This handler should not be notified");
```
cmelchior: Are you sure this will be reported correctly on the test thread? If I remember correctly the above code is more accurate (and also what we are doing in the `RunTestInLooperThread` class.

cmelchior:
This unit test still have concurrency problems IMO as we are trying to prove that something _didn't_ happen. Would one solution be to trigger 2 change listeners on the first test thread. Then you would be 100% sure that the closed handler would have been triggered (and failed).

realm/realm-library/src/androidTest/java/io/realm/NotificationsTest.java
beeender: Yes, verified, with `@RunWith(AndroidJUnit4.class)` it would just fail in the background thread. The behaviour you mentioned is the old android test behaviour.

cmelchior:
👍

PR #3230, opened 2016-08-01
Only execute the builds on slaves with the android label

emanuelez:


emanuelez:
@realm/java Instamerge

PR #3232, opened 2016-08-02
Disable change listeners on IntentService

### LINKED ISSUE #2875: Detect when adding a ChangeListener on an IntentService
IntentServices are Looper Threads, which means that Realm currently allow you to register change listeners there. However since every invocation of `onHandleIntent` is a single event. Registering a change listener will not work, since the Realm should be closed when exiting `onHandleIntent`.

We might be able to implement a heuristic that checks if the current thread is named `IntentServicer[XXX]` the same way we test for Looper threads now, and throw an exception in that case:

https://android.googlesource.com/platform/frameworks/base/+/master/core/java/android/app/IntentService.java#108

cmelchior:
Fixes #2875 

With this change, using any form of ChangeListener will now throw on an IntentService thread.

IntentService has always been a bastard child in the sense that it is a  Looper thread, but is not meant to be used as such, since the looping is just used to process the queue of jobs.

- All logic around the auto-refresh feature has also been moved to the HandlerController instead of being scattered all over the place.

- `RunTestInLooperThread` have been extended so it is now possible to define the name of the worker thread.

### DIFF
--- a/realm/realm-library/src/main/java/io/realm/HandlerController.java
+++ b/realm/realm-library/src/main/java/io/realm/HandlerController.java
@@ -769,10 +769,11 @@ final class HandlerController implements Handler.Callback {
         }
     }
 
+    /**
+     * Toggles the auto refresh flag. Will throw an {@link IllegalStateException} if auto-refresh is not available.
+     */
     public void setAutoRefresh(boolean autoRefresh) {
-        if (autoRefresh && Looper.myLooper() == null) {
-            throw new IllegalStateException("Cannot enabled autorefresh on a non-looper thread.");
-        }
+        checkCanBeAutoRefreshed();
         this.autoRefresh = autoRefresh;
     }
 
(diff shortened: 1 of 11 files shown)


cmelchior:
Ready for review @realm/java

realm/realm-library/src/main/java/io/realm/HandlerController.java
```diff
@@ -769,15 +769,48 @@ public void refreshSynchronousTableViews() {
...
+        checkCanBeAutoRefreshed();
         this.autoRefresh = autoRefresh;
     }
 
     public boolean isAutoRefreshEnabled() {
         return autoRefresh;
     }
 
+    /**
+     * Validate that the current thread can enable auto refresh. An {@link IllegalStateException} will be thrown if that
```
beeender: Validate -> Validates

beeender:
:+1:

stk1m1:
👍

realm/realm-library/src/main/java/io/realm/HandlerController.java
```diff
@@ -769,15 +769,48 @@ public void refreshSynchronousTableViews() {
...
+            return false;
+        }
+
+        return true;
+    }
+
+    private static boolean isIntentServiceThread() {
+        // Tries to determine if a thread is an IntentService thread. No public API can detect this,
+        // so use the thread name as a heuristic:
+        // https://android.googlesource.com/platform/frameworks/base/+/master/core/java/android/app/IntentService.java#108
```
kneth: cool

kneth:
:+1:

PR #3239, opened 2016-08-02
insert(): Correctly detect multiple objects with the same primary key 

### LINKED ISSUE #3212: @PrimaryKey is not handled correctly when using insert()

```java

public class PrimaryKeyActivity extends Activity {
    private static final String TAG = PrimaryKeyActivity.class.getSimpleName();

    @Override

        final Car car = new Car();
        car.wheels = new RealmList<>(new Wheel(), new Wheel());

        final Realm realm = Realm.getDefaultInstance();
        Log.d(TAG, "Number of wheels with id 1 before inserting: " + realm.where(Wheel.class).equalTo("id", 1).count());
        realm.beginTransaction();
        realm.insert(car);
        realm.commitTransaction();
        Log.d(TAG, "Number of wheels with id 1 after inserting: " + realm.where(Wheel.class).equalTo("id", 1).count());
        realm.close();
    }
}

public class Car extends RealmObject {
    @PrimaryKey
    private long id = 42;
    public RealmList<Wheel> wheels;
}

public class Wheel extends RealmObject {
    @PrimaryKey
    private long id = 1;
}

```

#### Expected log output:

>D/PrimaryKeyActivity: Number of wheels with id 1 before inserting: 0
>D/PrimaryKeyActivity: Number of wheels with id 1 after inserting: 1

#### Actiual log output:

>D/PrimaryKeyActivity: Number of wheels with id 1 before inserting: 0
>D/PrimaryKeyActivity: Number of wheels with id 1 after inserting: **2**

As you can see, we have two objects with the same primary key.
Note that `insertOrUpdate()` and `copyToRealm()` work correctly.

Realm version(s): 1.1.0

Android Studio version: 2.1.2

Which Android version and device: Android 4.2.2 on Samsung GT-I9105

cmelchior:
Fixes #3212 

This PR fixes a bug which happened when using `insert()` with two java instances with the same primary key. This now correctly throws a `RealmPrimaryKeyConstraintBrokenException`.

@realm/java

### DIFF
--- a/realm/realm-annotations-processor/src/test/resources/io/realm/AllTypesRealmProxy.java
+++ b/realm/realm-annotations-processor/src/test/resources/io/realm/AllTypesRealmProxy.java
@@ -672,7 +672,7 @@ public class AllTypesRealmProxy extends some.test.AllTypes
         String primaryKeyValue = ((AllTypesRealmProxyInterface) object).realmGet$columnString();
         long rowIndex = TableOrView.NO_MATCH;
         if (primaryKeyValue == null) {
-            rowIndex = table.findFirstNull(pkColumnIndex);
+            rowIndex = Table.nativeFindFirstNull(tableNativePtr, pkColumnIndex);
         } else {
             rowIndex = Table.nativeFindFirstString(tableNativePtr, pkColumnIndex, primaryKeyValue);
         }
@@ -735,7 +735,7 @@ public class AllTypesRealmProxy extends some.test.AllTypes
                 String primaryKeyValue = ((AllTypesRealmProxyInterface) object).realmGet$columnString();
                 long rowIndex = TableOrView.NO_MATCH;
                 if (primaryKeyValue == null) {
-                    rowIndex = table.findFirstNull(pkColumnIndex);
+                    rowIndex = Table.nativeFindFirstNull(tableNativePtr, pkColumnIndex);
                 } else {
                     rowIndex = Table.nativeFindFirstString(tableNativePtr, pkColumnIndex, primaryKeyValue);
                 }
@@ -795,7 +795,7 @@ public class AllTypesRealmProxy extends some.test.AllTypes
         String primaryKeyValue = ((AllTypesRealmProxyInterface) object).realmGet$columnString();
         long rowIndex = TableOrView.NO_MATCH;
         if (primaryKeyValue == null) {
-            rowIndex = table.findFirstNull(pkColumnIndex);
+            rowIndex = Table.nativeFindFirstNull(tableNativePtr, pkColumnIndex);
         } else {
             rowIndex = Table.nativeFindFirstString(tableNativePtr, pkColumnIndex, primaryKeyValue);
         }
@@ -863,7 +863,7 @@ public class AllTypesRealmProxy extends some.test.AllTypes
                 String primaryKeyValue = ((AllTypesRealmProxyInterface) object).realmGet$columnString();
                 long rowIndex = TableOrView.NO_MATCH;
                 if (primaryKeyValue == null) {
-                    rowIndex = table.findFirstNull(pkColumnIndex);
+                    rowIndex = Table.nativeFindFirstNull(tableNativePtr, pkColumnIndex);
                 } else {
                     rowIndex = Table.nativeFindFirstString(tableNativePtr, pkColumnIndex, primaryKeyValue);
                 }
--- a/realm/realm-annotations-processor/src/test/resources/io/realm/AllTypesRealmProxy.java
+++ b/realm/realm-annotations-processor/src/test/resources/io/realm/AllTypesRealmProxy.java
@@ -682,7 +682,7 @@ public class AllTypesRealmProxy extends some.test.AllTypes
                 Table.nativeSetString(tableNativePtr, pkColumnIndex, rowIndex, (String)primaryKeyValue);
             }
         } else {
-            table.throwDuplicatePrimaryKeyException(primaryKeyValue);
+            Table.throwDuplicatePrimaryKeyException(primaryKeyValue);
         }
         cache.put(object, rowIndex);
         Table.nativeSetLong(tableNativePtr, columnInfo.columnLongIndex, rowIndex, ((AllTypesRealmProxyInterface)object).realmGet$columnLong());
@@ -745,7 +745,7 @@ public class AllTypesRealmProxy extends some.test.AllTypes
                         Table.nativeSetString(tableNativePtr, pkColumnIndex, rowIndex, (String)primaryKeyValue);
                     }
                 } else {
-                    table.throwDuplicatePrimaryKeyException(primaryKeyValue);
+                    Table.throwDuplicatePrimaryKeyException(primaryKeyValue);
                 }
                 cache.put(object, rowIndex);
                 Table.nativeSetLong(tableNativePtr, columnInfo.columnLongIndex, rowIndex, ((AllTypesRealmProxyInterface)object).realmGet$columnLong());
--- a/realm/realm-annotations-processor/src/test/resources/io/realm/AllTypesRealmProxy.java
+++ b/realm/realm-annotations-processor/src/test/resources/io/realm/AllTypesRealmProxy.java
@@ -668,12 +668,23 @@ public class AllTypesRealmProxy extends some.test.AllTypes
         Table table = realm.getTable(some.test.AllTypes.class);
         long tableNativePtr = table.getNativeTablePointer();
         AllTypesColumnInfo columnInfo = (AllTypesColumnInfo) realm.schema.getColumnInfo(some.test.AllTypes.class);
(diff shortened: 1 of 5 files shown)


realm/realm-library/src/androidTest/java/io/realm/BulkInsertTests.java
```diff
@@ -317,7 +319,33 @@ public void insertToRealm_NullPrimaryKey() {
...
+    public void insert_duplicatedPrimaryKeyFails() {
+
+        // Single object with 2 references to two objects with the same ID
+        AllJavaTypes obj = new AllJavaTypes(2);
+        obj.setFieldList(new RealmList<AllJavaTypes>(new AllJavaTypes(1), new AllJavaTypes(1)));
+        realm.beginTransaction();
+        try {
+            realm.insert(obj);
+            fail();
+        } catch (RealmPrimaryKeyConstraintException ignored) {
```
beeender: Why would this throw? It is clearly valid that `RealmList` can have duplicated elements. I would say if use have two objects with same pk but different other fields, then it is an undefined behaviour and user should not do that.

cmelchior: Because you cannot `insert` two objects with the same PK. In that case you should have called `insertOrUpdate`. We cannot really just say it is undefined behaviour as it would be a serious data corruption if you suddenly had two objects with the same primary key.

cmelchior: Note. It would work, if the list referenced the same Java object, e.g. `new RealmList<>(obj, obj)`. Then we correctly detect that it is the same object, but if you have to distinct java objects we can no longer assume anything

beeender: Got it!

realm/realm-annotations-processor/src/test/resources/io/realm/AllTypesRealmProxy.java
```diff
@@ -668,12 +668,23 @@ public static long insert(Realm realm, some.test.AllTypes object, Map<RealmModel
...
+        } else {
+            rowIndex = Table.nativeFindFirstString(tableNativePtr, pkColumnIndex, primaryKeyValue);
+        }
+        if (rowIndex == TableOrView.NO_MATCH) {
+            rowIndex = Table.nativeAddEmptyRow(tableNativePtr, 1);
+            if (primaryKeyValue != null) {
+                Table.nativeSetString(tableNativePtr, pkColumnIndex, rowIndex, (String)primaryKeyValue);
+            }
+        } else {
+            table.throwDuplicatePrimaryKeyException(primaryKeyValue);
```
beeender: make `throwDuplicatePrimaryKeyException` static and use `Table` here

realm/realm-annotations-processor/src/test/resources/io/realm/AllTypesRealmProxy.java
```diff
@@ -668,12 +668,23 @@ public static long insert(Realm realm, some.test.AllTypes object, Map<RealmModel
...
-        String realmGet$columnString = ((AllTypesRealmProxyInterface)object).realmGet$columnString();
-        if (realmGet$columnString != null) {
-            Table.nativeSetString(tableNativePtr, columnInfo.columnStringIndex, rowIndex, realmGet$columnString);
+        long pkColumnIndex = table.getPrimaryKey();
+        String primaryKeyValue = ((AllTypesRealmProxyInterface) object).realmGet$columnString();
+        long rowIndex = TableOrView.NO_MATCH;
+        if (primaryKeyValue == null) {
+            rowIndex = table.findFirstNull(pkColumnIndex);
+        } else {
+            rowIndex = Table.nativeFindFirstString(tableNativePtr, pkColumnIndex, primaryKeyValue);
```
beeender: It is strange calling native method here but java method above. I think it is better call the java wrapper method only. Easy for maintaining.

beeender: Also, we can make less code by support `null` by `nativeFindFirstString`. But it is beyond the scope of this PR.

realm/realm-annotations-processor/src/test/resources/io/realm/AllTypesRealmProxy.java
```diff
@@ -668,12 +668,23 @@ public static long insert(Realm realm, some.test.AllTypes object, Map<RealmModel
...
+        }
+        if (rowIndex == TableOrView.NO_MATCH) {
+            rowIndex = Table.nativeAddEmptyRow(tableNativePtr, 1);
```
beeender: same, but calling the native method is for the purpose of insterting speed, i am OK with this.

realm/realm-annotations-processor/src/test/resources/io/realm/AllTypesRealmProxy.java
cmelchior: Not sure I follow? `((AllTypesRealmProxyInterface) object).realmGet$columnString();` is required because you need the data from the actual Java object since it is not a proxy. 

All other methods should ideally use the Table interface to avoid unnecessary allocations and checks (which was the entire point of `insert()`

realm/realm-annotations-processor/src/test/resources/io/realm/AllTypesRealmProxy.java
```diff
@@ -716,16 +727,27 @@ public static void insert(Realm realm, Iterator<? extends RealmModel> objects, M
...
-                cache.put(object, rowIndex);
-                String realmGet$columnString = ((AllTypesRealmProxyInterface)object).realmGet$columnString();
-                if (realmGet$columnString != null) {
-                    Table.nativeSetString(tableNativePtr, columnInfo.columnStringIndex, rowIndex, realmGet$columnString);
+                String primaryKeyValue = ((AllTypesRealmProxyInterface) object).realmGet$columnString();
+                long rowIndex = TableOrView.NO_MATCH;
+                if (primaryKeyValue == null) {
+                    rowIndex = table.findFirstNull(pkColumnIndex);
+                } else {
+                    rowIndex = Table.nativeFindFirstString(tableNativePtr, pkColumnIndex, primaryKeyValue);
```
beeender: same here

realm/realm-annotations-processor/src/test/resources/io/realm/AllTypesRealmProxy.java
```diff
@@ -716,16 +727,27 @@ public static void insert(Realm realm, Iterator<? extends RealmModel> objects, M
...
                 }
+                if (rowIndex == TableOrView.NO_MATCH) {
+                    rowIndex = Table.nativeAddEmptyRow(tableNativePtr, 1);
+                    if (primaryKeyValue != null) {
+                        Table.nativeSetString(tableNativePtr, pkColumnIndex, rowIndex, (String)primaryKeyValue);
+                    }
+                } else {
+                    table.throwDuplicatePrimaryKeyException(primaryKeyValue);
```
beeender: Table

realm/realm-annotations-processor/src/test/resources/io/realm/AllTypesRealmProxy.java
```diff
@@ -854,16 +870,10 @@ public static void insertOrUpdate(Realm realm, Iterator<? extends RealmModel> ob
                 if (rowIndex == TableOrView.NO_MATCH) {
                     rowIndex = Table.nativeAddEmptyRow(tableNativePtr, 1);
                     if (primaryKeyValue != null) {
-                        Table.nativeSetString(tableNativePtr, pkColumnIndex, rowIndex, ((AllTypesRealmProxyInterface) object).realmGet$columnString());
+                        Table.nativeSetString(tableNativePtr, pkColumnIndex, rowIndex, (String)primaryKeyValue);
```
beeender: just to confirm, the default value nullable string field is `null` not `""`?

cmelchior: Yup. The default is based on the `@Required` annotation.

I am actually not entirely sure we handle that case properly here: `@Required @PrimaryKey String  id`, should use `'"` as the default value.

realm/realm-annotations-processor/src/test/resources/io/realm/AllTypesRealmProxy.java
beeender: I mean it is calling java method `findFirstNull` at line 675, but native method `nativeFindFirstString` at line 677. which looks a bit strange.

cmelchior: Good point. I'll fix it.

kneth:
@beeender has already done the hard work: LGTM

cmelchior:
Updated @realm/java

beeender:
:+1:

kneth:
:+1:

PR #3241, opened 2016-08-02
Clarify how insertOrUpdate behave for objects with primary keys.

### LINKED ISSUE #3109: Fix insertOrUpdate Javadoc
Feedback: 
> from the javadoc of `insertOrUpdate()` We don't check if the provided objects are already managed or not, so inserting a managed object might duplicate it` (https://github.com/realm/realm-java/blob/master/realm/realm-library/src/main/java/io/realm/Realm.java#L863) this seems confusing in my opinion, it lead me to think that `insertOrUpdate()` might duplicate objects.

It is true that `insertOrUpdate` never duplicate objects with primary keys. We should clarify that.

cmelchior:
Fixes #3109 

@realm/java

### DIFF
--- a/realm/realm-library/src/main/java/io/realm/Realm.java
+++ b/realm/realm-library/src/main/java/io/realm/Realm.java
@@ -780,7 +780,10 @@ public final class Realm extends BaseRealm {
      * <p>
      * Please note:
      * <ul>
-     * <li>We don't check if the provided objects are already managed or not, so inserting a managed object will duplicate it</li>
+     * <li>
+     *     We don't check if the provided objects are already managed or not, so inserting a managed object might duplicate it.
+     *     Duplication will only happen if the object doesn't have a primary key. Objects with primary keys will never get duplicated.
+     * </li>
      * <li>We don't create (nor return) a managed {@link RealmObject} for each element</li>
      * <li>Copying an object will copy all field values. Any unset field in the object and child objects will be set to their default value if not provided</li>
      * </ul>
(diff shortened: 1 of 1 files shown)


beeender:
:+1:

stk1m1:
👍

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -840,7 +840,10 @@ public void insert(RealmModel object) {
      * <p>
      * Please note:
      * <ul>
-     * <li>We don't check if the provided objects are already managed or not, so inserting a managed object might duplicate it</li>
+     * <li>
+     *     We don't check if the provided objects are already managed or not, so inserting a managed object might duplicate it.
+     *     This will only happen if the object doesn't have a primary key. Objects with primary keys will never get duplicated.
```
kneth: "This" -> "Duplication"

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -871,7 +874,10 @@ public void insertOrUpdate(Collection<? extends RealmModel> objects) {
...
```
kneth: "This" -> "Duplication"

kneth:
:+1: with minor comment

realm/realm-library/src/main/java/io/realm/Realm.java
zaki50: Shouldn't we add [at]throws clause in this javadoc comment as well?

cmelchior: Done

PR #3255, opened 2016-08-03
Do not copy strings too many times.

kneth:
`Table::get_name()` returns `StringData`. Converting to `string` and back to `StringData` is really stupid. Let's run `git blame` and see how did that!

@realm/java


beeender:
:grin: :+1:

cmelchior:
👍

PR #3276, opened 2016-08-09
Static native method needs jclass

beeender:


kneth:
Isn't the rule that you need `jclass` when you access a static field, and otherwise you use `jobject` (as `this`)?

beeender:
Yes, rule is there. And also if the header file is generated from java class, this won't build because I added the warning checking.

But we always forget to update the header file when change native method declaration in java :(

We have to merge cmake!

cmelchior:
Ups. I think that one is mine 👍

kneth:
:+1:

Issue #3278, opened 2016-08-10
Add .equalTo(String fieldName, byte[] value) to RealmQuery


bkromhout:
#### Goal

`RealmQuery` is currently missing the ability to query against `byte[]` fields using `equalTo`, and since that *is* one of the supported field types, I believe such a thing should be possible.

#### Expected Results

I would like to be able to construct a `RealmQuery` which returns items based on whether a `byte[]` field's value matches a byte array I supply. For example:

```java
byte[] bytes = new byte[] {...}; // Some byte array.
RealmResults<Foo> foos = realm.where(Foo.class).equalTo("bar", bytes).findAll();
```

If this is something which I can help out with, I'd definitely be willing to (though I might need a few pointers as to where to look).

Thanks!

kneth:
@bkromhout It is true that we don't support `equalTo()` and `notEqualTo()` on binary data (`byte[]`). Implementing it, requires a methods like https://github.com/realm/realm-java/blob/master/realm/realm-jni/src/io_realm_internal_TableQuery.cpp#L832 and https://github.com/realm/realm-java/blob/master/realm/realm-jni/src/io_realm_internal_TableQuery.cpp#L888.

beeender:
#3285 is merged and released.

PR #3285, opened 2016-08-11
Add methods to RealmQuery to support byte arrays

bkromhout:
This is a PR to address issue #3278, which I opened yesterday.

The intent of this PR is to add methods to `RealmQuery` which support `byte[]` fields. The original issue was only for `equalTo()`, but if the core supports others, I'd be happy to add them as well (see my first comment for more info on that)

As far as I understand, these are the steps I need to take:
* Add methods to `RealmQuery.java`
* Add methods and native method signatures to `TableQuery.java`
* Regenerate JNI header files
* Add methods to `io_realm_internal_TableQuery.cpp`
* Add tests to `RealmQueryTests` (In my initial commits, I also added some to `JNIQueryTest`, not sure if that was needed?)


bkromhout:
@kneth: As I mentioned in my email, I'd prefer to keep all discussions in one central location, so let's keep it here :)

As you can see, I've already make a bit of progress here. I'm currently working on adding methods to `io_realm_internal_TableQuery.cpp`, and I had a couple questions.

#### Question 1 
You pointed to a couple places in your comment on #3278 in this file that I should look at. Those ones are the String methods, and it seems that they support not just `equalTo`, but also `notEqualTo`, `contains`, `beginsWith`, and `endsWith`.

As you can see, I've already added stuff for `notEqualTo`, since I have a hunch that the core supported that for `byte[]` if it already has support for `equalTo`. My question is, does the core also support the other 3 (`contains`, `beginsWith`, `endsWith`)? If so, I'd be happy to also add `byte[]` support for those as well 😊.

#### Question 2
While working on the `io_realm_internal_TableQuery.cpp` file, I'm trying to model my implementation off of the one for `String` methods. 

In the `TableQuery_StringPredicate` method, there is a big `if` block that handles things differently depending on whether or not this is a link query (if I understand it correctly). 

In the `else` part of that block, there are lines [such as this one](https://github.com/realm/realm-java/blob/master/realm/realm-jni/src/io_realm_internal_TableQuery.cpp#L869), which is the part I have a question about. So say the `String` version of the line looks like this:
```java
Q(nativeQueryPtr)->and_query(table_ref->column<String>(size_t(arr[arr_len-1])).equal(StringData(value2), is_case_sensitive));
```
For my version, should it look like this?
```java
Q(nativeQueryPtr)->and_query(table_ref->column<Binary>(size_t(arr[arr_len-1])).equal(BinaryData(value));
```
Or is that not *quite* correct?

Thanks!

bkromhout:
Alright, I've now hit a snag when trying to compile (using `./gradlew assmble`). I get this output:
```
...

:installRealm
Incremental java compilation is an incubating feature.
[buildinfo] Not using buildInfo properties file for this build.
:realm:realm-annotations-processor:generateVersionClass
:realm:realm-annotations-processor:compileJava UP-TO-DATE
:realm:realm-annotations-processor:processResources UP-TO-DATE
:realm:realm-annotations-processor:classes UP-TO-DATE
:realm:realm-annotations-processor:jar UP-TO-DATE
:realm:realm-annotations-processor:install
:realm:realm-jni:downloadCore
:realm:realm-jni:deployCore UP-TO-DATE
:realm:realm-jni:generateNdkToolchainArm UP-TO-DATE
:realm:realm-jni:buildAndroidJniArm
arm-linux-androideabi-gcc -Os -DNDEBUG -fPIC -DPIC -pthread -Wall -Wextra -ansi -pedantic -Wno-long-long -DANDROID -Wno-variadic-macros -DREALM_HAVE_CONFIG -DPIC -I/Users/bkromhout/Dev/realm-java/realm/realm-jni/build/core-1.4.2/include -Os -std=c++11 -Wmissing-declarations -Werror -fvisibility=hidden -ffunction-sections -fdata-sections -flto -mthumb -MMD -MP -c io_realm_internal_TableQuery.cpp -o io_realm_internal_TableQuery-arm.pic.o
io_realm_internal_TableQuery.cpp: In function 'void TableQuery_BinaryPredicate(JNIEnv*, jlong, jlongArray, jbyteArray, BinaryPredicate)':
io_realm_internal_TableQuery.cpp:959:96: error: 'class realm::Columns<realm::BinaryData>' has no member named 'equal'
                 Q(nativeQueryPtr)->and_query(table_ref->column<Binary>(size_t(arr[arr_len-1])).equal(value2));
                                                                                                ^
io_realm_internal_TableQuery.cpp:962:96: error: 'class realm::Columns<realm::BinaryData>' has no member named 'not_equal'
                 Q(nativeQueryPtr)->and_query(table_ref->column<Binary>(size_t(arr[arr_len-1])).not_equal(value2));
                                                                                                ^
make: *** [io_realm_internal_TableQuery-arm.pic.o] Error 1
:realm:realm-jni:buildAndroidJniArm FAILED
:installRealm FAILED

* What went wrong:
Execution failed for task ':realm-jni:buildAndroidJniArm'.
> Process 'command 'make'' finished with non-zero exit value 2
```

Clearly something I've done for the latter half of my `TableQuery_BinaryPredicate` method in the `io_realm_internal_TableQuery.cpp` file isn't correct, but I'm not sure what it should look like. I tried to make it similar to what the `String` method in that file looked like.

bdash:
`Columns<BinaryData>` has `operator==` and `operator!=`, rather than `equal` / `not_equal` methods. `Columns<StringData>` uses the named methods as it needs to accommodate a case-sensitivity flag being passed in.

bkromhout:
Thanks @bdash! I've gotten it to build properly and run all tests successfully.

I'm not sure why Jenkins indicates the build failed (though since I can't actually see anything on Jenkins, I can't really comment on why that might be).

beeender:
@bkromhout It is our CI problem, I just retriggerred it

kneth:
@bkromhout Thanks for the great work ❤️ 

`beginsWith()` etc. are nice to have for consistency. If you have time, it will be appreciated if you implement them.

I don't have time to do a proper review today, but from a quick read-through it does look good.

bkromhout:
@kneth, No problem!

I'll definitely implement those three as well then, either tomorrow or the day after that.

Thanks!

kneth:
@bkromhout I added the label "In Progress" which we use on issues we are working on but are not ready to review.

bkromhout:
I've run into another snag. I implemented what I assumed would be the correct way of calling the core methods in CPP for `contains`, `begins_with`, and `ends_with`, but it's not building and gives me these errors:
```
...

io_realm_internal_TableQuery.cpp: In function 'void TableQuery_BinaryPredicate(JNIEnv*, jlong, jlongArray, jbyteArray, BinaryPredicate)':
io_realm_internal_TableQuery.cpp:977:96: error: 'class realm::Columns<realm::BinaryData>' has no member named 'contains'
                 Q(nativeQueryPtr)->and_query(table_ref->column<Binary>(size_t(arr[arr_len-1])).contains(value2));
                                                                                                ^
io_realm_internal_TableQuery.cpp:980:96: error: 'class realm::Columns<realm::BinaryData>' has no member named 'begins_with'
                 Q(nativeQueryPtr)->and_query(table_ref->column<Binary>(size_t(arr[arr_len-1])).begins_with(value2));
                                                                                                ^
io_realm_internal_TableQuery.cpp:983:96: error: 'class realm::Columns<realm::BinaryData>' has no member named 'ends_with'
                 Q(nativeQueryPtr)->and_query(table_ref->column<Binary>(size_t(arr[arr_len-1])).ends_with(value2));
                                                                                                ^
make: *** [io_realm_internal_TableQuery-arm.pic.o] Error 1
make: *** Waiting for unfinished jobs....
:realm-jni:buildAndroidJniArm FAILED
```

@bdash, since you were able to point me in the correct direction with this before, do you have any advice? I understand the use of `==` and `!=` for the previous two, but there aren't really equivalent operators for these three, so I'm a bit baffled here. Am I missing something simple here?

bdash:
The query expression API (e.g., `Columns<Binary>` and such) doesn't support `contains` / `begins_with` / `ends_with` on binary columns. Those operations are only supported on `Query` directly (`Query::contains` and similar), which limits them to comparisons that don't traverse relationships. The lack of equivalent functions in the query expression API appears to be an oversight rather than an intentional choice. I see two options for dealing with this:

1. Only support `contains` / `begins_with` / `ends_with` when no relationships are involved.
2. Don't add support for these operators right now.

In either case the operators could be provided with full generality in the future once core's query expression API adds support for them.

bkromhout:
@bdash Interesting...I assume that `String`'s use of `contains` / `begins_with` / `ends_with` when traversing relationships must be a special case then.

As far as the options you mentioned, I personally would go with option 1. We could document that these three methods cannot be used when traversing relationships, then have the CPP file throw up an exception if someone tries to do so.  
I think it's better to add the functionality for non-relationship-traversing cases than to exclude it altogether. Plus, it would (hopefully) need minimal changes once the core adds support for this (only changing a few lines in the `TableQuery` CPP rather than in that + the java files).

However, I don't think I'm allowed to make the final call, so I'll defer to what @realm/java thinks is best here 😉

beeender:
CI error 
```
io_realm_internal_TableQuery.cpp: In function 'void TableQuery_BinaryPredicate(JNIEnv*, jlong, jlongArray, jbyteArray, BinaryPredicate)':
io_realm_internal_TableQuery.cpp:981:96: error: 'class realm::Columns<realm::BinaryData>' has no member named 'contains'
                 Q(nativeQueryPtr)->and_query(table_ref->column<Binary>(size_t(arr[arr_len-1])).contains(value2));
                                                                                                ^
io_realm_internal_TableQuery.cpp:984:96: error: 'class realm::Columns<realm::BinaryData>' has no member named 'begins_with'
                 Q(nativeQueryPtr)->and_query(table_ref->column<Binary>(size_t(arr[arr_len-1])).begins_with(value2));
                                                                                                ^
io_realm_internal_TableQuery.cpp:987:96: error: 'class realm::Columns<realm::BinaryData>' has no member named 'ends_with'
                 Q(nativeQueryPtr)->and_query(table_ref->column<Binary>(size_t(arr[arr_len-1])).ends_with(value2));
                                                                                                ^
```

bkromhout:
@beeender, yep, we were discussing that in the few comments above yours. What do you think about the options @bdash proposed?

beeender:
hmmm, got it. Let me check if those methods can be easily added in core.

kneth:
If `arr_len > 1` in `TableQuery_BinaryPredicate()` we might have to throw an exception saying it's not supported.

bmunkholm:
I think the original reason for not having contains, begins_with and ends_with was that we thought that they likely would be rarely needed, and then didn't want to pollute the API. So I suggest to finish this addition which in itself seems both needed and supported. Then if the other methods are needed we can add those later in another PR.

bkromhout:
@kneth Done, though if you'd prefer one generic exception message rather than the 5 more specific ones I included then I can do that instead.

@bmunkholm Makes sense. I'd be happy to add back support for queries which work across relationships at a later date once the core supports it.

bkromhout:
What was the CI error this time, every builds and all tests pass for me locally 😕

beeender:
Seems to be a disk full issue on CI machine. Retriggered.

bkromhout:
Since the CI is now building successfully, I guess this is ready for final review. What say you @realm/java?

realm/realm-jni/src/io_realm_internal_TableQuery.cpp
```diff
@@ -897,24 +897,139 @@ JNIEXPORT void JNICALL Java_io_realm_internal_TableQuery_nativeNotEqual__J_3JLja
     TableQuery_StringPredicate(env, nativeQueryPtr, columnIndexes, value, caseSensitive, StringNotEqual);
 }
 
-JNIEXPORT void JNICALL Java_io_realm_internal_TableQuery_nativeBeginsWith(
+JNIEXPORT void JNICALL Java_io_realm_internal_TableQuery_nativeBeginsWith__J_3JLjava_lang_String_2Z(
```
beeender: Which `javah` options creates this kind of style method name? `-jni`?

bkromhout: I just used the [regenerate-jni-headers.sh](https://github.com/realm/realm-java/blob/master/realm/realm-jni/generate-jni-headers.sh) script which is at `realm/realm-jni/`, but yes, it looks like that's the case.

For what it's worth, those generated headers (for the `String` methods) didn't change until I added the equivalent ones for `byte[]`. I assume that at that point javah had to output more explicit header names to help distinguish them for some reason. It seems that the `String` headers for `nativeEqualTo` and `nativeNotEqualTo` already had that extra suffix, so I assume this is a standard thing for javah output.

bkromhout: Also, I should note that according to [Oracle's `javah` documentation](http://docs.oracle.com/javase/8/docs/technotes/tools/windows/javah.html) the `-jni` option is the default behavior, so actually having it explicitly specified is a no-op anyway

realm/realm-library/src/androidTest/java/io/realm/RealmQueryTests.java
```diff
@@ -1618,6 +1630,95 @@ public void buildQueryFromResultsGC() {
         }
     }
 
+    private static byte[][] binaries = {{1, 2, 3}, {1, 2}, {1, 2, 3}, {2, 3}, {2}, {4, 5, 6}};
+
+    private void createBinaryOnlyDataSet() {
+        realm.beginTransaction();
+        realm.delete(BinaryOnly.class);
```
beeender: This is not needed. `TestRealmConfigurationFactory` will ensure every test starts with a empty Realm.

realm/realm-jni/src/io_realm_internal_TableQuery.cpp
beeender: hmmmm, interesting ...  from jni.h
```
class _jstring : public _jobject {};
class _jarray : public _jobject {};
class _jbyteArray : public _jarray {};
```

The `jstring` and `jbytearray` both have the same type `jobject`. So the method names have to be different.

bkromhout: Good ol' JVM 😅

beeender:
Needs tests about `contains`,`beginWith`,`endWith` with given `(byte[])null` for both nullable/non-nullable `byte[]` field.

Others LGTM!

realm/realm-jni/src/io_realm_internal_TableQuery.cpp
kneth: I think that `javah` does this to avoid overloading. It is basically C code, and overloading was introduced in C11. `javah` is much older than that. I have noticed a similar issue/annoyance in Android Studio 2.2 beta.

bkromhout:
Line removed and tests added, ready for next round of reviews @realm/java! 😄

bkromhout:
@kneth How could I forget haha, all done!

realm/realm-library/src/androidTest/java/io/realm/RealmQueryTests.java
```diff
@@ -1354,7 +1361,11 @@ public void equalTo_nullableFields() {
         assertEquals(1, realm.where(NullTypes.class).equalTo(NullTypes.FIELD_STRING_NULL, (String) null).findAll().size());
         assertEquals(1, realm.where(NullTypes.class).equalTo(NullTypes.FIELD_STRING_NULL, "Fish").findAll().size());
         assertEquals(0, realm.where(NullTypes.class).equalTo(NullTypes.FIELD_STRING_NULL, "Goat").findAll().size());
-        // 2 Bytes skipped, doesn't support equalTo query
+        // 2 Bytes
+        assertEquals(1, realm.where(NullTypes.class).equalTo(NullTypes.FIELD_BYTES_NULL, new byte[] {0}).findAll().size());
```
cmelchior: minor detail, but `findAll().size()` can be replaced by `count()` which is also faster :)

realm/realm-library/src/androidTest/java/io/realm/entities/BinaryOnly.java
```diff
@@ -0,0 +1,31 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.realm.entities;
+
+import io.realm.RealmObject;
+
+public class BinaryOnly extends RealmObject {
```
cmelchior: Any reason for creating a new class instead of reusing one of our standard ones (AllTypes/AllJavaTypes) ?

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -291,6 +291,25 @@ public boolean isValid() {
      * @param fieldName the field to compare.
      * @param value the value to compare with.
      * @return the query object.
+     * @throws java.lang.IllegalArgumentException if one or more arguments do not match class or field type, or if
+     * {@code fieldName} points to a linked field.
+     */
+    public RealmQuery<E> equalTo(String fieldName, byte[] value) {
```
cmelchior: Seems to be missing unit tests for the field name being invalid or not pointing to a byte field?

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -648,6 +667,25 @@ public boolean isValid() {
      * @param fieldName the field to compare.
      * @param value the value to compare with.
      * @return the query object.
+     * @throws java.lang.IllegalArgumentException if one or more arguments do not match class or field type, or if
+     * {@code fieldName} points to a linked field.
+     */
+    public RealmQuery<E> notEqualTo(String fieldName, byte[] value) {
```
cmelchior: Same, missing unit test?

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -1160,6 +1198,21 @@ public boolean isValid() {
...
     /**
+     * Condition that value of field contains the specified byte array.
+     *
+     * @param fieldName the field to compare.
+     * @param value the byte array.
+     * @return the query object.
+     * @throws java.lang.IllegalArgumentException if one or more arguments do not match class or field type, or if
+     * {@code fieldName} points to a linked field.
+     */
+    public RealmQuery<E> contains(String fieldName, byte[] value) {
```
cmelchior: Same, missing unit test for invalid field names

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -1187,6 +1240,21 @@ public boolean isValid() {
...
     /**
+     * Condition that the value of field begins with the specified byte array.
+     *
+     * @param fieldName the field to compare.
+     * @param value the byte array.
+     * @return the query object.
+     * @throws java.lang.IllegalArgumentException if one or more arguments do not match class or field type, or if
+     * {@code fieldName} points to a linked field.
+     */
+    public RealmQuery<E> beginsWith(String fieldName, byte[] value) {
```

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -1213,6 +1281,21 @@ public boolean isValid() {
...
+    /**
+     * Condition that the value of field ends with the specified byte array.
+     *
+     * @param fieldName the field to compare.
+     * @param value the byte array.
+     * @return the query object.
+     * @throws java.lang.IllegalArgumentException if one or more arguments do not match class or field type, or if
+     * {@code fieldName} points to a linked field.
+     */
+    public RealmQuery<E> endsWith(String fieldName, byte[] value) {
```

cmelchior:
Nice work 🎉 There seems to be missing some unit test that checks that the proper exceptions are thrown for invalid field names being used, but otherwise it looks good.

bkromhout:
@cmelchior I'll be sure to address your comments; progress might be a bit slower since University started again today, but hopefully I should have it done quickly :)

bmunkholm:
@realm/java: We should DEFINITELY not add new methods that throws "not implemented" at runtime. And worse in the changelog mention they are implemented. That's not your fault @bkromhout, but a major oversight by the reviewers.

bkromhout:
@bmunkholm, what would you suggest for those three methods then? Should this PR be narrowed in scope to just the `equalTo`, `notEqualTo` methods for byte arrays,  and the other three can be done in another PR once the core has support for doing them across relationships?

bmunkholm:
Yes - that was my original suggestion.
As it is now people look in the API and use the unfinished methods only to discover that they don't work at runtime. That's a really bad user experience. And I'm not even convinced those methods are used. So would rather add them later if people has a need for them.

realm/realm-jni/src/io_realm_internal_TableQuery.cpp
```diff
@@ -897,24 +897,139 @@ JNIEXPORT void JNICALL Java_io_realm_internal_TableQuery_nativeNotEqual__J_3JLja
...
+        }
+        else {
+            jbyte* bytePtr = env->GetByteArrayElements(value, NULL);
+            if (!bytePtr) {
+                ThrowException(env, IllegalArgument, "binaryPredicate");
+                return;
+            }
+            size_t dataLen = S(env->GetArrayLength(value));
+            value2 = BinaryData(reinterpret_cast<char*>(bytePtr), dataLen);
+            env->ReleaseByteArrayElements(value, bytePtr, 0);
```
zaki50: Is it OK to release `bytePtr` before using `value2`?

`BinaryData` class seems to simply hold passed pointer.

bkromhout:
@bmunkholm, Gotcha. When I get the chance I will remove those again so that this PR only adds `equalTo` and `notEqualTo`.

realm/realm-jni/src/io_realm_internal_TableQuery.cpp
zaki50: About the last argument of `ReleaseByteArrayElements`, I think `JNI_ABORT` is better since we don't modify the contents.

bkromhout: I believe so; I modeled my use of that off of what you guys had for the `tbl_nativeDoByteArray` method in [`tablebase_tpl.hpp`](https://github.com/realm/realm-java/blob/master/realm/realm-jni/src/tablebase_tpl.hpp), however, I admit that it might not be okay to do?
That being said, I would expect that the unit tests would fail hard if it was releasing it too early.

As far as JNI_ABORT, I can do that for sure.

realm/realm-library/src/androidTest/java/io/realm/RealmQueryTests.java
bkromhout: Would you like to me change all of the existing ones? Or just mine?

realm/realm-library/src/androidTest/java/io/realm/entities/BinaryOnly.java
bkromhout: You know... it seemed like I good idea at the time 😅; I'll get rid of it and use AllJavaTypes instead

realm/realm-library/src/main/java/io/realm/RealmQuery.java
bkromhout: Removing this method per later discussion.

realm/realm-library/src/main/java/io/realm/RealmQuery.java
bkromhout: Same

realm/realm-library/src/main/java/io/realm/RealmQuery.java
bkromhout: Same

bkromhout:
CI Fluke or actual problem this time? (Also, side note, is there some way for me to be able to view the CI build job for this PR so that I don't have to bother everyone by asking each time? 😄)

realm/realm-jni/src/io_realm_internal_TableQuery.cpp
zaki50: In `tbl_nativeDoByteArray`, `bytePtr` is not used after invoking `ReleaseByteArrayElements`.
`(pTable->*doBinary)( S(columnIndex), S(rowIndex), realm::BinaryData(reinterpret_cast<char*>(bytePtr), dataLen));` copies bytes to its own array and does not access the pointer from `BinaryData` after returning the function (the code is in the core  and not yet public though).

The official document of `ReleaseBooleanArrayElements` clearly says that this function frees `bytePtr` if the `mode` is `0` or `JNI_ABORT`.
http://docs.oracle.com/javase/1.5.0/docs/guide/jni/spec/functions.html#wp1544

I think `ReleaseByteArrayElements` must be at the end of this function.


And I confirmed that `Q(nativeQueryPtr)->equal(S(arr[0]), value2);` and `Q(nativeQueryPtr)->not_equal(S(arr[0]), value2);` in `TableQuery_BinaryPredicate` copy byte array in `value2`, 
but I'm not sure about `table_ref->column<Binary>(size_t(arr[arr_len-1])) == value2` and `table_ref->column<Binary>(size_t(arr[arr_len-1])) != value2`. What do you think @kneth ?

zaki50:
It's a issue of CI server. I'm asking to fix it.

beeender:
:+1:

realm/realm-library/src/androidTest/java/io/realm/RealmQueryTests.java
cmelchior: Primarily just yours, but while you're at might as well fix the rest of them 😄

cmelchior:
Looks like only https://github.com/realm/realm-java/pull/3285#discussion_r75214116 and https://github.com/realm/realm-java/pull/3285#discussion_r75286756 needs to be addressed. Then I think this looks ready to merge: 👏

realm/realm-library/src/androidTest/java/io/realm/RealmQueryTests.java
bkromhout: Haha it should be a simple find and replace; will do!

realm/realm-jni/src/io_realm_internal_TableQuery.cpp
bkromhout: Regardless of what @kneth's reply is, it certainly shouldn't *hurt* it to put that line at the end of the method if I'm understanding it right, so I'll just do it 😉

cmelchior:
Hi @bkromhout 
We would like to make a release today, so I am merging this now and will fix the last two remaining issues in a separate PR. Thanks a lot for all your hard work 🎉

bkromhout:
@cmelchoir, sounds good! Sorry I wasn't able to get to the last two parts!

PR #3295, opened 2016-08-16
getFieldIndex returns null for non-existing field

### LINKED ISSUE #3292: FATAL columnIndex is less than 0
#### Goal

Query on brand new DynamicRealm

#### Expected Results

Get an error similar to `Table "Thing" does not exist in schema`

#### Actual Results

Fatal crash `java.lang.ArrayIndexOutOfBoundsException: columnIndex is less than 0.`

Create a new DynamicRealm
Run something like: `dynamicRealm.where("Thing").equalTo("id", id).findFirst();`

Realm version(s): 1.1.1

Android Studio version: 2.2 Beta

Which Android version and device: Android 5.1 on Huawei Mate S

beeender:
Close #3292

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 * `insert()` now correctly throws an exception if two different objects have the same primary key (#3212).
 * Blackberry Z10 throwing "Function not implemented" (#3178).
 * Reduced the number of file descriptors used by Realm Core (#3197).
+* Throws a proper exception when operate on a non-existing field with dynamic API (#3292).
 
 ### Internal
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 1.1.2
+
+### Bug fixes
+
+* Throw a proper exception when operating on a non-existing field with the dynamic API (#3292).
+
 ## 1.1.1
 
 ### Enhancements
@@ -16,7 +22,6 @@
 * `insert()` now correctly throws an exception if two different objects have the same primary key (#3212).
 * Blackberry Z10 throwing "Function not implemented" (#3178).
 * Reduced the number of file descriptors used by Realm Core (#3197).
-* Throw a proper exception when operating on a non-existing field with the dynamic API (#3292).
 
 ### Internal
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@
 * `insert()` now correctly throws an exception if two different objects have the same primary key (#3212).
 * Blackberry Z10 throwing "Function not implemented" (#3178).
 * Reduced the number of file descriptors used by Realm Core (#3197).
-* Throws a proper exception when operate on a non-existing field with dynamic API (#3292).
+* Throw a proper exception when operating on a non-existing field with the dynamic API (#3292).
 
 ### Internal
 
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -16,6 +16,7 @@
 * `insert()` now correctly throws an exception if two different objects have the same primary key (#3212).
 * Blackberry Z10 throwing "Function not implemented" (#3178).
 * Reduced the number of file descriptors used by Realm Core (#3197).
+* Throws a proper exception when operate on a non-existing field with dynamic API (#3292).
```
cmelchior: Throw a proper exception when operating on a non-existing field with the dynamic API

realm/realm-library/src/androidTest/java/io/realm/RealmObjectSchemaTests.java
```diff
@@ -730,6 +731,28 @@ public void getPrimaryKey_nonExistFieldThrows() {
         schema.getPrimaryKey();
     }
 
+    @Test
+    public void getFieldIndex() {
```
cmelchior: This unit test targets an internal method. It would be more future proof if the unit test tested the expected behavior from the public API instead.

E.g, I think this PR will instead of "Class `Thing` does not exists in the Schema" say "Field `NoField` does not exists"?

realm/realm-library/src/main/java/io/realm/RealmObjectSchema.java
```diff
@@ -633,15 +633,16 @@ private void checkEmpty(String str) {
             }
             return columnIndices;
         } else {
-            if (getFieldIndex(fieldDescription) == null) {
+            Long fieldIndex = getFieldIndex(fieldDescription);
```
cmelchior: Wow, did we really call this 3 times 👍

cmelchior:
This change doesn't exactly cover #3292 from what I can tell? The internal changes looks good, but I am missing a test case for the actual issue reported. IMO having an error message saying that "Class "X" doesn't exist" would be superior.

realm/realm-library/src/androidTest/java/io/realm/DynamicRealmTests.java
```diff
@@ -664,4 +668,18 @@ public void run() throws Exception {
             }
         });
     }
+
+    @Test
+    public void equalTo_noFieldObjectShouldThrow() {
```
cmelchior: This still tests a non-existing field on an existing Table, but the issue reported was about an (obvious) non-existing field on a non-existing table.

beeender: No, the issue was about a non-existing field.
For a non-existing table, our exception message is correct, i did check it.

cmelchior: My brain short circuited. Just ignore the comment 😄

realm/realm-library/src/main/java/io/realm/RealmObjectSchema.java
```diff
@@ -633,15 +633,16 @@ private void checkEmpty(String str) {
...
+            if (fieldIndex == null) {
                 throw new IllegalArgumentException(String.format("Field '%s' does not exist.", fieldDescription));
```
kneth: Message isn't consistent with message in line 629. Should it be?

cmelchior:
Nice 👍

realm/realm-library/src/main/java/io/realm/RealmObjectSchema.java
beeender: uh... we might need to create a task to clean up exception messages ... OK to make it like this in this PR?

kneth: Yes, let's create a "exception message cleanup" issue.

PR #3301, opened 2016-08-16
Fix notification tests for following changes

beeender:
To support multiprocess & OS notifications, a dedicated thread will
monitor the Realm changes and send notifications to other thread. This
means our tests cannot assume notifications will be sent for every
transaction commit.

Also, the Realm.handler will moved to a lower layer, it's better to not
rely on it in the test as well.

To avoid a massive PR, tests changes are made separately first.


realm/realm-library/src/androidTest/java/io/realm/NotificationsTest.java
```diff
@@ -349,22 +349,23 @@ public Boolean call() throws Exception {
...
                 // at the end so we can evaluate results faster.
-                threadHandler.post(new Runnable() {
+                // 500 ms delay is to make sure the notification daemon thread gets time to send notification.
+                threadHandler.postDelayed(new Runnable() {
                     @Override
                     public void run() {
                         TestHelper.quitLooperOrFail();
                     }
-                });
+                }, 500);
```
cmelchior: Why the delay?

cmelchior: A bit hacky, but then this unit test was hacky to begin with 😄

cmelchior:
👍

nhachicha:
👍

PR #2727, opened 2016-05-03
Remove optional APIs if dependencies don't meet

beeender:
* Add @OptionalAPI to mark a method as optional with dependencies.
* New bytecode transformer to remove an optional API if it doesn't
  have enough dependencies.
* Modify gradle plugin.
* Mark the rx.Observable related APIs as optional.
* Add integration-tests module.

This is not the perfect solution since the RxObservableFactory is still
in the final result. But it is good enough for us to work around the
issue with Jackson and other lib which needs to do reflection on the
RealmObject.

To replace #2400 , to `releases` branch.

PR #3308, opened 2016-08-17
Revert "Remove optional APIs if dependencies don't meet (#2727)"

### LINKED ISSUE #3307: 'com.intellij:annotations' lib included as generated code, causing DEX conflicts
#### Goal

Build an application that depends on Realm and an `aar` dependency that includes `com.intellij:annotations` as a dependency.

#### Expected Results

The application builds successfully.

#### Actual Results

When assembling the application, I get this error.

```

* What went wrong:
Execution failed for task ':apps:questionnaire:app:transformClassesWithDexForDebug'.
> com.android.build.api.transform.TransformException: com.android.ide.common.process.ProcessException: java.util.concurrent.ExecutionException: com.android.dex.DexException: Multiple dex files define Lorg/jetbrains/annotations/NotNull;
```

App build file contains the following:

```
buildscript {
    repositories {
        classpath 'io.realm:realm-gradle-plugin:1.1.0'
    }
}

dependencies {
    compile project(':path:to:library:project')
}
```

Library build file contains:

```
dependencies {
    compile 'com.intellij:annotations:12.0'
}
```

#### Workaround

Since Realm generates the annotations library into the `build/intermediates/transforms/RealmOptionalAPITransformer/debug/folders/1/10/realm-optional-api/org/jetbrains`, there is no way to exclude it from the classpath using Gradle's typical `exclude group: 'com.intellij', module: 'annotations'` method, so in order to fix it I've had to do:

```
configurations {
    compile.exclude group: 'com.intellij', module: 'annotations'
}
```

This solution works, but is not ideal. If the annotations module gets an update that dependent libraries (or even my app) require, I won't be able to get it until I update to a version of realm that supplies the updated annotations library (if one even exists). I recommend adding `compile 'com.intellij:annotations:12.0` to the project through the `realm-gradle-plugin`, instead of generating a copy of the intellij library in `build/intermediates/`.

Realm version(s):

realm-gradle-plugin 1.1.0

beeender:
This reverts commit 6d7e9b0bf108efc3e95c8d23b2676c9aba1dc78b.

OptionalAPI is not used, so just remove it to make a smaller code base.
Also to close #3307.


CHANGELOG.md
```diff
@@ -68,10 +72,6 @@
 * Fixed a bug that allowed both `RealmConfiguration.Builder.assetFile()`/`deleteRealmIfMigrationNeeded()` to be configured at the same time, which leads to the asset file accidentally being deleted in migrations (#2933).
 * Realm crashed outright when the same Realm file was opened in two processes. Realm will now optimistically retry opening for 1 second before throwing an Error (#2459).
 
-### Enhancements
```
cmelchior: The changelog shouldn't be reverted IMO

README.md
```diff
@@ -137,10 +137,6 @@ so each sub-project must be opened in its own window.
 
 See [CONTRIBUTING.md](CONTRIBUTING.md) for more details!
 
-This project adheres to the [Contributor Covenant Code of Conduct](https://realm.io/conduct).
```
cmelchior: This shouldn't be removed either

build.gradle
```diff
@@ -86,19 +86,10 @@ task check {
     dependsOn checkExamples
 }
 
-task integrationTestsConnectedCheck(type:GradleBuild) {
```
cmelchior: This also looks problematic?

CHANGELOG.md
beeender: Oooops .... I am trying to cover my sins

build.gradle
beeender: Yeah, i think it is good to still have this since lots of stuff can only be tested through this way. But on current source tree, this is only used for optional API transformer.

realm/realm-library/src/main/java/io/realm/internal/TableView.java
```diff
@@ -676,7 +676,8 @@ public String toJson() {
     @Override
     public String toString() {
         long columnCount = getColumnCount();
-        StringBuilder stringBuilder = new StringBuilder("The TableView contains ");
+        StringBuilder stringBuilder = new StringBuilder("The TableView ");
+        stringBuilder.append("contains ");
```
zaki50: I think we don't need to change this.

zaki50:
:+1: with minor comment

build.gradle
nhachicha: Ok we can revert this task back when needed

nhachicha:
👍

PR #3311, opened 2016-08-17
DynamicRealmObject typed getters no longer seg faults

### LINKED ISSUE #3294: A/libc: Fatal signal 11 (SIGSEGV), code 1, fault addr 0x69 in tid 3715
#### Goal

Clearer error message.

I had accidentally auto-completed a `getString` as a `getShort` somewhere in my codebase.  Took a bit of time to track it down due to mystical error message.

This is using a `DynamicRealm`.

#### Expected Results

Error that says something like `Type error: cannot convert string field to short`

#### Actual Results

Get this single line error:
```
A/libc: Fatal signal 11 (SIGSEGV), code 1, fault addr 0x69 in tid 3715
```

Call `getShort("field")` on a field defined as `schema.addField("field", String.class)`.

#### Code Sample

```java

```

Realm version(s): **1.1.1**

Android Studio version: **2.2 Beta**

Which Android version and device: **Android 5.1** on a Huawei Mate S

cmelchior:
Closes #3294

DynamicRealmObject typed getters didn't check if the field name being retrieved was the correct type, which caused a seg fault if you called dynamicObj.getString(integerField). A proper IllegalArgumentException is now being throw instead.

Review completed in https://github.com/realm/realm-java/pull/3285

### DIFF
--- a/realm/realm-library/src/main/java/io/realm/DynamicRealmObject.java
+++ b/realm/realm-library/src/main/java/io/realm/DynamicRealmObject.java
@@ -115,7 +115,7 @@ public final class DynamicRealmObject extends RealmObject implements RealmObject
         long columnIndex = proxyState.getRow$realm().getColumnIndex(fieldName);
         RealmFieldType columnType = proxyState.getRow$realm().getColumnType(columnIndex);
         if (columnType != RealmFieldType.BOOLEAN) {
-            throw new IllegalArgumentException(fieldName + " is not a boolean, but a " + columnType);
+            throw new IllegalArgumentException("'" + fieldName + "' is not a '" + RealmFieldType.BOOLEAN + "', but a " + columnType);
         }
         return proxyState.getRow$realm().getBoolean(columnIndex);
     }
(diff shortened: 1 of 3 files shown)


realm/realm-library/src/main/java/io/realm/DynamicRealmObject.java
```diff
@@ -650,6 +658,18 @@ public RealmFieldType getFieldType(String fieldName) {
...
 
+    private void checkFieldType(String fieldName, long columnIndex, RealmFieldType expectedType) {
+        RealmFieldType columnType = proxyState.getRow$realm().getColumnType(columnIndex);
+        if (columnType != expectedType) {
+            String indefiniteVowel = "";
+            if (expectedType == RealmFieldType.INTEGER || expectedType == RealmFieldType.LIST) {
+                indefiniteVowel = "n";
+            }
+            throw new IllegalArgumentException(String.format("'%s' is not a%s '%s', but a%s '%s'.",
+                    fieldName, indefiniteVowel, expectedType, indefiniteVowel, columnType));
```
zaki50: we should have another vowel check for `columnType`.

realm/realm-library/src/main/java/io/realm/DynamicRealmObject.java
```diff
@@ -650,6 +658,18 @@ public RealmFieldType getFieldType(String fieldName) {
         return proxyState.getRow$realm().getColumnType(columnIndex);
     }
 
+    private void checkFieldType(String fieldName, long columnIndex, RealmFieldType expectedType) {
+        RealmFieldType columnType = proxyState.getRow$realm().getColumnType(columnIndex);
+        if (columnType != expectedType) {
+            String indefiniteVowel = "";
+            if (expectedType == RealmFieldType.INTEGER || expectedType == RealmFieldType.LIST) {
```
zaki50: `expectedType == RealmFieldType.LIST` should be `expectedType == RealmFieldType.OBJECT`?

PR #3296, opened 2016-08-16
DynamicRealmObject typed getters no longer seg faults

cmelchior:
Closes #3294 

DynamicRealmObject typed getters didn't check if the field name being retrieved was the correct type, which caused a seg fault if you called `dynamicObj.getString(integerField)`. A proper `IllegalArgumentException` is now being throw instead.

@realm/java

PR #3312, opened 2016-08-17
DynamicRealmObject typed getters

### LINKED ISSUE #3294: A/libc: Fatal signal 11 (SIGSEGV), code 1, fault addr 0x69 in tid 3715

cmelchior:
Closes #3294

DynamicRealmObject typed getters didn't check if the field name being retrieved was the correct type, which caused a seg fault if you called dynamicObj.getString(integerField). A proper IllegalArgumentException is now being throw instead.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * Throw a proper exception when operating on a non-existing field with the dynamic API (#3292).
 * `DynamicRealmObject.setList` should only accept `RealmList<DynamicRealmObject>` (#3280).
+* `DynamicRealmObject.getX(fieldName)` now throws a proper exception instead of a native crash when called with a field name of the wrong type (#3294).
 
 ### Internal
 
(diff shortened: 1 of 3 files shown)


cmelchior:
Review in https://github.com/realm/realm-java/pull/3296

PR #3313, opened 2016-08-17
Verify SharedGroup in handleAsyncTransactionCompleted

beeender:
In a rare case, if handleAsyncTransactionCompleted posted before set
handler to null, the handleAsyncTransactionCompleted will be called with
a closed Realm.

It is very difficult to construct a test case for this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * Throw a proper exception when operating on a non-existing field with the dynamic API (#3292).
 * `DynamicRealmObject.setList` should only accept `RealmList<DynamicRealmObject>` (#3280).
+* Fixed a concurrency crash which might happen when `Realm.executeTransactionAsync()` called with a `OnSuccess` and Realm closed before `OnSuccess` called.
 
 ### Internal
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 * Throw a proper exception when operating on a non-existing field with the dynamic API (#3292).
 * `DynamicRealmObject.setList` should only accept `RealmList<DynamicRealmObject>` (#3280).
+* Fixed a concurrency crash which might happen when `Realm.executeTransactionAsync()` called with a `OnSuccess` and Realm closed before `OnSuccess` called.
```
cmelchior: Fixed a concurrency crash which might happen when `Realm.executeTransactionAsync()` tried to cal `onSucess` after the Realm was closed.

realm/realm-library/src/main/java/io/realm/HandlerController.java
```diff
@@ -156,10 +156,13 @@ public boolean handleMessage(Message message) {
      * @param onSuccess onSuccess callback to run for the async transaction that completed.
      */
     public void handleAsyncTransactionCompleted(Runnable onSuccess) {
-        if (onSuccess != null) {
-            pendingOnSuccessAsyncTransactionCallbacks.add(onSuccess);
+        // Same reason with handleMessage
```
cmelchior: Same reason as handeMessage()

cmelchior:
Nice catch 👍

nhachicha:
👍

PR #3315, opened 2016-08-17
improve `Table` and schema cache.

zaki50:
This PR solves following problems.

* There are two `Class to Table` caches. One is in `Realm` class and another is in `RealmSchema` class.`
* The cache never hits against proxy classes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,10 @@
 * Throw a proper exception when operating on a non-existing field with the dynamic API (#3292).
 * `DynamicRealmObject.setList` should only accept `RealmList<DynamicRealmObject>` (#3280).
 * `DynamicRealmObject.getX(fieldName)` now throws a proper exception instead of a native crash when called with a field name of the wrong type (#3294).
-* Fixed a problem that `Table` and schema cache never hits against proxy classes (#3315).
+
+### Enhancements
+
+* Optimized internal caching of schema classes (#3315).
 
 ### Internal
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 * Throw a proper exception when operating on a non-existing field with the dynamic API (#3292).
 * `DynamicRealmObject.setList` should only accept `RealmList<DynamicRealmObject>` (#3280).
 * `DynamicRealmObject.getX(fieldName)` now throws a proper exception instead of a native crash when called with a field name of the wrong type (#3294).
+* Fixed a problem that `Table` and schema cache never hits against proxy classes (#3315).
 
 ### Internal
 
--- a/realm/realm-library/src/main/java/io/realm/Realm.java
+++ b/realm/realm-library/src/main/java/io/realm/Realm.java
@@ -51,7 +51,6 @@ import io.realm.internal.ColumnInfo;
 import io.realm.internal.RealmObjectProxy;
 import io.realm.internal.RealmProxyMediator;
 import io.realm.internal.Table;
-import io.realm.internal.Util;
 import io.realm.internal.log.RealmLog;
 import rx.Observable;
 
@@ -121,10 +120,6 @@ import rx.Observable;
 public final class Realm extends BaseRealm {
 
     public static final String DEFAULT_REALM_NAME = RealmConfiguration.DEFAULT_REALM_NAME;
-
-    // Caches Class objects (both model classes and proxy classes) to Realm Tables
-    private final Map<Class<? extends RealmModel>, Table> classToTable =
-            new HashMap<Class<? extends RealmModel>, Table>();
 
     private static RealmConfiguration defaultConfiguration;
 
@@ -604,7 +599,7 @@ public final class Realm extends BaseRealm {
             return null;
         }
         E realmObject;
-        Table table = getTable(clazz);
+        Table table = schema.getTable(clazz);
         if (table.hasPrimaryKey()) {
             // As we need the primary key value we have to first parse the entire input stream as in the general
             // case that value might be the last property :(
@@ -683,7 +678,7 @@ public final class Realm extends BaseRealm {
      */
     public <E extends RealmModel> E createObject(Class<E> clazz) {
         checkIfValid();
-        Table table = getTable(clazz);
+        Table table = schema.getTable(clazz);
         long rowIndex = table.addEmptyRow();
         return get(clazz, rowIndex);
     }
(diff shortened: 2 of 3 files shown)


realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -1404,13 +1399,7 @@ static String getCanonicalPath(File realmFile) {
     }
 
     Table getTable(Class<? extends RealmModel> clazz) {
```
cmelchior: Is this method even used anymore?

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 * Throw a proper exception when operating on a non-existing field with the dynamic API (#3292).
 * `DynamicRealmObject.setList` should only accept `RealmList<DynamicRealmObject>` (#3280).
 * `DynamicRealmObject.getX(fieldName)` now throws a proper exception instead of a native crash when called with a field name of the wrong type (#3294).
+* Fixed a problem that `Table` and schema cache never hits against proxy classes (#3315).
```
cmelchior: Not sure I understand this sentence. Isn't the problem more that we cached the information twice?

realm/realm-library/src/main/java/io/realm/RealmSchema.java
```diff
@@ -221,20 +221,44 @@ Table getTable(String className) {
     Table getTable(Class<? extends RealmModel> clazz) {
         Table table = classToTable.get(clazz);
         if (table == null) {
-            clazz = Util.getOriginalModelClass(clazz);
-            table = transaction.getTable(realm.configuration.getSchemaMediator().getTableName(clazz));
-            classToTable.put(clazz, table);
+            Class<? extends RealmModel> originalClass = Util.getOriginalModelClass(clazz);
+            if (originalClass != clazz) {
```
cmelchior: Add a comment here that if this is true, then `clazz` is a proxy class?

realm/realm-library/src/main/java/io/realm/RealmSchema.java
```diff
@@ -221,20 +221,44 @@ Table getTable(String className) {
...
+                table = classToTable.get(originalClass);
+            }
+            if (table != null) {
+                // table was not cached against proxy class. Add cache entry.
```
cmelchior: or the table wasn't cached at all

cmelchior:
Nice catch. Great to get this simplified

CHANGELOG.md
zaki50: There are two problems.

One is that as you said we have two different maps for cache.

And another is that we put only original model classes as a key and never put proxy classes as a key. This means that cache entry is never used and we always recreate the cache entry when we pass a proxy class as an argument of `getTable()`.

I only mentioned the latter one. Should I mention both?

cmelchior: I would probably chuck it under "Enhancements" instead with something like "Optimized internal caching of classes" or something like that, but I don't have any strong opinions on it.

realm/realm-library/src/main/java/io/realm/RealmSchema.java
zaki50: no I think.

If the table wasn't cached at all, `table` is always `null`.
Only when the `clazz` represents proxy class and the cache has and entry only for original class, we can reach here.

cmelchior: Ah right, In that case I would fold it into the above if. It looks a bit strange they are seperate now.

realm/realm-library/src/main/java/io/realm/RealmSchema.java
zaki50: updated

CHANGELOG.md
zaki50: updated

realm/realm-library/src/main/java/io/realm/RealmSchema.java
zaki50: right. I'll update the code.

zaki50: @cmelchior done. please review 08abc6

realm/realm-library/src/main/java/io/realm/Realm.java
zaki50: used by proxy classes.

zaki50:
retest this please

kneth:
LGTM

realm/realm-library/src/main/java/io/realm/RealmSchema.java
```diff
@@ -221,20 +221,38 @@ Table getTable(String className) {
...
-            classToTable.put(clazz, table);
+            Class<? extends RealmModel> originalClass = Util.getOriginalModelClass(clazz);
+            if (originalClass != clazz) {
+                table = classToTable.get(originalClass);
+            }
+            if (table == null) {
+                table = transaction.getTable(realm.configuration.getSchemaMediator().getTableName(originalClass));
+                classToTable.put(originalClass, table);
+            }
+            if (originalClass != clazz) {
```
beeender: This three lines seem to be useless. It will be added at line 230 anyway.

realm/realm-library/src/main/java/io/realm/RealmSchema.java
```diff
@@ -221,20 +221,38 @@ Table getTable(String className) {
...
+            Class<? extends RealmModel> originalClass = Util.getOriginalModelClass(clazz);
+            if (originalClass != clazz) {
+                classSchema = classToSchema.get(originalClass);
+            }
+            if (classSchema == null) {
+                Table table = getTable(clazz);
+                classSchema = new RealmObjectSchema(realm, table, columnIndices.getColumnInfo(originalClass).getIndicesMap());
+                classToSchema.put(originalClass, classSchema);
+            }
+            if (originalClass != clazz) {
```
beeender: Same i think?

realm/realm-library/src/main/java/io/realm/RealmSchema.java
zaki50: No. `originalClass` is always model class which is defined by the user.
But `clazz` can be both user defined model class (`originalClass == class`) and generated proxy class (`originalClass != class`).
In latter case, we need to create another cache entry.

beeender:
:+1:

realm/realm-library/src/main/java/io/realm/Realm.java
cmelchior: Sounds fine. We can always change that in another PR

realm/realm-library/src/main/java/io/realm/RealmSchema.java
cmelchior: Since `originalClass != clazz` is used twice i would abstract it into `boolean isProxyClass`. I would also make the code more readable.

nhachicha:
LGTM

PR #3317, opened 2016-08-18
Remove the dependency of Realm.Handler from tests

beeender:
Instead, those tests should use method from RunInLooperThread.
Fix this in advance for the coming OS notification integration.


cmelchior:
👍

nhachicha:
👍

PR #1900, opened 2015-12-07
Fix crash when close Realm in listener

beeender:

PR #3319, opened 2016-08-18
Wait all async tasks done before next test

beeender:
This is highly related with #1900.
Below things are still guaranteed by this change after commit async
transaction:
* When callback function called (onSuccess/onError), the background
  Realm is closed.
* When any change listeners called, the background Realm is closed.

What is true now but it is not guaranteed in the future:
* Background Realm might not be closed before REALM_CHANGED sent (not
  received).

Due to this, to avoid the flaky tests, we have to ensure all async tasks
quit peacefully before start the next test. This is implemented by wait
and check in the TestRealmConfigurationFactory.

NOTE: Any test, if the async tasks cannot be finished peacefully in a
certain time, it has to be considered as a problem and fixed.

This would be needed by OS notifications since Java won't have a precise
control of sending REALM_CHANGED anymore.

### DIFF
--- a/realm/realm-library/src/main/java/io/realm/Realm.java
+++ b/realm/realm-library/src/main/java/io/realm/Realm.java
@@ -1189,8 +1189,9 @@ public final class Realm extends BaseRealm {
 
                     if (!Thread.currentThread().isInterrupted()) {
                         bgRealm.commitAsyncTransaction();
-                        // The bgRealm needs to be closed before post event to caller's handler to avoid concurrency
-                        // problem. This is currently guaranteed by posting handleAsyncTransactionCompleted below.
+                        // The bgRealm needs to be closed before posting the REALM_CHANGED event to the caller's handler
+                        // to avoid currency problems. This is currently guaranteed by posting
+                        // handleAsyncTransactionCompleted below.
                         bgRealm.close();
                         transactionCommitted = true;
                     }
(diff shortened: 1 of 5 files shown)


nhachicha:
LGTM

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -1188,15 +1188,10 @@ public void run() {
...
-                            @Override
-                            public void run() {
-                                // The bgRealm needs to be closed before post event to caller's handler to avoid
-                                // concurrency problem. eg.: User wants to delete Realm in the callbacks.
-                                // This will close Realm before sending REALM_CHANGED.
-                                bgRealm.close();
-                            }
-                        });
+                        bgRealm.commitAsyncTransaction();
+                        // The bgRealm needs to be closed before post event to caller's handler to avoid concurrency
```
cmelchior: The bgRealm needs to be closed before posting the REALM_CHANGED event to the caller's handler to avoid currency problems.his is currently guaranteed by posting handleAsyncTransactionCompleted below.

cmelchior:
A bit hacky but 😄 . Besides a smaller comment 👍 

As a side note: Me and @nhachicha talked about exposing the executor in a way similar to AsyncTask.THREAD_POOL_EXECUTOR. This PR  also seems to support that idea. So perhaps something we should consider more (In another issue/PR).

PR #3323, opened 2016-08-19
Smaller fixes to binary queries

cmelchior:
Follow up on missing items in #https://github.com/realm/realm-java/pull/3285

### DIFF
--- a/realm/realm-jni/src/io_realm_internal_TableQuery.cpp
+++ b/realm/realm-jni/src/io_realm_internal_TableQuery.cpp
@@ -926,6 +926,7 @@ static void TableQuery_BinaryPredicate(JNIEnv *env, jlong nativeQueryPtr, jlongA
     jsize arr_len = arr.len();
     try {
         BinaryData value2;
+        jbyte* bytePtr;
         if (value == NULL) {
             if (!TBL_AND_COL_NULLABLE(env, getTableByArray(nativeQueryPtr, arr).get(), arr[arr_len-1])) {
                 return;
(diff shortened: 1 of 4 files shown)


realm/realm-jni/src/io_realm_internal_TableQuery.cpp
```diff
@@ -933,15 +934,13 @@ static void TableQuery_BinaryPredicate(JNIEnv *env, jlong nativeQueryPtr, jlongA
             value2 = BinaryData();
         }
         else {
-            jbyte* bytePtr = env->GetByteArrayElements(value, NULL);
-            if (!bytePtr) {
+            if (!bytes.ptr()) {
```
beeender: Will it ever get here? line 930 checked it already i think?

nhachicha:
LGTM

realm/realm-jni/src/io_realm_internal_TableQuery.cpp
cmelchior: I agree, but this here http://stackoverflow.com/questions/25852147/when-does-getbytearrayelements-return-null seems to indicate it might fail in other ways. At least this check isn't harmful.

zaki50: @beeender this checks the result of `env->GetLongArrayElements(javaArray, NULL)`. not `value ` itself.

zaki50:
:+1:

beeender:
:+1:

PR #3324, opened 2016-08-19
Remove improper incompatibleLockFile test

beeender:
This test is not constructed properly. It tries to modify the lock file
while the Realm is still opened. This would cause some undefined
behaviour in core. Especially when we are using the generic
external_commit_helper which is based on SharedGroup::wait_for_change()
the helper thread will freeze because of polling hangs forever.

Just remove the test, core should ensure the proper exception thrown and
we have other tests to ensure the exception conversion right.

### DIFF
--- a/realm/realm-library/src/androidTest/java/io/realm/RealmTests.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/RealmTests.java
@@ -3428,24 +3428,4 @@ public class RealmTests {
         TestHelper.awaitOrFail(bgRealmFished);
         assertFalse(bgRealmChangeResult.get());
     }
-
-    @Test
-    public void incompatibleLockFile() throws IOException {
-        // Replace .lock file with a corrupted one
-        File lockFile = new File(realmConfig.getPath() + ".lock");
-        assertTrue(lockFile.exists());
-        FileOutputStream fooStream = new FileOutputStream(lockFile, false);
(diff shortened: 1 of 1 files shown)


cmelchior:
But modifying the lock file while it is open is the entire point of the test?
Are you saying that doing it will cause the helper thread to just hang forever?

beeender:
Yes. And modifying the lock file should not happen ever in the real like since it is protected by core with exclusive lock. If user modified the lock file through some other ways than core, then the behaviour is not guaranteed anymore.

The original test is trying to emulate a case like:

realm opened by core version 1
then the other process opened by core version 2

which is quite difficult to make since we have to use native code to make a exclusive flock. (java side FileLock doesn't corporate with flock)

beeender:
hmmm, but maybe you are right, still need to spend some more time to figure out why it hangs forever.
Back to in progress

beeender:
Checked again, I still think this test is not proper. We assume lock file should not be modified while the shared lock is hold, and the lock file validation happens only when the Realm gets opened. When holding a flock, it doesn't mean other process/threads cannot modify the file, it only means they cannot obtain the lock if they obey the flock rules.

This test will hang when we `wait_for_change()` called, since that need to access the mmap of the lock file.

cmelchior:
OK. I'm fine with removing the test 👍

zaki50:
:+1:

Issue #3327, opened 2016-08-22
Misleading migration example


mrexmelle:
#### Goal

> Providing a correct example to the developers who want to implement Realm migration procedure.
#### Expected Results

> When a developer wants to upgrade from version 1 to version 2, only one migration should take place, i.e. migration from 1 to 2.
#### Actual Results

>  When a developer wants to upgrade from version 1 to version 2, the migration continues until version 3. 
#### Steps & Code to Reproduce

> Clear enough.
## Currently

``` java

if(oldVersion == 0)
{
 // migration from version 0 to 1 takes place
 oldVersion++;
}

if(oldVersion == 1)
{
 // migration from version 1 to 2 takes place
 oldVersion++;
}

if(oldVersion == 2)
{
 // migration from version 2 to 3 takes place
 oldVersion ++;
}

```
## Expected

``` java

if(oldVersion == 0 && oldVersion<newVersion)
{
 // migration from version 0 to 1 takes place
 oldVersion++;
}

if(oldVersion == 1 && oldVersion<newVersion)
{
 // migration from version 1 to 2 takes place
 oldVersion++;
}

if(oldVersion == 2 && oldVersion<newVersion)
{
 // migration from version 2 to 3 takes place
 oldVersion ++;
}

```
#### Version of Realm and tooling

Realm version(s): 1.1.1

Android Studio version: 2.1.3

Which Android version and device: Any

Zhuinden:
Well it [**does** say](https://github.com/realm/realm-java/blob/master/examples/migrationExample/src/main/java/io/realm/examples/realmmigrationexample/model/Migration.java#L27) it's an example of migrating a Realm from its previous versions to SchemaVersion 3, and I'm not sure why you'd ever want to stop at SchemaVersion 2, but technically it's true, even if you specified schema version 2, you'd still end up with schema version 3.

cmelchior:
Hi @mrexmelle 
Technically you are correct, but since Realm doesn't allow downgrades I am hard pressed to see a situation where you would ship migration code for something that doesn't even exists yet?

Zhuinden:
Yeah, to go from 3 to 2 you would have to make a 4th schema version that is an inverse of the operations you do in the migration to 3

mrexmelle:
Hello @cmelchior,

[...]

PR #3328, opened 2016-08-22
Misleading migration example

mrexmelle:
The PR is created as a solution to issue #3327 Misleading migration example.


Zhuinden:
I can guarantee that you'll run into checkstyle issues because of the lack of whitespace

cmelchior:
I commented on the issue. While this change is technically correct I think it is trying to cover a case that shouldn't happen for 99% of developers, so I am not sure it is a good idea to merge this. But lets take the discussion in #3327

cmelchior:
Hi @mrexmelle 
Following the discussion in #3327, we will adopt the pattern described there instead. This will also solve the issues you ran into. 

This unfortunately means we won't merge this PR, but thanks for kicking off the discussion.

mrexmelle:
I'm cool. Thanks!

Issue #3274, opened 2016-08-09
RealmRecyclerViewAdapter not accept RealmList<T>


shannavas:
RealmList<TimeStamp> result = new RealmList<>();
       result.add(realm.where(TimeStamp.class).findFirst());

   `  recyclerView.setAdapter(new MyRecyclerViewAdapter(this,result));`

 ` public MyRecyclerViewAdapter(RecyclerViewExampleActivity activity,`OrderedRealmCollection<TimeStamp> data) {
      `  super(activity ,data, true);`
      `  this.activity = activity;`
    `}`
Adapter does't show any thing.
**RealmRecyclerViewAdapter OrderedRealmCollection not accept RealmList , its only accept RealmResult**

kneth:
In order to help you, we need you to use the issue template. Otherwise, we d not have enough information.

Zhuinden:
Make sure you set a RecyclerView.LayoutManager for your recycler view as well.

shannavas:
Recyclerview  working for RealmResult on setadapter()
but not working for RealmList<T> on setadapter()

```
RealmResults<Person> person = realm.where(Person.class).findAll();
                adapter = new BooksAdapter(MainActivity.this,person);
                recycler.setAdapter(adapter);
```
the above example is working fine

```
RealmList<person> firstPerson = new RealmList<>();
Person person = realm.where(Person.class).findFirst();
firstPersion.add(person);
                adapter = new BooksAdapter(MainActivity.this,firstPersion);
                recycler.setAdapter(adapter);
```
it's not working.

my adapter constructor
```
 public MyRecyclerViewAdapter(RecyclerViewExampleActivity activity,OrderedRealmCollection data){
super(activity,data,true);
}
```

Zhuinden:
Oh, your `RealmList<>` is unmanaged, and `isValid()` returns false on unmanaged objects.

https://github.com/realm/realm-java/issues/3101

So yeah, **managed** RealmLists work.

shannavas:
hoe to create managed RealmLists for the above.

Zhuinden:
Customize your adapter.

PR #3341, opened 2016-08-24
isValid() now returns true for unmanaged object. Added isManaged()

### LINKED ISSUE #3101: Align RealmObject.isValid() with Cocoa/.NET
See https://github.com/realm/realm-dotnet/pull/668/files/cf8ba4f0d920f7ca5044a4a1e242ae134f84be28#r69203861

Right now we return `false` for unmanaged objects, both Cocoa and .NET return `true` in the same case.
This has the advantage that you can use `isValid()` to determine if you can use the data from the RealmObject.

Java has some extra considerations though as we also have to consider the loaded state. Right now we return `false` if the object isn't loaded yet. That makes sense to me I think.

In any case, any change to these semantics are a breaking change, thus should only be part of 2.0

cmelchior:
Closes #3101 

This aligns our API with similar methods from Cocoa/.NET

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,11 +2,13 @@
 
 ### Breaking Changes
 
-* `isValid()` now returns always return `true` instead of `false` for unmanaged objects. This puts it in line with the behaviour from the Cocoa and .NET API's (#3101).
+* `isValid()` now always returns `true` instead of `false` for unmanaged objects. This puts it in line with the behaviour from the Cocoa and .NET API's (#3101).
 
 ### Enhancements
 
 * Added `realmObject.isManaged()` and `RealmObject.isManaged(obj)` (#3101).
+
+## 1.2.1
 
 ### Internal
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,16 @@
-## 1.2.1
+## 2.0.0
+
+### Breaking Changes
+
+* `isValid()` now returns always return `true` instead of `false` for unmanaged objects. This puts it in line with the behaviour from the Cocoa and .NET API's (#3101).
+
+### Enhancements
+
+* Added `realmObject.isManaged()` and `RealmObject.isManaged(obj)` (#3101).
 
 ### Internal
 
-* Move JNI build to CMake.
+* Moved JNI build to CMake.
 
 ## 1.2.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,11 +2,11 @@
 
 ### Breaking Changes
 
-* `isValid()` now always returns `true` instead of `false` for unmanaged objects. This puts it in line with the behaviour of the Cocoa and .NET API's (#3101).
+* `isValid()` now always returns `true` instead of `false` for unmanaged `RealmObject` and `RealmList`. This puts it in line with the behaviour of the Cocoa and .NET API's (#3101).
 
 ### Enhancements
 
-* Added `realmObject.isManaged()` and `RealmObject.isManaged(obj)` (#3101).
+* Added `realmObject.isManaged()`, `RealmObject.isManaged(obj)` and `RealmCollection.isManaged()` (#3101).
 
 ## 1.2.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### Breaking Changes
 
-* `isValid()` now always returns `true` instead of `false` for unmanaged objects. This puts it in line with the behaviour from the Cocoa and .NET API's (#3101).
+* `isValid()` now always returns `true` instead of `false` for unmanaged objects. This puts it in line with the behaviour of the Cocoa and .NET API's (#3101).
 
 ### Enhancements
 
--- a/realm/realm-library/src/main/java/io/realm/Realm.java
+++ b/realm/realm-library/src/main/java/io/realm/Realm.java
@@ -1319,8 +1319,8 @@ public final class Realm extends BaseRealm {
         if (realmObject == null) {
             throw new IllegalArgumentException("Null objects cannot be copied from Realm.");
         }
-        if (!RealmObject.isValid(realmObject)) {
-            throw new IllegalArgumentException("RealmObject is not valid, so it cannot be copied.");
+        if (!(RealmObject.isManaged(realmObject) && RealmObject.isValid(realmObject))) {
+            throw new IllegalArgumentException("Only valid managed objects can be copied from Realm.");
         }
         if (realmObject instanceof DynamicRealmObject) {
             throw new IllegalArgumentException("DynamicRealmObject cannot be copied from Realm.");
--- a/realm/realm-library/src/main/java/io/realm/RealmCollection.java
+++ b/realm/realm-library/src/main/java/io/realm/RealmCollection.java
@@ -140,7 +140,7 @@ public interface RealmCollection<E extends RealmModel> extends Collection<E> {
 
     /**
      * Checks if the collection is still valid to use, i.e., the {@link io.realm.Realm} instance hasn't been closed. It
-     * will always return {@code true} for unmanaged collection.
+     * will always return {@code true} for an unmanaged collection.
      *
      * @return {@code true} if it is still valid to use or an unmanaged collection, {@code false} otherwise.
      */
--- a/realm/realm-library/src/main/java/io/realm/RealmCollection.java
+++ b/realm/realm-library/src/main/java/io/realm/RealmCollection.java
@@ -149,7 +149,7 @@ public interface RealmCollection<E extends RealmModel> extends Collection<E> {
     /**
      * Checks if the collection is managed by Realm. A managed collection is just a wrapper around the data in the
      * underlying Realm file. On Looper threads, a managed collection will be live-updated so it always points to the
-     * latest data. Managed collections are thread confined so the they cannot be accessed from other threads than the
+     * latest data. Managed collections are thread confined so that they cannot be accessed from other threads than the
      * one that created them.
      * <p>
      *
--- a/realm/realm-library/src/main/java/io/realm/RealmCollection.java
+++ b/realm/realm-library/src/main/java/io/realm/RealmCollection.java
@@ -139,12 +139,27 @@ public interface RealmCollection<E extends RealmModel> extends Collection<E> {
     boolean load();
 
     /**
-     * Checks if the collection is still valid to use e.g. the {@link io.realm.Realm} instance hasn't
-     * been closed.
+     * Checks if the collection is still valid to use, i.e., the {@link io.realm.Realm} instance hasn't been closed. It
+     * will always return {@code true} for unmanaged collection.
      *
-     * @return {@code true} if still valid to use, {@code false} otherwise.
+     * @return {@code true} if it is still valid to use or an unmanaged collection, {@code false} otherwise.
      */
     boolean isValid();
+
+    /**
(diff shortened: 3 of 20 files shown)


cmelchior:
@realm/java

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -1319,8 +1319,8 @@ private void checkMaxDepth(int maxDepth) {
         if (realmObject == null) {
             throw new IllegalArgumentException("Null objects cannot be copied from Realm.");
         }
-        if (!RealmObject.isValid(realmObject)) {
-            throw new IllegalArgumentException("RealmObject is not valid, so it cannot be copied.");
+        if (!(RealmObject.isManaged(realmObject) && RealmObject.isValid(realmObject))) {
```
beeender: `if (!RealmObject.isManaged(realmObject) || !RealmObject.isValid(realmObject))` is easier for me to understand. but it is fine to leave as it is.

realm/realm-library/src/main/java/io/realm/RealmObject.java
```diff
@@ -242,6 +242,46 @@ public final boolean isLoaded() {
...
+     * Realm file. On Looper threads, a managed object will be live-updated so it always points to the latest data. It
+     * is possible to register a change listener using {@link #addChangeListener(RealmChangeListener)} to be notified
+     * when this happens. Managed objects are thread confined so the they cannot be accessed from other threads than
+     * the one that created them.
+     *
+     * If this method returns {@code false}, the object is unmanaged. An unmanaged object is just a normal Java object,
+     * so it can be parsed freely across threads, but the data in the object is not connected to the underlying data,
+     * so it will not be live updated.
+     *
+     * It is possible to convert a object from managed to unmanaged using {@link Realm#copyFromRealm(RealmModel)}.
```
beeender: `to create a managed object from an unmanaged object.`
technically it is not converting, the original object stays the same.

realm/realm-library/src/main/java/io/realm/RealmObject.java
```diff
@@ -242,6 +242,46 @@ public final boolean isLoaded() {
...
+     * A unmanaged object can be converted to a managed object using {@link Realm#copyToRealm(RealmModel)}.
```
beeender: same, not convert.

realm/realm-library/src/main/java/io/realm/RealmObject.java
```diff
@@ -242,6 +242,46 @@ public final boolean isLoaded() {
...
+     * Realm file. On Looper threads, a managed object will be live-updated so it always points to the latest data. It
+     * is possible to register a change listener using {@link #addChangeListener(RealmModel, RealmChangeListener)} to be
+     * notified when this happens. Managed objects are thread confined so the they cannot be accessed from other threads
+     * than the one that created them.
+     *
+     * If this method returns {@code false}, the object is unmanaged. An unmanaged object is just a normal Java object,
+     * so it can be parsed freely across threads, but the data in the object is not connected to the underlying data,
+     * so it will not be live updated.
+     *
+     * It is possible to convert a object from managed to unmanaged using {@link Realm#copyFromRealm(RealmModel)}.
```
beeender: same

beeender:
Some exception messages need to be fixed, like `DynamicRealmObject.java:54`
`checkValidObjectForDetach` needs to check `isManaged`

beeender:
@realm/java

CHANGELOG.md
```diff
@@ -1,8 +1,18 @@
+## 2.0.0
+
+### Breaking Changes
+
+* `isValid()` now always returns `true` instead of `false` for unmanaged objects. This puts it in line with the behaviour from the Cocoa and .NET API's (#3101).
```
kneth: "from" -> "of"???

realm/realm-library/src/main/java/io/realm/Realm.java
kneth: But with `&&`, you can evaluate faster (if `isManaged()` returns `false`, the subexpression is `false` no matter what `isValid()` returns).

realm/realm-library/src/main/java/io/realm/RealmObject.java
```diff
@@ -242,6 +242,48 @@ public final boolean isLoaded() {
     }
 
     /**
+     * Checks if this object is managed by Realm. A managed object is just a wrapper around the data in the underlying
+     * Realm file. On Looper threads, a managed object will be live-updated so it always points to the latest data. It
+     * is possible to register a change listener using {@link #addChangeListener(RealmChangeListener)} to be notified
+     * when this happens. Managed objects are thread confined so the they cannot be accessed from other threads than
```
kneth: "this happens" -> "changes happen"

realm/realm-library/src/main/java/io/realm/RealmObject.java
```diff
@@ -242,6 +242,48 @@ public final boolean isLoaded() {
...
+     * the one that created them.
+     *
```
kneth: `<p>` missing

realm/realm-library/src/main/java/io/realm/Realm.java
beeender: Same i think? if `isManged(realmObject)` returns false, the condition after `||` won't be evaluated. but minor issue, i will leave it like this.

realm/realm-library/src/main/java/io/realm/RealmObject.java
```diff
@@ -242,6 +242,48 @@ public final boolean isLoaded() {
...
+     * If this method returns {@code false}, the object is unmanaged. An unmanaged object is just a normal Java object,
+     * so it can be parsed freely across threads, but the data in the object is not connected to the underlying Realm,
+     * so it will not be live updated.
+     *
```
kneth: `<p>` missing

realm/realm-library/src/main/java/io/realm/RealmObject.java
```diff
@@ -242,6 +242,48 @@ public final boolean isLoaded() {
...
+     * It is possible to create a managed object from an unmanaged object by using
+     * {@link Realm#copyFromRealm(RealmModel)}. A unmanaged object can be created from a managed object by using
```
kneth: Opposite?

zaki50:
How about `RealmCollection#isValid()`?
I think we should change the behavior of it as well.

cmelchior:
Good point. Forgot about that one

Zhuinden:
This will fix https://github.com/realm/realm-java/issues/3274

realm/realm-library/src/main/java/io/realm/RealmObject.java
```diff
@@ -242,6 +242,52 @@ public final boolean isLoaded() {
     }
 
     /**
+     * Checks if this object is managed by Realm. A managed object is just a wrapper around the data in the underlying
+     * Realm file. On Looper threads, a managed object will be live-updated so it always points to the latest data. It
+     * is possible to register a change listener using {@link #addChangeListener(RealmChangeListener)} to be notified
+     * when changes happen. Managed objects are thread confined so the they cannot be accessed from other threads than
```
stk1m1: `so the they` -> `so that they`

realm/realm-library/src/main/java/io/realm/RealmObject.java
```diff
@@ -242,6 +242,52 @@ public final boolean isLoaded() {
...
+     * the one that created them.
+     * <p>
+     *
+     * If this method returns {@code false}, the object is unmanaged. An unmanaged object is just a normal Java object,
+     * so it can be parsed freely across threads, but the data in the object is not connected to the underlying Realm,
+     * so it will not be live updated.
+     * <p>
+     *
+     * It is possible to create a managed object from an unmanaged object by using
+     * {@link Realm#copyToRealm(RealmModel)}. A unmanaged object can be created from a managed object by using
```
stk1m1: `A unmanaged` -> `An unmanaged`

realm/realm-library/src/main/java/io/realm/Realm.java
stk1m1: I think `||` could be more preferred in this _short-circuit OR_ ops, but most would prefer to see _De Morgan_ applied anyway.

realm/realm-library/src/main/java/io/realm/RealmObject.java
```diff
@@ -242,6 +242,52 @@ public final boolean isLoaded() {
...
+     */
+    public boolean isManaged() {
+        return isManaged(this);
+    }
+
+    /**
+     * Checks if this object is managed by Realm. A managed object is just a wrapper around the data in the underlying
+     * Realm file. On Looper threads, a managed object will be live-updated so it always points to the latest data. It
+     * is possible to register a change listener using {@link #addChangeListener(RealmModel, RealmChangeListener)} to be
+     * notified when changes happen. Managed objects are thread confined so the they cannot be accessed from other threads
```
stk1m1: `so the` -> `so that`

realm/realm-library/src/main/java/io/realm/RealmObject.java
```diff
@@ -242,6 +242,52 @@ public final boolean isLoaded() {
...
+     * than the one that created them.
+     * <p>
+     *
+     * If this method returns {@code false}, the object is unmanaged. An unmanaged object is just a normal Java object,
+     * so it can be parsed freely across threads, but the data in the object is not connected to the underlying Realm,
+     * so it will not be live updated.
+     * <p>
+     *
+     * It is possible to create a managed object from an unmanaged object by using
+     * {@link Realm#copyToRealm(RealmModel)}. A unmanaged object can be created from a managed object by using
```
stk1m1: `A unmanaged` -> `An unmanaged`

beeender:
@realm/java Added `RealmCollection.isManaged()`

realm/realm-library/src/main/java/io/realm/Realm.java
cmelchior: Boolean logic always gives me a headache 😄

cmelchior:
Thanks for cleaning up the rest. Looks good to me 👍

realm/realm-library/src/main/java/io/realm/RealmCollection.java
```diff
@@ -139,14 +139,29 @@
...
      *
-     * @return {@code true} if still valid to use, {@code false} otherwise.
+     * @return {@code true} if it is still valid to use or an unmanaged collection, {@code false} otherwise.
      */
     boolean isValid();
 
     /**
+     * Checks if the collection is managed by Realm. A managed collection is just a wrapper around the data in the
+     * underlying Realm file. On Looper threads, a managed collection will be live-updated so it always points to the
+     * latest data. Managed collections are thread confined so the they cannot be accessed from other threads than the
```
zaki50: `so the they are` -> `so they are`?

realm/realm-library/src/main/java/io/realm/RealmCollection.java
```diff
@@ -139,14 +139,29 @@
     boolean load();
 
     /**
-     * Checks if the collection is still valid to use e.g. the {@link io.realm.Realm} instance hasn't
-     * been closed.
+     * Checks if the collection is still valid to use, i.e., the {@link io.realm.Realm} instance hasn't been closed. It
+     * will always return {@code true} for unmanaged collection.
```
Zhuinden: an

zaki50:
:+1:

PR #3354, opened 2016-08-26
revive the version check of ndk

zaki50:
I wasted our time by using wrong version of NDK.
To prevent us from doing the same mistake, I revived the checks of NDK version.

@realm/java

### DIFF
--- a/realm/realm-library/build.gradle
+++ b/realm/realm-library/build.gradle
@@ -430,3 +430,69 @@ if (project.hasProperty('dontCleanJniFiles')) {
     }
     clean.dependsOn cleanExternalBuildFiles
 }
+
+project.afterEvaluate {
+    android.libraryVariants.all { variant ->
+        variant.externalNativeBuildTasks[0].dependsOn(checkNdk)
+    }
+}
+
+task checkNdk() << {
+    def ndkPathInEnvVariable = System.env.ANDROID_NDK_HOME
+    if (!ndkPathInEnvVariable) {
+        throw new GradleException("environment variable 'ANDROID_NDK_HOME' must be set.")
+    }
+    checkNdk(ndkPathInEnvVariable)
+
+    def localPropFile = rootProject.file('local.properties')
+    if (!localPropFile.exists()) {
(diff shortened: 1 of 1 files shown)


cmelchior:
👍

realm/realm-library/build.gradle
```diff
@@ -430,3 +430,69 @@ if (project.hasProperty('dontCleanJniFiles')) {
...
+project.afterEvaluate {
+    android.libraryVariants.all { variant ->
+        variant.externalNativeBuildTasks[0].dependsOn(checkNdk)
+    }
+}
+
+task checkNdk() << {
+    def ndkPathInEnvVariable = System.env.ANDROID_NDK_HOME
+    if (!ndkPathInEnvVariable) {
+        throw new GradleException("environment variable 'ANDROID_NDK_HOME' must be set.")
```
emanuelez: "environment" -> "The environment"

realm/realm-library/build.gradle
```diff
@@ -430,3 +430,69 @@ if (project.hasProperty('dontCleanJniFiles')) {
...
+
+    def localPropFile = rootProject.file('local.properties')
+    if (!localPropFile.exists()) {
+        // we can skip the checks since 'ANDROID_NDK_HOME' will be used instead.
+    } else {
+        def String ndkPathInLocalProperties = getValueFromPropertiesFile(localPropFile, 'ndk.dir')
+        if (!ndkPathInLocalProperties) {
+            throw new GradleException("'ndk.dir' must be set in ${localPropFile.getAbsolutePath()}.")
+        }
+        if (new File(ndkPathInLocalProperties).getCanonicalPath()
```
emanuelez: You could compare the object themselves instead of the path. That will take care of system dependent case sensitivity

emanuelez:
Minor comments, otherwise 👍

realm/realm-library/build.gradle
zaki50: `File.equals()` does not seem to handle case sensitivity of filesystem correctly.

`new File('/usr')` is not equal to `new File('/Usr')` on case insensitive HFS+ filesystem.

PR #3360, opened 2016-08-29
revive the version check of ndk (#3354)

beeender:
* revive the version check of ndk in order not to waste our time by using wrong version of NDK.

* minor fix

* minor fix

* move 'checkNdk(ndkPathInLocalProperties)' to ensure the existance of the directory when obtainig canonical path


beeender:
This is needed for releases branch as well.

PR #3366, opened 2016-08-30
DeleteLocalRef when the ref is created in loop

beeender:
Add wrapper class for JNI local reference to delete the local ref after
using it.

This is reported by user on helpscout:
https://secure.helpscout.net/conversation/244053233/6163/?folderId=366141

And some useful explanation can be found:
http://stackoverflow.com/questions/24289724/jni-deletelocalref-clarification

Normally the local ref doesn't have to be deleted since they will be
cleaned up when program returns to Java from native code. Using it in a
loop is obvious a corner case: the size of local ref table is relatively
small (512 on Android). To avoid it, the local ref should be deleted
when using it in a loop.

### DIFF
--- a/realm/realm-library/src/androidTest/java/io/realm/RealmAsyncQueryTests.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/RealmAsyncQueryTests.java
@@ -2121,8 +2121,8 @@ public class RealmAsyncQueryTests {
     @RunTestInLooperThread
     public void batchUpdate_localRefIsDeletedInLoopOfNativeBatchUpdateQueries() {
         final Realm realm = looperThread.realm;
-        // For Android, the size of local ref map is 512
-        final int TEST_COUNT = 512;
+        // For Android, the size of local ref map is 512. Use 1024 for more pressure.
+        final int TEST_COUNT = 1024;
         final AtomicBoolean updatesTriggered = new AtomicBoolean(false);
         // The first time onChange gets called for every results.
         final AtomicInteger firstOnChangeCounter = new AtomicInteger(0);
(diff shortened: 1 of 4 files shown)


realm/realm-library/src/main/cpp/util.hpp
```diff
@@ -696,6 +697,29 @@ class JniBooleanArray {
     jint                m_releaseMode;
 };
 
+// Wrap jobject and call the DeleteLocalRef when destruction.
```
cmelchior: Wraps jobject and automatically calls DeleteLocalRef when this object is destroyed.

realm/realm-library/src/main/cpp/util.hpp
```diff
@@ -696,6 +697,29 @@ class JniBooleanArray {
...
+// DeleteLocalRef is not necessary to be called in most cases since all local references will be cleaned up when program
```
cmelchior: the program

cmelchior:
Create catch. A unit test capturing it would be nice though. Creating one will be a bit tricky since we limit the number of concurrent async tasks in Java to 100.

So for us to write a unit test I think we need to 
1) Create Async queries in batches of 90
2) Wait for all queries in the batch to complete, before starting the next batch
3) Once we started above 512 queries, do a commit. That will trigger all 500+ queries to get rerun which should overflow the reference table.

cmelchior:
👍

realm/realm-library/src/androidTest/java/io/realm/RealmAsyncQueryTests.java
```diff
@@ -2111,6 +2114,57 @@ public void run() {
...
 
+    // This test reproduce the issue in https://secure.helpscout.net/conversation/244053233/6163/?folderId=366141
+    // First it creates 512 async queries, then trigger a transaction to make the queries gets update with
+    // nativeBatchUpdateQueries. It should not exceed the limits of local ref map size in JNI.
+    @Test
+    @RunTestInLooperThread
+    public void batchUpdate_localRefIsDeletedInLoopOfNativeBatchUpdateQueries() {
+        final Realm realm = looperThread.realm;
+        // For Android, the size of local ref map is 512
+        final int TEST_COUNT = 512;
```
zaki50: Shouldn't we test at least one bigger than the size of local ref map?

zaki50:
:+1: except one minor comment

PR #3368, opened 2016-08-30
RealmLog

### LINKED ISSUE #3326: Improve control of [possibly flawed] RealmLog default/custom logger
#### Goal/Expected Results

More reasonable control of RealmLog.LOGGERS; mostly the ability to remove the default logger.
RealmBase automatically adds during static initialization:
https://github.com/realm/realm-java/blob/master/realm/realm-library/src/main/java/io/realm/BaseRealm.java#L80
Possibly a RealmLog.set(Logger logger) or a RealmLog.clear() would be sufficient.

#### Actual Results

First of all, RealmBase's "BuildConfig.DEBUG ? DebugAndroidLogger : ReleaseAndroidLogger" logic is based off of "com.getkeepsafe.relinker.BuildConfig".
This logic seems like a bug...two actually.
1) The logic should probably be based off of "io.realm.BuildConfig".
2) Libraries by default build as Release, and I am assuming Realm is, so this will always evaluate to ReleaseAndroidLogger.

Second of all, RealmLog provides no way to remove this logger.
The result is that adding a custom logger ends up showing some duplicate messages.

```java
if (REALM_LOG_LEVEL >= RealmLog.VERBOSE)
{
    if (false)
    {
        // Spews lots of not very useful [to me] info...
        io.realm.internal.Util.setDebugLevel(1);
    }

    //
    // NOTE: Realm already adds a DebugAndroidLogger or ReleaseAndroidLogger:
    //  https://github.com/realm/realm-java/blob/master/realm/realm-library/src/main/java/io/realm/BaseRealm.java#L80
    // But, that logic is flawed because it uses com.getkeepsafe.relinker.BuildConfig and will always compile as
    // Release and will always add ReleaseAndroidLogger.
    // Realm provides no way to remove this Logger.
    //
    AndroidLogger androidLogger = new AndroidLogger();
    androidLogger.setMinimumLogLevel(REALM_LOG_LEVEL);
    androidLogger.setTag("MYREALM");
    RealmLog.add(androidLogger);
}
```

Realm version(s): 1.1.1
Android Studio version: 2.1.3

cmelchior:
This PR is preparing our API for the extended logging capabilities soon exposed in core, and it consists of a few changes.

- `RealmLog` is now moved to the public API and refactored slightly
- All JNI log events are now routed through the `RealmLog`
- JNI logging events are now categorised as VERBOSE rather than DEBUG
- Log methods now accept var args to `String.format`. This is an optimization to previously where we did string concatenation and then threw the result away.
- RealmLog exposes the Log classes from Log4J. This is a good compromise as it closely mirrors the levels in core and also means the log levels will closely match anything running on a normal JVM.
- The `AndroidLogger` class now maps between Realm Log levels and Android Log levels.
- Closes #3326 

The general idea is that all logging events should be routed through `RealmLog` so people have a chance of enabling/disabling events as needed. Slightly unorthodox is perhaps the `getMinimumLogLevel` on the `Logger` interface, but it is needed so we can more effectively disable logging in Core as otherwise sending a large amount of log events through JNI to get them discarded in Java would be too expensive. 

@realm/java


realm/realm-library/src/main/java/io/realm/log/LogLevel.java
```diff
@@ -0,0 +1,71 @@
...
+ */
+public class LogLevel {
+
+    /**
+     * The ALL has the lowest possible rank and is intended to turn on all logging.
+     */
+    public static final int ALL = 1;
+
+    /**
+     * The TRACE Level designates finer-grained informational events than the DEBUG
```
beeender: `DEBUG` -> `DEBUG.`

realm/realm-library/src/main/java/io/realm/log/Logger.java
```diff
@@ -0,0 +1,91 @@
...
+package io.realm.log;
+
+/**
+ * Interface for custom loggers that can be registered at {@link RealmLog#add(Logger)}.
+ * The different log levels are described in {@link LogLevel}.
+ */
+public interface Logger {
+
+    /**
+     * Define which {@link LogLevel} events this logger cares about. This value is used primarily by Realms underlying
```
beeender: Defines

realm/realm-library/src/main/java/io/realm/log/Logger.java
```diff
@@ -0,0 +1,91 @@
...
+     * storage engine to determine which events are sent to {@link RealmLog}.
+     *
```
beeender: <p>

realm/realm-library/src/main/java/io/realm/log/Logger.java
```diff
@@ -0,0 +1,91 @@
...
+     * If multiple loggers are registered, the minimum value among all loggers is used.
+     *
```
beeender: <p>

realm/realm-library/src/main/cpp/util.hpp
```diff
@@ -114,29 +114,36 @@ jclass GetClass(JNIEnv* env, const char* classStr);
...
-    #define TR_LEAVE() if (trace_level >= 3) { __android_log_print(ANDROID_LOG_DEBUG, log_tag, " <-- %s", __FUNCTION__); } else {}
+    #define TR_ENTER(env) if (trace_level <= 2) { env->CallStaticVoidMethod(realmlog_class, log_trace, " --> %s", __FUNCTION__); } else {}
+    #define TR_ENTER_PTR(env, ptr) if (trace_level <= 2) { env->CallStaticVoidMethod(realmlog_class, log_trace, " --> %s %", PRId64, __FUNCTION__, static_cast<int64_t>(ptr)); } else {}
+    #define TR(env, ...) if (trace_level <= 2) { env->CallStaticVoidMethod(realmlog_class, log_trace, __VA_ARGS__); } else {}
+    #define TR_ERR(env, ...) if (trace_level <= 6) { env->CallStaticVoidMethod(realmlog_class, log_error, __VA_ARGS__); } else {}
+    #define TR_LEAVE(env) if (trace_level <= 2) { env->CallStaticVoidMethod(realmlog_class, log_trace, " <-- %s", __FUNCTION__); } else {}
   #else // ANDROID
-    #define TR_ENTER()
-    #define TR_ENTER_PTR(ptr)
+    #define TR_ENTER(env)
```
beeender: I like to have `env` defined in the macro, but that requires lots of changes in jni. Are you sure all of them are changed? Maybe we can keep it like before and change it later.

beeender: Uh, never mind. env is not needed before. Just ignore my above comment.

realm/realm-library/src/main/cpp/io_realm_internal_Util.cpp
```diff
@@ -71,6 +84,17 @@ JNIEXPORT void JNI_OnUnload(JavaVM* vm, void*)
 
 JNIEXPORT void JNICALL Java_io_realm_internal_Util_nativeSetDebugLevel(JNIEnv*, jclass, jint level)
 {
+    /**
```
beeender: if you add `LogLevel.java` to the cmake `create_javah`, it will generate the header fine with all trace level defined in the cpp land. And just use them instead of using immediate numbers.

beeender:
GOOD!!! Except some comments :+1:

CHANGELOG.md
```diff
@@ -14,15 +14,16 @@
 
 * Added `realmObject.isManaged()`, `RealmObject.isManaged(obj)` and `RealmCollection.isManaged()` (#3101).
 * Added `RealmConfiguration.Builder.directory(File)`.
-
-### Internal
-
-* Moved JNI build to CMake.
+* `RealmLog` has been moved to the public API. It is now possible to control which events Realm emit to LogCat. See the `RealmLog` class for more details.
```
zaki50: `LogCat` -> `logcat`

CHANGELOG.md
```diff
@@ -14,15 +14,16 @@
...
 
 ### Bug fixes
 
 * Fixed a lint error in proxy classes when the 'minSdkVersion' of user's project is smaller than 11 (#3356).
 * Fixed a potential crash when there were lots of async queries waiting in the queue.
+s
```
kneth: s?

realm/realm-library/src/main/java/io/realm/log/LogLevel.java
```diff
@@ -0,0 +1,71 @@
...
+     */
+    public static final int TRACE = 2;
+
+    /**
+     * The DEBUG Level designates fine-grained informational events that are most useful to debug an application.
```
kneth: "Level" -> "level"

realm/realm-library/src/main/java/io/realm/log/LogLevel.java
kneth: "Level" -> "level"

realm/realm-library/src/main/java/io/realm/log/Logger.java
```diff
@@ -0,0 +1,91 @@
...
+     * @return the minimum {@link LogLevel} events this logger cares about.
+     */
+    int getMinimumLogLevel();
+
+    /**
+     * Handles a {@link LogLevel#TRACE} event.
+     *
+     * @param throwable optional exception to log.
+     * @param message optional additional message.
+     * @param args optional args used to format the message using {@link String#format(String, Object...)}.
```
kneth: "optional args" -> "optional arguments" (multiple places)

kneth:
:+1: with documentation comments

realm/realm-library/src/main/cpp/io_realm_internal_Util.cpp
cmelchior: Great idea. Done

cmelchior:
Lots of problems fixed. Round two @realm/java

realm/realm-library/src/main/java/io/realm/log/RealmLog.java
```diff
@@ -0,0 +1,300 @@
...
+ */
+public final class RealmLog {
+
+    private static final Logger[] NO_LOGGERS = new Logger[0];
+
+    // All of the below should be modified together under under a lock on LOGGERS.
+    private static final List<Logger> LOGGERS = new ArrayList<>();
+    private static volatile Logger[] loggersAsArray = NO_LOGGERS;
+    private static volatile int minimumNativeLogLevel = Integer.MAX_VALUE;
+    private static int minimumNativeDebugLevel;
```
zaki50: What's the difference of these `minimumNativeLogLevel` and `minimumNativeDebugLevel`?
`minimumNativeLogLevel` does not seem to be updated.

realm/realm-library/src/main/java/io/realm/log/RealmLog.java
```diff
@@ -0,0 +1,300 @@
...
+            loggersAsArray = NO_LOGGERS;
+        }
+    }
+
+    /**
+     * Logs a {@link LogLevel#TRACE} exception.
+     *
+     * @param throwable exception to log.
+     */
+    public static void trace(Throwable throwable) {
```
zaki50: `Keep` annotation is needed (called from JNI)

zaki50: uh, not this one

realm/realm-library/src/main/java/io/realm/log/RealmLog.java
```diff
@@ -0,0 +1,300 @@
...
+        trace(throwable, null);
+    }
+
+    /**
+     * Logs a {@link LogLevel#TRACE} event.
+     *
+     * @param message message to log.
+     * @param args optional args used to format the message using {@link String#format(String, Object...)}.
+     */
+    public static void trace(String message, Object... args) {
```
zaki50: `Keep` annotation is needed (called from JNI)

realm/realm-library/src/main/java/io/realm/log/RealmLog.java
```diff
@@ -0,0 +1,300 @@
...
+        debug(throwable, null);
+    }
+
+    /**
+     * Logs a {@link LogLevel#DEBUG} event.
+     *
+     * @param message message to log.
+     * @param args optional args used to format the message using {@link String#format(String, Object...)}.
+     */
+    public static void debug(String message, Object... args) {
```
zaki50: `Keep` annotation is needed (called from JNI)

realm/realm-library/src/main/java/io/realm/log/RealmLog.java
```diff
@@ -0,0 +1,300 @@
...
+        info(throwable, null);
+    }
+
+    /**
+     * Logs an {@link LogLevel#INFO} event.
+     *
+     * @param message message to log.
+     * @param args optional args used to format the message using {@link String#format(String, Object...)}.
+     */
+    public static void info(String message, Object... args) {
```
zaki50: `Keep` annotation is needed (called from JNI)

realm/realm-library/src/main/java/io/realm/log/RealmLog.java
```diff
@@ -0,0 +1,300 @@
...
+        warn(throwable, null);
+    }
+
+    /**
+     * Logs a {@link LogLevel#WARN} event.
+     *
+     * @param message message to log.
+     * @param args optional args used to format the message using {@link String#format(String, Object...)}.
+     */
+    public static void warn(String message, Object... args) {
```
zaki50: `Keep` annotation is needed (called from JNI)

realm/realm-library/src/main/java/io/realm/log/RealmLog.java
```diff
@@ -0,0 +1,300 @@
...
+        error(throwable, null);
+    }
+
+    /**
+     * Logs an {@link LogLevel#ERROR} event.
+     *
+     * @param message message to log.
+     * @param args optional args used to format the message using {@link String#format(String, Object...)}.
+     */
+    public static void error(String message, Object... args) {
```
zaki50: `Keep` annotation is needed (called from JNI)

realm/realm-library/src/main/java/io/realm/log/RealmLog.java
```diff
@@ -0,0 +1,300 @@
...
+        fatal(throwable, null);
+    }
+
+    /**
+     * Logs an {@link LogLevel#FATAL} event.
+     *
+     * @param message message to log.
+     * @param args optional args used to format the message using {@link String#format(String, Object...)}.
+     */
+    public static void fatal(String message, Object... args) {
```
zaki50: `Keep` annotation is needed (called from JNI)

realm/realm-library/src/main/java/io/realm/log/Logger.java
```diff
@@ -0,0 +1,91 @@
...
+    int getMinimumLogLevel();
+
+    /**
+     * Handles a {@link LogLevel#TRACE} event.
+     *
+     * @param throwable optional exception to log.
+     * @param message optional additional message.
+     * @param args optional arguments used to format the message using {@link String#format(String, Object...)}.
+     */
+    void trace(Throwable throwable, String message, Object... args);
```
zaki50: String formatting should be done by `RealmLog` since executing `String.format(message, args)` in each `Logger` is waste of CPU cycle.

zaki50:
second round done.

realm/realm-library/src/main/java/io/realm/log/Logger.java
cmelchior: Technically you are right, but right now each logger decided if they want to log or not. Which means that no work is done if the logger doesn't care about it. Also, it would prevent loggers from doing their own formatting if they wanted to.

realm/realm-library/src/main/java/io/realm/log/RealmLog.java
zaki50: I think we don't need `minimumNativeLogLevel`

realm/realm-library/src/main/java/io/realm/log/RealmLog.java
beeender: `Keep` should just be applied to the class, right?

beeender:
`AndroidLogger` actually can be implemented in JNI instead. In that way, we don't need env any more and can log in native thread without jvm.

cmelchior:
the `AndroidLogger` is also used by the binding.

beeender:
Yeah, `AndroidLogger` will still exist, but just implement it with `_android_log` in the JNI. So when need jni logging, no need to go through jni->java->android log implementation in jni->loggging, just jni->android log implementation in jni->loggging

beeender:
But that make logging level control a bit worse ... hmmm ...

realm/realm-library/src/main/java/io/realm/log/RealmLog.java
zaki50: Ah, right.

I think it's OK to apply @Keep` annotation to whole class.

cmelchior:
I agree that native logging becomes a bit awkward, but right now all the `TR` macros have been degraded to `VERBOSE/TRACE` instead of `DEBUG` and the default logging level is `WARNING` so it shouldn't be high traffic enough to matter.

I agree that if someone actually sets the level to `INFO` and we start logging a high amount of INFO events, it could have a performance penalty, but apart from that I think it should be fine.

Also note that it is possible to set the `getMinimumDebugLevel` to a higher value to disable native logging and still log lower java. Perhaps I should rename that to `getMinimumNativeDebugLevel()` just to make that more clear?

realm/realm-library/src/main/java/io/realm/log/RealmLog.java
cmelchior: Adding it to the entire class. Also safer going forward.

beeender:
:+1:

cmelchior:
Comments fixed @realm/java

realm/realm-library/src/main/java/io/realm/log/RealmLog.java
```diff
@@ -0,0 +1,300 @@
...
+ */
+@Keep
+public final class RealmLog {
+
+    private static final Logger[] NO_LOGGERS = new Logger[0];
+
+    // All of the below should be modified together under under a lock on LOGGERS.
+    private static final List<Logger> LOGGERS = new ArrayList<>();
+    private static volatile Logger[] loggersAsArray = NO_LOGGERS;
+    private static volatile int minimumNativeLogLevel = Integer.MAX_VALUE;
```
zaki50: no `volatile` needed

realm/realm-library/src/main/java/io/realm/log/RealmLog.java
```diff
@@ -0,0 +1,300 @@
...
+     */
+    public static void add(Logger logger) {
+        if (logger == null) {
+            throw new IllegalArgumentException("A non-null logger has to be provided");
+        }
+        synchronized (LOGGERS) {
+            LOGGERS.add(logger);
+            int minimumLogLevel = logger.getMinimumNativeDebugLevel();
+            if (minimumLogLevel < minimumNativeLogLevel) {
+                setMinimumNativeDebugLevel(minimumLogLevel);
```
zaki50: We need to update `minimumNativeLogLevel` in `setMinimumNativeDebugLevel()`

realm/realm-library/src/main/java/io/realm/log/RealmLog.java
cmelchior: Since you can remove/add loggers from any thread, I think it is needed?

realm/realm-library/src/main/java/io/realm/log/RealmLog.java
cmelchior: Ups, fixed

CHANGELOG.md.orig
```diff
@@ -0,0 +1,717 @@
+## 2.0.0
```
zaki50: 👀

realm/realm-library/src/main/java/io/realm/log/RealmLog.java
zaki50: any read and write of `minimumNativeLogLevel` is synchronized with `LOGGERS`.
I think no extra volatile is necessary

CHANGELOG.md.orig
zaki50: please remove this file

zaki50:
👍  other than `CHANGELOG.md.orig`

PR #3370, opened 2016-08-31
Integrate Object Store [PART4] - OS notifications

beeender:
* Use OS's notification mechanism to notify threads.
* Create RealmNotificer interface for decouple Android related handler
  logic.
* Create AndroidNotifier for the handler notifications.

The major change of this PR is about the timing. The notifications are
not sent immediately after transaction committed. Instead, there is a
dedicated thread monitoring changes and notify others when changes
happen.

The known problem is for every RealmConfiguration, a monitor thread will
be created which is not ideal for app which is using multiple
RealmConfiguration.

There are different implementations for the monitoring thread in OS. For
Android, we can choose from generic which is based on the core's
wait_for_change() and android which is used by dotnet based on the
named pipe.
To align with dotnet, we are using the named pipe for now which also
enables notifications between realm-java and realm-dotnet.


beeender:
@realm/java need review please.

realm/realm-library/src/main/cpp/java_binding_context.cpp
```diff
@@ -0,0 +1,60 @@
...
+        // Always try to attach here since this may be called in the finalizer/phantom thread where m_local_jni_env
+        // should not be used on. No need to call DetachCurrentThread since this thread should always be created by
+        // JVM.
+        JNIEnv *env;
+        m_jvm->AttachCurrentThread(&env, nullptr);
+        env->DeleteWeakGlobalRef(m_java_notifier);
+    }
+}
+
+void JavaBindingContext::changes_available()
```
cmelchior: Perhaps `notify_changes_available` instead for better readability?

beeender: uh. it is an overwrite function from OS :grinning:

realm/realm-library/src/main/java/io/realm/AndroidNotifier.java
```diff
@@ -0,0 +1,166 @@
...
+import io.realm.internal.HandlerControllerConstants;
+import io.realm.internal.RealmNotifier;
+import io.realm.internal.async.QueryUpdateTask;
+import io.realm.internal.log.RealmLog;
+
+/**
+ * Implementation of {@link RealmNotifier} for Android based on {@link Handler}.
+ */
+// FIXME: Please move me to the io.realm.internal when HandlerController is there.
+class AndroidNotifier implements RealmNotifier {
```
cmelchior: Great abstraction for JVM 👍

realm/realm-library/src/main/java/io/realm/AndroidNotifier.java
```diff
@@ -0,0 +1,166 @@
...
+
+    public AndroidNotifier(HandlerController handlerController) {
+        if (isAutoRefreshAvailable()) {
+            handler = new Handler(handlerController);
+        }
+    }
+
+    // Called by Java when transaction committed to send LOCAL_COMMIT to current thread's handler.
+    @Override
+    public void notifyByLocalThread() {
```
cmelchior: notifyCommitByLocalThread?

realm/realm-library/src/main/java/io/realm/AndroidNotifier.java
```diff
@@ -0,0 +1,166 @@
...
+    // | Thread A                                                      | Thread B     | Daemon Thread                                  |
+    // |---------------------------------------------------------------+--------------+------------------------------------------------|
+    // |                                                               | Make changes |                                                |
+    // |                                                               |              | Detect and notify thread A through JNI ALooper |
+    // | Call Realm::notify() in ALooper callback                      |              |                                                |
+    // | Realm::notify() calls JavaBindingContext:change_available()   |              |                                                |
+    // | change_available calls into this method to send REALM_CHANGED |              |                                                |
+    // |---------------------------------------------------------------+--------------+------------------------------------------------|
+    @Override
+    public void notifyByOtherThread() {
```
cmelchior: `notifyCommitByOtherThread()`?

realm/realm-library/src/main/java/io/realm/AndroidNotifier.java
```diff
@@ -0,0 +1,166 @@
...
+        Message msg = Message.obtain();
+        msg.what = HandlerControllerConstants.LOCAL_COMMIT;
+        if (!handler.hasMessages(HandlerControllerConstants.LOCAL_COMMIT)) {
+            handler.removeMessages(HandlerControllerConstants.REALM_CHANGED);
+            handler.sendMessageAtFrontOfQueue(msg);
+        }
+    }
+
+    // This is called by OS when other thread/process changes the Realm.
+    // This is getting called on the same thread which creates the Realm.
```
cmelchior: which created

realm/realm-library/src/main/java/io/realm/AndroidNotifier.java
```diff
@@ -0,0 +1,166 @@
...
+    // |                                                               |              | Detect and notify thread A through JNI ALooper |
+    // | Call Realm::notify() in ALooper callback                      |              |                                                |
+    // | Realm::notify() calls JavaBindingContext:change_available()   |              |                                                |
+    // | change_available calls into this method to send REALM_CHANGED |              |                                                |
+    // |---------------------------------------------------------------+--------------+------------------------------------------------|
+    @Override
+    public void notifyByOtherThread() {
+        if (handler == null) return;
+
+        // For all other threads, use the Handler
```
cmelchior: This line no longer makes sense. Just remove?

realm/realm-library/src/main/java/io/realm/AndroidNotifier.java
```diff
@@ -0,0 +1,166 @@
...
+
+    @Override
+    public boolean isValid() {
+        return handler != null;
+    }
+
+    @Override
+    public void close() {
+        if (handler != null) {
+            handler.removeCallbacksAndMessages(null);
```
cmelchior: Also set handler to null?

realm/realm-library/src/main/java/io/realm/AndroidNotifier.java
```diff
@@ -0,0 +1,166 @@
...
+    public void completeUpdateAsyncQueries(QueryUpdateTask.Result result) {
+        Looper looper = handler.getLooper();
+        if (looper.getThread().isAlive()) {     // The receiving thread is alive
+            handler.obtainMessage(HandlerControllerConstants.COMPLETED_UPDATE_ASYNC_QUERIES, result).sendToTarget();
+        }
+    }
+
+    private static boolean isAutoRefreshAvailable() {
+        return (Looper.myLooper() != null && !isIntentServiceThread());
+
```
cmelchior: Remove whitespace

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -76,6 +74,8 @@
     private final static Long INVALID_NATIVE_POINTER = 0L;
     private ArgumentsHolder argumentsHolder;
 
+    ;
```
cmelchior: remove

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -2175,24 +2172,39 @@ private void checkSortParameters(String fieldNames[], final Sort[] sortOrders) {
...
+                    notifier.completeAsyncResults((QueryUpdateTask.Result)obj);
+                    break;
+                case COMPLETE_ASYNC_OBJECT:
+                    notifier.completeAsyncObject((QueryUpdateTask.Result)obj);
+                    break;
+                case THROW_BACKGROUND_EXCEPTION:
+                    notifier.throwBackgroundException((Throwable)obj);
+                    break;
+                default:
+                    // Should not get here.
```
cmelchior: Throw `IllegalState` then

realm/realm-library/src/main/java/io/realm/internal/RealmNotifier.java
```diff
@@ -0,0 +1,66 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.realm.internal;
+
+import io.realm.internal.async.QueryUpdateTask;
+
+
```
cmelchior: Remove whitespace

realm/realm-library/src/main/java/io/realm/internal/RealmNotifier.java
```diff
@@ -0,0 +1,66 @@
...
+/**
+ * This interface needs to be implemented by Java and pass to OS in order to get notifications when other thread/process
+ * changes the Realm file.
+ */
+public interface RealmNotifier {
+    /**
+     * This is called from Java when the changes have been made on the same thread.
+     */
+    void notifyByLocalThread();
+
```
cmelchior: White space

realm/realm-library/src/main/java/io/realm/internal/async/QueryUpdateTask.java
```diff
@@ -93,9 +99,21 @@ public void run() {
...
+                        notifier.completeAsyncResults(result);
+                        break;
+                    case COMPLETE_ASYNC_OBJECT:
+                        notifier.completeAsyncObject(result);
+                        break;
+                    case COMPLETE_UPDATE_ASYNC_QUERIES:
+                        notifier.completeUpdateAsyncQueries(result);
+                        break;
+                    default:
+                        throw new RuntimeException("Wrong event " + event);
```
cmelchior: `IllegalArgument` instead?

realm/realm-library/src/main/java/io/realm/AndroidNotifier.java
```diff
@@ -0,0 +1,166 @@
...
+    // This is called by OS when other thread/process changes the Realm.
+    // This is getting called on the same thread which creates the Realm.
+    // FIXME: The whole calling routine is twisted and needs to be rewritten in the near future.
+    // |---------------------------------------------------------------+--------------+------------------------------------------------|
+    // | Thread A                                                      | Thread B     | Daemon Thread                                  |
+    // |---------------------------------------------------------------+--------------+------------------------------------------------|
+    // |                                                               | Make changes |                                                |
+    // |                                                               |              | Detect and notify thread A through JNI ALooper |
+    // | Call Realm::notify() in ALooper callback                      |              |                                                |
```
cmelchior: Perhaps clarify that this is happening in the OS? I was looking for the `notify()` method on the Java Realm.

cmelchior:
Mostly minor stuff regarding naming and layout. Apart from that it looks good, but we need to keep track of the remaining FIXME's somewhere.

realm/realm-library/src/main/cpp/java_binding_context.cpp
```diff
@@ -0,0 +1,60 @@
...
+JavaBindingContext::JavaBindingContext(const ConcreteJavaBindContext& concrete_context)
+    : m_local_jni_env(concrete_context.jni_env)
+{
+    jint ret = m_local_jni_env->GetJavaVM(&m_jvm);
+    if (ret != 0) {
+        throw std::runtime_error(util::format("Failed to get Java vm. Error: %d", ret));
+    }
+    if (concrete_context.java_notifier) {
+        m_java_notifier = m_local_jni_env->NewWeakGlobalRef(concrete_context.java_notifier);
+        jclass cls = m_local_jni_env->GetObjectClass(m_java_notifier);
```
kneth: Are checks on `GetObjectClass` and `GetMethodID` not required? Can we safely assume that they will succeed?

beeender: I think so. It should just crash during any of our tests. Or i can throw a c++ exception here if it is needed?

beeender: I will just throw a native exception here.

beeender: Uh... actually there is no need to do that. Those functions will never return `NULL`, instead it will just throw an exception.

beeender:
Created #3372 and https://github.com/realm/realm-object-store/issues/152 for FIXMEs.

beeender:
@realm/java Review again, thanks!

cmelchior:
👍

realm/realm-library/src/main/java/io/realm/internal/RealmNotifier.java
```diff
@@ -0,0 +1,66 @@
...
+ * limitations under the License.
+ */
+
+package io.realm.internal;
+
+import io.realm.internal.async.QueryUpdateTask;
+
+
+/**
+ * This interface needs to be implemented by Java and pass to OS in order to get notifications when other thread/process
```
kneth: "OS" -> "Realm Object Store"

realm/realm-library/src/main/java/io/realm/internal/RealmNotifier.java
```diff
@@ -0,0 +1,66 @@
...
+ */
+public interface RealmNotifier {
+    /**
+     * This is called from Java when the changes have been made on the same thread.
+     */
+    void notifyByLocalThread();
+
+
+    /**
+     * This is called in OS's JavaBindingContext::changes_available.
```
kneth: "OS" -> "Realm Object Store"

realm/realm-library/src/androidTest/java/io/realm/internal/PivotTest.java
```diff
@@ -19,6 +19,7 @@
 import junit.framework.TestCase;
 
 import io.realm.RealmFieldType;
+import io.realm.Sort;
```
zaki50: Can we remove this?

realm/realm-library/src/main/java/io/realm/internal/RealmNotifier.java
```diff
@@ -0,0 +1,66 @@
...
+     * This is getting called on the same thread which created this Realm when the same Realm file has been changed by
+     * other thread. The changes on the same thread should not trigger this call.
+     */
+    @SuppressWarnings("unused")
+    void notifyByOtherThread();
```
zaki50: This needs `@Keep` annotation

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
```diff
@@ -51,13 +54,16 @@ Java_io_realm_internal_SharedRealm_nativeCloseConfig(JNIEnv *, jclass, jlong con
...
 JNIEXPORT jlong JNICALL
-Java_io_realm_internal_SharedRealm_nativeGetSharedRealm(JNIEnv *env, jclass, jlong config_ptr)
+Java_io_realm_internal_SharedRealm_nativeGetSharedRealm(JNIEnv *env, jclass, jlong config_ptr, jobject notifier)
 {
     TR_ENTER_PTR(config_ptr)
 
     auto config = reinterpret_cast<realm::Realm::Config*>(config_ptr);
     try {
         auto shared_realm = Realm::get_shared_realm(*config);
+        shared_realm->m_binding_context = JavaBindingContext::create(env, notifier);
```
zaki50: no need to create global ref of `notifier`?

beeender: No, `JavaBindingContext` is holding a global weak ref there. And check the ref before using it. The life-cycle of `notifier` is totally controlled in java side -- as a member var of `SharedRealm.java`

zaki50: ah, constructor of `JavaBindingContext` is  creating weak global ref.
please ignore above comment.

realm/realm-library/src/main/cpp/java_binding_context.hpp
```diff
@@ -0,0 +1,66 @@
...
+#include "binding_context.hpp"
+
+namespace realm {
+
+namespace _impl {
+
+// Binding context which will be called from OS.
+class JavaBindingContext final : public BindingContext {
+private:
+    struct ConcreteJavaBindContext {
```
zaki50: just a question.

why do we need to create `ConcreteJavaBindContext` instead of using `env` and `notifier` directly?

beeender: To make the `JavaBindingContext` constructor private and still be able to use `std::make_shared`. See http://stackoverflow.com/questions/8147027/how-do-i-call-stdmake-shared-on-a-class-with-only-protected-or-private-const/8147213#8147213

zaki50:
What's the difference between `HandlerControllerConstants.COMPLETED_ASYNC_REALM_OBJECT` and `QueryUpdateTask.NotifyEvent.COMPLETE_ASYNC_OBJECT`.
Both needed?

beeender:
`HandlerControllerConstants.COMPLETED_ASYNC_REALM_OBJECT` is the event ID for Realm android handler, `QueryUpdateTask.NotifyEvent.COMPLETE_ASYNC_OBJECT` is used to decouple currently handler implementation from realm-java. They are not exact one-to-one matched. All of them might be able to be removed in #3372 . So I would like leave them in the next task.

zaki50:
:+1:

PR #3377, opened 2016-09-01
Use createObject(class, primaryKey) in test

beeender:
Since Realm.createObject(class) will be deprecated on master for Classes
with primary key defined, fix the test cases which use it first to avoid
more conflicts when merging.


zaki50:
There seems to be some other test cases that uses `createObject(Class)` for primary key classes.
All should be updated in this PR?

cmelchior:
👍

beeender:
@zaki50 Those in json need some other fixes as well... so i will fix them on master

zaki50:
@beeender How about `RealmModelTests#createObject()` ?

Issue #2536, opened 2016-03-31
Calling setter method in no argument constructor results in NPE


jonasrottmann:
#### Goal

Be able to use setter methods in the no arguments constructor.

#### Expected Results

No NPE and field set to the given value.

#### Actual Results

```
Caused by: java.lang.NullPointerException: Attempt to invoke virtual method 'void io.realm.BaseRealm.checkIfValid()' on a null object reference
                                                                                at io.realm.ModelRealmProxy.realmSet$string(ModelRealmProxy.java:62)
                                                                                at de.jonasrottmann.realmtest.Model.setString(Model.java:13)
                                                                                at de.jonasrottmann.realmtest.Model.<init>(Model.java:19)
                                                                                at io.realm.ModelRealmProxy.<init>(ModelRealmProxy.java:51)
                                                                                at io.realm.DefaultRealmModuleMediator.newInstance(DefaultRealmModuleMediator.java:80)
                                                                                at io.realm.BaseRealm.get(BaseRealm.java:521)
                                                                                at io.realm.Realm.createObject(Realm.java:709)
```

#### Code Sample

```java

public class Model extends RealmObject {
    public String string;



    public void setString(String string) {
        this.string = string;
    }



    public Model() {
        setString("Success!");
    }
}

```

#### Version of Realm and tooling
Realm version(s): 0.88.2

Android Studio version: 2.1 Preview 4

Which Android version and device: 6.0.1 (MHC19J) on Nexus 5X

cmelchior:
Hi @jonasrottmann 

I just tried the following which didn't fail:

```
public class DefaultValues extends RealmObject {
    private int age = 42;
    private String name;
    private String address;

    // Default constructor
    public DefaultValues() {
        this.name = "Foo";
        setAddress("Bar");
    }

    public int getAge() {
        return age;
    }

    public void setAge(int age) {
        this.age = age;
    }

    public String getName() {
        return name;
    }

    public void setName(String name) {
        this.name = name;
    }

    public String getAddress() {
        return address;
    }

    public void setAddress(String address) {
        this.address = address;
    }
}

    @Test
    public void defaultConstructor_standAlone() {
        DefaultValues obj = new DefaultValues();
        assertEquals(42, obj.getAge());
        assertEquals("Foo", obj.getName());
        assertEquals("Bar", obj.getAddress());
    }
```

zaki50:
no...

beeender:
[...]

PR #3379, opened 2016-09-01
Create object without setting PK is not allowed

beeender:
* createObject must be called with PK value when creating a object with
  PK defined.
* Creating objects from JSON must have have corresponding PK defined in
  the JSON object.

Known issue:
The default values for creating object from JSONStream will be differnt
from those created by createObject. Default values from default
constructor VS default values from core. This has to be addressed by #777

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## 2.0.0
 
+### Known issues
+
+* When create `RealmObject` from JSON stream, it will takes default values set by the non-managed `RealmObject`'s default constructor for those absent field in JSON object. This behaviour is different from other APIs when creating `RealmObject` (#3386).
+
 ### Breaking Changes
 
 * `isValid()` now always returns `true` instead of `false` for unmanaged `RealmObject` and `RealmList`. This puts it in line with the behaviour of the Cocoa and .NET API's (#3101).
@@ -9,7 +13,7 @@
   - `RealmIOExcpetion` has been removed and replaced by `RealmFileException`.
 * Removed `RealmConfiguration.Builder(Context, File)` and `RealmConfiguration.Builder(File)` constructors.
 * `RealmConfiguration.Builder.assetFile(Context, String)` has been renamed to `RealmConfiguration.Builder.assetFile(String)`.
-* `Realm.createObject(Class<E>)` and `DynamicRealm.createObjec(String)` throws `RealmException` if they are called to create an object with a primary key defined.
+* Object with primary key is now required to define it when the object is created. This means that `Realm.createObject(Class<E>)` and `DynamicRealm.createObject(String)` now throws `RealmException` if they are used to create an object with a primary key field. Use `Realm.createObject(Class<E>, Object)` or `DynamicRealm.createObject(String, Object)` instead.
 * Importing from JSON without the primary key field defined in the JSON object now throws `IllegalArgumentException`.
 
 ### Enhancements
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### Known issues
 
-* When create `RealmObject` from JSON stream, it will takes default values set by the non-managed `RealmObject`'s default constructor for those absent field in JSON object. This behaviour is different from other APIs when creating `RealmObject` (#3386).
+* When creating a `RealmObject` from a JSON stream, it will take the default values defined by its default constructor for those fields that are not defined in the JSON object. This behaviour is different from other APIs when creating `RealmObject`s.
 
 ### Breaking Changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,8 +9,8 @@
   - `RealmIOExcpetion` has been removed and replaced by `RealmFileException`.
 * Removed `RealmConfiguration.Builder(Context, File)` and `RealmConfiguration.Builder(File)` constructors.
 * `RealmConfiguration.Builder.assetFile(Context, String)` has been renamed to `RealmConfiguration.Builder.assetFile(String)`.
-* `Realm.createObject(Class<E>)` and `DynamicRealm.createObjec(String)` throws a `RealmException` if they are called to create an object with a primary key defined.
-* Importing from JSON without the primary key field defined in the JSON object now throws an exception.
+* `Realm.createObject(Class<E>)` and `DynamicRealm.createObjec(String)` throws `RealmException` if they are called to create an object with a primary key defined.
+* Importing from JSON without the primary key field defined in the JSON object now throws `IllegalArgumentException`.
 
 ### Enhancements
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@
   - `RealmIOExcpetion` has been removed and replaced by `RealmFileException`.
 * Removed `RealmConfiguration.Builder(Context, File)` and `RealmConfiguration.Builder(File)` constructors.
 * `RealmConfiguration.Builder.assetFile(Context, String)` has been renamed to `RealmConfiguration.Builder.assetFile(String)`.
+* `Realm.createObject(Class<E>)` and `DynamicRealm.createObjec(String)` throws a `RealmException` if they are called to create an object with a primary key defined.
+* Importing from JSON without the primary key field defined in the JSON object now throws an exception.
 
 ### Enhancements
 
--- a/realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmProxyClassGenerator.java
+++ b/realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmProxyClassGenerator.java
@@ -1632,7 +1632,6 @@ public class RealmProxyClassGenerator {
         writer.emitStatement("return obj");
         writer.endMethod();
         writer.emitEmptyLine();
-
     }
 
     private String columnInfoClassName() {
--- a/realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmProxyClassGenerator.java
+++ b/realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmProxyClassGenerator.java
@@ -1553,6 +1553,10 @@ public class RealmProxyClassGenerator {
         writer.emitEmptyLine();
     }
 
+    // FIXME: Since we need to check the PK in stream before create an object, this is now using copyToRealm instead of
+    // createObject() to avoid parse the stream twice. This brings a problem that the default value behaviour is
+    // different from those which are using the createObject. And it needs to be addressed by
+    // https://github.com/realm/realm-java/issues/777
     private void emitCreateUsingJsonStream(JavaWriter writer) throws IOException {
         writer.emitAnnotation("SuppressWarnings", "\"cast\"");
         writer.emitAnnotation("TargetApi", "Build.VERSION_CODES.HONEYCOMB");
@@ -1563,7 +1567,10 @@ public class RealmProxyClassGenerator {
                 Arrays.asList("Realm", "realm", "JsonReader", "reader"),
                 Collections.singletonList("IOException"));
 
-        writer.emitStatement("%s obj = realm.createObject(%s.class)",qualifiedClassName, qualifiedClassName);
+        if (metadata.hasPrimaryKey()) {
+            writer.emitStatement("boolean jsonHasPrimaryKey = false");
+        }
+        writer.emitStatement("%s obj = new %s()", qualifiedClassName, qualifiedClassName);
         writer.emitStatement("reader.beginObject()");
         writer.beginControlFlow("while (reader.hasNext())");
         writer.emitStatement("String name = reader.nextName()");
@@ -1601,7 +1608,7 @@ public class RealmProxyClassGenerator {
             } else {
                 RealmJsonTypeHelper.emitFillJavaTypeFromStream(
                         interfaceName,
-                        metadata.getSetter(fieldName),
+                        metadata,
                         fieldName,
                         qualifiedFieldType,
                         writer
@@ -1616,9 +1623,16 @@ public class RealmProxyClassGenerator {
         }
         writer.endControlFlow();
         writer.emitStatement("reader.endObject()");
+        if (metadata.hasPrimaryKey()) {
+            writer.beginControlFlow("if (!jsonHasPrimaryKey)");
+            writer.emitStatement(Constants.STATEMENT_EXCEPTION_NO_PRIMARY_KEY_IN_JSON, metadata.getPrimaryKey());
+            writer.endControlFlow();
+        }
+        writer.emitStatement("obj = realm.copyToRealm(obj)");
         writer.emitStatement("return obj");
         writer.endMethod();
         writer.emitEmptyLine();
+
     }
 
     private String columnInfoClassName() {
--- a/realm/realm-library/src/androidTest/java/io/realm/RealmJsonAbsentPrimaryKeyTests.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/RealmJsonAbsentPrimaryKeyTests.java
@@ -1,0 +1,161 @@
+/*
+ * Copyright 2016 Realm Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.realm;
+
+import org.json.JSONArray;
+import org.json.JSONException;
+import org.json.JSONObject;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+
+import java.io.IOException;
+import java.util.Arrays;
+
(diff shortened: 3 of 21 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,8 @@
   - `RealmIOExcpetion` has been removed and replaced by `RealmFileException`.
 * Removed `RealmConfiguration.Builder(Context, File)` and `RealmConfiguration.Builder(File)` constructors.
 * `RealmConfiguration.Builder.assetFile(Context, String)` has been renamed to `RealmConfiguration.Builder.assetFile(String)`.
+* `Realm.createObject(Class<E>)` and `DynamicRealm.createObjec(String)` throws a `RealmException` if they are called to create an object with a primary key defined.
+* Importing from JSON without the primary key field defined in the JSON object now throws an exception.
```
kneth: which exception?

CHANGELOG.md
```diff
@@ -9,6 +9,8 @@
   - `RealmIOExcpetion` has been removed and replaced by `RealmFileException`.
 * Removed `RealmConfiguration.Builder(Context, File)` and `RealmConfiguration.Builder(File)` constructors.
 * `RealmConfiguration.Builder.assetFile(Context, String)` has been renamed to `RealmConfiguration.Builder.assetFile(String)`.
+* `Realm.createObject(Class<E>)` and `DynamicRealm.createObjec(String)` throws a `RealmException` if they are called to create an object with a primary key defined.
```
kneth: "throws a" -> "throws"

realm/realm-library/src/androidTest/java/io/realm/RealmJsonAbsentPrimaryKeyTests.java
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+    @After
+    public void tearDown() {
+        if (realm != null) {
+            realm.close();
+        }
+    }
+
+    // parameters for testing absent primary key value. PrimaryKey field isabsent.
```
kneth: space missing after "is"

realm/realm-library/src/main/java/io/realm/internal/Table.java
```diff
@@ -373,25 +373,18 @@ public void moveLastOver(long rowIndex) {
         nativeMoveLastOver(nativePtr, rowIndex);
     }
 
-    public long addEmptyRow() {
+    /**
+     * Add an empty row to the table.
+     *
+     * @param checkPrimaryKey set to true to check if this table has a primary key defined. Throw an exception if yes.
```
kneth: Embed true in `@code`

realm/realm-library/src/main/java/io/realm/internal/Table.java
```diff
@@ -373,25 +373,18 @@ public void moveLastOver(long rowIndex) {
...
-                    break;
-                case INTEGER:
-                    if (findFirstLong(primaryKeyColumnIndex, INTEGER_DEFAULT_VALUE) != NO_MATCH) {
-                        throwDuplicatePrimaryKeyException(INTEGER_DEFAULT_VALUE);
-                    }
-                    break;
-                default:
-                    throw new RealmException("Cannot check for duplicate rows for unsupported primary key type: " + type);
-            }
+        if (!checkPrimaryKey && hasPrimaryKey()) {
```
kneth: In order to throw, both subexpressions must be true, right? So `checkPrimaryKey` must be false. This is not consistent with doc above. Or what am I missing :-)

beeender: Nice catch! will fix and add a test case.

zaki50:
Shouldn't we update Javadoc comments of `Realm#create*FromJson()`?

realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmProxyClassGenerator.java
```diff
@@ -1616,9 +1623,16 @@ private void emitCreateUsingJsonStream(JavaWriter writer) throws IOException {
...
+        if (metadata.hasPrimaryKey()) {
+            writer.beginControlFlow("if (!jsonHasPrimaryKey)");
+            writer.emitStatement(Constants.STATEMENT_EXCEPTION_NO_PRIMARY_KEY_IN_JSON, metadata.getPrimaryKey());
+            writer.endControlFlow();
+        }
+        writer.emitStatement("obj = realm.copyToRealm(obj)");
         writer.emitStatement("return obj");
         writer.endMethod();
         writer.emitEmptyLine();
+
```
zaki50: remove this?

realm/realm-library/src/androidTest/java/io/realm/RealmJsonNullPrimaryKeyTests.java
```diff
@@ -130,23 +125,28 @@ public void createOrUpdateObjectFromJson_primaryKey_isNullOrAbsent_fromJsonObjec
...
-        realm.commitTransaction();
-
+    public void createOrUpdateObjectFromJson_primaryKey_isNull_updateFromJsonObject() throws JSONException {
         // PrimaryKeyAsString
         if (clazz.equals(PrimaryKeyAsString.class)) {
+            realm.beginTransaction();
+            realm.createObject(clazz, null); // name = null, id =null
+            realm.createOrUpdateObjectFromJson(clazz, new JSONObject(jsonString));
+            realm.commitTransaction();
+
```
zaki50: Why do we have these same lines in both `if` and `else`?

beeender: HA, right, will change it. It was caused by that i thought the default value is `0` and `""`.

realm/realm-library/src/androidTest/java/io/realm/internal/JNITableTest.java
```diff
@@ -87,7 +87,7 @@ public void zeroColOperations() {
 
         // Add rows
         try { tableZeroCols.add("val");         fail("No columns in table"); } catch (IndexOutOfBoundsException ignored) {}
-        try { tableZeroCols.addEmptyRow();      fail("No columns in table"); } catch (IndexOutOfBoundsException ignored) {}
+        try { tableZeroCols.addEmptyRow(true);      fail("No columns in table"); } catch (IndexOutOfBoundsException ignored) {}
```
zaki50: please align `fail()` to other lines

realm/realm-library/src/main/java/io/realm/internal/Table.java
```diff
@@ -373,25 +373,18 @@ public void moveLastOver(long rowIndex) {
...
+            throw new RealmException(String.format("Class '%s' has a primary key defined and the primary key needs to" +
+                    " be set when creating new object.", getName()));
```
zaki50: `getName()` should be `Table.tableNameToClassName(getName())`

beeender:
@realm/java Updated.

CHANGELOG.md
```diff
@@ -9,6 +9,8 @@
   - `RealmIOExcpetion` has been removed and replaced by `RealmFileException`.
 * Removed `RealmConfiguration.Builder(Context, File)` and `RealmConfiguration.Builder(File)` constructors.
 * `RealmConfiguration.Builder.assetFile(Context, String)` has been renamed to `RealmConfiguration.Builder.assetFile(String)`.
+* `Realm.createObject(Class<E>)` and `DynamicRealm.createObjec(String)` throws `RealmException` if they are called to create an object with a primary key defined.
```
cmelchior: I would expand this a little I think. Something like

Objects with primary keys are now required to define it when the object is created. This means that `Realm.createObject(Class<E>)` and `DynamicRealm.createObject(String)` now throws `RealmException` if they are used to create an object with a primary key field. Use `createObject(Class, Object)` instead.

realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmProxyClassGenerator.java
```diff
@@ -1553,6 +1553,10 @@ private void emitCreateOrUpdateUsingJsonObject(JavaWriter writer) throws IOExcep
         writer.emitEmptyLine();
     }
 
+    // FIXME: Since we need to check the PK in stream before create an object, this is now using copyToRealm instead of
```
cmelchior: I am not sure we can ship this? It sounds like a major regression (at least for the ones using this API). 
I'm not sure what you mean by parsing the stream twice? Currently we convert to JSONObject, but that is the same as doing `copyToRealm`, just with JSON, no?

beeender: That is only for `createOrUpdateAllFromJson`, for updating. by creating directly from stream, we set the properties one by one. If this is not acceptable, I would like to implement the #777 first since it doesn't seem to be too difficult.

realm/realm-library/src/androidTest/java/io/realm/RealmJsonAbsentPrimaryKeyTests.java
```diff
@@ -0,0 +1,161 @@
...
+    public void createOrUpdateObjectFromJson_primaryKey_isNullOrAbsent_fromJsonObject() throws JSONException {
+        realm.beginTransaction();
+        thrown.expect(IllegalArgumentException.class);
+        realm.createOrUpdateObjectFromJson(clazz, new JSONObject(jsonString));
+        realm.commitTransaction();
+    }
+
+    // Testing absent primary key value for createAllFromJson()
+    @Test
+    public void createALlFromJson_primaryKey_isNullOrAbsent_fromJsonObject() throws JSONException {
```
cmelchior: ALl -> All

Also below

realm/realm-library/src/androidTest/java/io/realm/RealmJsonAbsentPrimaryKeyTests.java
```diff
@@ -0,0 +1,161 @@
...
+    final private String jsonString;
+
+    public RealmJsonAbsentPrimaryKeyTests(Class<? extends RealmObject> clazz, String jsonString) {
+        this.jsonString = jsonString;
+        this.clazz = clazz;
+    }
+
+    // Testing absent primary key value for createObjectFromJson()
+    @Test
+    public void createObjectFromJson_primaryKey_isNullOrAbsent_fromJsonObject() throws JSONException {
```
cmelchior: Shouldn't it just be `isAbsent`, it isn't null in any of the test data?

beeender: Uh... forget to change this name. Copied from the original test.

realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmProxyClassGenerator.java
cmelchior: It looks like it only effect the `create` part, not the `update` part right?
I see 3 choices:

1) Accept the current solution and document the breaking change 
2) Rewrite `createFromJsonStream` to create JSONObject instead and return the result from `createObjectFromJson(jsonObj)`
3) Add support for #777 

For expendiency, I would probably just go with 1) now make an issue for looking into 2/3 in th next week or two, because we need a better solution before releasing anything publically.

If 3) is indeed fairly easy to do that would be awesome, if not I think we should do 2.

beeender: See #3386

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -675,6 +687,11 @@ private Scanner getFullStringScanner(InputStream in) {
     public <E extends RealmModel> E createObject(Class<E> clazz) {
         checkIfValid();
         Table table = schema.getTable(clazz);
+        // Check and throw the exception earlier for a better exception message.
+        if (table.hasPrimaryKey()) {
+            throw new RealmException(String.format("'%s' has a primary key, use" +
+                    " 'createObject(Class<E>, Object)' to create it.", clazz.getSimpleName()));
```
zaki50: `clazz.getSimpleName()` returns obfuscated name if ProGuard was applied.
`Table.tableNameToClassName(table.getName())` is better I think.

realm/realm-library/src/main/java/io/realm/internal/Table.java
```diff
@@ -475,7 +469,7 @@ public long addEmptyRows(long rows) {
      * @return the row index of the appended row.
      */
     protected long add(Object... values) {
-        long rowIndex = addEmptyRow(false);
+        long rowIndex = addEmptyRow();
 
         checkImmutable();
```
zaki50: Though this is not related to the changes of PR, Is it OK to check immutability after adding a row?

zaki50: @beeender How about this?
Moved some checks first and supported primary key.

```
diff --git a/realm/realm-library/src/main/java/io/realm/internal/Table.java b/realm/realm-library/src/main/java/io/realm/internal/Table.java
index 5b3c1ba..4b557b7 100644
--- a/realm/realm-library/src/main/java/io/realm/internal/Table.java
+++ b/realm/realm-library/src/main/java/io/realm/internal/Table.java
@@ -481,8 +481,6 @@ public class Table implements TableOrView, TableSchema {
      * @return the row index of the appended row.
      */
     protected long add(Object... values) {
-        long rowIndex = addEmptyRow();
-
         checkImmutable();
 
         // Check values types
@@ -512,8 +510,14 @@ public class Table implements TableOrView, TableSchema {
             }
         }
 
+        final long pkIndex = getPrimaryKey();
+        long rowIndex = (pkIndex == NO_MATCH) ? addEmptyRow() : addEmptyRowWithPrimaryKey(values[(int) pkIndex]);
+
         // Insert values
         for (long columnIndex = 0; columnIndex < columns; columnIndex++) {
+            if (columnIndex == pkIndex) {
+                continue;
+            }
             Object value = values[(int)columnIndex];
             switch (colTypes[(int)columnIndex]) {
             case BOOLEAN:
```

zaki50:
Since primary key check is removed from `addEmptyRow()`, I think it's better to update `io.realm.internal.Table#add(Object...)` like:

```
diff --git a/realm/realm-library/src/main/java/io/realm/internal/Table.java b/realm/realm-library/src/main/java/io/realm/internal/Table.java
index 5b3c1ba..4b557b7 100644
--- a/realm/realm-library/src/main/java/io/realm/internal/Table.java
+++ b/realm/realm-library/src/main/java/io/realm/internal/Table.java
@@ -481,8 +481,6 @@ public class Table implements TableOrView, TableSchema {
      * @return the row index of the appended row.
      */
     protected long add(Object... values) {
-        long rowIndex = addEmptyRow();
-
         checkImmutable();
 
         // Check values types
@@ -512,8 +510,14 @@ public class Table implements TableOrView, TableSchema {
             }
         }
 
+        final long pkIndex = getPrimaryKey();
+        long rowIndex = (pkIndex == NO_MATCH) ? addEmptyRow() : addEmptyRowWithPrimaryKey(values[(int) pkIndex]);
+
         // Insert values
         for (long columnIndex = 0; columnIndex < columns; columnIndex++) {
+            if (columnIndex == pkIndex) {
+                continue;
+            }
             Object value = values[(int)columnIndex];
             switch (colTypes[(int)columnIndex]) {
             case BOOLEAN:
```

zaki50:
:+1: other than some minor comments

realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmProxyClassGenerator.java
zaki50: If we can fix #2536, fixing #777 is very easy (just enabling the transforming in the constructor).

beeender:
@zaki50 Yes, saw it. I was thinking the method should just be removed since i cannot see why it is useful. I will put some comments to the function.

realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmProxyClassGenerator.java
beeender: I was thinking something like we do byte code insertion a special constructor there, something like `Foo(RealmDefaultConstructor c)` if user doesn't have one there. With the construction, user can set the default values.

realm/realm-library/src/main/java/io/realm/Realm.java
beeender: @zaki50 But `clazz` here should have `@Keep` right? otherwise we cannot even get the table at line 689

zaki50: Model classes don't have `@Keep` annotation.
Table name is resolved by `RealmProxyMediator#getTableName()` and it uses `Class` itself, not it's name.

zaki50:
:+1:

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 ## 2.0.0
 
+### Known issues
+
+* When create `RealmObject` from JSON stream, it will takes default values set by the non-managed `RealmObject`'s default constructor for those absent field in JSON object. This behaviour is different from other APIs when creating `RealmObject` (#3386).
```
cmelchior: When creating a `RealmObject` from a JSON stream, it will take the default values defined by its default constructor for those fields that are not defined in the JSON object. This behaviour is different from other APIs when creating `RealmObject`s

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -675,6 +687,11 @@ private Scanner getFullStringScanner(InputStream in) {
     public <E extends RealmModel> E createObject(Class<E> clazz) {
         checkIfValid();
         Table table = schema.getTable(clazz);
+        // Check and throw the exception earlier for a better exception message.
+        if (table.hasPrimaryKey()) {
+            throw new RealmException(String.format("'%s' has a primary key, use" +
+                    " 'createObject(Class<E>, Object)' to create it.", Table.tableNameToClassName(table.getName())));
```
cmelchior: to create it -> instead.

cmelchior:
Except some minor comments and if CI is happy 👍

PR #3384, opened 2016-09-02
Update core to 2.0.0-rc4

beeender:
And with some code cleanup.
Throw an runtime exception when input java bytes array cannot be read.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@
 ### Internal
 
 * Moved JNI build to CMake.
+* Updated Realm Core to 2.0.0-rc2.
 
 ## 1.2.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,7 +31,7 @@
 ### Internal
 
 * Moved JNI build to CMake.
-* Updated Realm Core to 2.0.0-rc2.
+* Updated Realm Core to 2.0.0-rc4.
 
 ## 1.2.0
 
--- a/realm/realm-library/src/androidTest/java/io/realm/SortTest.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/SortTest.java
@@ -153,19 +153,19 @@ public class SortTest {
 
         assertEquals("Adam", results.get(0).getColumnString());
         assertEquals(4, results.get(0).getColumnLong());
-        assertEquals(2, ((TableView) results.getTable()).getSourceRowIndex(0));
+        assertEquals(2, ((TableView) results.getTableOrView()).getSourceRowIndex(0));
 
         assertEquals("Adam", results.get(1).getColumnString());
         assertEquals(5, results.get(1).getColumnLong());
-        assertEquals(0, ((TableView) results.getTable()).getSourceRowIndex(1));
+        assertEquals(0, ((TableView) results.getTableOrView()).getSourceRowIndex(1));
 
         assertEquals("Adam", results.get(2).getColumnString());
         assertEquals(5, results.get(2).getColumnLong());
-        assertEquals(3, ((TableView) results.getTable()).getSourceRowIndex(2));
+        assertEquals(3, ((TableView) results.getTableOrView()).getSourceRowIndex(2));
 
         assertEquals("Brian", results.get(3).getColumnString());
         assertEquals(4, results.get(3).getColumnLong());
-        assertEquals(1, ((TableView) results.getTable()).getSourceRowIndex(3));
+        assertEquals(1, ((TableView) results.getTableOrView()).getSourceRowIndex(3));
     }
 
     private void checkSortTwoFieldsIntString(RealmResults<AllTypes> results) {
@@ -179,19 +179,19 @@ public class SortTest {
 
         assertEquals("Adam", results.get(0).getColumnString());
         assertEquals(4, results.get(0).getColumnLong());
-        assertEquals(2, ((TableView) results.getTable()).getSourceRowIndex(0));
+        assertEquals(2, ((TableView) results.getTableOrView()).getSourceRowIndex(0));
 
         assertEquals("Brian", results.get(1).getColumnString());
         assertEquals(4, results.get(1).getColumnLong());
-        assertEquals(1, ((TableView) results.getTable()).getSourceRowIndex(1));
+        assertEquals(1, ((TableView) results.getTableOrView()).getSourceRowIndex(1));
 
         assertEquals("Adam", results.get(2).getColumnString());
         assertEquals(5, results.get(2).getColumnLong());
-        assertEquals(0, ((TableView) results.getTable()).getSourceRowIndex(2));
+        assertEquals(0, ((TableView) results.getTableOrView()).getSourceRowIndex(2));
 
         assertEquals("Adam", results.get(3).getColumnString());
         assertEquals(5, results.get(3).getColumnLong());
-        assertEquals(3, ((TableView) results.getTable()).getSourceRowIndex(3));
+        assertEquals(3, ((TableView) results.getTableOrView()).getSourceRowIndex(3));
     }
 
     private void checkSortTwoFieldsIntAscendingStringDescending(RealmResults<AllTypes> results) {
@@ -205,19 +205,19 @@ public class SortTest {
 
         assertEquals("Brian", results.get(0).getColumnString());
         assertEquals(4, results.get(0).getColumnLong());
-        assertEquals(1, ((TableView) results.getTable()).getSourceRowIndex(0));
+        assertEquals(1, ((TableView) results.getTableOrView()).getSourceRowIndex(0));
 
         assertEquals("Adam", results.get(1).getColumnString());
         assertEquals(4, results.get(1).getColumnLong());
-        assertEquals(2, ((TableView) results.getTable()).getSourceRowIndex(1));
+        assertEquals(2, ((TableView) results.getTableOrView()).getSourceRowIndex(1));
 
         assertEquals("Adam", results.get(2).getColumnString());
         assertEquals(5, results.get(2).getColumnLong());
-        assertEquals(0, ((TableView) results.getTable()).getSourceRowIndex(2));
+        assertEquals(0, ((TableView) results.getTableOrView()).getSourceRowIndex(2));
 
         assertEquals("Adam", results.get(3).getColumnString());
         assertEquals(5, results.get(3).getColumnLong());
-        assertEquals(3, ((TableView) results.getTable()).getSourceRowIndex(3));
+        assertEquals(3, ((TableView) results.getTableOrView()).getSourceRowIndex(3));
     }
 
     private void checkSortTwoFieldsStringAscendingIntDescending(RealmResults<AllTypes> results) {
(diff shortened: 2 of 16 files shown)


realm/realm-library/src/main/cpp/io_realm_internal_Table.cpp
```diff
@@ -712,15 +712,12 @@ JNIEXPORT void JNICALL Java_io_realm_internal_Table_nativeSetByteArray(
     if (!TBL_AND_INDEX_AND_TYPE_VALID(env, TBL(nativeTablePtr), columnIndex, rowIndex, type_Binary))
         return;
     try {
-        if (dataArray == NULL) {
-            if (!TBL_AND_COL_NULLABLE(env, TBL(nativeTablePtr), columnIndex)) {
+        if (dataArray == NULL && (!TBL_AND_COL_NULLABLE(env, TBL(nativeTablePtr), columnIndex))) {
```
MGaetan89: Why add extra parenthesis around the second part of the condition?

beeender: Ooooops!

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -2217,19 +2217,19 @@ private void checkQueryIsNotReused() {
...
-        if (this.view != null) {
-            return view.getTargetRowIndex(rowIndex);
-        } else if (table instanceof TableView){
-            return ((TableView) table).getSourceRowIndex(rowIndex);
-        } else {
-            return rowIndex;
-        }
+        long tableRowIndex = this.query.find();
+        return tableRowIndex;
+//        if (tableRowIndex < 0) {
```
beeender: @cmelchior Do still need those?

cmelchior: No, they can go. From what I can tell the implementation is correct, but I was on a goose chase because of the bug in core revealed by one of the unit tests.

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -1697,7 +1697,7 @@ public Long call() throws Exception {
                     try {
                         sharedRealm = SharedRealm.getInstance(realmConfiguration);
 
-                        // Run the query & handover the table view for the caller thread
+                        // Run the query & handover the table linkView for the caller thread
```
cmelchior: Looks like a refactor gone wrong

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -1814,7 +1814,7 @@ public Long call() throws Exception {
 
                         long columnIndex = getColumnIndexForSort(fieldName);
 
-                        // run the query & handover the table view for the caller thread
+                        // run the query & handover the table linkView for the caller thread
```
cmelchior: same

CHANGELOG.md
```diff
@@ -31,6 +31,7 @@
 ### Internal
 
 * Moved JNI build to CMake.
+* Updated Realm Core to 2.0.0-rc2.
```
zaki50: should be `rc4`

realm/realm-library/src/main/java/io/realm/internal/TableQuery.java
```diff
@@ -420,11 +420,15 @@ public TableQuery isNotEmpty(long[] columnIndices) {
 
     // Searching methods.
 
+    @Deprecated // Doesn't seem to be used
     public long find(long fromTableRow) {
```
zaki50: I think it's OK to remove this now since it's in the internal class.

realm/realm-library/src/androidTest/java/io/realm/SortTest.java
```diff
@@ -153,19 +153,19 @@ private void checkSortTwoFieldsStringAscendingIntAscending(RealmResults<AllTypes
 
         assertEquals("Adam", results.get(0).getColumnString());
         assertEquals(4, results.get(0).getColumnLong());
-        assertEquals(2, ((TableView) results.getTable()).getSourceRowIndex(0));
+        assertEquals(2, ((TableView) results.getTableOrView()).getSourceRowIndex(0));
```
kneth: Really?

cmelchior: Meaning?

realm/realm-library/src/main/cpp/util.hpp
```diff
@@ -572,6 +574,10 @@ class JniByteArray {
         , m_arrayLength(javaArray == NULL ? 0 : env->GetArrayLength(javaArray))
         , m_array(javaArray == NULL ? NULL : env->GetByteArrayElements(javaArray, NULL))
         , m_releaseMode(JNI_ABORT) {
+        if (m_javaArray != nullptr && m_array == nullptr) {
+            // javaArray is not null but GetByteArrayElements returns null, something is really wrong.
```
nhachicha: is the data copied in the case of `GetByteArrayElements` returning null? (you can find out using the flag `idCopy`)

realm/realm-library/src/main/cpp/CMakeLists.txt
```diff
@@ -80,9 +80,11 @@ endif()
 set(WARNING_CXX_FLAGS "-Werror -Wall -Wextra -pedantic -Wno-long-long -Wno-variadic-macros \
 -Wno-missing-field-initializers -Wmissing-declarations -Wno-error=uninitialized -Wno-error=maybe-uninitialized")
 set(REALM_COMMON_CXX_FLAGS "-DREALM_ANDROID -DREALM_HAVE_CONFIG -DPIC -pthread -fvisibility=hidden -std=c++14")
-set(CMAKE_CXX_FLAGS_RELEASE "-Os -DNDEBUG -flto")
+# There might be an issue with -Os of ndk gcc 4.9. It will hang the encryption related tests.
+# And this issue doesn't seem to impact the core compiling.
+set(CMAKE_CXX_FLAGS_RELEASE "-O2 -DNDEBUG -flto")
```
kneth: `-O2` might not for ARM 64 bit if we build with clang.

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -136,26 +136,26 @@ private RealmQuery(Realm realm, Class<E> clazz) {
...
     private RealmQuery(RealmResults<E> queryResults, Class<E> clazz) {
         this.realm = queryResults.realm;
         this.clazz = clazz;
         this.schema = realm.schema.getSchemaForClass(clazz);
-        this.table = queryResults.getTable();
-        this.view = null;
-        this.query = queryResults.getTable().where();
+        this.table = queryResults.getTableOrView();
+        this.linkView = null;
+        this.query = queryResults.getTableOrView().where();
```
zaki50: Do we need to execute `queryResults.getTableOrView()`?
We should use `this.table` instead if possible.

realm/realm-library/src/main/cpp/CMakeLists.txt
nhachicha: did you tried to build al ABI with this optimisation level?

realm/realm-library/src/main/cpp/util.hpp
beeender: It is `null` only when some serious thing happened in the jvm when calling `GetByteArrayElements`, merely our fault or user's fault. Basically it is something should never happen.

realm/realm-library/src/main/cpp/CMakeLists.txt
beeender: Tried build all ABIs but not test all of them since it is quite difficult for me to get a mips device. Will test those.

beeender: @kneth Uh... shall i add a FIXME there?

realm/realm-library/src/androidTest/java/io/realm/SortTest.java
beeender: UH... i actually have the same question .... if we use `Results` from OS, the `TableView` and strange `TableOrView` are going to be removed ..

realm/realm-library/src/main/cpp/io_realm_internal_Table.cpp
```diff
@@ -712,15 +712,12 @@ JNIEXPORT void JNICALL Java_io_realm_internal_Table_nativeSetByteArray(
     if (!TBL_AND_INDEX_AND_TYPE_VALID(env, TBL(nativeTablePtr), columnIndex, rowIndex, type_Binary))
         return;
     try {
-        if (dataArray == NULL) {
-            if (!TBL_AND_COL_NULLABLE(env, TBL(nativeTablePtr), columnIndex)) {
+        if (dataArray == NULL && !TBL_AND_COL_NULLABLE(env, TBL(nativeTablePtr), columnIndex)) {
```
bmunkholm: old performance thing: We already verified TBL a couple of lines above. We are now doing it again in TBL_AND_COL_NULLABLE. Maybe just add a FIXME for now.

realm/realm-library/src/androidTest/java/io/realm/SortTest.java
cmelchior: That is even better, but since we don't do it as part of this PR 😄 . This was a rename based on the fact that `getTable()` was a very wrong name especially when hunting bugs where the distinction between an actual table and tableview was important.

zaki50:
:+1: other than some minor comments.

realm/realm-library/src/androidTest/java/io/realm/SortTest.java
cmelchior: At least I spend some time going in the wrong direction based on it being named `getTable()`, at least with `getTableOrView()` you know it could be both.

But the sooner we get rid of it all together the better.

realm/realm-library/src/main/cpp/CMakeLists.txt
nhachicha: FWIW there is this https://community.imgtec.com/developers/mips/android/ nor sure though how to build a MIPS emulator

realm/realm-library/src/main/java/io/realm/internal/TableQuery.java
```diff
@@ -420,11 +420,15 @@ public TableQuery isNotEmpty(long[] columnIndices) {
...
         validateQuery();
         return nativeFind(nativePtr, fromTableRow);
     }
 
+    /**
+     * Return the table row index for the first element matching the query.
```
dalinaum: Shoud `Return` be `Returns`?

realm/realm-library/src/main/java/io/realm/internal/TableQuery.java
beeender: Let's remove this later, love removing code! :ghost:

realm/realm-library/src/main/cpp/io_realm_internal_Table.cpp
beeender: Added FIXME in `util.hpp`

realm/realm-library/src/main/cpp/CMakeLists.txt
beeender: Tested all ABIs except mips.

PR #3387, opened 2016-09-02
Mc/set unique

beeender:


beeender:
This needs #3384 #3379 be merged first.

beeender:
@realm/java Ready for review.

CHANGELOG.md
```diff
@@ -32,6 +32,7 @@
 
 * Moved JNI build to CMake.
 * Updated Realm Core to 2.0.0-rc4.
+* Call `set_int_unique()` and `set_string_unique()` for primary keys.
```
cmelchior: Not sure we need this in the changelog. It has zero impact on end users?

realm/realm-library/src/main/cpp/io_realm_internal_Table.cpp
```diff
@@ -684,6 +695,24 @@ JNIEXPORT void JNICALL Java_io_realm_internal_Table_nativeSetString(
...
+Java_io_realm_internal_Table_nativeSetStringUnique(JNIEnv *env, jclass, jlong nativeTablePtr, jlong columnIndex,
+                                                   jlong rowIndex, jstring value)
+{
+    if (!TBL_AND_INDEX_AND_TYPE_VALID(env, TBL(nativeTablePtr), columnIndex, rowIndex, type_String))
+        return;
+    try {
+        if (value == NULL) {
+            if (!TBL_AND_COL_NULLABLE(env, TBL(nativeTablePtr), columnIndex)) {
+                return;
+            }
```
cmelchior: We really need to agree on style guides for JNI. If statements are done in two different ways within 6 lines 😄

realm/realm-library/src/main/java/io/realm/internal/Table.java
```diff
@@ -385,9 +385,30 @@ public long addEmptyRow() {
...
+
+    /**
+     * Add an empty row to the table and set the primary key with the given value.
+     *
+     * @param primaryKeyValue the primary key value.
+     * @param skipChecking set to {@code true} to skip all validations. This is currently used by bulk insert which
+     *                     has its own validations.
+     * @return the row index.
+     */
+    public long addEmptyRowWithPrimaryKey(Object primaryKeyValue, boolean skipChecking) {
```
cmelchior: I would much prefer "positive" boolean instead of negatives, so this should be `validateValue` or something like that

realm/realm-library/src/main/java/io/realm/internal/Table.java
```diff
@@ -399,11 +420,12 @@ public long addEmptyRowWithPrimaryKey(Object primaryKeyValue) {
             switch (type) {
                 case STRING:
                 case INTEGER:
-                    if (findFirstNull(primaryKeyColumnIndex) != NO_MATCH) {
+                    if (!skipChecking && findFirstNull(primaryKeyColumnIndex) != NO_MATCH) {
                         throwDuplicatePrimaryKeyException("null");
                     }
                     rowIndex = nativeAddEmptyRow(nativePtr, 1);
                     row = getUncheckedRow(rowIndex);
+                    // FIXME: Use core's set_null_unique when core supports it.
```
cmelchior: Isn't this a big breaking change. It basically means we no longer support null primary keys?

beeender: No, we still support `null` for PK. Just not calling the right method from core. But for now, it is not a problem at all since we guarantee the uniqueness of PK in binding level

CHANGELOG.md
beeender: Loving removal things, will do

cmelchior:
Expect a minor point about reversing the boolean and the changelog. This looks good.

realm/realm-library/src/main/cpp/io_realm_internal_Table.cpp
beeender: Will change this when the `set_null_unique` PR.

kneth: The flow would be much nicer if the validation methods like `TBL_AND_INDEX_AND_TYPE_VALID` throw C++ exceptions instead.

kneth:
LGTM

cmelchior:
👍

PR #3390, opened 2016-09-03
fix Javadoc documentation about encryption key length (64 bit -> 64 bytes)

### LINKED ISSUE #3389: Incorrect javadoc
Javadoc for method <code>public Builder encryptionKey(byte[] key)</code> in <code>RealmConfiguration.Builder</code> states:

/* Sets the 64 bit key used to encrypt and decrypt the Realm file. */

That means, that client should provide 8-byte key, but actually 64-bytes key is required, otherwise exception is thrown:

java.lang.IllegalArgumentException: The provided key must be 64 bytes. Yours was: 8

zaki50:
fixes #3389

### DIFF
--- a/realm/realm-library/src/main/java/io/realm/RealmConfiguration.java
+++ b/realm/realm-library/src/main/java/io/realm/RealmConfiguration.java
@@ -443,7 +443,7 @@ public final class RealmConfiguration {
         }
 
         /**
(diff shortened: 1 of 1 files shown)


zaki50:
@realm/java

cmelchior:
:+1:

beeender:
:+1:

PR #3392, opened 2016-09-04
transform accesses to other model fields in model's constructors.

### LINKED ISSUE #3361: [SO] Realm-Transformer sometimes doesn't work
Ported from [Stack Overflow](http://stackoverflow.com/q/39205297/2413303)

#### Goal

#### Expected Results

When giving managed RealmObject to constructor, the value should be available through field access

#### Actual Results

The managed RealmObject's field access returns `null`

However, the managed RealmObject's getter/setter returns the proper value

(In log, not in debugger)

#### Code Sample

``` java
 realm = Realm.getDefaultInstance();
      @Override
            Stock stock1 = new Stock();
            stock1.id = "stock1";
            managedStock = realm.copyToRealm(stock1);
      }
});
//output here: managedStock.id = stock1
Log.d("myLogs", "managedStock.id = " + managedStock.id);
Goods goods = new Goods(managedStock);
```

and

``` java
public class Goods extends RealmObject {
    @PrimaryKey
    public String id;
    public Stock stock;

    public Goods() {}

    public Goods(Stock s) {
        //output here: s.id == null
        Log.d("myLogs", "s.id == " + s.id);
    }
}
```

and

``` java
public class Stock extends RealmObject {
    @PrimaryKey
    public String id;
    public String name;
}
```

Realm version(s): 1.2.0

Android Studio version: Android Studio 2.1.2

Which Android version and device: Android 5.0, Samsung Galaxy S4

zaki50:
fixes #3361 

@realm/java

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 * Fixed a lint error in proxy classes when the 'minSdkVersion' of user's project is smaller than 11 (#3356).
 * Fixed a potential crash when there were lots of async queries waiting in the queue.
+* Fixed a bug that Realm transformer does not transform any field accesses in model's constructors (#3361).
 
 ## 1.2.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 
 * Fixed a lint error in proxy classes when the 'minSdkVersion' of user's project is smaller than 11 (#3356).
 * Fixed a potential crash when there were lots of async queries waiting in the queue.
-* Fixed a bug that Realm transformer does not transform any field accesses in model's constructors (#3361).
+* Fixed a bug causing the Realm Transformer to not transform field access in the model's constructors (#3361).
 
 ## 1.2.0
 
--- a/realm-transformer/src/test/groovy/io/realm/transformer/BytecodeModifierTest.groovy
+++ b/realm-transformer/src/test/groovy/io/realm/transformer/BytecodeModifierTest.groovy
@@ -79,7 +79,7 @@ class BytecodeModifierTest extends Specification {
     def "UseRealmAccessors"() {
         setup: 'generate an empty class'
         def classPool = ClassPool.getDefault()
-        def ctClass = classPool.makeClass('testClass')
+        def ctClass = classPool.makeClass('TestClass')
 
         and: 'add a field'
         def ctField = new CtField(CtClass.intType, 'age', ctClass)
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 
 * Fixed a lint error in proxy classes when the 'minSdkVersion' of user's project is smaller than 11 (#3356).
 * Fixed a potential crash when there were lots of async queries waiting in the queue.
+* Fixed a bug that Realm transformer does not transform any field accesses in model's constructors (#3361).
```
cmelchior: Fixed a bug causing the Realm Transformer to not transform field access in the model's constructors

realm-transformer/src/test/groovy/io/realm/transformer/BytecodeModifierTest.groovy
```diff
@@ -98,18 +98,22 @@ class BytecodeModifierTest extends Specification {
...
+        def methodCalled = false
         def fieldIsUsed = false
         for (CodeIterator ci = codeAttribute.iterator(); ci.hasNext();) {
             int index = ci.next();
             int op = ci.byteAt(index);
             if (op == Opcode.GETFIELD) {
                 fieldIsUsed = true
+            } else if (op == Opcode.INVOKEVIRTUAL) {
+                // access to the getter
+                methodCalled = true
```
cmelchior: Why is this needed?

cmelchior:
Nice 👍 . Only minor comments and could use a little more comments in the code, but otherwise looks good.

realm-transformer/src/test/groovy/io/realm/transformer/BytecodeModifierTest.groovy
zaki50: We replace a field access with method call.
I think it's better to directly check whether there is a method call (`INVOKEVIRTUAL`) or not.

Another reason is to align all tests below those all have a check for method call.

zaki50: I think it's better to replace line 103-113 (for loop and check) with `!isFieldRead(ctMethod) && hasMethodCall(ctMethod)`.

zaki50: updated

emanuelez:
I remember implementing the transformation for the constructors as well but having to revert it for some problem it was causing. I really wish I could remember what such problem was :(

zaki50:
@emanuelez Was that NPE for field assignment in constructor?
Simple implementation causes a NPE for following model class.

```
    private String name = "unknown";
}
```

I added some check to avoid this NPE.

emanuelez:
Maybe it was that :)

beeender:
:+1:

cmelchior:
👍

PR #3397, opened 2016-09-05
Allow to specify default value of the field in model's constructor

### LINKED ISSUE #777: Default values for fields / assignment in default constructor

### LINKED ISSUE #2536: Calling setter method in no argument constructor results in NPE

zaki50:
This change allows to specify default value of the field.
And this change also enables to call setter methods in model's constructor.

fixes #777 
fixes #2536 

TODO

* [x] add tests 
* [x] update CHANGELING.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@
 * Importing from JSON without the primary key field defined in the JSON object now throws `IllegalArgumentException`.
 * Now `Realm.beginTransaction()`, `Realm.executeTransaction()` and `Realm.waitForChange()` throw `RealmMigrationNeededException` if a remote process introduces incompatible schema changes (#3409).
 * The primary key value of an object can no longer be changed after the object was created. Instead a new object must be created and all fields copied over.
-* Now `Realm.createObject(Class)` and `Realm.createObject(Class,Object)` take the values from the model's fields and default constructor (#777).
+* Now `Realm.createObject(Class)` and `Realm.createObject(Class,Object)` take the values from the model's fields and default constructor. `DynamicRealm` does not take these default values (#777).
 * When `Realm.create*FromJson()`s create a new `RealmObject`, now they take the default values defined by the field itself and its default constructor for those fields that are not defined in the JSON object.
 
 ### Enhancements
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,5 @@
 ## 2.0.0
 
-### Known issues
-
-* When creating a `RealmObject` from a JSON stream, it will take the default values defined by its default constructor for those fields that are not defined in the JSON object. This behaviour is different from other APIs when creating `RealmObject`s.
-
 ### Breaking Changes
 
 * `isValid()` now always returns `true` instead of `false` for unmanaged `RealmObject` and `RealmList`. This puts it in line with the behaviour of the Cocoa and .NET API's (#3101).
@@ -17,6 +13,8 @@
 * Importing from JSON without the primary key field defined in the JSON object now throws `IllegalArgumentException`.
 * Now `Realm.beginTransaction()`, `Realm.executeTransaction()` and `Realm.waitForChange()` throw `RealmMigrationNeededException` if a remote process introduces incompatible schema changes (#3409).
 * The primary key value of an object can no longer be changed after the object was created. Instead a new object must be created and all fields copied over.
+* Now `Realm.createObject(Class)` and `Realm.createObject(Class,Object)` take the values from the model's constructor (#777).
+* When `Realm.create*FromJson()`s create a new `RealmObject`, now they take the default values defined by its default constructor for those fields that are not defined in the JSON object.
 
 ### Enhancements
 
@@ -30,6 +28,7 @@
 * Fixed a lint error in proxy classes when the 'minSdkVersion' of user's project is smaller than 11 (#3356).
 * Fixed a potential crash when there were lots of async queries waiting in the queue.
 * Fixed a bug causing the Realm Transformer to not transform field access in the model's constructors (#3361).
+* Fixed a bug causing the `NullPointerException` when calling getters/setters in the model's constructors (#2536).
 
 ### Internal
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,8 +13,8 @@
 * Importing from JSON without the primary key field defined in the JSON object now throws `IllegalArgumentException`.
 * Now `Realm.beginTransaction()`, `Realm.executeTransaction()` and `Realm.waitForChange()` throw `RealmMigrationNeededException` if a remote process introduces incompatible schema changes (#3409).
 * The primary key value of an object can no longer be changed after the object was created. Instead a new object must be created and all fields copied over.
-* Now `Realm.createObject(Class)` and `Realm.createObject(Class,Object)` take the values from the model's constructor (#777).
-* When `Realm.create*FromJson()`s create a new `RealmObject`, now they take the default values defined by its default constructor for those fields that are not defined in the JSON object.
+* Now `Realm.createObject(Class)` and `Realm.createObject(Class,Object)` take the values from the model's fields and default constructor (#777).
+* When `Realm.create*FromJson()`s create a new `RealmObject`, now they take the default values defined by the field itself and its default constructor for those fields that are not defined in the JSON object.
 
 ### Enhancements
 
--- a/realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmProxyClassGenerator.java
+++ b/realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmProxyClassGenerator.java
@@ -725,7 +725,7 @@ public class RealmProxyClassGenerator {
                 .emitStatement("return object")
             .endControlFlow();
 
-        writer.emitStatement("final BaseRealm.RealmObjectContext objectContext = BaseRealm.objectContext.get();");
+        writer.emitStatement("final BaseRealm.RealmObjectContext objectContext = BaseRealm.objectContext.get()");
 
         writer.emitStatement("RealmObjectProxy cachedRealmObject = cache.get(object)");
         writer.beginControlFlow("if (cachedRealmObject != null)")
@@ -777,7 +777,7 @@ public class RealmProxyClassGenerator {
                             .emitStatement("realmObject = new %s()", qualifiedGeneratedClassName)
                             .emitStatement("cache.put(object, (RealmObjectProxy) realmObject)")
                         .nextControlFlow("finally")
-                            .emitStatement("objectContext.clear();")
+                            .emitStatement("objectContext.clear()")
                         .endControlFlow()
 
                     .nextControlFlow("else")
@@ -1579,7 +1579,7 @@ public class RealmProxyClassGenerator {
         writer.emitEmptyLine();
         writer.emitStatement("String path = proxyState.getRealm$realm().getPath()");
         writer.emitStatement("String otherPath = %s.proxyState.getRealm$realm().getPath()", otherObjectVarName);
-        writer.emitStatement("if (path != null ? !path.equals(otherPath) : otherPath != null) return false;");
+        writer.emitStatement("if (path != null ? !path.equals(otherPath) : otherPath != null) return false");
         writer.emitEmptyLine();
         writer.emitStatement("String tableName = proxyState.getRow$realm().getTable().getName()");
         writer.emitStatement("String otherTableName = %s.proxyState.getRow$realm().getTable().getName()", otherObjectVarName);
@@ -1629,12 +1629,12 @@ public class RealmProxyClassGenerator {
             }
             writer
                     .beginControlFlow("if (rowIndex != TableOrView.NO_MATCH)")
-                        .emitStatement("final BaseRealm.RealmObjectContext objectContext = BaseRealm.objectContext.get();")
+                        .emitStatement("final BaseRealm.RealmObjectContext objectContext = BaseRealm.objectContext.get()")
                         .beginControlFlow("try")
                             .emitStatement("objectContext.set(realm, table.getUncheckedRow(rowIndex), realm.schema.getColumnInfo(%s.class), false)", qualifiedClassName)
                             .emitStatement("obj = new %s()", qualifiedGeneratedClassName)
                         .nextControlFlow("finally")
-                            .emitStatement("objectContext.clear();")
+                            .emitStatement("objectContext.clear()")
                         .endControlFlow()
                     .endControlFlow()
                 .endControlFlow();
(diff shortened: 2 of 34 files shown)


realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmProxyClassGenerator.java
```diff
@@ -334,6 +343,44 @@ private void emitAccessors(JavaWriter writer) throws IOException {
...
+        {
+            writer.emitSingleLineComment("Called from model's constructor. Inject context.");
+            writer.emitStatement("injectObjectContext()");
+        }
+        writer.endControlFlow();
+        writer.emitEmptyLine();
+
+        if (isSetter) {
+            writer.beginControlFlow(
+                    "if (proxyState.isUnderConstruction() && !proxyState.getRealm$realm().isInTransaction())")
```
cmelchior: `isInTransaction` is a pretty expensive call. I am not sure it will be acceptable to call that for all accessors.

cmelchior: Also it calls `checkIfValid()` which means we are calling that method twice

zaki50: Right. and checking by `isInTransaction()` does not work as expected in some cases.
I'll remove the call of `isInTransaction()` and introduce a boolean field instead.

zaki50:
I created this PR against `releases` branch, but now I think it should be go to `master` since it includes an enhancement for `Realm#createObject()`.

beeender:
Anyway #777 is a new feature (also a breaking change), so it should go `master` IMO.

zaki50:
@beeender Right.

zaki50:
now all existing tests are green

zaki50:
retest this please

zaki50:
seems to need some more fixes 😢

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -684,10 +684,16 @@ private Scanner getFullStringScanner(InputStream in) {
 
     /**
      * Instantiates and adds a new object to the Realm.
+     * <p>
+     * This method is only available for the model classes that have no primary key defined
+     * even though the class provides a default value for it.
```
cmelchior: This method is only available for model classes with no `@PrimaryKey` annotation. If you like to create an object that has a primary key, use `createObject(Class, Object)` or `copyToRealm(RealmModel)` instead.

zaki50: thanks, done.

zaki50:
I rebased this against master.

zaki50:
one test fails.

I'd like to remove that test, but I'll try a bit more.

zaki50:
Now ready for review @realm/java

realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmProxyClassGenerator.java
```diff
@@ -371,6 +380,83 @@ private void emitAccessors(JavaWriter writer) throws IOException {
...
+                                field.getSimpleName().toString())
+                                .emitStatement("return")
+                                .endControlFlow();
+                        final String modelFqcn = Utils.getGenericTypeQualifiedName(field);
+                        writer.beginControlFlow("if (value != null && !value.isManaged())")
+                                .emitStatement("final Realm realm = (Realm) proxyState.getRealm$realm()")
+                                .emitStatement("final RealmList<%1$s> original = value", modelFqcn)
+                                .emitStatement("value = new RealmList<%1$s>()", modelFqcn)
+                                .beginControlFlow("for (%1$s item : original)", modelFqcn)
+                                    .beginControlFlow("if (item == null || RealmObject.isManaged(item))")
```
kneth: Is `RealmObject.isManaged(null)` valid?

kneth: Never mind. If `item == null`, that expression isn't evaluated.

kneth:
What is the performance impact of adding an extra layer?

CHANGELOG.md
```diff
@@ -17,6 +13,8 @@
 * Importing from JSON without the primary key field defined in the JSON object now throws `IllegalArgumentException`.
 * Now `Realm.beginTransaction()`, `Realm.executeTransaction()` and `Realm.waitForChange()` throw `RealmMigrationNeededException` if a remote process introduces incompatible schema changes (#3409).
 * The primary key value of an object can no longer be changed after the object was created. Instead a new object must be created and all fields copied over.
+* Now `Realm.createObject(Class)` and `Realm.createObject(Class,Object)` take the values from the model's constructor (#777).
```
cmelchior: model's default constructor.

cmelchior: Also what about if you assign a value directly on the field? I _think_ in bytecode that is included first in the default constructor,  but I might remember wrong. In any case we should mention it either way.

realm/realm-annotations-processor/src/test/resources/io/realm/AllTypesRealmProxy.java
```diff
@@ -106,73 +106,180 @@ public final AllTypesColumnInfo clone() {
...
         this.proxyState = new ProxyState(some.test.AllTypes.class, this);
+        proxyState.setRealm$realm(context.getRealm());
+        proxyState.setRow$realm(context.getRow());
+        proxyState.setAcceptDefaultValue$realm(context.getAcceptDefaultValue());
+        proxyState.setExcludeFields$realm(context.getExcludeFields());
     }
 
     @SuppressWarnings("cast")
     public String realmGet$columnString() {
+        if (proxyState == null) {
```
cmelchior: This check is also in the constructor. I assume the reason we need it here is because the super constructor might call these methods before we reach the similar statement in the proxy constructor?

realm/realm-annotations-processor/src/test/resources/io/realm/AllTypesRealmProxy.java
```diff
@@ -195,14 +318,36 @@ public final AllTypesColumnInfo clone() {
...
+            // Called from model's constructor. Inject context.
+            injectObjectContext();
+        }
+
+        if (proxyState.isUnderConstruction()) {
+            if (!proxyState.getAcceptDefaultValue$realm()) {
+                return;
+            }
+            if (proxyState.getExcludeFields$realm().contains("columnObject")) {
```
cmelchior: This looks like a potentially very expensive check, but it is guarded by a lot cheaper boolean checks. So guess it is ok.

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -684,28 +685,50 @@ private Scanner getFullStringScanner(InputStream in) {
...
     /**
      * Instantiates and adds a new object to the Realm.
+     * <p>
+     * This method is only available for model classes with no @PrimaryKey annotation.
+     * If you like to create an object that has a primary key, use {@link #createObject(Class, Object)}
+     * or {@link #copyToRealm(RealmModel)} instead.
      *
      * @param clazz the Class of the object to create.
      * @return the new object.
      * @throws RealmException if an object cannot be created.
```
cmelchior: Also document what it throws if you use a Model class with a primary key

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -717,9 +740,28 @@ private Scanner getFullStringScanner(InputStream in) {
...
+    /**
+     * Same as {@link #createObject(Class, Object)} but this does not check the thread.
+     *
+     * @param clazz the Class of the object to create.
+     * @param primaryKeyValue value for the primary key field.
+     * @param acceptDefaultValue if {@code true}, default value of the object will be applied and
+     *                           if {@code false}, it will be ignored.
+     * @return the new object.
+     * @throws RealmException if object could not be created due to the primary key being invalid.
+     * @throws IllegalStateException if the model clazz does not have an primary key defined.
```
cmelchior: These throw statements should also be on the public method instead of our internal one :)

CHANGELOG.md
cmelchior: Also document that this does not apply to DynamicRealmObjects.

zaki50: You are right.

From the view of the bytecode, 
```
public class Foo {
    private int bar = 1;
}
```
and
```
public class Foo {
    private int bar;
    public Foo() {
      bar = 1;
    }
}
```
are the same.

I've updated the CHANGELING.md

cmelchior:
Generally looks 👍 
A few general comments. I could really use a high-level description of how this works. We already have some high-level docs on the `RealmProcessor`. Perhaps add a section there?

Also we should really benchmark this. Generally it looks like we are adding about 3 boolean checks pr. accessor, which will probably be dwarfed by any JNI access, but just to be on the safe side. There should already be micro-benchmarks for testing accessors.

zaki50:
@kneth @cmelchior I'll benchmark the impact of this this change.

realm/realm-annotations-processor/src/test/resources/io/realm/AllTypesRealmProxy.java
zaki50: correct.

realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmProxyClassGenerator.java
```diff
@@ -371,6 +380,83 @@ private void emitAccessors(JavaWriter writer) throws IOException {
...
+        writer.beginControlFlow("if (proxyState == null)");
+        {
+            writer.emitSingleLineComment("Called from model's constructor. Inject context.");
+            writer.emitStatement("injectObjectContext()");
+        }
+        writer.endControlFlow();
+        writer.emitEmptyLine();
+
+        if (isSetter) {
+            writer.beginControlFlow(
```
zaki50: fixed

realm/realm-library/src/main/java/io/realm/Realm.java
zaki50: done

realm/realm-library/src/main/java/io/realm/Realm.java
zaki50: done

zaki50:
TODO: adding a high-level description of how this works in `RealmProcessor`

zaki50:
@cmelchior Added a description in 72cba0b8b75ee179e7ef390a89d78f5fbdb84781

CHANGELOG.md
```diff
@@ -13,8 +13,8 @@
 * Importing from JSON without the primary key field defined in the JSON object now throws `IllegalArgumentException`.
 * Now `Realm.beginTransaction()`, `Realm.executeTransaction()` and `Realm.waitForChange()` throw `RealmMigrationNeededException` if a remote process introduces incompatible schema changes (#3409).
 * The primary key value of an object can no longer be changed after the object was created. Instead a new object must be created and all fields copied over.
-* Now `Realm.createObject(Class)` and `Realm.createObject(Class,Object)` take the values from the model's constructor (#777).
-* When `Realm.create*FromJson()`s create a new `RealmObject`, now they take the default values defined by its default constructor for those fields that are not defined in the JSON object.
+* Now `Realm.createObject(Class)` and `Realm.createObject(Class,Object)` take the values from the model's fields and default constructor. `DynamicRealm` does not take these default values (#777).
```
cmelchior: Creating objects through the `DynamicRealm` does not use these values.

cmelchior:
Looks good to me. Just need some benchmark results to verify this doesn't cause major regressions in performance. Otherwise big 👍

zaki50:
Benchmark results of `1.2.0`, `master (3cbc084b3c888e9555629638e31adebbfd369c8f)` and `PR3397`. I measured 3 times each.

```
Benchmark benchmark = new Benchmark() {
    private Realm realm;
    private int i;

    @Override
    public void setUp() {
        realm = Realm.getInstance(realmConfiguration);
        i = 0;
    }

    @Override
    public void tearDown() {
        realm.close();
    }

    @Override
    protected void prepareRun() {
        realm.beginTransaction();
    }

    @Override
    public void run() {
        realm.createObject(Employee.class);
    }

    @Override
    protected void cleanupRun() {
        realm.cancelTransaction();
    }
};
```

On Nexus 6P (airplane mode)

nexus_6p


On genymotion (airplane mode)

genymotion

And I measured 100 times `createObject()` as well.

```
Benchmark benchmark = new Benchmark() {
    private Realm realm;
    private int i;

    @Override
    public void setUp() {
        realm = Realm.getInstance(realmConfiguration);
        i = 0;
    }

    @Override
    public void tearDown() {
        realm.close();
    }

    @Override
    protected void prepareRun() {
        realm.beginTransaction();
    }

    @Override
    public void run() {
        realm.createObject(Employee.class);
        realm.createObject(Employee.class);
        realm.createObject(Employee.class);
        // 100 times
    }

    @Override
    protected void cleanupRun() {
        realm.cancelTransaction();
    }
};
```

On genymotion (100 times `createObject()`)  (airplane mode)

genymotion-100

zaki50:
I can't see any big difference in one `createObject()`. And in 100 times `createObjects()`, PR3397 is about 10% slower than before.

cmelchior:
Yeah, and that is probably within the uncertainty of the underlying hardware anyway. Looks fine to me :shipit:

zaki50:
I also measured a time of creating `RealmObject` from `RealmResults`.

```
Benchmark benchmark = new Benchmark() {
    private Realm realm;
    private RealmResults<Employee> result;

    @Override
    public void setUp() {
        realm = Realm.getInstance(realmConfiguration);
        addObjects(realm);
        verify(realm);
        result = realm.where(Employee.class).findAll();
    }

    @Override
    public void tearDown() {
        deleteObjects(realm);
        realm.close();
    }

    @Override
    public void run() {
        final Iterator<Employee> iterator = result.iterator();
        while (iterator.hasNext()) {
            iterator.next();
        }
    }
```

The result is 10-15% slower than `master`.

I have an idea to improve this. So I'll create an another issue to do that later.

PR #3403, opened 2016-09-05
List<String> helper method for RealmQuery.In

ahmadalibaloch:
We always need to convert a `List<String>` into `String[]` to use RealmQuery.In method. This method will be handy all the time.

### DIFF
--- a/realm/realm-library/src/main/java/io/realm/RealmQuery.java
+++ b/realm/realm-library/src/main/java/io/realm/RealmQuery.java
@@ -425,6 +425,21 @@ public final class RealmQuery<E extends RealmModel> {
     // In
 
     /**
+     * In comparison. This allows you to test if objects match any value in a list of values.
+     *
+     * @param fieldName the field to compare.
+     * @param values list of values to compare with and it cannot be null or empty.
+     * @return the query object.
+     * @throws java.lang.IllegalArgumentException if the field isn't a String field or {@code values} is {@code null} or empty.
+     */
+    public RealmQuery<E> in(String fieldName, List<String> values) {
+        if(values == null || values.size() < 1) {
+            throw new IllegalArgumentException(EMPTY_VALUES);
+        }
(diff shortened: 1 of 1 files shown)


Zhuinden:
I'm a bit iffy about how it's only for `String`, although technically it could work for `List<T>`.

``` java
    public RealmQuery<E> in(String fieldName, Byte[] values) {
        if (values == null || values.length == 0) {
            throw new IllegalArgumentException(EMPTY_VALUES);
        }
        beginGroup().equalTo(fieldName, values[0]);
        for (int i = 1; i < values.length; i++) {
            or().equalTo(fieldName, values[i]);
        }
        return endGroup();
    }
```

I do wonder why it's only for `[]` and not for `List<T>` though. It would work pretty much the same way, and I wouldn't even convert the list to an array for that either.

Just my take, I'm not a Realm person :tongue:

ahmadalibaloch:
You are right. There should be methods for every type or a generic one. I was wondering if the generic type checking and then translation will be slow with Relfection so we need to create a method for List<Integer>, List<Byte> and so on?

zaki50:
We couldn't define `in(String fieldName, List<String> values)` and `in(String fieldName, List<Integer> values)` in the same class since those have the same erasure.

And we could not get actual `T` in the body of `in(String fieldName, List<T> values)`. Of course we could get the type of each value in the `values`, but it requires additional type checking.

Zhuinden:
Those list of primitives foiled it again!

zaki50:
In order to keep consistent API, I'd like to avoid providing only `in(String fieldName, List<String> values)`.
I think we need to think more to provide `in()` that accepts `List`.

So how about creating an issue for additional `in()` API and close this PR for now?

cmelchior:
Yes, that sounds fine. We need to consider how this will impact incoming features as well.

zaki50:
I've created an issue #3410

I'll close this PR and let's discuss on that issue.

PR #3409, opened 2016-09-07
Invalidate schema cache when the schema version of Realm is changed by other process.

zaki50:
This PR is a part of multi-process support.

Now schema cache referred by typed Realm instance is not shared and global schema cache is introduced instead in order to avoid validating schema multiple times for the same version.

This enables to change the column index of the defined fields during Realm instances are open.


zaki50:
Now ready for review @realm/java

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -1412,6 +1417,45 @@ Table getTable(Class<? extends RealmModel> clazz) {
...
+            final Set<Class<? extends RealmModel>> modelClasses = mediator.getModelClasses();
+            final Map<Class<? extends RealmModel>, ColumnInfo> map;
+            map = new HashMap<Class<? extends RealmModel>, ColumnInfo>(modelClasses.size());
+            try {
+                for (Class<? extends RealmModel> clazz : modelClasses) {
+                    final ColumnInfo columnInfo = mediator.validateTable(clazz, sharedRealm);
+                    map.put(clazz, columnInfo);
+                }
+            } catch (RealmMigrationNeededException e) {
+                // TODO incompatible schema change. what to do??
```
bmunkholm: ?? missing TODO implementation

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -74,7 +74,8 @@ static RealmCacheType valueOf(Class<? extends BaseRealm> clazz) {
 
     // Column indices are cached to speed up opening typed Realm. If a Realm instance is created in one thread, creating
     // Realm instances in other threads doesn't have to initialize the column indices again.
-    private ColumnIndices typedColumnIndices;
+    private static final int MAX_ENTRIES_IN_TYPED_COLUMN_INDICES_ARRAY = 4;
```
bmunkholm: why 4?

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@
 * Added `realmObject.isManaged()`, `RealmObject.isManaged(obj)` and `RealmCollection.isManaged()` (#3101).
 * Added `RealmConfiguration.Builder.directory(File)`.
 * `RealmLog` has been moved to the public API. It is now possible to control which events Realm emit to Logcat. See the `RealmLog` class for more details.
+* Now typed `RealmObject`s can access its fields properly if the column index was changed during the relevant Realm is open (#3409).
```
kneth: "during" -> "while" ???

realm/realm-library/src/androidTest/java/io/realm/ColumnIndicesTests.java
```diff
@@ -0,0 +1,119 @@
...
+    @Before
+    public void setUp() {
+        RealmConfiguration config = configFactory.createConfiguration();
+        realm = Realm.getInstance(config);
+        mediator = config.getSchemaMediator();
+    }
+
+    @After
+    public void tearDown() {
+        realm.close();
```
kneth: `if (realm != null)` maybe - just to be safe.

CHANGELOG.md
cmelchior: IMO this sounds like an internal change. For users it should perhaps be rephrased to 

Typed `RealmObject`s can now continue to access their fields properly even though the schema was changed while the Realm was open.

realm/realm-annotations-processor/src/test/resources/io/realm/AllTypesRealmProxy.java
```diff
@@ -31,49 +30,80 @@
...
             indicesMap.put("columnBinary", this.columnBinaryIndex);
-
             this.columnObjectIndex = getValidColumnIndex(path, table, "AllTypes", "columnObject");
             indicesMap.put("columnObject", this.columnObjectIndex);
-
             this.columnRealmListIndex = getValidColumnIndex(path, table, "AllTypes", "columnRealmList");
             indicesMap.put("columnRealmList", this.columnRealmListIndex);
 
-            setIndicesMap(indicesMap);
+            setIndicesMap(Collections.unmodifiableMap(indicesMap));
```
cmelchior: Any reason to make it unmodifiable? Isn't that just extra overhead we don't really need?

realm/realm-annotations-processor/src/test/resources/io/realm/AllTypesRealmProxy.java
```diff
@@ -31,49 +30,80 @@
...
+            this.columnDateIndex = original.columnDateIndex;
+            this.columnBinaryIndex = original.columnBinaryIndex;
+            this.columnObjectIndex = original.columnObjectIndex;
+            this.columnRealmListIndex = original.columnRealmListIndex;
+
+            setIndicesMap(original.getIndicesMap());
+        }
+
+        @Override
+        public final void copyFrom(ColumnInfo other) {
```
cmelchior: I would probably rename this to `copyColumnInfoFrom(other)`. `copyFrom` makes me think it will copy the entire RealmObject.

realm/realm-annotations-processor/src/test/resources/io/realm/AllTypesRealmProxy.java
```diff
@@ -31,49 +30,80 @@
...
+            if (!(other instanceof AllTypesColumnInfo)) {
```
cmelchior: Is this check nessassery? I assume we are the only ones calling this?

realm/realm-annotations-processor/src/test/resources/io/realm/AllTypesRealmProxy.java
```diff
@@ -31,49 +30,80 @@
...
+            this.columnDateIndex = otherInfo.columnDateIndex;
+            this.columnBinaryIndex = otherInfo.columnBinaryIndex;
+            this.columnObjectIndex = otherInfo.columnObjectIndex;
+            this.columnRealmListIndex = otherInfo.columnRealmListIndex;
+
+            setIndicesMap(otherInfo.getIndicesMap());
+        }
+
+        @Override
+        public final AllTypesColumnInfo copy() {
```
cmelchior: `copyColumnInfo` ?

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -342,4 +370,49 @@ private static void copyAssetFileIfNeeded(RealmConfiguration configuration) {
...
+                return candidate;
+            }
+        }
+        return null;
+    }
+
+    /**
+     * Stores the schema cache to the array.
+     * <p>
+     * If the {@code array} has empty ({@code == null}) slot, this method stores
```
kneth: "has" -> "is"

realm/realm-library/src/main/java/io/realm/Realm.java
cmelchior: From what I can tell, this will be thrown if e.g too many fields where deleted. The cache will allow fields to be re-arranged and added, but deleting them will break the typed binding (which makes sense).

Just re-throwing the RealmMigrationException is okay I guess. I don't know what else we can do?
- Throw `IllegalStateException`, but not sure that contains more meaning
- Throw an actual Error, but not sure that is better either.

realm/realm-library/src/main/java/io/realm/internal/ColumnInfo.java
```diff
@@ -34,11 +33,27 @@ protected final long getValidColumnIndex(String realmPath, Table table,
...
     }
+
+    protected final void setIndicesMap(Map<String, Long> indicesMap) {
+        this.indicesMap = indicesMap;
+    }
+
+    /**
+     * Copies the column index value from other {@link ColumnInfo} object.
+     *
+     * @param other The class of {@code other} must be exactly the the same with this instance.
```
kneth: "the the" -> "the"
"with" -> "as"

realm/realm-library/src/main/java/io/realm/internal/ColumnInfo.java
```diff
@@ -34,11 +33,27 @@ protected final long getValidColumnIndex(String realmPath, Table table,
...
+     *              Of course it must not be {@code null}.
```
kneth: "Of course it" -> "It"

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -274,6 +277,31 @@ static synchronized void invokeWithGlobalRefCount(RealmConfiguration configurati
...
+        }
+        final RefAndCount refAndCount = cache.refAndCountMap.get(RealmCacheType.TYPED_REALM);
+        if (refAndCount.localRealm.get() == null) {
+            // Called during initialization. just skip it.
+            // We can reach here if the DynamicRealm instance is initialized first.
+            return;
+        }
+        final ColumnIndices[] globalCacheArray = cache.typedColumnIndicesArray;
+        final ColumnIndices createdCacheEntry = realm.updateSchemaCache(
+                globalCacheArray);
```
cmelchior: Wrong wrap?

realm/realm-library/src/main/java/io/realm/internal/SharedRealm.java
```diff
@@ -161,10 +176,12 @@ long getNativePtr() {
 
     public void beginTransaction() {
         nativeBeginTransaction(nativePtr);
+        invokeListenerIfSchemaChanged();
     }
 
     public void commitTransaction() {
         nativeCommitTransaction(nativePtr);
+        invokeListenerIfSchemaChanged();
```
cmelchior: I am not sure I understand why these are triggered here? The schema will only change on advance_read right? (which means when we call `advance_read` or begin a transaction)

realm/realm-library/src/main/java/io/realm/internal/SharedRealm.java
```diff
@@ -175,6 +192,16 @@ public boolean isInTransaction() {
         return nativeIsInTransaction(nativePtr);
     }
 
+    public void setSchemaVersion(long schemaVersion) {
+        // FIXME migrate to ObjectStore
```
kneth: There is no easy way to set it. That is, you will have to implement it in Realm Object Store first.

realm/realm-library/src/main/java/io/realm/internal/SharedRealm.java
```diff
@@ -218,6 +245,7 @@ public boolean isEmpty() {
 
     public void refresh() {
         nativeRefresh(nativePtr);
+        invokeListenerIfSchemaChanged();
```
kneth: Can Realm Object Store's `Realm::add_schema_change_handler()` be used for anything useful here?

realm/realm-annotations-processor/src/test/resources/io/realm/AllTypesRealmProxy.java
zaki50: Right. Wrapping by unmodifiable was in `setIndicesMap` and just moved here to avoid wrapping multiple times.

But yes it's an overhead we can avoid. I'll remove this and add a comment on  `getIndicesMap()` something like `do not modify returne map`

realm/realm-library/src/androidTest/java/io/realm/ColumnIndicesTests.java
zaki50: done

CHANGELOG.md
zaki50: done

realm/realm-annotations-processor/src/test/resources/io/realm/AllTypesRealmProxy.java
zaki50: done

realm/realm-annotations-processor/src/test/resources/io/realm/AllTypesRealmProxy.java
zaki50: OK, I'll remove this check. This method is only used internally.

zaki50: done

realm/realm-library/src/main/java/io/realm/RealmCache.java
zaki50: fixed

realm/realm-library/src/main/java/io/realm/internal/ColumnInfo.java
zaki50: done

realm/realm-library/src/main/java/io/realm/internal/ColumnInfo.java
zaki50: done

realm/realm-library/src/main/java/io/realm/RealmCache.java
zaki50: I moved `({@code == null})` after `slot`

realm/realm-annotations-processor/src/test/resources/io/realm/AllTypesRealmProxy.java
zaki50: done

realm/realm-library/src/main/java/io/realm/internal/SharedRealm.java
zaki50: You are right. WE don't need this if we only support schema changes by remote process.

I added here for supporting the changing the schema version in *this* process.
No need to do this?

realm/realm-library/src/main/java/io/realm/internal/SharedRealm.java
zaki50: I found `set_schema_version()` in `object_store.cpp`. It does almost same thing here.

realm/realm-library/src/main/java/io/realm/internal/SharedRealm.java
zaki50: Thanks. I'll take a look.

realm/realm-library/src/main/java/io/realm/RealmCache.java
zaki50: The reason is `4` is not so big number and no specific reason...
Current implementation uses linear search when finding entry from array or storing entry to array. 

I think 2 or 3 is OK in most cases since schema change does not happen so often while the process of mobile app is living.

zaki50:
retest this please

realm/realm-library/src/main/java/io/realm/internal/SharedRealm.java
beeender: Does related dynamic APIs will refresh the schema cache as well?

realm/realm-library/src/main/java/io/realm/internal/ColumnInfo.java
```diff
@@ -34,11 +33,33 @@ protected final long getValidColumnIndex(String realmPath, Table table,
...
+     * @throws IllegalArgumentException if {@code other} has different class than this.
+     */
+    public abstract void copyColumnInfoFrom(ColumnInfo other);
+
+    /**
+     * Creates a copy of this instance.
+     *
+     * @return a copy.
+     */
+    public abstract ColumnInfo copyColumnInfo();
```
beeender: What about implement `Cloneable` instead?

realm/realm-library/src/main/java/io/realm/internal/ColumnIndices.java
```diff
@@ -50,4 +56,28 @@ public long getColumnIndex(Class<? extends RealmModel> clazz, String fieldName)
             return -1;
         }
     }
+
+    public ColumnIndices copyDeeply() {
```
beeender: `Cloneable` instead?

zaki50: I implemented that way at first, but `clone()` method of each class must call `super().clone()` in it and `super().clone()` has extra costs of some checks. So I stopped using `Cloneable`.

realm/realm-library/src/main/java/io/realm/internal/SharedRealm.java
kneth: True.

realm/realm-library/src/main/java/io/realm/internal/SharedRealm.java
cmelchior: Good point. We also need to do it here for synchronous transactions.

realm/realm-library/src/main/java/io/realm/internal/ColumnIndices.java
zaki50: I changed my mind and now ColumnInfo is Cloneable c0eb1b6db67f7254ff9d86348eafeb3e960294cf

realm/realm-library/src/main/java/io/realm/internal/SharedRealm.java
zaki50: Yes, but `set_schema_version` is not declared in `object_store.hpp`.

We need to modify ObjectStore itself to implement the method on `SharedRealm`. So I'd like to leave this as it is and make another PR to do that.

realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmProxyClassGenerator.java
```diff
@@ -414,8 +445,8 @@ private void emitValidateTableMethod(JavaWriter writer) throws IOException {
         writer.emitStatement("Table table = sharedRealm.getTable(\"%s%s\")", Constants.TABLE_PREFIX, this.simpleClassName);
 
         // verify number of columns
-        writer.beginControlFlow("if (table.getColumnCount() != " + metadata.getFields().size() + ")");
-        writer.emitStatement("throw new RealmMigrationNeededException(sharedRealm.getPath(), \"Field count does not match - expected %d but was \" + table.getColumnCount())",
+        writer.beginControlFlow("if (table.getColumnCount() < " + metadata.getFields().size() + ")");
+        writer.emitStatement("throw new RealmMigrationNeededException(sharedRealm.getPath(), \"Field count is less than expected - expected %d but was \" + table.getColumnCount())",
```
kneth: Should we write to the log if count > getFields().size()? It is not critical but might be nice to have when debugging.

realm/realm-library/src/main/java/io/realm/internal/SharedRealm.java
beeender: Currently, both `Realm` and `DynamicRealm` can exist at the same time, then following code has problems:
```java
dynamicRealm.beginTransaction();
dynamicRealm.removeField("foo");
realm.getFoo(); // Column doesn't exist anymore
dynamicRealm.commitTransaction();
```

I am not quite sure if it worth to spend time on this problem since having both at the same doesn't seem be to a real requirement.
An easy fix is we just don't allow having them at the same time.

zaki50: Current implementation, `DynamicRealm` on the same thread does not trigger any listener because I think that the `VersionID` of the each `SharedRealm` is independent.

If my thought is correct, schema changes by Dynamic API can be seen at the next `advance_read` or `beginTransaction()` by relevant typed Realm.

beeender: We used to use separated `SharedGroup` for dynamic realm and realm in one thread, but i think i changed that when switching to `SharedRealm`. I will confirm this.

realm/realm-library/src/main/java/io/realm/RealmCache.java
beeender: I am not quite convinced of caching old versions of `columnIndices`. Whenever we create a new realm instance, instead of getting a old schema cache, it will throw `MigrationException` first i think?

realm/realm-library/src/main/java/io/realm/internal/SharedRealm.java
zaki50: I'll remove `invokeListenerIfSchemaChanged()` since the API for changing schema version is not in public API and thus we can call it manually from our implementation when needed.

beeender: But `invokeListenerIfSchemaChanged()` in `refresh()` is still needed I think? It might be changed by other process.

cmelchior: @beeender Really? That would be rather unfortunate, because it introduces a lot of really weird situations

beeender: @cmelchior Uh. no, just confirmed, we are not actually using cache mechanism from OS.

realm/realm-library/src/main/java/io/realm/internal/SharedRealm.java
kneth: Indeed. The thought behind my initial comment was  that we need to either duplicate the method (adding it to object store's `SharedRealm`) or make it public in `ObjectStore`. Sorry for not making it clear.

realm/realm-library/src/main/java/io/realm/RealmCache.java
zaki50: Async API has a chance to go back to old schema when the query is complete.

realm/realm-library/src/main/java/io/realm/internal/SharedRealm.java
cmelchior: Making it public in the Object Store would make be feel much safer. This is a table that _must not_ be different than other platforms.

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -342,4 +369,49 @@ private static void copyAssetFileIfNeeded(RealmConfiguration configuration) {
...
     }
+
+    /**
+     * Finds an entry for specified schema version in the array.
+     *
+     * @param array target array of schema cache.
+     * @param schemaVersion requested version of the schema.
+     * @return {@link ColumnIndices} instance for specified schema version. {@code null} if not found.
+     */
+    public static ColumnIndices find(ColumnIndices[] array, long schemaVersion) {
```
beeender: a better name instead of `find`? Something like `findColumnIndices` or `getColumnIndices`? `find` causes trouble for me while reading the code ...

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -342,4 +369,49 @@ private static void copyAssetFileIfNeeded(RealmConfiguration configuration) {
...
+        for (int i = array.length - 1; 0 <= i; i--) {
```
beeender: I think foreach loop would be enough? and easy to write :)
especially after the array get wrapped, the latest cache might be at index 0.
Anyway, minor thing.

beeender: Uh... just ignore me, i think i didn't read the loop carefully

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -342,4 +369,49 @@ private static void copyAssetFileIfNeeded(RealmConfiguration configuration) {
...
+                return candidate;
+            }
+        }
+        return null;
+    }
+
+    /**
+     * Stores the schema cache to the array.
+     * <p>
+     * If the {@code array} has empty slot ({@code == null}), this method stores
```
beeender: Incomplete sentence

zaki50: Oh...
Updated.

beeender:
One thing in my mind is after this PR, if there is schema change while an Realm instance opened, it is not a problem anymore. The problem is after this, the schema version gets bumped. Then user restarted the app, then an `MigrationNeededException` will throw although the schema is compatible with the Realm file, but the schema version mismatched.

```java
      if (currentVersion != UNVERSIONED && currentVersion < requiredVersion && columnIndices == null) {
            realm.doClose();
            throw new RealmMigrationNeededException(configuration.getPath(), String.format("Realm on disk need to migrate from v%s to v%s", currentVersion, requiredVersion));
        }
```

But I guess this should be handled in another PR. What should we do? @realm/java introduce a new member in `RealmConfiguration`? or just don't throw?

beeender:
Another thing is i really like to avoid making copies of `ColumnIndices`. One of my idea is we can use an array to hold the `ColumnIndices` in `RealmSchema` and a same array ref in the object proxy. Something like:

```java
public final class RealmSchema {
//...
    ColumnIndices[1] columnIndices;
//...
}
```

```java
public class AllTypesRealmProxy {
    ColumnIndices[1] columnIndices;
    AllTypesRealmProxy(ColumnInfo[1] columnInfo) {
    }
}
```

So technically it still ref the one in the global cache, but even the one is removed from the cache, it is ok, array still hold a ref to it.
One thing i don't think through is with fine grained notifications, we might be able to know which part of schema is changing, so we probably still need something to do copy-on-write.
Also above should be thread safe, i have a quick look, it seems those are all called under the lock of `RealmCache`.

zaki50:
@beeender I was thinking the similar idea, but it requires almost the same number of array objects as the copied `ColumnIndices` and it also introduces some overhead in each time of the field access.

My conclusion is that the schema change is a rare case and it's better to avoid overheads of the field access.

beeender:
I think copy of `ColumnIndices` can be avoided as well. But anyway, current copying seems to be acceptable.

realm/realm-library/src/androidTest/java/io/realm/RealmTests.java
```diff
@@ -3385,4 +3386,49 @@ public void run(Realm realm) {
...
+                catTable.removeColumn(nameIndex);
+                final long newIndex = catTable.addColumn(RealmFieldType.STRING,
+                        Cat.FIELD_NAME, true);
+
+                realm.setVersion(realm.getConfiguration().getSchemaVersion() + 1);
+
+                nameIndexNew.set(newIndex);
+            }
+        });
+        // we need ↓ to update index cache if the schema version was changed in the same thread.
```
beeender: `↓` is this intended?

realm/realm-library/src/main/java/io/realm/Realm.java
zaki50: I removed this `TODO` comment and added `@throws RealmMigrationNeededException` to APIs that may trigger a schema change (f85bef7e4272ebdbf1e5f91eb019f1dd58e5a6c1).

realm/realm-library/src/main/java/io/realm/internal/ColumnIndices.java
```diff
@@ -50,4 +56,35 @@ public long getColumnIndex(Class<? extends RealmModel> clazz, String fieldName)
...
+            copy.put(entry.getKey(), entry.getValue().clone());
+        }
+        return copy;
+    }
+
+    public void copyFrom(ColumnIndices other, RealmProxyMediator mediator) {
+        for (Map.Entry<Class<? extends RealmModel>, ColumnInfo> entry : classes.entrySet()) {
+            final ColumnInfo otherColumnInfo = other.getColumnInfo(entry.getKey());
+            if (otherColumnInfo == null) {
+                throw new IllegalStateException("failed to copy ColumnIndices cache: "
```
beeender: `failed` -> `Failed`

realm/realm-library/src/androidTest/java/io/realm/RealmTests.java
zaki50: yes.. Should we avoid that?

realm/realm-library/src/main/java/io/realm/internal/ColumnIndices.java
zaki50: done

realm/realm-library/src/androidTest/java/io/realm/RealmTests.java
beeender: I guess it's fine? finally we can have colourful emoji in the source code!

beeender:
:+1:

realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmProxyClassGenerator.java
zaki50: done!

zaki50:
@realm/java I think that all comments are addressed.

Please give me (hopefully) final round review!

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@
 * `RealmConfiguration.Builder.assetFile(Context, String)` has been renamed to `RealmConfiguration.Builder.assetFile(String)`.
 * Object with primary key is now required to define it when the object is created. This means that `Realm.createObject(Class<E>)` and `DynamicRealm.createObject(String)` now throws `RealmException` if they are used to create an object with a primary key field. Use `Realm.createObject(Class<E>, Object)` or `DynamicRealm.createObject(String, Object)` instead.
 * Importing from JSON without the primary key field defined in the JSON object now throws `IllegalArgumentException`.
+* Now `Realm.beginTransaction()`, `Realm.executeTransaction()` and `Realm.waitForChange()` throw `RealmMigrationNeededException` if a remote process introduces an incompatible schema changes (#3409).
```
cmelchior: Minor, but I would put this under enhancements, since we didn't support interprocess changes before.

Note: But with the PR merged are we missing anything for claiming full interprocess support? /cc @beeender

realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmProxyClassGenerator.java
```diff
@@ -436,18 +437,29 @@ private void emitInitTableMethod(JavaWriter writer) throws IOException {
...
-        writer.emitStatement("throw new RealmMigrationNeededException(sharedRealm.getPath(), \"Field count is less than expected - expected %d but was \" + table.getColumnCount())",
-                metadata.getFields().size());
+        writer.emitStatement("final long columnCount = table.getColumnCount()");
+        writer.beginControlFlow("if (columnCount != %d)", metadata.getFields().size());
+            writer.beginControlFlow("if (columnCount < %d)", metadata.getFields().size());
+                writer.emitStatement("throw new RealmMigrationNeededException(sharedRealm.getPath(), \"Field count is less than expected - expected %d but was \" + columnCount)",
+                        metadata.getFields().size());
+            writer.endControlFlow();
+            writer.beginControlFlow("if (allowExtraColumns)");
+                writer.emitStatement("RealmLog.info(\"Field count is more than expected - expected %d but was %%1$d\", columnCount)",
```
cmelchior: Multiple levels of named arguments. I think my head goes 💥 😄

CHANGELOG.md
beeender: @cmelchior The only thing in my mind right now is `Realm.deleteRealm()`. With proper core support, it seems difficult to support it in a multiprocesses environment.
@finnschiermer

realm/realm-annotations-processor/src/test/resources/io/realm/AllTypesRealmProxy.java
```diff
@@ -275,11 +276,19 @@ public static Table initTable(SharedRealm sharedRealm) {
...
             Table table = sharedRealm.getTable("class_AllTypes");
-            if (table.getColumnCount() < 9) {
-                throw new RealmMigrationNeededException(sharedRealm.getPath(), "Field count is less than expected - expected 9 but was " + table.getColumnCount());
+            final long columnCount = table.getColumnCount();
+            if (columnCount != 9) {
+                if (columnCount < 9) {
+                    throw new RealmMigrationNeededException(sharedRealm.getPath(), "Field count is less than expected - expected 9 but was " + columnCount);
+                }
+                if (allowExtraColumns) {
+                    RealmLog.info("Field count is more than expected - expected 9 but was %1$d", columnCount);
```
cmelchior: I would put his under `debug`. It doesn't seem like something worth the `info` level?

CHANGELOG.md
cmelchior: Ok, I would be inclined to say that is a developer responsibility to not delete files being accessed (although throwing in that case would be nice).

cmelchior:
Besides a comment about the log level being to high. I think this is 👍

beeender:
:+1:

realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmProxyClassGenerator.java
zaki50: the diff of generated code under `realm/realm-annotations-processor/src/test/resources/io/realm/` help me very much!

beeender: WHAT??? there is a diff generated!!!! I have never known that!

cmelchior: Me neither 👯

realm/realm-annotations-processor/src/test/resources/io/realm/AllTypesRealmProxy.java
zaki50: done

PR #3416, opened 2016-09-08
Ignore callingOrdersOfListeners for now

beeender:
since it becomes a flaky test after #3370.

And it will be fixed very soon.


cmelchior:
👍

zaki50:
:+1:

PR #3418, opened 2016-09-09
Disallow changing PK after object created

beeender:
Thrown an exception if changing the pk after the object created.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 * `RealmConfiguration.Builder.assetFile(Context, String)` has been renamed to `RealmConfiguration.Builder.assetFile(String)`.
 * Object with primary key is now required to define it when the object is created. This means that `Realm.createObject(Class<E>)` and `DynamicRealm.createObject(String)` now throws `RealmException` if they are used to create an object with a primary key field. Use `Realm.createObject(Class<E>, Object)` or `DynamicRealm.createObject(String, Object)` instead.
 * Importing from JSON without the primary key field defined in the JSON object now throws `IllegalArgumentException`.
+* Primary key cannot be changing anymore after the `RealmObject` created.
 
 ### Enhancements
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@
 * `RealmConfiguration.Builder.assetFile(Context, String)` has been renamed to `RealmConfiguration.Builder.assetFile(String)`.
 * Object with primary key is now required to define it when the object is created. This means that `Realm.createObject(Class<E>)` and `DynamicRealm.createObject(String)` now throws `RealmException` if they are used to create an object with a primary key field. Use `Realm.createObject(Class<E>, Object)` or `DynamicRealm.createObject(String, Object)` instead.
 * Importing from JSON without the primary key field defined in the JSON object now throws `IllegalArgumentException`.
-* Primary key cannot be changing anymore after the `RealmObject` created.
+* The primary key value of an object can no longer be changed after the object was created. Instead a new object must be created and all fields copied over.
 
 ### Enhancements
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@
 * `RealmConfiguration.Builder.assetFile(Context, String)` has been renamed to `RealmConfiguration.Builder.assetFile(String)`.
 * Object with primary key is now required to define it when the object is created. This means that `Realm.createObject(Class<E>)` and `DynamicRealm.createObject(String)` now throws `RealmException` if they are used to create an object with a primary key field. Use `Realm.createObject(Class<E>, Object)` or `DynamicRealm.createObject(String, Object)` instead.
 * Importing from JSON without the primary key field defined in the JSON object now throws `IllegalArgumentException`.
-* Now `Realm.beginTransaction()`, `Realm.executeTransaction()` and `Realm.waitForChange()` throw `RealmMigrationNeededException` if a remote process introduces an incompatible schema changes (#3409).
+* Now `Realm.beginTransaction()`, `Realm.executeTransaction()` and `Realm.waitForChange()` throw `RealmMigrationNeededException` if a remote process introduces incompatible schema changes (#3409).
 * The primary key value of an object can no longer be changed after the object was created. Instead a new object must be created and all fields copied over.
 
 ### Enhancements
--- a/realm-annotations/src/main/java/io/realm/annotations/PrimaryKey.java
+++ b/realm-annotations/src/main/java/io/realm/annotations/PrimaryKey.java
@@ -25,10 +25,11 @@ import java.lang.annotation.Target;
  * The @PrimaryKey annotation will mark a field as a primary key inside Realm. Only one field in a
  * RealmObject class can have this annotation, and the field should uniquely identify the object.
  * Trying to insert an object with an existing primary key will result in an
- * {@link io.realm.exceptions.RealmPrimaryKeyConstraintException}.
- *
+ * {@link io.realm.exceptions.RealmPrimaryKeyConstraintException}. Primary key cannot be changed
+ * after the object created.
+ * <p>
  * Primary keys also count as having the {@link Index} annotation.
- *
+ * <p>
  * It is allowed to apply this annotation on the following primitive types: byte, short, int, and long.
  * String, Byte, Short, Integer, and Long are also allowed, and further permitted to have {@code null}
  * as a primary key value.
--- a/realm/realm-annotations-processor/src/main/java/io/realm/processor/Constants.java
+++ b/realm/realm-annotations-processor/src/main/java/io/realm/processor/Constants.java
@@ -30,6 +30,9 @@ public class Constants {
             "throw new IllegalArgumentException(\"Trying to set non-nullable field '%s' to null.\")";
     static final String STATEMENT_EXCEPTION_NO_PRIMARY_KEY_IN_JSON =
             "throw new IllegalArgumentException(\"JSON object doesn't have the primary key field '%s'.\")";
+    static final String STATEMENT_EXCEPTION_PRIMARY_KEY_CANNOT_BE_CHANGED =
+            "throw new io.realm.exceptions.RealmException(\"Primary key field '%s' cannot be changed after object" +
+                    " created.\")";
 
     static final Map<String, String> JAVA_TO_REALM_TYPES;
     static {
--- a/realm/realm-annotations-processor/src/main/java/io/realm/processor/Constants.java
+++ b/realm/realm-annotations-processor/src/main/java/io/realm/processor/Constants.java
@@ -32,7 +32,7 @@ public class Constants {
             "throw new IllegalArgumentException(\"JSON object doesn't have the primary key field '%s'.\")";
     static final String STATEMENT_EXCEPTION_PRIMARY_KEY_CANNOT_BE_CHANGED =
             "throw new io.realm.exceptions.RealmException(\"Primary key field '%s' cannot be changed after object" +
-                    " created.\")";
+                    " was created.\")";
 
     static final Map<String, String> JAVA_TO_REALM_TYPES;
     static {
--- a/realm/realm-library/src/androidTest/java/io/realm/DynamicRealmObjectTests.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/DynamicRealmObjectTests.java
@@ -37,6 +37,15 @@ import io.realm.entities.CyclicType;
 import io.realm.entities.Dog;
 import io.realm.entities.NullTypes;
 import io.realm.entities.Owner;
+import io.realm.entities.PrimaryKeyAsBoxedByte;
+import io.realm.entities.PrimaryKeyAsBoxedInteger;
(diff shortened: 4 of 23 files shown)


CHANGELOG.md
```diff
@@ -15,6 +15,7 @@
 * `RealmConfiguration.Builder.assetFile(Context, String)` has been renamed to `RealmConfiguration.Builder.assetFile(String)`.
 * Object with primary key is now required to define it when the object is created. This means that `Realm.createObject(Class<E>)` and `DynamicRealm.createObject(String)` now throws `RealmException` if they are used to create an object with a primary key field. Use `Realm.createObject(Class<E>, Object)` or `DynamicRealm.createObject(String, Object)` instead.
 * Importing from JSON without the primary key field defined in the JSON object now throws `IllegalArgumentException`.
+* Primary key cannot be changing anymore after the `RealmObject` created.
```
cmelchior: What about instead:

"The primary key value of an object can no longer be changed after the object was created. Instead a new object must be created and all fields copied over."

Zhuinden: Technically `realm.copyFromRealm()` should do exactly that, no?

cmelchior: Not really. You can do anything with in-memory copies, and if you do `realm.copyToRealmOrUpdate(copy)` it will still work as it just updates the existing object.

What will not work is:

```
Person p = realm.where(Person.class).equalTo("id", 42).findFirst():
realm.beginTransaction();
p.setId(43); // Change key of existing object
realm.commitTransaction();
```

Zhuinden: I meant that a new (unmanaged) object is created with all fields copied over, like so

``` java
_realm.executeTransaction((realm) -> {
    Person p = realm.where(Person.class).equalTo("id", 42).findFirst():
    // p.setId(43); // this would fail
    Person up = realm.copyFromRealm(p);
    up.setId(43);
    realm.insertOrUpdate(up); // this will work
});
```

So you don't need to _manually copy every field_. But I think we mean the same thing, after all :tongue:

TR4Android:
Why exactly would you disallow changing the primary key? I actually frequently change the temporary id after a sync with the server which assigns the final id. If this change is added I'd suddenly find myself deleting and recreating objects...

Zhuinden:
@TR4Android I'd assume it's because the Object Store's primary key implementation in the new version of the core does not allow it. Simple as that.

stk1m1:
@beeender 

I'd wish `PrimaryKeyAsString.java` has same field names as others for easy parameterization but the particular class was used in many tests that I wasn't able to modify it. Might be able to add `NotNullablePrimaryKey` interface to do the job a bit easy. The PR looks awesome!

@TR4Android 

AFAICT, There are few merits over fixing PK value. 
1. You'll be fool-proven from accidentally colliding two objects by letting them have the same PK. As of now, it will generate a dup-ed PK exception when you change the PK of an object or you'd update an unintended object.
2. [Realm-Cocoa](https://github.com/realm/realm-cocoa) has long supported this that Java binding will be aligned in behavior.
3. There are more important reasons why this is necessary but you'll see in time.

Zhuinden:
How would you define "unintended object" though? Having to copy out the data and saving it again, and deleting the old data seems counterproductive. I understand the duplication warning, but not the disallowing.

TR4Android:
@Zhuinden Yeah, I find it quite counterintuitive to need to copy the object, set the new id and then delete the old object. This doesn't exactly fool-prove me from colliding primary keys either @stk1m1.

Other databases like SQL support and update of the primary key as well, so I see no reason why Realm shouldn't. It should throw an error for duplicated primary keys though.

stk1m1:
@Zhuinden 
> (I don't see why `@NotNullablePrimaryKey` would make sense compared to just `@Required` and `@PrimaryKey`)

`NotNullablePrimaryKey` _interface_ is for streamlining test cases, and not related to PK annotation.

Zhuinden:
Ah, my mistake 😕

beeender:
@stk1m1 yes, i tried to use same `id` and `value`

beeender:
@TR4Android I am sorry this PR will cause some inconvenience for you, but we have to do this to adapt some changes from the underlying storage engine :(
In the future, this limitation might be relaxed, but now you need some workaround like @Zhuinden mentioned in https://github.com/realm/realm-java/pull/3418#discussion_r78190436

cmelchior:
@TR4Android I agree this change is unfortunate, but perhaps you can elaborate on how your application code handles changing ID's. Perhaps we can come up with something that works better in that case.

I mean if you have changing ID's it seems hard to use our various `*orUpdate()` methods? And if you don't use them, a natural solution could be to just remove the `@PrimaryKey` annotation from your models?

Zhuinden:
@cmelchior you're right, maybe it should just be `@Index` then for him instead of `@PrimaryKey`

realm-annotations/src/main/java/io/realm/annotations/PrimaryKey.java
```diff
@@ -25,10 +25,11 @@
  * The @PrimaryKey annotation will mark a field as a primary key inside Realm. Only one field in a
  * RealmObject class can have this annotation, and the field should uniquely identify the object.
  * Trying to insert an object with an existing primary key will result in an
- * {@link io.realm.exceptions.RealmPrimaryKeyConstraintException}.
- *
+ * {@link io.realm.exceptions.RealmPrimaryKeyConstraintException}. Primary key cannot be changed
```
cmelchior: A Primary key value cannot be changed after the object has been created.

Zhuinden: I kinda think that if you have a managed object with ID 42 and then you call `.setId(42)`, then the Realm code should detect that it is the same value and not an actual "change".

realm/realm-library/src/main/java/io/realm/DynamicRealmObject.java
```diff
@@ -794,4 +808,13 @@ public String toString() {
     public ProxyState realmGet$proxyState() {
         return proxyState;
     }
+
+    // Checks if the given field is primary key field. Throws if it is a PK field.
+    private void checkIsPrimaryKey(String fieldName) {
+        RealmObjectSchema objectSchema = proxyState.getRealm$realm().getSchema().getSchemaForClass(getType());
+        if (objectSchema.hasPrimaryKey() && objectSchema.getPrimaryKey().equals(fieldName)) {
+            throw new RealmException(String.format("Primary key field '%s' cannot be changed after object created."
```
cmelchior: Throw `IllegalArgumentException` instead?

realm-annotations/src/main/java/io/realm/annotations/PrimaryKey.java
beeender: @Zhuinden That requires additional check and it doesn't seem to be useful ... user simply should avoid to call `setXXX` to set PK, right?

cmelchior:
@Zhuinden If your primary keys are given by the server, the uniqueness constraint should be guaranteed as well. No need for the client to enforce it (albeit it does have value as documentation).

cmelchior:
@beeender Looks fine to me. Only had a small comment about the type of exception being thrown 👍

realm/realm-library/src/main/java/io/realm/DynamicRealmObject.java
beeender: eg.: `setId(1)` to set PK, so it is not really because of 1 is illegal, but the `id` cannot be changed. IAE is not quite feasible here i think?

cmelchior: This is only for DynamicRealmObject, so it is called from e.g `setInt("myId", 1)`, right? So the IllegalArgument is for the the "myId" argument.

beeender: Typed Realm needs to throw as well. So throw IAE here and `RealmExcpetion` there? Will that look better?

Zhuinden:
@cmelchior that's true, to be honest these "temporary IDs" seem weird to me, I've never seen anything like that before, so this change doesn't affect me.

But I still think trying to set the same value as primary key shouldn't throw an exception, and is something that Realm should detect.

realm/realm-library/src/main/java/io/realm/DynamicRealmObject.java
cmelchior: IMO yes 👍

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@
 * `RealmConfiguration.Builder.assetFile(Context, String)` has been renamed to `RealmConfiguration.Builder.assetFile(String)`.
 * Object with primary key is now required to define it when the object is created. This means that `Realm.createObject(Class<E>)` and `DynamicRealm.createObject(String)` now throws `RealmException` if they are used to create an object with a primary key field. Use `Realm.createObject(Class<E>, Object)` or `DynamicRealm.createObject(String, Object)` instead.
 * Importing from JSON without the primary key field defined in the JSON object now throws `IllegalArgumentException`.
+* Now `Realm.beginTransaction()`, `Realm.executeTransaction()` and `Realm.waitForChange()` throw `RealmMigrationNeededException` if a remote process introduces an incompatible schema changes (#3409).
```
Zhuinden: ~an~

TR4Android:
Well, my app syncs two databases, an app database and a server database. The user is able to enter data into the database, even when he is offline, which is why the app assigns a temporary (negative) id, as it does not know what the next id on the server database would be. After a sync, the server reports the final id that it assigned, and the app changes the temporary id assigned previously to the official id.

While I could remove the `@PrimaryKey` annotation, I'd really like the duplication check, as that has helped quite a lot finding issues in the code.

Hope this gets fixed in a future release, as I really hate workarounds... Still, it's a great library you guys have :+1:

realm/realm-annotations-processor/src/main/java/io/realm/processor/Constants.java
```diff
@@ -30,6 +30,9 @@
             "throw new IllegalArgumentException(\"Trying to set non-nullable field '%s' to null.\")";
     static final String STATEMENT_EXCEPTION_NO_PRIMARY_KEY_IN_JSON =
             "throw new IllegalArgumentException(\"JSON object doesn't have the primary key field '%s'.\")";
+    static final String STATEMENT_EXCEPTION_PRIMARY_KEY_CANNOT_BE_CHANGED =
+            "throw new io.realm.exceptions.RealmException(\"Primary key field '%s' cannot be changed after object" +
+                    " created.\")";
```
kneth: "created" -> "was created"

realm/realm-library/src/main/java/io/realm/DynamicRealmObject.java
```diff
@@ -794,4 +808,13 @@ public String toString() {
     public ProxyState realmGet$proxyState() {
         return proxyState;
     }
+
+    // Checks if the given field is primary key field. Throws if it is a PK field.
+    private void checkIsPrimaryKey(String fieldName) {
+        RealmObjectSchema objectSchema = proxyState.getRealm$realm().getSchema().getSchemaForClass(getType());
+        if (objectSchema.hasPrimaryKey() && objectSchema.getPrimaryKey().equals(fieldName)) {
+            throw new IllegalArgumentException(String.format("Primary key field '%s' cannot be changed after object created."
+                    , fieldName));
```
kneth: Move `,` to previous line
Either "created" -> "was created" or "created" -> "creation"

kneth:
:+1: with minor comments

realm/realm-library/src/androidTest/java/io/realm/DynamicRealmObjectTests.java
```diff
@@ -467,6 +519,28 @@ public void typedSetter_null() {
...
+            try {
+                if (pkClass.equals(PrimaryKeyAsString.CLASS_NAME)) {
+                    object = dynamicRealm.createObject(pkClass, "");
+                    object.setNull(PrimaryKeyAsString.FIELD_PRIMARY_KEY);
+                } else {
+                    object = dynamicRealm.createObject(pkClass, 0);
+                    object.setNull("id");
+                }
+                fail();
+            } catch (IllegalArgumentException ignored) {
```
zaki50: I think `try -catch` should not include `dynamicRealm.createObject()`.

realm/realm-library/src/androidTest/java/io/realm/DynamicRealmObjectTests.java
```diff
@@ -823,6 +897,37 @@ public void untypedSetter_illegalImplicitConversionThrows() {
...
+            try {
+                if (pkClass.equals(PrimaryKeyAsString.CLASS_NAME)) {
+                    object = dynamicRealm.createObject(pkClass, "");
+                    object.set(PrimaryKeyAsString.FIELD_PRIMARY_KEY, value);
+                } else {
+                    object = dynamicRealm.createObject(pkClass, 0);
+                    object.set("id", value);
+                }
+                fail();
+            } catch (IllegalArgumentException ignored) {
```
zaki50: same

zaki50:
:+1: with minor comment about test

zaki50:
@beeender pushed small fix

PR #3424, opened 2016-09-12
use the logic in Object Store to update schema version

### LINKED ISSUE #3417: Use ObjectStore functions for creating meta tables
To ensure full and stable compatibility between platforms we need to use Object Store functionality for creating the `meta` and `pk` tables.

See e.g https://github.com/realm/realm-java/pull/3409#discussion-diff-78084779

zaki50:
fixes #3417

### DIFF
--- a/realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
+++ b/realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
@@ -148,6 +148,24 @@ Java_io_realm_internal_SharedRealm_nativeGetVersion(JNIEnv *env, jclass, jlong s
     return -1;
 }
 
+JNIEXPORT void JNICALL
+Java_io_realm_internal_SharedRealm_nativeSetVersion(JNIEnv *env, jclass, jlong shared_realm_ptr, jlong version)
+{
+    TR_ENTER_PTR(env, shared_realm_ptr)
+
+    auto shared_realm = *(reinterpret_cast<SharedRealm*>(shared_realm_ptr));
+    try {
+        if (!shared_realm->is_in_transaction()) {
+            std::ostringstream ss;
+            ss << "Cannot set schema version when the realm is not in transaction.";
+            ThrowException(env, IllegalState, ss.str());
+            return;
+        }
+
+        ObjectStore::set_schema_version(shared_realm->read_group(), static_cast<uint64_t>(version));
+    } CATCH_STD()
+}
+
 JNIEXPORT jboolean JNICALL
 Java_io_realm_internal_SharedRealm_nativeIsEmpty(JNIEnv *env, jclass, jlong shared_realm_ptr)
 {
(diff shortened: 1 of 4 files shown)


zaki50:
ready for review @realm/java

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
```diff
@@ -148,6 +148,24 @@ Java_io_realm_internal_SharedRealm_nativeGetVersion(JNIEnv *env, jclass, jlong s
     return -1;
 }
 
+JNIEXPORT void JNICALL
+Java_io_realm_internal_SharedRealm_nativeSetVersion(JNIEnv *env, jclass, jlong shared_realm_ptr, jlong version)
```
beeender: Rename to `nativeSetSchemaVersion` instead since we have a method `nativeGetVersion` is to get the `SharedGroup` version.

realm/realm-library/src/main/java/io/realm/RealmSchema.java
```diff
@@ -87,7 +87,7 @@ public RealmObjectSchema get(String className) {
         Set<RealmObjectSchema> schemas = new LinkedHashSet<>(tableCount);
         for (int i = 0; i < tableCount; i++) {
             String tableName = realm.sharedRealm.getTableName(i);
-            if (Table.isMetaTable(tableName)) {
+            if (!Table.isModelTable(tableName)) {
```
beeender: Is this correct? For the old logic, the loop will be skipped if it is a primary key table.

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
zaki50: `nativeGetVersion()` seems to be also dealing with schema version.

```
JNIEXPORT jlong JNICALL
Java_io_realm_internal_SharedRealm_nativeGetVersion(JNIEnv *env, jclass, jlong shared_realm_ptr)
{
    TR_ENTER_PTR(env, shared_realm_ptr)

    auto shared_realm = *(reinterpret_cast<SharedRealm*>(shared_realm_ptr));
    try {
        return static_cast<jlong>(ObjectStore::get_schema_version(shared_realm->read_group()));
    } CATCH_STD()

    // FIXME: Use constant value
    return -1;
}
```

realm/realm-library/src/main/java/io/realm/RealmSchema.java
zaki50: I think modified code also skips primary key table since primary key table does not have `class_` prefix.
Do I have something overlooked?

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
beeender: Uh, the other one is `Java_io_realm_internal_SharedRealm_nativeGetVersionID()` ... Sorry... Maybe change them both to `SchemaVersion()`? But let's do it in another PR ... they are not related with this one.

realm/realm-library/src/main/java/io/realm/RealmSchema.java
beeender: Nevermind, my bad :)

beeender:
:+1:

kneth:
:+1:

PR #3433, opened 2016-09-13
Check if any object missed in the module

### LINKED ISSUE #3398: A module/schema must include all required classes
If class `A` links to class `B` but a module/schema does not include `B`, object store will crash during validation (calling `update_schema`). We must ensure that all required classes (direct and indirect) are included in the module/schema. If a class is missing, an exception must be thrown.

beeender:
Close #3398
If a linked object is not declared in the RealmModule, a RealmException
will be thrown. This is achieved by calling mediator's createTable
instead of Table's getTable which doesn't have proper check.


beeender:
This is strange, `RealmQueryTests.findFirst()` failed which should not be impacted at all by this PR. Need investigation.

zaki50:
https://github.com/realm/realm-java/pull/3433/files#diff-fe5ca9f6e0254b639a10bcc000406fdbR843 should be addressed before merging.

realm/realm-library/src/androidTest/java/io/realm/RealmMigrationTests.java
```diff
@@ -831,6 +840,9 @@ public void openPreNullWithRequired() throws IOException {
                 .schemaVersion(0)
                 .schema(AllTypes.class)
                 .build();
+        // FIXME this test should fail since AllTypes contains a link to Dog table.
+        // We should add other required classes to the schema above and also we should
+        // have another test to check if the mediator contains all required classes.
```
beeender: Created #3481 for this. Let's keep this issue out of this PR scope.

realm/realm-library/src/androidTest/java/io/realm/entities/CatOnlyModule.java
```diff
@@ -0,0 +1,24 @@
+/*
+ * Copyright 2015 Realm Inc.
```
zaki50: 2016

realm/realm-library/src/androidTest/java/io/realm/entities/DogOnlyModule.java
```diff
@@ -0,0 +1,23 @@
+/*
+ * Copyright 2015 Realm Inc.
```
zaki50: 2016

realm/realm-library/src/main/java/io/realm/internal/RealmProxyMediator.java
```diff
@@ -43,12 +43,25 @@
...
+    /**
      * Creates the backing table in Realm for the given RealmObject class.
      *
+     * @param rootMediator the {@link RealmProxyMediator} will be used when the {@code clazz} has a linked field needs
+     *                     to be created.
      * @param clazz the {@link RealmObject} model class to create backing table for.
      * @param sharedRealm the wrapper object of underlying native database.
      */
-    public abstract Table createTable(Class<? extends RealmModel> clazz, SharedRealm sharedRealm);
+    public abstract Table createTable(RealmProxyMediator rootMediator, Class<? extends RealmModel> clazz,
```
zaki50: Why do we need this?
I think `rootMediator` is always `this`.

beeender: See `CompositeMediator`

zaki50:
I think this PR should go to `master` since this is an enhancement.

beeender:
#4674 fixed this.

PR #3449, opened 2016-09-17
fix unstable test

zaki50:
@realm/java


realm/realm-library/src/androidTest/java/io/realm/RealmAsyncQueryTests.java
```diff
@@ -1125,7 +1125,7 @@ public void findAllSortedAsync_batchUpdate() {
             public boolean onInterceptInMessage(int what) {
                 switch (what) {
                     case HandlerControllerConstants.COMPLETED_ASYNC_REALM_RESULTS: {
-                        if (numberOfIntercept.incrementAndGet() == 1) {
+                        if (numberOfIntercept.incrementAndGet() == 2 /* 2 queries are both completed */) {
```
cmelchior: I thought we actually batched the async query results, but it doesn't look like it from reviewing the code in `HandlerController`, so yes I think this change is correct 👍

beeender:
:+1:

PR #3454, opened 2016-09-18
resurrect buildTargetABIs property support

zaki50:
This PR enables us to specify `buildTargetABIs` property in command-line or in `~/.gradle/gradle.properties` like before and it will be ignored when we execute gradle from AS (AS automatically set appropriate ABI to `android.injected.build.abi`).

@realm/java


zaki50:
@nhachicha yes, `android.injected.build.abi` is basically added by the AS and we don't import root project into AS. So I think we need to support that in the root project. We can use `buildTargetABIs ` property instead.

PR #3457, opened 2016-09-18
Introduce global init

cmelchior:
This PR changes slightly how Realm is initialized

1) `Context` is now removed from `RealmConfiguration.Builder()`
2) Instead a global `Realm.init(Context)` is introduced

It has a number of advantages:

- The `Context` is now available before touching _any_ Realm API. This is especially an advantage for future features that wants to interact with the Android system without having a configuration first.
- We now always have the context when loading JNI code, which means we can always use Relinker.
- Simplified Builder.
- `Realm.getDefaultInstance()` now works automatically without having to set one first
- The JNI code is now only loaded from one location.


realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -142,6 +147,52 @@
...
+     *
+     * @param context the Application Context.
+     * @throws IllegalArgumentException if null context is provided.
+     */
+    public static synchronized void init(Context context) {
+        if (context == null) {
+            throw new IllegalArgumentException("Non-null context required.");
+        }
+        RealmCore.loadLibrary(context);
+        BaseRealm.applicationContext = context;
```
beeender: Check if the context is application context, and throws.

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -142,6 +147,52 @@
...
+        defaultConfiguration = new RealmConfiguration.Builder().build();
+
+        if (!realmInitialized) {
```
beeender: this can be checked at the beginning of the method.

realm/realm-library/src/main/java/io/realm/Realm.java
beeender: Or we could just call `getApplicationContext` instead. And add some javadoc for it.

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -142,6 +147,52 @@
...
+            RealmLog.add(BuildConfig.DEBUG ? new AndroidLogger(Log.DEBUG) : new AndroidLogger(Log.WARN));
+        }
+        realmInitialized = true;
```
beeender: `realmInitialized` is not needed. And set `BaseRealm.applicationContext` has to be the last statement in this function, we can just use `BaseRealm.applicationContext` as the flag

examples/kotlinExample/src/main/kotlin/io/realm/examples/kotlin/MyApplication.kt
```diff
@@ -0,0 +1,30 @@
+/*
+ * Copyright 2014 Realm Inc.
```
zaki50: 2016

examples/migrationExample/src/main/java/io/realm/examples/realmmigrationexample/MigrationExampleActivity.java
```diff
@@ -59,7 +60,7 @@ protected void onCreate(Bundle savedInstanceState) {
 
         // When you create a RealmConfiguration you can specify the version of the schema.
         // If the schema does not have that version a RealmMigrationNeededException will be thrown.
-        RealmConfiguration config0 = new RealmConfiguration.Builder(this)
+        RealmConfiguration config0 = new RealmConfiguration.Builder()
                 .name("default0")
```
zaki50: not in the scope of this PR, filename should ends with `.realm`

realm/realm-library/src/main/java/io/realm/Realm.java
cmelchior: Doh, why didn't I think of that :)

zaki50: @beeender We should consider that unit tests would set own Context for changing base directory.
So `getApplicationContext()` is better if it works.

examples/jsonExample/src/main/java/io/realm/examples/json/JsonExampleActivity.java
```diff
@@ -44,9 +44,10 @@
     @Override
     protected void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
+        Realm.init(getApplicationContext());
```
cmelchior: It was mostly to avoid having to create the MyApplication subclass, but I was probably just beeing lazy, and I agree that this pattern is bad. I'll add the Application subclass to the other examples.

examples/migrationExample/src/main/java/io/realm/examples/realmmigrationexample/MigrationExampleActivity.java
cmelchior: Actually it is not a requirement they end with `.realm`, but it is a nice convention.

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -142,6 +147,50 @@
...
+     * @throws IllegalArgumentException if null context is provided.
+     */
+    public static synchronized void init(Context context) {
+        if (BaseRealm.applicationContext == null) {
+            if (context == null) {
+                throw new IllegalArgumentException("Non-null context required.");
+            }
+            BaseRealm.applicationContext = context.getApplicationContext();
+            RealmCore.loadLibrary(BaseRealm.applicationContext);
+            defaultConfiguration = new RealmConfiguration.Builder().build();
```
beeender: Uh... I am not quite sure if this is needed. Any good reason we also set a default configuration here?

cmelchior: Yes, that is so you can get started as quickly as possible:
```
Realm.init(context);
Realm realm = Realm.getDefaultInstance()
```
Would work.

All the other SDK's have a similar functionality

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -142,6 +147,50 @@
...
+     *
+     * @param context the Application Context.
+     * @throws IllegalArgumentException if null context is provided.
+     */
+    public static synchronized void init(Context context) {
+        if (BaseRealm.applicationContext == null) {
+            if (context == null) {
+                throw new IllegalArgumentException("Non-null context required.");
+            }
+            BaseRealm.applicationContext = context.getApplicationContext();
```
beeender: `BaseRealm.applicationContext = context.getApplicationContext();` is better to be called after `RealmCore.loadLibrary(BaseRealm.applicationContext);` since after it set, it is allowed to get a Realm instance in other threads, but `.so` might not be loaded yet.

cmelchior: Good catch

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -183,10 +184,10 @@ public static synchronized void init(Context context) {
             if (context == null) {
                 throw new IllegalArgumentException("Non-null context required.");
             }
-            BaseRealm.applicationContext = context.getApplicationContext();
-            RealmCore.loadLibrary(BaseRealm.applicationContext);
+            RealmCore.loadLibrary(context);
             defaultConfiguration = new RealmConfiguration.Builder().build();
             RealmLog.add(BuildConfig.DEBUG ? new AndroidLogger(Log.DEBUG) : new AndroidLogger(Log.WARN));
+            BaseRealm.applicationContext = context.getApplicationContext();
```
beeender: uh.. then the line 188 will throw since the context is `null` there

cmelchior: I don't think so? `context` is the input to the method and there is a nullcheck just above here?

beeender: `this(BaseRealm.applicationContext);` Builder takes the `applicationContext` as input and it is still `null` at line 188.

cmelchior: But only way to solve that would be to re-introduce the `isInitialized` variable or use the protected `Builder(Context)`. I was kinda hoping to get rid of the constructor, but seems easier to use that for now.

beeender: `RealmConfiguration.Builder(Context)` is package protected, so we can use `defaultConfiguration = new RealmConfiguration.Builder().build(context.getApplicationContext);` here?

cmelchior: Yep, that is possible. We can always refactor if we decide to remove the constructor at some point. It will involve _a lot_ of changes to unit tests though, which was why I kept it outside the scope of this PR

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -142,6 +148,50 @@
     }
 
     /**
+     * Initializes the Realm library. It is required to call this method before interacting with any other of the Realm
```
kneth: I think we should explain what we mean by "initialize". Besides loading `.so` files (not important for users of Realm), a default configuration is created and a logger is set up.

realm/realm-library/src/main/java/io/realm/internal/Table.java
```diff
@@ -52,10 +52,6 @@
     private final SharedRealm sharedRealm;
     private long cachedPrimaryKeyColumnIndex = NO_MATCH;
 
-    static {
```
kneth: If I only run the `internal` tests, will the `.so` be loaded?

realm/realm-library/src/main/java/io/realm/Realm.java
cmelchior: Do we want to go into that much detail? I mean this method is most likely the first method you encounter when using our library so diving into details like logging seems a bit much? Adding a note about the default instance might make sense though

CHANGELOG.md
```diff
@@ -2,12 +2,13 @@
...
 ### Breaking Changes
 
+* It is now required to call `Realm.init(Context)` before calling any other Realm API.
 * `isValid()` now always returns `true` instead of `false` for unmanaged `RealmObject` and `RealmList`. This puts it in line with the behaviour of the Cocoa and .NET API's (#3101).
 * armeabi is not supported anymore.
 * Added new `RealmFileException`.
   - `IncompatibleLockFileException` has been removed and replaced by `RealmFileException` with kind `INCOMPATIBLE_LOCK_FILE`.
   - `RealmIOExcpetion` has been removed and replaced by `RealmFileException`.
-* Removed `RealmConfiguration.Builder(Context, File)` and `RealmConfiguration.Builder(File)` constructors.
+* Removed `RealmConfiguration.Builder(Context)`, `RealmConfiguration.Builder(Context, File)` and `RealmConfiguration.Builder(File)` constructors.
```
bmunkholm: This should be right after the Realm.init() change comment.

realm/realm-library/src/main/java/io/realm/internal/Table.java
cmelchior: Probably not ...

bmunkholm:
Lacking a few tests to verify that you check that init() has been called before using other API's.

realm/realm-library/src/main/java/io/realm/internal/Table.java
cmelchior: I'll create an new issue for figuring out what to do with tests. It is not entirely trivial, with also https://github.com/realm/realm-java/pull/3457#issuecomment-247944547

PR #3462, opened 2016-09-19
add default value instruction support

zaki50:
TODO:
* [x] add SetDefault support to `Table` class
* [x] update annotation processor to use SetDefault
* [x] add tests

Current limitation: following operations as a default value are not supported.
* nullify a link
* all operations (such as adding element, removing element, and so on) to RealmList


zaki50:
Ready for review @realm/java

zaki50:
I added a parameter `isDefault` to `Table.set*()` and passed `false` to `isDefault` in all existing codes.
And then I added a code to pass `true` to `isDefault` in proxy's setters when it is executed in the model's constructor.

zaki50:
Hmm, I need to fix ..

realm/realm-annotations-processor/src/test/resources/io/realm/AllTypesRealmProxy.java
```diff
@@ -170,6 +171,9 @@ private void injectObjectContext() {
             if (!proxyState.getAcceptDefaultValue$realm()) {
                 return;
             }
+            final Row row = proxyState.getRow$realm();
+            row.getTable().setLong(columnInfo.columnLongIndex, row.getIndex(), value, true);
```
kneth: Can we cache the table so we don't have to get it in all setters?

realm/realm-library/src/androidTest/java/io/realm/RealmTests.java
```diff
@@ -2331,6 +2332,17 @@ public void execute(Realm realm) {
     }
 
     @Test
+    public void createObject_overwriteNullifiedLinkWithDefaultValue() {
+        final DefaultValueOverwriteNullLink created;
+        realm.beginTransaction();
+            created = realm.createObject(DefaultValueOverwriteNullLink.class);
```
kneth: Indentation.

realm/realm-library/src/main/cpp/io_realm_internal_Table.cpp
```diff
@@ -615,26 +615,31 @@ JNIEXPORT jlong JNICALL Java_io_realm_internal_Table_nativeGetLinkTarget
...
 
 JNIEXPORT void JNICALL Java_io_realm_internal_Table_nativeSetLink
-  (JNIEnv* env, jclass, jlong nativeTablePtr, jlong columnIndex, jlong rowIndex, jlong targetRowIndex)
+  (JNIEnv* env, jclass, jlong nativeTablePtr, jlong columnIndex, jlong rowIndex, jlong targetRowIndex, jboolean isDefault)
 {
     if (!TBL_AND_INDEX_AND_TYPE_INSERT_VALID(env, TBL(nativeTablePtr), columnIndex, rowIndex, type_Link))
         return;
     try {
-        TBL(nativeTablePtr)->set_link( S(columnIndex), S(rowIndex), S(targetRowIndex));
+        TBL(nativeTablePtr)->set_link( S(columnIndex), S(rowIndex), S(targetRowIndex), isDefault == JNI_TRUE ? true : false);
```
kneth: I believe we have `to_bool()` for casting from `jboolean` to `bool`.

realm/realm-library/src/main/cpp/io_realm_internal_TableView.cpp
```diff
@@ -317,6 +317,16 @@ JNIEXPORT jlong JNICALL Java_io_realm_internal_TableView_nativeGetLink
...
 }
 
+JNIEXPORT jboolean JNICALL Java_io_realm_internal_TableView_nativeIsNull
+        (JNIEnv* env, jobject, jlong nativeViewPtr, jlong columnIndex, jlong rowIndex)
+{
+    try {
+        if (!VIEW_VALID_AND_IN_SYNC(env, nativeViewPtr))
+            return 0;
+    } CATCH_STD()
+    return TV(nativeViewPtr)->get_parent().is_null( S(columnIndex), TV(nativeViewPtr)->get_source_ndx(S(rowIndex))) ? JNI_TRUE : JNI_FALSE;  // noexcept
```
kneth: Please move inside `try` block. I don't think any of these core methods will throw but just being safe.

realm/realm-library/src/androidTest/java/io/realm/RealmTests.java
```diff
@@ -2331,6 +2332,17 @@ public void execute(Realm realm) {
...
+        realm.commitTransaction();
+
+        assertEquals(created.getExpectedKeyOfFieldObject(),
+                created.getFieldObject().getFieldRandomPrimaryKey());
```
zaki50: fixed

realm/realm-annotations-processor/src/test/resources/io/realm/NullTypesRealmProxy.java
```diff
@@ -203,6 +204,12 @@ private void injectObjectContext() {
             if (!proxyState.getAcceptDefaultValue$realm()) {
                 return;
             }
+            final Row row = proxyState.getRow$realm();
+            if (value == null) {
+                throw new IllegalArgumentException("Trying to set non-nullable field 'fieldStringNotNull' to null.");
+            }
+            row.getTable().setString(columnInfo.fieldStringNotNullIndex, row.getIndex(), value, true);
+            return;
```
zaki50: We need to ignore setting primary key when under constructon in order to allow specifying primary key value for unmanaged objects like:
```
    @PrimaryKey
    public String id = UUID.randomUUID().toString();
}

Foo unmanaged = new Foo();
```

realm/realm-annotations-processor/src/test/resources/io/realm/AllTypesRealmProxy.java
```diff
@@ -197,6 +201,9 @@ private void injectObjectContext() {
             if (!proxyState.getAcceptDefaultValue$realm()) {
                 return;
             }
+            final Row row = proxyState.getRow$realm();
```
zaki50: Before this PR, setting value is always directly handled by `UncheckedRow` as non-default value.
Now it is handled by Table and treat it as a default value when called from its constructor.

realm/realm-library/src/androidTest/java/io/realm/internal/JNITableTest.java
```diff
@@ -513,4 +523,423 @@ public void isNullable() {
...
+                }
+            }
+
+        } finally {
+            sharedRealm.close();
+        }
+    }
+
+    @Test
+    public void defaultValue_setMultipleTimes() {
```
zaki50: To make sure we can specify default value multiple times.
For example, users have a chance to write following code:

```
public class Foo {
    public String name = "unknown"; // first default value
    public Foo() {
    }

    public Foo(String name) {
        this.name = name; // second default value. This must overwrite first value.
    }
```

zaki50: Oh, above code is not correct, but user has a chance to set value multiple times.

realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmProxyClassGenerator.java
```diff
@@ -389,51 +449,47 @@ private void emitCodeForInjectingObjectContext(JavaWriter writer, VariableElemen
...
+                                              CodeEmitter defaultValueCodeEmitter) throws IOException {
+        writer.beginControlFlow("if (proxyState.isUnderConstruction())");
+        if (isPrimaryKey) {
+            writer.emitSingleLineComment("default value of the primary key is always ignored.");
+            writer.emitStatement("return");
+        } else {
+            writer.beginControlFlow("if (!proxyState.getAcceptDefaultValue$realm())")
+                    .emitStatement("return")
+                    .endControlFlow();
+            if (Utils.isRealmModel(field)) {
```
zaki50: nice catch.
I simplified code around here.

realm/realm-annotations-processor/src/test/resources/io/realm/AllTypesRealmProxy.java
zaki50: Yes, but the `Table` is only used when the setters are called from its constructor.
And, `row.getTable()` is just returning `Row`'s private field (not involve JNI call).

So I would like to leave as it is.

realm/realm-library/src/androidTest/java/io/realm/RealmTests.java
zaki50: fixed

realm/realm-library/src/main/cpp/io_realm_internal_TableView.cpp
zaki50: fixed

realm/realm-library/src/main/cpp/io_realm_internal_Table.cpp
zaki50: I didn't know that. thanks.
I'll fix all other setters.

zaki50: @kneth I couldn't find `to_bool()` function.

I could only find `Accessor::to_bool(ContextType, ValueType &)`

zaki50: @kneth I found `B()` macro.
Can I use this?

`#define B(x)    static_cast<bool>(x)`

beeender: Isn't `isDefault == JNI_TRUE` enough?

zaki50: I use `B()`. If there is a better way, please comment this.

zaki50: @beeender I thought at first, but I aligned my new code  with existing code `value == JNI_TRUE ? true : false`.
Now I use `B` macro that was used to covert `jboolean` to `bool` in other places.

zaki50:
@realm/java I addressed all comments.

second round please.

realm/realm-annotations-processor/src/test/resources/io/realm/AllTypesRealmProxy.java
cmelchior: I'm not sure I follow why we need to call these methods on the table. Wouldn't it be better to expose them through the Row interface?

cmelchior:
One thing I am not sure of in this PR is what happens to values you didn't set in the constructor? Will they be set with `setDefault` as well or use the standard behavior? Ideally when you do `addEmptyRow` it should use `setDefault` on all columns.

See e.g https://github.com/realm/realm-cocoa-private/issues/230#issuecomment-248380891

realm/realm-annotations-processor/src/test/resources/io/realm/AllTypesRealmProxy.java
zaki50: Core provides SetDefault features only on `Table` class and I couldn't find a way to get `Table` instance from `Row` in JNI layer. `Row` seems to have the pointer to the `Table` as a private member and we can use it by modifying the Core.

Should I ask core team to expose SetDefault feature to `Row` class?

zaki50:
If users doesn't set default  value, the value of the column is its default value set by `addEmptyRow` e.g. `0` or `false` and current implementation does not call SetDefault automatically.

I need to read the comments in cocoa repo in order to align the behavior of `SetDefault`

zaki50:
> Ideally when you do addEmptyRow it should use setDefault on all columns.

But `nillifyLink()` currently does not support `SetDefault` in both `Row` and `Table`thus it's impossible to use `SetDefault` to **all** columns.

cmelchior:
Yup, but all the ones that do should do it.

realm/realm-annotations-processor/src/test/resources/io/realm/AllTypesRealmProxy.java
cmelchior: That would actually be the better design longterm, short term I was mostly thinking that we expose it in the row interface, since Unchecked row already have an reference to the table, we could expose `setDefaultString(index, val)` and friends, but it would inflate the method count, so perhaps it is just better to wait for core support.

For now, I think it is fine to just create a Core ticket requesting this.

cmelchior:
> If users doesn't set default value, the value of the column is its default value set by addEmptyRow e.g. 0 or false and current implementation does not call SetDefault automatically.

Seems that this is already handled automatically, so just ignore my comment

kneth:
LGTM 👍

zaki50:
merged latest master into this branch.
If CI is OK, I'll merge this.

zaki50:
I checked CI console and found that all tests are OK but failed to report it to GitHub.

```
Could not update commit status. Message: GitHub API rate limit exceeded

Finished: SUCCESS
```

I'll merge this.

PR #3481, opened 2016-09-22
Check if any object missed in the mediator

### LINKED ISSUE #3398: A module/schema must include all required classes

beeender:
Close #3398 If a linked object is not declared in the RealmModule, a
RealmException will be thrown. This is achieved by calling mediator's
createTable instead of Table's getTable which doesn't have proper check.


beeender:
#3398 needs to be merged first. And this is still in progress.
Need to spend some time on the old test fix and a new test case.

beeender:
#4674 fixed this.

PR #3512, opened 2016-09-28
Append the version number to the name of all the artifacts uploaded to Bintray

### LINKED ISSUE #3514: Release script to BinTray not working
Yesterday when we published our new artifacts, they were correctly uploaded to BinTray, but the maven-metadata.xml was not generated which means that 3rd parties could not access them.

We need to be able to release everything to BinTray with one gradle command and have it working.

emanuelez:
Fixes #3514.

This might solve an issue regarding Bintray not updating the `maven-metadata.xml` file in the Maven repository


cmelchior:
There is no way to test this sort of officially trying to release a version right?

emanuelez:
Unfortunately not. The `maven-metadata.xml` file is autogenerated when the artifacts are published.

Issue #3411, opened 2016-09-07
Problem with distinct in realmresults 1.20


michaelbukachi:
I've noticed some strange behaviour with the distinct method of a RealmResults object. Whenever I fetch data sorted in descending descending order and call the distinct value on the results, it seems the last distinct value is selected instead of the first e.g:
AA
AA
AA

in the situation above, the last AA would be selected instead of the first. In realm 1.1.1 it used to be the first. Was there a change made to realm cause I checked the changelog and didn't see anything

Zhuinden:
Well it **does** say that the selected element by `distinct()` can be pretty much any element that has that value as its value, and order is undefined... although the sort speed optimization could have changed it.

I don't really know about what happened there because that's in the Core.

michaelbukachi:
Well before it was random, it was always the first distinct element. So I built my query based on that and it has always worked till now

Zhuinden:
Okay wait a second, the docs say

distinct
> public RealmResults<E> distinct(String fieldName)
> Returns a distinct set of objects of a specific class. If the result is sorted, the first object will be returned in case of multiple occurrences, otherwise it is undefined which object is returned.
> Parameters:
> fieldName - the field name.
> Returns:
> a non-null RealmResults containing the distinct objects.
> Throws:
> IllegalArgumentException - if a field is null, does not exist, is an unsupported type, is not indexed, or > points to linked fields.

You **should** get the first element in case of sorted results.

Are you sure it no longer returns the first element consistently? Because that'd be quite the bug. :|

michaelbukachi:
`realm.where(Memo.class).equalTo("isFile", false).findAllSortedAsync("date", Sort.DESCENDING);`
This is the query for fetching data so the data should be sorted in descending order.

`private RealmChangeListener<RealmResults<Memo>> listener = new RealmChangeListener<RealmResults<Memo>>() {
        @Override
        public void onChange(RealmResults<Memo> element) {
            if (!element.isEmpty()) {
                adapter.updateData(element.distinct("connection"));
                adapter.notifyDataSetChanged();
            }
            empty.setVisibility(adapter.getItemCount() == 0 ? View.VISIBLE : View.GONE);
        }
    };`
Here the query results are filtered out. I've confirmed that the results are in descending order. but the last element in the results is selected instead of the first

beeender:
@michaelbukachi Can you please list the values of `RealmResults<Memo> element` and values of the `element.distinct("connection")` here?

[...]

PR #3522, opened 2016-09-28
Unit test combining sort() and distinct()

### LINKED ISSUE #3503: findAllSorted and distinct
#### Goal

I'm using `Kotlin` and I'm trying to retrieve objects with `findAllSorted` using a `Long` that's a timestamp to sort them, and then filtering with `distinct`.

I've also asked on [Stackoverflow](http://stackoverflow.com/questions/39708623/realm-findallsorted-and-distinct)
#### Expected Results

I expected to get the "newest objects  for each distinct value of the field used by `distinct`
#### Actual Results

I got the oldest ones.

```
override fun getStuff(): MutableList<Stuff> = realm.where(Stuff::class.java)
        .findAllSorted("timeStamp", Sort.DESCENDING)
        .distinct("otherId")

storage.saveStuff(Stuff("a", otherId=1, timeStamp=1))
storage.saveStuff(Stuff("b", otherId=1, timeStamp=2))
storage.saveStuff(Stuff("c", otherId=1, timeStamp=3))
// Other stuff with different otherId

storage.getStuff()[0].name == "c" // false
```

Realm version(s): 1.2.0 / 1.1.1

Android Studio version: 2.2

Kotlin Version: 1.0.4

kneth:
Fixes #3503

Connects to #3503
Connects to #3411


Zhuinden:
Any idea what could be causing this?

ironage:
The order of distinct and the order of sort are now two different things. This has to do with the ability to sort or distinct through a chain of linked properties. Looks like we are always passing `ascending=true` [for distinct](https://github.com/realm/realm-java/blob/63d082809b696a5c7f168e39a7b62ed01d2d23ae/realm/realm-library/src/main/cpp/io_realm_internal_TableView.cpp#L107). Maybe we need a way to let users supply the ordering of distinct in the Java binding?

cmelchior:
It sounds like either the Java binding should remember the sort order to provide it to distinct, but more ideally Core would look at the sort order on any restricting views and use that as the default. What do you think @ironage ?

kneth:
@Zhuinden Not yet, but my hypothesis is that `distinct()` resets the sort order of the table view. If we move to Object Store's `Results` class, we will have to apply the order before calling `findAll()`, and thereby preserve the order to any derived `Results`.

ironage:
I think the current implementation is the most flexible, but maybe doesn’t do what users expect. If Java could add a flag for sort on distinct, I could imagine the following:
```
override fun getStuff(): MutableList<Stuff> = realm.where(Stuff::class.java)
        .findAllSorted(“timeStamp", Sort.ASCENDING)
        .distinct("otherId”, Sort.DESCENDING)
```
That example could show a list of items containing only objects with the last unique `otherId` while the overall results are sorted by earliest time. By changing the two Sort flags you could get four different combinations of results.

If you always want the distinct order the same as in the sort order, we could do that core side. This is not currently supported because sort and distinct operate on different properties. But if this is actually what you want then we could implement it (you would lose the flexibility of the current implementation, but if you don't use it anyways that doesn't matter) is this a preferable approach? @cmelchior @kneth @tgoyne

cmelchior:
Hmm, yes good point. I could why sorting order might be interesting  for distinct alone. We could probably always remember the sorting order for the first sort in the binding and just use that as the default value for distinct if none was provided.

kneth:
We should migrate to object store (implementing `RealmResults` using `Results`) and wait for https://github.com/realm/realm-object-store/issues/183

Zhuinden:
Is this still broken? Results integration is complete

kneth:
@Zhuinden We have kept it open since we don't like core's behaviour. Or rather, we need to extend core's behaviour.

Zhuinden:
Realm-Core 3.0.0 will fix this in Realm-Java 4.0.0.

cmelchior:
This is now fixed with the core upgrade to 3.0.0. So we should merge this PR to get the unit test coverage for the behaviour. 

Ready for review @realm/java

PR #3523, opened 2016-09-28
Accelerate build with ccache and lcache

beeender:
See https://github.com/beeender/lcache for source code for lcache.
It is very simple right now. It identifies if the link target is cached by
using checksums from command line + checksums of all input files.


beeender:
`﻿⁠⁠⁠⁠./gradlew assemble  1.50s user 0.16s system 0% cpu 6:13.82 total﻿⁠⁠⁠⁠`
vs
`﻿⁠⁠⁠⁠./gradlew assemble  1.41s user 0.11s system 12% cpu 12.599 total﻿⁠⁠⁠⁠`
for building x86 only on my machine. Seems working.
Haven't tried with mac yet.

beeender:
@realm/java Before we setup CI to use this, every one can actually benefits from it on local building.
Just download https://github.com/beeender/lcache/releases/download/v0.0.0/lcache-mac for the mac, put it some where and do the setup as below:

The realm-library gradle check those paths from project properties first
, then the system env.

So those can be set in the ~/.gradle/gradle.properties like:
ccachePath=/usr/bin/ccache
lcachePath=/usr/bin/lcache

Or by system ENVs like:
REALM_CCACHE_PATH=/usr/bin/ccache
REALM_LCACHE_PATH=/usr/bin/lcache

If nothing changed in the native code, the native build now takes no time.

realm/realm-library/build.gradle
```diff
@@ -30,6 +30,10 @@ if (!ext.coreArchiveDir) {
 ext.coreArchiveFile = rootProject.file("${ext.coreArchiveDir}/realm-sync-android-${project.coreVersion}.tar.gz")
 ext.coreDistributionDir = file("${projectDir}/distribution/realm-core/")
 ext.coreDir = file("${project.coreDistributionDir.getAbsolutePath()}/core-${project.coreVersion}")
+ext.ccachePath = project.hasProperty('ccachePath') ?
```
emanuelez: This can be shortened to:
```
ext.ccachePath = project.findProperty('ccachePath') ?: System.getenv('REALM_CCACHE_PATH')
```
It's also safer in case of refactoring

realm/realm-library/build.gradle
```diff
@@ -30,6 +30,8 @@ if (!ext.coreArchiveDir) {
 ext.coreArchiveFile = rootProject.file("${ext.coreArchiveDir}/realm-sync-android-${project.coreVersion}.tar.gz")
 ext.coreDistributionDir = file("${projectDir}/distribution/realm-core/")
 ext.coreDir = file("${project.coreDistributionDir.getAbsolutePath()}/core-${project.coreVersion}")
+ext.ccachePath = project.findProperty('ccachePath') ?: System.getenv('REALM_CCACHE_PATH')
```
emanuelez: Actually... why using custom environment variables? The CMake toolchain already accepts `NDK_CCACHE`. Could we just use that and introduce `NDK_LCACHE` maybe?

beeender: OK!

beeender:
Sure, i will clean up that lcache repo later, maybe add some necessary features. at least add a readme there

PR #3527, opened 2016-09-28
Add support for the multiple flavors to the metrics collection stage

emanuelez:


emanuelez:
@cmelchior

zaki50:
@beeender In this case, adding `?w=1` at the end of URL makes much readable.

Jenkinsfile
```diff
@@ -141,24 +150,30 @@ def storeJunitResults(String path) {
...
-  sendMetrics('aar_size', aarFile.length as String)
-
-  def soFiles = findFiles(glob: 'realm/realm-library/build/outputs/aar/unzipped/jni/*/librealm-jni.so')
-  for (int i = 0; i < soFiles.length; i++) {
-      def abiName = soFiles[i].path.tokenize('/')[-2]
-      def libSize = soFiles[i].length as String
-      sendTaggedMetric('abi_size', libSize, 'type', abiName)
+  def flavors = ['base', 'objectServer']
+  for (def i = 0; i < flavors.size(); i++) {
+    def flavor = flavors[i]
```
zaki50: `['base', 'objectServer'].each { flavor ->` is better?

Jenkinsfile
```diff
@@ -141,24 +150,30 @@ def storeJunitResults(String path) {
...
+
+    def methods = readFile("realm/realm-library/build/outputs/aar/methods${flavor}")
+    sendMetrics('methods', methods, ['flavor':flavor])
+
+    def aarFile = findFiles(glob: "realm/realm-library/build/outputs/aar/realm-android-library-${flavor}-release.aar")[0]
+    sendMetrics('aar_size', aarFile.length as String, ['flavor':flavor])
+
+    def soFiles = findFiles(glob: "realm/realm-library/build/outputs/aar/unzipped${flavor}/jni/*/librealm-jni.so")
+    for (def j = 0; j < soFiles.size(); j++) {
+        def soFile = soFiles[j]
```
zaki50: we can use `each` against `soFiles `

zaki50:
:+1: with minor  comments

Jenkinsfile
emanuelez: Iterators are non-serializable, which is something you want to avoid in Jenkinsfiles, since Jenkins saves the state to disk in order to be able to recover from master reboots. See here: https://github.com/jenkinsci/pipeline-examples/blob/master/docs/BEST_PRACTICES.md

Jenkinsfile
emanuelez: See above

emanuelez:
@cmelchior It's actually doing the right thing now, sending bytes instead of KB. I fixed the scale in Graphana, but we still need to fix the old data

Jenkinsfile
zaki50: :+1:

Jenkinsfile
zaki50: :+1:

PR #3542, opened 2016-10-01
Clean .so files that were created by old build script

nhachicha:
Credit to @zaki50 to add this fix.

After a gradle clean `realm/realm-library/src/main/jniLibs` was not cleared causing the packager to use old generated _.so_  in the AAR instead of the freshly compiled inside `externalNativeBuild` ex `./realm/realm-library/.externalNativeBuild/cmake/objectServerDebug/obj/x86/librealm-jni.so`

without this fix it might include old _.so_ not containing the sync symbols.


nhachicha:
@realm/java

zaki50:
:+1:

Issue #3549, opened 2016-10-03
DynamicRealmObject#setList() causes native crash if the specified field is not a List


zaki50:
Calling `DynamicRealmObject#setList()` against an object from `DynamicRealm` causes a native crash if the field type is not a `List`.

zaki50:
fixed a test in #3550 to reproduce this issue.

zaki50:
This issue is related to #3404, but it needs more fix.

Current `CheckedRow#getLinkList(long)` does not check field type.
It's completely wrong and we need to fix it.

zaki50:
fixed in  2.0.1

PR #3551, opened 2016-10-03
Always use CheckedRow when creating DynamicRealmObject

### LINKED ISSUE #3404: DynamicRealmObject from RealmResults is using UncheckedRow
https://github.com/realm/realm-java/blob/v1.2.0/realm/realm-library/src/main/java/io/realm/BaseRealm.java#L590 sets `UncheckedRow` even if the `result` is `DynamicRealmObject`.

It causes unexpected crash when an inappropriate getter or setter is used.

Correction: crash was fixed in #3312 but using unchecked row is still a mistake.

### code to reproduce

```
DynamicRealm dynamicRealm = DynamicRealm.getInstance(conf);
dynamicRealm.where("Person").findFirst().getString("age"); // getString for INTEGER field
```

Realm version(s): current master, 1.2.0

Android Studio version: any

Which Android version and device: any

zaki50:
fixes #3404 

DynamicRealmObject should not have `UncheckedRow` instance as a `row`. This PR fix that issue.

And this also removes a check of the `row` instance if it is an instance of `UncheckedRow` in some constructors of DynamicRealmObject since CheckedRow is never passed to it.

### DIFF
--- a/realm/realm-library/src/main/java/io/realm/BaseRealm.java
+++ b/realm/realm-library/src/main/java/io/realm/BaseRealm.java
@@ -518,14 +518,14 @@ abstract class BaseRealm implements Closeable {
     // Used by RealmList/RealmResults
     // Invariant: if dynamicClassName != null -> clazz == DynamicRealmObject
     <E extends RealmModel> E get(Class<E> clazz, String dynamicClassName, long rowIndex) {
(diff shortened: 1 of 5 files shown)


zaki50:
This PR is required to fix #3549 

PR #3564, opened 2016-10-05
Add instructions for using the ROS

radu-tutueanu:
@cmelchior @emanuelez

This is the file included in the RMP bundle as `Android.md`. I figured this is the place it should Iive in.

### DIFF
--- a/OBJECT-SERVER.md
+++ b/OBJECT-SERVER.md
@@ -1,0 +1,27 @@
+# Using Realm Object Server on Android
+
+
+1) Add the following to the top-level `build.gradle` file:
+
+
+```groovy
+buildscript {
+repositories {
+jcenter()
+}
+dependencies {
+classpath "io.realm:realm-gradle-plugin:2.0.0"
+}
+}
+```
+
+
+2) Add the following to the app folders `build.gradle` file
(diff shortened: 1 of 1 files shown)


OBJECT-SERVER.md
```diff
@@ -0,0 +1,27 @@
...
+1) Add the following to the top-level `build.gradle` file:
+
+
+```groovy
+buildscript {
+repositories {
+jcenter()
+}
+dependencies {
+classpath "io.realm:realm-gradle-plugin:2.0.0"
```
Zhuinden: Update to 2.0.1

emanuelez: This is going to be a pain to keep updated so I'd prefer to just have a placeholder instead. Thoughts?

Zhuinden: Well, `classpath "io.realm:realm-gradle-plugin:<version>` would also make sense.

That's how the `-SNAPSHOT` is shown as an example.

radu-tutueanu:
Closing this in favour of a webloc pointing to https://realm.io/docs/java/latest/#enabling-realm-mobile-platform

PR #3573, opened 2016-10-06
Disable flaky test

cmelchior:
Life is to short for restarting CI all the time. Disabling this for now until we can figure out what is wrong with it.

/cc @nhachicha

PR #3579, opened 2016-10-06
OrderedRealmCollection.first()/last() can return a default value instead of throwing when the collection is empty

### LINKED ISSUE #3399: OrderedRealmCollection.first()/last() and default values
From http://stackoverflow.com/questions/35287974/android-realm-prevent-exception-with-first-method-and-empty-database/35314169?noredirect=1#comment65982070_35314169

Right now `RealmCollection.first()/last()` will throw `IndexOfOutRangeException` if called on an empty collection. This means that anyone using these API's have to write annoying boilerplate code around it like:

```
RealmResults<Foo> results = query.findAllSorted("endDate", Sort.DESCENDING);
Foo obj = (result.size() > 0) ? result.first() : null;
```

Instead we should make it easier to provide a default value in case of an empty list, this could e.g be:

```
RealmResults<Foo> results = query.findAllSorted("endDate", Sort.DESCENDING);
Foo obj = result.firstOrNull(); // Option A
Foo obj2 = result.first(null); // Option B, provide your own default value
```

Option B is more general, but I am not sure how useful it would be as `null` is probably used the majority of times.

Option A expresses the intent more clearly and is easier to autocomplete.

maxfurman:
Resolves #3399 

This PR adds overloads for the `.first` and `.last`methods to accept an object of the same type as the collection (or null) to be returned if the collection is empty.


realm/realm-library/src/androidTest/java/io/realm/OrderedRealmCollectionTests.java
```diff
@@ -221,6 +221,14 @@ public void first_emptyCollection() {
     }
 
     @Test
+    public void first_withDefault() {
+        AllJavaTypes defaultObject = collection.get(0);
```
cmelchior: Nitpick, but instead of getting an object from a collection defined outside this method you could just create a standalone object.

realm/realm-library/src/main/java/io/realm/OrderedRealmCollection.java
```diff
@@ -35,6 +35,13 @@
     E first();
 
     /**
+     * Gets the first object from the collection. If the collection is empty, the provided default will be used.
```
cmelchior: be used instead.

realm/realm-library/src/main/java/io/realm/OrderedRealmCollection.java
```diff
@@ -43,6 +50,13 @@
     E last();
 
     /**
+     * Gets the last object from the collection. If the collection is empty, the provided default will be used.
```
cmelchior: will be used instead.

realm/realm-library/src/main/java/io/realm/OrderedRealmCollection.java
```diff
@@ -35,6 +35,13 @@
     E first();
 
     /**
+     * Gets the first object from the collection. If the collection is empty, the provided default will be used.
+     *
+     * @return the first object or the provided default.
+     */
+    E first(E defaultObject);
```
cmelchior: This name should match the variable name in `RealmList`, so either both should be `defaultObject` or `defaultValue`.... I probably lean towards defaultValue since these collections might also support boxed primitive types down the line.

realm/realm-library/src/main/java/io/realm/OrderedRealmCollection.java
```diff
@@ -43,6 +50,13 @@
     E last();
 
     /**
+     * Gets the last object from the collection. If the collection is empty, the provided default will be used.
+     *
+     * @return the last object or the provided default.
+     */
+    E last(E defaultObject);
```
cmelchior: same

realm/realm-library/src/main/java/io/realm/RealmList.java
```diff
@@ -468,6 +468,21 @@ public E first() {
...
     /**
+    * {@inheritDoc}
+    */
+    public E first(E defaultValue) {
+        if (managedMode) {
+            checkValidView();
+            if (!view.isEmpty()) {
+                return get(0);
+            }
+        } else if (unmanagedList != null && unmanagedList.size() > 0) {
```
cmelchior: `!isEmpty()` instead of `size() > 0` ?

realm/realm-library/src/main/java/io/realm/RealmList.java
```diff
@@ -485,6 +500,21 @@ public E last() {
     /**
      * {@inheritDoc}
      */
+    public E last(E defaultValue) {
+        if (managedMode) {
+            checkValidView();
+            if (!view.isEmpty()) {
+                return get((int) view.size() - 1);
+            }
+        } else if (unmanagedList != null && unmanagedList.size() > 0) {
```
cmelchior: Same

realm/realm-library/src/main/java/io/realm/RealmResults.java
```diff
@@ -236,6 +236,18 @@ public E first() {
      * {@inheritDoc}
      */
     @Override
+    public E first(E defaultValue) {
+        if (size() > 0) {
```
cmelchior: `!isEmpty` ?

realm/realm-library/src/main/java/io/realm/RealmResults.java
```diff
@@ -249,6 +261,19 @@ public E last() {
      * {@inheritDoc}
      */
     @Override
+    public E last(E defaultValue) {
+        int size = size();
+        if (size > 0) {
```
cmelchior: Same

realm/realm-library/src/androidTest/java/io/realm/OrderedRealmCollectionTests.java
maxfurman: I thought to do that at first, but I wasn't sure how to determine and instantiate the class contained in the collection within the test.

cmelchior: In this case just doing `new AllJavaTypes()` should work I think? Since the default value is not manipulated in any way?

maxfurman:
I don't have access to the CI, so I can't see why it failed. @cmelchior Could you please either give me read access, or look into the test failure for me? Thanks.

zaki50:
CI failed because of a flaky test.
That was disabled by #3573
Could you merge current master into this branch?

realm/realm-library/src/main/java/io/realm/RealmList.java
```diff
@@ -456,30 +456,62 @@ public E get(int location) {
...
+    }
+
+    /**
+    * {@inheritDoc}
+    */
+    public E first(E defaultValue) {
+        return firstImpl(false, defaultValue);
+    }
+
+    private E firstImpl(boolean shouldThrow, E defaultValue) {
```
cmelchior: I would simplify this by just using `defaultValue != null` as a substitute for `shouldThrow`. No reason to create a boolean for that.

realm/realm-library/src/main/java/io/realm/RealmList.java
```diff
@@ -456,30 +456,62 @@ public E get(int location) {
...
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public E last(E defaultValue) {
+        return lastImpl(false, defaultValue);
+    }
+
+    private E lastImpl(boolean shouldThrow, E defaultValue) {
```
cmelchior: same

realm/realm-library/src/main/java/io/realm/RealmResults.java
```diff
@@ -225,10 +225,26 @@ public E get(int location) {
...
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public E first(E defaultValue) {
+        return firstImpl(false, defaultValue);
+    }
+
+    private E firstImpl(boolean shouldThrow, E defaultValue) {
```
cmelchior: same

realm/realm-library/src/main/java/io/realm/RealmResults.java
```diff
@@ -237,11 +253,27 @@ public E first() {
...
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public E last(E defaultValue) {
+        return lastImpl(false, defaultValue);
+        
+    }
+
+    private E lastImpl(boolean shouldThrow, E defaultValue) {
```
cmelchior: same

realm/realm-library/src/main/java/io/realm/RealmList.java
cmelchior: Doh, very good point. Just ignore me then 😄

cmelchior:
Hi @maxfurman 

I think this looks ready to merge. There seems to be a merge conflict with the `CHANGELOG.md` so if you just merge latest master into your branch we can merge this. Alternatively you can just give us edit rights and we can do it as well.

Thanks again for adding this 👍

Issue #3580, opened 2016-10-06
Is `kotlin-reflect` _really_ required?


joshfriend:
#### Goal

I would like to remove 11k methods from my dex count when using Realm in a Kotlin Android application.

#### Expected Results

Documentation should not require me to add `kotlin-reflect` as a dependency.

#### Actual Results

[Documentation states that `kotlin-reflect` is required](https://realm.io/docs/java/latest/#kotlin):

> Many of Realm APIs reference Java classes, and for that reason you must include `org.jetbrains.kotlin:kotlin-reflect:${kotlin_version}` as a compile dependency.

I believe this is false because the ability to get a java `Class<T>` from a Kotlin `KClass<T>` is [present in the Kotlin stdlib](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.jvm/java.html).

I've removed `kotlin-reflect` from my dependencies, ran a clean build, and the application still seemed to function normally.

MGaetan89:
I have been using Realm in my Kotlin application for a couple of months without the `kotlin-reflect` dependency, and it is working as expected.

cmelchior:
Website and example updated.

joshfriend:
Awesome, thank you for confirming!

PR #3588, opened 2016-10-07
Update kotlin example

cmelchior:
Part fix for #3580 

@realm/java

### DIFF
--- a/examples/kotlinExample/src/main/kotlin/io/realm/examples/kotlin/KotlinExampleActivity.kt
+++ b/examples/kotlinExample/src/main/kotlin/io/realm/examples/kotlin/KotlinExampleActivity.kt
@@ -30,11 +30,10 @@ import org.jetbrains.anko.async
 import org.jetbrains.anko.uiThread
 import kotlin.properties.Delegates
 
-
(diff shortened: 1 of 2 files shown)


examples/kotlinExample/src/main/kotlin/io/realm/examples/kotlin/KotlinExampleActivity.kt
```diff
@@ -30,11 +30,10 @@ import org.jetbrains.anko.async
 import org.jetbrains.anko.uiThread
 import kotlin.properties.Delegates
 
-
 class KotlinExampleActivity : Activity() {
 
     companion object {
-        val TAG: String = KotlinExampleActivity::class.qualifiedName as String
+        val TAG: String? = KotlinExampleActivity::class.java.simpleName
```
nhachicha: the class name can't be null? unless there's something wrong with the classloader?
`val TAG: String = KotlinExampleActivity::class.java.simpleName` instead?

cmelchior: Good catch

PR #3592, opened 2016-10-07
test for issur3140

zaki50:
This PR is just for checking if the test fails on CI server or not.

No need to review this for now


zaki50:
I've confirmed that the test failed at least on Nexus 5X.

PR #3602, opened 2016-10-10
Modern exception handling

kneth:
Migrating the old C macro based exception handling to modern C++ lambda functions. Only C++ exception is thrown in the C++ code, and they are converted to Java exception. The idea is that all JNI border functions must wrap all code in `try_catch()`. When writing C++ code (core, object store or JNI), only C++ exceptions should be thrown - no need to "throw" Java exception. The function `try_catch()` takes care of it.

I do dislike big PRs, and this is definitely a big one. Let me give a short guide to the reviewers:

* Focus on `util.hpp` and `util.cpp`
* Take one JNI method to see how it works

@realm/java

### DIFF
--- a/realm/realm-library/src/main/cpp/io_realm_SyncManager.cpp
+++ b/realm/realm-library/src/main/cpp/io_realm_SyncManager.cpp
@@ -54,9 +54,7 @@ public:
             case Level::fatal: log_method = log_fatal; break;
             case Level::all:
             case Level::off:
-                ThrowException(env, IllegalArgument,
-                        util::format("Unknown logger argument: %s.", util::Logger::get_level_prefix(level)));
-                return;
+                throw invalid_argument(util::format("Unknown logger argument: %s.", util::Logger::get_level_prefix(level)));
         }
         log_message(env, log_method, msg.c_str());
     }
@@ -97,9 +95,9 @@ JNIEXPORT void JNICALL Java_io_realm_SyncManager_nativeInitializeSyncClient
     (JNIEnv *env, jclass sync_manager_class)
 {
     TR_ENTER(env)
-    if (sync_client) return;
+    try_catch<void>(env, [&]() {
+        if (sync_client) return;
 
-    try {
         AndroidLogger::shared().set_level_threshold(util::Logger::Level::warn);
 
         sync::Client::Config config;
@@ -111,35 +109,36 @@ JNIEXPORT void JNICALL Java_io_realm_SyncManager_nativeInitializeSyncClient
         sync_manager_notify_error_handler = env->GetStaticMethodID(sync_manager,
                                                                    "notifyErrorHandler", "(ILjava/lang/String;)V");
         sync_client->set_error_handler(error_handler);
-    } CATCH_STD()
+    });
 }
 
 // Create the thread from java side to avoid some strange errors when native throws.
 JNIEXPORT void JNICALL
 Java_io_realm_SyncManager_nativeRunClient(JNIEnv *env, jclass)
 {
-    try {
+    try_catch<void>(env, [&]() {
         sync_client->run();
-    } CATCH_STD()
+    });
 }
 
 JNIEXPORT void JNICALL
 Java_io_realm_SyncManager_nativeSetSyncClientLogLevel(JNIEnv* env, jclass, jint logLevel)
 {
-    util::Logger::Level native_log_level;
-    switch(logLevel) {
-        case io_realm_log_LogLevel_ALL: native_log_level = util::Logger::Level::all; break;
-        case io_realm_log_LogLevel_TRACE: native_log_level = util::Logger::Level::trace; break;
-        case io_realm_log_LogLevel_DEBUG: native_log_level = util::Logger::Level::debug; break;
-        case io_realm_log_LogLevel_INFO: native_log_level = util::Logger::Level::info; break;
-        case io_realm_log_LogLevel_WARN: native_log_level = util::Logger::Level::warn; break;
-        case io_realm_log_LogLevel_ERROR: native_log_level = util::Logger::Level::error; break;
-        case io_realm_log_LogLevel_FATAL: native_log_level = util::Logger::Level::fatal; break;
-        case io_realm_log_LogLevel_OFF: native_log_level = util::Logger::Level::off; break;
-        default:
-            ThrowException(env, IllegalArgument, "Invalid log level: " + logLevel);
-            return;
-    }
-    // FIXME: This call is not thread safe. Switch to OS implementation to make it thread safe.
-    AndroidLogger::shared().set_level_threshold(native_log_level);
+    try_catch<void>(env, [&]() {
+        util::Logger::Level native_log_level;
+        switch(logLevel) {
+            case io_realm_log_LogLevel_ALL: native_log_level = util::Logger::Level::all; break;
+            case io_realm_log_LogLevel_TRACE: native_log_level = util::Logger::Level::trace; break;
+            case io_realm_log_LogLevel_DEBUG: native_log_level = util::Logger::Level::debug; break;
+            case io_realm_log_LogLevel_INFO: native_log_level = util::Logger::Level::info; break;
+            case io_realm_log_LogLevel_WARN: native_log_level = util::Logger::Level::warn; break;
+            case io_realm_log_LogLevel_ERROR: native_log_level = util::Logger::Level::error; break;
+            case io_realm_log_LogLevel_FATAL: native_log_level = util::Logger::Level::fatal; break;
+            case io_realm_log_LogLevel_OFF: native_log_level = util::Logger::Level::off; break;
+            default:
+                throw invalid_argument("Invalid log level: " + logLevel);
+        }
+        // FIXME: This call is not thread safe. Switch to OS implementation to make it thread safe.
+        AndroidLogger::shared().set_level_threshold(native_log_level);
+    });
 }
--- a/realm/realm-library/src/main/cpp/io_realm_internal_CheckedRow.cpp
+++ b/realm/realm-library/src/main/cpp/io_realm_internal_CheckedRow.cpp
@@ -14,6 +14,9 @@
  * limitations under the License.
  */
 
+#include <jni.h>
+#include <exception>
+
 #include "io_realm_internal_CheckedRow.h"
 #include "io_realm_internal_UncheckedRow.h"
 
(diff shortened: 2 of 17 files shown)


realm/realm-library/src/main/cpp/util.hpp
```diff
@@ -58,11 +61,73 @@ JNIEXPORT jint JNICALL JNI_OnLoad(JavaVM *vm, void *reserved);
...
+        }
     }
+    catch (std::exception& e) {
+        // we don't wish to do the exception handling in here; this is a templated function, and we might end up
+        // with multiple copies of the code for no good reason
+        ConvertException(env);
+    }
+
+    // when a Java exception is thrown, we return to Java with any return value; as soon as the execution point is
+    // back at the JVM, the Java exception is thrown, and the returned value isn't used for anything
```
beeender: Don't we have some cases that we need to return `-1` for some cases (eg.: the index)?

realm/realm-library/src/main/cpp/util.hpp
```diff
@@ -58,11 +61,73 @@ JNIEXPORT jint JNICALL JNI_OnLoad(JavaVM *vm, void *reserved);
...
+// Exception handling using lambdas
+// Credit: http://stackoverflow.com/questions/29202377/how-do-i-force-the-compiler-to-recognize-lambda-type-in-templated-function
+template<typename T, typename F>
+T try_catch(JNIEnv* env, F func) {
+    try {
+        // if an exception already has occurred, it is not possible to call any JNI
+        // functions, and we don't wish to try to execute the JNI function
+        // alternatively, we could clear the exception but then we can easily miss
+        // an error stage
+        if (env->ExceptionCheck() == JNI_FALSE) {
```
beeender: I think this checking is not necessary. If this ever happens, then we have a bug in our code, and an exception will be thrown anyway. It doesn't worth to check it all the time especially in some very frequent calls.

realm/realm-library/src/main/cpp/util.cpp
```diff
@@ -43,195 +43,170 @@ jmethodID session_error_handler;
...
-{
-    ostringstream ss;
+jthrowable JThrowableNew(JNIEnv* env, const char* class_name, const char* message) {
+    jclass j_exception_class = env->FindClass(class_name);
+    if (j_exception_class == nullptr) {
+        return nullptr;
+    }
+    jmethodID constructor = env->GetMethodID(j_exception_class, "<init>", "(Ljava/lang/String;)V");
+    if (constructor == nullptr) {
+        env->DeleteLocalRef(j_exception_class);
```
beeender: no need to call this I think? The local ref will be deleted anyway when the code returns the to java land. But it is minor, i am fine with this.

realm/realm-library/src/main/cpp/util.cpp
```diff
@@ -43,195 +43,170 @@ jmethodID session_error_handler;
...
-            break;
-    }
-    if (jExceptionClass != NULL) {
-        env->ThrowNew(jExceptionClass, message.c_str());
-        TR_ERR(env, "Exception has been throw: %s", message.c_str())
+    if (j_exception != nullptr) {
+        env->Throw(j_exception);
+        TR_ERR_NO_VA_ARG(env, "Exception has been throw")
     }
     else {
```
beeender: Something is really wrong here. Throw a native exception?

realm/realm-library/src/main/cpp/util.hpp
```diff
@@ -58,11 +61,73 @@ JNIEXPORT jint JNICALL JNI_OnLoad(JavaVM *vm, void *reserved);
...
+
+// class_not_found - looking on a class in JVM failed; see also java.lang.ClassNotFoundException
+class class_not_found : public std::runtime_error {
+public:
+    class_not_found(const std::string& msg) : std::runtime_error(msg) {};
+    class_not_found(const char* msg) : std::runtime_error(msg) {};
+};
+
+// null_value - trying to set a required field to null
+class null_value : public std::exception {
```
beeender: why `exception` instead of `runtime_error` here?

realm/realm-library/src/main/cpp/util.hpp
```diff
@@ -58,11 +61,73 @@ JNIEXPORT jint JNICALL JNI_OnLoad(JavaVM *vm, void *reserved);
...
+    }
+
+    // when a Java exception is thrown, we return to Java with any return value; as soon as the execution point is
+    // back at the JVM, the Java exception is thrown, and the returned value isn't used for anything
+    return static_cast<T>(0);
+}
+
+// we need to have specialized C++ exceptions so we know what kind of error state has been reached
+// illegal_state - the internal state of the Realm is wrong; it is not serious but execution cannot continue
+class illegal_state : public std::runtime_error {
```
beeender: For those exception names, i think we should use camel case naming. `IllegalState` instead? And others are the same.

beeender:
also, better to go to `releases` branch to avoid future merge conflicts.

realm/realm-library/src/main/cpp/io_realm_SyncManager.cpp
```diff
@@ -97,9 +95,9 @@ JNIEXPORT void JNICALL Java_io_realm_SyncManager_nativeInitializeSyncClient
     (JNIEnv *env, jclass sync_manager_class)
 {
     TR_ENTER(env)
-    if (sync_client) return;
+    try_catch<void>(env, [&]() {
+        if (sync_client) return;
```
bmunkholm: return on separate line

realm/realm-library/src/main/cpp/io_realm_internal_CheckedRow.cpp
```diff
@@ -24,214 +27,213 @@ using namespace realm;
 JNIEXPORT jlong JNICALL Java_io_realm_internal_CheckedRow_nativeGetColumnCount
   (JNIEnv* env, jobject obj, jlong nativeRowPtr)
 {
-    if (!ROW(nativeRowPtr)->is_attached())
-        return 0;
+    return try_catch<jlong>(env, [&]() {
+        Row* row = reinterpret_cast<Row*>(nativeRowPtr);
+        if (!row->is_attached())
+            return static_cast<jlong>(0);
```
bmunkholm: why the cast to return 0?

realm/realm-library/src/main/cpp/io_realm_internal_CheckedRow.cpp
```diff
@@ -24,214 +27,213 @@ using namespace realm;
 JNIEXPORT jlong JNICALL Java_io_realm_internal_CheckedRow_nativeGetColumnCount
   (JNIEnv* env, jobject obj, jlong nativeRowPtr)
 {
-    if (!ROW(nativeRowPtr)->is_attached())
-        return 0;
+    return try_catch<jlong>(env, [&]() {
+        Row* row = reinterpret_cast<Row*>(nativeRowPtr);
```
bmunkholm: Is it deliberate to not use the ROW macro for the casts all over in this file?

realm/realm-library/src/main/cpp/io_realm_internal_CheckedRow.cpp
```diff
@@ -24,214 +27,213 @@ using namespace realm;
...
 
 JNIEXPORT jlong JNICALL Java_io_realm_internal_CheckedRow_nativeGetColumnIndex
   (JNIEnv* env, jobject obj, jlong nativeRowPtr, jstring columnName)
 {
-    if (!ROW(nativeRowPtr)->is_attached())
-        return 0;
+    return try_catch<jlong>(env, [&]() {
+        Row *row = reinterpret_cast<Row *>(nativeRowPtr);
+        if (!row->is_attached())
+            return static_cast<jlong>(0);
```
bmunkholm: same here - and likely many other places - why cast 0 to jlong?

realm/realm-library/src/main/cpp/util.hpp
kneth: Methods in `CheckedRow` call methods in `UncheckedRow` and it is primarily for that.

realm/realm-library/src/main/cpp/util.hpp
kneth: Yes,  `get_source_ndx` seems to be the only case. So we don't throw in that situation.

realm/realm-library/src/main/cpp/util.hpp
kneth: Good point. Maybe we should prefix with `Realm`?

realm/realm-library/src/main/cpp/util.hpp
kneth: No particular reason. `logic_error` might actually be better?

realm/realm-library/src/main/cpp/util.hpp
beeender: Does `JavaIllegalState` make sense?

realm/realm-library/src/main/cpp/util.cpp
kneth: You mean, throw `std::runtime_error` if we cannot allocate an exception? Ignoring that silently is probably a bad idea.

realm/realm-library/src/main/cpp/util.hpp
beeender: I have no idea ... We have different categories of exception in both java and c++, perhaps stay with `runtime_error` in JNI? Also, see https://github.com/realm/realm-java/pull/3602#pullrequestreview-3622825 , are there any core exception we want them to be thrown in native code without converting for easier core debugging?

realm/realm-library/src/main/cpp/io_realm_internal_CheckedRow.cpp
kneth: Because of the template. `0` is `int` and we return a `jlong` also. The compiler will not allow multiple return types (and gcc cannot figure out that 0 also can be a `jlong`).

realm/realm-library/src/main/cpp/util.hpp
beeender: Uh, but it still doesn't worth it? In that case, this function shouldn't be called again i think? Can we check the return value when calling method in `UncheckedRow`?

realm/realm-library/src/main/cpp/util.cpp
beeender: yes, `j_exception` should not be `null`. If it is, the process should be terminated. `runtime_error` would be good enough.

kneth:
@beeender Segmentation faults are not exceptions (it's a signal), and we will get the stack in those situations. All the exceptions we catch in here is something the user should be able to fix within his/her app.

realm/realm-library/src/main/cpp/util.hpp
kneth: https://developer.android.com/training/articles/perf-jni.html#exceptions

beeender: Yeah, I know. I am worry about the overhead it adds. But anyway, just check the source code of `ExceptionCheck()`, it seems to be very cheap call. So :+1:

Zhuinden:
@kneth I don't like native crashes. :slightly_frowning_face:  I prefer exceptions.

kneth:
@Zhuinden In the case you cannot allocate/find an exception, it is better to get a native exception than silently ignore the error state ;-)

realm/realm-library/src/main/cpp/util.hpp
kneth: Now I remember. I wish to include a table and a column index as members. `runtime_error` has a string a member to I have to use a more generic class.

realm/realm-library/src/main/cpp/util.hpp
kneth: Won't work as we also "return `void`".

kneth:
I have created a PR against `release`: #3621

realm/realm-library/src/main/cpp/io_realm_internal_Table.cpp
```diff
@@ -15,6 +15,11 @@
  */
 
 #include <sstream>
+#include <exception>
```
beeender: unnecessary header?

realm/realm-library/src/main/cpp/util.cpp
```diff
@@ -43,195 +43,170 @@ jmethodID session_error_handler;
...
+            case realm::RealmFileException::Kind::IncompatibleLockFile:
+                kind_code = io_realm_internal_SharedRealm_FILE_EXCEPTION_KIND_IMCOMPATIBLE_LOCK_FILE;
+                break;
+            case realm::RealmFileException::Kind::FormatUpgradeRequired:
+                kind_code = io_realm_internal_SharedRealm_FILE_EXCEPTION_KIND_FORMAT_UPGRADE_REQUIRED;
+                break;
+        }
+        j_exception = JThrowableNew(env, realm_file_exception, e.what(), kind_code);
+    }
+    catch (null_value& e) {
```
beeender: Better to just overwrite the `null_value`'s `what` method.

realm/realm-library/src/main/cpp/util.hpp
beeender: need a overwritten `what()` method i think.

PR #3609, opened 2016-10-11
[CherryPick] Accelerate build with ccache and lcache (#3523)

beeender:
See https://github.com/beeender/lcache for source code for lcache.
It is very simple right now. It identify if the link target is cached by
using checksums from command line + checksums of all input files.

The realm-library gradle check those paths from project properties first
, then the system env.

So those can be set in the ~/.gradle/gradle.properties like:
ccachePath=/usr/bin/ccache
lcachePath=/usr/bin/lcache

Or by system ENVs like:
NDK_CCACHE=/usr/bin/ccache
NDK_LCACHE=/usr/bin/lcache


beeender:
My mistake, this should be committed to `releases` branch. Cherry pick it from master.

PR #3610, opened 2016-10-11
Use debug log level for transformer

### LINKED ISSUE #3608: Reduce Realm Transform Log Verbosity
Right now realm transform is taking a long time to process and a small part of that is the amount of print messages causing excess cpu usage on android studio. It is currently printing field/tokens for every single class/field even if they are not actually transformed.  One option is to only print those fields/tokens if they are  transformed. 

```
Behavior: nextToken
      Field being accessed: com.fasterxml.jackson.core.filter.FilteringParserDelegate._exposedContext
      Field being accessed: com.fasterxml.jackson.core.filter.FilteringParserDelegate._currToken
      Field being accessed: com.fasterxml.jackson.core.filter.FilteringParserDelegate._headContext
      Field being accessed: com.fasterxml.jackson.core.filter.FilteringParserDelegate._exposedContext
      Field being accessed: com.fasterxml.jackson.core.filter.FilteringParserDelegate.delegate
      Field being accessed: com.fasterxml.jackson.core.filter.FilteringParserDelegate._currToken
      Field being accessed: com.fasterxml.jackson.core.filter.FilteringParserDelegate._headContext
```
Realm Transform time: 6608 milliseconds

beeender:
The AS 2.2.1 seems enable info level log by default which makes the
build slow in the AS.
Close #3608

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ This release is not protocol-compatible with previous versions of the Realm Mobi
 
 * Build error when using Java 7 (#3563).
 
+### Enhancements
+
+* Reduce transformer logger verbosity (#3608).
+
 ## Internal
 
 * Upgraded to Realm Core 2.1.0 / Realm Sync 2.0-BETA. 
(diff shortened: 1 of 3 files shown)


zaki50:
:+1:

PR #3612, opened 2016-10-11
Ignore android.injected.build.abi

beeender:
It is set by AS randomly for some reasons, seems AS will inject ABIs
according to execution target. But it is not having the right ABI values
all the time. So just disable it, only check buildTargetABIs property.


zaki50:
:+1:

PR #3620, opened 2016-10-11
Call notifySessionError from native code

### LINKED ISSUE #3597: SyncSession#notifySessionError() seems to be unused
The comment for `SyncSession#notifySessionError()` says that it is called from `Session.cpp`, but I couldn't find `Session.cpp`.
There is a file `io_realm_internal_objectserver_SyncSession.cpp`, but it doesn't call the method.

If it's not needed, we should remove it.
If it should be called from JNI, we should fix it.

beeender:
Fix #3597

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * Those were not kept by ProGuard: names of native methods not in the `io.realm.internal` package, names of classes used in method signature (#3596).
 * Missing ProGuard configuration for libraries used by Sync extension (#3596).
+* Fixed a bug that error handler is not called when sync session fails (#3597).
 
 ### Enhancements
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 * Those were not kept by ProGuard: names of native methods not in the `io.realm.internal` package, names of classes used in method signature (#3596).
 * Missing ProGuard configuration for libraries used by Sync extension (#3596).
-* Fixed a bug that error handler is not called when sync session fails (#3597).
+* Error handler was not called when sync session failed (#3597).
 
 ### Enhancements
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 * Those were not kept by ProGuard: names of native methods not in the `io.realm.internal` package, names of classes used in method signature (#3596).
 * Missing ProGuard configuration for libraries used by Sync extension (#3596).
+* Fixed a bug that error handler is not called when sync session fails (#3597).
```
dalinaum: It looks like we usually omit `Fixed a bug ~`, so  is `error handler is not called when sync session fails (#3597).` enough?

realm/realm-library/src/main/cpp/objectserver_shared.hpp
```diff
@@ -59,6 +59,11 @@ class JniSession {
             g_vm->AttachCurrentThread(&local_env, nullptr);
             std::string log = num_to_string(error_code) + " " + message.c_str();
             log_message(local_env, log_debug, log.c_str());
```
cmelchior: Do we want to log this here? The default java implementation also log these?

beeender: you are right, there is no need to log it here. It has been logged in the sync as well.

realm/realm-library/src/main/cpp/objectserver_shared.hpp
```diff
@@ -54,11 +54,13 @@ class JniSession {
             }
         };
         auto error_handler = [&, global_obj_ref_tmp](int error_code, std::string message) {
-            // FIXME: Simplify this by moving log_message to AndroidLogger
             JNIEnv *local_env;
             g_vm->AttachCurrentThread(&local_env, nullptr);
-            std::string log = num_to_string(error_code) + " " + message.c_str();
-            log_message(local_env, log_debug, log.c_str());
+            jclass java_session_class = local_env->GetObjectClass(global_obj_ref_tmp);
```
kneth: Can `java_session_class` never be `NULL`?

beeender: never ... otherwise something is really wrong

bmunkholm: @beeender If it's allowed to return null we have to check for it. At least assert().
Also goes for the next method.

PR #3621, opened 2016-10-11
Encapsulate C++ exceptions

kneth:
Superseed #3602 

@realm/java

### DIFF
--- a/realm/realm-library/src/main/cpp/util.cpp
+++ b/realm/realm-library/src/main/cpp/util.cpp
@@ -43,7 +43,7 @@ jmethodID session_error_handler;
 
 void ThrowRealmFileException(JNIEnv* env, const std::string& message, realm::RealmFileException::Kind kind);
 
-jthrowable JThrowableNew(JNIEnv* env, const char* class_name, const char* message) {
+static jthrowable JThrowableNew(JNIEnv* env, const char* class_name, const char* message) {
     jclass j_exception_class = env->FindClass(class_name);
     if (j_exception_class == nullptr) {
         return nullptr;
@@ -59,7 +59,7 @@ jthrowable JThrowableNew(JNIEnv* env, const char* class_name, const char* messag
     return reinterpret_cast<jthrowable>(exception);
 }
 
-jthrowable JThrowableNew(JNIEnv* env, const char* class_name, const char* message, jbyte kind) {
+static jthrowable JThrowableNew(JNIEnv* env, const char* class_name, const char* message, jbyte kind) {
     jclass j_exception_class = env->FindClass(class_name);
     if (j_exception_class == nullptr) {
         return nullptr;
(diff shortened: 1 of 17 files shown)


beeender:
Didn't notice a new PR is created ... added some comments on #3602

bmunkholm:
What's the size and speed implications - if any? (I would assume we same some code - anything noticeable?)
Likely no speed implications - but if it's untested it's unknown :-)

realm/realm-library/src/main/cpp/util.cpp
```diff
@@ -43,195 +43,165 @@ jmethodID session_error_handler;
 
 void ThrowRealmFileException(JNIEnv* env, const std::string& message, realm::RealmFileException::Kind kind);
 
-void ConvertException(JNIEnv* env, const char *file, int line)
-{
-    ostringstream ss;
+jthrowable JThrowableNew(JNIEnv* env, const char* class_name, const char* message) {
```
beeender: should be local static method.

realm/realm-library/src/main/cpp/util.cpp
```diff
@@ -43,195 +43,165 @@ jmethodID session_error_handler;
...
+        env->DeleteLocalRef(j_exception_class);
+        return nullptr;
+    }
+    jstring j_str = env->NewStringUTF(message);
+    jobject exception = env->NewObject(j_exception_class, constructor, j_str);
+    env->DeleteLocalRef(j_exception_class);
+    return reinterpret_cast<jthrowable>(exception);
+}
+
+jthrowable JThrowableNew(JNIEnv* env, const char* class_name, const char* message, jbyte kind) {
```
beeender: should be local static method.

realm/realm-library/src/main/cpp/util.cpp
```diff
@@ -43,195 +43,165 @@ jmethodID session_error_handler;
...
+    const char index_out_of_bound_exception[]    = "java/lang/ArrayIndexOutOfBoundsException";
+    const char unsupported_operation_exception[] = "java/lang/UnsupportedOperationException";
+    const char class_not_found_exception[]       = "java/lang/ClassNotFoundException";
+    const char runtime_exception[]               = "java/lang/RuntimeException";
+
+    jthrowable j_exception;
+
+    // we are going to "throw" a Java exception so it is safest to clear
+    // already occurred exceptions
+    env->ExceptionClear();
```
beeender: If there is a Java exception before, we'd better either just return to throw the old one / or do nothing just wait for a jni crash. I'd like the just have a JNI crash since it is a bug from our side and it should not happen.

Hide the old java exception may mislead our debbuging a lot.

realm/realm-library/src/main/cpp/util.cpp
```diff
@@ -43,195 +43,165 @@ jmethodID session_error_handler;
...
+        return nullptr;
+    }
+    jstring j_str = env->NewStringUTF(message);
+    jobject exception = env->NewObject(j_exception_class, constructor, kind, j_str);
+    env->DeleteLocalRef(j_exception_class);
+    return reinterpret_cast<jthrowable>(exception);
+}
+
+void ConvertException(JNIEnv* env) {
+    const char out_of_memory_error[]             = "io/realm/internal/OutOfMemoryError";
```
beeender: `static const char` is better maybe? Although compiler may generate same code.

realm/realm-library/src/main/cpp/util.cpp
kneth: "Fail early" is probably better.

kneth:
@bmunkholm The results from the microbenchmark is shown below. The variance in the microbenchmark is big (20-30 %), and I have the Android emulator as test device. It seems that it is a bit slower to change the exception handling but we are also doing more work (checking is an exception has occurred earlier). For the result reported in nanoseconds (ns), I am not convinced that the Android timer is making much sense.

Test | PR | `releases` branch
-----|----|-------------------
read double |397 ns |274 ns
read long  | 356 ns |246 ns
read string |817 ns| 735 ns
write double |1.54 us| 1.15 us
write long| 1.38 us| 1.45 us
write string |2.31 us| 2.06 us
containsQuery |55.19 us |55.88 us
count| 8.15 us| 5.84 us
findAll |60.60 us |63.02 us
findAllSortedOneField |2.67 ms| 2.67 ms
average| 2.87 us| 2.47 us
get| 8.89 us |8.90 us
max| 3.73 us| 3.31 us
min |3.44 us| 3.82 us
size |346 ns |251 ns
sort |2.70 ms |2.63 ms
sum |2.88 us| 2.44 us

emanuelez:
Should somebody from core take a look at this as well?

emanuelez:
Also, how about the difference in size?

kneth:
Arch | PR | `releases` branch
----|----|-------------------
arm64-v8a | 46087696 | 45278448
armeabi-v7a | 32083420 | 31426548
mips | 39590140 | 38861044
x86 | 37743208 | 36996564
x86_64 | 45917696 | 45114016

beeender:
@kneth That is unstripped so, right?

kneth:
@beeender Yes, it's debug.

kneth:
By `static inline` the the new `try_catch()` method, is it possible to reduce the size a bit - see the updated tables above.

finnschiermer:
:+1:
But kind of scary that it adds ~2% to total size, isn't it?

emanuelez:
I added size increase percentages to the table

kneth:
I move it back to "In Progress" as the increase of size and the decrease of performance is a concern. I'll try to optimize it further in the near future.

dalinaum:
@kneth Should it be moved to "In Progress" again?

kneth:
@dalinaum Well, I am not working on it currently so "In Progress" is not really true ;-)

cmelchior:
After internal architecture review, we are parking this until we are able to unwind c++ exceptions correctly so we are not losing information when throwing Java exceptions from C++. @kneth will investigate and/or reach or to the NDK team as they are apparently working on this.

kneth:
A new, unified implementation of a library to unwind the C++ stack is likely to be in included in NDK r16.

Zhuinden:
Now you just need to figure out how to migrate to R16 from R10e 😅

kneth:
@Zhuinden I hope that we will switching to cmake just before r16 :smile:

kneth:
If we are ever going to do this, this PR will need more care than doing from scratch.

Issue #3639, opened 2016-10-17
Revoke server token on Logout


cmelchior:
ROS Authentication Server now supports revocation. We need to implement that endpoint when the user is logged out locally:

```
//Happy path
POST
<server>/auth/revoke
{ "token" : "<value>" }

Success response: 204
```

cmelchior:
@zaki50 What is the status on this?

cmelchior:
Fixed as part of #3642

PR #3640, opened 2016-10-17
SyncUser.currentUser is now correctly cleared on logout

### LINKED ISSUE #3638: Logging out doesn't remove the `currentUser`
From http://stackoverflow.com/questions/40058676/realm-cant-user-logout

cmelchior:
Fixes #3638  
Also fixed some compile errors. Not entirely sure how they where not caught by CI? 

@realm/java 

Also this impact #3639 (/cc @zaki50 )

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 * Renamed `User` to `SyncUser`, `Credentials` to `SyncCredentials` and `Session` to `SyncSession` to align names with Cocoa.
 
+### Bug fixes
+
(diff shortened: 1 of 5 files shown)


beeender:
I guess the compiling error is not caught by CI is because of the the both PR were verified be any of them merged.

pavelpoley:
> java.lang.IllegalStateException: Current user is not valid if more that one valid, logged-in user exists.

i am call SyncUser.current().logOut(); but when logging in (With different account) again this exception is thrown, with same account not thrown

Realm 5.9.1

Issue #3528, opened 2016-09-28
Exceptions in Custom loggers crash JNI


cmelchior:
Correction: Was caused by a Custom Logger that crashed, but we should provide a much better exception in that case.

```
art/runtime/check_jni.cc:65] JNI DETECTED ERROR IN APPLICATION: JNI GetStringChars called with pending exception 'java.lang.IllegalArgumentException' thrown in unknown throw location
art/runtime/check_jni.cc:65]     in call to GetStringChars
art/runtime/check_jni.cc:65]     from long io.realm.internal.SharedRealm.nativeCreateConfig(java.lang.String, byte[], byte,boolean, boolean, boolean, boolean, java.lang.String, java.lang.String)

art/runtime/check_jni.cc:65]   native: #08 pc 00055915  /data/app/com.invoice2go.dev-1/lib/arm/librealm-jni.so (???)
art/runtime/check_jni.cc:65]   native: #09 pc 0001f537  /data/app/com.invoice2go.dev-1/lib/arm/librealm-jni.so (Java_io_realm_internal_SharedRealm_nativeCreateConfig+54)
art/runtime/check_jni.cc:65]   native: #10 pc 006df8f7  /data/dalvik-cache/arm/data@app@com.invoice2go.dev-1@base.apk@classes.dex (Java_io_realm_internal_SharedRealm_nativeCreateConfig__Ljava_lang_String_2_3BBZZZZLjava_lang_String_2Ljava_lang_String_2+218)
art/runtime/check_jni.cc:65]   at io.realm.internal.SharedRealm.nativeCreateConfig(Native method)
art/runtime/check_jni.cc:65]   at io.realm.internal.SharedRealm.getInstance(SharedRealm.java:166)
art/runtime/check_jni.cc:65]   at io.realm.internal.SharedRealm.getInstance(SharedRealm.java:155)
```

From this configuration in Kotlin:

```
RealmConfiguration.Builder()
                        .name(realmName)
                        .modules(myModule)
                        .deleteRealmIfMigrationNeeded()
                        .build()
```

beeender:
i am making changes on loggers, will fix this together.

PR #3643, opened 2016-10-17
Refactor RealmLog

beeender:
The motivation:
* The previous implementaion for JNI log needs calls like:
  JNI -> Java -> JNI. It is not quite effecient for the most common case
  -- log in JNI and sync.
* The previous log levels are bit confusing.
* util.cpp gets messy.
* The Logger class in Java gets confusing with the Logger from
  core/sync.

So this commit makes below changes:
* Adds cpp namespace jni_impl and jni_util, and make log relevant code
  belong to these two namepsaces.
* Moves the RealmLog to JNI.
* Implenment a default Android logger in native code.
* Add tag support for realm-java Logger. Java side is using tag
  "REALM_JAVA", JNI side is using tag "REALM_JNI" and sync is using tag
  "REALM_CORE".
* All tags share the same log level.
* Some cleanups.

Benchmark:

|            |  Before NokiaX2 - 4.3           | After NokiaX2 - 4.3  |
| ------------- |:-------------:| -----:|
| readDouble      | 2.37 μs | 2.39 μs |
| readLong     | 2.53 μs      |  2.54 μs |
| readString | 10.36 μs      |   9.51 μs |


beeender:
@realm/java This is ready for review. I am checking https://github.com/realm/realm-java/issues/3528 , will update this PR if there is anything needed for that.

realm/realm-library/src/main/cpp/jni_util/log.cpp
```diff
@@ -0,0 +1,189 @@
...
+    JavaVM* m_jvm;
+    // Global ref of the logger object.
+    jobject m_java_logger;
+    jmethodID m_log_method;
+
+    inline JNIEnv* get_current_env() noexcept
+    {
+        JNIEnv *env;
+        if (m_jvm->GetEnv((void **)&env, JNI_VERSION_1_6) != JNI_OK) {
+            m_jvm->AttachCurrentThread(&env, nullptr); // Should never fail
```
nhachicha: this is a no-op if the thread is already attached right ?

realm/realm-library/src/main/cpp/jni_util/log.cpp
```diff
@@ -0,0 +1,189 @@
...
+        std::lock_guard<std::mutex> lock(m_mutex);
+        for (auto& logger : m_loggers) {
+            logger->log(level, tag, throwable, stacktrace, message);
+        }
+    }
+}
+
+void CoreLoggerBridge::do_log(realm::util::Logger::Level level, std::string msg)
+{
+    // Ignore the leve threshold from the root logger.
```
nhachicha: ..the level..

realm/realm-library/src/main/cpp/jni_util/log.cpp
```diff
@@ -0,0 +1,189 @@
...
+    Log::Level jni_level;
+    switch (level) {
+        case Level::trace: jni_level = Log::trace; break;
+        case Level::debug: // Map to same debug level.
```
nhachicha: don't we need a `break` or we will fall down to `default`?
same for `all` & `off`

realm/realm-library/src/main/cpp/jni_util/log.hpp
```diff
@@ -0,0 +1,193 @@
...
+        trace = io_realm_log_LogLevel_TRACE,
+        debug = io_realm_log_LogLevel_DEBUG,
+        info = io_realm_log_LogLevel_INFO,
+        warn = io_realm_log_LogLevel_WARN,
+        error = io_realm_log_LogLevel_ERROR,
+        fatal = io_realm_log_LogLevel_FATAL,
+        off = io_realm_log_LogLevel_OFF
+    };
+
+    // Add & Remove a Java Logger. An Java logger needs to be implemented from io.realm.log.Logger interface.
```
nhachicha: A Java

realm/realm-library/src/main/java/io/realm/log/Logger.java
```diff
@@ -16,75 +16,23 @@
...
-     * @param message optional additional message.
-     * @param args optional arguments used to format the message using {@link String#format(String, Object...)}.
-     */
-    void error(Throwable throwable, String message, Object... args);
-
-    /**
-     * Handles a {@link LogLevel#FATAL} event.
+     * Handles a log event
      *
+     * @param level  for this log event. It can only be a value between {@link LogLevel#TRACE} and
```
nhachicha: remove extra space

realm/realm-library/src/main/java/io/realm/log/RealmLog.java
```diff
@@ -29,12 +27,8 @@
 @Keep
 public final class RealmLog {
 
-    private static final Logger[] NO_LOGGERS = new Logger[0];
-
-    // All of the below should be modified together under under a lock on LOGGERS.
-    private static final List<Logger> LOGGERS = new ArrayList<>();
-    private static volatile Logger[] loggersAsArray = NO_LOGGERS;
-    private static int minimumNativeLogLevel = Integer.MAX_VALUE;
+    @SuppressWarnings("FieldCanBeLocal")
```
nhachicha: why is this accessed from JNI?

beeender: This is not. just to suppress the lint warning that AS asks to move it to a local var.

realm/realm-library/src/androidTest/java/io/realm/RealmLogTests.java
```diff
@@ -0,0 +1,53 @@
...
+    public void setUp() {
+        Realm.init(InstrumentationRegistry.getTargetContext());
+    }
+
+    @Test
+    public void add_remove() {
+        TestHelper.TestLogger testLogger = new TestHelper.TestLogger();
+        RealmLog.add(testLogger);
+        RealmLog.fatal("TEST");
+        assertEquals("TEST", testLogger.message);
```
beeender: log lever is checked in the `set_get_logLevel` test

realm/realm-library/src/main/cpp/jni_impl/android_logger.cpp
```diff
@@ -0,0 +1,95 @@
...
+    static std::shared_ptr<AndroidLogger> android_logger(new AndroidLogger());
+    return android_logger;
+}
+
+void AndroidLogger::log(Log::Level level, const char* tag, jthrowable, const char* stacktrace,
+        const char* message) {
+    android_LogPriority android_log_priority;
+    switch (level) {
+        case Log::Level::trace:
+            android_log_priority = ANDROID_LOG_VERBOSE;
```
beeender: yes!

realm/realm-library/src/main/cpp/jni_util/log.cpp
```diff
@@ -0,0 +1,193 @@
...
+{
+    std::lock_guard<std::mutex> lock(m_mutex);
+    for (auto it = m_loggers.begin(); it != m_loggers.end(); ++it) {
+        if (*it == logger) {
+            m_loggers.erase(it);
+        }
+    }
+}
+
+void Log::clear_loggers()
```
beeender: This will remove the default logger as well and user has no way to add it back. What do you expect? Keep the default logger when this called? or add another method `add_default_logger()`?

realm/realm-library/src/main/cpp/jni_util/log.hpp
```diff
@@ -0,0 +1,193 @@
...
+    // Used by JavaLogger.
+    JniLogger(bool is_java_logger);
+    // Indicate if this is a wrapper for Java Logger class. See JavaLogger
+    bool m_is_java_logger;
+
+protected:
+    // Overwrite this method to handle the log event.
+    // throwable is the Throwable passed from Java which could be null.
+    // statcktrace is the stracktrace string parsed from throwable by the Java layer which could be an empty string.
+    virtual void log(Log::Level level, const char* tag, jthrowable throwable, const char* stacktrace,
```
cmelchior: Why having both the throwable and the stacktrace?

realm/realm-library/src/main/cpp/util.cpp
```diff
@@ -94,7 +95,7 @@ void ThrowException(JNIEnv* env, ExceptionKind exception, const std::string& cla
     string message;
     jclass jExceptionClass = NULL;
 
-    TR_ERR(env, "jni: ThrowingException %d, %s, %s.", exception, classStr.c_str(), itemStr.c_str())
+    Log::e("jni: ThrowingException %1, %2, %3.", exception, classStr.c_str(), itemStr.c_str());
```
cmelchior: Good idea 👍

realm/realm-library/src/main/java/io/realm/log/AndroidLogger.java
```diff
@@ -1,150 +0,0 @@
...
- *     <td>{@link LogLevel#DEBUG}</td><td>{@link Log#DEBUG}</td>
- *     <td>{@link LogLevel#INFO}</td><td>{@link Log#INFO}</td>
- *     <td>{@link LogLevel#WARN}</td><td>{@link Log#WARN}</td>
- *     <td>{@link LogLevel#ERROR}</td><td>{@link Log#ERROR}</td>
- *     <td>{@link LogLevel#FATAL}</td><td>{@link Log#ERROR}</td>
- *     <td>{@link LogLevel#OFF}</td><td>Not supported. Remove the logger instead.</td>
- * </tr>
- * </table>
- */
-public class AndroidLogger implements Logger {
```
cmelchior: This class is in the public API, I don't think we can remove it in a minor release.

realm/realm-library/src/main/java/io/realm/log/Logger.java
```diff
@@ -16,75 +16,23 @@
 
 package io.realm.log;
 
+import io.realm.internal.KeepMember;
+
 /**
  * Interface for custom loggers that can be registered at {@link RealmLog#add(Logger)}.
  * The different log levels are described in {@link LogLevel}.
  */
 public interface Logger {
```
cmelchior: This interface is in the public API, we cannot just change it in a minor release without warning :(

realm/realm-library/src/main/cpp/jni_util/log.hpp
beeender: the idea is we parse the stacktrace in java and send it o JNI then JNI doesn' have to get stacktrace string. So the question is do we need to pass the `Throwable` to the customized logger?

realm/realm-library/src/main/cpp/jni_impl/android_logger.cpp
```diff
@@ -0,0 +1,95 @@
...
+            android_log_priority = ANDROID_LOG_ERROR;
+            break;
+        case Log::Level::fatal:
+            android_log_priority = ANDROID_LOG_FATAL;
+            break;
+        default:
+            android_log_priority = ANDROID_LOG_VERBOSE; // Cannot get here. Remove the warning only.
+            break;
+    }
+    if (stacktrace) {
```
beeender: Currently no. I would like to only have the message there. And format the message as stacktrace when the throwable existing?

cmelchior: 👍

realm/realm-library/src/main/cpp/jni_util/log.hpp
cmelchior: I think we need to parse the Throwable to custom loggers, yes. Especially since they are prime target for more validation/formatting in a custom logger.

realm/realm-library/src/main/cpp/jni_util/log.cpp
beeender: I will just keep the default logger when this gets called. Cannot figure out a case that why the default logger should be removed.

cmelchior: What if you want to set the debugging level low and redirect them to a file? Then you would still be flooding LogCat?

cmelchior: What if you want to set the debugging level low and redirect them to a file? Then you would still be flooding LogCat?

cmelchior: What if you want to set the debugging level low and redirect them to a file? Then you would still be flooding LogCat?

cmelchior: What if you want to set the debugging level low and redirect them to a file? Then you would still be flooding LogCat?

beeender: Hmmm. OK, then add a method `RealmLog.registerDefaultLogger()`? the problem is i have test case to call `RealmLog.clear()`  removes the default logger which might have impacts on other test cases.

cmelchior: That sounds like a test problem that shouldn't really leak into the API?  What about just re-adding the logger after the test?

beeender:
@realm/java help to review again.

realm/realm-library/src/main/java/io/realm/log/RealmLog.java
```diff
@@ -16,83 +16,154 @@
...
-    // All of the below should be modified together under under a lock on LOGGERS.
-    private static final List<Logger> LOGGERS = new ArrayList<>();
-    private static volatile Logger[] loggersAsArray = NO_LOGGERS;
-    private static int minimumNativeLogLevel = Integer.MAX_VALUE;
+    /**
+     * To convert the old {@link Logger} to the new {@link RealmLogger}.
+     */
+    private static class LoggerAdaptor implements RealmLogger {
+        private Logger logger;
+        private static final Map<Logger, LoggerAdaptor> loggerMap = new HashMap<Logger, LoggerAdaptor>();
```
zaki50: Shouldn't we use `IdentityHashMap` instead of `HashMap`?

beeender: Yes, you are right!

realm/realm-library/src/androidTest/java/io/realm/RealmLogTests.java
```diff
@@ -0,0 +1,157 @@
...
+        public void trace(Throwable throwable, String message, Object... args) {
+        }
+
+        @Override
+        public void debug(Throwable throwable, String message, Object... args) {
+        }
+
+        @Override
+        public void info(Throwable throwable, String message, Object... args) {
+
```
kneth: Please remove empty line

realm/realm-library/src/main/cpp/jni_util/log.cpp
```diff
@@ -0,0 +1,196 @@
...
+    std::lock_guard<std::mutex> lock(m_mutex);
+    if (std::find(m_loggers.begin(), m_loggers.end(), logger) == m_loggers.end()) {
+        m_loggers.push_back(logger);
+    }
+}
+
+void Log::remove_logger(std::shared_ptr<JniLogger> logger)
+{
+    std::lock_guard<std::mutex> lock(m_mutex);
+    for (auto it = m_loggers.begin(); it != m_loggers.end(); ++it) {
```
kneth: Why not use `remove_if()`?

realm/realm-library/src/main/java/io/realm/log/RealmLog.java
```diff
@@ -16,83 +16,154 @@
...
+     * Removes all loggers except the default logger. The default native logger will be removed as well. Use
+     * {@link #registerDefaultLogger()} to add it back.
      */
     public static void clear() {
-        synchronized (LOGGERS) {
-            LOGGERS.clear();
-            setMinimumNativeDebugLevel(Integer.MAX_VALUE);
-            loggersAsArray = NO_LOGGERS;
-        }
+        nativeClearLoggers();
```
zaki50: Shouldn't we clear `LoggerAdaptor#loggerMap` here?

zaki50: Shouldn't we clear `LoggerAdaptor#loggerMap` here?

beeender: Yes! forget that

CHANGELOG.md
```diff
@@ -4,6 +4,12 @@
 
 * Renamed `User` to `SyncUser`, `Credentials` to `SyncCredentials` and `Session` to `SyncSession` to align names with Cocoa.
 
+### Deprecated
+
+* `Logger`. Use `RealmLogger` instead.
+* `AndroidLogger`. The logger for Android is implemented in the native code instead.
```
cmelchior: in native code instead.

realm/realm-library/src/main/cpp/jni_util/log.cpp
cmelchior: @beeender Do you know how expensive this call is if already attached?

realm/realm-library/src/main/java/io/realm/log/AndroidLogger.java
```diff
@@ -36,6 +36,9 @@
  *     <td>{@link LogLevel#OFF}</td><td>Not supported. Remove the logger instead.</td>
  * </tr>
  * </table>
+ *
+ * @deprecated The new {@link RealmLogger} for Android is implemented in native code. Please don't to use this
```
cmelchior: in native code. This class will be removed in a future release.

realm/realm-library/src/main/java/io/realm/log/RealmLog.java
```diff
@@ -16,83 +16,154 @@
...
+    private static String REALM_JAVA_TAG = "REALM_JAVA";
 
-    // All of the below should be modified together under under a lock on LOGGERS.
-    private static final List<Logger> LOGGERS = new ArrayList<>();
-    private static volatile Logger[] loggersAsArray = NO_LOGGERS;
-    private static int minimumNativeLogLevel = Integer.MAX_VALUE;
+    /**
+     * To convert the old {@link Logger} to the new {@link RealmLogger}.
+     */
+    private static class LoggerAdaptor implements RealmLogger {
```
cmelchior: LoggerAdaptor -> LoggerAdapter?

realm/realm-library/src/main/java/io/realm/log/RealmLog.java
```diff
@@ -16,83 +16,154 @@
...
-        synchronized (LOGGERS) {
-            LOGGERS.add(logger);
-            int minimumLogLevel = logger.getMinimumNativeDebugLevel();
-            if (minimumLogLevel < minimumNativeLogLevel) {
-                setMinimumNativeDebugLevel(minimumLogLevel);
-            }
-            loggersAsArray = LOGGERS.toArray(new Logger[LOGGERS.size()]);
+        nativeAddLogger(logger);
+    }
+
```
cmelchior: Remove whitespace.

realm/realm-library/src/main/java/io/realm/log/RealmLog.java
```diff
@@ -16,83 +16,154 @@
...
+        synchronized (LoggerAdaptor.class) {
+            add(new LoggerAdaptor(logger));
         }
     }
 
-    private static void setMinimumNativeDebugLevel(int nativeDebugLevel) {
-        minimumNativeLogLevel = nativeDebugLevel;
-        Util.setDebugLevel(nativeDebugLevel); // Log level for Realm Core
+    /**
+     * Sets the current {@link LogLevel}.
```
cmelchior: Perhaps specify that this will effect all loggers?

`Sets the current LogLevel. Setting this will affect all registered loggers.` ?

realm/realm-library/src/main/java/io/realm/log/RealmLog.java
```diff
@@ -16,83 +16,154 @@
...
             }
-            setMinimumNativeDebugLevel(newMinLevel);
-            loggersAsArray = LOGGERS.toArray(new Logger[LOGGERS.size()]);
         }
         return true;
     }
 
     /**
-     * Remove all loggers.
+     * Removes all loggers except the default logger. The default native logger will be removed as well. Use
```
cmelchior: The first sentence contradicts the 2nd :)

`Remove all loggers.`

realm/realm-library/src/main/java/io/realm/log/RealmLog.java
```diff
@@ -16,83 +16,154 @@
...
-        synchronized (LOGGERS) {
-            LOGGERS.clear();
-            setMinimumNativeDebugLevel(Integer.MAX_VALUE);
-            loggersAsArray = NO_LOGGERS;
-        }
+        nativeClearLoggers();
+    }
+
+    /**
+     * Adds default native logger if it has been removed before.
```
cmelchior: Also specify what happens if this is called twice? Is the default logger added twice or only once?

cmelchior: We should also specify what the default logger actually does i.e logs to LogCat

realm/realm-library/src/main/java/io/realm/log/RealmLogger.java
```diff
@@ -0,0 +1,38 @@
...
+import io.realm.internal.KeepMember;
+
+/**
+ * Interface for custom loggers that can be registered at {@link RealmLog#add(RealmLogger)}.
+ * The different log levels are described in {@link LogLevel}.
+ */
+public interface RealmLogger {
+
+    /**
+     * Handles a log event
```
cmelchior: Missing `.`

realm/realm-library/src/objectServer/java/io/realm/SyncManager.java
```diff
@@ -232,14 +232,15 @@ static void notifyUserLoggedOut(SyncUser user) {
 
     /**
      * Sets the log level for the underlying.
+     *
      * @param logLevel
+     * @deprecated use {@link RealmLog#setLevel(int)} instead.
      */
     public static void setLogLevel(int logLevel) {
```
cmelchior: SyncManager is in Beta. It is OK to just remove this immediately I think.

beeender:
@cmelchior For `GetEnv` vs `AttachCurrentThread` on a thread which has a vm attached:
On Nokia X2 (davik) 470ns vs 576 ns
On Emulator 6.0 (art) 36ns vs 52ns

beeender:
@realm/java Please review again

realm/realm-library/src/main/java/io/realm/log/RealmLog.java
```diff
@@ -152,15 +155,19 @@ public static boolean remove(Logger logger) {
...
+        synchronized (LoggerAdapter.class) {
+            nativeClearLoggers();
+            LoggerAdapter.clear();
+        }
     }
 
     /**
-     * Adds default native logger if it has been removed before.
+     * Adds default native logger if it has been removed before. If the default logger has been registered already,
+     * it won't be added again. The default logger on Android will log to the logcat.
```
cmelchior: will log to Logcat.

PR #3591, opened 2016-10-07
allow to put Realm database file on external storage.

zaki50:
fixes #3140 
fixes #3633

This PR requires realm/realm-object-store#191 to be merged 

TODO:
* [x] update object-store
* [x] tests

PR #3653, opened 2016-10-18
fix temp directory name

zaki50:
I forgot to address this Muling's comment https://github.com/realm/realm-java/pull/3591/files#r83391423

@realm/java

### DIFF
--- a/realm/realm-library/src/main/java/io/realm/Realm.java
+++ b/realm/realm-library/src/main/java/io/realm/Realm.java
@@ -192,7 +192,7 @@ public final class Realm extends BaseRealm {
             defaultConfiguration = new RealmConfiguration.Builder(context).build();
             ObjectServerFacade.getSyncFacadeIfPossible().init(context);
             BaseRealm.applicationContext = context.getApplicationContext();
-            SharedRealm.initialize(new File(context.getFilesDir(), ".realmNamedPipes"));
+            SharedRealm.initialize(new File(context.getFilesDir(), ".realm.temp"));
         }
     }
 


realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -192,7 +192,7 @@ public static synchronized void init(Context context) {
             defaultConfiguration = new RealmConfiguration.Builder(context).build();
             ObjectServerFacade.getSyncFacadeIfPossible().init(context);
             BaseRealm.applicationContext = context.getApplicationContext();
-            SharedRealm.initialize(new File(context.getFilesDir(), ".realmNamedPipes"));
+            SharedRealm.initialize(new File(context.getFilesDir(), ".realm.temp"));
```
cmelchior: Shouldn't we use `context.getCacheDir()` instead?

cmelchior: Also, does this work if multiple Realm files use the same file?

cmelchior: This is the doc for it: https://developer.android.com/reference/android/content/Context.html#getCacheDir() 

Which doesn't really answer the question though. My main concern is what happens if a tmp file is backed up and restored I think.

zaki50: @cmelchior named pipe files created in it has the hash of corresponding realm file's path in its name. Hash can collide, but I think it is out of scope of this PR.

Back up and restore is also file since the named pipe does not contain anything in it.

cmelchior: Ok, in that case keeping them in `getFilesDir()` is probably the safest 👍

PR #3666, opened 2016-10-20
Single daemon thread for notification

beeender:
With changes in https://github.com/realm/realm-object-store/pull/197 ,
there is only one thread will be create for listening changes for all
different Realms. But an additional SharedGroup will be created in the
daemon thread for determine which SharedGroup has changed since last
time. That requires some changes in the java side especially the daemon
thread should not be created when Realm.compactRealm called.

realm/realm-object-store#197 needs to be merged first.


diegomontoya:
https://github.com/realm/realm-object-store/pull/197 has been merged. Hopefully this gets reviewed and merged soon.

realm/realm-library/src/main/java/io/realm/internal/SharedRealm.java
```diff
@@ -178,18 +178,20 @@ private SharedRealm(long nativePtr, RealmConfiguration configuration, RealmNotif
         objectServerFacade = null;
     }
 
+    // This will create a SharedRealm with autoChangeNotifications is false,
```
kneth: "with" -> "where"

PR #3678, opened 2016-10-24
Fix memory leak when unsubscribing from RxJava observables.

### LINKED ISSUE #3552: Potential memory leak when creating Observable from Where asObservable.
When doing mRealm.where(Person.class).findAllAsync().asObservable().subscribe(new Subscriber...); in a Fragment and unsubscribing and closing the Realm when leaving a Fragment there still seems to be multiple instances of Realm and RealmResults in the Heap.

I may be using Realm Where asObservable incorrectly but I posted on Stack Overflow and am yet to get an answer. I've created a basic empty project to illustrate the issue here: https://github.com/willjgriff/android-realm-leak

The original SO post with more details is here: http://stackoverflow.com/questions/39758862/possible-realm-memory-leak-in-android-when-creating-observable-from-where-clause

Realm version(s): 2.0.0
Android Studio version: 2.2
Which Android version and device: Tested on Api 18 and 23 on a Genymotion emulator. Tested on Api 23 on a Galaxy S6.

Any help understanding this issue is appreciated.

### LINKED ISSUE #3672: Rx off thread loading memory leak
Hello guys, I will be very grateful for any help in understanding this issue.

I really enjoy the way I can async subscribe to data changes with Rx, but I have next situation - I need to convert RealmObjects to another objects from my project. Because the count of objects is huge, I want to do this conversion in off thread also and subscribe to all this stuff from Main thread.

So I am trying to do something like this:
```kotlin
 val workerThreadScheduler = AndroidSchedulers.from(threadManager?.getDBThread()?.looper)
 subscription = Observable.defer {
            Log.d(LOG_TAG, "create realm in ${Thread.currentThread().name}")
            Observable.just(Realm.getDefaultInstance())
        }.flatMap { realm ->
            realm.where(Person::class.java)
                    .findAll()
                    .asObservable()
                    .doOnSubscribe { Log.d(LOG_TAG, "subscribe in ${Thread.currentThread().name}") }
                    .doOnUnsubscribe {
                        Log.d(LOG_TAG, "unsubscribe and close realm in ${Thread.currentThread().name}")
                        realm.close()
                    }
                    .map { persons ->
                        val emails = ArrayList<String>()
                        for (person in persons) emails.add(person.email)
                        return@map emails
                    }
        }
                .subscribeOn(workerThreadScheduler)
                .unsubscribeOn(workerThreadScheduler)
                .subscribe { emails -> Log.d(LOG_TAG, "got emails ${emails.size}") }
```

Here I am trying to subscribe to RealmResults from my own `HandlerThread` with looper, so that way I can manipulate with RealmObjects in off thread. I subscribe, unsubscribe, create and close Realm instance in this thread, ` .doOnUnsubscribe` method fired.

The code behaves exactly as I want, except all RealmResults are leaking.

Full example project is here https://github.com/gvoltr/android-realm-leak/tree/master/RealmLeakTest

Realm version: 2.0.2

cmelchior:
Fixes #3672 
Fixes #3552 

This PR fixes a bug that would leak all RealmResults/RealmObject instances on any long lived thread. Kind embarrassing since I looked into this a couple of weeks ago and couldn't find anything, but yay for a good repo case.

There is no unit test since the code in question is private and any test would be non-deterministic. If you think a unit test is required, I can move the helper class to our internal package and unit test it from there. That wouldn't have caught this particular case though as this bug was more about calling code and class not agreeing on the protocol for using the class.

@realm/java

### DIFF
--- a/realm/realm-library/src/main/java/io/realm/rx/RealmObservableFactory.java
+++ b/realm/realm-library/src/main/java/io/realm/rx/RealmObservableFactory.java
@@ -326,7 +326,7 @@ public class RealmObservableFactory implements RxObservableFactory {
                 references.put(object, count - 1);
             } else if (count == 1) {
                 references.remove(object);
-            } else (count <= 0) {
+            } else {
                 throw new IllegalStateException("Invalid reference count: " + count);
             }
         }
--- a/realm/realm-library/src/main/java/io/realm/rx/RealmObservableFactory.java
+++ b/realm/realm-library/src/main/java/io/realm/rx/RealmObservableFactory.java
@@ -322,10 +322,12 @@ public class RealmObservableFactory implements RxObservableFactory {
             Integer count = references.get(object);
             if (count == null) {
                 throw new IllegalStateException("Object does not have any references: " + object);
-            } else if (count > 0) {
+            } else if (count > 1) {
                 references.put(object, count - 1);
-            } else {
+            } else if (count == 1){
                 references.remove(object);
+            } else if (count <= 0) {
+                throw new IllegalStateException("Invalid reference count: " + count);
             }
         }
     }
--- a/realm/realm-library/src/main/java/io/realm/rx/RealmObservableFactory.java
+++ b/realm/realm-library/src/main/java/io/realm/rx/RealmObservableFactory.java
@@ -324,9 +324,9 @@ public class RealmObservableFactory implements RxObservableFactory {
                 throw new IllegalStateException("Object does not have any references: " + object);
             } else if (count > 1) {
                 references.put(object, count - 1);
-            } else if (count == 1){
+            } else if (count == 1) {
                 references.remove(object);
-            } else if (count <= 0) {
+            } else (count <= 0) {
                 throw new IllegalStateException("Invalid reference count: " + count);
             }
         }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 * Missing ProGuard configuration for libraries used by Sync extension (#3596).
 * Error handler was not called when sync session failed (#3597).
 * Permission error when a database file is located at external storage (#3140).
+* Memory leak when unsubscribing from a RealmResults/RealmObject RxJava Observable (#3552).
 
 ### Enhancements
 


realm/realm-library/src/main/java/io/realm/rx/RealmObservableFactory.java
```diff
@@ -322,10 +322,12 @@ public void releaseReference(K object) {
             Integer count = references.get(object);
             if (count == null) {
                 throw new IllegalStateException("Object does not have any references: " + object);
-            } else if (count > 0) {
+            } else if (count > 1) {
                 references.put(object, count - 1);
-            } else {
+            } else if (count == 1){
```
beeender: space between `)` and `{`

realm/realm-library/src/main/java/io/realm/rx/RealmObservableFactory.java
```diff
@@ -322,10 +322,12 @@ public void releaseReference(K object) {
...
                 references.remove(object);
+            } else if (count <= 0) {
```
emanuelez: Just `else` here might be better here. The current version might be error prone in case of future refactorings.

realm/realm-library/src/main/java/io/realm/rx/RealmObservableFactory.java
```diff
@@ -324,9 +324,9 @@ public void releaseReference(K object) {
                 throw new IllegalStateException("Object does not have any references: " + object);
             } else if (count > 1) {
                 references.put(object, count - 1);
-            } else if (count == 1){
+            } else if (count == 1) {
                 references.remove(object);
-            } else if (count <= 0) {
+            } else (count <= 0) {
```
Zhuinden: that doesn't look right

cmelchior:
@Zhuinden That is what you get when editing directly from Github 😄 . Fixed.

realm/realm-library/src/main/java/io/realm/rx/RealmObservableFactory.java
```diff
@@ -322,10 +322,12 @@ public void releaseReference(K object) {
...
             if (count == null) {
                 throw new IllegalStateException("Object does not have any references: " + object);
-            } else if (count > 0) {
+            } else if (count > 1) {
                 references.put(object, count - 1);
-            } else {
+            } else if (count == 1) {
                 references.remove(object);
+            } else {
+                throw new IllegalStateException("Invalid reference count: " + count);
```
zaki50: Shouldn't we remove the entry even if the `count` is less than `1`?

cmelchior: I'm not entirely sure what happens if you call `subscription.unsubscribe()` twice though. Testing right now to verify.

zaki50: I also think we should throwing an exception.
Removing the entry before throwing is much better, isn't it?

cmelchior: Wouldn't that send mixed signals when reading the code? If we reach the else state, you are using the class the wrong way (only called by our code btw). Leaking in that case seems fine vs. the code being more clear to read?

Note: Also tested unsubscribing twice. It works fine

PR #3693, opened 2016-10-26
fix a bug that proguard configuration keeps all class names.

### LINKED ISSUE #3689: [WARNING!] Obfuscation is not working with 2.1.0
**Beware, don't release to Google Play with Realm 2.1.0** if you want your code to be obfuscated. Realm 2.1.0 won't allow any of your class names to be obfuscated, even on classes that are not RealmObjects.

I'm using custom dictionaries for it, but it probably does not matter. I tested this multiple times and my code with 2.0.2 is always correctly obfuscated, code with 2.1.0 is not obfuscated at all. Luckily this happened to me once with another library and I'm checking the resulting apk with an apktool before releasing.

zaki50:
Our ProGuard configuration introduced in `2.1.0` unexpectedly keeps classes those do not have any fields with KeepMember annotation.

fixes #3689

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### bug fixes
 
-* ProGuard configuration introduced in 2.1.0 unexpectedly keeps classes those do not have any fields with KeepMember annotation.
+* ProGuard configuration introduced in 2.1.0 unexpectedly keeps classes those do not have any fields with KeepMember annotation (#3689).
 
 ## 2.1.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,8 @@
 ## 2.1.1
 
-### bug fixes
+### Bug fixes
 
-* ProGuard configuration introduced in 2.1.0 unexpectedly keeps classes those do not have any fields with KeepMember annotation (#3689).
+* ProGuard configuration introduced in 2.1.0 unexpectedly kept classes that did not have the @KeepMember annotation (#3689).
 
 ## 2.1.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 2.1.1
+
+### bug fixes
+
+* ProGuard configuration introduced in 2.1.0 unexpectedly keeps classes those do not have any fields with KeepMember annotation.
+
 ## 2.1.0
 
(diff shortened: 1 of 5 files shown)


zaki50:

realm/realm-library/src/main/java/io/realm/log/RealmLogger.java
```diff
@@ -22,6 +22,7 @@
  * Interface for custom loggers that can be registered at {@link RealmLog#add(RealmLogger)}.
  * The different log levels are described in {@link LogLevel}.
  */
+@KeepMember
```
beeender: So the `@KeepMember` has to be applied to the class as well? Is it still necessary to apply `@KeepMember` for the `log` method below?

zaki50: yes.
I'd like to achieve only requiring to add the annotation to methods, but I couldn't find the way.

beeender:
LGTM, but I know almost nothing about proguard :P

CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## 2.1.1
+
+### bug fixes
```
cmelchior: `bug fixes` -> `Bug Fixes`

CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
...
+
+* ProGuard configuration introduced in 2.1.0 unexpectedly keeps classes those do not have any fields with KeepMember annotation (#3689).
```
cmelchior: kept classes that did not have the @KeepMember annotation

realm/realm-library/src/main/java/io/realm/internal/KeepMember.java
```diff
@@ -24,9 +24,9 @@
 /**
  * This annotation is used to mark the fields and methods to be kept by ProGuard/DexGuard.
  * The ProGuard configuration must have '-keep class io.realm.internal.KeepMember'
- * and '-keep,includedescriptorclasses class * { @io.realm.internal.KeepMember *; }'.
+ * and '-keep,includedescriptorclasses @io.realm.internal.KeepMember class * { @io.realm.internal.KeepMember *; }'.
```
cmelchior: Also add that this annotation must be added to class as well to work

zaki50:
All comments are addressed

PR #3703, opened 2016-10-27
Add SortDescriptor to support sort by link's field

beeender:


beeender:
This is still in progress. Still have a few crashes & some docs are missing.

beeender:
There are lots of more work than I expected:
1. A simple wrapper of core's `SortDescriptor` won't work since it holds a pointer to a `Column` and it won't be thread safe for async queries.
2. Holding a array of column indices is not a perfect idea either. The column index could change when the schema changes and it leads to some horrible native crash -- this could happen to the current implementation as well.

So, for the `SortDescriptor` in java side:
1. It should hold a list of field name strings instead of column index.
2. Parse the field names to column index only before every core to `sort`/`distinct`/etc. since for every read transaction, the column index can be different.
3. The async queries are really tricky: For example, user starts a async query, and changed the schema when the async queries returns first time. But at the ascync query is running still in the background, it might crash in the background thread because of some sort field removed. User has no way to prevent this, even he is aware of this. There is no way to stop a running background async queries. So, in this case, the async query should just stop silently.

cmelchior:
@beeender This is being replaced by #3834, right?

beeender:
@cmelchior Yes. Close this one.

Issue #3732, opened 2016-11-02
StackOverflow when using insertOrUpdate to insert object with two-way relation


sipersso:
#### Goal

I want to copy an object that has a two way relation from one Realm to another using insertOrUpdate. 

#### Expected Results

Item is inserted to the new realm and relations are maintained.

#### Actual Results

Stack Overflow Error
```
 at io.realm.ParentRealmProxy.insertOrUpdate(ParentRealmProxy.java:534)
                      at io.realm.ChildRealmProxy.insertOrUpdate(ChildRealmProxy.java:506)
                      at io.realm.ParentRealmProxy.insertOrUpdate(ParentRealmProxy.java:534)
                      at io.realm.ChildRealmProxy.insertOrUpdate(ChildRealmProxy.java:506)
                      at io.realm.ParentRealmProxy.insertOrUpdate(ParentRealmProxy.java:534)
                      at io.realm.ChildRealmProxy.insertOrUpdate(ChildRealmProxy.java:506)
                      at io.realm.ParentRealmProxy.insertOrUpdate(ParentRealmProxy.java:534)
                      at io.realm.ChildRealmProxy.insertOrUpdate(ChildRealmProxy.java:506)
                      at io.realm.ParentRealmProxy.insertOrUpdate(ParentRealmProxy.java:534)
```

#### Steps & Code to Reproduce

See code sample

#### Code Sample
```
public class Parent extends RealmObject {
    @PrimaryKey
    @Required
    private String id;

    private RealmList<Child> childList;

    public RealmList<Child> getChildList() {
        return childList;
    }
}

public class Child extends RealmObject {

    @PrimaryKey
    @Required
    private String id;

    private Parent parent;

    public Parent getParent() {
        return parent;
    }

    public void setParent(Parent parent) {
        this.parent = parent;
    }
}

public class RealmOverflowBugActivity extends AppCompatActivity {

    @Override
    protected void onCreate(Bundle savedInstanceState) {
        super.onCreate(savedInstanceState);
        setContentView(R.layout.activity_realm_overflow_bug);

        Realm.init(this);

        RealmConfiguration config1 = new RealmConfiguration.Builder().name("config1").deleteRealmIfMigrationNeeded().build();
        RealmConfiguration config2 = new RealmConfiguration.Builder().name("config2").deleteRealmIfMigrationNeeded().build();

        Realm realm1 = Realm.getInstance(config1);
        Realm realm2 = Realm.getInstance(config2);

[...]

PR #3737, opened 2016-11-02
Unit test for bug with insertOrUpdate

cmelchior:
Expose bug described in https://github.com/realm/realm-java/issues/3732


nhachicha:
tracked in #3732 PR https://github.com/realm/realm-java/pull/3755 should fix this

PR #3749, opened 2016-11-05
DefaultRealmModule not created for empty Kotlin projects

### LINKED ISSUE #3746: Disable the default RealmModule generation if no realm model/module
#### Goal

Some developpers might want to create a library that will use Realm, but won't declare any model or module in it. The problem is that if we don't declare any of them, the compilation will failed because the generated Mediator will contains no classes and throw a `getMissingProxyClassException(clazz);`
The same issue can happens (I think) if we have a Library that declare the model in Realm + a LibModule, and we use only the LibModule in the application, without declaring any more models.

#### Expected Results

A flag in build.gradle or a check that disable the generation of realm class or module for project that use Realm without declaring any of both.

cmelchior:
Fixes #3746 

@realm/java

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,8 @@
 ## 2.1.2
 
-### Enhancement
+### Bug fixes
 
-* The default Realm module is no longer generated if no model classes exists (#3746).
+* Kotlin projects no longer creates the `RealmDefaultModule` if no Realm model classes are present (#3746).
 
 ## 2.1.1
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## 2.1.2
+
+### Bug fixes
+
+* Kotlin projects no longer creates the `RealmDefaultModule` if no Realm model classes are present (#3746).
```
kneth: "creates" -> "create"

cmelchior: Yes, it only effects Kotlin projects. It was still a bug in the annotation processor code, but it was only triggered by Kotlin.

Issue #1470, opened 2015-09-11
SerializedName annotation for `createAllFromJSON`


beeender:
We need something similar to GSON's `SerializedName` annotation which can map a different name field from JSON string to RealmObject's property.

eg:
```
{ "dog_name": "abc"}
```
```
class Dog {
    @SerializedName("dog_name")
    private String name;
}
```

navjotbedi:
Any update on this feature ?

TR4Android:
I'd really like to use Realms Json methods, but the way they currently work they are unfortunately too limited to be of any use to me. In addition to a `@SerializedName` annotation, I'd also need a `@SerializedObjectId(s)Name` annotation that allows linking of related objects using their ids. Most web APIs only return the id of the related object, which makes it impossible to use Realms Json methods (as far as I know).

So, for example, the following should work:
```json
{"id": 1, "relatedId": 1, "relatedIds": [1, 2, 3]}
```
```java
class TestObject extends RealmObject {
    // different serialized name in json representation
    @SerializedName("id") @PrimaryKey
    long mId;

    // one-to-one relationship in json representation
    @SerializedObjectIdName("relatedId")
    RelatedObject mRelatedObject;

    // one-to-many relationship in json representation
    @SerializedObjectIdsName("relatedIds")
    RealmList<RelatedObject> mRelatedObjects;
}
```
Using those annotations the Realm Json methods should be able to properly transfer the information encoded in the Json to the Realm object model, including its relationships. Having those annotations would solve quite a few issues for me with partially updating objects (see issue #3500) and would reduce the amount of code needed and make it more efficient.

cmelchior:
@TR4Android I have browsed through the code and compiled a list of things that needs to be done (I probably forgot something though). If you decide to start on this, I would strongly encourage you to open a PR as soon as possible so we can follow progress along the way and adjust if needed. It would make sense for us to make a feature branch which you can make smaller PR's against instead of trying to cram everything into one big PR against master. It will be a lot easier for us to review.

Let me know. Cheers.

**Overall design goal**

Make it possible to use a Java specific name when instead of being forced to adopt the name being used by a JSON API or another platform.

USE CASE: My JSON input does not match my preferred Java class. This is useful if using our various `createFromJson` methods

USE CASE: I'm sharing a schema with iOS and would like my own names as the chosen names break normal Java conventions.

With this approach we combine these two use cases. Is is that okay or will it bite us later so we don't want to mix JSON and Schema concerns?

[...]

PR #3758, opened 2016-11-08
RFC for Import/Export API

cmelchior:
Spinoff from #1470 

This RFC tries to describe two possible approaches to the problem, but perhaps I am missing something even better. 

What do you think @realm/java @TR4Android

### DIFF
--- a/rfcs/import-export-api.md
+++ b/rfcs/import-export-api.md
@@ -1,0 +1,243 @@
+# Import / Export API
+
+This document describe a possible new Import/Export API that is more generalized
+than our current JSON support and is intended to replace it.
+
+## Motivation
+
+Currently Realm Java exposes a number of JSON methods (12) for importing JSON
+data into Realm. These works fine, but are limited in functionality plus they
+bloat our public API quite considerably as well as adding a fair amount of code
+to our proxy classes.
+
+There is general agreement internally that Realm Java should not be a full JSON
+parser. Other JSON libraries are much more focused, better tested and with more
+features. The tipping point was https://github.com/realm/realm-java/issues/1470,
+which we all agree is a very useful feature, but also a slippery slope for
+turning into a full-fledged JSON library.
+
+So instead of just keep piling features onto our JSON support we would much
+rather look at this from a broader perspective of getting data in and out of
+Realm through a more generalized Import/Export API. Such a API should be flexible
+enough to support most common data formats.
+
+
+## Use cases - High level
+
+a) Import JSON from common sources: String, JSONObject/JSONArray, JsonReader.
+    1) Users of our current JSON API should be able to migrate without too much
+       trouble.
+	2) Support things like `@SerializedName`
+	3) Support automatically setting up references, e.g by just having the
+	   primary key value in JSON as a reference.
+
+b) Be compatible with common JSON libraries: GSON, Jackson, Moshi.
+	1) It should be possible to auto-generate TypeAdapters.
+
+c) Import data from other formats like CSV, XML and SQLite databases.
+
+e) Export data to common formats: CSV, XML, JSON
+
+
+## Use cases - Already supported
+
+Realm already have `copyToRealm()/copyFromRealm()` methods that uses in-memory
+Java objects as an intermediate representation.
+
+This means that any library that either output or accept in-memory object is
+supported by Realm Java today. An incomplete list is:
+
+- Retrofit
+- GSON
+- Jackson
+- Moshi
+- Other annotation processors: Realm does not consume annotations enabling other
+  tools to also generate code, e.g.: https://github.com/cmelchior/realmfieldnameshelper
+
+This means that any support we offer is mostly to be able to provide a more
+efficient/lightweight solution + support for features not offered by other
+libraries.
+
+## Use cases - Not currently supported
+
+- Export directly to JSON without needing a 3rd party library
+- Export to CSV and XML
+- Custom support for `@SerializableName`
+- Support for automatically find references from ID's.
+
+## Implementation
+
+### Current API's
+
+3 variants of the below 4 methods. JsonReader variant only supported on API 11+.
+
+```
+realm.createObjectFromJson(Class, <String/JSONObject/JsonReader>)
+realm.createOrUpdateObjectFromJson(Class, <String/JSONObject/JsonReader>)
+realm.createAllFromJson(Class, <String/JSONArray/JsonReader>)
+realm.createOrUpdateAllFromJson(Class, <String/JSONArray/JsonReader>)
+```
+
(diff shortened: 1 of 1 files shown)


TR4Android:
@cmelchior Thanks for the detailed proposal, I would almost say that option B is better as it's more concise if the import is done directly one the Realm database object. A few suggestions though:
* Just name the methods `importData` or similar. I think the keyword `import` conveys what those methods are meant to do: import data from another format into the database. (I would advise against the method name `import` itself as it's a reserved keyword in most languages.)
* Don't try to auto-detect the type in the import API. Most, if not all, developers know the type they're importing, adding detection just increases the likelihood that a developer uses the less efficient version of the API and introduces a lot of code that needs to be maintained.
* Make the imported type an enum, that way it's pretty clear what types are supported for importing and there's no need to check for invalid types as would be the case for `String`.

Also, still an open question for me is how to guarantee the order of execution of the annotation processors involved. We need to somehow make sure that the proxy classes get generated first, so the import/export processor can add its code to it.

cmelchior:
> Just name the methods importData or similar. I think the keyword import conveys what those methods are meant to do: import data from another format into the database. (I would advise against the method name import itself as it's a reserved keyword in most languages.)

`importData` is an option as well. My primary concern was the distinction between update or not. Right now most of our apis are called e.g `insertOrUpdate/createOrUpdate`, `importOrUpdate` just sounds wrong? 

> Make the imported type an enum, that way it's pretty clear what types are supported for importing and there's no need to check for invalid types as would be the case for String.

I also like a enum a lot more, but it makes the API less flexible since you are then constrained to the types we define. I'm not sure it is a problem in practise though as `XML/JSON/CSV` would probably support 99% of export formats.

Zhuinden:
> XML/JSON/CSV would probably support 99% of export formats.

I could define flatbuffer or something ridiculous like [flatbuffers](https://github.com/square/wire)

------

as for JSON support, we should also take into consideration the fact that [you still can't directly map between `RealmList<RealmInteger>`](https://github.com/realm/realm-java/issues/575), so you'd need custom serialization logic for array of integers to either convert to that, or  join them into a single String field like `|5|7|10|` if you so desire.

------------

---------

I have this wild guess that the right direction would be something like the Retrofit2 [Converter.Factory](https://github.com/square/retrofit/blob/master/retrofit-converters/gson/src/main/java/retrofit2/converter/gson/GsonConverterFactory.java) approach.

cmelchior:
> I have this wild guess that the right direction would be something like the Retrofit2 Converter.Factory approach.

Retrofit has the luxury of making the users define the interface, so they can annotate their API with `@json / @xml` if I remember correctly. I really like the factory approach as well, but there will be cases where it is impossible to tell the difference, e.g If I'm importing a String is that a json or xml string.

Zhuinden:
@cmelchior 

> there will be cases where it is impossible to tell the difference, e.g If I'm importing a String is that a json or xml string.

uh... isn't that why [the converter factory gets a list of annotations](https://github.com/aurae/retrofit-logansquare/blob/master/converter-logansquare/src/main/java/com/github/aurae/retrofit2/LoganSquareConverterFactory.java#L33) which can be used to determine if it's for `@Root` annotation (SimpleXML)` or `@JsonObject` annotation (LoganSquare) for example?

cmelchior:
Yes, and that is a great idea, but users doesn't control the Realm API, so they cannot annotate the import method.

Zhuinden:
They can annotate RealmObjects though. Maybe that changes things? The key here is "automating"  batch insertion, or so I would think. Although I do wonder how relationships would be handled, or custom serialization like for RealmList of RealmInts. 

This is a difficult issue, I'll try to think about it; although my wild guess is that you'll come up with something even better than any idea I'd have. :D /sleeptime

rfcs/import-export-api.md
```diff
@@ -0,0 +1,243 @@
+# Import / Export API
+
+This document describe a possible new Import/Export API that is more generalized
```
dalinaum: document -> describes?

rfcs/import-export-api.md
```diff
@@ -0,0 +1,243 @@
...
+
+There is general agreement internally that Realm Java should not be a full JSON
+parser. Other JSON libraries are much more focused, better tested and with more
+features. The tipping point was https://github.com/realm/realm-java/issues/1470,
+which we all agree is a very useful feature, but also a slippery slope for
+turning into a full-fledged JSON library.
+
+So instead of just keep piling features onto our JSON support we would much
+rather look at this from a broader perspective of getting data in and out of
+Realm through a more generalized Import/Export API. Such a API should be flexible
```
dalinaum: Such a API -> Such an API?

rfcs/import-export-api.md
```diff
@@ -0,0 +1,243 @@
...
+	1) It should be possible to auto-generate TypeAdapters.
+
+c) Import data from other formats like CSV, XML and SQLite databases.
+
+e) Export data to common formats: CSV, XML, JSON
+
+
+## Use cases - Already supported
+
+Realm already have `copyToRealm()/copyFromRealm()` methods that uses in-memory
```
dalinaum: Realm already have -> Realm already has?
that uses -> that use?

rfcs/import-export-api.md
```diff
@@ -0,0 +1,243 @@
...
+**Advantages**
+- API visible in main Realm API.
+- Actual implementation still fully opt-in.
+- New formats supported easily by creating `RealmXml`, `RealmCsv`, ...
+- Big reduction in Realm Proxy class size and complexity
+- Big reduction in API surface for the Realm class
+- Move towards a opt-in plugin-type architecture.
+
+**Disadvantages**
+- Methods will crash if no plugin is installed.
```
beeender: This disadvantage is really really minor ... I wouldn't say it is a disadvantage, it is just a bug needs to be fixed when it happens.

rfcs/import-export-api.md
```diff
@@ -0,0 +1,243 @@
...
+- Less flexible than a seperate class.
+
+
+## Open questions:
+
+* Can we enumerate possible input formats (XML, JSON, CSV), or does it need to
+  be flexible?
+
+* A more radical solution could also be to just remove our JSON support
+  completely and completely delegate it to 3rd party libraries?
```
beeender: The converter has to call APIs in io.realm.internal for speed concerns. I am not quite sure we have all the checks for every internal API (but ideally we should have all checks inside the internal APIs)

eygraber:
I like @zhuinden idea of a retrofit-esque solution. Define an interface in realm for importing data, create a library with implementations using popular 3rd party libraries (gson, Moshi, something for XML, etc...), and let the user pass an instance of that implementation to the import method.

PR #3782, opened 2016-11-12
fix artifactoryPublish task in realm-annotations-processor

zaki50:
fix NPE when executing `artifactoryPublish` task in `realm-annotations-processor`.

```
:****:****-annotations-processor:artifactoryPublish FAILED
:ojoRealm FAILED

* What went wrong:
Execution failed for task ':****-annotations-processor:artifactoryPublish'.
> java.lang.NullPointerException (no error message)

```

@realm/java @emanuelez

### DIFF
--- a/realm/build.gradle
+++ b/realm/build.gradle
@@ -12,8 +12,8 @@ buildscript {
         classpath 'com.novoda:gradle-android-command-plugin:1.3.0'
         classpath 'com.github.skhatri:gradle-s3-plugin:1.0.2'
         classpath 'org.kt3k.gradle.plugin:coveralls-gradle-plugin:2.4.0'
-        classpath 'org.jfrog.buildinfo:build-info-extractor-gradle:4.4.5'
-        classpath 'com.jfrog.bintray.gradle:gradle-bintray-plugin:1.7'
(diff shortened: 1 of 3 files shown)


realm/build.gradle
```diff
@@ -12,8 +12,8 @@ buildscript {
         classpath 'com.novoda:gradle-android-command-plugin:1.3.0'
         classpath 'com.github.skhatri:gradle-s3-plugin:1.0.2'
         classpath 'org.kt3k.gradle.plugin:coveralls-gradle-plugin:2.4.0'
-        classpath 'org.jfrog.buildinfo:build-info-extractor-gradle:4.4.5'
-        classpath 'com.jfrog.bintray.gradle:gradle-bintray-plugin:1.7'
+        classpath 'org.jfrog.buildinfo:build-info-extractor-gradle:3.1.1'
```
cmelchior: Do we need to downgrade?

zaki50: I got some exception in `realm/realm-annotations-processor` project and I couldn't find a way to solve it.
I checked other working projects such as `grade-plugin`, they still use `build-info-extractor-gradle:3.1.1` and `gradle-bintray-plugin:1.6`.

Downgrading is the only way that I could find..

PR #3789, opened 2016-11-14
Update IntroExampleActivity.java

NivedithaKabbur:
Initialize realm for the application context.


NivedithaKabbur:
Initialize realm for the application context.

PR #3790, opened 2016-11-14
Update IntroExampleActivity.java

NivedithaKabbur:
Initialize realm for the application context.


examples/introExample/src/main/java/io/realm/examples/intro/IntroExampleActivity.java
```diff
@@ -41,6 +41,7 @@
     protected void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
         setContentView(R.layout.activity_realm_basic_example);
+         Realm.init(this);
```
Zhuinden: white space is wrong and does not resemble the docs (if I wanted to call this in activity, I'd put it before `super.onCreate()` call)

cmelchior:
Hi @NivedithaKabbur 
Thanks for the PR, but we are already setting this in https://github.com/realm/realm-java/blob/master/examples/introExample/src/main/java/io/realm/examples/intro/MyApplication.java

So this is not nessassery.

PR #3798, opened 2016-11-15
Fix getLocalInstanceCount crashing when Realms are closed again.

### LINKED ISSUE #3791: Realm.getLocalInstanceCount() can crash
See https://github.com/realm/realm-java/issues/3788#issuecomment-260318809

cmelchior:
Fixes #3791 

@realm/java

### DIFF
--- a/realm/realm-library/src/main/java/io/realm/RealmCache.java
+++ b/realm/realm-library/src/main/java/io/realm/RealmCache.java
@@ -402,7 +402,8 @@ final class RealmCache {
         } else {
             int totalRefCount = 0;
             for (RealmCacheType type : RealmCacheType.values()) {
-                totalRefCount += cache.refAndCountMap.get(type).localCount.get();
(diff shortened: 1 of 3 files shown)


realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -402,7 +402,8 @@ static int getLocalThreadCount(RealmConfiguration configuration) {
         } else {
             int totalRefCount = 0;
             for (RealmCacheType type : RealmCacheType.values()) {
-                totalRefCount += cache.refAndCountMap.get(type).localCount.get();
+                Integer localCount = cache.refAndCountMap.get(type).localCount.get();
```
Zhuinden: Is there a chance for `cache.refAndCountMap.get(type)` to return null?

cmelchior: No, it is initialized in the constructor

Issue #3804, opened 2016-11-16
Realm crashes on Android 2.3 device


zaki50:
This issue is reported by @takke on twitter (http://twitter.com/takke/status/798782520749801472).

#### Goal

Works on Android 2.3.x devices.

#### Expected Results

no crash

#### Actual Results

When executing `Realm.getDefaultInstance()`, the app crash with following dump in logcat.

```
01-06 10:32:12.689 I/DEBUG   (  747): *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
01-06 10:32:12.689 I/DEBUG   (  747): Build fingerprint: 'google/passion/passion:2.3.6/GRK39F/189904:user/release-keys'
01-06 10:32:12.689 I/DEBUG   (  747): pid: 845, tid: 845  >>> jp.takke.realmandroid23issuesample <<<
01-06 10:32:12.689 I/DEBUG   (  747): signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr de5685a7
01-06 10:32:12.689 I/DEBUG   (  747):  r0 00000007  r1 00000001  r2 800a3d68  r3 00000000
01-06 10:32:12.689 I/DEBUG   (  747):  r4 0000ce60  r5 0000abe0  r6 de5685a7  r7 00000000
01-06 10:32:12.689 I/DEBUG   (  747):  r8 0000abe0  r9 4214cb28  10 0000abe0  fp 800a5368
01-06 10:32:12.689 I/DEBUG   (  747):  ip 800a56cc  sp beef23b0  lr 8004a719  pc 8004a718  cpsr 60000030
01-06 10:32:12.699 I/DEBUG   (  747):  d0  2e746c756166652e  d1  2f73656c69662f6e
01-06 10:32:12.699 I/DEBUG   (  747):  d2  646e616d6c61656f  d3  7369333264696f74
01-06 10:32:12.699 I/DEBUG   (  747):  d4  2f73656c69662f65  d5  2e746c7561666564
01-06 10:32:12.699 I/DEBUG   (  747):  d6  616d2e6d6c616572  d7  746e656d6567616e
01-06 10:32:12.699 I/DEBUG   (  747):  d8  0000000000000000  d9  0000000000000000
01-06 10:32:12.699 I/DEBUG   (  747):  d10 0000000000000000  d11 0000000000000000
01-06 10:32:12.699 I/DEBUG   (  747):  d12 0000000000000000  d13 0000000000000000
[...]

Issue #3735, opened 2016-11-02
RealmFileException: Function not implemented in /Users/cm/Realm/realm-java/realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp line 81 Kind: ACCESS_ERROR.


matomick:
I get this, it happens on: Android 4.3
Blackberry Leap device.
Realm 2.0.2
Any ideas for a fix?

io.realm.exceptions.RealmFileException: Function not implemented in /Users/cm/Realm/realm-java/realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp line 81 Kind: ACCESS_ERROR.
	at io.realm.internal.SharedRealm.nativeGetSharedRealm(Native Method)
	at io.realm.internal.SharedRealm.getInstance(Unknown Source)
	at io.realm.internal.SharedRealm.getInstance(Unknown Source)
	at io.realm.RealmCache.createRealmOrGetFromCache(Unknown Source)
	at io.realm.Realm.getInstance(Unknown Source)

cmelchior:
Are you saving the Realm on external storage?

matomick:
The realm file isn't even created, this happens at the execution of getInstance (when realm is supposed to create the realm file because it doesn't exists)

Zhuinden:
What is your `RealmConfiguration`?

matomick:
realmFileName = Realm.DEFAULT_REALM_NAME;
config = new RealmConfiguration.Builder()
                    .name(realmFileName)
                    .schemaVersion(versionSchema)
                    .encryptionKey(key)
                    .migration(new MigrationRealm())
                    .build();

Zhuinden:
@beeender Blackberry is broken again

diegomontoya:
@matomick Try again by disabling encryption. Run again without the ".encryptionKey(key)" part. At least this can exclude the encryption code path.

Anthonyeef:
Same thing happened here. From the log Fabric collected for me, I only saw this error happened on Blackberry's device(e.g, Passport and Z10).

Anthonyeef:
Also, I am currently running realm version 2.0.2 on my app.

Anthonyeef:
More detail with log:
```
Fatal Exception: io.realm.exceptions.RealmFileException: Function not implemented in /Users/cm/Realm/realm-java/realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp line 81
[...]

PR #3824, opened 2016-11-21
Added a support for androidTest on API 9 devices

zaki50:
This enables for us to reproduce #3804 on our unit test.

@realm/java 

After applying this change, our android test reproduces #3804 on ARM with API 9 device, and #3735 on x86  with API 9 emulator.


kneth:
@zaki50 Since you only modify, is it correct to assume this change has no impact on the final product?

zaki50:
@kneth yes. I'm sure this change has no impact for the product.

`realm/realm-library/src/main/AndroidManifest.xml` is the only file which has a change related to product, but it's just a simplification.
Should I revert that?

kneth:
@zaki50 Sounds great. No reason to revert.

zaki50:
@kneth I updated `realm/realm-library/build.gradle` to minimize the modification.

PR #3826, opened 2016-11-21
fix CI build on releases branch

zaki50:


zaki50:
fast-merge /cc @realm/java 
I only cherry-picked commits on master that fix CI build issue.

PR #3835, opened 2016-11-23
Upgrade to latest OS

cmelchior:
Upgrade to latest commit on the `realm-object-store/realm-java` branch

@realm/java


cmelchior:
Yes, they are part of the sync refactoring. I'm using them in another branch right now. This is is mostly to make sure that we compile with the latest commit ( @nhachicha ran into problems with it) .

PR #3838, opened 2016-11-24
Nh/objectstore userstore

nhachicha:
Uses the Object Store to save/retrieve SyncUsers

* The current ObjectStore implementation supports only one logged-in user, hence the simplification of the `UserStore` interface. After the _SyncManager_ is completely integrated & the ObjectStore updated the SyncUser `identity` could be used, to work with multiple users.


nhachicha:
@realm/java

realm/realm-library/src/main/cpp/io_realm_ObjectStoreUserStore.cpp
```diff
@@ -0,0 +1,138 @@
...
+        std::runtime_error("cannot be called if more that one valid, logged-in user exists.");//TODO user appropriate exception
+    } else if (all_users.size() < 1) {
+        std::runtime_error("no user logged-in yet.");//TODO user appropriate exception
+    } else {
+        return all_users.front();
+    }
+}
+
+JNIEXPORT jstring JNICALL Java_io_realm_ObjectStoreUserStore_getCurrentUser
+        (JNIEnv *env, jclass) {
```
kneth: Please, no indentation

realm/realm-library/src/main/cpp/io_realm_ObjectStoreUserStore.cpp
```diff
@@ -0,0 +1,138 @@
...
+    TR_ENTER()
+    const std::shared_ptr<SyncUser>& user = currentUserOrThrow();
+    if (user->state() == SyncUser::State::Active) {
+        return env->NewStringUTF(user->refresh_token().c_str());
```
kneth: Why not use `to_jstring()`?

realm/realm-library/src/main/cpp/io_realm_ObjectStoreUserStore.cpp
```diff
@@ -0,0 +1,138 @@
...
+    } else if (all_users.size() < 1) {
+        std::runtime_error("no user logged-in yet.");//TODO user appropriate exception
+    } else {
+        return all_users.front();
+    }
+}
+
+JNIEXPORT jstring JNICALL Java_io_realm_ObjectStoreUserStore_getCurrentUser
+        (JNIEnv *env, jclass) {
+    TR_ENTER()
```
kneth: No `try`/`catch`?

realm/realm-library/src/main/cpp/io_realm_ObjectStoreUserStore.cpp
```diff
@@ -0,0 +1,138 @@
...
+
+/*
+ * Class:     io_realm_ObjectStoreSyncManager
+ * Method:    logoutCurrentUser
+ * Signature: ()V
+ */
+JNIEXPORT void JNICALL Java_io_realm_ObjectStoreUserStore_logoutCurrentUser
+        (JNIEnv *, jclass) {
+    TR_ENTER()
+    const std::shared_ptr<SyncUser>& user = currentUserOrThrow();
```
kneth: No `try`/`catch`?

realm/realm-library/src/main/cpp/io_realm_ObjectStoreUserStore.cpp
```diff
@@ -0,0 +1,138 @@
...
+ * Method:    getAllUsers
+ * Signature: ()[Ljava/lang/String;
+ */
+JNIEXPORT jobjectArray JNICALL Java_io_realm_ObjectStoreUserStore_getAllUsers
+        (JNIEnv *env, jclass) {
+    TR_ENTER()
+    std::vector<std::shared_ptr<SyncUser>> all_users = SyncManager::shared().all_users();
+    if (!all_users.empty()) {
+        std::vector<jobject> valid_users;
+        for (auto& user : all_users) {
```
kneth: Consider to use `std::find_if()`

realm/realm-library/src/main/cpp/io_realm_ObjectStoreUserStore.cpp
```diff
@@ -0,0 +1,138 @@
...
+            if ((*user).state() == SyncUser::State::Active) {
+                valid_users.push_back(env->NewStringUTF((*user).refresh_token().c_str()));
+            }
+        }
+
+        std::vector<jobject>::size_type size_valid_users = valid_users.size();
+        jclass stringClass = env->FindClass("java/lang/String");
```
kneth: Please use `GetClass()` in `util.hpp`. Or look it up when the `.so` is loaded (see `JNI_OnLoad` in `io_realm_internal_Util.cpp`)

realm/realm-library/src/main/cpp/io_realm_ObjectStoreUserStore.cpp
```diff
@@ -0,0 +1,138 @@
...
+        jobjectArray users_token = env->NewObjectArray(size_valid_users, stringClass, 0);
+        if (users_token == NULL) {
+            ThrowException(env, OutOfMemory, "Could not allocate memory to return all users.");
+            return NULL;
+        }
+
+        for (std::vector<jobject>::size_type i = 0; i != size_valid_users; ++i) {
```
kneth: Can't you do `for (auto user : valid_users)` instead?

nhachicha:
@realm/java ready to review

realm/realm-library/src/main/cpp/io_realm_ObjectStoreUserStore.cpp
```diff
@@ -23,35 +23,47 @@
...
+
+static const std::shared_ptr<SyncUser>& currentUserOrThrow(JNIEnv *env) //throws
+{
     std::vector<std::shared_ptr<SyncUser>> all_users = SyncManager::shared().all_users();
     if (all_users.size() > 1) {
-        std::runtime_error("cannot be called if more that one valid, logged-in user exists.");//TODO user appropriate exception
+        ThrowException(env, RuntimeError, ERR_MULTIPLE_LOGGED_IN_USERS);
     } else if (all_users.size() < 1) {
-        std::runtime_error("no user logged-in yet.");//TODO user appropriate exception
+        ThrowException(env, RuntimeError, ERR_NO_LOGGED_IN_USER);
```
kneth: Won't your compiler not give you a warning? I mean, nothing is returned.

realm/realm-library/src/main/cpp/io_realm_ObjectStoreUserStore.cpp
```diff
@@ -23,35 +23,47 @@
...
+{
     TR_ENTER()
-    const std::shared_ptr<SyncUser>& user = currentUserOrThrow();
-    if (user->state() == SyncUser::State::Active) {
-        return env->NewStringUTF(user->refresh_token().c_str());
-    } else {
-        return NULL;
-    }
+    try {
+        const std::shared_ptr<SyncUser> &user = currentUserOrThrow(env);
```
kneth: Throwing a C++ exception in `currentUserOrThrow` will be caught by `CATCH_STD()`. If you only set the Java exception state (as in `ThrowException()`), you are not allowed to call any JNI functions. As you don't return anything in `currentUserOrThrow()` if you "throw" a Java exception, `user->state` might seg. fault - or `to_jstring()` might terminate the app.

realm/realm-library/src/main/cpp/io_realm_ObjectStoreUserStore.cpp
```diff
@@ -67,11 +79,13 @@ JNIEXPORT void JNICALL Java_io_realm_ObjectStoreUserStore_updateOrCreateUser
...
  */
-JNIEXPORT void JNICALL Java_io_realm_ObjectStoreUserStore_logoutCurrentUser
-        (JNIEnv *, jclass) {
+JNIEXPORT void JNICALL Java_io_realm_ObjectStoreUserStore_logoutCurrentUser (JNIEnv *env, jclass)
+{
     TR_ENTER()
-    const std::shared_ptr<SyncUser>& user = currentUserOrThrow();
-    user->log_out();
+    try {
+        const std::shared_ptr<SyncUser>& user = currentUserOrThrow(env);
```
kneth: Likewise here. Since `currentUserOrThrow()` just returns, you don't know if `user->log_out()` can be called.

realm/realm-library/src/main/cpp/io_realm_ObjectStoreUserStore.cpp
```diff
@@ -80,46 +94,44 @@ JNIEXPORT void JNICALL Java_io_realm_ObjectStoreUserStore_logoutCurrentUser
...
     std::vector<std::shared_ptr<SyncUser>> all_users = SyncManager::shared().all_users();
     if (!all_users.empty()) {
-        std::vector<jobject> valid_users;
-        for (auto& user : all_users) {
-            if ((*user).state() == SyncUser::State::Active) {
-                valid_users.push_back(env->NewStringUTF((*user).refresh_token().c_str()));
-            }
-        }
+        jsize array_length = 0;
+        std::vector<std::shared_ptr<SyncUser>>::iterator valid_users = std::find_if(all_users.begin(),all_users.end(),
```
kneth: Nice :-)

realm/realm-library/src/main/cpp/util.hpp
```diff
@@ -665,6 +665,7 @@ extern jmethodID java_lang_long_init;
 extern jclass java_lang_float;
 extern jmethodID java_lang_float_init;
 extern jclass java_lang_double;
+extern jclass java_lang_string;
```
kneth: Don't you need the constructor too?

kneth: No, you don't - sorry.

realm/realm-library/src/main/cpp/io_realm_ObjectStoreUserStore.cpp
```diff
@@ -0,0 +1,150 @@
...
+        return all_users.front();
+    }
+}
+
+/*
+ * Class:     io_realm_ObjectStoreUserStore
+ * Method:    getCurrentUser
+ * Signature: ()Ljava/lang/String;
+ */
+JNIEXPORT jstring JNICALL Java_io_realm_ObjectStoreUserStore_getCurrentUser (JNIEnv *env, jclass)
```
beeender: define the function like:
```
JNIEXPORT jstring JNICALL
Java_io_realm_ObjectStoreUserStore_getCurrentUser (JNIEnv *env, jclass) // new line for this.
```

Also, remove the comments before the function maybe? It will be generated in the header file and we don't have it for all of our current jni cpp files.

Also apply for other functions.

realm/realm-library/src/main/cpp/io_realm_ObjectStoreUserStore.cpp
```diff
@@ -0,0 +1,150 @@
...
+{
+    TR_ENTER()
+    try {
+        const std::shared_ptr<SyncUser> &user = currentUserOrThrow(env);
+        if (user->state() == SyncUser::State::Active) {
+            return to_jstring(env, user->refresh_token().data());
+        } else {
+            return NULL;
```
beeender: use `nullptr` instead?

realm/realm-library/src/main/cpp/io_realm_ObjectStoreUserStore.cpp
```diff
@@ -0,0 +1,150 @@
...
+        }
+    } CATCH_STD()
```
beeender: Also, return a `nullptr` here. Otherwise there will be a compiling warning.

realm/realm-library/src/main/cpp/io_realm_ObjectStoreUserStore.cpp
```diff
@@ -0,0 +1,150 @@
...
+ */
+JNIEXPORT void JNICALL Java_io_realm_ObjectStoreUserStore_updateOrCreateUser (JNIEnv *env, jclass, jstring identity, jstring jsonToken, jstring url)
+{
+    TR_ENTER()
+    try {
+        JStringAccessor user_identity(env, identity); // throws
+        JStringAccessor user_json_token(env, jsonToken); // throws
+        JStringAccessor auth_url(env, url); // throws
+
+        SyncManager::shared().get_user(user_identity, user_json_token, util::Optional<std::string>(auth_url));
```
beeender: `std::string(auth_url) ` should be enough i think.

realm/realm-library/src/main/cpp/io_realm_ObjectStoreUserStore.cpp
```diff
@@ -0,0 +1,150 @@
...
+}
+
+/*
+ * Class:     io_realm_ObjectStoreSyncManager
+ * Method:    getAllUsers
+ * Signature: ()[Ljava/lang/String;
+ */
+JNIEXPORT jobjectArray JNICALL Java_io_realm_ObjectStoreUserStore_getAllUsers (JNIEnv *env, jclass)
+{
+    TR_ENTER()
```
beeender: missing `try ... CATCH_STD`

realm/realm-library/src/main/cpp/io_realm_ObjectStoreUserStore.cpp
```diff
@@ -0,0 +1,150 @@
...
+    std::vector<std::shared_ptr<SyncUser>> all_users = SyncManager::shared().all_users();
+    if (!all_users.empty()) {
+        jsize array_length = 0;
+        std::vector<std::shared_ptr<SyncUser>>::iterator valid_users = std::find_if(all_users.begin(),all_users.end(),
```
beeender: this function logic doesn't seem to be correct. `find_if` will stop when it finds the first element matches the condition, so the `array_length` will be always 1(or 0) here. And the following `for_each` loop will iterate from the `valid_users` until the end of original vector, means you will get a `array_length == -1` in the loop. 

I think you can do something like:
```cpp
array_length = std::count_if(all_users.begin(),all_users.end(), [&] {...};
for (auto user : all_users) {
    if (user->state() == SyncUser::State::Active) {
         // Set the java array.
    }
}
```

realm/realm-library/src/objectServer/java/io/realm/ObjectStoreUserStore.java
```diff
@@ -0,0 +1,92 @@
...
+package io.realm;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+
+/**
+ * A User Store backed by the ObjectStore metadata Realm to store user.
+ */
+public class ObjectStoreUserStore implements UserStore {
```
beeender: This name is funny ... Can we have a better one? `RealmUserStore` ?

realm/realm-library/src/main/cpp/io_realm_ObjectStoreUserStore.cpp
```diff
@@ -0,0 +1,150 @@
...
+    } CATCH_STD()
+}
+
+
+/*
+ * Class:     io_realm_ObjectStoreSyncManager
+ * Method:    configureMetaDataSystem
+ * Signature: (Ljava/lang/String;[B)V
+ */
+JNIEXPORT void JNICALL Java_io_realm_ObjectStoreUserStore_configureMetaDataSystem (JNIEnv *env, jclass, jstring baseFile)
```
beeender: I believe this function should exist in the JNI `SyncManager` wrapper and called by the `SyncManager`. @cmelchior

realm/realm-library/src/objectServer/java/io/realm/ObjectStoreUserStore.java
```diff
@@ -0,0 +1,92 @@
...
+            for (String userJson : allUsers) {
+                users.add(SyncUser.fromJson(userJson));
+            }
+            return users;
+        }
+        return Collections.emptyList();
+    }
+
+    // init and load the Metadata Realm containing SyncUsers
+    protected static native void configureMetaDataSystem(String baseFile);
```
beeender: why those are `protected`?

realm/realm-library/src/objectServer/java/io/realm/UserStore.java
```diff
@@ -29,48 +28,41 @@
...
      *
      * @return Collection of all users. If no users exist, an empty collection is returned.
      */
     Collection<SyncUser> allUsers();
-
-
-    /**
-     * Removes all saved users.
-     */
-    void clear();
```
beeender: So this API is not needed anymore?

realm/realm-library/src/objectServer/java/io/realm/ObjectStoreUserStore.java
```diff
@@ -0,0 +1,92 @@
...
+
+    // return json data (token) of the current logged in user
+    protected static native String getCurrentUser ();
```
beeender: we always name the JNI function like `nativeGetCurrentUser`, shall we follow that naming rule? (Although I think `native` is redundant in the method name.)

realm/realm-library/src/objectServer/java/io/realm/ObjectStoreUserStore.java
```diff
@@ -0,0 +1,92 @@
...
+        // create or update token (userJson) using identity
+        updateOrCreateUser(user.getIdentity(), userJson, user.getSyncUser().getAuthenticationUrl().toString());
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public SyncUser get() {
+        String userJson = getCurrentUser();
```
beeender: From the cpp code, this returns a token string, so it is also a JSON string?

realm/realm-library/src/main/cpp/io_realm_ObjectStoreUserStore.cpp
nhachicha: I agree once the the `SyncManager` is fully integrated. Btw (Cocoa call it inside `RLMSyncManager.mm#initWithCustomRootDirectory`)

realm/realm-library/src/main/cpp/io_realm_ObjectStoreUserStore.cpp
nhachicha: why? which function throws? btw `currentUserOrThrow` is not called here

realm/realm-library/src/objectServer/java/io/realm/UserStore.java
nhachicha: yes. now we call `ObjectStoreUserStore.reset_for_testing();` as a helper for tests.

realm/realm-library/src/objectServer/java/io/realm/ObjectStoreUserStore.java
nhachicha: they will be accessed from https://github.com/realm/realm-android-user-store/ to provide an implementation that encrypt/decrypt user

realm/realm-library/src/objectServer/java/io/realm/ObjectStoreUserStore.java
nhachicha: Changed to `DefaultUserStore`

realm/realm-library/src/main/cpp/io_realm_DefaultUserStore.cpp
```diff
@@ -126,25 +97,33 @@ JNIEXPORT jobjectArray JNICALL Java_io_realm_ObjectStoreUserStore_getAllUsers (J
...
         if (users_token == NULL) {
             ThrowException(env, OutOfMemory, ERR_COULD_NOT_ALLOCATE_MEMORY);
-            return NULL;
+            return nullptr;
         }
 
-        std::for_each(valid_users, all_users.end(), [&](const std::shared_ptr<SyncUser>& user) {
-           env->SetObjectArrayElement(users_token, --array_length, to_jstring(env, user->refresh_token().data()));
-        });
+        for (auto user : all_users) {
```
beeender: I think you mean `valid_users` instead of `all_users` here?

beeender:
:+1:

realm/realm-library/src/objectServer/java/io/realm/DefaultUserStore.java
```diff
@@ -0,0 +1,92 @@
...
+ */
+
+package io.realm;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+
+/**
+ * A User Store backed by the ObjectStore metadata Realm to store user.
```
cmelchior: Perhaps rephrase so it is more meaningful to users since the ObjectStore is an entirely internal thing. What about `A User Store backed by a Realm file`

realm/realm-library/src/objectServer/java/io/realm/DefaultUserStore.java
```diff
@@ -0,0 +1,92 @@
...
+ */
+public class DefaultUserStore implements UserStore {
```
cmelchior: `Default` is a very non-descript word. What about `RealmFileUserStore`?

realm/realm-library/src/objectServer/java/io/realm/UserStore.java
```diff
@@ -29,48 +28,41 @@
...
- * @see SharedPrefsUserStore
+ * @see DefaultUserStore
  */
 @Beta
 public interface UserStore {
 
-    String CURRENT_USER_KEY = "realm$currentUser";
-
     /**
      * Saves a {@link SyncUser} object under the given key. If another user already exists, it will be replaced.
```
cmelchior: This is no longer correct.

realm/realm-library/src/objectServer/java/io/realm/UserStore.java
cmelchior: What about if we don't use the `DefaultUserStore`. There this still seems handy?

Issue #469, opened 2014-10-12
Document how to set an auto increment id?


djyde:

mpost:
I would also be interested in an auto id feature. Maybe not even an increment but a UUID.

vuhung3990:
i use this on android                    
                    realm.beginTransaction();
                    dbObj obj = realm.createObject(dbObj.class);

                    // increatement index
                    long lastID = realm.where(dbObj.class).maximumInt("id");
                    int nextID = lastID <= 0 ? 0 : (int) (lastID + 1);

                    // insert new value
                    obj.setId(nextID);
                    obj.setName("thang");
                    obj.setAge(10);

                    realm.commitTransaction();

rajuashok:
When I build a Messaging application I would like to create local fake message(s) with unique IDs and then update them with the actual unique key I get from the server after successful post. The way I do this in SQL is by setting an int id field to Primary Key and then setting String uniqueKey as part of my Unique Key.  I've just started experimenting with Realm today, but some of these limitations have really made it difficult for me to migrate to Realm.

markini:
I have a lot of local models which needs an unique identifier.
I pass the id to fragments and threads and fetch them then from realm. 

I'm doing something like this for now: `model.setId(UUID.randomUUID().toString());`
I'm still not sure what I should do when a collision occurs. 
- Overriding is no option in my case
- Letting the app crash could lead to unpredictable states. 

I'm thinking about either query every id before adding the model to realm or implement a rollback strategy (not sure how, or if possible). 
Has anyone similar use cases?

Rexota:
@markini Collision of UUID is very little chance.

markini:
@Rexota yeah, we decided to just go without collision detection.
I looked a little bit into the UUID documentation and I guess collisions of said ids are indeed the least of my worries.

AlokBansal8:
@bmunkholm any update on it?

ronanrodrigo:
I try using timestamp. What you do think about this? `Int(NSDate().timeIntervalSince1970)`

Arthez:
> I to solve this issue with timestamp. What you do think about this? Int(NSDate().timeIntervalSince1970)
Im afraid that this is bad way, I insert list of 8 and pairs of 3 have same ID, even if I put nanoseconds.

cmelchior:
Timestamps could work, but they only have millisecond precision which means that there is a small chance of two timestamps being created at the same time. Using an AtomicInteger would be safer:

[...]

PR #3875, opened 2016-12-05
Add the autoIncrement option for primary keys

larsgrefer:
see #469


zaki50:
I don't think this does not work as expected with synced Realm.

When the two devices use auto-increment feature, they get the same value and those are merged into one.

cmelchior:
Hi @larsgrefer 

Thank you for the pull request. Unfortunately, we cannot accept it for the reasons provided by @zaki50. Auto-incremented keys simply do not work in a distributed setting, so it is not something we would want to provide support for in the Realm library itself. The issue you reference is only for documenting one of the numerous work-arounds described in the issue, not about adding direct support for it. Reading the issue, I realize that is not entirely clear, and I'm sorry for that.

In general, before working on any of our bigger issues it would be safer to reach out to us with an outline of your solution so we can make sure it is aligned with our other requirements. That will prevent work like this getting wasted.

PR #3898, opened 2016-12-12
Zeroing memory before using it

kneth:
I have noticed that while debugging in Android Studio, strings contain garbage. Java strings can contain `NULL` characters so it can be a simple artefact of that. But in the light of our string issues in the last two years,  it might be worth zeroing the memory before using it.

The impact on performance is evaluated using the microbenchmark (included in the project)  on a Samsung S III mini (I-8190N). Impact is minimal (writeLong is included to evaluate the variance).

| Test | String length | w/o `memset` | with `memset` |
|------|--------------|---------------|-----------------|
| writeLong |N/A | 15.63 μs | 16.09 μs |
| writeLongString | 64 | 23.71 μs | 24.86 μs |
| writeMediumString | 16 | 20.8 μs | 21.55 μs |
| writeShortString | 3 | 20.31 μs | 20.40 μs |

### DIFF
--- a/realm/realm-library/src/main/cpp/util.cpp
+++ b/realm/realm-library/src/main/cpp/util.cpp
@@ -405,7 +405,9 @@ JStringAccessor::JStringAccessor(JNIEnv* env, jstring str)
         size_t error_code;
         buf_size = Xcode::find_utf8_buf_size(begin, end, error_code);
     }
-    m_data.reset(new char[buf_size]);  // throws
+    char *c = new char[buf_size]; // throws
+    std::memset(c, 0, buf_size);
+    m_data.reset(c);
     {
         const jchar* in_begin = chars.data();
         const jchar* in_end   = in_begin + chars.size();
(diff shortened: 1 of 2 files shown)


kneth:
@bmunkholm I have added additional micro benchmarks, narrowed the usage of `std::memset`, and updated the reported benchmarks in the description.

realm/realm-library/src/main/cpp/util.cpp
```diff
@@ -405,7 +405,8 @@ JStringAccessor::JStringAccessor(JNIEnv* env, jstring str)
         size_t error_code;
         buf_size = Xcode::find_utf8_buf_size(begin, end, error_code);
     }
-    m_data.reset(new char[buf_size]);  // throws
+    char* c = new char[buf_size]; // throws
```
bmunkholm: Maybe find a better name than `c`

beeender:
I think adding a `\0` at the end would be good enough, but from benchmark it doesn't add too much overhead ... Also, I feel `x4` should be always ensure we have a room for space .

PR #3899, opened 2016-12-13
Disabling sync/compact combo

kneth:
Due to https://github.com/realm/realm-core/issues/2345 we need to disable `Realm.compactRealm()` when the Realm is sync'ed.

### DIFF
--- a/realm/realm-library/src/main/java/io/realm/Realm.java
+++ b/realm/realm-library/src/main/java/io/realm/Realm.java
@@ -1555,10 +1555,13 @@ public final class Realm extends BaseRealm {
      *
      * @param configuration a {@link RealmConfiguration} pointing to a Realm file.
      * @return {@code true} if successful, {@code false} if any file operation failed.
-     * @throws IllegalArgumentException if the realm file is encrypted. Compacting an encrypted Realm file is not
-     *                                  supported yet.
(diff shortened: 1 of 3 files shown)


realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -1555,10 +1555,13 @@ public static boolean deleteRealm(RealmConfiguration configuration) {
...
      * @param configuration a {@link RealmConfiguration} pointing to a Realm file.
      * @return {@code true} if successful, {@code false} if any file operation failed.
-     * @throws IllegalArgumentException if the realm file is encrypted. Compacting an encrypted Realm file is not
-     *                                  supported yet.
+     * @throws IllegalArgumentException if Realm is synchronized.
      */
     public static boolean compactRealm(RealmConfiguration configuration) {
+        // FIXME: remove this restriction when https://github.com/realm/realm-core/issues/2345 is resolved
+        if (configuration.isSyncConfiguration()) {
+            throw new IllegalArgumentException("You cannot compact a sync'ed Realm.");
```
nhachicha: How about `UnsupportedOperationException("Compacting is not supported yet on synced Realm.  see https://github.com/realm/realm-core/issues/2345");`?

PR #3904, opened 2016-12-14
Update core, sync and object store

beeender:
Core to v2.2.0.
Sync to v1.0.0-BETA-5.0.

Also adapt changes from latest object store.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,11 +28,6 @@
 * Added version number to the native library, preventing ReLinker from accidentally loading old code (#3775).
 * `Realm.getLocalInstanceCount(config)` throwing NullPointerException if called after all Realms have been closed (#3791).
 
-### Internal
-
-* Upgraded Realm Core to 2.1.0.
-* Upgraded Realm Sync to 1.0.0-BETA-5.0.
-
 ## 2.2.0
 
 ### Object Server API Changes (In Beta)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,11 @@
 * Added version number to the native library, preventing ReLinker from accidentally loading old code (#3775).
 * `Realm.getLocalInstanceCount(config)` throwing NullPointerException if called after all Realms have been closed (#3791).
 
+### Internal
+
+* Upgraded Realm Core to 2.1.0.
+* Upgraded Realm Sync to 1.0.0-BETA-5.0.
+
 ## 2.2.0
 
 ### Object Server API Changes (In Beta)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,12 @@
+## 2.2.2
+
+* Fixed "operation not permitted" issue when create Realm file on some devices' external storage (#3629).
+
+### Internal
+
+* Upgraded Realm Core to 2.1.0.
+* Upgraded Realm Sync to 1.0.0-BETA-5.0.
+
 ## 2.2.1
 
 ### Object Server API Changes (In Beta)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 
 ### Bug fixes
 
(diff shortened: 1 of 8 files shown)


beeender:
Build sync fails. Since https://github.com/realm/realm-object-store/pull/179/commits/6b6012cb2d603e20bf82e19c453af0f9bf997894 is removed. What shall we do with it?

beeender:
Also need update the OS sub-module .

CHANGELOG.md
```diff
@@ -9,6 +9,11 @@
 * Added version number to the native library, preventing ReLinker from accidentally loading old code (#3775).
 * `Realm.getLocalInstanceCount(config)` throwing NullPointerException if called after all Realms have been closed (#3791).
 
+### Internal
```
MGaetan89: This should be in 2.2.2, no?
As 2.2.1 was already released.

beeender: Thanks! Fix it right away!

CHANGELOG.md
```diff
@@ -4,11 +4,18 @@
 
 * Disabled `Realm.compactRealm()` when sync is enabled as it might corrupt the Realm (https://github.com/realm/realm-core/issues/2345).
 
+### Bug fixes
+
+* Fixed "operation not permitted" issue when create Realm file on some devices' external storage (#3629).
```
cmelchior: - Remove "Fixed". 
- creating Realm files

realm/realm-library/src/main/cpp/CMakeLists.txt
```diff
@@ -120,7 +120,7 @@ set(WARNING_CXX_FLAGS "-Wall -Wextra -pedantic -Wno-long-long -Wno-variadic-macr
 -Wno-missing-field-initializers -Wmissing-declarations -Wno-error=uninitialized -Wno-error=maybe-uninitialized")
 set(REALM_COMMON_CXX_FLAGS "-DREALM_ANDROID -DREALM_HAVE_CONFIG -DPIC -pthread -fvisibility=hidden -std=c++14 -fsigned-char")
 if (build_SYNC)
-    set(REALM_COMMON_CXX_FLAGS "${REALM_COMMON_CXX_FLAGS} -DREALM_SYNC")
+    set(REALM_COMMON_CXX_FLAGS "${REALM_COMMON_CXX_FLAGS} -DREALM_ENABLE_SYNC=1")
```
cmelchior: Is this more idiomatic or?

realm/realm-library/src/main/cpp/CMakeLists.txt
```diff
@@ -162,13 +162,12 @@ file(GLOB objectstore_SRC
...
-    "object-store/src/util/*.cpp")
+    "object-store/src/util/*.cpp"
+    "object-store/src/impl/epoll/*.cpp"
+    "object-store/src/util/android/*.cpp")
 # Sync needed Object Store files
 if (build_SYNC)
-    file(GLOB objectstore_sync_SRC
-        "object-store/src/sync_manager.cpp"
-        "object-store/src/sync_session.cpp")
+    file(GLOB objectstore_sync_SRC "object-store/src/sync/*")
```
cmelchior: Isn't there also a `impl` directory in the sync folder? GLOB doesn't include that

beeender: We don't actually need anything from that directory. Add it in your SyncManager PR later ?

realm/realm-library/src/main/cpp/CMakeLists.txt
beeender: To align it with OS. And also core normally uses macro in this way.

kneth: Then you can do `#if REALM_ENABLE_SYNC` instead of `#ifdef REALM_ENABLE_SYNC`. Using `#ifdef` is a bit dangerous as `-DREALM_ENABLE_SYNC=NO` still defines `REALM_ENABLE_SYNC`.

PR #3913, opened 2016-12-15
Waiting longer time for checking auth server

beeender:


beeender:
retest this please

beeender:
From log i saw sometimes the ros testing server fails to send the response, maybe 20+50 = 1s sometimes is too short when the docker host is under heave loading. But it seems this only happened on Dec. 14th. Strange.

PR #3919, opened 2016-12-16
SyncUser.currentUser returns null if no current user.

dalinaum:
SyncUser.currentUser throws an exception instead of returning null now. It will be a breaking change.


cmelchior:
This is getting fixed in https://github.com/realm/realm-java/pull/3921 as well. So closing this as #3921 contains the full set of changes for `currentUser`

PR #3921, opened 2016-12-18
Throw correct exception for multiple logged in users

cmelchior:
`SyncUser.currentUser` should throw IllegalStateException if multiple users are logged in.

@realm/java

### DIFF
--- a/realm/realm-library/src/androidTestObjectServer/java/io/realm/SyncUserTests.java
+++ b/realm/realm-library/src/androidTestObjectServer/java/io/realm/SyncUserTests.java
@@ -20,7 +20,6 @@ import android.support.test.InstrumentationRegistry;
 import android.support.test.runner.AndroidJUnit4;
 
 import org.junit.After;
-import org.junit.Assert;
 import org.junit.BeforeClass;
 import org.junit.Rule;
 import org.junit.Test;
@@ -40,7 +39,6 @@ import io.realm.internal.network.AuthenticationServer;
 import io.realm.rule.RunInLooperThread;
 import io.realm.util.SyncTestUtils;
 
-import static io.realm.util.SyncTestUtils.createRandomTestUser;
 import static io.realm.util.SyncTestUtils.createTestUser;
 import static junit.framework.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
--- a/realm/realm-library/src/androidTestObjectServer/java/io/realm/SyncUserTests.java
+++ b/realm/realm-library/src/androidTestObjectServer/java/io/realm/SyncUserTests.java
@@ -104,7 +104,9 @@ public class SyncUserTests {
     }
 
     private AuthenticateResponse getNewRandomUser() {
-        return SyncTestUtils.createLoginResponse(UUID.randomUUID().toString(), Long.MAX_VALUE);
+        String identity = UUID.randomUUID().toString();
+        String userTokenValue = UUID.randomUUID().toString();
+        return SyncTestUtils.createLoginResponse(userTokenValue, identity, Long.MAX_VALUE);
     }
 
     // Test that current user is cleared if it is logged out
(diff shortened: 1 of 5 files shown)


beeender:
Why OS sub module is updated?

realm/realm-library/src/androidTestObjectServer/java/io/realm/SyncUserTests.java
```diff
@@ -20,23 +20,35 @@
 import android.support.test.runner.AndroidJUnit4;
 
 import org.junit.After;
+import org.junit.Assert;
```
zaki50: `Assert` is not used.

realm/realm-library/src/androidTestObjectServer/java/io/realm/SyncUserTests.java
```diff
@@ -20,23 +20,35 @@
...
+import java.net.URL;
 import java.util.Collection;
+import java.util.UUID;
 
+import io.realm.internal.network.AuthenticateResponse;
+import io.realm.internal.network.AuthenticationServer;
 import io.realm.rule.RunInLooperThread;
 import io.realm.util.SyncTestUtils;
 
+import static io.realm.util.SyncTestUtils.createRandomTestUser;
```
zaki50: `createRandomTestUser()` is not used in this class.

realm/realm-library/src/androidTestObjectServer/java/io/realm/SyncUserTests.java
```diff
@@ -74,6 +84,36 @@ public void currentUser_returnsNullIfUserExpired() {
         assertNull(SyncUser.currentUser());
     }
 
+    @Test
+    public void currentUser_throwsIfMultipleUsersLoggedIn() {
+        AuthenticationServer originalAuthServer = SyncManager.getAuthServer();
+        AuthenticationServer authServer = Mockito.mock(AuthenticationServer.class);
+        SyncManager.setAuthServerImpl(authServer);
```
zaki50: I think this line should be moved in the `try` block below to avoid a chance for mock `AuthenticationServer` to be left.

realm/realm-library/src/main/cpp/io_realm_RealmFileUserStore.cpp
```diff
@@ -31,11 +31,15 @@ Java_io_realm_RealmFileUserStore_nativeGetCurrentUser (JNIEnv *env, jclass)
...
-        } else {
+        try {
+            const std::shared_ptr<SyncUser> &user = SyncManager::shared().get_current_user();
+            if (user) {
+                return to_jstring(env, user->refresh_token().data());
+            } else {
+                return nullptr;
+            }
+        } catch (std::logic_error e) {
+            ThrowException(env, IllegalState, "Multiple users have been logged in. `currentUser` is only valid if one single user is logged in.");
```
beeender: The OS has sent a quite meaningful exception message:
```c++
if (std::find_if(std::next(it), m_users.end(), is_active_user) != m_users.end()) {
        throw std::logic_error("Current user is not valid if more that one valid, logged-in user exists.");
```

Do we really need to rephrase it? Or the exception message in the OS should be changed directly.

cmelchior: I'm not convinced (yet) that all logic errors should be converted to ISE, but if it comes up again that seems like a sensible thing to do.

beeender: I just feel really bad when we need to convert exceptions here and there. Or shall we use a different error class in OS in this case? (maybe just subclass the `logic_error` and throw it.) Then we still can have a centralized place to do exception converting.

beeender: I think so? To send a clear exception message is part of OS's responsibility i think.

cmelchior: I'll move it.

PR #3943, opened 2016-12-23
Wait forever in debugging mode

beeender:
It is quite annoy when we stop at a breakpoint the awaitOrFail timeout
happens. By checking if the debugger connected, we can know we are
actually debugging the test and don't want to be interrupted by the
timeout.


Zhuinden:
I ran into this a while ago, it really is annoying.

Issue #813, opened 2015-02-03
Catching JNI Exceptions


ImGauravC:
Following is a way to log JNI exceptions in the Google play console. Can this be baked in? 

http://blog.httrack.com/blog/2013/08/23/catching-posix-signals-on-android/

cmelchior:
Interesting. It is definitely something we will take a closer look at.

PR #3960, opened 2016-12-28
Log the native backtrace when exception throws

beeender:
It is always a pain for us that after converting c++ exception to
java's, all the native backtrace just lost.
This tries to hook into the __cxa_throw to log the native backtrace PCs
when exception throws.
The pc address can be then decoded through addr2line like:
addr2line -C -f -e librealm-jni.so 0xa0293 0xa0548 0xb8cd1 0x3acb3 0x15611

A more wicked idea would be totally implement a __cxa_throw function and
convert exceptions from there and unwind the callstack to where JNI borders
then return to java peacefully. It seems to be doable and we can get rid of
the `try CATCHSTD` blocks for every JNI call. Need more investigations.

### DIFF
--- a/realm/realm-library/src/main/cpp/jni_util/exception_utils.cpp
+++ b/realm/realm-library/src/main/cpp/jni_util/exception_utils.cpp
@@ -1,0 +1,65 @@
+#include <unwind.h>
+#include <dlfcn.h>
+#include <iomanip>
+#include <android/log.h>
+
+struct BacktraceState
+{
+    void** current;
+    void** end;
+};
+
+static _Unwind_Reason_Code unwindCallback(struct _Unwind_Context* context, void* arg)
+{
+    BacktraceState* state = static_cast<BacktraceState*>(arg);
+    uintptr_t pc = _Unwind_GetIP(context);
+    if (pc) {
+        if (state->current == state->end) {
+            return _URC_END_OF_STACK;
+        } else {
+            *state->current++ = reinterpret_cast<void*>(pc);
+        }
+    }
+    return _URC_NO_REASON;
+}
+
+size_t captureBacktrace(void** buffer, size_t max)
+{
+    BacktraceState state = {buffer, buffer + max};
+    _Unwind_Backtrace(unwindCallback, &state);
+
+    return state.current - buffer;
+}
+
+void dumpBacktrace(std::ostream& os, void** buffer, size_t count)
+{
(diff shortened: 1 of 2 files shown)


kneth:
It might be worth comparing with https://github.com/realm/realm-core/blob/master/src/realm/util/terminate.cpp

kneth:
We must also investigate how it works together with Crashlytics: https://docs.fabric.io/android/crashlytics/ndk.html

realm/realm-library/src/main/cpp/jni_util/exception_utils.cpp
```diff
@@ -0,0 +1,65 @@
+#include <unwind.h>
```
kneth: Does API < 16 have `libunwind`?

kneth:
Despite my concerns and comments, I like the idea!

realm/realm-library/src/main/cpp/jni_util/exception_utils.cpp
beeender: Did you see anything about the API 16 with libunwind? I cannot find too much information about that.

But! https://code.google.com/p/android/issues/detail?id=199936
> libunwind is not an NDK API. even if you get this to work now, in future releases you will be prevented from linking against libunwind at runtime.

:( :( :(

kneth:
Might be useful: http://stackoverflow.com/questions/8115192/android-ndk-getting-the-backtrace

beeender:
According to http://stackoverflow.com/a/35585744/1396606 , it seems it is safe for us to use `<unwind.h>` -- we are using gnustl static linked. Also, we are using `visibility=hidden` for build, but we don't need to parse symbols inside the `so`. Just get the backtrace addresses are good enough for us.

realm/realm-library/src/main/cpp/jni_util/exception_utils.cpp
kneth: `unwind.h` exists in both R10e and R13b.

kneth: See https://docs.fabric.io/android/crashlytics/ndk.html#one-line-stack-traces-for-native-crashes for API < 16.

beeender:
Some ideas in #813

kneth:
I think I have heard rumours that the NDK team is working on a new `libunwind`.

Zhuinden:
.

PR #3961, opened 2016-12-29
Enable -Werror and fix warnings

beeender:
Adapt the same warning options from object store

Depends on https://github.com/realm/realm-object-store/pull/308

### DIFF
--- a/realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
+++ b/realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
@@ -1,5 +1,7 @@
 #include "io_realm_internal_SharedRealm.h"
 
+#include <realm/util/features.h>
+
 #include "object_store.hpp"
 #include "shared_realm.hpp"
 
@@ -37,7 +39,7 @@ Java_io_realm_internal_SharedRealm_nativeInit(JNIEnv *env, jclass, jstring tempo
 JNIEXPORT jlong JNICALL
 Java_io_realm_internal_SharedRealm_nativeCreateConfig(JNIEnv *env, jclass, jstring realm_path, jbyteArray key,
         jbyte schema_mode, jboolean in_memory, jboolean cache, jboolean disable_format_upgrade,
-        jboolean auto_change_notification, jstring sync_server_url, jstring /*sync_user_token*/)
+        jboolean auto_change_notification, REALM_UNUSED jstring sync_server_url, jstring /*sync_user_token*/)
 {
     TR_ENTER()
 
@@ -52,13 +54,9 @@ Java_io_realm_internal_SharedRealm_nativeCreateConfig(JNIEnv *env, jclass, jstri
         config->cache = cache;
         config->disable_format_upgrade = disable_format_upgrade;
         config->automatic_change_notifications = auto_change_notification;
-#if REALM_ENABLE_SYNC
         if (sync_server_url) {
             config->force_sync_history = true;
         }
-#else
-        (void)sync_server_url;
-#endif
         return reinterpret_cast<jlong>(config);
     } CATCH_STD()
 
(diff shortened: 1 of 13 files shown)


realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
```diff
@@ -56,6 +56,9 @@ Java_io_realm_internal_SharedRealm_nativeCreateConfig(JNIEnv *env, jclass, jstri
         if (sync_server_url) {
             config->force_sync_history = true;
         }
+#else
+        (void)sync_server_url;
```
nhachicha: is this to remove warnings ?

realm/realm-library/src/main/cpp/util.cpp
```diff
@@ -165,11 +165,14 @@ void ThrowRealmFileException(JNIEnv* env, const std::string& message, realm::Rea
 
     jmethodID constructor = env->GetMethodID(cls, "<init>", "(BLjava/lang/String;)V");
     // Initial value to suppress gcc warning.
-    jbyte kind_code;
+    jbyte kind_code = -1; // To suppress compile warning.
     switch (kind) {
         case realm::RealmFileException::Kind::AccessError:
             kind_code = io_realm_internal_SharedRealm_FILE_EXCEPTION_KIND_ACCESS_ERROR;
             break;
+        case realm::RealmFileException::Kind::BadHistoryError:
```
nhachicha: Did yo update `shared_realm.cpp#translate_file_exception` with the new exception as well?

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
beeender: yes! unused vars

realm/realm-library/src/main/cpp/util.cpp
beeender: the new kind came from the `shared_realm.cpp#translate_file_exception`. a new emum value is added in OS.

dalinaum:
It looks like that #3970 should be merged before this PR, and this PR should update later. Right?

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
```diff
@@ -37,7 +37,7 @@ Java_io_realm_internal_SharedRealm_nativeInit(JNIEnv *env, jclass, jstring tempo
 JNIEXPORT jlong JNICALL
 Java_io_realm_internal_SharedRealm_nativeCreateConfig(JNIEnv *env, jclass, jstring realm_path, jbyteArray key,
         jbyte schema_mode, jboolean in_memory, jboolean cache, jboolean disable_format_upgrade,
-        jboolean auto_change_notification, jstring sync_server_url, jstring sync_user_token)
+        jboolean auto_change_notification, jstring sync_server_url, jstring /*sync_user_token*/)
```
beeender: I can remove `sync_user_token` if it is not going to be used by the OS sync manager.

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
```diff
@@ -37,7 +39,7 @@ Java_io_realm_internal_SharedRealm_nativeInit(JNIEnv *env, jclass, jstring tempo
 JNIEXPORT jlong JNICALL
 Java_io_realm_internal_SharedRealm_nativeCreateConfig(JNIEnv *env, jclass, jstring realm_path, jbyteArray key,
         jbyte schema_mode, jboolean in_memory, jboolean cache, jboolean disable_format_upgrade,
-        jboolean auto_change_notification, jstring sync_server_url, jstring sync_user_token)
+        jboolean auto_change_notification, REALM_UNUSED jstring sync_server_url, jstring /*sync_user_token*/)
```
cmelchior: This is slightly confusing, since it _is_ used for sync configurations, but guess it is acceptable vs. the gain we for catching all other warnings.

kneth: I agree: `REALM_MAYBE_UNUSED` would be a better name.

PR #3970, opened 2016-12-31
Update core to 2.3.0

beeender:
Also update object-store to 99570ba6e0 .

### DIFF
--- a/realm/realm-library/src/main/cpp/objectserver_shared.hpp
+++ b/realm/realm-library/src/main/cpp/objectserver_shared.hpp
@@ -54,14 +54,14 @@ public:
                 coordinator->wake_up_notifier_worker();
             }
         };
-        auto error_handler = [&, global_obj_ref_tmp](int error_code, std::string message) {
+        auto error_handler = [&, global_obj_ref_tmp](std::error_code error_code, bool /*is_fatal*/, const std::string message) {
             JNIEnv *local_env;
             g_vm->AttachCurrentThread(&local_env, nullptr);
             jclass java_session_class = local_env->GetObjectClass(global_obj_ref_tmp);
             jmethodID notify_error_handler = local_env->GetMethodID(java_session_class,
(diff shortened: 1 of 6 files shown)


dalinaum:
retest this please

realm/realm-library/src/main/cpp/objectserver_shared.hpp
```diff
@@ -60,7 +61,19 @@ class JniSession {
                 coordinator->wake_up_notifier_worker();
             }
         };
-        auto error_handler = [weak_session_ref](std::error_code error_code, bool /*is_fatal*/, const std::string message) {
+        auto error_handler = [weak_session_ref](std::error_code error_code, bool is_fatal, const std::string message) {
+            if (error_code.category() != realm::sync::protocol_error_category() ||
+                    error_code.category() != realm::sync::client_error_category()) {
+                // FIXME: Consider below when moving to the OS sync manager.
```
cmelchior: Wouldn't finding a way to kill the C++ thread after a Java exception fix this? It would mean we also need to find a way to restart the client (although I believe OS does that by itself), so sounds doable?

beeender: OS just returns https://github.com/realm/realm-object-store/pull/303/files#diff-f3165f1bbea4d4c8abf9c03a4b4e86f9R385 when this happens.

If it is an fatal error, we need to do something with it. But we are not doing it before this PR, so I think we should handle it in another PR or in the OS sync manager PR.

cmelchior: Ok, at least we are consistent with OS then. I find it problematic that we just ignore unknown errors, but lets have that discussion elsewhere.

Zhuinden: at least logging it would probably help

PR #3834, opened 2016-11-23
Integrate OS Results & Async query

beeender:
TODO:
- [x] Check the query's ownership. OS `Results` moves the param `Query` when construct. But in java the `RealmQuery` still need it. Consider clone it before create `Results`.
- [x] `RealmResults` in looper thread.
- [x] `RealmResults` in non-looper thread.
- [x] Behaviour of `RealmResults`'s iteration when deletion happens. (#3883)
- [x] Kill `TableOrView`, yeah.
- [x] `NotificationToken` destructor's thread safety.
- [x] To support `distinctAsync()`. This needs to be added in the OS side, maybe create a mode `Results.Mode.distinct()` in OS `Results`? Relevant PR https://github.com/realm/realm-object-store/pull/235 https://github.com/realm/realm-object-store/pull/294
- [x] Support `findFirstAsync()`.
- [x] Check if the `distinct()` returns a new `RealmResults` object, and should it return a new object.
- [x] `ProxyState` is public because of `RealmObjectProxy` is in the internal package.- (Do it later #4044)
- [x] Figure out how to deliver async transaction callbacks.
- [ ] Check the exception thrown problem in the external_commit_helper thread.
- [x] Memory with `SortDescriptor`
- [x] Aggregates (https://github.com/realm/realm-object-store/pull/288)

TODOs moved from #3372 

- [x] Use `Results` from ObejctStore
- [x] Just use async query from OS? Maybe remove FIXME in `RealmNotifier`?
- [x] Move `HandlerControler` and `AndroidNotifier` to `io.realm.internal`
- [x] FIXME in `AndroidNotifier`, simplify the calling routing?
- [x] Check the use case in https://github.com/realm/realm-java/issues/3668#issuecomment-255378795. `Realm.where(AllTypes.class).findAllAsync().where().findAllAsync()`. It is useful to build async queries for https://github.com/realm/realm-java/issues/3668#issuecomment-255351407

### See https://gist.github.com/beeender/9a37c81d56326f02be734a3087da0f84 for more details

- [x] For a local commit, Realm change listener will be called before the `beginTransaction()` returns or `commitTransaction()` returns. Is it acceptable?

## Things don't impact review but have to be solved before MERGE
- [ ] Close realm inside `did_change` callbacks https://github.com/realm/realm-object-store/pull/318
- [x] Update core to `v2.3.1` to solve the linkview base query handover.
- [ ] CHECK THE OBJECT STORE SUBMODULE VERSION IN THIS PR!!!
- [ ] Async and db size

PR #3977, opened 2017-01-03
RFC: Finegrained notifications API

cmelchior:
Proposal for the Java parts of https://github.com/realm/realm-java/pull/3834

All the code is just prototyping for the `FINEGRAINED_NOTIFICATIONS.md` document. This branch will be deleted after review is complete.

@realm/java 

Don't try to make this compile. It probably won't :)

### DIFF
--- a/realm/realm-library/src/main/java/io/realm/ObservableCollection.java
+++ b/realm/realm-library/src/main/java/io/realm/ObservableCollection.java
@@ -4,7 +4,7 @@ package io.realm;
  * Interface for collections that are able to tell interested parties how they
  * changed when they where updated.
  */
-public interface ObservableCollection<T, L> extends Observable<T> {
-    void addChangeListener(L listener);
-    void removeChangeListener(L listener);
+public interface ObservableCollection<T, S> extends Observable<T> {
+    void addChangeListener(S listener);
+    void removeChangeListener(S listener);
 }
--- a/realm/realm-library/src/main/java/io/realm/ObservableCollection.java
+++ b/realm/realm-library/src/main/java/io/realm/ObservableCollection.java
@@ -1,0 +1,10 @@
+package io.realm;
+
+/**
+ * Interface for collections that are able to tell interested parties how they
+ * changed when they where updated.
+ */
+public interface ObservableCollection<T, L> extends Observable<T> {
+    void addChangeListener(L listener);
+    void removeChangeListener(L listener);
+}
--- a/realm/realm-library/src/objectServer/java/io/realm/OrderedCollectionChange.java
+++ b/realm/realm-library/src/objectServer/java/io/realm/OrderedCollectionChange.java
@@ -11,7 +11,7 @@ public class OrderedCollectionChange {
     public Range[] getDeletionRanges() { return null; }
     public Range[] getInsertionRanges() { return null; }
     public Range[] getChangeRanges() { return null; }
-    Move[] getMoves() { return null; }
+    public Move[] getMoves() { return null; }
 
     public static class Range {
         public final long startIndex;
--- a/realm/realm-library/src/objectServer/java/io/realm/OrderedCollectionChange.java
+++ b/realm/realm-library/src/objectServer/java/io/realm/OrderedCollectionChange.java
@@ -1,0 +1,35 @@
+package io.realm;
+
+/**
+ * This interface describes the changes made to a collection during the last
+ * update.
+ */
+public class OrderedCollectionChange {
+    public long[] getDeletions() { return null; }
+    public long[] getInsertertions() { return null; }
+    public long[] getChanges()  { return null; }
+    public Range[] getDeletionRanges() { return null; }
+    public Range[] getInsertionRanges() { return null; }
+    public Range[] getChangeRanges() { return null; }
+    Move[] getMoves() { return null; }
+
+    public static class Range {
+        public final long startIndex;
+        public final long length;
+
+        public Range(long startIndex, long length) {
+            this.startIndex = startIndex;
+            this.length = length;
+        }
+    }
+
+    public static class Move {
+        public final long oldIndex;
+        public final long newIndex;
+
+        public Move(long oldIndex, long newIndex) {
+            this.oldIndex = oldIndex;
+            this.newIndex = newIndex;
+        }
+    }
+}
--- a/rfcs/FINEGRAINED_NOTIFICATIONS.md
+++ b/rfcs/FINEGRAINED_NOTIFICATIONS.md
@@ -10,9 +10,9 @@ Realm wants to promote a reactive pattern to working with your data. It should b
 
 This is interesting for a number of use cases:
 
-1) I want to animate or refresh a RecyclerView when elements are added, moved, modified or deleted in a `RealmResults`.
-2) I want to be notified about the result of an query changing so I can refresh my UI.
-3) I want my changelisteners to _only_ trigger when there is a relevant change.
+1. I want to animate or refresh a RecyclerView when elements are added, moved, modified or deleted in a `RealmResults`.
+2. I want to be notified about the result of an query changing so I can refresh my UI.
+3. I want my changelisteners to _only_ trigger when there is a relevant change.
 
 While you in some use cases probably want to compare old and new data, we have not yet found a compelling enough use case to support this for collections.
 Single object notifications will have this functionality.
@@ -61,12 +61,12 @@ public interface OrderedRealmCollectionChangeListener<T> {
 // RecyclerView uses the `get*Ranges()` methods.
 public class OrderedCollectionChange {
     public long[] getDeletions() { return null; }
-    public long[] getInsertertions() { return null; }
+    public long[] getInsertions() { return null; }
     public long[] getChanges()  { return null; }
     public Range[] getDeletionRanges() { return null; }
     public Range[] getInsertionRanges() { return null; }
     public Range[] getChangeRanges() { return null; }
-    Move[] getMoves() { return null; }
+    public Move[] getMoves() { return null; }
 
     public static class Range {
         public final long startIndex;
--- a/rfcs/FINEGRAINED_NOTIFICATIONS.md
+++ b/rfcs/FINEGRAINED_NOTIFICATIONS.md
@@ -1,0 +1,122 @@
+# Finegrained notifications
+
+## Summary
+
+This document describes an API proposal for a finegrained collection for Realm Java. Object-level notifications will be covered in another document.
+
+## Motivation
+
+Realm wants to promote a reactive pattern to working with your data. It should be possible to define a query for any change you are interested in and just depend on Realm figuring out when that change has happend and notify you. 
+
+This is interesting for a number of use cases:
+
+1) I want to animate or refresh a RecyclerView when elements are added, moved, modified or deleted in a `RealmResults`.
+2) I want to be notified about the result of an query changing so I can refresh my UI.
+3) I want my changelisteners to _only_ trigger when there is a relevant change.
+
+While you in some use cases probably want to compare old and new data, we have not yet found a compelling enough use case to support this for collections.
+Single object notifications will have this functionality.
+
+## Current situation
+
+- Right now Realm Java changelisteners are "table-based", which means that a change to any element in a table will trigger all changelisteners registered on that table, even though the element wasn't included in the query. This results in a lot of un-needed change events being triggered.
+
+- Single object change listeners work the same way, which is even worse.
+
+- Changelisteners are not supported on RealmList
+
+- Cocoa has support for fine-grained collection changes.
+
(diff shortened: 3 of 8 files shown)


rfcs/FINEGRAINED_NOTIFICATIONS.md
```diff
@@ -0,0 +1,122 @@
...
+
+
+## Detailed design
+
+Add these new interfaces and class:
+
+```
+// New generic interface for generic changes (= our current changelisteners)
+// All relevant classes should implement this: Realm, RealmResults, RealmList, RealmObject, DynamicRealm, DynamicRealmObject
+// TODO: Is Observable the best name?
```
beeender: I like the name but feel it may confuse user with RxJava's same name interface. But I don't have a better suggestion.

rfcs/FINEGRAINED_NOTIFICATIONS.md
```diff
@@ -0,0 +1,122 @@
...
+public interface Observable<T> {
+    void addChangeListener(RealmChangeListener<T> listener);
+    void removeChangeListener(RealmChangeListener<T> listener);
+    void removeAllChangeListeners(); // Current name is removeChangeListeners? IMO this name is better, but do we want the penalty of a rename?
+}
+
+// New interface for collections. Add another changelistener for fine-grained notifications
+public interface ObservableCollection<T, L> extends Observable<T> {
```
beeender: `L` is strange, see Type Parameter Naming Conventions http://docs.oracle.com/javase/tutorial/java/generics/types.html

rfcs/FINEGRAINED_NOTIFICATIONS.md
```diff
@@ -0,0 +1,122 @@
...
+    void addChangeListener(L listener);
```
beeender: I am a little confused about the interface inheritance mixed with generic types, so for the collection class, it has to implement 5 methods?

realm/realm-library/src/objectServer/java/io/realm/OrderedCollectionChange.java
```diff
@@ -0,0 +1,35 @@
...
+ * update.
+ */
+public class OrderedCollectionChange {
+    public long[] getDeletions() { return null; }
+    public long[] getInsertertions() { return null; }
+    public long[] getChanges()  { return null; }
+    public Range[] getDeletionRanges() { return null; }
+    public Range[] getInsertionRanges() { return null; }
+    public Range[] getChangeRanges() { return null; }
+    Move[] getMoves() { return null; }
```
zaki50: public?

rfcs/FINEGRAINED_NOTIFICATIONS.md
```diff
@@ -0,0 +1,122 @@
...
+// Primary purpose is to make RecyclerView animations easy, but should also be general purpose
+// RecyclerView uses the `get*Ranges()` methods.
+public class OrderedCollectionChange {
+    public long[] getDeletions() { return null; }
+    public long[] getInsertertions() { return null; }
+    public long[] getChanges()  { return null; }
+    public Range[] getDeletionRanges() { return null; }
+    public Range[] getInsertionRanges() { return null; }
+    public Range[] getChangeRanges() { return null; }
+    Move[] getMoves() { return null; }
```
zaki50: public?

rfcs/FINEGRAINED_NOTIFICATIONS.md
```diff
@@ -0,0 +1,122 @@
...
+    void onChange(T collection, OrderedCollectionChange changes);
+}
+
+// Class for describing changes
+// Primary purpose is to make RecyclerView animations easy, but should also be general purpose
+// RecyclerView uses the `get*Ranges()` methods.
+public class OrderedCollectionChange {
+    public long[] getDeletions() { return null; }
+    public long[] getInsertertions() { return null; }
+    public long[] getChanges()  { return null; }
```
beeender: `getChanges` or `getModifications`? the `modifications` is used by object store. If there is no strong reason to use  `changes` maybe we should stay with the name in object store? Also, think about the API name `addChangeListener`, `change` there means deletion and insertion as well.

rfcs/FINEGRAINED_NOTIFICATIONS.md
```diff
@@ -0,0 +1,122 @@
...
+// Using OrderedCollectionChange to prepare for change events from unordered data structures.
+public interface OrderedRealmCollectionChangeListener<T> {
+    void onChange(T collection, OrderedCollectionChange changes);
+}
+
+// Class for describing changes
+// Primary purpose is to make RecyclerView animations easy, but should also be general purpose
+// RecyclerView uses the `get*Ranges()` methods.
+public class OrderedCollectionChange {
+    public long[] getDeletions() { return null; }
```
beeender: This returns the old index? Probably i need to play with the recycle view to understand how does this work.

rfcs/FINEGRAINED_NOTIFICATIONS.md
cmelchior: Yes, my bad

rfcs/FINEGRAINED_NOTIFICATIONS.md
cmelchior: Yes, I had the same thought process. `RealmObservable` might also work, but ideas are welcome.

Zhuinden: `RealmObservable`?

realm/realm-library/src/main/java/io/realm/ObservableCollection.java
```diff
@@ -0,0 +1,10 @@
+package io.realm;
+
+/**
+ * Interface for collections that are able to tell interested parties how they
+ * changed when they where updated.
+ */
+public interface ObservableCollection<T, L> extends Observable<T> {
+    void addChangeListener(L listener);
```
zaki50: Why `L` instead of `OrderedRealmCollectionChangeListener<T>`?

rfcs/FINEGRAINED_NOTIFICATIONS.md
cmelchior: Yes, the Observable interface adds the generic change listeners we use today. This interface then extends it with support for change notifications. We will probably use a similar approach for object level notifications with an ObservableItem interface. At least that was my idea with this approach.

rfcs/FINEGRAINED_NOTIFICATIONS.md
cmelchior: I am not sure either, most likely the RecyclerView internally tracks the index of the items visible on the screen, so it knows which ones to remove

rfcs/FINEGRAINED_NOTIFICATIONS.md
cmelchior: The wording came from the RecyclerViewAdapter API: https://developer.android.com/reference/android/support/v7/widget/RecyclerView.Adapter.html#notifyItemChanged(int, java.lang.Object), but I see your point. I'm slightly leaning towards using the the RecyclerView API term `getChanges()`, but wonder what the rest say?

rfcs/FINEGRAINED_NOTIFICATIONS.md
cmelchior: Yes, my bad

realm/realm-library/src/objectServer/java/io/realm/OrderedCollectionChange.java
cmelchior: Yes, my bad

rfcs/FINEGRAINED_NOTIFICATIONS.md
```diff
@@ -0,0 +1,122 @@
...
+
+This document describes an API proposal for a finegrained collection for Realm Java. Object-level notifications will be covered in another document.
+
+## Motivation
+
+Realm wants to promote a reactive pattern to working with your data. It should be possible to define a query for any change you are interested in and just depend on Realm figuring out when that change has happend and notify you. 
+
+This is interesting for a number of use cases:
+
+1) I want to animate or refresh a RecyclerView when elements are added, moved, modified or deleted in a `RealmResults`.
```
cmelchior: Not entirely sure what you mean? How should it be changed?

beeender: use `.` instead of `)` to make it look better 🍡 
1. I want to animate or refresh a RecyclerView when elements are added, moved, modified or deleted in a `RealmResults`.
2. I want to be notified about the result of an query changing so I can refresh my UI.
3. I want my changelisteners to _only_ trigger when there is a relevant change.

realm/realm-library/src/objectServer/java/io/realm/OrderedCollectionChange.java
```diff
@@ -0,0 +1,35 @@
+package io.realm;
+
+/**
+ * This interface describes the changes made to a collection during the last
+ * update.
+ */
+public class OrderedCollectionChange {
+    public long[] getDeletions() { return null; }
+    public long[] getInsertertions() { return null; }
```
Zhuinden: `getInsertions` sounds more valid than `insertertion`

realm/realm-library/src/main/java/io/realm/ObservableCollection.java
cmelchior: I'm future-proofing this for other collection types, e.g. Set and Map are not ordered so if they want to expose their own change notification we don't need to have a `ObservableOrderedCollection` and ObservableUnorderdCollection` interface.

rfcs/FINEGRAINED_NOTIFICATIONS.md
tgoyne: I went with "modifications" mostly because I found that the word "changes" was getting too overloaded when writing the docs. Using the same word in the class name and one of the property names makes it hard to word things.

realm/realm-library/src/objectServer/java/io/realm/OrderedCollectionChange.java
```diff
@@ -11,7 +11,7 @@
     public Range[] getDeletionRanges() { return null; }
     public Range[] getInsertionRanges() { return null; }
     public Range[] getChangeRanges() { return null; }
-    Move[] getMoves() { return null; }
+    public Move[] getMoves() { return null; }
```
Zhuinden: btw, is there a reason why this class specifically is in `objectServer` flavor?

rfcs/FINEGRAINED_NOTIFICATIONS.md
```diff
@@ -0,0 +1,122 @@
...
+    public long[] getInsertions() { return null; }
+    public long[] getChanges()  { return null; }
+    public Range[] getDeletionRanges() { return null; }
+    public Range[] getInsertionRanges() { return null; }
+    public Range[] getChangeRanges() { return null; }
+    public Move[] getMoves() { return null; }
+
+    public static class Range {
+        public final long startIndex;
+        public final long length;
```
kneth: Why not `endIndex`? (to me, range is almost identical to interval)

rfcs/FINEGRAINED_NOTIFICATIONS.md
```diff
@@ -0,0 +1,122 @@
...
+
+### Future features that might impact this
+
+* Object-level notifications: Will probably be another interface that extends `Oberservable`
+* Unordered data sets: ObservableCollection does not put any restrictions there, and we can create `UnorderedCollectionChange` objects and `UnorderedRealmCollectionChangeListener` interfaces. This can be done without breaking changes.
+* Changes to what type of changes we want to expose. Since `OrderedCollectionChange` is a class and not an interface, extending it with new functionality can be done without breaking changes.
+
+## Work required
+
+- [ ] We need to move `RealmResults` to be backed by `Results` from Object Store instead.
```
kneth: In progress :-)

rfcs/FINEGRAINED_NOTIFICATIONS.md
cmelchior: Mostly because the RecyclerView API uses `startIndex` + `length` and that is our primary use case. `endIndex` can be calculated from `length` and vice versa.

kneth: Fine with me. Maybe rename `startIndex` to `offset`.

cmelchior: `offset` seems harder to understand than `startIndex`, but maybe that is just me?

Zhuinden: I think `startIndex` is more obvious, because I'm not sure by first look what `offset` refers to

rfcs/FINEGRAINED_NOTIFICATIONS.md
eygraber: What about `Changeable`?

cmelchior:
API merged. Closing

PR #3995, opened 2017-01-06
Fix memory leak (#3993)

### LINKED ISSUE #3993: Memory leak (Row) in Java_io_realm_internal_Table_nativeSetPrimaryKey
I just stumbled across this while investigating a Core crash 
seems we're allocating a Row on the heap 

https://github.com/realm/realm-java/blob/d547fde64271229cb6bb98df30deda05faeba9ce/realm/realm-library/src/main/cpp/io_realm_internal_Table.cpp#L1508

kneth:
Closes #3993

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 2.2.3
+
+### Bug fixes
+
+* Fixed memory leak in `Java_io_realm_internal_Table_nativeSetPrimaryKey` (#3993).
+
 ## 2.2.2
 
 ### Object Server API Changes (In Beta)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### Bug fixes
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## 2.2.3
+
+### Bug fixes
+
+* Fixed memory leak in `Java_io_realm_internal_Table_nativeSetPrimaryKey` (#3993).
```
cmelchior: Minor, could this be rephrased to be more meaningful for users since `Java_io_realm_internal_Table_nativeSetPrimaryKey` is just one internal (native) method?

kneth: Well, I considered something like "Fixed native memory leak when setting value of primary key"

cmelchior: More like "when creating primary key metadata"?. This leak only occurs when setting up the primary key table.

cmelchior: Oh wait, no. Sorry, your right.

kneth: I'll try to reformulate

Issue #3810, opened 2016-11-17
RealmThreadPoolExecutor: desired CORE_POOL_SIZE != calculated


diegomontoya:
```java
Runtime.getRuntime().availableProcessors()
```

Returns the number of online/active cores in Android. If the class is loaded while the phone was in heavy sleep, it may report value of 1 even when it has 8 cores which translates to a much lower CORE_POOL_SIZE for the duration of the app than desired. 

Referenced from http://stackoverflow.com/questions/10133570/availableprocessors-returns-1-for-dualcore-phones.

Zhuinden:
That's interesting because I specifically remember that that executor is initialized based on code in AsyncTask, but apparently that was also a lie already.

cmelchior:
Thanks @diegomontoya Viable implementation seems to be here: http://stackoverflow.com/questions/7962155/how-can-you-detect-a-dual-core-cpu-on-an-android-device-from-code/10377934#10377934

diegomontoya:
@cmelchior Yes. That linked implementation does work on all android we have tested but missing 1 optimization. It should compile the pattern in the class and re-use it for matching instead of potentially re-generating the regex on each filterFilter accept method call. In android, a regex pattern is a heavy duty object with native ptr and finalizer. 

@Zhuinden We take Android for what it is. 99% open, lots of bad code, lots of great code, and documentation written by robots. =)

diegomontoya:
1) Start app. Do nothing.
2) Have app thread.sleep for 60s then log core count.
3) Turn off screen.  (trigger device sleep mode)
4) Check log. 

I have not tried above but it should trigger the issue.

diegomontoya:
Come to think of it, the easier way to test is with adb. Force device to go into sleep mode via adb command and log the cpu core api output.

bmeike:
@diegomontoya Yeah... definitely.  I was hoping, though, for something that I could put in a unit test...

bmeike:
Closing, with the merge of 3997.  A new issue 4006, explores an explicit native solution.

PR #3997, opened 2017-01-06
Count 'cpuX' files in /sys/devices/system/cpu as a more accurate way

bmeike:
of counting processors and setting the max threads for the Executor. (#3810)

### DIFF
--- a/realm/realm-library/src/main/java/io/realm/internal/async/RealmThreadPoolExecutor.java
+++ b/realm/realm-library/src/main/java/io/realm/internal/async/RealmThreadPoolExecutor.java
@@ -59,12 +59,27 @@ public class RealmThreadPoolExecutor extends ThreadPoolExecutor {
         return new RealmThreadPoolExecutor(1, 1);
     }
 
+    /**
+     * Try using the number of files named 'cpuNN' in sysfs to figure out the number of
+     * processors on this device. `Runtime.getRuntime().availableProcessors()` may return
+     * a smaller number when the device is sleeping.
+     *
+     * @return the number of threads to be allocated for the executor pool
+     */
     @SuppressWarnings("DMI_HARDCODED_ABSOLUTE_FILENAME")
     private static int calculateCorePoolSize() {
         int cpus = countFilesInDir(SYS_CPU_DIR, "cpu[0-9]+");
+        if (cpus <= 0) {
+            cpus = Runtime.getRuntime().availableProcessors();
+        }
         return (cpus <= 0) ? 1 : (cpus * 2) + 1;
     }
 
+    /**
+     * @param dirPath A directory path
+     * @param pattern A regex
+     * @return the number of files, in the `dirPath` directory, whose names match `pattern`
+     */
     private static int countFilesInDir(String dirPath, String pattern) {
         final Pattern filePattern = Pattern.compile(pattern);
         try {
@@ -75,7 +90,7 @@ public class RealmThreadPoolExecutor extends ThreadPoolExecutor {
                 }
             });
             return (files == null) ? 0 : files.length;
-        } catch (Exception e) {
+        } catch (SecurityException ignore) {
         }
         return 0;
     }
--- a/realm/realm-library/src/main/java/io/realm/internal/async/RealmThreadPoolExecutor.java
+++ b/realm/realm-library/src/main/java/io/realm/internal/async/RealmThreadPoolExecutor.java
@@ -16,6 +16,8 @@
 
 package io.realm.internal.async;
 
+import java.io.File;
+import java.io.FileFilter;
 import java.util.concurrent.ArrayBlockingQueue;
 import java.util.concurrent.Callable;
 import java.util.concurrent.Future;
@@ -23,8 +25,7 @@ import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.locks.Condition;
 import java.util.concurrent.locks.ReentrantLock;
-
-import io.realm.Realm;
+import java.util.regex.Pattern;
 
 /**
  * Custom thread pool settings, instances of this executor can be paused, and resumed, this will also set
@@ -32,9 +33,10 @@ import io.realm.Realm;
  * <a href="https://developer.android.com/training/multiple-threads/define-runnable.html"> Androids recommendation</a>.
  */
 public class RealmThreadPoolExecutor extends ThreadPoolExecutor {
-    // reduce context switch by using a number of thread proportionate to the number of cores
-    // from AOSP https://android.googlesource.com/platform/frameworks/base/+/refs/heads/master/core/java/android/os/AsyncTask.java#182
-    private static final int CORE_POOL_SIZE = Runtime.getRuntime().availableProcessors() * 2 + 1;
+    private static final String SYS_CPU_DIR = "/sys/devices/system/cpu/";
+
+    // reduce context switching by using a number of thread proportionate to the number of cores
+    private static final int CORE_POOL_SIZE = countFilesInDir(SYS_CPU_DIR, "cpu[0-9]+") * 2 + 1;
     private static final int QUEUE_SIZE = 100;
 
     private boolean isPaused;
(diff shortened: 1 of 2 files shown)


Zhuinden:
But do you always have access to this data?

realm/realm-library/src/main/java/io/realm/internal/async/RealmThreadPoolExecutor.java
```diff
@@ -16,25 +16,29 @@
...
 /**
  * Custom thread pool settings, instances of this executor can be paused, and resumed, this will also set
  * appropriate number of Threads & wrap submitted tasks to set the thread priority according to
  * <a href="https://developer.android.com/training/multiple-threads/define-runnable.html"> Androids recommendation</a>.
  */
 public class RealmThreadPoolExecutor extends ThreadPoolExecutor {
-    // reduce context switch by using a number of thread proportionate to the number of cores
-    // from AOSP https://android.googlesource.com/platform/frameworks/base/+/refs/heads/master/core/java/android/os/AsyncTask.java#182
-    private static final int CORE_POOL_SIZE = Runtime.getRuntime().availableProcessors() * 2 + 1;
+    private static final String SYS_CPU_DIR = "/sys/devices/system/cpu/";
```
cmelchior: Is this guaranteed to always be here or can vendors change it? ( I'm looking at you Samsung)

realm/realm-library/src/main/java/io/realm/internal/async/RealmThreadPoolExecutor.java
```diff
@@ -55,6 +59,27 @@ public static RealmThreadPoolExecutor newSingleThreadExecutor() {
         return new RealmThreadPoolExecutor(1, 1);
     }
 
+    @SuppressWarnings("DMI_HARDCODED_ABSOLUTE_FILENAME")
+    private static int calculateCorePoolSize() {
+        int cpus = countFilesInDir(SYS_CPU_DIR, "cpu[0-9]+");
+        return (cpus <= 0) ? 1 : (cpus * 2) + 1;
+    }
+
+    private static int countFilesInDir(String dirPath, String pattern) {
```
cmelchior: Perhaps a comment about what this function does? Is there a file pr. CPU?

realm/realm-library/src/main/java/io/realm/internal/async/RealmThreadPoolExecutor.java
cmelchior: Also a comment about what this path represents would be nice.

bmeike: Yeah.  I was thinking about this.  What would you think about defaulting to `Runtime.getRuntime().availableProcessors()`

cmelchior: Sounds fine with me

bmeike:
Comments addressed.  Please revisit when you have a chance.

realm/realm-library/src/main/java/io/realm/internal/async/RealmThreadPoolExecutor.java
kneth: Sysfs as been part of Linux since 2.5.44, and Android Froyo was based on 2.6.32. A vendor might add custom SELinux rules to prevent access to `/sysfs`.

realm/realm-library/src/main/java/io/realm/internal/async/RealmThreadPoolExecutor.java
```diff
@@ -55,6 +59,42 @@ public static RealmThreadPoolExecutor newSingleThreadExecutor() {
...
     }
 
+    /**
+     * Try using the number of files named 'cpuNN' in sysfs to figure out the number of
+     * processors on this device. `Runtime.getRuntime().availableProcessors()` may return
+     * a smaller number when the device is sleeping.
+     *
+     * @return the number of threads to be allocated for the executor pool
+     */
+    @SuppressWarnings("DMI_HARDCODED_ABSOLUTE_FILENAME")
```
kneth: `@SuppressFBWarnings`?

realm/realm-library/src/main/java/io/realm/internal/async/RealmThreadPoolExecutor.java
```diff
@@ -55,6 +59,42 @@ public static RealmThreadPoolExecutor newSingleThreadExecutor() {
...
+        final Pattern filePattern = Pattern.compile(pattern);
+        try {
+            File[] files = new File(dirPath).listFiles(new FileFilter() {
+                @Override
+                public boolean accept(File file) {
+                    return filePattern.matcher(file.getName()).matches();
+                }
+            });
+            return (files == null) ? 0 : files.length;
+        } catch (SecurityException ignore) {
```
cmelchior: Is there any chance any other exception will be thrown, incl. runtime exceptions? Would it be safer to just catch the generic Exception, since getting this value correct is not critical the method should probably be as failsafe as possible?

realm/realm-library/src/main/java/io/realm/internal/async/RealmThreadPoolExecutor.java
```diff
@@ -55,6 +59,42 @@ public static RealmThreadPoolExecutor newSingleThreadExecutor() {
...
+
+    /**
+     * @param dirPath A directory path
+     * @param pattern A regex
+     * @return the number of files, in the `dirPath` directory, whose names match `pattern`
+     */
+    private static int countFilesInDir(String dirPath, String pattern) {
+        final Pattern filePattern = Pattern.compile(pattern);
+        try {
+            File[] files = new File(dirPath).listFiles(new FileFilter() {
```
nhachicha: We should check if the File  [exists()] (https://developer.android.com/reference/java/io/File.html#exists()) & readable`isDirectory()` & `canRead`

kneth:
I like @nhachicha's suggestion of using a C++11 function.

cmelchior:
I would think that Java just called down to native anyway? A quick search does reveal it does some magic there, but it turned out to be a bit more troublesome exactly what

nhachicha:
we're already shipping with the ObjectStore, that uses c++14.

I tested `std::thread::hardware_concurrency()` on Genymotion Froyo 2.3.7 & it reported 1 Core 
similar to the content of `/sys/devices/system/cpu/` 
```
ls -l /sys/devices/system/cpu/
-r--r--r-- root     root         4096 2017-01-09 16:08 online
-r--r--r-- root     root         4096 2017-01-09 16:08 possible
-r--r--r-- root     root         4096 2017-01-09 16:08 present
-r--r--r-- root     root         4096 2017-01-09 16:08 kernel_max
-r--r--r-- root     root         4096 2017-01-09 16:08 offline
drwxr-xr-x root     root              2017-01-09 15:20 cpuidle
drwxr-xr-x root     root              2017-01-09 15:20 cpu0
```

IMO it's better to use this supported standard NDK method.

realm/realm-library/src/main/java/io/realm/internal/async/RealmThreadPoolExecutor.java
bmeike: No need.  `listFiles` is documented to return null for any of those failures.

bmeike:
I like the idea of using some native method.  Counting files seems like a hack.  If there is some reliable source of truth on this, I'd be delighted to do use it instead.  Please enlighten me.

realm/realm-library/src/main/java/io/realm/internal/async/RealmThreadPoolExecutor.java
bmeike: Yes, exactly.  In that case we revert to `availableProcessors()`.

The whole mess seems pretty ad-hoc, to me.  Surely the kernel knows how many CPUs there are.  Surely with our super C powers, we can ask it.

realm/realm-library/src/main/java/io/realm/internal/async/RealmThreadPoolExecutor.java
bmeike: Sure.  Could do that.  I think this is pretty much equivalent, and it doesn't require a new `provided` library.

cmelchior:
@bmeike Was this merged by mistake?

bmeike:
I merged it.  I see green checkmarks from you and beeender.

cmelchior:
We took the discussion on Slack. A new issue will be created with the remaining work.

PR #4028, opened 2017-01-12
Ignore exceptions

### LINKED ISSUE #3559: [Sync] Crashes in the error handler is swallowed by JNI
If the SyncClients error handler is called, it will call back to Java, but if there is an exception in the Java code for whatever reason, that error gets completely swallowed. We need to fix that somehow, so some error information surface somewhere.

kneth:
If an exception is thrown in an async error handler, the app's behaviour is highly unpredictable. We swallow the exception and log that it happens but let the app continue.

Closes #3559

### DIFF
--- a/realm/realm-library/src/objectServer/java/io/realm/SyncUser.java
+++ b/realm/realm-library/src/objectServer/java/io/realm/SyncUser.java
@@ -44,7 +44,6 @@ import io.realm.internal.network.LogoutResponse;
 import io.realm.internal.objectserver.ObjectServerUser;
 import io.realm.internal.objectserver.Token;
 import io.realm.log.RealmLog;
-import io.realm.permissions.PermissionChange;
 import io.realm.permissions.PermissionModule;
 
 /**
@@ -203,7 +202,12 @@ public class SyncUser {
                     handler.post(new Runnable() {
                         @Override
                         public void run() {
-                            callback.onError(error);
+                            try {
+                                callback.onError(error);
+                            } catch (Exception e) {
+                                RealmLog.info("onError has thrown an exception but ignoring it: %s", e.getMessage());
+                                e.printStackTrace();
+                            }
                         }
                     });
                 }
--- a/realm/realm-library/src/objectServer/java/io/realm/SyncUser.java
+++ b/realm/realm-library/src/objectServer/java/io/realm/SyncUser.java
@@ -205,8 +205,8 @@ public class SyncUser {
                             try {
                                 callback.onError(error);
                             } catch (Exception e) {
-                                RealmLog.info("onError has thrown an exception but ignoring it: %s", e.getMessage());
-                                e.printStackTrace();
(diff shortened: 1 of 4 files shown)


realm/realm-library/src/objectServer/java/io/realm/SyncUser.java
```diff
@@ -203,7 +202,12 @@ private void postError(final ObjectServerError error) {
                     handler.post(new Runnable() {
                         @Override
                         public void run() {
-                            callback.onError(error);
+                            try {
+                                callback.onError(error);
+                            } catch (Exception e) {
+                                RealmLog.info("onError has thrown an exception but ignoring it: %s", e.getMessage());
```
cmelchior: "but is ignoring it."

Also use `io.realm.internal.Util.getStackTrace(Throwable)` instead of `e.getMessage()`. That will show the entire stack trace in logcat.

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/AuthTests.java
```diff
@@ -66,4 +66,31 @@ public void onError(ObjectServerError error) {
...
+            public void onError(ObjectServerError error) {
+                assertEquals(ErrorCode.INVALID_CREDENTIALS, error.getErrorCode());
+                throw new IllegalArgumentException("BOOM");
+            }
+        });
+
+        try {
+            Thread.sleep(2000);
+        } catch (InterruptedException e) {
+            e.printStackTrace();
```
cmelchior: Fail the test here

realm/realm-library/src/objectServer/java/io/realm/SyncUser.java
```diff
@@ -203,7 +202,12 @@ private void postError(final ObjectServerError error) {
                     handler.post(new Runnable() {
                         @Override
                         public void run() {
-                            callback.onError(error);
+                            try {
```
beeender: I this correct? `loginAsync` will called mostly from UI thread, and the this `Runnable` is posted to UI thread, it should be totally valid to throw an exception from here.

kneth: As noted in the issue, I actually think that throwing an exception in an async error handler is a really bad idea. For example RxJava does not allow it (or consider it bad practise). It might be difficult to catch an exception from an async task, so doing it might just terminate the app.

cmelchior: I chatted with @kneth about this yesterday as well, e.g. https://jira.mongodb.org/browse/JAVA-2129 do the same. It is less than ideal, but given that it is impossible for developers to reason about where the exception will finally surface, logging it if it happens seems the least bad solution.

beeender: 👍

Issue #4039, opened 2017-01-13
SyncClient Unable to renew Access Token


nhachicha:
REALM_SYNC_VERSION=1.0.0-BETA-7.0

ROS=v1.0.0-BETA-5.1

Using the RealmTasks you can reproduce the following error.

After you create a new user & login, wait a couple of minutes then add a new task (commit)
the SyncSession will stop syncing the changes because ROS send a fatal client error.

- The server first send an `Access token expired (202)` so the client can renew the access token 

- The client receive this error but ignore it because of this bug #4038 

- Even if the Sync Client is not ignoring the error, it tries to bind again because `session.isAuthenticated(session.configuration)` can return `true` if the client consider the token to be still valid (based on timestamps for example) this is not true for the server who consider the token to be expired (hence the first `202`)

- If the client continue binding with the expired token, the server will return a fatal error (`203, Bad user authentication (BIND, REFRESH)`

PR #4040, opened 2017-01-13
Nh/fixes token renew

### LINKED ISSUE #4038: SyncClient ignoring protocol error
Errors can be either client error or protocol error not both
the following [condition is false](https://github.com/realm/realm-java/blob/releases/realm/realm-library/src/main/cpp/objectserver_shared.hpp#L65-L66), which result in ignoring all recoverable errors including new tokens.

nhachicha:
Fixes #4038 and  #4039. preventing Sync client to renew the access token.

### DIFF
--- a/realm/realm-library/src/main/cpp/objectserver_shared.hpp
+++ b/realm/realm-library/src/main/cpp/objectserver_shared.hpp
@@ -61,8 +61,8 @@ public:
                 coordinator->wake_up_notifier_worker();
             }
         };
-        auto error_handler = [weak_session_ref](std::error_code error_code, bool is_fatal, const std::string message) {
-            if (error_code.category() != realm::sync::protocol_error_category() ||
+        auto error_handler = [weak_session_ref, local_realm_path](std::error_code error_code, bool is_fatal, const std::string message) {
+            if (error_code.category() != realm::sync::protocol_error_category() &&
                     error_code.category() != realm::sync::client_error_category()) {
                 // FIXME: Consider below when moving to the OS sync manager.
                 // Ignore this error since it may cause exceptions in java ErrorCode.fromInt(). Throwing exception there
--- a/realm/realm-library/src/main/cpp/objectserver_shared.hpp
+++ b/realm/realm-library/src/main/cpp/objectserver_shared.hpp
@@ -61,7 +61,7 @@ public:
                 coordinator->wake_up_notifier_worker();
             }
         };
-        auto error_handler = [weak_session_ref, local_realm_path](std::error_code error_code, bool is_fatal, const std::string message) {
+        auto error_handler = [weak_session_ref](std::error_code error_code, bool is_fatal, const std::string message) {
             if (error_code.category() != realm::sync::protocol_error_category() &&
                     error_code.category() != realm::sync::client_error_category()) {
                 // FIXME: Consider below when moving to the OS sync manager.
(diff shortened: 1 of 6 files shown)


realm/realm-library/src/main/cpp/objectserver_shared.hpp
```diff
@@ -62,7 +62,7 @@ class JniSession {
             }
         };
         auto error_handler = [weak_session_ref](std::error_code error_code, bool is_fatal, const std::string message) {
-            if (error_code.category() != realm::sync::protocol_error_category() ||
+            if (error_code.category() != realm::sync::protocol_error_category() &&
```
beeender: Seems like I put the trap here 😺

nhachicha:
@realm/java

realm/realm-library/src/objectServer/java/io/realm/internal/objectserver/BoundState.java
```diff
@@ -52,7 +52,22 @@ public void onError(ObjectServerError error) {
...
+            //  the session with this token. To be safe we remove the token that has been considered by the
+            //  the server to be invalid.
+            session.getUser().removeAccessToken(session.getServerUrl());
+
+            // stop the session to avoid sending a bind to the server which will cause it to return
+            // a fatal 203 (bad refresh)
+            session.stopNativeSession();
+
+            // Create a new session & bind it
+            session.createNativeSession();
```
beeender: Wouldn't it be better just go to `UNBOUND` state since in the constructor there will call `stopNativeSession` and `createNativeSession`, also it will transit to `BINDING` state there.
@cmelchior

realm/realm-library/src/objectServer/java/io/realm/SyncUser.java
```diff
@@ -392,6 +392,10 @@ private static String getManagementRealmUrl(URL authUrl) {
         }
     }
 
+    public void removeAccessToken (URI serverURL) {
```
kneth: Should a `public` method have documentation?

cmelchior: Yes, but I am doubtful we want this method to be public?

realm/realm-library/src/objectServer/java/io/realm/internal/objectserver/BoundState.java
```diff
@@ -52,7 +52,22 @@ public void onError(ObjectServerError error) {
...
+            //  still consider this token to be valid (based on timestamps for example)
+            //
+            //  this may cause the server to send a fatal error (203 bad refresh) if we try to bind
+            //  the session with this token. To be safe we remove the token that has been considered by the
+            //  the server to be invalid.
+            session.getUser().removeAccessToken(session.getServerUrl());
+
+            // stop the session to avoid sending a bind to the server which will cause it to return
+            // a fatal 203 (bad refresh)
+            session.stopNativeSession();
```
cmelchior: I would probably stop the native session before removing the access token. I'm not sure, but I think not doing it could introduce a race condition.

realm/realm-library/src/objectServer/java/io/realm/internal/objectserver/BoundState.java
cmelchior: It is true that UNBOUND will stop/crete the session, but it will not automatically transition to BINDING, so I think keeping this logic here makes sense

Issue #3138, opened 2016-07-05
findFirstAsync is not emitting object when queried object is deleted from Realm


viraj49:
#### Goal

Receive emitted object in following code when RealmObject is deleted from Realm.

```
realm.where(GitHubUserProfile.class)
         .equalTo("login", login)
         .findFirstAsync()
         .asObservable()
```
#### Expected Results

Want to receive a callback when Queried RealmObject is deleted from Realm, just like we do when RealmObject gets modified. I am getting a callback when I use findAllAsync() even when object is deleted.
#### Actual Results

Not getting a callback when RealmObject is deleted from Realm.
#### Code Sample

```
/* Realm NOT emitting when I delete the queried object */
private void sampleQuery1(String login) {
    realm.where(GitHubUserProfile.class)
             .equalTo("login", login)
             .findFirstAsync()
             .asObservable()
             .cast(GitHubUserProfile.class)
             .filter(realmObject -> realmObject.isLoaded())
             .filter(realmObject -> realmObject.isValid())
             .subscribeOn(AndroidSchedulers.mainThread())
             .observeOn(AndroidSchedulers.mainThread())
             .subscribe(gitHubUserProfile -> {
                  view.setData(gitHubUserProfile);
              });
}

/* Realm emitting when I delete the queried object */
private void sampleQuery2(String login) {
    realm.where(GitHubUserProfile.class)
             .equalTo("login", login)
             .findAllAsync()
             .asObservable()
             .cast(GitHubUserProfile.class)
             .filter(realmObject -> realmObject.isLoaded())
             .filter(realmObject -> realmObject.isValid())
             .subscribeOn(AndroidSchedulers.mainThread())
             .observeOn(AndroidSchedulers.mainThread())
             .subscribe(gitHubUserProfile -> {
                  view.setData(gitHubUserProfile);
              });
}

public void clearGitHubUserProfileFromRealm(String login) {
    GitHubUserProfile gitHubUserProfile = realm.where(GitHubUserProfile.class)
        .equalTo("login", login)
        .findFirst();

    Realm realm = Realm.getDefaultInstance();
    if (gitHubUserProfile != null) {
        realm.executeTransaction(realm1 -> gitHubUserProfile.deleteFromRealm());
    }
    realm.close();
}
```
#### Version of Realm and tooling

Realm version(s): 1.1.0

Android Studio version: 2.1.2

Which Android version and device: 6.0.1, One plus One

Zhuinden:
[...]

PR #4050, opened 2017-01-16
Remove deprecated Logger class

beeender:


Zhuinden:
Last time stuff like this waited until the next major version, the bug got pushed behind by sync stuff 👅 see https://github.com/realm/realm-java/issues/3138

PR #4051, opened 2017-01-16
Removing unused public constants

### LINKED ISSUE #4047: Default ports not correct
When setting a Realm URL it looks like it defaults to port 80 instead of 9080. This will cause it to throw a connection error if trying to connect to a default configured server.

```
// Will default to port 80, should default to 9080
SyncConfiguration config = new SyncConfiguration.Builder(user, "realm://ros.realm.io/~/default");
```

Also verify that `realms` default to the correct port (9443)

kneth:
Closes #4047

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ### Object Server API Changes (In Beta)
 
 * Exceptions thrown in error handlers are ignored but logged (#3559).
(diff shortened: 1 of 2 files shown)


kneth:
Thank you for your kind words :-)

kneth:
@beeender's comments in the issue saved me a lot of time.

PR #4052, opened 2017-01-16
Adding error code disabled_session

### LINKED ISSUE #4046: Ensure that Java ErrorCode's match Syncs
The Sync Error codes has shuffled around a bit in the last few releases, and we need to ensure that the Java equivalents are aligned.

kneth:
On par with protocol 16. The error code is handled by object store.

Closes #4046


kneth:
@beeender Because the sync team forgot to add it. There is currently a PR for updating the documentation.

PR #4053, opened 2017-01-17
Nh/init metadata

nhachicha:
This fix a native crash, when scheduling a client reset, because the ObjectStore metadata is not configure. (note this issue doesn't exist in master, because we init the metadata, when using the default UserStore - RealmUserStore)

### DIFF
--- a/realm/realm-library/src/main/cpp/CMakeLists.txt
+++ b/realm/realm-library/src/main/cpp/CMakeLists.txt
@@ -170,8 +170,10 @@ file(GLOB objectstore_SRC
 # Sync needed Object Store files
 if (build_SYNC)
     file(GLOB objectstore_sync_SRC
-        "object-store/src/sync/*"
-        "object-store/src/sync/impl/*")
+        "object-store/src/results.cpp"
+        "object-store/src/impl/results_notifier.cpp"
+        "object-store/src/sync/*.cpp"
+        "object-store/src/sync/impl/*.cpp")
 endif()
 
 add_library(realm-jni SHARED ${jni_SRC} ${objectstore_SRC} ${objectstore_sync_SRC})
--- a/realm/realm-library/src/main/cpp/io_realm_SyncManager.cpp
+++ b/realm/realm-library/src/main/cpp/io_realm_SyncManager.cpp
@@ -31,6 +31,9 @@
 
 #include "jni_util/log.hpp"
 #include "jni_util/jni_utils.hpp"
+#include "sync/sync_manager.hpp"
+#include "sync/sync_user.hpp"
+#include "util.hpp"
 
 using namespace realm;
 using namespace realm::sync;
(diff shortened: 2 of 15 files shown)


realm/realm-library/src/main/cpp/io_realm_SyncManager.cpp
```diff
@@ -59,3 +62,13 @@ Java_io_realm_SyncManager_nativeRunClient(JNIEnv *env, jclass)
...
     } CATCH_STD()
 }
+
+JNIEXPORT void JNICALL
+Java_io_realm_SyncManager_nativeConfigureMetaDataSystem(JNIEnv *env, jclass,
+                                                        jstring baseFile) {
+    TR_ENTER()
+    try {
+        JStringAccessor base_file_path(env, baseFile); // throws
+        SyncManager::shared().configure_file_system(base_file_path, SyncManager::MetadataMode::NoEncryption);
```
kneth: So we never encrypt metadata?

realm/realm-library/src/main/cpp/CMakeLists.txt
```diff
@@ -170,8 +170,10 @@ file(GLOB objectstore_SRC
 # Sync needed Object Store files
 if (build_SYNC)
     file(GLOB objectstore_sync_SRC
-        "object-store/src/sync/*"
-        "object-store/src/sync/impl/*")
+        "object-store/src/results.cpp"
+        "object-store/src/impl/results_notifier.cpp"
```
beeender: `results.cpp` and `results_notifier.cpp` don't seem to be necessary for this PR?

realm/realm-library/src/main/cpp/objectserver_shared.hpp
```diff
@@ -101,8 +101,8 @@ class JniSession {
                     realm::SyncFileActionMetadata(manager,
                                                   realm::SyncFileActionMetadata::Action::HandleRealmForClientReset,
                                                   original_path,
-                                                  nullptr,
-                                                  nullptr,
+                                                  "",
```
beeender: I don't quite understand how could this be useful if `url` and `user_identity` are empty string...

kneth: `original_path` is an absolute path so you don't have to know anything more.

beeender: Doesn't it need to know the sync Realm's url?

kneth: It will basically make a copy of the local file - and remove the local file.

realm/realm-library/src/main/cpp/io_realm_SyncManager.cpp
nhachicha: no, this is what motivated the creation of https://github.com/realm/realm-android-user-store 

on iOS they can obtain a key directly from the keychain, which is not the case for Android, since we need the KeyStore to be unlocked & does involve user interaction & lifecycle

realm/realm-library/src/main/cpp/CMakeLists.txt
nhachicha: we're using `SyncManager` which pulls the `metadata` dependency from the OS which need the definition of these 2 files to compile.

PR #4055, opened 2017-01-17
Try to reproduce nh crash

beeender:


nhachicha:
should be fixed in https://github.com/realm/realm-object-store/pull/343

PR #4058, opened 2017-01-17
Expose schemaVersion in SyncConfig

### LINKED ISSUE #3982: SyncConfiguration should expose schemaVersion
Cocoa exposes `SyncConfiguration.schemaVersion()`. We should do the same. 
The argument is that we want to catch the binding making unintentional changes.

However due to the unique working of sync there is a few caveats and corner cases that needs to be considered. At this point it is unclear how much of this is handled by the Object Store

1) `schemaVersion()` should be considered as `X or bigger`, so calling `realm.getVersion()` might return a number larger than X. It should never return a lower version.

Consider e.g. the following problematic case: 

1) Device A installs schemaVersion:1
2) Device B installs schemaVersion:3
3) Device A installs schemaVersion:2.

If 2/3 both are offline and A comes online later it will write 2 in the metarealm which due to syncs will cause B to suddenly have a lower version number. While not breaking, it will probably be weird for users to experience this. 

Basically we need a `MAX(verA, verB)` as conflict resolution for the metadata schema version number, which I'm not entirely sure how we can do without major refactoring.

bmeike:
Fixes #3982


bmeike:
@kneth Understood.  Not sure how that affects this PR.

As a slight side note, I posted this on the Java-team slack channel, a few days ago:

Ok, here’s what I think.  I am still missing something.  We are *depending* on a safe, increasing stream of version numbers.  I understand, now, that it cannot be supplied by a server.  This is not git, though.  We cannot tolerate schemas that are not strict supersets.  For instance if we have schema version 1, and two different clients create versions 2 and 2’, that are *not* mergeable (say, a property with the same name, but different type), something breaks (sync fails?).  We are depending on the owner of the schema, a human or an enterprise, to dole out version numbers in a reasonable way.  Under those circumstances, I can’t see how “last write wins” makes sense.  It seems to me there should be two versions for every client/server: schemaversion = max(observedVersions…) and viewVersion: subset (<= schemaversion) of schemaversion that this program uses.  … and it seems like this should be a cross-platform invariant, and so part of the core, not each binding. So, what am I missing?

bmeike:
Btw, the test that is failing is supposed to test the condition that Christian describes here: https://github.com/realm/realm-java/issues/3982#issue-198554711

Device A installs schemaVersion:1
Device B installs schemaVersion:3
Device A installs schemaVersion:2.

It tries to do this by replacing the realm db file while the db is closed.

cmelchior:
@bmeike I found some bugs in how we handle schema versions for SyncConfigurations, I'm adding some unit tests + modifications to your branch.

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -284,13 +284,16 @@ static Realm createAndValidate(RealmConfiguration configuration, ColumnIndices[]
...
-            throw new RealmMigrationNeededException(configuration.getPath(), String.format("Realm on disk need to migrate from v%s to v%s", currentVersion, requiredVersion));
-        }
-        if (currentVersion != UNVERSIONED && requiredVersion < currentVersion && columnIndices == null) {
-            realm.doClose();
-            throw new IllegalArgumentException(String.format("Realm on disk is newer than the one specified: v%s vs. v%s", currentVersion, requiredVersion));
+
+        if (!configuration.isSyncConfiguration()) {
+            if (currentVersion != UNVERSIONED && currentVersion < requiredVersion && columnIndices == null) {
+                realm.doClose();
+                throw new RealmMigrationNeededException(configuration.getPath(), String.format("Realm on disk need to migrate from v%s to v%s", currentVersion, requiredVersion));
```
kneth: `%s` -> `%d`

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -346,6 +349,13 @@ private static void initializeRealm(Realm realm) {
                 RealmSchema schema = new RealmSchema(realmObjectSchemas);
                 // Assumption: when SyncConfiguration then additive schema update mode
                 realm.sharedRealm.updateSchema(schema, version);
+                // FIXME - Begin: OS currently grabs its own write transaction which prevents us from doing it in Java
```
kneth: Object store's `update_schema()` has a parameter to say that you are already in a transaction.

realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -591,7 +591,8 @@ static boolean compactRealm(final RealmConfiguration configuration) {
     /**
      * Migrates the Realm file defined by the given configuration using the provided migration block.
      *
-     * @param configuration configuration for the Realm that should be migrated.
+     * @param configuration configuration for the Realm that should be migrated. If this is a SyncConfiguration this
+     *                      method does nothing.
```
nhachicha: I think it's better to throw an `UnsupportedOperationException` to prevent the user from providing a migration block for a SyncConfiguration

cmelchior: `migration()` is not part of the builder, so it should never happen.

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
```diff
@@ -436,9 +437,24 @@ Java_io_realm_internal_SharedRealm_nativeUpdateSchema(JNIEnv *env, jclass, jlong
...
 }
 
+JNIEXPORT jboolean JNICALL
+Java_io_realm_internal_SharedRealm_nativeRequiresMigration(JNIEnv *env, jclass, jlong nativePtr,
+                                                           jlong nativeSchemaPtr) {
+
+    TR_ENTER()
+    try {
+        auto shared_realm = *(reinterpret_cast<SharedRealm*>(nativePtr));
+        auto *schema = reinterpret_cast<Schema*>(nativeSchemaPtr);
```
beeender: `auto *` -> `auto* `. See https://github.com/realm/realm-core/blob/master/doc/development/coding_style_guide.cpp

realm/realm-library/src/main/java/io/realm/internal/SharedRealm.java
```diff
@@ -326,10 +327,18 @@ public boolean compact() {
...
 
+    /**
+     * Update the underlying schema based on the schema description.
+     * Calling this method must be done from inside a write transaction.
+     */
     public void updateSchema(RealmSchema schema, long version) {
         nativeUpdateSchema(nativePtr, schema.getNativePtr(), version);
     }
 
+    public boolean requiresMigration(RealmSchema schema) {
```
beeender: So this is not used by now? needed in the future?

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
```diff
@@ -436,9 +437,24 @@ Java_io_realm_internal_SharedRealm_nativeUpdateSchema(JNIEnv *env, jclass, jlong
...
+        const std::vector<SchemaChange> &change_list = shared_realm->schema().compare(*schema);
```
beeender: ditto

realm/realm-library/src/main/java/io/realm/internal/SharedRealm.java
cmelchior: It should be used in `Realm.java`

cmelchior: For some reason, GitHub is folding that file so it is almost hidden

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -281,103 +281,141 @@ static Realm createInstance(RealmConfiguration configuration, ColumnIndices[] gl
...
+        final boolean unversioned = currentVersion == UNVERSIONED;
 
         try {
-            if (!syncAvailable) {
-                realm.beginTransaction();
-                if (version == UNVERSIONED) {
-                    commitNeeded = true;
-                    realm.setVersion(realm.configuration.getSchemaVersion());
-                }
+            realm.beginTransaction();
```
beeender: now i remembered why the `commitNeeded` was introduced in the first place. If two thread both start init the Realm, one will win with the transaction lock and get here. the other one will just wait until the transaction is committed. BUT, it still have `unversioned == true`. Which means when it gets the transaction lock, it will happily change everything again in the transaction. So it is VERY IMPORTANT to call `getVersion()` again after the transaction begins! that is the whole point we cannot get rid of the `cancelTransaction` in the functions.

beeender: And they will the similar problem in the `initializeSyncedRealm`

cmelchior: All of this is behind a synchronized method though? `createRealmOrGetFromCache` ?

cmelchior: Hmm, i guess multi process access could still be a problem. I'll look into it

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -281,103 +281,146 @@ static Realm createInstance(RealmConfiguration configuration, ColumnIndices[] gl
...
+                columnInfoMap.put(modelClass, mediator.validateTable(modelClass, realm.sharedRealm, false));
+            }
+
+            realm.schema.columnIndices = new ColumnIndices(
+                    (unversioned) ? realm.configuration.getSchemaVersion() : currentVersion, columnInfoMap);
+
+            if (unversioned) {
+                final Transaction transaction = realm.configuration.getInitialDataTransaction();
+                if (transaction != null) {
+                    transaction.execute(realm);
```
beeender: This is called in a try catch block in `initializeSyncedRealm` but not this one. But I suggest to use a catch just at line 360 to prevent unexpected exception thrown eg.: `mediator.createTable`.

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -281,103 +281,146 @@ static Realm createInstance(RealmConfiguration configuration, ColumnIndices[] gl
...
-                        realm.executeTransaction(new Transaction() {
-                            @Override
-                            public void execute(Realm realm) {
-                                realm.setVersion(realm.configuration.getSchemaVersion());
-                            }
-                        });
-                    } else {
+                    try {
                         transaction.execute(realm);
+                    } catch (Exception e) {
```
beeender: move the catch block to 421 to catch other exceptions as well.

realm/realm-library/src/main/java/io/realm/Realm.java
cmelchior: Good idea

realm/realm-library/src/main/java/io/realm/Realm.java
cmelchior: Good idea

PR #4066, opened 2017-01-18
Get more integration test to work

beeender:
- Add a new TestRule RunWithRemoteService to allow a test case create a
  service running in a remote process. This is very much like what we have
  for the RealmInterProcessTests but with more flexibility, eg.:
  Separate service for separate test case.
- Enabled sync interaction tests: notify by simple commit, notify by
  lots of commits.
- Deprecate RemoteProcessService.


beeender:
This works great with running a signal test, but fails when running two or more ...
We need to figure out a way to peacefully restart sync service and client.

realm/realm-library/src/androidTest/java/io/realm/rule/RunWithRemoteService.java
```diff
@@ -0,0 +1,194 @@
...
+import static junit.framework.Assert.assertTrue;
+import static junit.framework.Assert.fail;
+
+/**
+ * This is a helper {@link TestRule} to do test which needs interaction with a remote process.
+ * To use this:
+ * 1. Define a subclass of {@link RemoteTestService} and create steps as static member of it. Those steps should be
+ * named as "stepA_doXXX", "stepB_doYYY", etc. to indicate the order of them.
+ * 2. Add the service into the AndroidManifest.xml. And the android:process property must be ":remote".
+ * 3. Annotated your test case by {@link RunTestWithRemoteService} with your remote service class.
```
cmelchior: Annotated -> Annotate

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/utils/HttpUtils.java
```diff
@@ -38,6 +38,16 @@
     // will forward this query to the host, running the integration test server on 8888
     private final static String START_SERVER = "http://127.0.0.1:8888/start";
     private final static String STOP_SERVER = "http://127.0.0.1:8888/stop";
+    private static boolean inited = false;
+
+    // Start the ROS service only once since there are problems with reseting it. Also it makes testing faster.
```
cmelchior: Can you elaborate on the problems resetting it? 
This pattern here seems a bit dangerous as it will leave the server running after the last integration test is done. We can perhaps live with that for now though.

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/utils/HttpUtils.java
```diff
@@ -38,6 +38,16 @@
...
+    public static void init() throws Exception {
```
cmelchior: This should be synchronized

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/utils/UserFactory.java
```diff
@@ -16,14 +16,38 @@
...
+import io.realm.log.RealmLog;
 
 // Must be in `io.realm.objectserver` to work around package protected methods.
 public class UserFactory {
+    private static final String PASSWORD = "myPassw0rd";
+    // Since the integration tests need use the same user for different processes, we create a new user name when the
+    // test starts and store it in a Realm. Then it can be retrieved for every process.
+    private String userName;
+    private static UserFactory instance;
+    private static RealmConfiguration configuration = new RealmConfiguration.Builder()
```
cmelchior: This require `Realm.init()` to be called before using this class.  Perhaps we should add a comment to the header about that?

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/utils/UserFactory.java
```diff
@@ -32,4 +56,43 @@ public static SyncUser createAdminUser(String authUrl) {
         SyncCredentials credentials = SyncCredentials.custom("admin", "debug", null);
         return SyncUser.login(credentials, authUrl);
     }
+
+    // Since we don't have a reliable way to reset the sync server and client, just use a new user for every test case.
+    public static void resetInstance() {
+        instance = null;
+        SecureRandom random = new SecureRandom();
```
cmelchior: Just use `UUID.random().toString()` as user name, it should be safe enough.

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/utils/UserFactory.java
```diff
@@ -32,4 +56,43 @@ public static SyncUser createAdminUser(String authUrl) {
         SyncCredentials credentials = SyncCredentials.custom("admin", "debug", null);
         return SyncUser.login(credentials, authUrl);
     }
+
+    // Since we don't have a reliable way to reset the sync server and client, just use a new user for every test case.
+    public static void resetInstance() {
```
cmelchior: There seems to be a mismatch between the name and description. Perhaps `generateNewUser()` instead?

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/utils/UserFactory.java
```diff
@@ -32,4 +56,43 @@ public static SyncUser createAdminUser(String authUrl) {
...
+    // The @Before method will be called before the looper tests finished. We need to find a better place to call this.
+    public static void clearInstance()  {
+        Realm realm = Realm.getInstance(configuration);
+        realm.beginTransaction();
+        realm.delete(UserFactoryStore.class);
+        realm.commitTransaction();
+        realm.close();
+    }
+
+    public static UserFactory getInstance() {
```
cmelchior: Needs to be synchronized

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/utils/HttpUtils.java
beeender: It might be my fault ... After clean up everything, the original solution (start/stop in BeforeClass/AfterClass) works. Will revert this change.

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/utils/HttpUtils.java
beeender: will remove this.

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/utils/UserFactory.java
beeender: This is to match the method name `getInstance()`. Fixed the comment.

beeender:
Finally got this to work. @realm/java review please.

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/ProcessCommitTests.java
```diff
@@ -16,159 +16,202 @@
...
-                    Looper.loop();
-
-                } catch (Throwable e) {
-                    exception[0] = e;
-                    testFinished.countDown();
-                }
+            public void onChange(RealmResults<ProcessInfo> element) {
+                assertEquals(1, all.size());
+                assertEquals("Background_Process1", all.get(0).getName());
+                realm.close();
```
cmelchior: Can these clean up methods somehow be moved to the Rule? These tests are so complex that anything we can do to reduce it would be nice.

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/ProcessCommitTests.java
```diff
@@ -16,159 +16,202 @@
...
+            @Override
+            protected void run() {
+                user = UserFactory.getInstance().loginWithDefaultUser(Constants.AUTH_URL);
+                String realmUrl = Constants.SYNC_SERVER_URL;
+
+                final SyncConfiguration syncConfig = new SyncConfiguration.Builder(user, realmUrl)
+                        .directory(getService().getRoot())
+                        .name(UUID.randomUUID().toString() + ".realm")
+                        .build();
+                getService().realm = Realm.getInstance(syncConfig);
```
cmelchior: I think @bmeike would prefer if this was `setRealm()` and getRealm()` instead :)

realm/realm-library/src/androidTest/java/io/realm/rule/RunWithRemoteService.java
```diff
@@ -0,0 +1,180 @@
...
+            super(looper);
+            localMessenger = new Messenger(this);
+        }
+
+        @Override
+        public void handleMessage(Message msg) {
+            Bundle bundle = msg.getData();
+            String error = bundle.getString(RemoteTestService.BUNDLE_KEY_ERROR);
+            if (error != null) {
+                // Assert and show error from value process
```
beeender: I don't understand either, probably i wanted to say remote process.

realm/realm-library/src/androidTest/java/io/realm/services/RemoteProcessService.java
```diff
@@ -34,6 +34,7 @@
 
 /**
  * Helper service for multi-processes support testing.
+ * @deprecated use {@link RemoteTestService} instead.
```
beeender: It is still used in `RealmInterprocessTest` which I don't have time to refactor yet.

realm/realm-library/src/androidTest/java/io/realm/services/RemoteTestService.java
```diff
@@ -0,0 +1,176 @@
...
+import java.io.PrintWriter;
+import java.io.StringWriter;
+import java.util.HashMap;
+import java.util.Map;
+
+import io.realm.Realm;
+import io.realm.log.RealmLog;
+
+/**
+ * Helper class for multi-processes support testing.
```
beeender: Description in `RunWithRemoteService`. Added preferred naming rules there, and added a link here.

realm/realm-library/src/androidTest/java/io/realm/services/RemoteTestService.java
```diff
@@ -0,0 +1,176 @@
...
+
+        protected Step(int base, int id) {
+            this.message = base + id;
+            stepMap.put(this.message, this);
+        }
+
+        protected abstract void run();
+
+        protected RemoteTestService getService() {
+            return RemoteTestService.thiz;
```
beeender: This is a method of `Step` which is a static inner class.

realm/realm-library/src/androidTest/java/io/realm/services/RemoteTestService.java
```diff
@@ -0,0 +1,176 @@
...
+            } catch (RemoteException e) {
+                RealmLog.error(e);
+            }
+        }
+    }
+
+    public static final String BUNDLE_KEY_ERROR = "error";
+    @SuppressLint("UseSparseArrays")
+    private static Map<Integer, Step> stepMap = new HashMap<Integer, Step>();
+    public static RemoteTestService thiz;
```
beeender: All the `Step` are defined as static method in the child classes for a reason which I cannot remember... looks very much because of my comment here:
>     // There is no easy way to dynamically ensure step IDs have same value for different processes. So, use the stupid
>    // way.

And the step instance still need to access the `RemoteTestService` instance sometimes.

realm/realm-library/src/androidTest/java/io/realm/services/RemoteTestService.java
```diff
@@ -0,0 +1,176 @@
...
+
+    @Override
+    public IBinder onBind(Intent intent) {
+        return messenger.getBinder();
+    }
+
+    @Override
+    public boolean onUnbind(Intent intent) {
+        stopSelf();
+        recursiveDelete(rootFolder);
```
beeender: I think it doesn't, doc from `createTempFile()`:
> This method provides only part of a temporary-file facility. To arrange for a file created by this method to be deleted automatically, use the deleteOnExit() method.
> 
> The prefix argument must be at least three characters long. It is recommended that the prefix be a short, meaningful string such as "hjb" or "mail". The suffix argument may be null, in which case the suffix ".tmp" will be used. 

Also, the `TemporaryFolder` is doing the same.

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/ProcessCommitTests.java
```diff
@@ -16,159 +16,202 @@
...
+                Realm realm = getService().realm;
+
+                realm.beginTransaction();
+                ProcessInfo processInfo = realm.createObject(ProcessInfo.class);
+                processInfo.setName("Background_Process1");
+                processInfo.setPid(android.os.Process.myPid());
+                processInfo.setThreadId(Thread.currentThread().getId());
+                realm.commitTransaction();
+                // FIXME: If we close the Realm here, the data won't be able to synced to the main process. Is it a bug
```
beeender: Can we get a `SyncSession` instance without any Realm instance opened?

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/ProcessCommitTests.java
beeender: I think it could be achieved by inherit the `RunInLooperThread` which would do the user logout when clean up. I can do it in a following PR.

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/utils/UserFactory.java
```diff
@@ -16,18 +16,44 @@
...
 
+import java.util.UUID;
+
+import io.realm.Realm;
+import io.realm.RealmConfiguration;
 import io.realm.SyncCredentials;
 import io.realm.SyncUser;
+import io.realm.log.RealmLog;
 
+// Helper class to retrieve users with same IDs even in multi-processes.
```
beeender: Should this store the user? `SyncUser.currentUser()` can get the current login user I think?

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/ProcessCommitTests.java
cmelchior: Not currently, no. This is also something we need to get fixed, because other features depend on this being possible

realm/realm-library/src/androidTest/java/io/realm/rule/RunWithRemoteService.java
```diff
@@ -170,7 +174,7 @@ public void triggerServiceStep(RemoteTestService.Step step) {
         ActivityManager manager = (ActivityManager)getContext().getSystemService(Context.ACTIVITY_SERVICE);
         List<ActivityManager.RunningAppProcessInfo> processInfoList = manager.getRunningAppProcesses();
         for (ActivityManager.RunningAppProcessInfo info : processInfoList) {
-            if (info.processName.equals(getContext().getPackageName() + ":remote")) {
+            if (info.processName.equals(getContext().getPackageName() + REMOTE_PROCESS_POSTFIX)) {
```
cmelchior: I think the word you are looking for is SUFFIX

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/ProcessCommitTests.java
```diff
@@ -16,159 +16,207 @@
...
+            @Override
+            protected void run() {
+                getService().getRealm().close();
+                user.logout();
+            }
+        };
+    }
+
+    // 1. Open a sync Realm and listen to changes.
+    // A. Open the same sync Realm and add on object.
```
nhachicha: add one object

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/ProcessCommitTests.java
```diff
@@ -16,159 +16,207 @@
...
-                    testFinished.countDown();
+            public void onChange(RealmResults<TestObject> element) {
+                int counter = listenerCalledCounter.incrementAndGet();
+                int size = all.size();
+                if (size == 0) {
+                    listenerCalledCounter.decrementAndGet();
+                    return;
+                }
+                assertEquals(0, size % 100); // Added 100 objects every time.
+                assertEquals(counter * 100 -1, all.last().getIntProp());
```
nhachicha: space after `-`

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/ProcessCommitTests.java
```diff
@@ -16,159 +16,207 @@
...
+                assertEquals("Str" + (counter * 100 -1), all.last().getStringProp());
```
nhachicha: space after `-`

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/utils/UserFactory.java
```diff
@@ -16,18 +16,44 @@
...
 // Must be in `io.realm.objectserver` to work around package protected methods.
+// This require Realm.init() to be called before using this class.
 public class UserFactory {
+    private static final String PASSWORD = "myPassw0rd";
+    // Since the integration tests need use the same user for different processes, we create a new user name when the
```
nhachicha: need to use

PR #4078, opened 2017-01-19
fix transaction error of synced Realm.

zaki50:


realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -422,6 +422,8 @@ private static void initializeSyncedRealm(Realm realm) {
         } finally {
             if (commitChanges) {
                 realm.commitTransaction(false);
+            } else {
```
cmelchior: Also in `initializeRealm`?

cmelchior: Nvm, it does it already

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -422,6 +422,8 @@ private static void initializeSyncedRealm(Realm realm) {
...
+                realm.cancelTransaction();
```
nhachicha: I think it's safe to check first if we're in a transaction `realm.isInTransaction()`

beeender: it should never happens ... if i happens, then it is a serious bug ...

cmelchior: The transaction is started when you enter the method?

PR #4096, opened 2017-01-24
Free the SharedRealm in phantom daemon

### LINKED ISSUE #3730: NullPointerException: Null reference used for synchronization. SharedRealm.finalize()
#### Actual Results

> E/System: Uncaught exception thrown by finalizer
> E/System: java.lang.NullPointerException: Null reference used for synchronization (monitor-enter)
> 	at io.realm.internal.SharedRealm.finalize(SharedRealm.java:318)
> 	at java.lang.Daemons$FinalizerDaemon.doFinalize(Daemons.java:217)
> 	at java.lang.Daemons$FinalizerDaemon.run(Daemons.java:200)
> 	at java.lang.Thread.run(Thread.java:818)

>Was reproduced by client. Appeared multiple times in logcat.
>"Once in a while a popup error message is shown telling that the app has crashed."

Realm version(s): 2.0.2

Which Android version and device: Samsung Galaxy A5 (Android 6.0.1)

beeender:
Treat the SharedRealm the same as other native objects.
SharedRealm.close will only call Object Store Realm::close without
deleting the ShareRealm pointer.
Then we don't need the finalizer anymore. Fix #3730 .
This is related with
https://github.com/realm/realm-object-store/pull/318
as well. It is possible that java close the Realm in any of the Object Store's 
callbacks. To avoid Object Store operating on a invalid SharedRealm
pointer, binding should try to make sure after callbacks. However, it
cannot be totally avoided since user could set the Realm instance to
null and the instance can be GCed at any time. It is still something
should be considered in the Object Store implementation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 2.3.1
+
+### Bug fixes
+
+* Fixed NPE problem happened in SharedRealm.finalize() (#3730).
+
 ## 2.3.0
 
 ### Object Server API Changes 
--- a/realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
+++ b/realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
@@ -29,6 +29,8 @@ static_assert(SchemaMode::Additive ==
 static_assert(SchemaMode::Manual ==
               static_cast<SchemaMode>(io_realm_internal_SharedRealm_SCHEMA_MODE_VALUE_MANUAL), "");
 
+static void finalize_shared_realm(jlong ptr);
+
 JNIEXPORT void JNICALL
 Java_io_realm_internal_SharedRealm_nativeInit(JNIEnv *env, jclass, jstring temporary_directory_path)
 {
@@ -98,8 +100,9 @@ Java_io_realm_internal_SharedRealm_nativeCloseSharedRealm(JNIEnv*, jclass, jlong
 {
     TR_ENTER_PTR(shared_realm_ptr)
 
-    auto ptr = reinterpret_cast<SharedRealm*>(shared_realm_ptr);
-    delete ptr;
+    auto shared_realm = *(reinterpret_cast<SharedRealm*>(shared_realm_ptr));
+    // Close the SharedRealm only. Let the finalizer daemon thread free the SharedRealm
+    shared_realm->close();
 }
 
 JNIEXPORT void JNICALL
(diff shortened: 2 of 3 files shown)


Zhuinden:
Does this mean your Realm instances can get released even if you don't explicitly call `close()` on them?

beeender:
@Zhuinden Actually it was released even before in the `finalize()`. But still, explicit call to `close()` is necessary since `SharedRealm` will bypass the `RealmCache`.

realm/realm-library/src/main/java/io/realm/internal/SharedRealm.java
```diff
@@ -215,10 +217,6 @@ public static SharedRealm getInstance(RealmConfiguration config, RealmNotifier r
         }
     }
 
-    long getNativePtr() {
```
kneth: Just moving it.

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
```diff
@@ -29,6 +29,8 @@ static_assert(SchemaMode::Additive ==
 static_assert(SchemaMode::Manual ==
               static_cast<SchemaMode>(io_realm_internal_SharedRealm_SCHEMA_MODE_VALUE_MANUAL), "");
 
+static void finalize_shared_realm(jlong ptr);
```
kneth: Do we need this prototype?

beeender: Yes, for the `-Wmissing-declarations`.

realm/realm-library/src/main/java/io/realm/internal/SharedRealm.java
beeender: it looks better to be together with `getNativeFinalizerPtr()` and close to `close()`. :)

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
kneth: Really? OK

CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## 2.3.1
+
+### Bug fixes
+
+* Fixed NPE problem happened in SharedRealm.finalize() (#3730).
```
Zhuinden: I'm not sure if that `happened` is really needed.

beeender: doh .... too late ... it is all your fault, you comment 10 seconds earlier !! I will fix it next time when modify the changelog.

Zhuinden: I'm sorry 😭

PR #4111, opened 2017-01-26
Fix RealmCollection.contains not respecting custom equal methods

cmelchior:
Got reported directly, and fixing it was easier than creating the issue :D

EDIT: `static.realm.io` is still insanely slow so CI times out on the unit tests. Will move into review once that is fixed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ### Bug fixes
 
 * NPE problem in SharedRealm.finalize() (#3730).
-* `RealmList.contains()` and `RealmResults.contains()` now correctly uses custom `equals()` methods on Realm model classes.
+* `RealmList.contains()` and `RealmResults.contains()` now correctly use custom `equals()` method on Realm model classes.
 * Build error when the project is using Kotlin (#4087).
 * Bug causing classes to be replaced by classes already in Gradle's classpath (#3568).
 * NullPointerException when notifying a single object that it changed (#4086).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### Bug fixes
 
-* NPE problem happened in SharedRealm.finalize() (#3730).
+* NPE problem in SharedRealm.finalize() (#3730).
 * `RealmList.contains()` and `RealmResults.contains()` now correctly uses custom `equals()` methods on Realm model classes.
 * Build error when the project is using Kotlin (#4087).
 * Bug causing classes to be replaced by classes already in Gradle's classpath (#3568).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### Bug fixes
 
 * Fixed NPE problem happened in SharedRealm.finalize() (#3730).
+* `RealmList.contains()` and `RealmResults.contains()` now correctly uses custom `equals()` methods on Realm model classes.
 
 ## 2.3.0
 
--- a/realm/realm-library/src/androidTest/java/io/realm/RealmCollectionTests.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/RealmCollectionTests.java
@@ -30,6 +30,7 @@ import java.util.Iterator;
 import java.util.List;
 
 import io.realm.entities.AllJavaTypes;
+import io.realm.entities.CustomMethods;
 import io.realm.entities.Dog;
 import io.realm.entities.NullTypes;
 import io.realm.rule.TestRealmConfigurationFactory;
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -2,8 +2,10 @@
 
 ### Bug fixes
 
-* Fixed NPE problem happened in SharedRealm.finalize() (#3730).
+* NPE problem happened in SharedRealm.finalize() (#3730).
```
Zhuinden: are you sure you need `happened`?

CHANGELOG.md
```diff
@@ -2,9 +2,10 @@
 
 ### Bug fixes
 
-* Fixed NPE problem happened in SharedRealm.finalize() (#3730).
-* Fixed a build error when the project is using Kotlin (#4087).
-* Fixed a bug causing classes to be replaced by classes already in Gradle's classpath (#3568).
+* NPE problem in SharedRealm.finalize() (#3730).
+* `RealmList.contains()` and `RealmResults.contains()` now correctly uses custom `equals()` methods on Realm model classes.
```
kneth: "uses" -> "use"
"methods" -> "method"

realm/realm-library/src/androidTest/java/io/realm/RealmCollectionTests.java
```diff
@@ -181,6 +214,18 @@ public void contains_null() {
...
 
+    // Test that the custom equal methods is being used when testing if an object is part of the
+    // collection
+    @Test
+    public void contains_customEqualMethod() {
+        RealmCollection<CustomMethods> collection = createCustomMethodsCollection(realm, collectionClass);
+        // This custom equals method will only consider the field `name` when comparing objects.
+        // So this unmanaged version should be equal to any object with the same value, managed
+        // or not.
+        CustomMethods unmanaged = new CustomMethods("Child0");
```
kneth: Only test first element? Maybe testing `Child + TEST_SIZE - 1` too (checking the iterators)?

zaki50: And how about testing if `Child + TEST_SIZE` is not contained?

cmelchior:
Updated @kneth

realm/realm-library/src/androidTest/java/io/realm/entities/CustomMethods.java
```diff
@@ -33,20 +42,20 @@ public void setName(String name) {
...
 
-    @Ignore
-    public boolean reverseEquals;
+    public RealmList<CustomMethods> getMethods() {
+        return methods;
+    }
 
     @Override
     public boolean equals(Object o) {
         if (!(o instanceof CustomMethods)) {
```
nhachicha: - you're not checking for equal identity 
- passed object could be null

realm/realm-library/src/androidTest/java/io/realm/entities/CustomMethods.java
```diff
@@ -33,20 +42,20 @@ public void setName(String name) {
...
         CustomMethods other = (CustomMethods) o;
-        if (isManaged() == other.isManaged() && other.name.equals(name)) {
-            return !reverseEquals;
-        } else {
-            return reverseEquals;
-        }
+
+        // Only compare name. Managed and unmanaged objects will be equal as long as they have the
+        // same value
+        return other.name.equals(name);
```
nhachicha: - `other.name` is not required could be null

nhachicha: I suggest the following semi-generated impl
```java
@Override
public boolean equals(Object o) {
    if (this == o) return true;
    if (o == null || !(o instanceof CustomMethods)) return false;

    CustomMethods that = (CustomMethods) o;

// Only compare name. Managed and unmanaged objects will be equal as long as they have the same value
    return name != null ? name.equals(that.name) : that.name == null;
}
```

realm/realm-library/src/main/java/io/realm/RealmList.java
```diff
@@ -720,19 +720,26 @@ public boolean load() {
...
-        boolean contains = false;
         if (managedMode) {
             realm.checkIfValid();
+
+            // Deleted objects can never be part of a RealmList
             if (object instanceof RealmObjectProxy) {
                 RealmObjectProxy proxy = (RealmObjectProxy) object;
-                if (proxy.realmGet$proxyState().getRow$realm() != null && realm.getPath().equals(proxy.realmGet$proxyState().getRealm$realm().getPath()) && proxy.realmGet$proxyState().getRow$realm() != InvalidRow.INSTANCE) {
-                    contains = view.contains(proxy.realmGet$proxyState().getRow$realm().getIndex());
+                if (proxy.realmGet$proxyState().getRow$realm() == InvalidRow.INSTANCE) {
```
nhachicha: why removing the comparison  of the path ? we need to assure we're comparing managed object from the same Realm?

realm/realm-library/src/main/java/io/realm/RealmResults.java
```diff
@@ -189,14 +189,22 @@ public boolean isManaged() {
      */
     @Override
     public boolean contains(Object object) {
-        boolean contains = false;
-        if (isLoaded() && object instanceof RealmObjectProxy) {
-            RealmObjectProxy proxy = (RealmObjectProxy) object;
-            if (realm.getPath().equals(proxy.realmGet$proxyState().getRealm$realm().getPath()) && proxy.realmGet$proxyState().getRow$realm() != InvalidRow.INSTANCE) {
-                contains = (table.sourceRowIndex(proxy.realmGet$proxyState().getRow$realm().getIndex()) != TableOrView.NO_MATCH);
```
nhachicha: we're not using `Java_io_realm_internal_TableView_nativeFindBySourceNdx` for table views? and instead doing a linear comparison with all `RealmResults` elements? what's the performance impact ?

cmelchior: Probably quite big, but we are kinda forced to do that if you want to respect custom equal methods

realm/realm-library/src/main/java/io/realm/RealmList.java
cmelchior: Our autogenerated `equals` method does that

realm/realm-library/src/androidTest/java/io/realm/entities/CustomMethods.java
cmelchior: Yeah, I might have missed a few nuances here. I'll double check

cmelchior:
Updated @nhachicha

realm/realm-library/src/main/java/io/realm/RealmResults.java
cmelchior: https://github.com/realm/realm-java/issues/4139 have been created to track a potential improvement to this.

PR #4114, opened 2017-01-26
Write and read in many threads

kneth:


realm/realm-library/src/androidTest/java/io/realm/RealmTests.java
```diff
@@ -3802,4 +3802,55 @@ public boolean accept(File dir, String name) {
...
         realmOnExternalStorage.close();
     }
+
+    // Stress testing multiple readers/writers on threads
+    // Related to https://github.com/realm/realm-core/issues/2383 and https://github.com/realm/realm-java/issues/2567
+    @Test
+    public void multipleReadersAndWriters() {
+        final int Nreaders = 10;
+        final int NWriters = 10;
+        final int Niterations = 1000;
```
zaki50: Should be `nReaders`, `nWriters` and `nIterations`?

kneth:
We don't need this test since it can be reproduced with core only.

PR #4116, opened 2017-01-27
remove duplicate init of the OS metadata

nhachicha:
calling `nativeConfigureMetaDataSystem` once to init the ObjectStore metadata table needed for client reset action & store Sync users


cmelchior:
Actually this should go into `releases`

PR #4120, opened 2017-01-29
Feature : Add a Fields Class generator.

Mulham-Raee:
### Specifications
- A Fields class will be generated for each **RealmObject**  with the name className + Fields. (example: **PersonsFields**).
- The generated class contains all the fields as a **public static String** variables with a value of the field name itself.

### Goals
- Reduce bloated code required for quires, and the need to use literal Strings as query keys.
- Safer code and errors detection when changing fields name, as it will normally break all queries using this field. 
- This will also adds **autocomplete** which allows navigating through all fields of the class to select one to be queried.

### Example Code
```java
       private int age;
 }

       private int id;

       @QueryFieldBy(fields = "age")
       private Dog dog;
 }
 
// Generated classes.
public final class DogFields {
       public static final String AGE = "age";
 }

public final class PersonFields {
       public static final String ID = "id";
       public static final String DOG_AGE = "dog.age";
 }
 
// Query the person with an id of 1.
realm.where(Person.class).equalTo(PersonFields.ID, 1).findFirst();
// Query all people who have a dog which is 4 years old.
realm.where(Person.class).equalTo(PersonFields.DOG_AGE, 4).findAll();
```

### DIFF
--- a/realm-annotations/src/main/java/io/realm/annotations/QueryFieldBy.java
+++ b/realm-annotations/src/main/java/io/realm/annotations/QueryFieldBy.java
@@ -1,0 +1,44 @@
+package io.realm.annotations;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+
+/**
+ * This annotation will generate an extra subset of fields inside the perspective Fields class
+ * to allow the annotated Object to be queried by its fields.
+ * <p>
+ *
+ * Example:
+ * * <pre>
+ * {@code
+ *   public class Dog extends RealmObject {
+ *       private int age;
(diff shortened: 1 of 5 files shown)


Zhuinden:
Any reason why you didn't rely on https://github.com/cmelchior/realmfieldnameshelper ?

cmelchior:
Hi @Mulham-Raee 
The reason that is not part of the offical Realm library is that user code that depends on auto-generated code comes with a number of pitfalls which we feel is not suited for all level of programmers. And one of our prime design goals is being as simple to use as possible.

The library https://github.com/cmelchior/realmfieldnameshelper does exactly what you propose though and I suggest using that instead.

Mulham-Raee:
@Zhuinden I wasn't aware of this library!
@cmelchior thanks for your explanation.

PR #4129, opened 2017-01-31
allow to skip native build by setting buildTargetABIs to empty

zaki50:
@realm/java


kneth:
I like the idea even though it is a bit of a hack :+1:

Issue #4093, opened 2017-01-23
RealmMigrationNeededException after renaming an indexed field and adding a new field with the old name


edgars-supe:
Hello!
Recently, I had to rename an indexed field and add a new field with the same name as the old name for the existing one, but without an index. Everything was okay until we noticed that users are getting an exception and can't even launch the app. 
Stack trace:
```java
io.realm.exceptions.RealmMigrationNeededException: Field count is less than expected - expected 13 but was 12
    at io.realm.EventRealmProxy.validateTable(EventRealmProxy.java:701)
    at io.realm.DefaultRealmModuleMediator.validateTable(DefaultRealmModuleMediator.java:78)
    at io.realm.Realm.initializeRealm(Realm.java:348)
    at io.realm.Realm.createAndValidate(Realm.java:313)
    at io.realm.Realm.createInstance(Realm.java:278)
    at io.realm.RealmCache.createRealmOrGetFromCache(RealmCache.java:143)
    at io.realm.Realm.getDefaultInstance(Realm.java:209)
    at [redacted]
```
Event class before the changes:
```java
public class Event extends RealmObject implements Parcelable, MediaObject {
    @PrimaryKey
    private long id;
    @Index
    private int channelId;
    private Channel channel;
    private String title;
    private String description;
    private String picture;
    private Long timeStartUnix;
    private Long timeStopUnix;
    private String category1;
    private String sysStatus;
    private boolean isArchive;
    private String continueWatchingTime;
    //...
}
```
And after:
```java
public class Event extends RealmObject implements Parcelable, MediaObject {
    @PrimaryKey
    private long id;
    @Index
    private int channelXprsId;
    private Long channelId;
    private Channel channel;
    private String title;
    private String description;
    private String picture;
    private Long timeStartUnix;
    private Long timeStopUnix;
    private String category1;
    private String sysStatus;
    private boolean isArchive;
    private String continueWatchingTime;
    //...
}
```
Here's the relevant part of the migration:
```java
if(oldVersion == 16) {
    schema.get("Event")
        .renameField("channelId", "channelXprsId")
        .addField("channelId", Long.class);
    oldVersion++;
}
```
While debugging, I noticed that the static FIELD_NAMES list contains all 13 columns, but during validation I saw that the Table object has only 12 columns (from Table.toString(), I assume), which is what the issue is about.
[...]

PR #4142, opened 2017-02-02
Add a unit tests for scenarios described in #4093

bmeike:


realm/realm-library/src/androidTest/java/io/realm/RealmMigrationTests.java
```diff
@@ -1209,6 +1211,70 @@ public void migrate(DynamicRealm realm, long oldVersion, long newVersion) {
...
     }
 
+    @Test
+    public void renameAndAddField() {
+        RealmMigration migration = new RealmMigration() {
+            @Override
+            public void migrate(DynamicRealm realm, long oldVersion, long newVersion) {
+                realm.schema.get("MigrationFieldRenameAndAdd")
+                        .renameField("string1", "string2")
+                        .addField("string1", String.class);
```
cmelchior: It would be more interesting if it had another type than String. The original `string1` was also a String, so if something goes wrong this probably just works anyway.

realm/realm-library/src/androidTest/java/io/realm/RealmMigrationTests.java
```diff
@@ -1209,6 +1211,70 @@ public void migrate(DynamicRealm realm, long oldVersion, long newVersion) {
         Realm.migrateRealm(config, migration);
     }
 
+    @Test
+    public void renameAndAddField() {
+        RealmMigration migration = new RealmMigration() {
+            @Override
+            public void migrate(DynamicRealm realm, long oldVersion, long newVersion) {
+                realm.schema.get("MigrationFieldRenameAndAdd")
```
cmelchior: use `getSchema()` instead of `schema`. Not sure why `schema` is package protected to be honest 😕

realm/realm-library/src/androidTest/java/io/realm/RealmMigrationTests.java
```diff
@@ -1209,6 +1211,70 @@ public void migrate(DynamicRealm realm, long oldVersion, long newVersion) {
...
+
+    @Test
+    public void renameAndAddIndexedField() {
+        final int oldTestVal = 7;
+        final Long testVal = Long.valueOf(293);
+
+        RealmMigration migration = new RealmMigration() {
+            @Override
+            public void migrate(DynamicRealm realm, long oldVersion, long newVersion) {
+                realm.schema.get("MigrationIndexedFieldRenamed") // WTF??  NO SUCH SCHEMA?
```
cmelchior: Remove comment

realm/realm-library/src/androidTest/java/io/realm/RealmMigrationTests.java
```diff
@@ -1209,6 +1211,70 @@ public void migrate(DynamicRealm realm, long oldVersion, long newVersion) {
...
+
+        RealmConfiguration config = configFactory.createConfigurationBuilder()
+                .schema(MigrationFieldRenameAndAdd.class)
+                .schemaVersion(2)
+                .migration(migration)
+                .assetFile("rename-and-add.realm")
+                .build();
+
+        Realm realm = Realm.getInstance(config);
+        assertTrue(realm.getSchema().get("MigrationFieldRenameAndAdd").hasField("string1"));
```
cmelchior: Perhaps extract this string since it is used 3 times. Doing `className = MigrationFieldRenameAndAdd.class.getSimpleName()` would also make it more resilient towards people refactoring the class.

realm/realm-library/src/androidTest/java/io/realm/RealmMigrationTests.java
```diff
@@ -1209,6 +1211,70 @@ public void migrate(DynamicRealm realm, long oldVersion, long newVersion) {
...
+        realm = Realm.getInstance(config);
+
+        realm.beginTransaction();
+        MigrationIndexedFieldRenamed obj = realm.createObject(MigrationIndexedFieldRenamed.class, 2);
+        obj.oldTestField = oldTestVal;
+        obj.testField = testVal;
+        realm.commitTransaction();
+
+        assertTrue(realm.getSchema().get("MigrationIndexedFieldRenamed").hasField("testField"));
+        assertTrue(realm.getSchema().get("MigrationIndexedFieldRenamed").hasField("oldTestField"));
```
kneth: Maybe check if index is still preserved. Since the original issue is about wrong number of columns, it might be worth to check that too.

kneth:
@bmeike It looks good to be but I suggest that you add a couple of extra checks so we can easier argue that the issue is not for real ;-)

realm/realm-library/src/androidTest/java/io/realm/RealmMigrationTests.java
bmeike: Yep.  Good call.

realm/realm-library/src/androidTest/java/io/realm/RealmMigrationTests.java
bmeike: THis is @kneth 's test.  I didn't mess with it...

realm/realm-library/src/androidTest/java/io/realm/entities/migration/MigrationIndexedFieldRenamed.java
```diff
@@ -0,0 +1,29 @@
+/*
+ * Copyright 2014 Realm Inc.
```
bmeike: Whoops.  Corrected in template file as well

realm/realm-library/src/androidTest/java/io/realm/RealmMigrationTests.java
bmeike: Right.  Done.

realm/realm-library/src/androidTest/java/io/realm/RealmMigrationTests.java
bmeike: Yes.  Good call. Done.

realm/realm-library/src/androidTest/java/io/realm/RealmMigrationTests.java
bmeike: Rrrr... Gone.

realm/realm-library/src/androidTest/java/io/realm/RealmMigrationTests.java
```diff
@@ -1209,6 +1211,74 @@ public void migrate(DynamicRealm realm, long oldVersion, long newVersion) {
...
+        RealmConfiguration config = configFactory.createConfigurationBuilder()
+                .schema(MigrationIndexedFieldRenamed.class)
+                .schemaVersion(2)
+                .migration(migration)
+                .assetFile("rename-and-add-indexed.realm")
+                .build();
+        realm = Realm.getInstance(config);
+
+        realm.beginTransaction();
+        MigrationIndexedFieldRenamed obj = realm.createObject(MigrationIndexedFieldRenamed.class, 2);
```
cmelchior: Mixing CLASS_NAME and this is a bit confusing. My previous comment was mostly to set `CLASSNAME = MigrationIndexedFieldNamed.class.getSimpleName()`, but having `CLASS_NAME = "MigrationIndexedFieldNamed" is fine as well.

bmeike: Yes.  I kind of agree about this.  I found it odd to be passing a String to `RealmSchema.get()`.  Perhaps we should introduce a convenience method `RealmSchema.get(Class<? extends RealmObject> klass)`

cmelchior: Perhaps. I think the reason we originally didn't want to do that was because the rest of the dynamic API is string based, and referencing a class by class reference would break if you renamed the class during a migration since you would always reference the new name instead of the old one actually in the schema. By forcing people to use Strings they "lock" down the name to a particular point in time.

And yes I know that the above argumentation is slightly against by review comments for this PR, but hey have slightly different concerns 😄

PR #4143, opened 2017-02-02
Add minVersion and targetVersion to Realm Transformer metrics

bmeike:
Fixes https://github.com/realm/realm-java-private/issues/206

### DIFF
--- a/realm-transformer/src/main/groovy/io/realm/transformer/RealmTransformer.groovy
+++ b/realm-transformer/src/main/groovy/io/realm/transformer/RealmTransformer.groovy
@@ -174,11 +174,14 @@ class RealmTransformer extends Transform {
             it.getPackageName()
         }
 
+        def targetSdk = project?.android?.defaultConfig?.targetSdkVersion?.mApiLevel as String;
+        def minSdk = project?.android?.defaultConfig?.minSdkVersion?.mApiLevel as String;
+
         def env = System.getenv()
         def disableAnalytics = env["REALM_DISABLE_ANALYTICS"]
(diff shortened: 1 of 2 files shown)


realm-transformer/src/main/groovy/io/realm/transformer/RealmTransformer.groovy
```diff
@@ -174,11 +174,14 @@ class RealmTransformer extends Transform {
             it.getPackageName()
         }
 
+        def targetSdk = project?.android?.defaultConfig?.targetSdkVersion?.mApiLevel as String;
+        def minSdk = project?.android?.defaultConfig?.minSdkVersion?.mApiLevel as String;
```
zaki50: `targetSdkVersion` and `minSdkVersion` depend on product flavor. To collect more precise information,  we should obtain them from the product flavor currently building.

I'm looking for a way to determine the product flavor.

bmeike: Well, only if they are changed by the flavor.  I believe all flavors inherit from the defaultConfig, yes?

zaki50: yes, but each flavor can override those versiin values.

bmeike: @zaki50 : Looks like pretty much a hack:  http://stackoverflow.com/questions/30621183/how-to-get-current-flavor-in-gradle

Issue #4094, opened 2017-01-23
Access token should be refreshed before the current one expires


radu-tutueanu:
I tested the 2.3.0 release with RealmTasks for Android and ROS 1.0.0 with the expiration time set to 60s.

Right now, when the Android binding tries to send an UPLOAD message with and expired token, the server sends an error(`Sending: ERROR(error_code=202, message_size=20, try_again=0, session_ident=1)`) and closes the session.

The client then connects to the auth server and gets a new token.

**This behaviour is incorrect, the client should request an access token before the current access token expires and send a REFRESH message to the server, so that the session does not get disconnected.**

Not sure how much of this behaviour should be in object-store and how much in the binding.

cc @austinzheng @mrackwitz

bdash:
The object store doesn't really have much in the way of event loop support, so setting a timer to refresh the token before it expires is currently up to the binding.

nhachicha:
we should implement a timer that refresh regularly the `access_token` using the `refresh_token` like in [cocoa](https://github.com/realm/realm-cocoa/blob/fef9c95554bb52b9f48c59249bc38a14565fdc16/Realm/RLMSyncSessionRefreshHandle.mm#L63)

The refresh interval should be based on the [accessToken.tokenData.expires](https://github.com/realm/realm-cocoa/blob/fef9c95554bb52b9f48c59249bc38a14565fdc16/Realm/RLMSyncSessionRefreshHandle.mm#L111)

@radu-tutueanu what should happen when we try to refresh an `access_token` with an expired(revoked) `refreh_token` does the server return `203` or `202` ? I ask this because one error is fatal & one is recoverable from the client perspective

radu-tutueanu:
It should be a `202`, but I am not sure. @kvap, what happens in this case?

kvap:
The server does not care about the client's old token when it is a `bind` or `refresh` message. It checks the new token that the client provided in the message and returns `203 (bad_authentication)` if the new token has expired.

nhachicha:
@kvap so `203` should force the user to log in again & try to acquire a new `refresh_token`?

kvap:
Well, it does not necessarily mean that the user has to log in again. The situation when the client provides a **new** token which is already **expired**, is strange and should be avoided. Try increasing `refreshBuffer`, or generating a fresh token without re-login just in case the user is still authentic but the new token has expired.

radu-tutueanu:
[...]

PR #4147, opened 2017-02-03
Nh/refresh access token

nhachicha:
fixes https://github.com/realm/realm-java/issues/4094

### DIFF
--- a/realm/realm-library/src/androidTestObjectServer/java/io/realm/AuthenticateRequestTests.java
+++ b/realm/realm-library/src/androidTestObjectServer/java/io/realm/AuthenticateRequestTests.java
@@ -22,6 +22,7 @@ import io.realm.util.SyncTestUtils;
 
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 import static org.mockito.Matchers.any;
 import static org.mockito.Mockito.when;
@@ -60,10 +61,10 @@ public class AuthenticateRequestTests {
     @Test
     public void userRefresh() throws URISyntaxException, JSONException {
         Token t = SyncTestUtils.createTestUser().getSyncUser().getUserToken();
-        AuthenticateRequest request = AuthenticateRequest.userRefresh(t);
+        AuthenticateRequest request = AuthenticateRequest.userRefresh(t, new URI("realm://objectserver/" + t.value() + "/default"));
 
         JSONObject obj = new JSONObject(request.toJson());
-        assertFalse(obj.has("path"));
+        assertTrue(obj.has("path"));
         assertEquals(t.value(), obj.get("data"));
         assertEquals("realm", obj.get("provider"));
     }
--- a/realm/realm-library/src/androidTestObjectServer/java/io/realm/AuthenticateRequestTests.java
+++ b/realm/realm-library/src/androidTestObjectServer/java/io/realm/AuthenticateRequestTests.java
@@ -40,10 +40,10 @@ public class AuthenticateRequestTests {
     @Test
     public void realmLogin() throws URISyntaxException, JSONException {
         Token t = SyncTestUtils.createTestUser().getSyncUser().getUserToken();
-        AuthenticateRequest request = AuthenticateRequest.realmLogin(t, new URI("realm://objectserver/" + t.value() + "/default"));
+        AuthenticateRequest request = AuthenticateRequest.realmLogin(t, new URI("realm://objectserver/" + t.identity() + "/default"));
 
         JSONObject obj = new JSONObject(request.toJson());
-        assertEquals("/" + t.value() + "/default", obj.get("path"));
+        assertEquals("/" + t.identity() + "/default", obj.get("path"));
         assertEquals(t.value(), obj.get("data"));
         assertEquals("realm", obj.get("provider"));
     }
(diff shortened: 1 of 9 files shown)


nhachicha:
@realm/java

realm/realm-library/src/objectServer/java/io/realm/internal/objectserver/ObjectServerSession.java
```diff
@@ -290,6 +324,76 @@ protected void onError(AuthenticateResponse response) {
         networkRequest = new RealmAsyncTaskImpl(task, SyncManager.NETWORK_POOL_EXECUTOR);
     }
 
+    private void scheduleRefreshAccessToken(long expireDateInMs) {
+        // calculate the delay time before which we should refresh the access_token,
+        // we adjust to 10 second to proactively refresh the access_token before the session
+        // hit the expire date on the token
+        long refreshAfter =  expireDateInMs - System.currentTimeMillis() - TimeUnit.SECONDS.toMillis(10);
```
kneth: Should `10` be a constant?

realm/realm-library/src/androidTestObjectServer/java/io/realm/AuthenticateRequestTests.java
```diff
@@ -60,10 +61,10 @@ public void userLogin() throws URISyntaxException, JSONException {
     @Test
     public void userRefresh() throws URISyntaxException, JSONException {
         Token t = SyncTestUtils.createTestUser().getSyncUser().getUserToken();
-        AuthenticateRequest request = AuthenticateRequest.userRefresh(t);
+        AuthenticateRequest request = AuthenticateRequest.userRefresh(t, new URI("realm://objectserver/" + t.value() + "/default"));
```
cmelchior: I thought it was `identity` not `value` used here?

realm/realm-library/src/objectServer/java/io/realm/internal/objectserver/ObjectServerSession.java
```diff
@@ -290,6 +324,76 @@ protected void onError(AuthenticateResponse response) {
...
+    // Authenticate by getting access tokens for the specific Realm
+    private void refreshAccessToken() {
+        // Authenticate in a background thread. This allows incremental backoff and retries in a safe manner.
+        if (refreshTokenNetworkRequest != null) {
+            refreshTokenNetworkRequest.cancel();
+        }
+        Future<?> task = SyncManager.NETWORK_POOL_EXECUTOR.submit(new ExponentialBackoffTask<AuthenticateResponse>() {
+            @Override
+            protected AuthenticateResponse execute() {
+                return authServer.refreshUser(user.getUserToken(), configuration.getServerUrl(), user.getAuthenticationUrl());
```
cmelchior: It doesn't seem to retry if something goes wrong?

realm/realm-library/src/objectServer/java/io/realm/internal/objectserver/ObjectServerSession.java
```diff
@@ -108,6 +113,8 @@
     private SyncSession userSession;
     private SyncSession publicSession;
 
+    private final static ScheduledThreadPoolExecutor REFRESH_TOKENS_EXECUTOR = new ScheduledThreadPoolExecutor(1);
```
cmelchior: Is one thread enough?

realm/realm-library/src/objectServer/java/io/realm/internal/objectserver/ObjectServerSession.java
```diff
@@ -290,6 +324,76 @@ protected void onError(AuthenticateResponse response) {
...
+                    );
+                    user.addRealm(configuration.getServerUrl(), desc);
+                    // schedule the next refresh
+                    scheduleRefreshAccessToken(response.getAccessToken().expiresMs());
+                }
+            }
+
+            @Override
+            protected void onError(AuthenticateResponse response) {
+                // unrecoverable error, reschedule will not happen
```
cmelchior: Log this to INFO/WARN/ERROR?

realm/realm-library/src/objectServer/java/io/realm/internal/objectserver/ObjectServerSession.java
nhachicha: Theoretically there should be one logged in user, so one token to refresh max

cmelchior: Good point. Multiple users are supported though, but you are right that e.g `currentUsers()` will crash for multiple logged in users, so perhaps starting out with 1 thread is fine. Then we can adjust if need be.

realm/realm-library/src/objectServer/java/io/realm/internal/objectserver/ObjectServerSession.java
nhachicha: This uses `ExponentialBackoffTask` and retry only in case of network error (`IO_EXCEPTION`), any other errors should stop the refresh. Any use of this invalid `access_token` should force the server to return a `202` which will cause a new authentication that will obtain a new `access_token`, and the refresh process will be resumed again

beeender:
Session's stop is a synchronized method, but the token scheduler is not synchronized. It seems to be a problem when stop gets called and scheduler time out at the same time?

realm/realm-library/src/objectServer/java/io/realm/internal/objectserver/ObjectServerSession.java
```diff
@@ -290,6 +325,76 @@ protected void onError(AuthenticateResponse response) {
...
+            refreshTokenNetworkRequest.cancel();
+        }
+        Future<?> task = SyncManager.NETWORK_POOL_EXECUTOR.submit(new ExponentialBackoffTask<AuthenticateResponse>() {
+            @Override
+            protected AuthenticateResponse execute() {
+                return authServer.refreshUser(user.getUserToken(), configuration.getServerUrl(), user.getAuthenticationUrl());
+            }
+
+            @Override
+            protected void onSuccess(AuthenticateResponse response) {
```
beeender: This will be called in the bg thread and `stop` can be called at the same time :(
So it would be easier to define a `synchronized` method like `private void doOnSuccess() synchronized` to do things here or just put below into a `synchronized (this)` block.

cmelchior:
Also this should go into `releases` ?

realm/realm-library/src/objectServer/java/io/realm/internal/objectserver/ObjectServerSession.java
nhachicha: good catch 👍🏼

PR #4155, opened 2017-02-08
Use CMake's ability to do transitive dependency resolution

emanuelez:
Cmake has the ability to do transitive dependency resolution. This means that we can simply say that core depends on sync and that sync depends on core and cmake will do the rest, including ordering libraries correctly for the linker to consume. This also avoids some redundancy.

This PR came out of a debugging session with @kneth where we were both facing linking issues with atomic on MIPS because of the wrong ordering of the libraries during the invocation of `ld`.


emanuelez:
Hmm.. maybe this should do this in `releases`

Issue #3833, opened 2016-11-23
RealmMigrationNeededException: Field count is more than expected - expected 10 but was 11


jakubjodelka:
Hi, 
I have problem with migration in my application. All time logcat gives me "io.realm.exceptions.RealmMigrationNeededException: Field count is more than expected - expected 10 but was 11 after migration". 

My old model:
```
    @PrimaryKey
    private int sid;
    private String title;
    private String noSpecialCharsTitle;
    private String lang;
    private String img;
    private String icon;
    private boolean owned;
    private boolean favorite;
    private int favouritePosition = Integer.MAX_VALUE;
```

My new model: 
```
    @PrimaryKey
    private int sid;
    private String title;
    private String noSpecialCharsTitle;
    private String lang;
    private String img;
    private String icon;
    private boolean owned;
    private boolean favorite;
    private int favouritePosition = Integer.MAX_VALUE;
    private int stationPosition = Integer.MAX_VALUE;
```

Migration: 
```Realm.init(ApplicationClass.getAppContext());
        RealmConfiguration realmConfig =  new RealmConfiguration.Builder()
                .schemaVersion(SCHEMA_VERSION)
                .migration((realm, oldVersion, newVersion) -> {
                    if(oldVersion == 0) {
                        final RealmObjectSchema station = realm.getSchema().get("Station");
                        station.addField("stationPosition", int.class);
                        oldVersion++;
                    }
                })
                .build();
        Realm.setDefaultConfiguration(realmConfig);
```

What's wrong with this migration? Why app always throws RealmMigrationNeededException?

Zhuinden:
You can try

        RealmConfiguration realmConfig =  new RealmConfiguration.Builder()
                .schemaVersion(SCHEMA_VERSION)
                .migration((realm, oldVersion, newVersion) -> {
                    if(oldVersion == 0) {
                        final RealmObjectSchema station = realm.getSchema().get("Station");
                        if(!station.hasField("stationPosition") {
                            station.addField("stationPosition", int.class);
                        }
                        oldVersion++;
                    }
                })
                .build();
        Realm.setDefaultConfiguration(realmConfig);

jakubjodelka:
Still same crash. 
```
RealmMigrationNeededException: Field count is more than expected - expected 10 but was 11. 
```

Zhuinden:
Out of curiousity, what happens if you don't specify a migration at all?

jakubjodelka:
Without any migration: 
```
[...]

PR #4164, opened 2017-02-09
RealmResults is always live-to-updated

beeender:
This is the precondition of fine grained notifications. OS will trigger
the collection notification immediately when transaction begins on the
local thread to compute the change set if there is any. This conflicts
with Java's original RealmResults behavior -- the original RealmResults
would only be synced in the next event loop.
Also, there are some edge cases don't work well with the original
RealmResults behavior, see details in #3833.

So:
- RealmResults becomes always up-to-date again which means it will
  never contains a invalid row.
- Behavior of iteration on a RealmResults still just works, it will just
  iterate on snapshot of collection. This means user can still delete
  elements from a RealmResults inside iteration.
- Deletion & Modification on RealmResults in simple-for-loop won't work
  as expected if the changes will impact the order/elements of the
  results. This could be solved by the future new Collection type
  RealmCollectionSnapshot.
- Add Collection.load() and Collection.isLoaded() to support java sync
  queries.
- Test fix.
  - https://github.com/realm/realm-android-adapters/issues/48 won't be an
    issue anymore since the RealmResults is always up to date and it won't
    contain any invalid rows. So remove the related tests.
  - Failure tests caused by listener being triggred with beginTransaction()
  - Remove realmResultsListenerAddedAfterCommit.
    when add listener to the OS Results after commit transaction, the
    OS CollectionNotifier will be created at the SharedGroup version
    of transaction committed. So the listener won't be called anymore since
    the all changes already exist in current SharedGroup.


realm/realm-library/src/main/java/io/realm/RealmResults.java
```diff
@@ -102,7 +99,7 @@ public boolean isValid() {
     }
 
     /**
-     * A {@link RealmResults} is always a managed collection.
+     * A {@link RealmResults} is always a managed iteratorCollection.
```
cmelchior: Looks like a mistake?

realm/realm-library/src/main/java/io/realm/internal/Collection.java
```diff
@@ -42,27 +40,158 @@ public void onChange(T observer) {
...
+        }
+
+        /**
+         * Not supported by Realm collection iterators.
+         *
+         * @throws UnsupportedOperationException
+         */
+        @Deprecated
+        public void remove() {
+            throw new UnsupportedOperationException("remove() is not supported by RealmResults iterators.");
```
cmelchior: This is also used by RealmList iterators right?

realm/realm-library/src/main/java/io/realm/internal/Collection.java
```diff
@@ -42,27 +40,158 @@ public void onChange(T observer) {
...
+        // like what realm-cocoa does, we will have a massive overhead since all the iterators created in the previous
+        // event loop need to be detached.
+        void invalidate() {
+            iteratorCollection = null;
+        }
+
+        void checkValid() {
+            if (iteratorCollection == null)  {
+                throw new ConcurrentModificationException(
+                        "No outside changes to a Realm is allowed while iterating a RealmResults." +
```
cmelchior: RealmList? Also we don't have `Realm.refresh()`

realm/realm-library/src/main/java/io/realm/internal/Collection.java
```diff
@@ -42,27 +40,158 @@ public void onChange(T observer) {
...
+        @Deprecated
+        public void remove() {
+            throw new UnsupportedOperationException("remove() is not supported by RealmResults iterators.");
+        }
+
+        void detach() {
+            iteratorCollection = iteratorCollection.createSnapshot();
+        }
+
+        // The iterator become invalid after receiving a remote change notification. In Java, the destruction of
```
cmelchior: Wouldn't it make sense to invalidate the iterator if it runs to completion (hasNext = false) ?

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ### Breaking changes
 
 * `RealmResults.distinct()` returns a new `RealmResults` object instead of filtering on the original object (#2947).
+* `RealmResults` is auto-updated all the time. Any transaction on the caller thread which may have impact on the order or elements of the `RealmResults` will change the `RealmResults` immediately instead of change it in the next event loop. Iterator behavior of `RealmResults` stays the same, transaction inside the iterating still works as expected.
```
Zhuinden: Does that mean that even the good ol' `for(int i = results.size()-1; i >= 0; i--)` won't work either?

cmelchior: No that would probably work

realm/realm-library/src/main/java/io/realm/internal/Collection.java
beeender: not now. `RealmList` still based on the old implementation (calling methods on LinkView). We can do it later since all behavior of `RealmList` stays the same.

realm/realm-library/src/main/java/io/realm/internal/Collection.java
beeender: But then calling `hasNext()` the 2nd time will throw exception with `checkValid()`. That doesn't sounds like an expected behavior.

cmelchior: No, that would be bad. I was mostly thinking of it as an optimization, but if it introduces all kinds of weird behaviour in our implementation it might not be worth it.

realm/realm-library/src/main/cpp/io_realm_internal_Collection.cpp
```diff
@@ -420,10 +350,8 @@ Java_io_realm_internal_Collection_nativeDeleteFirst(JNIEnv *env, jclass, jlong n
 
     try {
         auto wrapper = reinterpret_cast<ResultsWrapper*>(native_ptr);
-        if (wrapper->get_results().size() > 0) {
-            wrapper->get_results().get_tableview().remove(0);
-            // Refresh snapshot
-            wrapper->refresh_snapshot();
+        if (wrapper->m_results.size() > 0) {
+            wrapper->m_results.get_tableview().remove(0);
```
beeender: Great point! And also the `get_tableview()` creates some unnecessary overhead.
Changed to use `Row::move_last_over()` instead.

beeender: @kneth By using the `move_last_over()`, `deleteFirstFromRealm` will fail because of `createNonCyclicCollection` doesn't sort the Results. But yes, we can fix it by calling the `remove()` instead. But shall we keep the original behavior to make it slower?

```
notSortedResults = {0, 1, 2, 3}
notSortedResults.deleteFromRealm(1);
// By using move_last_over it will become
notSortedResults = {0, 3, 2}
```

IMO, we should use `move_last_over()` since we never guarantee the order of unsorted `Results` is stable.

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ### Breaking changes
 
 * `RealmResults.distinct()` returns a new `RealmResults` object instead of filtering on the original object (#2947).
+* `RealmResults` is auto-updated continuously. Any transaction on the caller thread which may have an impact on the order or elements of the `RealmResults` will change the `RealmResults` immediately instead of change it in the next event loop. Iterator behavior of `RealmResults` stays the same, transactions inside the iterating still works as expected.
```
cmelchior: caller -> current ?

realm/realm-library/src/main/java/io/realm/internal/Collection.java
```diff
@@ -42,27 +40,157 @@ public void onChange(T observer) {
...
```
dalinaum: The iterator becomes?

CHANGELOG.md
cmelchior: What does it mean they stay the same? You mean the Iterator uses a snapshot or?

realm/realm-library/src/androidTest/java/io/realm/OrderedRealmCollectionIteratorTests.java
```diff
@@ -801,6 +800,7 @@ public void run() {
     }
 
     @Test
+    @Ignore("Enable this test when support RealmCollectionSnapshot")
```
cmelchior: I don't think this test is re-enabled in your #4172 ?

CHANGELOG.md
beeender: @cmelchior Yes, the iterator is actually using a snapshot underneath. would it be clear like:

> Iterator behavior of `RealmResults` stays the same, you can still delete/modify the `RealmResults` while iterating the it through iterators.

realm/realm-library/src/androidTest/java/io/realm/OrderedRealmCollectionIteratorTests.java
beeender: It is, see https://github.com/realm/realm-java/pull/4172/files#diff-07a5d547f084d754d3b6e6b0fdfeff26R834

CHANGELOG.md
cmelchior: Something like that yes, but we still need to be very clear when it now breaks. Perhaps "The standard `RealmResults.iterator()` will continue to work as normal, which means that you can still delete or modify elements without impacting the iterator. The same is not true for simple for loops. In some cases a simple for loop will not work (link-to-example), and you must either iterate backwards (link-to-example) or use the new `createSnapshot()` method."

What do you think?

beeender: Yeah, it seems your English teacher is much better than mine!

realm/realm-library/src/androidTest/java/io/realm/RealmAsyncQueryTests.java
```diff
@@ -1104,23 +1105,36 @@ public void badVersion_syncTransaction() throws NoSuchFieldException, IllegalAcc
...
             }
         });
 
         // 2. Advances the caller Realm, invalidating the version in the handover object.
         realm.beginTransaction();
+        assertTrue(result.isLoaded());
         realm.createObject(AllTypes.class);
         realm.commitTransaction();
 
         // 3. The async query should now (hopefully) fail with a BadVersion.
```
beeender: I put a `NOTE` on top of this test:
>     // NOTE: This test is not checking the same thing after the OS results integration. Just keep it for an additional test for async.
To make it less confusing, i decided to keep the original comments there :P

Issue #1694, opened 2015-10-29
Support for custom TypeAdapters


saket:
Have you considered a way to support custom data types? Maybe by registering a type converter? ORMs like GreenDao support this feature and its very useful. We're currently using the `@ignore` workaround but its not clean.

cmelchior:
Yes we have considered it, and it is not impossible it will appear at some point. Once we add support for custom methods (https://github.com/realm/realm-java/issues/909) this will become a lot easier though as you can then do the conversion in the getter/setter.

That said, having a custom type adapter will reduce a lot of boilerplate code across model classes.

Saketme:
We're already using custom getters for conversion by using `@Ignore` annotation, but with TypeAdapters, it'll become super easy :D

cmelchior:
I'm reopening this, because we didn't have a public issue tracking this. Might as well be this one :)

AndreasBackx:
@cmelchior Any update on this? This would be awesome for `java.time` classes which can be used on Android using [ThreeTenABP](https://github.com/JakeWharton/ThreeTenABP).

cmelchior:
Hi @AndreasBackx 
Sorry, no update. Right now we are focusing on other features, but after 0.88 this use case should be a lot easier to handle as you can make the conversion in the getter/setters themselves.

AndreasBackx:
@cmelchior yes. I'm using custom getters and setters right now. Thank you for the reply.

AndreasBackx:
@cmelchior This is a reply to the comment made in #1072.

I don't see why having a simple yet atrocious workaround makes for a low priority. The example you're giving me is what I gave and what has already been talked about in this issue. The problem is that you'll have to the getter/setter method for every single instance. This is a pain and if you change on thing, you're gonna have to change all of the other ones. This can cause bugs if one forgets to change one of them.

Making an intermediary object is, as I have previously stated in my previous comment(s), adding a barrier to the data layer of your code. The few barriers people seemingly have to face with Realm is why Realm is getting more popular. But it kind of feels like I'm working with raw SQL again when I keep having to add _barriers_ to my code.

I would much rather have something supported by Realm itself than having to do a workaround everywhere and it's not as if I could ever be the only one. I'm sure there are a lot of people (and I'm not solely talking about people running Realm) that use types that are not supported by Realm. Why should we have to create hacky workaround for what is normally regarded as good code?

I'm willing to help make a PR for this, but I would have to know what your vision is on this first.

[...]

PR #4165, opened 2017-02-09
RFC: Primitive List API

cmelchior:
With work in Core slowly picking up speed regarding Primitive List support, we need to start thinking about how the public API is going to look like. Also, so it can guide the underlying implementation.

This document tries to flesh out the various proposals that have been discusse so we can get feedback on the direction as well as guide Cores design decisions.

Thoughts @realm/java ?

### DIFF
--- a/rfcs/primitive_lists_api.md
+++ b/rfcs/primitive_lists_api.md
@@ -234,10 +234,11 @@ sum // only numbers // only numbers
 * Schemas can already be restrict using modules, making `<E extends RealmModel>` inaccurate anyway.
 * RealmCollection will have to contain optional methods no matter what, so adding them to RealmQuery doesn't break
   that pattern.
+* Would allow us to remove the requirement that people implements `RealmModel`, they could just use the `@RealmClass`
+  annotation.
 
 **Disadvantages**
-* People can now accidentally try to save non-realm classes, which will not break before runtime (IntelliJ plugin might
-  help here).
+* API gets a lot less type safe (IntelliJ plugin might help here).
 * `RealmQuery` will suddenly have a ton of new methods that will only work for primitive arrays (lint check?)
 
 
@@ -257,14 +258,25 @@ sum // only numbers // only numbers
 * Type safety is not guaranteed anyway if we allow `<Object>`` as generic.
 
 
+# Work required
+
+- [ ] Extend JSON API to support primitive lists.
+- [ ] Extend RealmObjectSchema to support primitive lists
+- [ ] Extend DynamicRealmObject to support primitive lists (Design required).
+- [ ] Annotation processor must be updated
+- [ ] Add support for primitive arrays using RealmTransformer (maybe not part of this RFC?)
+- [ ] Implement new classes (depends on chosen solution).
+- [ ] Refactor current Collection API's + unit tests.
+
 # Unresolved questions
 
-We need more feedback on the two proposals.
+1. We need more feedback on the two proposals.
 
-* Unclear how we are going to support lists-of-lists, e.g. what kind of queries do you want to run on a `int[][]`
+2. Unclear how we are going to support lists-of-lists, e.g. what kind of queries do you want to run on a `int[][]`
   (if any) ?
 
-None.
+3. Should we support `int[]` using bytecode transformation? Basically replace `int[]` and friends with `RealmList<Integer/.../...>` ?
+
 
 # Version history
 
--- a/rfcs/primitive_lists_api.md
+++ b/rfcs/primitive_lists_api.md
@@ -1,0 +1,271 @@
+- Feature Name: Primitive List API
+- Start Date: 2017-02-07
+- RFC PR:
+- Version: 1
+
+# Summary
+
+Currently Realm only supports lists of Realm Objects, but we should also support lists
+of primitive types like `int`, `String` and `boolean`. This document describes how such
+an API might look like.
+
+# Motivation
+
+With Core working on support for primitive lists, we need to consider how the public API is going to look like as it
+might influence some of the design choices for the underlying implementation. This RFC is mostly to explore the
+possible options for the public API and elicit feedback.
+
+# Variant 1: Relax generic type arguments
+
+In this variant, we relax the type requirement on RealmList, RealmQuery and RealmResults to `Object`.
+
+1) Relax type requirement for `RealmCollection`, `OrderedRealmCollection`, `RealmResults`, `RealmList` and `RealmQuery`
+
+```
+// Current signature
+public class RealmList<E extends RealmModel> extends AbstractList<E> implements OrderedRealmCollection<E>
+public class RealmQuery<E extends RealmModel> {
+public class RealmResults<E extends RealmModel> extends AbstractList<E> implements OrderedRealmCollection<E>
+public interface OrderedRealmCollection<E extends RealmModel> extends List<E>, RealmCollection<E>
+public interface RealmCollection<E extends RealmModel> extends Collection<E>
+
+// New signature
+public class RealmList<E> extends AbstractList<E> implements OrderedRealmCollection<E>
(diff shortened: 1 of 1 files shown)


rfcs/primitive_lists_api.md
```diff
@@ -0,0 +1,283 @@
...
+public interface RealmCollection<E> extends Collection<E>
+```
+
+1b) Add new classes with `Primitive*` as an abstract super class:
+
+Precedence: SparseIntArray, SparseBooleanArray, etc.
+
+```
+protected abstract class PrimitiveRealmQuery<Object>
+public class PrimitiveIntegerRealmQuery extends PrimitiveRealmQuery<Integer>
```
zaki50: Shouldn't we have `public class PrimitiveLongRealmQuery extends PrimitiveRealmQuery<Long>`?

cmelchior: Ups, yes.

kneth:
I prefer option 1. From a user's perspective is it easier: `RealmList<Long>` and `RealmList<Cat>` are "just" lists. Forcing the users to think about details of the types (`Long` and `Cat`) feels wrong.

We are pushing the fluent interface of the query engine. `RealmQuery` is becoming more and more bloated, and I think it is worth considering switching to the same query system as Realm JS.

Zhuinden:
I think laxing the `extends RealmModel` and just using `Object` instead will help with and is actually a requirement of https://github.com/realm/realm-java/issues/1694

cmelchior:
Depending on how Type Adapters are implemented that might not be necessary, but it would probably make a lot of things easier. Good point. 

Regarding the query system. We won't remove the typed API, but it is starting to squeak a bit under the weight of operators. So a string based API would live next to the typed one. At least that is the current idea.

nirinchev:
For .NET, option 1 will probably work better - we use the standard `IList<T>` on which we can relax the generic constraint to allow all types, naturally, issuing a compile time error if used with anything other than `RealmObject` or valid primitive type. Since we're using LINQ, the RealmQuery method explosion doesn't concern us. The API will be type-safe, but it might be harder to expose some functionality, e.g. `like` searches. cc @kristiandupont @fealebenpae

nhachicha:
I'm in favour of relaxing the type requirement on `RealmList`. As discussed with Zaki we can still check at compile time that the user doesn't abuse the generic, by using an unsupported type.

Example: we can detect at compile time this `RealmnList<BigInteger>` and throw an exception

cmelchior:
Good point. We should be able to catch all wrong types used in the annotation processor. Didn't think of that 👍 

If we relax the type parameters, I'm also secretly playing around with the idea of splitting out DynamicRealmObject from RealmObject since having to deal with both of them everywhere is a big hassle and put some rather annoying restrictions on what we can do in both classes.

Zhuinden:
Ah, then ability to get Realm / DynamicRealm from their corresponding classes will be a possibility :3

zaki50:
OK, I will start adding APIs with `option 1`.
Any objections? @realm/java

zaki50:
In regards to `Design1, 4a`, how about accepting `null` as field name for primitive queries instead of adding new methods? @realm/java

beeender:
> In regards to Design1, 4a, how about accepting null as field name for primitive queries instead of adding new methods?

I can accept it. Feels kind of nature to me.

cmelchior:
I think the 4a) option was not thought through. Querying would have to work like you would do with a `RealmIntegerWrapper` today but without the linked query.  Consider
```
public class ListDemo extends RealmObject {
  public RealmList<String> firstNames;
  public RealmList<String> lastNames;
  public String name;
}

// Currently with a wrapper type it would be 
realm.where(ListDemo.class).equalTo("firstNames.value", "John").findAll(); 

// With primitive support it would be
realm.where(ListDemo.class).equalTo("firstNames", "John").findAll();

// Breaks if you have multiple primitive lists
realm.where(ListDemo.class).equalTo("John").findAll();
```

cmelchior:
A crap, now I remember:

```
RealmList.getFirstNames().where().equalTo(null, "John");
```

rfcs/primitive_lists_api.md
```diff
@@ -0,0 +1,283 @@
...
+* RealmQuery/RealmResults already contain methods that doesn't work in certain circumstances.
+* Schemas can already be restrict using modules, making `<E extends RealmModel>` inaccurate anyway.
+* RealmCollection will have to contain optional methods no matter what, so adding them to RealmQuery doesn't break
+  that pattern.
+* Would allow us to remove the requirement that people implements `RealmModel`, they could just use the `@RealmClass`
+  annotation.
+
+**Disadvantages**
+* API gets a lot less type safe (IntelliJ plugin might help here).
+* `RealmQuery` will suddenly have a ton of new methods that will only work for primitive arrays (lint check?)
```
bmeike: Right.  I can see that I am in a minority, here, but this is what types are *for*.  I think this argument, all by itself, should be a stopper.

rfcs/primitive_lists_api.md
```diff
@@ -0,0 +1,283 @@
...
+* No changes to existing code. Primitive arrays will be a complete new addon.
+* With 1b) we statically check everything using the type system on compile time.
+
+**Disadvantages**
+* Explosion in classes in the Realm API. Maintenance and test will be problematic.
+* It is not clear how to handle primitive arrays in the current Collection hierarchy. It seems a bit pointless to allow
+  optional methods for primitive lists in the interface if we go to great lengths in the actual class to make it type
+  safe.
+* Type safety is not guaranteed anyway if we allow `<Object>`` as generic.
+
```
bmeike: Note that, to the contrary, type safety is still guaranteed for all of RealmList.  It just isn't guaranteed here.

rfcs/primitive_lists_api.md
cmelchior: Note, that even with types, the current system `<? extends RealmModel>` isn't perfect as a Schema can restrict valid types even further.

bmeike: Noted.

zaki50:
I made two PRs to define Primitive List API.
One is Design 1 (#5145) and another is Design 2 (#5139).

I was thinking that I had a problem in `Design 1`, but now I solved most ob them and now I'm leaning towards `Design 1`.

cmelchior:
I strongly prefer Design 1 as well. There are a few caveats like `findFirstAsync()` not working on value lists and nullability annotations forcing null checks more places than strictly needed, but I think that is acceptable.

zaki50:
OK, I start implementing the feature on `Design 1`.

rfcs/primitive_lists_api.md
```diff
@@ -0,0 +1,271 @@
...
+notEqualTo(String)
+notEqualTo(String, Case)
+sum()
+```
+
+Benefit: Intent expressed in method params. All queries use the same class Downside: Ton of new methods. More chances for
+people to use the "wrong" method. Today people still have to remember not to call e.g. `sum` on a String field though.
+
+
+4b) Add support for a special keyword for column names. NSPredicate apparently uses `SELF`, `$` could also work.
```
Zhuinden: `$` isn't that good because you need to put it in backticks in kotlin.

rfcs/primitive_lists_api.md
```diff
@@ -0,0 +1,271 @@
...
+  optional methods for primitive lists in the interface if we go to great lengths in the actual class to make it type
+  safe.
+* Type safety is not guaranteed anyway if we allow `<Object>`` as generic.
+
+
+# Unresolved questions
+
+We need more feedback on the two proposals.
+
+* Unclear how we are going to support lists-of-lists, e.g. what kind of queries do you want to run on a `int[][]`
```
Zhuinden: oh, so that's what "nested subtables" means.

Issue #3883, opened 2016-12-07
Stable iterators for RealmCollection


beeender:
### Problem

This has been discussed before
https://github.com/realm/realm-java/issues/640
https://github.com/realm/realm-java/issues/602

https://github.com/realm/realm-wiki/wiki/Java-Iterators (Internal wiki page)

So #2124 was introduced to solve below problem:

```java
RealmResults<Person> persons = realm.allObjects(Person.class);
realm.beginTransaction();
for (int i = 0; persons.size(); i++) {
	// Will only delete half the elements due to size() calling
	// sync_if_needed.
	persons.get(i).removeFromRealm(); 
 }
realm.commitTransaction();
```

Then we added #2990 to make sure the `RealmResults` gets updated in very next event loop by calling `handler.postAtFrontOfQueue()`. 

Usually above works fine, this becomes a topic again for below reasons:

#### 1. However there are still corner cases which could trigger an exception:

> java.lang.IllegalStateException: Object is no longer managed by Realm. Has it been deleted?

eg.: Delete an element in `RealmResults` then calling the corresponding `ListView.measure()` method. `measure()` will try to get views by calling adapter's `getView()` to measure, but at this time, since it is still in the same event loop the elements get deleted, the `RealmResults` has not been updated yet.

#### 2. PR #3834 met some difficulties to support the exact same behavior with OS `Results`.

`Results` will sync the `TableView` when ever any accessor called unless it is an snapshot.

### `Results::snapshot()`

This call returns a special `Results` object that never updates to reflect changes in the underlying data.

Since `Results` can be created on `Query`/`LinkView`/`TableView`, so a snapshot can be created from those as well.

#### 1. Change all `RealmResults`'s  backend to a snapshot when `beginTransaction()` called. 
And change them back in the very next event loop. This could allow us to make #3834 and stay with current `RealmResults`'s behaviors.
But this won't solve the problem 2 and 3.

#### 2. Introduce a new `RealmCollection` type `RealmSnapshotCollection`. 

`RealmSnapshotCollection` is backed by a snapshot of OS `Results`. So the collection itself won't be updated, size and order stay the same forever (elements inside are still live objects.).
It could be created by `RealmResults.snapshot()` or `RealmList.snapshot()`.
Since it won't be auto-updated, so there is no point to allow listener related APIs on it. Calling `addChangeListener` will just throw.
The iteration (both simple loop and foreach loop) will be always stable.
[...]

PR #4172, opened 2017-02-11
Add OrderedRealmCollectionSnapshot

beeender:
Introduce OrderedRealmCollectionSnapshot as another type of
OrderedRealmCollection.
A snapshot can be created from RealmResults or RealmList.
A snapshot is backed by a snapshot of OS Results. So the snapshot
itself won't be updated. The size and order stay the same forever
(elements inside are still live objects.).

Since the RealmResults is auto-updated all the time, snapshot will be
usefull when changing the results in a simple loops.

This commit also moves the common code from RealmResults to
OrderedRealmCollectionImpl since those can be shared with snapshot
implementation.

Implement #3883


Zhuinden:
I'm guessing snapshots cannot be passed between threads either, right?

But they also don't auto-update anymore?

beeender:
No, it cannot be passed between threads. The objects in the snapshot are still auto-updated, but the snapshot itself is not.

beeender:
CI fails with case:
close Realm in `RealmResults`'s listener then call `Realm.refresh()`. 

Needs to be fixed in Object Store.

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 ### Enhancements
 
 * Added support for sorting by link's field (#672).
+* Added `OrderedRealmCollectionSnapshot` class and `OrderedRealmCollection.createSnapshot()` methods. `OrderedRealmCollectionSnapshot` would be useful when changing `RealmResults` or `RealmList` in simple loops.
```
kneth: "methods" -> "method"
"would" -> "is"

realm/realm-library/src/androidTest/java/io/realm/CollectionTests.java
```diff
@@ -34,7 +34,9 @@
 
     // Enumerates all known collection classes from the Realm API.
     protected enum CollectionClass {
-        MANAGED_REALMLIST, UNMANAGED_REALMLIST, REALMRESULTS
+        MANAGED_REALMLIST, UNMANAGED_REALMLIST, REALMRESULTS,
+        REALMRESULTSSNAPSHOT_RESULTS_BASE, REALMRESULTSSNAPSHOT_LIST_BASE
```
kneth: `REALMRESULTS_SNAPSHOT_RESULTS_BASE` and `REALMRESULTS_SNAPSHOT_LIST_BASE` maybe

realm/realm-library/src/androidTest/java/io/realm/ManagedRealmCollectionTests.java
```diff
@@ -500,6 +560,29 @@ public void minDate() {
         assertEquals(new Date(-YEAR_MILLIS * 20 * TEST_SIZE / 2), collection.minDate(AllJavaTypes.FIELD_DATE));
     }
 
+    // Delete the last row in the collection then test the aggregates methods.
+    // Since deletion will turn the corresponding object into invalid for collection snapshot, this tests if the
+    // aggregates methods ignores the invalid rows and returns the correct result.
```
kneth: "ignores" -> "ignore"
"returns" -> "return"

realm/realm-library/src/androidTest/java/io/realm/OrderedRealmCollectionSnapshotTests.java
```diff
@@ -0,0 +1,88 @@
...
+        for (int i = 0; i < testSize; i++) {
+            AllTypes allTypes = realm.createObject(AllTypes.class);
+            allTypes.setColumnLong(i);
+        }
+        realm.commitTransaction();
+    }
+
+    @Test
+    public void deleteFromRealm_twice() {
+        realm.beginTransaction();
```
kneth: No assertions?

realm/realm-library/src/main/java/io/realm/OrderedRealmCollectionSnapshot.java
```diff
@@ -0,0 +1,191 @@
+package io.realm;
+
+import io.realm.internal.Collection;
+import io.realm.internal.UncheckedRow;
+
+/**
+ * An {@link OrderedRealmCollectionSnapshot} is a special type of {@link OrderedRealmCollection}. It can be created by
+ * calling {@link OrderedRealmCollection#createSnapshot()}. Unlike {@link RealmResults} and {@link RealmList}, it's
+ * size and orders of elements will never be changed after creation.
```
kneth: Should we mention something about invalid objects? And that `createSnapshot()` is almost idempotent?

realm/realm-library/src/main/java/io/realm/OrderedRealmCollection.java
```diff
@@ -137,4 +137,14 @@
      * @throws UnsupportedOperationException if the collection is unmanaged.
      */
     boolean deleteLastFromRealm();
+
+    /**
+     * Creates a snapshot from this {@link OrderedRealmCollection}.
```
cmelchior: Perhaps add a bit more what a snapshot actually means, i.e. element positions not updating and even an iterator example would probably be nice.

realm/realm-library/src/main/java/io/realm/OrderedRealmCollection.java
```diff
@@ -137,4 +137,14 @@
...
+     *
+     * @return the snapshot of this collection.
+     * @see OrderedRealmCollectionSnapshot
+     * @throws java.lang.IllegalStateException if the Realm is closed or the method is called from the wrong thread.
+     * @throws UnsupportedOperationException if the collection is unmanaged.
+     */
+    OrderedRealmCollectionSnapshot<E> createSnapshot();
```
cmelchior: Still debating with my myself whether `createSnapshot`, `getSnapshot` or `asSnapshot` is better. `createSnapshot` is probably fine though and also indicate that it is not a "free" method to call.

realm/realm-library/src/main/java/io/realm/OrderedRealmCollectionImpl.java
```diff
@@ -0,0 +1,604 @@
...
+import io.realm.internal.Collection;
+import io.realm.internal.InvalidRow;
+import io.realm.internal.RealmObjectProxy;
+import io.realm.internal.SortDescriptor;
+import io.realm.internal.Table;
+import io.realm.internal.UncheckedRow;
+
+/**
+ * General implementation for {@link OrderedRealmCollection} which is based on the {@link Collection}.
+ * Currently only {@link RealmResults} and {@link OrderedRealmCollectionSnapshot} is inherited from this class. But
```
cmelchior: is inherited from -> extends

realm/realm-library/src/main/java/io/realm/OrderedRealmCollectionImpl.java
```diff
@@ -0,0 +1,604 @@
...
+
+    /**
+     * {@inheritDoc}
+     */
+    public boolean isValid() {
+        return collection.isValid();
+    }
+
+    /**
+     * A {@link RealmResults} is always a managed collection.
```
cmelchior: This is not a RealmResults?

realm/realm-library/src/main/java/io/realm/OrderedRealmCollectionImpl.java
```diff
@@ -0,0 +1,604 @@
...
+     *
+     * @return {@code true}.
+     * @see RealmCollection#isManaged()
+     */
+    public boolean isManaged() {
+        return true;
+    }
+
+    /**
+     * Searches this {@link RealmResults} for the specified object.
```
cmelchior: Same

realm/realm-library/src/main/java/io/realm/OrderedRealmCollectionImpl.java
```diff
@@ -0,0 +1,604 @@
...
+            }
+        }
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void deleteFromRealm(int location) {
+        // TODO: Implement the deleteLast in OS level and do check there!
```
cmelchior: deleteLast?

realm/realm-library/src/main/java/io/realm/OrderedRealmCollectionImpl.java
```diff
@@ -0,0 +1,604 @@
...
+        collection.delete(location);
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public boolean deleteAllFromRealm() {
+        realm.checkIfValid();
+        if (size() > 0) {
```
cmelchior: Why this check? Deleting the empty list is still a success?

realm/realm-library/src/main/java/io/realm/OrderedRealmCollectionImpl.java
```diff
@@ -0,0 +1,604 @@
...
+
+    // Sorting
+
+    // aux. method used by sort methods
+    private long getColumnIndexForSort(String fieldName) {
+        if (fieldName == null || fieldName.isEmpty()) {
+            throw new IllegalArgumentException("Non-empty field name required.");
+        }
+        if (fieldName.contains(".")) {
+            throw new IllegalArgumentException("Sorting using child object fields is not supported: " + fieldName);
```
cmelchior: This limitation is being removed in another PR right?

realm/realm-library/src/main/java/io/realm/OrderedRealmCollectionImpl.java
```diff
@@ -0,0 +1,604 @@
...
+     * @throws UnsupportedOperationException
+     */
+    @Deprecated
+    @Override
+    public E remove(int index) {
+        throw new UnsupportedOperationException(NOT_SUPPORTED_MESSAGE);
+    }
+
+    /**
+     * Not supported by RealmResults.
```
cmelchior: This class might also be used by RealmList right?

realm/realm-library/src/main/java/io/realm/OrderedRealmCollectionImpl.java
```diff
@@ -0,0 +1,604 @@
...
+     * @throws UnsupportedOperationException
+     */
+    @Deprecated
+    @Override
+    public boolean remove(Object object) {
+        throw new UnsupportedOperationException(NOT_SUPPORTED_MESSAGE);
+    }
+
+    /**
+     * Not supported by RealmResults.
```
cmelchior: Same

realm/realm-library/src/main/java/io/realm/OrderedRealmCollectionImpl.java
```diff
@@ -0,0 +1,604 @@
...
+     *
+     * @throws UnsupportedOperationException
+     */
+    @Deprecated
+    @Override
+    public E set(int location, E object) {
+        throw new UnsupportedOperationException(NOT_SUPPORTED_MESSAGE);
+    }
+
```
cmelchior: Whitespace

realm/realm-library/src/main/java/io/realm/OrderedRealmCollectionSnapshot.java
```diff
@@ -0,0 +1,191 @@
...
+ * <p>
+ * {@link OrderedRealmCollectionSnapshot} is useful when making changes which may impact on the size or orders of the
```
cmelchior: remove "on"
orders -> order

realm/realm-library/src/main/java/io/realm/OrderedRealmCollectionSnapshot.java
```diff
@@ -0,0 +1,191 @@
...
+ * <pre>
+ * {@code
+ * final RealmResults<Dog>  dogs = realm.where(Dog.class).findAll();
+ * final OrderedRealmCollectionSnapshot<Dog> snapshot = dogs.createSnapshot();
+ * final int dogsCount = snapshot.size(); // dogs.size() == snapshot.size() == 10
+ * realm.executeTransaction(new Realm.Transaction() {
+ *     @Override
+ *     public void execute(Realm realm) {
+ *         for (int i = 0; i < dogsCount; i++) {
+ *         // This won't work since RealmResults is always up-to-day, its size gets decreased by 1 after every loop. An
```
cmelchior: day -> date

realm/realm-library/src/main/java/io/realm/OrderedRealmCollectionSnapshot.java
```diff
@@ -0,0 +1,191 @@
...
+     *
+     * @throws UnsupportedOperationException
+     */
+    @Override
+    public RealmResults<E> sort(String fieldName) {
+        throw getUnsupportedException("sort");
+    }
+
+    /**
+     * Not supported by {@link OrderedRealmCollectionSnapshot}. Use 'sort()' on the original
```
cmelchior: Why?

realm/realm-library/src/main/java/io/realm/OrderedRealmCollectionSnapshot.java
```diff
@@ -0,0 +1,191 @@
...
+     *
+     * @throws UnsupportedOperationException
+     */
+    @Override
+    public RealmResults<E> sort(String fieldName1, Sort sortOrder1, String fieldName2, Sort sortOrder2) {
+        throw getUnsupportedException("sort");
+    }
+
+    @Override
+    public RealmResults<E> sort(String[] fieldNames, Sort[] sortOrders) {
```
cmelchior: Mising Javadoc

CHANGELOG.md
dalinaum: "would be" -> "is"?

beeender:
@cmelchior queries and sorting on snapshot is meaningless, since if we want to support that, basically it will do query and sorting on the original TableView (the TableView has to be synced before query and sorting). IMO, it will confuse users. So, user should just do query and sorting on the original `RealmResults` instead. the purpose of snapthot ideally should be limited to only simple loops (maybe other use cases it is useful as well, but i cannot tell right now.).

realm/realm-library/src/main/java/io/realm/OrderedRealmCollectionImpl.java
beeender: yes, might be. `RealmList` may need to hold the `LinkView` as well as the `Collection`. The `add` and `remove` are still APIs from `LinkView`. But this class can enable the `RealmList` for APIs it doesn't have right now, for example, listeners related API.

realm/realm-library/src/androidTest/java/io/realm/ManagedRealmCollectionTests.java
```diff
@@ -500,6 +560,29 @@ public void minDate() {
         assertEquals(new Date(-YEAR_MILLIS * 20 * TEST_SIZE / 2), collection.minDate(AllJavaTypes.FIELD_DATE));
     }
 
+    // Delete the last row in the collection then test the aggregates methods.
```
dalinaum: `Deletes last row in the collection then tests the aggregates methods.`

realm/realm-library/src/main/java/io/realm/OrderedRealmCollectionImpl.java
dalinaum: extend? Since its subject is plural, `{@link RealmResults} and {@link OrderedRealmCollectionSnapshot}`.

realm/realm-library/src/main/java/io/realm/OrderedRealmCollectionImpl.java
```diff
@@ -0,0 +1,604 @@
...
+    /**
+     * Returns a distinct set of objects from a specific class. When multiple distinct fields are
+     * given, all unique combinations of values in the fields will be returned. In case of multiple
+     * matches, it is undefined which object is returned. Unless the result is sorted, then the
+     * first object will be returned.
+     *
+     * @param firstFieldName first field name to use when finding distinct objects.
+     * @param remainingFieldNames remaining field names when determining all unique combinations of field values.
+     * @return a non-null {@link RealmResults} containing the distinct objects.
+     * @throws IllegalArgumentException if field names is empty or {@code null}, does not exist,
```
dalinaum: if field names are empty?

realm/realm-library/src/main/java/io/realm/OrderedRealmCollectionSnapshot.java
```diff
@@ -0,0 +1,191 @@
+package io.realm;
+
+import io.realm.internal.Collection;
+import io.realm.internal.UncheckedRow;
+
+/**
+ * An {@link OrderedRealmCollectionSnapshot} is a special type of {@link OrderedRealmCollection}. It can be created by
+ * calling {@link OrderedRealmCollection#createSnapshot()}. Unlike {@link RealmResults} and {@link RealmList}, it's
```
dalinaum: it's size -> its size?

realm/realm-library/src/main/java/io/realm/OrderedRealmCollectionSnapshot.java
dalinaum: orders -> order?

realm/realm-library/src/main/java/io/realm/OrderedRealmCollectionSnapshot.java
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    @Override
+    public OrderedRealmCollectionSnapshot<E> createSnapshot() {
+        realm.checkIfValid();
+        return this;
+    }
+
+    /**
+     * Deletes the object at the given index from the Realm. The object at the given index becomes invalid. Just returns
```
dalinaum: The object at the given index will become invalid?

realm/realm-library/src/main/java/io/realm/OrderedRealmCollectionSnapshot.java
```diff
@@ -0,0 +1,191 @@
...
+    public void deleteFromRealm(int location) {
+        realm.checkIfValidAndInTransaction();
+        UncheckedRow row = collection.getUncheckedRow(location);
+        if (row.isAttached()) {
+            collection.delete(location);
+        }
+    }
+
+    /**
+     * Deletes the first object from the Realm. The first object becomes invalid.
```
dalinaum: The first object will become invalid?

Same above. I don't know whether `The first object becomes invalid.` is suitable or not.

realm/realm-library/src/main/java/io/realm/OrderedRealmCollectionSnapshot.java
```diff
@@ -0,0 +1,191 @@
...
+     * Deletes the last object from the Realm. The last object becomes invalid.
+     *
+     * @return {@code true} if an object was deleted, {@code false} otherwise.
+     * @throws java.lang.IllegalStateException if the Realm is closed or the method is called from the wrong thread.
+     */
+    @Override
+    public boolean deleteLastFromRealm() {
+        realm.checkIfValidAndInTransaction();
+        UncheckedRow row = collection.lastUncheckedRow();
+        return  row != null && row.isAttached() && collection.deleteLast();
```
dalinaum: double white spaces between `return` and `row`.

realm/realm-library/src/androidTest/java/io/realm/ManagedOrderedRealmCollectionTests.java
```diff
@@ -151,38 +150,60 @@ public void tearDown() {
...
-        throw new AssertionError("Unknown collection: " + collectionClass);
+        if (isSnapshot(collectionClass)) {
+            orderedCollection = orderedCollection.createSnapshot();
+        }
+        return orderedCollection;
     }
 
     @Test
     public void sort_twoFields() {
+        if (isSnapshot(collectionClass)) {
```
beeender: https://github.com/realm/realm-java/pull/4172#issuecomment-280559194

realm/realm-library/src/main/java/io/realm/OrderedRealmCollection.java
beeender: YEAH, it is not FREE! But actually the snapshot `TableView` is created through move constructor. So if user doesn't operate on the original Results until the next notification, it is free. Since the bg thread will handover a new `TableView` when next notification arrives.

realm/realm-library/src/main/java/io/realm/OrderedRealmCollectionImpl.java
beeender: This is the current behavior https://github.com/realm/realm-java/blob/e318dca510b807495550820643fa5c0938e13acd/realm/realm-library/src/main/java/io/realm/RealmResults.java#L304

realm/realm-library/src/main/java/io/realm/OrderedRealmCollectionImpl.java
beeender: This is for the aggregates used in the class. I changed the exception message in this PR and created an issue https://github.com/realm/realm-java/issues/4194

realm/realm-library/src/main/java/io/realm/OrderedRealmCollectionSnapshot.java
beeender: https://github.com/realm/realm-java/pull/4172#issuecomment-280559194

realm/realm-library/src/main/java/io/realm/OrderedRealmCollectionImpl.java
beeender: I didn't get it ... if the `remainingFieldNames` is empty, then it will call `public RealmResults<E> distinct(String fieldName) ` instead. right?

realm/realm-library/src/main/java/io/realm/OrderedRealmCollectionSnapshot.java
Zhuinden: I think `becomes` is fine here

realm/realm-library/src/main/java/io/realm/OrderedRealmCollection.java
beeender: Added javadoc for `OrderedRealmCollection`

realm/realm-library/src/main/java/io/realm/OrderedRealmCollectionImpl.java
```diff
@@ -237,7 +238,7 @@ private long getColumnIndexForSort(String fieldName) {
             throw new IllegalArgumentException("Non-empty field name required.");
         }
         if (fieldName.contains(".")) {
-            throw new IllegalArgumentException("Sorting using child object fields is not supported: " + fieldName);
+            throw new IllegalArgumentException("Aggregates on child object fields is not supported: " + fieldName);
```
Zhuinden: are not supported*?

realm/realm-library/src/main/java/io/realm/OrderedRealmCollection.java
```diff
@@ -23,6 +23,79 @@
...
+ *     s = dogs.size(); // This will be decreased by 1 every time after a dog is removed.
+ * }
+ * realm.commitTransaction();
+ * s = dogs.size(); // 0
+ * }
+ * </pre>
+ *
+ * You can delete the element while iterating without impacting the iterator. But whenever the element in the
+ * {@link RealmResults} changes, the {@link RealmResults} will be impacted. Like the above example, during the iterating
+ * after a element gets deleted, the {@link RealmResults}'s size become different. So does the order if modifications
```
kneth: "after a" -> "after an"

realm/realm-library/src/main/java/io/realm/OrderedRealmCollection.java
```diff
@@ -23,6 +23,79 @@
...
+ * applied to the element can impact the elements order of the {@link RealmResults}. {@link RealmList} has the same
+ * behaviour with {@link RealmResults} since they are both live collections.
+ *
+ * <p>
+ *
+ * Simple for-loop is different. See below example:
```
kneth: "Simple" -> "A simple"

realm/realm-library/src/main/java/io/realm/OrderedRealmCollectionImpl.java
cmelchior: Yeah, I remember where we get it from. The Collection API in Java use the boolean to indicate if a change happened, e.g. it should only return true if items where actually deleted.

realm/realm-library/src/main/java/io/realm/OrderedRealmCollection.java
```diff
@@ -23,6 +23,79 @@
  * element in the {@code OrderedRealmCollection} has an index. Each element can thus be accessed by its
  * index, with the first index being zero. Normally, {@code OrderedRealmCollection}s allow duplicate
  * elements, as compared to Sets, where elements have to be unique.
+ *
+ * <p>
+ *
+ * There are three types of {@link OrderedRealmCollection}. {@link RealmResults} and {@link RealmList} are live
+ * collections. They are up-to-date all the time and they will never contain a invalid {@link RealmObject}.
```
cmelchior: an invalid

realm/realm-library/src/main/java/io/realm/OrderedRealmCollection.java
```diff
@@ -23,6 +23,79 @@
...
+ * {@link OrderedRealmCollectionSnapshot} is different. An {@link OrderedRealmCollectionSnapshot} can be created from
+ * another {@link OrderedRealmCollection}. Its size and elements order stay the same as the original collection's when
+ * it was created. {@link OrderedRealmCollection} may contain invalid {@link RealmObject} if the object gets deleted.
```
cmelchior: OrderedRealmCollection -> OrderedRealmCollectionSnapshot 

{@link RealmObject} if the object gets deleted. -> `{@link RealmObject}'s if the objects gets deleted.`?

realm/realm-library/src/main/java/io/realm/OrderedRealmCollection.java
cmelchior: I'm confused now. Does this means that if I do

```
RealmResults<Person> persons = realm.where(Person.class).findAllSorted("age", Sort.ASC);
realm.beginTransaction();
for (Person p : persons) {
  p.setAge(persons.max("age") + 1);
}
realm.commitTransaction();
```
That I will keep moving items to the back of the list so they get processed twice?

beeender: no. above code just works. But if accessing the `persons` during the iterating is not stable. So the iterator will be always stable, but `RealmResults` is not.

cmelchior: I assume you mean that the individual objects will change, but their order in the collection will not? But `So does the order if modifications` specially say that the order changes?

PR #4181, opened 2017-02-14
Add example for showing how to store encryption key

beeender:


beeender:
Some more examples from @kotomisak 

https://github.com/kotomisak/db-showcase-android

More specifically:
- https://github.com/kotomisak/db-showcase-android/blob/develop/mobile/src/main/java/cz/koto/misak/dbshowcase/android/mobile/model/ModelProvider.java
- https://github.com/kotomisak/db-showcase-android/blob/develop/mobile/src/main/java/cz/koto/misak/dbshowcase/android/mobile/ui/control/ControlRootViewModel.java

Zhuinden:
Is this related to https://github.com/realm/realm-android-user-store ?

cmelchior:
Not really, although it uses the same concept. The idea is to show how to use a double layered key approach to store Realm encryption keys since the keystore does not allow you to extract the key material again, which is required by Realm.

So you 
1) Create a keystore and secure it using a fingerprint (because it's fun/easy to show, but you can use pin or whatever)

2) Then create a AES key outside the keystore and put the key inside the keystore. From the keystores point of view it will just see random bytes which it stores quite fine.

3) Show that on startup you can use the fingerprint to get the Realm key out of the keystore again and you can now use it to unlock the Realm.

This works quite nicely offline and _is_ secure _if_ you trust that the keystore does its job correctly.

PR #4206, opened 2017-02-20
improve performance of getters and setters in proxy classes

zaki50:
This change is a part of fixes of #3809.

Before this PR:

 ```
public class Model implements Realm Model {
    private String foo;
    public Model() {
        super();
        // 2: context is not set yet
        realmGet$foo(); // 3: this getter injects the context
    }
}

public class ModelRealmProxy extends Model {
    public ModelRealmProxy() {
        super(); // 1: execute the parent's constructor
        // 4: proxyState == null should be false since getter already injected the context.
        // If parent's constructor does not use any setters/getters at all, context is injected here.
        if (proxyState == null) {
            injectObjectContext();
        }
    }
}
```

After this PR:
```
public class Model implements Realm Model  {
    private String foo;
    public Model() {
        super();
        injectObjectContext(); //2: this line is injected by our transformer.
        realmGet$foo(); // 3: no need to call injectObjectContext(); in it
    }
}

public class ModelRealmProxy extends Model {
    public ModelRealmProxy() {
        super(); // 1: execute the parent's constructor
        //4: no need to inject the context. parent's constructor always calls injectObjectContext();
    }
}
```

@realm/java

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 
 ### Enhancements
 
-* Improved performance of getters and setters in proxy classes. As a side-effect of this improvement, now all proxy classes have a public method of `realm$injectObjectContext()`.
+* Improved performance of getters and setters in proxy classes.
 
 ## 2.3.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,12 +5,16 @@
 * Updated to Realm Sync v1.0.4.
 * Updated to Realm Core v2.3.1.
 
+### Enhancements
+
+* Improved performance of getters and setters in proxy classes. As a side-effect of this improvement, now all proxy classes have a public method of `realm$injectObjectContext()`.
 
 ## 2.3.1
 
 ### Enhancements
 
 * [ObjectServer] The `serverUrl` given to `SyncConfiguration.Builder()` is now more lenient and will also accept only paths as argument (#4144).
+* [ObjectServer] Add a timer to refresh periodically the access_token.
 
 ### Bug fixes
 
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

**Comment by zaki50** [comment]
Should this go to `master` only?

**Comment by beeender** [comment]
BTW, i think `releases` branch is the right place to go.

File: CHANGELOG.md [hunk 1]
```diff
@@ -5,12 +5,16 @@
 * Updated to Realm Sync v1.0.4.
 * Updated to Realm Core v2.3.1.
 
+### Enhancements
+
+* Improved performance of getters and setters in proxy classes. As a side-effect of this improvement, now all proxy classes have a public method of `realm$injectObjectContext()`.
```
cmelchior: No reason to name the public method I think? Proxy classes are not visible anyway?

realm-transformer/src/main/groovy/io/realm/transformer/BytecodeModifier.groovy
```diff
@@ -85,6 +85,13 @@ class BytecodeModifier {
         clazz.addInterface(proxyInterface)
     }
 
+    public static void callInjectObjectContextFromDefaultConstructor(CtClass clazz) {
+        def defaultConstructor = clazz.getDeclaredConstructor()
+        defaultConstructor.insertBeforeBody('if ($0 instanceof io.realm.internal.RealmObjectProxy) {' +
```
cmelchior: This doesn't match the code you posted in the description 😉 . Instead of the instance check would it make more sense to have the method being part of getter/setter interface we add to all model classes anyway? That way you could call `realm$injectObjectContext` directly since it would just be a no-op for unmanaged classes? 

EDIT: http://stackoverflow.com/questions/103564/the-performance-impact-of-using-instanceof-in-java I guess either approach is fine so probably no reason to change this.

CHANGELOG.md
zaki50: I'll remove that.

kneth:
Any indication of the performance gain?

zaki50:
@kneth not so big. This PR just remove null check from getters/setters.

kneth:
@zaki50 No worries. I agree, getters/setters must be as lean as possible.

PR #4210, opened 2017-02-21
Deprecate distinctXxx of RealmResults

beeender:
Those methods should only exist in the RealmResults. And maybe those
should be deprecated since all they are doing is distinct on the
where().
And annotate methods with @Override in the OrderedRealmCollectionImpl.

### DIFF
--- a/realm/realm-library/src/main/java/io/realm/RealmResults.java
+++ b/realm/realm-library/src/main/java/io/realm/RealmResults.java
@@ -198,4 +198,53 @@ public class RealmResults<E extends RealmModel> extends OrderedRealmCollectionIm
             throw new UnsupportedOperationException(realm.getClass() + " does not support RxJava.");
         }
     }
+
+    /**
+     * Returns a distinct set of objects of a specific class. If the result is sorted, the first
+     * object will be returned in case of multiple occurrences, otherwise it is undefined which
+     * object is returned.
+     *
+     * @param fieldName the field name.
+     * @return a new non-null {@link RealmResults} containing the distinct objects.
+     * @throws IllegalArgumentException if a field is null, does not exist, is an unsupported type,
+     * is not indexed, or points to linked fields.
+     */
+    public RealmResults<E> distinct(String fieldName) {
+        SortDescriptor distinctDescriptor = SortDescriptor.getInstanceForDistinct(collection.getTable(), fieldName);
+        Collection distinctCollection = collection.distinct(distinctDescriptor);
+        return createLoadedResults(distinctCollection);
+    }
+
+    /**
+     * Asynchronously returns a distinct set of objects of a specific class. If the result is
+     * sorted, the first object will be returned in case of multiple occurrences, otherwise it is
+     * undefined which object is returned.
+     *
+     * @param fieldName the field name.
+     * @return immediately a {@link RealmResults}. Users need to register a listener
+     * {@link io.realm.RealmResults#addChangeListener(RealmChangeListener)} to be notified when the
+     * query completes.
+     * @throws IllegalArgumentException if a field is null, does not exist, is an unsupported type,
+     * is not indexed, or points to linked fields.
+     */
+    public RealmResults<E> distinctAsync(String fieldName) {
+        realm.sharedRealm.capabilities.checkCanDeliverNotification(RealmQuery.ASYNC_QUERY_WRONG_THREAD_MESSAGE);
+        return where().distinctAsync(fieldName);
+    }
+
+    /**
+     * Returns a distinct set of objects from a specific class. When multiple distinct fields are
+     * given, all unique combinations of values in the fields will be returned. In case of multiple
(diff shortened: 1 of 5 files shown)


realm/realm-library/src/main/java/io/realm/RealmResults.java
```diff
@@ -198,4 +198,53 @@ public void removeChangeListeners() {
...
+     * first object will be returned.
+     *
+     * @param firstFieldName first field name to use when finding distinct objects.
+     * @param remainingFieldNames remaining field names when determining all unique combinations of field values.
+     * @return a non-null {@link RealmResults} containing the distinct objects.
+     * @throws IllegalArgumentException if field names is empty or {@code null}, does not exist,
+     * is an unsupported type, or points to a linked field.
+     */
+    public RealmResults<E> distinct(String firstFieldName, String... remainingFieldNames) {
+        return where().distinct(firstFieldName, remainingFieldNames);
```
cmelchior: The reason we are doing it on `where()` is because it is returning a copy, previously it was done in-place without the `where()`.

beeender:
I removed it from `OrderedRealmCollectionImpl` is because of i don't want the snapshot to support this method. That is the same reason with `Snapshot.where()`.

Actually this method doesn't do more that `RealmResults.where().distinct()`. So instead of adding those methods into the `RealmResults` and `RealmList`, I think it will be better just educate user to use `RealmResults.where()` instead. Also reduce the API methods count.

cmelchior:
Its a trade-off I think. Having one way of doing things is nice as it reduces methods and make it easier to document, but if getting to that functionality is hard you will annoy people anyway. 

We have different examples in the API:
1) `realm.allObjects(MyClass)` were removed in favor of `realm.where(MyClass).findAll();`
2) `realm.find(MyClass, Id)` is being added instead of using `realm.where(MyClass).equalTo("id", Id).findFirst()`
3) Adding `first()/last()` instead of `getIndex(0)/getIndex(size() - 1)`

In this case I would probably say that having distinct on `OrderedRealmCollection` and throwing for `OrderedRealmCollectionSnapshot` would be the most pragmatic since the snapshot isn't supposed to be widely used + having to force people throught `where()` feels really annoying. But it is probably good to have the discussion.

beeender:
Sure, but the question is "is the distinct important enough to waste method count?"

I would say compared with all the APIs listed above, distinct is quite a minor feature that doesn't worth to expose an API from `RealmResults`.

BTW, It makes me feel really uncomfortable if we have:
`RealmResults.distinct()` and `RealmResults.distinctAsync()`
but we don't have
`RealmResults.sortAsync()` although `RealmResults.sort()` is there.

zaki50:
The title of this PR should be updated?

realm/realm-library/src/main/java/io/realm/RealmResults.java
```diff
@@ -202,4 +203,30 @@ public void removeChangeListener(OrderedRealmCollectionChangeListener<RealmResul
             throw new UnsupportedOperationException(realm.getClass() + " does not support RxJava.");
         }
     }
+
+    /**
+     * @deprecated use {@link RealmQuery#distinct(String)} on the return value of {@link #where()} instead. This will
+     * be removed in coming 3.x.x minor releases.
+     */
+    public RealmResults<E> distinct(String fieldName) {
```
zaki50: We should have Deprecated annotation as well to align other deprecated methods.

realm/realm-library/src/main/java/io/realm/RealmResults.java
```diff
@@ -202,4 +203,30 @@ public void removeChangeListener(OrderedRealmCollectionChangeListener<RealmResul
...
+        SortDescriptor distinctDescriptor = SortDescriptor.getInstanceForDistinct(collection.getTable(), fieldName);
+        Collection distinctCollection = collection.distinct(distinctDescriptor);
+        return createLoadedResults(distinctCollection);
+    }
+
+    /**
+     * @deprecated use {@link RealmQuery#distinctAsync(String)} on the return value of {@link #where()} instead. This
+     * will be removed in coming 3.x.x minor releases.
+     */
+    public RealmResults<E> distinctAsync(String fieldName) {
```
zaki50: We should have Deprecated annotation as well to align other deprecated methods.

realm/realm-library/src/main/java/io/realm/RealmResults.java
```diff
@@ -202,4 +203,30 @@ public void removeChangeListener(OrderedRealmCollectionChangeListener<RealmResul
...
+        return where().distinctAsync(fieldName);
+    }
+
+    /**
+     * @deprecated use {@link RealmQuery#distinct(String, String...)} on the return value of {@link #where()} instead.
+     * This will be removed in coming 3.x.x minor releases.
+     */
+    public RealmResults<E> distinct(String firstFieldName, String... remainingFieldNames) {
```
zaki50: We should have Deprecated annotation as well to align other deprecated methods.

PR #4212, opened 2017-02-21
fix README format issue.

daimajia:


zaki50:
Thanks @daimajia 

To align the format, I think that it's enough to add empty line after line 182.
Can you confirm that?

daimajia:
@zaki50 It seems that won't work if just add an new line after line 182, the leading spaces must to be convert to `tab`.

PR #3900, opened 2016-12-13
Use ObjectStore SyncManager/SyncSession

cmelchior:
Still WIP

PR #4214, opened 2017-02-21
Using ObjectStore SyncManager & Session

nhachicha:
This PR integrates the ObjectStore SyncManager & Session, and get rid of the Java state machine/wrapper around the native OS session.

This is a continuation of the work started by @cmelchior in https://github.com/realm/realm-java/pull/3900

During the integration I found two bugs/limitation in the ObjectStore that I fixed temporarily in this [branch](https://github.com/realm/realm-object-store/tree/nh/sync_manager_java_integration), while waiting for a proper fix for: 

- https://github.com/realm/realm-object-store/pull/370
- https://github.com/realm/realm-object-store/issues/379

The remaining tasks are:
- [x] [Re]Implementing refresh token 
- [x] Testing the Client reset
- [x] Merging OS changes into `object-store/master`
- [x] Merging `java/releases` _behold conflict wrath_ 👹 
- [x] Fixing Tests


nhachicha:
@realm/java

realm/realm-library/src/main/cpp/objectserver_shared.hpp
```diff
@@ -19,123 +19,106 @@
...
-            }
+#ifdef REALM_ENABLE_SYNC
+        m_config = config;
+
+        // error handler will be called form the sync client thread
+        auto error_handler = [=](std::shared_ptr<SyncSession> session, SyncError error) {
+            realm::jni_util::Log::d("error_handler lambda invoked");
+
+            JNIEnv *env = realm::jni_util::JniUtils::get_env(true);
+            env->CallStaticVoidMethod(java_syncmanager,
```
kneth: What if a (Java) exception is thrown?

realm/realm-library/src/main/cpp/objectserver_shared.hpp
```diff
@@ -19,123 +19,106 @@
...
+                                      java_error_callback_method,
+                                      error.error_code.value(),
+                                      env->NewStringUTF(error.message.c_str()),
+                                      env->NewStringUTF(session.get()->path().c_str()));
```
kneth: Maybe `to_jstring(env, session.get()->path())`

realm/realm-library/src/main/cpp/objectserver_shared.hpp
```diff
@@ -19,123 +19,106 @@
...
-                        });
-                }
+        // path on disk of the Realm file.
+        // the sync configuration object.
+        // the session which should be bound.
+        auto bind_handler = [=](const std::string& path, const SyncConfig& syncConfig, std::shared_ptr<SyncSession> session) {
+            realm::jni_util::Log::d("Callback to Java requesting token for path");
+
+            JNIEnv *env = realm::jni_util::JniUtils::get_env(true);
+            jstring access_token_string = (jstring) env->CallStaticObjectMethod(java_syncmanager,
```
kneth: What if an Java exception is thrown?

realm/realm-library/src/main/cpp/objectserver_shared.hpp
```diff
@@ -19,123 +19,106 @@
...
             }
         };
-        m_sync_session->set_sync_transact_callback(sync_transact_callback);
-        m_sync_session->set_error_handler(std::move(error_handler));
+        // Get logged in user
+        JStringAccessor user_identity(env, sync_user_identity);
+        JStringAccessor realm_url(env, sync_realm_url);
+        std::shared_ptr<SyncUser> user = SyncManager::shared().get_existing_logged_in_user(user_identity);
+        if (!user)
+        {
```
kneth: Same line as `if`.

realm/realm-library/src/main/cpp/objectserver_shared.hpp
```diff
@@ -19,123 +19,106 @@
...
+            JStringAccessor realm_auth_url(env, sync_realm_auth_url);
+            JStringAccessor refresh_token(env, sync_refresh_token);
+            user = SyncManager::shared().get_user(user_identity, refresh_token, realm::util::Optional<std::string>(realm_auth_url));
+        }
+        SyncConfig sc = {user,realm_url,SyncSessionStopPolicy::Immediately,std::move(bind_handler),std::move(error_handler)};
```
kneth: Spaces after `,`

realm/realm-library/src/main/java/io/realm/internal/NativeObject.java
```diff
@@ -21,7 +21,7 @@
  * It specifies the operations common to all such objects.
  * All Java classes wrapping a core class should implement NativeObject.
  */
-interface NativeObject {
+public interface NativeObject {
```
kneth: Why making it public?

realm/realm-library/src/main/cpp/objectserver_shared.hpp
nhachicha: This will propagate the error to Java via `SyncSession.notifySessionError`. The default error handler is a no-op, however if the user specify a custom error handler that may throw, then the caller thread will crash causing probably the sync client thread to crash as well

realm/realm-library/src/main/cpp/objectserver_shared.hpp
nhachicha: 1- The ObjectStore sync session state will remain unchanged (ex: it will still be blocked at `WaitingForAccessToken`. This is ok as another thread may `revive`/`reuse` the session 

2- The `bind_handler` is usually triggered from the UI thread so this will crash the app.

realm/realm-library/src/main/java/io/realm/internal/NativeObject.java
nhachicha: _undskyld_ 👻

realm/realm-library/src/main/cpp/objectserver_shared.hpp
kneth: But you are not allowed to use any JNI function when a Java exception has been thrown.

nhachicha: if the `CallStaticObjectMethod` fails we will not invoke any further JNI method (not even `JStringAccessor`) since `access_token_string` is not valid & the lambda execution finishes.

cmelchior: Didn't we agree to catch all exceptions on the Java side and log it there as an Error in order to prevent this scenario. RxJava has similar restrictions. Callbacks are not allowed to throw.

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -172,10 +171,6 @@ private RealmCache(RealmConfiguration config) {
         @SuppressWarnings("unchecked")
         E realm = (E) refAndCount.localRealm.get();
 
-        // Notifies SyncPolicy that the Realm has been opened for the first time
-        if (refAndCount.globalCount == 1) {
```
cmelchior: Kinda sad to see this go, but I guess we need to implement it at the OS level when we want it back.

realm/realm-library/src/objectServer/java/io/realm/ErrorCode.java
```diff
@@ -105,9 +105,9 @@ public int intValue() {
      * <p>
      * FATAL: The session cannot be recovered and needs to be re-created. A likely cause is that the User does not
      * have access to this Realm. Check that the {@link SyncConfiguration} is correct. Any fatal error will cause
-     * the session to be become {@link SessionState#STOPPED}.
+     * the session to become SessionState#STOPPED.
```
cmelchior: SessionState is deleted?

realm/realm-library/src/objectServer/java/io/realm/ErrorCode.java
```diff
@@ -105,9 +105,9 @@ public int intValue() {
...
      * <p>
-     * RECOVERABLE: Temporary error. The session becomes {@link SessionState#UNBOUND}, but will automatically try to
+     * RECOVERABLE: Temporary error. The session becomes @link SessionState#UNBOUND, but will automatically try to
```
cmelchior: SessionState is deleted

realm/realm-library/src/objectServer/java/io/realm/SyncManager.java
```diff
@@ -84,7 +84,7 @@ public void onError(SyncSession session, ObjectServerError error) {
             }
         }
     };
-
+    private static Map<String, SyncSession> sessions = new HashMap<String, SyncSession>();
```
cmelchior: Add comment about what the string is.

realm/realm-library/src/objectServer/java/io/realm/SyncManager.java
```diff
@@ -173,6 +160,11 @@ public static void setDefaultSessionErrorHandler(SyncSession.ErrorHandler errorH
      * Gets any cached {@link SyncSession} for the given {@link SyncConfiguration} or create a new one if
      * no one exists.
      *
+     * Note: This will not create a new native (Object Store) session, this will only associate a Realm's path
```
cmelchior: This is a public method, so talking about implementation details here is probably not a good idea. It is still relevant info, so perhaps inline inside the method?

realm/realm-library/src/objectServer/java/io/realm/SyncManager.java
```diff
@@ -182,25 +174,30 @@ public static synchronized SyncSession getSession(SyncConfiguration syncConfigur
...
+
+        return session;
     }
 
-    public static AuthenticationServer getAuthServer() {
+    /**
+     * Remove the wrapped Java session.
+     * @param syncConfiguration configuration object for the synchronized Realm.
+     */
+    public static synchronized void removeSession(SyncConfiguration syncConfiguration) {
```
cmelchior: Do this need to be public? The Javadoc sounds very internal.

realm/realm-library/src/objectServer/java/io/realm/SyncManager.java
```diff
@@ -230,8 +227,57 @@ static void notifyUserLoggedOut(SyncUser user) {
         }
     }
 
+    // This is called from SyncManager.cpp from the worker thread the Sync Client is running on
+    // Right now Core doesn't send these errors to the proper session, so instead we need to notify all sessions
```
cmelchior: Some work has done in this regard. We probably need to ask @kspangsege about the current state.

realm/realm-library/src/objectServer/java/io/realm/SyncManager.java
```diff
@@ -230,8 +227,57 @@ static void notifyUserLoggedOut(SyncUser user) {
...
+     * the {@code access_token} expires.
+     *
+     * @throws IllegalStateException if the wrapped Java session is not found.
+     * @param sessionPath The path to the previously Java wraped session.
+     * @return a valid cached {@code access_token} if available or null.
+     */
+    private synchronized static String bindSessionWithConfig(String sessionPath) {
+        final SyncSession syncSession = sessions.get(sessionPath);
+        if (syncSession == null) {
+            throw new IllegalStateException("Matching Java SyncSession could not be found for: " + sessionPath);
```
cmelchior: Will this propagate correctly? I assume it is being called from the SyncClient thread?

realm/realm-library/src/objectServer/java/io/realm/SyncManager.java
```diff
@@ -230,8 +227,57 @@ static void notifyUserLoggedOut(SyncUser user) {
...
+        } else {
+            return syncSession.accessToken(authServer);
+        }
+    }
+
+    /**
+     * Resets the SyncManger and clear all existing users.
+     * This will also terminate all sessions.
+     *
+     * Only call this method when testing
```
cmelchior: Missing .

realm/realm-library/src/objectServer/java/io/realm/SyncSession.java
```diff
@@ -17,33 +17,52 @@
...
-import io.realm.internal.objectserver.ObjectServerSession;
 
 /**
  * This class represents the connection to the Realm Object Server for one {@link SyncConfiguration}.
  * <p>
- * A Session is created by either calling {@link SyncManager#getSession(SyncConfiguration)} or by opening
- * a Realm instance using that configuration. Once a session has been created, it will continue to exist until the app
- * is closed or the {@link SyncConfiguration} is no longer used.
+ * A Session is created by opening a Realm instance using that configuration. Once a session has been created,
+ * it will continue to exist until the app is closed or all Threads using this {@link SyncConfiguration} closes their respective {@link Realm}s.
```
cmelchior: Threads -> threads

realm/realm-library/src/objectServer/java/io/realm/SyncSession.java
```diff
@@ -71,29 +90,26 @@ public SyncUser getUser() {
...
-     *
-     * @return the current {@link SessionState} for this session.
-     */
-    public SessionState getState() {
-        return osSession.getState();
-    }
-
-    ObjectServerSession getOsSession() {
-        return osSession;
+    // Called from native code.
```
cmelchior: No longer true? This is being called from `SyncManager` ?

realm/realm-library/src/objectServer/java/io/realm/SyncManager.java
```diff
@@ -230,8 +227,57 @@ static void notifyUserLoggedOut(SyncUser user) {
...
+
+    /**
+     * All errors from native Sync is reported to this method. From the path we can determine which
+     * session to contact. If {@code path == null} all sessions are effected.
+     */
+    @SuppressWarnings("unused")
+    private static synchronized void notifyErrorHandler(int errorCode, String errorMessage, String path) {
+        for (SyncSession syncSession : sessions.values()) {
+            if (path == null || path.equals(syncSession.getConfiguration().getPath())) {
+                syncSession.notifySessionError(errorCode, errorMessage);
```
cmelchior: After discussions with @beeender, I believed we agreed that we need to catch all exceptions in callbacks from the OS and log them to logcat instead of throwing them, since it is close to impossible to make C++ terminate correctly for the Java session to actually terminate the app (instead of being swallowed in JNI causing other crashes). Also other libraries like RxJava have similar policies.

realm/realm-library/src/objectServer/java/io/realm/SyncSession.java
```diff
@@ -71,29 +90,26 @@ public SyncUser getUser() {
...
+        }
     }
 
-    @Override
-    protected void finalize() throws Throwable {
-        super.finalize();
-        if (osSession.getState() != SessionState.STOPPED) {
-            RealmLog.warn("Session was not closed before being finalized. This is a potential resource leak.");
-            osSession.stop();
+    // This is called from a synchronized block
```
cmelchior: You cannot guarantee that since the method is public.

realm/realm-library/src/objectServer/java/io/realm/SyncSession.java
```diff
@@ -114,5 +130,185 @@ protected void finalize() throws Throwable {
...
+        }
+        clearScheduledAccessTokenRefresh();
+
+        // Authenticate in a background thread. This allows incremental backoff and retries in a safe manner.
+        Future<?> task = SyncManager.NETWORK_POOL_EXECUTOR.submit(new ExponentialBackoffTask<AuthenticateResponse>() {
+            @Override
+            protected AuthenticateResponse execute() {
+                if (!isClosed && !Thread.currentThread().isInterrupted()) {
+                    return authServer.loginToRealm(
+                            getUser().getAccessToken(),//refresh token in fact
```
cmelchior: Yes, the mixed terminology is a pain :/

realm/realm-library/src/objectServer/java/io/realm/SyncSession.java
```diff
@@ -114,5 +130,185 @@ protected void finalize() throws Throwable {
...
+                            configuration.getPath(),
+                            configuration.shouldDeleteRealmOnLogout()
+                    );
+                    getUser().getSyncUser().addRealm(configuration.getServerUrl(), desc);
+                    // schedule a token refresh before it expires
+                    if (nativeRefreshAccessToken(configuration.getPath(), getUser().getSyncUser().getAccessToken(configuration.getServerUrl()).value(), configuration.getServerUrl().toString())) {
+                        scheduleRefreshAccessToken(authServer, response.getAccessToken().expiresMs());
+
+                    } else {
+                        // token not applied, no refresh will be scheduled
```
cmelchior: What does this mean exactly?

realm/realm-library/src/objectServer/java/io/realm/SyncSession.java
```diff
@@ -114,5 +130,185 @@ protected void finalize() throws Throwable {
...
+            // hit the expire date on the token
+            long refreshAfter =  expireDateInMs - System.currentTimeMillis() - REFRESH_MARGIN_DELAY;
+            if (refreshAfter < 0) {
+                // Token already expired
+                RealmLog.debug("Expires time already reached for the access token, refresh soon");
+                // we avoid refreshing directly to avoid an edge case where the client clock is ahead
+                // of the server, causing all access_token received from the server to be always
+                // expired, we will flood the server with refresh token requests then, so adding
+                // a bit of delay is the best effort in this case.
+                refreshAfter = REFRESH_MARGIN_DELAY;
```
cmelchior: Nitpick: 10 seconds seem like quite a while to wait there, or?

realm/realm-library/src/objectServer/java/io/realm/SyncSession.java
```diff
@@ -114,5 +130,185 @@ protected void finalize() throws Throwable {
...
+    }
+
+    private void scheduleRefreshAccessToken(final AuthenticationServer authServer, long expireDateInMs) {
+            // calculate the delay time before which we should refresh the access_token,
+            // we adjust to 10 second to proactively refresh the access_token before the session
+            // hit the expire date on the token
+            long refreshAfter =  expireDateInMs - System.currentTimeMillis() - REFRESH_MARGIN_DELAY;
+            if (refreshAfter < 0) {
+                // Token already expired
+                RealmLog.debug("Expires time already reached for the access token, refresh soon");
```
cmelchior: Perhaps `Refesh soon -> Refresh as soon as possible` ?

realm/realm-library/src/objectServer/java/io/realm/SyncSession.java
```diff
@@ -114,5 +130,185 @@ protected void finalize() throws Throwable {
...
+                    return authServer.refreshUser(getUser().getSyncUser().getUserToken(), configuration.getServerUrl(), getUser().getSyncUser().getAuthenticationUrl());
+                }
+                return null;
+            }
+
+            @Override
+            protected void onSuccess(AuthenticateResponse response) {
+                synchronized (SyncSession.this) {
+                    if (!isClosed && !Thread.currentThread().isInterrupted()) {
+                        RealmLog.debug("Access Token refreshed successfully");
```
cmelchior: Perhaps add the Realm Sync Url as well

realm/realm-library/src/objectServer/java/io/realm/SyncUser.java
```diff
@@ -462,6 +455,10 @@ ObjectServerUser getSyncUser() {
         return syncUser;
     }
 
+    public String getAuthURL () {
```
cmelchior: Isn't there already a `getAuthentificationUrl()`?

realm/realm-library/src/androidTestObjectServer/java/io/realm/SyncManagerTests.java
```diff
@@ -153,14 +139,14 @@ public void loggedOut(SyncUser user) {
...
-
     @Test
-    public void session() {
+    public void session() throws IOException {
         SyncUser user = createTestUser();
         String url = "realm://objectserver.realm.io/default";
         SyncConfiguration config = new SyncConfiguration.Builder(user, url)
                 .build();
-
+        // This will trigger the creation of the session
```
nhachicha: The ObjectStore creates automatically the session when you first open a Realm(SharedRealm) 

`SyncManager.getSession(config)` does only create the Java wrapper (if it doesn't exist) around the previously ObjectStore session.

Hence the need to call `Realm.getInstance(config);`

realm/realm-library/src/main/cpp/io_realm_SyncSession.cpp
```diff
@@ -0,0 +1,44 @@
...
+                                                   jstring accessToken, jstring sync_realm_url) {
+    TR_ENTER()
+    try {
+        JStringAccessor local_realm_path(env, localRealmPath);
+        std::shared_ptr<SyncSession> session = SyncManager::shared().get_existing_session(local_realm_path);
+        if (session) {
+            JStringAccessor access_token(env, accessToken);
+            JStringAccessor realm_url(env, sync_realm_url);
+            session->refresh_access_token(access_token, realm::util::Optional<std::string>(realm_url));
+            return true;
```
kneth: `true` -> `JNI_TRUE`

realm/realm-library/src/main/cpp/io_realm_SyncManager.cpp
```diff
@@ -24,42 +24,56 @@
...
 JNIEXPORT void JNICALL Java_io_realm_SyncManager_nativeInitializeSyncClient
-    (JNIEnv *env, jclass)
+    (JNIEnv* env, jclass)
 {
     TR_ENTER()
     if (sync_client) return;
 
     try {
+        // Setup SyncManager
+        g_binding_callback_thread_observer = &s_client_thread_listener;
```
nhachicha: yes it's a global used in the ObjectStore https://github.com/realm/realm-object-store/blob/14c2c7e7038850302f60f6fa1e36a22bceb3ab94/src/sync/impl/sync_client.hpp#L43

realm/realm-library/src/main/cpp/io_realm_SyncSession.cpp
```diff
@@ -0,0 +1,44 @@
...
+        }
+        else {
+            realm::jni_util::Log::d("nativeRefreshAccessToken no active/inactive session found");
+        }
+    } CATCH_STD()
+    return false;
```
kneth: `false` -> `JNI_FALSE`

realm/realm-library/src/main/cpp/io_realm_SyncSession.cpp
```diff
@@ -0,0 +1,44 @@
...
+        JStringAccessor local_realm_path(env, localRealmPath);
+        std::shared_ptr<SyncSession> session = SyncManager::shared().get_existing_session(local_realm_path);
+        if (session) {
+            JStringAccessor access_token(env, accessToken);
+            JStringAccessor realm_url(env, sync_realm_url);
+            session->refresh_access_token(access_token, realm::util::Optional<std::string>(realm_url));
+            return true;
+        }
+        else {
+            realm::jni_util::Log::d("nativeRefreshAccessToken no active/inactive session found");
```
nhachicha: nope!

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
```diff
@@ -79,9 +77,15 @@ Java_io_realm_internal_SharedRealm_nativeCreateConfig(JNIEnv *env, jclass, jstri
...
         config->automatic_change_notifications = auto_change_notification;
         if (sync_server_url) {
-            config->force_sync_history = true;
+            return reinterpret_cast<jlong>(new JniConfigWrapper(env,
+                                                                config,
+                                                                sync_server_url,
+                                                                sync_server_auth_url,
+                                                                sync_user_identity,
+                                                                sync_refresh_token));
+        } else {
```
kneth: `else {` on a separate line

Zhuinden: so it's `} else {` in Java but `}\n else {` in C++?

Ergh 😄

realm/realm-library/src/objectServer/java/io/realm/ErrorCode.java
nhachicha: yes, we don't track(do any logic) regarding the session's state, it's up to the ObjectStore to call the `binder_handler` if it needs an `access_token` or when reporting an error (via `error_handler`) 
I'll update the comment though to no hurt your memories 👻

realm/realm-library/src/objectServer/java/io/realm/SyncManager.java
nhachicha: unfortunately yes, this is called from `SyncObjectServerFacade#realmClosed` which is an internal package

realm/realm-library/src/objectServer/java/io/realm/SyncSession.java
nhachicha: we interface with the ObjectStore using the `realm_path`, so if for some use case (see below) the ObjectStore session is no longer found `nativeRefreshAccessToken` and we couldn't apply the `access_token` we just fetched, then stop the refreshing scheduling.

One use case, is the user open a Realm, we start refreshing, by the time the network query returns, the user have already closed the last instance of this Realm & the ObjectStore destroys the session, so when calling `get_session` in `nativeRefreshAccessToken` you'll get nothing.

realm/realm-library/src/objectServer/java/io/realm/SyncManager.java
kspangsege: Yes, this was fixed some time ago. See https://github.com/realm/realm-sync/blob/cfa258df3203cac5b9395769ab01764a730bfeb6/CHANGELOG.md#breaking-changes-18.

realm/realm-library/src/objectServer/java/io/realm/SyncSession.java
nhachicha: There's a use case where the user can end-up flooding the ROS with queries.

If the user/device clock is ahead compared to the server, then the token you just received from the server will be considered always as expired & will issue immediately another refresh request. 10 seconds feels like a safety net for me to avoid flooding, but I'm open to suggestions

nhachicha:
@kneth  @cmelchior updated

> Does the OS log state changes? That was pretty nice for debugging?

Unfortunately not, we might added it in a separate OS PR. 

> Are we confident that the number of unit tests we have cover this refactoring?

Most of the tests were done using RealmTasks, including the following use cases
- Upload/Download changeset
- Reusing session
- Access token expires
- Client reset
- SyncSession resume uploading when network is back (although after ~10min timeout from Sync client)
- Refresh Token 
- Error reporting

realm/realm-library/src/main/cpp/util.cpp
```diff
@@ -43,8 +43,11 @@ jclass java_lang_string;
 jmethodID java_lang_double_init;
 jclass java_util_date;
 jmethodID java_util_date_init;
-jclass session_class_ref;
-jmethodID session_error_handler;
+#ifdef REALM_ENABLE_SYNC
+jclass java_syncmanager;
+jmethodID java_bind_session_method;
```
beeender: Use `jni_util::JavaMethod` and remove those.

realm/realm-library/src/main/cpp/objectserver_shared.hpp
```diff
@@ -19,123 +19,105 @@
...
+        config->sync_config = std::make_shared<SyncConfig>(sc);
+#endif
+    }
+
+    inline Realm::Config* get_config() {
+        return m_config;
     }
 
-    inline realm::sync::Session* get_session() const noexcept
+
```
beeender: Two blank lines.

realm/realm-library/src/objectServer/java/io/realm/SyncSession.java
```diff
@@ -114,5 +129,185 @@ protected void finalize() throws Throwable {
...
+    private void clearScheduledAccessTokenRefresh() {
+        if (refreshTokenTask != null) {
+            refreshTokenTask.cancel();
+        }
+        if (refreshTokenNetworkRequest != null) {
+            refreshTokenNetworkRequest.cancel();
+        }
+    }
+
+    private native boolean nativeRefreshAccessToken(String path, String accessToken, String authURL);
```
beeender: should be static method.

realm/realm-library/src/objectServer/java/io/realm/SyncManager.java
beeender: Create an `io.realm.internal.objectserver.SessionManager` and move `sessions` there? I am quite uncomfortable to expose `removeSession()` to public.

realm/realm-library/src/objectServer/java/io/realm/SyncManager.java
```diff
@@ -230,8 +228,65 @@ static void notifyUserLoggedOut(SyncUser user) {
...
+     * the session specified by sessionPath.
+     *
+     * This will also schedule a timer to proactively refresh the {@code access_token} regularly, before
+     * the {@code access_token} expires.
+     *
+     * @throws IllegalStateException if the wrapped Java session is not found.
+     * @param sessionPath The path to the previously Java wraped session.
+     * @return a valid cached {@code access_token} if available or null.
+     */
+    private synchronized static String bindSessionWithConfig(String sessionPath) {
```
beeender: `@SuppressWarnings("unused")` this method.

realm/realm-library/src/objectServer/java/io/realm/SyncSession.java
beeender: I don't think this method should be public ...

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/AuthTests.java
```diff
@@ -100,6 +102,9 @@ public void onError(ObjectServerError error) {
     @Test
     @RunTestInLooperThread
     public void loginAsync_errorHandlerThrows() {
+        // set log level to info to make sure the IllegalArgumentException
+        // thrown in the test is visible in Logcat
+        RealmLog.setLevel(LogLevel.INFO);
```
beeender: The log level needs to be reset to default after this test done.

realm/realm-library/src/objectServer/java/io/realm/SyncSession.java
cmelchior: I don't have anything I could convincingly argue would be better, so this is probably fine.

realm/realm-library/src/main/cpp/io_realm_SyncSession.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+#include "io_realm_SyncSession.h"
+#include "objectserver_shared.hpp"
+
+using namespace std;
+using namespace realm;
+using namespace sync;
+
+JNIEXPORT jboolean JNICALL
+Java_io_realm_SyncSession_nativeRefreshAccessToken(JNIEnv *env, jobject, jstring localRealmPath,
```
nhachicha: good news `clang-format` caught this !

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
```diff
@@ -62,7 +59,8 @@ Java_io_realm_internal_SharedRealm_nativeInit(JNIEnv *env, jclass, jstring tempo
 JNIEXPORT jlong JNICALL
 Java_io_realm_internal_SharedRealm_nativeCreateConfig(JNIEnv *env, jclass, jstring realm_path, jbyteArray key,
         jbyte schema_mode, jboolean in_memory, jboolean cache, jlong /* schema_version */, jboolean disable_format_upgrade,
-        jboolean auto_change_notification, REALM_UNUSED jstring sync_server_url, jstring /*sync_user_token*/)
+        jboolean auto_change_notification, REALM_UNUSED jstring sync_server_url, REALM_UNUSED jstring sync_server_auth_url,
+                                                      REALM_UNUSED jstring sync_user_identity,REALM_UNUSED jstring sync_refresh_token)
```
nhachicha: will apply `clang-format` for the entire file

realm/realm-library/src/main/cpp/io_realm_internal_Util.cpp
```diff
@@ -53,6 +53,12 @@ JNIEXPORT jint JNICALL JNI_OnLoad(JavaVM* vm, void*)
         java_lang_double_init = env->GetMethodID(java_lang_double, "<init>", "(D)V");
         java_util_date        = GetClass(env, "java/util/Date");
         java_util_date_init   = env->GetMethodID(java_util_date, "<init>", "(J)V");
+#ifdef REALM_ENABLE_SYNC
```
beeender: aslo, this should be `#if REALM_ENABLE_SYNC` instead of `#ifdef`

nhachicha:
@beeender comments addressed, only point missing is concerning [SyncSession#close](https://github.com/realm/realm-java/pull/4214#discussion_r105090466) any suggestions are welcome.

CI is acting up (need to reattach Android device) tests are passing locally though.

realm/realm-library/src/objectServer/java/io/realm/SyncSession.java
cmelchior: I have a hunch we need to make the lifecyle methods of a Session available when we expose a `SyncPolicy` so if we are forced to do it now it wouldn't be that harmful, although less ideal.

realm/realm-library/src/main/cpp/io_realm_SyncSession.cpp
```diff
@@ -0,0 +1,51 @@
...
+                                                                              jstring sync_realm_url)
+{
+    TR_ENTER()
+    try {
+        JStringAccessor local_realm_path(env, localRealmPath);
+        auto session = SyncManager::shared().get_existing_session(local_realm_path);
+        if (session) {
+            JStringAccessor access_token(env, accessToken);
+            JStringAccessor realm_url(env, sync_realm_url);
+            session->refresh_access_token(access_token, realm::util::Optional<std::string>(realm_url));
```
bdash: I'm guessing you can write the second argument as simply `std::string(realm_url)`, which should then be implicitly convertible to `Optional<std::string>`.

realm/realm-library/src/main/cpp/jni_util/log.hpp
```diff
@@ -181,6 +181,10 @@ class CoreLoggerBridge : public realm::util::RootLogger {
     void do_log(Logger::Level, std::string msg) override;
     static CoreLoggerBridge& shared();
 
+    // Only used by the sync client to create it's own copy. In most other cases use
+    // CoreLoggerBridge::shared()
+    CoreLoggerBridge() = default;
```
beeender: `CoreLoggerBridge::shared()` is commented out and the constructor should be private :(

PR #4216, opened 2017-02-21
Enable listeners on RealmList

beeender:
The RealmList holds a Collection which is used for listeners.
Other RealmList APIs are still calling from LinkView.

- [x] Add tests for RealmList and RealmChangeListener.


beeender:
I is quite difficult for me to make all the `RealmResults` and `RealmList` listeners related test in one parameterized test because of the generics. So just add the new tests into `RealmListTests`.

realm/realm-library/src/androidTest/java/io/realm/RealmListTests.java
```diff
@@ -974,4 +978,113 @@ public void add_set_dynamicObjectCreatedFromTypedRealm() {
...
+        // This should trigger the listener.
+        realm.beginTransaction();
+        realm.cancelTransaction();
+        assertEquals(2, listenerCalledCount.get());
+        looperThread.testComplete();
+    }
+
+    @Test
+    @RunTestInLooperThread
+    public void removeALlChangeListeners() {
```
cmelchior: ALl -> All

realm/realm-library/src/main/java/io/realm/OrderedRealmCollectionImpl.java
```diff
@@ -15,8 +15,7 @@
 
 /**
  * General implementation for {@link OrderedRealmCollection} which is based on the {@code Collection}.
- * Currently only {@link RealmResults} and {@link OrderedRealmCollectionSnapshot} extend this class. But
- * {@link RealmList} could also extend this to share the same iterator implementation.
+ * Currently only {@link RealmResults} and {@link OrderedRealmCollectionSnapshot} extend this class.
```
cmelchior: Any value in this line at all?

realm/realm-library/src/main/java/io/realm/RealmList.java
```diff
@@ -838,6 +842,60 @@ public String toString() {
...
+        checkForAddRemoveListener(listener);
+        collection.removeListener(this, listener);
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void removeAllChangeListeners() {
+        realm.checkIfValid();
```
cmelchior: Would be nice if this could be incorporated into the `checkForAddRemoveListener` method? Perhaps `checkForAddRemoveListener(listener, checkForNull)` ?

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@
 
 * Added support for sorting by link's field (#672).
 * Added `OrderedRealmCollectionSnapshot` class and `OrderedRealmCollection.createSnapshot()` method. `OrderedRealmCollectionSnapshot` is useful when changing `RealmResults` or `RealmList` in simple loops.
+* Added support for adding listeners on `RealmList`.
```
cmelchior: What happens with the iterators?

beeender: `RealmList`'s iterators stay the same. They have quite different meanings compared with `RealmResults`. e.g.: `RealmList` is actually mutable for users, but `RealmResults` is not.

realm/realm-library/src/main/java/io/realm/OrderedRealmCollectionImpl.java
beeender: no. will remove.

realm/realm-library/src/main/java/io/realm/internal/Collection.java
```diff
@@ -348,6 +348,7 @@ public Collection(SharedRealm sharedRealm, LinkView linkView, SortDescriptor sor
         this.context = sharedRealm.context;
         this.table = linkView.getTable();
         this.context.addReference(this);
+        this.loaded = true;
```
zaki50: We should not initialize `loaded` field twice.
`private boolean loaded = false;` should be `private boolean loaded;` and assign in each constructor.

PR #4221, opened 2017-02-22
Link to debug core lib for debug build

beeender:
This failed before when linking. It seems to be fixed after removing
lto.


beeender:
CI will pass after core release https://github.com/realm/realm-core/pull/2466

kneth:
@beeender After upgrading to ROS 1.1.0 it did pass on my computer.

beeender:
It seems all java tests become very slow when using debug core which will cause CI timeout :(

beeender:
I hate gradle! 

So:
- Running tests with debug version of core is too slow, which easily takes 1+ hour.

I am trying:
1 . linking debug core with debug variant for local & CI
2. Run tests with release variant on CI
3. For local development, we can switch between debug & release in android studio

The problem is by setting `testCoverageEnabled = true` for release build will cause mocking fail which is not acceptable at al..

The idea is creating a new build type `citest` which links with release version of core and only for ci testing. But without further modification, the CI will build 3 types `debug`, `release` and `citest`. That means build the native code 3 times. 

I kind of give up for now, let me know if anyone has better ideas.

beeender:
`utf8Tests` takes 7+ minutes. By disabling it, the duration of CI might still be acceptable (< 1hour).

kneth:
@beeender Interesting observation.

realm/realm-library/src/androidTest/java/io/realm/RealmTests.java
```diff
@@ -781,6 +781,7 @@ public void realmConfiguration_fileName() {
     }
 
     @Test
+    @Ignore
```
kneth: Do we really want to disable it? Can we ignore it when using a custom core only?

beeender: By testing ObjectServer only for PR CI, i think 7+ minutes is acceptable?

beeender: What do you mean `using a custom core`? With this PR, the CI will always run testing with debug core.

kneth: Sorry I wasn't clear. You are disabling this test since it will take too long, right? Is there  a way where we can run it?

beeender: actually based on what the test want to check, we might be able to make it much faster, eg.: do not do queries, just write to the object then read and check.

beeender:
`waitForInitialData_resilientInCaseOfRetries` sometimes crashes, but it seems to be fixed with new sync release `1.10.0`.

beeender:
by testing only ObjectServer flour, the CI still takes 59 minutes, that is too long time :(

kneth:
@beeender One hour is not an option when checking PRs. If we run the tests on a x86 emulator on fast hardware, could we get the time down?

beeender:
@kneth definitely yes. Let me try it on CI.

beeender:
@realm/java 
Ideally we should run CI testing with debug enabled for core. But because of the assertions slow down the testing too much (1 hour for one flavour testing), we just add a opt-out gradle option to enable it when needed. So CI would just have the same speed as before by linking with release core.

But I suggest we enable it on our develop env all the time. Since a lot of issues will just assert earlier and all the debug symbols in core make debugging much easier for daily life.
just add below to your `~/.gradle/gradle.properties`
```
enableDebugCore=true
```

bmeike:
I am *so* with you on that.  I just *really* hate gradle.

That said, we are using it in a way it was not really intended to be used.  As dumb as it may seem gradle does not really understand how to compose independent projects.  Some of our pain arises from the fact we attempt to do exactly that.  We might, at some point, consider re-structuring our build (if that is even possible: I admit that I do not understand it)

... and, I wonder about @kneth 's idea.  It has bugged me for a while that CI thinks we have a single acceptance test.  Could we separate the test into pieces and throw hardware at it?

Zhuinden:
@bmeike Gradle is going to be much less painful once they switch language from Groovy to Kotlin.

realm/realm-library/src/main/cpp/CMakeLists.txt
```diff
@@ -62,23 +62,25 @@ create_javah(TARGET jni_headers
...
 
-# TODO: Ideally the debug build should link with core's debug build. But core dbg lib has
-# some compile options problems with arm, especially with macro REALM_DEBUG. Link to core
-# dbg for debug build when that gets solved.
-# We always link to the non-dbg version of core libs for now.
-# This means only JNI part has debugging symbols with debug build.
-# Debugging with core source code will also be done though anther approach -- compiling the core
-# with cmake inside android project.
+# Link to core/sync debug lib for debug build if it is debug build and linking with debug core is enabled.
+# FIXME: Ideally we should linking to debug core on CI for testing, but the assertions slow down the CI a lot.
```
cmelchior: Maybe TODO instead? Since we cannot really fix it right now?

PR #4226, opened 2017-02-23
Cast to int64_t when logging

beeender:
Otherwise is will be casted to bool when formatting.


zaki50:
@beeender We should fix this comment as well.
 https://github.com/realm/realm-java/blob/v2.3.1/realm/realm-library/src/main/cpp/util.hpp#L79

zaki50:
:+1:

kneth:
Did we really only use `VOID_PTR` twice?

beeender:
@kneth yeah!! removed them!

Issue #4128, opened 2017-01-31
Database corruption after SEGV_MAPERR


bios-seiji:
Hello,

We're experiencing high incidents of SIGSEGV SEGV_MAPERR crashes on recent versions of Realm. We can reliably reproduce these crashes by opening realms on multiple threads and conducting operations. Crashing multiple times (1-5) will usually cause the database to corrupt. Most of our tests have been with encrypted realms, however, I was able to reproduce the same results using an unencrypted realm with much effort.

#### Observations

- The probability of crash increases with the number of global realm instances open (N > 2 usually leads to crash).
- copyFromRealm increases probability of crashes
- Crashes usually occur during GC
- Crashes eventually lead to database corruption
- Corruption spreads, once a realm has corruption, doing operations on that realm will crash and eventually prevent the realm from opening entirely.
- Write operations on main thread increase crashes
- Encryption increases the probability but does not appear to be the cause

#### Results

This is a small set of crashes from production. I have hundreds, but they're much of the same.

```
signal 11 (SIGSEGV), code 1, fault addr 0x674db176 in tid 27384 (m.messenger.app)
Revision: '0'
ABI: 'arm'
pid: 27384, tid: 27384, name: m.messenger.app  >>> com.messenger.app <<<
signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x674db176
    r0 674db172  r1 b7989f78  r2 00000861  r3 0000004e
    r4 676e6172  r5 b7900f90  r6 00000068  r7 bef3c350
    r8 00000001  r9 bef3c37c  sl 9ec858c0  fp b7fd5e78
    ip b6cb05dc  sp bef3c348  lr b3b7152d  pc b3b5e6c2  cpsr 800e0030

Stack Trace:
  RELADDR   FUNCTION                                                                             FILE:LINE
  000846c2  realm::BpTreeNode::get_bptree_leaf(unsigned int) const+82                            unwind-c.c:?
  00085b0b  realm::BpTree<realm::util::Optional<long long> >::get(unsigned int) const+20         unwind-c.c:?
  0008da9f  realm::TimestampColumn::get(unsigned int) const+14                                   unwind-c.c:?
  000babe9  realm::TimestampNode<realm::Equal>::find_first_local(unsigned int, unsigned int)+64  unwind-c.c:?
  000c3215  realm::ParentNode::find_first(unsigned int, unsigned int)+44                         unwind-c.c:?
[...]

PR #4238, opened 2017-02-23
Upgrading to sync v1.2.1 and core v2.3.2

kneth:
@realm/java

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,11 +3,12 @@
 ### Bug fixes
 
 * Fixed log levels in JNI layer (#4204).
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -3,11 +3,12 @@
 ### Bug fixes
 
 * Fixed log levels in JNI layer (#4204).
+* Fixed a bug in encryption (#4128).
```
cmelchior: Should we write more? Like it causing apps in rare cases wh n using many threads?

kneth: In #4128 it was reported to crash with only 2 threads (it just took longer).

PR #4243, opened 2017-02-24
stabilize flaky test

zaki50:
@realm/java


zaki50:
@beeender listeners are registered after `assertTrue(dogs.load());`

```
         realm.beginTransaction();
         akamaru.setAge(17);
         realm.commitTransaction();
```
causes executing of both `RealmChangeListener<Realm>` and `RealmChangeListener<RealmResults<Dog>>` but sometimes `RealmChangeListener<Realm>` is executed first and sometimes `RealmChangeListener<RealmResults<Dog>>` is executed first.

And next run loop is sometimes too early to check the count.

PR #4249, opened 2017-02-24
Expose bug with default constructors

cmelchior:
Apparently the recent change to getters/setters (https://github.com/realm/realm-java/pull/4206)  caused a regression not caught by our unit tests, but it was revealed by our Kotlin example (we should really re-enable checking those again).

/cc @zaki50

### DIFF
--- a/realm/realm-library/src/androidTest/java/io/realm/entities/DefaultValueFromOtherConstructor.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/entities/DefaultValueFromOtherConstructor.java
@@ -1,0 +1,45 @@
+/*
+ * Copyright 2016 Realm Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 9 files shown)


cmelchior:
Note this is blocking our 2.3.2 release, so we need to either fix it fast or revert the offending change.

cmelchior:
Instead of only injecting the context in the default constructor I changed it to injecting it in every constructor.

zaki50:
@cmelchior Yes. We need to inject from all constructors.
I'll review this.

realm/realm-library/src/androidTest/java/io/realm/entities/DefaultValueFromOtherConstructor.java
```diff
@@ -0,0 +1,45 @@
+/*
+ * Copyright 2016 Realm Inc.
```
zaki50: 2017

realm/realm-library/src/androidTest/java/io/realm/entities/DefaultValueFromOtherConstructor.java
```diff
@@ -0,0 +1,45 @@
...
+
+package io.realm.entities;
+
+import java.util.Date;
+import java.util.UUID;
+
+import io.realm.RealmList;
+import io.realm.RealmObject;
+import io.realm.annotations.Ignore;
+import io.realm.annotations.PrimaryKey;
```
zaki50: `RealmList` and `PrimaryKey` is unused.

PR #4256, opened 2017-02-27
Fixing package name

kneth:
I think the package name is wrong. @nhachicha

### DIFF
--- a/examples/secureTokenAndroidKeyStore/src/main/AndroidManifest.xml
+++ b/examples/secureTokenAndroidKeyStore/src/main/AndroidManifest.xml
@@ -4,8 +4,8 @@
 
     <application android:allowBackup="true" android:icon="@mipmap/ic_launcher"
         android:label="@string/app_name" android:supportsRtl="true" android:theme="@style/AppTheme"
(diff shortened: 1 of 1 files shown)


examples/secureTokenAndroidKeyStore/src/main/AndroidManifest.xml
```diff
@@ -4,8 +4,8 @@
 
     <application android:allowBackup="true" android:icon="@mipmap/ic_launcher"
         android:label="@string/app_name" android:supportsRtl="true" android:theme="@style/AppTheme"
-        android:name="examples.io.realm.securetokenandroidkeystore.MyApplication">
-        <activity android:name="examples.io.realm.securetokenandroidkeystore.MainActivity">
+        android:name="io.realm.examples.securetokenandroidkeystore.MyApplication">
```
cmelchior: We probably need to restructure the java code as well

kneth: With this change, the example can be executed on a device from Android Studio.

cmelchior: Ups. Seems your right.

PR #4257, opened 2017-02-27
Add thread check to methods in RealmQuery.

### LINKED ISSUE #4228: Do thread checking in `RealmQuery`
Somehow most APIs in `RealmQuery` don't do thread checking, so user can:
`RealmQuery q1 = realm.where()` in thread1, then do `q2.equalTo()` in thread2.

zaki50:
Throw IllegalStateException instead of process crash when any of thread confined methods in RealmQuery is called from wrong thread.

fixes #4228 

@realm/java

### DIFF
--- a/realm/realm-library/src/androidTest/java/io/realm/RealmQueryTests.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/RealmQueryTests.java
@@ -165,7 +165,6 @@ public class RealmQueryTests {
         IN_STRING,
         IN_STRING_WITH_CASE,
         IN_BYTE,
-        //IN_BYTE_ARRAY,
         IN_SHORT,
         IN_INTEGER,
         IN_LONG,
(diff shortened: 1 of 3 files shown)


realm/realm-library/src/androidTest/java/io/realm/RealmQueryTests.java
```diff
@@ -147,6 +149,322 @@ private void populateNoPrimaryKeyNullTypesRows() {
...
+    public void callThreadConfinedMethodsFromWrongThread() throws Throwable {
+        final RealmQuery<AllJavaTypes> query = realm.where(AllJavaTypes.class);
+
+        final AtomicReference<Throwable> throwableFromThread = new AtomicReference<Throwable>();
+        final CountDownLatch testFinished = new CountDownLatch(1);
+
+        final String expectedMessage;
+        //noinspection TryWithIdenticalCatches
+        try {
+            final Field expectedMessageField = BaseRealm.class.getDeclaredField("INCORRECT_THREAD_MESSAGE");
```
cmelchior: Clever

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -1022,11 +1151,15 @@ public boolean isValid() {
...
      */
     public RealmQuery<E> lessThan(String fieldName, Date value) {
+        realm.checkIfValid();
+
         long columnIndices[] = schema.getColumnIndices(fieldName, RealmFieldType.DATE);
         this.query.lessThan(columnIndices, value);
         return this;
     }
 
+    // Less Than Or Equal To
```
cmelchior: Nitpick: Not a big fan of these kind of sections since they are kinda impossible to maintain and nothing detects if methods are moved outside them

realm/realm-library/src/androidTest/java/io/realm/RealmQueryTests.java
```diff
@@ -147,6 +149,322 @@ private void populateNoPrimaryKeyNullTypesRows() {
...
+        EQUAL_TO_LONG,
+        EQUAL_TO_DOUBLE,
+        EQUAL_TO_FLOAT,
+        EQUAL_TO_BOOLEAN,
+        EQUAL_TO_DATE,
+
+        IN_STRING,
+        IN_STRING_WITH_CASE,
+        IN_BYTE,
+        //IN_BYTE_ARRAY,
```
nhachicha: aren't we sure about this one? similar comments for the below commented enum

realm/realm-library/src/androidTest/java/io/realm/RealmQueryTests.java
```diff
@@ -147,6 +149,322 @@ private void populateNoPrimaryKeyNullTypesRows() {
...
+        }
+
+        final Thread thread = new Thread("callThreadConfinedMethodsFromWrongThread") {
+            @Override
+            public void run() {
+                try {
+                    for (ThreadConfinedMethods method : ThreadConfinedMethods.values()) {
+                        try {
+                            callThreadConfinedMethod(query, method);
+                            fail("IllegalStateException must be thrown.");
```
nhachicha: are we sure `fail` on this Thread will be propagated to the test thread?

realm/realm-library/src/androidTest/java/io/realm/RealmQueryTests.java
```diff
@@ -147,6 +149,322 @@ private void populateNoPrimaryKeyNullTypesRows() {
...
+                        } catch (Throwable e) {
+                            if (e instanceof IllegalStateException && expectedMessage.equals(e.getMessage())) {
+                                // expected exception
+                                continue;
+                            }
+                            throwableFromThread.set(e);
+                            return;
```
nhachicha: shouldn't you call `testFinished.countDown();`?

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -191,7 +191,7 @@ private RealmQuery(BaseRealm realm, LinkView linkView, String className) {
      * @return {@code true} if still valid to use, {@code false} otherwise.
      */
     public boolean isValid() {
-        if (realm == null || realm.isClosed()) {
+        if (realm == null || realm.isClosed() /* this includes thread checking */) {
```
nhachicha: I don't understand the comment `this includes thread checking`, `realm.isClosed` doesn't check for thread confinement.

zaki50: `Realm.isClosed()` checks thread confinement.

realm/realm-library/src/main/java/io/realm/RealmQuery.java
zaki50: Should I remove all existing section headers?

realm/realm-library/src/androidTest/java/io/realm/RealmQueryTests.java
zaki50: yes.

`fail()` throws `AssertionError` and then caught by following `} catch (Throwable e) {`.

Finally that error is re-thrown at
```
        final Throwable throwable = throwableFromThread.get();
        if (throwable != null) {
            throw throwable;
        }
```

realm/realm-library/src/androidTest/java/io/realm/RealmQueryTests.java
zaki50: following `finally` block is doing that.

realm/realm-library/src/androidTest/java/io/realm/RealmQueryTests.java
zaki50: OK, I'll removed them.

realm/realm-library/src/main/java/io/realm/RealmQuery.java
kneth: Yes. I think they are from a time where the IDEs/editors didn't have a function to show all the methods in a separate pane.

zaki50:
@cmelchior @nhachicha all comments are addressed.

realm/realm-library/src/main/java/io/realm/RealmQuery.java
zaki50: removed.

PR #4273, opened 2017-03-02
update release dates in CHANGELOG.md

zaki50:
This PR includes #1636

@realm/java

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,14 @@
+## 3.0.1 (YYYY-MM-DD)
+
+### Enhancements
+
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,14 @@
+## 3.0.1 (YYYY-MM-DD)
+
+### Enhancements
```
cmelchior: Good idea, then we make sure the order is always the same and we can always delete the empty sections when doing final edits.

zaki50: Yes, that's what I intended!

PR #4280, opened 2017-03-02
Formatting C++ source code using clang-format using core's rules.

kneth:
@realm/java 

To do: 

* [x] Add description on how to install and run clang-format from Android Studio
* [x] Curly braces around single lines (`if` and `for`)


realm/realm-library/src/main/cpp/.clang-format
```diff
@@ -0,0 +1,89 @@
+---
```
cmelchior: Can we reference the one in cores repo instead? Then we don't have to maintain a copy here?

realm/realm-library/src/main/cpp/io_realm_SyncManager.cpp
```diff
@@ -41,34 +41,35 @@ using namespace realm::jni_util;
...
 std::unique_ptr<Client> sync_client;
 
-JNIEXPORT void JNICALL Java_io_realm_SyncManager_nativeInitializeSyncClient
-    (JNIEnv *env, jclass)
+JNIEXPORT void JNICALL Java_io_realm_SyncManager_nativeInitializeSyncClient(JNIEnv* env, jclass)
 {
     TR_ENTER()
-    if (sync_client) return;
+    if (sync_client)
+        return;
```
zaki50: I think we should add curly brace around single line block if we give it independent line.
@realm/java What do you think?

kneth: I absolutely agree. And I think we should have curly braces always - even for single lines.

realm/realm-library/src/main/cpp/.clang-format
kneth: Not sure how easy it will be.

realm/realm-library/src/main/cpp/io_realm_SyncManager.cpp
cmelchior: @zaki50 Agree 👍

realm/realm-library/src/main/cpp/.clang-format
cmelchior: Ok

realm/realm-library/src/main/cpp/io_realm_SyncManager.cpp
diegomontoya: Style issue for core: https://github.com/realm/realm-core/issues/2294. Library and core should have the same fixed/uniform style.

cmelchior:
@kneth Did you see https://github.com/realm/realm-core/issues/2294 .. While I personally agree that our bracer style is "better", being uniform is probably more important?

kneth:
@cmelchior Yes, and I have added `{}` to all places where it was missing.

realm/realm-library/src/main/cpp/io_realm_internal_Table.cpp
```diff
@@ -1235,15 +1305,17 @@ JNIEXPORT jlong JNICALL Java_io_realm_internal_Table_nativeFindFirstNull(JNIEnv*
                                                                          jlong columnIndex)
 {
     Table* pTable = TBL(nativeTablePtr);
-    if (!TBL_AND_COL_INDEX_VALID(env, pTable, columnIndex))
-        return jlong(-1);
-    if (!TBL_AND_COL_NULLABLE(env, pTable, columnIndex))
-        return jlong(-1);
+    if (!TBL_AND_COL_INDEX_VALID(env, pTable, columnIndex)) {
+        return to_jlong_or_not_found(-1);
```
beeender: `to_jlong_or_not_found` is a confusing name to me...
what about remove the `to_jlong_or_not_found` and use:
`reinteprete_cast<jlong>(realm::not_found)` here?

kneth:
Closing - see #4307

PR #4292, opened 2017-03-06
Fail Realm.migrateRealm() if a SyncConfiguration is used.

### LINKED ISSUE #4075: calling Realm#migration with a SyncConf should throw
Sync Realm perform automatic migration, we shouldn't mislead the user to call [migrateRealm](https://github.com/realm/realm-java/blob/master/realm/realm-library/src/main/java/io/realm/BaseRealm.java#L608) with a specified migration block, this should throw `UnsupportedOperationException` instead of silently ignoring the user migration block

cmelchior:
Fixes #4075 

I used `IllegalArgumentException` instead of `UnsupportedOperationException`, since the later should only really be used if the method fails for all input IMO.

### DIFF
--- a/realm/realm-library/src/androidTestObjectServer/java/io/realm/SyncedRealmMigrationTests.java
+++ b/realm/realm-library/src/androidTestObjectServer/java/io/realm/SyncedRealmMigrationTests.java
@@ -43,23 +43,6 @@ public class SyncedRealmMigrationTests {
 
     @Rule
     public final TestSyncConfigurationFactory configFactory = new TestSyncConfigurationFactory();
-    @Rule
-    public final ExpectedException thrown = ExpectedException.none();
-
-    private Realm realm;
-    private Context context;
-
-    @Before
-    public void setup() {
-        context = InstrumentationRegistry.getInstrumentation().getContext();
-    }
-
-    @After
-    public void tearDown() {
-        if (realm != null) {
-            realm.close();
-        }
-    }
 
     @Test
     public void migrateRealm_syncConfigurationThrows() {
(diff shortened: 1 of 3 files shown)


realm/realm-library/src/androidTestObjectServer/java/io/realm/SyncedRealmMigrationTests.java
```diff
@@ -0,0 +1,76 @@
...
+ */
+@RunWith(AndroidJUnit4.class)
+public class SyncedRealmMigrationTests {
+
+    @Rule
+    public final TestSyncConfigurationFactory configFactory = new TestSyncConfigurationFactory();
+    @Rule
+    public final ExpectedException thrown = ExpectedException.none();
+
+    private Realm realm;
```
nhachicha: Not used

realm/realm-library/src/androidTestObjectServer/java/io/realm/SyncedRealmMigrationTests.java
```diff
@@ -0,0 +1,76 @@
...
+    private Context context;
```
nhachicha: Not used

realm/realm-library/src/androidTestObjectServer/java/io/realm/SyncedRealmMigrationTests.java
```diff
@@ -0,0 +1,76 @@
...
+
+    @Before
+    public void setup() {
```
nhachicha: `setup` include unused code, could be removed

realm/realm-library/src/androidTestObjectServer/java/io/realm/SyncedRealmMigrationTests.java
```diff
@@ -0,0 +1,76 @@
...
+        context = InstrumentationRegistry.getInstrumentation().getContext();
+    }
+
+    @After
+    public void tearDown() {
```
nhachicha: `tearDown ` include unused code, could be removed

PR #4295, opened 2017-03-07
Throw IllegalArgumentException if mixing TLS and non-TLS connections

### LINKED ISSUE #3710: Enforce TLS if it was used for authentication
Right now it is possible to e.g login to ROS using `https` while using `realm` as the sync protocol.
While technical possible to use `https` for auth and `realm` for sync, it is counter productive and against security best practises.

We should enforce people using `realms` if using `https` by throwing an `IllegalArgumentException` if it happens.

cmelchior:
Fixes #3710

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### Enhancements
 
 * Now `targetSdkVersion` is 25.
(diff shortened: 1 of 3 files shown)


cmelchior:
Closing. This decision should be up to sysadmins, not the app SDK.

PR #4313, opened 2017-03-13
Correctly report Client Reset

### LINKED ISSUE #4080: Sync- User should be notified about client reset
currently error code related to the sync client reset are managed directly in JNI
https://github.com/realm/realm-java/blob/master/realm/realm-library/src/main/cpp/objectserver_shared.hpp#L87-L110

the current implementation will schedule a Realm backup/reset when the app starts again, we need also to return an error code so the user have a chance to stop the app.

cmelchior:
Fixes #4080 

TODO
- [x] Unit tests
- [x] Cleanup Javadoc
- [x] Implement native code for Testing error
- [x] Implement native code for executing the Client Reset manually

------

Previously Client Reset was hidden among other error codes and not very visible. This PR adds explicit support for Client Reset similar to Cocoa. See #4080 for alternatives to the API in this PR.

I had a discussion with @nhachicha, and we both agreed that breaking the API (in a small way) by adding a new error callback method was acceptable. Client Reset is such a destructive event that forcing devs to handle it explicitly is probably a good idea, and we get rid of all the ugly casts in https://github.com/realm/realm-java/issues/4080#issuecomment-285331612

I'm not entirely sure about the name / methods for the error class yet, and have been switching between `ClientResetError`, `ClientResetException` and `ClientResetRequiredError`. I settled on the latter as it seems more descriptive and fits with the `executeClientReset` method, but interested in feedback on naming.

Also, if anyone feels strongly against breaking the `ErrorHandler` interface, please suggest an alternative.

Ready for review @realm/java


realm/realm-library/src/main/cpp/io_realm_SyncManager.cpp
```diff
@@ -96,3 +96,23 @@ JNIEXPORT void JNICALL Java_io_realm_SyncManager_nativeConfigureMetaDataSystem(J
...
+JNIEXPORT void JNICALL Java_io_realm_SyncManager_nativeSimulateSyncError(JNIEnv* env, jclass, jstring localRealmPath,
+                                                                         jint errorCode, jstring errorMessage,
+                                                                         jboolean isFatal)
+{
+    TR_ENTER()
+    try {
+        JStringAccessor local_realm_path(env, localRealmPath);
+        JStringAccessor error_message(env, errorMessage);
+
+        std::shared_ptr<SyncSession> session = SyncManager::shared().get_existing_active_session(local_realm_path);
```
beeender: good place for using `auto` -> `auto session = SyncManager::shared().get_existing_active_session(local_realm_path);`

realm/realm-library/src/objectServer/java/io/realm/ClientResetRequiredError.java
```diff
@@ -0,0 +1,87 @@
...
+     * will be deleted.
+     *
+     * @return a reference to the location of the original Realm file. After Client Reset has been executed this file
+     *         will no longer exists. Use {@code file.exists()} to check this.
+     */
+    public File getOriginalFile() {
+        return originalFile;
+    }
+
+
```
beeender: two blank lines

realm/realm-library/src/objectServer/java/io/realm/ClientResetRequiredError.java
```diff
@@ -0,0 +1,87 @@
...
+        this.configuration = configuration;
+        this.backupFile = new File(backupFilePath);
+        this.originalFile = new File(configuration.getPath());
+    }
+
+    /**
+     * Calling this method will execute the Client Reset manually instead of waiting until next app restart. This will
+     * only be possible if all instances of that Realm has been closed, otherwise a {@link IllegalStateException} will
+     * be thrown.
+     *
```
beeender: <p> here?

realm/realm-library/src/objectServer/java/io/realm/ClientResetRequiredError.java
```diff
@@ -0,0 +1,87 @@
...
+     * After this method returns, the backup file can be found in the location returned by {@link #getBackupFile()}.
+     * The file at {@link #getOriginalFile()} have been deleted, but will be recreated from scratch next time a
```
beeender: `have been` -> `is`

realm/realm-library/src/objectServer/java/io/realm/ClientResetRequiredError.java
```diff
@@ -0,0 +1,87 @@
...
+ * @see io.realm.SyncSession.ErrorHandler#onClientReset(SyncSession, ClientResetRequiredError) for more information
+ *      about when and why Client Reset occur and how to deal with it.
+ */
+public class ClientResetRequiredError extends ObjectServerError {
+
+    private final RealmConfiguration configuration;
+    private final File backupFile;
+    private final File originalFile;
+
+    public ClientResetRequiredError(ErrorCode errorCode, String errorMessage, String backupFilePath, RealmConfiguration configuration) {
```
beeender: Why this is public?

realm/realm-library/src/main/cpp/io_realm_ClientResetRequiredError.cpp
```diff
@@ -0,0 +1,41 @@
...
+JNIEXPORT void JNICALL Java_io_realm_ClientResetRequiredError_nativeExecuteClientReset(JNIEnv* env, jobject,
+                                                                               jstring localRealmPath)
+{
+    TR_ENTER()
+    try {
+        JStringAccessor local_realm_path(env, localRealmPath);
+        if (!SyncManager::shared().immediately_run_file_actions(std::string(local_realm_path))) {
+            ThrowException(
+                env, IllegalState,
+                concat_stringdata("Realm was not configured correctly. Client Reset could not be run for Realm at: ",
```
beeender: `concat_stringdata()` this function is strange. I think we should remove it, and use `format.hpp` from object store instead. But minor issue. I can do it in another PR.

realm/realm-library/src/androidTestObjectServer/java/io/realm/SessionTests.java
```diff
@@ -52,4 +60,76 @@ public void get_syncValues() {
...
+                .errorHandler(new SyncSession.ErrorHandler() {
+                    @Override
+                    public void onError(SyncSession session, ObjectServerError error) {
+                        fail("Wrong error " + error.toString());
+                    }
+
+                    @Override
+                    public void onClientReset(SyncSession session, ClientResetRequiredError error) {
+                        try {
+                            error.executeClientReset();
```
cmelchior: Yes, this is why I'm concerned about the naming of things. Perhaps the method should be `onClientResetRequired` instead, to make it more clear that when this method is called it means the Client should do a Reset, but one hasn't actually been done yet.

realm/realm-library/src/objectServer/java/io/realm/ClientResetRequiredError.java
cmelchior: Mistake. Will fix.

realm/realm-library/src/androidTestObjectServer/java/io/realm/SessionTests.java
beeender: `onClientResetRequired` is better for me to understand.

realm/realm-library/src/androidTestObjectServer/java/io/realm/SessionTests.java
```diff
@@ -52,4 +60,76 @@ public void get_syncValues() {
...
+
+        Realm realm = Realm.getInstance(config);
+        looperThread.testRealms.add(realm);
+
+        // Trigger error
+        SyncManager.simulateClientReset(SyncManager.getSession(config));
+    }
+
+    // Check that we can manually execute the Client Reset.
+    @Test
```
cmelchior: Good catch

realm/realm-library/src/main/cpp/io_realm_SyncManager.cpp
cmelchior: I don't mind changing it, but I'm not sure I fully understand why, especially since we later do `*session` which would only work if this was a pointer type?

beeender: it's just for making the line shorter. The below two lines are same for compiler:
`auto session = SyncManager::shared().get_existing_active_session(local_realm_path);`
`std::shared_ptr<SyncSession> session = SyncManager::shared().get_existing_active_session(local_realm_path);`

realm/realm-library/src/main/cpp/io_realm_ClientResetRequiredError.cpp
```diff
@@ -0,0 +1,41 @@
...
+                                  local_realm_path));
+            return;
```
kneth: Why this `return`? Nothing happens after this point.

realm/realm-library/src/main/cpp/io_realm_SyncManager.cpp
```diff
@@ -96,3 +96,23 @@ JNIEXPORT void JNICALL Java_io_realm_SyncManager_nativeConfigureMetaDataSystem(J
...
+        if (!session) {
+            ThrowException(env, IllegalArgument, concat_stringdata("Session not found: ", local_realm_path));
+            return;
+        }
+        std::error_code code = std::error_code{static_cast<int>(errorCode), realm::sync::protocol_error_category()};
+        SyncSession::OnlyForTesting::handle_error(*session, {code, std::string(error_message), isFatal == JNI_TRUE});
```
kneth: `to_bool(isFatal)` maybe?

realm/realm-library/src/objectServer/java/io/realm/ClientResetRequiredError.java
```diff
@@ -0,0 +1,87 @@
...
+
+package io.realm;
+
+import java.io.File;
+
+/**
+ * Class encapsulating information needed for handling a Client Reset event.
+ *
+ * @see io.realm.SyncSession.ErrorHandler#onClientReset(SyncSession, ClientResetRequiredError) for more information
+ *      about when and why Client Reset occur and how to deal with it.
```
kneth: "occur" -> "occurs"

realm/realm-library/src/objectServer/java/io/realm/ClientResetRequiredError.java
```diff
@@ -0,0 +1,87 @@
...
+    public ClientResetRequiredError(ErrorCode errorCode, String errorMessage, String backupFilePath, RealmConfiguration configuration) {
+        super(errorCode, errorMessage);
+        this.configuration = configuration;
+        this.backupFile = new File(backupFilePath);
+        this.originalFile = new File(configuration.getPath());
+    }
+
+    /**
+     * Calling this method will execute the Client Reset manually instead of waiting until next app restart. This will
+     * only be possible if all instances of that Realm has been closed, otherwise a {@link IllegalStateException} will
```
kneth: "has" -> "have"

realm/realm-library/src/objectServer/java/io/realm/SyncSession.java
```diff
@@ -127,6 +135,46 @@ void close() {
...
+         *     <li>
+         *         If Client Reset isn't executed manually, it will automatically be carried out the next time all
+         *         Realm instances have been closed and re-opened. This will most likely be
+         *         when the app is restarted.
+         *     </li>
+         * </ol>
+         *
+         * <b>WARNING:</b>
+         * Any writes to the Realm file between this callback and Client Reset has been executed, will not be
+         * synchronized to the Object Server. Those changes will only be present in the backed up file. It is therefor
```
kneth: "therefor" -> "therefore"

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
```diff
@@ -85,10 +85,19 @@ class JniConfigWrapper {
...
 
-            JNIEnv* env = realm::jni_util::JniUtils::get_env(true);
+            auto error_message = error.message;
+            auto error_code = error.error_code.value();
+            if (error.is_client_reset_requested()) {
+                // Hack the error message to send information about the location of the backup.
+                // If more uses of the user_info map surfaces. Refactor this to send the full
+                // map instead.
+                error_message = error.user_info[SyncError::c_recovery_file_path_key];
+                error_code = 7; // See ErrorCode.java
```
kneth: Won't the Java enum be recreated as C `#define`s?

realm/realm-library/src/main/cpp/io_realm_ClientResetRequiredError.cpp
cmelchior: `format.hpp` doesn't work with a `JStringAccessor` right now, so I'll prefer to refactor this in another PR?

realm/realm-library/src/main/cpp/io_realm_SyncManager.cpp
cmelchior: I know, but perhaps I just like actually having the type information, because then I don't have to think when reading it :) So some sort of guideline when to use `auto` vs. not would be nice?

Anyway, I'll change it.

realm/realm-library/src/objectServer/java/io/realm/ClientResetRequiredError.java
cmelchior: I think `have been` is correct here since I'm talking about something that already happened

realm/realm-library/src/main/cpp/io_realm_ClientResetRequiredError.cpp
beeender: `JStringAccessor` can be implicitly convert to `std::string`. But never mind, i can do it in another PR.

realm/realm-library/src/main/cpp/io_realm_ClientResetRequiredError.cpp
cmelchior: Mostly to just follow the pattern we use everywhere else. Also prevents bugs in case of a refactor, but I'll be happy to remove it if you think it really isn't needed?

realm/realm-library/src/main/cpp/io_realm_SyncManager.cpp
cmelchior: Missed we had that one 👍

realm/realm-library/src/main/cpp/io_realm_SyncManager.cpp
beeender: I have been asked to use `auto` in the similar case when modify core or object-store (i cannot remember ) ... anyway, this is not documented. only thing about `auto` is https://github.com/realm/realm-core/blob/master/doc/development/coding_style_guide.cpp#L189

realm/realm-library/src/objectServer/java/io/realm/ClientResetRequiredError.java
beeender: then `has been` please

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
cmelchior: I'm not sure it is done out of the box, but otherwise we can probably make it happen. It would be much safer than this at least

cmelchior: Actually, I'll revert that statement. They are not re-created and even if we can make it happen, I'm pretty sure we cannot get to the value `7` since that is a class variable of the enum, not the ordinal.

cmelchior:

PR #4315, opened 2017-03-13
Restoring Sync logging 

nhachicha:
- Sync Client is no longer created from the Java binding, but from the OS
- Registering a log factory so the OS can create a logger to attach to the sync client


nhachicha:
@realm/java

realm/realm-library/src/main/cpp/io_realm_SyncManager.cpp
```diff
@@ -58,24 +53,14 @@ struct AndroidClientListener : public realm::BindingCallbackThreadObserver {
...
-        // Create SyncClient
-        sync::Client::Config config;
-        config.logger = &CoreLoggerBridge::shared();
-        sync_client = std::make_unique<Client>(std::move(config)); // Throws
+struct AndroidSyncLoggerFactory : public realm::SyncLoggerFactory {
+    virtual std::unique_ptr<util::Logger> make_logger(Logger::Level level) override
+    {
+        auto logger = std::make_unique<CoreLoggerBridge>();
+        logger->set_level_threshold(level);
+        return std::move(logger);
```
beeender: no need to call `std::move` here, compiler will optimize the code to avoid constructors here. It will just return a rvalue. See http://stackoverflow.com/questions/17473753/c11-return-value-optimization-or-move

realm/realm-library/src/main/cpp/jni_util/log.hpp
```diff
@@ -184,7 +184,6 @@ class CoreLoggerBridge : public realm::util::RootLogger {
     static CoreLoggerBridge& shared();
```
beeender: Because of `SyncLoggerFactory`, singleton is not suitable for this class anymore. Just remove `shared()`. Actually we can also separate tags for core and sync. But it is not relevant to this PR. I can do it later.

realm/realm-library/src/main/cpp/io_realm_SyncManager.cpp
```diff
@@ -58,24 +53,14 @@ struct AndroidClientListener : public realm::BindingCallbackThreadObserver {
...
-    try {
-        // Setup SyncManager
-        g_binding_callback_thread_observer = &s_client_thread_listener;
-
-        // Create SyncClient
-        sync::Client::Config config;
-        config.logger = &CoreLoggerBridge::shared();
-        sync_client = std::make_unique<Client>(std::move(config)); // Throws
+struct AndroidSyncLoggerFactory : public realm::SyncLoggerFactory {
+    virtual std::unique_ptr<util::Logger> make_logger(Logger::Level level) override
```
beeender: virtual is redundant when the function is marked override

realm/realm-library/src/main/cpp/io_realm_SyncManager.cpp
```diff
@@ -93,6 +78,12 @@ JNIEXPORT void JNICALL Java_io_realm_SyncManager_nativeConfigureMetaDataSystem(J
     try {
         JStringAccessor base_file_path(env, baseFile); // throws
         SyncManager::shared().configure_file_system(base_file_path, SyncManager::MetadataMode::NoEncryption);
+
+        // Register Sync Client thread start/stop callback
+        g_binding_callback_thread_observer = &s_client_thread_listener;
```
cmelchior: It looks a bit weird that this setup is in a method named `configureMetaDataSystem`. Perhaps rename this method to `initializeSync` or `initializeSyncManager` ?

realm/realm-library/src/main/cpp/io_realm_SyncManager.cpp
beeender: my bad. This is actually correct since the `unique_ptr` needs to be converted to ``std::unique_ptr<util::Logger>

beeender:
@realm/java PR updated.

realm/realm-library/src/main/cpp/io_realm_SyncManager.cpp
```diff
@@ -59,24 +46,15 @@ struct AndroidClientListener : public realm::BindingCallbackThreadObserver {
...
-        g_binding_callback_thread_observer = &s_client_thread_listener;
-
-        // Create SyncClient
-        sync::Client::Config config;
-        config.logger = &CoreLoggerBridge::shared();
-        sync_client = std::make_unique<Client>(std::move(config)); // Throws
+struct AndroidSyncLoggerFactory : public realm::SyncLoggerFactory {
+    std::unique_ptr<util::Logger> make_logger(Logger::Level level) override
+    {
+        auto logger = std::make_unique<CoreLoggerBridge>(std::string("REALM_SYNC"));
```
cmelchior: I was trying to figure out how `RealmLog.setLevel()` works for these factory created loggers. From what I can tell, changing the LogLevel after they are created doesn't work currently?

realm/realm-library/src/objectServer/java/io/realm/SyncManager.java
```diff
@@ -303,9 +300,7 @@ static void simulateClientReset(SyncSession session) {
                 true);
     }
 
-    private static native void nativeInitializeSyncClient();
-    // init and load the Metadata Realm containing SyncUsers
-    protected static native void nativeConfigureMetaDataSystem(String baseFile);
+    protected static native void nativeInitializeSyncManager(String baseFile);
```
cmelchior: baseFile -> `syncBaseDir/syncRootDir`?

realm/realm-library/src/main/cpp/io_realm_SyncManager.cpp
beeender: i will fix this in #4337

PR #4319, opened 2017-03-14
Cancel transaction explicitly

zaki50:
If `Realm.getInstance()` is called in the 'transaction' object, calling `close()` does not cancel current transaction. To ensure that the transaction is not left open, cancel it explicitly if needed.

@realm/java

### DIFF
--- a/realm/realm-library/src/main/java/io/realm/Realm.java
+++ b/realm/realm-library/src/main/java/io/realm/Realm.java
@@ -1440,8 +1440,13 @@ public class Realm extends BaseRealm {
                 } catch (final Throwable e) {
                     exception[0] = e;
                 } finally {
-                    // SharedGroup::close() will cancel the transaction if needed.
-                    bgRealm.close();
+                    try {
+                        if (isInTransaction()) {
+                            bgRealm.cancelTransaction();
+                        }
+                    } finally {
+                        bgRealm.close();
+                    }
                 }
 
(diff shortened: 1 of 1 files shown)


beeender:
transaction will be cancelled when destroy core `SharedGroup`. Or did i miss anything here？

zaki50:
@beeender I think users can call `Realm.getInstance()` in the transaction ( https://github.com/realm/realm-java/blob/v3.0.0/realm/realm-library/src/main/java/io/realm/Realm.java#L1430 ) and it can prevent for `close()` to destroy `SharedGroup`.

beeender:
But then there will be leak which is even worse than a transaction not cancelled exception since it would be difficult to be found.

I am thinking about disallowing user to get instance in the async transaction block from `RealmCache`. What do you think?

cmelchior:
Since we provide the Realm instance used in the transaction block, it would make sense to disallow getting it yourself, but detecting that sounds like tracking some rather nasty state. Is that worth it?

I lean slightly towards no (but I'm happy to be convinced otherwise), in which case this fix looks good to me.

zaki50:
I agree with @cmelchior 
There are no reason to allow for user to obtain another Realm instance in the transaction, but prohibiting it requires a cost to every `getInstance()` call.
I think just cancelling the transaction is enough.

What do you think? @beeender

Issue #4101, opened 2017-01-24
Single Object Notifications


cmelchior:
**Motivation**
https://github.com/realm/realm-object-store/pull/324 added support for single object notifications to the Object Store

https://github.com/realm/realm-cocoa/pull/4537 is adding the public API's on the Cocoa side.

We should add similar capabilities to Android. 

**Use cases**
1) Object change listeners will now only be called if the object actually changed. It should reduce the number of unnecessary updates which is happening right now due to the change notification being table based.

2) It should make it possible to register listeners for a single field value, e.g. if the UI should only be refreshed if a single field changed. 
 
3) It should be possible to register a listener on an object and be give the old and new values, e.g. useful if you only need to refresh the UI if a String changed in a particular way.

3) It should now be possible to be notified when an object is deleted. See https://github.com/realm/realm-java/issues/3138. Realm classes only implementing "RealmModel"  might be tricky to support.

4) We should strive for easy integration with Databinding https://developer.android.com/topic/libraries/data-binding/index.html. How this should be done is still undecided.

This issue consists of the following tasks:

- [ ] Design the public API for this.
    * It is not possible to register listeners on linked objects (verify this).
    * Changes to linked objects and lists cannot be reported except "they changed" (verify this).
    * Figure out how to best parse the map of new and old values for primitive values.
    * Should ideally support Databinding out of the box or at least make integration very easy.
    * Should work for both classes that `extends RealmObject` and `implements RealmModel`.
- [x] It is unclear at this point if we depend on #3834 being merged first. This needs to be clarified 
- [ ] Implement API based on the Object Store implementation
- [ ] Javadoc
- [ ] Unit tests 
- [ ] Website documentation

beeender:
This has been almost implemented with #3834 but gets reverted to make PR smaller. When start this, check 4079726324ead77d730109b2e458c234c66e1feb fist. It is the commit removed the `RowNotifier`.

zaki50:
Does this API provide an ability to get notification when a linked object is changed?

cmelchior:
I don't believe so. If I remember correctly that is the default behaviour on .NET and the rationale being that if you care about changes to a linked object you should register the listener there instead. 

That argument makes sense, although I can see supporting it would probably lead to a more natural syntax in code:

```
// Version A: Listeners on each object
Person p = getPerson();
p.addFieldChangeListener("name", new RealmFieldChangeListener() {
  public void onChange(Person p, ChangeSet change) {
    updateUI(p);
[...]

PR #4331, opened 2017-03-16
Add detailed notification for RealmObject

beeender:
- Add RowObject for wrapping Object Store's Object class.
- Enabled detailed object notification.
- Add RealmObjectChangeListener, ObjectChangeSet.
- Fix false positive change notifications for RealmObject. Now the
  notification will only be sent when the object actually changes.

See #4101


beeender:
For use cases in #4101 

1) Object change listeners will now only be called if the object actually changed. It should reduce the number of unnecessary updates which is happening right now due to the change notification being table based.

2) It should make it possible to register listeners for a single field value, e.g. if the UI should only be refreshed if a single field changed. 
 
3) It should be possible to register a listener on an object and be give the old and new values, e.g. useful if you only need to refresh the UI if a String changed in a particular way.

4) It should now be possible to be notified when an object is deleted. See https://github.com/realm/realm-java/issues/3138. Realm classes only implementing "RealmModel"  might be tricky to support.

5) We should strive for easy integration with Databinding https://developer.android.com/topic/libraries/data-binding/index.html. How this should be done is still undecided.

Currently this will solve 1) 4) need more investigation on 5)

2) I don't say a strong requirement right now, so we can discuss it when needed in the future. Also based on current API, it is easy to get a 3rd party lib to subscribe on field changes.

3) Yes, cocoa has it. My major concern is that would require to store the old value in memory and convert to java values. Problems will happen when there is some huge strings/binary existing, then the memory consumption + encoding performance are something really should be considered. Also, generic type for field value might also be a problem.
I don't see there is case that old value is important, but if we need to implement it, based on current API, maybe we can have something like:
```java
    class FieldChange<T> {
        /**
         * The name of the field which changed.
         */
        final String fieldName;
        final T oldValue;
        final T value;

        public FieldChange(String fieldName, T oldValue, T value) {
            this.fieldName = fieldName;
            this.oldValue = oldValue;
            this.value = value;
        }
    }
```

beeender:
@realm/java I am still working on this to fix some corner issues. But the API part is ready for review. Please take a look.

CHANGELOG.md
```diff
@@ -15,6 +15,8 @@
   * Migration for linking objects is not yet supported.
   * Backlink verification is incomplete.  Evil code can cause native crashes.
 * [ObjectServer] In case of a Client Reset, information about the location of the backed up Realm file is now reported through the `ErrorHandler` interface (#4080).
+* The listener on `RealmObject` will only be triggered if the object changes (#3894).
+* Added `RealmObjectChangeListener` to get detailed information about `RealmObejct` changes.
```
kneth: "RealmObejct" -> "RealmObject"

cmelchior: RealmObjct -> RealmObject

realm/realm-library/src/main/java/io/realm/ObjectChangeSet.java
```diff
@@ -0,0 +1,49 @@
...
+ * limitations under the License.
+ */
+
+package io.realm;
+
+/**
+ * Information about the changes made to an object.
+ * @see RealmObject#addChangeListener(RealmObjectChangeListener) .
+ */
+public interface ObjectChangeSet {
```
cmelchior: Mirroring the `CollectionChangeSet` seems like the correct approach

realm/realm-library/src/main/java/io/realm/ObjectChangeSet.java
```diff
@@ -0,0 +1,49 @@
...
+
+        public FieldChange(String fieldName) {
+            this.fieldName = fieldName;
+        }
+    }
+
+    /**
+     * @return {@code null} if the object has been deleted. Otherwise returns the information about changed fields.
+     */
+    FieldChange[] getFieldChanges();
```
cmelchior: This wouldn't very well if you had e.g an `int` and a `String` field, as this would basically be type-erased to `FieldChange<Object>`.

Figuring out how to represent the changes are probably the crux of the problem, since the rest of the API's fall fairly natural from what we did for Collection notifications. Also, we probably need to consider what is most likely if interested in multiple changes. Do you register 1 listener and do switches or register multiple listeners?

What about this, it adds a few more helper methods people can use, that could provide a little more type safety. 

```
public interface ObjectChangeset {
  boolean isDeleted();
  // I'm not sure this is very useful in practical code. I cannot think of any generic code
  // where you wouldn't need the type information somehow?
  FieldChange[] getChanges(); 
  String[] getFieldNames;
  boolean didChange(String fieldName);
  <E> FieldChange<E> getChange(E type, String fieldName);

   /// Alternatively
   FieldChange<Long> getLongChange(String fieldName);
   FieldChange<String> getStringChange(String fieldName);
   FieldChange<Double> getDoubleChange(String fieldName);
   FieldChange<Float> getFloatChange(String fieldName);
   FieldChange<Boolean> getBooleanChange(String fieldName);
   FieldChange<Date> getDateChange(String fieldName);
   <E> FieldChange<E> getRealmObjectChange(E type, String fieldName);
   <E> FieldChange<RealmList<E>> getRealmListChange(E type, String fieldName);
}
```
Generally, I think we should try to make it as easy as possible for people to get the typed information since that is most likely what you want to operate in client code.

realm/realm-library/src/main/java/io/realm/RealmObjectChangeListener.java
```diff
@@ -0,0 +1,53 @@
...
+     * When this gets called to return the results of an asynchronous query made by {@link RealmQuery#findFirstAsync()},
+     * {@code changeSet} will be {@code null}.
+     * <p>
+     * When this gets called when the object is deleted, {@code changeSet.isDeleted()} will return {@code true} and
+     * {@code changeSet.getFieldChanges()} will return {@code null}.
+     * <p>
+     * When this gets called when the object is modified, {@code changeSet.isDeleted()} will return {@code false} and
+     * {@code changeSet.getFieldChanges()} will return the detailed information about the fields' changes.
+     * <p>
+     * This will be called if the {@link RealmObject} field gets set or removed, but it won't be called if the
```
cmelchior: Perhaps rewriting this would make it a bit more clear:

````
If a field points to another RealmObject this listener will only be triggered if the field is set to a new object or null. Updating the referenced RealmObject will not trigger this listener. 

If a field points to a RealmList, this listener will only be triggered if a objects are inserted, removed or moved within the List. Updating the objects in the RealmList will no trigger this listener.
```

realm/realm-library/src/main/java/io/realm/internal/RowObject.java
```diff
@@ -0,0 +1,117 @@
+package io.realm.internal;
+
+import io.realm.ObjectChangeSet;
+import io.realm.RealmObjectChangeListener;
+
+/**
+ * Java wrapper for Object Store's {@code Object} class. Currently it is only used for object notifications.
+ */
+public class RowObject implements NativeObject {
```
kneth: Is `RowObject` really the best name? I know that `RealmObject` is taken but that's what it is.

realm/realm-library/src/main/java/io/realm/ObjectChangeSet.java
Zhuinden: > <E> FieldChange<E> getChange(E type, String fieldName);

`E` as type, how? You'd be able to use `int.class` and the like, but that falls apart with `RealmList<E>` which requires the ugly hack `new TypeToken<RealmList<E>>() {};`

cmelchior: If we want to use generics we are forced to use the boxed types, e.g `Integer` instead of `int`. Alternatively you are forced to create a `FieldChange` class for each type we support, which also seems like massive overkill: `IntegerFieldChange`, `intFieldChange` ....

realm/realm-library/src/androidTest/java/io/realm/RealmObjectTests.java
```diff
@@ -1593,7 +1590,7 @@ public void run() throws Exception {
 
                 //noinspection TryFinallyCanBeTryWithResources
                 try {
-                    dog.addChangeListener(null);
+                    dog.addChangeListener((RealmChangeListener)null);
```
dalinaum: `dog.addChangeListener((RealmChangeListener) null);`

realm/realm-library/src/main/java/io/realm/internal/RowObject.java
beeender: I am really out of ideas ... First i tried `Object`, but apparently it will conflict a lot with java's `Object` ... But `ObjectStoreObject` sounds really stupid... any suggestions are welcome.

cmelchior: Just throwing ideas at the wall: `OsObject`, `NativeRealmObject`, `InternalObject`, `OsRealmObject` ... nothing is really great, so `RowObject` is probably fine for now.

But perhaps we should find a common prefix for all the ObjectStore classes?

beeender: common prefix makes sense. So `OsObject`? and `Collection` -> `OsResults`?

cmelchior: Yes and `OsSharedRealm` I assume?

realm/realm-library/src/main/java/io/realm/ObjectChangeSet.java
beeender: 1. I prefer one listener on the `RealmObject` instead of multiple listeners on the object's fields. It makes the API simpler and it is still quite simple for real use cases, eg.: databinding. We/Users can always inherit the `RealmObjectChangeListener` for specific use cases. See my example code of using it in data binding https://github.com/realm/realm-android-adapters/pull/99/files#diff-58a0e399a032fe72057d0e65269ebdb6

2. I would choose something with less methods, so probably:

```java
public interface ObjectChangeset {
  boolean isDeleted();
  String[] getChangedField();
 
  public static class FieldChange<T> {
    T oldValue;
    T newValue;
  }

  // This will always return null for RealmList field.
  <E> FieldChange<E> getChange(E type, String fieldName);
}
```

3. `FieldChange` and `getChange()` will not be implemented in this PR. Also if we are going to have them, there are some restrictions:

a) For local transactions, there is no way to get the old values.
b) store the old values will be slow and consuming memory when the old value is a big String or Binary -- The listener is on the `RealmObject` instead of on the field.

cmelchior: Having `isDeleted()`, `getChangedFields()`, `isChanged(String fieldName)` and `getChange(type, fieldName)` probably covers the use cases I can think of 👍 

It seems a rather big (and arbitrary) limitation that this doesn't work for local transactions. Any reason why? 

IMO we should just document that you need to be careful if using this with objects with _big_  field values. You shouldn't register 100's of these anyway, but doing it on an object with a 16MB image is probably a bad idea (if it changes). Cocoa has a feature request for ignoring certain fields I believe, but this is currently out of scope.

realm/realm-library/src/main/java/io/realm/ObjectChangeSet.java
```diff
@@ -0,0 +1,36 @@
...
+    /**
+     * @return true if the object has been deleted from the Realm.
+     */
+    boolean isDeleted();
+
+    /**
+     * @return the names of changed fields if the object still exists and there are field changes. Returns {@code null}
+     * if the object has been deleted.
+     */
+    String[] getChangedFields();
```
beeender: @cmelchior I am not quite convinced that `boolean isFieldChanged(String fieldName)` is useful here:
- If user wants, it is very easy to do with `Arrays.asList(getChangedFields()).contains(fieldName)`
- It would be ambiguous when the object is delete, what should it return in that case? `false`?

But I am totally fine if you strongly think it is needed. 🍏

beeender:
@realm/java This is ready for review.
@bmeike I removed some tests from backlink test set because of those notification on backlinked object should not be triggered when the parent changes. They passed before simply because of we fire lots of false positive notification on RealmObject before. But let me know if there are any tests related to this PR need to be added to the backlink tests.

realm/realm-library/src/main/java/io/realm/ObjectChangeSet.java
cmelchior: I would think a very common pattern is:

```
if (changeset.isFieldChanged("name")) {
  updateName(changeSet.getChange(String.class, "name").newValue());
}
```

That should be as easy as possible.

Agree, that it gets slightly ambigious with deleted objects, but `getChangedFields()` would also return nothing, so I think that will be acceptable?

beeender: OK, i will add it in another PR.

realm/realm-library/src/androidTest/java/io/realm/LinkingObjectsManagedTests.java
```diff
@@ -151,44 +151,6 @@ public void basic_multipleReferencesFromParentList() {
         assertEquals(parent, child.getListParents().last());
     }
 
-    // A listener registered on the backlinked object should be called when a commit adds a backlink
-    @Test
-    @RunTestInLooperThread
-    public void notification_onCommitModelObject() {
```
cmelchior: It feels like a mistake to delete this test? Perhaps a restriction in the Object Store?

realm/realm-library/src/androidTest/java/io/realm/LinkingObjectsManagedTests.java
```diff
@@ -224,81 +186,6 @@ public void run(Realm realm) {
             child, parent);
     }
 
-    // A listener registered on the backlinked object should be called when a backlinked object is deleted
-    @Test
-    @RunTestInLooperThread
-    public void notification_onDeleteModelObject() {
```
cmelchior: Same

realm/realm-library/src/androidTest/java/io/realm/LinkingObjectsManagedTests.java
```diff
@@ -224,81 +186,6 @@ public void run(Realm realm) {
...
-            },
-            child, parent);
-    }
-
-    // A listener registered on the backlinked object is called
-    // for an unrelated change on the an object of the same type!!
-    // This test exists only to document existing (but odd) behavior.
-    @Test
-    @RunTestInLooperThread
-    public void notification_notSentOnUnrelatedChangeModelObject() {
```
cmelchior: This was only here to document the old behaviour so 👍

realm/realm-library/src/androidTest/java/io/realm/LinkingObjectsManagedTests.java
```diff
@@ -498,63 +385,6 @@ public void json_updateList() {
         assertTrue(parents.contains(parent));
     }
 
-    // A JSON update should generate a notifcation
-    @Test
-    @RunTestInLooperThread
-    public void json_jsonUpdateCausesNotification() {
```
cmelchior: Not sure if this should be deleted?

realm/realm-library/src/androidTest/java/io/realm/ObjectChangeSetTests.java
```diff
@@ -0,0 +1,281 @@
...
+        AllTypes allTypes = realm.where(AllTypes.class).findFirst();
+        checkChangedField(allTypes, AllTypes.FIELD_FLOAT);
+        realm.beginTransaction();
+        allTypes.setColumnFloat(42.0f);
+        realm.commitTransaction();
+    }
+
+    @Test
+    @RunTestInLooperThread(before = PopulateOneAllTypes.class)
+    public void changeDoubleField() {
```
cmelchior: All types doesn't seem to be covered. Missing is at least: `boolean`, `float`, `byte[]`, `Date`

realm/realm-library/src/androidTest/java/io/realm/TypeBasedNotificationsTests.java
```diff
@@ -737,6 +694,17 @@ public void run() {
                 @Override
                 public void onChange(Dog object) {
                     typebasedCommitInvocations.incrementAndGet();
+                    if (typebasedCommitInvocations.get() > NUMBER_OF_LISTENERS) {
+                        fail();
+                    } else if (typebasedCommitInvocations.get() == NUMBER_OF_LISTENERS) {
+                        // Delayed post in case the listener gets triggered more time than expected.
```
cmelchior: Not sure I understand the logic behind this statement?

realm/realm-library/src/main/java/io/realm/ObjectChangeSet.java
```diff
@@ -0,0 +1,36 @@
...
+ */
+public interface ObjectChangeSet {
+
+    /**
+     * @return true if the object has been deleted from the Realm.
+     */
+    boolean isDeleted();
+
+    /**
+     * @return the names of changed fields if the object still exists and there are field changes. Returns {@code null}
```
cmelchior: Just return `String[0]` if no fields changed IMO. No reason to force a bunch of null-checks onto people.

realm/realm-library/src/main/java/io/realm/RealmObject.java
```diff
@@ -313,6 +313,19 @@ public final boolean load() {
     }
 
     /**
+     * Adds a change listener to this RealmObject to get detailed information about the changes.
+     *
+     * @param listener the change listener to be notified.
+     * @throws IllegalArgumentException if the change listener is {@code null} or the object is an unmanaged object.
+     * @throws IllegalArgumentException if object is an unmanaged RealmObject.
```
cmelchior: You just stated that above.

realm/realm-library/src/main/java/io/realm/RealmObjectChangeListener.java
```diff
@@ -0,0 +1,55 @@
...
+ * @see Realm#removeAllChangeListeners()
+ * @see Realm#removeChangeListener(RealmChangeListener)
+ */
+public interface RealmObjectChangeListener<T> {
+
+    /**
+     * When this gets called to return the results of an asynchronous query made by {@link RealmQuery#findFirstAsync()},
+     * {@code changeSet} will be {@code null}.
+     * <p>
+     * When this gets called because of the object is deleted, {@code changeSet.isDeleted()} will return {@code true}
```
cmelchior: because the object was deleted

realm/realm-library/src/main/java/io/realm/RealmObjectChangeListener.java
```diff
@@ -0,0 +1,55 @@
...
+     * and {@code changeSet.getFieldChanges()} will return {@code null}.
+     * <p>
+     * When this gets called because of the object is modified, {@code changeSet.isDeleted()} will return {@code false}
```
cmelchior: because the object was modified

realm/realm-library/src/main/java/io/realm/ObjectChangeSet.java
cmelchior: Can you make a list of outstanding issues? or known problems?

realm/realm-library/src/androidTest/java/io/realm/LinkingObjectsManagedTests.java
beeender: no, it is not. And it is expected behavior from object store.
Line 175 called `parent.setFieldObject(child);`, the changed object is parent, but the original test added listener on `child` which won't be trigger since nothing changed on the `child`. I am not quite sure what is the purpose of this test, maybe @bmeike can help to clarify this.

realm/realm-library/src/androidTest/java/io/realm/TypeBasedNotificationsTests.java
beeender: we post call of `parent.setFieldObject(child);` just in case the listener gets called more than `NUMBER_OF_LISTENERS` times. In that case, it will fail on line 698.

Zhuinden:
Databinding support requires an ignored field. But that is a question of treating `transient` the same as `@Ignore`, and also that inheriting from a parent that has only ignored fields needs to be allowed by the annotation processor

realm/realm-library/src/main/java/io/realm/ObjectChangeSet.java
beeender: Sure, see #4366

realm/realm-library/src/androidTest/java/io/realm/LinkingObjectsManagedTests.java
cmelchior: Technically the `RealmResults` field on `child` is a field with a (backlink) reference to the parent, so one could argue it should trigger the same type of notification as when a `RealmList` changes. I'm fine with leaving it as it is right now, but we should probably document that somewhere.

realm/realm-library/src/main/cpp/io_realm_internal_OsObject.cpp
```diff
@@ -0,0 +1,205 @@
...
+        auto table = m_wrapper->m_object.row().get_table();
+        for (size_t i = 0; i < change_set.columns.size(); ++i) {
+            if (change_set.columns[i].empty()) {
+                continue;
+            }
+            // FIXME: After full integration of the OS schema, parse the column name from
+            // wrapper->m_object.get_object_schema() will be faster.
+            field_names.push_back(to_jstring(env, table->get_column_name(i)));
+        }
+        m_field_names_array = env->NewObjectArray(field_names.size(), java_lang_string, 0);
```
cmelchior: Shouldn't this be released again, after you sent it to Java in the `CallVoidMethod()` ?

realm/realm-library/src/main/cpp/io_realm_internal_OsObject.cpp
```diff
@@ -0,0 +1,205 @@
...
+        }
+    }
+
+private:
+    ObjectWrapper* m_wrapper;
+    bool m_deleted = false;
+    jobjectArray m_field_names_array = nullptr;
+};
+
+static void finalize_object(jlong ptr);
```
cmelchior: Mistake?

realm/realm-library/src/main/java/io/realm/ProxyState.java
```diff
@@ -30,18 +30,57 @@
...
 
     private Row row;
+    private OsObject osObject;
     private BaseRealm realm;
     private boolean acceptDefaultValue;
     private List<String> excludeFields;
 
-    private final List<RealmChangeListener<E>> listeners = new CopyOnWriteArrayList<RealmChangeListener<E>>();
-    protected long currentTableVersion = -1;
+    private ObserverPairList<OsObject.ObjectObserverPair> observerPairs =
```
cmelchior: I'm trying to convince myself that moving the CopyOnWriteArrayList inside the ObserverPairList is correct, but I'm still not sure? I'm not sure it will fail either though.

I would think that synchronizing the methods on ObserverPairList would be safer?

realm/realm-library/src/main/java/io/realm/RealmObject.java
```diff
@@ -354,6 +365,30 @@ public final boolean load() {
     }
 
     /**
+     * Adds a change listener to a RealmObject.
```
cmelchior: "Adds a change listener to a RealmObject that will be triggered if any field or referenced object is changed" ?

realm/realm-library/src/main/java/io/realm/RealmObjectChangeListener.java
```diff
@@ -0,0 +1,55 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.realm;
+
+/**
+ * {@code RealmObjectChangeListener} can be registered with a {@link RealmModel} or {@link RealmObject} to receive
```
cmelchior: registered with a -> registered on a

realm/realm-library/src/main/java/io/realm/RealmObjectChangeListener.java
```diff
@@ -0,0 +1,55 @@
...
+ * detailed notification about object changes.
```
cmelchior: notifications when a object changes.

realm/realm-library/src/main/java/io/realm/RealmObjectChangeListener.java
```diff
@@ -0,0 +1,55 @@
...
+ * <p>
+ * Realm instances on a thread without an {@link android.os.Looper} cannot register a {@code RealmObjectChangeListener}.
+ * <p>
+ *
+ * @param <T> The <a href="https://realm.io/docs/java/latest/#auto-updating-objects">live object</a> being returned
```
cmelchior: Should this be `T extends RealmModel` in this case, since a `RealmObjectChangeListener` only works on models

realm/realm-library/src/main/java/io/realm/RealmObjectChangeListener.java
```diff
@@ -0,0 +1,55 @@
...
+ * @see Realm#removeAllChangeListeners()
+ * @see Realm#removeChangeListener(RealmChangeListener)
+ */
+public interface RealmObjectChangeListener<T> {
+
+    /**
+     * When this gets called to return the results of an asynchronous query made by {@link RealmQuery#findFirstAsync()},
+     * {@code changeSet} will be {@code null}.
+     * <p>
+     * When this gets called because of the object was deleted, {@code changeSet.isDeleted()} will return {@code true}
```
cmelchior: because of the -> because the

realm/realm-library/src/main/java/io/realm/RealmObjectChangeListener.java
```diff
@@ -0,0 +1,55 @@
...
+     * and {@code changeSet.getFieldChanges()} will return {@code null}.
+     * <p>
+     * When this gets called because of the object was modified, {@code changeSet.isDeleted()} will return {@code false}
```
cmelchior: because of the -> because the

realm/realm-library/src/main/java/io/realm/RealmObjectChangeListener.java
```diff
@@ -0,0 +1,55 @@
...
+     * and {@code changeSet.getFieldChanges()} will return the detailed information about the fields' changes.
+     * <p>
+     * If a field points to another RealmObject this listener will only be triggered if the field is set to a new object
+     * or null. Updating the referenced RealmObject will not trigger this listener.
+     * <p>
+     * If a field points to a RealmList, this listener will only be triggered if a objects are inserted, removed or
+     * moved within the List. Updating the objects in the RealmList will no trigger this listener.
+     *
```
cmelchior: Also add a description about backlink fields.

Perhaps something like:

```
Changes to @LinkingObjects RealmResults fields will not be monitored, nor reported through this change listener.
```

realm/realm-library/src/main/java/io/realm/internal/SharedRealm.java
```diff
@@ -428,6 +429,38 @@ void invalidateIterators() {
         iterators.clear();
     }
 
+    // addPendingRow, removePendingRow and executePendingRow queries is to solve that the listener cannot be added
```
cmelchior: Add a test exposing that listeners cannot be added inside transactions

realm/realm-library/src/main/java/io/realm/ProxyState.java
beeender: `ObserverPairList` should never be used across thread. using `CopyOnWriteArrayList` is mainly because of user may call `xxx.removeListener` inside a listener callback.

realm/realm-library/src/main/cpp/io_realm_internal_OsObject.cpp
beeender: forward declaration ... i can remove it . Build failed without it before if I remembered correctly... but it builds now.

realm/realm-library/src/main/java/io/realm/RealmObject.java
beeender: "referenced object is changed" this is not correct I think?

realm/realm-library/src/main/cpp/io_realm_internal_OsObject.cpp
beeender: This is quite annoy :( Although the `NewObjectArray` returns an `LocalRef` and `LocalRef` will be auto released when it gets back to java, we still may reach the limitation of max size of local ref table (512 entries). Which means in this case, it will crash with more than 511 fields change. I am going to add a new JNI wrapper `JavaString` for this -- release the `LocalRef` and acquire a `GlobalRef` in that to solve the problem. Shall I do it in another PR?

cmelchior: You only create one ObjectArray pr. notification right? So the number of fields shouldn't matter? But the number of objects registered notifications for will?

cmelchior: Also note, that the autorelease will only work if the thread is attached to Java. This is being run on the notifier thread which is a pure C++ thread, right?

beeender: No, this one runs in the thread where the listener registered, so it must be a java thread.

It needs to return a `ObjectArray` with `String`s, every individual entry unfortunately is a java `String` object.

realm/realm-library/src/androidTest/java/io/realm/ObjectChangeSetTests.java
```diff
@@ -0,0 +1,316 @@
...
+        AllTypes allTypes = realm.where(AllTypes.class).findFirstAsync();
+        allTypes.addChangeListener(new RealmObjectChangeListener<AllTypes>() {
+            @Override
+            public void onChange(AllTypes object, ObjectChangeSet changeSet) {
+                assertTrue(object.isValid());
+                assertNull(changeSet);
+                looperThread.testComplete();
+            }
+        });
+    }
```
kneth: What if multiple fields are changed in the same transaction?

realm/realm-library/src/main/cpp/io_realm_internal_OsObject.cpp
```diff
@@ -0,0 +1,203 @@
...
+        for (size_t i = 0; i < field_names.size(); ++i) {
+            env->SetObjectArrayElement(m_field_names_array, i, field_names[i]);
+        }
+    }
+
+    JNIEnv* check_env()
+    {
+        JNIEnv* env = JniUtils::get_env(false);
+        if (!env || env->ExceptionCheck()) {
+            // JVM dettached or java exception has been thrown before.
```
kneth: "dettached" -> "detached"

realm/realm-library/src/main/java/io/realm/RealmObject.java
```diff
@@ -313,11 +313,22 @@ public final boolean load() {
     }
 
     /**
+     * Adds a change listener to this RealmObject to get detailed information about the changes.
```
kneth: "about the" -> "about"

realm/realm-library/src/main/java/io/realm/RealmObjectChangeListener.java
```diff
@@ -0,0 +1,57 @@
...
+ * limitations under the License.
+ */
+
+package io.realm;
+
+import io.realm.annotations.LinkingObjects;
+
+/**
+ * {@code RealmObjectChangeListener} can be registered on a {@link RealmModel} or {@link RealmObject} to receive
+ * detailed notification when a object changes.
```
kneth: "notification" -> "notifications"

realm/realm-library/src/main/java/io/realm/RealmObjectChangeListener.java
```diff
@@ -0,0 +1,57 @@
...
+ * <p>
+ * Realm instances on a thread without an {@link android.os.Looper} cannot register a {@code RealmObjectChangeListener}.
+ * <p>
+ *
+ * @param <T> The type of {@link RealmModel} where your listener will be registered on.
```
kneth: "where" -> "on which"?

realm/realm-library/src/main/java/io/realm/internal/SharedRealm.java
kneth: "is" -> "are"

realm/realm-library/src/main/java/io/realm/internal/SharedRealm.java
```diff
@@ -428,6 +429,38 @@ void invalidateIterators() {
...
+    // inside a transaction. For the findFirstAsync(), listener is registered on a Object Store Results first, then move
```
kneth: "a Object" -> "an Object"

realm/realm-library/src/main/java/io/realm/internal/SharedRealm.java
```diff
@@ -428,6 +429,38 @@ void invalidateIterators() {
...
+    // the listeners to the Object when the query for Results returns. When beginTransaction() called, all listeners'
+    // on the results will be triggered first, that leads to the registration of listeners on the Object which will
+    // will throw because of the transaction has began already. So here we execute all PendingRow queries first before
```
kneth: remove "will"
"has began already" -> "has already begun"

bmeike:
I am just getting to this.  I think that '@' to me, somehow did not actually notify me.
So, I'm not qualified to talk about "expected behavior" as far as core is concerned.  I *definitely* would expect to be notified when the contents of a backlinked field changes.  If that is not the case, then I think we need documentation that says that notifications do not work for backlinks.

realm/realm-library/src/androidTest/java/io/realm/LinkingObjectsManagedTests.java
bmeike: Yeah... I'm with Christian on this.  Something has, most definitely, changed, for the child.  child.backLinkedField.size(), for instance, is changes.  That is, clearly, a change.

If notifications on backlinks are not supported, that's life.  We need clear documentation of the fact, though.

realm/realm-library/src/androidTest/java/io/realm/ObjectChangeSetTests.java
beeender: `long` is just the same as `int`, and there is no int field in the `AllTypes` :(

realm/realm-library/src/androidTest/java/io/realm/ObjectChangeSetTests.java
beeender: Added a test case.

realm/realm-library/src/androidTest/java/io/realm/ObjectChangeSetTests.java
```diff
@@ -0,0 +1,316 @@
...
+            public void onChange(RealmModel object, ObjectChangeSet changeSet) {
+                fail();
+            }
+        });
+        looperThread.postRunnableDelayed(new Runnable() {
+            @Override
+            public void run() {
+                looperThread.testComplete();
+            }
+        }, 100);
```
beeender: The event might be sent through the notifier thread and we have no control for that thread unfortunately.  I don't like the delay here, and I don't like to rely on the fact we are triggering on the `commitTransaction` will trigger the listener either . So post delay is more "common" solution here.

realm/realm-library/src/androidTest/java/io/realm/ObjectChangeSetTests.java
```diff
@@ -0,0 +1,316 @@
...
+        allTypes.setColumnRealmObject(realm.createObject(Dog.class));
+        realm.commitTransaction();
+    }
+
+    @Test
+    @RunTestInLooperThread(before = PopulateOneAllTypes.class)
+    public void changeLinkFieldRemoveObject() {
+        Realm realm = looperThread.realm;
+        AllTypes allTypes = realm.where(AllTypes.class).findFirst();
+        checkChangedField(allTypes, AllTypes.FIELD_REALMOBJECT);
```
beeender: This one is actually testing deletion. Added a test `changeLinkFieldSetNull`

realm/realm-library/src/main/cpp/io_realm_internal_OsObject.cpp
```diff
@@ -0,0 +1,203 @@
...
+    CATCH_STD()
+}
+
+JNIEXPORT void JNICALL Java_io_realm_internal_OsObject_nativeStopListening(JNIEnv* env, jobject, jlong native_ptr)
+{
+    TR_ENTER_PTR(native_ptr)
+
+    try {
+        auto wrapper = reinterpret_cast<ObjectWrapper*>(native_ptr);
+        wrapper->m_notification_token = {};
```
beeender: `m_notification_token` is an instance but not a pointer. so you cannot give a `nullptr` value to it. `{}` is the efficient way here since it will destruct the old instance and give it empty values. See definition of `NotificationToken& operator=(NotificationToken&&);`

realm/realm-library/src/main/java/io/realm/RealmObject.java
```diff
@@ -354,6 +365,30 @@ public final boolean load() {
...
     /**
+     * Adds a change listener to a RealmObject.
+     *
+     * @param object RealmObject to add listener to.
+     * @param listener the change listener to be notified.
+     * @throws IllegalArgumentException if the {@code object} is {@code null} or an unmanaged object, or the change
+     * listener is {@code null}.
+     * @throws IllegalStateException if you try to add a listener from a non-Looper or {@link IntentService} thread.
+     */
+    public static <E extends RealmModel> void addChangeListener(E object, RealmChangeListener<E> listener) {
```
beeender: yes, the idea is we still support to add the `RealmListener` on `RealmObject`.

realm/realm-library/src/main/java/io/realm/ProxyState.java
```diff
@@ -30,18 +30,57 @@
  * {@link RealmObject} and {@link DynamicRealmObject}.
  */
 public final class ProxyState<E extends RealmModel> implements PendingRow.FrontEnd {
+
+    static class RealmChangeListenerWrapper<T extends RealmModel> implements RealmObjectChangeListener<T> {
```
beeender: yes, see above

realm/realm-library/src/main/java/io/realm/RealmObjectChangeListener.java
```diff
@@ -0,0 +1,57 @@
...
+
+    /**
+     * When this gets called to return the results of an asynchronous query made by {@link RealmQuery#findFirstAsync()},
+     * {@code changeSet} will be {@code null}.
+     * <p>
+     * When this gets called because the object was deleted, {@code changeSet.isDeleted()} will return {@code true}
+     * and {@code changeSet.getFieldChanges()} will return {@code null}.
+     * <p>
+     * When this gets called because the object was modified, {@code changeSet.isDeleted()} will return {@code false}
+     * and {@code changeSet.getFieldChanges()} will return the detailed information about the fields' changes.
```
beeender: it means `fields's` here.

realm/realm-library/src/androidTest/java/io/realm/LinkingObjectsManagedTests.java
beeender: Clarified in the javadoc of `LinkingObjects`.

realm/realm-library/src/androidTest/java/io/realm/RealmObjectTests.java
```diff
@@ -1637,9 +1638,49 @@ public void onChange(Dog object) {
...
+                }
+            });
+            fail("adding change listener on unmanaged object must throw an exception.");
+        } catch (IllegalArgumentException ignore) {
+        }
+
+        looperThread.testComplete();
+    }
+
+    // Object Store will throw when adding change listener inside a transaction.
```
beeender: Added javadoc. removing is allowed inside a transaction.

realm-annotations/src/main/java/io/realm/annotations/LinkingObjects.java
```diff
@@ -57,6 +57,7 @@
  *     <li>They are ignored when doing a `copyToRealm().`</li>
  *     <li>They are ignored when doing a `copyFromRealm().`</li>
  *     <li>They are ignored when using the various `createObjectFromJson*` and `createAllFromJson*` methods.</li>
+ *     <li>The listeners on parent object won't be triggered when the linking objects changes.</li>
```
cmelchior: Perhaps rephase. I especially find "parent object" confusion since I assume it means the object pointing to this object.

Suggestion: 
Listeners on an object with a @LinkingObject field will not be triggered if the linking objects change, e.g if another object drops a reference to this object.

realm/realm-library/src/main/java/io/realm/RealmObject.java
cmelchior: That is how the current change listeners work? The fine-grained object notifications work slightly different that way

beeender: Does "referenced object" mean the `RealmObject` field?

cmelchior: Yes and RealmList

beeender: what about ""Adds a change listener to a RealmObject that will be triggered if any value field or referenced RealmObject field is changed, or the `RealmList` field itself is changed.` -- since change of the object in the `RealmList` field will not trigger the notification.

cmelchior: Really, I thought it did that as well at least for the global ChangeListener, but if not, then yes, that wording is fine 👍

realm/realm-library/src/main/java/io/realm/internal/SharedRealm.java
```diff
@@ -429,11 +429,11 @@ void invalidateIterators() {
...
     }
 
-    // addPendingRow, removePendingRow and executePendingRow queries is to solve that the listener cannot be added
-    // inside a transaction. For the findFirstAsync(), listener is registered on a Object Store Results first, then move
+    // addPendingRow, removePendingRow and executePendingRow queries are to solve that the listener cannot be added
+    // inside a transaction. For the findFirstAsync(), listener is registered on an Object Store Results first, then move
     // the listeners to the Object when the query for Results returns. When beginTransaction() called, all listeners'
     // on the results will be triggered first, that leads to the registration of listeners on the Object which will
-    // will throw because of the transaction has began already. So here we execute all PendingRow queries first before
+    // throw because of the transaction has already began. So here we execute all PendingRow queries first before
```
kneth: "began" -> "begun"

realm-annotations/src/main/java/io/realm/annotations/LinkingObjects.java
```diff
@@ -57,7 +57,6 @@
  *     <li>They are ignored when doing a `copyToRealm().`</li>
  *     <li>They are ignored when doing a `copyFromRealm().`</li>
  *     <li>They are ignored when using the various `createObjectFromJson*` and `createAllFromJson*` methods.</li>
- *     <li>The listeners on parent object won't be triggered when the linking objects changes.</li>
```
Zhuinden: oh, so they **are** triggered? Cool.

realm/realm-library/src/androidTest/java/io/realm/ObjectChangeSetTests.java
Zhuinden: Technically calling listeners for local commit in `commitTransaction()` will fix some silly bugs like https://github.com/realm/realm-java/issues/4384 so it is a good thing.

realm-annotations/src/main/java/io/realm/annotations/LinkingObjects.java
beeender: no, but the next line in this file is talking about the same thing :p

Zhuinden: oh, so I should learn how to read 😄

bmeike:
Thanks for adding the doc.  I think that will reduce the number of surprised developers!

Issue #4030, opened 2017-01-12
Make introExample fail on warnings


bmeike:
Many enterprises have a "no warnings" policy.  Realm should not introduce warnings.  introExample should be a canary, failing if there are any warnings

cmelchior:
We added the ErrorProne plugin in #4342, which should cover this. Closing

PR #4338, opened 2017-03-17
fix warnings in generated code

### LINKED ISSUE #4329: Add @Generated("realm") annotation to generated classes
#### Goal

Be able to suppress ErrorProne warnings on Realm generated classes.

#### Expected Results

Right now, generated classes miss a lot of `@Overidde` annotation which makes warnings (and error when treating warnings as errors) when compiler process Realm generated classes (like ModelRealmProxy).

ErrorProne has a way to ignore any validations for classes that are generated. But to work, the generated classes must have a `@Generated(<value>)` annotation.

I request that the code generator project adds the `@Generated("realm")` annotation to all generated classes.

I checked `JavaWriter` class and it should in theory simple as adding `emitAnnotation("Generated", "realm")` at right location, like at top of class. For `RealmProxy`, it would be around here probably: https://github.com/realm/realm-java/blob/master/realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmProxyClassGenerator.java#L97

I maybe would have time to propose a PR, is it possible to only recompile realm-annotation-processor and replace it in my project to test it? I would like to avoid having to compile everything.

zaki50:
Add `Override` annotation to accessors and stop using raw type in proxy classes in order to remove warnings from java.

fixes #4329

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 * `Realm.migrateRealm(RealmConfiguration)` now fails correctly with an `IllegalArgumentException` if a `SyncConfiguration` is provided (#4075).
 * Fixed a potential cause for Realm file corruptions (never reported).
+* Add `@Override` annotation to accessors and stop using raw type in proxy classes in order to remove warnings from javac (#4329).
 
 ### Deprecated
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 
 * `Realm.migrateRealm(RealmConfiguration)` now fails correctly with an `IllegalArgumentException` if a `SyncConfiguration` is provided (#4075).
 * Fixed a potential cause for Realm file corruptions (never reported).
-* Add `@Override` annotation to accessors and stop using raw type in proxy classes in order to remove warnings from javac (#4329).
+* Add `@Override` annotation to proxy class accessors and stop using raw type in proxy classes in order to remove warnings from javac (#4329).
 
 ### Deprecated
 
--- a/realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmProxyClassGenerator.java
+++ b/realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmProxyClassGenerator.java
@@ -1608,6 +1608,7 @@ public class RealmProxyClassGenerator {
             return;
         }
         writer.emitAnnotation("Override");
+        writer.emitAnnotation("SuppressWarnings", "\"ArrayToString\"");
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 
 * `Realm.migrateRealm(RealmConfiguration)` now fails correctly with an `IllegalArgumentException` if a `SyncConfiguration` is provided (#4075).
 * Fixed a potential cause for Realm file corruptions (never reported).
+* Add `@Override` annotation to accessors and stop using raw type in proxy classes in order to remove warnings from javac (#4329).
```
cmelchior: to proxy class accessors

zaki50:
@cmelchior As I commented in https://github.com/realm/realm-java/issues/4329#issuecomment-287333402, I couldn't reproduce the override warnings on my environment.

cmelchior:
Did you try installing the error prone plugin, since that was what reported it? https://github.com/google/error-prone/tree/master/examples/gradle

zaki50:
@cmelchior Thanks! I've got the warnings and confirmed that this PR removed them.

I'm considering the way to the test.

Adding the error prone plugin to Realm Java itself is the first choice, but that produces many warnings. If we choose this way, we should fix or suppress all warnings.

Another choice is adding an example project which uses error prone plugin and in which `-Werror` option is enabled.

What do you think @cmelchior and @realm/java ?

cmelchior:
We have this issue as well for this: https://github.com/realm/realm-java/issues/4030
Perhaps adding the errorProne plugin to the `introExample`? or would that create to much noise?

zaki50:
I'd like to keep `introExample` as lean as possible.
Adding `-Werror` to `introExample ` is acceptable for me, but I think adding error prone plugin is too much.

I'm trying to add error prone plugin to Realm Java project.
I've found that the warnings are not so many and it reports some useful warnings (for example, I've found a test case that missing \@Test annotation).

maoueh:
@zaki50 @cmelchior Thanks guy, so quick :)

And glad you've found some small stuff by adding ErrorProne to the project. That's the point of it.

kneth:
How many warning do we get, if we enable Error Prone? I have heard nice things about that too, and can it prevent just one bug, it is worth it.

zaki50:
@kneth I made a PR to fix https://github.com/realm/realm-java/pull/4339 but there are more tests to be fixed.
I'l made another PR to fix the rest.

There are still many warnings to be fixed such as missing `break` at the end of `case`.

We should introduce `ErrorProne`.

zaki50:
I've found another error of calling toString to array object in proxy class.
I'll create another PR to fix that.

zaki50:
@realm/java added more fixes.

PR #4341, opened 2017-03-17
fix warnings reported by ErrorProne (second PR)

zaki50:
@realm/java

### DIFF
--- a/realm/realm-library/src/androidTest/java/io/realm/ManagedRealmCollectionTests.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/ManagedRealmCollectionTests.java
@@ -760,13 +760,16 @@ public class ManagedRealmCollectionTests extends CollectionTests {
                     case REMOVE_ALL:
                     case RETAIN_ALL:
                         expected = UnsupportedOperationException.class;
+                        break;
+                    default:
+                        // use default exception
                 }
             }
 
             try {
                 switch (method) {
                     case DELETE_ALL: collection.deleteAllFromRealm(); break;
-                    case ADD_OBJECT: collection.add(new AllJavaTypes());
+                    case ADD_OBJECT: collection.add(new AllJavaTypes()); break;
                     case ADD_ALL_OBJECTS: collection.addAll(Collections.singletonList(new AllJavaTypes())); break;
                     case CLEAR: collection.clear(); break;
                     case REMOVE_OBJECT: collection.remove(new AllJavaTypes()); break;
@@ -847,6 +850,9 @@ public class ManagedRealmCollectionTests extends CollectionTests {
                         case REMOVE_ALL:
                         case RETAIN_ALL:
                             expected = UnsupportedOperationException.class;
+                            break;
+                        default:
+                            // use default exception
                     }
                 }
 
@@ -854,8 +860,8 @@ public class ManagedRealmCollectionTests extends CollectionTests {
                     switch (method) {
                         case ADD_OBJECT: collection.add(new AllJavaTypes()); break;
                         case ADD_ALL_OBJECTS: collection.addAll(Collections.singletonList(new AllJavaTypes())); break;
-                        case CLEAR: collection.clear();
-                        case CONTAINS:
+                        case CLEAR: collection.clear(); break;
+                        case CONTAINS: collection.contains(tempObject); break;
                         case CONTAINS_ALL: collection.containsAll(Collections.singletonList(tempObject)); break;
                         case EQUALS:
                             //noinspection ResultOfMethodCallIgnored
--- a/realm/realm-library/src/androidTest/java/io/realm/NotificationsTest.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/NotificationsTest.java
@@ -564,7 +564,8 @@ public class NotificationsTest {
 
         // Starts background looper and let it hang.
         ExecutorService executorService = Executors.newSingleThreadExecutor();
-        executorService.submit(new Runnable() {
+        //noinspection unused
+        final Future<?> future = executorService.submit(new Runnable() {
             @Override
             public void run() {
                 Looper.prepare(); // Fake background thread with a looper, eg. a IntentService.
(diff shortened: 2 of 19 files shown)


realm/realm-library/src/androidTest/java/io/realm/ManagedRealmCollectionTests.java
```diff
@@ -760,13 +760,16 @@ public void mutableMethodsOutsideTransactions() {
...
+                    default:
+                        // use default exception
                 }
             }
 
             try {
                 switch (method) {
                     case DELETE_ALL: collection.deleteAllFromRealm(); break;
-                    case ADD_OBJECT: collection.add(new AllJavaTypes());
+                    case ADD_OBJECT: collection.add(new AllJavaTypes()); break;
```
bmeike: Would you be willing to put these `break`s on new lines?  It would be more consistent, if less compact.

realm/realm-library/src/androidTest/java/io/realm/ManagedRealmCollectionTests.java
```diff
@@ -760,13 +760,16 @@ public void mutableMethodsOutsideTransactions() {
                     case REMOVE_ALL:
                     case RETAIN_ALL:
                         expected = UnsupportedOperationException.class;
+                        break;
+                    default:
```
bmeike: I'm not sure I see the advantage in adding an explicit default clause...

realm/realm-library/src/androidTest/java/io/realm/ManagedRealmCollectionTests.java
```diff
@@ -847,15 +850,18 @@ public Boolean call() throws Exception {
                         case REMOVE_ALL:
                         case RETAIN_ALL:
                             expected = UnsupportedOperationException.class;
+                            break;
+                        default:
```
bmeike: Again, I'm not sure why the `default`

realm/realm-library/src/androidTest/java/io/realm/NotificationsTest.java
```diff
@@ -564,7 +564,8 @@ public void looperThreadQuitsLooperEarly() throws InterruptedException {
 
         // Starts background looper and let it hang.
         ExecutorService executorService = Executors.newSingleThreadExecutor();
-        executorService.submit(new Runnable() {
+        //noinspection unused
+        final Future<?> future = executorService.submit(new Runnable() {
```
bmeike: Why keep the `future`?  It is never used, I believe...

realm/realm-library/src/androidTest/java/io/realm/RealmQueryTests.java
```diff
@@ -99,14 +99,14 @@ private void populateTestRealm(Realm testRealm, int dataSize) {
             allTypes.setColumnBinary(new byte[]{1, 2, 3});
             allTypes.setColumnDate(new Date(DECADE_MILLIS * (i - (dataSize / 2))));
```
bmeike: If you would be willing, I would prefer that all of those lower case 'f' float indicators were uppercased.  It isn't so important with the 'f', but getting in the habit of using upper case prevents having to distinguish between `54321` and `5432l`

realm/realm-library/src/androidTest/java/io/realm/rule/RunInLooperThread.java
```diff
@@ -108,7 +109,8 @@ public Thread newThread(Runnable runnable) {
                             return new Thread(runnable, threadName);
                         }
                     });
-                    executorService.submit(new Runnable() {
+                    //noinspection unused
+                    final Future<?> submit = executorService.submit(new Runnable() {
```
bmeike: Yea... I don't get this.  Why not just ignore it?

realm/realm-library/src/androidTest/java/io/realm/ManagedRealmCollectionTests.java
zaki50: The `case`s in this `switch` are not covering all values in the `Enum` and that causes some warnings of ErrorProne.
This `default` means all other `case`s can be ignored.

realm/realm-library/src/androidTest/java/io/realm/ManagedRealmCollectionTests.java
zaki50: I aligned it to existing code.
I think reformatting should be done in separate PR.

realm/realm-library/src/androidTest/java/io/realm/NotificationsTest.java
zaki50: `ErrorProne` reports a warning that the return value of `submit()` should not be ignored.
To suppress that, just assigning the return value to unused local variable.

realm/realm-library/src/androidTest/java/io/realm/RealmQueryTests.java
zaki50: I agree with that, but this PR is focusing on fixing warnings from `ErrorProne`.
We should have another PR to fix  that.

realm/realm-library/src/androidTest/java/io/realm/rule/RunInLooperThread.java
zaki50: same

realm/realm-library/src/androidTest/java/io/realm/NotificationsTest.java
zaki50: Of course we can suppress the ErrorProne warning directly, but it affects entire method.
I think suppressing scope should be as narrow as possible.

realm/realm-library/src/androidTest/java/io/realm/RealmQueryTests.java
cmelchior: That is actually a great argument. I think right now we are fairly inconsistent, but I'm all for enforcing uppercasing the type suffix 👍

PR #4342, opened 2017-03-17
Introduce ErrorProne plugin

zaki50:
This PR requires merge of #4338 and #4341.
After merging of those PRs, all checks should be passed.


cmelchior:
This should also fix: #4030

PR #4358, opened 2017-03-21
findFirstAsync returns invalid row if no object

### LINKED ISSUE #4352: Issue with addChangeListener on async query, without objects created.
#### Goal

In Realm `3.0.0`, the following code, documented in `isLoaded()` method, doesn't seem to work, if there aren't any `Person` objects created in the database.

```Java
Person person = realm.where(Person.class).findFirstAsync();
person.isLoaded(); // == false
person.addChangeListener(new RealmChangeListener() {
    @Override
    public void onChange(Person person) {
        person.isLoaded(); // Always true here
        if (person.isValid()) {
            // It is safe to access the person.
        }
    }
});
```

#### Expected Results

`onChange()` should be called with `person.isLoaded() == true` and `person.isValid() == false`.

#### Actual Results

`onChange()` is never called.

Realm version: 3.0.0

beeender:
Fix #4352
The findFirstAsync()'s behavior doesn't match the javadoc from the first
day the API was introduced. It kept running the query until it could
find a row match the query condition. This behavior create difficulties
if user want to check if there is no object in the db. Also it was not
consistent with the behavior of findFirst() which will return an invalid
row in the same condition.

### DIFF
--- a/realm/realm-library/src/main/java/io/realm/internal/PendingRow.java
+++ b/realm/realm-library/src/main/java/io/realm/internal/PendingRow.java
@@ -58,12 +58,12 @@ public class PendingRow implements Row {
                         Row row = returnCheckedRow ? CheckedRow.getFromRow(uncheckedRow) : uncheckedRow;
                         // Ask the front end to reset the row and stop async query.
                         frontEnd.onQueryFinished(row);
-                        clearPendingCollection();
+                    } else {
+                        frontEnd.onQueryFinished(InvalidRow.INSTANCE);
                     }
-                } else {
-                    // The Realm is closed. Do nothing then.
-                    clearPendingCollection();
                 }
+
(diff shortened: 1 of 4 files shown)


beeender:
@realm/java

realm/realm-library/src/main/java/io/realm/internal/PendingRow.java
```diff
@@ -58,12 +58,12 @@ public void onChange(PendingRow pendingRow) {
                         Row row = returnCheckedRow ? CheckedRow.getFromRow(uncheckedRow) : uncheckedRow;
```
zaki50: we should update the comment a few lines above.

`// If no rows returned by the query, just wait for the query updates until it returns a valid row.`

PR #4359, opened 2017-03-21
Fix code comments

beeender:


beeender:
Only comments change.

zaki50:
:+1:

PR #4365, opened 2017-03-23
Add JNI helper class JavaGlabalRef and JavaClass

beeender:
- Added two class to help manage the jobject lifecycle.
- Use a static local JavaClass for SyncManager.

Since to support old/new values for object notification, we need to create java objects from JNI frequently. Clean up this part to make future work easier.

### DIFF
--- a/realm/realm-library/src/main/cpp/jni_util/jni_utils.hpp
+++ b/realm/realm-library/src/main/cpp/jni_util/jni_utils.hpp
@@ -19,6 +19,10 @@
 
 #include <jni.h>
 
+#include <vector>
+
+#include "java_global_ref.hpp"
+
 namespace realm {
 namespace jni_util {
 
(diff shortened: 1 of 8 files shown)


realm/realm-library/src/main/cpp/jni_util/jni_utils.hpp
```diff
@@ -31,12 +35,16 @@ class JniUtils {
 
     // Call this only once in JNI_OnLoad.
     static void initialize(JavaVM* vm, jint vm_version) noexcept;
+    // Call this in JNI_OnUnload.
+    static void release();
```
cmelchior: `releaseGlobalRefs()` would probably be a better name. Right now it is very generic?

realm/realm-library/src/main/cpp/jni_util/jni_utils.hpp
```diff
@@ -31,12 +35,16 @@ class JniUtils {
...
     // When attach_if_needed is false, returns the JNIEnv if there is one attached to this thread. Assert if there is
     // none. When attach_if_needed is true, try to attach and return a JNIEnv if necessary.
     static JNIEnv* get_env(bool attach_if_needed = false);
     // Detach the current thread from the JVM. Only required for C++ threads that where attached in the first place.
     // Failing to do so is a resource leak.
     static void detach_current_thread();
+    // Keep the given global reference until the JNI_OnUnload called.
```
cmelchior: until JNI_OnUnload is called.

realm/realm-library/src/main/cpp/jni_util/jni_utils.hpp
beeender: `release()` is corresponding to the above `initialize()`. So it is supposed to do more clean up work if needed. (But actually there is no practical way to unload a so, so the `JNI_OnUnload()` will never be called 😁  )

PR #4367, opened 2017-03-23
Javadoc for findFirstAsync

### LINKED ISSUE #4360: ChangeListener no longer triggered on a findFirstAsync, after delete
#### Goal

I have a single Token object, that I need to insert/update/delete and be notified every time this happens.
Here's what I do:

```java
        mTokenObservable = database.getRealm().where(Token.class).findFirstAsync();
        mTokenObservable.addChangeListener((RealmChangeListener<Token>) element -> {
            if (element.isValid()) {
                setAccessToken(element.accessToken);
            } else {
                // The token has been deleted
                setAccessToken(null);
            }
        });
```

- This work after the first insert, and also after updates.
- This also works after the first delete.
- But when I insert again after that, I am no longer called.
Note: I confirmed by using Realm Browser that the insert really worked.

#### Expected Results

I should be called.

#### Actual Results

On insertion after the delete, the listener is not triggered

FYI: If I change my `mTokenObservable` to a list and use `findAll` instead of `findFirstAsync`, this works.

#### Code Sample

See above.

Realm version(s): 3.0.0

Android Studio version: Android Studio 2.4 Preview 2 Build #AI-171.3804685, built on March 15, 2017

Which Android version and device: Emulator API 23

beeender:
Fix #4360

### DIFF
--- a/realm/realm-library/src/main/java/io/realm/RealmQuery.java
+++ b/realm/realm-library/src/main/java/io/realm/RealmQuery.java
@@ -1918,7 +1918,7 @@ public class RealmQuery<E extends RealmModel> {
      * Similar to {@link #findFirst()} but runs asynchronously on a worker thread. An listener should be registered to
      * the returned {@link RealmObject} to get the notification when query completes. The registered listener will also
      * be triggered if there are changes made to the queried {@link RealmObject}. If the {@link RealmObject} is deleted,
-     * the listener will be called the last time and the query will not be re-executed.
+     * the listener will be called one last time and then stop. The query will not be re-run.
      *
      * @return immediately an empty {@link RealmObject} with {@code isLoaded() == false}. Trying to access any field on
      * the returned object before it is loaded will throw an {@code IllegalStateException}.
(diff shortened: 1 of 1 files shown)


realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -1915,15 +1915,14 @@ public E findFirst() {
...
-     * @return immediately an empty {@link RealmObject}. Trying to access any field on the returned object
-     * before it is loaded will throw an {@code IllegalStateException}. Use {@link RealmObject#isLoaded()} to check if
-     * the object is fully loaded or register a listener {@link io.realm.RealmObject#addChangeListener}
-     * to be notified when the query completes. If no RealmObject was found after the query completed, the returned
-     * RealmObject will have {@link RealmObject#isLoaded()} set to {@code true} and {@link RealmObject#isValid()} set to
-     * {@code false}.
+     * Similar to {@link #findFirst()} but runs asynchronously on a worker thread. An listener should be registered to
+     * the returned {@link RealmObject} to get the notification when query completes. The registered listener will also
+     * be triggered if there are changes made to the queried {@link RealmObject}. If the {@link RealmObject} is deleted,
+     * the listener will be called the last time and the query will not be re-executed.
```
cmelchior: called one last time and then stop. The query will not be re-run.

PR #4368, opened 2017-03-23
Trigger collection listeners when commit

### LINKED ISSUE #4245: Predictive UI animations and local transactions
Without https://github.com/realm/realm-java/issues/4225 , user will still meets the un-synced list views when doing local commit.

Take below sequences for example:
list - 10 items
local commit delete the 1st item
touch event comes, since the list is always live, list.size() return 9 and did a refresh(), everything is up to date
fine grained notification comes, with deletion = [0], 0 should be the index in the old collection, but now, since the list has been updated already, it will delete the 1st in the updated view, which is wrong

The cocoa's solution about this: https://realm.io/docs/objc/latest/#interface-driven-writes

My thoughts, the current cocoa solution (suppress the next chosen listener) is still the best option:
1. It avoids the overhead to trigger all other listeners on the same thread, which is not necessary when other list view is not in the front
2. Not too complex to users, user just need to update the view according to what he changed in the local transaction, also, this gives user more flexibility for more complex animations (I assume there are requirements than what we can handle in the change set)

Another quick solutions can be delivered before #4225:
1. Exposing `Refresh()` to trigger the listeners when commit local transaction and educating user to use it. (So when we ship the listener suppressing, user could know they can use suppressing instead of refresh to make it better.)
2. Do `refresh()` in the `commitTransaction`. This requires less work from user side, but a) it creates overhead for cases which doesn't need to trigger listeners. b) when we deliver the listener suppress, the `refresh` in `commitTransaction` has to be removed which should be considered as a breaking change.

Examples here https://github.com/realm/realm-android-adapters/tree/mc/drag-drop-testing shows the problem, it is for drag & drop. But it could be applied to other cases that the view loose consistency with data.

beeender:
- The listeners on RealmList/RealmResults will be tiggered immediately
  when the transaction committed on the same thread if the collections
  are changed or the async query should return.
- Listeners on the RealmObject will have same behaviour after #4331
  merged.

This is to solve the problem for predictive UI animations and local
transactions.
Fix #4245

### DIFF
--- a/realm/realm-library/src/androidTest/java/io/realm/RealmListTests.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/RealmListTests.java
@@ -1053,17 +1053,24 @@ public class RealmListTests extends CollectionTests {
                 fail();
             }
         });
+
+        collection.removeAllChangeListeners();
+
+        // This one is added after removal, so it should be triggered.
+        collection.addChangeListener(new RealmChangeListener<RealmList<Dog>>() {
+            @Override
+            public void onChange(RealmList<Dog> element) {
+                listenerCalledCount.incrementAndGet();
+                looperThread.testComplete();
+            }
+        });
+
+        // This should trigger the listener if there is any.
         realm.beginTransaction();
         collection.get(0).setAge(42);
         realm.commitTransaction();
 
-        collection.removeAllChangeListeners();
-
-        // This should trigger the listener if there is any.
-        realm.beginTransaction();
-        realm.cancelTransaction();
-        assertEquals(0, listenerCalledCount.get());
-        looperThread.testComplete();
+        assertEquals(1, listenerCalledCount.get());
     }
 
     @Test
@@ -1083,21 +1090,19 @@ public class RealmListTests extends CollectionTests {
                     @Override
                     public void onChange(RealmList<Dog> collection, OrderedCollectionChangeSet changes) {
                         assertEquals(0, listenerCalledCount.getAndIncrement());
+                        looperThread.testComplete();
                     }
                 };
 
         collection.addChangeListener(listener1);
         collection.addChangeListener(listener2);
-        realm.beginTransaction();
-        collection.get(0).setAge(42);
-        realm.commitTransaction();
 
         collection.removeChangeListener(listener1);
 
         // This should trigger the listener if there is any.
         realm.beginTransaction();
-        realm.cancelTransaction();
+        collection.get(0).setAge(42);
+        realm.commitTransaction();
         assertEquals(1, listenerCalledCount.get());
-        looperThread.testComplete();
     }
 }
--- a/realm/realm-library/src/androidTest/java/io/realm/internal/CollectionTests.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/internal/CollectionTests.java
@@ -378,13 +378,15 @@ public class CollectionTests {
         addRowAsync();
     }
 
-    // Local commit will trigger the listener first when beginTransaction gets called then again in the next event loop.
+    // Local commit will trigger the listener first when beginTransaction gets called then again when transaction
+    // committed.
     @Test
     @RunTestInLooperThread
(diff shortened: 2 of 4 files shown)


realm/realm-library/src/androidTest/java/io/realm/RealmListTests.java
```diff
@@ -1053,17 +1053,24 @@ public void onChange(RealmList<Dog> collection, OrderedCollectionChangeSet chang
...
             }
         });
-        realm.beginTransaction();
-        collection.get(0).setAge(42);
-        realm.commitTransaction();
 
         collection.removeAllChangeListeners();
 
+        // This one is added after removal, so it should be triggered.
+        collection.addChangeListener(new RealmChangeListener<RealmList<Dog>>() {
```
cmelchior: Any reason to add this in a test named `removeAllChangeListeners()` ?

realm/realm-library/src/androidTest/java/io/realm/internal/CollectionTests.java
```diff
@@ -396,15 +398,16 @@ public void onChange(Collection collection1) {
...
                         sharedRealm.close();
-                        looperThread.testComplete();
                         break;
                 }
             }
         });
         addRow(sharedRealm);
-        assertEquals(collection.size(), 5);
+        transactionCommitted.set(true);
+        looperThread.testComplete();
```
cmelchior: This seems to change the flow of this test. Before it was depending on the notification arriving through the looper, now it is triggered synchronously. 

Wouldn't it be better to make a new test case `listenerTriggeredOnCommit()` or something that tests that the change listener is triggered _both_ on `beginTransaction()` and `commitTransaction()` ?

realm/realm-library/src/androidTest/java/io/realm/RealmListTests.java
beeender: because of i don't want to rely on the behavior "listeners triggered when commit" in this test. By adding another listener after all listeners removed, it can ensure that `removeAllChangeListeners` works, and the listener added after will be triggered. Which means the same test should work even when "listeners are triggered in the next event loop."

realm/realm-library/src/androidTest/java/io/realm/internal/CollectionTests.java
beeender: See the diff of the comments at the beginning of the test, the original test is to test the `commitTransaction` will trigger the listener in the next event loop. But now, the behavior to be tested is changed.

zaki50: I think we should check if that the listener is called **twice**.
I don't think this updated test fails in case of not being called when committed.

zaki50: I think we should check if that the listener is called twice.
I don't think this updated test fails in case of not being called when committed.

beeender: Right! fixed!

realm/realm-library/src/androidTest/java/io/realm/internal/CollectionTests.java
```diff
@@ -398,13 +399,16 @@ public void onChange(Collection collection1) {
                     case 1:
                         assertEquals(collection1.size(), 5);
                         sharedRealm.close();
-                        looperThread.testComplete();
+                        break;
+                    default:
+                        fail();
```
zaki50: What will happen when reached to here after `looperThread.testComplete();` is called?
But yes, it's difficult to check that the listener is **not called after** `looperThread.testComplete();` is executed.
It's just a question from my curiosity.

beeender: If it reached here after `testComplete` called, the test will still fail with assertion.

Zhuinden:
This reminds me of the commit/notification behavior in 0.88.3

Issue #4215, opened 2017-02-21
Common prefix for Object Store Classes


beeender:
To match the same `Results` in Object Store.
Blame me. I made the confusing name.

Use common prefix `Os`

`Collection` -> `OsResults`
`SharedRealm` -> `OsSharedRealm`

cmelchior:
`Collection` -> `OsResults` have been merged. Only missing `SharedRealm` -> `OsSharedRealm` now.

beeender:
DONE

PR #4382, opened 2017-03-25
Refactor Schemas into separate native and realm-based implementations

bmeike:
I would like your opinion on this refactor.  I think it is mostly right, but, at this point I have done no test beyond making sure it compiles.  It illustrates my intention, though, without taking all the time to fix details.

I think this is a significant improvement and I think it would make it much easier to migrate the code, slowly, to using ObjectStore.

Have a look, if you would, and see if you agree.  Note that I am not particularly interested in bugs smaller than design-level problems.  I'm only trying to find out, at this point, if the basic idea is sound.


realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -389,14 +390,13 @@ private static void initializeSyncedRealm(Realm realm) {
             final Set<Class<? extends RealmModel>> modelClasses = mediator.getModelClasses();
 
             final ArrayList<RealmObjectSchema> realmObjectSchemas = new ArrayList<>();
-            final RealmSchema realmSchemaCache = new RealmSchema();
             for (Class<? extends RealmModel> modelClass : modelClasses) {
-                RealmObjectSchema realmObjectSchema = mediator.createRealmObjectSchema(modelClass, realmSchemaCache);
+                RealmObjectSchema realmObjectSchema = mediator.createRealmObjectSchema(modelClass, null);
```
Zhuinden: Null schema cache? interesting

realm/realm-library/src/main/java/io/realm/RealmNativeSchema.java
```diff
@@ -0,0 +1,127 @@
...
+
+/**
+ * Class for interacting with the Realm schema using a dynamic API. This makes it possible
+ * to add, delete and change the classes in the Realm.
+ * <p>
+ * All changes must happen inside a write transaction for the particular Realm.
+ *
+ * @see RealmMigration
+ */
+public class RealmNativeSchema extends RealmBaseSchema {
```
Zhuinden: this is part of public API, so I don't think it's that simple to rename it just like that.

realm/realm-library/src/main/java/io/realm/RealmSchema.java
```diff
@@ -36,121 +33,69 @@
...
      * @return the set of all classes in this Realm or no RealmObject classes can be saved in the Realm.
      */
+    @Override
     public Set<RealmObjectSchema> getAll() {
-        if (realm == null) {
-            long[] ptrs = nativeGetAll(nativePtr);
-            Set<RealmObjectSchema> schemas = new LinkedHashSet<RealmObjectSchema>(ptrs.length);
-            for (int i = 0; i < ptrs.length; i++) {
-                schemas.add(new RealmObjectSchema(ptrs[i]));
+        int tableCount = (int) realm.getSharedRealm().size();
```
Zhuinden: does this mean `realm == null` could never be true before?

realm/realm-library/src/main/java/io/realm/RealmSchema.java
```diff
@@ -36,121 +33,69 @@
...
-            String internalClassName = TABLE_PREFIX + className;
-            if (realm.sharedRealm.hasTable(internalClassName)) {
-                Table table = realm.sharedRealm.getTable(internalClassName);
-                RealmObjectSchema.DynamicColumnMap columnIndices = new RealmObjectSchema.DynamicColumnMap(table);
-                return new RealmObjectSchema(realm, table, columnIndices);
-            } else {
-                return null;
-            }
-        }
+
```
Zhuinden: so the `if(realm == null) { ... } else { ... }` branches were divided into two different classes?

When could `realm == null` be true for a RealmSchema? DynamicRealm vs typed Realm?

realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -651,6 +651,10 @@ protected void finalize() throws Throwable {
         super.finalize();
     }
 
+    public SharedRealm getSharedRealm() {
```
Zhuinden: I'm not sure this should be visible from the outside :smile:

realm/realm-library/src/main/cpp/CMakeLists.txt
```diff
@@ -36,7 +36,7 @@ set(classes_LIST
     io.realm.internal.Table io.realm.internal.CheckedRow
     io.realm.internal.LinkView io.realm.internal.Util io.realm.internal.UncheckedRow
     io.realm.internal.TableQuery io.realm.internal.SharedRealm io.realm.internal.TestUtil
-    io.realm.log.LogLevel io.realm.log.RealmLog io.realm.Property io.realm.RealmSchema
+    io.realm.log.LogLevel io.realm.log.RealmLog io.realm.Property io.realm.RealmNativeSchema
```
beeender: Also, move to internal package enables native object life control through `NativeObject` interface which will expose `getNativePtr()` publicly.

realm/realm-library/src/main/java/io/realm/RealmBaseSchema.java
```diff
@@ -0,0 +1,112 @@
...
+
+/**
+ * Class for interacting with the Realm schema using a dynamic API. This makes it possible
+ * to add, delete and change the classes in the Realm.
+ * <p>
+ * All changes must happen inside a write transaction for the particular Realm.
+ *
+ * @see RealmMigration
+ */
+public abstract class RealmBaseSchema {
```
beeender: I don't quite understand this abstract class. Why it is inherited by both `RealmSchema` and `RealmNativeSchema`? IMO, the object schema should be built on the top of OS `Schema` and remove the column indices cache from java since basically they are doing the same thing.

realm/realm-library/src/main/java/io/realm/Realm.java
kneth: Object Store is also caching.

realm/realm-library/src/main/java/io/realm/RealmBaseSchema.java
```diff
@@ -0,0 +1,112 @@
+/*
+ * Copyright 2015 Realm Inc.
```
kneth: 2017

realm/realm-library/src/main/java/io/realm/RealmBaseSchema.java
kneth: Currently Object Store does not have much support for dynamic schemas so `Realm` and `DynamicRealm` will have different needs when it comes to schemas.

realm/realm-library/src/main/java/io/realm/RealmNativeSchema.java
```diff
@@ -0,0 +1,127 @@
+/*
+ * Copyright 2015 Realm Inc.
```
kneth: 2017

realm/realm-library/src/main/java/io/realm/RealmBaseSchema.java
beeender: Why the needs are different? I saw `Schema(std::initializer_list<ObjectSchema> types);` which means we can actually build a `Schema` from a list of `ObjectSchema`?

beeender:
Some thoughts about what we are going to achieve by moving to the OS schema:
- Cache the schema in both java & JNI
   - JNI layer also needs to access the object schema for some cases, eg.: get the old & new values for object notifications from a column index. See https://github.com/realm/realm-java/pull/4331/files#diff-0b02758b73386cd32b206d69e081e258R79
   - Java needs the Schema object for create `SortDescriptor`, currently they are all retrieved from `Table`. See the `SortDescriptor.java`.
- Properly to handle the schema changes in the near future. This is currently not handled well, if a dynamic Realm changes the schema, other opened Realm instance won't know the schema changes, and they continue to operate the db based on the old column indices. This may not be an urgent thing to do, but we need to consider this when integrate the OS schema.
- Life cycle. Currently all native schema are closed when the Realm instance closed which relies on the `Realm.close()`. It would be nice if the native object can extend the `NativeObject` to let the daemon thread to free it. The benefit is if there is any possibility we can get rid of the `Realm.close()` method, we don't have to change this part of code anymore.

realm/realm-library/src/main/java/io/realm/RealmBaseSchema.java
cmelchior: A DynamicRealm doesn't have any schema you can build, so all information must be gathered dynamically on every access (possible cached).

realm/realm-library/src/main/java/io/realm/RealmNativeSchema.java
```diff
@@ -0,0 +1,127 @@
...
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.LinkedHashSet;
+import java.util.Map;
+import java.util.Set;
+
+
+/**
+ * Class for interacting with the Realm schema using a dynamic API. This makes it possible
```
cmelchior: It is a bit hard to guess what the purpose of these classes are when the Javadoc is wrong

realm/realm-library/src/main/java/io/realm/RealmBaseSchema.java
cmelchior: The idea about having an abstract super class for the Schema looks like a good idea to me. It means we can split the implementation and gradually move things over/around.

I have some concerns with the naming though. Right now we only expose `RealmSchema` and we should continue to do that in order to not break compatibility.
This means the inheritence hiearchy should looks something like this:

```
public abstract class io.realm.RealmSchema {
	
}

// Our current impl. bikeshed the name
public class StandardRealmSchema extends RealmSchema {
	
}

// The new Object store schema
public class OsRealmSchema extends RealmSchema {
	
}

// Dynamic Realm schema impl if required
public class DynamicRealmSchema extends RealmSchema {
	
}
```

That way we can just return the proper implementation subclass depending on the situation.

Keeping the public interface clean is more important than everything is in the correct package. Right now we have liberal use of package protected methods, which sometimes causes headaches. Classes in `io.realm.internal` are not considered part of the public API so can have whatever modifier makes sense.

Also I'm not sure how this approach solves the backlink field name issue? That seems slightly orthogonal to splitting the code.
A Backlinks field is a kind of ephemeral field, I don't see this proposal addressing this?

One idea would be to have that concept directly in the schema definition. Like a kind of shortcut. That would also enable dynamic realms to take advantage of it, instead of always taking the long route with `linkingObjects(Person.dog).name`:

```
// Just throwing ideas at the wall
//No idea how the concept should be named.
dynamicRealmSchema.setLinkingObjectField("owners", LinkingObjects(dynamicRealmSchema, "Person", "dog"));


// Now a dynamic query can use the same syntax as a typed Realm.
dynamicRealm.where("Dog").equalTo("owners.name", "John")

```

realm/realm-library/src/main/java/io/realm/Realm.java
bmeike: As far as I can tell, that parameter is never used.  Did I miss something?

realm/realm-library/src/main/java/io/realm/RealmSchema.java
bmeike: Yeah, exactly.  And this is precisely the question I hoped would come up.  I looked at the places that the variable `realm` was assigned, and I don't think it is possible that the value changes, outside the constructor.  If I've got that wrong, none of this stuff will work.

realm/realm-library/src/main/java/io/realm/BaseRealm.java
bmeike: Excellent.  Fixing.  Sadly, with the way our code is organized, package and protected visibility are, essentially, exactly the same.

realm/realm-library/src/main/cpp/CMakeLists.txt
bmeike: 1) Where is that agreement??  ;-P  I can, totally, fix, but don't see either any examples of types whose name begins with OS or of documentation that prescribes it.
2) Move which to "internal"?  Just `RealmNativeSchema`?  I think that is doable!

realm/realm-library/src/main/java/io/realm/RealmBaseSchema.java
bmeike: @beeender I don't exactly understand why this works the way it does, either.  I suspect it is because it is in transition.  I suspect that, over time, functionality will be moved to the `RealmNativeSchema` class.  Eventually the RealmSchema class will shrivel up and die.  When it does, the abstract base class can go away too.

realm/realm-library/src/main/java/io/realm/RealmBaseSchema.java
bmeike: Fixed

realm/realm-library/src/main/java/io/realm/RealmNativeSchema.java
bmeike: Yeah... I had guessed at that, but I wasn't positive.  It looked to me as if the code that ended up in `RealmNativeSchema didn't support DynamicRealm much.  Glad to get the confirmation.  How should it work?  How should it work until `ObjectStoreSchema` supports `DynamicRealm`?

Fixed the CR date.

realm/realm-library/src/main/java/io/realm/RealmNativeSchema.java
bmeike: Yep.  I spent as little time as I could, doing this.  I can think of no other way to get team input on a significant change.  It is just a trial balloon, though...

realm/realm-library/src/main/java/io/realm/RealmBaseSchema.java
bmeike: Yes!!  That's a huge improvement!  Excellent call.

bmeike: The relationship between this change and the backlinks code is ephemeral but present.  The information about the backlinked fields, necessary for querying, has to live in some schema-like-object.  With the current organization, two implementations separated by if-then-elses, it was difficult for me (still trying to understand this thing) to figure out which of the two schema-like objects to choose: the `RealmObjectSchema`, or the `ColumnInfo` object.  Once I'd done the work to figure it out I thought I'd propose this code change, which will make this sort of change (which seems likely to happen a lot, as we move to the ObjectStore) much easier to do.

realm/realm-library/src/main/cpp/CMakeLists.txt
cmelchior: Perhaps it was a Slack discussion, but #4215 captures the conclusion. I think it sprang from the fine-grained single object notifications that needed a wrapper around Object Stores `Object` class, which obviously doesn't work. `NativeObject` was already taken by our phantom reference interface so prefixing all wrappers around Object Store classes with `Os*` seemed the next best thing.

#4215 Just captured the conclusion.

bmeike:
Many thanks for the comments, gents.  Really appreciated.

realm/realm-library/src/main/java/io/realm/Realm.java
bmeike: I did miss something.  Fixing.

kneth:
Should we have: `BaseRealm` + `BaseRealmSchema`, `Realm` + `RealmSchema` and `DynamicRealm` + `DynamicRealmSchema` (I don't fancy `StandardRealmSchema` to be honest).

Zhuinden:
I think that makes sense.

cmelchior:
Except it would break the public API, `RealmSchema` is the current public class and should remain so, that means the base abstract class must be `RealmSchema`. That the typed schema class then gets a slightly less readable name is just unfortunate. Perhaps `TypedRealmSchema` would be better?

realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -85,7 +85,7 @@ public void onSchemaVersionChanged(long currentVersion) {
                                 RealmCache.updateSchemaCache((Realm) BaseRealm.this);
                             }
                         }, true);
-        this.schema = new RealmSchema(this);
+        this.schema = new StandardRealmSchema(this);
```
cmelchior: Shouldn't this be delegated to subclasses for construction? Since right now a DynamicRealm would also get this schema?

realm/realm-library/src/main/java/io/realm/OsRealmSchema.java
```diff
@@ -0,0 +1,159 @@
...
+
+/**
+ * Class for interacting with the Realm schema using a dynamic API. This makes it possible
+ * to add, delete and change the classes in the Realm.
+ * <p>
+ * All changes must happen inside a write transaction for the particular Realm.
+ *
+ * @see RealmMigration
+ */
+public class OsRealmSchema extends RealmSchema {
```
cmelchior: Making this class public makes it part of the public API since it lives in `io.realm` which I doubt we want? Can it be moved to `io.realm.internal`, otherwise it should be made package private.

realm/realm-library/src/main/java/io/realm/OsRealmSchema.java
```diff
@@ -0,0 +1,159 @@
...
+            schema.put(className, realmObjectSchema);
+            return realmObjectSchema;
+        }
+
+        @Override
+        public boolean contains(String className) {
+            return schema.containsKey(className);
+        }
+    }
+
```
cmelchior: To much whitespace? Did you run the formatter?

realm/realm-library/src/main/java/io/realm/RealmSchema.java
```diff
@@ -34,324 +31,82 @@
...
-    public void close() {
-        if (nativePtr != 0) {
-            Set<RealmObjectSchema> schemas = getAll();
-            for (RealmObjectSchema schema : schemas) {
-                schema.close();
-            }
-            nativeClose(nativePtr);
-        }
-    }
+    public abstract void close();
```
cmelchior: Since all these methods are what users will see, it is important they have proper Javadoc. It is actually less important for the actual implementation classes.

kneth:
@cmelchior I see your point. We can leave it as it is, and change the names to a consistent naming convention in version 4.0.0.

realm/realm-library/src/main/java/io/realm/OsRealmSchema.java
bmeike: I will package protect it.  I tried to move it to `internal`, (that is, surely, the right thing to do), but it is so entangled with other package protected stuff that it is just too much for one commit.

realm/realm-library/src/main/java/io/realm/RealmSchema.java
bmeike: Right!  Of course!

realm/realm-library/src/main/java/io/realm/BaseRealm.java
bmeike: I think it should be.  I tried to do it and the number of changes required was just huge.  I'm leaving myself a note that this *does* need to happen.  If you can stand putting it off for a bit, we can get this hunk in...

realm/realm-library/src/main/java/io/realm/OsRealmSchema.java
bmeike: Huh!  I did!  Fixing!

realm/realm-library/src/main/java/io/realm/BaseRealm.java
cmelchior: Sure. That's fine

bmeike:
I don't like `StandardRealmSchema` either.  It is in *no* way, descriptive.  In fact, it is probably anti-descriptive, since it is the one we will, most likely, deprecate.  I tried `LegacyRealmSchema`, and `JavaRealmSchema` and couldn't abide either of them.  I welcome suggestions!!

bmeike:
How embarrassing....   Fixing.

realm/realm-library/src/main/java/io/realm/OsRealmSchema.java
bmeike: Ok.  Making the class private required a fairly awful kludge, or some very complicated restructuring.  I opted for the kludge and a comment.  It needs to be fixed, but at least there is nothing new in the public API.

kneth:
@bmeike Maybe `PlainOldRealmSchema` 😄

Zhuinden:
Seriously why not TypedRealmSchema?

kneth:
`TypedRealmSchema` can do for now, and we can sort it out when we get to 4.0.0 (which should be month into the future).

bmeike:
Woo!  Thank you!

bmeike:
Akk.  This is failing on CI, with a native exception, that I can't repro on my laptop.

PR #4349, opened 2017-03-20
Sync Progress Notifications

cmelchior:
Fixes #4104 

This PR adds support for Sync Progress Listeners. The public API mirrors the Cocoa one with one exception. Instead of having a `Direction` enum, I added a two methods `addDownloadProgressListener` and `addUploadProgressListener`. It reduces the complixity of the method and is generally considered good practise: https://martinfowler.com/bliki/FlagArgument.html

TODO
- [x] Finalize public API
- [x] Cleanup Javadoc
- [x] Implement integration tests.
- [x] Add sample implementation to the `objectServerExample`

Needs fixing (possible in other PR's):
- [ ] Figure out how to handle ProgressListeners being added before the Realm instance is open. Right now it isn't possible, which is an odd restriction. Lifting it seems to require unbundling the session life cycle from Realm instances. We probably want to do that anyway though.

PR #4415, opened 2017-04-03
Re-add Sync Progress Notifications 

cmelchior:
Blocked by https://github.com/realm/realm-sync/issues/1273#issuecomment-291092074
Original PR: #4349


monajafi:
Hi @cmelchior 
I'm looking everyday to this feature's progress, but unfortunately it seems will take time.
When do you estimate conflict will be resolved?

cmelchior:
Fixing this require a protocol breaking change for the Sync Protocol, it is being postponed to 3.3 were a bunch of other protocol changes will happen

Zhuinden:
Protocol break? Does this have something to do with the stable IDs?

cmelchior:
Not really, but fixing this required changes to the sync protocol as it needs to send some more information. That Stable ID's also break some things is just a coincidence (but also why we want to release both together)

cmelchior:
Ready for final review. The bug that caused this to be pulled last has been fixed in Core/OS.
Note most (if not all review) has already been done in #4349

realm/realm-library/src/objectServer/java/io/realm/SyncSession.java
```diff
@@ -115,6 +136,105 @@ void notifySessionError(int errorCode, String errorMessage) {
         }
     }
 
+    // Called from native code
```
zaki50: Native code is calling `SyncManager#notifyProgressListener()` instead of this.
Shouldn't we move this comment and following `@KeepMember` annotation to `SyncManager#notifyProgressListener()`?

realm/realm-library/src/androidTestObjectServer/java/io/realm/SessionTests.java
```diff
@@ -61,6 +61,54 @@ public void get_syncValues() {
         assertEquals(configuration, session.getConfiguration());
     }
 
+    @Test
+    public void addDownloadProgressListener_nullThrows() {
+        SyncSession session = SyncManager.getSession(configuration);
+        try {
```
kneth: Can't `ExpectedException` be used?

realm/realm-library/src/syncIntegrationTest/java/io/realm/BaseIntegrationTest.java
```diff
@@ -104,4 +103,4 @@ private static void deleteFile(File file) throws IOException {
             throw new IllegalStateException("Failed to delete file or directory: " + file.getAbsolutePath());
         }
     }
-}
+}
```
zaki50: We should have CR at the end of the file.

realm/realm-library/src/objectServer/java/io/realm/SyncManager.java
```diff
@@ -279,6 +279,24 @@ private static synchronized void notifyNetworkIsBack() {
...
+     * find the progress listener), but it means we don't have to cache anything on the C++ side which
+     * can leak since we don't have control over the session lifecycle.
+     */
+    @SuppressWarnings("unused")
+    private static synchronized void notifyProgressListener(String localRealmPath, long listenerId, long transferedBytes, long transferableBytes) {
+        SyncSession session = sessions.get(localRealmPath);
+        if (session != null) {
+            try {
+                session.notifyProgressListener(listenerId, transferedBytes, transferableBytes);
+            } catch (Exception exception) {
```
beeender: If the exception has been caught here, how could https://github.com/realm/realm-java/pull/4415/files#diff-e7e3b3778e84279d269f220324c75ccfR92 return true?

realm/realm-library/src/objectServer/java/io/realm/SyncManager.java
```diff
@@ -91,7 +91,7 @@ public void onError(SyncSession session, ObjectServerError error) {
         }
     };
     // keeps track of SyncSession, using 'realm_path'. Java interface with the ObjectStore using the 'realm_path'
-    private static Map<String, SyncSession> sessions = new HashMap<String, SyncSession>();
+    private static Map<String, SyncSession> sessions = new ConcurrentHashMap<>();
```
beeender: Even the newly added `notifyProgressListener` is synchronized on `SyncManager.class`, why do you need to use `ConcurrentHashMap`?

realm/realm-library/src/main/cpp/io_realm_internal_Util.cpp
```diff
@@ -56,6 +56,10 @@ JNIEXPORT jint JNICALL JNI_OnLoad(JavaVM* vm, void*)
         java_lang_double_init = env->GetMethodID(java_lang_double, "<init>", "(D)V");
         java_util_date = GetClass(env, "java/util/Date");
         java_util_date_init = env->GetMethodID(java_util_date, "<init>", "(J)V");
+#if REALM_ENABLE_SYNC
+        java_syncmanager_class = GetClass(env, "io/realm/SyncManager");
```
kneth: Can we use our `realm::jni_util::JavaMethod` for this?

realm/realm-library/src/objectServer/java/io/realm/SyncSession.java
```diff
@@ -115,6 +136,105 @@ void notifySessionError(int errorCode, String errorMessage) {
...
+    @SuppressWarnings("unused")
+    @KeepMember
```
beeender: This is NOT called from native code i think?

realm/realm-library/src/objectServer/java/io/realm/Progress.java
```diff
@@ -0,0 +1,131 @@
...
+     *
+     * @return a number between {@code 0.0} and {@code 1.0}, where {@code 0.0} represents that no data has been
+     *         transferred yet, and {@code 1.0} that all data has been transferred.
+     */
+    public double getFractionTransferred() {
+        if (transferableBytes == 0) {
+            return 1.0D;
+        } else {
+            double percentage = (double) transferredBytes / (double) transferableBytes;
+            return percentage > 1.0D ? 1.0D : percentage;
```
kneth: Really?

realm/realm-library/src/main/cpp/io_realm_SyncSession.cpp
```diff
@@ -28,22 +29,21 @@
...
 
-using namespace std;
 using namespace realm;
 using namespace jni_util;
 using namespace sync;
 
 JNIEXPORT jboolean JNICALL Java_io_realm_SyncSession_nativeRefreshAccessToken(JNIEnv* env, jclass,
-                                                                              jstring localRealmPath,
-                                                                              jstring accessToken,
+                                                                              jstring j_local_realm_path,
```
cmelchior: It stands for `java`. It is mostly a work-around so I don't have to different names for the same thing when the first I do is call `JStringAccessor`.

If it is confusing I guess I can go back to finding somewhat less sensible names for the method params.

cmelchior: Not having it on `sync_realm_url` was a mistake.

realm/realm-library/src/objectServer/java/io/realm/SyncSession.java
```diff
@@ -53,6 +59,8 @@
 public class SyncSession {
     private final static ScheduledThreadPoolExecutor REFRESH_TOKENS_EXECUTOR = new ScheduledThreadPoolExecutor(1);
     private final static long REFRESH_MARGIN_DELAY = TimeUnit.SECONDS.toMillis(10);
+    private final static int DIRECTION_DOWNLOAD = 1;
+    private final static int DIRECTION_UPLOAD = 2;
```
cmelchior: Have any good ideas for what that class might be. I hadn't which was why I settled on this.

realm/realm-library/src/objectServer/java/io/realm/SyncSession.java
```diff
@@ -115,6 +136,105 @@ void notifySessionError(int errorCode, String errorMessage) {
...
+    // Called from native code
+    @SuppressWarnings("unused")
+    @KeepMember
+    synchronized void notifyProgressListener(long listenerId, long transferredBytes, long transferableBytes) {
+        Pair<ProgressListener, Progress> listener = listenerIdToProgressListenerMap.get(listenerId);
+        if (listener != null) {
+            Progress newProgressNotification = new Progress(transferredBytes, transferableBytes);
+            if (!newProgressNotification.equals(listener.second)) {
+                listener.first.onChange(newProgressNotification);
+                listener.second = newProgressNotification;
```
cmelchior: Good catch. Yeah, I should assign it first. We catch exceptions in the method calling this.

realm/realm-library/src/objectServer/java/io/realm/SyncSession.java
```diff
@@ -115,6 +136,105 @@ void notifySessionError(int errorCode, String errorMessage) {
...
+     * Adds a progress listener tracking changes that need to be downloaded from the Realm Object
+     * Server.
+     * <p>
+     * The {@link ProgressListener} will be triggered immediately when registered, and periodically
+     * afterwards.
+     *
+     * @param mode type of mode used. See {@link ProgressMode} for more information.
+     * @param listener the listener to register.
+     */
+    public synchronized void addDownloadProgressListener(ProgressMode mode, ProgressListener listener) {
```
cmelchior: I guess it is a tradeoff: Keep creating different monitor types which will be more fine-grained and complicated vs. just using `synchronized` which is more coarse-grained, but less complicated.

I don't see the SyncSession being used in a highly threaded manner, but if you feel strongly about it, I can introduce a specific monitor type?

realm/realm-library/src/main/cpp/io_realm_SyncSession.cpp
```diff
@@ -56,6 +56,69 @@ JNIEXPORT jboolean JNICALL Java_io_realm_SyncSession_nativeRefreshAccessToken(JN
...
+
+            auto path = to_jstring(local_env, local_realm_path);
+            local_env->CallStaticVoidMethod(java_syncmanager_class, java_notify_progress_listener, path, listener_id,
+                                            static_cast<jlong>(transferred), static_cast<jlong>(transferrable));
+
+            // All exceptions will be caught on the Java side of handlers, but errors will still end
+            // up here, so we need to do something sensible with them.
+            // Throwing a C++ exception will terminate the sync thread and cause the pending Java
+            // exception to become visible. For some (unknown) reason Logcat will not see the C++
+            // exception, only the Java one.
```
cmelchior: Exceptions are getting caught on the Java side and converted to a LogCat statement, true. This is about `Error` type exceptions. I believe I saw this with `AssertionError`s. Perhaps the javadoc is unclear?

realm/realm-library/src/objectServer/java/io/realm/Progress.java
cmelchior: Mostly to be extra safe and there is no restriction to `transferredBytes` / `transferableBytes`. If we lift this, we should probably guard the constructor?

realm/realm-library/src/objectServer/java/io/realm/SyncManager.java
cmelchior: Changing progress listeners inside a callback caused ConcurrentModificationExceptions

realm/realm-library/src/objectServer/java/io/realm/SyncManager.java
cmelchior: Errors are not caught by this. We could change this to `Throwable`, but that could potentially mask some very serious errors.

realm/realm-library/src/objectServer/java/io/realm/SyncManager.java
```diff
@@ -285,6 +286,7 @@ private static synchronized void notifyNetworkIsBack() {
      * can leak since we don't have control over the session lifecycle.
      */
     @SuppressWarnings("unused")
+    @KeepMember
```
beeender: `@KeepMember` needs to be on the class as well. but, wait, the class has been already annotated with `@Keep` so this is not necessary ....

realm/realm-library/src/main/java/io/realm/internal/util/Pair.java
```diff
@@ -27,8 +27,8 @@
  * objects.
  */
 public class Pair<F, S> {
-    public final F first;
-    public final S second;
+    public F first;
```
beeender: Why this cannot be `final` anymore? I cannot find code changing it after construction.

realm/realm-library/src/objectServer/java/io/realm/SyncManager.java
beeender: So if an `Error` thrown here, the jvm won't be stopped, instead it will get back to JNI where calls this function and continue there? interesting ...

realm/realm-library/src/main/java/io/realm/internal/util/Pair.java
cmelchior: I'm changing it here: https://github.com/realm/realm-java/pull/4415/files#diff-0bc822b1fa3365c424a52d324b4f0bd4R156

examples/objectServerExample/src/main/java/io/realm/examples/objectserver/CounterActivity.java
```diff
@@ -86,12 +121,21 @@ public void onChange(CRDTCounter counter) {
...
+            session.addDownloadProgressListener(ProgressMode.INDEFINITELY, downloadListener);
+            session.addUploadProgressListener(ProgressMode.INDEFINITELY, uploadListener);
         }
     }
 
     @Override
     protected void onStop() {
         super.onStop();
+        if (session != null) {
+            session.removeProgressListener(downloadListener);
```
cmelchior: The are being added in `onStart`? https://github.com/realm/realm-java/pull/4415/files#diff-f293d30335c94eee737ce7a8c1fd2790R127

realm/realm-library/src/main/cpp/util.cpp
```diff
@@ -45,6 +45,10 @@ jclass java_lang_string;
 jmethodID java_lang_double_init;
 jclass java_util_date;
 jmethodID java_util_date_init;
+#if REALM_ENABLE_SYNC
+jclass java_syncmanager_class;
```
cmelchior: The progress listener callback is happening on a pure C++ thread. It might not be able to lookup those classes correctly according to: https://developer.android.com/training/articles/perf-jni.html#faq_FindClass

realm/realm-library/src/main/cpp/io_realm_internal_Util.cpp
```diff
@@ -56,6 +56,10 @@ JNIEXPORT jint JNICALL JNI_OnLoad(JavaVM* vm, void*)
...
+        java_notify_progress_listener = env->GetStaticMethodID(java_syncmanager_class, "notifyProgressListener", "(Ljava/lang/String;JJJ)V");
```
cmelchior: https://developer.android.com/training/articles/perf-jni.html#faq_FindClass same as below

examples/objectServerExample/src/main/java/io/realm/examples/objectserver/CounterActivity.java
nhachicha: I think the code block is part of `onCreate` https://github.com/realm/realm-java/pull/4415/files#diff-f293d30335c94eee737ce7a8c1fd2790R83

nhachicha: I think the code block is part of `onCreate` https://github.com/realm/realm-java/pull/4415/files#diff-f293d30335c94eee737ce7a8c1fd2790R83

realm/realm-library/src/main/cpp/util.cpp
nhachicha: It's the same situation in https://github.com/realm/realm-java/blob/abf24d36e3354253ba2742c9a36252b9edf1b56e/realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp#L83-L106

you do the lookup (and cache using `static`) in the VM thread, then capture it by copy in the lambda executed in C++

examples/objectServerExample/src/main/java/io/realm/examples/objectserver/CounterActivity.java
cmelchior: It isn't. GitHub is just folding it weirdly. If you see the full source you can see it is `onStart`

realm/realm-library/src/main/cpp/util.cpp
cmelchior: Fixed

realm/realm-library/src/main/cpp/io_realm_internal_Util.cpp
cmelchior: Fixed

examples/objectServerExample/src/main/res/values/realm_colors.xml
```diff
@@ -20,4 +20,9 @@
     <color name="flamingo_darker">#d64881</color>
     <color name="touch_area_pressed">#dadada</color>
 
+    // Progress bar colors
```
dalinaum: Should we use XML style comments such as like `<!-- Progress bar colors -->`.

realm/realm-library/src/main/cpp/io_realm_SyncSession.cpp
```diff
@@ -56,6 +56,72 @@ JNIEXPORT jboolean JNICALL Java_io_realm_SyncSession_nativeRefreshAccessToken(JN
...
+        if (session) {
+            session->unregister_progress_notifier(static_cast<uint64_t>(listener_token));
+        }
+    }
+    CATCH_STD()
+}
+
 JNIEXPORT jboolean JNICALL Java_io_realm_SyncSession_nativeWaitForDownloadCompletion(JNIEnv* env,
                                                                                      jobject session_object,
                                                                                      jstring localRealmPath)
```
dalinaum: Should it be also changed to `j_local_realm_path`?

realm/realm-library/src/objectServer/java/io/realm/Progress.java
```diff
@@ -0,0 +1,139 @@
...
+    public long getTransferableBytes() {
+        return transferableBytes;
+    }
+
+    /**
+     * The fraction of bytes transferred out of all transferable bytes. Counting from since the {@link ProgressListener}
+     * was added.
+     *
+     * @return a number between {@code 0.0} and {@code 1.0}, where {@code 0.0} represents that no data has been
+     *         transferred yet, and {@code 1.0} that all data has been transferred.
```
dalinaum: Should not use indent this line? Since it is not a parameter description.

PR #4421, opened 2017-04-04
Acquire a global ref of jstring for field changes

### LINKED ISSUE #4378: Reach the Local Ref Table size limit when 512+ fields changes for the same object.
Currently the `String` array of changed field names are created from JNI which may reach the local ref table size limits. 

Idea to fix it is creating `JavaString` class and acquire a global ref then release the local ref.

Test to reproduce it:

```java
    @Test
    @RunTestInLooperThread
    public void moreFieldsChangedThanLocalRefTableSize() {
        final String CLASS_NAME = "ManyFields";
        final int FIELD_COUNT = 1024;
        RealmConfiguration config = looperThread.createConfiguration("many_fields");
        DynamicRealm realm = DynamicRealm.getInstance(config);

        realm.beginTransaction();
        RealmObjectSchema objectSchema = schema.create(CLASS_NAME);
        for (int i = 0; i < FIELD_COUNT; i++) {
            objectSchema.addField("field" + i, int.class);
        }
        DynamicRealmObject obj = realm.createObject(CLASS_NAME);
        realm.commitTransaction();

        obj.addChangeListener(new RealmObjectChangeListener<DynamicRealmObject>() {
            @Override
            public void onChange(DynamicRealmObject object, ObjectChangeSet changeSet) {
                assertEquals(FIELD_COUNT, changeSet.getChangedFields().length);
                looperThread.testComplete();
            }
        });

        realm.beginTransaction();
        for (int i = 0; i < FIELD_COUNT; i++) {
            obj.setInt("field" + i, 42);
        }
        realm.commitTransaction();
    }
```

beeender:
When there are more than 512 fields change, the JNI local ref table size
limitation may be reached. Fix #4378

### DIFF
--- a/realm/realm-library/src/main/cpp/io_realm_internal_OsObject.cpp
+++ b/realm/realm-library/src/main/cpp/io_realm_internal_OsObject.cpp
@@ -70,7 +70,8 @@ struct ChangeCallback {
             return;
         }
 
-        std::vector<jstring> field_names;
+        // The local ref of jstring needs to be released to void reach the local ref table size limitation.
+        std::vector<JavaGlobalRef> field_names;
         auto table = m_wrapper->m_object.row().get_table();
         for (size_t i = 0; i < change_set.columns.size(); ++i) {
             if (change_set.columns[i].empty()) {
@@ -78,11 +79,11 @@ struct ChangeCallback {
             }
             // FIXME: After full integration of the OS schema, parse the column name from
             // wrapper->m_object.get_object_schema() will be faster.
-            field_names.push_back(to_jstring(env, table->get_column_name(i)));
+            field_names.push_back(JavaGlobalRef(env, to_jstring(env, table->get_column_name(i)), true));
         }
         m_field_names_array = env->NewObjectArray(field_names.size(), java_lang_string, 0);
         for (size_t i = 0; i < field_names.size(); ++i) {
-            env->SetObjectArrayElement(m_field_names_array, i, field_names[i]);
+            env->SetObjectArrayElement(m_field_names_array, i, field_names[i].get());
         }
     }
 
--- a/realm/realm-library/src/main/cpp/io_realm_internal_OsObject.cpp
+++ b/realm/realm-library/src/main/cpp/io_realm_internal_OsObject.cpp
@@ -70,7 +70,7 @@ struct ChangeCallback {
             return;
         }
 
-        // The local ref of jstring needs to be released to void reach the local ref table size limitation.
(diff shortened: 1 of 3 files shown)


realm/realm-library/src/main/cpp/io_realm_internal_OsObject.cpp
```diff
@@ -70,19 +70,20 @@ struct ChangeCallback {
...
+        std::vector<JavaGlobalRef> field_names;
         auto table = m_wrapper->m_object.row().get_table();
         for (size_t i = 0; i < change_set.columns.size(); ++i) {
             if (change_set.columns[i].empty()) {
                 continue;
             }
             // FIXME: After full integration of the OS schema, parse the column name from
             // wrapper->m_object.get_object_schema() will be faster.
-            field_names.push_back(to_jstring(env, table->get_column_name(i)));
+            field_names.push_back(JavaGlobalRef(env, to_jstring(env, table->get_column_name(i)), true));
```
zaki50: Who will delete created global references?
I think we need to call `DeleteGlobalRef` against elements in `field_names`.

beeender: When `field_names` gets destructed (when it gets out of scope of this function), all the element will be destructed. The `DeleteGlobalRef` is called in the `JavaGlobalRef`'s destructor.

realm/realm-library/src/main/cpp/io_realm_internal_OsObject.cpp
```diff
@@ -70,19 +70,20 @@ struct ChangeCallback {
             return;
         }
 
-        std::vector<jstring> field_names;
+        // The local ref of jstring needs to be released to void reach the local ref table size limitation.
```
dalinaum: to avoid?

realm/realm-library/src/main/cpp/io_realm_internal_OsObject.cpp
zaki50: I couldn't find `DeleteGlobalRef` in the destructor of `JavaGlobalRef`.

beeender: @zaki50 here https://github.com/realm/realm-java/blob/mc/bug/4378/realm/realm-library/src/main/cpp/jni_util/java_global_ref.cpp#L27

zaki50: Ah, I was reading wrong file.

Yes, `DeleteGlobalRef` is called!

PR #4422, opened 2017-04-04
Fix OsRealmSchema leak

beeender:


realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -411,6 +411,7 @@ private static void initializeSyncedRealm(Realm realm) {
                 realm.setVersion(newVersion);
                 commitChanges = true;
             }
+            schema.close();
```
cmelchior: Shouldn't the schema lifecycle follow the Realm instance lifecycle instead? E.g. release it when we release the last instance?

beeender: I have no idea, it seems OsRealmSchema is used for sync schema updating only?

zaki50: If we need `schema.close()` here, we also need to close it at line 405.
Shouldn't we use `try - finally`?

cmelchior: Yes, right now, but `close()` is on the abstract class, and functionality will probably shift between the two classes, so we cannot make any assumptions

PR #4435, opened 2017-04-05
Merge releases back into master

cmelchior:
Not sure why this wasn't done automatically, so here is a manual PR


beeender:
you need to solve the conflict

Issue #4279, opened 2017-03-02
Support transient keyword


cmelchior:
From this article https://medium.com/@ffvanderlaan/realm-auto-updated-objects-what-you-need-to-know-b2d769d12d76#.hzmdv62a9

Would it make sense to allow the `transient` keyword and just treat any field with `transient` as having `@Ignore`? On the top of my head I cannot come with any negatives, but plenty of positives:

* `transitive` literally means "don't serialize this" = "Don't save to disk" = "@Ignore"
* Better support for GSON.
* Closer to language semantics instead of using custom annotations.
* Open question: Should we deprecate/remove `@Ignore` ?

Some good counter points here http://stackoverflow.com/questions/2154622/why-does-jpa-have-a-transient-annotation, but I'm not entirely convinced they apply to the Android world.

```
public class Person extends RealmObject {
  // Today
  @Ignore
  public String foo;

  // With change
  public transient String foo;
}
```

Thoughts @realm/java ?

**TODO**
- [ ] Decide if `@Ignore` should be deprecated
- [ ] Modify annotation processor so `transient` are treated as `@Ignore`
- [ ] Unit tests

zaki50:
Allowing  `transient` and treating it as `@Ignore` is :+1:, but I don't think we should remove `@Ignore`.

cmelchior:
Merged in #4436

Zhuinden:
Awesome!

AntonAndreevEpam:
Awesome, but I have case  where I am using `transient` to remove concrete field from GSON serialization and at the same moment I required that field in database. So right now that case is broken because, if there is a transient keyword on field, Realm ignores it even if I add `@Required` annotation

Zhuinden:
No, you are just not using `@Expose` and trying to hack around it.

Use `@Expose`, Realm works just fine this way.

eygraber:
Using `@Expose` is not the only way of doing things with GSON; it's equally valid to take the approach of including everything, and excluding unwanted fields using `transient`, and I'm sure that most use cases take this approach.

Realm shouldn't make interop with GSON hinge on forcing use of `@Expose`.

Zhuinden:
Hmm it's been a year but I think I might have been rash; although in general I wouldn't use RealmObjects as the descriptors of a JSON API.

[...]

PR #4436, opened 2017-04-05
Add support for transient fields

cmelchior:
While looking into how to best support databinding, I ran across #4279 , which was easily fixable.
It doesn't make it possible to extend `BaseObservable` yet, but it is a step in the right direction.


realm/realm-annotations-processor/src/main/java/io/realm/processor/ClassMetaData.java
```diff
@@ -374,40 +360,43 @@ private boolean checkForVolatileFields() {
...
 
-        if (variableElement.getAnnotation(Ignore.class) != null) { return true; }
+        // Ignore fields marked with @Ignore or if they are transient
+        if (field.getAnnotation(Ignore.class) != null || field.getModifiers().contains(Modifier.TRANSIENT)) {
+            return true;
+        }
 
-        if (variableElement.getAnnotation(Index.class) != null) {
-            if (!categorizeIndexField(element, variableElement)) { return false; }
+        if (field.getAnnotation(Index.class) != null) {
```
bmeike: would this be clearer if it were an &&, as the previous test is an ||

cmelchior:
Good catch @zaki50. I'll adjust that.

cmelchior:
Updated @zaki50

PR #4444, opened 2017-04-06
Hide StandardRealmSchema class from public API.

### LINKED ISSUE #4443: realm getSchema incorrect return type
realm.getSchema() function return object from type StandardRealmSchema 
this type is object have packcge only access so it is impossible to access his methods,
the only way is to convert this object to (RealmSchema), i think it is mistake 
the return type need to be RealmSchema and not StandardRealmSchema 

realm version in use :3.1.0

zaki50:
fixes #4443 

@realm/java

### DIFF
--- a/realm/realm-library/src/main/java/io/realm/OsRealmSchema.java
+++ b/realm/realm-library/src/main/java/io/realm/OsRealmSchema.java
@@ -196,12 +196,12 @@ class OsRealmSchema extends RealmSchema {
     }
 
     @Override
-    RealmObjectSchema getSchemaForClass(Class<? extends RealmModel> clazz) {
+    OsRealmObjectSchema getSchemaForClass(Class<? extends RealmModel> clazz) {
         throw new UnsupportedOperationException();
     }
 
     @Override
-    StandardRealmObjectSchema getSchemaForClass(String className) {
+    OsRealmObjectSchema getSchemaForClass(String className) {
         throw new UnsupportedOperationException();
     }
 
--- a/realm/realm-library/src/main/java/io/realm/OsRealmSchema.java
+++ b/realm/realm-library/src/main/java/io/realm/OsRealmSchema.java
@@ -21,6 +21,8 @@ import java.util.LinkedHashSet;
 import java.util.Map;
 import java.util.Set;
 
+import io.realm.internal.Table;
+
 
 /**
  * Class for interacting with the Realm schema using a dynamic API. This makes it possible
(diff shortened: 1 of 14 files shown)


zaki50:
To save method count, should we use `getSchema()` and cast it to `StandardRealmSchema` when using it internally?

Zhuinden:
That seems scary long-term.

bmeike:
I totally agree with @Zhuinden: not using Java (especially typing) in the interests of method count or some other Colt McAnliss-y metric is, absolutely, misguided (a variety of premature) optimization.

zaki50:
@bmeike 👍

zaki50:
@bmeike updated.

realm/realm-library/src/main/java/io/realm/OsRealmSchema.java
```diff
@@ -62,6 +64,26 @@ public boolean contains(String className) {
...
+            throw new UnsupportedOperationException();
+        }
+
+        @Override
+        RealmObjectSchema getSchemaForClass(Class<? extends RealmModel> clazz) {
+            throw new UnsupportedOperationException();
+        }
+
+        @Override
+        StandardRealmObjectSchema getSchemaForClass(String className) {
```
cmelchior: This should just return `RealmObjectSchema`?

realm/realm-library/src/main/java/io/realm/OsRealmSchema.java
```diff
@@ -163,6 +185,26 @@ public boolean contains(String className) {
...
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    RealmObjectSchema getSchemaForClass(Class<? extends RealmModel> clazz) {
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    StandardRealmObjectSchema getSchemaForClass(String className) {
```
cmelchior: Same

realm/realm-library/src/main/java/io/realm/RealmSchema.java
```diff
@@ -145,4 +145,9 @@ private void checkIndices() {
             throw new IllegalStateException("Attempt to use column index before set.");
         }
     }
+
+    abstract Table getTable(Class<? extends RealmModel> clazz);
+    abstract Table getTable(String className);
+    abstract RealmObjectSchema getSchemaForClass(Class<? extends RealmModel> clazz);
+    abstract StandardRealmObjectSchema getSchemaForClass(String className);
```
cmelchior: Return `RealmObjectSchema`?

zaki50: Right!

realm/realm-library/src/main/java/io/realm/OsRealmSchema.java
zaki50: This should be `OsRealmObjectSchema`.
Narrowing return type from parent class is valid since Java5.

realm/realm-library/src/main/java/io/realm/OsRealmSchema.java
zaki50: This should be `OsRealmObjectSchema`.

zaki50:
@cmelchior fixed.

realm/realm-library/src/main/java/io/realm/OsRealmSchema.java
cmelchior: Missed this

zaki50: done.

beeender:
please merge this after #4450 merged. thanks

bmeike:
Thanks for doing that.  I think it will give us a better clue about what we need to implement for OsRealmsSchema.

PR #4455, opened 2017-04-08
fix API incompatibility introduced in 3.1.0

### LINKED ISSUE #4443: realm getSchema incorrect return type

zaki50:
fixes #4443 

I fixed incompatibility of `RealmSchema.getAll()` in addition to `BaseRealm.getSchema()`.

@realm/java

### DIFF
--- a/realm/realm-library/src/main/java/io/realm/OsRealmSchema.java
+++ b/realm/realm-library/src/main/java/io/realm/OsRealmSchema.java
@@ -34,7 +34,7 @@ import io.realm.internal.Table;
  */
 class OsRealmSchema extends RealmSchema {
     static final class Creator extends RealmSchema {
-        private final Map<String, OsRealmObjectSchema> schema = new HashMap<>();
+        private final Map<String, RealmObjectSchema> schema = new HashMap<>();
 
         @Override
         public void close() { }
@@ -46,7 +46,7 @@ class OsRealmSchema extends RealmSchema {
         }
 
         @Override
-        public Set<OsRealmObjectSchema> getAll() {
+        public Set<RealmObjectSchema> getAll() {
             return new LinkedHashSet<>(schema.values());
         }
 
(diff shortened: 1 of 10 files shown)


realm/realm-library/src/main/java/io/realm/OsRealmSchema.java
```diff
@@ -34,7 +34,7 @@
  */
```
bmeike: I don't think you should change this class.  `OsRealmSchema` can not handle generic `RealmObjectSchema`.  It is only able to handle `OsRealmObjectSchema`.  It should use typing to make that clear.

realm/realm-library/src/main/java/io/realm/RealmSchema.java
```diff
@@ -54,7 +54,7 @@
      *
      * @return the set of all classes in this Realm or no RealmObject classes can be saved in the Realm.
      */
-    public abstract Set<? extends RealmObjectSchema> getAll();
```
bmeike: I believe this is correct as it stands.

bmeike:
Thanks for fixing this.

bmeike:
Hey @zaki50, let me know if you agree with my suggestion that we ought to keep the more specific type in OsRealmSchema.  Also, happy to discuss if you want.  A million thanks for fixing these bugs I introduced.

realm/realm-library/src/main/java/io/realm/RealmSchema.java
zaki50: The point is that we can't introduce any breaking changes without bumping major version.

Changing return type from `Set <RealmObjectSchema>` to `Set<? extends RealmObjectSchema>` is a breaking change. Users might have a code `Set<RealmObjectSchema> allObjectSchema = realm.getSchema().getAll();` and this line causes compile error in 3.1.

bmeike: Yep, I guess that is right.  That is a serious bummer.

PR #4465, opened 2017-04-11
My/fix memory leaking

zaki50:
Fixed memory leaks when synced Realm was initialize.
And it also contains some refactorings.

Those changes are split into small commits. I recommend to review each commit instead of entire change.

@realm/java

### DIFF
--- a/realm/realm-library/src/main/java/io/realm/Realm.java
+++ b/realm/realm-library/src/main/java/io/realm/Realm.java
@@ -396,13 +396,18 @@ public class Realm extends BaseRealm {
             final RealmProxyMediator mediator = configuration.getSchemaMediator();
             final Set<Class<? extends RealmModel>> modelClasses = mediator.getModelClasses();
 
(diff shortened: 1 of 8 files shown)


realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -396,13 +396,18 @@ private static void initializeSyncedRealm(Realm realm) {
...
+                for (Class<? extends RealmModel> modelClass : modelClasses) {
+                    mediator.createRealmObjectSchema(modelClass, schemaCreator);
+                }
 
-            // Assumption: When SyncConfiguration then additive schema update mode.
-            schema = new OsRealmSchema(schemaCreator);
+                // Assumption: When SyncConfiguration then additive schema update mode.
+                schema = new OsRealmSchema(schemaCreator);
+            } finally {
+                schemaCreator.close();
```
beeender: move this `schemaCreator.close();` to line 443? We have a finally block there already.

PR #4491, opened 2017-04-13
added field name information to exception messages

### LINKED ISSUE #4484: When inserting an object with a null required field: say which one
#### Goal

Currently, when trying to insert an object that has a `null` field which is required, the exception looks like this:

```
java.lang.IllegalArgumentException: Illegal Argument: This field is not nullable.
```
Unfortunately my object has many fields, and so it's a waste of time trying to debug what the issue is.

#### Expected Results

Please change the message to something like `The field 'fooBar' is not nullable.`

Realm version(s): 3.1.0

Android Studio version: 2.4 Preview 5

Which Android version and device: Emulator API level 23

zaki50:
fixes #4484

@realm/java

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,14 +3,11 @@
 ### Bug Fixes
 
 * `equals()` and `hashCode()` of managed `RealmObject`s that come from linking objects don't work correctly (#4487).
+* Field name was missing in exception message when `null` was set to required field (#4484).
 
 ### Internal
 
 * Upgraded to Realm Sync 1.5.2.
-
-### Bug Fixes
-
-* Field name was missing in exception message when `null` was set to required field (#4484).
 
 ## 3.1.2 (2017-04-12)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 * Upgraded to Realm Sync 1.5.2.
 
+### Bug Fixes
+
+* Field name was missing in exception message when `null` was set to required field (#4484).
+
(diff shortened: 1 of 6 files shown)


zaki50:
@realm/java

realm/realm-library/src/main/cpp/util.hpp
```diff
@@ -343,7 +343,7 @@ inline bool TypeValid(JNIEnv* env, T* pTable, jlong columnIndex, int expectColTy
     int colType = pTable->get_column_type(col);
     if (colType != expectColType) {
         realm::jni_util::Log::e("Expected columnType %1, but got %2.", expectColType, pTable->get_column_type(col));
-        ThrowException(env, IllegalArgument, "ColumnType invalid.");
+        ThrowException(env, IllegalArgument, "ColumnType of '" + ((std::string) pTable->get_column_name(col)) + "' is invalid.");
```
beeender: shall we use `std::string(pTable->get_column_name(col))` instead to avoid c-style casting?

zaki50: right!

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@
 
 * Upgraded to Realm Sync 1.5.2.
 
+### Bug Fixes
```
beeender: It seems the changelog is from the latest. But why didn't github show a conflict warning?

zaki50: I've merged latest `releases`.

zaki50:
Thanks. I'll merge this after CI build.

PR #4492, opened 2017-04-13
Introduce DynamicRealmObject#linkingObjects(String srcClassName, String srcFieldName)

zaki50:
TODO: 
- [x] need to discuss whether we introduce this API.
- [x] more tests
- [x] add a check if the object is attached or not (after the merge of #4500).
- [x] CHANGELOG


cmelchior:
I like the idea 👍

Zhuinden:
Hmm @cmelchior this is for defining named backlinks, right? I assume you don't need a migration for unnamed backlinks, because it's automanaged

zaki50:
Not for defining named backlinks, but it returns actual result of backlinks specified by `srcClassName` and `srcFieldName`.

Zhuinden:
Oh! I see now, thanks

zaki50:
@realm/java If there are any reasons to avoid to introduce this, please comment it to this PR.

zaki50:
@realm/java ready for review

realm/realm-library/src/androidTest/java/io/realm/LinkingObjectsDynamicTests.java
```diff
@@ -20,34 +20,484 @@
...
+        }
+
+        final DynamicRealmObject target3 = dynamicRealm.where(BacklinksTarget.CLASS_NAME).equalTo(BacklinksTarget.FIELD_ID, 3).findFirst();
+        final RealmResults<DynamicRealmObject> target3Sources = target3.linkingObjects(BacklinksSource.CLASS_NAME, BacklinksSource.FIELD_CHILD);
+        assertNotNull(target3Sources);
+        assertTrue(target3Sources.isEmpty());
+    }
+
+    @Test
+    public void linkingObjects_OBJECT_notDefinedInModel() {
```
cmelchior: Just reading the method name it isn't exactly clear what is being tested. Perhaps rephrase or add a comment about it? 

EDIT: I assume you mean that `notDefinedInModel` refers to the linkingObjects not being defined in the model class, but only accessed dynamically?

realm/realm-library/src/androidTest/java/io/realm/LinkingObjectsDynamicTests.java
```diff
@@ -20,34 +20,484 @@
...
+        final RealmResults<DynamicRealmObject> cat1Owners = cat1.linkingObjects(Owner.CLASS_NAME, Owner.FIELD_CAT);
+        assertNotNull(cat1Owners);
+        assertEquals(numOwnersOfCat1, cat1Owners.size());
+        for (DynamicRealmObject cat1Owner : cat1Owners) {
+            assertEquals(Owner.CLASS_NAME, cat1Owner.getType());
+            assertTrue(cat1Owner.getString(Owner.FIELD_NAME).endsWith("_cat1"));
+            assertEquals(cat1, cat1Owner.getObject(Owner.FIELD_CAT));
+        }
+
+        final DynamicRealmObject cat2 = dynamicRealm.where(Cat.CLASS_NAME).equalTo(Cat.FIELD_NAME, "cat2").findFirst();
```
cmelchior: What does this test that `cat1` doesn't?

realm/realm-library/src/androidTest/java/io/realm/LinkingObjectsDynamicTests.java
```diff
@@ -20,34 +20,484 @@
...
+        assertTrue(cat3Owners.isEmpty());
+    }
+
+    @Test
+    public void linkingObjects_LIST() {
+        //           source100          source200        source300
+        //            //    \\          \\ || //
+        //        target1   target2     target2
+        //
+        //  /  = object ref
```
cmelchior: This isn't used?

realm/realm-library/src/androidTest/java/io/realm/LinkingObjectsDynamicTests.java
```diff
@@ -20,34 +20,484 @@
...
+        assertEquals(AllJavaTypes.CLASS_NAME, target1Sources.first().getType());
+        assertEquals(100L, target1Sources.first().getLong(AllJavaTypes.FIELD_ID));
+        assertTrue(target1Sources.first().getList(AllJavaTypes.FIELD_LIST).contains(target1));
+        assertTrue(target1Sources.first().getList(AllJavaTypes.FIELD_LIST).contains(target2));
+        assertFalse(target1Sources.first().getList(AllJavaTypes.FIELD_LIST).contains(target3));
+
+        // tests sources of target2
+        final RealmResults<DynamicRealmObject> target2Sources = target2.linkingObjects(AllJavaTypes.CLASS_NAME, AllJavaTypes.FIELD_LIST);
+        assertNotNull(target2Sources);
+        // TODO Is ↓ an expected behavior?
```
cmelchior: Yes, the is the defined behaviour

realm/realm-library/src/androidTest/java/io/realm/LinkingObjectsDynamicTests.java
```diff
@@ -20,34 +20,484 @@
...
+            @Override
+            public void execute(Realm realm) {
+                final BacklinksTarget target1 = realm.createObject(BacklinksTarget.class);
+                target1.setId(1);
+
+                final BacklinksSource source = realm.createObject(BacklinksSource.class);
+                source.setChild(target1);
+            }
+        });
+
```
cmelchior: Whitespace

realm/realm-library/src/androidTest/java/io/realm/LinkingObjectsDynamicTests.java
```diff
@@ -20,34 +20,484 @@
...
+            thrown.expect(IllegalStateException.class);
+            target.linkingObjects(BacklinksSource.CLASS_NAME, BacklinksSource.FIELD_CHILD);
+            fail();
+        } finally {
+            dynamicRealm.close();
+        }
     }
 
     @Test
+    @Ignore
```
cmelchior: Why is this being ignored?

realm/realm-library/src/androidTest/java/io/realm/LinkingObjectsUnmanagedTests.java
```diff
@@ -65,41 +65,18 @@ public void copyFromRealm() {
         assertEquals(parent, child.getObjectParents().first());
 
         AllJavaTypes unmanagedChild = realm.copyFromRealm(child);
-        assertEquals(new AllJavaTypes().getObjectParents(), unmanagedChild.getObjectParents());
-    }
-
-    // When managed, an object's backlinks fields get live.
-    @Test
```
cmelchior: Why delete this test?

realm/realm-library/src/main/java/io/realm/DynamicRealmObject.java
```diff
@@ -35,6 +35,7 @@
  */
 @SuppressWarnings("WeakerAccess")
 public class DynamicRealmObject extends RealmObject implements RealmObjectProxy {
+    private static final String MSG_LINK_QUERY_NOT_SUPPORTED = "Field name must not contain '.'";
```
cmelchior: Wouldn't `Queries across relationships are not supported` be more accurate?

realm/realm-library/src/main/java/io/realm/DynamicRealmObject.java
```diff
@@ -925,6 +926,42 @@ public String toString() {
         return sb.toString();
     }
 
+    /**
+     * Returns {@link RealmResults} containing all {@code srcClassName} class objects that have a relationship
+     * to this object from {@code srcFieldName} field.
+     *
```
cmelchior: Also add that an entry is added for each reference, e.g. if the same reference is in a list multiple times, the `src` object will show up here multiple times.

realm/realm-library/src/main/java/io/realm/DynamicRealmObject.java
```diff
@@ -925,6 +926,42 @@ public String toString() {
...
+     * @param srcClassName name of the class returned objects belong to.
+     * @param srcFieldName name of the field to use. Field type must be either {@code io.realm.RealmFieldType.OBJECT} or
+     * {@code io.realm.RealmFieldType.LIST}.
+     * @return the result.
+     */
```
cmelchior: Also list when `IllegalArgumentException` will be thrown

realm/realm-library/src/main/java/io/realm/DynamicRealmObject.java
```diff
@@ -925,6 +926,42 @@ public String toString() {
         return sb.toString();
     }
 
+    /**
+     * Returns {@link RealmResults} containing all {@code srcClassName} class objects that have a relationship
+     * to this object from {@code srcFieldName} field.
+     *
+     * @param srcClassName name of the class returned objects belong to.
+     * @param srcFieldName name of the field to use. Field type must be either {@code io.realm.RealmFieldType.OBJECT} or
```
cmelchior: Perhaps `name of the field in the source class that holds a reference to this object. Field type ...`

realm/realm-library/src/androidTest/java/io/realm/LinkingObjectsDynamicTests.java
zaki50: It doesn't test anything. We need to update this test case once we implement backlinks query.

cmelchior: Perhaps add a `@Ignore("TODO: Enable again when we add support for backlink queries")` then

zaki50: Nice idea!

realm/realm-library/src/androidTest/java/io/realm/LinkingObjectsDynamicTests.java
nhachicha: maybe the links are well defined (owners that only points to this ca)? and not mixed up with the previous cat's owners

realm/realm-library/src/androidTest/java/io/realm/LinkingObjectsDynamicTests.java
nhachicha: @cmelchior  is there a `distinct` that can aggregate the result?

cmelchior: Yes, Backlinks are exposed as a `RealmResults` so you can use `distinct` on it if you want to only get unique references.

realm/realm-library/src/androidTest/java/io/realm/LinkingObjectsDynamicTests.java
```diff
@@ -20,34 +20,484 @@
...
+
+
+        final DynamicRealm dynamicRealm = DynamicRealm.getInstance(looperThread.realmConfiguration);
+        try {
+            final DynamicRealmObject targetAsync = dynamicRealm.where(BacklinksTarget.CLASS_NAME)
+                    .equalTo(BacklinksTarget.FIELD_ID, 1L).findFirstAsync();
+            // precondition
+            assertFalse(targetAsync.isLoaded());
+
+            thrown.expect(IllegalStateException.class);
```
beeender: I am not quite sure `thrown` works in looper thread. also, if it works, then `fail()` at line 453 is not needed.

beeender: Same for the case below.

realm/realm-library/src/androidTest/java/io/realm/LinkingObjectsDynamicTests.java
```diff
@@ -20,34 +20,484 @@
...
+    public void linkingObjects_ClassIsNull() throws Exception {
+        dynamicRealm.beginTransaction();
+        final DynamicRealmObject object = dynamicRealm.createObject(AllJavaTypes.CLASS_NAME, 1L);
+        dynamicRealm.commitTransaction();
+
+        try {
+            object.linkingObjects(null, AllJavaTypes.FIELD_INT);
+            fail();
+        } catch (IllegalArgumentException expected) {
+            final Field messageField = StandardRealmSchema.class.getDeclaredField("EMPTY_STRING_MSG");
```
beeender: just package protect the `EMPTY_STRING_MSG`?

realm/realm-library/src/main/java/io/realm/RealmResults.java
```diff
@@ -67,6 +67,17 @@
                 srcTableType);
     }
 
+    static RealmResults<DynamicRealmObject> createBacklinkResults(DynamicRealm realm, Row row, Table srcTable, String srcFieldName) {
+        if (!(row instanceof UncheckedRow)) {
```
beeender: If this method return a dynamic results always, then the `row` should only be `CheckedRow`. Also, the comments at https://github.com/realm/realm-java/blob/master/realm/realm-library/src/main/java/io/realm/DynamicRealmObject.java#L72 is not correct :P

realm/realm-library/src/androidTest/java/io/realm/LinkingObjectsUnmanagedTests.java
```diff
@@ -53,7 +53,7 @@ public void tearDown() {
         }
     }
 
-    // When unmanaged, an object's backlinks fields have their initialized value (probably null).
```
bmeike: Is it a requirement that the initialized value be null, now?  It certainly is not forced by any code that I wrote...  If it is a requirement, I'd like to see a test that demonstrates that we enforce it.  If it is not a requirement, then I would prefer that we not claim that it is.

realm/realm-library/src/main/java/io/realm/DynamicRealmObject.java
bmeike: Yes.  I think our docs will be much clearer if we stick to a single set of terms.  The "source class" contains the "source field" . They contain the forward reference.  The "target field" in the "target class" contains the backlink to the source class.

realm/realm-library/src/main/java/io/realm/DynamicRealmObject.java
```diff
@@ -925,6 +926,42 @@ public String toString() {
...
+        proxyState.getRow$realm().checkIfAttached();
+
+        final RealmSchema schema = realm.getSchema();
+        final RealmObjectSchema realmObjectSchema = schema.get(srcClassName);
+        if (realmObjectSchema == null) {
+            throw new IllegalArgumentException("Class not found: " + srcClassName);
+        }
+
+        if (srcFieldName != null && srcFieldName.contains(".")) {
+            throw new IllegalArgumentException(MSG_LINK_QUERY_NOT_SUPPORTED);
```
bmeike: There, really, is no need for defining constants for strings, like this.  They will be interned on compile...

realm/realm-library/src/androidTest/java/io/realm/LinkingObjectsUnmanagedTests.java
Zhuinden: Well as far as I know you can't really obtain an "empty RealmResults", and that's the only other thing that would make sense for it to be.

bmeike: @Zhuinden This is certainly legal, as things stand now: 

```
public class InitializedBacklinks extends RealmObject {
    private InitializedBacklinks child;

    @LinkingObjects("child")
    private final RealmResults<InitializedBacklinks> parents;

    public InitializedBacklinks() {
        this((RealmResults<InitializedBacklinks>) null);
    }

    public InitializedBacklinks(Realm realm) {
        this(realm.where(InitializedBacklinks.class).findAll());
    }

    private InitializedBacklinks(RealmResults<InitializedBacklinks> parents) {
        this.parents = parents;
    }

    public InitializedBacklinks getChild() {
        return child;
    }

    public void setChild(InitializedBacklinks child) {
        this.child = child;
    }
}

```

I think that might even be meaningful, when used Dynamically.

realm/realm-library/src/androidTest/java/io/realm/LinkingObjectsDynamicTests.java
bmeike: I'm on that. These are reminders of tests I need to write.  I didn't realize the @Ignore allowed a comment.  I will fix shortly.

realm/realm-library/src/androidTest/java/io/realm/LinkingObjectsUnmanagedTests.java
```diff
@@ -109,9 +86,11 @@ public void copyToAndFromRealm() {
...
         assertEquals(parent, parents.first());
 
+        // backlinks fields in unmanaged object created from managed object are null.
+
         unmanagedChild = realm.copyFromRealm(child);
         assertEquals(unmanagedChild.getFieldId(), 1);
-        assertEquals(new AllJavaTypes().getObjectParents(), unmanagedChild.getObjectParents());
+        assertNull(unmanagedChild.getObjectParents());
 
         RealmResults<AllJavaTypes> queryResults = realm.where(AllJavaTypes.class).equalTo("fieldId", 1).findAll();
```
bmeike: I am unclear on why it is a good idea to change the semantics of these tests.  Did we change the semantics of backlinks???

Zhuinden: I think it was that `RealmResults` field via `copyFromRealm()` is ignored

realm/realm-library/src/androidTest/java/io/realm/LinkingObjectsDynamicTests.java
zaki50: yes!

realm/realm-library/src/androidTest/java/io/realm/LinkingObjectsDynamicTests.java
zaki50: OK, I'm reverting this change.

realm/realm-library/src/androidTest/java/io/realm/LinkingObjectsDynamicTests.java
zaki50: @nhachicha yes. We should check if the only owners who point the cat are returned.
Just checking the owners of `cat1`is enough?

realm/realm-library/src/androidTest/java/io/realm/LinkingObjectsUnmanagedTests.java
zaki50: I deleted it because the code is completely the same with the top half of the next test case.
Should we keep this?

realm/realm-library/src/androidTest/java/io/realm/LinkingObjectsUnmanagedTests.java
zaki50: @bmeike I was not thinking that a backlinks field of unmanaged object can return non-null object.
As you said in other comment, users can keep non-null RealmResults object to backlink field of unmanaged objects.

I'll revert this change.

realm/realm-library/src/main/java/io/realm/DynamicRealmObject.java
zaki50: added

realm/realm-library/src/main/java/io/realm/DynamicRealmObject.java
zaki50: done

realm/realm-library/src/main/java/io/realm/DynamicRealmObject.java
zaki50: It makes easy to check the message in test cases.

realm/realm-library/src/main/java/io/realm/RealmResults.java
zaki50: right. updated

zaki50:
addressed all comments.
ready for the next round. @realm/java

realm/realm-library/src/main/java/io/realm/DynamicRealmObject.java
```diff
@@ -925,6 +926,49 @@ public String toString() {
...
+     * to this object from {@code srcFieldName} field.
+     * <p>
+     * An entry is added for each reference, e.g. if the same reference is in a list multiple times,
+     * the src object will show up here multiple times.
+     *
+     * @param srcClassName name of the class returned objects belong to.
+     * @param srcFieldName name of the field in the source class that holds a reference to this object.
+     *                    Field type must be either {@code io.realm.RealmFieldType.OBJECT} or {@code io.realm.RealmFieldType.LIST}.
+     * @return the result.
+     * @throws IllegalArgumentException if the source field name is {@code null}.
```
cmelchior: or the `srcClassName` doesn't exists.

Zhuinden:
the change log is missing a `

zaki50:
@Zhuinden thanks. fixed in #4608

PR #4497, opened 2017-04-14
Clear pending collection first for findFirstAsync

### LINKED ISSUE #4495: NullPointerException on RealmChangeListener with async queries workflow
#### Goal

By making use of two `RealmChangeListeners` and one async query, my goal is to reutilize objects when possible, or create a new ones if needed.

#### Results

- Before Realm `3.0` this workflow worked well;
- On `3.0.0` it stopped working, because of https://github.com/realm/realm-java/issues/4352;
- On `3.1.1` it throws a `NullPointerException`.

```
java.lang.NullPointerException: Attempt to invoke virtual method 'void io.realm.internal.Collection.removeListener(java.lang.Object, io.realm.RealmChangeListener)' on a null object reference
	at io.realm.internal.PendingRow.clearPendingCollection(PendingRow.java:205)
	at io.realm.internal.PendingRow.notifyFrontEnd(PendingRow.java:236)
	at io.realm.internal.PendingRow.access$000(PendingRow.java:18)
	at io.realm.internal.PendingRow$1.onChange(PendingRow.java:46)
	at io.realm.internal.PendingRow$1.onChange(PendingRow.java:43)
	at io.realm.internal.Collection$RealmChangeListenerWrapper.onChange(Collection.java:62)
	at io.realm.internal.Collection$CollectionObserverPair.onChange(Collection.java:43)
	at io.realm.internal.Collection$Callback.onCalled(Collection.java:87)
	at io.realm.internal.Collection$Callback.onCalled(Collection.java:77)
	at io.realm.internal.ObserverPairList.foreach(ObserverPairList.java:108)
	at io.realm.internal.Collection.notifyChangeListeners(Collection.java:524)
	at android.os.MessageQueue.nativePollOnce(Native Method)
	at android.os.MessageQueue.next(MessageQueue.java:143)
	at android.os.Looper.loop(Looper.java:122)
	at android.app.ActivityThread.main(ActivityThread.java:5254)
	at java.lang.reflect.Method.invoke(Method.java:372)
	at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:903)
	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:698)
```

- The bug seems to occur when the last `Transaction` tries to call its listeners, after `item.reset();`
- `doThings(item);` is never called.

##### Classes

```java
public class Item extends RealmObject {

    public static Item create(Long id) {
        Item item = new Item();
        item.id = id;
        return item;
    }

    @PrimaryKey
    public Long id;

    public String name;

    public Feed parentFeed;


    public void reset() {
        name = "";
    }
}
```
```java
public class Feed extends RealmObject {

    @PrimaryKey
    public Long id;

    public String name;
}
```

##### Code Sample

```java
final Item item = getRealm().where(Item.class).isNull("parentFeed").findFirstAsync();

// Create a new item, if no valid item is found.
item.addChangeListener(new RealmChangeListener<Item>() {

    @Override
    public void onChange(Item item) {
        if (!item.isValid()) {
            item.removeChangeListener(this);

            getRealm().executeTransaction(new Realm.Transaction() {

                @Override
                    getRealm().copyToRealm(Item.create(getLastId()));
                }
            });
        }
    }
});

// Wait for valid item, reset, and use it to do things.
item.addChangeListener(new RealmChangeListener<Item>() {

    @Override
    public void onChange(final Item item) {
        if (item.isValid()) {
            item.removeChangeListener(this);

            getRealm().executeTransaction(new Realm.Transaction() {

                @Override
                    item.reset();
                }
            });

            doThings(item);
        }
    }
});
```

#### Notes

I could use just a single `RealmChangeListener`, but I'm using Kotlin in my project, and I'm making use of two different extensions functions, which is why I end up using two different listeners.

#### Version of Realm

Realm version(s): 3.1.1

beeender:
Otherwise the pending query will be executed again if there is a local
transaction in the listener. That would cause a infinite recursion or
NPE (like #4495).

Fix #4495

### DIFF
--- a/realm/realm-library/src/main/java/io/realm/internal/PendingRow.java
+++ b/realm/realm-library/src/main/java/io/realm/internal/PendingRow.java
@@ -227,8 +227,8 @@ public class PendingRow implements Row {
         if (pendingCollection.isValid()) {
             // PendingRow will always get the first Row of the query since we only support findFirst.
             UncheckedRow uncheckedRow = pendingCollection.firstUncheckedRow();
-            // Clear the pending collection immediately in case beginTransaction called in the listener which execute
-            // the query again.
+            // Clear the pending collection immediately in case beginTransaction is called in the listener which will
+            // execute the query again
             clearPendingCollection();
             // If no rows returned by the query, notify the frontend with an invalid row.
             if (uncheckedRow != null) {
--- a/realm/realm-library/src/main/java/io/realm/internal/PendingRow.java
+++ b/realm/realm-library/src/main/java/io/realm/internal/PendingRow.java
@@ -222,6 +222,9 @@ public class PendingRow implements Row {
         if (pendingCollection.isValid()) {
             // PendingRow will always get the first Row of the query since we only support findFirst.
             UncheckedRow uncheckedRow = pendingCollection.firstUncheckedRow();
+            // Clear the pending collection immediately in case beginTransaction called in the listener which execute
+            // the query again.
+            clearPendingCollection();
             // If no rows returned by the query, notify the frontend with an invalid row.
             if (uncheckedRow != null) {
                 Row row = returnCheckedRow ? CheckedRow.getFromRow(uncheckedRow) : uncheckedRow;
@@ -231,9 +234,10 @@ public class PendingRow implements Row {
                 // No row matches the query, return a invalid row.
                 frontEnd.onQueryFinished(InvalidRow.INSTANCE);
             }
+        } else {
+            clearPendingCollection();
         }
 
-        clearPendingCollection();
     }
 
     // Execute the query immediately and call frontend's onQueryFinished().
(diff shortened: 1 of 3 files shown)


realm/realm-library/src/main/java/io/realm/internal/PendingRow.java
```diff
@@ -227,6 +227,9 @@ private void notifyFrontEnd() {
         if (pendingCollection.isValid()) {
             // PendingRow will always get the first Row of the query since we only support findFirst.
             UncheckedRow uncheckedRow = pendingCollection.firstUncheckedRow();
+            // Clear the pending collection immediately in case beginTransaction called in the listener which execute
```
cmelchior: is called in the listener which will execute the query again

realm/realm-library/src/main/java/io/realm/internal/PendingRow.java
```diff
@@ -236,9 +239,10 @@ private void notifyFrontEnd() {
                 // No row matches the query, return a invalid row.
                 frontEnd.onQueryFinished(InvalidRow.INSTANCE);
             }
+        } else {
+            clearPendingCollection();
```
cmelchior: The naming of these variables / classes could use a refactor -> `Collection` -> `OsResults` and `clearPendingCollection()` -> `clearPendingOsResults()` but that is being tracked elsewhere

PR #4498, opened 2017-04-14
Sync reconnect

### LINKED ISSUE #4141: Sync Session not retrying upload after Network resume
Steps:
Using RealmTask with Realm `2.3.0` & realm-object-server `1.0.2`

- install RealmTasks in both emulators 
screen shot 2017-02-02 at 14 49 09

- Disable network by using flight mode on one emulator, then create some offline data
screen shot 2017-02-02 at 14 51 21

- Resume network by disabling the flight mode
screen shot 2017-02-02 at 14 54 57

The server doesn't receive any _UPLOAD_  from the client, the session doesn't seem to retry uploading in the background as it should be.

Note that upon a restart (and the creation of a new session) the client sync back correctly the changes created offline. This hints a problem with either the session lifecycle or the retry mechanism in the sync client.

nhachicha:
fixes #4141

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## 3.1.3 (YYYY-MM-DD)
 
+### Enhancements
+
+* [ObjectServer] Resume synchronization as soon as the network is back (#4141).
+
 ### Bug Fixes
 
 * `equals()` and `hashCode()` of managed `RealmObject`s that come from linking objects don't work correctly (#4487).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### Enhancements
 
-* [ObjectServer] Resume synchronization as soon as the network is back (#4141).
+* [ObjectServer] Resume synchronization as soon as the connectivity is back (#4141).
 
 ### Bug Fixes
 
--- a/realm/realm-library/src/objectServer/java/io/realm/SyncManager.java
+++ b/realm/realm-library/src/objectServer/java/io/realm/SyncManager.java
@@ -26,6 +26,7 @@ import java.util.concurrent.TimeUnit;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import io.realm.internal.Keep;
 import io.realm.internal.network.AuthenticationServer;
+import io.realm.internal.network.NetworkStateReceiver;
 import io.realm.internal.network.OkHttpAuthenticationServer;
 import io.realm.log.RealmLog;
 
@@ -97,6 +98,19 @@ public class SyncManager {
     // Right now it just lives and dies together with the process.
     private static volatile AuthenticationServer authServer = new OkHttpAuthenticationServer();
     private static volatile UserStore userStore;
+
+    private static NetworkStateReceiver.ConnectionListener networkListener = new NetworkStateReceiver.ConnectionListener() {
+        @Override
+        public void onChange(boolean connectionAvailable) {
+            if (connectionAvailable) {
+                RealmLog.debug("NetworkListener: Connection available");
+                // notify all sessions
+                notifyNetworkIsBack();
+            } else {
+                RealmLog.debug("NetworkListener: Connection lost");
+            }
+        }
+    };
 
     static volatile SyncSession.ErrorHandler defaultSessionErrorHandler = SESSION_NO_OP_ERROR_HANDLER;
 
@@ -181,6 +195,10 @@ public class SyncManager {
         if (session == null) {
             session = new SyncSession(syncConfiguration);
             sessions.put(syncConfiguration.getPath(), session);
+            if (sessions.size() == 1) {
+                RealmLog.debug("first session created add network listener");
+                NetworkStateReceiver.addListener(networkListener);
+            }
         }
 
         return session;
@@ -198,6 +216,10 @@ public class SyncManager {
         SyncSession syncSession = sessions.remove(syncConfiguration.getPath());
         if (syncSession != null) {
             syncSession.close();
+        }
+        if (sessions.isEmpty()) {
+            RealmLog.debug("last session dropped, remove network listener");
+            NetworkStateReceiver.removeListener(networkListener);
         }
     }
 
@@ -245,6 +267,14 @@ public class SyncManager {
                     RealmLog.error(exception);
                 }
             }
+        }
+    }
+
+    private static synchronized void notifyNetworkIsBack() {
+        try {
+            nativeReconnect();
+        } catch (Exception exception) {
+            RealmLog.error(exception);
         }
     }
 
@@ -303,4 +333,5 @@ public class SyncManager {
     protected static native void nativeInitializeSyncManager(String syncBaseDir);
     private static native void nativeReset();
     private static native void nativeSimulateSyncError(String realmPath, int errorCode, String errorMessage, boolean isFatal);
+    private static native void nativeReconnect();
 }
--- a/realm/realm-library/src/main/cpp/io_realm_SyncManager.cpp
+++ b/realm/realm-library/src/main/cpp/io_realm_SyncManager.cpp
@@ -100,3 +100,12 @@ JNIEXPORT void JNICALL Java_io_realm_SyncManager_nativeSimulateSyncError(JNIEnv*
     }
     CATCH_STD()
 }
+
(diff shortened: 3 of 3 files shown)


nhachicha:
@realm/java

realm/realm-library/src/objectServer/java/io/realm/SyncManager.java
```diff
@@ -181,6 +208,11 @@ public static synchronized SyncSession getSession(SyncConfiguration syncConfigur
         if (session == null) {
             session = new SyncSession(syncConfiguration);
             sessions.put(syncConfiguration.getPath(), session);
+            if (sessions.size() == 1) {
+                // first session add network listener
+                System.out.println(">>>>>>>>>>>>>>>>>>>>>>>>>>>>> first session add network listener");
```
Zhuinden: Not a fond of `System.out.println()`

realm/realm-library/src/objectServer/java/io/realm/SyncManager.java
```diff
@@ -199,6 +231,11 @@ private static synchronized void removeSession(SyncConfiguration syncConfigurati
         if (syncSession != null) {
             syncSession.close();
         }
+        if (sessions.size() == 0) {
+            // last session removed, remove network listener
+            System.out.println(">>>>>>>>>>>>>>>>>>>>>>>>>>>>> last session removed, remove network listener");
```
Zhuinden: Not fond of `System.out.println()`

nhachicha: that's an old commit :)

realm/realm-library/src/objectServer/java/io/realm/SyncManager.java
nhachicha: that's an old commit :)

realm/realm-library/src/objectServer/java/io/realm/SyncManager.java
```diff
@@ -303,4 +336,5 @@ static void simulateClientReset(SyncSession session) {
     protected static native void nativeInitializeSyncManager(String syncBaseDir);
     private static native void nativeReset();
     private static native void nativeSimulateSyncError(String realmPath, int errorCode, String errorMessage, boolean isFatal);
+    private static native void nativeNotifyNetworkIsBack();
```
beeender: why name it as `NotifyNetworkIsBack()` instead of `reconnect()` to match the same thing in the Object Store? Are we going to do more things than reconnect in this function?

realm/realm-library/src/objectServer/java/io/realm/SyncManager.java
```diff
@@ -181,6 +198,10 @@ public static synchronized SyncSession getSession(SyncConfiguration syncConfigur
         if (session == null) {
             session = new SyncSession(syncConfiguration);
             sessions.put(syncConfiguration.getPath(), session);
+            if (sessions.size() == 1) {
+                RealmLog.debug("first session created add network listener");
+                NetworkStateReceiver.addListener(networkListener);
```
beeender: `NetworkStateReceiver.addListener()` is not `synchronized` method, but `NetworkStateReceiver.removeListener()` is . Is it correct?

realm/realm-library/src/objectServer/java/io/realm/SyncManager.java
Zhuinden: whoops :smile:

realm/realm-library/src/objectServer/java/io/realm/SyncManager.java
```diff
@@ -98,6 +100,21 @@ public void onError(SyncSession session, ObjectServerError error) {
     private static volatile AuthenticationServer authServer = new OkHttpAuthenticationServer();
     private static volatile UserStore userStore;
 
+    private static NetworkStateReceiver.ConnectionListener networkListener = new NetworkStateReceiver.ConnectionListener() {
+        @Override
+        public void onChange(boolean connectionAvailable) {
+            if (connectionAvailable) {
+                if (NetworkStateReceiver.isOnline(SyncObjectServerFacade.getApplicationContext())) {
```
beeender: This checking seems not necessary? See https://github.com/realm/realm-java/blob/79152436a06c7fd92806483fa4f07eb9a675702b/realm/realm-library/src/objectServer/java/io/realm/internal/network/NetworkStateReceiver.java#L82

realm/realm-library/src/objectServer/java/io/realm/SyncManager.java
nhachicha: `addListener ` is called from a synchronized path as well (from `public static synchronized SyncSession getSession`)

realm/realm-library/src/objectServer/java/io/realm/SyncManager.java
```diff
@@ -98,6 +100,19 @@ public void onError(SyncSession session, ObjectServerError error) {
     private static volatile AuthenticationServer authServer = new OkHttpAuthenticationServer();
     private static volatile UserStore userStore;
 
+    private static NetworkStateReceiver.ConnectionListener networkListener = new NetworkStateReceiver.ConnectionListener() {
+        @Override
+        public void onChange(boolean connectionAvailable) {
+            if (connectionAvailable) {
+                RealmLog.debug("networkListener connection available");
```
cmelchior: Perhaps: `NetworkListener: Connection available` and `NetworkListener: Connection lost` for better readability.

realm/realm-library/src/objectServer/java/io/realm/SyncManager.java
```diff
@@ -199,6 +218,10 @@ private static synchronized void removeSession(SyncConfiguration syncConfigurati
         if (syncSession != null) {
             syncSession.close();
         }
+        if (sessions.size() == 0) {
```
zaki50: I prefer `isEmpty()` to `size() == 0`.

realm/realm-library/src/objectServer/java/io/realm/SyncManager.java
```diff
@@ -303,4 +334,5 @@ static void simulateClientReset(SyncSession session) {
     protected static native void nativeInitializeSyncManager(String syncBaseDir);
     private static native void nativeReset();
     private static native void nativeSimulateSyncError(String realmPath, int errorCode, String errorMessage, boolean isFatal);
+    private static native void reconnect();
```
zaki50: method name should be `nativeReconnect`?

realm/realm-library/src/objectServer/java/io/realm/SyncManager.java
nhachicha: `isEmpty` is doing the same thing http://androidxref.com/7.0.0_r1/xref/libcore/ojluni/src/main/java/java/util/HashMap.java#324 but +1 for clarity

realm/realm-library/src/objectServer/java/io/realm/SyncManager.java
nhachicha: @cmelchior 
I'm not sure about the Slack example.
 
I believe calling reconnect is a _blocking_ call in Slack kind of UI. Whereas `reconnect` here just reset the reconnect_delay, and send a `ping` to the server. The SyncClient may not resume Uploading directly.

I'm also afraid users will try to register another `NetworkListener` or JobScheduler to listen to network changes in order to call `reconnect`, which is redundant.

If something network related changed then our listener will be triggered (Slack maybe doing polling hence the force refresh now ...)

IMO. either we expose the method only and let the user, manually call (or register a NetworkListener to do it for him) *or* rely on our internal listener to call the method.

cmelchior: I was mostly worried that the NetworkListener might not catch all cases that would be considered "offline", but for a public `reconnect()` to be truly useful we would need to expose other methods like `isOnline()` and `getReconnectDelay()` (or similar).

But I'm fine with just keeping it private for now, then we can evaluate from there.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 ## 3.1.3 (YYYY-MM-DD)
 
+### Enhancements
+
+* [ObjectServer] Resume synchronization as soon as the network is back (#4141).
```
kneth: "network" -> "connectivity"?

nhachicha:
@kneth tested manually using `Realm Tasks`

PR #4500, opened 2017-04-14
Fix exception thrown from backlinks field

### LINKED ISSUE #4499: Accessing backlinks field should throw IllegalState instead of IllegalArgument if the object is deleted or not loaded
accessing backlinks of deleted or not loaded object is now throwing `IllegalArgumentException`.

This should be `IllegalStateException`.

zaki50:
fixes #4499

- [x] add test case that reproduce the issue
- [x] fix the code
- [x] change log

### DIFF
--- a/realm/realm-library/src/androidTest/java/io/realm/LinkingObjectsManagedTests.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/LinkingObjectsManagedTests.java
@@ -465,6 +465,45 @@ public class LinkingObjectsManagedTests {
         fail();
     }
 
+    @Test
+    @RunTestInLooperThread
+    public void linkingObjects_IllegalStateException_ifDeletedIndirectly() {
+        final Realm realm = looperThread.realm;
+
+        realm.executeTransaction(new Realm.Transaction() {
+            @Override
+            public void execute(Realm realm) {
+                final BacklinksTarget target1 = realm.createObject(BacklinksTarget.class);
+                target1.setId(1);
+
+                final BacklinksSource source = realm.createObject(BacklinksSource.class);
+                source.setChild(target1);
+            }
+        });
+
+        final BacklinksTarget target = realm.where(BacklinksTarget.class)
+                .equalTo(BacklinksTarget.FIELD_ID, 1L).findFirst();
+
+        realm.executeTransaction(new Realm.Transaction() {
+            @Override
+            public void execute(Realm realm) {
+                // delete target object indirectly
+                realm.where(BacklinksTarget.class)
+                        .equalTo(BacklinksTarget.FIELD_ID, 1L)
+                        .findFirst()
+                        .deleteFromRealm();
+            }
+        });
+
+        // precondition
+        assertFalse(target.isValid());
+
+        thrown.expect(IllegalStateException.class);
+        //noinspection ResultOfMethodCallIgnored
+        target.getParents();
+        fail();
+    }
+
     /**
      * Table validation should fail if the backinked column already exists in the target table.
      * The realm `backlinks-fieldInUse.realm` contains the classes `BacklinksSource` and `BacklinksTarget`
--- a/realm/realm-library/src/androidTest/java/io/realm/LinkingObjectsManagedTests.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/LinkingObjectsManagedTests.java
@@ -404,7 +404,7 @@ public class LinkingObjectsManagedTests {
 
     @Test
     @RunTestInLooperThread
-    public void linkingObjects_IllegalStateException_ifNotYetYetLoaded() {
+    public void linkingObjects_IllegalStateException_ifNotYetLoaded() {
         final Realm realm = looperThread.realm;
 
         realm.executeTransaction(new Realm.Transaction() {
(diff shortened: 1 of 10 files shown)


zaki50:
ready for review @realm/java

realm/realm-library/src/androidTest/java/io/realm/entities/BacklinksSource.java
```diff
@@ -18,6 +18,9 @@
 import io.realm.RealmObject;
 
 public class BacklinksSource extends RealmObject {
+    public static final String CLASS_NAME = "BacklinksSource";
```
dalinaum: Are `CLASS_NAME` and `FIELD_CHILD` used?

realm/realm-library/src/androidTest/java/io/realm/entities/BacklinksTarget.java
```diff
@@ -20,6 +20,10 @@
 import io.realm.annotations.LinkingObjects;
 
 public class BacklinksTarget extends RealmObject {
+    public static final String CLASS_NAME = "BacklinksTarget";
```
dalinaum: Same?

realm/realm-library/src/androidTest/java/io/realm/LinkingObjectsManagedTests.java
```diff
@@ -401,6 +402,105 @@ public void json_updateList() {
...
 
+    @Test
+    @RunTestInLooperThread
+    public void linkingObjects_IllegalStateException_ifNotYetLoaded() {
+        final Realm realm = looperThread.realm;
+
+        realm.executeTransaction(new Realm.Transaction() {
+            @Override
+            public void execute(Realm realm) {
+                final BacklinksTarget target1 = realm.createObject(BacklinksTarget.class);
```
dalinaum: Is there other target variable?

realm/realm-library/src/androidTest/java/io/realm/LinkingObjectsManagedTests.java
```diff
@@ -401,6 +402,105 @@ public void json_updateList() {
...
+
+    @Test
+    @RunTestInLooperThread
+    public void linkingObjects_IllegalStateException_ifDeleted() {
+        final Realm realm = looperThread.realm;
+
+        realm.executeTransaction(new Realm.Transaction() {
+            @Override
+            public void execute(Realm realm) {
+                final BacklinksTarget target1 = realm.createObject(BacklinksTarget.class);
```
dalinaum: Same

realm/realm-library/src/androidTest/java/io/realm/LinkingObjectsManagedTests.java
```diff
@@ -401,6 +402,105 @@ public void json_updateList() {
...
+                target1.setId(1);
+
+                final BacklinksSource source = realm.createObject(BacklinksSource.class);
+                source.setChild(target1);
+            }
+        });
+
+        final BacklinksTarget target = realm.where(BacklinksTarget.class)
```
dalinaum: Oh. I just saw this line. `target1` and `target` are little weird. Maybe there are better names.

realm/realm-library/src/androidTest/java/io/realm/entities/BacklinksSource.java
zaki50: Not used now, but will be used in https://github.com/realm/realm-java/pull/4492

realm/realm-library/src/main/java/io/realm/internal/UncheckedRow.java
```diff
@@ -188,6 +188,13 @@ public LinkView getLinkList(long columnIndex) {
         return new LinkView(context, parent, columnIndex, nativeLinkViewPtr);
     }
 
+    @Override
+    public void checkIfBacklinkAvailable() {
```
bmeike: This function has the wrong name.  It is not checking to see if backlinks are available.  It is checking to see if the object is attached.

bmeike:
I would prefer that you change the name of the new method, to reflect its actual function.  Otherwise, LGTM.

realm/realm-library/src/androidTest/java/io/realm/LinkingObjectsManagedTests.java
zaki50: Both are the same object, but one is typed, another is dynamic.
In this test case, we don't need to use different variable name for them.
I'll update them to use the same variable name.

realm/realm-library/src/main/java/io/realm/internal/UncheckedRow.java
zaki50: The main concern now is whether we can provide a backlinks object to users, but yes, that's the same with whether it's attached or not. And latter is more versatile. 

How about `checkIfAttached()`?

zaki50:
@bmeike Is `Row.checkIfAttached()` fine?
If yes, I'll merge this PR.

zaki50:
@bmeike To progress #4492, I'm merging this PR.
If there is a better name for `Row.checkIfAttached()`, I'll update it in another PR.

PR #4515, opened 2017-04-18
Re-add Realm.refresh()

### LINKED ISSUE #3476: Feature Request: Manual "Refresh" (synchronization of Realm instance to latest version on background thread)
#### Goal

I feel a certain set of de-ja-vu, but I'd like to execute a periodic task in an `IntentService` and ensure that the Realm is always up to date for querying data without having to execute a transaction on non-looper (or IntentService) background thread.
#### Expected Results

Call `realm.refresh()` on a background non-looper (or IntentService) thread, and the Realm instance should be synchronized to its latest version.
#### Actual Results

This method was removed in 0.89.0 in favor of `waitForChange()`, but that method is difficult to work with [waitForChange()/stopWaitForChange()](https://gist.github.com/cmelchior/8786ec511f7e1f47673935a26f568106) because you need to use your own thread, and you cannot wake it up periodically while it waits for a Realm change.
#### Code Sample

http://stackoverflow.com/questions/39600315/realm-not-fetching-data
or
http://stackoverflow.com/questions/38833163/realmchangelistener-does-not-get-called-when-realm-gets-updated-in-notificationl
or
http://stackoverflow.com/questions/39613453/realm-and-the-lack-of-limit-how-a-to-do-a-stepped-delete-operation/39618296#39618296

Realm version(s): 1.2.0

Android Studio version: 2.2.0

Which Android version and device: Nexus 5X, 7.0.0

cmelchior:
Closes #3476 

This PR adds back `refresh()`. Mostly for two reasons:

1) With the adoption of the ObjectStore and Snapshots for iterations, the reasons we originally removed `refresh()` in favour of `waitForChange()` for are no longer valid.

2) We keep getting support requests by people using event buses and such instead of RealmChangeListeners. In some of those cases `refresh()` does make sense, so this would lessen the support burden as well.

3) @Zhuinden no longer have to keep various hacks up to date ( Just kidding mate, we really appreciate what you are doing 😄  💯 ) 

Feel free to tell me if this is just stupid, but then we should probably also close #3476 with a Wont Fix.

### DIFF
--- a/realm/realm-library/src/main/java/io/realm/BaseRealm.java
+++ b/realm/realm-library/src/main/java/io/realm/BaseRealm.java
@@ -115,9 +115,11 @@ abstract class BaseRealm implements Closeable {
 
     /**
      * Refreshes the Realm instance and all the RealmResults and RealmObjects instances coming from it.
-     * It also calls the listeners associated to the Realm instance.
-     *
-     * Calling this on a thread with async queries will turn those queries into synchronous queries.
+     * It also calls any listeners associated with the Realm if neeeded.
+     * <p>
+     * WARNING: Calling this on a thread with async queries will turn those queries into synchronous queries.
+     * In most cases it is better to use {@link RealmChangeListener}s to be notified about changes to the
+     * Realm on a given thread than it is to use this method. 
      *
      * @throws IllegalStateException if attempting to refresh from within a transaction.
      */
--- a/realm/realm-library/src/main/java/io/realm/BaseRealm.java
+++ b/realm/realm-library/src/main/java/io/realm/BaseRealm.java
@@ -114,6 +114,22 @@ abstract class BaseRealm implements Closeable {
     }
 
     /**
+     * Refreshes the Realm instance and all the RealmResults and RealmObjects instances coming from it.
+     * It also calls the listeners associated to the Realm instance.
+     *
+     * Calling this on a thread with async queries will turn those queries into synchronous queries.
+     *
+     * @throws IllegalStateException if attempting to refresh from within a transaction.
+     */
+    public void refresh() {
+        checkIfValid();
+        if (isInTransaction()) {
+            throw new IllegalStateException("Cannot refresh a Realm instance inside a transaction.");
+        }
+        sharedRealm.refresh();
+    }
+
+    /**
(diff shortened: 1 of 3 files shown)


Zhuinden:
I still think there are valid use-cases for `realm.refresh()` where you probably don't want to do a `realm.beginTransaction(); realm.cancelTransaction();`, otherwise I wouldn't have to maintain the hacks for people :stuck_out_tongue_closed_eyes:

realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -114,6 +114,22 @@ public boolean isAutoRefresh() {
...
+     * It also calls the listeners associated to the Realm instance.
+     *
+     * Calling this on a thread with async queries will turn those queries into synchronous queries.
+     *
+     * @throws IllegalStateException if attempting to refresh from within a transaction.
+     */
+    public void refresh() {
+        checkIfValid();
+        if (isInTransaction()) {
+            throw new IllegalStateException("Cannot refresh a Realm instance inside a transaction.");
```
beeender: I think Object Store will throw this anyway? It would be no need to do it here .

cmelchior: It doesn't, it just return `false`. I checked.

beeender: UH. no. OS just returns false. ignore me then!

realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -114,6 +114,22 @@ public boolean isAutoRefresh() {
     }
 
     /**
+     * Refreshes the Realm instance and all the RealmResults and RealmObjects instances coming from it.
+     * It also calls the listeners associated to the Realm instance.
+     *
```
kneth: add `<p>`

cmelchior: I'll try to come up with a blob of text regarding the usage of this class.

cmelchior:
@kneth I updated the Javadoc a little.

Issue #1598, opened 2015-10-14
Collection aggregate queries


cmelchior:
It should be possible to query on aggregates from RealmLists in a model.

```
public class Foo extends RealmObject {
  private RealmList<Bar> bars;
  // ...
}
```

API will probably be a bit tricky as we don't have NSPredicate to piggyback on:

```
1) realm.where(Foo.class).equalTo("bars.count()", 10).findAll()
2) realm.where(Foo.class).equalTo("bars@min()", 0).findAll()
3) realm.where(Foo.class).maxEqualTo("bars", 0).findAll()
```

Good ideas are welcome that fits well in a fluent interface, some discussion can be found here: https://github.com/realm/realm-java/issues/975#issuecomment-147443908

Cocoa issue here: https://github.com/realm/realm-cocoa/pull/2476/files

astigsen:
I prefer `realm.where(Foo.class).equalTo("bars.count()", 10).findAll()`. It is also more flexible in the way that you can add arguments, like `realm.where(Foo.class).equalTo("bars.sum(bas)", 10).findAll()`.

But you might want to consider adding a specific method for testing if the list is empty since that is such a common request: realm.where(Foo.class).isEmpty("bars").findAll()

kneth:
With `realm.where(Foo.class).equalTo("bars.count()", 10).findAll()` it is a strange mixture of a fluent interface and a query language. I think `realm.where(Foo.class).count("bars").equalTo(10).findAll()` is better.

cmelchior:
I agree adding methods inside strings is a bit dangerous, and very hidden if you don't know it exists. However doing `count("bars").equalTo(10)` just moves ambiguity to the object level, e.g. 

`count("bars").equalTo(10)` is legal, but `count("bars").beginGroup()` would not be. It is not unsolvable but will also introduce other headaches....

beeender:
I believe we can restrict `count("bars").equalTo(10)` by using builder pattern? Or simply `count("bar", 10)`?

cmelchior:
The problem with the builder pattern is that you need "close" methods to exit the inner scope. It will be possible but you would have code that looks like this:

`
realm.where(Foo.class)
  .count("bars")
     .equalTo("10")
  .done() // exit()/end()/whatever to exit scope
  .equalTo("name", "bar").
  .findAll();
```

 
 



Regarding `count("bars",10)`. We have many more methods than equalTo: min/max/average + lessThan/greaterThan, so

[...]

PR #4519, opened 2017-04-20
LinkingObject queries

### LINKED ISSUE #4321: Allow backlink fields in queries
#### Goal

Extend @LinkingObjects feature to support queries.

#### Code Sample

This should work

```java
   String name;
   Dog dog;
 }

   String name;
   @LinkingObjects("dog")
   final RealmResults<Person> owners = null;
 }

 // Find all Dogs with at least one owner named John
 realm.where(Dog.class).equalTo("owners.dog.name", "Fido").findAll();
```
-----
### Tasks
- [x] The byte code transformation must change the field accessor (cat.owners) to getting a `RealmResults` using `Table::get_backlink_view()` making something like ```cat.owners.where().equalTo("name", "John").findAll()``` possible
- [x] Add support in `RealmQuery`
- [ ] Extend query predicates in `RealmQuery` (`equalTo()` etc.) to allow querying backlinks
- [x] Add UTs for queries
- [ ] Documentation on link queries should be extended to describe how this can be used instead to make these kind of queries much simpler.

### References

#2904 

Realm version(s): 3.1

bmeike:
Code complete for backlink queries.  Fixes #4321

Needs testing and, as anticipated, DynamicRealm tests are failing.

Let's get the review started, though.  Bring it!!  ;-P


realm/realm-library/src/main/java/io/realm/StandardRealmObjectSchema.java
```diff
@@ -642,8 +672,8 @@ StandardRealmObjectSchema add(String name, RealmFieldType type, RealmObjectSchem
      * @param fieldName field name to find index for.
      * @return column index or null if it doesn't exists.
```
Zhuinden: this can't be null anymore, then

realm/realm-library/src/main/java/io/realm/internal/ColumnIndices.java
```diff
@@ -25,67 +25,107 @@
...
+
+    /**
+     * Convenience method to return the column index for a given field on a clazz
+     * or {@code -1} if no such field exists.
+     *
+     * @param clazz the class to search.
+     * @param fieldName the name of the field whose index is needed.
+     * @return the index in clazz of the field fieldName.
      */
+    @Deprecated
```
cmelchior: Describe what to use instead

realm/realm-library/src/main/java/io/realm/internal/ColumnIndices.java
```diff
@@ -25,67 +25,107 @@
...
     }
 
-    public void copyFrom(ColumnIndices other, RealmProxyMediator mediator) {
-        for (Map.Entry<Class<? extends RealmModel>, ColumnInfo> entry : classes.entrySet()) {
-            final ColumnInfo otherColumnInfo = other.getColumnInfo(entry.getKey());
-            if (otherColumnInfo == null) {
-                throw new IllegalStateException("Failed to copy ColumnIndices cache: "
-                        + Table.tableNameToClassName(mediator.getTableName(entry.getKey())));
+    @Override
+    public String toString() {
```
cmelchior: 👏

realm/realm-library/src/main/java/io/realm/internal/ColumnInfo.java
```diff
@@ -16,60 +16,197 @@
...
+    @SuppressWarnings("unused")
+    protected final long addColumnDetails(SharedRealm realm, Table table, String columnName, RealmFieldType columnType) {
+        long columnIndex = table.getColumnIndex(columnName);
+        if (columnIndex < 0) {
+            throw new RealmMigrationNeededException(
+                    realm.getPath(),
+                    "Field '" + columnName + "' not found for type " + table.getClassName());
+        }
+
+        // expensive and unnecessary?
```
cmelchior: We need to check this if a user accidentally changed the type of a field, e.g
```
private String age -> private int age
```

realm/realm-library/src/main/java/io/realm/internal/TableQuery.java
```diff
@@ -16,29 +16,33 @@
 
 package io.realm.internal;
 
+import android.util.Log;
```
cmelchior: Not used

realm/realm-library/src/main/java/io/realm/internal/TableQuery.java
```diff
@@ -16,29 +16,33 @@
...
     private final Context context;
+    private final long nativePtr;
 
     // All actions (find(), findAll(), sum(), etc.) must call validateQuery() before performing
     // the actual action. The other methods must set queryValidated to false in order to enforce
     // the first action to validate the syntax of the query.
     private boolean queryValidated = true;
 
     // TODO: Can we protect this?
+    // FIXME: Either explain why we need the context, of quit holding on to it.
```
cmelchior: I believe it it is because the Context is the `root` element and if we fail to hold onto it, some C++  resources might be released too soon causing crashes. It might also be redundant after we introduced phantom references. @kneth or @beeender might remember

realm/realm-library/src/main/java/io/realm/internal/TableQuery.java
```diff
@@ -16,29 +16,33 @@
...
     public TableQuery(Context context, Table table, long nativeQueryPtr) {
         if (DEBUG) {
-            System.err.println("++++++ new TableQuery, ptr= " + nativeQueryPtr);
+            Log.d(TAG, "New TableQuery: ptr= " + nativeQueryPtr);
```
cmelchior: Use RealmLog not Android log

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/ProcessCommitTests.java
```diff
@@ -52,10 +52,9 @@
 @RunWith(AndroidJUnit4.class)
 public class ProcessCommitTests extends BaseIntegrationTest {
 
-    // FIXME: Ignore for now. They do still not work. It might be caused by two processes each creating
     // a Sync Client, but it needs to be investigated.
     @Test
-    @Ignore
+    @Ignore("Ignore for now. Still doesn't work. It might be caused by two processes each creating")
```
cmelchior: Missing a line of explanation.

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/ProcessCommitTests.java
```diff
@@ -109,14 +108,13 @@ public void onChange(RealmResults<ProcessInfo> element) {
...
 
-    // FIXME: Ignore for now. They do still not work. It might be caused by two processes each creating
     // a Sync Client, but it needs to be investigated.
     //TODO send string from service and match
     //     replicate integration tests from Cocoa
     //     add gradle task to start the sh script automatically (create pid file, ==> run or kill existing process
     //     check the requirement for the issue again
     @Test
-    @Ignore
+    @Ignore("Ignore for now. Still doesn't work. It might be caused by two processes each creating")
```
cmelchior: Same

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -1697,7 +1698,7 @@ ColumnIndices updateSchemaCache(ColumnIndices[] globalCacheArray) {
                     map.put(clazz, columnInfo);
                 }
             } catch (RealmMigrationNeededException e) {
-                throw e;
+                throw e; // WTF??
```
Zhuinden: It probably was here just to show that a RealmMigrationNeededException can be thrown here in this block.

realm/realm-library/src/main/java/io/realm/RealmFieldType.java
```diff
@@ -42,17 +42,16 @@
     FLOAT(9),
     DOUBLE(10),
     OBJECT(12),
-    LIST(13);
-    // BACKLINK(14); Not exposed until needed.
+    LIST(13),
+    BACKLINK(14);
```
Zhuinden: technically it's a backlink that points to the linking objects (afaik) but it's called `linking objects` everywhere else.

realm/realm-annotations-processor/src/main/java/io/realm/processor/ClassMetaData.java
```diff
@@ -195,10 +192,7 @@ public boolean isPrimaryKey(VariableElement variableElement) {
      */
     public boolean isModelClass() {
         String type = classType.toString();
-        if (type.equals("io.realm.DynamicRealmObject")) {
-            return false;
-        }
-        return (!type.endsWith(".RealmObject") && !type.endsWith("RealmProxy"));
+        return !type.equals("io.realm.DynamicRealmObject") && (!type.endsWith(".RealmObject") && !type.endsWith("RealmProxy"));
```
bmeike: Yep, fixed.

realm/realm-annotations-processor/src/main/java/io/realm/processor/Constants.java
```diff
@@ -37,73 +38,58 @@
...
-        JAVA_TO_REALM_TYPES.put("java.lang.Long", "Long");
-        JAVA_TO_REALM_TYPES.put("java.lang.Float", "Float");
-        JAVA_TO_REALM_TYPES.put("java.lang.Double", "Double");
-        JAVA_TO_REALM_TYPES.put("java.lang.Boolean", "Boolean");
-        JAVA_TO_REALM_TYPES.put("java.lang.String", "String");
-        JAVA_TO_REALM_TYPES.put("java.util.Date", "Date");
-        JAVA_TO_REALM_TYPES.put("byte[]", "BinaryByteArray");
-        // TODO: add support for char and Char
-    }
+    public enum RealmFieldType {
```
bmeike: Yes.  This was ill thought out.  Fixed.

realm/realm-annotations-processor/src/main/java/io/realm/processor/Constants.java
```diff
@@ -37,73 +38,58 @@
...
+            this.javaType = javaType;
+        }
+
+        public String getJavaType() {
+            return javaType;
+        }
+
+        @Override
+        public String toString() {
+            return "RealmFieldType." + super.toString();
```
bmeike: Yes.  This was ill thought out.  Fixed.

realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmProxyClassGenerator.java
```diff
@@ -604,32 +601,37 @@ private void emitCreateRealmObjectSchemaMethod(JavaWriter writer) throws IOExcep
...
+                case LIST:
+                    String genericTypeSimpleName = Utils.getGenericTypeSimpleName(field);
+                    writer.beginControlFlow("if (!realmSchema.contains(\"" + genericTypeSimpleName + "\"))")
+                            .emitStatement("%s%s.createRealmObjectSchema(realmSchema)", genericTypeSimpleName, Constants.PROXY_SUFFIX)
+                            .endControlFlow()
+                            .emitStatement("realmObjectSchema.add(\"%s\", RealmFieldType.LIST, realmSchema.get(\"%s\"))",
+                                    fieldName, genericTypeSimpleName);
+                    break;
+
+                default:
```
bmeike: I thought that was just a bug.  By all means, correct me if I mis-understand something.

realm/realm-annotations-processor/src/test/resources/io/realm/AllTypesRealmProxy.java
```diff
@@ -31,65 +31,57 @@
...
-
-            setIndicesMap(otherInfo.getIndicesMap());
+        protected final ColumnInfo copy(boolean mutable) {
+            return new AllTypesColumnInfo(this, mutable);
         }
 
         @Override
-        public final AllTypesColumnInfo clone() {
-            return (AllTypesColumnInfo) super.clone();
+        protected final void copy(ColumnInfo rawSrc, ColumnInfo rawDst) {
```
bmeike: Yeah, there is a reason. It is kind of a long story, though.  In most contexts, `protected` means "this is visible to subclasses".  In our code `public`, `<package>`, and `protected` all mean pretty much the same thing.  None the less, because this method is meant *only* for subclasses, I made it `protected` in the superclass.  An override cannot reduce visibility, so this is protected too.

realm/realm-library/src/androidTest/java/io/realm/ColumnIndicesTests.java
```diff
@@ -17,6 +17,7 @@
 
 import android.support.annotation.NonNull;
 import android.support.test.runner.AndroidJUnit4;
+import android.util.Log;
```
bmeike: fixed

realm/realm-library/src/androidTest/java/io/realm/ColumnIndicesTests.java
```diff
@@ -81,18 +82,20 @@ public void copyDeeply() {
         final long schemaVersion = 100;
 
         final ColumnIndices columnIndices = create(schemaVersion);
-        final ColumnIndices deepCopy = columnIndices.clone();
+        final ColumnIndices deepCopy = new ColumnIndices(columnIndices, true);
```
bmeike: That could work, but then you don't have the `copy` method.  You need copy for `copyFrom`, and you need `copyFrom` to make @zaki50 's clever fast update scheme to work.

realm/realm-library/src/androidTest/java/io/realm/LinkingObjectsManagedTests.java
```diff
@@ -19,6 +19,7 @@
 import android.content.Context;
 import android.support.test.InstrumentationRegistry;
 import android.support.test.runner.AndroidJUnit4;
+import android.util.Log;
```
bmeike: fixed

realm/realm-library/src/androidTest/java/io/realm/LinkingObjectsManagedTests.java
```diff
@@ -569,7 +570,7 @@ public void migration_backlinkedSourceClassDoesntExist() throws IOException {
         } catch (IOException e) {
             fail("Failed copying realm");
         } catch (RealmMigrationNeededException expected) {
-            assertTrue(expected.getMessage().contains("Cannot find source class"));
+            assertTrue(expected.getMessage().contains("Cannot find source field"));
```
bmeike: Yeah.  Let me look into this.  I don't know why it changed...

bmeike: This appears to be a legitimate test failure.  Will fix.

realm/realm-library/src/androidTest/java/io/realm/LinkingObjectsManagedTests.java
```diff
@@ -592,6 +593,7 @@ public void migration_backlinkedSourceClassDoesntExist() throws IOException {
      * validate its table.  If we have been living clean lives, though, the validator for
      * `BacklinksMissingFieldTarget` should notice that there is no field named `BacklinksMissingFieldSource.xxxchild`.
      */
+    @Ignore("Need to rebuild the test library")
```
bmeike: Yeah.  This needs to be repaired before this PR is committed.  It will take a while, because the UT depends on building a test library, bound at runtime, that is different than the one used at compile time.  Building that library will take maybe an hour, so I didn't want to do it until the rest of the changes were stable

realm/realm-library/src/androidTest/java/io/realm/LinkingObjectsManagedTests.java
```diff
@@ -681,7 +684,7 @@ public void query_multipleReferencesWithDistinct() {
     //  /  = object ref
     //  // = list ref
     @Test
-    @Ignore
+    @Ignore("Backlinks in queries are not supported yet")
```
bmeike: Yep!  These tests need to be added, along with a couple more...

realm/realm-library/src/androidTest/java/io/realm/RealmQueryTests.java
```diff
@@ -2461,10 +2461,10 @@ public void isNotNull_linkField() {
         // 10 Date
         assertEquals(1, realm.where(NullTypes.class).isNotNull(
                 NullTypes.FIELD_OBJECT_NULL + "." + NullTypes.FIELD_DATE_NULL).count());
+        // FIXME: Currently, Realm Core does not support isNotNull() query for nested link field.
```
bmeike: I don't know.  I just noticed that this code was commented out and added the "FIXME" to make it more visible.  Will check with core...

realm/realm-library/src/androidTest/java/io/realm/RealmQueryTests.java
```diff
@@ -2719,6 +2719,7 @@ public void isValid_removedParent() {
         list.remove(RealmFieldType.UNSUPPORTED_MIXED);
         list.remove(RealmFieldType.UNSUPPORTED_TABLE);
         list.remove(RealmFieldType.UNSUPPORTED_DATE);
+        list.remove(RealmFieldType.BACKLINK); //FIXME!!! GBM
```
bmeike: I removed the BACKLINK type from the list, in order to get the test to pass.  I don't know yet, whether that is the right thing to do or not.  It is a reminder that I need to revisit this, before I commit.

realm/realm-library/src/androidTest/java/io/realm/RealmSchemaTests.java
```diff
@@ -190,8 +191,8 @@ public void remove_invalidArgumentThrows() {
 
     // Test that it if { A -> B  && B -> A } you should remove the individual fields first before removing the entire
     // class. This also include transitive dependencies.
-    // FIXME: Disabled until https://github.com/realm/realm-core/pull/1475#issuecomment-185192434 is fixed.
-    // @Test
+    @Ignore("Disabled until https://github.com/realm/realm-core/pull/1475#issuecomment-185192434 is fixed.")
```
bmeike: Awesome!!!  Enabling.

realm/realm-library/src/androidTest/java/io/realm/RealmTests.java
```diff
@@ -1104,7 +1104,7 @@ public void copyToRealm() {
 
         AllTypes allTypes = new AllTypes();
         allTypes.setColumnString("String");
-        allTypes.setColumnLong(1l);
+        allTypes.setColumnLong(1L);
         allTypes.setColumnFloat(1f);
```
bmeike: Right.  Good call.  Fixed.

realm/realm-library/src/androidTest/java/io/realm/RealmTests.java
```diff
@@ -1104,7 +1104,7 @@ public void copyToRealm() {
...
         allTypes.setColumnDouble(1d);
```
bmeike: done

realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -696,6 +695,7 @@ public void clear() {
         }
     }
 
+    //FIXME!! This stuff doesn't appear to be used.  It should be deleted or explained.
```
Zhuinden: https://github.com/realm/realm-java/blob/4fce41319de8e69302e84e018e13c27ff5dd51c5/realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmProxyMediatorGenerator.java#L217

realm/realm-library/src/androidTestObjectServer/java/io/realm/SyncConfigurationTests.java
```diff
@@ -377,18 +378,17 @@ public void directory_dirIsAFile() throws IOException {
         file.delete(); // clean up
     }
 
-    /* FIXME: deleteRealmOnLogout is not supported by now
+    @Ignore("deleteRealmOnLogout is not supported yet")
```
bmeike: Yes.  Fair enough.  I was trying to find out which UTs were not being run.

realm/realm-library/src/main/cpp/io_realm_internal_TableQuery.cpp
```diff
@@ -62,12 +62,18 @@ JNIEXPORT jstring JNICALL Java_io_realm_internal_TableQuery_nativeValidateQuery(
 // helper functions
 
 // Return TableRef used for build link queries
-static TableRef getTableForLinkQuery(jlong nativeQueryPtr, JniLongArray& indicesArray)
+static TableRef getTableForLinkQuery(jlong nativeQueryPtr, JniLongArray& tablesArray, JniLongArray& indicesArray)
```
bmeike: Yes.  Good call.  Will do.

realm/realm-library/src/main/java/io/realm/DynamicRealmObject.java
```diff
@@ -905,12 +905,11 @@ public String toString() {
...
                     sb.append(proxyState.getRow$realm().isNullLink(columnIndex)
                             ? "null"
-                            : Table.tableNameToClassName(proxyState.getRow$realm().getTable().getLinkTarget(columnIndex).getName()));
+                            : proxyState.getRow$realm().getTable().getLinkTarget(columnIndex).getClassName());
                     break;
                 case LIST:
-                    final String tableName = proxyState.getRow$realm().getTable().getLinkTarget(columnIndex).getName();
-                    String targetType = Table.tableNameToClassName(tableName);
-                    sb.append(String.format("RealmList<%s>[%s]", targetType, proxyState.getRow$realm().getLinkList(columnIndex).size()));
+                    String tableClassName = proxyState.getRow$realm().getTable().getLinkTarget(columnIndex).getClassName();
```
bmeike: Yeah, point taken.  This is the `className`, not the `tableName`.  Unfortunately, there is, already, a variable named "className" in scope.  I thought `tableClassName` was better than what AS suggested: `className1`.

realm/realm-library/src/main/java/io/realm/OsRealmSchema.java
```diff
@@ -199,12 +208,20 @@ OsRealmObjectSchema getSchemaForClass(Class<? extends RealmModel> clazz) {
         throw new UnsupportedOperationException();
     }
 
+    /**
+     * @inheritDoc Unimplemented.
```
bmeike: Copy/Pasta.  Fixed.

realm/realm-library/src/main/java/io/realm/Realm.java
bmeike: Removed and commented.

realm/realm-library/src/main/java/io/realm/RealmFieldType.java
bmeike: I just uncommented it, but I think that we should probably be consistent.  Btw, I think that this is probably a breaking change...

Zhuinden: ....uh... why would it be breaking anything?

realm/realm-library/src/main/java/io/realm/StandardRealmObjectSchema.java
```diff
@@ -749,81 +689,28 @@ private void checkEmpty(String str) {
...
-            throw new UnsupportedOperationException();
+        public long getColumnIndex(String columnName) {
+            return table.getColumnIndex(columnName);
         }
 
         @Override
-        public int size() {
-            throw new UnsupportedOperationException();
+        protected ColumnInfo copy(boolean immutable) {
+            return this;
```
bmeike: Yes.  I had had another plan in mind, but this may be the way to fix the `DynamicRealm`

realm/realm-library/src/main/java/io/realm/internal/ColumnIndices.java
bmeike: Duh.  Will do.

realm/realm-library/src/main/java/io/realm/internal/ColumnInfo.java
bmeike: Yes, but I don't believe we need to do it here.  The check happens in Validation, I think.

realm/realm-library/src/main/java/io/realm/internal/TableQuery.java
bmeike: It is!!!

```        if (DEBUG) {
            Log.d(TAG, "New TableQuery: ptr= " + nativeQueryPtr);
        }
```

bmeike: fixed

realm/realm-library/src/main/java/io/realm/internal/TableQuery.java
bmeike: fixed

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/ProcessCommitTests.java
bmeike: I have no idea what's going on here.  I just explicitly ignored the test, instead of commenting it out.

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/ProcessCommitTests.java
bmeike: Exactly.  Instead of commenting it out, it is now explicitly ignored.  I have no idea what the problem is.

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/ProcessCommitTests.java
Zhuinden: as in, the TODO was only partly moved, and the `// a Sync Client, but it needs to be investigated.` was truncated from the end :P

realm/realm-library/src/main/java/io/realm/StandardRealmObjectSchema.java
bmeike: RIght.  Good catch.  Fixed.

realm/realm-library/src/androidTest/java/io/realm/RealmQueryTests.java
bmeike: Done.

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/ProcessCommitTests.java
Zhuinden: as in, the TODO was only partly moved, and the `// a Sync Client, but it needs to be investigated.` was truncated from the end :P

realm/realm-library/src/main/java/io/realm/internal/TableQuery.java
beeender: It is described in the comments. See https://github.com/realm/realm-java/blob/master/realm/realm-library/src/main/java/io/realm/internal/Context.java#L22
So i think the FIXME is not needed here?

realm/realm-library/src/main/java/io/realm/internal/TableQuery.java
```diff
@@ -20,25 +20,28 @@
 
 import io.realm.Case;
 import io.realm.Sort;
+import io.realm.log.RealmLog;
 
 
 public class TableQuery implements NativeObject {
-    protected boolean DEBUG = false;
+    private static final String TAG = "TABQ";
```
beeender: This is not used

realm/realm-library/src/main/java/io/realm/internal/TableQuery.java
```diff
@@ -768,9 +771,9 @@ private void throwImmutable() {
 
     private native Long nativeMinimumTimestamp(long nativeQueryPtr, long columnIndex, long start, long end, long limit);
 
-    private native void nativeIsNull(long nativePtr, long[] columnIndices);
+    private native void nativeIsNull(long nativePtr, long[] columnIndices, long[] tablePtrs);
 
-    private native void nativeIsNotNull(long nativePtr, long[] columnIndices);
+    private native void nativeIsNotNull(long nativePtr, long[] columnIndice, long[] tablePtrss);
```
beeender: `ss` in `tablePtrss`

realm/realm-library/src/main/java/io/realm/RealmSchema.java
```diff
@@ -89,13 +91,67 @@
...
+    abstract RealmObjectSchema getSchemaForClass(String className);
+
+    /**
+     * Parse the passed field description into its components.
+     * This must be standard across implementations and is, therefore, implemented in the base class.
+     *
+     * @param fieldDescription a field description.
+     * @return the parse tree: a list of column names
+     */
+    protected final List<String> parseFieldDescription(String fieldDescription) {
```
beeender: Consider to use `FieldDescriptor`? https://github.com/realm/realm-java/blob/master/realm/realm-library/src/main/java/io/realm/internal/FieldDescriptor.java

That class was initially designed to handle the cases like this, but at that time, the schema implementation is not completed, so it has to read the Table every time to get the column index. Maybe it is the time to optimize that class?

realm/realm-library/src/main/java/io/realm/internal/TableQuery.java
bmeike: Ok, I got sucker-punched here.  I had no idea that this was not an Android Context.    I really suggest changing the name, at some point, to "RealmContext".  I will remove the FIXME, but add a TODO for the rename.

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/ProcessCommitTests.java
bmeike: yep.  fixed.

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/ProcessCommitTests.java
bmeike: yep.  fixed.

realm/realm-library/src/main/java/io/realm/internal/TableQuery.java
bmeike: Nice catch.  I still have a habit, from eclipse days, of hitting splat-s, to save the file.

realm/realm-library/src/main/java/io/realm/RealmSchema.java
bmeike: Yes!  That is brilliant!

I will put a "TODO" in both classes, noting that they should be consolidated.

realm/realm-annotations-processor/src/main/java/io/realm/processor/Constants.java
Zhuinden: Aren't those new string parameters essentially equal to `name()`?

realm/realm-library/src/main/java/io/realm/StandardRealmObjectSchema.java
```diff
@@ -584,7 +584,7 @@ StandardRealmObjectSchema add(String name, RealmFieldType type, RealmObjectSchem
      * Returns the column index in the underlying table for the given field name.
      *
      * @param fieldName field name to find index for.
-     * @return column index or null if it doesn't exists.
+     * @return column index or -1 if it doesn't exists.
```
Zhuinden: exist*

Zhuinden: exist*

bmeike:
All comments addressed. ~10 failing UTs.  Need 4-6 more UTs.

bmeike:
Fixed that error message @Zhuinden.  Thanks.

realm/realm-library/src/main/java/io/realm/internal/ColumnInfo.java
```diff
@@ -16,60 +16,260 @@
...
+     * NOTE: there is no protection against calling this method with an "immutable" instance as dst!
+     *
+     * @param src The source for the copy
+     * @param dst The destination of the copy.  Will be an exact copy of src after return.
+     */
+    protected abstract void copy(ColumnInfo src, ColumnInfo dst);
+
+    /**
+     * Add a new column to the indexMap.
+     * <b>For use only by subclasses!</b>.
```
kneth: `<p>`

realm/realm-library/src/main/java/io/realm/internal/ColumnInfo.java
```diff
@@ -16,60 +16,260 @@
...
+     * @param columnType Type RealmType of the column.
+     * @return the index of the column in the table
+     */
+    @SuppressWarnings("unused")
+    protected final long addColumnDetails(SharedRealm realm, Table table, String columnName, RealmFieldType columnType) {
+        long columnIndex = table.getColumnIndex(columnName);
+        if (columnIndex < 0) {
+            throw new RealmMigrationNeededException(
+                    realm.getPath(),
+                    "Field '" + columnName + "' not found for type " + table.getClassName());
```
kneth: " not found" -> " was not found"

realm/realm-library/src/main/java/io/realm/internal/ColumnInfo.java
```diff
@@ -16,60 +16,260 @@
...
+        }
+
+        // FIXME!!! GBM - expensive and unnecessary?
```
kneth: It is only unnecessary if it is validated elsewhere

realm/realm-library/src/main/java/io/realm/internal/Table.java
```diff
@@ -502,7 +498,7 @@ protected long add(Object... values) {
                     ") does not match the number of columns in the table (" +
                     String.valueOf(columns) + ").");
         }
-        RealmFieldType colTypes[] = new RealmFieldType[columns];
+        RealmFieldType[] colTypes = new RealmFieldType[columns];
```
kneth: Oh dear, the old C programmer is still around.

realm/realm-library/src/main/java/io/realm/internal/ColumnInfo.java
bmeike: Exactly.  I believe, though, that it is.
I will confirm.

realm/realm-library/src/androidTestObjectServer/java/io/realm/SyncConfigurationTests.java
```diff
@@ -377,18 +378,17 @@ public void directory_dirIsAFile() throws IOException {
...
     @Test
     public void deleteOnLogout() {
-        User user = createTestUser();
+        SyncUser user = createTestUser();
         String url = "realm://objectserver.realm.io/default";
 
         SyncConfiguration config = new SyncConfiguration.Builder(user, url)
-                .deleteRealmOnLogout()
+                //.deleteRealmOnLogout()
```
bmeike: I really don't know.  All of this code had been commented out.  Naturally, it no longer compiles.   Since the test is now @Ignored instead of being commented out, I had to add that one line comment because it doesn't compile and I don't want to delete information.  Grrrrrr.....

realm/realm-library/src/main/cpp/io_realm_internal_TableQuery.cpp
```diff
@@ -62,12 +62,22 @@ JNIEXPORT jstring JNICALL Java_io_realm_internal_TableQuery_nativeValidateQuery(
...
+// indicesArray entry is the source column index in the source table.
+static TableRef getTableForLinkQuery(jlong nativeQueryPtr, JniLongArray& tablesArray, JniLongArray& indicesArray)
 {
     TableRef table_ref = Q(nativeQueryPtr)->get_table();
     jsize link_element_count = indicesArray.len() - 1;
     for (int i = 0; i < link_element_count; i++) {
-        table_ref->link(size_t(indicesArray[i]));
+        auto colIndex = size_t(indicesArray[i]);
+        auto tablePtr = TBL(tablesArray[i]);
+        if (tablePtr == nullptr) {
```
bmeike: Yes, that is the intention.  Though, I have to say, in trying to come up with test cases, it doesn't seem all that useful to me...

realm/realm-library/src/main/java/io/realm/internal/ColumnIndices.java
```diff
@@ -24,68 +24,149 @@
...
-public final class ColumnIndices implements Cloneable {
+public final class ColumnIndices {
+    private final Map<Class<? extends RealmModel>, ColumnInfo> classes;
+    private final Map<String, ColumnInfo> classesByName;
+    private final boolean mutable;
     private long schemaVersion;
-    private Map<Class<? extends RealmModel>, ColumnInfo> classes;
 
+    /**
+     * Create a mutable ColumnIndices containing the ColumnInfo objects in the passed map
```
bmeike: Fixed

realm/realm-library/src/main/java/io/realm/internal/ColumnIndices.java
```diff
@@ -24,68 +24,149 @@
...
+     * @param schemaVersion the schema version
+     * @param classes a map of table classes to their column info
+     * @throws IllegalArgumentException if any of the ColumnInfo object is immutable.
+     */
     public ColumnIndices(long schemaVersion, Map<Class<? extends RealmModel>, ColumnInfo> classes) {
+        this(schemaVersion, new HashMap<>(classes), true);
+    }
+
+    /**
+     * Create a copy of the passed ColumnIndices
```
bmeike: fixed

realm/realm-library/src/main/java/io/realm/internal/ColumnInfo.java
bmeike: fixed.

realm/realm-library/src/main/cpp/io_realm_internal_TableQuery.cpp
Zhuinden: Well it sounds tricky but nice to have.

For example Dog -> owners -> cats?

kneth: So "dogs which live together with cats?"

realm/realm-library/src/main/java/io/realm/internal/ColumnInfo.java
bmeike: Confirmed.  All these validations are performed in Proxy.validate...

cmelchior: Great find. Purge it with 🔥 then 👍

realm/realm-library/src/main/cpp/io_realm_internal_TableQuery.cpp
Zhuinden: Pretty much! I know cats like that, I may as well query for then :stuck_out_tongue_winking_eye:

bmeike:
Only UTs left.  Praying they all work...

realm/realm-library/src/androidTest/java/io/realm/LinkingObjectsManagedTests.java
cmelchior: Perhaps put that info in the ignore message?

realm/realm-library/src/androidTest/java/io/realm/RealmQueryTests.java
```diff
@@ -2758,6 +2957,9 @@ public void isEmpty() {
                 case LIST:
                     assertEquals(1, realm.where(AllJavaTypes.class).isEmpty(AllJavaTypes.FIELD_LIST).count());
                     break;
+                case LINKING_OBJECTS:
+                    // FIXME!!! GBM - write test;
```
Zhuinden: This definitely looks important :D

realm/realm-library/src/androidTest/java/io/realm/ColumnInfoTests.java
```diff
@@ -135,8 +145,18 @@ public void clone_hasSameValue() {
...
-        // Current implementation shares the indices map between copies.
-        assertSame(columnInfo.getIndicesMap(), copy.getIndicesMap());
+    @Test
+    public void copy_immutableThrows() {
+        final CatRealmProxy.CatColumnInfo original
+                = (CatRealmProxy.CatColumnInfo) mediator.validateTable(Cat.class, realm.sharedRealm, false);
+
+        CatRealmProxy.CatColumnInfo copy = (CatRealmProxy.CatColumnInfo) original.copy(false);
+        try {
+            copy.copyFrom(original);
```
bmeike: Good idea, but out of scope.  This change is already way too big.

realm/realm-library/src/androidTest/java/io/realm/LinkingObjectsManagedTests.java
```diff
@@ -630,14 +633,15 @@ public void migration_backlinkedSourceFieldDoesntExist() {
      * for `BacklinksWrongTypeTarget` should notice, though, that its `parents` field points to an object
      * of the wrong type, `Integer`, instead of `BacklinksWrongTypeSource`.
      */
+    @Ignore("Need to rebuild the test library")
```
bmeike: I will try that.  I'm not sure it will work, but if it does, that would be fantastic.

realm/realm-library/src/androidTest/java/io/realm/LinkingObjectsManagedTests.java
```diff
@@ -796,13 +800,21 @@ private void verifyPostConditions(final Realm realm, final PostConditions test,
...
+                new Runnable() {
+                    @Override
+                    public void run() {
+                        test.run(realm);
+                        looperThread.testComplete();
+                    }
+                });
+    }
+
+    @Test
```
bmeike: Yes.  In fact, adding two more, now.

realm/realm-library/src/androidTest/java/io/realm/RealmQueryTests.java
bmeike: Yes.

realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -714,6 +713,7 @@ public void clear() {
         }
     }
 
+    // FIXME: This stuff doesn't appear to be used.  It should be deleted or explained.
```
bmeike: Not removing that comment until it can be replaced with a description of the use of that stuff!  ;-P

realm/realm-library/src/main/java/io/realm/internal/Table.java
```diff
@@ -1169,11 +1172,24 @@ public long getVersion() {
...
+    public static String getClassNameForTable(String name) {
+        if (name == null) { return null; }
+        if (!name.startsWith(TABLE_PREFIX)) {
+            return name;
+        }
+        return name.substring(TABLE_PREFIX.length());
+    }
+
+    public static String getTableNameForClass(Class<?> clazz) {
+        return (clazz == null) ? null : getTableNameForClass(clazz.getSimpleName());
```
bmeike: Whoah!  Great catch!!!

bmeike:
I would like to take @cmelchior 's suggestion and merge this wo/the UTs.  I have added temporary code in `StandardRealmSchema` that disables LinkingObjects in queries.  I will submit ~12 UTs and any changes they necessitate, as a separate PR.

realm/realm-library/src/main/java/io/realm/internal/Table.java
bmeike: Turns out that nobody is using that method.  Deleted.

realm/realm-library/src/androidTest/java/io/realm/LinkingObjectsManagedTests.java
bmeike: After thinking about that, I am pretty sure that it will not work.  I need two different versions of exactly the same class.  Unless I'm missing something, there is no way to do that in a single compilation unit.

cmelchior:
@bmeike We shouldn't merge anything to `master` that doesn't have unit tests. Here you can just branch of `feature/backlinks` into `feature/backlinks-test` , add the tests there and then make a PR with `feature/backlinks` as base. That will make a clean, small PR, without having to merge this to `master`

realm/realm-library/src/androidTest/java/io/realm/LinkingObjectsManagedTests.java
cmelchior: Some scenarios might be too much trouble to be worth unit testing. This is such a far fetched corner case anyway, and the user would it immediately when using the app. So not having these tests are probably fine 👍

bmeike:
Got it.  ... now I'll have two branches to merge with master, every morning!

cmelchior:
Umm, no? You shouldn't merge `master` into a branch that has been made from another branch. That would kinda defeat the purpose?

`master` <--- `feature/backlinks` <--- `feature/backlinks-test

1) `master should only be merged into `feature/backlinks`
2) Then `feature/backlinks` should be merged into `feature/backlinks-tests`. Ideally, this should cause no or very few merge conflicts,

bmeike:
Right.  I mean that I have to merge *from* master.  ... and I agree that the second merge should be painless.  Just don't want to deal with merge-confilict hell when this sucker is actually ready to go.  I've been lucky, so far....

cmelchior:
@bmeike Yes, it does suck. Primarily we don't want to change that workflow because it is a nice motivator to not have branches hanging around for too long. You probably got hit by it a bit more than usual 😢

realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmProxyClassGenerator.java
bmeike: Fixed.  Will now ignore unrecognized field types.

realm/realm-library/src/androidTest/java/io/realm/LinkingObjectsManagedTests.java
bmeike: As I think I mentioned, elsewhere, this probably won't work.  I need the exact same class to have two different definitions.  I think the only way to do that is with a separate compilation unit.  The test library needs to be rebuilt.  I'll be doing that before the commit.

realm/realm-library/src/androidTestObjectServer/java/io/realm/SyncConfigurationTests.java
bmeike: Fixing this is outside the scope of this PR

realm/realm-library/src/androidTest/java/io/realm/LinkingObjectsManagedTests.java
bmeike: Ok, I'll make these tests low priority.  I'm guessing, though, that this isn't going to be uncommon at all, once people build frameworks that depend on Realm.

realm/realm-library/src/androidTest/java/io/realm/ColumnInfoTests.java
bmeike: Hey @beeender !  You will be glad to know that this is back in scope, done, and a huge improvement!

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -326,7 +326,7 @@ private static RealmCache getCache(String realmPath, boolean createIfNotExist) {
 
             if (realmClass == Realm.class && refAndCount.globalCount == 0) {
                 // Stores a copy of local ColumnIndices as a global cache.
-                RealmCache.storeColumnIndices(cache.typedColumnIndicesArray, realm.schema.getImmutableColumnIndicies());
+                RealmCache.storeColumnIndices(typedColumnIndicesArray, realm.schema.getImmutableColumnIndicies());
```
beeender: Is github doing something wrong? Is this part of commit?

realm/realm-library/src/main/java/io/realm/internal/fields/FieldDescriptor.java
```diff
@@ -0,0 +1,120 @@
...
+    public static FieldDescriptor createLegacyDescriptor(Table table, String fieldDescription, boolean allowLink, boolean allowList) {
+        return new LegacyFieldDescriptor(table, fieldDescription, allowLink, allowList);
+    }
+
+    /**
+     * TODO:
+     * I suspect that choosing the parsing strategy based on whether there is a ref to a ColumnIndices
+     * around or not, is bad architecture.  Almost certainly, there should be a schema that has
+     * ColumnIndices and one that does not and the strategies below should belong to the first
+     * and second, respectively.  --gbm
```
beeender: No offence and I am not against leaving name in the comment, but it looks quite old style :P Like something people did before git was invented. To know who left the comment there, `git blame` should just work.

marchy:
Would this feature include supporting subqueries in regular queries (i.e.: not across inverse relationships)? Tracked in https://github.com/realm/realm-java/issues/1598

bmeike:
Abandoned.  Any useful code from this PR was included in the merge of #4576

bmeike:
All comments addressed.  Dismissing reviews.

bmeike:
Merge and close.  Continuing work on this feature tracked in #4620

PR #4528, opened 2017-04-20
Fix flaky test caused by non-closed Realm

beeender:
When the case is using ExpectedException, TestRealmConfigurationFactory
won't be able to detect the Realm instance is not closed.


cmelchior:
> When the case is using ExpectedException, TestRealmConfigurationFactory
won't be able to detect the Realm instance is not closed.

But none of the changes are are in a test cases that uses the `ExpectedException`?

beeender:
@cmelchior https://github.com/realm/realm-java/pull/4528/files#diff-feb202ebcef8ca848c17b2f2f81e78e7R372 and https://github.com/realm/realm-java/pull/4528/files#diff-feb202ebcef8ca848c17b2f2f81e78e7R392

PR #4536, opened 2017-04-23
Add support for SyncConfiguration.waitForServerChanges()

### LINKED ISSUE #4270: SyncConfiguration.waitForServerChanges()
When using synced Realms it can sometimes be beneficial to wait until all changes from the server have been downloaded. Perhaps to avoid many rapid UI refreshes or because it only makes sense to view the data as a whole.

To address this issue we want to add two new API's:

1. New SyncConfiguration parameter: `SyncConfiguration.Builder.waitForServerChanges()`

This is a new boolean that enforces the behavior that it shouldn't be possible to open a Realm until  all changes have been downloaded. This should use the Object Store API described here: https://github.com/realm/realm-object-store/pull/385

* Setting this boolean will cause `getInstance(config)` to throw an Exception on the UI thread. It is probably okay to make blocking calls everywhere else.

* Setting this boolean should cause Realm to block the Realm file from being opened until the server says all changes have been downloaded.

2. New async method for opening instances: `Realm.getInstanceAsync(RealmConfiguration config)`

In order to provide access to Realms configured this way on the UI thread we need an async way of opening the Realm since network operations are not allowed there. Fortunately we already discussed the option of a `Realm.getInstanceAsync(config)` method: #2299 

This method should defer all long-running actions to a background Realm before allowing the caller thread to open the Realm instance. This includes: Validating the schema, running any migration logic, compacting (https://github.com/realm/realm-java/issues/3739) and downloading server changes as required. #2299 will be tracking that part.

**Usage:**
```java
// Create configuration
SyncConfiguration config = new SyncConfiguration.Builder(user, server)
  .waitForServerChanges()
  .build();

// Get instance on a Looper thread
Realm.getInstanceAsync(config, new Realm.Callback() {
  public void OnSuccess(Realm realm) {
    // Use Realm
  }
  // Do we really want to pollute the API with this method?
  // Perhaps make Realm.Callback abstract and provide default impl for this?
  // Should we instead provide `getInstanceAsync(config, onSuccess, onError)`?
    // Something went wrong? 
  }
})

// Get instance on a background thread (blocking call)
// This must throw if on the UI thread
````

This issue tracks the 1st part. `getInstanceAsync()` is tracked here #2299 

The equivalent Cocoa implementation is tracked here: https://github.com/realm/realm-cocoa/pull/4681. Their implementation will differ a little due to different constraints. 

**TODO**
- [ ] Add new setting to `SyncConfiguration.Builder.waitForServerChanges()` ... Name is not set in stone, so can still be bikeshedded.
- [ ] Add `SyncConfiguration.willWaitForServerChanges()` getter.
- [ ] Setting this boolean should block the Realm from being opened until the server says it has all changes. This is naturally racy, so use the semantics defined by OS (determine "done" when starting download).
- [ ] Calling a blocking `getInstance()` should throw on the UI thread since networking isn't allowed there. 
- [ ] Decide if it is acceptable to make blocking calls on other Looper threads. Initial hunch would be yes.
- [ ] Decide how to handle being offline. Is it acceptable to wait indefinetetely, or perhaps provide a timeout using e.g `waitForServerChanges(1, TimeUnit.MINUTE), after which the Realm file will be opened anyway. It is still unclear if there is a use case for this, so can possible be punted.
- [ ] Unit tests
- [ ] Integration test towards Sync server.
- [ ] Website docs

### PR DESCRIPTION
**Pull request by cmelchior**
Fixes #4270 

This feature doesn't really make a lot of sense without #2299, so probably shouldn't be merged until after that is done, but it can be reviewed without it.

This condition is only checked the first time a Realm is opened on a Thread, i.e. if multiple instances are opened, the subsequent calls will just return the cached instance. Is this the semantics we want? or should we do this on each call to `getInstance()`?

TODO
- [x] Integration tests
- [x] Unit tests
- [x] Convert links to `getInstanceAsync` from `@code` to `@link` once it has been merged

### CODE REVIEW

**Comment by cmelchior** [comment]
Ready for review @realm/java

File: realm/realm-library/src/main/java/io/realm/BaseRealm.java [hunk 1]
```diff
@@ -86,6 +87,11 @@ public void onSchemaVersionChanged(long currentVersion) {
                             }
                         }, true);
         this.schema = new StandardRealmSchema(this);
+
+        // If waitForServerChanges() was enabled, we need to make sure that all data is downloaded
+        // before proceeding. We need to open the Realm instance first to start any potential underlying
+        // SyncSession so this will work. TODO: This needs to be decoupled.
+        ObjectServerFacade.getSyncFacadeIfPossible().downloadServerChangesIfNeeded(configuration);
```
beeender: Currently all operations in the `RealmCache` is locked on the `RealmCache.class`, which means this will block opening all other different `RealmConfiguration`s. But never mind, i think having it here is OK, i will make the add a lock on individual `RealmCache` in the `getInstanceAsync()` PR.

realm/realm-library/src/main/cpp/io_realm_SyncSession.cpp
```diff
@@ -49,3 +53,47 @@ JNIEXPORT jboolean JNICALL Java_io_realm_SyncSession_nativeRefreshAccessToken(JN
...
+    TR_ENTER()
+    try {
+        JStringAccessor local_realm_path(env, localRealmPath);
+        auto session = SyncManager::shared().get_existing_session(local_realm_path);
+
+        if (session) {
+            static jni_util::JavaClass java_sync_session_class(env, "io/realm/SyncSession");
+            static jni_util::JavaMethod java_notify_result_method(
+                env, java_sync_session_class, "notifyAllChangesDownloaded", "(Ljava/lang/Long;Ljava/lang/String;)V");
+            jobject java_session_object = env->NewGlobalRef(session_object);
```
beeender: Use our wrapper class `JavaGlobalRef` instead, but you need to add copy constructor to the `JavaGlobalRef` class
Then you don't need line 84 and 87-90

realm/realm-library/src/main/cpp/jni_util/java_local_ref.hpp
```diff
@@ -46,11 +46,6 @@ class JavaLocalRef {
         m_env->DeleteLocalRef(m_jobject);
     }
 
-    JavaLocalRef(const JavaLocalRef&) = delete;
-    JavaLocalRef& operator=(const JavaLocalRef&) = delete;
-    JavaLocalRef(JavaLocalRef&& rhs) = delete;
-    JavaLocalRef& operator=(JavaLocalRef&& rhs) = delete;
```
beeender: The default move/copy constructor won't work. So remove those lines will cause crash when you copy/move a non-null `JavaLocalRef`. (Now it doesn't crash just because of you copy a non-null `JavaLocalRef` to a null one)

So if the copy/move constructor is needed, then it needs the implementation as well.

realm/realm-library/src/objectServer/java/io/realm/SyncSession.java
```diff
@@ -118,6 +124,63 @@ void close() {
...
+    public void downloadAllServerChanges() {
+        checkMainThread("downloadAllServerChanges() cannot be called from the main thread.");
+
+        // Blocking only happens at the Java layer. To prevent deadlocking the underlying SyncSession we register
+         // an async listener there and let it callback to the Java Session when done. This feel icky at best, but
+         // since all operations on the SyncSession operate under a shared mutex, we would prevent all other actions on the
+         // session, including trying to stop it.
+         // In Java we cannot lock on the Session object either since it will prevent any attempt at modifying the
+         // lifecycle while it is in a waiting state. Thus we use a specialised mutex.
+        synchronized (waitForChangesMutex) {
```
beeender: is it allowed to call `downloadAllServerChanges` more than once?

cmelchior: Nothing prevents you from doing it right now, since the `downloadAllServerChanges()` method is public. I'm actually a bit sceptical about having the `downloadAllServerChanges()` method public, so feedback on that would be nice as well.

realm/realm-library/src/main/cpp/io_realm_SyncSession.cpp
cmelchior: `JavaGlobalRef` doesn't seem to call `DeleteGlobalRef` anywhere?

beeender: in `java_global_ref.cpp`

realm/realm-library/src/main/cpp/io_realm_SyncSession.cpp
```diff
@@ -49,3 +54,42 @@ JNIEXPORT jboolean JNICALL Java_io_realm_SyncSession_nativeRefreshAccessToken(JN
...
+                        java_error_message_ref =
+                            jni_util::JavaLocalRef<jstring>(env, env->NewStringUTF(error.message().c_str()));
+                    }
+                    env->CallVoidMethod(java_session_object_ref.get(), java_notify_result_method,
+                                        java_error_code_ref.get(), java_error_message_ref.get());
+                });
+
+            return to_jbool(listener_registered);
+        }
+        else {
```
kneth: Not needed - you're also returning after `CATCH_STD()`

realm/realm-library/src/main/java/io/realm/internal/Capabilities.java
```diff
@@ -35,4 +35,10 @@
      * @param exceptionMessage message which is contained in the exception.
      */
     void checkCanDeliverNotification(String exceptionMessage);
+
+    /**
+     * Multiple threads might be able to deliver notifications, but the Main thread in GUI applications often have
+     * special rules that needs to be enforced.
```
kneth: "needs" -> "need"

realm/realm-library/src/main/java/io/realm/internal/android/AndroidCapabilities.java
```diff
@@ -50,6 +50,15 @@ public void checkCanDeliverNotification(String exceptionMessage) {
         }
     }
 
+    @Override
+    public boolean isMainThread() {
+        return looper == Looper.getMainLooper();
```
kneth: `looper != null && looper == Looper.getMainLooper()` (if `getMainLooper()` is expensive)

realm/realm-library/src/objectServer/java/io/realm/SyncConfiguration.java
```diff
@@ -210,6 +218,17 @@ public boolean shouldDeleteRealmOnLogout() {
         return deleteRealmOnLogout;
     }
 
+
+    /**
+     * Returns {@code true} if the Realm will download all known changes from the remote server before being opened.
+     *
+     * @return {@code true} if all changes will be downloaded before the Realm can be opened. {@code false} if the Realm
+     * can be opened right away.
```
kneth: "right away" -> "instantaneously"????

realm/realm-library/src/objectServer/java/io/realm/SyncConfiguration.java
```diff
@@ -557,6 +577,27 @@ public Builder errorHandler(SyncSession.ErrorHandler errorHandler) {
...
+        /**
+         * Setting this will cause the Realm to not open until all known changes from the server has been downloaded.
+         * <p>
+         * Since downloading all changes can be an lengthy operation that might block the UI thread, Realms with this
+         * setting enabled should only be opened on background threads or with
+         * {@code Realm.getInstanceAsync(config, onSuccess)} on the UI thread.
+         * <p>
+         * If a device is offline, the Realm will open immediately.
+         * <p>
+         * This check is only enforced the first time a Realm is opened on each thread, i.e. it will be enforced
```
kneth: ", i.e." -> "i.e., "

realm/realm-library/src/objectServer/java/io/realm/SyncSession.java
```diff
@@ -118,6 +124,63 @@ void close() {
         clearScheduledAccessTokenRefresh();
     }
 
+    // This method will be called once all changes have been downloaded.
+    // This method might be called on another thread than the one that called `downloadAllServerChanges`
```
kneth: `.` at the end

realm/realm-library/src/objectServer/java/io/realm/SyncSession.java
```diff
@@ -118,6 +124,63 @@ void close() {
...
+    // Be very careful with synchronized blocks.
+    // If the native listener was successfully registered, Object Store guarantee that this method will be called at
```
kneth: "guarantee" -> "guarantees"

realm/realm-library/src/objectServer/java/io/realm/SyncSession.java
```diff
@@ -118,6 +124,63 @@ void close() {
...
+    // least once, even if the session is closed.
+    void notifyAllChangesDownloaded(Long errorcode, String errorMessage) {
+        if (waitingForServerChanges != null) {
+            waitingForServerChanges.handleResult(errorcode, errorMessage);
+        }
+    }
+
+    /**
+     * Calling this method will block until all known remote changes have been download and applied to the Realm.
```
kneth: "download" -> "downloaded"

realm/realm-library/src/objectServer/java/io/realm/SyncSession.java
```diff
@@ -118,6 +124,63 @@ void close() {
...
+     * <p>
+     * This method cannot be called before the session has been started.
+     *
+     * @throws IllegalStateException if called on the Android main thread.
+     */
+    public void downloadAllServerChanges() {
+        checkMainThread("downloadAllServerChanges() cannot be called from the main thread.");
+
+        // Blocking only happens at the Java layer. To prevent deadlocking the underlying SyncSession we register
+         // an async listener there and let it callback to the Java Session when done. This feel icky at best, but
```
kneth: A space too much
"feel" -> "feel"

realm/realm-library/src/objectServer/java/io/realm/SyncSession.java
```diff
@@ -118,6 +124,63 @@ void close() {
...
+         // session, including trying to stop it.
+         // In Java we cannot lock on the Session object either since it will prevent any attempt at modifying the
+         // lifecycle while it is in a waiting state. Thus we use a specialised mutex.
+        synchronized (waitForChangesMutex) {
+            if (!isClosed) {
+                waitingForServerChanges = new WaitForServerChangesWrapper();
+                boolean listenerRegistered = nativeWaitForDownloadCompletion(configuration.getPath());
+                if (!listenerRegistered) {
+                    waitingForServerChanges = null;
+                    throw new ObjectServerError(ErrorCode.UNKNOWN, "It was not possible to download all changes. Have the Sync Client been started?");
```
kneth: "Sync Client" -> "SyncClient" or "sync client"

realm/realm-library/src/objectServer/java/io/realm/SyncSession.java
```diff
@@ -351,6 +414,60 @@ private void clearScheduledAccessTokenRefresh() {
...
+                } catch (InterruptedException e) {
+                    throw new ObjectServerError(ErrorCode.UNKNOWN, e);
+                }
+            }
+        }
+
+        /**
+         * Process the result of a waiting action. This will also unblock anyone who called {@link #waitForChanges}.
+         *
+         * @param errorCode error code if an error occured, {@code null} if changes where successfully downloaded.
```
kneth: "where" -> "were"

realm/realm-library/src/objectServer/java/io/realm/SyncSession.java
```diff
@@ -442,7 +442,7 @@ public void waitForServerChanges() {
         /**
          * Process the result of a waiting action. This will also unblock anyone who called {@link #waitForChanges}.
          *
-         * @param errorCode error code if an error occured, {@code null} if changes where successfully downloaded.
+         * @param errorCode error code if an error occurred, {@code null} if changes were successfully downloaded.
```
Zhuinden: Has?

realm/realm-library/src/objectServer/java/io/realm/SyncSession.java
```diff
@@ -149,18 +149,18 @@ public void downloadAllServerChanges() {
...
+        // In Java we cannot lock on the Session object either since it will prevent any attempt at modifying the
+        // lifecycle while it is in a waiting state. Thus we use a specialised mutex.
         synchronized (waitForChangesMutex) {
             if (!isClosed) {
                 waitingForServerChanges = new WaitForServerChangesWrapper();
                 boolean listenerRegistered = nativeWaitForDownloadCompletion(configuration.getPath());
                 if (!listenerRegistered) {
                     waitingForServerChanges = null;
-                    throw new ObjectServerError(ErrorCode.UNKNOWN, "It was not possible to download all changes. Have the Sync Client been started?");
+                    throw new ObjectServerError(ErrorCode.UNKNOWN, "It was not possible to download all changes. Have the SyncClient been started?");
```
Zhuinden: Has?

realm/realm-library/src/objectServer/java/io/realm/SyncSession.java
```diff
@@ -351,6 +414,60 @@ private void clearScheduledAccessTokenRefresh() {
...
 
+    // Wrapper class for handling the async operations of the underlying SyncSession calling `async_wait_for_download_completion`
+    private static class WaitForServerChangesWrapper {
+
+        private final CountDownLatch waitForChanges = new CountDownLatch(1);
+        private volatile boolean resultReceived = false;
+        private Long errorCode = null;
+        private String errorMessage;
+
+        public void cancelWait() {
```
beeender: This is not called by anyone I think? Actually this can just be removed. `getInstanceAsync()` will be using `RealmAsyncTask`, when calling `RealmAsyncTask.cancel()`, `CountDownLatch.await()` will just throw a `InterruptedException`. see line 436 in this file. Also, line 437 should log the exception instead of throw it.

realm/realm-library/src/objectServer/java/io/realm/SyncSession.java
```diff
@@ -351,6 +414,60 @@ private void clearScheduledAccessTokenRefresh() {
...
+
+        /**
+         * Block until the wait either completes or is terminated for other reasons.
+         */
+        public void waitForServerChanges() {
+            if (!resultReceived) {
+                try {
+                    waitForChanges.await();
+                } catch (InterruptedException e) {
+                    throw new ObjectServerError(ErrorCode.UNKNOWN, e);
```
beeender: Just log the exception as a warning here, since this happens because of user calls `RealmAsyncTask.cancel()`. It is not an unexpected error.

beeender: Re-thought about this, this might need to throw a checked `InterruptedException` or a special exception for `RealmAsyncTask` returned by `getInstanceAsync()` for cancelling. I will revisit this part after #4570 merged.

cmelchior: I'll log a warning for now then

realm/realm-library/src/objectServer/java/io/realm/SyncConfiguration.java
```diff
@@ -600,13 +600,12 @@ public Builder errorHandler(SyncSession.ErrorHandler errorHandler) {
          * setting enabled should only be opened on background threads or with
          * {@code Realm.getInstanceAsync(config, onSuccess)} on the UI thread.
          * <p>
-         * If a device is offline, the Realm will open immediately.
+         * If this is enabled the Realm will not open until it have talked to the server at least once. This means that
```
Zhuinden: has

realm/realm-library/src/objectServer/java/io/realm/SyncSession.java
```diff
@@ -139,7 +139,7 @@ void notifyAllChangesDownloaded(Long errorcode, String errorMessage) {
      * Calling this method will block until all known remote changes have been downloaded and applied to the Realm.
      * This will involve network access, so calling this method should only be done from a non-UI thread.
      * <p>
-     * If the device is offline, this method will return immediately.
+     * If the device is offline, this method might never return.
```
Zhuinden: well that's quite different! 😄

cmelchior:
`getInstanceAsync` is now on master so this has been updated for final review. Only contention point I have right now is that `SyncSession.downloadAllServerChanges()` is part of the public API. This might or might not be okay (I probably slightly lean towards it is okay, since anybody using the session directly is after power features). If we want to keep it private, we need to move the `SyncObjectServerFacade` from `io.realm.internal` to `io.realm`, which is doable. Let me know your thoughts on this.

Please review @realm/java

realm/realm-library/src/objectServer/java/io/realm/SyncSession.java
```diff
@@ -118,6 +124,63 @@ void close() {
...
+                    if (!waitingForServerChanges.isSuccess()) {
+                        waitingForServerChanges.throwExceptionIfNeeded();
+                    }
+                }
+                waitingForServerChanges = null;
+            }
+        }
+    }
+
+    private void checkMainThread(String errorMessage) {
```
zaki50: `checkMainThread()` should be `checkNotMainThread`.

We already have `check*()` methods such as `checkIfValid()`, `checkHasPrimaryKey()` and most of them throw an exception is the condition in its name is not satisfied.

realm/realm-library/src/objectServer/java/io/realm/SyncSession.java
```diff
@@ -118,6 +124,63 @@ void close() {
...
+                if (!listenerRegistered) {
+                    waitingForServerChanges = null;
+                    throw new ObjectServerError(ErrorCode.UNKNOWN, "It was not possible to download all changes. Has the SyncClient been started?");
+                }
+                waitingForServerChanges.waitForServerChanges();
+
+                // This might return after the session was closed. In that case, just ignore any result
+                if (!isClosed) {
+                    if (!waitingForServerChanges.isSuccess()) {
+                        waitingForServerChanges.throwExceptionIfNeeded();
```
zaki50: no need to clear `waitingForServerChanges` when a exception is thrown here?

realm/realm-library/src/objectServer/java/io/realm/SyncSession.java
```diff
@@ -351,6 +414,56 @@ private void clearScheduledAccessTokenRefresh() {
...
+
+        /**
+         * Block until the wait either completes or is terminated for other reasons.
+         */
+        public void waitForServerChanges() {
+            if (!resultReceived) {
+                try {
+                    waitForChanges.await();
+                } catch (InterruptedException e) {
+                    RealmLog.warn("Downloading changes have been interrupted");
```
zaki50: Why don't we just rethrow `InterruptedException`?
I think that it's better to handle Interrupted status upper layer.

realm/realm-library/src/objectServer/java/io/realm/SyncSession.java
```diff
@@ -351,6 +414,56 @@ private void clearScheduledAccessTokenRefresh() {
...
+        private String errorMessage;
+
+        /**
+         * Block until the wait either completes or is terminated for other reasons.
+         */
+        public void waitForServerChanges() {
+            if (!resultReceived) {
+                try {
+                    waitForChanges.await();
+                } catch (InterruptedException e) {
```
beeender: I think the `waitForServerChanges()` should throws a checked `InterruptedException`instead and move the call this in `RealmCache` to make it works better with `getInstanceAsync()`

realm/realm-library/src/objectServer/java/io/realm/SyncSession.java
```diff
@@ -351,6 +414,56 @@ private void clearScheduledAccessTokenRefresh() {
...
+         * Process the result of a waiting action. This will also unblock anyone who called {@link #waitForChanges}.
+         *
+         * @param errorCode error code if an error occurred, {@code null} if changes were successfully downloaded.
+         * @param errorMessage error message (if any).
+         */
+        public void handleResult(Long errorCode, String errorMessage) {
+            this.errorCode = errorCode;
+            this.errorMessage = errorMessage;
+            this.resultReceived = true;
+            waitForChanges.countDown();
```
beeender: I assume `handlerResult` will always be called even the sync session is closed. But if it is not, then it is more like something should be fixed in the ObjectStore/sync.

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/SyncedRealmTests.java
```diff
@@ -0,0 +1,111 @@
...
+
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+
+
+/**
+ * Catch all class for tests that not naturally fit anywhere else.
+ */
+public class SyncedRealmTests extends BaseIntegrationTest {
+
```
beeender: need tests for `getInstanceAsync()` and cancel the `getInstanceAsync()` if possible.

realm/realm-library/src/objectServer/java/io/realm/SyncSession.java
cmelchior: `SyncSession` needs to be completely thread-safe, and `notifyAllChangesDownloaded` uses the presence of this class as an indicator. This probably wasn't thread safe either though, so I converted it to an AtomicReference instead.

cmelchior:
After internal discussion, we are changing the semantics slightly. `waitForServerChanges` is getting renamed to `waitForInitialData()` and will only be invoked the first time the Realm is opened. This fits the use cases we currently care about better and we can always add something like `waitForServerChanges` later if we need it.

realm/realm-library/src/main/cpp/io_realm_SyncSession.cpp
```diff
@@ -49,3 +54,39 @@ JNIEXPORT jboolean JNICALL Java_io_realm_SyncSession_nativeRefreshAccessToken(JN
...
+
+            bool listener_registered =
+                session->wait_for_download_completion([java_session_object_ref](std::error_code error) {
+                    JNIEnv* env = realm::jni_util::JniUtils::get_env(true);
+                    jni_util::JavaLocalRef<jobject> java_error_code_ref(env, nullptr);
+                    jni_util::JavaLocalRef<jstring> java_error_message_ref(env, nullptr);
+                    if (error != std::error_code{}) {
+                        java_error_code_ref = jni_util::JavaLocalRef<jobject>(env, NewLong(env, error.value()));
+                        java_error_message_ref =
+                            jni_util::JavaLocalRef<jstring>(env, env->NewStringUTF(error.message().c_str()));
```
cmelchior: to_jstring only work on StringData not c strings

realm/realm-library/src/objectServer/java/io/realm/SyncSession.java
cmelchior: Yes, OS guarantee that once the listener has been registered.

realm/realm-library/src/main/cpp/io_realm_SyncSession.cpp
beeender: `std::string` can be implicitly converted to `StringData` https://github.com/realm/realm-core/blob/master/src/realm/string_data.hpp#L84

but yeah, `NewStringUTF`is fine

beeender: uh, when construct `StringData` from `std::string`, a `std::string` instance needs to be retained to make sure the data is valid. So `NewStringUTF` is fine here. just ignore me.

realm/realm-library/src/main/java/io/realm/internal/Context.java
```diff
@@ -1,43 +0,0 @@
-/*
```
zaki50: mistake?

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -281,10 +282,23 @@ private static RealmCache getCache(String realmPath, boolean createIfNotExist) {
...
                 sharedRealm = SharedRealm.getInstance(configuration);
+
+                // If waitForInitialRemoteData() was enabled, we need to make sure that all data is downloaded
+                // before proceeding. We need to open the Realm instance first to start any potential underlying
+                // SyncSession so this will work. TODO: This needs to be decoupled.
+                if (!fileExists) {
+                    try {
+                        ObjectServerFacade.getSyncFacadeIfPossible().downloadRemoteChanges(configuration);
+                    } catch (InterruptedException e) {
+                        throw new DownloadingRealmInterruptedException(e);
```
cmelchior: Yes, good point 👍

realm/realm-library/src/objectServer/java/io/realm/exceptions/DownloadingRealmInterruptedException.java
```diff
@@ -0,0 +1,27 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.realm.exceptions;
+
+import io.realm.SyncConfiguration;
+
+
+public class DownloadingRealmInterruptedException extends RuntimeException {
```
beeender: javadoc for this

realm/realm-library/src/main/java/io/realm/internal/ObjectServerFacade.java
```diff
@@ -88,4 +88,8 @@ public static ObjectServerFacade getSyncFacadeIfPossible() {
     // If no session yet exists for this path. Wrap a new Java Session around an existing OS one.
     public void wrapObjectStoreSessionIfRequired(RealmConfiguration config) {
     }
+
+    public void downloadRemoteChanges(RealmConfiguration config) {
```
beeender: add javadoc to say this will throw then interrupted exception.

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/SyncedRealmTests.java
```diff
@@ -0,0 +1,153 @@
...
+                .build();
+        Random randomizer = new Random();
+
+        for (int i = 0; i < 10; i++) {
+            Thread t = new Thread(new Runnable() {
+                @Override
+                public void run() {
+                    Realm realm = null;
+                    try {
+                        realm = Realm.getInstance(config);
```
beeender: i think it is actually possible that just call `Thread.currentThread().interrupt()` here. As long as the downloading latch is the first latch the `getInstance()` reaches, it will just throw there. So no need to loop 10 times.

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/SyncedRealmTests.java
```diff
@@ -0,0 +1,153 @@
...
+
+    @Rule
+    public RunInLooperThread looperThread = new RunInLooperThread();
+
+    @Rule
+    public final UiThreadTestRule uiThreadTestRule = new UiThreadTestRule();
+
+    @Test
+    @UiThreadTest
+    public void waitForInitialRemoteData_mainThreadThrows() {
```
beeender: Also need a test to check if the realm has been downloaded, then it won't throw this anymore.

beeender: I am actually wondering if it is needed to have an API `Realm.isDownloaded(SyncConfiguration)`.

cmelchior: You can use `new File(syncConfiguration.getPath()).exists()`, but having a more general shorthand through `configuration.realmExists()` is probably not a bad idea.

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/SyncedRealmTests.java
cmelchior: But once you hit the interrupt, the download has already completed?

cmelchior:
Ready for another round of review @realm/java 
Note that naming and functionality changed slightly after discussions with @zaki50 and @beeender.

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -152,13 +156,18 @@ public void run() {
...
                 RealmLog.warn(e, "`CreateRealmRunnable` has been interrupted.");
             } catch (final Throwable e) {
-                RealmLog.error(e, "`CreateRealmRunnable` failed.");
-                notifier.post(new Runnable() {
-                    @Override
-                    public void run() {
-                        callback.onError(e);
-                    }
-                });
+                // DownloadingRealmInterruptedException is treated specially.
```
beeender: Why don't you just:
```java
} catch (DownloadingRealmInterruptedException ignored) {
} catch (final Throwable e) {
// ...
}
```

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/SyncedRealmTests.java
beeender: You can call `Thread.currentThread().interrupt()` before the `Realm.getInstance(config);`, the `latch.await()` will just throw if it found the interrupted flag has been set.

realm/realm-library/src/main/cpp/jni_util/java_local_ref.hpp
```diff
@@ -41,15 +41,26 @@ class JavaLocalRef {
...
+    {
+        this->~JavaLocalRef();
+        new (this) JavaLocalRef(std::move(rhs));
+        return *this;
+    }
+
+    inline JavaLocalRef(JavaLocalRef&& rhs)
+    {
+        m_env = rhs.m_env;
+        m_jobject = (rhs.m_jobject != nullptr) ? m_env->NewLocalRef(rhs.m_jobject) : nullptr;
```
beeender: This will leak `rhs.m_jobject`. well, in this special case, it may not leak the local ref since all the local ref will be released when jni return to java. But in the C++ wise, this create a new local ref which means there are two local refs to the same object, and then the `rhs.m_jobject` is set to null, which is a leak.

Won't code below work?
```c++
inline JavaLocalRef(JavaLocalRef&& rhs) 
    : m_env(rhs.m_env), m_jobject(rhs.m_jobject)
{
    m_jobject = nullptr;
}
```

initialization list is preferred since it will avoid setting m_jojbect twice.

realm/realm-library/src/main/java/io/realm/RealmCache.java
cmelchior: Because `DownloadingRealmInterruptedException` is a Sync exception, so it cannot be referenced directly from Base.

realm/realm-library/src/main/cpp/jni_util/java_global_ref.cpp
```diff
@@ -34,3 +34,8 @@ JavaGlobalRef& JavaGlobalRef::operator=(JavaGlobalRef&& rhs)
     new (this) JavaGlobalRef(std::move(rhs));
     return *this;
 }
+
+JavaGlobalRef::JavaGlobalRef(JavaGlobalRef& rhs)
```
beeender: Sorry, forgot this on in the last review.
```c++
JavaGlobalRef::JavaGlobalRef(JavaGlobalRef& rhs) 
    : m_ref(rhs.m_ref ? jni_util::JniUtils::get_env(true)->NewGlobalRef(rhs.m_ref) : nullptr)
{
}

```

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 * [ObjectServer] Added support for `SyncUser.isAdmin()` (#4353).
 * [ObjectServer] Added support for changing passwords through `SyncUser.changePassword()` (#4423).
+* [ObjectServer] Added support for `SyncConfigration.Builder.waitForInitialRemoteData()` (#4270).
```
dalinaum: SyncConfiguration

realm/realm-library/src/main/java/io/realm/RealmConfiguration.java
```diff
@@ -200,11 +200,25 @@ InputStream getAssetFile() throws IOException {
...
     public String getPath() {
         return canonicalPath;
     }
 
     /**
+     * Checks if the Realm file defined by this configuration already exists.
+     *
+     * @return {@code true} if the Realm file exists, {@code false} otherwise.
+     */
+    public boolean realmExists() {
```
zaki50: This method is useful in some cases, but I don't think that we should make this public API.
Without proper lock, this method does not guarantee anything.

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -260,6 +260,8 @@ private static void checkFilesDirAvailable(Context context) {
      * @throws RealmMigrationNeededException if no migration has been provided by the default configuration and the
      * RealmObject classes or version has has changed so a migration is required.
      * @throws RealmFileException if an error happened when accessing the underlying Realm file.
+     * @throws io.realm.exceptions.DownloadingRealmInterruptedException if {@link SyncConfiguration.Builder#waitForInitialRemoteData()}
```
dalinaum: Is full package needed? `RealmMigrationNeededException` and `RealmFileException` doesn't start with `io.realm.exceptions`.

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -277,6 +279,8 @@ public static Realm getDefaultInstance() {
      * classes or version has has changed so a migration is required.
      * @throws RealmFileException if an error happened when accessing the underlying Realm file.
      * @throws IllegalArgumentException if a null {@link RealmConfiguration} is provided.
+     * @throws io.realm.exceptions.DownloadingRealmInterruptedException if {@link SyncConfiguration.Builder#waitForInitialRemoteData()}
```
dalinaum: Same above.

realm/realm-library/src/main/java/io/realm/RealmConfiguration.java
cmelchior: I kinda see your point, but I'm not entirely sure what kind of locks you think could make this better? This method has the exact same race condition as `new File(config.getPath()).exists()` which is what people would be using today?

But I'll just remove it from the public API for now.

realm/realm-library/src/main/java/io/realm/Realm.java
zaki50: That exception class is in `objectServer` flavor.
We can't use `import` for that.

cmelchior: Yeah, I just discovered that  as well :)

cmelchior:
@zaki This method should work even if no file exists on the server. ROS will currently always create a server side file when a session is created (there is an issue for not doing this in some cases).

It would just synchronise an empty file which would crash with a RealmMigrationException or upgrade the schema automatically depending on the value of `SyncConfiguration.schemaVersion()`. Just like for local Realms copied from assets.

zaki50:
:+1:

zaki50:
@cmelchior https://github.com/realm/realm-java/pull/4536/files#r115412260 there is a typo in changelog

PR #4538, opened 2017-04-23
Add support for changing a users password.

### LINKED ISSUE #4423: Change Password API
Sync Users should be able to change their passwords easily. 

ROS recently merged a new REST endpoint at `/auth/password` for changing the user password to help with this: https://github.com/realm/realm-object-server/pull/1028

One obvious API idea:
```
SyncUser user = SyncUser.login(credentials, user);

// Re-use existing Callback
user.changePassword("newPassword", new Callback() {
       @Override
       public void onSuccess(SyncUser user) {
         // Password successfully changed
       }
       @Override
       public void onError(ObjectServerError error) {
         // Something went wrong
       }
});
```
- [ ] Agree on public API with other bindings
- [x] Determine how much functionality should be in OS ->  None, all REST calls live in the bindings (for now)
- [x] Implement public API + Javadoc
- [x] Unit tests
- [x] Integration tests

cmelchior:
Fixes #4423 

TODO
- [x] Align on if the API should be named `changePassword` or `updatePassword` -> `changePassword` it is


realm/realm-library/src/objectServer/java/io/realm/SyncUser.java
```diff
@@ -221,63 +222,26 @@ public static SyncUser login(final SyncCredentials credentials, final String aut
...
      * @throws IllegalArgumentException if not on a Looper thread.
      */
     public static RealmAsyncTask loginAsync(final SyncCredentials credentials, final String authenticationUrl, final Callback callback) {
-        if (Looper.myLooper() == null) {
-            throw new IllegalStateException("Asynchronous login is only possible from looper threads.");
-        }
-        final Handler handler = new Handler(Looper.myLooper());
-        ThreadPoolExecutor networkPoolExecutor = SyncManager.NETWORK_POOL_EXECUTOR;
-        Future<?> authenticateRequest = networkPoolExecutor.submit(new Runnable() {
+        checkLooperThread("Asynchronous login is only possible from looper threads.");
```
beeender: Use this instead ? `sharedRealm.capabilities.checkCanDeliverNotification("Callback cannot be delivered on current thread.");`

realm/realm-library/src/objectServer/java/io/realm/SyncUser.java
```diff
@@ -346,14 +310,58 @@ protected void onError(LogoutResponse response) {
     }
 
     /**
+     * Change this users password. This is done synchronously and involves the network, so calling this method on the
+     * Android UI thread will always crash.
+     *
+     * @param newPassword the users new password.
+     * @throws IllegalArgumentException if not on a Looper thread.
```
beeender: This restriction is strange, I don't see why user cannot call this method on a non-UI looper thread. Also, calling this on UI thread will surely block the UI thread, but it is user's responsibility to use the async version in this case I think.

realm/realm-library/src/objectServer/java/io/realm/SyncUser.java
```diff
@@ -476,8 +484,68 @@ ObjectServerUser getSyncUser() {
...
+
+        private void postError(final ObjectServerError error) {
+            if (callback != null) {
+                handler.post(new Runnable() {
+                    @Override
+                    public void run() {
+                        try {
+                            callback.onError(error);
+                        } catch (Exception e) {
+                            RealmLog.debug("onError has thrown an exception that is being ignored: %s", Util.getStackTrace(e));
```
beeender: `RealmLog.debug(e, "onError has thrown an exception that is being ignored");`

beeender: Also, the log should be called even when the `callback == null`

realm/realm-library/src/objectServer/java/io/realm/SyncUser.java
cmelchior: Ah yes. Forgot about that one. Much better since it gets rid of the Android reference

realm/realm-library/src/objectServer/java/io/realm/SyncUser.java
cmelchior: Ups, my bad (copy-paste). This does work on a non-looper thread, that is the entire purpose of this method.

realm/realm-library/src/objectServer/java/io/realm/SyncUser.java
cmelchior: Hmm, looking at this. This is actually different than the Realm notifications. This is the SyncUser which can use these functions independently of a Realm.

This indicate we need to abstract the `AndroidRealmNotifier` even further somehow, but it isn't clear to me right now how to do that. Would it be okay to punt this in another issue?

beeender: Actually i met the similar problem right now with `getInstanceAsync()`, so i need a handler even before the Realm created. My currently plan is making an interface `Notifier` with `post()` method, which will only handle sending event.

Similar for the `Capabilities`, it is actually doesn't require a `SharedRealm` instance, maybe pass a `null` for now? I will change it later.

cmelchior: If I can parse `null` as a SharedRealm, I can do that as well 👍

realm/realm-library/src/objectServer/java/io/realm/SyncUser.java
```diff
@@ -489,12 +489,12 @@ ObjectServerUser getSyncUser() {
     private static abstract class Request {
 
         private final Callback callback;
-        private final Handler handler;
+        private final AndroidRealmNotifier handler;
```
Zhuinden: I wonder if that should be called `notifier` instead

realm/realm-library/src/androidTestObjectServer/java/io/realm/SyncUserTests.java
```diff
@@ -268,4 +268,34 @@ public void login_appendAuthSegment() {
             SyncManager.setAuthServerImpl(originalServer);
         }
     }
+
+    @Test
+    public void changePassword_nullThrows() {
+        SyncUser user = createTestUser();
+        try {
+            user.changePassword(null);
```
beeender: `ExpectedException` to make it shorter?

realm/realm-library/src/androidTestObjectServer/java/io/realm/SyncUserTests.java
```diff
@@ -268,4 +268,34 @@ public void login_appendAuthSegment() {
...
+                    fail();
+                }
+
+                @Override
+                public void onError(ObjectServerError error) {
+                    fail();
+                }
+            });
+            fail();
+        } catch (IllegalStateException ignored) {
```
beeender: ditto

realm/realm-library/src/objectServer/java/io/realm/SyncUser.java
```diff
@@ -346,14 +312,56 @@ protected void onError(LogoutResponse response) {
     }
 
     /**
+     * Change this users password. This is done synchronously and involves the network, so calling this method on the
+     * Android UI thread will always crash.
```
beeender: > so calling this method on the Android UI thread will always crash.

Is this still true?

realm/realm-library/src/objectServer/java/io/realm/SyncUser.java
```diff
@@ -476,8 +484,70 @@ ObjectServerUser getSyncUser() {
...
+        // Start the request
+        public RealmAsyncTask start() {
+            Future<?> authenticateRequest = networkPoolExecutor.submit(new Runnable() {
+                @Override
+                public void run() {
+                    try {
+                        postSuccess(Request.this.run());
+                    } catch (ObjectServerError e) {
+                        postError(e);
+                    }
```
beeender: catch other `RuntimeException` and log it.

realm/realm-library/src/objectServer/java/io/realm/SyncUser.java
```diff
@@ -476,8 +484,70 @@ ObjectServerUser getSyncUser() {
...
+
+        private void postError(final ObjectServerError error) {
+            if (callback != null) {
+                handler.post(new Runnable() {
+                    @Override
+                    public void run() {
+                        try {
+                            callback.onError(error);
+                        } catch (Exception e) {
+                            RealmLog.error(e, "onError has thrown an exception that is being ignored.");
```
beeender: I don't think the exception should be ignored here. the exception is mostly caused by user's `onError` callback, and they should just be thrown. remove this catch block?

realm/realm-library/src/objectServer/java/io/realm/SyncUser.java
```diff
@@ -476,8 +484,70 @@ ObjectServerUser getSyncUser() {
...
+                        }
+                    }
+                });
+            } else {
+                RealmLog.error(error, "An error was thrown, but no callback was available");
```
beeender: Just log the error at line 520 before everything. The `post` could return `false` if the caller thread is terminated, in that case, at least we have the error log.

realm/realm-library/src/objectServer/java/io/realm/internal/network/AuthenticationServer.java
```diff
@@ -56,4 +56,9 @@
      * logged out as well.
      */
     LogoutResponse logout(Token userToken, URL authenticationUrl);
+
+    /**
+     * Change a users password.
```
beeender: Change -> Changes

realm/realm-library/src/objectServer/java/io/realm/internal/network/OkHttpAuthenticationServer.java
```diff
@@ -79,47 +81,63 @@ public AuthenticateResponse refreshUser(Token userToken, URI serverUrl, URL auth
...
         }
     }
 
-    private static URL buildLogoutUrl(URL authenticationUrl) {
+    @Override
+    public ChangePasswordResponse changePassword(Token userToken, String newPassword, URL authenticationUrl) {
+        try {
+            String requestBody = ChangePasswordRequest.create(userToken, newPassword).toJson();
+            return changePassword(buildActionUrl(authenticationUrl, ACTION_CHANGE_PASSWORD), requestBody);
+        } catch (Exception e) {
```
beeender: The `Error` will be silently suppressed by the executor service as well i think, catch `Throwable` instead? Or actually maybe it is better that we override `ThreadPoolExecutor.afterExecute` to just throw all uncaught throwables and terminated the process -- when those happen, it cannot be recovered anymore. Then it could be anther PR.

realm/realm-library/src/objectServer/java/io/realm/internal/network/OkHttpAuthenticationServer.java
```diff
@@ -79,47 +81,63 @@ public AuthenticateResponse refreshUser(Token userToken, URI serverUrl, URL auth
...
+    // Builds the URL for a specific auth endpoint
+    private static URL buildActionUrl(URL authenticationUrl, String action) {
         final String baseUrlString = authenticationUrl.toExternalForm();
         try {
-            if (baseUrlString.endsWith("/")) {
-                return new URL(baseUrlString + "revoke");
-            } else {
-                return new URL(baseUrlString + "/revoke");
-            }
+            // Assertion: baseUrlString must not end with `/`
```
beeender: So not end by `/` is checked somewhere?

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/AuthTests.java
```diff
@@ -157,4 +158,40 @@ public void run() {
...
+        String username = UUID.randomUUID().toString();
+        String password = "password";
+        SyncCredentials credentials = SyncCredentials.usernamePassword(username, password, true);
+        SyncUser user = SyncUser.login(credentials, Constants.AUTH_URL);
+        user.logout();
+
+        try {
+            user.changePassword("new-password");
+            fail();
+        } catch (ObjectServerError ignored) {
```
beeender: `ExpectedException`?

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/BaseIntegrationTest.java
```diff
@@ -45,6 +47,7 @@ public static void setUp () throws Exception {
     public static void tearDown () throws Exception {
         try {
             HttpUtils.stopSyncServer();
+            RealmLog.setLevel(LogLevel.WARN);
```
beeender: Save the log level by calling `RealmLog.getLevel()` and restore to that.

realm/realm-library/src/objectServer/java/io/realm/SyncUser.java
cmelchior: I didn't test it this time around, but Android has crashed for some years now if you do networking on the UI thread. I'll add a unit test to verify.

realm/realm-library/src/objectServer/java/io/realm/internal/network/AuthenticationServer.java
Zhuinden: and `user's`

realm/realm-library/src/objectServer/java/io/realm/SyncUser.java
Zhuinden: @beeender `NetworkOnMainThreadException`, yes

realm/realm-library/src/objectServer/java/io/realm/SyncUser.java
cmelchior: This was the pattern we agreeded upon when JNI was involved, but since this is happening in pure Java, I guess throwing the exception would be nicer.

realm/realm-library/src/objectServer/java/io/realm/SyncUser.java
cmelchior: I don't think logging `error` that way is a good idea. Since some people do monitor the log, and it would be a false positive. I changed the logic slightly instead so we should now log in all cases where something else doesn't handle it.

realm/realm-library/src/objectServer/java/io/realm/internal/network/OkHttpAuthenticationServer.java
cmelchior: I'll change to catch `Throwable` 👍 . Your right, it seems the Network thread pool is not propagating errors correctly. I'll create another issue for tracking and fixing that.

realm/realm-library/src/objectServer/java/io/realm/internal/network/OkHttpAuthenticationServer.java
cmelchior: Hmm, I thought it was, but double checked, and apparently we don't normalize the URL. I'll re-add this.

realm/realm-library/src/androidTestObjectServer/java/io/realm/SyncUserTests.java
```diff
@@ -272,30 +280,26 @@ public void login_appendAuthSegment() {
...
 
-                @Override
-                public void onError(ObjectServerError error) {
-                    fail();
-                }
-            });
-            fail();
-        } catch (IllegalStateException ignored) {
-        }
+        thrown.expect(IllegalArgumentException.class);
```
beeender: `IllegalStateException`

realm/realm-library/src/objectServer/java/io/realm/SyncUser.java
```diff
@@ -346,14 +313,56 @@ protected void onError(LogoutResponse response) {
     }
 
     /**
+     * Change this users password. This is done synchronously and involves the network, so calling this method on the
```
kneth: "users" -> "user's"

realm/realm-library/src/objectServer/java/io/realm/SyncUser.java
```diff
@@ -346,14 +313,56 @@ protected void onError(LogoutResponse response) {
...
+     * Android UI thread will always crash.
+     *
+     * @param newPassword the users new password.
```
kneth: "users" -> "user's"

Is `newPassword` in clear text? Or is encrypted somehow? If it is encrypted, encryption method should be mentioned. If not encrypted, it is insanely insecure!!!!

realm/realm-library/src/objectServer/java/io/realm/SyncUser.java
```diff
@@ -346,14 +313,56 @@ protected void onError(LogoutResponse response) {
...
+     * @throws ObjectServerError if the password could not be changed for some reason.
```
kneth: remove "for some reason"

realm/realm-library/src/objectServer/java/io/realm/SyncUser.java
```diff
@@ -346,14 +313,56 @@ protected void onError(LogoutResponse response) {
...
+        }
+        AuthenticationServer authServer = SyncManager.getAuthServer();
+        ChangePasswordResponse response = authServer.changePassword(getSyncUser().getUserToken(), newPassword, getAuthenticationUrl());
+        if (!response.isValid()) {
+            throw response.getError();
+        }
+    }
+
+    /**
+     * Change this users password asynchronously.
```
kneth: "users" -> "user's"

realm/realm-library/src/objectServer/java/io/realm/SyncUser.java
```diff
@@ -346,14 +313,56 @@ protected void onError(LogoutResponse response) {
...
+     *
+     * @param newPassword the users new password.
```
kneth: "users" -> "user's"

realm/realm-library/src/objectServer/java/io/realm/SyncUser.java
```diff
@@ -346,14 +313,56 @@ protected void onError(LogoutResponse response) {
...
 
     /**
+     * Change this users password. This is done synchronously and involves the network, so calling this method on the
+     * Android UI thread will always crash.
+     *
+     * @param newPassword the users new password.
+     * @throws ObjectServerError if the password could not be changed for some reason.
+     */
+    public void changePassword(String newPassword) throws ObjectServerError {
+        if (newPassword == null) {
```
nhachicha: can you use an empty password? what's the password requirement from the server POV? (number of char etc.)
or do you rely on the server returning the error about the size?

realm/realm-library/src/objectServer/java/io/realm/SyncUser.java
```diff
@@ -346,14 +313,56 @@ protected void onError(LogoutResponse response) {
...
+            throw response.getError();
+        }
+    }
+
+    /**
+     * Change this users password asynchronously.
+     *
+     * @param newPassword the users new password.
+     * @param callback callback when login has completed or failed. The callback will always happen on the same thread
+     * as this this method is called on.
```
nhachicha: > this this

realm/realm-library/src/objectServer/java/io/realm/SyncUser.java
cmelchior: Yes, I didn't want to encode any requirements on the client side. The server should return an error if the password isn't good enough. It is arguable if we should even throw on `null`

Zhuinden: You should probably throw on null, that's most likely a user error at that point.

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/AuthTests.java
```diff
@@ -145,16 +144,40 @@ public void onSuccess(SyncUser user) {
             @Override
             public void onError(ObjectServerError error) {
                 assertEquals(ErrorCode.INVALID_CREDENTIALS, error.getErrorCode());
-                throw new IllegalArgumentException("BOOM");
+                looperThread.testComplete();
```
cmelchior: Point taken. I rewritten the test so it now test what the method actually says it does.

realm/realm-library/src/objectServer/java/io/realm/SyncUser.java
```diff
@@ -321,15 +288,15 @@ public void logout() {
             syncUser.clearTokens();
             syncUser.localLogout();
 
-            // Finally revoke server token. The local user is logged out in any case.
+            // Finally create server token. The local user is logged out in any case.
```
cmelchior: Bad refactor

cmelchior:
Updated. @kneth I added a section about security to the changePassword methods.

realm/realm-library/src/objectServer/java/io/realm/SyncUser.java
```diff
@@ -101,7 +106,7 @@ private SyncUser(ObjectServerUser user) {
      * A user is invalidated when he/she logs out or the user's access token expires.
      *
      * @return current {@link SyncUser} that has logged in and is still valid. {@code null} if no user is logged in or the user has
-     *         expired.
+     * expired.
```
dalinaum: Does we indent lines only for parameter descriptions? If not, we should update https://github.com/realm/realm-java/blob/master/CONTRIBUTING.md

realm/realm-library/src/objectServer/java/io/realm/SyncUser.java
```diff
@@ -221,63 +225,26 @@ public static SyncUser login(final SyncCredentials credentials, final String aut
      * @param credentials credentials to use.
      * @param authenticationUrl server that the user is authenticated against.
      * @param callback callback when login has completed or failed. The callback will always happen on the same thread
-     *                 as this this method is called on.
+     * as this this method is called on.
```
dalinaum: Parameter descriptions should be indented.

realm/realm-library/src/objectServer/java/io/realm/SyncUser.java
dalinaum: Nvm. Sorry. I've misread this line.

realm/realm-library/src/objectServer/java/io/realm/SyncUser.java
```diff
@@ -346,14 +313,62 @@ protected void onError(LogoutResponse response) {
     }
 
     /**
+     * Change this user's password. This is done synchronously and involves the network, so calling this method on the
```
dalinaum: Changes this user's password?

realm/realm-library/src/objectServer/java/io/realm/SyncUser.java
```diff
@@ -346,14 +313,62 @@ protected void onError(LogoutResponse response) {
...
+        }
+        AuthenticationServer authServer = SyncManager.getAuthServer();
+        ChangePasswordResponse response = authServer.changePassword(getSyncUser().getUserToken(), newPassword, getAuthenticationUrl());
+        if (!response.isValid()) {
+            throw response.getError();
+        }
+    }
+
+    /**
+     * Change this user's password asynchronously.
```
dalinaum: Changes this user's password?

realm/realm-library/src/objectServer/java/io/realm/SyncUser.java
cmelchior: The new auto-formatter formats it this way. We need to fix the formatter instead then, which is outside the scope of this PR.

PR #4547, opened 2017-04-25
Fix schemaVersion docs for synced Realms

### LINKED ISSUE #4466: Does realm provides a "needs update schema / version" notification for synched realms?
Hello,
I'm using a SyncServer that already has a realm with version 3.0.0.
After the release of realm-java 3.1.1 I want to update my code and my schemas but I have the problem of the right mechanism to do it on clients and the server without losing data. 
I don't know if the following sequence is right:
-First Case:
Synched realm version: 3.0.0. Synched  Realm schema version 1. Client Local realm version: 3.0.0. Realm Schema version 1.
The client already has a right "SyncUser.currentUser" so it can connects and syncs with the synched realm

-Second Case
Synched realm version: 3.0.0. Synched  Realm schema version 1. Client Local realm version: 3.1.1. Realm Schema version 1.
The first time the client tries to connect to the synched realm with the currentUser, it will upgrade synched version to 3.1.1. Is that right? And since then the sync version is 3.1.1

-Third Case
Synched realm version: 3.1.1. Synched  Realm schema version 1. Client Local realm version: 3.0.0. Realm Schema version 1.
This is a client that haven upgrade the app so it holds and old realm version. And tries to connect to the synched realm version. what is going to happen here?

Thank you in advance.
Gabriel

cmelchior:
Fixes #4466

### DIFF
--- a/realm/realm-library/src/objectServer/java/io/realm/SyncConfiguration.java
+++ b/realm/realm-library/src/objectServer/java/io/realm/SyncConfiguration.java
@@ -471,6 +471,9 @@ public class SyncConfiguration extends RealmConfiguration {
          * <p>
          * Failing to increment the schema version will cause Realm to throw a {@link io.realm.exceptions.RealmMigrationNeededException}
          * when the Realm is opened and the changed schema will not be applied.
+         * <p>
+         * <b>WARNING:</b> There is no guarantee that the value inserted here is the same returned by {@link Realm#getVersion()}.
+         * Due to the nature of synced Realms, the value can both be higher and lower.
          *
          * @param schemaVersion the schema version.
          * @throws IllegalArgumentException if schema version is invalid.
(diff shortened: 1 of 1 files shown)


realm/realm-library/src/objectServer/java/io/realm/SyncConfiguration.java
```diff
@@ -464,8 +463,14 @@ public Builder encryptionKey(byte[] key) {
...
-         * Sets the schema version of the Realm. This must be equal to or higher than the schema version of the existing
-         * Realm file, if any. If the schema version is higher than the already existing Realm, a migration is needed.
+         * Sets the schema version of the Realm.
+         * <p>
+         * While synced Realms only support additive schema changes which can be applied without requiring a manual
+         * migration, the schema version must still be incremented as an indication to Realm that the change was
+         * intentional.
+         * <p>
+         * Failing to increment the schema version will cause Realm to throw a {@link io.realm.exceptions.RealmMigrationNeededException}
+         * when the Realm is opened and the changed schema will not be applied.
```
beeender: What if the server side Realm schema version is higher than the client side? That needs to be documented here as well.

cmelchior: Done

PR #4549, opened 2017-04-25
Add manualClean task, which works without requiring any dependency artifacts

zaki50:
Current `clean` task depends on the artifacts of the project itself. That means `clean` task only works after successful build.

I've added `manualClean` task, which removes all build files without requiring any artifacts.

@realm/java

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -289,6 +289,12 @@ task manualClean {
             workingDir "${rootDir}"
             commandLine 'find', '.', '-mindepth', '2', '-type', 'd', '-name', '.gradle', '-print', '-exec', 'rm', '-rf', '{}', ';', '-prune'
         }
+
+        // clean ${System.env.HOME}/.m2/repository/io/realm
+        exec {
+            workingDir "${rootDir}"
+            commandLine 'sh', '-c', "echo \"${System.env.HOME}/.m2/repository/io/realm\" && rm -rf \"${System.env.HOME}/.m2/repository/io/realm\""
+        }
     }
 }
 
--- a/build.gradle
+++ b/build.gradle
@@ -267,6 +267,31 @@ task clean {
     dependsOn cleanLocalMavenRepos
 }
 
+task manualClean {
+    description = 'Clean build files manually, which means that this task works without clean tasks'
+    group = 'Clean'
+
+    doLast {
+        // clean 'build' directories
+        exec {
+            workingDir "${rootDir}"
+            commandLine 'find', '.', '-type', 'd', '-name', 'build', '-print', '-exec', 'rm', '-rf', '{}', ';', '-prune'
+        }
+
+        // clean '.externalNativeBuild' directories
+        exec {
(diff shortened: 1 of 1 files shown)


emanuelez:
Are we ok to make this not work on Windows?

zaki50:
I don't think current build.gradle works on Windows...

build.gradle
```diff
@@ -267,6 +267,31 @@ task clean {
     dependsOn cleanLocalMavenRepos
 }
 
+task manualClean {
+    description = 'Clean build files manually, which means that this task works without clean tasks'
```
cmelchior: This doesn't look right.

build.gradle
```diff
@@ -267,6 +267,31 @@ task clean {
...
+    group = 'Clean'
+
+    doLast {
+        // clean 'build' directories
+        exec {
+            workingDir "${rootDir}"
+            commandLine 'find', '.', '-type', 'd', '-name', 'build', '-print', '-exec', 'rm', '-rf', '{}', ';', '-prune'
```
kneth: Can't you use `-delete` instead of `-exec`? Does it collide with `-prune`?

zaki50: I didn't know `-delete` option.

Then I tried `find . -type d -name build -delete` , but `build` directories are not deleted.
What's wrong?

build.gradle
zaki50: I've updated the description. Is that what you mean?

build.gradle
kneth: `-delete` is only deleting the contents of the directory, while your `-exec rm` deletes the directory (and its contents).

zaki50: Ah, I got it.

zaki50:
@realm/java Sorry, I've add a code to remove `${System.env.HOME}/.m2/repository/io/realm`.
Could you review it?

build.gradle
```diff
@@ -267,6 +267,37 @@ task clean {
...
+        // clean '.gradle' directories except one in the root
+        exec {
+            workingDir "${rootDir}"
+            commandLine 'find', '.', '-mindepth', '2', '-type', 'd', '-name', '.gradle', '-print', '-exec', 'rm', '-rf', '{}', ';', '-prune'
+        }
+
+        // clean ${System.env.HOME}/.m2/repository/io/realm
+        exec {
+            workingDir "${rootDir}"
+            commandLine 'sh', '-c', "echo \"${System.env.HOME}/.m2/repository/io/realm\" && rm -rf \"${System.env.HOME}/.m2/repository/io/realm\""
```
beeender: `"echo \"${System.env.HOME}/.m2/repository/io/realm\"` doesn't seem to be useful. Does it?

zaki50: You mean that also `-print` in `find` should be removed?

beeender: uh, you just want to print the path here, got it!

Issue #2299, opened 2016-02-18
Realm.getInstanceAsync()


beeender:
**Edit by CM**: Changing the focus slightly of this issue. Discussions around exposing info and callbacks for custom migrations have been moved to #4083 

---

As described in #4270 we need to expose a `Realm.getInstanceAsync(config)` in order to allow synced Realms to be downloaded first.

**Usage**
```java
// Create configuration
RealmConfiguration config = new RealmConfiguration.Builder()
  .migration(new MyMigration())
  .schemaVersion(42)
  .build();

// Get instance on a Looper thread
Realm.getInstanceAsync(config, new Realm.Callback() {
  public void OnSuccess(Realm realm) {
    // Use Realm
  }
  // Do we really want to force both onSuccess and onError?
  // Perhaps make Realm.Callback abstract and provide default impl for this?
  // Should we instead provide `getInstanceAsync(config, onSuccess, onError)`?
  public void onError(Exception e) {
    // Something went wrong? Migration failed, problems opening file?
  }
})
```

**TODO**
This issue is therefore now about the following:

- [x] Add new API `Realm.getInstanceAsync(config, callback)` and `DynamicRealm.getInstanceAsync(config, callback)`

- [x] If used, all schema validation and migration logic should run on the background thread to avoid blocking the UI thread at all. Once ready the caller thread should be allowed to open it's SharedRealm. I don't think there is any need to handover SharedRealm instances. I doubt it is supported in any case.

- [x] Should throw if called on a non-looper thread.
- [x] Unit tests
- [ ] Website docs

----

**Old description**
Since the migration could take some time it might block the UI, users probably needs to show a progress bar when that happens.
Also, an async `Realm.migrate(RealmConfiguration config, RealmCallback onSuccess)` will be very handy to this situation.

cmelchior:
Adding a asynchronous initializer for Realm would also benefit startup times, and I have been thinking something along the same lines, but wouldn't an asynchronous `getInstance` be enough? Is there any use case for people needing to know when a migration is happening?

```
Realm.getInstanceAsync(config, onSuccess); 
Realm.getDefaultInstanceAsync(config, onSuccess); // We probably also need this one then?
```
vs.
```
Realm.isMigrationNeeded(config);
Realm.migrateRealm(config, onSuccess);

// Usage
if (Realm.isMigrationNeeded(config)) {
  Realm.migrateRealm(config, onsuccess);
} else {
  Realm.getInstance(config);
}
```

[...]

PR #4551, opened 2017-04-25
Fine grained locks for RealmCache

beeender:
- Separated lock for different RealmConfiguration instead of one lock on
  the RealmCache class. So Opening Realm instances from different
  configurations won't block each other.
- DynamicRealm which is created during opening type Realm will not be
  associated to any RealmCache to avoid recursive locks and multiple
  times initial block. (Also make the code easier.)

This is for #4536 and part of implementation of #2299 .


realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -69,20 +68,33 @@
 
     final long threadId;
     protected final RealmConfiguration configuration;
+    // Which RealmCache is this Realm associated to. It could be null if the Realm instance is opened without putting it
```
cmelchior: It is null if the Realm instance is opened without being put into a cache. It is also null if the Realm is closed.

realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -605,7 +625,9 @@ public void onResult(int count) {
                 RealmMigration realmMigration = (migration == null) ? configuration.getMigration() : migration;
                 DynamicRealm realm = null;
                 try {
-                    realm = DynamicRealm.getInstance(configuration);
+                    // Create a DynamicRealm WITHOUT putting it into a RealmCache to avoid recursive locks and call init
+                    // steps multiple times (copy asset file / initialData transaction).
+                    realm = DynamicRealm.createInstance(configuration);
```
cmelchior: Wouldn't this theoretically make it possible to open a Realm inside a migration block with a different configuration for the same path? If there are clear benefits of doing it this way and it makes the code much cleaner, not catching that particular error is probably fine.

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -254,16 +254,13 @@ public static void removeDefaultConfiguration() {
...
-     * from the Realm file.
+     * @param cache the {@link RealmCache} where to create the realm in.
      * @return a {@link Realm} instance.
      */
-    static Realm createInstance(RealmConfiguration configuration, ColumnIndices[] globalCacheArray) {
+    static Realm createInstance(RealmCache cache) {
+        RealmConfiguration configuration = cache.getConfiguration();
         try {
-            return createAndValidate(configuration, globalCacheArray);
+            return createAndValidate(cache);
```
cmelchior: Something about the naming of this method/variable bugs me, but I cannot set the finger on it. I guess at this point it sounds like we are validating the cache, instead of a new Realm instance.

What about `createAndValidateRealmFromCache(cache)` or something like that?

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -76,7 +76,8 @@ static RealmCacheType valueOf(Class<? extends BaseRealm> clazz) {
     // Separated references and counters for typed Realm and dynamic Realm.
     private final EnumMap<RealmCacheType, RefAndCount> refAndCountMap;
 
-    final private RealmConfiguration configuration;
+    // This will be only valid if refAndCount.globalCount > 0
```
cmelchior: You mean `valid` -> `!= null` ?

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -85,13 +86,14 @@ static RealmCacheType valueOf(Class<? extends BaseRealm> clazz) {
 
     // Realm path will be used as the key to store different RealmCaches. Different Realm configurations with same path
     // are not allowed and an exception will be thrown when trying to add it to the cache map.
+    // The RealmCache won't be removed from the map even when all Realm instances associated to it get closed. Instead,
```
cmelchior: What case is optimised by this? Not that I expect people to open 1000's of Realms, but some apps might be opening quite a lot of different Realms (think synced chat app with a Realm pr. user/channel)

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -105,25 +107,29 @@ private RealmCache(RealmConfiguration config) {
...
+                cache = new RealmCache();
+                cachesMap.put(configuration.getPath(), cache);
+            }
         }
 
-        RefAndCount refAndCount = cache.refAndCountMap.get(RealmCacheType.valueOf(realmClass));
+        return cache.doCreateRealmOrGetFromCache(configuration, realmClass);
+    }
+
+    private synchronized <E extends BaseRealm> E doCreateRealmOrGetFromCache(RealmConfiguration configuration,
```
cmelchior: You don't seem to gain much by moving the synchronized method from `createRealmOrGetFromCache` to `doCreateRealmOrGetFromCache` ?

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -134,6 +140,12 @@ private RealmCache(RealmConfiguration config) {
                 }
             }
             sharedRealm.close();
+
+            // We are holding the lock, and we can set the invalidated configuration since there is no global ref to it.
+            this.configuration = configuration;
+        } else {
+            // Throws the exception if validation failed.
```
cmelchior: Throws exception if validation failed

realm/realm-library/src/main/java/io/realm/BaseRealm.java
beeender: `createInstance` is package protected, so user shouldn't call it to create Realm bypass cache. But if user use `Realm.getInstance()` in the migration block, the cache counter actually still works like before, so it is just the same like before.

realm/realm-library/src/main/java/io/realm/RealmCache.java
beeender: `createRealmOrGetFromCache` is a static method, before change, the lock is on the `RealmCache.class` for the whole method. Now, the lock on the `RealmCache.class` only protect the adding to map, the others are protected by the lock on the `RealmCache` instance. `doCreateRealmOrGetFromCache` is a member method now.

realm/realm-library/src/main/java/io/realm/RealmCache.java
beeender: add comments.

realm/realm-library/src/main/java/io/realm/RealmCache.java
beeender: Not removing it just because of lock on `RealmCache.class` in `release` makes everything much much complex. See the new commit which is using a weak ref to release unused cache from map.

CHANGELOG.md
```diff
@@ -12,6 +12,8 @@
 
 ### Internal
 
+* Use separated locks for different `RealmCache` ($4551).
```
cmelchior: Use separate locks for different `RealmCache`s

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -76,7 +78,10 @@ static RealmCacheType valueOf(Class<? extends BaseRealm> clazz) {
     // Separated references and counters for typed Realm and dynamic Realm.
     private final EnumMap<RealmCacheType, RefAndCount> refAndCountMap;
 
-    // This will be only valid if refAndCount.globalCount > 0
+    // This will be only valid if refAndCount.globalCount > 0.
```
cmelchior: Not sure what you mean by valid in this context? `not null` or something else?

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -85,45 +91,64 @@ static RealmCacheType valueOf(Class<? extends BaseRealm> clazz) {
...
+    }
 
-        RefAndCount refAndCount = cache.refAndCountMap.get(RealmCacheType.valueOf(realmClass));
+    private synchronized <E extends BaseRealm> E doCreateRealmOrGetFromCache(RealmConfiguration configuration,
+            Class<E> realmClass) {
+
+        RefAndCount refAndCount = refAndCountMap.get(RealmCacheType.valueOf(realmClass));
 
         if (refAndCount.globalCount == 0) {
+            copyAssetFileIfNeeded(configuration);
```
beeender: Bug here, need to check the dyn global count + typed global count == 0 here.

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -85,45 +91,64 @@ static RealmCacheType valueOf(Class<? extends BaseRealm> clazz) {
...
-    private RealmCache(RealmConfiguration config) {
-        configuration = config;
+    private RealmCache() {
         refAndCountMap = new EnumMap<>(RealmCacheType.class);
         for (RealmCacheType type : RealmCacheType.values()) {
             refAndCountMap.put(type, new RefAndCount());
         }
     }
 
+    private static synchronized RealmCache getCache(String realmPath, boolean createIfNotExist) {
```
bmeike: Why is this synchronized?  It looks like it might is because there is a concern about multi-threaded access to `cacheMap`.  If that is the case, then you *must* synchronize all other access to `cacheMap` on the same lock... and I suggest that you might consider using a lock other than the Class object.

bmeike:
I realize that it is late in the game, but I really think that it would be possible to improve the design of `BaseRealm`.

realm/realm-library/src/main/java/io/realm/RealmCache.java
beeender: Yes, this function is the only place accessing the `cacheMap`. I can change it to lock on the `cacheMap` itself.

beeender:
@bmeike The design of `Realm`/`DynamicRealm`/`RealmCache` is a long story,  and also be noticed that there is a similar thing inside the Object Store which is doing the same for caching Realm instances called `RealmCoordinator`. We are not using it just because of `RealmCache` was added before that existed in Object Store, and there would be some breaking changes I would like to have to redesign the Realm instances mechanism:

- Semantics of `Realm.close()`
 Current `Realm.close()` is a confusing API, eg:
```java
  Realm realm1 = Realm.getDefaultInstance();
  Realm realm2 = Realm.getDefaultInstance();
  realm1.close();
  print("Is realm1 closed?" + realm1.isClosed()) // false, confusing
  realm2.close();
  print("Is realm1 closed?" + realm1.isClosed()) // true, OK, the thread ref count reaches 0
```
  it would be much clearer if:
```java
  Realm realm1 = Realm.getDefaultInstance();
  Realm realm2 = Realm.getDefaultInstance();
  realm1.close();
  print("Is realm1 closed?" + realm1.isClosed()) // true
  print("Is realm2 closed?" + realm2.isClosed()) // false
  realm2.close();
  print("Is realm2 closed?" + realm1.isClosed()) // true
```
  We can just using the OS `RealmCoordinator` to achieve this quite easier, then also it is possible to have more debug information for users to check which specific Realm instance is leaked.

- `DynamicRealm` and `Realm` should 
  Currently `DynamicRealm` and `Realm` can be created at the same time, and even on the same thread, they are using different `SharedRealm` instance. There was a reason that which i don't recall :) And that creates complexity for the cache management. Also, it creates lot more corner cases we are not handled well right now: think about Realm instance and DynamicRealm instance both exist, then the DynamicRealm changes the schema and typed Realm won't be aware of that, the column indices will be wrong then. (I have seen a user on SO is doing that)
  This behavior should be change, allowing them both exist is not a valid use case at all, we should just disallow that.

- Garbage collect Realm instance
  by fixing the semantics of `Realm.close()`, we might be able to support GC better which is always a headache for users to remember to call `close()`. But this need to be investigated more since I don't really know if the GC is stressed enough to collect un-refed Realm instances on time

- `Realm.deleteRealm()` 
  This API heavily relies on the `RealmCache` and it is not multi-processes safe. It is not ideal at all. But actually the implementation of it should exist in core to have a proper multi-processes support. See core issue https://github.com/realm/realm-core/issues/2165

So, yes, redesign is something I want to do as well, but then we need to introduce some breaking changes in the API level (and I think they are good breaking changes) also a few dependencies from others.

bmeike:
Approving changes.  Please, though, open an issue to fix this.

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -85,45 +91,66 @@ static RealmCacheType valueOf(Class<? extends BaseRealm> clazz) {
 
     // Realm path will be used as the key to store different RealmCaches. Different Realm configurations with same path
     // are not allowed and an exception will be thrown when trying to add it to the cache map.
-    private static final Map<String, RealmCache> cachesMap = new HashMap<>();
+    // A weak ref is used to hold the RealmCache instance. The weak ref entry will be cleared if and only if there
+    // is no Realm instance holding a strong ref to it and there is no Realm instance associated it is BEING created.
+    private static final Map<String, WeakReference<RealmCache>> cachesMap =
```
zaki50: Now we always iterate all entries when we get an entry from the map.
I think `ArrayList` instead of `HashMap` is enough.

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -222,20 +243,18 @@ static synchronized void release(BaseRealm realm) {
...
 
             // No more local reference to this Realm in current thread, close the instance.
             realm.doClose();
 
-            // No more instance of typed Realm and dynamic Realm. Remove the configuration from cache.
-            if (totalRefCount == 0) {
-                cachesMap.remove(canonicalPath);
+            // No more instance of typed Realm and dynamic Realm.
+            if (getTotalGlobalRefCount() == 0) {
+                // We keep the cache in the map even when its global counter reaches 0. It will be reused when next time
```
zaki50: Which map does the `in the map` mean? `cachesMap`?

realm/realm-library/src/main/java/io/realm/RealmCache.java
zaki50: I don't think we can use weak reference here.
If a user lost all reference to a Realm instance without calling `close()`,  corresponding `RealmCache` instance seems to be GCed since the strong reference to the `RealmCache`is only exists in the Realm instance.

beeender: The first version is not using the weak ref, and the `RealmCache` will not be cleared, instead, it will be re-used next time. 
I think the case you mentioned is more important than clearing unused cache from memory since we want to show the debug information about how many instances are not closed. I will revert the weak ref commit.

zaki50: How about using `AtomicReference` instead?

zaki50: Ah, no. It doesn't work..

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -85,45 +91,66 @@ static RealmCacheType valueOf(Class<? extends BaseRealm> clazz) {
...
+        return cache.doCreateRealmOrGetFromCache(configuration, realmClass);
+    }
+
+    private synchronized <E extends BaseRealm> E doCreateRealmOrGetFromCache(RealmConfiguration configuration,
+            Class<E> realmClass) {
 
-        RefAndCount refAndCount = cache.refAndCountMap.get(RealmCacheType.valueOf(realmClass));
+        RefAndCount refAndCount = refAndCountMap.get(RealmCacheType.valueOf(realmClass));
+
+        if (getTotalGlobalRefCount() == 0) {
```
Zhuinden: I think `globalRefCount()` could be 0 if `Realm realm = Realm.getDefaultInstance(); realm.close();`

realm/realm-library/src/main/java/io/realm/RealmCache.java
zaki50: No, it diesn't work..

beeender: I got some other idea, having a `RealmCache.leakedCachesList`, and add the `RealmCache` to the list in the `BaseRealm.finalize()` if it was not closed properly. So for users who called the `Realm.close()`, the weak ref will just be cleaned when they are not needed, but if user doesn't close it, the cache will be there for debugging.

realm/realm-library/src/main/java/io/realm/RealmCache.java
beeender: `configuration` could be non-null if `Realm.getDefaultInstance()` throws an exception before returning the instance. In that case, `globalRefCount()` is 0.

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -164,17 +193,15 @@ private RealmCache(RealmConfiguration config) {
             if (realmClass == Realm.class && refAndCount.globalCount == 0) {
```
beeender: You are right, since `localCount` will be reset when `release()`. But i actually prefer the current code, it is easier to read without knowing `localCount` will be reset :P

realm/realm-library/src/main/java/io/realm/RealmCache.java
beeender: yes, `cahcesMap`! will change that to a list.

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -453,4 +474,23 @@ private static int storeColumnIndices(ColumnIndices[] array, ColumnIndices colum
...
+    public ColumnIndices[] getTypedColumnIndicesArray() {
+        return typedColumnIndicesArray;
+    }
+
+    /**
+     * @return the total global ref count.
+     */
+    private int getTotalGlobalRefCount() {
+        int totalRefCount = 0;
+        for (RealmCacheType type : RealmCacheType.values()) {
```
beeender: `refAndCountMap` is an `EnumMap`, and
> EnumMap is implemented with a pair of arrays of the same length as the number of enum entries.

So `EnumMap.values()` just return the values array without creating anything:
```
    public Collection<V> values() {
        Collection<V> vs = values;
        if (vs != null)
            return vs;
        else
            return values = new Values();
    }
```

I just knew this, interesting. I guess no need to change this.

zaki50: But you are using Enum.vaues() (RealnCachType.values())?

beeender: Yeah, right, i will change that to iterate on `refAndCountMap.values()` instead. Thanks!

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -76,7 +78,11 @@ static RealmCacheType valueOf(Class<? extends BaseRealm> clazz) {
     // Separated references and counters for typed Realm and dynamic Realm.
     private final EnumMap<RealmCacheType, RefAndCount> refAndCountMap;
 
-    final private RealmConfiguration configuration;
+    // This will be only valid if refAndCount.globalCount > 0.
```
beeender: This should be `getTotalGlobalRefCount() > 0`, will change the comments.

realm/realm-library/src/main/java/io/realm/RealmCache.java
cmelchior: Good idea 👍

beeender:
@realm/java updated.

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -489,4 +502,26 @@ public RealmConfiguration getConfiguration() {
...
+        int totalRefCount = 0;
+        for (RefAndCount refAndCount : refAndCountMap.values()) {
+            totalRefCount += refAndCount.globalCount;
+        }
+
+        return totalRefCount;
+    }
+
+    /**
+     * If a Realm instance is GCed but `Realm.close()` is not called before, we still want to track the cache for
```
cmelchior: Right now that information isn't available to users though?

beeender: It is available. First there will be a log to say Realm instance is not closed.
then user can call `Realm.getLocalInstanceCount()` and `Realm.getGlobalInstanceCount()` for how many references are not released just like before.

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -76,64 +81,112 @@ static RealmCacheType valueOf(Class<? extends BaseRealm> clazz) {
...
-            sharedRealm.close();
+
+            // We are holding the lock, and we can set the invalidated configuration since there is no global ref to it.
+            this.configuration = configuration;
+        } else {
+            // Throws exception if validation failed.
+            validateConfiguration(configuration);
         }
 
         if (refAndCount.localRealm.get() == null) {
```
zaki50: if `refAndCount.localRealm.get() != null` and `refAndCount.localCount.get() == 0` here, 
`refAndCount.localRealm.get()` might return an instance which is already closed (for reuse)?

Can we resurrect such a instance?

beeender: `refAndCount.localRealm.get() != nul`l and `refAndCount.localCount.get() == 0` should never happen, but i see your point, it is more readable if i merge the block 212-220 to line 192 as you mentioned before.

zaki50: Why not happen? 

How about following case?
```
Realm.getDefaultInstance().close();
Realm r = Realm.getDefaultInstance();
```

zaki50: Ah, `localRealm` is cleared in `RealmCache.release()`. I got it.

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -504,4 +514,14 @@ private int getTotalGlobalRefCount() {
...
         return totalRefCount;
     }
+
+    /**
+     * If a Realm instance is GCed but `Realm.close()` is not called before, we still want to track the cache for
+     * debugging. Adding them to the list to keep the strong ref of the cache to prevent the cache gets GCed.
+     */
+    void leak() {
+        if (!isLeaked.getAndSet(true)) {
+            leakedCaches.add(this);
```
zaki50: The following is my understanding. Is that correct?

If a user writes code below, `leakedCaches` keeps a strong reference to the `RealmCache` instance and it never release it. Of course the referenced `RealmCache` instance never be GCed, but it works as expected (== Realm instance in it is GCed) even if this kind of leak happens.

```
Realm.getDefaultInstance();
System.gc(); // leak() is called.
Realm r = Realm.getDefaultInstance();
r.close();
r.close();
```

beeender: hmmm, this case is interesting, actually the behaviour is the same even before this change:
We are holding `ThreadLocal` to the Realm instance, the `ThreadLocal` is holding a weak ref to it underneath. So when above happens, the 3rd line (`Realm r = Realm.getDefaultInstance();`) will just return a `null` instance since the weak ref to the Realm instance has been just released. 

So I think it is better we just throw exception when we detect this leak to prevent user from getting a `null` instance?

beeender: aha, that is actually the case we discussed in https://github.com/realm/realm-java/pull/4551#discussion_r114300963 !!!
So shall we let user continue in this case?

zaki50: I think the `ThreadLocal` keeps strong ref.

beeender: Would throwing in this case be a breaking change? I think it is ...

beeender: hmm, the key (thread id?) is weak ref
```
        static class Entry extends WeakReference<ThreadLocal> {
            /** The value associated with this ThreadLocal. */
            Object value;

            Entry(ThreadLocal k, Object v) {
                super(k);
                value = v;
            }
        }
```

So the case you show won't happen as long as the thread is not terminated? `leak()` won't be called in that case.

beeender: OK, checked the `ThreadLocal` implementation, I think the `leak()` will only be called when:
1. User doesn't call `Realm.close()` and doesn't hold any strong ref the Realm instance.
2. The thread where the Realm instances is created on is terminated.

zaki50: Yes! That is perfect.

zaki50:
👏

PR #4556, opened 2017-04-26
Use target table to create snapshot from LinkView

### LINKED ISSUE #4554: `createSnapshot()` causes parent object to be deleted
#### Goal

Delete the child Feed here:

```java
// Empty database before this.

getRealm().executeTransaction(new Realm.Transaction() {
    @Override
        App app = realm.copyToRealm(App.create((long) 0));
        Feed feed = realm.copyToRealm(Feed.create((long) 23));
        app.feeds.add(feed);
    }
});

final App app = getRealm().where(App.class).equalTo("id", 0).findFirst();
final List<Feed> appFeeds = app.feeds.createSnapshot();

getRealm().beginTransaction();
appFeeds.get(0).deleteFromRealm();
getRealm().commitTransaction();

List<App> apps = getRealm().where(App.class).findAll();
List<Feed> feeds = getRealm().where(Feed.class).findAll();

Log.i("LOG", "apps " + apps);
Log.i("LOG", "feeds " + feeds);
```

#### Expected Results

Child deleted. I works as intended **without** `createSnapshot()`.

```
I/LOG: apps [App = [{id:0},{feeds:RealmList<Feed>[0]}]]
I/LOG: feeds []
```

#### Actual Results

Parent deleted.

```
I/LOG: apps []
I/LOG: feeds [Feed = [{id:23}]]
```

#### Classes

```java
public class App extends RealmObject {

    public static App create(Long id) {
        App app = new App();
        app.id = id;
        app.feeds = new RealmList<>();
        return app;
    }

    @PrimaryKey
    public Long id;

    public RealmList<Feed> feeds;
}
```
```java
public class Feed extends RealmObject {

    public static Feed create(Long id) {
        Feed feed = new Feed();
        feed.id = id;
        return feed;
    }

    @PrimaryKey
    public Long id;
}
```

Realm version: 3.1.3

beeender:
- Fix #4554 .
- Remove useless confusing `LinkView.getTable()`.


realm/realm-library/src/main/java/io/realm/internal/Collection.java
```diff
@@ -367,7 +367,7 @@ public Collection(SharedRealm sharedRealm, LinkView linkView, SortDescriptor sor
 
         this.sharedRealm = sharedRealm;
         this.context = sharedRealm.context;
-        this.table = linkView.getTable();
+        this.table = linkView.getTargetTable();
```
zaki50: This constructor seems to be used not only in snapshot but in constructors of `RealmList`. Is this change valid for `RealmList` as well?

Zhuinden: Good question, possibly a unit test for both cases is a good idea

beeender: Yes, it is. it is used to construct the `RealmList.collection`, the `RealmList.collection` is only used for create/remove listeners. listeners won't access the target table, so the bug hasn't been exposed from there.

PR #4558, opened 2017-04-26
Permission API's

### LINKED ISSUE #4296: Expose Permission Realm
Expose the Permission Realm in a similar way to Cocoa: https://github.com/realm/realm-cocoa/pull/4563

See https://docs.google.com/document/d/1RN6xFxtnBFZQaIuJ3rBXFmxxOjDM3ftZcrJn5Pv2shw/edit

cmelchior:
Fixes #4296 

This is a feature branch. The work will be split into smaller PR's that will be merged into this branch incrementally.

TODO:
- [x] Check changelog before final merge.
- [x] Javadoc
- [x] Timing issue for permission Realm not being ready
- [x] Non-public PermissionManager constructor
- [x] applyPermissions method ( #4731 )
- [x] getDefaultPermissions method ( #4850 )
- [x] Remove FIXME's
- [x] Deprecate Management Realm
- [x] Remove task before calling success (in case it crashes)

Optional
- [x] makeOffer method
- [x] acceptOffer method 
- [x] revokeOffer method
- [x] getOffers method


CHANGELOG.md
```diff
@@ -82,6 +82,7 @@
 ### Enhancements
 
 * [ObjectServer] Added support for `SyncUser.isAdmin()` (#4353).
+* [ObjectServer] New set of Permission API's have been added to `SyncUser` through `SyncUser.getPermissionManager()` (#4296).
```
kneth: API's APIs

realm/realm-library/src/objectServer/java/io/realm/PermissionManager.java
```diff
@@ -1092,6 +1103,88 @@ private void notifyCallbackWithSuccess(RealmResults<PermissionOffer> permissions
...
+        }
+
+        @Override
+        public void run() {
+            if (checkAndReportInvalidState()) {
+                return;
+            }
+            matchingOffers = managementRealm.where(PermissionOffer.class)
+                    .equalTo("token", offerToken)
+                    .findAllAsync();
```
Zhuinden: what thread is this on that it has a looper but is called `AsyncTask`?

realm/realm-library/src/objectServer/java/io/realm/PermissionManager.java
```diff
@@ -376,10 +442,10 @@ private static String getRealmUrl(RealmType type, URL authUrl) {
             scheme = "realms";
         }
         try {
-            return new URI(scheme, authUrl.getUserInfo(), authUrl.getHost(), authUrl.getPort(),
-                    "/~/" + type.getName(), null, null).toString();
+            String path = (type.isGlobalRealm() ? "/" : "/~/") + type.getName();
```
bigfish24: Is this a safe assumption to be making?

bigfish24:
Want to highlight that we still have a fair amount of discrepancy with our APIs as seen here: https://docs.google.com/a/realm.io/spreadsheets/d/19I-AmlluZxnhaUxmS_it2oYlN0W622feM0fetESMRMY/edit?usp=sharing

Also, I would prefer we reevaluate whether we want to add the default Permissions Realm since I spoke with @mbalex99 about getting rid of this with 2.0.

cmelchior:
Note that the differences in names are hugely down to Java having everything inside a `PermissionManager` abstraction since that is required for our Realm lifecycle to work.

I would be more than happy to remove the `getDefaultPermissions`, but it is still a use case we need to support. I'm touching base with @mbalex99 to hear what the plans are.

PR #4559, opened 2017-04-26
Try to create getFilesDir if it doesn't exist on startup.

### LINKED ISSUE #4493: Cannot get Realm instance: io.realm.exceptions.RealmFileException: Unable to open a realm at path
#### Goal

Call `Realm realm = Realm.getInstance(mRealmConfig);` where

`
mRealmConfig = RealmConfiguration.Builder()
                .schemaVersion(2)
                .deleteRealmIfMigrationNeeded()
                .build();
`

#### Expected Results

Get a Realm without exception.

#### Actual Results

The following exception is thrown:
`io.realm.exceptions.RealmFileException: Unable to open a realm at path '/data/data/com.dropbox.paper/files/default.realm.management': make_dir() failed: No such file or directory. (make_dir() failed: No such file or directory) (/data/data/com.dropbox.paper/files/default.realm.management) in /home/cc/repo/realm/release/realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp line 217 Kind: ACCESS_ERROR.`

More here:
[realm-exception.txt](https://github.com/realm/realm-java/files/920336/realm-exception.txt)

This happens on first launch of the app.

Realm version(s): 2.2.1

Tried upgrading to 3.1.2 but still seeing the same issue.

Realm sync feature enabled: unsure. Would this affect this? How can I turn this on/off?

Android Studio version: 2.3

Which Android version and device:

- Nexus 5 on 6.0.1
- Google Pixel 7.1.1

cmelchior:
Closes #4493 

We cannot "fix" it, but this way we try as hard as possible to recover from the situation, and if that fails we throw a better exception, explaining the situation.

/cc @rajuashok

### DIFF
--- a/realm/realm-library/src/main/java/io/realm/Realm.java
+++ b/realm/realm-library/src/main/java/io/realm/Realm.java
@@ -245,7 +245,7 @@ public class Realm extends BaseRealm {
 
         // One final check before giving up
         if (context.getFilesDir() == null || !context.getFilesDir().exists()) {
-            throw new IllegalStateException("Context.getFilesDir() could not be found. See https://issuetracker.google.com/issues/36918154");
+            throw new IllegalStateException("Context.getFilesDir() returns " + context.getFilesDir() + " which is not an existing directory. See https://issuetracker.google.com/issues/36918154");
         }
     }
 
--- a/realm/realm-library/src/main/java/io/realm/Realm.java
+++ b/realm/realm-library/src/main/java/io/realm/Realm.java
@@ -211,13 +211,17 @@ public class Realm extends BaseRealm {
      */
     private static void checkUserDirectoryAvailable(Context context) {
         File userDir = context.getFilesDir();
-        if (userDir != null && !userDir.exists()) {
-            try {
-                // This was reported as working on some devices, which I really hope is just the race condition
-                // kicking in, otherwise something is seriously wrong with the permission system on those devices.
-                // We will try it anyway, since starting a loop will be slower by many magnitudes.
-                userDir.mkdirs();
-            } catch (SecurityException ignored) {
+        if (userDir != null) {
+            if (userDir.exists()) {
+                return; // Everything is fine. Escape as soon as possible
+            } else {
+                try {
+                    // This was reported as working on some devices, which I really hope is just the race condition
+                    // kicking in, otherwise something is seriously wrong with the permission system on those devices.
+                    // We will try it anyway, since starting a loop will be slower by many magnitudes.
+                    userDir.mkdirs();
+                } catch (SecurityException ignored) {
+                }
             }
         }
         if (userDir == null || !userDir.exists()) {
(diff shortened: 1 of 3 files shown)


realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -195,6 +198,58 @@ public static synchronized void init(Context context) {
...
+                currentTotalWaitMs += waitMs;
+                if (currentTotalWaitMs > maxTotalWaitMs) {
+                    break;
+                }
+            }
+        }
+
+        // One final check before giving up
+        if (context.getFilesDir() == null || !context.getFilesDir().exists()) {
+            throw new IllegalStateException("Context.getFilesDir() could not be found. See https://issuetracker.google.com/issues/36918154");
```
zaki50: do we need to force to crash the app?

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -195,6 +198,58 @@ public static synchronized void init(Context context) {
...
+     * <p>
+     * This method attempts to fix that situation. If this doesn't work an {@link IllegalStateException} will be
+     * thrown.
+     * <p>
+     * See these links for further details:
+     * https://issuetracker.google.com/issues/36918154
+     * https://github.com/realm/realm-java/issues/4493#issuecomment-295349044
+     */
+    private static void checkUserDirectoryAvailable(Context context) {
+        File userDir = context.getFilesDir();
```
zaki50: Why `userDir`?

cmelchior: What else would you call it?

realm/realm-library/src/main/java/io/realm/Realm.java
cmelchior: Well, now users can catch the IllegalStateException if they think they can recover better than we do. If we don't throw this exception here, it becomes a lot more random how the error surfaces (NullPointer / Core Exception / Somewhere else).

realm/realm-library/src/main/java/io/realm/Realm.java
zaki50: How about `filesDir` or `defaultDir`?

cmelchior: `filesDir` is fine with me

realm/realm-library/src/main/java/io/realm/Realm.java
zaki50: If we expect that a user handles the exception in this senario, we should throw specific exception. IllegalState is too generic.

cmelchior: I don't mind throwing some more specific exception, but what should it be? `IllegalState` feels pretty accurate here to me? It should definitely not be a checked exception though since that would A) break the API and B) In most users cannot do anything about it anyway

cmelchior: It shouldn't be an `Error` either since, you _might_ be able to recover from it.

zaki50: Hmm, `IllegalStateException` is fine then. I think most developers don't care this scenario.

beeender: Change the exception to something more specific to get more information when that happens. Something like:
```java
throw new IllegalStateException("Context.getFilesDir() returns " + Context.getFilesDir() + " which is not an existing dir. See https://issuetracker.google.com/issues/36918154");
```

PR #4563, opened 2017-04-26
Fix threading bugs in RunInLooperThread rule

bmeike:
During the process of fixing the last couple bugs in the backlinks UTs, I discovered that the `RunInLooperThread` `Rule` was both difficult to read and, probably because of that, had fundamental threading bugs.  This PR fixes both of those problems.

### DIFF
--- a/realm/realm-library/src/androidTest/java/io/realm/TestHelper.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/TestHelper.java
@@ -779,7 +779,7 @@ public class TestHelper {
     }
 
     public static void awaitOrFail(CountDownLatch latch) {
-        awaitOrFail(latch, 60);
+        awaitOrFail(latch, 300);
     }
 
     public static void awaitOrFail(CountDownLatch latch, int numberOfSeconds) {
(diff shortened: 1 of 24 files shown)


realm/realm-library/src/androidTest/java/io/realm/rule/RunInLooperThread.java
```diff
@@ -229,11 +265,177 @@ public void postRunnableDelayed(Runnable runnable, long delayMillis) {
...
+
+                TestHelper.exitOrThrow(executorService, signalTestCompleted, test);
+            } catch (Exception error) {
+                // These exceptions should only come from TestHelper.awaitOrFail()
+                testException = error;
+            } finally {
+                // Tries as hard as possible to close down gracefully, while still keeping all exceptions intact.
+                cleanUp(testException);
+
+                // Only TestHelper.awaitOrFail() threw an exception
```
cmelchior: Any reason this block of code isn't inside `cleanup` it looks slightly weird to me right now?

bmeike:
Ok.  All tests are passing, all comments addressed.  I would like to take @cmelchior 's suggestion and commit this mess, with the phrase in the field scanner that forbids backlink fields in queries.  I will follow this up with a smaller, more reasonable commit, with about 12 unit tests and whatever (presumably small) changes they require.

realm/realm-library/src/androidTest/java/io/realm/rule/RunInLooperThread.java
```diff
@@ -40,186 +40,222 @@
...
+    public void addTestRealm(Realm realm) {
+        synchronized (lock) {
+            testRealms.add(realm);
+        }
+    }
+
+    /**
+     * Explicitly close all held realms.
+     * <p>
+     * 'testRealms' is ccessed from both test and main threads.
```
bmeike: fixed.

realm/realm-library/src/androidTest/java/io/realm/rule/RunInLooperThread.java
bmeike: Yeh.  That and all those noinspections.  Fixed.

bmeike:
All comments addressed.

realm/realm-library/src/androidTest/java/io/realm/TestHelper.java
```diff
@@ -779,7 +779,7 @@ public static void populateForDistinctFieldsOrder(Realm realm, long numberOfBloc
     }
 
     public static void awaitOrFail(CountDownLatch latch) {
-        awaitOrFail(latch, 60);
+        awaitOrFail(latch, 300);
```
beeender: isn't 300 too long? 60 s is long enough for any test to pass, otherwise there is a bug in the test.

bmeike: Doggone it!  Yes.  That's a mistake.  I sometimes need that for my local tests, but it shouldn't be standard.  WIll fix.

PR #4570, opened 2017-04-27
Implement getInstanceAsync

### LINKED ISSUE #2299: Realm.getInstanceAsync()

beeender:
Fix #2299
- Add APIs to get Realm instance asynchronously.
- Remove some useless methods.

There are some work need to be done before create the first Realm instance
in the process, like creating schema table, doing migration, etc.. Those
could block the UI thread quite badly. This commit tries to do those
initialization work in the background and hold a Realm instance in the
background until the 2nd instance created in the caller thread.

A better solution than this would be do initialization in the
background and only deliver a column indices cache to caller thread
without holding a Realm instance in the background. But that is not
possible since from the current database design, we cannot know if the
schema changes compared with the last time it was opened.

Also create a SharedGroup in the background and handover it to the
caller thread is not ideal as well. That not only requires some design
changes in the Object Store RealmCoordinator, but also is a very
special use case of SharedGroup which core is not designed for.
SharedGroup Leaking during the handover is another flaw for this
solution -- we can only rely on the GC to collect the leaked SharedGroup
during handover then.

- [x] Check and fix javadoc for `Realm.Callback` `DynamicRealm.Callback`.

### DIFF
--- a/realm/realm-library/src/androidTest/java/io/realm/RealmCacheTests.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/RealmCacheTests.java
@@ -28,22 +28,29 @@ import org.junit.runner.RunWith;
 import java.io.File;
 import java.io.IOException;
 import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicReference;
 
 import io.realm.entities.AllTypes;
 import io.realm.entities.StringOnly;
 import io.realm.exceptions.RealmFileException;
+import io.realm.rule.RunInLooperThread;
+import io.realm.rule.RunTestInLooperThread;
 import io.realm.rule.TestRealmConfigurationFactory;
 
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertSame;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 
 @RunWith(AndroidJUnit4.class)
 public class RealmCacheTests {
 
+    @Rule
+    public final RunInLooperThread looperThread = new RunInLooperThread();
     @Rule
     public final TestRealmConfigurationFactory configFactory = new TestRealmConfigurationFactory();
 
@@ -328,6 +335,184 @@ public class RealmCacheTests {
         assertNull(realmA.sharedRealm);
         dynamicRealmA.close();
         assertNull(realmA.sharedRealm);
+    }
+
+    @Test
+    @RunTestInLooperThread
+    public void getInstanceAsync_typedRealm() {
+        final RealmConfiguration configuration = looperThread.createConfiguration();
+        final AtomicBoolean realmCreated = new AtomicBoolean(false);
+        Realm.getInstanceAsync(configuration, new Realm.Callback() {
+            @Override
+            public void onSuccess(Realm realm) {
+                realmCreated.set(true);
+                assertEquals(1, Realm.getLocalInstanceCount(configuration));
+                realm.close();
+                looperThread.testComplete();
+            }
+        });
+        assertFalse(realmCreated.get());
+    }
+
+    @Test
+    @RunTestInLooperThread
+    public void getInstanceAsync_dynamicRealm() {
+        final RealmConfiguration configuration = looperThread.createConfiguration();
+        final AtomicBoolean realmCreated = new AtomicBoolean(false);
+        DynamicRealm.getInstanceAsync(configuration, new DynamicRealm.Callback() {
+            @Override
+            public void onSuccess(DynamicRealm realm) {
+                realmCreated.set(true);
+                assertEquals(1, Realm.getLocalInstanceCount(configuration));
+                realm.close();
+                looperThread.testComplete();
+            }
+        });
+        assertFalse(realmCreated.get());
+    }
+
+    @Test
+    @RunTestInLooperThread
+    public void getInstanceAsync_callbackDeliveredInFollowingEventLoopWhenLocalCacheExist() {
+        final RealmConfiguration configuration = looperThread.createConfiguration();
+        final AtomicBoolean realmCreated = new AtomicBoolean(false);
+        final Realm localRealm = Realm.getInstance(configuration);
+        Realm.getInstanceAsync(configuration, new Realm.Callback() {
+            @Override
+            public void onSuccess(Realm realm) {
+                realmCreated.set(true);
+                assertEquals(2, Realm.getLocalInstanceCount(configuration));
+                assertSame(realm, localRealm);
+                realm.close();
+                localRealm.close();
+                looperThread.testComplete();
+            }
+        });
+        assertFalse(realmCreated.get());
+    }
+
+    @Test
+    @RunTestInLooperThread
+    public void getInstanceAsync_callbackDeliveredInFollowingEventLoopWhenGlobalCacheExist() throws InterruptedException {
+        final RealmConfiguration configuration = looperThread.createConfiguration();
+        final AtomicBoolean realmCreated = new AtomicBoolean(false);
+        final CountDownLatch globalRealmCreated = new CountDownLatch(1);
+        final CountDownLatch getAsyncFinishedLatch = new CountDownLatch(1);
+
+        final Thread thread = new Thread(new Runnable() {
+            @Override
+            public void run() {
+                Realm realm = Realm.getInstance(configuration);
+                globalRealmCreated.countDown();
+                TestHelper.awaitOrFail(getAsyncFinishedLatch);
+                realm.close();
+            }
+        });
+        thread.start();
+
+        TestHelper.awaitOrFail(globalRealmCreated);
+        Realm.getInstanceAsync(configuration, new Realm.Callback() {
+            @Override
+            public void onSuccess(Realm realm) {
+                realmCreated.set(true);
+                assertEquals(1, Realm.getLocalInstanceCount(configuration));
+                realm.close();
+                getAsyncFinishedLatch.countDown();
+                try {
+                    thread.join();
+                } catch (InterruptedException e) {
+                    fail();
+                }
+                looperThread.testComplete();
+            }
+        });
+        assertFalse(realmCreated.get());
+    }
+
+    @Test
+    @RunTestInLooperThread
+    public void getInstanceAsync_typedRealmShouldStillBeInitializedInBGIfOnlyDynamicRealmExists() {
+        final RealmConfiguration configuration = looperThread.createConfiguration();
+        final DynamicRealm dynamicRealm = DynamicRealm.getInstance(configuration);
+        final AtomicBoolean realmCreated = new AtomicBoolean(false);
+
+        Realm.getInstanceAsync(configuration, new Realm.Callback() {
+            @Override
+            public void onSuccess(Realm realm) {
+                realmCreated.set(false);
+                assertEquals(2, Realm.getLocalInstanceCount(configuration));
+                dynamicRealm.close();
+                realm.close();
+                looperThread.testComplete();
+            }
+        });
+        // Callback should not be called immediately since we need to create column indices cache in bg thread.
+        // Only a local dynamic Realm instance existing at this time.
+        assertFalse(realmCreated.get());
+        assertEquals(1, Realm.getLocalInstanceCount(configuration));
+    }
+
+    @Test
+    @RunTestInLooperThread
+    public void getInstanceAsync_onError() {
+        final RealmConfiguration configuration =
+                looperThread.createConfigurationBuilder()
+                .assetFile("NotExistingFile")
+                .build();
+        Realm.getInstanceAsync(configuration, new Realm.Callback() {
+            @Override
+            public void onSuccess(Realm realm) {
+                fail();
+            }
+
+            @Override
+            public void onError(Throwable exception) {
+                assertTrue(exception instanceof RealmFileException);
+                looperThread.testComplete();
+            }
+        });
+    }
+
+    // If the async task is canceled before the posted event to create Realm instance in caller thread, the event should
+    // just be ignored.
+    @Test
+    @RunTestInLooperThread
+    public void getInstanceAsync_cancelBeforePostShouldNotCreateRealmInstanceOnTheCallerThread() {
+        final AtomicReference<RealmAsyncTask> realmAsyncTasks = new AtomicReference<>();
+        final Runnable finishedRunnable = new Runnable() {
+            @Override
+            public void run() {
+                looperThread.testComplete();
+            }
+        };
+        final RealmConfiguration configuration = looperThread.createConfigurationBuilder()
+                .name("will_be_canceled")
+                .initialData(new Realm.Transaction() {
+                    @Override
+                    public void execute(Realm realm) {
+                        // The BG thread started to initial the first Realm instance. Post an event to the caller's
+                        // queue to cancel the task before the event to create the Realm instance in caller thread.
+                        looperThread.postRunnable(new Runnable() {
+                            @Override
+                            public void run() {
+                                assertNotNull(realmAsyncTasks.get());
+                                realmAsyncTasks.get().cancel();
+                                // Wait the async task to be terminated.
+                                TestHelper.waitRealmThreadExecutorFinish();
+                                // Finish the test.
+                                looperThread.postRunnable(finishedRunnable);
+                            }
+                        });
+                    }
+                })
+                .build();
+
+        realmAsyncTasks.set(Realm.getInstanceAsync(configuration, new Realm.Callback() {
+            @Override
+            public void onSuccess(Realm realm) {
+                fail();
+            }
+        }));
     }
 
     // The DynamicRealm and Realm with the same Realm path should share the same RealmCache
--- a/realm/realm-library/src/androidTest/java/io/realm/rule/RunInLooperThread.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/rule/RunInLooperThread.java
@@ -239,6 +239,10 @@ public class RunInLooperThread extends TestRealmConfigurationFactory {
 
     @Override
     protected void after() {
+        // Wait for all async tasks to have completed to ensure a successful deleteRealm call.
+        // If it times out, it will throw.
+        TestHelper.waitRealmThreadExecutorFinish();
+
         super.after();
 
         // probably belt *and* suspenders...
--- a/realm/realm-library/src/main/java/io/realm/BaseRealm.java
+++ b/realm/realm-library/src/main/java/io/realm/BaseRealm.java
@@ -26,6 +26,7 @@ import java.util.Collections;
 import java.util.List;
 import java.util.concurrent.atomic.AtomicBoolean;
 
+import io.realm.exceptions.RealmException;
 import io.realm.exceptions.RealmFileException;
 import io.realm.exceptions.RealmMigrationNeededException;
 import io.realm.internal.CheckedRow;
@@ -747,4 +748,86 @@ abstract class BaseRealm implements Closeable {
     }
 
     public static final ThreadLocalRealmObjectContext objectContext = new ThreadLocalRealmObjectContext();
+
+    /**
+     * The Callback used when reporting back the result of loading a Realm asynchronously using either
+     * {@link Realm#getInstanceAsync(RealmConfiguration, Realm.Callback)} or
+     * {@link DynamicRealm#getInstanceAsync(RealmConfiguration, DynamicRealm.Callback)}.
+     * <p>
+     * Before creating the first Realm instance in a process, there are some initialization work that needs to be done
+     * such as creating or validating schemas, running migration if needed,
+     * copy asset file if {@link RealmConfiguration.Builder#assetFile(String)} is supplied and execute the
+     * {@link RealmConfiguration.Builder#initialData(Realm.Transaction)} if necessary. This work may take time
+     * and block the caller thread for a while. To avoid the {@code getInstance()} call blocking the main thread, the
+     * {@code getInstanceAsync()} can be used instead which will do the initialization work in the background thread and
+     * deliver a Realm instance to the caller thread.
+     * <p>
+     * In general, this method is most useful on the UI thread since that should be blocked as little as possible. On
+     * any other Looper threads or other threads that don't support callbacks, using the standard {@code getInstance()}
+     * should be fine.
+     * <p>
+     * Here is an example of using {@code getInstanceAsync()} when the app starts the first activity:
+     * <pre>
+     * public class MainActivity extends Activity {
+     *
+     *   private Realm realm = null;
+     *   private RealmAsyncTask realmAsyncTask;
+     *
+     *   \@Override
+     *   protected void onCreate(Bundle savedInstanceState) {
+     *     super.onCreate(savedInstanceState);
+     *     setContentView(R.layout.layout_main);
+     *     realmAsyncTask = Realm.getDefaultInstanceAsync(new Callback() {
+     *         \@Override
+     *         public void onSuccess(Realm realm) {
+     *             if (isDestroyed()) {
+     *                 // If the activity is destroyed, the Realm instance should be closed immediately to avoid leaks.
+     *                 // Or you can call realmAsyncTask.cancel() in onDestroy() to stop callback delivery.
+     *                 realm.close();
+     *             } else {
+     *                 MainActivity.this.realm = realm;
+     *                 // Remove the spinner and start the real UI.
+     *             }
+     *         }
+     *     });
+     *
+     *     // Show a spinner before Realm instance returned by the callback.
+     *   }
+     *
+     *   \@Override
+     *   protected void onDestroy() {
+     *     super.onDestroy();
+     *     if (realm != null) {
+     *         realm.close();
+     *         realm = null;
+     *     } else {
+     *         // Calling cancel() on the thread where getInstanceAsync was called on to stop the callback delivery.
+     *         // Otherwise you need to check if the activity is destroyed to close in the onSuccess() properly.
+     *         realmAsyncTask.cancel();
+     *     }
+     *   }
+     * }
+     * </pre>
+     *
+     * @param <T> {@link Realm} or {@link DynamicRealm}.
+     */
+    public abstract static class InstanceCallback<T extends BaseRealm> {
+
+        /**
+         * Deliver a Realm instance to the caller thread.
+         *
+         * @param realm the Realm instance for the caller thread.
+         */
+        public abstract void onSuccess(T realm);
+
+        /**
+         * Deliver the error happens when creating the Realm instance to the caller thread. The default implementation
+         * will throw the exception on the caller thread.
+         *
+         * @param exception happened while initializing Realm on a background thread.
+         */
+        public void onError(Throwable exception) {
+            throw new RealmException("Exception happens when initializing Realm in the background thread.", exception);
+        }
+    }
 }
--- a/realm/realm-library/src/main/java/io/realm/BaseRealm.java
+++ b/realm/realm-library/src/main/java/io/realm/BaseRealm.java
@@ -754,15 +754,15 @@ abstract class BaseRealm implements Closeable {
      * {@link Realm#getInstanceAsync(RealmConfiguration, Realm.Callback)} or
      * {@link DynamicRealm#getInstanceAsync(RealmConfiguration, DynamicRealm.Callback)}.
      * <p>
-     * Before creating the first Realm instance in a process, there are some initialization work that needs to be done
+     * Before creating the first Realm instance in a process, there are some initialization work that need to be done
      * such as creating or validating schemas, running migration if needed,
      * copy asset file if {@link RealmConfiguration.Builder#assetFile(String)} is supplied and execute the
      * {@link RealmConfiguration.Builder#initialData(Realm.Transaction)} if necessary. This work may take time
      * and block the caller thread for a while. To avoid the {@code getInstance()} call blocking the main thread, the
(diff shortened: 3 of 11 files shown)


realm/realm-library/src/main/java/io/realm/RealmInstanceCallback.java
```diff
@@ -0,0 +1,98 @@
...
+ *         // Needs to close the Realm instance in the onSuccess() callback if it doesn't.
+ *         realmAsyncTask.cancel();
+ *     }
+ *   }
+ * }
+ * </pre>
+ *
+ * @param <T> {@link Realm} or {@link DynamicRealm}.
+ */
+public abstract class RealmInstanceCallback<T extends BaseRealm> {
```
beeender: maybe https://github.com/realm/realm-java/issues/2299#issuecomment-297647926 is better? @realm/java

realm/realm-library/src/main/java/io/realm/internal/android/AndroidRealmNotifier.java
```diff
@@ -27,6 +28,7 @@ public AndroidRealmNotifier(SharedRealm sharedRealm, Capabilities capabilities)
 
     @Override
     public boolean post(Runnable runnable) {
+        RealmLog.error("RRRRR + " + handler);
```
Zhuinden: RRRRR

realm/realm-library/src/main/java/io/realm/RealmInstanceCallback.java
```diff
@@ -0,0 +1,98 @@
...
+     */
+    public abstract void onSuccess(T realm);
+
+    /**
+     * Deliver the error happens when creating the Realm instance to the caller thread. The default implementation will
+     * throw the exception on the caller thread.
+     *
+     * @param exception happens when creating the Realm instance.
+     */
+    public void onError(RuntimeException exception) {
```
Zhuinden: Why specifically `RuntimeException`?

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -75,6 +84,82 @@ static RealmCacheType valueOf(Class<? extends BaseRealm> clazz) {
...
+            } finally {
+                if (instance != null) {
+                    instance.close();
+                }
+            }
+        }
+    }
+
+    private static final String ASYNC_NOT_ALLOWED_MSG =
+            "The Realm instance cannot be got asynchronously on current thread.";
```
Zhuinden: Something is strange about `cannot be got`, also the error message doesn't specify *why*.

realm/realm-library/src/main/java/io/realm/RealmInstanceCallback.java
beeender: Other checked exception should be handled by async task. But the `Transaction.OnError()` is actually taking a `Throwable` ... So ... no idea ...

beeender:
@Zhuinden It can cancel it, but it doesn't guarantee it will be cancelled. But I think some small improvement can be made to guarantee if the `cancel()` is called on the same thread where the results will be returned on, and before the results returned, then the `onSuccess()` won't be called anymore.

Calling `cancel()` on a different thread will be impossible to guarantee that. 

I will update the PR.

realm/realm-library/src/androidTest/java/io/realm/RealmCacheTests.java
```diff
@@ -331,6 +335,181 @@ public void releaseCacheInOneThread() {
...
+                realm.close();
+                looperThread.testComplete();
+            }
+        });
+        assertFalse(realmCreated.get());
+    }
+
+    @Test
+    @RunTestInLooperThread
+    public void getInstanceAsync_callbackImmediatelyWhenLocalCacheExist() {
```
cmelchior: IMO I don't like this semantic. As discussed in https://github.com/realm/realm-java/issues/4270#issuecomment-285330690 

While technically it isn't wrong per the contract of async to do that, it can and probably will mess up people's mental model. 

I would much rather have all callbacks happen through the looper queue, which will allow all remaining setup code in e.g. `onCreate` to run before this is called.

realm/realm-library/src/androidTest/java/io/realm/rule/RunInLooperThread.java
```diff
@@ -76,6 +76,11 @@ protected void before() throws Throwable {
     @Override
     protected void after() {
         super.after();
+
+        // Waits all async tasks done to ensure successful deleteRealm call.
```
cmelchior: Wait for all async tasks to have completed to ensure a successful deleteRealm call.
This will throw when if it times out. <I'm not sure I understand that last sentance?>

realm/realm-library/src/main/java/io/realm/DynamicRealm.java
```diff
@@ -71,6 +71,26 @@ public static DynamicRealm getInstance(RealmConfiguration configuration) {
     }
 
     /**
+     * The creation of the first Realm instance for per {@link RealmConfiguration} in the process takes time to do
```
cmelchior: The creation of the first Realm instance per {@link RealmConfiguration} in a process can take some time as all initialization code need to run at that point (Setting up the Realm, validating schemas and creating initial data).

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -212,6 +215,21 @@ public static Realm getDefaultInstance() {
...
 
     /**
+     * Calls {@link #getInstanceAsync(RealmConfiguration, RealmInstanceCallback)} with default configuration.
+     *
+     * @param callback invoked to return the results.
+     * @return a {@link RealmAsyncTask} representing a cancellable task.
+     * @see RealmInstanceCallback for how to handle the results.
+     * @see #setDefaultConfiguration(RealmConfiguration) about default configurations.
+     */
+    public static RealmAsyncTask getDefaultInstanceAsync(RealmInstanceCallback<Realm> callback) {
```
cmelchior: Pr comment in issue, I would just remove this method

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -224,12 +242,32 @@ public static Realm getDefaultInstance() {
...
     public static Realm getInstance(RealmConfiguration configuration) {
         if (configuration == null) {
-            throw new IllegalArgumentException("A non-null RealmConfiguration must be provided");
+            throw new IllegalArgumentException(NULL_CONFIG_MSG);
         }
         return RealmCache.createRealmOrGetFromCache(configuration, Realm.class);
     }
 
     /**
+     * The creation of the first Realm instance for per {@link RealmConfiguration} in the process takes time to do
```
cmelchior: See above

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -75,6 +84,82 @@ static RealmCacheType valueOf(Class<? extends BaseRealm> clazz) {
         }
     }
 
+    private static class CreateRealmRunnable<T extends BaseRealm> implements Runnable {
+        private RealmConfiguration configuration;
+        private RealmInstanceCallback<T> callback;
+        private Class<T> realmClass;
+        private CountDownLatch fgRealmCreatedLatch = new CountDownLatch(1);
```
cmelchior: fg?

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -75,6 +84,82 @@ static RealmCacheType valueOf(Class<? extends BaseRealm> clazz) {
...
+                        // instance on the caller thread.
+                        // Thread.isInterrupted() cannot be used for checking here since CountDownLatch.await() will
+                        // will clear interrupted status.
+                        if (taskCanceled.get()) {
+                            return;
+                        }
+                        T instanceToReturn = null;
+                        try {
+                            instanceToReturn = createRealmOrGetFromCache(configuration, realmClass);
+                        }catch (RuntimeException e) {
```
cmelchior: whitespace

cmelchior: Catch `Throwable` instead, otherwise some errors will not propagate to the Looper thread?

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -75,6 +84,82 @@ static RealmCacheType valueOf(Class<? extends BaseRealm> clazz) {
...
+                        }
+                        if (instanceToReturn != null) {
+                            callback.onSuccess(instanceToReturn);
+                        }
+                    }
+                });
+                if (!results) {
+                    fgRealmCreatedLatch.countDown();
+                }
+                // There is a small chance that the posted runnable cannot be executed because of the thread terminated
```
cmelchior: But the thread cannot terminate while it is in `wait` state?

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -75,6 +84,82 @@ static RealmCacheType valueOf(Class<? extends BaseRealm> clazz) {
...
+                // before the runnable gets fetched from the event queue.
+                if (!fgRealmCreatedLatch.await(5, TimeUnit.SECONDS)) {
+                    RealmLog.warn("Timeout for creating Realm instance in foreground thread in `CreateRealmRunnable` ");
+                }
+            } catch (InterruptedException e) {
+                taskCanceled.set(true);
+                RealmLog.warn(e, "`CreateRealmRunnable` has been interrupted.");
+            } catch (final RuntimeException e) {
```
cmelchior: Throwable?

realm/realm-library/src/main/java/io/realm/RealmCache.java
cmelchior: What about "Realm instances cannot be loaded asynchronously on a non-looper thread." ?

realm/realm-library/src/main/java/io/realm/RealmInstanceCallback.java
cmelchior: Background threads should never swallow exceptions IMO. This should be `Throwable`

realm/realm-library/src/main/java/io/realm/RealmInstanceCallback.java
```diff
@@ -0,0 +1,99 @@
...
+ *     setContentView(R.layout.layout_main);
+ *     realmAsyncTask = Realm.getDefaultInstanceAsync(new RealmInstanceCallback<Realm>() {
+ *         \@Override
+ *         public void onSuccess(Realm realm) {
+ *             if (isDestroyed()) {
+ *                 // If the activity is destroyed, the Realm instance should be closed immediately to avoid leaks.
+ *                 // Or you can call realmAsyncTask.cancel() in onDestroy() to stop callback delivery.
+ *                 realm.close();
+ *             } else {
+ *                 MainActivity.this.realm = realm;
```
beeender: Will this cause misunderstanding for users? remember you don't have to call `close()` for the `Transaction.onSuccess(Realm realm)`. But here, the realm instance won't be closed automatically when the callback returns. @realm/java

realm/realm-library/src/main/java/io/realm/RealmInstanceCallback.java
cmelchior: I think I would prefer this being an inner class in `Realm/DynamicRealm` instead. This name is really long and fugly.

An inner class would enable us to do this:

```
Realm.getInstanceAsync(config, new Realm.Callback() {
  @Override
  public void onSuccess(Realm realm) {
  }
});
```

Also we are fairly inconsistent with our callbacks. In some places we split `OnSuccess/OnError` into two interfaces (transactions), in other we combine them in one interface (network requests).

I would be tempted to say that for loading instances, people don't want to handle errors which for the sake of consistency would indicate two interfaces `Realm.Callback.OnSuccess` and `Realm.Callback.OnError`, this would also mean we need more `getInstanceAsync()` variants like we do for transactions.

realm/realm-library/src/main/java/io/realm/RealmCache.java
Zhuinden: most likely `foreground`

realm/realm-library/src/androidTest/java/io/realm/rule/RunInLooperThread.java
Zhuinden: If it times out, it will throw?

realm/realm-library/src/main/java/io/realm/RealmCache.java
Zhuinden: `foreground` I think

realm/realm-library/src/main/java/io/realm/RealmCache.java
beeender: It can, if other thread call `thatThread.interrupt()` or `asyncTask.cancel()`
This is majorly to protect below case:
```
Event QUEUE
1 some other event
2 getInstance in foreground
```

If 1 some other event clean the event queue and quits the looper, the 2 will never be executed then the async task hangs forever.

realm/realm-library/src/main/java/io/realm/RealmCache.java
beeender: "non-looper" is not generic for other java platforms, but not a issue for now, i will use that.

realm/realm-library/src/androidTest/java/io/realm/RealmCacheTests.java
beeender: OK, that even makes the code simpler

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -124,22 +124,22 @@ public void run() {
...
                     }
                 });
                 if (!results) {
-                    fgRealmCreatedLatch.countDown();
+                    createdInForegroundLatch.countDown();
                 }
                 // There is a small chance that the posted runnable cannot be executed because of the thread terminated
                 // before the runnable gets fetched from the event queue.
-                if (!fgRealmCreatedLatch.await(5, TimeUnit.SECONDS)) {
+                if (!createdInForegroundLatch.await(5, TimeUnit.SECONDS)) {
```
cmelchior: 5 seconds will cause an ANR that crashes the app. If we intend to let the app recover from this the value should be less, 1-2 seconds?

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -75,6 +83,88 @@ static RealmCacheType valueOf(Class<? extends BaseRealm> clazz) {
...
+                            callback.onSuccess(instanceToReturn);
+                        }
+                    }
+                });
+                if (!results) {
+                    createdInForegroundLatch.countDown();
+                }
+                // There is a small chance that the posted runnable cannot be executed because of the thread terminated
+                // before the runnable gets fetched from the event queue.
+                if (!createdInForegroundLatch.await(5, TimeUnit.SECONDS)) {
```
cmelchior: 5 seconds will cause ANR, that crashes the app, so if we intend users to recover from this, the value must be less. Perhaps 1-2 seconds?

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -75,6 +83,88 @@ static RealmCacheType valueOf(Class<? extends BaseRealm> clazz) {
...
+                    RealmLog.warn("Timeout for creating Realm instance in foreground thread in `CreateRealmRunnable` ");
+                }
+            } catch (InterruptedException e) {
+                RealmLog.warn(e, "`CreateRealmRunnable` has been interrupted.");
+            } catch (final RuntimeException e) {
```
cmelchior: The only errors that could end up seems to be really critical ones. Shouldn't this be `Throwable` instead? Also, why not post the `InterruptedException` here as well?

realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -744,4 +745,86 @@ protected RealmObjectContext initialValue() {
...
+     *         // Otherwise you need to check if the activity is destroyed to close in the onSuccess() properly.
+     *         realmAsyncTask.cancel();
+     *     }
+     *   }
+     * }
+     * </pre>
+     *
+     * @param <T> {@link Realm} or {@link DynamicRealm}.
+     */
+    public abstract static class InstanceCallback<T extends BaseRealm> {
```
Zhuinden: If the user needs to create this, then it ought to be public

cmelchior: It is the later: https://github.com/realm/realm-java/pull/4570/files#diff-241792b8b6ea216e6edbac593a7861b1R267

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -126,6 +126,9 @@
  */
 public class Realm extends BaseRealm {
 
+    private static final String NULL_CONFIG_MSG = "A non-null RealmConfiguration must be provided";
+
+
```
zaki50: too many empty lines?

realm/realm-library/src/main/java/io/realm/BaseRealm.java
beeender: the problem is if it is `protected`, then the `@inheritdoc` doesn't work. With current setup, the `onError` and `onSuccess` javadoc cannot be generated, I haven't found a way to solve that yet.

cmelchior: Just copy the Javadoc and move on?

realm/realm-library/src/main/java/io/realm/RealmCache.java
beeender: `InterruptedException` will happen if user calls `RealmAsyncTask.cancel()` while the latch is await. Since it is an intended behaviour and the default `onError` will rethrow, so `InterruptedException` should not be thrown.

Will catch `Throwable` here.

realm/realm-library/src/main/java/io/realm/BaseRealm.java
cmelchior: Having this public puts it in the public API.. that should be fixed somehow IMO.

beeender: `BaseRealm` is packaged protected, so even this is public, it is not part of public API. (same for other methods in BaseRealm like `BaseRealm.isInTransaction()`)

cmelchior: Good point. Ignore me 👍

realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -747,4 +748,86 @@ protected RealmObjectContext initialValue() {
     }
 
     public static final ThreadLocalRealmObjectContext objectContext = new ThreadLocalRealmObjectContext();
+
+    /**
+     * The Callback used when reporting back the result of loading a Realm asynchronously using either
+     * {@link Realm#getInstanceAsync(RealmConfiguration, Realm.Callback)} or
+     * {@link DynamicRealm#getInstanceAsync(RealmConfiguration, DynamicRealm.Callback)}.
+     * <p>
+     * Before creating the first Realm instance in a process, there are some initialization work that needs to be done
```
kneth: Should "needs" be "need" (refers back to "some initialization work" - plural?)

realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -747,4 +748,86 @@ protected RealmObjectContext initialValue() {
...
+     * such as creating or validating schemas, running migration if needed,
+     * copy asset file if {@link RealmConfiguration.Builder#assetFile(String)} is supplied and execute the
+     * {@link RealmConfiguration.Builder#initialData(Realm.Transaction)} if necessary. This work may take time
+     * and block the caller thread for a while. To avoid the {@code getInstance()} call blocking the main thread, the
+     * {@code getInstanceAsync()} can be used instead which will do the initialization work in the background thread and
```
kneth: "which will" -> "to"

realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -747,4 +748,86 @@ protected RealmObjectContext initialValue() {
...
+     * deliver a Realm instance to the caller thread.
+     * <p>
+     * In general, this method is most useful on the UI thread since that should be blocked as little as possible. On
```
kneth: "most" -> "mostly"

realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -747,4 +748,86 @@ protected RealmObjectContext initialValue() {
...
+
+        /**
+         * Deliver a Realm instance to the caller thread.
+         *
+         * @param realm the Realm instance for the caller thread.
+         */
+        public abstract void onSuccess(T realm);
+
+        /**
+         * Deliver the error happens when creating the Realm instance to the caller thread. The default implementation
```
kneth: "the error" -> "an error"

realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -747,4 +748,86 @@ protected RealmObjectContext initialValue() {
...
+         * will throw the exception on the caller thread.
```
kneth: "the exception" -> "an exception"

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -279,12 +281,33 @@ public static Realm getDefaultInstance() {
...
         if (configuration == null) {
-            throw new IllegalArgumentException("A non-null RealmConfiguration must be provided");
+            throw new IllegalArgumentException(NULL_CONFIG_MSG);
         }
         return RealmCache.createRealmOrGetFromCache(configuration, Realm.class);
     }
 
     /**
+     * The creation of the first Realm instance per {@link RealmConfiguration} in a process can take some time as all
+     * initialization code need to run at that point (Setting up the Realm, validating schemas and creating initial
```
kneth: "Setting" -> "setting"

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -78,6 +86,88 @@ static RealmCacheType valueOf(Class<? extends BaseRealm> clazz) {
...
+                if (!results) {
+                    createdInForegroundLatch.countDown();
+                }
+                // There is a small chance that the posted runnable cannot be executed because of the thread terminated
+                // before the runnable gets fetched from the event queue.
+                if (!createdInForegroundLatch.await(2, TimeUnit.SECONDS)) {
+                    RealmLog.warn("Timeout for creating Realm instance in foreground thread in `CreateRealmRunnable` ");
+                }
+            } catch (InterruptedException e) {
+                RealmLog.warn(e, "`CreateRealmRunnable` has been interrupted.");
```
zaki50: Shouldn't we restore the interrupted state to current thread?

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -279,12 +281,33 @@ public static Realm getDefaultInstance() {
...
+     * data). This method places the initialization work in a background thread and deliver the Realm instance
+     * to the caller thread asynchronously after the initialization is finished.
+     *
+     * @param configuration {@link RealmConfiguration} used to open the Realm.
+     * @param callback invoked to return the results.
+     * @throws IllegalArgumentException if a null {@link RealmConfiguration} is provided.
```
zaki50: also `null` callback causes `IAE`

realm/realm-library/src/main/java/io/realm/DynamicRealm.java
```diff
@@ -71,6 +71,27 @@ public static DynamicRealm getInstance(RealmConfiguration configuration) {
...
 
     /**
+     * The creation of the first Realm instance per {@link RealmConfiguration} in a process can take some time as all
+     * initialization code need to run at that point (Setting up the Realm, validating schemas and creating initial
+     * data). This method places the initialization work in a background thread and deliver the Realm instance
+     * to the caller thread asynchronously after the initialization is finished.
+     *
+     * @param configuration {@link RealmConfiguration} used to open the Realm.
+     * @param callback invoked to return the results.
+     * @throws IllegalArgumentException if a null {@link RealmConfiguration} is provided.
```
zaki50: also `null` callback causes `IAE`

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -119,6 +209,30 @@ private RealmCache(String path) {
         }
     }
 
+    static <T extends BaseRealm> RealmAsyncTask createRealmOrGetFromCacheAsync(
```
zaki50: Moving these two methods after `getCache()` is better?

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -78,6 +86,88 @@ static RealmCacheType valueOf(Class<? extends BaseRealm> clazz) {
...
+                    @Override
+                    public void run() {
+                        // If the RealmAsyncTask.cancel() is called before, we just return without creating the Realm
+                        // instance on the caller thread.
+                        // Thread.isInterrupted() cannot be used for checking here since CountDownLatch.await() will
+                        // will clear interrupted status.
+                        // Using the future to check which this runnable belongs to is to ensure if it is canceled from
+                        // the caller thread before, the callback will never be delivered.
+                        if (future == null || future.isCancelled()) {
+                            return;
```
zaki50: No need to call `createdInForegroundLatch.countDown();`?

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -78,6 +86,88 @@ static RealmCacheType valueOf(Class<? extends BaseRealm> clazz) {
...
+                        }
+                        T instanceToReturn = null;
+                        try {
+                            instanceToReturn = createRealmOrGetFromCache(configuration, realmClass);
+                        } catch (Throwable e) {
+                            callback.onError(e);
+                        } finally {
+                            createdInForegroundLatch.countDown();
```
zaki50: Shouldn't we call `createdInForegroundLatch.countDown();` before `callback.onError(e);`?

```
T instanceToReturn = null;
Throwable error = null;
try {
    instanceToReturn = ...;
} catch (Throwable e) {
    error = e;
} finally {
    createdInForegroundLatch.countDown();
}

if (instanceToReturn != null) {
    callback.onSuccess(instanceToReturn);
} else {
    callback.onError(error);
}
```

realm/realm-library/src/androidTest/java/io/realm/RealmCacheTests.java
```diff
@@ -330,6 +337,184 @@ public void releaseCacheInOneThread() {
         assertNull(realmA.sharedRealm);
     }
 
+    @Test
+    @RunTestInLooperThread
+    public void getInstanceAsync_typedRealm() {
```
beeender: It is more like testing `RealmCache`, i think here is better.

realm/realm-library/src/main/java/io/realm/RealmCache.java
beeender: if the future is cancelled, then `CounterDownLatch.await()` will just throw. But yeah, not harmful to have `countDown()` here, i will add.

zaki50: If needed, `createdInForegroundLatch` should be `canReleaseBackgroundInstanceLatch`?

realm/realm-library/src/main/java/io/realm/RealmCache.java
beeender: yeah, this is better, will change.

realm/realm-library/src/main/java/io/realm/RealmCache.java
beeender: it will be reset by the `await()` 

     * <p>If the current thread:
     * <ul>
     * <li>has its interrupted status set on entry to this method; or
     * <li>is {@linkplain Thread#interrupt interrupted} while waiting,
     * </ul>
     * then {@link InterruptedException} is thrown and the current thread's
     * interrupted status is cleared.

zaki50: Yes, and I think we should restore it by calling `Thread.currentThread().interrupt();` to tell to the caller of this method.

PR #4574, opened 2017-04-28
Removed io.realm.internal.Util#javaPrint(String)

zaki50:
b0e188145c0d7762cd3edd9725d65e0be38cd614 already removed all calls to `io.realm.internal.Util#javaPrint` from JNI code.

This PR also removes `Util.getNativeMemUsage()`. See https://github.com/realm/realm-java/pull/4574#issuecomment-298296872


zaki50:
Also `Util.getNativeMemUsage()` is not called. Can we remove it? @realm/java

cmelchior:
It seems like a nice utility method to have around in case of debugging, so it should probably stay?

kneth:
@zaki50 On Android it is virtually impossible to calculate the amount of memory used. So yeah, go ahead and remove `Util.getNativeMemUsage()`.

PR #4575, opened 2017-04-28
Support readOnly() on Configurations

### LINKED ISSUE #1147: Readonly Realms
Cocoa support opening Realms in ReadOnly mode. We should do the same.

cmelchior:
Fixes #1147 

This PR adds support for read only Realms. This was needed by #4558.

The semantics differ slightly across local and synced Realms:

**Local Realms**
It should only work in conjunction with `assetFile()` since that is the only place it makes sense. Schemas will not be created, so the file must provide the correct schema.

**Synced Realms**
It should only work with `waitForInitialRemoteData(bool)`, since that would be the equivalent to the local `assetFile()`. If the downloaded file do not provide the correct schema, the Realm will fail to open and throw RealmMigrationNeededException.

Note the Object Store has support for read only Realms, but it require us using the Object Store schema, which we don't support (yet).

TODO
- [x] More unit tests
- [x] Doesn't make sense to merge this until #4536 has been merged, since people need to download the schema from the server.


cmelchior:
Ready for review @realm/java

realm/realm-library/src/syncIntegrationTest/java/io/realm/SyncedRealmTests.java
```diff
@@ -0,0 +1,274 @@
...
+    public void waitForInitialData_resilientInCaseOfRetriesAsync() {
+        SyncCredentials credentials = SyncCredentials.usernamePassword(UUID.randomUUID().toString(), "password", true);
+        SyncUser user = SyncUser.login(credentials, Constants.AUTH_URL);
+        final SyncConfiguration config = new SyncConfiguration.Builder(user, Constants.USER_REALM)
+                .waitForInitialRemoteData(false)
+                .build();
+        Random randomizer = new Random();
+
+        for (int i = 0; i < 10; i++) {
+            final int iteration = i;
```
bmeike: iteration???  Were you using this for an error message or something?

realm/realm-library/src/objectServer/java/io/realm/SyncConfiguration.java
```diff
@@ -92,6 +92,7 @@ private SyncConfiguration(File directory,
                                 RealmProxyMediator schemaMediator,
                                 RxObservableFactory rxFactory,
                                 Realm.Transaction initialDataTransaction,
+                                boolean readOnly,
```
Zhuinden: I wrote a builder based initialization and I find that having a N length constructor is pretty much unavoidable.

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -457,7 +467,9 @@ private static void initializeRealm(Realm realm) {
             if (commitChanges) {
                 realm.commitTransaction();
             } else {
-                realm.cancelTransaction();
+                if (realm.isInTransaction()) {
```
beeender: ->
```
} else if (realm.isInTransaction()) {
```

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -413,38 +414,47 @@ private static void initializeRealm(Realm realm) {
         // the Realm is initialized.
         boolean commitChanges = false;
         try {
-            realm.beginTransaction();
+            // We need to start a transaction no matter readOnly mode, because it acts as an interprocess lock.
+            // TODO: For proper inter-process support we also need to move e.g copying the asset file under an
+            // interprocess lock. This lock can obviously not be created by a Realm instance so we probably need
+            // to implement it in Object Store. When this happens, the `beginTransaction(true)` can be removed again.
+            realm.beginTransaction(true);
```
beeender: I feel it is not right to deal with the read-only in this case. If it is read-only, we should just open the Realm and validate the schema without involving any write transaction. A read lock will be held on the version when the realm is opened. Multi-processes is not something we can help anyway since, after Realm open in process A, process B could still start a write transaction and change the schema. The schema change handling for read-only will just be the same for read-only realm (well, we don't handle it well now.).

beeender: But the schema validation procedure might be changed for OS schema integration, not sure what is the best to do right now.

realm/realm-library/src/main/java/io/realm/RealmConfiguration.java
```diff
@@ -615,16 +633,36 @@ public Builder initialData(Realm.Transaction transaction) {
...
+         * <p>
+         * WARNING: This could potentially be a lengthy operation and should ideally be done on a background thread.
+         *
+         * @param assetFile path to the asset database file.
+         * @param readOnly if {@code true} no local changes can be made to the Realm once copied. All write
+         * transactions will fail with an {@link IllegalStateException}. Marking a Realm as read-only only applies to
+         * the Realm in this process. Other processes can still write to the Realm.
+         * @throws IllegalStateException if this is configured to clear its schema by calling {@link #deleteRealmIfMigrationNeeded()}.
+         */
+        public Builder assetFile(final String assetFile, boolean readOnly) {
```
beeender: I fell this API is redundant, one method for an abnormal use case which can be achieved by extra one line of code.

realm/realm-library/src/main/java/io/realm/RealmConfiguration.java
```diff
@@ -605,6 +620,9 @@ public Builder rxFactory(RxObservableFactory factory) {
          * @param transaction transaction to execute.
          */
         public Builder initialData(Realm.Transaction transaction) {
+            if (readOnly) {
```
beeender: Check and throw in the `Builder.build()` instead. Not worth to check here plus this doesn't deal with `initialData()` first then `readOnly()`.

realm/realm-library/src/main/java/io/realm/RealmConfiguration.java
```diff
@@ -237,6 +240,16 @@ public RxObservableFactory getRxFactory() {
...
+     * {@link IllegalStateException} if {@link Realm#beginTransaction()} is called on it.
+     *
+     * @return {@code true} if this Realm is read only, {@code false} if not.
+     */
+    public boolean isReadOnly() {
+        return readOnly;
+    }
+
     @Override
     public boolean equals(Object obj) {
```
beeender: So it is allowed that user to create a read-only realm in thread A, but writeable Realm in thread B? I don't think this matches the behaviour from Object Store -- we need to ensure the behaviour is the same after we migrate to the Object Store schema, otherwise there is going to be a breaking change.

realm/realm-library/src/main/java/io/realm/RealmConfiguration.java
```diff
@@ -237,6 +240,16 @@ public RxObservableFactory getRxFactory() {
         return rxObservableFactory;
     }
 
+    /**
+     * Returns whether this Realm is read only or not. Read-only Realms cannot be modified and will throw an
```
kneth: Is it read only or read-only?

realm/realm-library/src/main/java/io/realm/internal/SharedRealm.java
```diff
@@ -243,6 +243,15 @@ public static SharedRealm getInstance(RealmConfiguration config, SchemaVersionLi
     }
 
     public void beginTransaction() {
+        beginTransaction(false);
+    }
+
+    public void beginTransaction(boolean ignoreReadOnly) {
```
kneth: Can it be `private`?

realm/realm-library/src/main/java/io/realm/RealmConfiguration.java
cmelchior: The `final` modifier looks like a mistake. I'll remove it.
Not sure what you mean @beeender? Removing `assetFile(String)` would be a breaking change, which is why I added this one on top. We should probably deprecate the other one though?

realm/realm-library/src/objectServer/java/io/realm/SyncConfiguration.java
cmelchior: Yes, it is way too long, but it is only called from one place (inside the Builder). Only other alternative would be having a no-arg constructor and a bunch of setter methods, but that looks even worse?

realm/realm-library/src/objectServer/java/io/realm/SyncConfiguration.java
```diff
@@ -257,6 +259,7 @@ boolean isSyncConfiguration() {
         private SharedRealm.Durability durability = SharedRealm.Durability.FULL;
         private boolean deleteRealmOnLogout = false;
         private final Pattern pattern = Pattern.compile("^[A-Za-z0-9_\\-\\.]+$"); // for checking serverUrl
+        private boolean readOnly = false;
```
cmelchior: The pattern is final? I do agree that the mix of uninitialized and default values is rather ugly though.

realm/realm-library/src/objectServer/java/io/realm/SyncConfiguration.java
```diff
@@ -605,9 +611,17 @@ public Builder errorHandler(SyncSession.ErrorHandler errorHandler) {
          * <p>
          * This check is only enforced the first time a Realm is created. If you otherwise want to make sure a Realm
          * has the latest changes, use {@link SyncSession#downloadAllServerChanges()}.
+         *
+         * @param readOnly if {@code true} no local changes can be made to the Realm once downloaded. All write
+         * transactions will fail with an {@link IllegalStateException}. Marking a Realm as read-only only applies to
+         * the Realm in this process. Other processes or devices can still write to the Realm.
          */
-        public Builder waitForInitialRemoteData() {
+        public Builder waitForInitialRemoteData(boolean readOnly) {
```
cmelchior: Not entirely sure what you mean? This setting is only available on the SyncConfiguration Builder which only works for synced Realms?

realm/realm-library/src/syncIntegrationTest/java/io/realm/SyncedRealmTests.java
cmelchior: Yes, which was later removed. I'll fix this.

realm/realm-library/src/main/java/io/realm/RealmConfiguration.java
beeender: I mean, why don't just remove `assetFile(final String assetFile, boolean readOnly)` and keep `assetFile(final String assetFile)` and `readOnly()`

realm/realm-library/src/main/java/io/realm/Realm.java
cmelchior: I agree, this needs to be changed with the OS schema integration. Which was why I went this route as it seemed the approach with the lowest friction.

It is not true that we cannot prevent multi-process interactions here. This is the entire reason the setup/validate method is inside a write-transaction. That is cross-process safe. If another process then later changes the Schema, the SharedRealm in this process will recieve a schema change callback.

The downside of course is that you need to know _not_ to write anything inside this method.
Having an interprocess lock that doesn't depend on Realm would be safer, and also required for our `assetFile()` to work. I don't think Object Store has this kind of logic yet. Nor do I know how difficult it is to add. 

Java doesn't officially support interprocess yet though, so your point is valid. In read-only mode, just open the Realm and verify. Then we can tackle interprocess later? That would also address @bmeike issues around code complexity in this method.

realm/realm-library/src/main/java/io/realm/RealmConfiguration.java
cmelchior: No, the schema cache should detect different readOnly modes and throw.  It is a bug it wasn't added to `equals()`.

We can never detect this across processes though.

realm/realm-library/src/main/java/io/realm/Realm.java
beeender: See my comment in https://github.com/realm/realm-java/pull/4575#pullrequestreview-37205956 which is relevant. In Object Store, it seems that creating the schemas is allowed in read-only mode?

realm/realm-library/src/main/java/io/realm/RealmConfiguration.java
cmelchior: In e.g. `assetFile()` we do the check if `initialDataTransaction` is set and throw. I guess you can argue both ways. Throwing as soon as possible is nicer, but the Builder isn't technically "done" until you call `build()`.

For consistency, I would prefer to throw here for now as changing when `assetFile()` throws is technically a breaking change. I wouldn't be terribly opposed to changing it at the next major version bump though. Perhaps I should just create an issue for it?

realm/realm-library/src/syncIntegrationTest/java/io/realm/SyncedRealmTests.java
```diff
@@ -0,0 +1,274 @@
...
+        SyncUser user = SyncUser.login(credentials, Constants.AUTH_URL);
+        final SyncConfiguration configNew = configurationFactory.createSyncConfigurationBuilder(user, Constants.USER_REALM)
+                .waitForInitialRemoteData(true)
+                .schema(StringOnly.class)
+                .build();
+        assertFalse(configNew.realmExists());
+
+        Realm realm = null;
+        try {
+            // This will fail, because the server Realm is completely empty and the Client is not allowed to write the
```
kneth: What's the different between "empty" and "completely empty"?

realm/realm-library/src/syncIntegrationTest/java/io/realm/SyncedRealmTests.java
```diff
@@ -0,0 +1,274 @@
...
+            user.logout();
+        }
+    }
+
+    @Test
+    public void waitForInitialRemoteData_readOnlyFalse_upgradeSchema() {
+        SyncCredentials credentials = SyncCredentials.usernamePassword(UUID.randomUUID().toString(), "password", true);
+        SyncUser user = SyncUser.login(credentials, Constants.AUTH_URL);
+        final SyncConfiguration config = configurationFactory.createSyncConfigurationBuilder(user, Constants.USER_REALM)
+                .waitForInitialRemoteData(false) // Not readonly so Client should be allowed to write schema
```
kneth: "readonly" -> "read-only"

realm/realm-library/src/main/java/io/realm/RealmConfiguration.java
beeender: UH... It is my mistake, I thought there is a method `Builder.readOnly()`. OK, do we really want to restrict users from copying realm file by himself?

realm/realm-library/src/main/java/io/realm/RealmConfiguration.java
beeender: It is my misunderstanding. See comment in the `assertFile()`

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/SyncedRealmTests.java
```diff
@@ -63,7 +63,7 @@ public void waitForInitialRemoteData_mainThreadThrows() {
         final SyncUser user = loginUser();
 
         SyncConfiguration config = new SyncConfiguration.Builder(user, Constants.USER_REALM)
-                .waitForInitialRemoteData()
+                .waitForInitialRemoteData(false)
```
kneth: I must admit that `false` to me carries little information. Maybe a constant with `READ_ONLY` in the name could increase the readability.

cmelchior: Yes, this annoys me as well. Adding named constants is probably a good idea.

realm/realm-library/src/main/java/io/realm/RealmConfiguration.java
beeender: BTW, `readOnly()` doesn't work with `deleteIfMigrationNeeded` as well.

realm/realm-library/src/main/java/io/realm/RealmConfiguration.java
cmelchior: In the original design `readOnly()` was a separate builder method, but I don't see it being useful unless you have an assetFile? I would actually argue that setting `readOnly()` in all other cases is probably a programmer error.
It would be more future proof though if we decide to allow it for other use cases and it removes the ugliness that is a un-named boolean on `waitForInitialRemoteData(false)` which just reads wrong on many levels. 

I briefly considered allowing it for `initialTransaction()` as well, but that will definitely conflict with the current semantics from Object Store, but we might want to change that when we migrate?

I won't be opposed to add `Builder.readOnly()` as it massively increases readability, but at an increased chance of people using it wrongly. Thoughts? @realm/java

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -413,38 +414,47 @@ private static void initializeRealm(Realm realm) {
...
             long currentVersion = realm.getVersion();
             boolean unversioned = currentVersion == UNVERSIONED;
             commitChanges = unversioned;
 
             RealmConfiguration configuration = realm.getConfiguration();
 
-            if (unversioned) {
-                realm.setVersion(configuration.getSchemaVersion());
-            }
+            // Only allow creating the schema if not in read-only mode
```
zaki50: How about throwing some exception if `unversioned && configuration.isReadOnly()` is `true`?
I think it makes following code much simpler.

realm/realm-library/src/main/java/io/realm/internal/SharedRealm.java
cmelchior: Not, it lives in `io.realm.internal` and is called from `io.realm`

zaki50:
@cmelchior If a user calls `DynamicRealm.getInstance()` first with read-only `RealmConfiguration`, then he  calls `Realm.getInstance()`, Realm file in asset does not seem to be copied since `getTotalGlobalRefCount()` is not 0. 

https://github.com/realm/realm-java/blob/cm/readonly/realm/realm-library/src/main/java/io/realm/RealmCache.java#L292

Should we throw a exception if `DynamicRealm.getInstance()` is called with read-only when the Realm file does not exist?

zaki50:
Ah, no. Just ignore my comment above.
file is copied when `DynamicRealm.getInstance()` was called.

realm/realm-library/src/syncIntegrationTest/java/io/realm/SyncedRealmTests.java
```diff
@@ -0,0 +1,275 @@
...
+        SyncCredentials credentials = SyncCredentials.usernamePassword(UUID.randomUUID().toString(), "password", true);
+        SyncUser user = SyncUser.login(credentials, Constants.AUTH_URL);
+        SyncConfiguration config = new SyncConfiguration.Builder(user, Constants.USER_REALM)
+                .waitForInitialRemoteData()
+                .build();
+
+        Realm realm = null;
+        try {
+            realm = Realm.getInstance(config);
+            assertTrue(realm.isEmpty());
```
zaki50: Why empty? And what does this test?

I think there are no difference if we don't call `.waitForInitialRemoteData();`.

realm/realm-library/src/objectServer/java/io/realm/SyncConfiguration.java
bmeike: I am so sorry!  I totally missed the `private` on there.

realm/realm-library/src/androidTest/java/io/realm/RealmTests.java
```diff
@@ -3993,4 +3990,40 @@ public void refresh_insideTransactionThrows() {
...
+                .name("readonly.realm")
+                .schema(StringOnly.class)
+                .assetFile("readonly.realm")
+                .readOnly()
+                .build();
+        realm = Realm.getInstance(config);
+
+        try {
+            realm.beginTransaction();
+        } catch (IllegalStateException e) {
```
beeender: Don't forget this one

realm/realm-library/src/main/java/io/realm/Realm.java
beeender: This one is not addressed?

realm/realm-library/src/main/java/io/realm/RealmConfiguration.java
```diff
@@ -672,6 +702,20 @@ Builder schema(Class<? extends RealmModel> firstClass, Class<? extends RealmMode
...
+            // an assetFile is configured
+            if (readOnly) {
+                if (initialDataTransaction != null) {
+                    throw new IllegalStateException("This Realm is marked as read-only. Read-only Realms cannot use initialData(Realm.Transaction).");
+                }
+                if (assetFilePath == null) {
+                    throw new IllegalStateException("Only Realms provided using 'assetFile(path)' can be marked read-only. No such Realm was provided.");
+                }
+                if (deleteRealmIfMigrationNeeded) {
+                    throw new IllegalStateException("'deleteRealmIfMigrationNeeded()' and read-only Realms cannot be combined");
```
beeender: read-only has to be used together with `assertFilePath`, so if a user has an old version asset copied, what should be done? delete the old one and copy the new one?

zaki50: Exactly.
Suppose an app contains an master data as an read-only Realm database.
If the new version of the app need to use a master with updated schema,
I think the app need to delete the old one and then copy the new Realm file 
from assets.

`deleteIfMigrationNeeded()` helps those scenario much.

cmelchior: Yeah, I can see the value in that. I'll change the semantics so they can be combined

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -413,42 +414,45 @@ private static void initializeRealm(Realm realm) {
...
 
-            if (unversioned) {
-                final Transaction transaction = configuration.getInitialDataTransaction();
-                if (transaction != null) {
-                    transaction.execute(realm);
-                }
+            // Finally add any initial data
+            final Transaction transaction = configuration.getInitialDataTransaction();
+            if (transaction != null) {
+                transaction.execute(realm);
```
zaki50: We can execute `transaction.execute(realm);` only when `unversioned` is true.

cmelchior: Yes, I just had another unit test complaining about this 😄

realm/realm-library/src/main/java/io/realm/RealmConfiguration.java
cmelchior: @zaki50 This change will be a bit bigger than I would like since our current code path only assume you can copy the file if it doesn't exist already. I would prefer creating a separate issue for this we can track?

zaki50: Sounds fine.
Allowing `deleteIfMagrationNeeded()` later breaks nothing.

cmelchior:
All comments addressed. I have created issues for the subtasks we discussed. But current semantics should be fine for a v1 for this. Merging.

PR #4576, opened 2017-04-29
Linking Objects unit tests.

bmeike:
Tests for Linking Objects.


realm/realm-library/src/androidTest/java/io/realm/LinkingObjectsQueryTests.java
```diff
@@ -0,0 +1,469 @@
...
+        AllJavaTypes gen3 = realm.createObject(AllJavaTypes.class, 3);
+        RealmList<AllJavaTypes> parents = gen3.getFieldList();
+        parents.add(gen2A);
+        parents.add(gen2B);
+
+        realm.commitTransaction();
+
+        RealmResults<AllJavaTypes> result = realm.where(AllJavaTypes.class)
+                .distinct(AllJavaTypes.FIELD_LO_OBJECT);
+        assertEquals(1, result.size());
```
cmelchior: I'm not sure I understand this test. What would the semantics be of calling `distinct()` on a backlink field? If it was even possible I would expect it to return all objects with unique backlinks? In this case it would be:
- gen1: Since it has gen2A/gen2B that reference it.
- gen2A/gen2B/gen3 has no-one referencing them through FIELD_LO_OBJECT, so it would be random which of these should be picked alongside gen1

realm/realm-library/src/androidTest/java/io/realm/LinkingObjectsQueryTests.java
```diff
@@ -0,0 +1,469 @@
...
+    //             gen1
+    //             / \
+    //         gen2A gen2B
+    //           \\   //
+    //            gen3
+    //  /  = object ref
+    //  // = list ref
+    @Test
+    public void query_startWithBacklink() {
+        realm.beginTransaction();
```
cmelchior: Extract this tree generation into a method since it is being used in multiple tests?

realm/realm-library/src/androidTest/java/io/realm/LinkingObjectsQueryTests.java
```diff
@@ -0,0 +1,469 @@
...
+
+        AllJavaTypes gen3 = realm.createObject(AllJavaTypes.class, 3);
+        RealmList<AllJavaTypes> parents = gen3.getFieldList();
+        parents.add(gen2A);
+        parents.add(gen2B);
+
+        realm.commitTransaction();
+
+        RealmResults<AllJavaTypes> result = realm.where(AllJavaTypes.class)
+                .greaterThan(AllJavaTypes.FIELD_LO_OBJECT + "." + AllJavaTypes.FIELD_OBJECT + "." + AllJavaTypes.FIELD_ID, 1)
```
cmelchior: Perhaps a comment that this filters out gen2B? I had to re-read this multiple times to figure that out 😄

realm/realm-library/src/androidTest/java/io/realm/LinkingObjectsQueryTests.java
```diff
@@ -0,0 +1,469 @@
...
+        realm.commitTransaction();
+
+        RealmResults<AllJavaTypes> result = realm.where(AllJavaTypes.class)
+                .lessThan(AllJavaTypes.FIELD_OBJECT + "." + AllJavaTypes.FIELD_LO_OBJECT + "." + AllJavaTypes.FIELD_ID, 2)
+                .findAll();
+        assertEquals(1, result.size());
+        assertTrue(result.contains(gen2A));
+    }
+
+    // FIXME!!! ???
```
cmelchior: What does this mean? The tests look good enough?

realm/realm-library/src/androidTest/java/io/realm/LinkingObjectsQueryTests.java
```diff
@@ -0,0 +1,469 @@
...
+                NullTypes.FIELD_LO_OBJECT + "." + NullTypes.FIELD_FLOAT_NULL).count());
+        // 9 Double
+        assertEquals(1, realm.where(NullTypes.class).isNull(
+                NullTypes.FIELD_LO_OBJECT + "." + NullTypes.FIELD_DOUBLE_NULL).count());
+        // 10 Date
+        assertEquals(1, realm.where(NullTypes.class).isNull(
+                NullTypes.FIELD_LO_OBJECT + "." + NullTypes.FIELD_DATE_NULL).count());
+    }
+
+    // FIXME!!! ???
```
cmelchior: Same

realm/realm-library/src/androidTest/java/io/realm/RealmQueryTests.java
```diff
@@ -2529,17 +2548,45 @@ public void isNull_linkField() {
...
+        } catch (IllegalArgumentException expected) {
+            assertEquals("Illegal Argument: isNull() by nested query for link field is not supported.", expected.getMessage());
+        }
+
+        // 13 LIST
+        try {
+            result = realm.where(NullTypes.class).isNull(NullTypes.FIELD_OBJECT_NULL + "." + NullTypes.FIELD_LIST_NULL);
+            fail("isNull should throw on nested linked fields (LIST)");
+        } catch (IllegalArgumentException expected) {
+            assertEquals("Illegal Argument: RealmList(fieldListNull) is not nullable.", expected.getMessage());
```
cmelchior: Any reason we throw in this case, it is trivially `false` since nothing will ever match this criteria, but do we really want to throw here. I could see some use cases where generic code might end constructing this kind of queries? Same in the others below

realm/realm-library/src/androidTest/java/io/realm/RealmQueryTests.java
```diff
@@ -2746,45 +2822,11 @@ public void isNotNull_linkFieldNotNullable() {
...
-        try {
-            realm.where(Cat.class).isNull("owner.dogs");
-            fail();
-        } catch (IllegalArgumentException expected) {
-            assertEquals("Illegal Argument: RealmList(dogs) is not nullable.", expected.getMessage());
-        }
-    }
-
-    // Calling isNotNull on fields with the RealmList type will trigger an exception.
+    @Ignore("Disabled because of time to run.")
```
cmelchior: I don't understand this comment?

realm/realm-library/src/main/cpp/io_realm_internal_TableQuery.cpp
```diff
@@ -93,6 +93,20 @@ static TableRef getTableByArray(jlong nativeQueryPtr, JniLongArray& indicesArray
     return table_ref;
 }
 
+//FIXME!!! this needs a clue.
```
cmelchior: Not sure what you mean?

realm/realm-library/src/main/java/io/realm/StandardRealmSchema.java
```diff
@@ -16,6 +16,8 @@
 
 package io.realm;
 
+import android.util.Log;
```
cmelchior: Not used

realm/realm-library/src/androidTest/java/io/realm/OrderedCollectionChangeSetTests.java
```diff
@@ -42,6 +40,7 @@
 import static junit.framework.Assert.fail;
 import static org.junit.Assert.assertArrayEquals;
 
+
```
beeender: Wouldn't those changes have been merged already?

realm/realm-library/src/androidTest/java/io/realm/QueryTests.java
```diff
@@ -0,0 +1,146 @@
+package io.realm;
+/*
```
beeender: license before the package

realm/realm-library/src/androidTest/java/io/realm/RealmQueryTests.java
beeender: `RealmList` is not nullable, i think it is current behaviour. This is just merging the test below into this one.

realm/realm-library/src/androidTest/java/io/realm/RealmQueryTests.java
beeender: This test was disabled even before.

realm/realm-library/src/main/java/io/realm/SchemaConnector.java
```diff
@@ -0,0 +1,55 @@
...
+ * methods on Schema, visible outside the io.realm package.
+ *
+ * The class is in the package, so it has access to package protected methods.
+ * The class is <b>not</b> outside the package.
+ * The class implements one or more interfaces visible to package-external clients, that need them.
+ *
+ * I suggest creating instances of this through a factory method in the service class.
+ * That will make it easy to lazily instantiate a singleton should that become advisable.
+ */
+class SchemaConnector implements FieldDescriptor.SchemaProxy {
```
cmelchior: I assume this was the kind of decoupling you where talking about at the architecture meeting?

bmeike: Yeah, exactly!

realm/realm-library/src/androidTest/java/io/realm/OrderedCollectionChangeSetTests.java
bmeike: I just did a diff.  They appear to have been merged already.

realm/realm-library/src/androidTest/java/io/realm/LinkingObjectsQueryTests.java
bmeike: Yeah... I thought about that.  The problem is that I need references to all of the objects that are added to the DB.

realm/realm-library/src/androidTest/java/io/realm/LinkingObjectsQueryTests.java
```diff
@@ -0,0 +1,572 @@
...
+            }
+            // 10 Date
+            nullTypes.setFieldDateNull(dates[i]);
+            if (dates[i] != null) {
+                nullTypes.setFieldDateNotNull(dates[i]);
+            }
+
+            nullTypesArray[i] = testRealm.copyToRealm(nullTypes);
+        }
+        nullTypesArray[0].setFieldObjectNull(nullTypesArray[0]);
```
kneth: What will `nullTypeArray[0].getFieldListNull()` return? To increase readability, I think all three times `fieldObjectNull` and `fieldListNull` should be set.

realm/realm-library/src/androidTest/java/io/realm/LinkingObjectsQueryTests.java
```diff
@@ -0,0 +1,572 @@
...
+                    fail("Unknown type: " + type);
+            }
+        }
+    }
+
+    // Similar to the version in TestHelper, but with more Backlinks
+    // Creates 3 NullTypes objects. The objects are self-referenced (link) in
+    // order to test link queries.
+    //
+    // +-+--------+------+---------+--------+--------------------+
```
kneth: Can you include `object` in the table too?

realm/realm-library/src/main/cpp/io_realm_internal_TableQuery.cpp
```diff
@@ -66,17 +66,18 @@ JNIEXPORT jstring JNICALL Java_io_realm_internal_TableQuery_nativeValidateQuery(
 // If the corresponding entry in tablesArray is anything other than a nullptr, the link is a backlink.
 // In that case, the tablesArray element is the pointer to the backlink source table and the
 // indicesArray entry is the source column index in the source table.
+// FIXME!!!  This doesn't actually seem to be following backlinks.
```
kneth: I think it's `getTableByArray()` which doesn't follow backlinks. See https://github.com/realm/realm-core/pull/2597

realm/realm-library/src/main/cpp/io_realm_internal_TableQuery.cpp
```diff
@@ -93,6 +94,21 @@ static TableRef getTableByArray(jlong nativeQueryPtr, JniLongArray& indicesArray
     return table_ref;
 }
 
+// FIXME!!!  This is a hasty attempt to fix the nullable queries.
+// I am not at all sure that it is even the right idea, let alone correct code. --gbm
+static bool isNullable(JNIEnv* env, Table* src_table_ptr, TableRef table_ref, jlong column_idx)
```
kneth: See my comment about `getTableByArray()`.

realm/realm-library/src/androidTest/java/io/realm/LinkingObjectsQueryTests.java
bmeike: Yeah, this could be improved.  I copied the existing tests.

realm/realm-library/src/androidTest/java/io/realm/LinkingObjectsQueryTests.java
bmeike: That makes sense. I grabbed the tests in TestHelper and extended them to test Backlinks. More testing would be even better.

realm/realm-library/src/main/java/io/realm/RealmResults.java
```diff
@@ -56,22 +58,24 @@
...
-        return new RealmResults<T>(
+        return new RealmResults<>(
                 realm,
-                Collection.createBacklinksCollection(realm.sharedRealm, row, srcTable, srcFieldName),
+                Collection.createBacklinksCollection(realm.sharedRealm, uncheckedRow, srcTable, srcFieldName),
                 srcTableType);
     }
 
-    static RealmResults<DynamicRealmObject> createBacklinkResults(DynamicRealm realm, CheckedRow row, Table srcTable, String srcFieldName) {
+    static RealmResults<DynamicRealmObject> createDynamicBacklinkResults(BaseRealm realm, Row row, Table srcTable, String srcFieldName) {
```
zaki50: If this method is only for dynamic, it only accepts `CheckedRow`.
This change should be reverted.

realm/realm-library/src/main/java/io/realm/RealmResults.java
```diff
@@ -56,22 +58,24 @@
  * @see Realm#executeTransaction(Realm.Transaction)
  */
 public class RealmResults<E extends RealmModel> extends OrderedRealmCollectionImpl<E> {
-    static <T extends RealmModel> RealmResults<T> createBacklinkResults(Realm realm, UncheckedRow row, Class<T> srcTableType, String srcFieldName) {
+    @SuppressLint("unused")
+    static <T extends RealmModel> RealmResults<T> createBacklinkResults(BaseRealm realm, Row row, Class<T> srcTableType, String srcFieldName) {
```
zaki50: Why `BaseRealm`?
This method is for typed Realm.
This change should be reverted.

bmeike: I strongly disagree.  There is no reason to require a `Realm` type here.

realm/realm-library/src/main/java/io/realm/RealmResults.java
bmeike: That is patently not true.  This compiles, runs, and all tests pass.  If there is a reason for using a `CheckRow` then it should be in the code.

bmeike: BTW, I didn't make this change just on a whim.  Choosing which version of an overloaded method to use, based purely on casts at the call, is incredibly fragile.  In fact, it broke as soon as I merged it with the backlinks branch.

realm/realm-library/src/main/java/io/realm/RealmResults.java
zaki50: I don't think that `DynamicRealm` with class object as a source type makes sense.

realm/realm-library/src/main/java/io/realm/RealmResults.java
zaki50: renaming the method is :+1:

cmelchior: The whole separation between DynamicRealm and Realm is pretty awful (and I wrote most of it) when it comes to our internal API's. I guess it boils down to two things: 1) DynamicRealmObject is a subclass of RealmObject which actually complicate things quite a lot 2) The difference between using Strings and Classes is permutating almost all code paths to the underlying code.

I wanted to revisit it for quite a while, but don't really have any good ideas on how to fix it.

realm/realm-library/src/main/java/io/realm/RealmResults.java
Zhuinden: @zaki50 but then what about https://github.com/realm/realm-java/blob/master/realm/realm-library/src/main/java/io/realm/OrderedRealmCollectionImpl.java#L31-L35

realm/realm-library/src/main/java/io/realm/DynamicRealmObject.java
```diff
@@ -967,7 +967,7 @@ public String toString() {
                     RealmFieldType.OBJECT.name(), RealmFieldType.LIST.name()));
         }
 
-        return RealmResults.createBacklinkResults(realm, (CheckedRow) proxyState.getRow$realm(), realmObjectSchema.getTable(), srcFieldName);
+        return RealmResults.createDynamicBacklinkResults(realm, proxyState.getRow$realm(), realmObjectSchema.getTable(), srcFieldName);
```
Zhuinden: Now that you mention it, separating the call based on a cast to `(UncheckedRow)` OR `(CheckedRow)` **is** super brittle.

realm/realm-library/src/main/java/io/realm/internal/fields/FieldDescriptor.java
```diff
@@ -156,20 +165,28 @@ public final int length() {
...
-     * After the field description (@see parseFieldDescription(String) is parsed, this method
-     * returns a java array.  For most columns the table will be the 'current' table, so this
-     * array will contain NativeObject.NULLPTR.  If a column is a LinkingObjects column, however,
-     * the array contains the native pointer to the <b>source</b> table.
+     * Return a java array of native table pointers.  For most columns the table will be identified by
+     * the type of the column: no further information is needed.  In that case, this array will contain
+     * NativeObject.NULLPTR.  If, however, a column is a LinkingObjects column the <b>source</b> table
+     * cannot be inferred, so the returned array contains the native pointer to it.
+     *
+     * The return is an array because it will be, immediately, passed to native code
```
Zhuinden: .

realm/realm-library/src/main/java/io/realm/internal/ColumnInfo.java
```diff
@@ -246,9 +245,9 @@ protected final long addColumnDetails(Table table, String columnName, RealmField
      */
     @SuppressWarnings("unused")
     protected final void addBacklinkDetails(SharedRealm realm, String columnName, String sourceTableName, String sourceColumnName) {
-        Table sourceTable = realm.getTable(Table.getTableNameForClass(sourceTableName));
-        long columnIndex = sourceTable.getColumnIndex(sourceColumnName);
-        indicesMap.put(columnName, new ColumnDetails(columnIndex, RealmFieldType.LINKING_OBJECTS, sourceTableName));
+//        Table sourceTable = realm.getTable(Table.getTableNameForClass(sourceTableName));
+//        long columnIndex = sourceTable.getColumnIndex(sourceColumnName);
+//        indicesMap.put(columnName, new ColumnDetails(columnIndex, RealmFieldType.LINKING_OBJECTS, sourceTableName));
```
Zhuinden: well, maybe some other time

realm/realm-library/src/main/java/io/realm/RealmResults.java
bmeike: Ok. I've reverted this.

I *hate* it.  In the call, we have to cast to get the parameter type right and then there is *nothing* inside the method that requires any of the more constrained types.  This is an abomination.  After looking at the code, though, I agree that it is better with the constrained params than it is without.

I think that the root problem is that there is no generic parameter on `DynamicObject`, so we lose all typing information as soon as we create one.

Out of scope.  Holding my nose.

realm/realm-library/src/main/java/io/realm/internal/ColumnInfo.java
bmeike: ;-P

realm/realm-library/src/androidTest/java/io/realm/LinkingObjectsQueryTests.java
bmeike: Will start an issue to track this.  It should be part of completing the backlinks work

realm/realm-library/src/main/java/io/realm/DynamicRealmObject.java
bmeike: Wildly, wildly, wildly.  Grosses me right out.

realm/realm-library/src/androidTest/java/io/realm/LinkingObjectsQueryTests.java
bmeike: Opened  #4620

realm/realm-library/src/main/cpp/io_realm_internal_TableQuery.cpp
kneth: Actually, we need a little more work on that - https://github.com/realm/realm-core/pull/2597 isn't the complete solution.

Issue #4561, opened 2017-04-26
Enable encryption with Sync


nhachicha:
We’re not using the encryption key for Synced Realm https://github.com/realm/realm-java/blob/fef417cb35a9920838a07f1ab93c8a84645b4d6f/realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp#L134

this is because the ObjectStore has a different `realm_encryption_key` https://github.com/realm/realm-object-store/blob/3eada170f380174992b47b6023f375f3f372a9d2/src/sync/sync_config.hpp#L108 parameter that we need to set (as opposed to the non sync encryption param `encryption_key` https://github.com/realm/realm-object-store/blob/563f2e9993ae4f4a8ab6f524a1ba71e707ebe78f/src/shared_realm.hpp#L145)

https://github.com/realm/realm-object-store/pull/414

nhachicha:
fixed inhttps://github.com/realm/realm-java/pull/4746

Issue #4371, opened 2017-03-23
Sync certificate verify failed only in android


probepark:
- realm-mobile-platform version :  1.2.1
- realm-java version : 3.0.0 
- realm-cocoa version : 2.4.4
- ream-browser(Mac APP) version : 2.1.5

This issue exists only in Android.

```
Connection[1]: SSL handshake failed: certificate verify failed
```

realm-cocoa / realm-browser has been confirmed to work..

istx25:
Hi @probepark. Would you be able to share a sample project that reproduces/demonstrates the SSL handshake failure? From there we'll be able to better assist you in getting the problem resolved.

cmelchior:
Hi @probepark It would especially be interesting know what URL you are trying to reach, but a sample project would be even better.

probepark:
I will make a sample.
However, it will take some time to set up ssl.
@istx25 @cmelchior

probepark:
https://github.com/probepark/realm-object-server-ssl-example

It took a long time. Sorry.
You can run this example.
Please confirm

@istx25 @cmelchior

probepark:
any updates?

nhachicha:
Hi,
 
I was able to see the `certificate verify failed` issue 

the error code `336134278` indicates an issue with OpenSSL validating the certificate 

I've seen similar issue with the CA "Let's Encrypt Authority X3" 
https://community.letsencrypt.org/t/certificate-not-trusted-on-android/29132

I created an issue internally, to track this.

probepark:
remind : 
https is working
realms isn't working.

I think that auth is using okhttp.
what is using realm url?

nhachicha:
Yes the auth is using OkHttp & it's working ( I can see the `access_token` obtained in the log)

Realm URL points to the **R**ealm **O**bject **S**erver, it's used by the sync client (C++), which uses OpenSSL for Android. 

Is it also possible to send the log of the server, when you start the Android App (you can get it from the ROS dashboard) or if you have access to the server (on Linux `/var/log/realm-object-server.log` on macOS they are also shown on stdout. )

probepark:
@nhachicha 
```
2017-03-29T15:29:56.174Z - debug: http:  127.0.0.1 - POST /auth HTTP/1.1 200 741 - 2382.266 ms
2017-03-29T15:29:56.175Z - detail: client: Using already open Realm file: /var/lib/realm/object-server/internal_data/auth.realm
[...]

PR #4591, opened 2017-05-02
Nh/android support ssl

nhachicha:
Fixing #4371 

This adds two options:

- `trustedRootCA` to provide the set of trusted root CA that Sync client should use

- `withoutSSLVerification` to disable OpenSSL certificate validation (for debug purpose or when using a self signed certificate in dev environment) 

TODO
- [x] changelog
- [x] tests


zaki50:
This seems new feature.
Shouldn't this target `master`?

zaki50:
Changed the target to `master`

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
```diff
@@ -130,8 +132,19 @@ class JniConfigWrapper {
             user = SyncManager::shared().get_user(user_identity, refresh_token,
                                                   realm::util::Optional<std::string>(realm_auth_url));
         }
-        m_config.sync_config = std::make_shared<SyncConfig>(SyncConfig{
-            user, realm_url, SyncSessionStopPolicy::Immediately, std::move(bind_handler), std::move(error_handler)});
+        if (sync_ssl_trust_certificate_path) {
```
cmelchior: Any reason for this `if`. Couldn't the last parameter just be made optional?

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -15,6 +15,8 @@
  */
 package io.realm;
 
+import android.text.TextUtils;
```
cmelchior: Android import, please don't use this

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -527,55 +529,71 @@ synchronized void invokeWithLock(Callback0 callback) {
...
 
-                outputStream = new FileOutputStream(realmFile);
-                byte[] buf = new byte[4096];
-                int bytesRead;
-                while ((bytesRead = inputStream.read(buf)) > -1) {
-                    outputStream.write(buf, 0, bytesRead);
-                }
-            } catch (IOException e) {
+        // Copy Sync Server certificate path if available
+
```
cmelchior: Unnecessary whitespace?

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -527,55 +529,71 @@ synchronized void invokeWithLock(Callback0 callback) {
...
+        String syncServerCertificateAssetName = ObjectServerFacade.getFacade(configuration.isSyncConfiguration()).getSyncServerCertificateAssetName(configuration);
+        if (!TextUtils.isEmpty(syncServerCertificateAssetName)) {
+            String syncServerCertificateFilePath = ObjectServerFacade.getFacade(configuration.isSyncConfiguration()).getSyncServerCertificateFilePath(configuration);
+
+            // using getRealmDirectory avoid file collision between same filename from different users (Realms)
```
cmelchior: This doesn't seem to match the code?

realm/realm-library/src/main/java/io/realm/RealmConfiguration.java
```diff
@@ -185,13 +184,12 @@ boolean hasAssetFile() {
     }
 
     /**
-     * Returns input stream object to the Realm asset file.
+     * Returns the path to the Realm asset file.
      *
-     * @return input stream to the asset file.
-     * @throws IOException if copying the file fails.
+     * @return absolute path to the asset file.
```
cmelchior: Is this true? Historically it is just the path for files in assets, and I believe that is a zipped file, so talking about "absolute path" doesn't feel correct?

realm/realm-library/src/main/java/io/realm/internal/SharedRealm.java
```diff
@@ -211,11 +211,13 @@ public static SharedRealm getInstance(RealmConfiguration config) {
...
-        String syncUserIdentifier = syncUserConf[0];
-        String syncRealmUrl = syncUserConf[1];
-        String syncRealmAuthUrl = syncUserConf[2];
-        String syncRefreshToken = syncUserConf[3];
+        Object[] syncUserConf = ObjectServerFacade.getSyncFacadeIfPossible().getUserAndServerUrl(config);
+        String syncUserIdentifier = (String) syncUserConf[0];
+        String syncRealmUrl = (String) syncUserConf[1];
+        String syncRealmAuthUrl = (String) syncUserConf[2];
+        String syncRefreshToken = (String) syncUserConf[3];
+        Boolean syncClientValidateSsl = (Boolean) syncUserConf[4];
```
cmelchior: why `Boolean` and not `boolean` ?

realm/realm-library/src/objectServer/java/io/realm/SyncConfiguration.java
```diff
@@ -17,6 +17,7 @@
 package io.realm;
 
 import android.content.Context;
+import android.text.TextUtils;
```
cmelchior: Please don't use this

realm/realm-library/src/objectServer/java/io/realm/SyncConfiguration.java
```diff
@@ -219,6 +234,37 @@ public boolean shouldDeleteRealmOnLogout() {
         return deleteRealmOnLogout;
     }
 
+    /**
+     * Returns the name of the certificate stored under the {@code assets}, to be used to validate
```
cmelchior: remove 'the'.

realm/realm-library/src/objectServer/java/io/realm/SyncConfiguration.java
```diff
@@ -596,6 +646,42 @@ public Builder errorHandler(SyncSession.ErrorHandler errorHandler) {
...
+        }
+
+        /**
+         * This will disable TLS certificate verification for the remote Realm Object Server.
+         * It is not recommended to use this in production.
+         * <p>
+         * This might be useful in non-production environments where you use a self-signed certificate
+         * for testing for example.
+         */
+        public Builder withoutSSLVerification() {
```
cmelchior: I much prefer `disableSSLVerification`

realm/realm-library/src/objectServer/java/io/realm/SyncConfiguration.java
```diff
@@ -596,6 +646,42 @@ public Builder errorHandler(SyncSession.ErrorHandler errorHandler) {
...
+            this.serverCertificateAssetName = filename;
+            return this;
+        }
+
+        /**
+         * This will disable TLS certificate verification for the remote Realm Object Server.
+         * It is not recommended to use this in production.
+         * <p>
+         * This might be useful in non-production environments where you use a self-signed certificate
+         * for testing for example.
```
cmelchior: Since you start with `this might` , i would probably just remove `for example`.

realm/realm-library/src/objectServer/java/io/realm/SyncConfiguration.java
```diff
@@ -733,6 +819,16 @@ public SyncConfiguration build() {
                 throw new IllegalStateException("Could not create directory for saving the Realm: " + realmFileDirectory);
             }
 
+            if (!TextUtils.isEmpty(serverCertificateAssetName)) {
+                if (!syncClientValidateSsl) {
+                    throw new IllegalStateException("SSL Verification is disable, server certificate should not be provided");
```
cmelchior: It feels a bit weird to me to have this check as well. The code for this would look like:

```
if (BuildType.DEBUG) {
  builder.disableSSLVerification()
} else {
  builder.trustedCA("myserver.pem")
}
```
Without this check it would be 
```
builder.trustedCA("myserver.pem")
if (BuildType.DEBUG) {
  builder.disableSSLVerification()
}
```
Which feels nicer to me?

realm/realm-library/src/syncIntegrationTest/java/io/realm/SSLConfigurationTests.java
```diff
@@ -0,0 +1,198 @@
+/*
+ * Copyright 2016 Realm Inc.
```
cmelchior: 2017

realm/realm-library/src/syncIntegrationTest/java/io/realm/SSLConfigurationTests.java
```diff
@@ -0,0 +1,198 @@
...
+    public Timeout globalTimeout = Timeout.seconds(10);
+
+    @Rule
+    public final TestSyncConfigurationFactory configurationFactory = new TestSyncConfigurationFactory();
+
+    @Test
+    public void trustedRootCA() throws InterruptedException {
+        String username = UUID.randomUUID().toString();
+        String password = "password";
+        SyncUser user = SyncUser.login(SyncCredentials.usernamePassword(username, password, true), Constants.AUTH_URL);
```
cmelchior: It feels slightly weird that this will use the Java defined CA's instead of our custom one, but that is probably outside the scope of this PR.

realm/realm-library/src/syncIntegrationTest/java/io/realm/SSLConfigurationTests.java
```diff
@@ -0,0 +1,198 @@
...
+
+        // 2. Local state should now be completely reset. Open the Realm again with a new configuration which should
+        // download the uploaded changes.
+        user = SyncUser.login(SyncCredentials.usernamePassword(username, password), Constants.AUTH_URL);
+        SyncConfiguration config = new SyncConfiguration.Builder(user, Constants.USER_REALM_SECURE)
+                .schema(StringOnly.class)
+                .build();
+        realm = Realm.getInstance(config);
+        try {
+            assertTrue(realm.isEmpty());
```
cmelchior: Isn't this reporteed as not working somewhere else, like through the ErrorHandler? It feels problematic that Sync just silently fails.

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/SyncedRealmTests.java
```diff
@@ -1,184 +0,0 @@
...
-
-    @Rule
-    public RunInLooperThread looperThread = new RunInLooperThread();
-
-    @Rule
-    public final UiThreadTestRule uiThreadTestRule = new UiThreadTestRule();
-
-    @Test
-    @UiThreadTest
-    public void waitForInitialRemoteData_mainThreadThrows() {
```
cmelchior: All these tests are in the other class, right?

realm/realm-library/src/main/java/io/realm/RealmCache.java
nhachicha: we're already using `TextUtils` in the code base ex: https://github.com/realm/realm-java/blob/9cea9356930b4357403b46ef4ca9995598a078ce/realm/realm-library/src/main/java/io/realm/RealmConfiguration.java#L184

cmelchior: That is also a mistake 😉

realm/realm-library/src/syncIntegrationTest/java/io/realm/SSLConfigurationTests.java
cmelchior: We should file a change request against Sync then. That doesn't sem very user friendly.

CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## 3.3.0
+
+### Enhancements
+
+* [ObjectServer] Added two options to `SyncConfiguration` to provide a trusted root CA `trustedRootCA` and to disable SSL validation `disableSSLVerification`. fixes (#4371).
```
cmelchior: I would just remove `fixes`. We haven't done that elsewhere.

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
cmelchior: I would probably go with the `util:none` approach in this case, especially since the list of parameters are so long

realm/realm-library/src/main/java/io/realm/RealmCache.java
cmelchior: Still present?

realm/realm-library/src/main/java/io/realm/internal/SharedRealm.java
cmelchior: what about `boolean syncClientValidateSsl = (syncUserConf[4] == Boolean.TRUE)`?

nhachicha:
@realm/java

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
```diff
@@ -130,8 +132,17 @@ class JniConfigWrapper {
...
+
+        util::Optional<std::string> ssl_trust_certificate_path = util::none;
+        if (sync_ssl_trust_certificate_path) {
+            ssl_trust_certificate_path =
+                realm::util::Optional<std::string>(JStringAccessor(env, sync_ssl_trust_certificate_path));
+        }
         m_config.sync_config = std::make_shared<SyncConfig>(SyncConfig{
-            user, realm_url, SyncSessionStopPolicy::Immediately, std::move(bind_handler), std::move(error_handler)});
+            user, realm_url, SyncSessionStopPolicy::Immediately, std::move(bind_handler), std::move(error_handler),
+            nullptr, util::none, sync_client_validate_ssl, ssl_trust_certificate_path});
```
beeender: is the `realm_encryption_key` newly added param to build `SyncConfig`? Don't we need to pass the key here?

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -527,55 +528,69 @@ synchronized void invokeWithLock(Callback0 callback) {
...
+        String syncServerCertificateAssetName = ObjectServerFacade.getFacade(configuration.isSyncConfiguration()).getSyncServerCertificateAssetName(configuration);
+        if (!Util.isEmptyString(syncServerCertificateAssetName)) {
+            String syncServerCertificateFilePath = ObjectServerFacade.getFacade(configuration.isSyncConfiguration()).getSyncServerCertificateFilePath(configuration);
+
+            File certificateFile = new File(syncServerCertificateFilePath);
+            copyFileIfNeeded(syncServerCertificateAssetName, certificateFile);
+        }
+    }
+
+    private static void copyFileIfNeeded(String assetFileName, File certificateFile) {
```
beeender: name for `certificateFile` is confusing, this method is used for copying asset realm file as well.

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -527,55 +528,69 @@ synchronized void invokeWithLock(Callback0 callback) {
...
-                    outputStream.write(buf, 0, bytesRead);
-                }
-            } catch (IOException e) {
+        // Copy Sync Server certificate path if available
+        String syncServerCertificateAssetName = ObjectServerFacade.getFacade(configuration.isSyncConfiguration()).getSyncServerCertificateAssetName(configuration);
+        if (!Util.isEmptyString(syncServerCertificateAssetName)) {
+            String syncServerCertificateFilePath = ObjectServerFacade.getFacade(configuration.isSyncConfiguration()).getSyncServerCertificateFilePath(configuration);
+
+            File certificateFile = new File(syncServerCertificateFilePath);
+            copyFileIfNeeded(syncServerCertificateAssetName, certificateFile);
```
beeender: corner case: what if user wants to change the cert file to a different one with the same name?

realm/realm-library/src/main/java/io/realm/internal/SharedRealm.java
```diff
@@ -232,7 +234,9 @@ public static SharedRealm getInstance(RealmConfiguration config, SchemaVersionLi
                 syncRealmUrl,
                 syncRealmAuthUrl,
                 syncUserIdentifier,
-                syncRefreshToken);
+                syncRefreshToken,
+                syncClientValidateSsl,
+                syncSslTrustCertificatePath);
```
beeender: 14 params for creating config :P 🎉 we may want to find a way to make this look better, but it is not the scope of this PR.

realm/realm-library/src/objectServer/java/io/realm/SyncConfiguration.java
```diff
@@ -596,6 +647,42 @@ public Builder errorHandler(SyncSession.ErrorHandler errorHandler) {
         }
 
         /**
+         * Provide the trusted root certificate(s) authority (CA) in {@code PEM} format, that should be used to
```
beeender: `Provide` -> `Provides`

realm/realm-library/src/objectServer/java/io/realm/SyncConfiguration.java
```diff
@@ -596,6 +647,42 @@ public Builder errorHandler(SyncSession.ErrorHandler errorHandler) {
...
+         * validate the TLS connections to the Realm Object Server.
+         * <p>
+         * The file should be stored under {@code assets}, it will be copied at runtime into the internal storage.
+         * <p>
+         * Note: This is similar to passing the parameter {@code CAfile} to {@code SSL_CTX_load_verify_locations},
+         *       Therefor it is recommended to include only the root CA you trust, and not the entire list of root CA
```
beeender: `Therefore`

realm/realm-library/src/syncIntegrationTest/java/io/realm/SSLConfigurationTests.java
```diff
@@ -0,0 +1,204 @@
...
+        // make sure the changes gets to the server
+        SystemClock.sleep(TimeUnit.SECONDS.toMillis(2));  // FIXME: Replace with Sync Progress Notifications once available.
+        realm.close();
+        user.logout();
+        Realm.deleteRealm(configOld);
+
+        // 2. Local state should now be completely reset. Open the Realm again with a new configuration which should
+        // download the uploaded changes.
+        user = SyncUser.login(SyncCredentials.usernamePassword(username, password), Constants.AUTH_URL);
+        SyncConfiguration config = new SyncConfiguration.Builder(user, Constants.USER_REALM_SECURE)
```
beeender: why not using `configurationFactory.createSyncConfigurationBuilder`  here?

realm/realm-library/src/syncIntegrationTest/java/io/realm/SSLConfigurationTests.java
```diff
@@ -0,0 +1,204 @@
...
```
beeender: ditto

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
nhachicha: yes setting the `realm_encryption_key` here should fix https://github.com/realm/realm-java/issues/4561 as well

realm/realm-library/src/main/java/io/realm/RealmCache.java
nhachicha: This is a similar use case with the existing [implementation](https://github.com/realm/realm-java/blob/deb25a87d08d93ebc77c5c8d9e81b8a632485b83/realm/realm-library/src/main/java/io/realm/RealmCache.java#L533) i.e what happen if you want to change the Realm shipped in the asset file with the same name. 

For now both use cases will need, the user to uninstall or clear internal data

realm/realm-library/src/main/java/io/realm/internal/SharedRealm.java
nhachicha: Kotlin to the help with named params! 😅

PR #4596, opened 2017-05-03
Remove onSuccess callback before calling it

### LINKED ISSUE #4595: OnSuccess callback called infinite times for Async Transaction
#### Goal

Do an action on onSuccess of async transaction 

#### Expected Results

onSuccess listener invoked once 

#### Actual Results

onSuccess Listener invoked infinite times

#### Code Sample

```java
  realm.executeTransactionAsync(r -> {

            ChatDialog dialog = r.where(ChatDialog.class).equalTo(
                    ChatDialog.FIELDS.DIALOG_ID, id).findFirst();
            if (dialog != null) {
                dialog.getChatInfo().setDealerChatStatus(status);
                dialog.getChatInfo().setConsumerChatStatus(status);
            }
        }, success);
```

And here is the onSuccess Code 
```
 realm.executeTransaction(r -> {
            List<ChatDialog> ChatDialogs = realm.where(ChatDialog.class).equalTo(
                    ChatDialog.FIELDS.DEALER_CHAT_STATUS, localStatus).findAll();
            if (ChatDialogs.size() > 0) {
                for (ChatDialog carChatDialog :ChatDialogs) {
                    carChatDialog.getChatInfo().setDealerChatStatus(status);
                   ChatService.start(context,
                            SyncDialogStatusCommand.getSyncDialogStatusCommandIntent(
                                    carChatDialog.getDialogId()));
                }
            }
        });
```

Realm version(s): 3.0.0 but this issue not produceable on same code using version 1.2.0

Android Studio version: 2.3.1

Which Android version and device: 6.0.0, galaxy note 5

beeender:
Otherwise it will cause infinite recursion if a transaction is committed
in the callback.
Fix #4595

### DIFF
--- a/realm/realm-library/src/main/java/io/realm/internal/RealmNotifier.java
+++ b/realm/realm-library/src/main/java/io/realm/internal/RealmNotifier.java
@@ -18,6 +18,7 @@ package io.realm.internal;
 
 import java.io.Closeable;
 import java.util.ArrayList;
+import java.util.Iterator;
 import java.util.List;
 
 import io.realm.RealmChangeListener;
@@ -97,10 +98,15 @@ public abstract class RealmNotifier implements Closeable {
     // called from java_binding_context.cpp
     void didChange() {
         realmObserverPairs.foreach(onChangeCallBack);
-        for (Runnable runnable : transactionCallbacks) {
+
+        Iterator<Runnable> it = transactionCallbacks.iterator();
+        while (it.hasNext()) {
+            Runnable runnable = it.next();
+            // The callback needs to be removed from list before calling to avoid sync transactions in the callback
+            // triggers it recursively.
+            it.remove();
             runnable.run();
         }
-        transactionCallbacks.clear();
     }
 
     // Called from JavaBindingContext::before_notify.
--- a/realm/realm-library/src/main/java/io/realm/internal/RealmNotifier.java
+++ b/realm/realm-library/src/main/java/io/realm/internal/RealmNotifier.java
@@ -99,13 +99,14 @@ public abstract class RealmNotifier implements Closeable {
     void didChange() {
         realmObserverPairs.foreach(onChangeCallBack);
 
-        Iterator<Runnable> it = transactionCallbacks.iterator();
-        while (it.hasNext()) {
-            Runnable runnable = it.next();
-            // The callback needs to be removed from list before calling to avoid sync transactions in the callback
+        if (!transactionCallbacks.isEmpty()) {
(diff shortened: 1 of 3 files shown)


realm/realm-library/src/main/java/io/realm/internal/RealmNotifier.java
```diff
@@ -97,10 +98,15 @@ protected RealmNotifier(SharedRealm sharedRealm) {
...
     void didChange() {
         realmObserverPairs.foreach(onChangeCallBack);
-        for (Runnable runnable : transactionCallbacks) {
+
+        Iterator<Runnable> it = transactionCallbacks.iterator();
+        while (it.hasNext()) {
+            Runnable runnable = it.next();
+            // The callback needs to be removed from list before calling to avoid sync transactions in the callback
+            // triggers it recursively.
+            it.remove();
```
zaki50: Does this work if transactionCallbacks contains multiple callbacks?
If previous code causes recursive call, I would think ConcurrentModificationException will be thrown.

beeender: hmmm, good catch! will create an issue for that,

Zhuinden: This reminds me of the good old times of

    while(!transactionCallbacks.isEmpty()) {
         Runnable runnable = transactionCallbacks.remove(0);
         runnable.run();
    }

realm/realm-library/src/main/java/io/realm/internal/RealmNotifier.java
```diff
@@ -97,10 +98,16 @@ protected RealmNotifier(SharedRealm sharedRealm) {
     // called from java_binding_context.cpp
     void didChange() {
         realmObserverPairs.foreach(onChangeCallBack);
-        for (Runnable runnable : transactionCallbacks) {
-            runnable.run();
+
+        if (!transactionCallbacks.isEmpty()) {
+            // The callback list needs to be cleared before calling to avoid sync transactions in the callback
```
zaki50: `sync transactions` is a bit ambiguous since we have `sync` feature and it also uses transaction.

Does `synced transactions` or `synchronized transactions` solve this?

realm/realm-library/src/main/java/io/realm/internal/RealmNotifier.java
```diff
@@ -18,6 +18,7 @@
 
 import java.io.Closeable;
 import java.util.ArrayList;
+import java.util.Iterator;
```
zaki50: Not used anymore?

PR #4597, opened 2017-05-03
Rename Context to NativeContext

cmelchior:
Renames `io.realm.internal.Context` to `io.realm.internal.NativeContext` (started out as RealmContext) to avoid confusion with `android.content.Context`

### DIFF
--- a/realm/realm-library/src/main/java/io/realm/internal/CheckedRow.java
+++ b/realm/realm-library/src/main/java/io/realm/internal/CheckedRow.java
@@ -33,7 +33,7 @@ public class CheckedRow extends UncheckedRow {
     @SuppressWarnings({"unused", "FieldCanBeLocal"})
     private UncheckedRow originalRow;
 
-    private CheckedRow(Context context, Table parent, long nativePtr) {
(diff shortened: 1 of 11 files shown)


cmelchior:
Both seems fitting to me, so I can change it if you think `NativeContext` is a better fit?

beeender:
I think `NativeContext` matches the `NativeObject` better. So please change it? Thanks!

bmeike:
I'm curious about what the thinking is behind the name "NativeContext".  My first thought would be "RealmContext", because, well, that's sort of what it is.  What is it that is clarified by calling it the "NativeContext"?  Native what?

I have no objection at all to the name.  I certainly don't understand the choice, though.

Zhuinden:
@bmeike it's the thing that keeps track of Java objects that are orphaned and then releases their references once they are finalized by the ~~GC~~ finalizer thread. 

So it's that thing that keeps track of native references and then finalizes them.

cmelchior:
Yes, `NativeContext` kinda matches `NativeObject` and is a bit more focused on the native part of the code base, but I could have lived with both as well.

PR #4599, opened 2017-05-04
Fix flaky test

beeender:


beeender:
It was because of waiting for all `RealmAsyncTask` done has been moved the looperthread instead of the configfactory.

PR #4603, opened 2017-05-04
    Wait for async tasks finish to solve flaky test

beeender:
Apparently waiting some random millisecond is never a good idea.

### DIFF
--- a/realm/realm-library/src/androidTest/java/io/realm/RealmAsyncQueryTests.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/RealmAsyncQueryTests.java
@@ -37,6 +37,7 @@ import io.realm.entities.AnnotationIndexTypes;
 import io.realm.entities.Dog;
 import io.realm.entities.NonLatinFieldNames;
 import io.realm.entities.Owner;
+import io.realm.internal.async.RealmThreadPoolExecutor;
 import io.realm.log.LogLevel;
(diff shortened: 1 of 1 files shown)


beeender:
DoH!

cmelchior:
> Apparently waiting some random millisecond is never a good idea.

Shocking 😄

realm/realm-library/src/androidTest/java/io/realm/RealmAsyncQueryTests.java
```diff
@@ -458,6 +454,17 @@ public void onSuccess() {
                 looperThread.testComplete();
             }
         });
+
+        // Wait for all async tasks finish to ensure the async transaction posted callback will arrive first.
+        TestHelper.resetRealmThreadExecutor();
```
zaki50: Shouldn't we use `try - finally` to ensure that executor is restored?

beeender: I think when that happens, we should fix it. `NoSuchFieldException` and `IllegalAccessException` doesn't sounds like anything can be recovered...

cmelchior: try...finally will not fix it, the exception will still be thrown, but it will ensure we are not breaking 100 other tests.

PR #4607, opened 2017-05-08
Clean up SharedRealm source

beeender:
- Remove unused methods.
- Fix double negatives.
- Use auto& to avoid creating temp objects.

### DIFF
--- a/realm/realm-library/src/main/java/io/realm/internal/SharedRealm.java
+++ b/realm/realm-library/src/main/java/io/realm/internal/SharedRealm.java
@@ -128,6 +128,7 @@ public final class SharedRealm implements Closeable, NativeObject {
 
         @Override
         public int compareTo(@SuppressWarnings("NullableProblems") VersionID another) {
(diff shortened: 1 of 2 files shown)


realm/realm-library/src/main/java/io/realm/internal/SharedRealm.java
```diff
@@ -214,17 +215,18 @@ public static SharedRealm getInstance(RealmConfiguration config, SchemaVersionLi
         String syncRealmUrl = syncUserConf[1];
         String syncRealmAuthUrl = syncUserConf[2];
         String syncRefreshToken = syncUserConf[3];
-        boolean enable_caching = false; // Handled in Java currently
-        boolean disableFormatUpgrade = false; // TODO Double negatives :/
+
+        final boolean enableCaching = false; // Handled in Java currently
+        final boolean enableFormatUpgrade = true;
```
kneth: Thank you!

PR #4609, opened 2017-05-08
Use shorthand for generic arguments for better AS support

cmelchior:
By using a shorthand for the generic argument when creating Changelisteners, IntelliJ (and Android Studio) will expand that automatically to be a camel-cased version of the generic argument. This makes the code more readable with minimal effort. Existing code will not be affected.

Example:
```
// Before
RealmResults<Permission> permissions = realm.where(Permission.class).findAll();
permissions.addChangeListener(new RealmChangeListener<RealmResults<Permission>>() {
  @Override
  public void onChange(RealmResults<Permission> element) {
     // user forced to manually rename `element` to something more descriptive
  }
});

// Now 
permissions.addChangeListener(new RealmChangeListener<RealmResults<Permission>>() {
  @Override
  public void onChange(RealmResults<Permission> permissions) {
     // IntelliJ will automatically rename the variable to 'permissions' even though it is really named 't'.
  }
});
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 * Added `Realm.refresh()` and `DynamicRealm.refresh()` (#3476).
 * Added `Realm.getInstanceAsync()` and `DynamicRealm.getInstanceAsync()` (#2299).
 * Added `DynamicRealmObject#linkingObjects(String,String)` to support linking objects on `DynamicRealm` (#4492).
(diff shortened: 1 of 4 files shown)


zaki50:
I think that it's better to target `releases` branch, but it's OK

beeender:
yeah, I think it should go to `releases` branch.

cmelchior:
Yeah, it probably should have gone to `releases`, but given that the next release will be a minor one, I'll take the liberty to merge this to `master`.

PR #4613, opened 2017-05-09
Fixed @LinkingObjects notation for Kotlin

### LINKED ISSUE #4611: Inverse Relationships not compatible with Kotlin
We are trying to implement [Inverse Relationships](https://news.realm.io/news/realm-java-3-1) in our Kotlin codebase but when we decorate our model with the '@LinkingObjects' annotation we get the following compilation error:
```
Error:Field 'messages' of type "io.realm.RealmResults<xxxxx.models.Message>" is not supported.

Warning:Unclosed files for the types '[io.realm.ChatRealmProxy]'; these types will not undergo annotation processing
```

The definition for the objects involved is the following:
``` Kotlin
@RealmClass
open class Chat : RealmModel {
    ...
    @LinkingObjects( "chat" )
    var messages:RealmResults<Message>? = null
    ...
}


@RealmClass
open class Message : RealmModel {
    ...
    open var chat:Chat? = null
    ...
}
```

Is the use of annotations to implement the Inverse Relationships incompatible with Kotlin?

Realm version: 3.1.4
Android Studio version: 2.4 Preview 7
Which Android version and device: Any

cmelchior:
Fixes #4611 

Apparently, annotations need CLASS retention in order to work with Kotlin. I'm guessing some combination of how annotation processors work with the Kotlin compiler.

Also updated the example code to be more correct (don't need open on fields / updated Anko). And yes that should probably have been in another PR 😄

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+# 3.1.5 (YYYY-MM-DD)
+
+## Bug fixes
+
+* `@LinkingObjects` annotation now also works with Kotlin (#4611).
+
+
 ## 3.1.4 (2017-05-04)
 
 ## Bug fixes
--- a/examples/kotlinExample/build.gradle
+++ b/examples/kotlinExample/build.gradle
@@ -46,5 +46,5 @@ android {
 
 dependencies {
     compile "org.jetbrains.kotlin:kotlin-stdlib:${kotlin_version}"
-    compile 'org.jetbrains.anko:anko-sdk15:0.8.2'
+    compile 'org.jetbrains.anko:anko-sdk15:0.9.1'
 }
--- a/examples/kotlinExample/src/main/kotlin/io/realm/examples/kotlin/KotlinExampleActivity.kt
+++ b/examples/kotlinExample/src/main/kotlin/io/realm/examples/kotlin/KotlinExampleActivity.kt
@@ -26,7 +26,7 @@ import io.realm.Sort
 import io.realm.examples.kotlin.model.Cat
 import io.realm.examples.kotlin.model.Dog
 import io.realm.examples.kotlin.model.Person
-import org.jetbrains.anko.async
+import org.jetbrains.anko.doAsync
 import org.jetbrains.anko.uiThread
 import kotlin.properties.Delegates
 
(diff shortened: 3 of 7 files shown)


marchy:
Oh interesting, so only the class itself needs to be marked as 'open'?
As per the Kotlin documentation ([Overriding Methods](https://kotlinlang.org/docs/reference/classes.html) section) "_...unlike Java, Kotlin requires explicit annotations for overridable members (we call them open) and for overrides_".

Furthermore "_If there is no open annotation on a function, like Base.nv(), declaring a method with the same signature in a subclass is illegal, either with override or without it._"

I take it that the proxy generation mechanism that Realm uses somehow circumvents these restrictions? (If so, it may be something that could change in the future, but in this case... would be very useful)
PS: The whole 'open' decision is heavily disputed in the Kotlin community and am personally a strong believer it is mistake in Kotlin – for these exact kinds of reasons.


As for the example code it's a part we struggled with a lot as it feels quite misaligned with Kotlin/Swift best-practices. In particular the guidance about defining properties:
``` Kotlin
// You can put properties in the constructor as long as all of them are initialized with
// default values. This ensures that an empty constructor is generated.
```
is against Kotlin/Swift non-nullable initialization practices. In Swift we use "!" to denote a late-bound initialization and in Kotlin "lateinit".

This seems to work in our codebase, but if you can confirm that 'lateinit' works with all your test cases I can file a separate ticket to have the docs updated.

i.e.: Non-nullable Realm properties should be defined as:
``` Kotlin
@PrimaryKey lateinit var id: Long
lateinit var name: String
lateinit var age: Int

// Other objects in a one-to-one relation must also subclass RealmObject
var dog: Dog? = null
  		  
// One-to-many relations is simply a RealmList of the objects which also subclass RealmObject
var cats = RealmList<Cat>(),
```
instead of 
``` Kotlin
@PrimaryKey var id: Long = 0  		  
var name: String = "",  		  
var age: Int = 0,

// Other objects in a one-to-one relation must also subclass RealmObject
var dog: Dog? = null,
  		  
// One-to-many relations is simply a RealmList of the objects which also subclass RealmObject
var cats: RealmList<Cat> = RealmList(),
```

Zhuinden:
Class needs to be open in order to allow generated `____RealmProxy` classes to exist.

Shahroz16:
Any update on this?

cmelchior:
This has been merged so will be part of the next release. Until then you can use the `3.2.0-SNAPSHOT` if you want.

PR #4623, opened 2017-05-10
Make the proxy toString methods reveal themselves

bmeike:
Make the `toString` methods in the DynamicObject Proxy and the Realm Proxy more up-front about who  they are.

### DIFF
--- a/realm/realm-library/src/androidTest/java/io/realm/DynamicRealmObjectTests.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/DynamicRealmObjectTests.java
@@ -1220,7 +1220,7 @@ public class DynamicRealmObjectTests {
         // Checks that toString() doesn't crash, and does simple formatting checks. We cannot compare to a set String as
         // eg. the byte array will be allocated each time it is accessed.
         String str = dObjTyped.toString();
-        assertTrue(str.startsWith("AllJavaTypes = dyn["));
+        assertTrue(str.startsWith("AllJavaTypes = dynamic["));
         assertTrue(str.endsWith("}]"));
     }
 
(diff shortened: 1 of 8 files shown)


zaki50:
I feel this is an enhancement and should go to `master`, but it's OK.
Could you add an entry for this to CHANGELOG?

realm/realm-library/src/androidTest/java/io/realm/DynamicRealmObjectTests.java
```diff
@@ -1220,7 +1220,7 @@ public void toString_test() {
         // Checks that toString() doesn't crash, and does simple formatting checks. We cannot compare to a set String as
         // eg. the byte array will be allocated each time it is accessed.
         String str = dObjTyped.toString();
-        assertTrue(str.startsWith("AllJavaTypes = ["));
+        assertTrue(str.startsWith("AllJavaTypes = dyn["));
```
cmelchior: No reason to save chars here. I would strongly prefer `dynamic`

cmelchior: Also nitpicking I would personally prefer `AllJavaTypes (Proxy) = [ ]` and `AllJavaTypes (DynamicObject) = [ ]` since the right-hand side of the `=` is the objects data and the left hand side is the meta information.

Zhuinden: I think bmeike's is a bit more concise which makes it a bit more expressive, but it might just be me.

bmeike: Changed to "dynamic["

bmeike:
@kneth : Where is the documentation of which you speak?  Will open a separate issue to fix.

PR #4632, opened 2017-05-11
Refactor object creation into OsObject

beeender:
- Hide the addEmptyRow from java to support future stable ID.
- Add bulk insertion benchmark.

This won't be the final design of internal object creation API, when
integration of OS object accessor, the internal API might be changed a
bit since it doesn't look nice at all -- 5 params for the JNI call to
create an object with integer primary key!


beeender:
Benchmark for bulk insertion:
Left is before without this change.
image

beeender:
Also, some internal APIs are not needed anymore, I will clean them in another PR.

realm/realm-library/src/main/cpp/io_realm_internal_OsObject.cpp
```diff
@@ -152,6 +155,87 @@ static void finalize_object(jlong ptr)
...
+    auto& shared_realm = *(reinterpret_cast<SharedRealm*>(shared_realm_ptr));
+    auto& table = *(reinterpret_cast<realm::Table*>(table_ptr));
+    shared_realm->verify_in_write(); // throws
+    if (is_pk_null && !TBL_AND_COL_NULLABLE(env, &table, pk_column_ndx)) {
+        return npos;
+    }
+
+    if (is_pk_null) {
+        if (table.find_first_null(pk_column_ndx) != npos) {
+            throw_duplicated_primary_key_exception(env, "'null'");
```
kneth: Why not throw a C++ exception and let `CATCH_STD()` do the work?

realm/realm-library/src/main/cpp/io_realm_internal_OsObject.cpp
```diff
@@ -152,6 +155,87 @@ static void finalize_object(jlong ptr)
...
+            return npos;
+        }
+    }
+    else {
+        if (table.find_first_int(pk_column_ndx, pk_value) != npos) {
+            throw_duplicated_primary_key_exception(env, reinterpret_cast<long long>(pk_value));
```
kneth: Like above

realm/realm-library/src/main/cpp/io_realm_internal_OsObject.cpp
beeender: I just came up with some ideas for a better way to throw java exceptions from JNI. will fix this in a separated PR.

realm/realm-library/src/main/java/io/realm/internal/OsObject.java
```diff
@@ -156,4 +243,25 @@ private void notifyChangeListeners(String[] changedFields) {
...
+    private static native long nativeCreateNewObject(long sharedRealmPtr, long tablePtr);
+
+    private static native long nativeCreateRow(long sharedRealmPtr, long tablePtr);
+
+
+    private static native long nativeCreateNewObjectWithPrimaryKey(long sharedRealmPtr,
+                                                                   long tablePtr, long pk_column_index,
+                                                                   long primaryKeyValue, boolean isNullValue);
+
+    private static native long nativeCreateRowWithPrimaryKey(long sharedRealmPtr,
```
zaki50: The meanings of the return values of `nativeCreateNewObject*` and `nativeCreateRow*` is a bit confusing.
`nativeCreateNewObject*` returns a pointer to native `Row` object and `nativeCreateRow*` returns an index of the created row in the `Table` (this method does not create `Row` instance).

Shouldn't we add a javadoc comment at least?

zaki50: I think it's better to use different method name for `long` and `String` (`nativeCreateRowWithLongPrimaryKey `)?

realm/realm-library/src/main/cpp/io_realm_internal_OsObject.cpp
kneth: Can't wait :-)

realm/realm-library/src/benchmarks/java/io/realm/benchmarks/RealmInsertBenchmark.java
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    @Benchmark
+    public void insertNoPrimaryKey(long reps) {
+        AllTypes allTypes = new AllTypes();
+        realm.beginTransaction();
+        for (long i = 0; i < reps; i++) {
+            realm.insert(allTypes);
+        }
+        realm.commitTransaction();
```
cmelchior: Note that this method probably is probably the biggest contributor to this benchmark. If at all possible, I would try to move `beginTransaction/commitTransaction/cancelTransaction()` outside this method.

realm/realm-library/src/benchmarks/java/io/realm/benchmarks/RealmInsertBenchmark.java
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    @Benchmark
+    public void insertNoPrimaryKeyList(long reps) {
+        realm.beginTransaction();
+        for (long i = 0; i < reps; i++) {
+            realm.insert(noPkObjects);
+        }
+        realm.commitTransaction();
```
cmelchior: same

realm/realm-library/src/main/cpp/io_realm_internal_OsObject.cpp
```diff
@@ -204,3 +291,78 @@ JNIEXPORT void JNICALL Java_io_realm_internal_OsObject_nativeStopListening(JNIEn
...
+
+JNIEXPORT jlong JNICALL Java_io_realm_internal_OsObject_nativeCreateRowWithPrimaryKey__JJJJZ(
+    JNIEnv* env, jclass, jlong shared_realm_ptr, jlong table_ptr, jlong pk_column_ndx, jlong pk_value,
+    jboolean is_pk_null)
+{
+    try {
+        return do_create_row_with_primary_key(env, shared_realm_ptr, table_ptr, pk_column_ndx, pk_value, is_pk_null);
+    }
+    CATCH_STD()
+    return npos;
```
cmelchior: What value is npos? Wouldn't -1 be more readable?

realm/realm-library/src/main/cpp/io_realm_internal_OsObject.cpp
```diff
@@ -204,3 +291,78 @@ JNIEXPORT void JNICALL Java_io_realm_internal_OsObject_nativeStopListening(JNIEn
...
+
+JNIEXPORT jlong JNICALL Java_io_realm_internal_OsObject_nativeCreateRowWithPrimaryKey__JJJLjava_lang_String_2(
+    JNIEnv* env, jclass, jlong shared_realm_ptr, jlong table_ptr, jlong pk_column_ndx, jstring pk_value)
+{
+    try {
+        return do_create_row_with_primary_key(env, shared_realm_ptr, table_ptr, pk_column_ndx, pk_value);
+    }
+    CATCH_STD()
+
+    return npos;
```
cmelchior: Same

realm/realm-library/src/main/java/io/realm/internal/OsObject.java
cmelchior: 👍 It confused me as well.

realm/realm-library/src/main/cpp/io_realm_internal_OsObject.cpp
beeender: It is actually a value defined in core. See `array.hpp`. Will it be less confused with namespace here? like `return realm::npos`

realm/realm-library/src/main/cpp/io_realm_internal_OsObject.cpp
```diff
@@ -152,14 +155,98 @@ static void finalize_object(jlong ptr)
...
+    return row_ndx;
+}
+
+static inline size_t do_create_row_with_primary_key(JNIEnv* env, jlong shared_realm_ptr, jlong table_ptr,
+                                                    jlong pk_column_ndx, jstring pk_value)
+{
+    auto& shared_realm = *(reinterpret_cast<SharedRealm*>(shared_realm_ptr));
+    auto& table = *(reinterpret_cast<realm::Table*>(table_ptr));
+    JStringAccessor str_accessor(env, pk_value); // throws
+    shared_realm->verify_in_write();             // throws
```
beeender: changed the `Table.throwImmutable` to the same message. I think it is fine to not mention it in the changelog, the two sentences basically have the same meaning.

realm/realm-library/src/main/cpp/io_realm_internal_OsObject.cpp
kneth: Adding name space will increase readability.

realm/realm-library/src/main/cpp/io_realm_internal_OsObject.cpp
zaki50: The only way for users to distinguish `IllegalStateException`s is to check the message in it.
I guess some users in the world is actually doing that. Since the message itself is not a public interface, I don't think that changing it is a breaking change, but shouldn't we notice it at least?

PR #4636, opened 2017-05-12
Helper class for throwing java exception from JNI

beeender:


realm/realm-library/src/main/cpp/jni_util/java_exception_thrower.hpp
```diff
@@ -0,0 +1,49 @@
...
+public:
+    JavaExceptionThrower(JNIEnv* env, const char* class_name, std::string message)
+        : std::runtime_error(std::move(message))
+        , m_exception_class(env, class_name)
+    {
+    }
+
+    virtual void throw_java_exception(JNIEnv* env)
+    {
+        env->ThrowNew(m_exception_class, what());
```
zaki50: No need to print the message to error log?
`ThrowException` is has the following code before calling `ThrowNew()`.
```
        Log::e("Exception has been thrown: %1", message.c_str());
```

realm/realm-library/src/main/cpp/jni_util/java_exception_thrower.hpp
```diff
@@ -0,0 +1,49 @@
...
+namespace realm {
+namespace jni_util {
+
+// Class to help throw a Java exception from JNI code.
+// This exception will be called from CATCH_STD and throw a Java exception there.
+class JavaExceptionThrower : public std::runtime_error {
+public:
+    JavaExceptionThrower(JNIEnv* env, const char* class_name, std::string message)
+        : std::runtime_error(std::move(message))
+        , m_exception_class(env, class_name)
```
zaki50: Do we need to enable `free_on_unload`? Shouldn't it be `m_exception_class(env, class_name, false)`?

realm/realm-library/src/main/cpp/jni_util/java_exception_thrower.hpp
cmelchior: 1) If the Exception is caught and handled, it would be annoying to print it to the log.
2) If the Exception isn't caught, the app will crash and print the error to the log anyway.

IMO we shouldn't print the log?

realm/realm-library/src/main/cpp/io_realm_internal_OsObject.cpp
```diff
@@ -26,11 +26,15 @@
 #include "jni_util/java_global_weak_ref.hpp"
 #include "jni_util/java_method.hpp"
 #include "jni_util/java_class.hpp"
+#include "jni_util/java_exception_thrower.hpp"
 
 using namespace realm;
 using namespace realm::jni_util;
 using namespace realm::_impl;
 
+static const char* PK_CONSTRAINT_EXCEPTION_CLASS = "io/realm/exceptions/RealmPrimaryKeyConstraintException";
```
cmelchior: Would it make sense to move all Java exception types into the java_exception_thrower class instead of having them scattered around?

realm/realm-library/src/main/cpp/jni_util/java_exception_thrower.hpp
kneth: It won't hurt to log. Maybe at level `info` or `warning`?

realm/realm-library/src/main/cpp/io_realm_internal_OsObject.cpp
```diff
@@ -182,14 +179,13 @@ static inline size_t do_create_row_with_primary_key(JNIEnv* env, jlong shared_re
 
     if (is_pk_null) {
         if (table.find_first_null(pk_column_ndx) != npos) {
-            throw_duplicated_primary_key_exception(env, "'null'");
-            return npos;
+            throw JavaExceptionThrower(env, PK_CONSTRAINT_EXCEPTION_CLASS, format(PK_EXCEPTION_MSG_FORMAT, "'null'"));
```
kneth: Should we add `__FILE__` and `__LINE__` (or another unique identifier) to be can easily find where the exception is thrown?

realm/realm-library/src/main/cpp/io_realm_internal_OsObject.cpp
beeender: If the exception is used by all modules, yes, I think it is a good idea to have all exceptions in a centralized place. But currently, this exception is only used in this file. Let's see what is the best to do in the future maybe when we get more and more exceptions thrown in the JNI.

realm/realm-library/src/main/cpp/jni_util/java_exception_thrower.hpp
beeender: It actually doesn't matter I think ... `free_on_unload` is something will never be called on Android :P (JNI_OnUnload() is not called in android) And all most all the exceptions we throw is something should be fixed but not caught.

zaki50: If `free_on_unload ` is true, `JavaGlobalRef` object is pushed to static `vector` every time an exception is thrown and never be released.
Shouldn't we avoid those kind of memory leak as much as possible?

realm/realm-library/src/main/cpp/io_realm_internal_OsObject.cpp
bmeike: I still think it would be a great idea to put them all in one place.  Suppose that I am about to add a PK_CONSTRAINT_EXCEPTION to the code.  Why would I go looking in io_real_internal_OsObject.cpp, to find it?

beeender: OK, I will move those to another file in the other PR, together with other java exception thrown in JNI.

PR #4649, opened 2017-05-16
Support minor release for release script

beeender:


zaki50:
Could you remove a comment in line 5?
` # FIXME: Only patch release is supported now.`

tools/release.sh
```diff
@@ -93,14 +93,35 @@ prepare_branch() {
...
+        while true
+        do
+            read -r -p "Type the command to edit CHANGELOG.md, default(vim):" editor
+            if [ -z "$editor" ] ; then
+                editor="vim"
+            fi
+            "$editor" CHANGELOG.md
+
+            read -r -p "Please merge the unreleased entries in the 'CHANGELOG.md' and then press any key to continue..." _
+            if [ "$(grep -c "YYYY-MM-DD" CHANGELOG.md)" -eq 1 ] ; then
```
zaki50: We can write just`if grep -q "YYYY-MM-DD" CHANGELOG.md; then` here.

beeender: this is to check if the changelog entries are merged correctly. eg.:
```
2.2.0 (YYYY-MM-DD)

...

2.1.3 (YYYY-MM-DD)
```

The 2.1.3 should be removed and merge to 2.2.0 sections.

zaki50: Ah, `-eq 1` is important. Got it.

PR #4655, opened 2017-05-17
Vertical slice of PermissionManager + getPermissionsAsync

cmelchior:
First step getting the new PermissionManager into place.

This is vertical slice that functions as a proof of concept. Only `getPermissionsAsync()` have been implemented and alongside that the required infrastructure that the rest of the methods will also use.

There is a bunch of FIXME's in the code, they are being tracked in #4558. They will be fixed in separate PR's in order to keep this manageable.


cmelchior:
Ready for review @realm/java.
Most important class to review is `PermissionManager`. Keep in mind this is just an intermediate commit to `feature/permission-wrapper`

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -302,6 +302,7 @@ private static RealmCache getCache(String realmPath, boolean createIfNotExist) {
                 if (!fileExists) {
                     try {
                         ObjectServerFacade.getSyncFacadeIfPossible().downloadRemoteChanges(configuration);
+                        sharedRealm.refresh();
```
beeender: why calling `refresh()` here?

realm/realm-library/src/objectServer/java/io/realm/SyncUser.java
```diff
@@ -462,6 +477,15 @@ private static String getManagementRealmUrl(URL authUrl) {
         }
     }
 
+    /**
+     * FIXME Javadoc
+     *
+     * @return
+     */
+    public PermissionManager getPermissionManager() {
```
beeender: What is the designed way for user to get the `PermissionManager`? through this API or the constructor? It is confusing to me if the `close()` should be called on the return value of this API.

realm/realm-library/src/objectServer/java/io/realm/permissions/PermissionManager.java
```diff
@@ -0,0 +1,396 @@
...
+                    "accessed on the thread it was created on.");
+        }
+    }
+
+    /**
+     * FIXME Add Javadoc
+     */
+    @Override
+    public void close() {
+        // If Realms are still being opened, abort that task
```
zaki50: Shouldn't we call `checkIfValidThread()`?

realm/realm-library/src/objectServer/java/io/realm/permissions/PermissionManager.java
```diff
@@ -0,0 +1,396 @@
...
+        closed = true;
+    }
+
+    /**
+     * Checks if this PermissionManager is closed or not. If it is closed, all methods will report back an error
+     *
+     * @return {@code true} if the PermissionManager is closed, {@code false} if it is still open.
+     */
+    public boolean isClosed() {
+        return closed;
```
zaki50: Shouldn't we call `checkIfValidThread()`?

realm/realm-library/src/objectServer/java/io/realm/SyncUser.java
zaki50: Behavior of this method is a bit confusing.
This method returns cached instance if the returned object of previous call on the same thread was not closed.

Why do we need to provide `PermissionManager.close()` to users?

cmelchior: Ideally, the constructor wont be public. Trying to figure out a way to make it not so. The PermissionManager should be closed just Realms. I'll expand the javadoc description.

cmelchior: Because the PermissionManager is strongly tied to two underlying Realms. As long as they are open, the user cannot call logout and hopefully people would be used to this pattern if using Realm already. I agree it isn't as nice as one could hope. 

The other alternative would be to make `close()` private and then have a finalizer call it. We would also need to determine how `syncuser.logout()` would behave in that case. I would assume it would just forcibly close the PermissionManager. Right now it would throw an error if you tried to logout while the permissionmanager was open.

What do you think?

cmelchior: I had to move the PermissionManager to `io.realm` to solve this. `io.realm.permissions` now only contain Realm modules and model classes, which I guess is acceptable?

cmelchior:
All comments addressed. Ready for round 2 @realm/java

realm/realm-library/src/androidTest/java/io/realm/rule/RunInLooperThread.java
```diff
@@ -81,6 +83,14 @@
     // Access guarded by 'lock'
     private List<Realm> testRealms;
```
bmeike: There's an Annotation for that...

realm/realm-library/src/objectServer/java/io/realm/Permission.java
```diff
@@ -0,0 +1,123 @@
...
+
+        /**
+         * FIXME: add description
+         */
+        READ(true, false, false),
+
+        /**
+         * FIXME: add description
+         */
+        WRITE(true, true, false),
```
bmeike: I think this is mis-named.  It is ReadWrite permission.  I kind of wonder if it wouldn't be better to represent access as a set, possibly empty, of permissions...

realm/realm-library/src/objectServer/java/io/realm/Permission.java
```diff
@@ -0,0 +1,123 @@
...
+
+    public String getUserId() {
+        return userId;
+    }
+
+    public String getPath() {
+        return path;
+    }
+
+    public boolean isMayRead() {
```
bmeike: isMayWrite???  isMay15th?  isMayCute?

realm/realm-library/src/objectServer/java/io/realm/PermissionManager.java
```diff
@@ -0,0 +1,388 @@
...
+    // Activate a task. All tasks are controlled by the Handler in order to make it asynchronous.
+    // PRECONDITION: Underlying Realms are no longer in the process of being opened.
+    private void activateTask(PermissionManagerAsyncTask task) {
+        activeTasks.add(task);
+        handler.post(task);
+    }
+
+    // Open both underlying Realms asynchronously. Once they are both ready, all tasks added in the meantime are
+    // started. If the Realms failed to open correctly any error will be reported through the `Callback.onError` callback.
+    private void openRealms() {
```
bmeike: Single threaded?

realm/realm-library/src/objectServer/java/io/realm/PermissionManager.java
```diff
@@ -0,0 +1,388 @@
...
+                    permissionRealm = realm;
+                    checkIfRealmsAreOpened();
+                }
+
+                @Override
+                public void onError(Throwable exception) {
+                    synchronized (errorLock) {
+                        permissionRealmError = new ObjectServerError(ErrorCode.UNKNOWN, exception);
+                    }
+                    checkIfRealmsAreOpened();
```
bmeike: Since you already have the lock, it might be cheaper to just hold it.  ...unless you think it safer to release it.

realm/realm-library/src/objectServer/java/io/realm/PermissionManager.java
```diff
@@ -0,0 +1,388 @@
...
+
+        @Override
+        public void run() {
+            if (isCancelled()) { return; }
+            if (permissions != null || isClosed()) {
+                notifyCallbackWithResult();
+            } else {
+                synchronized (errorLock) {
+                    if (permissionRealmError != null) {
+                        callback.onError(permissionRealmError);
```
bmeike: This seems dangerous to me.  You are invoking the callback while holding the lock.  Would it be better to get a reference to the lock and then do the callback outside the synchronized block?

realm/realm-library/src/objectServer/java/io/realm/PermissionManager.java
```diff
@@ -0,0 +1,388 @@
...
+                    }
+                }
+            }
+        }
+
+        private void notifyCallbackWithResult() {
+            synchronized (errorLock) {
+                if (isClosed()) {
+                    IllegalStateException reason = new IllegalStateException("PermissionManager has been closed.");
+                    callback.onError(new ObjectServerError(ErrorCode.UNKNOWN, reason));
```
bmeike: same...  If the callback is called with the lock, that ought to be part of the documented contract.

realm/realm-library/src/objectServer/java/io/realm/SyncUser.java
```diff
@@ -61,14 +60,30 @@
...
+    }
+    private ThreadLocal<PermissionManager> permissionManager = new ThreadLocal<PermissionManager>() {
+        @Override
+        public PermissionManager get() {
+            PermissionManager pm = super.get();
+            if (pm == null || pm.isClosed()) {
+                pm = new PermissionManager(SyncUser.this);
+                set(pm);
+                return pm;
+            } else {
```
bmeike: I would like to see the smaller clause first.  ... in fact, I don't see why you don't just factor it out...

realm/realm-library/src/objectServer/java/io/realm/SyncUser.java
bmeike: Can I suggest making PermissionsManager an interface, in `io.realm`, and them moving most of this stuff to `io.realm.internal.somewhere`.  `io.realm` is already a swamp...

cmelchior: I  could, but what would the point be? It wouldn't it reduce the swamp that is `io.realm`? Originally I had everything in `io.realm.permissions()`, but then you would be forced to have a public constructor which wasn't supposed to be used (gief Java 9 modules)

realm/realm-library/src/androidTest/java/io/realm/rule/RunInLooperThread.java
cmelchior: Not sure what you mean?

realm/realm-library/src/objectServer/java/io/realm/Permission.java
Zhuinden: oh, May is a girl. Heh.

realm/realm-library/src/objectServer/java/io/realm/PermissionManager.java
```diff
@@ -0,0 +1,388 @@
...
+    private Handler handler = new Handler();
+    private final SyncConfiguration managementRealmConfig;
+    private final SyncConfiguration permissionRealmConfig;
+    private Realm permissionRealm;
+    private Realm managementRealm;
+
+    // Task list used to queue tasks until the underlying Realms are done opening (or fail doing so).
+    private Deque<PermissionManagerAsyncTask> delayedTasks = new LinkedList<>();
+
+    // List of tasks that are running. Used to keep strong references for listeners to work.
```
kneth: They are not running in the operating system meaning of the word, right? I don't have any better suggestions.

realm/realm-library/src/objectServer/java/io/realm/PermissionManager.java
```diff
@@ -0,0 +1,388 @@
...
+     * @param callback
+     * @return
+     */
+    public RealmAsyncTask getPermissionsAsync(final Callback<RealmResults<Permission>> callback) {
+        checkIfValidThread();
+        checkCallbackNotNull(callback);
+        return addTask(new PermissionsAsyncTask(callback));
+    }
+
+    // Queue the task if the underlying Realms are not ready yet, otherwise
```
kneth: Enqueue?

realm/realm-library/src/objectServer/java/io/realm/PermissionManager.java
```diff
@@ -0,0 +1,388 @@
...
+     * @return {@code true} if the PermissionManager is closed, {@code false} if it is still open.
+     */
+    public boolean isClosed() {
+        checkIfValidThread();
+        return closed;
+    }
+
+    @Override
+    protected void finalize() throws Throwable {
+        super.finalize();
```
kneth: Call last?

realm/realm-library/src/objectServer/java/io/realm/internal/permissions/PermissionModule.java
```diff
@@ -0,0 +1,25 @@
+/*
+ * Copyright 2016 Realm Inc.
```
kneth: 2017?

realm/realm-library/src/objectServer/java/io/realm/PermissionManager.java
```diff
@@ -0,0 +1,475 @@
...
+    }
+
+    /**
+     * FIXME Add Javadoc
+     */
+    @Override
+    public void close() {
+        checkIfValidThread();
+        // If Realms are still being opened, abort that task
+        if (openInProgress) {
```
beeender: just remove this check and the corresponding `else`. `managementRealmOpenTask` `permissionRealmOpenTask` may finished at the different time, so you could have `managementRealm != null` but `permissionRealm == null`

realm/realm-library/src/objectServer/java/io/realm/SyncUser.java
beeender: The whole `PermissionManager` is restricted to be used only on thread which has a looper. Would it be acceptable?

realm/realm-library/src/objectServer/java/io/realm/PermissionManager.java
```diff
@@ -0,0 +1,475 @@
...
+        this.user = user;
+        threadId = Thread.currentThread().getId();
+        managementRealmConfig = new SyncConfiguration.Builder(
+                user, getRealmUrl(RealmType.MANAGEMENT_REALM, user.getAuthenticationUrl()))
+                .errorHandler(new SyncSession.ErrorHandler() {
+                    @Override
+                    public void onError(SyncSession session, ObjectServerError error) {
+                        // FIXME: How to handle Client Reset?
+                        synchronized (errorLock) {
+                            managementRealmError = error;
```
beeender: what should user do with errors?

realm/realm-library/src/objectServer/java/io/realm/SyncUser.java
beeender: Another concern is this is not consistent with `Realm.getInstance()` which is based on a counter. (But actually I think `Realm.getInstance()` is the one should be changed ...) maybe we should just doc it here.

realm/realm-library/src/androidTest/java/io/realm/RunTestInLooperThreadLifeCycleTest.java
```diff
@@ -0,0 +1,102 @@
...
+ * Current order is:
+ * - @RunTestInLooperThread(before = <classRef>)
+ * - @Before()
+ * - @RunTestInLooperThread/@Test
+ * - @After : This is called when exiting the test method
+ * - looperThread.runAfterTest(Runnable) : This is called when the LooperTest either succeed for fails.
+ */
+
+@RunWith(AndroidJUnit4.class)
+public class RunTestInLooperThreadLifeCycleTest {
```
cmelchior: Will do

realm/realm-library/src/objectServer/java/io/realm/SyncUser.java
cmelchior: I don't think there is any way around requiring a looper thread for now. Waiting for specific events from the server is going to be really painful otherwise. At least for a v1 I think it is acceptable.

cmelchior: I see the problem with the PermissionManager lifecycle. I guess it would make sense to give it the same behavior as Realm instance counting. At least the behavior is consistent then.

realm/realm-library/src/androidTest/java/io/realm/RunTestInLooperThreadLifeCycleTest.java
cmelchior: #4740

realm/realm-library/src/objectServer/java/io/realm/PermissionManager.java
cmelchior: Yup, but that boolean will not change unless both Realms are either opened or failed. See `checkIfRealmsAreOpened()`

cmelchior: Oh wait. I see what you mean. Will fix

realm/realm-library/src/objectServer/java/io/realm/PermissionManager.java
cmelchior: Yes, this method is ever only be called on the thread that opened the PermissionManager

realm/realm-library/src/objectServer/java/io/realm/PermissionManager.java
```diff
@@ -39,6 +39,37 @@
  */
 public class PermissionManager implements Closeable {
 
+    // Reference counted cache equivalent to how Realm instances work.
```
kneth: Will this always be true? If yes, how do we ensure changes one place are reflected the other place?

realm/realm-library/src/syncIntegrationTest/java/io/realm/SyncedRealmTests.java
```diff
@@ -25,6 +25,7 @@
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
```
nhachicha: unused import?

realm/realm-library/src/objectServer/java/io/realm/PermissionManager.java
```diff
@@ -0,0 +1,517 @@
...
+                    synchronized (errorLock) {
+                        permissionRealmError = new ObjectServerError(ErrorCode.UNKNOWN, exception);
+                        checkIfRealmsAreOpened();
+                    }
+                }
+            });
+        }
+    }
+
+    private void checkIfRealmsAreOpened() {
```
nhachicha: maybe rename to something that reflects the `runDelayedTasks` behavior as well. ex `checkIfRealmsAreOpenedAndRunDelayedTasks` 🙃

realm/realm-library/src/objectServer/java/io/realm/PermissionManager.java
```diff
@@ -0,0 +1,517 @@
...
+            throw new IllegalStateException("PermissionManager was accessed from the wrong thread. It can only be " +
+                                            "accessed on the thread it was created on.");
+        }
+    }
+
+    /**
+     * FIXME Add Javadoc
+     */
+    @Override
+    public void close() {
```
nhachicha: what about `delayedTask` do we need to clear it ?

realm/realm-library/src/objectServer/java/io/realm/PermissionManager.java
```diff
@@ -0,0 +1,517 @@
...
+     */
+    public boolean isClosed() {
+        checkIfValidThread();
+        return closed;
+    }
+
+    @Override
+    protected void finalize() throws Throwable {
+        if (managementRealm != null || permissionRealm != null) {
+            RealmLog.error("PermissionManager was not correctly closed before being finalized.");
```
nhachicha: Should a warning suffice?

realm/realm-library/src/objectServer/java/io/realm/PermissionManager.java
```diff
@@ -0,0 +1,517 @@
...
+            close();
+        }
+        super.finalize();
+    }
+
+    // Creates the URL to the permission/management Realm based on the authentication URL.
+    private static String getRealmUrl(RealmType type, URL authUrl) {
+        String scheme = "realm";
+        if (authUrl.getProtocol().equalsIgnoreCase("https")) {
```
nhachicha: I get the idea, but the client may not want to activate TLS for sync. I don't have a suggestion though, so will see how the API evolve

realm/realm-library/src/objectServer/java/io/realm/PermissionManager.java
```diff
@@ -0,0 +1,517 @@
...
+        // Prevent permissions from being GC'ed until fully loaded.
+        private RealmResults<Permission> loadingPermissions;
+
+        GetPermissionsAsyncTask(PermissionManager permissionManager, Callback<RealmResults<Permission>> callback) {
+            super(permissionManager, callback);
+        }
+
+        @Override
+        public void run() {
+            if (checkAndReportInvalidState()) { return; }
```
nhachicha: a warning log could be helpful for debugging?

realm/realm-library/src/objectServer/java/io/realm/PermissionManager.java
```diff
@@ -0,0 +1,517 @@
...
+            if (permissions != null) {
+                // Permissions already loaded
+                notifyCallbackWithSuccess(permissions);
+            } else {
+                // Start loading permissions.
+                // TODO Right now multiple getPermission() calls will result in multiple
+                // queries being executed. The first one to return will the be one returned
```
nhachicha: > will the be

will be the

realm/realm-library/src/objectServer/java/io/realm/PermissionManager.java
```diff
@@ -0,0 +1,517 @@
...
+                    }
+                });
+            }
+        }
+    }
+
+    // Class encapsulating all async tasks exposed by the PermissionManager.
+    // All subclasses are responsible for removing themselves from the activeTaskList when done.
+    // Made package protected instead of private to facilitate testing
+    abstract static class PermissionManagerAsyncTask<T> implements RealmAsyncTask, Runnable {
```
nhachicha: the name might be redundant (already calling it as an inner class of `PermissionManager`) `AsyncTask` IMO should be clear enough (`new PermissionManager.AsyncTask`)

realm/realm-library/src/objectServer/java/io/realm/PermissionManager.java
```diff
@@ -0,0 +1,517 @@
...
+            if (isCancelled()) { return true; }
+            if (permissionManager.isClosed()) {
+                ObjectServerError error = new ObjectServerError(ErrorCode.UNKNOWN,
+                        new IllegalStateException("PermissionManager has been closed"));
+                notifyCallbackError(error);
+                return true;
+            }
+
+            // We are juggling two different Realms. If only one fail, expose that error directly.
+            // Otherwise try to sensible join the two error message before returning it to the user.
```
nhachicha: error messages

realm/realm-library/src/objectServer/java/io/realm/Permission.java
```diff
@@ -0,0 +1,154 @@
...
+import io.realm.RealmModel;
+import io.realm.RealmObject;
+import io.realm.annotations.Ignore;
+import io.realm.annotations.PrimaryKey;
+import io.realm.annotations.Required;
+
+
+/**
+ * FIXME Javadoc
+ * Public description of what a a Permission is and how it works
```
cmelchior: Will fix this when I add the Javadoc

realm/realm-library/src/syncIntegrationTest/java/io/realm/PermissionManagerTests.java
```diff
@@ -0,0 +1,313 @@
...
+
+    @Rule
+    public final TestSyncConfigurationFactory configurationFactory = new TestSyncConfigurationFactory();
+
+    private SyncUser user;
+
+    @Before
+    public void setUpTest() {
+        String username = UUID.randomUUID().toString();
+        RealmLog.error("Username %s", username);
```
cmelchior: Ups, should have been deleted.

realm/realm-library/src/syncIntegrationTest/java/io/realm/PermissionManagerTests.java
```diff
@@ -0,0 +1,313 @@
...
+
+    @Test
+    @RunTestInLooperThread
+    public void getPermissions_noLongerValidWhenPermissionManagerIsClosed() {
+        final PermissionManager pm = user.getPermissionManager();
+        looperThread.closeAfterTest(pm);
+        pm.getPermissions(new PermissionManager.Callback<RealmResults<Permission>>() {
+            @Override
+            public void onSuccess(RealmResults<Permission> permissions) {
+                RealmLog.error("OnSuccess");
```
cmelchior: Ups, should have been deleted.

realm/realm-library/src/objectServer/java/io/realm/PermissionManager.java
```diff
@@ -0,0 +1,520 @@
...
+
+        return task;
+    }
+
+    // Park the task until all underlying Realms are ready
+    private void delayTask(AsyncTask task) {
+        delayedTasks.add(task);
+    }
+
+    // Run any tasks that where delayed while the underlying Realms where being opened.
```
kneth: "where" -> "were"
"where being" -> "were"

realm/realm-library/src/objectServer/java/io/realm/PermissionManager.java
```diff
@@ -0,0 +1,520 @@
...
+
+    // Activate a task. All tasks are controlled by the Handler in order to make it asynchronous.
+    // PRECONDITION: Underlying Realms are no longer in the process of being opened.
+    private void activateTask(AsyncTask task) {
+        activeTasks.add(task);
+        handler.post(task);
+    }
+
+    // Open both underlying Realms asynchronously. Once they are both ready, all tasks added in the meantime are
+    // started. If the Realms failed to open correctly any error will be reported through the `Callback.onError` callback.
```
kneth: Move "If the Realms failed to open correctly" to the end of the sentence.

realm/realm-library/src/objectServer/java/io/realm/PermissionManager.java
```diff
@@ -0,0 +1,520 @@
...
+            managementRealm.close();
+        }
+        if (permissionRealm != null) {
+            permissionRealm.close();
+        }
+        closed = true;
+    }
+
+    /**
+     * Checks if this PermissionManager is closed or not. If it is closed, all methods will report back an error
```
kneth: Add `.` at the end

kneth:
@cmelchior Can we safely ignore the failing tests when doing a review?

realm/realm-library/src/objectServer/java/io/realm/PermissionManager.java
```diff
@@ -0,0 +1,520 @@
...
+            throw new IllegalArgumentException("Non-null 'callback' required.");
+        }
+    }
+
+    private boolean isReady() {
+        return managementRealm != null && permissionRealm != null;
+    }
+
+    private void checkIfValidThread() {
+        // Checks if we are in thread that created the PermissionManager.
```
kneth: "in" -> "on"

cmelchior:
No, the failing tests are a problem. I'm still trying to figure out why they happen.

realm/realm-library/src/syncIntegrationTest/java/io/realm/PermissionManagerTests.java
```diff
@@ -69,6 +70,12 @@ public void getPermissions_returnLoadedResults() {
             @Override
             public void onSuccess(RealmResults<Permission> permissions) {
                 assertTrue(permissions.isLoaded());
+                if (permissions.size() > 1) {
```
kneth: Just curious: when will this happen?

cmelchior: There appears to be a timing issue. We have both a `__management` and `__permission` Realm and there is a timing issue for when the entries are being created in the `__permission` Realm. My implementation of the logic is wrong, and right now I'm trying to wait for both the `__permission` and `__management` Realm to appear, but for reasons, I don't fully understand yet,  one of these entries are never created.

bmeike:
Lgtm

bmeike:
Lgtm

realm/realm-library/src/objectServer/java/io/realm/PermissionManager.java
```diff
@@ -0,0 +1,525 @@
...
+        }
+    };
+
+    /**
+     * Return a thread confined, reference counted instance of the PermissionManager.
+     *
+     * @param syncUser user to create the PermissionManager for.
+     * @return a thread confined PermissionManager instance for the provided user.
+     */
+    static synchronized PermissionManager getInstance(SyncUser syncUser) {
```
beeender: this doesn't have to be `synchronized`?

realm/realm-library/src/objectServer/java/io/realm/PermissionManager.java
```diff
@@ -0,0 +1,525 @@
...
+    public boolean isClosed() {
+        checkIfValidThread();
+        return closed;
+    }
+
+    @Override
+    protected void finalize() throws Throwable {
+        if (managementRealm != null || permissionRealm != null) {
+            RealmLog.warn("PermissionManager was not correctly closed before being finalized.");
+            close();
```
beeender: `finalize()` will run on the finalizer thread, calling `close()` will just crash because of the thread checking. Just remove the call to `close()`

cmelchior: Good catch. I did fix this earlier, but apparently forgot to commit it

PR #4663, opened 2017-05-18
Update android gradle plugin to 3.0.0

zaki50:
https://developer.android.com/studio/preview/features/new-android-plugin.html
https://developer.android.com/studio/preview/features/new-android-plugin-migration.html

When you switch android gradle plugin from 2.x to 3.x, you need to stop running gradle daemon by executing `./gradlew --stop`.


cmelchior:
It also appears that `annotationProcessor` will automatically delegate to `kapt` if needed. That will reduce some complexity which is nice. We also need to completely remove `apt` since that won't work at all.

zaki50:
@cmelchior nice!

zaki50:
@cmelchior You mean we would drop the support for older Android Cradle Plugin?

cmelchior:
I guess we need to figure out how to go about it. If we could handle both versions inside the same plugin that would obviously be preferable at least during a transition period, but I'm not sure how we can detect the difference. I assume we will have to rely on reflection

zaki50:
I talked with Xav at the Office Hour in Google I/O and he said that he will add a feature to solve this compatibility issue.

cmelchior:
Woop woop

Zhuinden:
That's nice of them

zaki50:
Now realm plugin supports both old and new android gradle plugins.

zaki50:
https://github.com/robolectric/robolectric/issues/3169 seems to cause the test failure.

zaki50:
I confirmed that disabling aapt2 solved the failure, but I don't disable it for a while to wait the fix of upstream projects.

zaki50:
Disabled `aapt2` in example project to start merge process of this PR.

@realm/java Could you review this?

gradle-plugin/src/main/groovy/io/realm/gradle/Realm.groovy
```diff
@@ -40,7 +41,9 @@ class Realm implements Plugin<Project> {
         }
 
         def syncEnabledDefault = false
-        project.extensions.create('realm', RealmPluginExtension, project, syncEnabledDefault)
+        def dependencyConfigurationName = getDependencyConfigurationName(project)
+        project.extensions.create('realm', RealmPluginExtension,
```
cmelchior: Couldn't this fit into one line?

gradle-plugin/src/main/groovy/io/realm/gradle/Realm.groovy
```diff
@@ -80,6 +83,17 @@ class Realm implements Plugin<Project> {
         }
     }
 
+    private static String getDependencyConfigurationName(Project project) {
```
cmelchior: Perhaps add a comment here that `newDependencyName` is from `3.0.0-*` ?

realm-transformer/src/main/groovy/io/realm/transformer/RealmTransformer.groovy
```diff
@@ -66,6 +66,7 @@ class RealmTransformer extends Transform {
 
     @Override
     Set<Scope> getReferencedScopes() {
+        // Scope.PROJECT_LOCAL_DEPS and Scope.SUB_PROJECTS_LOCAL_DEPS is only for the compatibility with AGP 1.x, 2.x
```
cmelchior: only for the -> only for

zaki50:
@cmelchior I tested ithis with old plugin  before, but will test it again to make sure.

beeender:
I got `Error:This Gradle plugin requires Studio 3.0 minimum` on AS 2.3.3 . Is it OK for everyone to update the AS? @realm/java I am OK to update it :P

zaki50:
@beeender Ah, good catch!

Recently I always use AS 3.0 Canary and didn't see any major problems. @realm/java

cmelchior:
I'm already using 3.0.0-alpha4 so... :)

zaki50:
@cmelchior I tested the local build of this branch with Java and Kotlin projects that were using android gradle plugin 2.3.3. Worked fine!

PR #4665, opened 2017-05-18
schemaVersion no longer required for Synced Realms

cmelchior:
Due to miscommunication, the Java side mistakenly added a requirement that the schema version should be bumped for schema upgrades. This is not the case.

TODO:
- [ ] Changelog
- [ ] Tests
- [x] Remove version check
- [ ] Verify if the sharedrealm->update_schema set the schema version internally


cmelchior:
Based on wrong branch. Closing

PR #4666, opened 2017-05-18
SchemaVersion for synced Realms where required by mistake

### LINKED ISSUE #4618: Allow removing fields from Synced schemas
Cocoa added support for removing properties from a model class again. It will not remove the field, but just hide it. We should implement something similar.

See https://realm.io/docs/swift/latest/#syncing-migrations

### LINKED ISSUE #4658: Upgrading Synced Realms breaks existing clients
### Problem
Changing the Realm Schema (e.g. Adding a new property on an object) breaks existing clients.  

### Steps to reproduce using [RealmTasks](https://github.com/realm-demos/realm-tasks) Android project.
1. Clone RealmTasks
1. Open the Android Project
1. Open `UserManager.java` and add `.schemaVersion(1)` to the SyncConfiguration.
1. Install RealmTasks onto 2 device simulators. (I used a Nexus 6 and 6P)
1. Exit the app on both simulators.
1. Add a property to the `Task.java` Realm object.
1. Open `UserManager.java` and change the schema version to 2, i.e. `.schemaVersion(2)`.
1. Build and run on the 1st simulator
1. Just launch the old version of the app on the 2nd simulator.  (do not install the new version)

### Actual Result
The updated client (1st simulator) - Runs with the new schema.
The old client (2nd simulator) - Crashes after a sync happens with ROS and then every time you launch the app with the exception `IllegalArgumentException: The schema was changed but the schema version was not updated.  The configured schema version (1) must be greater than the version in the Realm file (2) in order to update the schema`

### Expected Result
The updated client (1st simulator) - Runs with the new schema.
The old client (2nd simulator) - Runs with the old schema just fine.

Would also be nice not to need to supply the schema version, but failure to do so prevents the upgrade.  You wouldn't be able to install the new version on the 1st simulator.  This also might be a bug in the Java binding, because the other platforms don't require schema version numbers.

cmelchior:
Fixes #4658 
Fixes #4618 

Due to miscommunication, the Java side mistakenly added a requirement that the schema version should be bumped for schema upgrades. This is not the case.

TODO:
- [x] Changelog
- [x] Unittests
- [x] Remove version check
- [x] Verify if the sharedrealm->update_schema set the schema version internally


cmelchior:
Ready for review @realm/java 

Note, extending the unit tests showcased another error regarding indexes. Right now our annotation processor add code that strictly require them to be present or not depending on your local schema, but synced Realms allow you to remove/add them at will. 

There are unit tests showing the current behavior. I'll create another issue to track the Index problems.

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -501,21 +501,9 @@ private static void initializeSyncedRealm(Realm realm) {
...
-                                        "The configured schema version (%d) must be greater than the version " +
-                                        " in the Realm file (%d) in order to update the schema.",
-                                newVersion, currentVersion));
-                    }
-                    realm.sharedRealm.updateSchema(schema.getNativePtr(), newVersion);
-                    // The OS currently does not handle setting the schema version. We have to do it manually.
-                    realm.setVersion(newVersion);
-                    commitChanges = true;
-                }
+                // Object Store handle all update logic
```
beeender: handle -> handles

realm/realm-library/src/androidTestObjectServer/java/io/realm/SyncedRealmMigrationTests.java
```diff
@@ -56,4 +64,186 @@ public void migrateRealm_syncConfigurationThrows() {
...
+        schema.create(className)
+                .addField(IndexedFields.FIELD_INDEXED_STRING, String.class) // No index
+                .addField(IndexedFields.FIELD_NON_INDEXED_STRING, String.class);
+        dynamicRealm.setVersion(43);
+        dynamicRealm.commitTransaction();
+        dynamicRealm.close();
+
+        try {
+            Realm realm = Realm.getInstance(config); // Opening at different schema version (42) should rebuild indexes
+            fail();
```
beeender: I cannot tell the difference between this test and `sameSchemaVersion_doNotRebuildIndexes`

realm/realm-library/src/androidTestObjectServer/java/io/realm/SyncConfigurationTests.java
```diff
@@ -477,61 +477,6 @@ public void schemaVersion_throwsIfLessThanCurrentVersion() throws IOException {
...
-        String url = "realm://ros.realm.io/~/default";
-        @SuppressWarnings("unchecked")
-        SyncConfiguration config = configFactory.createSyncConfigurationBuilder(user, url)
-                .schema(AllJavaTypes.class, StringOnly.class)
-                .name("schemaversion_v1.realm")
-                .schemaVersion(1)
-                .build();
-
-        // Add v1 of the Realm to the file system. v1 is missing the class `StringOnly`
-        configFactory.copyRealmFromAssets(context, "schemaversion_v1.realm", config);
```
beeender: `schemaversion_v1.realm` this file can be removed?

cmelchior: Yes 👍

realm/realm-library/src/androidTestObjectServer/java/io/realm/SyncedRealmMigrationTests.java
cmelchior: The difference is in schema versions, the first has 42 on disk and opens a typed Realm with 42. This second one has 43 on disk and opens with 42.

beeender: But the results are the same? Both are supposed to throw `RealmMigrationNeededException`? what should user do with this case?

cmelchior: Nothing, they will be in deep trouble, because this might happen at runtime. It is quite a big bug we should fix as soon as possible, but it is outside the scope of the PR (and luckily no-one hit it yet).

To work around it right now you need to define all indexes up front just like primary keys and they cannot be changed later.

cmelchior: Actually, it looks like a bug in Object Store. I added a new test and it seems that indexes are being correctly rebuilt if there are other changes as well, but only index changes will not be processed.

realm/realm-library/src/androidTestObjectServer/java/io/realm/SyncedRealmMigrationTests.java
```diff
@@ -56,4 +64,216 @@ public void migrateRealm_syncConfigurationThrows() {
...
+        dynamicRealm.commitTransaction();
+        dynamicRealm.close();
+
+        try {
+            Realm realm = Realm.getInstance(config); // Opening at same schema version (42) will not rebuild indexes
+            fail();
+        } catch (RealmMigrationNeededException ignored) {
+        }
+
+// FIXME: This is the intended behaviour
```
bmunkholm: What does it require to fix this.

cmelchior: That is a bit unclear at the moment and depends on the semantics that are unclear as described in : https://github.com/realm/realm-java/issues/4672

But it looks like there might be bugs in both Object Store and the annotation processor.

beeender:
@cmelchior ship it? :shipit:

PR #4674, opened 2017-05-19
Let Object Store handle table creations

beeender:
This tries to progressively move more things about schemas to Object
Store.

First the concept of Schema in Object Store is not the same as what we
have in Java. It is very much just a schema information holder and won't
take care of the schema modifications. That says it is more like the
ColumnIndices cache in the Java binding.
So this commit try to:
- Instead of inheriting from the RealmSchema, change the
OsRealmSchema/OsRealmObjectSchema to OsSchemaInfo/OsObjectSchemaInfo.
They behave as a simple Java wrapper to the relevant OS objects.
- Add functions to Proxy classes which will create its own
OsObjectSchemaInfo and those info can be used to create a OsSchemaInfo
through the mediator.
- Call `SharedRealm.updateSchema` with the OsSchemaInfo which is got
from the proxy interface to do table initialization.
 - This will also fix a minor bug we have before, All tables are created
 even if the class is not in the module.
- Migration is still handled in the old way, and it will be solved in
the future, to let Object Store handle it.
- ColumnIndices are still kept for now, but it should be replaced by the
OsSchemaInfo/OsObjectSchemaInfo in the future.


Zhuinden:
> Migration is still handled in the old way, and it will be solved in the future, to let Object Store handle it.

Hmm... Multiprocess  migration comes with OS migration, no?

beeender:
@realm/java This is ready for review. See design doc https://docs.google.com/document/d/18Tr5oiKxH2c7DN7h48BsYjk2axdnf9eesTPFP5kjZ-U/edit#

realm/realm-library/src/main/cpp/io_realm_internal_OsObjectSchemaInfo.cpp
```diff
@@ -15,15 +15,21 @@
...
+#include "io_realm_internal_OsObjectSchemaInfo.h"
 
 #include <object-store/src/object_schema.hpp>
 #include <object-store/src/property.hpp>
 
 #include "util.hpp"
 using namespace realm;
 
-JNIEXPORT jlong JNICALL Java_io_realm_OsRealmObjectSchema_nativeCreateRealmObjectSchema(JNIEnv* env, jclass,
+static void finalize_object_schema(jlong ptr)
```
kneth: Nice move :smile:

realm/realm-library/src/main/cpp/io_realm_internal_Property.cpp
```diff
@@ -63,7 +63,7 @@ JNIEXPORT jlong JNICALL Java_io_realm_Property_nativeCreateProperty__Ljava_lang_
     return 0;
 }
 
-JNIEXPORT void JNICALL Java_io_realm_Property_nativeClose(JNIEnv* env, jclass, jlong property_ptr)
+JNIEXPORT void JNICALL Java_io_realm_internal_Property_nativeClose(JNIEnv* env, jclass, jlong property_ptr)
```
kneth: No point in migrating to phantom references and the native object abstraction?

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
```diff
@@ -597,13 +597,27 @@ JNIEXPORT jboolean JNICALL Java_io_realm_internal_SharedRealm_nativeCompact(JNIE
...
         auto& shared_realm = *(reinterpret_cast<SharedRealm*>(shared_realm_ptr));
         auto* schema = reinterpret_cast<Schema*>(schema_ptr);
-        shared_realm->update_schema(*schema, static_cast<uint64_t>(version), nullptr, true);
+
+        if (migration_callback == nullptr) {
+            shared_realm->update_schema(std::move(*schema), static_cast<uint64_t>(version), nullptr, true);
+        }
+        else {
+            Realm::MigrationFunction callback = [=](SharedRealm old_realm, SharedRealm realm, Schema &) {
+                static JavaMethod do_migrate_method(env, migration_callback, "doMigrate", "(JJ)V");
```
kneth: Why can't we call `migrate()` from the `RealmMigration` object directly?

realm/realm-library/src/main/cpp/io_realm_internal_Property.cpp
beeender: That is more like for the future, the `RealmObject` will hold a `OsObjectSchemaInfo` because of lots of function calls from Object Store take a `ObjectSchema` as an input.
Also, Proxy hold a static `OsObjectSchemaInfo` which in theory doesn't have to be closed, but phantom ref makes it literally correct.
BTW, by following the phantom ref pattern makes our memory management life much easier.

beeender: Uh, sorry, i totally misunderstood your comments!! I can move it to phantom refs in this PR as well!! Surely can do!!

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
beeender: I probably should remove this and put it into the next PR, this PR doesn't handle migration :)

CHANGELOG.md
```diff
@@ -15,7 +17,8 @@
 
 ### Internal
 
-* Factor out internal interface ManagedObject
+* Factor out internal interface ManagedObject.
+* Use Object Store to do table initialization.
```
kneth: Maybe it is an enhancement but object store will throw an exception with a detailed with of everything which needs to migrated.

realm/realm-library/src/main/cpp/io_realm_internal_OsObjectSchemaInfo.cpp
```diff
@@ -37,18 +43,14 @@ JNIEXPORT jlong JNICALL Java_io_realm_OsRealmObjectSchema_nativeCreateRealmObjec
...
+JNIEXPORT jlong JNICALL Java_io_realm_internal_OsObjectSchemaInfo_nativeGetFinalizerPtr(JNIEnv*, jclass)
 {
-    TR_ENTER_PTR(native_ptr)
-    try {
-        ObjectSchema* object_schema = reinterpret_cast<ObjectSchema*>(native_ptr);
-        delete object_schema;
-    }
-    CATCH_STD()
+    TR_ENTER()
+    return reinterpret_cast<jlong>(&finalize_object_schema);
```
cmelchior: I don't understand this?  Why return here?

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -491,78 +501,6 @@ private static void initializeRealm(Realm realm) {
         }
     }
 
-    // Everything in this method needs to be behind a transaction lock
-    // to prevent multi-process interaction while the Realm is initialized.
-    private static void initializeSyncedRealm(Realm realm) {
```
cmelchior: 👏 💃

realm/realm-library/src/main/java/io/realm/internal/OsObjectSchemaInfo.java
```diff
@@ -0,0 +1,99 @@
...
+
+
+import io.realm.RealmFieldType;
+
+/**
+ * Java wrapper for Object Store ObjectSchema.
+ *
+ * @see OsSchemaInfo
+ */
+public class OsObjectSchemaInfo implements NativeObject {
```
cmelchior: This class is generally considered immutable right? Would it make sense to use the Builder pattern instead to create instances of it, then we only need to expose the getter methods afterwards?

CHANGELOG.md
Zhuinden: If detailed message appears as java exception message instead of the "expected field count 15 but was 14" then that is super-nice

realm/realm-library/src/main/java/io/realm/internal/OsObjectSchemaInfo.java
```diff
@@ -0,0 +1,99 @@
...
+
+    /**
+     * Adds a linked property to this {@code OsObjectSchema}.
+     *
+     * @param name the name of the linked property.
+     * @param type the type of the linked property.
+     * @return this {@code OsObjectSchemaInfo}.
+     */
+    public OsObjectSchemaInfo add(String name, RealmFieldType type, String linkedClassName) {
+        final Property property = new Property(name, type, linkedClassName);
```
cmelchior: Should we check that you only use OBJECT/LIST here?

realm/realm-library/src/main/java/io/realm/internal/OsObjectSchemaInfo.java
```diff
@@ -0,0 +1,99 @@
...
+     *
+     * @param name the name of the property.
+     * @param type the type of the property.
+     * @param primary set to true if this property is the primary key.
+     * @param indexed set to true if this property needs an index.
+     * @param required set to false if this property is not nullable.
+     * @return this {@code OsObjectSchemaInfo}.
+     */
+    public OsObjectSchemaInfo add(String name, RealmFieldType type, boolean primary, boolean indexed, boolean required) {
+        final Property property = new Property(name, type, primary, indexed, required);
```
cmelchior: Should we check that you dont use OBJECT/LIST here?

realm/realm-library/src/main/java/io/realm/internal/OsSchemaInfo.java
```diff
@@ -0,0 +1,61 @@
...
+     *
+     * @param objectSchemaInfoList all the object schemas should be contained in this {@code OsObjectSchemaInfo}.
+     */
+    public OsSchemaInfo(List<OsObjectSchemaInfo> objectSchemaInfoList) {
+        long[] schemaNativePointers = new long[objectSchemaInfoList.size()];
+        for (int i = 0; i < objectSchemaInfoList.size(); i++) {
+            schemaNativePointers[i] = objectSchemaInfoList.get(i).getNativePtr();
+        }
+        this.nativePtr = nativeCreateFromList(schemaNativePointers);
+        NativeContext.dummyContext.addReference(this);
```
cmelchior: Not sure why this is called `dummyContext`. It seems a rather weird name, but probably outside the scope of this PR

realm/realm-library/src/main/cpp/io_realm_internal_OsObjectSchemaInfo.cpp
kneth: You need to keep track of the finalizer (or what ever you wish to call it). The function pointer is stored in `NativeObject` instance.

CHANGELOG.md
kneth: I think you will get a list of all the changes.

realm/realm-library/src/main/java/io/realm/internal/OsSchemaInfo.java
```diff
@@ -0,0 +1,61 @@
...
+import java.util.List;
+
+/**
+ * Java wrapper for the Object Store Schema object.
+ * <p>
+ * When it is created from java binding, it is used for initializing/validating the schemas through Object Store. It
+ * won't contain the column indices information.
+ * <p>
+ * When this is get from the Object Store {@code SharedRealm} instance, this represents the real schema of the Realm
+ * file. It will contain the all the schema information as well as the information about the column indices.
```
nhachicha: > It will contain all the schema

CHANGELOG.md
beeender: The information about migration errors is not included in this PR. This will only contain the detailed information about the schema initialization issues (RealmModule issues) such as you defined ad `RealmObjectB` as a field of `RealmObjectA`, But `RealmObjectB` is not part of the `RealmModule`.

realm/realm-annotations-processor/src/test/resources/io/realm/AllTypesRealmProxy.java
```diff
@@ -398,27 +399,22 @@ protected final void copy(ColumnInfo rawSrc, ColumnInfo rawDst) {
...
-            realmObjectSchema.add("columnObject", RealmFieldType.OBJECT, realmSchema.get("AllTypes"));
-            if (!realmSchema.contains("AllTypes")) {
-                AllTypesRealmProxy.createRealmObjectSchema(realmSchema);
-            }
-            realmObjectSchema.add("columnRealmList", RealmFieldType.LIST, realmSchema.get("AllTypes"));
-            return realmObjectSchema;
-        }
-        return realmSchema.get("AllTypes");
+    private static OsObjectSchemaInfo createExpectedObjectSchemaInfo() {
+        OsObjectSchemaInfo info = new OsObjectSchemaInfo("AllTypes");;
```
beeender: your eyes are quite sensitive to `;` !!!!! 👍

realm/realm-annotations-processor/src/test/resources/io/realm/AllTypesRealmProxy.java
```diff
@@ -398,27 +399,22 @@ protected final void copy(ColumnInfo rawSrc, ColumnInfo rawDst) {
...
+        info.add("columnString", RealmFieldType.STRING, Property.PRIMARY_KEY, Property.INDEXED, !Property.REQUIRED);
+        info.add("columnLong", RealmFieldType.INTEGER, !Property.PRIMARY_KEY, !Property.INDEXED, Property.REQUIRED);
+        info.add("columnFloat", RealmFieldType.FLOAT, !Property.PRIMARY_KEY, !Property.INDEXED, Property.REQUIRED);
+        info.add("columnDouble", RealmFieldType.DOUBLE, !Property.PRIMARY_KEY, !Property.INDEXED, Property.REQUIRED);
+        info.add("columnBoolean", RealmFieldType.BOOLEAN, !Property.PRIMARY_KEY, !Property.INDEXED, Property.REQUIRED);
+        info.add("columnDate", RealmFieldType.DATE, !Property.PRIMARY_KEY, !Property.INDEXED, Property.REQUIRED);
+        info.add("columnBinary", RealmFieldType.BINARY, !Property.PRIMARY_KEY, !Property.INDEXED, Property.REQUIRED);
+        info.add("columnObject", RealmFieldType.OBJECT, "AllTypes");
```
beeender: sure, can do!

Zhuinden:
> How is the column caches updates when a sync client changes the schema?

This question does look important, after all, the schema cache needs to be updated if another process makes a modification to the schema.

Is there a way to listen for schema changes with OS schema?

realm/realm-library/src/main/java/io/realm/internal/OsSchemaInfo.java
beeender:
```
    // Dummy context which will be used by native objects which's destructors are always thread safe.
    static final NativeContext dummyContext = new NativeContext();
```

Do you have a better name in mind?

cmelchior: `threadSafeContext` ?

beeender:
@Zhuinden yes, see https://github.com/realm/realm-object-store/blob/a0f0d274edbaa77f633c31f956af3314961fffd2/src/shared_realm.cpp#L459

realm/realm-library/src/main/java/io/realm/internal/OsSchemaInfo.java
beeender: will do in another PR

beeender:
@realm/java @bmeike All comments addressed.

realm/realm-library/src/main/java/io/realm/internal/OsObjectSchemaInfo.java
```diff
@@ -16,14 +16,72 @@
...
+         * Adds a property to this builder.
+         *
+         * @param name the name of the property.
+         * @param type the type of the property.
+         * @param primary set to true if this property is the primary key.
+         * @param indexed set to true if this property needs an index.
+         * @param required set to false if this property is not nullable.
+         * @return this {@code OsObjectSchemaInfo}.
+         */
+        public Builder addProperty(String name, RealmFieldType type, boolean primary, boolean indexed,
```
cmelchior: primary -> primaryKey

Zhuinden: How about `isPrimary` and `isIndexed`?

realm/realm-library/src/main/java/io/realm/internal/OsObjectSchemaInfo.java
```diff
@@ -47,14 +47,14 @@ public Builder(String className) {
...
+         * @param isPrimaryKey set to true if this property is the primary key.
+         * @param isIndexed set to true if this property needs an index.
          * @param required set to false if this property is not nullable.
          * @return this {@code OsObjectSchemaInfo}.
          */
-        public Builder addProperty(String name, RealmFieldType type, boolean primary, boolean indexed,
+        public Builder addProperty(String name, RealmFieldType type, boolean isPrimaryKey, boolean isIndexed,
                                    boolean required) {
-            final Property property = new Property(name, type, primary, indexed, required);
+            final Property property = new Property(name, type, isPrimaryKey, isIndexed, required);
```
Zhuinden: welp, you'll be annoyed with me for this, but `isRequired` would make it consistent with this change.

Sorry!

realm/realm-annotations-processor/src/test/resources/io/realm/NullTypesRealmProxy.java
```diff
@@ -820,36 +821,34 @@ protected final void copy(ColumnInfo rawSrc, ColumnInfo rawDst) {
...
-                NullTypesRealmProxy.createRealmObjectSchema(realmSchema);
-            }
-            realmObjectSchema.add("fieldObjectNull", RealmFieldType.OBJECT, realmSchema.get("NullTypes"));
-            return realmObjectSchema;
-        }
-        return realmSchema.get("NullTypes");
+    private static OsObjectSchemaInfo createExpectedObjectSchemaInfo() {
+        OsObjectSchemaInfo.Builder builder = new OsObjectSchemaInfo.Builder("NullTypes");
+        builder.addProperty("fieldStringNotNull", RealmFieldType.STRING, !Property.PRIMARY_KEY, !Property.INDEXED, Property.REQUIRED);
+        builder.addProperty("fieldStringNull", RealmFieldType.STRING, !Property.PRIMARY_KEY, !Property.INDEXED, !Property .REQUIRED);
```
zaki50: @beeender Where did the space right before `.REQUIRED` come from?
Did you edit this test code by hand?

zaki50:
This PR also enable us to deprecate `RealmSchema.close()` and `RealmObjectSchema.close()`.
Those should not in our API since there are no case that users can call those explicitly.

realm/realm-annotations-processor/src/test/resources/io/realm/NullTypesRealmProxy.java
beeender: Opps, the editor split the long line into two automatically. Then I did a vim `J`. That caused the problem.

realm/realm-library/src/main/java/io/realm/internal/OsObjectSchemaInfo.java
beeender: Haha, OK. I guess I am going to hate you 🗡

realm/realm-library/src/main/cpp/io_realm_internal_OsSchemaInfo.cpp
```diff
@@ -42,9 +48,8 @@ JNIEXPORT jlong JNICALL Java_io_realm_OsRealmSchema_nativeCreateFromList(JNIEnv*
     return 0;
```
beeender: It actually does! Will change it.

Zhuinden:
Is https://github.com/realm/realm-java/pull/3066 superceded by this one?

beeender:
@Zhuinden yes!

beeender:
@bmeike , I need this change now, so i will just merge it. Let me know if there is any thing i need to fix for it, i will do it afterwards.

PR #4683, opened 2017-05-22
check the thread by using Thread instance instead of thread id

zaki50:
Thread instance itself is enough and a bit faster?

### DIFF
--- a/realm/realm-annotations-processor/src/test/resources/io/realm/SimpleRealmProxy.java
+++ b/realm/realm-annotations-processor/src/test/resources/io/realm/SimpleRealmProxy.java
@@ -263,7 +263,7 @@ public class SimpleRealmProxy extends some.test.Simple
     }
 
     public static some.test.Simple copyOrUpdate(Realm realm, some.test.Simple object, boolean update, Map<RealmModel,RealmObjectProxy> cache) {
-        if (object instanceof RealmObjectProxy && ((RealmObjectProxy) object).realmGet$proxyState().getRealm$realm() != null && ((RealmObjectProxy) object).realmGet$proxyState().getRealm$realm().threadId != realm.threadId) {
+        if (object instanceof RealmObjectProxy && ((RealmObjectProxy) object).realmGet$proxyState().getRealm$realm() != null && ((RealmObjectProxy) object).realmGet$proxyState().getRealm$realm().creatorThread != realm.creatorThread) {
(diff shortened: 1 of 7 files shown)


realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -67,7 +67,7 @@
     // Thread pool for all async operations (Query & transaction)
     static final RealmThreadPoolExecutor asyncTaskExecutor = RealmThreadPoolExecutor.newDefaultExecutor();
 
-    final long threadId;
```
bmeike: While I do believe this change is probably safe under the circumstances, it seems to me to increase the fragility of the code.  There is no reason to hold a reference to that gigantic, extremely heavy-weight Thread object.  Unless there is a really good reason for doing this, It seems safer to not to.

realm/realm-annotations-processor/src/test/resources/io/realm/SimpleRealmProxy.java
```diff
@@ -263,7 +263,7 @@ public static String getTableName() {
     }
 
     public static some.test.Simple copyOrUpdate(Realm realm, some.test.Simple object, boolean update, Map<RealmModel,RealmObjectProxy> cache) {
-        if (object instanceof RealmObjectProxy && ((RealmObjectProxy) object).realmGet$proxyState().getRealm$realm() != null && ((RealmObjectProxy) object).realmGet$proxyState().getRealm$realm().threadId != realm.threadId) {
+        if (object instanceof RealmObjectProxy && ((RealmObjectProxy) object).realmGet$proxyState().getRealm$realm() != null && ((RealmObjectProxy) object).realmGet$proxyState().getRealm$realm().creatorThread != realm.creatorThread) {
```
bmeike: I cannot think of any reason that this won't work (... and I *did* try!).  None the less, it is just weird enough (comparing objects with == instead of .equals) that I would be inclined to put it into a small method, and comment the method.  Better yet, unless there is a *really* good reason for doing it, let's just use .equals.

cmelchior: If you do that I think you lose the entire point of this refactor since you would introduce the method call back that was just removed?

zaki50:
I don't have strong opinion.
The only reason I tried to make this PR is that Google is checking thread validity in this way.

zaki50:
I'll close this PR.
I don't think this PR improves performance much.
 We should take natural code.

PR #4684, opened 2017-05-22
Kotlin Extension Library

### LINKED ISSUE #4092: `.in()` not working with Kotlin
Hello, the method `.in(fieldName, [])` is not working with Kotlin because the language has this parameter, common used on the control flow of data (loop, when).

For example, this code doesn't compile:

```kotlin
val counts = arrayOf(0,3,5,10)
return realm.where(User::class.java).in("count", counts).findAll()
```

You must add `.'in'()` to use this method. This is not a big problem but can generate problems in some point of the future.

### LINKED ISSUE #4701: Request for community comments: Better Kotlin support
With an awesome I/O17 being over, one of the big announcements were official support for [Kotlin](https://blog.jetbrains.com/kotlin/2017/05/kotlin-on-android-now-official/).

Realm is already fully interoperable with Kotlin and we have a [Kotlin example here](https://github.com/realm/realm-java/tree/master/examples/kotlinExample) and [here](https://github.com/realm-demos/realm-surveys/tree/master/Kotlin) that demonstrates how it works.

But we want to make that support even better and In order to do that, we are looking for feedback on what you would most like to see changed so Realm works even better with Kotlin

We have a few ideas ourselves:

1) Add extension functions for all classes with methods that currently accept `Class<? extends RealmModel>` so `KClass` can be used directly, e.g. `realm.where(Person::class)` instead of `realm.where(Person::class.java)`.

2) **(DONE)** Detect nullability automatically in model classes. Right now you are required to use the Realm  `@Required` annotation, but we can add support for JetBrains `@NotNull` as well which is how the Kotlin bytecode indicate not-null values. Example `var name : String` instead of of `@Required var name : String`

3) **(DONE)** Annotate our public API with JSR305 annotations in a [similar way to Okio/OKHttp](https://medium.com/square-corner-blog/rolling-out-nullable-42dd823fbd89). This means that Realm will expose correctly the nullability of return values and input paramaters. For example: `val result : RealmResults<Person> = realm.where(Person::class).findAll()` instead of `val results : RealmResults<Person>? = realm.where(Person::class).findAll()`. Also being tracked here: https://github.com/realm/realm-java/issues/4643

 But we are very interested what problems _you_ are running into when using Realm from Kotlin.

cmelchior:
Closes #4092
Closes #4701

This PR contains a new dependency `realm-android-kotlin-extensions` that is automatically included in all projects including Kotlin. The primary purpose to provide extension functions for methods so they become more Kotlin friendly.

TODO 
- [x] Figure out how and what kind of docs we should distribute. KDoc or Javadoc? KDoc seems to work fine since IntelliJ apparently uses the sources for the inline docs. CMD + F1 is showing correct and jumping to the definition also has the docs as expected.
- [x] Fix wrong classes included in KDoc: Currently not possible https://youtrack.jetbrains.com/issue/KT-11333
- [x] Test upload to OJO
- [x] Test upload to Bintray
- [x] Correctly report unit test failures
- [x] Cleanup
- [x] KDoc for all methods
- [x] Fix unit tests
- [x] Add unit tests for all extension methods
- [ ] Update website docs

Usage of these extensions can be seen here: https://github.com/realm/realm-java/pull/4684/files#diff-a9bdfddc109df70caf521cb459b12d15

---- 

OUTDATED

This PR is just a proof of concept for extending our current Kotlin support and is not intended to be merged as it is, but more like a starting point for a discussion. With Kotlin now being an official language for Android, we should explore the options that don't require rewriting everything in Kotlin ;)

This PR outlines an approach where we create a new `io.realm:realm-android-kotlin-extension` artifact that our Gradle plugin can automatically include if Kotlin is used, just like we detect when to use `kapt` today.

The question then remains what to put into such a plugin. Low hanging fruits would be:
- Add extension functions for all methods taking `Class` and allow `KClass` instead.
- Consider adding support for referencing field names directly for type safe queries. Will not work for chained queries and require a dependency for `kotlin-reflect` :/
- Add support for Kotlins null/non-null types. Cannot be part of the plugin, but must live in the annotation processor. It just require us detecting `org.jetbrains.annotations.NotNull` as a substitute for `@Required`.
- Other ideas?


realm/kotlin-extensions/build.gradle
```diff
@@ -0,0 +1,49 @@
+apply plugin: 'com.android.library'
+import io.realm.transformer.RealmTransformer
+android.registerTransform(new RealmTransformer())
+apply plugin: 'kotlin-android'
```
zaki50: It's safer to apply `kotlin-android` plugin before registering out transformer.

kneth:
If we support referencing field names in queries, operator overloading might be a nice thing: `realm.where(AllKotlinTypes::class).requiredString == "FooBar".findAll()`.

realm/realm-library/build.gradle
```diff
@@ -74,7 +74,8 @@ android {
 
     buildTypes {
         debug {
-            testCoverageEnabled = true
+            // FIXME: If enabled, crashes with https://issuetracker.google.com/issues/37116868
+            testCoverageEnabled = false
```
beeender: when it is enabled, debugger cannot show local vars which is another bug ... so disable this is 👍

realm/kotlin-extensions/build.gradle
Zhuinden: Typically people needed to use

```
apply plugin: 'kotlin-android'
apply plugin: 'kotlin-kapt'
apply plugin: 'realm-android'
```

in that order

realm/gradle.properties
```diff
@@ -1 +1,4 @@
 org.gradle.jvmargs=-Xms512m -Xmx2048m
+kotlin.incremental=false;
```
zaki50: `;` should be removed

Zhuinden:
Interestingly, the Realm World Tour slides were showing `inline fun <reified T : RealmModel>` extension methods for the Realm query (`realm.where<Dog>().findAll()`) 😄

zaki50:
Following change is needed to merge current master.
```
diff --git a/realm/kotlin-extensions/build.gradle b/realm/kotlin-extensions/buil
index 531e46d80..cc0ed80e7 100644
--- a/realm/kotlin-extensions/build.gradle
+++ b/realm/kotlin-extensions/build.gradle
@@ -16,10 +16,7 @@ android {

         testInstrumentationRunner "android.support.test.runner.AndroidJUnitRunn

-        flavorSelection 'api', 'base'
-        // flavorSelection is replaced by missingDimensionStrategy in beta4. Us
-        // https://androidstudio.googleblog.com/2017/09/android-studio-30-beta-
-        //missingDimensionStrategy 'api', 'base'
+        missingDimensionStrategy 'api', 'base'
     }
     buildTypes {
         debug {
```

cmelchior:
Regarding using `reified` vs. a `KClass` parameter I have no idea what is most ideomatic.
https://kotlinlang.org/docs/reference/inline-functions.html#reified-type-parameters seems to indicate that reified is, but I could be wrong?

```
// Java
public class Foo {
	public void bar(Class clazz) {
		...
	}
}

// Kotlin Extension methods
// Option 1: https://kotlinlang.org/docs/reference/inline-functions.html#reified-type-parameters
inline fun <reified T> Foo.bar(): Unit {
    return this.bar(T::class.java)
}

// Option 2
fun <T> Realm.delete(clazz: KClass<T>) {
    this.bar(clazz.java);
}
```

Thoughts @realm/java, @zaki50, @ericmaxwell2003 ?

nhachicha:
I think `reified` is better in this use case since the inlining has a better performance, moreover, the compiler is not copying large amount of code, just replacing each call with one overload

Zhuinden:
Definitely

``` kotlin
inline fun <reified T> Foo.bar(): Unit {
    this.bar(T::class.java)
}
```

cmelchior:
Ready for review @realm/java @ericmaxwell2003 . Note that most of this PR was reviewed in #5553, so it is mostly the Gradle parts missing.

realm/kotlin-extensions/build.gradle
```diff
@@ -0,0 +1,370 @@
...
+        defaults {
+             publications('basePublication', 'objectServerPublication')
+            publishPom = true
+            publishIvy = false
+        }
+    }
+}
+
+artifacts {
+    archives javadocJar
```
beeender: So the kotlin docs will be merged into the current javadoc?

cmelchior: Well, they replace them in the jar file. I'm not 100% who actually use that jar file though. IntelliJ uses the source code docs for all the ways you can access method documentation. If the jar file is just for bundling then html-pages, then it should work just fine

PR #4685, opened 2017-05-22
add architectureComponentsExample

zaki50:
Added an example project that uses ArchitectureComponents (`ViewModel` and `LiveData`).

I think that I should update the UI and output fancier.


examples/architectureComponentsExample/src/main/java/io/realm/examples/arch/ExampleViewModel.java
```diff
@@ -0,0 +1,45 @@
...
+package io.realm.examples.arch;
+
+import android.arch.lifecycle.ViewModel;
+
+import io.realm.Realm;
+import io.realm.examples.arch.livemodel.LiveRealmObject;
+import io.realm.examples.arch.model.Person;
+
+
+public class ExampleViewModel extends ViewModel {
```
cmelchior: This definitely work if you don't have a `Repository`, but if you introduce a Repository to hide DB/Network access (which they also recommend). Then it gets slightly more complicated, but perhaps it is fine to keep that out of this example?

examples/architectureComponentsExample/src/main/java/io/realm/examples/arch/ExampleViewModel.java
```diff
@@ -0,0 +1,45 @@
...
+    private final Realm realm;
+
+    private final LiveRealmObject<Person> livePerson;
+
+    public ExampleViewModel() {
+        realm = Realm.getDefaultInstance();
+        livePerson = new LiveRealmObject<>(realm.where(Person.class).findAllSorted("name").first());
```
cmelchior: This will run the query when the object is created on the UI thread which is is kinda bad. Ideally, it should be run inside `LiveObject.onActive()` method instead. Possible by overriding an abstract method?

examples/architectureComponentsExample/src/main/java/io/realm/examples/arch/livemodel/LiveRealmObject.java
```diff
@@ -0,0 +1,52 @@
...
+import android.arch.lifecycle.LiveData;
+import android.support.annotation.MainThread;
+
+import io.realm.ObjectChangeSet;
+import io.realm.RealmModel;
+import io.realm.RealmObject;
+import io.realm.RealmObjectChangeListener;
+
+
+public class LiveRealmObject<T extends RealmModel> extends LiveData<T> {
```
cmelchior: I created almost the exact same class in an example I was working on 😆 
There is few design choices for this class though, that might warrant different variations of it. In my case I was also opening the Realm in `onActive` and closing it in `onInActive`. Then had an abstract `RealmResults<T> onActive(Realm)` so yours could create their query inside that method.

Zhuinden: What a strange solution (from my perspective), I went with this approach https://gist.github.com/Zhuinden/7d49a2b136f6f478de09f0e79f4b0e0b

cmelchior: I don't think either approach is wrong, to be honest. Your approach put all of the responsibility in the ViewModel vs. this one where it lives in the LiveData object instead.

For small ViewModels, your approach seems fine to me, but I would be worried about how well it scales since you need to manually manage all the listeners. Having that inside the LiveData object means you don't need to write that code manually each time.

Zhuinden: Well in each new ViewModel the difference would be the query used, so u could make a base class based on that.

cmelchior: Yup, there are many ways you could split the responsibility. I have only played around with ViewModel/LiveData in toy examples so far and at least based on that I could see arguments for controlling the listeners in either the ViewModel or the LiveData. Neither seems inherently wrong to me.

One argument that favors LiveData is from Yirgits talk: The LocationLiveData class encapsulate the listeners in the LiveData class: https://youtu.be/bEKNi1JOrNs?t=14m43s

Zhuinden: Yeah, that makes sense, I just didn't think about the possibility of extending LiveData :smile: 

I based my example on [this](https://github.com/googlesamples/android-architecture-components/blob/178fe541643adb122d2a8925cf61a21950a4611c/BasicSample/app/src/main/java/com/example/android/persistence/viewmodel/ProductListViewModel.java#L44.L56) because it creates the DB and shares the products from its DAO inside the ViewModel, and the whole "open realm, share RealmResults" thing seemed eerily the same thing.

cmelchior: Not sure I agree with that. In that example, the Dao exposes a LiveData object which I would assume handles changes as well. That isn't done by the `ProductListViewModel`

Making the ViewModel handle the lifecycle of Realm seems like a good idea though.

Zhuinden: After thinking about it, the hybrid of these two approaches is the conceptually proper solution.

ViewModel should manage lifecycle of Realm, but `RealmResults<T>` **is a** `LiveData<List<T>>` so extending it is the reasonable choice.

Zhuinden: @ericmaxwell2003 actually used this proper hybrid approach in his article https://news.realm.io/news/android-architecture-components-and-realm/ 

I guess we're all soon thinking the same page 😄

examples/architectureComponentsExample/build.gradle
```diff
@@ -0,0 +1,41 @@
+apply plugin: 'com.android.application'
+apply plugin: 'android-command'
+apply plugin: 'realm-android'
+
+repositories {
+    maven {
+      url 'https://maven.google.com'
```
bmeike: incorrect indentation

examples/architectureComponentsExample/src/main/java/io/realm/examples/arch/LifecycleExampleActivity.java
```diff
@@ -0,0 +1,92 @@
...
+            @Override
+            public void onClick(View v) {
+                updater.stop();
+            }
+        });
+    }
+
+    @MainThread
+    private void updateViews(@Nullable Person person) {
+        if (person == null) {
```
bmeike: don't you mean:
`ageView.setText((person == null) ? "" : String.format(Locale.ENGLISH, "%1$s: %2$d", person.name, person.age));`

examples/architectureComponentsExample/src/main/java/io/realm/examples/arch/MyApplication.java
```diff
@@ -0,0 +1,43 @@
...
+
+
+public class MyApplication extends Application {
+
+    @Override
+    public void onCreate() {
+        super.onCreate();
+
+        Realm.init(this);
+        final RealmConfiguration defaultConfig = new RealmConfiguration.Builder().initialData(new Realm.Transaction() {
```
bmeike: I don't understand this `final`.  Why introduce an un-necessary variable, but make if final?

examples/architectureComponentsExample/src/main/java/io/realm/examples/arch/Updater.java
```diff
@@ -0,0 +1,83 @@
...
+       stop();
+    }
+
+    final class IncrementThread extends Thread {
+        IncrementThread() {
+            super("increment thread");
+        }
+
+        @Override
+        public void run() {
```
bmeike: I would prefer that this be a `Runnable`, passed to the a vanilla thread, on initialization.

examples/architectureComponentsExample/src/main/java/io/realm/examples/arch/Updater.java
```diff
@@ -0,0 +1,83 @@
...
+            final Realm realm = Realm.getDefaultInstance();
+            //noinspection TryFinallyCanBeTryWithResources
+            try {
+                final RealmResults<Person> all = realm.where(Person.class).findAll();
+                Realm.Transaction transaction = new Realm.Transaction() {
```
bmeike: Yeah... I really don't understand the use of `final`.  I am familiar with three styles of use: 1) use it only when required; 2) use it whenever possible on fields and only when required, elsewhere and 3) use it whenever possible.  There is actually a 4th, seldom used style that sometimes accompanies 1-3 above: Always use final on method parameters.  Are you proposing a different style?  Can you describe it?  Should we adopt it across our code?

examples/architectureComponentsExample/src/main/res/layout/activity_realm_basic_example.xml
```diff
@@ -0,0 +1,33 @@
...
+
+    <TextView
+        android:id="@+id/age"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:gravity="center_horizontal"
+        android:textSize="18sp"
+        tools:text="16"/>
+
+    <Button
```
bmeike: maybe wanna put these buttons side by side?

examples/architectureComponentsExample/src/main/java/io/realm/examples/arch/MyApplication.java
zaki50: I like `final` (single assignment) and hate assigning multiple times to one variable.

It's just my style. I'll remove `final`.

examples/architectureComponentsExample/src/main/java/io/realm/examples/arch/Updater.java
zaki50: Why?

Zhuinden: Generally we have been taught that you *can* extend Thread but honestly though conceptually it's a bit odd to reason about why you have a subclass of a thread that can do additional things, instead of just passing it a Runnable

> composition over inheritance

basically yes

examples/architectureComponentsExample/src/main/java/io/realm/examples/arch/Updater.java
Zhuinden: In Java you need to make a variable final in order to allow access to it inside an anonymous inside class created in the method.

If you don't specify it, then it's "implicitly final", but it's still final.

examples/architectureComponentsExample/src/main/java/io/realm/examples/arch/LifecycleExampleActivity.java
Zhuinden: DRY :smile:

examples/architectureComponentsExample/src/main/java/io/realm/examples/arch/ExampleViewModel.java
Zhuinden: In the example I linked below, I just used `RealmChangeListener` to initialize the `MutableLiveData` after it's been loaded, using async query.

Maybe that could be used as reference here as well

examples/architectureComponentsExample/src/main/java/io/realm/examples/arch/ExampleViewModel.java
Zhuinden: Honestly I don't really understand the `Repository` layer. I think data access and loading remote data are separate in the sense that you might need to access the db to know if something *has to be* downloaded, but the whole "get it from network directly" thing is a mess!

Just observe the db and then write into the db from network call, if you do that then all data will update everywhere else.

Binding the two data sources together is a recipe for disaster, but maybe that's just what I think?

cmelchior: That is the point of the Repository pattern. The rest of the application should not care where the data comes from. It could either be the network of the database, but that is an implementation detail.

For true offline support, you should always read data from the database, but that is orthogonal to using the repository pattern.

I tried to visualize it here: https://www.slideshare.net/ChristianMelchior/painless-persistence-with-realm#12 and an example of how it could be implemented is here: https://github.com/realm/realm-java/blob/master/examples/newsreaderExample/src/main/java/io/realm/examples/newsreader/model/Repository.java#L77

Zhuinden: > For true offline support, you should always read data from the database, but that is orthogonal to using the repository pattern.

oh so **that's why I never got it**! We always had to support offline mode, or even when we didn't, I architected it in an "offline-first" way.

Not much point to `forceReload`ing a `RealmResults` though, it's not like it can be out of date, heh.

cmelchior:
@zaki50 Any update on this one?

Zhuinden:
What happened to this one?

Zhuinden:
In the meantime, I started experimenting with Lifecycle management entirely through LiveData onActive/onInactive.

You can see it [here](https://github.com/Zhuinden/realm-monarchy/blob/938386a5d6afe4909037a5541169efa3fcf94693/monarchy-example/src/main/java/com/zhuinden/monarchyexample/MainActivity.java#L52-L60).

Zhuinden:
@cmelchior now that I have a Mac too, I can probably compile Realm, which means I could possibly finish this PR.

Should I?

cmelchior:
@Zhuinden Congratulations 🎉 

Yes, that would be extremely nice :+1:

Zhuinden:
@cmelchior I have a branch ready on my PC but I'm getting 403 when I'm trying to push to realm-java.

What am I missing? I even created my own branch as I figured I can't directly push *here*.

Zhuinden:
This is the last step :smile:

PR #4689, opened 2017-05-23
Split createTable & getTable

beeender:
To support Stable IDs, the primary key has to be created while the table
is created. This will require a new internal API createTableWithPK in
the future.
The previous getTable behaviour would be quite confusing to support
that.

- getTable will only return the table if it exists.
- createTable is the one to be used to create tables.
- Add JavaExceptionDef to have all common Java exception names in a
  single place.
- Remove useless tests.


realm/realm-library/src/main/cpp/java_exception_def.cpp
```diff
@@ -0,0 +1,21 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "java_exception_def.hpp"
+
+using namespace realm::_impl;
+
+const char* JavaExceptionDef::IllegalState = "java/lang/IllegalStateException";
```
cmelchior: A comment explaining this class would be nice

cmelchior: Couldn't this be inlined in the hpp file?

beeender: no, see https://stackoverflow.com/questions/18860895/how-to-initialize-static-members-in-the-header

beeender: added to the header

realm/realm-library/src/main/java/io/realm/internal/SharedRealm.java
```diff
@@ -527,8 +546,12 @@ private static native long nativeCreateConfig(String realmPath, byte[] key, byte
 
     private static native long[] nativeGetVersionID(long nativeSharedRealmPtr);
 
+    // Throw IAE if the table doesn't exist.
     private static native long nativeGetTable(long nativeSharedRealmPtr, String tableName);
 
+    // Native assertion if the table with given name exists.
```
zaki50: `RealmError` seems to be thrown if the name conflict happens. correct?

realm/realm-library/src/main/java/io/realm/StandardRealmSchema.java
```diff
@@ -110,7 +110,7 @@ public RealmObjectSchema create(String className) {
         if (realm.getSharedRealm().hasTable(internalTableName)) {
```
zaki50: I think we can skip this test if the `createTable()` throws a proper exception against the name conflict.

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
```diff
@@ -397,18 +400,34 @@ JNIEXPORT jlong JNICALL Java_io_realm_internal_SharedRealm_nativeGetTable(JNIEnv
     try {
         JStringAccessor name(env, table_name); // throws
         auto& shared_realm = *(reinterpret_cast<SharedRealm*>(shared_realm_ptr));
-        if (!shared_realm->read_group().has_table(name) && !shared_realm->is_in_transaction()) {
-            std::ostringstream ss;
-            ss << "Class " << name << " doesn't exist and the shared Realm is not in transaction.";
-            ThrowException(env, IllegalState, ss.str());
-            return static_cast<jlong>(NULL);
+        if(!shared_realm->read_group().has_table(name)) {
```
kneth: space is missing

realm/realm-library/src/main/java/io/realm/StandardRealmSchema.java
kneth: And `LangBindHelper::add_table()` will probably throw and we will eventually see it here.

realm/realm-library/src/main/java/io/realm/internal/SharedRealm.java
```diff
@@ -288,8 +288,27 @@ public boolean hasTable(String name) {
         return nativeHasTable(nativePtr, name);
     }
 
+    /**
+     * Gets a existing {@link Table} with the given name.
```
kneth: "a" -> "an"

realm/realm-library/src/main/java/io/realm/internal/Table.java
```diff
@@ -753,6 +753,12 @@ private Table getPrimaryKeyTable() {
         if (sharedRealm == null) {
             return null;
         }
+
+        // FIXME: The PK table creation should be handle by Object Store after integration of OS Schema.
+        if (!sharedRealm.hasTable(PRIMARY_KEY_TABLE_NAME)) {
```
kneth: Doesn't `create_metadata_tables()` create this table?

realm/realm-library/src/main/java/io/realm/StandardRealmSchema.java
beeender: YEAH, changed to throw from there.

realm/realm-library/src/main/java/io/realm/internal/Table.java
beeender: The PK table may still get created from here in some cases, eg.:`moreFieldsChangedThanLocalRefTableSize` test. So I would like to keep the old logic in this PR and clean it up in the future.

PR #4690, opened 2017-05-23
Fix queries not working on proguarded Realms model classes

### LINKED ISSUE #4673: Invalid query: table not found in this schema.
Hi there!
Faced a strange issue: ProGuard build is crashing with next message:

```
java.lang.IllegalArgumentException: Invalid query: table 'TableNameHere' not found in this schema.
at aig.a(CachedFieldDescriptor.java:72)
at aii.c(FieldDescriptor.java:284)
at aii.a(FieldDescriptor.java:177)
at ahb.b(RealmQuery.java:347)
at ahb.a(RealmQuery.java:339)
```

Proguard main configuration part:

```
-dontskipnonpubliclibraryclassmembers
-optimizationpasses 7
-allowaccessmodification
-repackageclasses
-dontpreverify
```

Realm version(s): 3.2.1
Android Studio version: 2.3.2
Which Android version and device: 7.1.2, Nexus 5X, Pixel XL

cmelchior:
Fixes #4673 

The new ColumnIndices wrapper cached names using `class.getSimpleName()` which breaks when the class is obfuscated (We should really have caught this in code review 😢 )

I modified it so a pair consisting of `<class, realSimpleName>` is now being passed around instead. It complicated the internals of `ColumnIndices` a little more than I would like (We could really use a MultiKeyMap implementation).

Creating/Copying this class is now also more expensive as we have to go through the mediator for the proper non-obfuscated name, but given how rarely this happens, it is probably acceptable for now.

We have an example `moduleExample/app` that used a proguarded Realm and we test the Release build as part of the release process, but unfortunately, that example didn't use any query methods that triggered the bug.

It is apparently possible to run unit tests with ProGuard: https://stackoverflow.com/questions/26482119/android-unit-tests-with-proguard-enabled so I'll be creating an issue for enabling that going forward.

For now, I modified the example to use a query that will catch this bug.

### DIFF
--- a/realm/realm-library/src/main/java/io/realm/internal/ColumnIndices.java
+++ b/realm/realm-library/src/main/java/io/realm/internal/ColumnIndices.java
@@ -79,11 +79,10 @@ public final class ColumnIndices {
         this(other.schemaVersion, new HashMap<Pair<Class<? extends RealmModel>, String>, ColumnInfo>(other.classesToColumnInfo.size()), mutable);
         for (Map.Entry<Pair<Class<? extends RealmModel>, String>, ColumnInfo> entry : other.classesToColumnInfo.entrySet()) {
             ColumnInfo columnInfo = entry.getValue().copy(mutable);
-            Pair<Class<? extends RealmModel>, String> oldKey = entry.getKey();
-            Pair<Class<? extends RealmModel>, String> newKey = Pair.<Class<? extends RealmModel>, String>create(oldKey.first, oldKey.second);
-            this.classes.put(newKey.first, columnInfo);
-            this.classesByName.put(newKey.second, columnInfo);
-            this.classesToColumnInfo.put(newKey, columnInfo);
+            Pair<Class<? extends RealmModel>, String> key = entry.getKey();
+            this.classes.put(key.first, columnInfo);
+            this.classesByName.put(key.second, columnInfo);
+            this.classesToColumnInfo.put(key, columnInfo);
         }
     }
 
--- a/realm/realm-library/src/main/java/io/realm/internal/ColumnIndices.java
+++ b/realm/realm-library/src/main/java/io/realm/internal/ColumnIndices.java
@@ -64,13 +64,13 @@ public final class ColumnIndices {
                 throw new IllegalArgumentException("ColumnInfo mutability does not match ColumnIndices");
             }
             Pair<Class<? extends RealmModel>, String> classDescription = entry.getKey();
-            this.classes.put(classDescription.first, entry.getValue());
-            this.classesByName.put(classDescription.second, entry.getValue());
+            this.classes.put(classDescription.first, columnInfo);
+            this.classesByName.put(classDescription.second, columnInfo);
         }
     }
 
     /**
-     * Create a copy of the passed ColumnIndices with the specified mutablity.
+     * Create a copy of the passed ColumnIndices with the specified mutability.
      *
      * @param other the ColumnIndices object to copy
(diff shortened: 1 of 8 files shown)


realm/realm-library/src/main/java/io/realm/internal/util/Pair.java
```diff
@@ -0,0 +1,90 @@
...
+ * Copy from the Android framework to avoid the dependency on Android classes + slight adjustment
+ * to support older versions of Android.
+ *
+ * Original source: https://android.googlesource.com/platform/frameworks/base/+/master/core/java/android/util/Pair.java
+ *
+ * Container to ease passing around a tuple of two objects. This object provides a sensible
+ * implementation of equals(), returning true if equals() is true on each of the contained
+ * objects.
+ */
+public class Pair<F, S> {
```
MGaetan89: Any reason to not use [`android.util.Pair`](https://developer.android.com/reference/android/util/Pair.html) directly?

Zhuinden: I assume it has to do with future plans to go non-android too with java api

Zhuinden: I assume it has to do with future plans to go non-android too with java api

cmelchior: Yes, it is an Android dependency, so when we move the library to the JVM it wouldn't be available. This is also why we try to have as few android dependencies as possible in general, because those need to be stubbed for the JVM

MGaetan89: I see, thanks for the info.

realm/realm-library/src/main/java/io/realm/internal/util/Pair.java
```diff
@@ -0,0 +1,90 @@
...
+    public F first;
+    public S second;
```
zaki50: should be `final`?

realm/realm-library/src/main/java/io/realm/internal/ColumnIndices.java
```diff
@@ -39,49 +40,59 @@
...
     public ColumnIndices(ColumnIndices other, boolean mutable) {
-        this(other.schemaVersion, new HashMap<Class<? extends RealmModel>, ColumnInfo>(other.classes.size()), mutable);
-        for (Map.Entry<Class<? extends RealmModel>, ColumnInfo> entry : other.classes.entrySet()) {
+        this(other.schemaVersion, new HashMap<Pair<Class<? extends RealmModel>, String>, ColumnInfo>(other.classesToColumnInfo.size()), mutable);
+        for (Map.Entry<Pair<Class<? extends RealmModel>, String>, ColumnInfo> entry : other.classesToColumnInfo.entrySet()) {
             ColumnInfo columnInfo = entry.getValue().copy(mutable);
-            this.classes.put(entry.getKey(), columnInfo);
-            this.classesByName.put(entry.getKey().getSimpleName(), columnInfo);
+            Pair<Class<? extends RealmModel>, String> oldKey = entry.getKey();
+            Pair<Class<? extends RealmModel>, String> newKey = Pair.<Class<? extends RealmModel>, String>create(oldKey.first, oldKey.second);
```
beeender: why do you need to create a new `Pair`? the `first` and `second` are both immutable anyway?

cmelchior: Good point 👍

PR #4693, opened 2017-05-23
Support stable IDs for sync

beeender:
A special column to store stable IDs are added by sync. See
https://github.com/realm/realm-sync/pull/1170 for details.

- Calling sync::create_object instead of add_empty_row to create new
  object.
- Calling sync::create_table/sync::create_table_with_primary_key instead
  of add_table to create a new object schema.
- addPrimaryKey() is not allowed for synced Realm anymore.
- New API RealmSchema.createWithPrimaryKeyField is added.

TODO:
- [x] Check the version in `dependencies.list`.
- [x] Check the object store submodule after https://github.com/realm/realm-object-store/pull/434 merged.
- [x] Testing with new ROS release which supports stable IDs.


realm/realm-library/src/androidTestObjectServer/java/io/realm/SchemaTests.java
```diff
@@ -145,4 +147,54 @@ public void allow_addField() {
...
+    public void addPrimaryKey_notAllowed() {
+        String className = "StringOnly";
+        Realm realm = Realm.getInstance(config);
+
+        realm.beginTransaction();
+        RealmObjectSchema objectSchema = realm.getSchema().get(className);
+        objectSchema.addField("foo", String.class);
+
+        try {
+            objectSchema.addPrimaryKey("foo");
```
cmelchior: If we are releasing this as part of a 4.0, why not just remove this method completely?

realm/realm-library/src/androidTestObjectServer/java/io/realm/SchemaTests.java
```diff
@@ -145,4 +147,54 @@ public void allow_addField() {
...
+        String className = "StringOnly";
+        Realm realm = Realm.getInstance(config);
+
+        realm.beginTransaction();
+        RealmObjectSchema objectSchema = realm.getSchema().get(className);
+
+        try {
+            objectSchema.addField("foo", String.class, FieldAttribute.PRIMARY_KEY);
+            fail();
+        } catch (UnsupportedOperationException ignored) {
```
cmelchior: Would IllegalArgumentException be better since it is the `FieldAttribute` that is causing this?

realm/realm-library/src/androidTestObjectServer/java/io/realm/SchemaTests.java
```diff
@@ -145,4 +147,54 @@ public void allow_addField() {
...
+        } finally {
+            realm.commitTransaction();
+            realm.close();
+        }
+    }
+
+    // Special column "__OID" should be hidden from users.
```
cmelchior: Hopefully, this will be changed to `!IOD` before release to avoid clashes with user columns.

realm/realm-library/src/main/cpp/io_realm_internal_OsObject.cpp
```diff
@@ -165,7 +165,11 @@ static inline size_t do_create_row(jlong shared_realm_ptr, jlong table_ptr)
     auto& shared_realm = *(reinterpret_cast<SharedRealm*>(shared_realm_ptr));
     auto& table = *(reinterpret_cast<realm::Table*>(table_ptr));
     shared_realm->verify_in_write();
+#if REALM_ENABLE_SYNC
+    return sync::create_object(shared_realm->read_group(), table);
```
cmelchior: Didn't OS add some feature flag so the same API could be used for both 1.* and 2.* or did I misunderstood something?

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
```diff
@@ -433,22 +435,83 @@ JNIEXPORT jlong JNICALL Java_io_realm_internal_SharedRealm_nativeGetTable(JNIEnv
...
-        JStringAccessor name(env, table_name); // throws
-        name_str = name;
+        table_name_str = JStringAccessor(env, table_name); // throws
         auto& shared_realm = *(reinterpret_cast<SharedRealm*>(shared_realm_ptr));
         shared_realm->verify_in_write(); // throws
-        Table* table = LangBindHelper::add_table(shared_realm->read_group(), name); // throws
+        Table* table;
+        auto& group = shared_realm->read_group();
+        if (is_pk_table) {
+            // sync::create_table() will add an extra column for stable ID which is not allowed for pk table.
```
cmelchior: Why isn't that allowed for the primary key table?

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
```diff
@@ -433,22 +435,83 @@ JNIEXPORT jlong JNICALL Java_io_realm_internal_SharedRealm_nativeGetTable(JNIEnv
...
+
+JNIEXPORT jlong JNICALL Java_io_realm_internal_SharedRealm_nativeCreateTableWithPrimaryKeyField(
+    JNIEnv* env, jclass, jlong shared_realm_ptr, jstring table_name, jstring field_name, jboolean is_string_type,
+    jboolean is_nullable)
+{
+    TR_ENTER_PTR(shared_realm_ptr)
+
+    std::string class_name_str;
+    try {
+        std::string table_name_str(JStringAccessor(env, table_name));
```
cmelchior: We really need to find a common pattern for the naming of java parameters that get converted to c++ ones and then document it. Here we use `field_name` -> `field_name_str`. In other places it is `fieldName` -> `field_name` and yet others it is `j_field_name` -> `field_name`. I don't particularly care which to use, although I would probably prefer if the c++ name was as short as possible i.e. `field_name`.

realm/realm-library/src/main/java/io/realm/OsRealmSchema.java
```diff
@@ -64,6 +64,12 @@ public RealmObjectSchema create(String className) {
         }
 
         @Override
+        public RealmObjectSchema createWithPrimaryKeyField(String className, String primaryKeyFieldName,
```
cmelchior: Just delete this functon?

realm/realm-library/src/main/java/io/realm/OsRealmSchema.java
```diff
@@ -168,6 +174,12 @@ public RealmObjectSchema create(String className) {
     }
 
     @Override
+    public RealmObjectSchema createWithPrimaryKeyField(String className, String primaryKeyFieldName, Class<?> fieldType,
```
cmelchior: Same

realm/realm-library/src/main/java/io/realm/StandardRealmObjectSchema.java
```diff
@@ -676,6 +687,12 @@ private void checkFieldExists(String fieldName) {
         }
     }
 
+    private void checkAddPrimaryKeyForSync() {
```
cmelchior: IMO we should strive to keep the rules for synced and non-synced Realms as close as possible. Wouldn't it make sense to disallow this for normal Realms as well?

realm/realm-library/src/main/java/io/realm/StandardRealmSchema.java
```diff
@@ -102,15 +104,40 @@ public RealmObjectSchema get(String className) {
...
+        StandardRealmObjectSchema.checkLegalName(primaryKeyFieldName);
+        String internalTableName = checkAndGetTableNameFromClassName(className);
+
+        StandardRealmObjectSchema.FieldMetaData metadata = SUPPORTED_SIMPLE_FIELDS.get(fieldType);
+        if (metadata == null || (metadata.realmType != RealmFieldType.STRING &&
+                metadata.realmType != RealmFieldType.INTEGER)) {
+            throw new IllegalArgumentException(String.format("Realm doesn't support primary key field type '%s'.",
+                    fieldType));
+        }
+        boolean isStringField = metadata.realmType == RealmFieldType.STRING;
```
cmelchior: `(metadata.realmType == RealmFieldType.STRING);` would be more readable IMO

realm/realm-library/src/main/java/io/realm/internal/SharedRealm.java
```diff
@@ -305,7 +306,37 @@ public Table getTable(String name) {
...
+     * Creates a primary key table with then given name. Native assertion will happen if the table with the same name
+     * exists. This function is different from {@link #createTable(String)} which will call {@code create_table()} from
+     * sync to do the creation. This will always call the core's {@code add_table()} to avoid creating the stable id
+     * column for pk table.
+     *
+     * @param name the name of table.
+     * @return a created {@link Table} object.
+     */
+    public Table createPkTable(String name) {
+        return new Table(this, nativeCreateTable(nativePtr, name, true));
```
cmelchior: Is this how the other bindings do this as well?

realm/realm-library/src/main/cpp/io_realm_internal_OsObject.cpp
beeender: Yes, OS did. But those similar logic exist in `object_accessor.hpp` which we are not ready to use right now. Mostly because of the OS schema. See https://github.com/realm/realm-object-store/pull/434/files#diff-99a0771b4ad4157fdbcf45dfdc8f1e24R184

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
beeender: If `__0ID` is added first, then the column index for the PK table will be changed which are hard coded everywhere. Object Store is not using sync's APK to create PK table either. https://github.com/realm/realm-object-store/blob/2aa939f1c0a2f849f67f76ec6a736e7dae36c1ea/src/object_store.cpp#L61

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
beeender: `j_field_name` is good, i will change.

realm/realm-library/src/main/java/io/realm/StandardRealmObjectSchema.java
beeender: I totally think we should disallow adding primary for non-sync in 4.0. But we need to align with other bindings.

realm/realm-library/src/main/java/io/realm/internal/SharedRealm.java
beeender: the `name` param should be removed, this is only used for creating PK table which the name is constantly `PRIMARY_KEY_TABLE_NAME`. Yes, Object Store is using the same logic. https://github.com/realm/realm-object-store/blob/2aa939f1c0a2f849f67f76ec6a736e7dae36c1ea/src/object_store.cpp#L57
But, we really really really should call OS to do that. After OS schema migration, i will change this.

realm/realm-library/src/main/java/io/realm/OsRealmSchema.java
beeender: Cannot delete it now, it is part of the `RealmSchema` interface. will remove it in OS schema migration PR.

realm/realm-library/src/main/java/io/realm/OsRealmSchema.java
beeender: ditto

realm/realm-library/src/main/java/io/realm/StandardRealmObjectSchema.java
cmelchior: Since migrations work very differently across bindings, I'm not sure that is needed?

cmelchior:
The CI fail looks legit enough, but apart from that I think this looks good (missing resolving merge conflicts though)

realm/realm-library/src/androidTest/java/io/realm/DynamicRealmObjectTests.java
```diff
@@ -1153,14 +1155,16 @@ public void isNull_false() {
...
-        // After the stable ID support, primary key field will be inserted first before others. So even FEILD_STRING is
-        // the first defined field in the class, it will be inserted after FIELD_ID.
-        String[] expectedKeys = {AllJavaTypes.FIELD_ID, AllJavaTypes.FIELD_STRING, AllJavaTypes.FIELD_LONG,
+        String[] expectedKeys = {AllJavaTypes.FIELD_STRING, AllJavaTypes.FIELD_ID, AllJavaTypes.FIELD_LONG,
                 AllJavaTypes.FIELD_SHORT, AllJavaTypes.FIELD_INT, AllJavaTypes.FIELD_BYTE, AllJavaTypes.FIELD_FLOAT,
                 AllJavaTypes.FIELD_DOUBLE, AllJavaTypes.FIELD_BOOLEAN, AllJavaTypes.FIELD_DATE,
                 AllJavaTypes.FIELD_BINARY, AllJavaTypes.FIELD_OBJECT, AllJavaTypes.FIELD_LIST};
         String[] keys = dObjTyped.getFieldNames();
-        assertArrayEquals(expectedKeys, keys);
+        // After the stable ID support, primary key field will be inserted first before others. So even FEILD_STRING is
```
Zhuinden: FEILD -> FIELD?

realm/realm-library/src/androidTestObjectServer/java/io/realm/SchemaTests.java
cmelchior: Thoughts on this @beeender ?

realm/realm-library/src/androidTestObjectServer/java/io/realm/SchemaTests.java
cmelchior: This will be changed in RC11

realm/realm-library/src/androidTestObjectServer/java/io/realm/SchemaTests.java
beeender: @cmelchior The place to throw exception is the same with `addPrimaryKey()`. So is it also feasible to throw `IllegalArgumentException` for `addPrimaryKey()`? it sounds a bit weird since `addPrimaryKey()` takes the field name as param, and it is not the reason of exception.

realm/realm-library/src/androidTestObjectServer/java/io/realm/SchemaTests.java
beeender: will change this after we RC11.

realm/realm-library/src/androidTest/java/io/realm/RealmSchemaTests.java
```diff
@@ -115,6 +116,86 @@ public void create_duplicatedNameThrows() {
     }
 
     @Test
+    public void createWithPrimaryKeyField_string() {
+        // Not nullable
+        realmSchema.createWithPrimaryKeyField("FooNonNull", "pkField", String.class, FieldAttribute.REQUIRED);
+        RealmObjectSchema objectSchema = realmSchema.getSchemaForClass("FooNonNull");
```
beeender: we didn't overwrite the `RealmObjectSchema.equals()`, so the `assertSame` will have no point here i think?

realm/realm-library/src/androidTest/java/io/realm/RealmSchemaTests.java
```diff
@@ -115,6 +116,86 @@ public void create_duplicatedNameThrows() {
...
+                fail();
+            } catch (IllegalArgumentException ignored) {
+            }
+            assertFalse(String.format("'%s' failed", name), realmSchema.contains(name));
+        }
+    }
+
+    @Test
+    public void createWithPrimaryKeyField_invalidFieldNameThrows() {
+        String[] fieldNames = new String[] { null, "", "foo.bar", TestHelper.getRandomString(65) };
```
beeender: I think i changed `TestHelper.getRandomString()` return value somewhere, it will only return capital letters string.

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
```diff
@@ -433,22 +435,83 @@ JNIEXPORT jlong JNICALL Java_io_realm_internal_SharedRealm_nativeGetTable(JNIEnv
...
+            table = LangBindHelper::get_table(group, table_ref->get_index_in_group());
+#else
+            table = LangBindHelper::add_table(group, table_name); // throws
+#endif
+        }
+        return reinterpret_cast<jlong>(table);
+    }
+    catch (TableNameInUse& e) {
+        // We need to print the table name, so catch the exception here.
+        std::string class_name_str(table_name.substr(TABLE_PREFIX.length()));
```
beeender: > Is it always OK to remove first a few characters?
yes, otherwise we made something really wrong.

> Is it sure that this method is not used for pk or metadata table?
This method is used for pk table as well, but we have check before, see 
https://github.com/realm/realm-java/pull/4693/files#diff-282e4aca0c4608335667e0c13e86e699L612

> I've found that there is a special method to create pk table.
> We don't have metadata table already?
The pk table is a bit messy, ideally it should always be handled in Object Store. And by moving move schema things there, we are quite close to the target.
meta table is better, we always calling `ObjectStore::set_schema_version()` which will create the meta table if it doesn't exist.

realm/realm-library/src/androidTest/java/io/realm/RealmSchemaTests.java
```diff
@@ -115,6 +116,86 @@ public void create_duplicatedNameThrows() {
...
+        objectSchema = realmSchema.getSchemaForClass("FooNull");
+        assertEquals("pkField", objectSchema.getPrimaryKey());
+        assertEquals(RealmFieldType.STRING, objectSchema.getFieldType("pkField"));
+        assertTrue(objectSchema.isNullable("pkField"));
+    }
+
+    @Test
+    public void createWithPrimaryKeyField_integer() {
+        // Not nullable
+        realmSchema.createWithPrimaryKeyField("FooNonNull", "pkField", int.class);
```
beeender: you comments lead me to the same stupid bug twice!! thanks a lot! https://github.com/realm/realm-java/pull/4884/commits/5bd9b2b7c4667256cac9030c7208006fddc7a16d

beeender: added a lot more test.

realm/realm-library/src/androidTest/java/io/realm/RealmSchemaTests.java
beeender: merge it back from the closed PR ... stupid me

realm/realm-library/src/androidTest/java/io/realm/TestHelper.java
```diff
@@ -346,11 +346,12 @@ public void log(int level, String tag, Throwable throwable, String message) {
         };
     }
 
+    // Generate a random string with only capital letters which is always a valid class/field name.
     public static String getRandomString(int length) {
         Random r = new Random();
         StringBuilder sb = new StringBuilder(length);
         for (int i = 0; i < length; i++) {
-            sb.append((char) r.nextInt(128)); // Restrict to standard ASCII chars.
+            sb.append((char) r.nextInt(25) + 'A'); // Restrict to capital letters
```
zaki50: `r.nextInt(25)` should be `r.nextInt(26)` or `r.nextInt('Z' + 1 - 'A')`.

`r.nextInt(x)` returns a value between `0` to `x - 1`.

realm/realm-library/src/androidTest/java/io/realm/RealmSchemaTests.java
zaki50: Ah, `createWithPrimaryKeyField()` doesn't update schema cache!
:+1:

realm/realm-library/src/androidTestObjectServer/java/io/realm/SchemaTests.java
zaki50: Can users migrate from this version to RC11?

beeender: nope, but we are not going to make an official release with this.

realm/realm-library/src/androidTestObjectServer/java/io/realm/SchemaTests.java
Zhuinden: Because migration of primary key in non-sync Realm would be quite a pain if you can't remove and re-add primary key to a different field. Mapping all objects to a new object schema just to change PK is rough...

cmelchior: Yup, we also agreeded that we should keep this for non-synced Realms. While we generally want to keep the API as similar as possible, it doesn't seem possible for migrations where synced Realms have a lot of other restrictions as well.

PR #4705, opened 2017-05-26
Turn off the column # check for synced realms

bmeike:
Fix for column count bug


bmeike:
Please review this.  It appears that there are no tests, either way, for the affected use cases.

zaki50:
We need to have an entry for this in CHANGELOG at least.

zaki50:
superseded by #4706

PR #4706, opened 2017-05-26
Turn off the column # check for synced realms

zaki50:


beeender:
I was going to change it in the stable IDs PR anyway :P

beeender:
So benchmark fails on those tests? are you sure? I think CI is running benchmarks.

PR #4707, opened 2017-05-26
Convert exception on sync client to java exception

beeender:
- Update realm-sync to 1.10.1
- update object-store to 1e3cbb1789
- Convert exception on sync client to java exception
  This will give us some better information when excpetion happens
  on the sync client thread.


realm/realm-library/src/main/cpp/io_realm_SyncManager.cpp
```diff
@@ -44,7 +50,23 @@ struct AndroidClientListener : public realm::BindingCallbackThreadObserver {
...
         JniUtils::detach_current_thread();
     }
-} s_client_thread_listener;
+
+    void handle_error(std::exception const& e) override
+    {
+        JNIEnv* env = JniUtils::get_env(true);
+        std::string msg = format("An exception has been thrown on the sync client thread:\n%1", e.what());
+        Log::f(msg.c_str());
+        // Since user has no way to handle exceptions thrown on the sync client thread, we just convert it to an Java
```
nhachicha: to a Java

nhachicha:
you have a compile error
```
  /tmp/realm-java/realm/realm-library/src/main/cpp/io_realm_SyncManager.cpp:54:10: error: 'void AndroidClientListener::handle_error(const std::exception&)' marked override, but does not override
       void handle_error(std::exception const& e) override
```
did you update the object store commit to point to realm/realm-object-store#456 ?

beeender:
@nhachicha No i didn't ... unfortunately object-store master needs `1.9.1` whose corresponding ROS release is not in public yet. We might need to branch out the Object Store for a while to merge this ....

realm/realm-library/src/main/cpp/io_realm_SyncManager.cpp
```diff
@@ -25,11 +25,17 @@
 #include "util.hpp"
 #include "jni_util/jni_utils.hpp"
 #include "jni_util/java_method.hpp"
+#include "jni_util/java_class.hpp"
```
bmeike: Aren't these supposed to be in alpha order?

realm/realm-library/src/main/cpp/io_realm_SyncManager.cpp
```diff
@@ -25,11 +25,17 @@
...
 
 using namespace realm;
+using namespace realm::util;
 using namespace realm::jni_util;
```
bmeike: alpha order?

realm/realm-library/src/main/cpp/io_realm_SyncManager.cpp
```diff
@@ -44,7 +50,23 @@ struct AndroidClientListener : public realm::BindingCallbackThreadObserver {
...
+        std::string msg = format("An exception has been thrown on the sync client thread:\n%1", e.what());
+        Log::f(msg.c_str());
+        // Since user has no way to handle exceptions thrown on the sync client thread, we just convert it to an Java
+        // exception to get more debug information for ourself.
+        // FIXME: We really need to find a universal and clever way to get the native backtrace when exception thrown
+        env->ThrowNew(m_realm_exception_class, msg.c_str());
+    }
+
+private:
+    // For some reasons, FindClass() doesn't work in the native thread even when the jvm is attached before. Get the
```
bmeike: jvm or JVM.  pick one.

beeender:
@realm/java This did a core update as well.

beeender:
This goes to master since it won't be compatible with ROS < 1.6.0 which is considered as a breaking change i think?

realm/realm-library/src/main/cpp/io_realm_SyncManager.cpp
```diff
@@ -44,7 +50,23 @@ struct AndroidClientListener : public realm::BindingCallbackThreadObserver {
...
+        std::string msg = format("An exception has been thrown on the sync client thread:\n%1", e.what());
+        Log::f(msg.c_str());
+        // Since user has no way to handle exceptions thrown on the sync client thread, we just convert it to an Java
+        // exception to get more debug information for ourself.
+        // FIXME: We really need to find a universal and clever way to get the native backtrace when exception thrown
+        env->ThrowNew(m_realm_exception_class, msg.c_str());
+    }
+
+private:
+    // For some reasons, FindClass() doesn't work in the native thread even when the JVM is attached before. Get the
```
kneth: Is it a bug in Android? Is it a know bug? Or does it also apply to any JVM implementation?

beeender: no idea, I cannot find it documented anywhere. I feel like it is an Android only thing. Strange thing is, if you have the method id and the jclass/jobject, the java method can be called without any problems.

kneth:
Should we disable the failing unit test if we cannot get a core hot fix?

Otherwise I think we can merge it.

beeender:
@kneth I guess we cannot ... It will make the master not-releasable ...

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ### Breaking Changes
 
+* [ObjectServer] Update protocol version to 18 which is only compatible with ROS > 1.6.0.
```
dalinaum: Updated protocol?

realm/realm-library/src/main/cpp/io_realm_SyncManager.cpp
```diff
@@ -72,8 +94,9 @@ JNIEXPORT void JNICALL Java_io_realm_SyncManager_nativeInitializeSyncManager(JNI
         JStringAccessor base_file_path(env, sync_base_dir); // throws
         SyncManager::shared().configure_file_system(base_file_path, SyncManager::MetadataMode::NoEncryption);
 
+        static AndroidClientListener client_thread_listner(env);
```
dalinaum: It should be `client_thread_listener`.

realm/realm-library/src/main/cpp/io_realm_SyncManager.cpp
```diff
@@ -72,8 +94,9 @@ JNIEXPORT void JNICALL Java_io_realm_SyncManager_nativeInitializeSyncManager(JNI
...
         // Register Sync Client thread start/stop callback
-        g_binding_callback_thread_observer = &s_client_thread_listener;
+        g_binding_callback_thread_observer = &client_thread_listner;
```
dalinaum: `client_thread_listner`

realm/realm-library/src/main/cpp/io_realm_SyncManager.cpp
cmelchior: It is documented here: https://developer.android.com/training/articles/perf-jni.html#faq_FindClass

PR #4716, opened 2017-05-26
Add RealmInteger type and simple UTs

bmeike:
All new, with feature branch.

Replaces #4703


realm/realm-library/src/androidTest/java/io/realm/RealmIntegerTests.java
```diff
@@ -0,0 +1,286 @@
...
+import static junit.framework.Assert.assertTrue;
+import static junit.framework.Assert.fail;
+
+@RunWith(AndroidJUnit4.class)
+public class RealmIntegerTests {
+
+    @Rule
+    public final TestRealmConfigurationFactory configFactory = new TestRealmConfigurationFactory();
+    @Rule
+    public final RunInLooperThread looperThread = new RunInLooperThread();
```
zaki50: Maybe we don't need Looper in this test?

realm/realm-library/src/androidTest/java/io/realm/RealmIntegerTests.java
```diff
@@ -0,0 +1,286 @@
...
+
+        assertEquals(6.1488962E18F, ri.floatValue());
+        assertEquals(6.1488959259517348E18, ri.doubleValue());
+
+        // negative
+        ri.set(0x8888444483338281L);
+        assertEquals(0x8888444483338281L, ri.longValue());
+        assertEquals(0x083338281, ri.intValue());
+        assertEquals((short) 0xf8281, ri.shortValue());
+        assertEquals((byte) 0xf81, ri.byteValue());
```
zaki50: How about this? 

```
final long NEGATIVE_VALUE = 0x8888444483338281L;
ri.set(NEGATIVE_VALUE);
assertEquals(NEGATIVE_VALUE, ri.longValue());
assertEquals((int) NEGATIVE_VALUE, ri.intValue());
assertEquals((short) NEGATIVE_VALUE, ri.shortValue());
assertEquals((byte) NEGATIVE_VALUE, ri.byteValue());
```

realm/realm-library/src/androidTest/java/io/realm/RealmIntegerTests.java
```diff
@@ -0,0 +1,286 @@
...
+    @Test
+    public void compareTo() {
+        RealmInteger ri1 = new RealmInteger(10);
+        RealmInteger ri2 = new RealmInteger("11");
+
+        assertEquals(ri1.compareTo(ri2), -1);
+        ri2.decrement(1);
+        assertEquals(ri1.compareTo(ri2), 0);
+        ri2.decrement(1);
+        assertEquals(ri1.compareTo(ri2), 1);
```
zaki50: Incorrect order of the arguments.
Expected value must come first.

realm/realm-library/src/androidTest/java/io/realm/RealmIntegerTests.java
bmeike: We will!

realm/realm-library/src/androidTest/java/io/realm/RealmIntegerTests.java
bmeike: If you can explain to me any way at all that that is clearer, I'll do it in a heartbeat.

realm/realm-library/src/androidTest/java/io/realm/RealmIntegerTests.java
bmeike: Right.  Will fix.

realm/realm-library/src/androidTest/java/io/realm/RealmIntegerTests.java
zaki50: In current code, we need to think about the meaning of each different values.
I think that casting the same value is much easier to understand.

realm/realm-library/src/main/java/io/realm/RealmInteger.java
```diff
@@ -0,0 +1,160 @@
...
+
+    /**
+     * Creates a new realm integer, with the specified initial value.
+     *
+     * @param value initial value: parsed by {@code Long.parseLong}.
+     */
+    public RealmInteger(String value) {
+        this.value = Long.parseLong(value);
+    }
+
```
kneth: No default value (zero maybe)?

bmeike: Not quite sure what you mean.  Java fields are initialized to 0, so `this.value` is 0, until it is set.  That shouldn't matter, though, because `Long.parseLong` throws an (unchecked) exception if it cannot parse its argument.   If you call `new RealmInteger("can't parse me!")` the object will never be constructed, so default values shouldn't be necessary...

realm/realm-library/src/androidTest/java/io/realm/RealmIntegerTests.java
bmeike: Yes, but the test is, exactly, about the meanings of the values.  Your version does not remove the cast.  It simply obfuscates the meaning of the test.  I do not see that as an improvement.

zaki50:
Now this PR contains diffs of unrelated changes. Perhaps we need to merge master  into `feature/counters` branch as well?

bmeike:
gbm/counters is the branch I want to merge to the feature branch.  The feature branch is feature/counters.  feature/counters is, at this point, exactly, master from about a week ago

the branch gbm/counters contains absolutely nothing that is unrelated to counters.  if you see things that you think are unrelated, they are simply the merges from master, necessary to avoid merge conflicts.

bmeike:
Oh... and, btw, I have no permission to merge anything, yet.... even to the feature branch.

bmeike:
Just merged master into the feature branch.  The diffs between this branch and the feature branch are now only the RealmInteger related changes.

PR #4717, opened 2017-05-27
Move benchmarks to separate library

cmelchior:
Trying to add Kotlin to our unit tests moved us above the dex limit. After a little investigation, it turns out that our benchmarks pull in ~30K methods (Apache commons and Guava).

So I wanted to split the benchmark + dependencies into a separate library module under `/realm/benchmarks`, so we could always run the benchmarks against any current version under `/realm/realm-library`. Unfortunately, this turns out not to be possible since the benchmark module has to depend on a specific configuration like e.g `baseDebug`, but after that is configured, Android Studio no longer allows you to change the build variant for the `realm-library` module to anything but `baseDebug`. 

This was kinda useless, so instead, I moved all the benchmark code to a separate top-level project `/library-benchmarks`. This makes it slightly harder to access, but IMO it is acceptable since we rarely run them anyway.

This change brought the current test APK down to ~30K methods which should give us plenty of room for Kotlin + more unit tests.

Current method count
image

With this change 
image


library-benchmarks/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -0,0 +1,6 @@
+#Sat May 27 11:51:11 CEST 2017
+distributionBase=GRADLE_USER_HOME
+distributionPath=wrapper/dists
+zipStoreBase=GRADLE_USER_HOME
+zipStorePath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.0-milestone-1-all.zip
```
zaki50: Shouldn't we use the same Gradle version with other modules (3.5 now?).

library-benchmarks/build.gradle
```diff
@@ -0,0 +1,48 @@
...
+}
+
+dependencies {
+    androidTestCompile 'com.android.support.test:runner:0.5'
+    androidTestCompile 'com.android.support.test:rules:0.5'
+    androidTestCompile 'junit:junit:4.12'
+    androidTestCompile 'dk.ilios:spanner:0.6.0'
+    androidTestCompile 'com.opencsv:opencsv:3.4'
+    androidTestCompile 'junit:junit:4.12'
+}
```
zaki50: We should have wrapper task.

```
allprojects {
    def props = new Properties()
    props.load(new FileInputStream("${rootDir}/../realm.properties"))
    props.each { key, val ->
        project.ext.set(key, val)
    }
}

task wrapper(type: Wrapper) {
    gradleVersion = project.gradleVersion
}
```

library-benchmarks/gradle/wrapper/gradle-wrapper.properties
cmelchior: Yes. Good catch 👍

library-benchmarks/gradlew
```diff
@@ -0,0 +1,160 @@
+#!/usr/bin/env bash
```
zaki50: The mode of this file seems `100644`.
It should be `100755`.

zaki50:
We should have `rootProject.name = 'realm-library-benchmarks'` in `library-benchmarks/settings.gradle` to give `realm-` prefix when the project is imported to Android Studio.

`examples/` project already have this configuration.

PR #4719, opened 2017-05-27
Enable Kotlin for unit tests

cmelchior:
This depends on #4717 being merged first otherwise this will breach the method limit.

This PR adds support for Kotlin in our unit tests and I added a small unit test describing some null weirdness with the schema, but that is mostly proof-of-concept.

The reasoning for this is primarily making sure we got the infrastructure in place to develop any features specifically for Kotlin that might come out of  #4701. 

Some implementation notes:
- Kotlin files are placed in `src/androidTest/kotlin`. This is mostly because anything under there should only test Kotlin features. 
- I added a dependency to `kotlin-reflect`. This adds about ~10K to the method count, but allow easier reference to Class/Field names. Along side the Kotlin standard library this puts the method count at about ~48K


zaki50:
Can we support defining model classes in both Java and Kotlin in a same app?
I don't think we can support it due to the limitation of kapt (kapt handles only kotlin files).

cmelchior:
`kapt` handles both .kt and .java files just fine from the looks of it.

PR #4732, opened 2017-05-31
fix crash when authentication error happens

### LINKED ISSUE #4726: java.util.IllegalFormatConversionException: d != io.realm.ErrorCode error
I used the latest version(3.2.0) of the realm-java.
And I found a bug on the SyncSession.java 327 line.

```java
RealmLog.debug("Session[%s]: Failed to get access token (%d)", configuration.getPath(), response.getError().getErrorCode());

```

The 2nd format of message is "%d", but 2nd argument is ErrorCode enum.
So, a exception was fire at the String.format(message, args).

I hope the bugs will be fixed as soon as possible.

Thanks.

Realm version(s): 3.2.0

Realm sync feature enabled: yes

Android Studio version: 2.3.2

Which Android version and device: Android version 7.0, Samsung Galaxy S8

zaki50:
fixes #4726

@realm/java

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 ### Bug Fixes
 
+* [ObjectServer] Fixed a crash when an authentication error happend (#4726).
+
 ### Internal
 
 
@@ -13,7 +15,7 @@
 
 ### Bug Fixes
 
-* Accept extra columns against synced Realm (#4706).
+* [ObjectServer] Accepted extra columns against synced Realm (#4706).
 
 
 ## 3.3.0 (2017-05-24)
@@ -66,7 +68,7 @@
 
 ### Internal
 
-* Use separated locks for different `RealmCache`s ($4551).
+* Use separated locks for different `RealmCache`s (#4551).
 
 ## 3.1.4 (2017-05-04)
 
--- a/realm/realm-library/src/objectServer/java/io/realm/SyncSession.java
+++ b/realm/realm-library/src/objectServer/java/io/realm/SyncSession.java
@@ -324,7 +324,8 @@ public class SyncSession {
             @Override
             protected void onError(AuthenticateResponse response) {
                 onGoingAccessTokenQuery.set(false);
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -6,14 +6,16 @@
 
 ### Bug Fixes
 
+* [ObjectServer] Fixed a crash when an authentication error happend (#4726).
```
cmelchior: This isn't entirely correct. I would clarify this so "Fixed a crash when debug level logging was enabled and an authentication error happend"

zaki50: Is that true? Message formatting seems to always happen.

realm/realm-library/src/objectServer/java/io/realm/SyncSession.java
```diff
@@ -324,7 +324,8 @@ protected void onSuccess(AuthenticateResponse response) {
             @Override
             protected void onError(AuthenticateResponse response) {
                 onGoingAccessTokenQuery.set(false);
-                RealmLog.debug("Session[%s]: Failed to get access token (%d)", configuration.getPath(), response.getError().getErrorCode());
+                RealmLog.debug("Session[%s]: Failed to get access token (%s)", configuration.getPath(),
+                        response.getError().getErrorCode());
```
kneth: `toString()`?

CHANGELOG.md
cmelchior: The logger should only run the formatter code if that level is enabled, otherwise it will just ignore it. At least that was the intent.

zaki50: Yes, but current code is not.

I created separate issue (#4734).

realm/realm-library/src/objectServer/java/io/realm/SyncSession.java
zaki50: `String.format()` will do it automatically. 

 I think we should not call `toString()` since that call is always executed if we fixed #4734.

CHANGELOG.md
Zhuinden: `happend` => `happened`?

PR #4737, opened 2017-06-01
Add Realm.getDefaultConfiguration()

### LINKED ISSUE #4725: Add Realm.getDefaultConfiguration()
We added `Realm.getInstanceAsync()` and `DynamicRealm.getInstanceAsync()` and those always required `RealmConfiguration` ( or `SyncConfiguration`).

In order to pass default configuration to those methods, it's better to have `Realm.getDefaultConfiguration()`.

zaki50:
closes #4725

### DIFF
--- a/realm/realm-library/src/androidTest/java/io/realm/RealmConfigurationTests.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/RealmConfigurationTests.java
@@ -103,15 +103,10 @@ public class RealmConfigurationTests {
 
     @Test
     public void getDefaultConfiguration_returnsTheSameObjectThatSetDefaultConfigurationSet() {
-        final RealmConfiguration defaultConfiguration = Realm.getDefaultConfiguration();
-        try {
-            final RealmConfiguration config = new RealmConfiguration.Builder().build();
-            Realm.setDefaultConfiguration(config);
+        final RealmConfiguration config = new RealmConfiguration.Builder().build();
+        Realm.setDefaultConfiguration(config);
 
-            assertSame(config, Realm.getDefaultConfiguration());
-        } finally {
-            Realm.setDefaultConfiguration(defaultConfiguration);
-        }
+        assertSame(config, Realm.getDefaultConfiguration());
     }
 
     @Test
@@ -123,7 +118,8 @@ public class RealmConfigurationTests {
             assertNull(Realm.getDefaultConfiguration());
         } finally {
             Realm.setDefaultConfiguration(defaultConfiguration);
-        }}
+        }
+    }
 
     @Test
     public void getInstance_nullConfigThrows() {
@@ -316,16 +312,15 @@ public class RealmConfigurationTests {
 
     @Test
     public void setDefaultConfiguration() {
-        Realm.removeDefaultConfiguration();
         Realm.setDefaultConfiguration(defaultConfig);
         realm = Realm.getDefaultInstance();
-        assertEquals(realm.getPath(), defaultConfig.getPath());
+        assertEquals(defaultConfig, realm.getConfiguration());
     }
 
     @Test
     public void getInstance() {
         realm = Realm.getInstance(defaultConfig);
-        assertEquals(realm.getPath(), defaultConfig.getPath());
+        assertEquals(defaultConfig, realm.getConfiguration());
     }
 
     @Test
--- a/realm/realm-library/src/androidTest/java/io/realm/RealmConfigurationTests.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/RealmConfigurationTests.java
@@ -59,6 +59,7 @@ import static org.junit.Assert.assertNotEquals;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNotSame;
 import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertSame;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 import static org.mockito.Mockito.mock;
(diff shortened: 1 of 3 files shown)


realm/realm-library/src/androidTest/java/io/realm/RealmConfigurationTests.java
```diff
@@ -91,15 +92,8 @@ public void tearDown() throws Exception {
...
-    private void clearDefaultConfiguration() throws NoSuchFieldException, IllegalAccessException {
-        final Field field = Realm.class.getDeclaredField("defaultConfiguration");
-        field.setAccessible(true);
-        field.set(null, null);
-    }
-
     @Test
-    public void setDefaultConfiguration_nullThrows() throws NoSuchFieldException, IllegalAccessException {
-        clearDefaultConfiguration();
+    public void setDefaultConfiguration_nullThrows() {
```
zaki50: I think that there is no need to clear default configuration in this test. Correct?

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -134,6 +133,8 @@
 
     public static final String DEFAULT_REALM_NAME = RealmConfiguration.DEFAULT_REALM_NAME;
 
+    private static final Object monitorForDefaultConfiguration = new Object();
```
beeender: uh... what about name it with `defaultConfigurationLock`?

realm/realm-library/src/androidTest/java/io/realm/RealmConfigurationTests.java
```diff
@@ -108,6 +102,30 @@ public void setDefaultConfiguration_nullThrows() throws NoSuchFieldException, Il
     }
 
     @Test
+    public void getDefaultConfiguration_returnsTheSameObjectThatSetDefaultConfigurationSet() {
+        final RealmConfiguration defaultConfiguration = Realm.getDefaultConfiguration();
```
cmelchior: Doing this implies that a default configuration is usable across multiple unit tests, which definitely shouldn't be the case. Each test should be isolated. So even though it is correct, it just adds noise IMO?

realm/realm-library/src/androidTest/java/io/realm/RealmConfigurationTests.java
```diff
@@ -108,6 +102,30 @@ public void setDefaultConfiguration_nullThrows() throws NoSuchFieldException, Il
...
+    @Test
+    public void getDefaultConfiguration_returnsNullAfterRemoveDefaultConfiguration() {
+        final RealmConfiguration defaultConfiguration = Realm.getDefaultConfiguration();
+        try {
+            Realm.removeDefaultConfiguration();
+
+            assertNull(Realm.getDefaultConfiguration());
+        } finally {
+            Realm.setDefaultConfiguration(defaultConfiguration);
+        }}
```
cmelchior: End bracer should be on a new line

realm/realm-library/src/androidTest/java/io/realm/RealmConfigurationTests.java
```diff
@@ -297,8 +315,8 @@ public void modules() {
     }
 
     @Test
-    public void setDefaultConfiguration() throws NoSuchFieldException, IllegalAccessException {
-        clearDefaultConfiguration();
+    public void setDefaultConfiguration() {
+        Realm.removeDefaultConfiguration();
```
cmelchior: No need to call this when you call `setDefaultConfiguation` right after?

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -325,15 +331,31 @@ public static void setDefaultConfiguration(RealmConfiguration configuration) {
...
-        defaultConfiguration = configuration;
+        synchronized (monitorForDefaultConfiguration) {
+            defaultConfiguration = configuration;
+        }
+    }
+
+    /**
+     * Returns the default configuration for {@link #getDefaultInstance()}.
+     *
+     * @return default configuration object or {@code null} if {@link #init(Context)} was not called yet
```
cmelchior: Smaller issue: As a general contract `Realm.init()` is required to be called before touching any other Realm API, so just throwing in this method if that hasn't been done would also be fine, and that is what most of our other methods do.

realm/realm-library/src/main/java/io/realm/Realm.java
zaki50: It's actually a monitor https://en.wikipedia.org/wiki/Monitor_(synchronization) , but yes we should use the single word for the same things.
I'll fix this.

realm/realm-library/src/androidTest/java/io/realm/RealmConfigurationTests.java
zaki50: OK, I'll remove this.
Any tests should not rely on the default configuration.

realm/realm-library/src/main/java/io/realm/Realm.java
zaki50: You mean we should check if `BaseRealm.applicationContext` is `null` or not in `Realm.getDefaultConfiguration()`?
`defaultConfiguration == null` does not mean that user did not call `Realm.init()` since we provide `Realm.removeDefaultConfiguration()`

cmelchior: Good point. It was just the reference to `Realm.init()` not being called that confused me. I would probably just remove that?

zaki50: Is `@return default configuration object or {@code null}.` fine?

cmelchior: `@return default configuration object or {@code null} if no default configuration is specified.` would be better

zaki50:
all comments are addressed.
I'll merge this once CI is fine.

PR #4744, opened 2017-06-01
Fix bug while downloading server changes

cmelchior:
Fixing bug experienced #4655 and discovered by @nhachicha 

Note this should be covered by this unit test: https://github.com/realm/realm-java/blob/79ca7666432a7a17b0dc29a1dfa7185f27b4f216/realm/realm-library/src/syncIntegrationTest/java/io/realm/SyncedRealmTests.java#L218

And I still need to uncover why this doesn't fail :/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 ### Bug Fixes
 
 * [ObjectServer] Fixed a crash when an authentication error happend (#4726).
+* [ObjectServer] Fixed race condition with `waitForInitialRemoteData` that could result in `RealmMigrationNeededExceptions` to be thrown.
 
 ### Internal
 
--- a/realm/realm-library/src/main/java/io/realm/RealmCache.java
+++ b/realm/realm-library/src/main/java/io/realm/RealmCache.java
@@ -303,6 +303,9 @@ final class RealmCache {
                 if (!fileExists) {
                     try {
                         ObjectServerFacade.getSyncFacadeIfPossible().downloadRemoteChanges(configuration);
+                        // Data is downloaded on background thread, so we need to make sure to advance
+                        // to latest version before continuing.
+                        sharedRealm.refresh();
                     } catch (Throwable t) {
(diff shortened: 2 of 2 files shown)


nhachicha:
@cmelchior here's the failing [test](https://github.com/realm/realm-java/pull/4746/files#diff-f68ea0a094e51b016576cb57154bd6bdR234) (needs `refresh`)

nhachicha:
https://github.com/realm/realm-java/blob/79ca7666432a7a17b0dc29a1dfa7185f27b4f216/realm/realm-library/src/syncIntegrationTest/java/io/realm/SyncedRealmTests.java#L218 doesn't cover the use case exposed by issue since in this unit test you delete the Realm which will cause this [code path](https://github.com/realm/realm-java/blob/79ca7666432a7a17b0dc29a1dfa7185f27b4f216/realm/realm-library/src/main/java/io/realm/RealmCache.java#L303-L306) to redownload all the latest commits from ROS. What you need instead is to keep the Realm on disk, upload more commits to this Realm (via a shared or admin user) then open again.

Here's my [unit test](https://github.com/realm/realm-java/pull/4746/files#diff-f68ea0a094e51b016576cb57154bd6bdR234) that mimic these steps

cmelchior:
Hmm, I'm not sure that is correct. That is the entire point of the test. `waitForInitialRemoteData()` should only be triggered when no file is there. `Session.downloadAllServerChanges()` will not update any Realms automatically though. So if you want the changes available right after it completes, I don't find it unsurprising you need to call `refresh()`. I don't think that has been properly documented though.

zaki50:
@cmelchior Shouldn't this PR target `releases` instead of `master`?

cmelchior:
I made a brain fart. The reason my original unit test is working is because we start a write transaction after downloading the changes which will automatically refresh the SharedRealm. I need to dig deeper into why the Permission Realm tests are failing it seems. 

Closing

PR #4746, opened 2017-06-01
Enable encryption with Sync

### LINKED ISSUE #4561: Enable encryption with Sync

nhachicha:
fixes #4561


nhachicha:
@realm/java

zaki50:
What will happen if an app of old version is already using `encryptionKey()` and the app is updated to fixed version?
Existing Realm database file will be encrypted automatically? Just an exception will be thrown?

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
```diff
@@ -145,7 +145,11 @@ class JniConfigWrapper {
             user, realm_url, SyncSessionStopPolicy::Immediately, std::move(bind_handler), std::move(error_handler),
             nullptr, util::none, sync_client_validate_ssl, ssl_trust_certificate_path});
 
-
+        if (!m_config.encryption_key.empty()) {
```
beeender: What about init the key before construct SyncConfig to make it less confusing?
```c++
        util::Optional<std::array<char, 64>> sync_encryption_key(util::none);
        if (!m_config.encryption_key.empty()) {
            sync_encryption_key = std::array<char, 64>();
            std::copy_n(m_config.encryption_key.begin(), 64, sync_encryption_key->begin());
        }

        m_config.sync_config = std::make_shared<SyncConfig>(SyncConfig{
            user, realm_url, SyncSessionStopPolicy::Immediately, std::move(bind_handler), std::move(error_handler),
            nullptr, sync_encryption_key, sync_client_validate_ssl, ssl_trust_certificate_path});

```

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/EncryptedSynchronizedRealmTests.java
```diff
@@ -0,0 +1,254 @@
...
+
+        // STEP 1: open a synced Realm using a local encryption key
+        String username = UUID.randomUUID().toString();
+        String password = "password";
+        SyncUser user = SyncUser.login(SyncCredentials.usernamePassword(username, password, true), Constants.AUTH_URL);
+
+        final byte[] randomKey = TestHelper.getRandomKey();
+
+        SyncConfiguration configWithEncryption = configurationFactory.createSyncConfigurationBuilder(user, Constants.USER_REALM)
+                .modules(new StringOnlyModule())
```
beeender: Won't `schemas(StringOnly.class)` work? uh... it is not available for `SyncConfiguration.Builder` ... shall we add it?

cmelchior: It should be there, but it is package protected so only works if the test is in `io.realm`

cmelchior:
> What will happen if an app of old version is already using encryptionKey() and the app is updated to fixed version? Existing Realm database file will be encrypted automatically? Just an exception will be thrown?

This is actually kinda problematic. I don't think we are able to do it automatically because we cannot detect if a file is encrypted with the wrong key or not encrypted at all. Also, I think that  `writeEncryptedCopyTo` will write a compacted version, which means the transaction log is lost, making it un-syncable.

Only viable solution would be to make sure that all changes are uploaded, then delete the file and re-download everything again 😢 . We don't have a `Session.uploadAllServerChanges()` either yet.

IMO, we should just merge this as is, because this is really a bug that should be fixed ASAP

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 ### Bug Fixes
 
 * [ObjectServer] Fixed a crash when an authentication error happend (#4726).
+* [ObjectServer] Enable encryption with Sync (#4561).
```
kneth: "Enabled"?

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/EncryptedSynchronizedRealmTests.java
```diff
@@ -0,0 +1,254 @@
...
+public class EncryptedSynchronizedRealmTests extends BaseIntegrationTest {
+    @Rule
+    public Timeout globalTimeout = Timeout.seconds(10);
+
+    @Rule
+    public final TestSyncConfigurationFactory configurationFactory = new TestSyncConfigurationFactory();
+
+    @Before
+    public void before() {
+        // This will set the 'm_metadata_manager' in 'sync_manager.cpp' to be 'null'
```
kneth: Backquote like in line 47?

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/EncryptedSynchronizedRealmTests.java
nhachicha: ☝🏼

beeender: OK, that is good enough

realm/realm-library/src/androidTestObjectServer/java/io/realm/util/SyncTestUtils.java
```diff
@@ -116,4 +129,14 @@ public static AuthenticateResponse createRefreshResponse() {
     public static AuthenticateResponse createErrorResponse(ErrorCode code) {
         return AuthenticateResponse.from(new ObjectServerError(code, "dummy"));
     }
+
+    public static void resetSyncMetadata() {
+        try {
+            SYNC_MANAGER_RESET_METHOD.invoke(null);
+        } catch (InvocationTargetException e) {
```
nhachicha: it requires API 19 unfortunately 
screen shot 2017-06-02 at 12 51 03

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/EncryptedSynchronizedRealmTests.java
```diff
@@ -0,0 +1,254 @@
...
+
+        // STEP 2: make sure the changes gets to the server
+        SystemClock.sleep(TimeUnit.SECONDS.toMillis(2));  // FIXME: Replace with Sync Progress Notifications once available.
+        realm.close(); // Realm is not deleted, just closed
+        user.logout();
+
+        // STEP 3: try to open again the Realm without the encryption key should fail
+        user = SyncUser.login(SyncCredentials.usernamePassword(username, password, false), Constants.AUTH_URL);
+        SyncConfiguration configWithoutEncryption = configurationFactory.createSyncConfigurationBuilder(user, Constants.USER_REALM)
+                .modules(new StringOnlyModule())
```
nhachicha: I want to keep the test a bit structured under `objectserver`

PR #4751, opened 2017-06-02
Fix test case with missing linking object schema

beeender:
Those tests pass with current schema implementation since currently it
will create extra object schemas which are not in the module. This
behaviour is not right and will fail those tests in the future schema
refactor.

So always define all the needed object schemas in the module.

This is needed for #4674


beeender:
@realm/java

kneth:
Yes, object store is actually checking if all classes are included instead of the currently behaviour where missing classes are silently included.

bmeike:
Hey... why did all the comments get deleted from the test?

PR #4764, opened 2017-06-08
Don't strip debug symbols for debug build!

beeender:
From Android Plugin Version 2.2.0, it tries to strip debug symbols of
native libs for all build types (even debug build!!??).
I don't like something try to be smart, so just let our cmake handler
stripping and backup the unstripped libs.

### DIFF
--- a/realm/realm-library/build.gradle
+++ b/realm/realm-library/build.gradle
@@ -92,6 +92,9 @@ android {
     packagingOptions {
         exclude 'META-INF/NOTICE.txt'
         exclude 'META-INF/LICENSE.txt'
+        // We did strip with cmake for release build.
+        // Please, Gradle, you are not that smart! Pleas DO NOT strip debug symbols for debug build!
+        doNotStrip "*/*/*.so"
     }
 
     lintOptions {


beeender:
@nhachicha you will be happy the next time.

realm/realm-library/build.gradle
```diff
@@ -92,6 +92,9 @@ android {
     packagingOptions {
         exclude 'META-INF/NOTICE.txt'
         exclude 'META-INF/LICENSE.txt'
+        // We did strip with cmake for release build.
+        // Please, Gradle, you are not that smart! Pleas DO NOT strip debug symbols for debug build!
+        doNotStrip "*/*/*.so"
```
emanuelez: Wouldn't this avoid stripping also release builds?

cmelchior: I wonder why they kept this. https://google.github.io/android-gradle-dsl/current/com.android.build.gradle.internal.dsl.PackagingOptions.html#com.android.build.gradle.internal.dsl.PackagingOptions:doNotStrip isn't particular helpful on that front. Perhaps useful if not using CMake

beeender: after cmake, the release lib has been already stripped. We did stripping for release even before Android Plug has this smart move. https://github.com/realm/realm-java/blob/master/realm/realm-library/src/main/cpp/CMakeLists.txt#L207

cmelchior: Yes, but we do it from CMake: https://github.com/realm/realm-java/blob/master/realm/realm-library/src/main/cpp/CMakeLists.txt#L207

emanuelez: Oh, so the problem is that Gradle was trying to strip binaries that were already stripped and failing to handle the case where they were already stripped?

PR #4765, opened 2017-06-08
Upgrade to latest Sync 1.9.1 / Core 2.8.0

cmelchior:


cmelchior:
Ready for review @realm/java

dependencies.list
```diff
@@ -1,7 +1,7 @@
 # Realm Sync Core release used by Realm Java
 # https://github.com/realm/realm-sync/releases
-REALM_SYNC_VERSION=1.8.5
-REALM_SYNC_SHA256=71e70f83b1604672bbfd7c04fcf984ecc8ac683864943ef3de262994c2e7b7fc
+REALM_SYNC_VERSION=1.9.1
```
nhachicha: OS commit below reference a dependency to `1.9.0`

dependencies.list
```diff
@@ -10,4 +10,4 @@ REALM_SYNC_SHA256=71e70f83b1604672bbfd7c04fcf984ecc8ac683864943ef3de262994c2e7b7
 # /tools/sync_test_server/Dockerfile specify which repo (apt) we should
 # install/use between 'realm' and 'realm-testing', the version below should
 # correspond to an existing version on the *specified* repo.
-REALM_OBJECT_SERVER_DE_VERSION=1.6.0-35
+REALM_OBJECT_SERVER_DE_VERSION=1.7.5-180
```
nhachicha: carriage return removed

dependencies.list
nhachicha: OS commit below reference the `1.9.0`

tools/sync_test_server/Dockerfile
```diff
@@ -5,8 +5,8 @@ ARG ROS_DE_VERSION
 # Add realm repo
 RUN apt-get update -qq \
     && apt-get install -y curl npm \
-    # && curl -s https://packagecloud.io/install/repositories/realm/realm/script.deb.sh \
-    && curl -s https://packagecloud.io/install/repositories/realm/realm-testing/script.deb.sh | bash
+    && curl -s https://packagecloud.io/install/repositories/realm/realm/script.deb.sh | bash
```
nhachicha: were you testing with stable build of ROS?

beeender:
Oh, it should go to master since it is not compatible with previous ROS. So 👍

tools/sync_test_server/Dockerfile
cmelchior: Yes. Testing releases are not getting correctly deployed, so the stable release is the latest available one.

dependencies.list
cmelchior: I bumped the commit to the latest possible. It seems that OS was never built against 1.9.1, but there is no breaking changes so It should be okay.

PR #4770, opened 2017-06-09
Clean up the Proxies a bit, in prep for RealmInteger mods

bmeike:
I would like to be able to read the proxy code easily, so that I can make the (possibly somewhat complicated) changes for `RealmInteger`.  There should be absolutely no functional changes here: just syntax.

### DIFF
--- a/realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmProxyClassGenerator.java
+++ b/realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmProxyClassGenerator.java
@@ -25,6 +25,7 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.EnumSet;
+import java.util.List;
 import java.util.Set;
 
 import javax.annotation.processing.ProcessingEnvironment;
@@ -37,6 +38,37 @@ import javax.tools.JavaFileObject;
 
 public class RealmProxyClassGenerator {
     private static final String BACKLINKS_FIELD_EXTENSION = "Backlinks";
+
+    private static final List<String> IMPORTS;
+    static {
+        List<String> l = Arrays.asList(
+            "android.annotation.TargetApi",
+            "android.os.Build",
+            "android.util.JsonReader",
+            "android.util.JsonToken",
+            "io.realm.exceptions.RealmMigrationNeededException",
+            "io.realm.internal.ColumnInfo",
+            "io.realm.internal.LinkView",
+            "io.realm.internal.OsObject",
+            "io.realm.internal.RealmObjectProxy",
+            "io.realm.internal.Row",
+            "io.realm.internal.SharedRealm",
+            "io.realm.internal.Table",
+            "io.realm.internal.android.JsonUtils",
+            "io.realm.log.RealmLog",
+            "java.io.IOException",
+            "java.util.ArrayList",
+            "java.util.Collections",
+            "java.util.List",
+            "java.util.Iterator",
+            "java.util.Date",
+            "java.util.Map",
+            "java.util.HashMap",
+            "org.json.JSONObject",
+            "org.json.JSONException",
+            "org.json.JSONArray");
+        IMPORTS = Collections.unmodifiableList(l);
+    }
 
     private final ProcessingEnvironment processingEnvironment;
     private final ClassMetaData metadata;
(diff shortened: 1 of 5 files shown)


realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmProxyClassGenerator.java
```diff
@@ -65,43 +97,16 @@ public void generate() throws IOException, UnsupportedOperationException {
...
-        imports.add("org.json.JSONObject");
-        imports.add("org.json.JSONException");
-        imports.add("org.json.JSONArray");
-
         writer.emitImports(imports)
                 .emitEmptyLine();
 
         // Begin the class definition
-        writer.beginType(
+        writer.emitAnnotation("SuppressWarnings(\"all\")")
```
zaki50: This change should be discussed in another PR.

realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmProxyClassGenerator.java
```diff
@@ -599,7 +604,10 @@ private void emitCreateRealmObjectSchemaMethod(JavaWriter writer) throws IOExcep
                 EnumSet.of(Modifier.PUBLIC, Modifier.STATIC), // Modifiers
                 "RealmSchema", "realmSchema"); // Argument type & argument name
 
-        writer.beginControlFlow("if (!realmSchema.contains(\"" + this.simpleClassName + "\"))");
+        writer.beginControlFlow("if (realmSchema.contains(\"%s\"))", this.simpleClassName)
```
zaki50: I like this change! :+1:

realm/realm-annotations-processor/src/test/resources/io/realm/AllTypesRealmProxy.java
```diff
@@ -399,26 +398,26 @@ protected final void copy(ColumnInfo rawSrc, ColumnInfo rawDst) {
     }
 
     public static RealmObjectSchema createRealmObjectSchema(RealmSchema realmSchema) {
+        if (realmSchema.contains("AllTypes")) {
```
cmelchior: Any reason to add this? `createRealmObjectSchema` is only called if the Realm file is unversioned (=just created). So this will always be false? https://github.com/realm/realm-java/blob/master/realm/realm-library/src/main/java/io/realm/Realm.java#L461

realm/realm-annotations-processor/src/test/resources/io/realm/AllTypesRealmProxy.java
```diff
@@ -399,26 +398,26 @@ protected final void copy(ColumnInfo rawSrc, ColumnInfo rawDst) {
...
-                AllTypesRealmProxy.createRealmObjectSchema(realmSchema);
-            }
-            realmObjectSchema.add("columnRealmList", RealmFieldType.LIST, realmSchema.get("AllTypes"));
-            return realmObjectSchema;
-        }
-        return realmSchema.get("AllTypes");
+            AllTypesRealmProxy.createRealmObjectSchema(realmSchema);
+        }
+        realmObjectSchema.add("columnObject", RealmFieldType.OBJECT, realmSchema.get("AllTypes"));
+        if (!realmSchema.contains("AllTypes")) {
```
cmelchior: If you keep the if above, couldn't this just be removed then?

realm/realm-annotations-processor/src/test/resources/io/realm/AllTypesRealmProxy.java
kneth: Yes, if the cyclic model classes, and `createRealmObjectSchema()` calls itself. @beeender has a PR which changes that.

realm/realm-annotations-processor/src/test/resources/io/realm/AllTypesRealmProxy.java
kneth: You are tricked into thinking this since `AllTypes` has a field of type `AllTypes`. Of course, we can detect this in the proxy generator, but I can not sure it is worth doing. `createRealmObjectSchema()` is only called when initializing. Optimizing it and have a more complex proxy generator seems to be the wrong way to go.

realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmProxyClassGenerator.java
```diff
@@ -1225,10 +1230,12 @@ private void emitInsertListMethod(JavaWriter writer) throws IOException {
 
         writer.beginControlFlow("while (objects.hasNext())")
                 .emitStatement("object = (%s) objects.next()", qualifiedClassName);
-        writer.beginControlFlow("if(!cache.containsKey(object))");
+        writer.beginControlFlow("if (cache.containsKey(object))")
```
bmeike: I really don't know.  My guess would be that this makes not perceptible difference in execution time.  It almost certainly shaves a few bytes off the size of the executable, though.

I agree that it would be nice to have metrics for time to construct the proxy and time to run common proxy operations.

bmeike:
Moved the SupressWarnings to a separate PR.  Can I get an "amen"?

realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmProxyClassGenerator.java
kneth: You could compare results from our micro benchmark.

cmelchior: I would be very surprised if this change was detectable even by the micro benchmarks. Anything below ~20% change is hidden by variance

PR #4775, opened 2017-06-12
Update kotlin to 1.1.2-5

zaki50:
Updated kotlin to `1.1.2-5`.
And now example project uses `kotlin-stdlib-jre7` instead of `kotlin-stdlib`.

Chnagelog of kotlin 1.1.2-5 is https://github.com/JetBrains/kotlin/blob/1.1.2/ChangeLog.md#112-5


zaki50:
Official documentation says that the `kotlin-stdlib-jre7` 


https://github.com/JetBrains/kotlin/tree/master/libraries/stdlib/jre7/src/kotlin

PR #4778, opened 2017-06-13
Add RealmObject.getRealm() and DynamicRealmObject.getDynamicRealm().

### LINKED ISSUE #4720: get Realm instance  from RealmObject
#### Goal

To get Realm instance from  RealmObject inheriting object when it is managed.
It was possible in the previous version ( e.g. 0.87.2).

#### Use Case

Our application have something like news feed.
The feed item has several kind of information.
In the new feed list view, almost same information is displayed for all kind of item.
When an user tap the feed list item in the feed list view,
a kind specific detail feed information is displayed in addition to common feed item information.

I am realizing following data structure on Realm 0.87.2.

realm_meet_up__9____speaker_deck

`FeedItem` class is a realm object for common feed list item information,
A realm object for kind specific detail feed information is defined for each kind,
and implements `FeedItemDetail` interface.

To realize polymorphism like behaviour which is not supported by realm object,
I add `getFeedDetail()` method with `FeedItemDetail` return type to `FeedItem` class with `@ignored` annotation. 
In the `getFeedDetail()`, I get Realm instance and find feed detail object instance with realm query.
It was possible, because I could get Realm from RealmObject.
So I want make RealmObject to get realm instance.

Sample codes are like this.

```
public interface FeedItemDetail{
   /* ... */
}

public class CardUpdateDetail extends RealmObject implements FeedItemDetail{

    @PrimaryKey
    private String feedId;

    /* ... */
}

public class SharedLinkDetail extends RealmObject implements FeedItemDetail{

    @PrimaryKey
    private String feedId;

    /* ... */
}

public class FeedItem extends RealmObject{

    /** kind of feed item */
    private int kind;
    @Ignored
    private FeedItemDetail feedItemDetail;

    FeedItemDetail getFeedItemDetail(){
         if(this.feedItemDetail != null){
            return this.feedItemDetail;
        }
        switch(this.kind){
            case FEED_ITEM_KIND_CARD_UPDATE:
                this.feedItemDetail = getRealm().where(CardUpdateDetail::class).equals("id", id).first()
                return this.feedItemDetail;
            case FEED_ITEM_KIND_SHARED_LINK:
                this.feedItemDetail = getRealm().where(SharedLinkDetail::class).equals("id", id).first()
                return this.feedItemDetail;
            case .....
        }
    }
}
```

zaki50:
closes #4720 

In the discussion of #4720, we found some problems in regard to `Realm` and `DynamicRealm`.
To avoid that problem, I implemented `RealmObject.getConfiguration()` instead.

Now I only add `public static RealmConfiguration getConfiguration(RealmModel)`.
Should we have non static version `public RealmConfiguration getConfiguration()` as well?


realm/realm-library/src/androidTest/java/io/realm/RealmObjectTests.java
```diff
@@ -1923,4 +1925,103 @@ public void execute(Realm realm) {
...
+        AllTypes object = realm.createObject(AllTypes.class);
+        realm.commitTransaction();
+
+        realm.close();
+        realm = null;
+
+        RealmObject.getConfiguration(object);
+    }
+
+    @Test(expected = IllegalStateException.class)
```
bmeike: We generate IllegalState and IllegalArgument exceptions for all manner of things.  I think all of these tests could be made more robust by checking to be sure that the Exception that is thrown is the one that you expect.

zaki50: will do

zaki50: done

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 * Added support for querying inverse relationships (#2904).
 * Moved inverse relationships out of beta stage.
 * Added `Realm.getDefaultConfiguration()` (#4725).
+* Added `RealmObject.getConfiguration(RealmModel)` (#4720).
```
cmelchior: We definitely need to duplicate the methods across the `RealmObject` base class and the static methods. People should be able to use both interchangeably.

Adding it to the `RealmObject` base class could be potentially breaking though if people have their own `getConfiguration()` method. For that reason alone, I'm not sure it is worth sneaking this change into a minor release, and we should probably wait for 4.0

Also, I'm still trying to convince myself that this is truly useful compared to exposing `RealmObject.getRealm()`.

```
public foo(Person p) {
  Realm realm = Realm.getInstance(p.getConfiguration());
  realm.executeTransaction({ p.deleteFromRealm(); });
  realm.close();
}

vs.

public foo(Person p) {
  Realm realm = p.getRealm();
  realm.executeTransaction({ p.deleteFromRealm(); });
}
```

I understand the argument that exposing a `getRealm()` method makes it harder to reason about the Realm lifecycle + we don't have to battle the `DynamicRealm/Realm` issues, but it still feels wrong to me somehow. It is like requiring people to always do one extra step of work, only because we want them to not accidentally call `realm.close()`. What are your thoughts on this @realm/java?

realm/realm-library/src/androidTest/java/io/realm/RealmObjectTests.java
```diff
@@ -1923,4 +1925,129 @@ public void execute(Realm realm) {
...
+            assertEquals(RealmObject.MSG_NOT_MANAGED_OBJECT, e.getMessage());
+        }
+    }
+
+    @Test
+    public void getRealmConfiguration_unmanagedRealmModelThrows() {
+        try {
+            RealmObject.getConfiguration(new AllTypesRealmModel());
+            fail();
+        } catch (IllegalArgumentException e) {
```
cmelchior: Is this too harsh? Should we just return `null` in this case, e.g `isValid()` returns `true` for unmanaged objects instead of crashing for example?

realm/realm-library/src/main/java/io/realm/RealmObject.java
```diff
@@ -68,6 +68,9 @@
 
 @RealmClass
 public abstract class RealmObject implements RealmModel, ManagableObject {
+    static final String MSG_NULL_OBJECT = "'model' is null.";
```
cmelchior: Any reason for these? They are only used in one place?

zaki50: This constant is used in test to check if the thrown exception is expected or not.

https://github.com/realm/realm-java/pull/4778#discussion_r121644858

realm/realm-library/src/main/java/io/realm/RealmObject.java
```diff
@@ -692,16 +691,16 @@ public final void removeAllChangeListeners() {
...
+     * @throws IllegalArgumentException if the {@code model} is {@code null}.
      * @throws IllegalStateException if the model is deleted, the corresponding Realm of the model is closed or in an incorrect thread.
      */
     public static RealmConfiguration getConfiguration(RealmModel model) {
         if (model == null) {
             throw new IllegalArgumentException(MSG_NULL_OBJECT);
         }
         if (!(model instanceof RealmObjectProxy)) {
-            throw new IllegalArgumentException(MSG_NOT_MANAGED_OBJECT);
+            return null;
```
Zhuinden: @cmelchior are you sure this is a good idea?

cmelchior: No, but I'm not sure the other one is a good idea either. This at lest is more consistent with the behavior of our other methods: `isValid/isLoaded` that return values for unmanaged objects instead of crashing.

realm/realm-library/src/main/java/io/realm/RealmObject.java
```diff
@@ -287,6 +289,40 @@ public boolean isManaged() {
     }
 
     /**
+     * Gets {@link RealmConfiguration} from this object.
+     *
+     * @return {@link RealmConfiguration} of the Realm where this object belongs to or {@code null} if unmanaged.
+     * @throws IllegalStateException if this object was already deleted from Realm or the corresponding Realm of this object was already closed.
+     */
+    public RealmConfiguration getConfiguration() {
```
nhachicha: will this return `SyncConfiguration` as well? can we add a test?

realm/realm-library/src/main/java/io/realm/RealmObject.java
```diff
@@ -287,6 +289,40 @@ public boolean isManaged() {
...
+     * @param model managed Realm object.
+     * @return {@link RealmConfiguration} of the Realm where the {@code model} belongs to or {@code null} if the {@code model} is unmanaged.
+     * @throws IllegalArgumentException if the {@code model} is {@code null}.
+     * @throws IllegalStateException if the model was deleted or the corresponding Realm of the model was already closed.
+     */
+    public static RealmConfiguration getConfiguration(RealmModel model) {
+        if (model == null) {
+            throw new IllegalArgumentException(MSG_NULL_OBJECT);
+        }
+        if (!(model instanceof RealmObjectProxy)) {
```
nhachicha: this check is done inside `isValid` why not doing 
```java
if (model == null) {
             throw new IllegalArgumentException(MSG_NULL_OBJECT);
 }

if (!RealmObject.isValid(model)) {
           throw new IllegalStateException("Unmanaged or deleted object");
}
final BaseRealm realm = ((RealmObjectProxy) model).realmGet$proxyState().getRealm$realm();
realm.checkIfValid();
return realm.getConfiguration();

```

zaki50: We need to return `null` when the object is managed even if the object is valid.

zaki50:
As a result of an internal discussion, we will provide `RealmObject.getRealm()` and `DynamicRealm#getDynamicRealm()` instead of `getConfiguration()`.

zaki50:
I've updated the PR.

@realm/java

realm/realm-library/src/androidTest/java/io/realm/DynamicRealmObjectTests.java
```diff
@@ -1259,4 +1260,102 @@ public void testExceptionMessage() {
...
+        final CountDownLatch threadFinished = new CountDownLatch(1);
+        final AtomicReference<Throwable> throwable = new AtomicReference<>();
+        final Thread thread = new Thread(new Runnable() {
+            @Override
+            public void run() {
+                try {
+                    object.getDynamicRealm();
+                    fail();
+                } catch (Throwable t) {
+                    throwable.set(t);
```
beeender: Why don't you just assert here to check the exception message? the failed assertion should fail the test for JUnit4.

zaki50: Doesn't that require one more try-catch to set thrown `AssertionFailureError` to `throwable`?

PR #4779, opened 2017-06-13
Suppress warnings in generated code.

bmeike:
There is no reason that generated code, in production, should ever generate a warning.

We may decide for any of a number of reasons, that we wish to generated code that would otherwise generate a warning.  We may also decide to pay attention to some warnings, an eliminate them.

The user, who can *never* do anything about warnings in the generated code, though, should never, ever see them.

### DIFF
--- a/realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmProcessor.java
+++ b/realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmProcessor.java
@@ -124,7 +124,7 @@ import io.realm.annotations.RealmModule;
         "io.realm.annotations.RealmModule",
         "io.realm.annotations.Required"
 })
-@SupportedOptions(value = {"debug"})
+@SupportedOptions(value = {"realm.suppressWarnings"})
 public class RealmProcessor extends AbstractProcessor {
 
     // Don't consume annotations. This allows 3rd party annotation processors to run.
--- a/realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmProcessor.java
+++ b/realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmProcessor.java
@@ -25,6 +25,7 @@ import java.util.Set;
 import javax.annotation.processing.AbstractProcessor;
 import javax.annotation.processing.RoundEnvironment;
 import javax.annotation.processing.SupportedAnnotationTypes;
+import javax.annotation.processing.SupportedOptions;
 import javax.lang.model.SourceVersion;
 import javax.lang.model.element.Element;
 import javax.lang.model.element.ElementKind;
@@ -123,7 +124,9 @@ import io.realm.annotations.RealmModule;
         "io.realm.annotations.RealmModule",
         "io.realm.annotations.Required"
 })
+@SupportedOptions(value = {"debug"})
 public class RealmProcessor extends AbstractProcessor {
+
     // Don't consume annotations. This allows 3rd party annotation processors to run.
     private static final boolean CONSUME_ANNOTATIONS = false;
 
--- a/realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmProxyClassGenerator.java
+++ b/realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmProxyClassGenerator.java
@@ -36,6 +36,7 @@ import javax.tools.JavaFileObject;
 
 
 public class RealmProxyClassGenerator {
+    private static final String OPTION_DEBUG = "debug";
     private static final String BACKLINKS_FIELD_EXTENSION = "Backlinks";
 
     private final ProcessingEnvironment processingEnvironment;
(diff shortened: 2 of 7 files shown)


zaki50:
How about introducing this as an opt-in feature?
E.g. in `build.gradle`, 

```
realm {
    suppressWarnings = true
}
```
I'm not sure it is possible though.

bmeike:
I like something like what @zaki50  is suggesting.  I think it should be the other sense, though.  Unless you ask for warnings on the generated code, you don't get them.

cmelchior:
```
realm {
    suppressWarnings = true
}
```

Would require some way of sending options to an annotation processor. I don't know if that is possible. I kinda like the idea that you have to enable warnings instead of disabling them.

Zhuinden:
@cmelchior there's a pretty ugly way that GreenRobot/EventBus does which passes an argument to their annotation processor at http://greenrobot.org/eventbus/documentation/subscriber-index/

``` groovy
android {
    defaultConfig {
        javaCompileOptions {
            annotationProcessorOptions {
                arguments = [ eventBusIndex : 'com.example.myapp.MyEventBusIndex' ]
            }
        }
    }
}
 
dependencies {
    compile 'org.greenrobot:eventbus:3.0.0'
    annotationProcessor 'org.greenrobot:eventbus-annotation-processor:3.0.1'
}
```

bmeike:
Implemented it this evening.  Will commit tomorrow.

realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmProcessor.java
```diff
@@ -123,7 +124,9 @@
         "io.realm.annotations.RealmModule",
         "io.realm.annotations.Required"
 })
+@SupportedOptions(value = {"debug"})
```
zaki50: I think we can use `OPTION_DEBUG` constant instead of hard coded value `"debug"` like `@SupportedOptions(value = {RealmProxyClassGenerator.OPTION_DEBUG})`.

realm/realm-library/build.gradle
```diff
@@ -65,6 +65,16 @@ android {
                 }
             }
         }
+
+        buildTypes {
+            debug {
+                javaCompileOptions {
+                    annotationProcessorOptions {
+                        arguments = [ debug : 'true' ]
```
cmelchior: Would `suppressWarnings` with a value of `false` be more readable? I feel a bit uneasy about using the word `debug` since it implies a lot more. At least in my mind?

bmeike:
Added compile time control of the generation of the @SupressWarnings at the head of the synthetic proxy class.  Annotation is added everywhere except the the proxies for the `realm-library` unit tests.

realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmProxyClassGenerator.java
```diff
@@ -36,6 +36,7 @@
 
 
 public class RealmProxyClassGenerator {
+    private static final String OPTION_DEBUG = "debug";
```
zaki50: I think we should have more Realm specific name for annotation processor option name since it is seen by all annotation processors.

How about adding `realm.` prefix?

realm/realm-library/build.gradle
zaki50: If `arguments` is a dictionary, I think that it's better to just add an entry instead of replacing dictionary itself.
I guess there is a chance that other plugins already setting their options.

realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmProcessor.java
bmeike: Nah.  That's just excess baggage.

realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmProxyClassGenerator.java
bmeike: Yep.  Good call.

realm/realm-library/build.gradle
bmeike: That's a good call too!  Will do.

bmeike: kk.  Renaming as `realm.suppressWarnings`.  Default is `true`

realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmProcessor.java
```diff
@@ -124,7 +124,7 @@
         "io.realm.annotations.RealmModule",
         "io.realm.annotations.Required"
 })
-@SupportedOptions(value = {"debug"})
+@SupportedOptions(value = {"realm.suppressWarnings"})
```
Zhuinden: I think this value could use the constant defined in `RealmProcessor` if that is made package visibility.

realm/realm-library/build.gradle
```diff
@@ -70,7 +70,7 @@ android {
             debug {
                 javaCompileOptions {
                     annotationProcessorOptions {
-                        arguments = [ debug : 'true' ]
+                        arguments += [ 'realm.suppressWarnings' : 'false' ]
```
Zhuinden: Good fix

Issue #4788, opened 2017-06-15
Case.INSENSITIVE query broken at 3.1.3


51systems:
It appears that case-insensitive queries are broken in some cases.

It appears that if you have a field with the same string in different cases, querying on the field in a case-insensitive manner returns no results, rather than the expected two results.

This behaviour was working in 3.1.2 but broke in 3.1.3. I suspect this may have been due to a change to realm core. It is still broken as of 3.3.2.

See the Repo here with a failing test:
https://github.com/51systems/RealmBugs

The following test fails:

```java
realm.beginTransaction();
Dog dog1 = realm.copyToRealm(Dog.create("ROVER"));
Dog dog2 = realm.copyToRealm(Dog.create("Rover"));
realm.commitTransaction();

assertThat(realm.where(Dog.class).equalTo("name", "ROVER", Case.SENSITIVE).findAll(),
        containsInAnyOrder(dog1));

assertThat(realm.where(Dog.class).equalTo("name", "Rover", Case.SENSITIVE).findAll(),
        containsInAnyOrder(dog2));

//Fails here
assertThat(realm.where(Dog.class).equalTo("name", "rover", Case.INSENSITIVE).findAll(),
        containsInAnyOrder(dog1, dog2));
```

51systems:
Update:
removing `@Index` from the `Dog.name` field allows the test to pass.

Zhuinden:
You might be correct, there was an optimization a while ago that ended up in a deadlock when the string contained only numbers and the search was case-insensitive on an indexed field, maybe this is also a side-effect of said optimization and is most likely a bug, see https://github.com/realm/realm-core/pull/2578#issuecomment-297844618

@danielpovlsen 

Nice find by the way

cmelchior:
Hi @51systems 
Thank you very much for the example project. We can confirm that something definitely isn't right. We are looking into it.

Zhuinden:
https://github.com/realm/realm-core/pull/2675 was merged but has to be released in core

Zhuinden:
Realm Core 2.8.6 was released but Realm-Java has to make a release where Core version is increased.

Zhuinden:
Realm-Java 3.5.0 is released that is updated to Core 2.8.6 and should fix the indexed case insensitive query results.

beeender:
@Zhuinden you are fast ....
This is fixed in `3.5.0`.

PR #4791, opened 2017-06-15
Expose bug when querying an indexed field with Case.INSENSITIVE

cmelchior:
Exposes #4788 
Waiting for https://github.com/realm/realm-core/pull/2675 to be fixed


Zhuinden:
It is merged in core

Zhuinden:
Realm Core 2.8.6 is released where this is pretty much the only update.

Zhuinden:
If this is merged against the latest vers then this should pass?

beeender:
@Zhuinden thanks! this change has been merged with another PR already.

PR #4792, opened 2017-06-15
Clean object schema tests

beeender:


realm/realm-library/src/androidTest/java/io/realm/entities/StringOnly.java
```diff
@@ -20,6 +20,7 @@
 
 public class StringOnly extends RealmObject {
```
bmeike: Should this be `StringOnly.class.getSimpleName()`?

beeender: That doesn't work with proguard if we have a way to enable it in the testing.
Also, it is kind of convention we are using in the testing models.

PR #4794, opened 2017-06-15
BLOB's default value when convert to not-nullable

beeender:
"" is a char* type takes 1 byte of memory.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 * [ObjectServer] Fixed a crash when an authentication error happens (#4726).
 * [ObjectServer] Enabled encryption with Sync (#4561).
 * [ObjectServer] Admin users did not connect correctly to the server (#4750).
+* When converting nullable BLOB field to required, `null` values should be converted to `byte[0]` instead of `byte[1]`.
 
 ### Internal
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 3.3.2 (2017-06-09)
+## 3.3.3 (YYYY-MM-DD)
 
 ### Breaking Changes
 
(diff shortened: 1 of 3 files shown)


realm/realm-library/src/androidTest/java/io/realm/RealmObjectSchemaTests.java
```diff
@@ -507,6 +506,47 @@ public void setRemoveRequired() {
...
+                    // Skip not-nullable fields .
+                    if (!fieldType.isNullable()) {
+                        break;
+                    }
+                    schema.addField(fieldName, fieldType.getType());
+                    DynamicRealmObject object = realm.createObject(schema.getClassName());
+                    assertTrue(object.isNull(fieldName));
+                    schema.setRequired(fieldName, true);
+                    assertFalse(object.isNull(fieldName));
+                    if (fieldType == FieldType.BLOB) {
```
kneth: funny to see all these `else if` within a `switch` on the same variable 😄

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 * [ObjectServer] Fixed a crash when an authentication error happens (#4726).
 * [ObjectServer] Enabled encryption with Sync (#4561).
 * [ObjectServer] Admin users did not connect correctly to the server (#4750).
+* When converting nullable BLOB field to required, `null` values should be converted to `byte[0]` instead of `byte[1]`.
```
kneth: Good catch!

cmelchior: Seems rewriting code from time to time is useful 😛

cmelchior: This should be part of the 3.3.3 changelog not 3.3.2

beeender: very, and I am fixing another one~~ we should do rewriting more often!

PR #4803, opened 2017-06-18
Correctly reset integration tests.

cmelchior:
According to @mrackwitz, we need to delete the `realm-object-server` directory between each integration test to correctly reset ROS ( https://github.com/realm/realm-object-server/issues/1225#issuecomment-308216359 ) .

This error caused #4655 to be extremely flaky. After this change, it seems to run consistently.

### DIFF
--- a/tools/sync_test_server/ros-testing-server.js
+++ b/tools/sync_test_server/ros-testing-server.js
@@ -3,6 +3,7 @@
 var winston = require('winston'); //logging
 const temp = require('temp');
 const spawn = require('child_process').spawn;
+const exec = require('child_process').exec;
 var http = require('http');
 var dispatcher = require('httpdispatcher');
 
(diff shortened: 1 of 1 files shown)


tools/sync_test_server/ros-testing-server.js
```diff
@@ -64,10 +65,16 @@ function stopRealmObjectServer() {
     if (syncServerChildProcess) {
         syncServerChildProcess.kill();
         syncServerChildProcess = null;
+        exec('rm -r ' + 'realm-object-server', function (err, stdout, stderr) {
```
beeender: really? It is not creating this dir in `--root path`? That sounds incorrect .....

cmelchior: We need to do this according to @mrackwitz . I agree It does feel rather odd though since `--path` is defined as `--root ROOT_PATH`.

mrackwitz: This is a known issue, see https://github.com/realm/realm-js/issues/706.

PR #4798, opened 2017-06-16
Converting nullable PK fields with null values

beeender:
- RealmObjectSchema.setRequired() will throw if the PK field has null
  values stored.
- Call set_xxx_unique when converting nullability on a PK field.
- Remove useless check in the proxy generator which caused inconsistency
  exception.
- Add relevant test cases.

PR #4810, opened 2017-06-19
Revert "Converting nullable PK fields with null values (#4798)"

beeender:
This reverts commit 80d3c1f80c4e070356e106ec2722976e11e651ba.

Try if this one caused lots of CI failures.


beeender:
this is not the cause of the CI failur. Close it.

PR #4812, opened 2017-06-19
Removed unused methods from internal Table class

zaki50:
This PR migrated all codes using `Table#addEmptyRow()` to `OsObject.create*()` in order to remove `Table#Table()`, `Table#addEmptyRow()`, `Table#addEmptyRows(long)`, `Table#add(Object...)`, `Table#pivot()` and `Table#createNative()`.

And this PR also removed `SharedRealm` argument from `OsObject.create*(SharedRealm, Table, ...)` methods since Table always contains the same `SharedRealm` instance in it.


realm/realm-library/src/main/cpp/io_realm_internal_OsObject.cpp
```diff
@@ -162,18 +162,22 @@ static void finalize_object(jlong ptr)
 
 static inline size_t do_create_row(jlong shared_realm_ptr, jlong table_ptr)
 {
-    auto& shared_realm = *(reinterpret_cast<SharedRealm*>(shared_realm_ptr));
     auto& table = *(reinterpret_cast<realm::Table*>(table_ptr));
-    shared_realm->verify_in_write();
+    if (shared_realm_ptr) {
```
beeender: `SharedRealm.nativePtr` is final, and `close()` won't free the memory. `shared_realm_ptr` should never be `null` i think. This check is redundant.

zaki50: You are correct, but `SharedRealm` in `Table` can be `null` and in that case, `0` is passed to this function as a `shared_realm_ptr`.

https://github.com/realm/realm-java/pull/4812/files#diff-cb8c4bdde24368ab202c6405c065a7fdR168

That case only happens when the `Table` is created in our unit tests without `SharedRealm`.
We will be able to remove this check after retirement of old `JNI*Test`.

beeender: right ... but this function could be on a very hot path (the bulk insertion for example.), add the check to support test cases may not worth it.

I think i have the same problem before, what I did is create a `SharedRealm` for the relevant JNI tests before.

IMO, some JNI tests could be just removed, like the `PivotTest.java`, some should be moved to upper layer tests. For those which are difficult to move upper, keep it in the JNI tests and create a `SharedRealm` instead. Support `Table` without `SharedRealm` doesn't have too much value for java binding.

realm/realm-library/src/main/java/io/realm/internal/Table.java
```diff
@@ -1026,7 +868,7 @@ public static String getTableNameForClass(String name) {
         return TABLE_PREFIX + name;
     }
 
-    protected native long createNative();
+    static native long nativeCreate();
```
beeender: I think no one is using this anymore? can we remove it?

realm/realm-library/src/main/cpp/io_realm_internal_OsObject.cpp
```diff
@@ -204,10 +204,10 @@ static inline size_t do_create_row_with_primary_key(JNIEnv* env, jlong shared_re
 static inline size_t do_create_row_with_primary_key(JNIEnv* env, jlong shared_realm_ptr, jlong table_ptr,
                                                     jlong pk_column_ndx, jstring pk_value)
 {
-    auto& shared_realm = *(reinterpret_cast<SharedRealm*>(shared_realm_ptr));
+    auto& shared_realm = *(reinterpret_cast<SharedRealm *>(shared_realm_ptr));
```
beeender: space

realm/realm-library/src/main/cpp/io_realm_internal_OsObject.cpp
```diff
@@ -171,7 +171,7 @@ static inline size_t do_create_row(jlong shared_realm_ptr, jlong table_ptr)
 static inline size_t do_create_row_with_primary_key(JNIEnv* env, jlong shared_realm_ptr, jlong table_ptr,
                                                     jlong pk_column_ndx, jlong pk_value, jboolean is_pk_null)
 {
-    auto& shared_realm = *(reinterpret_cast<SharedRealm*>(shared_realm_ptr));
+    auto& shared_realm = *(reinterpret_cast<SharedRealm *>(shared_realm_ptr));
```
beeender: space

realm/realm-library/src/androidTest/java/io/realm/internal/TableIndexAndDistinctTest.java
```diff
@@ -85,26 +132,37 @@ public void shouldTestIndexOnWrongColumnType(Long index) {
         t.addSearchIndex(index);
     }*/
 
-    public void testShouldCheckIndexIsOkOnColumn() {
+    @Test
+    public void shouldCheckIndexIsOkOnColumn() {
         init();
+        sharedRealm.beginTransaction();
```
beeender: can we call `beginTransaction()` in the `setUp()`  and `cancelTransaction()` in the `tearDown()`?

CHANGELOG.md
```diff
@@ -12,6 +12,9 @@
 
 ### Internal
 
+* Removed `Table#addEmptyRow()`, `Table#addEmptyRows()`, `Table#add(Object...)` and `Table#pivot(long,long,PivotType)`.
```
beeender: and the `Table#Table()`?

realm/realm-library/src/main/java/io/realm/internal/Table.java
zaki50: Good catch!
`Table.pivot()` was the only user of this method, but I removed it!

realm/realm-library/src/androidTest/java/io/realm/internal/TableIndexAndDistinctTest.java
zaki50: Nice idea. Will try.

realm/realm-library/src/androidTest/java/io/realm/TestHelper.java
```diff
@@ -106,22 +107,165 @@ public static RealmFieldType getColumnType(Object o) {
...
+            Object value = values[(int) columnIndex];
+            switch (colTypes[(int) columnIndex]) {
+                case BOOLEAN:
+                    table.setBoolean(columnIndex, rowIndex, (Boolean) value, false);
+                    break;
+                case INTEGER:
+                    if (value == null) {
+                        table.setNull(columnIndex, rowIndex, false);
+                    } else {
+                        long intValue = ((Number) value).longValue();
```
cmelchior: a bit strange to have this called `intValue` ( I assume you mean the core INTEGER, but still). IMO `longValue` would be better.

realm/realm-library/src/androidTest/java/io/realm/TestHelper.java
```diff
@@ -106,22 +107,165 @@ public static RealmFieldType getColumnType(Object o) {
...
+                        table.setLong(columnIndex, rowIndex, intValue, false);
+                    }
+                    break;
+                case FLOAT:
+                    table.setFloat(columnIndex, rowIndex, (Float) value, false);
```
cmelchior: Don't we support nullable floats?

realm/realm-library/src/androidTest/java/io/realm/TestHelper.java
```diff
@@ -106,22 +107,165 @@ public static RealmFieldType getColumnType(Object o) {
...
+                    break;
+                case DOUBLE:
+                    table.setDouble(columnIndex, rowIndex, (Double) value, false);
```
cmelchior: Same for doubles

realm/realm-library/src/androidTest/java/io/realm/TestHelper.java
```diff
@@ -106,22 +107,165 @@ public static RealmFieldType getColumnType(Object o) {
...
+                    break;
+                case STRING:
+                    if (value == null) {
+                        table.setNull(columnIndex, rowIndex, false);
+                    } else {
+                        table.setString(columnIndex, rowIndex, (String) value, false);
+                    }
+                    break;
+                case DATE:
+                    table.setDate(columnIndex, rowIndex, (Date) value, false);
```
cmelchior: Same for date

realm/realm-library/src/androidTest/java/io/realm/TestHelper.java
zaki50: I just copied this method from `Table#add(...)`.

I'll double check the implementation of the `Table#setFloat()`.

zaki50: Our proxy generates the following code for nullable Float field.
That means we should use `setNull()` for float.

I'll fix it.

zaki50:
Thanks. I'll merge this if the CI will be happy with the latest change.

PR #4814, opened 2017-06-19
Use fail() instead of assert(false) in tests

zaki50:
`assert(false)` checks nothing...

### DIFF
--- a/realm/realm-library/src/androidTest/java/io/realm/internal/JNIQueryTest.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/internal/JNIQueryTest.java
@@ -444,57 +444,60 @@ public class JNIQueryTest extends TestCase {
 
         // Compares strings in non string columns.
         for (int i = 0; i <= 6; i++) {
-            try { query.equalTo(new long[]{i}, oneNullTable, "string");                 assert(false); } catch(IllegalArgumentException e) {}
-            try { query.notEqualTo(new long[]{i}, oneNullTable, "string");              assert(false); } catch(IllegalArgumentException e) {}
-            try { query.beginsWith(new long[]{i}, oneNullTable, "string");              assert(false); } catch(IllegalArgumentException e) {}
-            try { query.endsWith(new long[]{i}, oneNullTable, "string");                assert(false); } catch(IllegalArgumentException e) {}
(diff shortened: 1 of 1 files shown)


realm/realm-library/src/androidTest/java/io/realm/internal/JNIQueryTest.java
```diff
@@ -444,70 +444,73 @@ public void testQueryWithWrongDataType() {
...
         // Compares strings in non string columns.
         for (int i = 0; i <= 6; i++) {
-            try { query.equalTo(new long[]{i}, oneNullTable, "string");                 assert(false); } catch(IllegalArgumentException e) {}
-            try { query.notEqualTo(new long[]{i}, oneNullTable, "string");              assert(false); } catch(IllegalArgumentException e) {}
-            try { query.beginsWith(new long[]{i}, oneNullTable, "string");              assert(false); } catch(IllegalArgumentException e) {}
-            try { query.endsWith(new long[]{i}, oneNullTable, "string");                assert(false); } catch(IllegalArgumentException e) {}
-            try { query.like(new long[]{i}, oneNullTable, "string");                    assert(false); } catch(IllegalArgumentException e) {}
-            try { query.contains(new long[]{i}, oneNullTable, "string");                assert(false); } catch(IllegalArgumentException e) {}
+            if (i != 6) {
+                try { query.equalTo(new long[]{i}, oneNullTable, "string");    fail(); } catch(IllegalArgumentException ignore) {}
```
beeender: `ignore` should be `expected` to suppress the AS lint warnings.

zaki50: `expected` doesn't suppress `Empty 'catch' block` warnings on my environment (AS 3.0 Preview4).

beeender: uh... what?? that is shocking ... then we need to change all the `expected` to `ignore`???

zaki50: Seems yes, or something wrong in my environment?

beeender: `ignore` is fine with my AS 2.3.3 too! I must remember something wrong ... 👍

PR #4830, opened 2017-06-22
Local ref needs to be cleaned on client thread

beeender:
- Add support for null JavaLocalRef.
- Clean the local ref after notifyAllChangesDownloaded.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@
 
 ### Bug Fixes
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -14,6 +14,8 @@
 
 ### Bug Fixes
 
+* [ObjectServer] Fixed a bug which may crash when the JNI local reference limitation reached on sync client thread.
```
cmelchior: was reached on the Sync Client thread.

PR #4731, opened 2017-05-31
Add ApplyPermissions method

cmelchior:
Part of #4558 
Adds `applyPermissions()` to PermissionManager.

This adds 3 new classes:
- `PermissionRequest` for wrapping request parameters. Mostly to avoid a 4-level parameter method which is considered bad practice. Clean Code by Bob Martin recommends 3 at most.
- `UserCondition` for creating the "query" that matches users on the server. Can be extended with more functionality in the future.
- AccessLevel to wrap the `mayRead/mayWrite/mayManage` levels in a sensible manner.

The above matches the approaches taken by Cocoa and .NET expect for the addition of the `PermissionRequest` class

TODO 
- [x] Redirect to #4558 once #4655 is merged
- [x] Add integration unit tests
- [x] Javadoc for applyPermission
- [x] Split out PR into smaller chuncks

PR #4842, opened 2017-06-23
PermissionManger should not return same instance for different users

cmelchior:
Discovered this embarrassing bug while writing unit tests for #4731 ... apparently caching is hard.

I modified the cache to be per user.

### DIFF
--- a/realm/realm-library/src/objectServer/java/io/realm/PermissionManager.java
+++ b/realm/realm-library/src/objectServer/java/io/realm/PermissionManager.java
@@ -62,8 +62,12 @@ public class PermissionManager implements Closeable {
             String userId = syncUser.getIdentity();
             ThreadLocal<Cache> threadLocalCache = cache.get(userId);
             if (threadLocalCache == null) {
-                threadLocalCache = new ThreadLocal<>();
-                threadLocalCache.set(new Cache());
+                threadLocalCache = new ThreadLocal<Cache>() {
+                    @Override
+                    protected Cache initialValue() {
+                        return new Cache();
+                    }
+                };
                 cache.put(userId, threadLocalCache);
             }
             Cache c = threadLocalCache.get();
@@ -322,7 +326,7 @@ public class PermissionManager implements Closeable {
         }
         closed = true;
     }
-    
+
     /**
      * Checks if this PermissionManager is closed or not. If it is closed, all methods will report back an error.
      *
--- a/realm/realm-library/src/objectServer/java/io/realm/PermissionManager.java
+++ b/realm/realm-library/src/objectServer/java/io/realm/PermissionManager.java
@@ -24,8 +24,10 @@ import java.net.URISyntaxException;
 import java.net.URL;
 import java.util.ArrayList;
 import java.util.Deque;
+import java.util.HashMap;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.Map;
 
 import io.realm.internal.permissions.ManagementModule;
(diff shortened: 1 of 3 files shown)


realm/realm-library/src/objectServer/java/io/realm/PermissionManager.java
```diff
@@ -40,34 +42,37 @@
...
+    static PermissionManager getInstance(SyncUser syncUser) {
+        synchronized (cacheLock) {
+            String userId = syncUser.getIdentity();
+            ThreadLocal<Cache> threadLocalCache = cache.get(userId);
+            if (threadLocalCache == null) {
+                threadLocalCache = new ThreadLocal<>();
+                threadLocalCache.set(new Cache());
+                cache.put(userId, threadLocalCache);
+            }
+            Cache c = threadLocalCache.get();
```
zaki50: `c` is `null` if the `getInstance()` is called other threads than one allocated `threadLocalCache`.
We should override `ThreadLocal#initialValue()` instead of setting initial value by calling `set()`.

realm/realm-library/src/objectServer/java/io/realm/PermissionManager.java
```diff
@@ -315,7 +322,7 @@ public void close() {
         }
         closed = true;
     }
-
+
```
zaki50: spaces

realm/realm-library/src/objectServer/java/io/realm/SyncUser.java
```diff
@@ -531,6 +531,7 @@ private static String getManagementRealmUrl(URL authUrl) {
      * @return
      */
     public PermissionManager getPermissionManager() {
+        new AndroidCapabilities().checkCanDeliverNotification("The PermissionManager can only opened on a Looper thread.");
```
zaki50: Commit log and PR description does not have a mention about this change.
Is it an intentional change?

cmelchior: This is being merged to a feature branch which is still WIP, so no changelog needed (yet)

realm/realm-library/src/objectServer/java/io/realm/PermissionManager.java
cmelchior: There is a null check just above that prevent it from ever being `null` here. I'll fix the `initialValue`

zaki50: In old code, you checked only `threadLocalCache` but also `c` need to be checked since you allocated `Cache` object only for the first thread.

But, current code is fine! `initialValue()` is executed on every thread.

cmelchior: Ah yes. Good point 👍

realm/realm-library/src/objectServer/java/io/realm/PermissionManager.java
```diff
@@ -40,34 +42,41 @@
 public class PermissionManager implements Closeable {
 
     // Reference counted cache equivalent to how Realm instances work.
-    private static ThreadLocal<PermissionManager> permissionManager = new ThreadLocal<PermissionManager>() {
-        @Override
-        protected PermissionManager initialValue() {
-            return null;
-        }
-    };
+    private static Map<String, ThreadLocal<Cache>> cache = new HashMap<>();
```
beeender: The lock can be removed if it is `ThreadLocal<Map<String, cache>>`. But this is fine to me.

PR #4843, opened 2017-06-23
PermissionWrapper unit test helper methods

cmelchior:
This PR adds a few helper methods required for the PermissionManager unit tests in #4731

### DIFF
--- a/realm/realm-library/src/androidTest/java/io/realm/TestHelper.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/TestHelper.java
@@ -157,7 +157,7 @@ public class TestHelper {
     public static String getRandomEmail() {
         StringBuilder sb = new StringBuilder(UUID.randomUUID().toString().toLowerCase());
         sb.append('@');
-        sb.append("realm.io");
+        sb.append("androidtest.realm.io");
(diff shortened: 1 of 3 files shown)


realm/realm-library/src/androidTest/java/io/realm/TestHelper.java
```diff
@@ -153,6 +154,13 @@ public static InputStream stringToStream(String str) {
         return key;
     }
 
+    public static String getRandomEmail() {
+        StringBuilder sb = new StringBuilder(UUID.randomUUID().toString().toLowerCase());
+        sb.append('@');
+        sb.append("realm.io");
```
zaki50: How about using a domain something like `javatest.realm.io`?

realm/realm-library/src/androidTest/java/io/realm/TestHelper.java
```diff
@@ -153,6 +154,13 @@ public static InputStream stringToStream(String str) {
         return key;
     }
 
+    public static String getRandomEmail() {
```
zaki50: This method does not seem to be used in this commit.
Do we actually need this?

PR #4847, opened 2017-06-26
Explicitly specify Locale for String.format() instead of implicit current Locale

zaki50:
Added explicit Locale information to follow `lint`s advice.

### DIFF
--- a/realm/realm-library/src/objectServer/java/io/realm/internal/network/AuthenticateResponse.java
+++ b/realm/realm-library/src/objectServer/java/io/realm/internal/network/AuthenticateResponse.java
@@ -20,6 +20,7 @@ import org.json.JSONException;
 import org.json.JSONObject;
 
 import java.io.IOException;
+import java.util.Locale;
 
 import io.realm.ErrorCode;
 import io.realm.ObjectServerError;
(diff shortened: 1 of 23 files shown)


realm/realm-library/src/objectServer/java/io/realm/internal/network/AuthenticateResponse.java
```diff
@@ -132,14 +133,14 @@ private AuthenticateResponse(String serverResponse) {
             if (accessToken == null) {
                 message = "accessToken = null";
             } else {
-                message = String.format("Identity %s; Path %s", accessToken.identity(), accessToken.path());
+                message = String.format(Locale.ENGLISH, "Identity %s; Path %s", accessToken.identity(), accessToken.path());
```
beeender: Will this still work if `accessToken.path()` return a non-english string? like CJK?

kneth:
Like @beeender I am a bit concerned how it will work with non-English letters. Even model classes can be in non-English - a Greek developer might use `σκύλοι` instead of `dog`.

realm/realm-library/src/objectServer/java/io/realm/internal/network/AuthenticateResponse.java
zaki50: I tested following test case to make sure Strings are formatted as expected.

```
@Test
public void formatNonAsciiString() {
    final String testee = String.format(Locale.ENGLISH,
            "format %s, %s and %s.", "σκύλοι", "領域", "ангелы");
    assertEquals("format σκύλοι, 領域 and ангелы.", testee);
}
```
After testing, I realized that the Android documentation recommends to use `Locale.US`.
https://developer.android.com/reference/java/util/Locale.html#default_locale

Should I update this PR to use `Locale.US`?

zaki50:
I've pushed a commit to use `Locale.US` instead of `Locale.ENGLISH`.
https://developer.android.com/reference/java/util/Locale.html#default_locale

zaki50:
@beeender @kneth 
I tested following test case, I didn't commit the test though.

```
@Test
public void formatNonAsciiString() {
    final String testee = String.format(Locale.US,
            "format %s, %s and %s.", "σκύλοι", "領域", "ангелы");
    assertEquals("format σκύλοι, 領域 and ангелы.", testee);
}
```

Zhuinden:
I wonder how well a Hungarian class name would break this.

zaki50:
@Zhuinden Thanks. Model class names are mostly used in annotation processor.
This PR only contains changes for library code.
I'll make another PR for annotation processor and transformer.

I'll add that test case in the PR.

PR #4850, opened 2017-06-26
PermissionWrapper - GetDefaultPermissions

cmelchior:
Part of #4558 

This exposes the default permissions for the ROS. Ideally, this should have been part of `getPermissions()`, but architectural decisions on the server side makes this impossible for the time being.


cmelchior:
Ready for review @realm/java

realm/realm-library/src/objectServer/java/io/realm/PermissionManager.java
```diff
@@ -183,6 +212,20 @@ public RealmAsyncTask getPermissions(final Callback<RealmResults<Permission>> ca
         return addTask(new GetPermissionsAsyncTask(this, callback));
     }
 
+    /**
+     * Returns default permissions for all Realms. The default permissions are the ones that will be used if no
+     * more user specific permissions is in effect.
```
kneth: remove "more"

realm/realm-library/src/objectServer/java/io/realm/PermissionManager.java
```diff
@@ -451,28 +556,43 @@ protected final boolean checkAndReportInvalidState() {
...
+            if (managementErrorHappened && !permissionErrorHappened && !defaultPermissionErrorHappened) {
                 notifyCallbackError(managementError);
                 return true;
             }
 
-            if (permissionErrorHappened && managementErrorHappened) {
-                notifyCallbackError(combineRealmErrors(managementError, permissionError));
+            if (defaultPermissionErrorHappened && !permissionErrorHappened && !managementErrorHappened) {
+                notifyCallbackError(defaultPermissionError);
+            }
```
zaki50: missing `return true;`?

realm/realm-library/src/objectServer/java/io/realm/PermissionManager.java
```diff
@@ -451,28 +556,43 @@ protected final boolean checkAndReportInvalidState() {
...
+            ObjectServerError defaultPermissionError;
             synchronized (permissionManager.errorLock) {
                 // Only hold lock while making a safe copy of current error state
                 managementErrorHappened = (permissionManager.managementRealmError != null);
                 permissionErrorHappened = (permissionManager.permissionRealmError != null);
+                defaultPermissionErrorHappened = (permissionManager.defaultPermissionsRealmError != null);
                 managementError = permissionManager.managementRealmError;
                 permissionError = permissionManager.permissionRealmError;
+                defaultPermissionError = permissionManager.defaultPermissionsRealmError;
             }
```
zaki50: I think it's better to have following early return here to avoid `LinkedHashMap` allocation on no errors.

```
if (!permissionErrorHappened && !managementErrorHappened && !defaultPermissionErrorHappened) {
    // Everything seems valid
    return false;
}
```

realm/realm-library/src/objectServer/java/io/realm/PermissionManager.java
beeender: lines 573 - 586 don't seems to be necessary. I think `combineRealmErrors()` could handle the single error case well.

realm/realm-library/src/objectServer/java/io/realm/PermissionManager.java
```diff
@@ -80,34 +81,44 @@ protected Cache initialValue() {
...
+            return globalRealm;
+        }
     }
 
     private final SyncUser user;
 
     // Used to track the lifecycle of the PermissionManager
     private RealmAsyncTask managementRealmOpenTask;
     private RealmAsyncTask permissionRealmOpenTask;
+    private RealmAsyncTask defaultPermissionsRealmOpenTask;
```
dalinaum: Other `~RealmOpenTask`s are singular and only `defaultPermissionsRealmOpenTask` is plural. Other `RealmOpenTask`s should be plural or `defaultPermissionsRealmOpenTask` should also be singular?

dalinaum: If it should be singular, `defaultPermissionsRealmConfig` and `defaultPermissionsRealm` should also be singular.

realm/realm-library/src/syncIntegrationTest/java/io/realm/PermissionManagerTests.java
```diff
@@ -159,6 +160,80 @@ public void onError(ObjectServerError error) {
 
     @Test
     @RunTestInLooperThread
+    public void getDefaultPermissions_returnLoadedResults() {
+        PermissionManager pm = user.getPermissionManager();
+        looperThread.closeAfterTest(pm);
+        pm.getPermissions(new PermissionManager.Callback<RealmResults<Permission>>() {
```
dalinaum: I think if this test's name is correct, it should be `pm.getDefaultPermissions`. Otherwise, its test's name should be `getPermissions_returnLoadedResults`.

realm/realm-library/src/objectServer/java/io/realm/PermissionManager.java
```diff
@@ -183,6 +212,20 @@ public RealmAsyncTask getPermissions(final Callback<RealmResults<Permission>> ca
         return addTask(new GetPermissionsAsyncTask(this, callback));
     }
 
+    /**
+     * Returns default permissions for all Realms. The default permissions are the ones that will be used if no
+     * more user specific permissions is in effect.
+     *
+     * @param callback callback notified when the permissions are ready. The returned {@link RealmResults} is a fully
+     * live query result, that will be auto-updated like any other {@link RealmResults}.
```
dalinaum: This line should be indented.

realm/realm-library/src/objectServer/java/io/realm/PermissionManager.java
```diff
@@ -123,10 +134,12 @@ public String getName() {
     private final Object errorLock = new Object();
     private volatile ObjectServerError permissionRealmError = null;
     private volatile ObjectServerError managementRealmError = null;
+    private volatile ObjectServerError globalPermissionRealmError = null;
 
     // Cached result of the permission query. This will be filled, once the first PermissionAsyncTask has loaded
     // the result.
-    private RealmResults<Permission> permissions;
+    private RealmResults<Permission> userPermissions;
+    private RealmResults<Permission> defaultPermissions;
```
beeender: don't you want to renaming this to `globalPermissions`? Actually i am bit confused about the reason why renaming the `default` to `global.

cmelchior: 👍  mostly because the Realm on the server is a global Realm, shared between all users. `defaultPermissionsRealm` still sounded like it was unique to each user, but I don't have strong feelings either way. So if you prefer `defaultPermissionsRealm` I can rename it back?

beeender: I feel no reason to have two different names internally, if we expose it as `defaultPermission`, i would like to use the unique name `default` here. So, please, rename it back 🥇

cmelchior: Will do 👍

realm/realm-library/src/objectServer/java/io/realm/PermissionManager.java
```diff
@@ -183,6 +212,20 @@ public RealmAsyncTask getPermissions(final Callback<RealmResults<Permission>> ca
         return addTask(new GetPermissionsAsyncTask(this, callback));
     }
 
+    /**
+     * Returns default permissions for all Realms. The default permissions are the ones that will be used if no
+     * user specific permissions is in effect.
+     *
+     * @param callback callback notified when the permissions are ready. The returned {@link RealmResults} is a fully
+     * live query result, that will be auto-updated like any other {@link RealmResults}.
```
dalinaum: Should this line be indented?

cmelchior: The automatic code formatter we are using are not doing this. So no.

realm/realm-library/src/objectServer/java/io/realm/PermissionManager.java
```diff
@@ -615,31 +717,47 @@ protected final void notifyCallbackError(ObjectServerError e) {
...
+                errorMsg.append(entry.getKey());
+                errorMsg.append('\n');
+                errorMsg.append(entry.getValue().toString());
+            }
             return errorMsg.toString();
         }
+
+        private ErrorCode combineErrorCodes(Map<String, ObjectServerError> errors) {
+            ErrorCode finalErrorCode = null;
+            for (ObjectServerError error : errors.values()) {
```
kneth: You could also check if `errors` has more than one element.

cmelchior: That doesn't catch the case where all Realms failed with the same error code. Which is actually likely if someone was wrong with the server.

PR #4853, opened 2017-06-27
Explicitly specify Locale for String.format() in our annotation processor

zaki50:
Explicitly specify Locale for String.format() in our annotation processor.
And also added a test for non-latin model name and field names.


realm/realm-annotations-processor/src/main/java/io/realm/processor/Backlink.java
```diff
@@ -129,7 +131,7 @@ public String getSimpleSourceClass() {
     public boolean validateSource() {
         // A @LinkingObjects cannot be @Required
         if (backlink.getAnnotation(Required.class) != null) {
-            Utils.error(String.format(
+            Utils.error(String.format(Locale.US,
```
dalinaum: How about adding new line before `LOCALE.US,`?

zaki50: done

PR #4862, opened 2017-06-27
fixes #4822

### LINKED ISSUE #4822: IllegalArgumentException: Could not parse user json
#### Goal

I'm trying to create a synced realm database

#### Expected Results

That the database should not crash even if it failed to login

#### Actual Results

SyncUser.currentUser throws an exception, crashing the whole application and it's not possible to delete that invalid user.
The error is something like:
Caused by:
```
 java.lang.IllegalArgumentException: Could not parse user json: eyJpZGVudGl0eSI6IjQ0ZWU5Y2ZlNGMwNDc5YTVmZDU4MWMwOTQ5OTI5Y2YxIiwiZXhwaXJlcyI6MTgxMzM1NDkzNCwiYXBwX2lkIjoiY29tLmxvc3RwaXhlbHMuZmllbGRzZXJ2aWNlIiwiYWNjZXNzIjpbInJlZnJlc2giXSwiaXNfYWRtaW4iOmZhbHNlLCJzYWx0IjowLjY1OTEwMjg3OTYyMzExNTN9:P1KRa4IykFyVWdd9CUt84jytP7r+O3X8P7bgwtXjJBrgbwLtf0Fz8+ezfKP01G5mjumsjx9bc4ZCY+JkaEOjSu1U+kDck0Oi7fOgvhpf2cpPjPryrEMF/XXkNq8EBzUfumdNRsxoBEJhNsGcOui/UO8kEI1y0qlxUdjVwEupqv62BPlz4AzGCC27MWsKUACDLJUekC/pdMTRYcrWEVTfmZXZHrm90QtXwmJVSUKePmwWhE5xTvDQKw49bT1FJzX45RhRZez3pP2SAYpwNH8IVpnMqN1M4+OruwTt//eRwO8VQdPJKgXLk3T2Cx1AWnjL+AkuW0dcKBmbjJt3Cs3v1g==
                                                                                 at io.realm.SyncUser.fromJson(SyncUser.java:157)
                                                                                 at io.realm.RealmFileUserStore.toSyncUserOrNull(RealmFileUserStore.java:84)
                                                                                 at io.realm.RealmFileUserStore.getCurrent(RealmFileUserStore.java:44)
                                                                                 at io.realm.SyncUser.currentUser(SyncUser.java:110)

```

I first start with an offline realm database. Then I create a synced user with this code
```
GoogleSignInResult result = Auth.GoogleSignInApi.getSignInResultFromIntent(data);
        if (result.isSuccess()) {
            GoogleSignInAccount acct = result.getSignInAccount();
            String token = acct.getIdToken();
            SyncCredentials myCredentials = SyncCredentials.google(token);
            String authURL = serverUrl + "/auth";
            return SyncUser.login(myCredentials, authURL);
        }
```

I then merge the offline and online database

```
if (user != null && user.isValid()) {
            RealmConfiguration configuration = getSyncConfig(user);
            Realm offlineRealm = Realm.getDefaultInstance();
            Realm.setDefaultConfiguration(configuration);
            Realm onlineRealm = Realm.getDefaultInstance();

            // Merge
            try {
                onlineRealm.refresh();
                // Code for merging offline database to online database
            } finally {
                offlineRealm.close();
                onlineRealm.close();
            }
        }
```

If an exception occurs I catch it and logout that user
```
SyncUser user = SyncUser.currentUser();
        if (user != null && user.isValid()) {
            user.logout();
        }
```

Now if an exception occur, then next time the app is started SyncUser.currentUser() will throw the above exception and it's not possible to use the app until the data of the app is cleared.

A local Realm object is a member of the activity and is closed after login/merge is done and set to default instance again to now point to correct realm database.

Is it possible to clear the invalid user without deleting all data? What can cause this to happen? Encryption is not applied to the databases. The realm server is running locally on my Mac. I can't reproduce the problem every time, but 1 out of 3 times when I try, I get this problem and needs to clear app data to be able to use the app again.

Realm version(s): 3.3.2

Realm sync feature enabled: yes

Android Studio version: 2.3

Which Android version and device: Emulator, android API 24

nhachicha:
- `refresh_token` was saved instead of the entire `SyncUser` JSON representation, causing some use case to cause a parsing error
-  Now calling `SyncUser#logout` will impact other instances of the same user to be `invalid`. This, for example, will prevent the creation of `SyncConfiguration` with a user already logged out (using `SyncUser#currentUser`)
- Adding more integration tests to cover edge cases


beeender:
The CI failure looks relevant to this change. @nhachicha

realm/realm-library/src/objectServer/java/io/realm/SyncUser.java
```diff
@@ -456,7 +456,7 @@ public String toJson() {
      */
     public boolean isValid() {
         Token userToken = getSyncUser().getUserToken();
-        return syncUser.isLoggedIn() && userToken != null && userToken.expiresMs() > System.currentTimeMillis();
+        return syncUser.isLoggedIn() && SyncManager.getUserStore().isActive(syncUser.getIdentity()) && userToken != null && userToken.expiresMs() > System.currentTimeMillis();
```
cmelchior: Doesn't the native `isActive` also take into account loggedin/expireDate? If so, this statement can be greatly reduced.

realm/realm-library/src/objectServer/java/io/realm/UserStore.java
```diff
@@ -68,4 +68,12 @@
...
      */
     Collection<SyncUser> allUsers();
+
+    /**
+     * Returns the state of the specified user: {@code true} if active (not logged out), {@code false} otherwise.
+     *
+     * @param identity identity of the user.
+     * @return {@code true} if the user is not logged out, {@code false} otherwise.
+     */
+    boolean isActive(String identity);
```
cmelchior: Wouldn't it be better for this to accept a `SyncUser`? This is the abstraction that people normally work with? (Also the rest of the methods in the interface)

realm/realm-library/src/objectServer/java/io/realm/internal/SyncObjectServerFacade.java
```diff
@@ -86,11 +87,17 @@ public void realmClosed(RealmConfiguration configuration) {
...
             SyncConfiguration syncConfig = (SyncConfiguration) config;
-            String rosServerUrl = syncConfig.getServerUrl().toString();
-            String rosUserIdentity = syncConfig.getUser().getIdentity();
-            String syncRealmAuthUrl = syncConfig.getUser().getAuthenticationUrl().toString();
-            String rosRefreshToken = syncConfig.getUser().getAccessToken().value();
-            return new Object[]{rosUserIdentity, rosServerUrl, syncRealmAuthUrl, rosRefreshToken, syncConfig.syncClientValidateSsl(), syncConfig.getServerCertificateFilePath()};
+            // make sure the user is still valid
+            SyncUser user = syncConfig.getUser();
+            if (!user.isValid()) {
+                throw new IllegalStateException("SyncUser is no longer valid, probably logged out.");
```
cmelchior: I think this is problematic, because I think this will cause `getInstance(syncConfig)` to throw if the user is no longer considered valid. Right now we only do this check when creating the `SyncConfiguration` because there you can do something about it. When opening the Realm instance it is too late, and we should just open it and let people work offline IMO.

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/AuthTests.java
```diff
@@ -266,4 +270,130 @@ public void changePassword_throwWhenUserIsLoggedOut() {
         user.changePassword("new-password");
     }
 
+    // Cached instances of RealmConfiguration should not be allowed to be used if the user is no longer valid
+    @Test
+    public void cachedInstanceShouldThrowIfUserBecomeInvalid() throws InterruptedException {
```
cmelchior: This very much depends on why the user becomes invalid. If it only happens if you manually call `logout()`, this is OK, but if it also happens if the user token just "expires", then this is problematic.

realm/realm-library/src/objectServer/java/io/realm/SyncUser.java
nhachicha: no this just checks if the ObjectStore user is marked for logout or not. OS doesn't check token validity

realm/realm-library/src/objectServer/java/io/realm/UserStore.java
nhachicha: we can but what about `get(String identity);`?

cmelchior: Ah yeah, both `get/remove()` uses `identity`. This is probably fine then.

realm/realm-library/src/objectServer/java/io/realm/internal/SyncObjectServerFacade.java
nhachicha: the issue is when the `refresh_token` is not usable the user need to login again, this is different from the use case where the `access_token` expires & we can wait for the client to refresh it.

btw this will throw only when creating the first instance in the current thread, when returning a "thread cached" instance we don't do any check. 

I can add a warning/error log instead if you feel strongly against.

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/AuthTests.java
nhachicha: as stated previously will add a logging instead, if you think it's too much to throw ...

realm/realm-library/src/objectServer/java/io/realm/internal/SyncObjectServerFacade.java
cmelchior: By your description it sounds like this only happens if the user manually called `SyncUser.logout()`, correct? If that is the case, this if fine IMO 👍

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/AuthTests.java
cmelchior: If this can only happen if the user manually have called `logout()`, throwing is fine 👍

nhachicha:
@realm/java

realm/realm-library/src/main/cpp/io_realm_RealmFileUserStore.cpp
```diff
@@ -85,6 +85,20 @@ JNIEXPORT void JNICALL Java_io_realm_RealmFileUserStore_nativeLogoutUser(JNIEnv*
     CATCH_STD()
 }
 
+JNIEXPORT jboolean JNICALL Java_io_realm_RealmFileUserStore_nativeIsActive(JNIEnv* env, jclass, jstring identity)
+{
+    TR_ENTER()
+    try {
+        JStringAccessor id(env, identity); // throws
```
beeender: `identity` -> `j_identity` , `id` -> `identity`

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/AuthTests.java
```diff
@@ -266,4 +270,130 @@ public void changePassword_throwWhenUserIsLoggedOut() {
...
+        final CountDownLatch backgroundThread = new CountDownLatch(1);
+        final IllegalStateException[] backgroundException = new IllegalStateException[1];
+        // Should throw when using the invalid configuration form a different thread
+        new Thread() {
+            @Override
+            public void run() {
+                try {
+                    Realm.getInstance(configuration);
+                } catch (IllegalStateException expected) {
+                    backgroundException[0] = expected;
```
beeender: you can assert here, `assertNotNull(backgroundException[0]);`. JUnit4 will fail the test when assertion fails in the bg thread.

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/AuthTests.java
```diff
@@ -266,4 +270,130 @@ public void changePassword_throwWhenUserIsLoggedOut() {
...
+        try {
+            // We should not be able to build a configuration with an invalid/logged out user
+            new SyncConfiguration.Builder(currentUser, Constants.USER_REALM).build();
+            fail("Invalid currentUser, it should not be possible to create a SyncConfiguration");
+        } catch (IllegalArgumentException expected) {
+            // User not authenticated or authentication expired.
+        }
+    }
+
+    // using a logout user should
```
beeender: should **throw**

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/AuthTests.java
```diff
@@ -266,4 +270,130 @@ public void changePassword_throwWhenUserIsLoggedOut() {
...
+    // Cached instances of RealmConfiguration should not be allowed to be used if the user is no longer valid
+    @Test
+    public void cachedInstanceShouldThrowIfUserBecomeInvalid() throws InterruptedException {
+        String username = UUID.randomUUID().toString();
+        String password = "password";
+
+        SyncCredentials credentials = SyncCredentials.usernamePassword(username, password, true);
+        SyncUser user = SyncUser.login(credentials, Constants.AUTH_URL);
+        final RealmConfiguration configuration = new SyncConfiguration.Builder(user, Constants.USER_REALM).build();
+        Realm.getInstance(configuration);
```
beeender: This instance needs to be closed, otherwise assertion will happen if you build with debug sync.

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/AuthTests.java
```diff
@@ -266,4 +270,130 @@ public void changePassword_throwWhenUserIsLoggedOut() {
...
+        }.start();
+
+        backgroundThread.await();
+        assertNotNull(backgroundException[0]);
+
+        // it is ok to return the cached instance, since this use case is legit
+        // user refresh token can timeout, or the token can be revoked from ROS
+        // while running the Realm instance. So it doesn't make sense to break this behaviour
+        Realm cachedInstance = Realm.getInstance(configuration);
+        assertNotNull(cachedInstance);
```
beeender: close the instance

realm/realm-library/src/main/cpp/io_realm_RealmFileUserStore.cpp
```diff
@@ -85,6 +85,20 @@ JNIEXPORT void JNICALL Java_io_realm_RealmFileUserStore_nativeLogoutUser(JNIEnv*
...
+        const std::shared_ptr<SyncUser>& user = SyncManager::shared().get_existing_logged_in_user(id);
+        if (user) {
+            return user->state() == SyncUser::State::Active;
```
kneth: Consider to use `to_jbool()`.

realm/realm-library/src/objectServer/java/io/realm/SyncUser.java
kneth: Please consider to order the terms so the most unlikely is first.

realm/realm-library/src/objectServer/java/io/realm/internal/SyncObjectServerFacade.java
```diff
@@ -86,10 +88,22 @@ public void realmClosed(RealmConfiguration configuration) {
     public Object[] getUserAndServerUrl(RealmConfiguration config) {
         if (config instanceof SyncConfiguration) {
             SyncConfiguration syncConfig = (SyncConfiguration) config;
+            // make sure the user is still valid
+            SyncUser user = syncConfig.getUser();
+            if (!user.isValid()) {
+                if (user.getAccessToken() == null) {
+                    throw new IllegalStateException("The SyncUser is already logged out, can not use the provided configuration");
```
kneth: Missing `.` at the end

realm/realm-library/src/objectServer/java/io/realm/UserStore.java
```diff
@@ -68,4 +68,12 @@
      * @return Collection of all users. If no users exist, an empty collection is returned.
      */
     Collection<SyncUser> allUsers();
+
+    /**
+     * Returns the state of the specified user: {@code true} if active (not logged out), {@code false} otherwise.
```
cmelchior: Perhaps also clarify that this will return true if the user has expired but not actively logged out.

realm/realm-library/src/objectServer/java/io/realm/internal/SyncObjectServerFacade.java
```diff
@@ -86,10 +88,22 @@ public void realmClosed(RealmConfiguration configuration) {
     public Object[] getUserAndServerUrl(RealmConfiguration config) {
         if (config instanceof SyncConfiguration) {
             SyncConfiguration syncConfig = (SyncConfiguration) config;
+            // make sure the user is still valid
+            SyncUser user = syncConfig.getUser();
+            if (!user.isValid()) {
+                if (user.getAccessToken() == null) {
+                    throw new IllegalStateException("The SyncUser is already logged out, can not use the provided configuration.");
```
cmelchior: out, can -> out and can

realm/realm-library/src/objectServer/java/io/realm/internal/SyncObjectServerFacade.java
```diff
@@ -86,10 +88,22 @@ public void realmClosed(RealmConfiguration configuration) {
...
+                } else {
+                    // user was not logged out but the `refresh_token` is not longer valid
+                    // the user will still get a stall version of Realm, that will work offline
+                    // but not sync.
+                    RealmLog.warn("SyncUser is no longer valid.");
```
cmelchior: Without context, this isn't much help. Perhaps  "Opening Realm at <path>, but user <identity> is longer valid" ?

realm/realm-library/src/objectServer/java/io/realm/internal/SyncObjectServerFacade.java
```diff
@@ -86,10 +88,22 @@ public void realmClosed(RealmConfiguration configuration) {
...
+                }
+            }
             String rosServerUrl = syncConfig.getServerUrl().toString();
-            String rosUserIdentity = syncConfig.getUser().getIdentity();
-            String syncRealmAuthUrl = syncConfig.getUser().getAuthenticationUrl().toString();
-            String rosRefreshToken = syncConfig.getUser().getAccessToken().value();
+            String rosUserIdentity = user.getIdentity();
+            String syncRealmAuthUrl = user.getAuthenticationUrl().toString();
+            String rosRefreshToken = user.toJson();
```
cmelchior: something looks wrong with this variable name? Perhaps `rosSerializedUser` instead?

cmelchior: Also, I guess this is actually the real fix?

PR #4866, opened 2017-06-28
Returns a schema info map in mediator

beeender:
Using map to avoid duplicated OsSchemaInfo in mediator, especially it is
an common case for composite mediator.
The key is the module class. Not using class name string is because of
we may want to support same name class in different packages in the
future.

### DIFF
--- a/realm/realm-library/src/main/java/io/realm/internal/OsSchemaInfo.java
+++ b/realm/realm-library/src/main/java/io/realm/internal/OsSchemaInfo.java
@@ -16,8 +16,6 @@
 
 package io.realm.internal;
 
-import java.util.List;
-
 /**
  * Java wrapper for the Object Store Schema object.
  * <p>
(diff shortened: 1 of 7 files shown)


realm/realm-library/src/main/java/io/realm/internal/OsSchemaInfo.java
```diff
@@ -16,8 +16,6 @@
 
 package io.realm.internal;
 
-import java.util.List;
```
dalinaum: If List was imported as `java.util.List` here, can `Collection` be imported here by `import java.util.Collection;`? Since there is no other Collection in this file, it seems not affecting `OsSchemaInfo`'s namespace.

beeender: `io.realm.internal.Collection` will have a name conflict in this file, since they are in the same package.

PR #4868, opened 2017-06-28
Ignore beta releases in the version checker.

cmelchior:
Ignore certain types of versions in the version checker (=BETA). This will probably come in handy for the 4.0 branch.

### DIFF
--- a/realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmVersionChecker.java
+++ b/realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmVersionChecker.java
@@ -21,6 +21,7 @@ import java.io.IOException;
 import java.io.InputStreamReader;
 import java.net.HttpURLConnection;
 import java.net.URL;
+import java.util.Locale;
 
 import javax.annotation.processing.ProcessingEnvironment;
 import javax.tools.Diagnostic;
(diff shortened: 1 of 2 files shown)


realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmVersionChecker.java
```diff
@@ -70,11 +71,35 @@ public void run() {
...
+     * @param currentVersion current version string.
+     * @return returns {@code true} if the new version detected should be reported to the user, {@code false} if not.
+     */
+
+    public static boolean reportNewVersion(String newVersion, String currentVersion) {
+        if (newVersion == null || newVersion.equals("")) {
+            // Invalid version strings should be ignored
+            return false;
+        }
+        if (newVersion.toLowerCase(Locale.US).endsWith("beta")) {
```
beeender: `endswith` -> `contains` to make it more robust

MGaetan89: Are you planning only beta version for Realm 4.0? Or also alpha/RC or anything else?
Why not checking if there is a prefix (`-something`) after the version number?

beeender: yes, i guess we should just use regex like `/[0-9]*\./[0-9]*\.[0-9]*$/`

cmelchior: Good idea. I was just going with minimal effort, but it is probably safe to assume that anything other than a number in the end is not an "official" release.

zaki50:
What's the point of this change?

In original code, `checkLatestVersion()` only returns a string that matches regex `\\d+\\.\\d+\\.\\d+` or the `REALM_VERSION`.
Isn't the return value of `checkLatestVersion()` always equal to `REALM_VERSION` if it ends with `-beta`?

realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmVersionChecker.java
```diff
@@ -70,11 +71,37 @@ public void run() {
...
+    /**
+     * Checks if the latest release should be reported to the user.
+     *
+     * Note: Public for testing.
+     *
+     * @param newVersion new version string.
+     * @param currentVersion current version string.
+     * @return returns {@code true} if the new version detected should be reported to the user, {@code false} if not.
+     */
+
```
zaki50: space

cmelchior:
@zaki50 
You're right. I must be blind :man_facepalming: . This PR doesn't change the semantics of the current behaviour. I missed that because the unit test doesn't touch the network code :(

Issue #4864, opened 2017-06-27
Realm 3.4.0: isEmpty for linked relationships causes SIGSEGV


polgacsa:
#### Goal

Query for empty inverse relationships: get all dogs without owners.

#### Expected Results

Query should return the dogs without owners.

#### Actual Results

SIGSEGV:
```
06-27 22:21:19.670 3461-3461/com.realm.bug.realmbug A/libc: Fatal signal 11 (SIGSEGV), code 2, fault addr 0x2429a227 in tid 3461 (lm.bug.realmbug)
                                                            
                                                            [ 06-27 22:21:19.670  1256: 1256 W/         ]
                                                            debuggerd: handling request: pid=3461 uid=10079 gid=10079 tid=3461
```
Full log attached: [full_log.txt](https://github.com/realm/realm-java/files/1106998/full_log.txt)

#### Steps & Code to Reproduce

The first query (`equalTo("owner.id", "person")`) is successful, and works as expected:
`06-27 21:52:47.999 31101-31101/com.realm.bug.realmbug I/MainActivity: size=1`
However the second (`isEmpty("owner")`) causes segfault.

#### Code Sample

`Dog` and `Person` are simple `RealmObject`s with `@PrimaryKey` id.

```java
private void testRealmBug(Context context) {
    Realm.init(context);
    RealmConfiguration realmConfiguration = new RealmConfiguration.Builder()
            .inMemory()
            .schemaVersion(2)
            .build();
    Realm realm = Realm.getInstance(realmConfiguration);

    realm.executeTransaction(new Realm.Transaction() {
        @Override
        public void execute(Realm realm) {
            Dog dog = realm.createObject(Dog.class, "dog");
            Person person = realm.createObject(Person.class, "person");
            person.getDogs().add(dog);
        }
    });

    // works as expected
    RealmResults<Dog> dogs = realm.where(Dog.class)
            .equalTo("owner.id", "person")
            .findAll();
    Log.i(TAG, "size=" + dogs.size());

    // SIGSEGV
    dogs = realm.where(Dog.class)
            .isEmpty("owner")
            .findAll();
    Log.i(TAG, "size=" + dogs.size());
}
```

```java
@RealmClass
public class Person extends RealmObject {

    @PrimaryKey
    private String id;

    private RealmList<Dog> dogs;

    public Person() {
    }

    public RealmList<Dog> getDogs() {
        return dogs;
    }

    public String getId() {
        return id;
    }
}
```

```java
@RealmClass
[...]

PR #4870, opened 2017-06-28
Fixing isNull() for link queries

### LINKED ISSUE #4856: Realm 3.4.0: exception for nested isNull() query
#### Goal
I had a nested _isNull_ query on a child object (like described in the [documentation](https://realm.io/docs/java/latest/#chaining-queries)). The query worked well until I updated to realm version 3.4.0.

#### Expected Results
I expected that the query is executed without exception using realm version 3.4.0 like it was with realm version 3.3.2.

Documentation of _isNull_:
```java
 /**
     * Tests if a field is {@code null}. Only works for nullable fields.
     * <p>
     * For link queries, if any part of the link path is {@code null} the whole path is considered to be {@code null}
     * e.g., {@code isNull("linkField.stringField")} will be considered to be {@code null} if either {@code linkField} or
     * {@code linkField.stringField} is {@code null}.
     *
     * @param fieldName the field name.
     * @return the query object.
     * @throws java.lang.IllegalArgumentException if the field is not nullable.
     * @see Required for further infomation.
     */
```

#### Actual Results
I received an exception using the nested _isNull_ query
```
java.lang.IllegalArgumentException: Illegal Argument: isNull() by nested query for link field is not supported.
 at io.realm.internal.TableQuery.nativeIsNull(Native Method)
 at io.realm.internal.TableQuery.isNull(TableQuery.java:611)
 at io.realm.RealmQuery.isNull(RealmQuery.java:205)
 at com.bz.realmchildquery.MainActivity.executeQuery(MainActivity.java:63)
```

#### Code Sample
You can download a sample project from: 
https://github.com/baloghz/RealmChildQuery.git

For demonstration I slightly modified the example from the [documentation](https://realm.io/docs/java/latest/#chaining-queries). A person can have one optional dog and the dog has an optional parameters (ate_at).

```java
    private Long ate_at;  // optional
    // getters & setters ...
}

    private Dog dog; // optional
    // getters & setters ...
}
```
I would like to find the Persons whose dog's ate_at is null
```java
RealmQuery<Person> query = realm.where(Person.class).isNull("dog.ate_at");
```

Try to change realm versions and you can see that everything works fine in 3.3.2 and not in 3.4.0

Realm version when the query worked: 3.3.2 
Realm version when the query gave an exception: 3.4.0

Android Studio version: 2.3.3

Which Android version and device: API21, Samsung SM-N9005

kneth:
If the link query is on a nullable field, it seems to query the wrong field.

This is embarrassing bug, and the solution is simple: Follow links between tables and not queries links.

Fixes #4856

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 
 ### Bug Fixes
 
-* Fixed a bug related to link queries (#4856).
+* Fixed a bug in `isNull()`, `isNotNull()`, `isEmpty()`, and `isNotEmpty()` when queries involve nullable fields in link queries (#4856).
 
 ### Internal
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### Bug Fixes
 
+* Fixed a bug related to link queries (#4856).
+
 ### Internal
 
 * Removed `Table#Table()`, `Table#addEmptyRow()`, `Table#addEmptyRows()`, `Table#add(Object...)`, `Table#pivot(long,long,PivotType)` and `Table#createnative()`.
--- a/realm/realm-library/src/androidTest/java/io/realm/RealmLinkTests.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/RealmLinkTests.java
@@ -534,6 +534,9 @@ public class RealmLinkTests {
 
         RealmResults<Owner> owners2 = testRealm.where(Owner.class).isNull("cat").findAll();
         assertEquals(1, owners2.size());
+
+        RealmResults<Owner> owners3 = testRealm.where(Owner.class).isNull("dogs.birthday").findAll();
+        assertEquals(0, owners3.size());
     }
 
     @Test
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,8 @@
 
 ### Bug Fixes
 
+* Fixed a bug related to link queries (#4856).
```
Zhuinden: well that's not very specific

beeender:
https://github.com/realm/realm-java/issues/4864 could be fixed as well i think?

kneth:
@beeender Maybe, it is probably worth adding a test inspired by #4864.

CHANGELOG.md
kneth: Fixed a bug in `isNull()`, `isNotNull()`, `isEmpty()` and `isNotEmpty()` when queries involve nullable fields in link queries.

Zhuinden:
Will there be a patch release with this change (3.4.1)?

kneth:
@Zhuinden I guess so :smile:

PR #4871, opened 2017-06-28
PermissionWrapper - acceptOffer

cmelchior:
Part of #4558 

Adds support for `acceptOffer`, see https://docs.google.com/document/d/1T-kkNnNXwqgED2Q0GRsObX10Q9RrHz_j7ycMBeLknUU/edit

Also this PR refactors the callbacks. We was now up to 3 different callbacks with 0, 1, and 2 return values. Instead of trying generify these which would result in something like `Callback0`, `Callback1`, `Callback2`, opted for readability instead so introduced a callback for each command. This also makes it possible to document and name parameters more clearly.


cmelchior:
Ready for review @realm/java

CHANGELOG.md
```diff
@@ -30,7 +30,7 @@
 
 ### Bug Fixes
 
-* Fixed a bug in `isNull()`, `isNotNull()`, `isEmpty()`, and `isNotEmpty()` when queries involve nullable fields in link queries (#4856).
```
zaki50: This fix should go to `releases` branch.
I made a PR for this #4923

realm/realm-library/src/objectServer/java/io/realm/PermissionManager.java
```diff
@@ -900,13 +1041,68 @@ private ErrorCode combineErrorCodes(Map<String, ObjectServerError> errors) {
 
     }
 
+    private interface ErrorCallback {
```
zaki50: The name of this interface should be `PermissionBaseCallback` or similar.
`PermissionsCallback is a ErrorCallback` is odd, isn't it?

realm/realm-library/src/objectServer/java/io/realm/PermissionManager.java
```diff
@@ -279,29 +281,36 @@ public RealmAsyncTask makeOffer(PermissionOffer offer, final Callback<String> ca
     }
 
     /**
-     * FIXME
-     * @param offerToken
-     * @param callback
-     * @return
+     * Accepts a permission offer sent by another user. Once this offer is accepted successfully, the permissions
+     * described by the token will be granted.
```
beeender: > A PermissionOffer allows to create a token, which can be shared via any communication channel to let any receiver obtain permissions for a Realm at a specific URL.

So it is user's responsibility to pass the token to the other?

realm/realm-library/src/objectServer/java/io/realm/PermissionManager.java
```diff
@@ -542,17 +553,24 @@ public void onChange(RealmResults <Permission> loadedPermissions) {
                 });
             }
         }
+
+        private void notifyCallbackWithSuccess(RealmResults<Permission> permissions) {
+            callback.onSuccess(permissions);
+            activeTasks.remove(this);
```
beeender: it would be good we can abstract this into the parent, at least the `activeTasks.remove(this);` is the common thing. But I am OK with what it is right now.

realm/realm-library/src/objectServer/java/io/realm/PermissionManager.java
```diff
@@ -730,6 +772,110 @@ public void onError(Throwable error) {
...
+                        @Override
+                        public void onChange(final PermissionOfferResponse response) {
+                            if (checkAndReportInvalidState()) {
+                                RealmObject.removeChangeListener(managedResponse, this);
+                                return;
+                            }
+                            handleServerStatusChanges(response, new Runnable() {
+                                @Override
+                                public void run() {
+                                    RealmObject.removeAllChangeListeners(managedResponse);
```
beeender: first, this has been called already at line 974
2nd, why call `removeAllChangeListeners` instead of `removeChangeListener()`?

realm/realm-library/src/objectServer/java/io/realm/PermissionManager.java
cmelchior: Yes. The offer/accept API is intended for async use cases like email/chat/app-links etc.

realm/realm-library/src/objectServer/java/io/realm/PermissionManager.java
cmelchior: Yes, it originally lived there, but with multiple different callbacks, it became impossible to keep there.

realm/realm-library/src/objectServer/java/io/realm/PermissionManager.java
cmelchior: 974?

realm/realm-library/src/objectServer/java/io/realm/PermissionManager.java
```diff
@@ -705,12 +742,17 @@ public void onSuccess() {
...
                     RealmObject.addChangeListener(managedOffer, new RealmChangeListener<PermissionOffer>() {
                         @Override
-                        public void onChange(PermissionOffer permissionOffer) {
+                        public void onChange(final PermissionOffer permissionOffer) {
                             if (checkAndReportInvalidState()) {
                                 RealmObject.removeChangeListener(managedOffer, this);
                                 return;
                             }
-                            handleServerStatusChanges(permissionOffer, permissionOffer.getToken());
+                            handleServerStatusChanges(permissionOffer, new Runnable() {
```
bmeike: Again, each one of these anonymous classes is an entirely new class.  It might be worth creating a single named class for this, and just creating instances as you need them.  Lambdas will do a better job of this, when we can use them.

realm/realm-library/src/objectServer/java/io/realm/PermissionManager.java
```diff
@@ -239,13 +241,13 @@ public RealmAsyncTask getDefaultPermissions(final Callback<RealmResults<Permissi
...
      * Once the request is successfully handled, a {@link Permission} entry is created in each user's
-     * {@link PermissionManager} and can be found using {@link PermissionManager#getPermissions(Callback)}.
+     * {@link PermissionManager} and can be found using {@link PermissionManager#getPermissions(PermissionsCallback)}.
      *
      * @param request request object describing which permissions to grant and to what Realm(s).
      * @param callback callback when the request either succeeded or failed.
      * @return async task representing the request. This can be used to cancel it if needed.
      */
-    public RealmAsyncTask applyPermissions(PermissionRequest request, final Callback<Void> callback) {
+    public RealmAsyncTask applyPermissions(PermissionRequest request, final ApplyPermissionsCallback callback) {
```
Zhuinden: Better question is why `PermissionRequest` is not final if the Callback is.

realm/realm-library/src/syncIntegrationTest/java/io/realm/PermissionManagerTests.java
```diff
@@ -576,6 +590,155 @@ public void onError(ObjectServerError error) {
...
+    @Test
+    @RunTestInLooperThread
+    @Ignore("FIXME: Figure out how the time differs between emulator and server")
+    public void acceptOffer_expiredThrows() {
+        // Trying to guess how long CI is to process this. The offer cannot be created if it
+        // already expired.
+        long delayMillis = TimeUnit.SECONDS.toMillis(10);
+        Date expiresAt = new Date(new Date().getTime() + delayMillis);
+        final String offerToken = createOffer(user, "test", AccessLevel.WRITE, expiresAt);
+        SystemClock.sleep(delayMillis); // Make sure that the offer expires.
```
beeender: Even you can guess how long it expires, it is still not good to run on CI since it takes too much time.
Better to find another way to test this.

realm/realm-library/src/syncIntegrationTest/java/io/realm/PermissionManagerTests.java
```diff
@@ -576,6 +590,155 @@ public void onError(ObjectServerError error) {
...
+                fail(error.toString());
+            }
+        };
+
+        pm2.acceptOffer(offerToken, callback);
+        pm3.acceptOffer(offerToken, callback);
+    }
+
+    /**
+     * Creates a offer for a newly created Realm.
```
beeender: minor, `a` -> `an`

realm/realm-library/src/syncIntegrationTest/java/io/realm/PermissionManagerTests.java
cmelchior: Yes, this is extremely flaky. I'll remove the test for now while thinking about how to fix it.

PR #4874, opened 2017-06-29
Better control of ROS instance used by the integration tests

cmelchior:
Courtesy of @alebsack.

This fixes a bunch of flaky integration tests. Mostly seen during the development of #4558 , but could also be the root cause for a bunch of other stability issues we have seen lately with integration tests timing out.

### DIFF
--- a/tools/sync_test_server/ros-testing-server.js
+++ b/tools/sync_test_server/ros-testing-server.js
@@ -32,6 +32,7 @@ function handleRequest(request, response) {
 }
 
 var syncServerChildProcess = null;
+var syncServerDir = null;
 
 function startRealmObjectServer(done) {
     // Hack for checking the ROS is fully initialized.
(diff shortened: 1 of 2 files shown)


bmunkholm:
@cmelchior Is this specific to Java or should any of this be reused for other SDKs?

cmelchior:
It is specific to how Java set up its test environment, but it does indicate that we need a description of what exactly a ROS instance does with regards to the file system.

alebsack:
I think we should eventually roll all of this into a common test harness provided by ROS.

tools/sync_test_server/ros-testing-server.js
```diff
@@ -40,50 +41,65 @@ function startRealmObjectServer(done) {
...
-                    if (logFindingCounter == 1) {
-                        done()
+    stopRealmObjectServer(function(err) {
+        if(err) {
+          return;
+        }
+        temp.mkdir('ros', function(err, path) {
+            if (!err) {
+                var oldCwd = process.cwd();
+                process.chdir(path);
```
beeender: spawn can set the `cwd` as well, would it be better to pass the `path` there instead of change the current process's `cwd`? https://nodejs.org/api/child_process.html#child_process_child_process_spawn_command_args_options

tools/sync_test_server/ros-testing-server.js
```diff
@@ -40,50 +41,65 @@ function startRealmObjectServer(done) {
...
-        syncServerChildProcess = null;
-        exec('rm -r ' + 'realm-object-server', function (err, stdout, stderr) {
-            if (err) {
-                winston.error(err);
-            } else {
-                winston.info("realm-object-server directory deleted");
-            }
+        syncServerChildProcess.on('exit', function() {
+            syncServerChildProcess = null;
+            exec('rm -r ' + syncServerDir, function (err, stdout, stderr) {
```
beeender: if the the `path` is always a temporary unique path, maybe keeping it is a better idea? for debugging purpose.

cmelchior: Since those dirs would never be deleted, I would rather avoid this for now. Mostly for selfish reasons, since I'm battling disk space daily, but also because I'm not sure we generally need that level of debugging?

beeender: The path should located in your system's `/tmp` dir which is cleaned every 3 days for macOS i think? https://superuser.com/a/187105

cmelchior: Ok, removed the deletion of the temp folders

cmelchior:
Ok, so I _think_ I finally solved the stability issues plaguing the integration tests. I don't have a full understanding of all the moving parts yet, but at least this setup doesn't crash randomly anymore on any of my permission tests (3 tries) :crossed_fingers: 

So far I identified 3 different issues:
1) We had race conditions when setting up the ROS server
2) ROS is saving data in the the process root dir, this was not cleaned up correctly befor. 
2) Not logging users out is apparently causing problems for reasons I don't fully understand since we are wiping the file-system and resetting as much state as possible. Most likely something in Object Store is maintaining old state. Logging all users out manually after each test seems to fix this.

PR #4876, opened 2017-06-29
Merge standard schema classes to their parents

beeender:
Since we changed the architecture, no need for this extra abstraction
anymore.


beeender:
I believing my vimming skills 😁

PR #4882, opened 2017-06-29
[Sync] Adding user account lookup

### LINKED ISSUE #4828: Expose User Lookup API
### Goal

Our users need a way to look up the random generated user IDs to connect those with their own user models of existing backends and can interact with user-specific Realms. We've an API which covers a minimum useful solution to achieve a user lookup. As this is subject to change, it would be desirable to wrap this API endpoint, so that our users can rely on a stable API and we can change the implementation details.

#### `SyncUser SyncUser.retrieveUser(String provider, String providerId)`

Retrieve another user by the provider identifier of one of their accounts. This is done synchronously and involves the network, so calling this method on the Android UI thread will always crash.
This user needs admin privilege in order to retrieve another user.

#### `RealmAsyncTask SyncUser.retrieveUserAsync(String provider, String providerId, SyncUser.Callback callback)`

Retrieve another user asynchronously by the provider identifier of one of their accounts.
This user needs admin privilege in order to retrieve another user.

nhachicha:
fixes #4828


nhachicha:
@realm/java

CHANGELOG.md
```diff
@@ -21,6 +21,8 @@
 
 ### Enhancements
 
+* [ObjectServer] Adding user lookup API for admin (#4828).
```
cmelchior: for administrators.

realm/realm-library/src/objectServer/java/io/realm/SyncUser.java
```diff
@@ -426,6 +427,68 @@ public SyncUser run() {
         }.start();
     }
 
+    /**
+     * Helper method for Admin users in order to look up a {@code SyncUser} using the identity provider and the used username.
+     * @param provider identity providers {@link io.realm.SyncCredentials.IdentityProvider}.
```
cmelchior: Whitespace between description and param missing.

realm/realm-library/src/objectServer/java/io/realm/SyncUser.java
```diff
@@ -426,6 +427,68 @@ public SyncUser run() {
...
+     * @param providerId username or email used to create the account first time.
```
cmelchior: I would expand this description to e.g. `username or email used to create the account for the first. What is needed will depend on what type of {@link SyncCredentials} was used.`

realm/realm-library/src/objectServer/java/io/realm/SyncUser.java
```diff
@@ -426,6 +427,68 @@ public SyncUser run() {
...
+     * @return {@code SyncUser} associated with the given identity provider and providerId.
+     * @throws ObjectServerError if the user could not be found.
```
cmelchior: It feels really strong to throw an error in this case, since that seems a really common scenario. Wouldn't just returning `null` be better?

realm/realm-library/src/objectServer/java/io/realm/SyncUser.java
```diff
@@ -426,6 +427,68 @@ public SyncUser run() {
...
+        if (Util.isEmptyString(provider)) {
+            throw new IllegalArgumentException("Not-null 'provider' required.");
+        }
+
+        if (Util.isEmptyString(providerId)) {
+            throw new IllegalArgumentException("None empty 'providerId' required.");
+        }
+
+        if (!isAdmin()) {
+            throw new IllegalStateException("User need to be admin in order to lookup ID.");
```
cmelchior: in order to lookup other users.

realm/realm-library/src/objectServer/java/io/realm/SyncUser.java
```diff
@@ -426,6 +427,68 @@ public SyncUser run() {
...
+                Token refreshToken = new Token(null, response.getUserId(), null, 0, null, response.isAdmin());
+                ObjectServerUser objectServerUser = new ObjectServerUser(refreshToken, getAuthenticationUrl());
+                objectServerUser.localLogout();
+                return new SyncUser(objectServerUser);
+            }
+        }
+    }
+
+    /**
+     * Asynchronously lookup up a {@code SyncUser} using the identity provider and the used username.
```
cmelchior: remove up

realm/realm-library/src/objectServer/java/io/realm/SyncUser.java
```diff
@@ -426,6 +427,68 @@ public SyncUser run() {
...
+     * This is for Admin users only.
+     *
+     * @param provider identity providers {@link io.realm.SyncCredentials.IdentityProvider}.
```
cmelchior: {@link io.realm.SyncCredentials.IdentityProvider} used when the account was created.

realm/realm-library/src/objectServer/java/io/realm/internal/network/LookupUserIdResponse.java
```diff
@@ -0,0 +1,127 @@
...
+    private LookupUserIdResponse(String serverResponse) {
+        ObjectServerError error;
+        String userId;
+        Boolean isAdmin;
+        String message;
+        try {
+            JSONObject obj = new JSONObject(serverResponse);
+            JSONObject jsonUser = obj.getJSONObject(JSON_FIELD_USER);
+            if (jsonUser != null) {
+                userId = jsonUser.has(JSON_FIELD_USER_ID) ? jsonUser.getString(JSON_FIELD_USER_ID) : null;
```
cmelchior: `jsonUser.optString(JSON_FIELD_USER_ID, null);` is a lot shorter 😉  Also below with `optBoolean`

realm/realm-library/src/objectServer/java/io/realm/internal/network/OkHttpAuthenticationServer.java
```diff
@@ -117,6 +127,18 @@ private static URL buildActionUrl(URL authenticationUrl, String action) {
         }
     }
 
+    private static URL buildLookupUserIdUrl(URL authenticationUrl, String action, String provider, String providerId) {
+        String authURL = authenticationUrl.toExternalForm();
+        // we need the base URL without the '/auth' part
+        String baseUrlString = authURL.substring(0, authURL.indexOf(authenticationUrl.getPath()));
```
cmelchior: Wouldn't it be easier to read if you use `ACTION_LOGOUT = "auth/revoke"` as a general pattern instead?

realm/realm-library/src/objectServer/java/io/realm/SyncUser.java
```diff
@@ -426,6 +427,68 @@ public SyncUser run() {
...
     }
 
+    /**
+     * Helper method for Admin users in order to look up a {@code SyncUser} using the identity provider and the used username.
+     * @param provider identity providers {@link io.realm.SyncCredentials.IdentityProvider}.
+     * @param providerId username or email used to create the account first time.
+     * @return {@code SyncUser} associated with the given identity provider and providerId.
+     * @throws ObjectServerError if the user could not be found.
+     */
+    public SyncUser retrieveUser(final String provider, final String providerId) throws ObjectServerError {
```
cmelchior: would `lookupUser()` be a better API name for this? `retrieve` to me sounds like something you lost or thrown away.

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/AuthTests.java
```diff
@@ -266,4 +267,110 @@ public void changePassword_throwWhenUserIsLoggedOut() {
...
+        final SyncUser user = SyncUser.login(credentials, Constants.AUTH_URL);
+        assertTrue(user.isValid());
+
+        try {
+            adminUser.retrieveUser("invalid", "username");
+            fail("Should not be possible to the userId of an unknown provider");
+        } catch (ObjectServerError expected) {
+        }
+    }
+
```
cmelchior: Whitespace

realm/realm-library/src/objectServer/java/io/realm/internal/network/AuthenticationServer.java
```diff
@@ -66,4 +66,9 @@
      * Changes a user's password using admin account.
      */
     ChangePasswordResponse changePassword(Token adminToken, String userID, String newPassword, URL authenticationUrl);
+
+    /**
+     * Look up the generated user ID using the selected identity provider.
```
beeender: Looks up

realm/realm-library/src/objectServer/java/io/realm/SyncUser.java
nhachicha: As per our discussion on Slack the error code `404` is not handled in `ErrorCode` (with a bunch of other potential that could be sent from ROS) this behaviour will change once we add the missing codes.

realm/realm-library/src/objectServer/java/io/realm/SyncUser.java
beeender: I think `retrieve` is fine here, since it returns only one user which matches but not a list of users. what is the name being used for cocoa?

nhachicha: I agree, as long as we align with other bindings ... the original proposed name was `retrieveUser` https://github.com/realm/realm-java/issues/4828

realm/realm-library/src/objectServer/java/io/realm/SyncUser.java
cmelchior: Not sure I follow? Wouldn't that just mean that different `RealmObjectServer` error would be returned? I was wondering if we should throw an error at all? Or do you mean that until we can correctly detect 404, we have no way of correctly knowing that was what happened?

realm/realm-library/src/objectServer/java/io/realm/internal/network/LookupUserIdResponse.java
nhachicha: cool for the String, the Boolean doesn't allow the user of `null` (for the boxed Boolean) as default value.

realm/realm-library/src/objectServer/java/io/realm/internal/network/OkHttpAuthenticationServer.java
nhachicha: It sure does. We need to update all our example/tests so `Constants.AUTH_URL` will not contain `auth` anymore as part of the URL. This could be done in a separate PR I believe

realm/realm-library/src/objectServer/java/io/realm/SyncUser.java
cmelchior: I don't have a strong opinion on this, so :man_shrugging:, but given that we describe the term as a "user lookup API". It feels weird then to use the word `retrieveUser`

realm/realm-library/src/objectServer/java/io/realm/internal/network/OkHttpAuthenticationServer.java
cmelchior: Thats fine 👍

Issue #4662, opened 2017-05-18
javassist.NotFoundException on Realm classes when running connectedAndroidTest


NitroG42:
Using last android plugin (3.0.0-alpha1) running the connectedAndroidTest task makes it crash on compile with a Error:Execution failed for task ':app:transformClassesWithRealmTransformerForDebugAndroidTest'.
> javassist.NotFoundException: com.package.myapp.model.MyModel

the app compilation itself works, but the androidTest can't find the classes.
It's a 100% Kotlin project.
If you need a sample project I can give one for you I think (it is reproductible on the projects I work on)

#### Version of Realm and tooling
Realm version(s): 3.2.0
Realm sync feature enabled: no
Android Studio version: 3.0.0
Using Kotlin and gradle 4.0.0-m2

bmeike:
We'll definitely have a look at this.  We don't support the alpha1 plugin, at this point.

gabrieloshiro:
Hi,

I get the same thing

100% Kotlin project
Realm version(s): 3.2.1
Realm sync feature enabled: no
Android Studio version: 3.0.0 canary 2
Gradle wrapper 4.0.0-m2
Kotlin-Gradle 1.1.2-4

`./gradlew clean test check assemble` passes.
`./gradlew clean test check connectedAndroidTest assemble` fails with

> javassist.NotFoundException: com.package.myapp.model.MyModel

isaac-udy-8i:
I'm also having the same problem.

Is there any sense of when we might see a fix for this? I'd rather not re-write our model classes in Java if this is going to be fixed in the short term, but I might have to if this issue is going to take a long time to solve.

cmelchior:
Have you tried using the latest stable version 2.3.2 instead? The `3.0.0-alpha` has quite a number of bugs, so ruling that out first would be good.

NitroG42:
It works on stable (2.3.2) and latest 2.4.0-alpha7. I made an issue here because I'm not sure if it's a bug or a need to implement some new features (like the fact they changed how flavors and dimensions works for example)

cmelchior:
Most likely it is a bug in the 3.0.0 plugin. We have not heard of any changes required by our end yet. I would recommend that you create an issue in the google bugtracker instead:  https://issuetracker.google.com/issues?q=componentid:192709%2B

We will keep this issue open just to keep track of it on our end as well.

NitroG42:
Done: https://issuetracker.google.com/issues/62152896

[...]

PR #4883, opened 2017-06-30
added a test that reproduce #4662

zaki50:
This PR exposes issue #4662


zaki50:
Ah, we don't execute `connectedCheck` against examples on our CI...

zaki50:
this issue was fixed in

PR #4884, opened 2017-06-30
Add RealmSchema.createWithPrimaryKeyField()

beeender:
And deprecate RealmObject.addPrimaryKey()/removePrimaryKey().

### DIFF
--- a/realm/realm-library/src/androidTest/java/io/realm/RealmSchemaTests.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/RealmSchemaTests.java
@@ -121,6 +121,7 @@ public class RealmSchemaTests {
         assertEquals("pkField", objectSchema.getPrimaryKey());
         assertEquals(RealmFieldType.STRING, objectSchema.getFieldType("pkField"));
         assertFalse(objectSchema.isNullable("pkField"));
+        assertTrue(objectSchema.hasIndex("pkField"));
 
         // Nullable
         realmSchema.createWithPrimaryKeyField("FooNull", "pkField", String.class);
@@ -128,6 +129,7 @@ public class RealmSchemaTests {
         assertEquals("pkField", objectSchema.getPrimaryKey());
         assertEquals(RealmFieldType.STRING, objectSchema.getFieldType("pkField"));
         assertTrue(objectSchema.isNullable("pkField"));
+        assertTrue(objectSchema.hasIndex("pkField"));
     }
 
     @Test
@@ -138,6 +140,7 @@ public class RealmSchemaTests {
         assertEquals("pkField", objectSchema.getPrimaryKey());
         assertEquals(RealmFieldType.INTEGER, objectSchema.getFieldType("pkField"));
         assertFalse(objectSchema.isNullable("pkField"));
+        assertTrue(objectSchema.hasIndex("pkField"));
 
         // Nullable
         realmSchema.createWithPrimaryKeyField("FooNull", "pkField", Integer.class);
@@ -145,13 +148,34 @@ public class RealmSchemaTests {
         assertEquals("pkField", objectSchema.getPrimaryKey());
         assertEquals(RealmFieldType.INTEGER, objectSchema.getFieldType("pkField"));
         assertTrue(objectSchema.isNullable("pkField"));
+        assertTrue(objectSchema.hasIndex("pkField"));
+    }
+
+    @Test
+    public void createWithPrimaryKeyField_fieldTypeIndexed() {
+        realmSchema.createWithPrimaryKeyField("Foo", "pkField", int.class,
+                FieldAttribute.INDEXED);
+        RealmObjectSchema objectSchema = realmSchema.getSchemaForClass("Foo");
+        assertEquals("pkField", objectSchema.getPrimaryKey());
+        assertEquals(RealmFieldType.INTEGER, objectSchema.getFieldType("pkField"));
+        assertTrue(objectSchema.hasIndex("pkField"));
+    }
+
+    @Test
+    public void createWithPrimaryKeyField_fieldTypePrimaryKey() {
+        realmSchema.createWithPrimaryKeyField("Foo", "pkField", int.class,
+                FieldAttribute.PRIMARY_KEY);
+        RealmObjectSchema objectSchema = realmSchema.getSchemaForClass("Foo");
+        assertEquals("pkField", objectSchema.getPrimaryKey());
+        assertEquals(RealmFieldType.INTEGER, objectSchema.getFieldType("pkField"));
+        assertTrue(objectSchema.hasIndex("pkField"));
     }
 
     @Test
     public void createWithPrimaryKeyField_invalidClassNameThrows() {
-        String[] names = { null, "", TestHelper.getRandomString(57) };
+        String[] invalidNames = { null, "", TestHelper.getRandomString(57) };
 
-        for (String name : names) {
+        for (String name : invalidNames) {
             try {
                 realmSchema.createWithPrimaryKeyField(name, "pkField", int.class);
                 fail();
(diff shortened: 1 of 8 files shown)


beeender:
This moves part of the `RealmSchema.createWithPrimaryKeyField()` implementation from #4693

realm/realm-library/src/main/java/io/realm/FieldAttribute.java
```diff
@@ -33,7 +33,10 @@
      * Marks a field as a primary key. This also implicitly mark it as {@link #INDEXED} and {@link #REQUIRED}.
      *
      * @see io.realm.annotations.PrimaryKey
+     * @deprecated adding or removing primary key after object schema gets created will be disallowed in
+     * realm-java 4.0.0.
      */
+    @Deprecated
```
beeender: I am not quite sure it is good to mark this as `@Deprecated`, currently the only public API using this is `addField()`, so removing this in the 4.0.0 is accepted? Or do we need this in the future API like `RealmObjectSchema.getFieldAttribute()`?

Zhuinden:
So now it's only possible to make a `long` primary key field into a `String` primary key field is by renaming the class, creating a new class with all existing fields (and the new ID of course), using transform to map from the old object schema  to the new object schema,  and removing the old object schema? 

Harsh.

beeender:
@Zhuinden yes, :( it will be very much like what you described above. however some helper function could be made to support that in a better way .

realm/realm-library/src/androidTest/java/io/realm/RealmSchemaTests.java
```diff
@@ -115,6 +115,87 @@ public void create_duplicatedNameThrows() {
...
+        realmSchema.createWithPrimaryKeyField("FooNull", "pkField", Integer.class);
+        objectSchema = realmSchema.getSchemaForClass("FooNull");
+        assertEquals("pkField", objectSchema.getPrimaryKey());
+        assertEquals(RealmFieldType.INTEGER, objectSchema.getFieldType("pkField"));
+        assertTrue(objectSchema.isNullable("pkField"));
+    }
+
+    @Test
+    public void createWithPrimaryKeyField_invalidClassNameThrows() {
+        String[] names = { null, "", TestHelper.getRandomString(57) };
```
zaki50: Shouldn't we avoid to have U+0000 or any other chars not valid for table name in the name?
I think we should update `TestHelper.getRandomString(int)` to exclude U+0000 at least.

zaki50: `invalidNames` is better than just `names`.

realm/realm-library/src/androidTest/java/io/realm/RealmSchemaTests.java
```diff
@@ -115,6 +115,87 @@ public void create_duplicatedNameThrows() {
...
+                fail();
+            } catch (IllegalArgumentException ignored) {
+            }
+            assertFalse(String.format("'%s' failed", name), realmSchema.contains(name));
+        }
+    }
+
+    @Test
+    public void createWithPrimaryKeyField_invalidFieldNameThrows() {
+        String[] fieldNames = new String[] { null, "", "foo.bar", TestHelper.getRandomString(65) };
```
zaki50: We must make sure that the return value of `TestHelper.getRandomString(65)` does not contain `.`.

realm/realm-library/src/androidTest/java/io/realm/RealmSchemaTests.java
```diff
@@ -115,6 +115,87 @@ public void create_duplicatedNameThrows() {
...
+        assertEquals(RealmFieldType.INTEGER, objectSchema.getFieldType("pkField"));
+        assertFalse(objectSchema.isNullable("pkField"));
+
+        // Nullable
+        realmSchema.createWithPrimaryKeyField("FooNull", "pkField", Integer.class);
+        objectSchema = realmSchema.getSchemaForClass("FooNull");
+        assertEquals("pkField", objectSchema.getPrimaryKey());
+        assertEquals(RealmFieldType.INTEGER, objectSchema.getFieldType("pkField"));
+        assertTrue(objectSchema.isNullable("pkField"));
+    }
```
zaki50: We should have tests that pass `FieldAttribute.PRIMARY_KEY` or `FieldAttribute.INDEXED`.

realm/realm-library/src/main/java/io/realm/internal/SharedRealm.java
```diff
@@ -308,6 +308,24 @@ public Table createTable(String name) {
...
+     * @param isNullable if the primary key field is nullable or not.
+     * @return a creatd {@link Table} object.
+     */
+    public Table createTableWithPrimaryKey(String tableName, String primaryKeyFieldName, boolean isStringType,
+                                           boolean isNullable) {
+        return new Table(this, nativeCreateTableWithPrimaryKeyField(nativePtr, tableName, primaryKeyFieldName,
+                isStringType, isNullable));
+    }
+
+
```
zaki50: too many empty lines

realm/realm-library/src/main/java/io/realm/RealmObjectSchema.java
```diff
@@ -38,7 +38,7 @@
  */
 public class RealmObjectSchema {
 
-    private static final Map<Class<?>, FieldMetaData> SUPPORTED_SIMPLE_FIELDS;
+    static final Map<Class<?>, FieldMetaData> SUPPORTED_SIMPLE_FIELDS;
```
bmeike: Consider a static method to access this.  Just exposing the implementation is a pretty drastic thing to do.

realm/realm-library/src/main/java/io/realm/RealmSchema.java
```diff
@@ -27,6 +27,8 @@
 import io.realm.internal.Util;
 import io.realm.internal.util.Pair;
 
+import static io.realm.RealmObjectSchema.SUPPORTED_SIMPLE_FIELDS;
```
bmeike: Please consider a normal import, and qualification at each use.  It makes the code so much clearer...

realm/realm-library/src/main/java/io/realm/RealmSchema.java
```diff
@@ -117,6 +116,42 @@ public RealmObjectSchema create(String className) {
...
+    public RealmObjectSchema createWithPrimaryKeyField(String className, String primaryKeyFieldName, Class<?> fieldType,
+                                                       FieldAttribute... attributes) {
+        checkEmpty(className, EMPTY_STRING_MSG);
+        RealmObjectSchema.checkLegalName(primaryKeyFieldName);
+        String internalTableName = checkAndGetTableNameFromClassName(className);
+
+        RealmObjectSchema.FieldMetaData metadata = RealmObjectSchema.getSupportedSimpleFields().get(fieldType);
+        if (metadata == null || (metadata.realmType != RealmFieldType.STRING &&
+                metadata.realmType != RealmFieldType.INTEGER)) {
+            throw new IllegalArgumentException(String.format("Realm doesn't support primary key field type '%s'.",
```
zaki50: Locale (`Locale.US`) is missing.

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
```diff
@@ -455,6 +457,38 @@ JNIEXPORT jlong JNICALL Java_io_realm_internal_SharedRealm_nativeCreateTable(JNI
...
+    try {
+        std::string table_name(JStringAccessor(env, j_table_name));
+        class_name_str = std::string(table_name.substr(TABLE_PREFIX.length()));
+        JStringAccessor field_name(env, j_field_name); // throws
+        auto& shared_realm = *(reinterpret_cast<SharedRealm*>(shared_realm_ptr));
+        shared_realm->verify_in_write(); // throws
+        DataType pkType = is_string_type ? DataType::type_String : DataType::type_Int;
+        Table* table;
+        auto& group = shared_realm->read_group();
+        table = LangBindHelper::add_table(group, table_name);
```
kneth: Doesn't it make sense to have this and the next lines as a function in object store? I mean, to ensure we create primary key fields the same way?

beeender:
Deprecating those may create troubles for current users. close this for now. Maybe we can try to avoid those for non-sync users.

PR #4889, opened 2017-06-30
Last field might be a backlink field.

### LINKED ISSUE #4864: Realm 3.4.0: isEmpty for linked relationships causes SIGSEGV

kneth:
When querying a `LinkingObjects` field (aka backlink field), it might be the last. `isEmpty()` is a pretty useful predicate, and we need to track down the source table.

The changes to the C++ code is only for code style consistency.

Closes #4864


realm/realm-library/src/main/java/io/realm/internal/fields/CachedFieldDescriptor.java
```diff
@@ -82,8 +82,8 @@ protected void compileFieldDescription(List<String> fields) {
             columnType = tableInfo.getColumnType(columnName);
             if (i < nFields - 1) {
                 verifyInternalColumnType(currentTable, columnName, columnType);
-                currentTable = tableInfo.getLinkedTable(columnName);
             }
+            currentTable = tableInfo.getLinkedTable(columnName);
```
bmeike: Does this work?  I absolutely don't remember but that was in the conditional for a reason, at one time.  What happens if you call `schema.getNativeTablePtr(null)` a couple lines down?

kneth: `schema.getNativeTablePtr()` is only called when the type is `LINKING_OBJECTS`, and in that case, `currentTable` cannot be `null`.

realm/realm-library/src/main/java/io/realm/internal/fields/CachedFieldDescriptor.java
```diff
@@ -80,6 +80,7 @@ protected void compileFieldDescription(List<String> fields) {
             }
 
             columnType = tableInfo.getColumnType(columnName);
+            // we don't check the type of the last field in the chain since it done in the C++ code
```
Zhuinden: +is

realm/realm-library/src/main/cpp/io_realm_internal_TableQuery.cpp
```diff
@@ -1668,9 +1668,10 @@ JNIEXPORT void JNICALL Java_io_realm_internal_TableQuery_nativeIsEmpty(JNIEnv* e
         auto column_idx = static_cast<size_t>(index_arr[arr_len - 1]);
 
         // Support a backlink as the last column in a field descriptor
-        Table* last = TBL(table_arr[arr_len-1]);
+        auto last = reinterpret_cast<Table*>(table_arr[arr_len-1]);
         if (last != nullptr) {
-            pQuery->and_query(src_table_ref->column<BackLink>(*last, column_idx).count() == 0);
+            TableRef last_ref = TableRef(last);
```
beeender: `TableRef` is not needed here, i like less lines of code :)

realm/realm-library/src/main/cpp/io_realm_internal_TableQuery.cpp
```diff
@@ -1739,9 +1740,10 @@ Java_io_realm_internal_TableQuery_nativeIsNotEmpty(JNIEnv *env, jobject, jlong n
         auto column_idx = static_cast<size_t>(index_arr[arr_len - 1]);
 
         // Support a backlink as the last column in a field descriptor
-        auto last = reinterpret_cast<Table *>(table_arr[arr_len-1]);
+        auto last = reinterpret_cast<Table*>(table_arr[arr_len-1]);
         if (last != nullptr) {
-            pQuery->and_query(src_table_ref->column<BackLink>(*last, column_idx).count() != 0);
+            TableRef last_ref = TableRef(last);
```
beeender: same

realm/realm-library/src/main/java/io/realm/internal/fields/CachedFieldDescriptor.java
beeender: Will, for example, `query.equalTo("abc", "backlinkField")` still throw? I guess so?

realm/realm-library/src/main/cpp/io_realm_internal_TableQuery.cpp
bmeike: I think this is a matter of taste.  I had an off-line discussion with @kneth about this and I agree with him that the explicit cast is better than my original code, which just used `last`, a `Table*`.  Being succinct is good, for sure.  I don't think this cast costs much of anything and, for me at least, it makes it clearer what's going on.   I'm good with either but, FWIW, I prefer this.

kneth:
@beeender Yeah, `equalTo()` will throw `IllegalArgumentException`. I add a test.

PR #4898, opened 2017-07-03
Remove io.realm.internal.OutOfMemoryError

beeender:


bmunkholm:
@cmelchior If this is a breaking change, it should no go into 3.5, but 4.0, right?
(although I don't expect this will break anyones code in practice as they shouldn't catch this)

Zhuinden:
`internal` is not part of public api in the sense that it is not considered a breaking change if internal classes are changed, afaik

cmelchior:
@bmunkholm We don't make any promises for anything in `io.realm.internal.*`. You could argue it was a mistake it was in there in the first place, but it is. That, plus people should never catch these, makes it fine IMO.

PR #4899, opened 2017-07-03
Update android gradle plugin to 3.0.0-beta7

zaki50:
I've fixed all issues and we can merge this PR after the review. @realm/java 

PR contains following changes.

1. Updating Android Gradle Plugin to 3.0.0-beta7
2. Removing work-around in gradle wrapper
3. Updatig Gradle wrapper to 4.2.1
4. Enabling new dex compiler (D8) in example projects

~~This PR still has some issues.~~

1. ~~First execution of `publishToMavenLocal` task in `realm/realm-library` project after `clean` fails if gradle daemon is not used. Gradle doesn't execute `assemble` task in that case (`build.gradle` contains explicit dependency to `assemble` task,  no idea why not executed).
Second time s always fine.~~ <- fixed by adding more task dependencies

2. ~~CMake build fails. This is caused by https://issuetracker.google.com/issues/65336086 and the status of the issue has become fixed. I think this build failure will be fixed in coming android gradle plugin.~~ <- fixed in `3.0.0-beta7`


zaki50:
3.0.0-beta4 has been released but it does not seem to allow specifying `-DCMAKE_TOOLCHAIN_FILE` and that causes compilation error.

I've made an issue 
https://issuetracker.google.com/issues/65336086

zaki50:
:tada:

PR #4913, opened 2017-07-05
Fix flaky emptyChangeSet_findAllAsync

beeender:
There is no guarantee that async query will be finished later after the
transaction. So the original assertEquals(9, collection.size()) might
not be true always.
Just remove the object deletion logic since it has nothing to do with
the test purpose.


beeender:
Uh, OK, that is another issue, will fix. that leads to a timeout, the one i fixed lead to an assertion failure which can be seen https://ci.realm.io/job/realm/job/realm-java/job/PR-4221/22/display/redirect

PR #4915, opened 2017-07-05
Fix merge conflict resulting in duplicate error code

cmelchior:
Happened because another feature was merged into `master` that also added this code. The automatic merge could not detect this.


cmelchior:
Fast merge @realm/java

PR #4920, opened 2017-07-05
PermissionWrapper - Add support for retrieving  list of offers

cmelchior:
Part of #4558 

Adds support for returning list of active offers. Mostly to enable to get tokens again or revoke them.

TODO 
- [x] Redirect to `feature/permission-wrapper` once #4871 is merged.


realm/realm-library/src/objectServer/java/io/realm/PermissionManager.java
```diff
@@ -1041,6 +1047,49 @@ private ErrorCode combineErrorCodes(Map<String, ObjectServerError> errors) {
...
+                            offers = loadedOffers;
+                        }
+                        notifyCallbackWithSuccess(offers);
+                    }
+                });
+            }
+        }
+
+        private void notifyCallbackWithSuccess(RealmResults<PermissionOffer> permissions) {
+            callback.onSuccess(permissions);
```
zaki50: Is it OK to call `onSuccess` from async task thread?

zaki50: Ah, all tasks in `PermissionManager` are executed by handler and that means the thread is the same.
Am I correct?

cmelchior: Yes, all callbacks happen on the Looper thread that opened the PermissionManager.

realm/realm-library/src/syncIntegrationTest/java/io/realm/PermissionManagerTests.java
```diff
@@ -699,6 +699,43 @@ public void onError(ObjectServerError error) {
...
+                looperThread.keepStrongReference(offers);
+                filteredOffers.addChangeListener(new RealmChangeListener<RealmResults>() {
+                    @Override
+                    public void onChange(RealmResults results) {
+                        switch (results.size()) {
+                            case 0: return;
+                            case 1:
+                                looperThread.testComplete();
+                                break;
+                            case 2:
```
kneth: `default` instead? If more than 2?

realm/realm-library/src/objectServer/java/io/realm/PermissionManager.java
bmeike: Is this the right order?  What will happen if the callback throws?  If it doesn't create an inconsistent state, in the callback, it might be better to do the cleanup first.  Else maybe at `try {} finally {}`?

cmelchior:
Ready for review @realm/java

realm/realm-library/src/syncIntegrationTest/java/io/realm/PermissionManagerTests.java
```diff
@@ -699,6 +699,43 @@ public void onError(ObjectServerError error) {
...
+        final String offerToken = createOffer(user, "test", AccessLevel.WRITE, null);
+        PermissionManager pm = user.getPermissionManager();
+        looperThread.closeAfterTest(pm);
+
+        pm.getCreatedOffers(new PermissionManager.OffersCallback() {
+            @Override
+            public void onSuccess(RealmResults<PermissionOffer> offers) {
+                RealmResults filteredOffers = offers.where()
+                        .equalTo("token", offerToken)
+                        .findAllAsync();
```
zaki50: Why do we need to use `Async` here?

cmelchior: Because I wanted to avoid race conditions on the server as the offers need to be downloaded to the other users `__management` realm

Issue #4266, opened 2017-02-28
Support MutableRealmInteger


cmelchior:
(The original initial comment for this issue is [here](https://github.com/realm/realm-java/issues/4266#issuecomment-308772718).  This heading reflects the current understanding of the project)

The basic semantics for this feature, now called `MutableRealmInteger` have finally been sorted out [here](https://github.com/realm/realm-java/issues/4266#issuecomment-308772718).

This issue now tracks implementation.

**TODO:**
- [x] Support unmanaged MutableRealmIntegers
- [x] Support basic managed MutableRealmIntegers (atomic increment and decrement)
- [x] Support @Required
- [x] Support @Index
- [x] Support JSON and streams
- [ ] Support Dynamic API
- [ ] Support RealmList<ManagedRealmInteger>

cmelchior:
Cocoa implementation is here: https://github.com/realm/realm-cocoa/pull/4744

Zhuinden:
I actually am not sure what this is for exactly.

cmelchior:
It is only useful in a synchronised setting, but there it is very useful any time you want to count anything, since a normal integer field would get overridden with `last-write-wins` which means you could "lose" something you counted. A Counter [CRDT](https://en.wikipedia.org/wiki/Conflict-free_replicated_data_type) is guaranteed to always converge with the semantics that increments from all devices are included. 

Some use cases:
- Passenger tracking
- Facebook likes
 - Inventory tracking

bmeike:
Are `RealmCounter` and `RealmInteger` different things?  I like the name `RealmCounter` better...  It seems to me that the feature of this thing is that it is a counter, not that it is an integer,

cmelchior:
@bmeike See https://github.com/realm/realm-object-store/issues/357 for the discussion with the rationale for the name.

bmeike:
Implementation plan:
- [ ] add RealmInteger class
- [ ] make RealmInteger a working alias for Long
- [ ] add incr and decr functionality.

cmelchior:
Some description of use cases for the difference between managed/unmanaged states
```
// Unmanaged variants
RealmInteger ri = new RealmInteger()
ri.isManaged() == false
ri.isValid() == false

// Managed variant
realm.beginTransaction();
AllTypes obj = realm.createObject(AllTypes.class);
realm.commitTransaction();
RealmInteger managedRi = obj.getRealmInteger();
managedRi.isValid() == true
managedRi.isManaged() == true;

[...]

PR #4922, opened 2017-07-06
Counters - First pass at implementation

bmeike:
Rewrite of the implementation of counters as MutableRealmIntegers.
Preliminary implementation of #4266


realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmProxyClassGenerator.java
```diff
@@ -556,6 +571,8 @@ private void emitCodeForUnderConstruction(JavaWriter writer, boolean isPrimaryKe
                 .emitEmptyLine();
     }
 
+    // Note that because of bytecode hackery, this method may run before the constructor!
+    // It may ever run before fields have been initialized.
```
cmelchior: ever - even

cmelchior: Also, I don't believe it can run _before_ the constructor,  but it will be invoked during the constructor. But I guess the effect is the same. Fields might not initialized.

realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmProxyInterfaceGenerator.java
```diff
@@ -54,24 +54,29 @@ public void generate() throws IOException {
...
+            String fieldName = field.getSimpleName().toString();
+            String fieldTypeCanonicalName = field.asType().toString();
+            writer
+                    .beginMethod(
+                            fieldTypeCanonicalName,
+                            metaData.getInternalGetter(fieldName),
+                            EnumSet.of(Modifier.PUBLIC))
+                    .endMethod();
+
+            // Counters do not have setters.
```
cmelchior: Counters -> MutableRealmIntegers

realm/realm-annotations-processor/src/test/java/io/realm/processor/RealmProcessorTest.java
```diff
@@ -94,7 +94,8 @@ public void compileProcessedEmptyFile() throws Exception {
                 .failsToCompile();
     }
 
-    // Disabled because it does not seem to find the generated interface file @Test
+    @Ignore("Disabled because it does not seem to find the generated interface file")
```
cmelchior: This looks a bit worrying?

realm/realm-annotations-processor/src/test/resources/io/realm/AllTypesRealmProxy.java
```diff
@@ -662,6 +685,13 @@ public static String getTableName() {
                 ((AllTypesRealmProxyInterface) obj).realmSet$columnBinary(JsonUtils.stringToBytes(json.getString("columnBinary")));
             }
         }
+        if (json.has("columnCounter")) {
+            if (json.isNull("columnCounter")) {
+                throw new IllegalArgumentException("Trying to set non-nullable field 'columnCounter' to null.");
```
cmelchior: Is this the semantics we want? By doing this, it looks impossible to initialize a counter to `null` from JSON?

cmelchior: Also, in the streaming version we set the value to `null`, not throw?

realm/realm-annotations-processor/src/main/java/io/realm/processor/ClassMetaData.java
```diff
@@ -385,12 +392,18 @@ private boolean categorizeField(Element element) {
...
 
-        // Check @LinkingObjects last since it is not allowed to be either @Index, @Required or @PrimaryKey
+        // @LinkingObjects cannot be @PrimaryKey or @Index.
         if (field.getAnnotation(LinkingObjects.class) != null) {
+            // Do not add backlinks to fields list.
             return categorizeBacklinkField(field);
         }
 
-        // Standard field that appear valid (more fine grained checks might fail later).
+        // Similarly, a RealmInteger cannot be a @PrimaryKey, @Index or @LinkingObject.
```
cmelchior: Also, any reason we cannot index it?

realm/realm-library/src/androidTest/java/io/realm/RealmIntegerTests.java
```diff
@@ -59,228 +63,318 @@ public void tearDown() throws Exception {
...
+        assertEquals(r1, r2);
+
+        MutableRealmInteger r3 = r1;
+        r1.set(19);
+        assertEquals(19, r3.get().intValue());
+    }
+
+    @SuppressWarnings({"ReferenceEquality", "EqualsIncompatibleType"})
+    private void testEquality(Counters c1, Counters c2) {
+        c1.columnCounter.set(7);
```
cmelchior: Maybe `assertFalse(c1 == c2)` to clarify contract for using this method?

realm/realm-library/src/main/java/io/realm/MutableRealmInteger.java
```diff
@@ -0,0 +1,309 @@
...
+package io.realm;
+
+import io.realm.annotations.Beta;
+import io.realm.internal.ManagableObject;
+import io.realm.internal.Row;
+
+
+/**
+ * A MutableRealmInteger is a mutable, {@link java.lang.Long}-like numeric quantity.
+ * MutableRealmInteger implementations are not thread safe.  Like all Realm objects,
```
cmelchior: That only applies to unmanaged MutableRealmIntegers, no. Managed integers are thread-confined.

realm/realm-library/src/main/java/io/realm/MutableRealmInteger.java
```diff
@@ -0,0 +1,309 @@
...
+ * managed RealmIntegers may not be moved across threads.  Unmanaged RealmObject
+ * may be moved across threads but require safe publication.
```
cmelchior: Not sure what "safe publication" means?

realm/realm-library/src/androidTest/java/io/realm/RealmIntegerTests.java
```diff
@@ -59,228 +63,318 @@ public void tearDown() throws Exception {
...
+        assertTrue(n.equals(Long.valueOf(8)));
+        assertTrue(n.equals(c1.columnCounter.get()));
+        assertTrue(n.intValue() == c1.columnCounter.get().intValue());
+
+        c1.columnCounter.increment(1);
+        assertFalse(n.intValue() == c1.columnCounter.get().intValue());
+        assertFalse(n.intValue() == r1.get().intValue());
+    }
+
+    private void testNullability(Counters c1) {
```
cmelchior: `testBasic`, `testEquality` and `testNullability` are all pretty complicated and seem to have some specific variants not really expressed in the name, but only halfway down the tests. Could assumptions about input be clarified in Javadoc for these methods?

realm/realm-library/src/main/java/io/realm/MutableRealmInteger.java
```diff
@@ -0,0 +1,309 @@
...
+    /**
+     * Managed Implementation.
+     */
+    // FIXME Counters: wire up to native increment method
+    static class Managed extends MutableRealmInteger {
+        private final ProxyState<?> proxyState;
+        private final BaseRealm realm;
+        private final Row row;
+        private final long columnIndex;
+        private Long value;
```
cmelchior: This looks dangerous. The value should be read from the `row` every time.

realm/realm-library/src/main/java/io/realm/MutableRealmInteger.java
```diff
@@ -0,0 +1,309 @@
...
+     * Decrements the {@code MutableRealmInteger}, subtracting the value of the argument.
+     * Increment/decrement from all devices are reflected in the new value, which is guaranteed to converge.
+     *
+     * @param dec quantity to be subtracted from the counter.
+     */
+    public abstract void decrement(long dec);
+
+
+    /**
+     * @return true if and only iff {@code get()} will return {@code null}.
```
cmelchior: iff -> if

realm/realm-library/src/main/java/io/realm/MutableRealmInteger.java
```diff
@@ -0,0 +1,309 @@
...
+        Long thisValue = get();
+        Long otherValue = o.get();
+        return (otherValue == null) ? ((thisValue != null) ? 1 : 0)
+                : (thisValue == null) ? ((otherValue != null) ? -1 : 0)
+                : thisValue.compareTo(otherValue);
+    }
+
+    /**
+     * A MutableRealmInteger's hash code depends only on its value.
+     * Will NPE if either value is null
```
cmelchior: Thats not true?

realm/realm-library/src/main/java/io/realm/MutableRealmInteger.java
```diff
@@ -0,0 +1,309 @@
...
+     */
+    @Override
+    public final int hashCode() {
+        Long thisValue = get();
+        return (thisValue == null) ? 0 : thisValue.hashCode();
+    }
+
+    /**
+     * Two RealmIntegers are {@code .equals} if and only if their longValues are equal.
+     * Will NPE if either value is null
```
cmelchior: Same?

realm/realm-library/src/main/java/io/realm/MutableRealmInteger.java
Zhuinden: I'd be surprised if `hashCode()` NPE'd o-o

realm/realm-library/src/main/java/io/realm/MutableRealmInteger.java
Zhuinden: I'd be surprised if `equals()` NPE'd o-O

realm/realm-annotations-processor/src/test/java/io/realm/processor/RealmProcessorTest.java
```diff
@@ -479,6 +480,36 @@ public void compileWithInterfaceForObject() {
...
+        // FIXME Counters: Add test
+    }
+
+    @Test
+    public void failOnPKCounter() {
+        // FIXME Counters: Add test
+    }
+
+    @Test
+    public void failOnIndexedCounter() {
```
beeender: why? index should be allowed i think?

realm/realm-annotations-processor/src/main/java/io/realm/processor/ClassMetaData.java
```diff
@@ -385,12 +392,18 @@ private boolean categorizeField(Element element) {
...
+        if (Utils.isRealmInteger(field)) {
+            if (!categorizeMutableRealmIntegerField(field)) { return false; }
```
zaki50: To avoid returning when `categorizeMutableRealmIntegerField(field)` returns `true`?

realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmJsonTypeHelper.java
```diff
@@ -340,6 +353,40 @@ public void emitGetObjectWithPrimaryKeyValue(
...
+                   .beginControlFlow("if (json.isNull(\"%s\"))", fieldName)
+                       .emitStatement(Constants.STATEMENT_EXCEPTION_ILLEGAL_NULL_VALUE, fieldName)
+                   .nextControlFlow("else")
+                       .emitStatement("((%s) obj).%s().set(json.getLong(\"%s\"))", interfaceName, getter, fieldName)
+                   .endControlFlow()
+               .endControlFlow();
+       }
+       // @formatter:on
+
+        // @formatter:off
```
kneth: Formatted incorrectly :smile: - a space too many

realm/realm-library/src/androidTest/java/io/realm/RealmIntegerTests.java
```diff
@@ -59,228 +63,318 @@ public void tearDown() throws Exception {
...
+        ri.increment(17);
+        ri.decrement(7);
+        realm.commitTransaction();
+
+        assertEquals(Long.valueOf(42L), unmanagedRI.get());
+        assertEquals(Long.valueOf(47L), managedRI.get());
+    }
+
+    /**
+     * Assure that a MutableRealmInteger acquired from an unmanaged object is not affected by changes in the DB.
```
kneth: "DB" -> "Realm"

realm/realm-library/src/main/java/io/realm/MutableRealmInteger.java
kneth: or "if and only iff" -> "iff"

realm/realm-library/src/main/java/io/realm/MutableRealmInteger.java
```diff
@@ -0,0 +1,309 @@
...
+        return get() == null;
+    }
+
+
+    /**
+     * RealmIntegers compare strictly by their values.
+     * Null is a legal value for a MutableRealmInteger: null < non-null
+     *
+     * @param o the compare target
+     * @return -1, 0, or 1, depending on whether this object's value is &gt;, =, or &lt; the target's.
```
kneth: Won't it return `1` if greater-than?

realm/realm-library/src/main/java/io/realm/MutableRealmInteger.java
cmelchior: No, `iff` is not a good term if we want to make our documentation easy to read.

realm/realm-annotations-processor/src/main/java/io/realm/processor/ClassMetaData.java
bmeike: I'm not sure it is check completely, yet.  It will be though.  Perhaps a *FIXME* is in order...

As far as indexing, I don't actually see a reason to forbid it.  Seems odd to me but I can certainly allow it!

realm/realm-annotations-processor/src/main/java/io/realm/processor/ClassMetaData.java
bmeike: Yeah...  I think that would change the meaning.  What @zaki50 said.

realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmProxyClassGenerator.java
bmeike: That seems odd to me.  I have never seen a constructor see uninitialized fields.  I can guarantee that this method sometimes runs before fields are initialized.  I may be missing something but I don't see how it can run after the constructor, but before field initialization.  It is possible that it runs after superclass constructor.  Didn't check that.

realm/realm-annotations-processor/src/test/java/io/realm/processor/RealmProcessorTest.java
bmeike: I would say so.  At least it isn't totally hidden, now.

realm/realm-annotations-processor/src/test/resources/io/realm/AllTypesRealmProxy.java
bmeike: Yeah... I just roughed this stuff in.  There are no JSON tests yet.

realm/realm-library/src/androidTest/java/io/realm/RealmIntegerTests.java
bmeike: Yeah.  Good call.

realm/realm-library/src/androidTest/java/io/realm/RealmIntegerTests.java
bmeike: I can do this if you like, but I think it is a horrible idea.  As you well know, defining the semantics for this feature has been a long and painful discussion, often plagued by misunderstanding.  The idea that English which is never checked for accuracy, better represents the semantic definition, than the code, seems completely mistaken, to me.

realm/realm-library/src/main/java/io/realm/MutableRealmInteger.java
bmeike: Exactly.  Neither one can be passed, safely, between threads.  Perhaps I should word that differently?

realm/realm-library/src/main/java/io/realm/MutableRealmInteger.java
bmeike: Standard term.  "Java Concurrency in Practice", section 3.5.
tl;dr, just what it sounds like: you have to do some kind of synchronization to pass something to another thread.

realm/realm-library/src/main/java/io/realm/MutableRealmInteger.java
bmeike: Yes, surely.  This is an intermediate commit.  I needed some way to make existing tests work.  Value gets removed.  Added FIXME comment

realm/realm-library/src/main/java/io/realm/MutableRealmInteger.java
bmeike: I lost this discussion once before.  iff -> if.

realm/realm-library/src/main/java/io/realm/MutableRealmInteger.java
bmeike: Ya.  As you can see, it won't NPE.  Copy pasta.  Deleted.

realm/realm-library/src/main/java/io/realm/MutableRealmInteger.java
bmeike: Yeah.  Leftovers from the previous implementation.  Good catch.

realm/realm-annotations-processor/src/test/java/io/realm/processor/RealmProcessorTest.java
bmeike: This is not in the spec.  Yes, I know that is because no spec exists.

Ok.  Will add to the spec.  Allowing both @Required and @Index seem like good choices.

realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmJsonTypeHelper.java
bmeike: Fixed

realm/realm-library/src/main/java/io/realm/MutableRealmInteger.java
bmeike: bleh!  Yes.  Good catch.

realm/realm-annotations-processor/src/test/java/io/realm/processor/RealmProcessorTest.java
cmelchior: I would also add a FIXME, just so it isn't forgotten.

realm/realm-annotations-processor/src/main/java/io/realm/processor/ClassMetaData.java
```diff
@@ -332,11 +337,13 @@ private boolean checkDefaultConstructor() {
 
     private boolean checkForFinalFields() {
         for (VariableElement field : fields) {
-            if (field.getModifiers().contains(Modifier.FINAL)) {
-                Utils.error(String.format(Locale.US,
-                        "Class \"%s\" contains illegal final field \"%s\".", className, field.getSimpleName().toString()));
-                return false;
-            }
+            if (!field.getModifiers().contains(Modifier.FINAL)) { continue; }
```
bmunkholm: please have `continue`and `return` on it's own line.

cmelchior: We introduced an automatic formatter that allows this kind of code. No-one on the team found it hindered readability, nor could lead to errors.

bmeike: Sure.  I am the person that introduced the code formatter and I did run it on this code.

I believe that the team policy is that if the formatter accepts it, it is ok.  We can change the formatter, if you think it important.  The point, however, is that delaying a commit for yet another day, for an un-automated style complaint, is a significant impediment to cross-time-zone development.

bmeike:
@cmelchior Sure.  IMO, that is the whole point of the @Ignore.  We should not have @Ignore d tests!  Every @Ignore documents a broken test, in a way that is visible *every* time you run the tests.

IMO, commenting out the @Test is criminal and should be punished with a severe lashing with a damp spaghetti noodle.  @Ignore -ing a test indicates that it is broken and needs to be fixed or removed.

That said, will add the FIXME

Re the merge, sorry.  I forgot to push the merged feature branch to origin.  Should be fixed now.

realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmProxyInterfaceGenerator.java
bmeike: fixed

realm/realm-annotations-processor/src/test/java/io/realm/processor/RealmProcessorTest.java
bmeike: ok!  done.

realm/realm-library/src/androidTest/java/io/realm/RealmIntegerTests.java
bmeike: fixed.  you got your own personal commit for this...

cmelchior:
Good point about `@Ignore` being very visible in the log. That is probably just my habit of searching for them 😄

PR #4928, opened 2017-07-06
Update sync 2.0.0-rc11

beeender:
There are some breaking changes from sync/core in this release:

- Object ID column name has been renamed from __OID to !OID
  Instead of hardcoding it in java, we read it from sync if it is built
  with sync. Otherwise hardcode it in JNI.
- SortDescriptor and DistinctDescriptor are separated into two classes.
  Change the related JavaSortDescriptor to adapt this change.
- Since core started using cmake, the openssl objects are not a part of
  core release anymore. Instead, it will be released to s3
  independently.

**DON'T MERGE THIS before followings done!!!!**
- [x] https://github.com/realm/realm-object-store/pull/467 needs to be merged first
- [x] Update the Object Store submodule.
- [x] Update ROS version.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,14 +1,15 @@
-> We LOVE to help with any issues or bug you have!
+<!---
 
-> **Questions**: If you have questions about how to use Realm, please ask on [SO](http://stackoverflow.com/questions/ask?tags=realm) - we monitor the Realm tag.
+**Questions**: If you have questions about how to use Realm, ask on
+[StackOverflow](http://stackoverflow.com/questions/ask?tags=realm).
+We monitor the `realm` tag.
 
-> **Feature Request**: Just fill in the first two sections below.
+**Feature Request**: Just fill in the first two sections below.
 
-> **Bugs**: To help you as fast as possible with an issue or bug please describe your issue and the steps you have taken to reproduce it in as many details as possible.
->
-> Thanks for helping us help you :-)
->
-> Remove this and above before submitting.
+**Bugs**: To help you as fast as possible with an issue please describe your issue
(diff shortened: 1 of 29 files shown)


beeender:
@kneth I thought there would be a breaking change for distinct, but it seems all tests are fine?

beeender:
@emanuelez help to take a look at the cmake part.

Zhuinden:
@beeender core had a fix/breaking change/whatever that allows `realm.where().findAll().distinct()` to use `distinct()` on the query results themselves instead of the full table.

beeender:
Also, since the `SortDescriptor` and `DistinctDescriptor` are separated in core, it would be good to do the same in java layer. I will do that in another PR -- not that urgent, we loose the type safety checking when calling into JNI anyway.

beeender:
> If that core change is merged, I think @Deprecated should be removed from RealmResults.distinct().

I would still like to remove `Results.distinct()` majorly because of it can be achieved easily by contaminating other APIs:
```
RealmQuery query = realm.where(someclass).equalTo(1, "field);
RealmResults results = query.findAll();
results.distinct();
query.distinct(); // same
```
plus the semantics on `RealmResults.distinct()` is so unclear, it is not doing distinct on the original results, instead, it will create a new one.

also, to support that, also need to have `RealmResults.sort()`, same story. But all those can be achieved by `RealmQuery.distinct()`.

kneth:
@beeender The new `distinct` will not be available before https://github.com/realm/realm-object-store/pull/467 is merged.

beeender:
@kneth This is actually using commits from https://github.com/realm/realm-object-store/pull/467

Zhuinden:
@kneth https://github.com/realm/realm-object-store/pull/467 is merged 😄

PR #4934, opened 2017-07-07
Upgrade Realm Core to 2.8.6

cmelchior:
This introduces a breaking change to how SSL is configured. https://github.com/realm/realm-object-store/pull/493 addressed this.

Close #4791 #4788


Zhuinden:
so close to fixing indexed case insensitive queries, too!

beeender:
@zaki50 it is only a internal API breaking change. Nothing is changed for users.

PR #4942, opened 2017-07-11
Create ProxyUtils class to simplify proxies

bmeike:
New class ProxyUtils to reduce size of proxies.


realm/realm-library/src/main/java/io/realm/internal/ProxyUtils.java
```diff
@@ -0,0 +1,37 @@
+package io.realm.internal;
+/*
```
beeender: license before `package`

bmeike:
Merging with the updated releases was dangerously complicated.  Since there is only a single commit in the branch, I, instead, cherry-picked it onto the new `releases` branch.  Both comments have been addressed.  Thanks @dalinaum and @beeender

PR #4951, opened 2017-07-11
Add multiple user UT

bmeike:
Add one more UT in light of the Cartasite work.

### DIFF
--- a/realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/AuthTests.java
+++ b/realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/AuthTests.java
@@ -405,8 +405,7 @@ public class AuthTests extends BaseIntegrationTest {
         final SyncUser[] users = new SyncUser[3];
 
         for (int i = 0; i < users.length; i++) {
-            SyncCredentials credentials = SyncCredentials.usernamePassword(UUID.randomUUID().toString(), password,
-                    true);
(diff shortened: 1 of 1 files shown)


realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/AuthTests.java
```diff
@@ -447,6 +446,21 @@ public void singleUserCanBeLoggedInAndOutRepeatedly() {
         }
     }
 
+    // verify that multiple users can be logged in at the same time
```
beeender: I don't quite get the point of this UT. From the description, it would be the same with `multipleUsersCanBeLoggedInSimultaneously`. But from the implementation, it tries to test login-logout with different users? Maybe change the description of the test?

bmeike: Trying to duplicate a problem that Cartasite is having.  The distinction is whether all of the users are logged in at once, or whether they are logged in serially.  Neither seems to drive the Cartasite bug, so maybe it isn't useful.

cmelchior: Seems the Cartasite issue was captured by another unit test by @beeender. Perhaps we should just close this PR?

bmeike:
abandoned

PR #4953, opened 2017-07-12
Counters: Wired to core

bmeike:
Next step in  #4266


realm/realm-annotations-processor/src/main/java/io/realm/processor/ClassMetaData.java
```diff
@@ -414,21 +413,31 @@ private boolean categorizeField(Element element) {
...
-            switch (realmType) {
-                case STRING:
-                case DATE:
-                case INTEGER:
-                case BOOLEAN:
-                    indexedFields.add(variableElement);
-                    return true;
+        boolean indexable = false;
+
+        if (Utils.isMutableRealmInteger(variableElement)) {
```
beeender: Is this necessary? wouldn't `variableElement.asType()` be `INTEGER` for `RealmMutableInteger` as well?

realm/realm-annotations-processor/src/test/resources/io/realm/AllTypesRealmProxy.java
```diff
@@ -900,9 +899,9 @@ public static long insert(Realm realm, some.test.AllTypes object, Map<RealmModel
         if (realmGet$columnBinary != null) {
             Table.nativeSetByteArray(tableNativePtr, columnInfo.columnBinaryIndex, rowIndex, realmGet$columnBinary, false);
         }
-        io.realm.MutableRealmInteger realmGet$columnMutableRealmInteger = ((AllTypesRealmProxyInterface) object).realmGet$columnMutableRealmInteger();
-        if ((realmGet$columnMutableRealmInteger != null) && (realmGet$columnMutableRealmInteger.get() != null)) {
-            Table.nativeSetLong(tableNativePtr, columnInfo.columnMutableRealmIntegerIndex, rowIndex, realmGet$columnMutableRealmInteger.get().longValue(), false);
+        Long realmGet$columnMutableRealmInteger = ((AllTypesRealmProxyInterface) object).realmGet$columnMutableRealmInteger().get();
+        if (realmGet$columnMutableRealmInteger != null) {
```
beeender: Don't you need set `null` when `realmGet$columnMutableRealmInteger` is `null`?

realm/realm-annotations-processor/src/test/resources/io/realm/AllTypesRealmProxy.java
```diff
@@ -968,9 +967,9 @@ public static void insert(Realm realm, Iterator<? extends RealmModel> objects, M
             if (realmGet$columnBinary != null) {
                 Table.nativeSetByteArray(tableNativePtr, columnInfo.columnBinaryIndex, rowIndex, realmGet$columnBinary, false);
             }
-            io.realm.MutableRealmInteger realmGet$columnMutableRealmInteger = ((AllTypesRealmProxyInterface) object).realmGet$columnMutableRealmInteger();
-            if ((realmGet$columnMutableRealmInteger != null) && (realmGet$columnMutableRealmInteger.get() != null)) {
-                Table.nativeSetLong(tableNativePtr, columnInfo.columnMutableRealmIntegerIndex, rowIndex, realmGet$columnMutableRealmInteger.get().longValue(), false);
+            Long realmGet$columnMutableRealmInteger = ((AllTypesRealmProxyInterface) object).realmGet$columnMutableRealmInteger().get();
+            if (realmGet$columnMutableRealmInteger != null) {
```
beeender: same

realm/realm-annotations-processor/src/test/resources/io/realm/AllTypesRealmProxy.java
```diff
@@ -772,11 +772,14 @@ public static String getTableName() {
     }
 
     public static some.test.AllTypes copyOrUpdate(Realm realm, some.test.AllTypes object, boolean update, Map<RealmModel,RealmObjectProxy> cache) {
-        if (object instanceof RealmObjectProxy && ((RealmObjectProxy) object).realmGet$proxyState().getRealm$realm() != null && ((RealmObjectProxy) object).realmGet$proxyState().getRealm$realm().threadId != realm.threadId) {
-            throw new IllegalArgumentException("Objects which belong to Realm instances in other threads cannot be copied into this Realm instance.");
-        }
-        if (object instanceof RealmObjectProxy && ((RealmObjectProxy) object).realmGet$proxyState().getRealm$realm() != null && ((RealmObjectProxy) object).realmGet$proxyState().getRealm$realm().getPath().equals(realm.getPath())) {
-            return object;
+        if (object instanceof RealmObjectProxy && ((RealmObjectProxy) object).realmGet$proxyState().getRealm$realm() != null) {
```
beeender: good improvement! But you may have conflicts when merging `feature/counters` ... this can actually be merged to `releases` now... also make this PR smaller.

realm/realm-annotations-processor/src/test/resources/some/test/Counter_Ignored.java
```diff
@@ -0,0 +1,16 @@
+package some.test;
+
+import io.realm.MutableRealmInteger;
+import io.realm.RealmObject;
+import io.realm.annotations.Ignore;
+
+public class Counter_Ignored extends RealmObject {
```
beeender: For those kind of test, you can actually use `TestRealmObjectFileObject` which doesn't need to create a model classes anymore and make the tests much easier. See `compileInvalidRequiredTypes` for example.

realm/realm-library/src/androidTest/java/io/realm/MutableRealmIntegerTests.java
```diff
@@ -366,12 +363,12 @@ private void testNullability(MutableRealmIntegerTypes c1) {
             c1.columnMutableRealmInteger.increment(5);
             fail("Attempt to increment a null valued MutableRealmInteger should throw NPE");
         }
-        catch (NullPointerException ignore) { }
+        catch (IllegalArgumentException ignore) { }
```
beeender: Isn't `IlegalStateException` more accurate? Also the message for `fail()` needs to be changed.

realm/realm-library/src/main/cpp/io_realm_internal_Table.cpp
```diff
@@ -666,6 +666,25 @@ JNIEXPORT void JNICALL Java_io_realm_internal_Table_nativeSetLong(JNIEnv* env, j
...
+JNIEXPORT void JNICALL Java_io_realm_internal_Table_nativeIncrementLong(JNIEnv* env, jclass, jlong nativeTablePtr,
+                                                                  jlong columnIndex, jlong rowIndex, jlong value)
+{
+    if (!TBL_AND_INDEX_AND_TYPE_VALID(env, TBL(nativeTablePtr), columnIndex, rowIndex, type_Int)) {
+        return;
+    }
+
+    Table* table = TBL(nativeTablePtr);
+    if (table->is_null(columnIndex, rowIndex)) {
+        ThrowException(env, IllegalArgument, "Cannot increment a MutableRealmInteger whose value is null. Set its value first.");
```
beeender: we try to use the new way to throw java exception in JNI code.
just move line 676-680 to try block and then change this line to:
```
THROW_JAVA_EXCEPTION(env, JavaExceptionDef::IllegalArgument,
                         "Cannot increment a MutableRealmInteger whose value is null. Set its value first.");
```

You can search for `THROW_JAVA_EXCEPTION` to get some examples.

realm/realm-library/src/main/java/io/realm/MutableRealmInteger.java
```diff
@@ -119,45 +120,53 @@ public boolean isValid() {
...
-//            row.getTable().setLong(columnIndex, row.getIndex(), value, true);
-//            return;
-//        }
-//        row.setLong(columnIndex,value);
         @Override
         public void set(Long value) {
             realm.checkIfValidAndInTransaction();
-            this.value = value;
+
+            if (!proxyState.isUnderConstruction()) {
```
beeender: I assume this method won't be used by the dynamic API, is it correct?

realm/realm-library/src/main/java/io/realm/MutableRealmInteger.java
```diff
@@ -57,41 +68,31 @@ public Long get() {
...
 
     /**
      * Managed Implementation.
      */
-    // FIXME MutableRealmIntegers: wire up to native increment method
     static class Managed extends MutableRealmInteger {
         private final ProxyState<?> proxyState;
         private final BaseRealm realm;
         private final Row row;
         private final long columnIndex;
```
zaki50: `columnIndex` can be updated by `SharedRealm.refresh()` because of additive schema change from ROS.
We always need to refer `FooColumnInfo` in proxy class if we use columnIndex cache. Or proxy class must update this `columnIndex` field.

realm/realm-annotations-processor/src/main/java/io/realm/processor/ClassMetaData.java
bmeike: It would be nice, but that won't work.  It is treated differently than an Integer in a number of places, so putting it into the JAVA_TO_REALM_TYPES list won't work.

realm/realm-annotations-processor/src/main/java/io/realm/processor/ClassMetaData.java
```diff
@@ -505,6 +514,7 @@ private boolean categorizeMutableRealmIntegerField(VariableElement field) {
 
         containsMutableRealmInteger = true;
 
+
```
bmeike: Fixed.

realm/realm-annotations-processor/src/test/resources/io/realm/AllTypesRealmProxy.java
bmeike: Yes.  I *have* been trying to do that.  The problem is that @zaki50 will then block it for at least a day, causing this project to be blocked.  Being in the -8 time zone is a pain! ;-)

realm/realm-annotations-processor/src/test/resources/io/realm/AllTypesRealmProxy.java
bmeike: I'm not sure about that.  This is exactly the same semantics as a "Long".  I need to look into this.  Will add "FIXME"

realm/realm-annotations-processor/src/test/resources/some/test/Counter_Ignored.java
bmeike: Cool!  Will have a look.

realm/realm-library/src/androidTest/java/io/realm/MutableRealmIntegerTests.java
bmeike: Well, close.  NPE would be the most accurate.  To do that, though, I'd have to catch the IAE thrown by the native code, check the message, and throw the NPE.  I even think that might be the right thing to do.

realm/realm-library/src/main/cpp/io_realm_internal_Table.cpp
bmeike: Yeah... I really hate having multiple ways of doing something, in a single file.  I argue strongly for keeping a consistent style in a single file.  If we are going to change style, lets do it across the file.

realm/realm-annotations-processor/src/test/resources/io/realm/AllTypesRealmProxy.java
bmeike: Yep.

realm/realm-library/src/androidTest/java/io/realm/MutableRealmIntegerTests.java
bmeike: Using the suggestion below to change to ISE.

realm/realm-library/src/main/java/io/realm/MutableRealmInteger.java
bmeike: I'm not sure.  Adding a task to verify that this all works with the Dynamic API.

realm/realm-library/src/main/java/io/realm/MutableRealmInteger.java
```diff
@@ -57,41 +68,31 @@ public Long get() {
         @Override
         public void increment(long inc) {
             if (value == null) {
-                throw new NullPointerException("Attempt to increment a null valued MutableRealmInteger");
+                throw new IllegalArgumentException("Attempt to increment a null valued MutableRealmInteger");
```
bmeike: Should be NPE.  Settling for ISE.

realm/realm-library/src/main/java/io/realm/MutableRealmInteger.java
```diff
@@ -57,41 +68,31 @@ public Long get() {
...
             }
             value = Long.valueOf(value + inc);
```
bmeike: You are attempting to argue *against* clarity?  Sorry.  I prefer this.

realm/realm-library/src/main/java/io/realm/MutableRealmInteger.java
bmeike: Wow!  If that is true that is an awesome catch.  Can you please show how that could happen?

realm/realm-library/src/main/java/io/realm/MutableRealmInteger.java
```diff
@@ -119,45 +120,53 @@ public boolean isValid() {
 
         @Override
         public Long get() {
-            return value;
+            row.checkIfAttached();
+            return (isNull()) ? null : row.getLong(columnIndex);
```
bmeike: The "hot path" argument doesn't seem very significant.  On the other hand, being able to pull `isNull` back up to the super class totally convinces me.  Fixed.

realm/realm-library/src/main/java/io/realm/MutableRealmInteger.java
```diff
@@ -119,45 +120,53 @@ public boolean isValid() {
...
+
+        @Override
+        public boolean isNull() {
+            row.checkIfAttached();
+            return row.isNull(columnIndex);
+        }
+
+        private void setValue(Long value, boolean isDefault) {
+            Table t = row.getTable();
+            long i = row.getIndex();
```
bmeike: Changing it just for you!

realm/realm-annotations-processor/src/main/java/io/realm/processor/ClassMetaData.java
bmeike: I think you mean "too many spaces".  Fixed!

realm/realm-library/src/main/java/io/realm/MutableRealmInteger.java
```diff
@@ -57,41 +68,31 @@ public Long get() {
...
-    // FIXME MutableRealmIntegers: wire up to native increment method
     static class Managed extends MutableRealmInteger {
         private final ProxyState<?> proxyState;
         private final BaseRealm realm;
         private final Row row;
         private final long columnIndex;
-        private Long value; // !!! FIXME: remove!
 
         /**
          * Inject the proxy state into this managed MutableRealmInteger.
```
bmeike: Yeah.  Fixed.

realm/realm-library/src/main/cpp/io_realm_internal_Table.cpp
Zhuinden: I find it odd that one change is to `reinterpret_cast` from Macros, and the other is from `throwException` to a macro?

bmeike: You and me, both, pal.  When it comes to this stuff, I just do what they tell me to do.

realm/realm-annotations-processor/src/test/resources/io/realm/AllTypesRealmProxy.java
beeender: uh. ok, this is for creating a new managed-object. nullable field will have default null value is core. so this is fine 👍

realm/realm-library/src/androidTest/java/io/realm/MutableRealmIntegerTests.java
```diff
@@ -366,12 +379,12 @@ private void testNullability(MutableRealmIntegerTypes c1) {
             c1.columnMutableRealmInteger.increment(5);
             fail("Attempt to increment a null valued MutableRealmInteger should throw NPE");
         }
-        catch (NullPointerException ignore) { }
+        catch (IllegalStateException ignore) { }
```
zaki50: this is C++ style. it should be `} catch (IllegalStateException ignore) { }`

realm/realm-library/src/androidTest/java/io/realm/MutableRealmIntegerTests.java
```diff
@@ -366,12 +379,12 @@ private void testNullability(MutableRealmIntegerTypes c1) {
...
         try {
             c1.columnMutableRealmInteger.decrement(5);
             fail("Attempt to decrement a null valued MutableRealmInteger should throw NPE");
         }
-        catch (NullPointerException ignore) { }
+        catch (IllegalStateException ignore) { }
```
zaki50: this is C++ style. it should be `} catch (IllegalStateException ignore) { }`

realm/realm-library/src/main/java/io/realm/MutableRealmInteger.java
```diff
@@ -57,104 +68,92 @@ public Long get() {
...
-        public void increment(long inc) {
-            Long val = get();
-            if (val == null) {
-                throw new NullPointerException("Attempt to increment a null valued MutableRealmInteger");
-            }
-            set(val + inc);
+        public final void increment(long inc) {
+            getRealm().checkIfValidAndInTransaction();
+            Row row = getRow();
+            row.getTable().incrementLong(getColumnIndex(), row.getIndex(), inc);
```
zaki50: If you intend for `MutableRealmInteger` to support dynamic API, we must check if the column is not primary key.
Currently, counter does not support dynamic API, right?

realm/realm-library/src/androidTest/java/io/realm/MutableRealmIntegerTests.java
bmeike: Yep.  will fix.

realm/realm-library/src/androidTest/java/io/realm/MutableRealmIntegerTests.java
bmeike: yep.  will fix.

realm/realm-library/src/main/java/io/realm/MutableRealmInteger.java
bmeike: Yes.  I answered this above, for @beeender .  There is a task, [here](https://github.com/realm/realm-java/issues/4266#issuecomment-313474899) to make sure this all works for the dynamic API.

bmeike:
This is the bulk of the implementation of https://github.com/realm/realm-object-store/issues/357 for Java

PR #4960, opened 2017-07-12
Verify schema for backlinks by OS

beeender:
- Add computed properties in the proxy's expected schema info, so the
  schema validation in Object Store will check for linking objects.
- Throw IllegalStateException if the backlink schema validation fails.
  ISE is more suitable here since backlink does not exist in the schema.
  Normally the way to handle this issue is not by adding a migration
  block, instead, just changing the declaration in the model class.
- Rewrite and enable the tests which are related to the linking objects
  schema validation by mocking the expected schema info.


realm/realm-library/src/main/java/io/realm/internal/Property.java
```diff
@@ -33,17 +33,18 @@
     private static final long nativeFinalizerPtr = nativeGetFinalizerPtr();
 
     Property(String name, RealmFieldType type, boolean isPrimary, boolean isIndexed, boolean isRequired) {
-        this.nativePtr = nativeCreateProperty(name, type.getNativeValue(), isPrimary, isIndexed, !isRequired);
+        this.nativePtr = nativeCreatePersistedProperty(name, type.getNativeValue(), isPrimary, isIndexed, !isRequired);
```
zaki50: Shouldn't we have a check to make sure that `type` is not `LINKING_OBJECTS`?

beeender: Object Store has validation. It will throw later anyway. Also, it is something should never happen to user.

realm/realm-library/src/androidTest/java/io/realm/LinkingObjectsManagedTests.java
```diff
@@ -542,124 +544,111 @@ public void migration_backlinkedFieldInUse() {
...
-     * {@code
-     * @LinkingObjects("child")
-     *     private final RealmResults<XXXBacklinksSource> parents = null;
-     * }
-     * </pre>
-     * If the both classes were used in the configuration, the test would fail because of the
-     * missing class.  Since the configuration contains only the single class `BacklinksTarget`,
-     * basic validation passes.  Backlink validation, however, should fail, seeking the
-     * `BacklinksSource` table.
+     * Schema validation should fail if the backinked column points to a non-existent class.
```
cmelchior: backinked -> backing

realm/realm-library/src/main/java/io/realm/internal/Property.java
zaki50: then :+1:

realm/realm-library/src/androidTest/java/io/realm/LinkingObjectsManagedTests.java
dalinaum: There are other `backinked colum`, should they be changed too?

PR #4969, opened 2017-07-14
Counters: Required, Index and Documentation

bmeike:
Add tests for @Required and @Index
Document ManagedRealmInteger

### DIFF
--- a/realm/realm-library/src/androidTest/java/io/realm/MutableRealmIntegerTests.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/MutableRealmIntegerTests.java
@@ -24,6 +24,7 @@ import org.junit.Test;
 import org.junit.runner.RunWith;
 
 import io.realm.entities.MutableRealmIntegerTypes;
+import io.realm.internal.Table;
 import io.realm.rule.RunInLooperThread;
 import io.realm.rule.TestRealmConfigurationFactory;
 
@@ -34,7 +35,6 @@ import static junit.framework.Assert.fail;
 
 
 // FIXME MutableRealmInteger: Need JSON tests.
-// FIXME MutableRealmInteger: Need tests for @Required, @Index
 
 @RunWith(AndroidJUnit4.class)
 public class MutableRealmIntegerTests {
@@ -105,7 +105,7 @@ public class MutableRealmIntegerTests {
         realm.commitTransaction();
 
         realm.beginTransaction();
-        testBasic(c1.columnMutableRealmInteger, c2.columnMutableRealmInteger);
+        testBasic(c1.columnNullableMutableRealmInteger, c2.columnNullableMutableRealmInteger);
         realm.commitTransaction();
     }
 
@@ -152,6 +152,54 @@ public class MutableRealmIntegerTests {
         realm.commitTransaction();
     }
 
+    /**
+     * @Required MutableRealmIntegers should not be nullable.
+     * There are other tests testing nullabilty: just need to test @Required here.
+     * There is no attempt to control the nullability of an unmanaged MutableRealmInteger.
+     * And attempt to copy an unmanaged model object with a null-valued MutableRealmInteger
+     * into an @Required field should fail.
+     */
+    @Test
+    public void required() {
+        realm.beginTransaction();
+        MutableRealmIntegerTypes c1 = realm.createObject(MutableRealmIntegerTypes.class);
+        realm.commitTransaction();
+
+        Table table = realm.getTable(MutableRealmIntegerTypes.class);
+        assertFalse(table.isColumnNullable(table.getColumnIndex(MutableRealmIntegerTypes.FIELD_NONNULLABLE_MUTABLEREALMINTEGER)));
+
+        try {
+            c1.columnNonNullableMutableRealmInteger.set(null);
+            fail("should not be able to set an @Required MutableRealmInteger null");
+        } catch(IllegalStateException ignore) {
+        }
+
+        c1 = new MutableRealmIntegerTypes();
+        c1.columnNonNullableMutableRealmInteger.set(null);
+        realm.beginTransaction();
+        try {
+            MutableRealmIntegerTypes c2 = realm.copyToRealm(c1);
+            fail("should not be able to copy an null valute to a @Required MutableRealmInteger");
+        } catch(IllegalArgumentException ignore) {
+        }
+        realm.commitTransaction();
+    }
+
+
+    /**
+     * @Indexed MutableRealmIntegers should have indices.
+     * Without @Index they should not.
+     */
+    @Test
+    public void indexed() {
+        realm.beginTransaction();
+        MutableRealmIntegerTypes c1 = realm.createObject(MutableRealmIntegerTypes.class);
+        realm.commitTransaction();
+
+        Table table = realm.getTable(MutableRealmIntegerTypes.class);
+        assertTrue(table.hasSearchIndex(table.getColumnIndex(MutableRealmIntegerTypes.FIELD_INDEXED_MUTABLEREALMINTEGER)));
+        assertFalse(table.hasSearchIndex(table.getColumnIndex(MutableRealmIntegerTypes.FIELD_NULLABLE_MUTABLEEALMINTEGER)));
+    }
 
     /**
      * Be absolutely certain that we can actually compare two longs.
@@ -175,9 +223,9 @@ public class MutableRealmIntegerTests {
     @Test
     public void compareTo_managed() {
         realm.beginTransaction();
-        MutableRealmInteger ri1 = realm.createObject(MutableRealmIntegerTypes.class).getColumnMutableRealmInteger();
+        MutableRealmInteger ri1 = realm.createObject(MutableRealmIntegerTypes.class).getColumnNullableMutableRealmInteger();
         ri1.set(0);
-        MutableRealmInteger ri2 = realm.createObject(MutableRealmIntegerTypes.class).getColumnMutableRealmInteger();
+        MutableRealmInteger ri2 = realm.createObject(MutableRealmIntegerTypes.class).getColumnNullableMutableRealmInteger();
         ri2.set(Long.MAX_VALUE);
         realm.commitTransaction();
         assertEquals(-1, ri1.compareTo(ri2));
@@ -199,10 +247,10 @@ public class MutableRealmIntegerTests {
     @Test
     public void updateOutsideTransactionThrows() {
         realm.beginTransaction();
-        realm.createObject(MutableRealmIntegerTypes.class).getColumnMutableRealmInteger().set(42);
+        realm.createObject(MutableRealmIntegerTypes.class).getColumnNullableMutableRealmInteger().set(42);
         realm.commitTransaction();
 
-        MutableRealmInteger managedRI = realm.where(MutableRealmIntegerTypes.class).findFirst().getColumnMutableRealmInteger();
+        MutableRealmInteger managedRI = realm.where(MutableRealmIntegerTypes.class).findFirst().getColumnNullableMutableRealmInteger();
         try {
             managedRI.set(1);
             fail("Setting a managed MutableRealmInteger outside a transaction should fail");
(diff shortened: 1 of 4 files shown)


realm/realm-library/src/main/java/io/realm/MutableRealmInteger.java
```diff
@@ -23,14 +23,68 @@
 
 /**
  * A MutableRealmInteger is a mutable, {@link java.lang.Long}-like numeric quantity.
- * MutableRealmInteger implementations are not thread safe.  Like all Realm objects,
- * managed MutableRealmIntegers may not be moved across threads.  Unmanaged MutableRealmIntegers
- * may be moved across threads but require safe publication.
+ * It behaves almost exactly as a reference to a {@link java.lang.Long}. More specifically:
+ * <ul>
+ * <li>The MutableRealmInteger may have the value {@code null}.</li>
```
bmeike: Should be "A MutableRealmInteger..."?

realm/realm-library/src/androidTest/java/io/realm/MutableRealmIntegerTests.java
```diff
@@ -152,6 +152,54 @@ public void validAndManaged_managed() {
...
+     * And attempt to copy an unmanaged model object with a null-valued MutableRealmInteger
+     * into an @Required field should fail.
+     */
+    @Test
+    public void required() {
+        realm.beginTransaction();
+        MutableRealmIntegerTypes c1 = realm.createObject(MutableRealmIntegerTypes.class);
+        realm.commitTransaction();
+
+        Table table = realm.getTable(MutableRealmIntegerTypes.class);
```
beeender: Use `realm.getSchema().get("MutableRealmIntegerTypes").isNullable(MutableRealmIntegerTypes.FIELD_NONNULLABLE_MUTABLEREALMINTEGER)` instead. We need to reduce the usage of low level APIs in the test if it is possible. Since those low level API might be changed in the future (we might move more things to Object Store.)

realm/realm-library/src/androidTest/java/io/realm/MutableRealmIntegerTests.java
```diff
@@ -152,6 +152,54 @@ public void validAndManaged_managed() {
...
+     * @Indexed MutableRealmIntegers should have indices.
+     * Without @Index they should not.
+     */
+    @Test
+    public void indexed() {
+        realm.beginTransaction();
+        MutableRealmIntegerTypes c1 = realm.createObject(MutableRealmIntegerTypes.class);
+        realm.commitTransaction();
+
+        Table table = realm.getTable(MutableRealmIntegerTypes.class);
```
beeender: ditto, use `hasIndex()`

realm/realm-library/src/androidTest/java/io/realm/MutableRealmIntegerTests.java
```diff
@@ -152,6 +152,54 @@ public void validAndManaged_managed() {
         realm.commitTransaction();
     }
 
+    /**
+     * @Required MutableRealmIntegers should not be nullable.
+     * There are other tests testing nullabilty: just need to test @Required here.
+     * There is no attempt to control the nullability of an unmanaged MutableRealmInteger.
+     * And attempt to copy an unmanaged model object with a null-valued MutableRealmInteger
```
cmelchior: And -> An

realm/realm-library/src/androidTest/java/io/realm/MutableRealmIntegerTests.java
cmelchior: You can use `realm.getSchema().get("MutableRealmIntegerTypes").isNullable(field)` which is our public API. I would avoid using private API's in unit tests if at all possible.

realm/realm-library/src/main/java/io/realm/MutableRealmInteger.java
```diff
@@ -23,14 +23,68 @@
...
+ * MutableRealmInteger counter = realm.where(Users.class)
+ *     .equalTo("name", Fred)
+ *     .counter.increment(1);
+ * }
+ * </pre>
+ * Note that the {@code .set} operator must be used with extreme care. It destroys any guarantees of consistency.
  * <p>
- * TODO: More complete docs, including examples of use.
+ *
+ * MutableRealmIntegers may not be primary keys. Their implementations are not thread safe.
```
beeender: `thread safe` -> `thread confined`? well, the unmanged `MutableRealmIntegers` is not thread safe, but the managed `MutableRealmIntegers` is, but thread confined.

realm/realm-library/src/main/java/io/realm/MutableRealmInteger.java
```diff
@@ -23,14 +23,68 @@
...
 /**
  * A MutableRealmInteger is a mutable, {@link java.lang.Long}-like numeric quantity.
- * MutableRealmInteger implementations are not thread safe.  Like all Realm objects,
- * managed MutableRealmIntegers may not be moved across threads.  Unmanaged MutableRealmIntegers
- * may be moved across threads but require safe publication.
+ * It behaves almost exactly as a reference to a {@link java.lang.Long}. More specifically:
+ * <ul>
+ * <li>The MutableRealmInteger may have the value {@code null}.</li>
+ * <li>The {@code .equals} operator compares the contained Long values. null-valued MutableRealmInteger are {@code .equals}</li>
+ * <li>The {@code .compareTo} operator compares the contained Long values.  It considers {@code null} &lt; any value.</li>
```
cmelchior: Not sure I understand the `any value` part of this?

realm/realm-library/src/main/java/io/realm/MutableRealmInteger.java
```diff
@@ -23,14 +23,68 @@
...
+ * {@code
+ * @Required
+ * public final MutableRealmInteger counter = MutableRealmInteger.valueOf(0L);
+ * }
+ * </pre>
+ *
+ *<p>
+ * A reference to a managed MutableRealmInteger is subject to all of the constraints that apply to the model object
+ * from which it was obtained: It can only be mutated within a transaction. It becomes invalid if
+ * the Realm backing it is closed. And so on. Use the {@code .isManaged} and {@code .isValid} operators to
```
cmelchior: I would remove `And so on`. If any other properties are important we should spell them out. 
You can use {@link #isManaged} instead of `{@code .isManaged}`, same for `.isValid`

realm/realm-library/src/main/java/io/realm/MutableRealmInteger.java
```diff
@@ -23,14 +23,68 @@
...
+ * executed on two different devices, simultaneously, even if connected by only a slow, unreliable network,
+ * will <b>always</b> cause the value of {@code counter} to converge, eventually on the value 2.
+ * <pre>
+ * {@code
+ * MutableRealmInteger counter = realm.where(Users.class)
+ *     .equalTo("name", Fred)
+ *     .counter.increment(1);
+ * }
+ * </pre>
+ * Note that the {@code .set} operator must be used with extreme care. It destroys any guarantees of consistency.
```
cmelchior: Not entirely true, but it gets a lot harder to reason about since `set` will override previous `increment/decrement` calls, so it will suddenly depend on the merge order what happens.

Perhaps something like

---

Note that the {@link #set(Long)} operator must be used with extreme care. It will override any method calls to {@link #increment(Long)} or {@link #decrement(Long)}. The value will always converge to the same result across all devices, but the value will depend on the order the operations where merged. Mixing {@link #set(Long)} with {@link #increment(Long)} is therefor not advised, unless fuzzy counting is acceptable.

realm/realm-library/src/main/java/io/realm/MutableRealmInteger.java
cmelchior: `A MutableInteger` is probably better in the header description IMO

bmeike: ?  It has to be "MutableRealmInteger", right?

realm/realm-library/src/androidTest/java/io/realm/MutableRealmIntegerTests.java
bmeike: awesome.  Will do.

bmeike: @beeender made the same point, I think.  Definitely will fix.

realm/realm-library/src/main/java/io/realm/MutableRealmInteger.java
bmeike: I mean to say that null == null, null < x, for any x.  Suggest an alternative wording?

realm/realm-library/src/main/java/io/realm/MutableRealmInteger.java
bmeike: Sure.  Will replace.

realm/realm-library/src/main/java/io/realm/MutableRealmInteger.java
bmeike: Ok.  Will do.

realm/realm-library/src/androidTest/java/io/realm/MutableRealmIntegerTests.java
bmeike: Yes.  Will fix.

realm/realm-library/src/main/java/io/realm/MutableRealmInteger.java
bmeike: Is what I said inaccurate?  If so, will, totally, change it.

realm/realm-library/src/androidTest/java/io/realm/MutableRealmIntegerTests.java
bmeike: Yep.  Will fix.

realm/realm-library/src/main/java/io/realm/MutableRealmInteger.java
cmelchior: Yes, sorry.

realm/realm-library/src/main/java/io/realm/MutableRealmInteger.java
cmelchior: I think we just generally have an aversion against calling it `not thread safe` when it actually means `thread confined`. In my mind `not thread safe` means you _can_ access it from multiple threads, but need external synchronization to do so, while `thread confined`, means you cannot do that, period.

cmelchior: There is a small distinction here between managed and unmanaged variants, but generally we have only been describing the managed behaviour, unless there are really good reasons to do otherwise.

realm/realm-library/src/main/java/io/realm/MutableRealmInteger.java
cmelchior: Ah sorry, it was the html escaping that threw me off. This is fine 👍

realm/realm-library/src/main/java/io/realm/MutableRealmInteger.java
cmelchior: Feel free to rephrase it. It was mostly the `It destroys any guarantees` part of your statement I had a problem with since that isn't exactly true.

bmeike:
Penultimate commit for https://github.com/realm/realm-java/issues/4266

realm/realm-library/src/main/java/io/realm/MutableRealmInteger.java
```diff
@@ -74,7 +78,7 @@
  *<p>
  * A reference to a managed MutableRealmInteger is subject to all of the constraints that apply to the model object
  * from which it was obtained: It can only be mutated within a transaction. It becomes invalid if
- * the Realm backing it is closed. And so on. Use the {@code .isManaged} and {@code .isValid} operators to
+ * the Realm backing it is closed. Use the {@link #isManaged}  and {@link #isManaged}  operators to
```
Zhuinden: isManaged and isManaged? Is that intentional?

bmeike: Whoah!  No!  Definitely not.

PR #4985, opened 2017-07-18
Add support for SyncSession.uploadAllLocalChanges()

cmelchior:
Mostly to mirror `SyncSession.downloadAllServerChanges()`, but this should also prove quite helpful for a lot of our Sync integration tests.


realm/realm-library/src/objectServer/java/io/realm/SyncSession.java
```diff
@@ -609,12 +660,12 @@ private void clearScheduledAccessTokenRefresh() {
          */
         public void waitForServerChanges() throws InterruptedException {
             if (!resultReceived) {
-                waitForChanges.await();
+                waiter.await();
```
cmelchior: Good catch. Will fix that.

PR #4989, opened 2017-07-19
Ignore callbacks from Object Store if waiting was canceled.

cmelchior:
Followup from https://github.com/realm/realm-java/pull/4985#discussion_r128153265

It does not appear possible to cancel the notification, so instead, we just ignore the callback.

Note that Sync has a synchronous method for waiting for upload/download to complete, but since all Object Store methods are behind a session mutex, exposing that would result in us blocking _all_ other interaction with the session.

This is against `master` as the refactor was also merged to `master`


realm/realm-library/src/objectServer/java/io/realm/SyncSession.java
```diff
@@ -387,7 +388,12 @@ private void waitForChanges(int direction) throws InterruptedException {
 
                 throw new ObjectServerError(ErrorCode.UNKNOWN, errorMsg + " Has the SyncClient been started?");
             }
-            wrapper.waitForServerChanges();
+            try {
+                wrapper.waitForServerChanges();
+            } catch(InterruptedException e) {
+                waitingForServerChanges.set(null); // Ignore any results being sent if the wait was interrupted.
```
zaki50: This is not enough.
There is a chance that users add another listener after the cancellation.

```
SyncSession session = ...;
try {
    session.downloadAllServerChanges();
} catch (InterruptedException e) {
    // cancelled
}

...  // so some works
session.uploadAllLocalChanges(); // this may return too early because of callback caused by downloadAllServerChanges() before
```

cmelchior: Hmm, good point. It looks like this will require changes to object store as it stands though. Right now we don't have any way to cancel this by the looks of it

cmelchior: So, canceling was neither supported in Object Store or Sync. So I introduced a callback Id instead that is used to determine if a notification should be used or not.

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/SyncSessionTests.java
```diff
@@ -106,4 +109,69 @@ public void uploadDownloadAllChanges() throws InterruptedException {
...
+            @Override
+            public void run() {
+                Realm userRealm = Realm.getInstance(userConfig);
+                userRealm.beginTransaction();
+                userRealm.createObject(AllTypes.class);
+                userRealm.commitTransaction();
+                SyncSession userSession = SyncManager.getSession(userConfig);
+                try {
+                    // 1. Start download (which will be interrupted)
+                    userInterruptLatch.countDown();
```
beeender: this will be a little bit flaky. there will still be a small chance the thread set to interrupted state after `downloadAllServerChanges()`. Actually the interrupted state can be set inside the current thread.
I think just call `Thread.currentThread().interrupt()` before the `downloadAllServerChanges()` would be enough and we can remove the latches.

Issue #4785, opened 2017-06-14
Breaking change Wishlist for 4.0


beeender:
This ticket is for general discussion of the breaking changes in the coming 4.0 release.

**Please edit the main post to add a topic with a sequence number, and we can discuss it in the comments. If the breaking change is decided to be worthy to add, then create a separated issue for it.**

1) [POSTPONED] Do not allow having both typed Realm instance and Dynamic Realm instance on the same thread. Throw an exception instead.
    This creates lots of troubles with Realm instance cache management. By disallowing it, the `RealmCache` implementation can be much simpler and give us an option to totally manage the Realm instance through Object Store.

2) [POSTPONED]  Change the behavior of `distinct()` to work against query result. See #4774

3) [DONE] ~~Replace RxJava1 with RxJava2. This issue here #3497 describes an approach where we split RxJava support into a separate library, but after trying that out in a sample project, it just feels really clunky. Doing `realm.where(Person.class).findAll().asFlowable()` is just so much nicer than doing `RxRealm.asFlowable(realm.where(Person.class).findAll())` and if one thing Kotlin has proved, then it is that small things matter~~ 😄 

4) [POSTPONED] Remove the mips so file since there is NO mips device in the market.

5) [DONE](https://github.com/realm/realm-java/pull/5314) [NH] ~~Rename [getAccessToken](https://github.com/realm/realm-java/blob/691c0f4720156187bb5cde1dfa4cc1272a0aaa11/realm/realm-library/src/objectServer/java/io/realm/SyncUser.java#L569) either rename it to reflect the fact that it's returning a `refresh_token`, or abstract the name to the end user.~~

6) [DONE](https://github.com/realm/realm-java/pull/5314) [NH] ~~Remove deprecated methods `retrieveUser` and `retrieveUserAsync` as well as the deprecated `SyncUser#Callback`~~

cmelchior:
> Do not allow having both type Realm instance and Dynamic Realm instance on the same thread. 

I assume you mean for the same `RealmConfiguration` right? I would probably tend to agree. I'm not sure there is a single use case where having a typed Realm and a Dynamic open at the same time make sense. Although I believe we do it during a migration, that is purely an implementation detail we can probably fix.

beeender:
> Do not allow having both type Realm instance and Dynamic Realm instance on the same thread.
Yes, for the same `RealmConfiguration`! Migration is different where we control the Realm instance internally.

Zhuinden:
[...]

Issue #4291, opened 2017-03-06
Better RxJava support


cmelchior:
Right now the interop between RxJava and Realm is not as good as it could be, which is also the feedback we are constantly getting.

This issue is mostly an attempt at summarizing the situation and making sure we have the information in one place instead of being spread out across multiple issues that do not address the full picture.

# The Current Situation
Currently, we have 3 major issues preventing great compatibility with RxJava:

1. No support for [Schedulers](https://github.com/ReactiveX/RxJava/wiki/Scheduler).
2. Streams work on immutable objects. Realm uses [live objects](https://realm.io/news/live-objects-fine-grained-notifications-realm-update/).
3. Thread confinement makes it hard to move work across threads.

I'll try to describe each case below and possible solutions. 

# Challenges

## 1. Custom Schedulers

Realm's async query methods do not allow the use of custom schedulers.

Take this example.

```java
realm.where(Person.class).findAllAsync().asObservable()
  .subscribeOn(Schedulers.io())
  .subscribe(...);
```

This will not run the query on the `io()` thread as one might expect, but on Realm's worker thread. 
On closer inspection though, this is natural, as `findAllAsync()` is executed outside the 
context of RxJava.

The obvious solution is to make `RealmQuery` observable.

```java
realm.where(Person.class).asObservable()
  .flatMap(new Func1<RealmQuery<Person>, Observable<RealmResults<Person>>>() {
        @Override
        public Observable<RealmResults<Person>> call(RealmQuery<Person> query) {
            return query.findAll().asObservable();
        }
    })
  .subscribeOn(Schedulers.io())
  .subscribe(...);
```

An implementation was started, but never completed here: https://github.com/realm/realm-java/pull/1978

This implementation has a lot of benefts:
- The flatmap function could also do a `copyFromRealm` for those who wants that. And it would happen on a background thread.
- The flatmap operation decide if `findAll/findAllSorted/...` is used. The alternative would be to duplicate all `find*` methods as `find*Observable()`, which feels like overkill, especially considering RxJava2 which have more stream types.
- Full interop with all RxJava Schedulers, also test schedulers.  
- `findAllAsync` and friends would still be available to those not caring about schedulers or uses RxJava.

**Challenges**
[...]

Issue #3497, opened 2016-09-27
RxJava 2 support


tonyshkurenko:
#### Goal

Use RxJava 2.0 with realm.io, since it will be released soon
#### Expected Results

Add support for the RxJava 2 same as it was added for the RxJava 1

Zhuinden:
https://github.com/realm/realm-java/issues/3507

cmelchior:
Hi @tonyshkurenko 
I haven't dug that deep into RxJava 2 yet, but Realm uses very little of the RxJava 1.* API's, so I would actually expect that we would automatically be compatible with RxJava 2.

I haven't tested it yet though. Are you thinking of something in particular with regards to RxJava2 ?

tonyshkurenko:
Hey! Thanks for the answer!

I have a pet project with realm and rxjava1 where I test everything, I wanted to refactor it for the rxjava2, to understand it. Currently, I don't have any info about it

cmelchior:
Ok. Testing it should be easy though. 
Just add `io.reactivex:rxjava:2.0.0-RC3` a dependency to your project and try to:

```
Realm realm = Realm.getDefaultInstance();
realm.where(Foo.class).findAllAsync().asObservable().subscribe(...);
```

cmelchior:
RxJava 2 Observables live under `io.reactivex.Observable` while RxJava 1 Observables live under `rx.Observable`. This is a big problem for us as it not only breaks our `RxObservableFactory` interface, but also means we cannot have one method `asObservable()` on RealmResults/Realm/RealmObject that can support both RxJava 1 and 2.

Looks like we have to rethink how to support RxJava going forward.

cmelchior:
For reference this was the old design document: https://github.com/realm/realm-java/pull/1651

The primary reason for going with the current approach was that at the time RxJava and RxJava2 were expected to be interface compatible, at least for the `Observable` class which is the only public class we expose.

Unfortunately this turned out to not be the case ( http://jakewharton.com/java-interoperability-policy-for-major-version-updates/ ), which while good for app developers, are very breaking for libraries like Realm that expose types from these libraries.

This leaves us with two choices IMO:

1) Keep the current approach but add a new `asObservable2()` or similar (bikeshed the name all you want)
 
```
Example
realm.where(Person.class).findAll().asObservable();
realm.where(Person.class).findAll().asObservable2();
```

[...]

PR #4991, opened 2017-07-19
Feature: RxJava2 Support

cmelchior:
Part of #4785 

This feature branch contains all changes required to convert our current RxJava1 support to RxJava2.

There has been a lot of discussion about how our RxJava support could look like in #3497 and #4291 . Mostly it came down to a choice between splitting into a separate library like `RxRealm` or continue to use our "optional dependency" approach. 

After thinking about this a lot and trying to prototype the various approaches, I came to the conclusion that we should continue to use our "optional dependency" approach.

Reasons:

* Splitting RxJava out is mostly a benefit for our codebase, not so much for users of Realm.
* Having a fluent API is extremely nice: `realm.where(Person.class).findAll().asFlowable()` vs. `RxRealm.asFlowable(realm.where(Person.class).findAll())`.
* With Kotlin we could introduce the fluent API as extension functions, but right now the numbers still only show ~15% of Realm uses having Kotlin in their code base.
* People not using RxJava are not really affected either way. We had a few cases where reflection based libraries broke, like Jackson, but apart from that, we had no complaints about the optional dependency approach.
* RxJava1 support is still possible through a `RxRealm` abstraction
* We don't have to spend time setting up a new repo, CI, documentation, and distribution for a new library.

@realm/java Let me know if you disagree strongly with this.

TODO:
- [x] [In Review] Convert existing interfaces and method to RxJava2. Mostly Observable -> Flowable.
- [x] [In Review] Implement `RealmResults.asChangesetObservable()`, `RealmList.asChangesetObservable()` and `RealmObject.asChangesetObservable()`. Wil require moving `Pair` to public API.
- [ ] [POSTPONED] Implement `Realm.asCompletable()` and `DynamicRealm.asCompletable()` for transactions.
- [x] Consider if `RxObservableFactory` is still the correct name. `Observable` has a specific meaning in the context of RxJava 2. `RxStreamFactory` perhaps? -> Not immediately clear that changing this name is a massive improvement, so no reason to break it for a lot of people.
- [x] Update example app to RxJava2 ( @dalinaum)
- [x] Go through wording and Javadoc to ensure it matches RxJava2


CHANGELOG.md
```diff
@@ -1,3 +1,21 @@
+## 4.0.0 (YYYY-MM-DD)
+
+### Breaking Changes
+
+* Realm has upgraded its RxJava1 support to RxJava2 (#3497)
+  * `Realm.asObservable()` has been renamed to `Realm.asFlowable()`.
+  * `RealmList.asObservable()` has been renamed to `Realm.asFlowable()`.
```
Zhuinden: `Realm.`?

CHANGELOG.md
```diff
@@ -1,3 +1,21 @@
...
+  * `RealmResults.asObservable()` has been renamed to `Realm.asFlowable()`.
```
Zhuinden: `Realm.`?

CHANGELOG.md
```diff
@@ -1,3 +1,21 @@
...
+  * `RealmObject.asObservable()` has been renamed to `Realm.asFlowable()`.
```
Zhuinden: `Realm.`?

CHANGELOG.md
```diff
@@ -1,3 +1,21 @@
...
+
+### Enhancements
+
+  * `RealmResults.asObservable()` has been added that publish a `Pair<RealmResults, OrderedCollectionChangeSet>` (#4277).
+  * `RealmList.asObservable()` has been added that publish a `Pair<RealmList, OrderedCollectionChangeSet>` (#4277).
```
Zhuinden: Is Pair from `android.util` or your own class?

Zhuinden:
I assume `asFlowable` with `LATEST` strategy, right?

CHANGELOG.md
cmelchior: Our own. We cannot depend on `android.util.Pair` on the JVM, sadly :(

cmelchior:
Yes, it would be `Flowable` with `LATEST` ... changeset observers would have to be `Observable()` though as you will need all `OrderedCollectionChangeset` events in order to animate things correctly.

Zhuinden:
@curiousily you can query Realm for changes only on looper threads, which means the UI thread, and other handler threads.

So it's the same as now except with `Flowable` and `Observable` where applicable, with additional suport for `changeSets` as well (fine-grained notifications - `OrderedRealmCollectionChangeListener` and `RealmObjectChangeListener`)

PR #4992, opened 2017-07-19
Convert RxJava1 to RxJava2

### LINKED ISSUE #4095: RealmObservableFactory unsubscription subscriber order and addChangeListener should be inverted
#### Goal

Use `take(1)`

#### Expected Results

Unsubscribe as needed.

#### Actual Results

Not sure, but according to Jake Wharton, I think the code in [RealmObservableFactory](https://github.com/realm/realm-java/blob/master/realm/realm-library/src/main/java/io/realm/rx/RealmObservableFactory.java#L81-L82) might not work as expected with `take(1)` and this change in RxBinding can be seen in https://github.com/JakeWharton/RxBinding/commit/3ccbd006152bfd282e75304836744971521b661c

Realm version(s): 2.3.0

cmelchior:
Fixes #4095
Part of #4991 

Convert the existing API's to RxJava 2 types:

* `asObservable()` -> `asFlowable()` ... Since Realm is a hot stream, dropping notifications and only using the latest is fine. The naming in the public API reflects this.

* RealmQuery is exposed as `Single()`. Note this is still not used in our public API, but the primary use case would move the query to a background scheduler, run the query, and convert all types using `copyFromRealm()`. Single()` fits that nicely. Also, I don't see any use cases where you would be interested in listening to changes to the query construction.


realm/realm-library/src/main/java/io/realm/rx/RealmObservableFactory.java
```diff
@@ -29,331 +35,337 @@
...
+                        if (!emitter.isCancelled()) {
+                            emitter.onNext(realm);
                         }
                     }
                 };
                 observableRealm.addChangeListener(listener);
-                subscriber.add(Subscriptions.create(new Action0() {
+
+                // Cleanup when stream is disposed
+                emitter.setDisposable(Disposables.fromRunnable(new Runnable() {
```
Zhuinden: Well at least if I know right then RxJava2 solves the issue that `addChangeListener` actually had to be called **before** `subscriber.add()`

cmelchior: I'm not entirely sure what you are referring to?

Zhuinden: https://github.com/realm/realm-java/issues/4095

realm/realm-library/src/main/java/io/realm/rx/RealmObservableFactory.java
```diff
@@ -29,331 +35,337 @@
...
     };
-    ThreadLocal<StrongReferenceCounter<RealmModel>> objectRefs = new ThreadLocal<StrongReferenceCounter<RealmModel>>() {
+    private ThreadLocal<StrongReferenceCounter<RealmModel>> objectRefs = new ThreadLocal<StrongReferenceCounter<RealmModel>>() {
         @Override
         protected StrongReferenceCounter<RealmModel> initialValue() {
             return new StrongReferenceCounter<RealmModel>();
         }
     };
 
+    private static final BackpressureStrategy BACK_PRESSURE_STRATEGY = BackpressureStrategy.LATEST;
```
curiousily: Is this the only sensible back pressure strategy that Realm can support?

cmelchior: It is the one that makes the most sense. The UI generally doesn't care _what_ changed, just that something changed, so yes. 

Do you have any concerns about this?

Zhuinden: @curiousily it is the only backpressure strategy that makes sense for RealmResults. It will always show its latest value, why enqueue events of the same thing?

curiousily: Thanks for the clarification! Eagerly waiting to try out Realm with RxJava2. Will write back if some use case comes to mind 👍

realm/realm-library/src/main/java/io/realm/rx/RealmObservableFactory.java
```diff
@@ -29,331 +35,337 @@
...
+                        if (!emitter.isCancelled()) {
+                            emitter.onNext(obj);
                         }
                     }
                 };
                 RealmObject.addChangeListener(object, listener);
-                subscriber.add(Subscriptions.create(new Action0() {
+
+                // Cleanup when stream is disposed
+                emitter.setDisposable(Disposables.fromRunnable(new Runnable() {
```
alexrwegener: Will automatically closing the local realm instance when disposing cause problems for longer running subscriptions on the same thread? Say you have a longer running `Flowable` and a shorter running `Flowable` using the same Realm instance and both are subscribed to. The shorter running `Flowable` is disposed which closes the realm instance. Would the longer running flowable no longer get updates at that point?

Zhuinden: They won't use the same instance because each `Realm.getInstance()` call is reference counted.

Zhuinden:
The only thing that nags me a bit inside is that there is a [rxjava3-preview](https://github.com/akarnokd/RxJava3-preview) which will *eventually* supercede RxJava2, and `Flowable` adheres to the `reactive-streams` specification that would mean you could just use `org.reactivestreams.Publisher` instead of RxJava2 specifically, which would be more future-proof.

realm/realm-library/src/main/java/io/realm/rx/RealmObservableFactory.java
alexrwegener: Missed that before in `RealmCache`. Thanks!

cmelchior:
`Publisher` only has the `subscribe` method which is really limited :( Also we thought the same thing when we used `rx.Observable` as the interface for the original `RxObsevableFactory`, and then it changed.

Personally, I would rather have really great integration with the "most common" Rx library, and then extensions for any other (RxJava1, RxJava3, <Insert>). The `RxRealm` extension approach should work for anything and with Kotlin you even add them as extension methods for the fluent API.

Zhuinden:
I never realized the default `Publisher` and `Subscriber` specification is so *stupid* that you get the `Subscription` not as a return value (like with `subscribeWith()`) but when you override `onSubscribe()` in the `Subscriber`? Yeah, that won't do. Let us be thankful that Flowable hides this.

cmelchior:
Ready for review @realm/java

Zhuinden:
@bmeike `Flowable.LATEST` to the best of my knowledge means that if there was an event emission, then only the latest is kept and sent to the subscribers.

Which makes sense because `RealmResults` when a notification is sent, only the latest is needed (every other event would also be the latest).

Jenkinsfile
```diff
@@ -51,7 +51,7 @@ try {
                 }
               } finally {
                 storeJunitResults 'realm/realm-annotations-processor/build/test-results/test/TEST-*.xml'
-                storeJunitResults 'examples/unitTestExample/build/test-results/**/TEST-*.xml'
+                // storeJunitResults 'examples/unitTestExample/build/test-results/**/TEST-*.xml'
```
zaki50: Needs `FIXME` comment?

realm/realm-library/src/main/java/io/realm/rx/RealmObservableFactory.java
```diff
@@ -29,331 +35,337 @@
...
+                        if (!emitter.isCancelled()) {
+                            emitter.onNext(realm);
                         }
                     }
                 };
                 observableRealm.addChangeListener(listener);
-                subscriber.add(Subscriptions.create(new Action0() {
+
+                // Cleanup when stream is disposed
+                emitter.setDisposable(Disposables.fromRunnable(new Runnable() {
```
Zhuinden: I think they kinda do the same thing and `setDisposable` works just fine.

cmelchior: After reading https://github.com/ReactiveX/RxJava/issues/4812 `setDisposable` seemed like the right choice, but if anyone has any reason for why `setCancellable` is better, I'll be happy to change it.

realm/realm-library/src/androidTest/java/io/realm/RxJavaTests.java
```diff
@@ -557,42 +559,42 @@ public void call(Realm rxRealm) {
...
         });
 
-        subscription.unsubscribe();
+        subscription.dispose();
         assertTrue(dynamicRealm.isClosed());
     }
 
     @Test
     @UiThreadTest
     public void realmResults_closeInDoOnUnsubscribe() {
```
cmelchior: It seems to me that `doOnUnsubscribe` is more accurate, perhaps `doOnDispose`? From what I can tell `unsubscribe` is the general word used for telling a stream that you no longer want it around, you can do that by either calling `dispose`, not `cancel`. Also, subscribing returns a `Disposable`, not a `Cancellable`

realm/realm-library/src/main/java/io/realm/rx/RealmObservableFactory.java
Zhuinden: I'm pretty sure it's `dispose()` for Realm's Observables.

PR #4999, opened 2017-07-20
add Nullable annotation to methods that can return 'null'

zaki50:
This improves type inference in Kotlin.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### Enhancements
 
 * [ObjectServer] Added `SyncSession.uploadAllLocalChanges()`.
+* Added `Nullable` annotation to methods that may return `null` in order to improve Kotlin usability. This also introduced a dependency to `com.google.code.findbugs:jsr305`.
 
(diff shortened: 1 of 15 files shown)


beeender:
Does this work with java 6?

cmelchior:
It is purely used by lint tools, so I don't see any reason why it shouldn't?

zaki50:
@cmelchior Kotlin seems to treat every *parameter* as nullable, but it infers every *return type* as non-null (at least AS uses non-null type when it generates a variable with explicit type for return value ).
If we add `ParametersAreNonnullByDefault` annotation, we at the same time need to add `Nullable` annotation to all nullable parameters. I'd like to give annotations to parameters in another PR.
Is it acceptable?

Zhuinden:
I know my lib got a request to have @NonNull added to parameters consistently, especially on callback parameters (such as Realm realm; for executeTransactionAsync)

zaki50:
I've changed target branch to `master`. And also added a changelog entry and fixed an error in kotlinExample.

PR #5000, opened 2017-07-20
fixes #4975

### LINKED ISSUE #4975: logout and `currentUser()`
Below test case will fail with:
```
java.lang.IllegalStateException: Current user is not valid if more that one valid, logged-in user exists.
```

```java
    @Test
    public void logout_currentUserMoreThanOne(){
        final CountDownLatch latch = new CountDownLatch(1);
        SyncUser user1 = UserFactory.createUniqueUser(Constants.AUTH_URL);
        final RealmConfiguration config1 = configurationFactory.createSyncConfigurationBuilder(user1, Constants.USER_REALM)
                .build();
        SyncUser.currentUser().logout();
        SyncUser user2 = UserFactory.createUniqueUser(Constants.AUTH_URL);
        Realm realm = Realm.getInstance(config1);
        SyncUser user3 = SyncUser.currentUser();
    }
```

nhachicha:


nhachicha:
@realm/java

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@
 * Broken case insensitive query with indexed field (#4788).
 * [ObjectServer] Bug related to the behaviour of `SyncUser#logout` and the use of invalid `SyncUser` with `SyncConfiguration` (#4822).
 * [ObjectServer] Not all error codes from the server were recognized correctly, resulting in UNKNOWN being reported instead.
+* [ObjectServer] Prevent the use of a logged out `SyncUser` to open a Realm (#4975).
```
beeender: So we decided to not allow this?

realm/realm-library/src/objectServer/java/io/realm/SyncUser.java
```diff
@@ -535,7 +535,16 @@ public String toJson() {
...
+        return userToken != null && userToken.expiresMs() > System.currentTimeMillis() && !isLoggedOut();
+    }
+
+    /**
+     * Returns whether the user is logged out from the Realm Object Server or not.
+     *
+     * @return  {@code true} if the user was logged out,  {@code false} otherwise.
+     */
+    public boolean isLoggedOut() {
+        return !syncUser.isLoggedIn() || !SyncManager.getUserStore().isActive(syncUser.getIdentity());
```
beeender: `syncUser.isLoggedIn()` is not necessary here i think? `isActive` will return false when `syncUser.isLoggedIn()` returns false i guess?

realm/realm-library/src/objectServer/java/io/realm/internal/SyncObjectServerFacade.java
```diff
@@ -91,7 +91,7 @@ public void realmClosed(RealmConfiguration configuration) {
             // make sure the user is still valid
             SyncUser user = syncConfig.getUser();
             if (!user.isValid()) {
-                if (user.getAccessToken() == null) {
+                if (user.isLoggedOut()) {
```
beeender: So how can we get to the branch at line 100 after this change?

CHANGELOG.md
cmelchior: I don't believe that was the conclusion, no. Cocoa allows this: https://realmio.slack.com/archives/C033Z2Y8T/p1500308286022326

realm/realm-library/src/objectServer/java/io/realm/internal/SyncObjectServerFacade.java
cmelchior: I think this is the wrong change. Users should be to open a Realm even if not logged in. Otherwise, they might not be able to open the Realm if offline.

CHANGELOG.md
cmelchior: Perhaps `Prevent the use of a `SyncUser` that explicitly logged out to open a Realm`

realm/realm-library/src/objectServer/java/io/realm/SyncUser.java
```diff
@@ -535,7 +535,16 @@ public String toJson() {
      */
     public boolean isValid() {
         Token userToken = getSyncUser().getUserToken();
-        return syncUser.isLoggedIn() && userToken != null && userToken.expiresMs() > System.currentTimeMillis() && SyncManager.getUserStore().isActive(syncUser.getIdentity());
+        return userToken != null && userToken.expiresMs() > System.currentTimeMillis() && !isLoggedOut();
+    }
+
+    /**
+     * Returns whether the user is logged out from the Realm Object Server or not.
```
cmelchior: I would extend this description with:
```
Returns whether the user logged out from the Realm Object Server or not. 
Note, a user can be logged in and still be invalid if his/her access has expired. 
<p>
Use {@link #isValid()`} to check if the user is still valid. 
```

realm/realm-library/src/objectServer/java/io/realm/internal/SyncObjectServerFacade.java
nhachicha: `RealmLog.warn("Can not use the provided configuration to open a Realm, the SyncUser is no longer valid.");` will be triggered if the SyncUser, didn't call explicitly `logout` but became invalid, mostly because of an expired `refresh_token`, for example

realm/realm-library/src/objectServer/java/io/realm/SyncUser.java
nhachicha: True, this will be cleaned in the PR, that will try to remove `ObjectServerUser`

realm/realm-library/src/objectServer/java/io/realm/SyncUser.java
beeender: I feel having both `isValid` and `isLoogedOut` will confuse user a lot. I think `isLoggedOut()` doesn't have to be existing in the public API? For any cases, user should just call `isValid()` to check. Can we make this API package protected?

realm/realm-library/src/objectServer/java/io/realm/SyncManager.java
```diff
@@ -242,7 +242,7 @@ static void setAuthServerImpl(AuthenticationServer authServerImpl) {
     }
 
     // Return the currently configured User store.
-    static UserStore getUserStore() {
+    public static UserStore getUserStore() {
```
bmeike: That makes this part of the public API?  Is that ok?

realm/realm-library/src/objectServer/java/io/realm/internal/SyncObjectServerFacade.java
```diff
@@ -91,13 +91,13 @@ public void realmClosed(RealmConfiguration configuration) {
...
             if (!user.isValid()) {
-                if (user.getAccessToken() == null) {
+                if (!SyncManager.getUserStore().isActive(user.getIdentity())) {
                     throw new IllegalStateException("The SyncUser is already logged out and can not use the provided configuration to open a Realm.");
                 } else {
                     // user was not logged out but the `refresh_token` is not longer valid
                     // the user will still get a stall version of Realm, that will work offline
                     // but not sync.
-                    RealmLog.warn("Can not use the provided configuration to open a Realm, the SyncUser is no longer valid.");
+                    RealmLog.warn("The provided configuration uses an expired SyncUser token, this Realm instance will work offline.");
```
bmeike: Better.  Do we expect users to know what "tokens" are and what to do about them?  Is there a clearer way to say this?

nhachicha: We do expose currently `getAccessToken`  as a public API ... `¯\_(ツ)_/¯`

realm/realm-library/src/objectServer/java/io/realm/SyncManager.java
nhachicha: that was the tradeoff compared to adding a new public API `isLoggedOut` that might be confusing for the user with `isValid`. FWIW we do expose `setUserStore` as a public API as well

PR #5003, opened 2017-07-21
Immutable RealmSchema and RealmObjectSchema

beeender:
Before this change, Realm.getSchema() and DynamicRealm.getSchema() both
return a mutable version RealmSchema object. That would cause issue when
changing the typed Realm's schema then continue to use typed inferface
to access the Realm where the column indices might be changed already
and had not been refreshed before the transaction commited.

After this change:
Realm.getSchema() returns an immutable RealmSchema object. And all
RealmObjectSchema objects retrieved from that will be immutable as
well.

### DIFF
--- a/realm/realm-library/src/androidTest/java/io/realm/RealmObjectSchemaTests.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/RealmObjectSchemaTests.java
@@ -240,6 +240,12 @@ public class RealmObjectSchemaTests {
                 fail();
             } catch (UnsupportedOperationException ignore) {
             }
+
+            try {
+                schema.removeField("test");
+                fail();
+            } catch (UnsupportedOperationException ignore) {
+            }
             return;
         }
         for (FieldType fieldType : FieldType.values()) {
--- a/realm/realm-library/src/androidTest/java/io/realm/RealmObjectSchemaTests.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/RealmObjectSchemaTests.java
@@ -16,8 +16,6 @@
 
 package io.realm;
 
-import android.support.test.runner.AndroidJUnit4;
-
 import org.hamcrest.CoreMatchers;
 import org.junit.After;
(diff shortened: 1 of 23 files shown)


realm/realm-library/src/androidTest/java/io/realm/RealmObjectSchemaTests.java
```diff
@@ -194,6 +224,24 @@ public boolean isNullable() {
 
     @Test
     public void addRemoveField() {
+        if (type == ObjectSchemaType.IMMUTABLE) {
+            try {
+                schema.addField("test", int.class);
```
cmelchior: the test is called `addRemove()` but you only test `add*` for the immutable version. Also add a test for `removeField`?

realm/realm-library/src/androidTest/java/io/realm/RealmSchemaTests.java
```diff
@@ -184,12 +238,21 @@ public void rename_shouldChangeInfoInPKTable() {
 
     @Test
     public void remove() {
-        realmSchema.remove(AllJavaTypes.CLASS_NAME);
-        assertFalse(realmSchema.contains(AllJavaTypes.CLASS_NAME));
+        if (type == SchemaType.MUTABLE) {
```
cmelchior: Nitpick: In most other tests you test for IMMUTABLE first

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -158,6 +160,16 @@ private Realm(RealmCache cache) {
     }
 
     /**
+     * Returns the immutable schema for this Realm.
```
cmelchior: I would probably rephrase this to.
```
Returns the schema for this Realm. The schema is immutable. 
Any attempt to modify it will result in an {@link UnsupportedOperationException}.
<p>
The schema can only be modified using {@link DynamicRealm.getSchema()} or through an migration.
```

PR #5007, opened 2017-07-22
Counters: JSON for MutableRealmIntegers

bmeike:
Another step in the implementation of #4266 :

* Clean up documentation
* Add JSON parsing for MutableRealmIntegers
* Add UTs for new code.


realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmProxyClassGenerator.java
```diff
@@ -1928,50 +1927,55 @@ private void emitCreateOrUpdateUsingJsonObject(JavaWriter writer) throws IOExcep
...
-                    primaryKeyFieldType, primaryKeyFieldName, writer);
+            RealmJsonTypeHelper.emitCreateObjectWithPrimaryKeyValue(
+                    qualifiedClassName, qualifiedGeneratedClassName, primaryKeyFieldType, primaryKeyFieldName, writer);
             writer.endControlFlow();
         }
         //@formatter:on
 
+        writer
+                .emitEmptyLine()
+                .emitStatement("final %1$s objProxy = (%1$s) obj", interfaceName);
```
zaki50: Nice!

realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmJsonTypeHelper.java
```diff
@@ -349,52 +343,52 @@ public void emitStreamTypeConversion(String interfaceName, String setter, String
...
+        // @formatter:on
+
+        // @formatter:off
+        @Override
+        public void emitStreamTypeConversion(String varName, String accessor, String fieldName, String fieldType, JavaWriter writer, boolean isPrimaryKey)
+                throws IOException {
+            writer
+                .emitStatement("Long val = null")
+                .beginControlFlow("if (reader.peek() != JsonToken.NULL)")
+                    .emitStatement("val = reader.nextLong()")
```
zaki50: In most cases, `val` is assigned twice (first time is `null` and the second time is the value from json).
Assigning twice is not only less efficient, but also less readable.
How about moving assigning `null` to `else` clause?

realm/realm-library/src/androidTest/java/io/realm/MutableRealmIntegerTests.java
```diff
@@ -346,8 +355,93 @@ public void copyFromIsNotLive() {
...
+        realm.commitTransaction();
+    }
+
+    private void checkTransactionException(Exception e) {
+        checkException(e, "only be done from inside a transaction");
+    }
+
+    private void checkException(Exception e, String expected) {
+        System.out.println("Expecting: " + expected + ", got: " + e);
+        assertTrue(e.getMessage().contains(expected));
```
zaki50: We don't need to use `System.out.println()` here.
These two lines should be `assertTrue("Expecting: " + expected + ", got: " + e, e.getMessage().contains(expected));`.

realm/realm-library/src/androidTest/java/io/realm/MutableRealmIntegerTests.java
```diff
@@ -169,19 +174,23 @@ public void required() {
                 realm.getSchema().get("MutableRealmIntegerTypes")
                         .isNullable(MutableRealmIntegerTypes.FIELD_NONNULLABLE_MUTABLEREALMINTEGER));
 
+        realm.beginTransaction();
```
zaki50: good catch

realm/realm-library/src/main/java/io/realm/MutableRealmInteger.java
```diff
@@ -22,72 +22,69 @@
...
- * A MutableRealmInteger is a mutable, {@link java.lang.Long}-like numeric quantity.
- * It behaves almost exactly as a reference to a {@link java.lang.Long}. More specifically:
+ * A {@code MutableRealmInteger} is a mutable, {@link Long}-like numeric quantity.
+ * It behaves almost exactly as a reference to a {@link Long}. More specifically:
  * <ul>
- * <li>A MutableRealmInteger may have the value {@code null}.</li>
- * <li>The {@code .equals} operator compares the contained Long values. null-valued MutableRealmInteger are {@code .equals}</li>
- * <li>The {@code .compareTo} operator compares the contained Long values.  It considers {@code null} &lt; any value.</li>
- * <li>The {@code .increment} and {@code .decrement} operators throw {@code IllegalStateException} when applied to a null-valued MutableRealmInteger.</li>
+ * <li>A {@code MutableRealmInteger} may have the value {@code null}.</li>
```
zaki50: Why not `@link`?

bmeike: That's a fair point.  I didn't do it because I am not specifically referring to any particular implementation of the `.equals` method.  I don't think the link is useful.  But I might be wrong.  Will fix.

realm/realm-library/src/main/java/io/realm/MutableRealmInteger.java
```diff
@@ -22,72 +22,69 @@
...
+ * <li>The {@link #equals} operator compares the contained {@link Long} values. {@code null}-valued {@code MutableRealmInteger} are {@code .equals}</li>
```
bmeike: I felt this was clearer.  "are equal" could mean "==" or ".equals".  I mean, specifically, the latter, so I said that.

realm/realm-library/src/main/java/io/realm/MutableRealmInteger.java
```diff
@@ -22,72 +22,69 @@
...
- * executed on two different devices, simultaneously, even if connected by only a slow, unreliable network,
- * will <b>always</b> cause the value of {@code counter} to converge, eventually on the value 2.
+ * For instance, if the value of {@code counter} field for the object representing user "Fred" is currently 0,
+ * then the following code, executed on two different devices, simultaneously, even if connected by only a slow,
+ * unreliable network, will <b>always</b> cause the value of {@code counter} to converge, eventually on the value 2.
  * <pre>
- * {@code
- * MutableRealmInteger counter = realm.where(Users.class)
+ * <code> MutableRealmInteger counter = realm.where(Users.class)
  *     .equalTo("name", Fred)
```
bmeike: Definitely.  Will fix.

realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmJsonTypeHelper.java
bmeike: Huh!  I was thinking that it was much more readable this way!
The tradeoff of "efficiency" against code size, seems to be to be pretty irrelevant.  Readability is the thing.  If you can get a second on the readability, I will, totally, change it.

realm/realm-library/src/androidTest/java/io/realm/MutableRealmIntegerTests.java
bmeike: Totally.  This is leftovers from testing the tests.  Will remove.

bmeike:
All comments addressed.

PR #5019, opened 2017-07-25
Fix class name length check

zaki50:
`RealmSchema.create(String)` and `RealmObjectSchema.setClassName(String)` did not accept class name whose length was 51 to ~56~ 57.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ### Bug Fixes
 
 * Potential crash after using `Realm.getSchema()` to change the schema of a typed Realm. `Realm.getSchema()` now returns an immutable `RealmSchema` instance.
-* `RealmSchema.create(String)` and `RealmObjectSchema.setClassName(String)` did not accept class name whose length was 51 to 56.
+* `RealmSchema.create(String)` and `RealmObjectSchema.setClassName(String)` did not accept class name whose length was 51 to 57.
 
 ### Internal
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### Bug Fixes
 
 * Potential crash after using `Realm.getSchema()` to change the schema of a typed Realm. `Realm.getSchema()` now returns an immutable `RealmSchema` instance.
+* `RealmSchema.create(String)` and `RealmObjectSchema.setClassName(String)` did not accept class name whose length was 51 to 56.
 
 ### Internal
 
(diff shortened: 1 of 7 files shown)


realm/realm-library/src/main/java/io/realm/MutableRealmSchema.java
```diff
@@ -45,11 +45,11 @@ public RealmObjectSchema create(String className) {
         checkNotEmpty(className, EMPTY_STRING_MSG);
 
         String internalTableName = Table.getTableNameForClass(className);
-        if (internalTableName.length() > Table.TABLE_MAX_LENGTH) {
+        if (className.length() > Table.CLASS_NAME_MAX_LENGTH) {
```
beeender: Isn't it the same as before?

zaki50: No. `internalTableName` has prefix `class_` but `className` is not.

realm/realm-library/src/main/java/io/realm/internal/Table.java
```diff
@@ -30,13 +30,14 @@
  */
 public class Table implements TableSchema, NativeObject {
 
-    public static final int TABLE_MAX_LENGTH = 56; // Max length of class names without prefix
+    private static final String TABLE_PREFIX = Util.getTablePrefix();
+    public static final int TABLE_NAME_MAX_LENGTH = 62; // Max length of table names
```
nhachicha: isn't `63` in [here](https://github.com/realm/realm-core/blob/655595ac410fd98318c9e84ad5b7a75ac88232ab/src/realm/group.hpp#L328)?

beeender: https://github.com/realm/realm-core/blob/655595ac410fd98318c9e84ad5b7a75ac88232ab/src/realm/group.hpp#L328 should be `63` i think?
`StringData::size()` doesn't count the `\0`. 

https://realm.io/docs/objc/latest/#current-limitations
> Class names are limited to a maximum of 57 UTF8 characters.

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ### Bug Fixes
 
 * Potential crash after using `Realm.getSchema()` to change the schema of a typed Realm. `Realm.getSchema()` now returns an immutable `RealmSchema` instance.
+* `RealmSchema.create(String)` and `RealmObjectSchema.setClassName(String)` did not accept class name whose length was 51 to 56.
```
beeender: should be `57`

realm/realm-library/src/main/java/io/realm/internal/Table.java
zaki50: Thanks. I was thinking the same thing.
In current commits, I kept original value since I wasn't sure if it's just a mistake.

I'll update this PR to use correct value.

zaki50:
@beeender Sure!

PR #5023, opened 2017-07-26
copyToRealmOrUpdate cause dup values in list field

### LINKED ISSUE #4957: Duplicate PrimaryKeys with CopyToRealmOrUpdate - how is this possible?
Hi,
I've got some nested RealmObjects which I try to insert into Realm via copytoRealmOrUpdate. This is my object

```
public class Site extends RealmObject {

    @Required
    @PrimaryKey
    private String SiteUUID;

    private RealmList<AttributeObject> mAttributeObjects;
    private RealmList<Campaign> Campaigns;

    public Site() {}

    //getter und setter
}

public class Campaign extends RealmObject {

    @Required
    @PrimaryKey
    private String CampaignUUID;

    @Expose
    @SerializedName("name")
    private String Name;

    public Campaign() {}
}

public class AttributeObject extends RealmObject {

    @PrimaryKey
    @Required
    private String AttributeObjectUUID;

    private Site site;

    public AttributeObject() {}
}
```

The tricky part for this object relies in this other class attributes which could contain a copy of the site object themselfs. I try to clarify this with the JSON which gets parsed into the Site object.

```
{
    "siteUUID": "4d310340-1dd1-436b-ab05-723eaea9730f",
    "attributes": [
        {
            "attributeUUID": "270568ce-f31a-4343-9a53-ba32f3a0ded8",
            "site": {
                "siteUUID": "4d310340-1dd1-436b-ab05-723eaea9730f",
                "campaigns": [
                    {
                        "campaignUUID": "2c2890d0-b15b-497b-946d-affce01c0bfa",
                        "name": "Campaign1"
                    }
                ]
            }
        }
    ],
    "campaigns": [
        {
            "campaignUUID": "2c2890d0-b15b-497b-946d-affce01c0bfa",
            "name": "Campaign1"
        }
    ]
}
```
As we can see the root Site object got a campaign and some other objects called "attributes". However these objects need to have a reference to its parent site and this site also contains the Campaign object. So we have two Sites with the same PrimaryKey and two Campaign objects with the same PrimaryKey. 

After I've parsed the object from JSON to RealmObject I tried to store the object in the database with my appendToRealm method

```
public static RealmObject appendToRealm(final RealmObject object) {
       try (Realm realm = Realm.getDefaultInstance()) {

           final RealmObject[] managedObject = new RealmObject[1];
               @Override
                    managedObject[0] = realm.copyToRealmOrUpdate(object);
               }
           });

           return managedObject[0];
       } catch (Exception e) {
           e.printStackTrace();
           HockeyCrashHelper.getInstance().sendCrash(e);
       }

       return null;
}
```
If my understanding is correct I expect Realm to loop the object tree recursively and inserts or updates all objects to the database. And because both duplicate objects contain the same primary key I expected Realm to only contain one Site object and only one Campaign object in the database after the appendToRealm method. But the Campaign object was inserted twice. If I see the logs in my app and inspect the database in RealmBrowser they show two instances of the Campaign object with the same uuid as primary key. The Site object is present only once, although it has been twice in the JSON, too. How is this even possible? For all other objects copyToRealmOrUpdate works as expected. Maybe you can clarify this problem?

Thanks in advance :)

Realm version(s): 3.3.2
Android Studio version: 2.3.3
Which Android version and device: Sony Xperia Z5, Android 7.0

beeender:
Close #4957
We are using HashMap to maintain a cache of added managed RealmObject
when doing copyOrUpdate. But if the given object graph contains objects
which have same PK but are different java object, cache missing will
happen. That caused the object got set multiple times. For non-list
fields, it is not a problem since it will eventually has the expected
value if user's objects graph is consistent. For list, that results the
elements in the list are added multiple times since we don't clear the
RealmList in copy().

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### Bug Fixes
 
 * Potential crash after using `Realm.getSchema()` to change the schema of a typed Realm. `Realm.getSchema()` now returns an immutable `RealmSchema` instance.
+* `Realm.copyToRealmOrUpdate()` might result a `RealmList` field contains duplicated elements (#4957).
 
 ### Internal
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ### Bug Fixes
 
 * Potential crash after using `Realm.getSchema()` to change the schema of a typed Realm. `Realm.getSchema()` now returns an immutable `RealmSchema` instance.
-* `Realm.copyToRealmOrUpdate()` might result a `RealmList` field contains duplicated elements (#4957).
+* `Realm.copyToRealmOrUpdate()` might cause a `RealmList` field to contain duplicated elements (#4957).
 
 ### Internal
 
--- a/realm/realm-annotations-processor/src/test/resources/io/realm/AllTypesRealmProxy.java
+++ b/realm/realm-annotations-processor/src/test/resources/io/realm/AllTypesRealmProxy.java
@@ -820,6 +820,7 @@ public class AllTypesRealmProxy extends some.test.AllTypes
         RealmList<some.test.AllTypes> columnRealmListList = realmObjectSource.realmGet$columnRealmList();
         if (columnRealmListList != null) {
             RealmList<some.test.AllTypes> columnRealmListRealmList = realmObjectCopy.realmGet$columnRealmList();
+            columnRealmListRealmList.clear();
             for (int i = 0; i < columnRealmListList.size(); i++) {
                 some.test.AllTypes columnRealmListItem = columnRealmListList.get(i);
                 some.test.AllTypes cachecolumnRealmList = (some.test.AllTypes) cache.get(columnRealmListItem);
--- a/realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmProxyClassGenerator.java
+++ b/realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmProxyClassGenerator.java
@@ -1580,6 +1580,8 @@ public class RealmProxyClassGenerator {
                         .beginControlFlow("if (%sList != null)", fieldName)
                             .emitStatement("RealmList<%s> %sRealmList = realmObjectCopy.%s()",
                                 genericType, fieldName, getter)
+                             // Clear is needed. See bug https://github.com/realm/realm-java/issues/4957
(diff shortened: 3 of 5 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ### Bug Fixes
 
 * Potential crash after using `Realm.getSchema()` to change the schema of a typed Realm. `Realm.getSchema()` now returns an immutable `RealmSchema` instance.
+* `Realm.copyToRealmOrUpdate()` might result a `RealmList` field contains duplicated elements (#4957).
```
Zhuinden: *might cause a `RealmList` to contain

Zhuinden:
Does this also happen with `insertOrUpdate()`?

beeender:
@Zhuinden nope, https://github.com/realm/realm-java/pull/5023/files#diff-eb98b3546f9b692ff01aaff903a3bdb3R1955

realm/realm-annotations-processor/src/test/resources/io/realm/AllTypesRealmProxy.java
```diff
@@ -820,6 +820,7 @@ public static String getTableName() {
         RealmList<some.test.AllTypes> columnRealmListList = realmObjectSource.realmGet$columnRealmList();
         if (columnRealmListList != null) {
             RealmList<some.test.AllTypes> columnRealmListRealmList = realmObjectCopy.realmGet$columnRealmList();
+            columnRealmListRealmList.clear();
```
nhachicha: I don't understand why this is necessary? isn't the list already empty? since just created from 
```
some.test.AllTypes realmObject = realm.createObjectInternal(some.test.AllTypes.class, ((AllTypesRealmProxyInterface) newObject).realmGet$columnString(), false, Collections.<String>emptyList());
...
AllTypesRealmProxyInterface realmObjectCopy = (AllTypesRealmProxyInterface) realmObject;
```

https://github.com/realm/realm-java/pull/5023/files#diff-9bbf8752325de6eed4a407ae41857948R795

beeender: The list will be created before when create `Object4957.child`. https://github.com/realm/realm-java/pull/5023/files/caac7175c9074e42aa23d657be8e193dbf576d04#diff-693596b26c39426696c02bf1cbf4a647R28 It is quite a minor case.

PR #5037, opened 2017-07-28
Fix a bug that RealmSchema.remove() and RealmSchema.rename() don't update internal cache

zaki50:


realm/realm-library/src/main/java/io/realm/RealmSchema.java
```diff
@@ -49,7 +49,7 @@
     // Caches Class objects (both model classes and proxy classes) to their Schema object
     private final Map<Class<? extends RealmModel>, RealmObjectSchema> classToSchema = new HashMap<>();
     // Caches Class Strings to their Schema object
-    private final Map<String, RealmObjectSchema> dynamicClassToSchema = new HashMap<>();
+    final Map<String, RealmObjectSchema> dynamicClassToSchema = new HashMap<>();
```
bmeike: If you are going to make this other than private, you need to control access to it.  Please wrap it in an `immutableHashMap`.

zaki50: But it's not immutable and actually modified from subclasses.

bmeike: I suggest, then, that you add `protected` methods to the superclass, that control the kinds of modifications that the subclass can perform.  A good rule is "`private` or `immutable`".  That is, even, in our lint config.

zaki50: @bmeike I've hidden the `dynamicClassToSchema` map again!

PR #5043, opened 2017-07-29
Add tests that check removed or renamed classes are removed from internal cache

zaki50:
This PR required for #5037 to be merged first.

There seems to be a bug in core.


realm/realm-library/src/androidTest/java/io/realm/RealmSchemaTests.java
```diff
@@ -322,4 +324,43 @@ public void remove_shouldRemoveInfoFromPKTable() {
...
+
+    @Test
+    public void rename_then_cancel() {
+        if (type == SchemaType.IMMUTABLE) {
+            return;
+        }
+
+        realmSchema.create("foo");
+        // getSchemaForClass is an internal method, but used from DynamicRealmObject and RealmQuery
+        final RealmObjectSchema foo = realmSchema.getSchemaForClass("foo");
```
cmelchior: Wouldn't using the public `realmSchema.get()` be the same?

realm/realm-library/src/androidTest/java/io/realm/RealmSchemaTests.java
```diff
@@ -322,4 +324,43 @@ public void remove_shouldRemoveInfoFromPKTable() {
...
+
+    @Test
+    public void remove_then_cancel() {
+        if (type == SchemaType.IMMUTABLE) {
+            return;
+        }
+
+        realmSchema.create("foo");
+        // getSchemaForClass is an internal method, but used from DynamicRealmObject and RealmQuery
+        realmSchema.getSchemaForClass("foo"); // make cache entry
```
cmelchior: Wouldn't using the public realmSchema.get() be the same?

zaki50: No. `realmSchema.get()` is not using the cache. `getSchemaForClass(Class)` also has the same issue, but that method is only used for typed models and no classes are removed actually.

realm/realm-library/src/androidTest/java/io/realm/RealmSchemaTests.java
zaki50: Same

PR #5047, opened 2017-07-30
Add SyncUser.allSessions()

### LINKED ISSUE #4783: Add SyncUser.allSessions()
Swift docs mention that they have a `SyncUser.allSession()` method. 
We should add a similar method.

They also have a `SyncUser.getSessionForUrl()` method, but that seems covered by our `SyncManager.getSession(configuration)`.

dalinaum:
Fixes #4783

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,8 @@
 * [ObjectServer] Added `SyncSession.uploadAllLocalChanges()`.
 * [ObjectServer] APIs of `UserStore` have been changed to support same user identity but different authentication server scenario.
 * Added `Nullable` annotation to methods that may return `null` in order to improve Kotlin usability. This also introduced a dependency to `com.google.code.findbugs:jsr305`.
-* Added support for new data type `MutableRealmIntegers`. The new type behaves almost exactly as a reference to a Long (mutable nullable, etc) but supports `increment` and `decrement` methods, which implement a Conflict Free Replicated Data Type, whose value will converge even when changed across distributed devices with poor connections. (#4266).
+* Added support for new data type `MutableRealmIntegers`. The new type behaves almost exactly as a reference to a Long (mutable nullable, etc) but supports `increment` and `decrement` methods, which implement a Conflict Free Replicated Data Type, whose value will converge even when changed across distributed devices with poor connections (#4266).
+* Added `SyncUser.allSessions` to retrive the all valid sessions belonging to the user (#4783).
 
 ### Bug Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,9 +4,9 @@
 
 * [ObjectServer] Added `SyncSession.uploadAllLocalChanges()`.
 * [ObjectServer] APIs of `UserStore` have been changed to support same user identity but different authentication server scenario.
+* [ObjectServer] Added `SyncUser.allSessions` to retrive the all valid sessions belonging to the user (#4783).
 * Added `Nullable` annotation to methods that may return `null` in order to improve Kotlin usability. This also introduced a dependency to `com.google.code.findbugs:jsr305`.
 * Added support for new data type `MutableRealmIntegers`. The new type behaves almost exactly as a reference to a Long (mutable nullable, etc) but supports `increment` and `decrement` methods, which implement a Conflict Free Replicated Data Type, whose value will converge even when changed across distributed devices with poor connections (#4266).
-* Added `SyncUser.allSessions` to retrive the all valid sessions belonging to the user (#4783).
 
 ### Bug Fixes
 
--- a/realm/realm-library/src/androidTestObjectServer/java/io/realm/SyncUserTests.java
+++ b/realm/realm-library/src/androidTestObjectServer/java/io/realm/SyncUserTests.java
@@ -369,9 +369,8 @@ public class SyncUserTests {
 
         SyncConfiguration configuration1 = new SyncConfiguration.Builder(user, url1).build();
         Realm realm1 = Realm.getInstance(configuration1);
-        assertEquals(1, user.allSessions().size());
-
         List<SyncSession> allSessions = user.allSessions();
+        assertEquals(1, allSessions.size());
         Iterator<SyncSession> iter = allSessions.iterator();
         SyncSession session = iter.next();
         assertEquals(user, session.getUser());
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -5,7 +5,7 @@
 * [ObjectServer] Added `SyncSession.uploadAllLocalChanges()`.
 * Added `Nullable` annotation to methods that may return `null` in order to improve Kotlin usability. This also introduced a dependency to `com.google.code.findbugs:jsr305`.
 * Added support for new data type `MutableRealmIntegers`. The new type behaves almost exactly as a reference to a Long (mutable nullable, etc) but supports `increment` and `decrement` methods, which implement a Conflict Free Replicated Data Type, whose value will converge even when changed across distributed devices with poor connections. (#4266)
-* Added `SyncUser.allSessions` to retrive the all valid sessions belonging to the user. (#4783)
+* Added `SyncUser.allSessions()` to retrive the all valid sessions belonging to the user. (#4783)
```
Zhuinden: retrieve*

dalinaum:

realm/realm-library/src/androidTestObjectServer/java/io/realm/SyncUserTests.java
```diff
@@ -356,4 +358,51 @@ public void changePassword_noneAdminThrows() {
...
+        assertEquals(user, session.getUser());
+        assertEquals(url1, session.getServerUrl().toString());
+
+        SyncConfiguration configuration2 = new SyncConfiguration.Builder(user, url2).build();
+        Realm realm2 = Realm.getInstance(configuration2);
+        assertEquals(2, user.allSessions().size());
+
+        allSessions = user.allSessions();
+        iter = allSessions.iterator();
+        String indivisualUrl = url2.replace("~", user.getIdentity());
```
nhachicha: typo in `indivisualUrl` -> individualUrl

realm/realm-library/src/androidTestObjectServer/java/io/realm/SyncUserTests.java
```diff
@@ -356,4 +358,51 @@ public void changePassword_noneAdminThrows() {
...
+        while (iter.hasNext()) {
+            session = iter.next();
+            assertEquals(user, session.getUser());
+            if (indivisualUrl.equals(session.getServerUrl().toString())) {
+                foundCount++;
+            }
+        }
+        assertEquals(1, foundCount);
+        realm1.close();
+        assertEquals(1, user.allSessions().size());
```
nhachicha: maybe store `allSessions = user.allSessions();` here ,so you can assert the sie then after access the iterator

realm/realm-library/src/objectServer/java/io/realm/SyncManager.java
```diff
@@ -230,6 +232,25 @@ private static synchronized void removeSession(SyncConfiguration syncConfigurati
         }
     }
 
+    /**
+     * Retruns the all valid sessions belonging to the user.
```
nhachicha: Preferer `active` to `valid` since the ObjectStore have uses these 3 states https://github.com/realm/realm-object-store/blob/f462534d94f09a1c6ccb8e161a4eae61a98ebd17/src/sync/sync_session.hpp#L57-L62

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 * [ObjectServer] Added `SyncSession.uploadAllLocalChanges()`.
 * Added `Nullable` annotation to methods that may return `null` in order to improve Kotlin usability. This also introduced a dependency to `com.google.code.findbugs:jsr305`.
 * Added support for new data type `MutableRealmIntegers`. The new type behaves almost exactly as a reference to a Long (mutable nullable, etc) but supports `increment` and `decrement` methods, which implement a Conflict Free Replicated Data Type, whose value will converge even when changed across distributed devices with poor connections. (#4266)
+* Added `SyncUser.allSessions()` to retrieve the all valid sessions belonging to the user. (#4783)
```
nhachicha: Preferer active to valid since the ObjectStore have uses these 3 states https://github.com/realm/realm-object-store/blob/f462534d94f09a1c6ccb8e161a4eae61a98ebd17/src/sync/sync_session.hpp#L57-L62

realm/realm-library/src/objectServer/java/io/realm/SyncManager.java
```diff
@@ -230,6 +232,25 @@ private static synchronized void removeSession(SyncConfiguration syncConfigurati
...
+     *
+     * @param syncUser the user to use.
+     * @return the all valid sessions belonging to the user.
```
nhachicha: valid -> active

realm/realm-library/src/objectServer/java/io/realm/SyncManager.java
beeender: I think sessions with other state could also be returned? eg.: `inactive`?

dalinaum: Since `SynUser::all_sessions` checks whether each session is in error state, all sessions which are not in error state should be returned. So I think some sessions are not `active`.

dalinaum:
@hachicha Thanks. I should look into `all_sessions()` more. I will update this PR.

realm/realm-library/src/objectServer/java/io/realm/SyncManager.java
beeender: `SynUser::all_sessions` are the sessions maintained by Object Store, but here the `sessions` is maintained by the java layer. So what is our plan? Are we going to remove the `SyncManager.sessions` to use the `SyncUser::all_sessions` instead?

dalinaum: I found that @cmelchior sent a DM to me at July 20th as follows:

> I don’t think using the Object Store `SyncUser::all_sessions()` is worth it. We track all of them in Java already.

How about using just `sessions` in this PR and discuss it later?

PS: I should modify this PR's implementation to exclude error state sessions.

beeender: @realm/java If we are sure that we won't use `ObjectStore`'s `all_sessions()` in the future, i am fine with this implementation. But if we do plan to migrate more things which might have impact on this feature, i suggest we postpone this ... According to my experience, all features we implemented in java first then move to Object Store is quite painful ... It would be a huge waste of time later ... (I am struggling with the last step of integration Object Store schema now ...)

cmelchior: Right now the Object Store controls the session lifecycle, which is actually the biggest pain around sessions. We need to decouple this down the line for other functionality like start/stopping it.

This was the primary reason I said to @dalinaum that just using the Java map was fine. Maintaining two Session objects (in SyncManager and this one) will surely also cause a lot of problems. Ideally, we would only have one Java object representing the session.

realm/realm-library/src/objectServer/java/io/realm/SyncManager.java
```diff
@@ -230,6 +232,25 @@ private static synchronized void removeSession(SyncConfiguration syncConfigurati
...
+     */
+    static synchronized List<SyncSession> getAllSessions(SyncUser syncUser) {
```
bmeike: I think you get rid of the synchronization on this method.
`sessions` is already concurrent.
... and synchronizing this method does *nothing* if `getState` or `getUser` returns mutable state...

CHANGELOG.md
dalinaum: @nhachicha @cmelchior Since ObjectStore's implementation returns sessions which are not in error, I think `valid` is more suitable for this PR.

bmeike:
Please verify that `getState` and `getUser` are returning thread safe things...

dalinaum:
@bmeike Since `getState` uses `m_session_mutex` and `getUser` retrieve a value from an immutable config, it seems safe.

PR #5058, opened 2017-08-02
Update Object Store

beeender:
Breaking changes from Object Store:

to 50cf5ee583
- To identify a user, Object Store needs both user ID and auth URL. Thus
  API has been changed in UserStore.
- Property::PropertyType has been changed which is not the same with
  field types in the core anymore.

### DIFF
--- a/realm/realm-library/src/main/cpp/io_realm_internal_Property.cpp
+++ b/realm/realm-library/src/main/cpp/io_realm_internal_Property.cpp
@@ -1,8 +1,7 @@
 /*
  * Copyright 2016 Realm Inc.
  *
- * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with the License.
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
  * http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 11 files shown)


beeender:
`getState_inactive` failed. Looks like caused by Object Store changes. Checking now.

beeender:
@realm/java ready for review.

realm/realm-library/src/main/cpp/io_realm_internal_Property.cpp
```diff
@@ -1,8 +1,7 @@
 /*
  * Copyright 2016 Realm Inc.
  *
- * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with the License.
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with the License.
```
nhachicha: removed line break?

realm/realm-library/src/main/java/io/realm/internal/Property.java
```diff
@@ -29,23 +29,92 @@
     public static final boolean REQUIRED = true;
     public static final boolean INDEXED = true;
 
+    @SuppressWarnings("WeakerAccess")
```
nhachicha: The following fields are used internally AFAICT, they don't need to be public to be generated in the header file

beeender: public because of the static assertion check here https://github.com/realm/realm-java/pull/5058/files/857c93b34938a1553fc304e4979fa6662618600d#diff-aafc664a99eb76a4271b81acd4f4a44cR27

PR #5065, opened 2017-08-03
Final step for OS schema integration

beeender:
- Build the column indices cache from OsSchemaInfo.
  Not like before, the column indices are not built when the Realm
  instance is created. Instead, they will only be built when the relevant
  RealmObject needs to be accessed.
- Column indices cache system has been changed. Different Realm instance
  will not share the same cache. Instead, every Realm instance will have
  its own cache. So we don't rely on the schema version any more. The
  cache can handle the situation when the schema versions are the same but
  schemas are not.
- Refresh column indices cache when schema changes. This is also
  supported in multi-processes environment.
- Almost all of the Realm initialization work is handled by Object
  Store. So we have the same routine for both sync/non-sync Realm. Also
  the initialization routine has be simplified a lot in java side.
- A new class OsRealmConfig is introduced to solve the 20+ arguments
  need to be passed to JNI to create Realm config.
- Due to the Object Store's behaviour, read-only sync Realm is not
  supported anymore.


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 ## 3.6.0 (YYYY-MM-DD)
 
+### Breaking Changes
+
+* [ObjectServer] Read-only is not supported by synced Realm. "SyncConfigration.readOnly()" will always throw `UnsupportedOperationException`.
```
beeender: Well, this has to be changed here. Is it acceptable for current master? or should those just go to master-4.0?

realm/realm-library/src/main/java/io/realm/internal/ColumnInfo.java
```diff
@@ -25,13 +25,14 @@
...
- * </ul>
+ * The Subclasses of this are a fast cache of column indices, for proxy object.
+ * <p>
  * The fast cache functionality is implemented in the Proxy classes generated by {@code RealmProxyClassGenerator}.
- * Be sure to understand what is going on there, before changing things here.
+ * Every proxy object will hold an reference of its {@code ColumnInfo}. The ref to the {@code ColumnInfo} instance is
+ * maintained by the {@link ColumnIndices}. As long as the reference in proxy object has been set, it should never be
+ * changed. When schema changes, the relevant {@code ColumnInfo} instance's content will be refreshed therefore the
+ * proxy object could have the latest column indices. Be sure to understand what is going on there,
+ * before changing things here.
```
beeender: @zaki50 I cannot figure out why this needs to be immutable before, since the cache system changes, can we just remove the mutable setting?

realm/realm-library/src/main/java/io/realm/internal/OsRealmConfig.java
```diff
@@ -0,0 +1,250 @@
...
+        String syncRealmUrl = (String) syncUserConf[1];
+        String syncRealmAuthUrl = (String) syncUserConf[2];
+        String syncRefreshToken = (String) syncUserConf[3];
+        boolean syncClientValidateSsl = (Boolean.TRUE.equals(syncUserConf[4]));
+        String syncSslTrustCertificatePath = (String) syncUserConf[5];
+        if (syncRealmUrl != null) {
+            nativeCreateAndSetSyncConfig(nativePtr, syncRealmUrl, syncRealmAuthUrl, syncUserIdentifier,
+                    syncRefreshToken);
+            // FIXME: shouldn't we throw earlier for this case?
+            if (syncClientValidateSsl && syncSslTrustCertificatePath != null) {
```
beeender: @nhachicha Is this correct?

beeender:
Generally this should be faster to open Realm instance. I will do some benchmark later to show it.

realm/realm-library/src/main/java/io/realm/internal/OsRealmConfig.java
nhachicha: Yes. btw we do log a warning when the user disable the SSL validation and provide a `rootCA` path https://github.com/realm/realm-java/blob/930e1d00cac47149733d17df8688d49fb983d5ac/realm/realm-library/src/objectServer/java/io/realm/SyncConfiguration.java#L828-L836

realm/realm-library/src/main/cpp/io_realm_internal_OsRealmConfig.cpp
```diff
@@ -0,0 +1,332 @@
...
+    REALM_ASSERT((sync_client_validate_ssl && j_sync_ssl_trust_certificate_path) ||
+                 (!sync_client_validate_ssl && !j_sync_ssl_trust_certificate_path));
+
+    try {
+        config.sync_config->client_validate_ssl = sync_client_validate_ssl;
+        if (j_sync_ssl_trust_certificate_path) {
+            JStringAccessor cert_path(env, j_sync_ssl_trust_certificate_path);
+            config.sync_config->ssl_trust_certificate_path = realm::util::Optional<std::string>(cert_path);
+        }
+        else {
```
nhachicha: no need for this block, since the ctor provide this default anyway (for `ssl_trust_certificate_path`)

realm/realm-library/src/main/cpp/io_realm_internal_OsRealmConfig.cpp
```diff
@@ -0,0 +1,332 @@
...
+            user = SyncManager::shared().get_user(sync_user_identifier, refresh_token);
+        }
+
+        util::Optional<std::array<char, 64>> sync_encryption_key(util::none);
+        if (!config.encryption_key.empty()) {
+            sync_encryption_key = std::array<char, 64>();
+            std::copy_n(config.encryption_key.begin(), 64, sync_encryption_key->begin());
+        }
+
+        // Disable ssl by default
```
nhachicha: why? SSL should be on by default. Currently, if the user wants to use SSL [s]he needs to provide the trusted `rootCA`, this will change in the future.

If you want to defer the SSL configuration for later (via `nativeSetSyncConfigSslSettings`), then these value `sync_client_validate_ssl` and `ssl_trust_certificate_path` have valid defaults in the ctor, you can skip them in `std::make_shared<SyncConfig>` call

realm/realm-library/src/main/cpp/io_realm_internal_OsRealmConfig.cpp
```diff
@@ -0,0 +1,332 @@
...
+
+JNIEXPORT void JNICALL Java_io_realm_internal_OsRealmConfig_nativeSetEncryptionKey(JNIEnv* env, jclass,
+                                                                                   jlong native_ptr,
+                                                                                   jbyteArray j_key_array)
+{
+    TR_ENTER_PTR(native_ptr)
+    try {
+        JniByteArray key_array(env, j_key_array);
+        auto& config = *reinterpret_cast<Realm::Config*>(native_ptr);
+        // Encryption key should be set before creating sync_config.
```
nhachicha: Why?

realm/realm-annotations-processor/src/test/resources/io/realm/AllTypesRealmProxy.java
```diff
@@ -436,111 +438,8 @@ public static OsObjectSchemaInfo getExpectedObjectSchemaInfo() {
...
-        ProxyUtils.verifyField(sharedRealm, columnTypes, "columnRealmList", RealmFieldType.LIST, "AllTypes");
-        if (!sharedRealm.hasTable("class_AllTypes")) {
-            throw new RealmMigrationNeededException(sharedRealm.getPath(), "Missing class 'class_AllTypes' for field 'columnRealmList'");
-        }
-        Table table_9 = sharedRealm.getTable("class_AllTypes");
-        if (!table.getLinkTarget(columnInfo.columnRealmListIndex).hasSameSchema(table_9)) {
-            throw new RealmMigrationNeededException(sharedRealm.getPath(), "Invalid RealmList type for field 'columnRealmList': '" + table.getLinkTarget(columnInfo.columnRealmListIndex).getName() + "' expected - was '" + table_9.getName() + "'");
-        }
-
-        long backlinkFieldIndex;
```
bmeike: where do these checks happen now?

realm/realm-library/src/main/java/io/realm/internal/ColumnIndices.java
```diff
@@ -42,147 +47,91 @@
...
 
-    private ColumnIndices(long schemaVersion, Map<Pair<Class<? extends RealmModel>, String>, ColumnInfo> classesMap, boolean mutable) {
-        this.schemaVersion = schemaVersion;
-        this.classesToColumnInfo = classesMap;
-        this.mutable = mutable;
-        this.classes = new HashMap<>(classesMap.size());
-        this.classesByName = new HashMap<>(classesMap.size());
-    }
 
     /**
```
bmeike: Boy.   This looks scary, to me.  I put that mutability thing in there, because we most *definitely* passed these things between threads.  If we still do that they *absolutely* must be effectively immutable.
Unless you are quite sure this is never passed between threads, I request that you store the name of the creating thread in the constructor, check it in every mutator, and throw if they don't match.

realm/realm-library/src/main/java/io/realm/internal/ColumnInfo.java
bmeike: Because these things are passed between threads.  Some of them have to be mutable.  Some of them absolutely cannot be.

realm/realm-library/src/main/java/io/realm/internal/ColumnIndices.java
beeender: No, we never pass the column indices across threads anymore. no more maps to hold a column indices cache globally. 

Realm -> RealmSchema -> ColumnIndices -> ColumnInfo
   | create Realm Object                                             ^
   V                                                                             |
RealmObject  -------------------------------------------------------

realm/realm-library/src/main/cpp/io_realm_internal_OsRealmConfig.cpp
beeender: Because of we need to get the key from `Realm::config` and set it to `SyncConfig`. https://github.com/realm/realm-java/pull/5065/files/e9f892215966160978abbe428ad24de5e23ce9b9#diff-99d5c7b90ca048df8fa6511c6daec6c7R291

This is actually a very strange API design in Object Store :(

realm/realm-library/src/main/cpp/io_realm_internal_OsRealmConfig.cpp
beeender: Yes, you are right. I was thinking `sync_client_validate_ssl` is redundant since we have `ssl_trust_certificate_path` to show if we are using ssl or not. But that is wrong, we could support ssl without a customized ssl cert path. Will change this.

realm/realm-library/src/main/cpp/io_realm_internal_OsRealmConfig.cpp
beeender: I am think about should we allow `nativeSetSyncConfigSslSettings` to be called multi-times. This `else` is mostly to make it safe in that case. I can add some assert to prevent it from being called more than once.

CHANGELOG.md
beeender: I think maybe I could work around this to make the behave the same as current `master` if it is _really not acceptable_ for `master`. But this has to get the same behaviour as the Object Store and other bindings at least from `master-4.0`.

cmelchior: Why do this? We have real use cases for it, it and it is easy to keep the checking in the Java layer for now?

beeender: Do other bindings have the same realm use cases? If yes, we should just change the behaviour in Object Store.  relevant discussion on slack https://realmio.slack.com/archives/C0340J37W/p1494485682035749 https://realmio.slack.com/archives/C053279FS/p1494422645218828

realm/realm-library/src/main/java/io/realm/internal/ColumnIndices.java
bmeike: Really?  That is incredibly awesome!!!

CHANGELOG.md
cmelchior: I don't know how Cocoa detect read only files, but they definitely have the same use case where people put Realm files on read-only storage. 

The second conversation you reference describes the motivation for treating missing tables as being present but just empty. I don't think we want these semantics in Java since it would complicate our query class quite a bit, and people can just create an explicit Schema to avoid the issue. Most likely this will require two different READ_ONLY modes in Object Store, which is a bit unfortunate.

realm/realm-library/src/main/java/io/realm/internal/ColumnIndices.java
zaki50: `ColumnInfo`  is not shared between threads, but shared between `RealmObject`s for the same table and `Realm` instance where those `RealmObject`s belongs.
When an index of some field is updated, `Realm` just updates a `ColumnInfo` instance that the `Realm` is referring, and the change is seen from all `RealmObject`s because all the `RealmObject`s are referring the same instance.

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -196,11 +193,6 @@ public void run() {
     // doCreateRealmOrGetFromCache is called with globalCount == 0.
     private RealmConfiguration configuration;
 
-    // Column indices are cached to speed up opening typed Realm. If a Realm instance is created in one thread, creating
-    // Realm instances in other threads doesn't have to initialize the column indices again.
-    private static final int MAX_ENTRIES_IN_TYPED_COLUMN_INDICES_ARRAY = 4;
-    private final ColumnIndices[] typedColumnIndicesArray = new ColumnIndices[MAX_ENTRIES_IN_TYPED_COLUMN_INDICES_ARRAY];
```
cmelchior: Did you measure how long it takes to generate the column indices pr. thread? I would assume it isn't much, but having some idea would be nice. Given how much simpler the code is gets, it is probably fine either way.

realm/realm-library/src/main/java/io/realm/internal/ColumnIndices.java
```diff
@@ -42,147 +47,91 @@
...
-     * Returns the {@link ColumnInfo} for the passed class or ({@code null} if there is no such class).
+     * Returns the {@link ColumnInfo} for the passed class.
      *
      * @param clazz the class for which to get the ColumnInfo.
-     * @return the corresponding {@link ColumnInfo} object, or {@code null} if not found.
+     * @return the corresponding {@link ColumnInfo} object.
+     * @throws io.realm.exceptions.RealmException if the class cannot be found in the schema.
      */
-    @Nullable
+    @Nonnull
```
cmelchior: We need to establish some conventions about these annotations. I see 3 options:

1) Everything should be marked with either @Nonnull or @Nullable
2) Everything is @Nonnull unless marked explicitly with @Nullable
3) Everything is @NUllable unless marked explicitly with @Nonnull

I prefer 2, but picking one and sticking to it is more important.

realm/realm-library/src/main/java/io/realm/internal/OsRealmConfig.java
```diff
@@ -0,0 +1,248 @@
...
+         * @return this {@link OsRealmConfig.Builder}.
+         */
+        public Builder autoUpdateNotification(boolean autoUpdateNotification) {
+            this.autoUpdateNotification = autoUpdateNotification;
+            return this;
+        }
+
+        // Package private because of the OsRealmConfig needs to carry the NativeContext. This should only be called
+        // by the SharedRealm.
+        OsRealmConfig build() {
```
cmelchior: Nitpick. It feels a bit weird that the rest of  the methods are public, but this is not. Given that this class is already in `io.realm.internal` keeping this public would be fine, no?

realm/realm-library/src/objectServer/java/io/realm/SyncConfiguration.java
```diff
@@ -703,18 +702,12 @@ public Builder waitForInitialRemoteData() {
...
         /**
-         * Setting this will cause the Realm to become read only and all write transactions made against this Realm will
-         * fail with an {@link IllegalStateException}.
-         * <p>
-         * This in particular mean that {@link #initialData(Realm.Transaction)} will not work in combination with a
-         * read only Realm and setting this will result in a {@link IllegalStateException} being thrown.
-         * </p>
-         * Marking a Realm as read only only applies to the Realm in this process. Other processes and devices can still
-         * write to the Realm.
+         * @deprecated readOnly is not supported by Synced Realm.
```
cmelchior: See discussion elsewhere, but this should not be deprecated IMO

realm/realm-library/src/syncIntegrationTest/java/io/realm/SyncedRealmTests.java
```diff
@@ -181,73 +181,6 @@ public void onError(Throwable exception) {
     }
 
     @Test
-    public void waitForInitialRemoteData_readOnlyTrue() throws InterruptedException {
```
cmelchior: See above

realm/realm-library/src/main/java/io/realm/internal/OsRealmConfig.java
beeender: See the comments on line 120. `io.realm` package should never call this method.

realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -76,30 +76,53 @@
     private RealmCache realmCache;
     protected SharedRealm sharedRealm;
     private boolean shouldCloseSharedRealm;
+    private SharedRealm.SchemaChangedCallback schemaChangedCallback = new SharedRealm.SchemaChangedCallback() {
+        @Override
+        public void onSchemaChanged() {
+            RealmSchema schema = getSchema();
+            if (schema != null) {
+                schema.refresh();
```
cmelchior: Ups, by mistake. I thought this on on `SharedRealm`. This is fine 👍

realm/realm-library/src/main/cpp/io_realm_internal_OsRealmConfig.cpp
```diff
@@ -0,0 +1,326 @@
...
+                "(JLio/realm/internal/OsRealmConfig;Lio/realm/internal/SharedRealm$InitializationCallback;)V", true);
+            JavaGlobalRef j_init_callback_ref(env, j_init_callback, true);
+            JavaGlobalRef j_config_ref(env, j_config, true);
+            config.initialization_function = [j_init_callback_ref, j_config_ref](SharedRealm realm) {
+                JNIEnv* env = JniUtils::get_env(false);
+                if (env->ExceptionCheck()) {
+                    return;
+                }
+                // Java needs a new pointer for the SharedRealm life control.
+                SharedRealm* new_shared_realm_ptr = new SharedRealm(realm);
```
beeender: The new pointer will be constructed to a new `SharedRealm` java object in the following callback, which will be freed when the `SharedRealm` instance gets GCed just like other `SharedRealm` object.

beeender: actually in the first implementation, i didn't alloc a new `ShareRealm` ptr here, but just using the pointer to `&realm`. In that way, we gain a very little performance (no need to new a `SharedRealm`), but the code in Java will be a bit confusing. So, do this way to make it looks like other places how we handle the native pointers.

realm/realm-library/src/main/cpp/io_realm_internal_OsRealmConfig.cpp
```diff
@@ -0,0 +1,326 @@
...
+    }
+    CATCH_STD()
+}
+
+JNIEXPORT void JNICALL Java_io_realm_internal_OsRealmConfig_nativeEnableChangeNotification(
+    JNIEnv*, jclass, jlong native_ptr, jboolean enable_auto_change_notification)
+{
+    TR_ENTER_PTR(native_ptr)
+
+    // No throws
```
beeender: "no throws" for the whole method body.

realm/realm-library/src/main/java/io/realm/BaseRealm.java
beeender: maybe change the name of `RealmSchema.refresh()` to `RealmSchema.invalidate()`? doesn't that sound better?

realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -76,30 +76,53 @@
...
+
+        OsRealmConfig.Builder configBuilder = new OsRealmConfig.Builder(configuration)
+                .autoUpdateNotification(true)
+                .migrationCallback(migrationCallback)
+                .schemaInfo(schemaInfo)
+                .initializationCallback(initializationCallback);
+        this.sharedRealm = SharedRealm.getInstance(configBuilder);
         this.shouldCloseSharedRealm = true;
+
+        sharedRealm.registerSchemaChangedCallback(schemaChangedCallback);
```
beeender: As an internal API, it should never be called twice.

realm/realm-library/src/main/java/io/realm/RealmCache.java
beeender: The benchmark results of insertion actually shows it. Although it is running in single thread mode, the original implementation cached the column indices before insertion, current implementation only caches it when insertion called on the object. The benchmark doesn't show noticeable difference.

realm/realm-library/src/main/java/io/realm/BaseRealm.java
cmelchior: Either is fine with me 👍

realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -690,9 +694,19 @@ public void onMigrationNeeded(SharedRealm sharedRealm, long oldVersion, long new
         }
     }
 
+    private static SharedRealm.MigrationCallback createMigrationCallback(final RealmMigration migration) {
+        return new SharedRealm.MigrationCallback() {
+            @Override
+            public void onMigrationNeeded(SharedRealm sharedRealm, long oldVersion, long newVersion) {
+                migration.migrate(
+                        DynamicRealm.createInstance(sharedRealm), oldVersion, newVersion);
```
beeender: what do you mean?

cmelchior: Couldn't this fit into one line?

realm/realm-library/src/main/java/io/realm/internal/ColumnIndices.java
bmeike: Sadly, that is not true.  The method RealmSchema.setInitialColumnIndices is called from multiple threads, with the same ColumnIndices.

beeender: `setInitialColumnIndices` is removed ;D

realm/realm-library/src/main/java/io/realm/internal/ColumnIndices.java
beeender: Yeah, following https://github.com/realm/realm-java/pull/5044/files#diff-d3925675e94190bc74665a93fda12ab3

realm/realm-library/src/main/cpp/io_realm_internal_OsRealmConfig.cpp
```diff
@@ -0,0 +1,326 @@
...
+        auto& config = *reinterpret_cast<Realm::Config*>(native_ptr);
+        if (j_compact_on_launch) {
+            static JavaClass compact_on_launch_class(env, "io/realm/CompactOnLaunchCallback");
+            static JavaMethod should_compact(env, compact_on_launch_class, "shouldCompact", "(JJ)Z");
+            JavaGlobalRef java_compact_on_launch_ref(env, j_compact_on_launch);
+
+            config.should_compact_on_launch_function = [java_compact_on_launch_ref](uint64_t totalBytes,
+                                                                                    uint64_t usedBytes) {
+                JNIEnv* env = JniUtils::get_env(false);
+                return env->CallBooleanMethod(java_compact_on_launch_ref.get(), should_compact,
```
beeender: Great catch! Fixed in 70f0a8e1cd4ea8dabc3fbe2c507d8b4b9ddd38d0

beeender:
@realm/java All comments are addressed.

bmeike:
Let's get this dude in.  It is gonna play havoc with lots of other code...

realm/realm-library/src/main/java/io/realm/internal/ColumnInfo.java
```diff
@@ -217,39 +222,28 @@ public String toString() {
...
      * <b>For use only by subclasses!</b>.
      * Must be called from within the subclass constructor, to maintain the effectively-final contract.
      *
-     * @param realm The shared realm.
+     * @param schemaInfo the {@link OsSchemaInfo} of the corresponding {@code Realm} instance.
      * @param columnName The name of the backlink column.
      * @param sourceTableName The name of the backlink source class.
      * @param sourceColumnName The name of the backlink source field.
      */
-    @SuppressWarnings("unused")
```
dalinaum: Doesn't need it anymore? It looks like invoking `addBacklinkDetails` are similar.

realm/realm-library/src/main/java/io/realm/internal/ColumnInfo.java
```diff
@@ -217,39 +222,28 @@ public String toString() {
      * <p>
      * No validation done here.  Presuming that all necessary validation takes place in {@code Proxy.validateTable}.
      *
-     * @param table The table to search for the column.
      * @param columnName The name of the column whose index is sought.
-     * @param columnType Type RealmType of the column.
+     * @param objectSchemaInfo the {@link OsObjectSchemaInfo} for the corresponding {@code RealmObject}.
      * @return the index of the column in the table
      */
-    @SuppressWarnings("unused")
```
dalinaum: Isn't it needed anymore? It looks like invoking `addColumnDetails` is similar with the previous code.

realm/realm-library/src/main/java/io/realm/internal/ColumnIndices.java
bmeike: I'm *really* thrilled to hear that this thing doesn't move between threads any more.  There is a bunch of silly code in `ColumnIndices` that is there to assure that the thing is effectively final.  It could go away now. Really really cool.

I wonder, though, if you would consider, at least in the debug version, adding checks to the columnIndices methods, to verify that they are not being called from a second thread.  The check should be incredibly cheap.  This code is somewhat hairy and it would be really great to run with the test in place for a version or so, just to be sure.  Call me nervous Nelly.

realm/realm-library/src/main/java/io/realm/internal/ColumnIndices.java
bmeike: I vote for #2, if it is enforceable...

realm/realm-library/src/main/java/io/realm/internal/ColumnInfo.java
beeender: No idea why it was added at the first place. Maybe it was used in the proxies, so AS will report lint issues before building tests.

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -150,6 +150,23 @@ private Realm(RealmCache cache) {
...
+        // FIXME: This is to work around the different behaviour between the read only Realms in the Object Store and
+        // in current java implementation. Opening a read only Realm with some missing schemas is allowed by Object
+        // Store and realm-cocoa. In that case, any query based on the missing schema should just return an empty
+        // results. Fix this together with https://github.com/realm/realm-java/issues/2953
+        if (configuration.isReadOnly()) {
+            RealmProxyMediator mediator = configuration.getSchemaMediator();
+            Set<Class<? extends RealmModel>> classes = mediator.getModelClasses();
+            for (Class<? extends RealmModel> clazz  : classes) {
+                String tableName = mediator.getTableName(clazz);
+                if (!sharedRealm.hasTable(tableName)) {
```
nhachicha: I presume other schema mismatches (_(adding/removing a column?) _) are being checked in `validateTable`?

cmelchior: I thought all of this validation logic lived in Object Store with the secondary `ReadOnlyAlternative` mode?

beeender: Yes! that is checked by Object Store https://github.com/realm/realm-object-store/pull/511/files#diff-94099fefbd1f17776b45b9f0998e8c61R1360

realm/realm-annotations-processor/src/test/resources/io/realm/AllTypesRealmProxy.java
beeender: @bmeike happened here https://github.com/realm/realm-java/pull/4960

PR #5067, opened 2017-08-03
Fix class name in not found message

bmeike:
The message in the nativeGetTable wants to talk about a `class` but, in stead, uses the name of the `table`.

### DIFF
--- a/realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
+++ b/realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
@@ -435,6 +435,9 @@ JNIEXPORT jlong JNICALL Java_io_realm_internal_SharedRealm_nativeGetTable(JNIEnv
         auto& shared_realm = *(reinterpret_cast<SharedRealm*>(shared_realm_ptr));
         if (!shared_realm->read_group().has_table(name)) {
             std::string name_str = name;
+            if (name_str.find(TABLE_PREFIX) == 0) {
+                name_str = name_str.substr(TABLE_PREFIX.length());
+            }
             THROW_JAVA_EXCEPTION(env, JavaExceptionDef::IllegalArgument,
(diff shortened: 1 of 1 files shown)


bmeike:
There may, easily, be a better way to do this.  Let me know.

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
```diff
@@ -435,6 +435,9 @@ JNIEXPORT jlong JNICALL Java_io_realm_internal_SharedRealm_nativeGetTable(JNIEnv
         auto& shared_realm = *(reinterpret_cast<SharedRealm*>(shared_realm_ptr));
         if (!shared_realm->read_group().has_table(name)) {
             std::string name_str = name;
+            if (name_str.find(TABLE_PREFIX) == 0) {
```
beeender: Java should always pass a `table_name` here which means it should always have `TABLE_PREFIX` in the string. What is the case triggered this issue?

(Due to historical reasons, we sometimes mixed use of table_name and class_name. Ideally the `TABLE_PREFIX` should be hidden from java -- java doesn't have to know that. By integrating more object_store stuff, we are slowly moving to the target.)

beeender: Opps, sorry. this change it correct!

bmeike: I totally think you are absolutely right.  Eventually the whole table prefix should disappear from Java.  If the core/os folks decide to change it to "foobar_", I don't even want to hear about it.  ...until then, though...

bmeike:
Hey, I have a question.  Does the second assignment to `name_str` do an un-necessary copy?  If not, why not?  The first one assignment made a copy...

kneth:
@bmeike It does make a copy: http://www.cplusplus.com/reference/string/string/substr/

PR #5076, opened 2017-08-08
Update gradle wrappers to 4.1

zaki50:
Gradle 4.1 seems to have a bug that `GradleBuild` task sometimes hangs up.
To work around that bug, we need to add `--console=plain` option when you execute gradle wrapper.

`tools/update_gradle_wrapper.sh` modifies the Gradle Wrapper script in the root directory to add that option automatically.


zaki50:
Could you review this ? @realm/java

tools/update_gradle_wrapper.sh
```diff
@@ -10,12 +10,22 @@ HERE=`pwd`
...
     pwd
     ./gradlew wrapper
     sed -E -i '' s/-bin\\.zip\$/-all.zip/ gradle/wrapper/gradle-wrapper.properties
 done
 
+popd
+
+sed -i '' '/^APP_ARGS=/a\
+# Realm'"'"'s work-around for a bug in Gradle 4.1\
+APP_ARGS="\${APP_ARGS} '"'"'--console=plain'"'"' \\\\\
```
beeender: Is the bug reported to upstream somewhere? Does it hang with any error message?

zaki50: Not yet. 
It doesn't print anything...

gradlew
```diff
@@ -160,6 +160,9 @@ save () {
     echo " "
 }
 APP_ARGS=$(save "$@")
+# Realm's work-around for a bug in Gradle 4.1
```
cmelchior: Would be nice to have a reference to a bug report somewhere, so we can know when to remove this again

tools/update_gradle_wrapper.sh
```diff
@@ -10,12 +10,22 @@ HERE=`pwd`
...
     cd $(dirname $i)
     pwd
     ./gradlew wrapper
     sed -E -i '' s/-bin\\.zip\$/-all.zip/ gradle/wrapper/gradle-wrapper.properties
 done
 
+popd
+
+sed -i '' '/^APP_ARGS=/a\
+# Realm'"'"'s work-around for a bug in Gradle 4.1\
```
cmelchior: `Realm'"'"'s` looks like copy paste error

gradlew
cmelchior: Also, we have more `gradlew` files than this, e.g. in `examples`, `realm-plugin`, `realm` and maybe more. Shouldn't this fix be in all of them?

tools/update_gradle_wrapper.sh
zaki50: No. we can't use ' in single quote string. `'"'"'` closes single quote string then open double quote string that contains one single quote in it the reopen single quote string again.
These three strings are concatenated.

Is there any efficient way to contain single quote in single quote string?

cmelchior: Not that I know of. So 👍

tools/update_gradle_wrapper.sh
zaki50: https://github.com/gradle/gradle/issues/2673

gradlew
zaki50: Added a link to bug report.

Currently, we are using `GradleBuild` task only in root project.
So modifying just `gradlew` in root directory is enough.

Do you prefer fixing all `gradlew` files?

cmelchior: No, just doing it here is fine.

Issue #4366, opened 2017-03-23
Old/New values paris for Single Object Notifications


beeender:
This is for following enhancement of #4101 which are not implemented in #4331 

Cocoa API will notify user when the object changes with old/new value pair:
```objc
@interface RLMPropertyChange : NSObject

/**
 The name of the property which changed.
 */
@property (nonatomic, readonly, strong) NSString *name;

/**
 The value of the property before the change occurred. This will always be `nil`
 if the change happened on the same thread as the notification and for `RLMArray`
 properties.

 For object properties this will give the object which was previously linked to,
 but that object will have its new values and not the values it had before the
 changes. This means that `previousValue` may be a deleted object, and you will
 need to check `invalidated` before accessing any of its properties.
 */
@property (nonatomic, readonly, strong, nullable) id previousValue;

/**
 The value of the property after the change occurred. This will always be `nil`
 for `RLMArray` properties.
 */
@property (nonatomic, readonly, strong, nullable) id value;
@end

```

By extending current java interface `ObjectChangeSet`, realm-java could have something similar.

The proposed API 1:

```java
public interface ObjectChangeset {
  boolean isDeleted();
  String[] getChangedField();
  isFieldChanged(String fieldName);
 
  public static class FieldChange<T> {
    T oldValue;
    T newValue;
  }

  // This will always return null for RealmList field.
  <E> FieldChange<E> getChange(E type, String fieldName);
}
```

The proposed API 2:

```java
public interface ObjectChangeset {
  boolean isDeleted();
  String[] getChangedField();
  isFieldChanged(String fieldName);
 
  Long getOldLongValue(String fieldName);
  Long getNewLongValue(String fieldName);
  Date getOldDateValue(String fieldName);
  Date getNewDateValue(String fieldName);
  // ... Other getters
}
```

Potential problems & limitations:

- [ ] Storing old values requires memory and potentially both memory & cpu consuming operation since there might be big strings needed to be converted from `StringData` to Java `String`. We may want to give user an option to disable the old/new values in the notification. Something like `RealmObject.addChangeListener(RealmObjectChangeListener listener, boolean needValues)`.
- [ ] Same like cocoa,  the value of the property before the change occurred. This will always be `nil`
 if the change happened on the same thread as the notification. 
- [ ] All primitive types will be returned as boxed type. This is because of we have to use generic for `FieldChange<T>`.
[...]

PR #5081, opened 2017-08-09
Move Pair into public API

cmelchior:
Part of #3497 

This PR moves `Pair` into the public API. This is to prepare for changeset observables that need to emit items that look like this: `Pair<RealmResults<E>, OrderedCollectionChangeset>`:

```
realm.where(Person.class).findAll().asChangesetObservable()
   .subscribe(new Consumer<Pair<RealmResults<E>, OrderedCollectionChangeset>>() {
       @Override
       public void accept(Pair<RealmResults<E>, OrderedCollectionChangeset> change) {
          change.first; // RealmResults
          change.second; // Changeset
       }
   });
```

An alternative could also be to create our own custom tuple class for this, but I'm not sure it is worth the effort?


beeender:
How to deal with the new/old value pairs in the future? https://github.com/realm/realm-java/issues/4366

We need to pass `RealmModel<E>`, `String[]` for changed fields, `Object[]` for both old values and new values.

cmelchior:
I assume that is going to be handled by the `OrderedCollectionChangeset` class somehow? `Pair` is purely used to support RxJava2 that can only emit 1 item instead of `(result,change)`

beeender:
Uh, right. by `ObjectChangeset`

realm/realm-library/src/main/java/io/realm/util/Pair.java
```diff
@@ -14,19 +14,22 @@
  * limitations under the License.
  */
 
-package io.realm.internal.util;
+package io.realm.util;
```
beeender: This package doesn't generate public javadoc. Before doing it, some existing classes need to be moved to the `io.realm.internal.util` since they are not necessary for public at all. But that could be another PR.

cmelchior: Not sure I understand why we need to move other existing classes? Which ones?
Also, the Javadoc should be generated for this package, we just exclude `internal`:
```
    exclude '**/internal/**'
    exclude '**/BuildConfig.java'
    exclude '**/R.java'
```

beeender: hmm? But i cannot see `io.realm.util` package here https://realm.io/docs/java/3.5.0/api/

UH... my fault, there is a `io.realm.util` in the `androidTest`. This is OK then 👍

PR #5082, opened 2017-08-09
Public Javadoc for PermissionManager

cmelchior:
Part of #4558 

This PR adds the remaining public Javadoc.

### DIFF
--- a/realm/realm-library/src/objectServer/java/io/realm/PermissionManager.java
+++ b/realm/realm-library/src/objectServer/java/io/realm/PermissionManager.java
@@ -43,10 +43,21 @@ import io.realm.permissions.PermissionRequest;
 
 
 /**
- * FIXME: Better Javadoc
- * Helper class for interacting with Realm permissions.
- * This class has connections to underlying Realms, so all data coming from this class is thread-confined and must be
+ * Helper class for interacting with Realm Object Server permissions for a {@link SyncUser}.
+ * <p>
+ * Current functionality supported by this class:
+ * <ul>
+ *     <li>List users existing permissions.</li>
+ *     <li>List default permissions.</li>
+ *     <li>Modify permissions for a Realm.</li>
+ *     <li>Create a permission offer that can be sent to others.</li>
+ *     <li>Accept permission offers sent by other users.</li>
+ * </ul>
+ * </p>
+ * This class is depending on underlying Realms, so all data coming from this class is thread-confined and must be
  * closed after use to avoid leaking resources.
+ *
+ * @see <a href="https://realm.io/docs/java/latest/#access-control">How to work with Access Controls</a>
  */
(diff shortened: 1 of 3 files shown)


realm/realm-library/src/objectServer/java/io/realm/PermissionManager.java
```diff
@@ -210,10 +222,11 @@ public void onError(SyncSession session, ObjectServerError error) {
     }
 
     /**
-     * FIXME: Add Javadoc
+     * Returns the list of permissions for all Realms available to this user.
```
beeender: `Returns` is not accurate here, it returns a async task. change to `Retrieves`?

realm/realm-library/src/objectServer/java/io/realm/PermissionManager.java
```diff
@@ -43,10 +43,21 @@
...
+ * Current functionality supported by this class:
+ * <ul>
+ *     <li>List users existing permissions.</li>
+ *     <li>List default permissions.</li>
+ *     <li>Modify permissions for a Realm.</li>
+ *     <li>Create a permission offer that can be sent to others.</li>
+ *     <li>Accept permission offers sent by other users.</li>
+ * </ul>
+ * </p>
+ * This class is depending on underlying Realms, so all data coming from this class is thread-confined and must be
```
bmeike: "is depending" > "depends"?

realm/realm-library/src/objectServer/java/io/realm/PermissionManager.java
```diff
@@ -151,10 +162,9 @@ public boolean isGlobalRealm() {
     private RealmResults<PermissionOffer> offers;
 
     /**
-     * FIXME Javadoc
      * Creates a PermissionManager for the given user.
      *
-     * Implementation notes: This class is thread safe since all public methods have thread confined checks in them
+     * This class is thread safe since all public methods have thread confined checks in them
```
bmeike: We should consider avoiding the use of "thread safe" in this context. While completely accurate, it seems to be confusing. I did it too.

Maybe something along the lines of "this class is thread confined, so thread safety is not a concern"?  I don't really know...

Zhuinden: Yeah they downvote me to oblivion on Reddit when I say they're thread-safe because they're thread-confined. I guess people expect "thread-safe == immutable" so I'd just say "this class is thread-confined, and can only be accessed on the thread it was obtained on" or something like that.

cmelchior: My bad, yes we should use thread confined over thread safe IMO, because I agree that thread safe normally means either immutable or synchronized somehow.

PR #5083, opened 2017-08-09
Deprecate Management Realm

cmelchior:
Part of #4558 

Deprecates `getManagementRealm()` in favour of `getPermissionManager()`. The implementation should be deleted in 4.0


Zhuinden:
You'll probably need to change [this section quite a bit then](https://realm.io/docs/java/latest/#access-control).

cmelchior:
Yup, that is on the TODO list before we ship this officially.

cmelchior:
Updated @beeender

PR #5084, opened 2017-08-09
PermissionWrapper - Remove tasks if callback crashes

cmelchior:
Part of #4558 

Found during another review. This ensures that we correctly remove tasks from the task queue even though onSuccess/OnError callbacks throw an exception.

### DIFF
--- a/realm/realm-library/src/objectServer/java/io/realm/PermissionManager.java
+++ b/realm/realm-library/src/objectServer/java/io/realm/PermissionManager.java
@@ -566,8 +566,8 @@ public class PermissionManager implements Closeable {
         }
 
         void notifyCallbackWithSuccess(RealmResults<Permission> permissions) {
-            callback.onSuccess(permissions);
             activeTasks.remove(this);
+            callback.onSuccess(permissions);
         }
     }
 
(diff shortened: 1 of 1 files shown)


beeender:
CI failed @cmelchior

realm/realm-library/src/objectServer/java/io/realm/PermissionManager.java
```diff
@@ -1095,11 +1095,8 @@ public void onChange(RealmResults<PermissionOffer> loadedOffers) {
         }
 
         void notifyCallbackWithSuccess(RealmResults<PermissionOffer> permissions) {
-            try {
-                callback.onSuccess(permissions);
-            } finally {
-                activeTasks.remove(this);
-            }
+            activeTasks.remove(this);
```
Zhuinden: maybe just using try-finally in the other cases would have also fixed the problem if the order matters for some reason

cmelchior: It doesn't matter if the task is removed before or after really, so I opted for having to write less code :)

cmelchior:
The intention is important. I changed it to try-finally.

beeender:
uh, nevermind, saw the comments above

PR #5090, opened 2017-08-10
Refactor compileFieldDescription() in CachedFieldDescriptor and DynamicFieldDescriptor

zaki50:
This PR did not change the logic at all, but gave a common prefix `current` to variables those are updated in the loop.

The most important improvement in this PR is correct usage of `table` and `class`.


bmeike:
@zaki50 : are we going to start making the use of `final` a standard for stack vars?  I would love that!  It has been discussed before, though, and I don't think we agreed on it.  If so, I will, *defiinitely* start requesting it in CRs!

Issue #4323, opened 2017-03-14
Verify `@LinkingObjects` in dynamic and migration scenarios


bmeike:
#### Goal

Reliable, consistent `@LinkingObjectes`

There are, currently no Unit Tests for DynamicObjects that have `@LinkingObjects` fields, nor are there tests for migration scenarios.  There should be.

-----
### Tasks
- [x] Add UTs for dynamic objects
- [x] Add UTs for migration scenarios
- [x] How live are `RealmResults` and how should `ChangeListeners` be triggered ?
- [x] Consider how `CheckedRow` should handle BackLinks. 

#### Version of Realm and tooling

Realm version(s): 3.2

cmelchior:
@bmeike Since @LinkingObject fields are not expressed in the Realm Schema, no migrations will be triggered either way. Is there anything left to do in this issue or can it just be closed?

bmeike:
It can, probably, be closed.  This was a place-holder for a checkmark in the original Issue. Will verify ASAP.

cmelchior:
@bmeike Any update on this?

bmeike:
@zaki50 added a complete suite of DynamicRealm tests.  They are good to go.

bmeike:
Opened a new ticket to track the last two, highly ephemeral items in this ticket

PR #5092, opened 2017-08-10
Fix Backlink UTs

bmeike:
Fix the backlink UTs so that they pass, Issue #4323 .

This will have to be revisited after @beeender 's commit moving to Object Store.

On hold for now.  There is probably little value in reviewing this, at this point.


bmeike:
Can we close this?

Zhuinden:
At a quick glance, I'd say this is definitely outdated and no longer relevant.

The one thing missing with backlinks that I know of is support for unnamed backlinks in Java query api, but that never happened.

Despite that I contributed some example code update, I'm still not an official though :stuck_out_tongue_winking_eye:

PR #5098, opened 2017-08-11
Fix class name in exception message

### LINKED ISSUE #5096: Class name in error message from query condition methods is wrong
Following test case will fail with `org.junit.ComparisonFailure: expected:<...eld 'age' in table '[Cat]' is of invalid type...> but was:<...eld 'age' in table '[Owner]' is of invalid type...>`

```
@Test
public void isEmpty_acrossLink_wrongTypeThrows() {
    for (RealmFieldType type : RealmFieldType.values()) {
        if (SUPPORTED_IS_EMPTY_TYPES.contains(type)) {
            continue;
        }

        RealmQuery<Owner> query = realm.where(Owner.class);
        try {
            query.isEmpty(Owner.FIELD_CAT + "." + Cat.FIELD_AGE);
            fail();
        } catch (IllegalArgumentException expected) {
            assertEquals(String.format(Locale.US,
                    "Invalid query: field '%s' in table '%s' is of invalid type '%s'.",
                    Cat.FIELD_AGE, Cat.CLASS_NAME, RealmFieldType.INTEGER.name()),
                    expected.getMessage());
        }
    }
}
```

The point is that the test case uses link query.

zaki50:
This PR fixes #5096

The important change is in `CachedFieldDescriptor#compileFieldDescription()`.
At the end of the method, they passed `className` as an first argument of `setCompilationResults()`. However, that argument should be the name of the container class of final field and actually the `className` is not the one. 

This PR also improve performance of `CachedFieldDescriptor#compileFieldDescription()`.
In the method, map lookup happened three times with the same key for each field. This PR replaced it with just one lookup for each field.

### DIFF
--- a/realm/realm-library/src/main/java/io/realm/internal/ColumnInfo.java
+++ b/realm/realm-library/src/main/java/io/realm/internal/ColumnInfo.java
@@ -33,7 +33,7 @@ import io.realm.RealmFieldType;
  * <p>
  * While the use of the fields in {@code ColumnDetails} is consistent, there are three subtly different cases:
  * <ul>
- * <li>If the column type is a simple type, the {@code peerClassName} field is empty (0L / NULLPTR)</li>
+ * <li>If the column type is a simple type, the {@code peerClassName} field is {@code null}</li>
  * <li>If the column type is OBJECT or LINK, the {@code peerClassName} field is the class name of the OBJECT/LINK type</li>
  * <li>If the column type is LINKING_OBJECT, the {@code peerClassName} field is the class name of the backlink source table
  * and the column index field is the index of the backlink source field, in the source table</li>
@@ -134,7 +134,7 @@ public abstract class ColumnInfo {
     }
 
     /**
-     * Returns the {@link ColumnDetails}, in the described table, for the named column.
+     * Returns the {@link ColumnDetails}, in the described table, of the named column.
      *
      * @return {@link ColumnDetails} or {@code null} if not found.
      */
--- a/realm/realm-library/src/main/java/io/realm/internal/ColumnInfo.java
+++ b/realm/realm-library/src/main/java/io/realm/internal/ColumnInfo.java
@@ -33,9 +33,9 @@ import io.realm.RealmFieldType;
  * <p>
  * While the use of the fields in {@code ColumnDetails} is consistent, there are three subtly different cases:
  * <ul>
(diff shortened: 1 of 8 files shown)


realm/realm-library/src/main/java/io/realm/internal/ColumnInfo.java
```diff
@@ -60,23 +60,24 @@
 public abstract class ColumnInfo {
 
     // Immutable column information
-    private static final class ColumnDetails {
+    public static final class ColumnDetails {
         public final long columnIndex;
         public final RealmFieldType columnType;
-        public final String linkTable;
+        public final String peerClassName;
```
cmelchior: Not sure I understand this name. Wouldn't `linkedClassName` be better?

realm/realm-library/src/main/java/io/realm/internal/ColumnInfo.java
```diff
@@ -33,9 +33,9 @@
  * <p>
  * While the use of the fields in {@code ColumnDetails} is consistent, there are three subtly different cases:
  * <ul>
- * <li>If the column type is a simple type, the link table field is empty (0L / NULLPTR)</li>
- * <li>If the column type is OBJECT or LINK, the link table field is the class name of the OBJECT/LINK type</li>
- * <li>If the column type is LINKING_OBJECT, the link table field is the class name of the backlink source table
+ * <li>If the column type is a simple type, the {@code peerClassName} field is {@code null}</li>
```
bmeike: This is not true.  It needs to be 0L / NULLPTR.  This value is use only by native code and it must appear to the native code, to be NULLPTR.

zaki50: I don't think we should describe Javadoc comment in C++ terms.
And this class should not care about whether the value is used only by native code or not.

Since the type of `peerClassName` is `String`, shouldn't we use the term `null` instead of `0L/NULLPTR`?

realm/realm-library/src/main/java/io/realm/internal/ColumnInfo.java
zaki50: fixed

zaki50:
@bmeike What do you think? https://github.com/realm/realm-java/pull/5098#discussion_r132810980
If you are fine with my comment, I'll merge this.

bmeike:
Approved quite a while ago....

PR #5102, opened 2017-08-13
PermissionWrapper - Support ClientReset

cmelchior:
Part of https://github.com/realm/realm-java/pull/4558

This PR adds support for Client Reset happening to any of the underlying Realms. It is a fairly complex error scenario, so I tried to describe it as accurately as possible in the code comments. So please read those for reasoning + clarification.


dalinaum:
There are `notifyCallbackError` and `notifyCallbackWithSuccess`. It doesn't seem consistent. Should it be 1) `notifyCallbackError` and `notifyCallbackSuccess` or 2) `notifyCallbackWithError` and `notifyCallbackWithSuccess`?

realm/realm-library/src/objectServer/java/io/realm/PermissionManager.java
```diff
@@ -1001,12 +1012,48 @@ protected final boolean checkAndReportInvalidState() {
                 return false;
             }
 
+            // Handle Client Reset if it happened in any of the Realms.
+            // A Client Reset is a fatal error for the PermissionManager, so all current and
+            // future tasks will exit as soon as posible after this event happened and report it
+            // through the error callback. Only action a user can take is to close the
+            // PermissionManager and re-open it again Some data might be lost (like permission
```
dalinaum: it again. Some data

realm/realm-library/src/objectServer/java/io/realm/PermissionManager.java
```diff
@@ -1001,12 +1012,48 @@ protected final boolean checkAndReportInvalidState() {
                 return false;
             }
 
+            // Handle Client Reset if it happened in any of the Realms.
+            // A Client Reset is a fatal error for the PermissionManager, so all current and
+            // future tasks will exit as soon as posible after this event happened and report it
```
nhachicha: posible -> possible

realm/realm-library/src/objectServer/java/io/realm/PermissionManager.java
```diff
@@ -1001,12 +1012,48 @@ protected final boolean checkAndReportInvalidState() {
...
+            // through the error callback. Only action a user can take is to close the
+            // PermissionManager and re-open it again Some data might be lost (like permission
+            // offers not yet processed). This is currently unavoidable.
+            // TODO: Eventually we might be able to recover the permission manager from this event
+            // but it will require some serious task management as we would need to do a full
+            // close, reschedule all tasks, and re-open behind users back. This is out of scope for
+            // now.
+            if (managementErrorHappened && managementError instanceof ClientResetRequiredError) {
+                ClientResetRequiredError cr = (ClientResetRequiredError) managementError;
+                permissionManager.managementRealm.close();
```
nhachicha: are we sure this is the only instance open?

realm/realm-library/src/objectServer/java/io/realm/PermissionManager.java
```diff
@@ -1001,12 +1012,48 @@ protected final boolean checkAndReportInvalidState() {
...
+                cr.executeClientReset();
+                permissionManager.clientReset = true;
+            }
+
+            if (permissionErrorHappened && permissionError instanceof ClientResetRequiredError) {
```
nhachicha: should you check if the `permissionManager.clientReset` is set already to avoid performing the `executeClientReset` multiple time?

realm/realm-library/src/syncIntegrationTest/java/io/realm/PermissionManagerTests.java
```diff
@@ -172,6 +172,84 @@ public void onError(ObjectServerError error) {
...
+                SyncManager.simulateClientReset(SyncManager.getSession(pm.permissionRealmConfig));
+                pm.getPermissions(new PermissionManager.PermissionsCallback() {
+                    @Override
+                    public void onSuccess(RealmResults<Permission> permissions) {
+                        fail();
+                    }
+
+                    @Override
+                    public void onError(ObjectServerError error) {
+                        assertEquals(ErrorCode.CLIENT_RESET, error.getErrorCode());
```
nhachicha: is the `error` populated with the location of the saved Realm? should we check the file exist?

cmelchior: Yes, it has the location, but we have general tests for Client Reset that verifies that, so It shouldnt' be needed here.

realm/realm-library/src/objectServer/java/io/realm/PermissionManager.java
cmelchior: I was about to say "Yes", because the PermissionManager is a singleton, but it is only a thread-singleton, so technically you could have a PermissionManager open on two different looper threads. I would estimate the chance of that ever happening to be virtually zero, but your right it is a concern.

In that case both PermissionManagers would receive the error callback and trigger this code. Both of them would "move" the file, which is unneassery work, but afaik it should be safe, since they would treat the error as fatal and not use the Realm anymore, so any data shouldn't be accessed on either thread. 

I'll create a unit test to verify that assumption.

realm/realm-library/src/objectServer/java/io/realm/PermissionManager.java
cmelchior: This code will only be called once. There is a check for the flag at the top of this method that will abort quickly if set.

realm/realm-library/src/objectServer/java/io/realm/PermissionManager.java
```diff
@@ -1001,12 +1012,48 @@ protected final boolean checkAndReportInvalidState() {
...
+            }
+
             // Handle errors
             Map<String, ObjectServerError> errors = new LinkedHashMap<>();
-            if (managementErrorHappened) { errors.put("Management Realm", managementError); }
-            if (permissionErrorHappened) { errors.put("Permission Realm", permissionError); }
-            if (defaultPermissionErrorHappened) { errors.put("Default Permission Realm", defaultPermissionError); }
-            notifyCallbackError(combineRealmErrors(errors)); // This will remove the task from the task list
+            if (permissionManager.clientReset) {
+                errors.put("ClientReset", new ObjectServerError(ErrorCode.CLIENT_RESET, ERROR_MESSAGE_CLIENT_RESET));
```
beeender: Add some more information about which realm has been performed client reset.

realm/realm-library/src/objectServer/java/io/realm/PermissionManager.java
beeender: In that case, the first one will throw if the 2nd one Realm has not been closed yet.

beeender: You need to handle the client reset case in the `PermissionManager.close()`. And need a test case for that.

cmelchior: Object store is actually crashing hard in that case. For now I'll implement a check that only allow the `PermissionManager to be opened from the main thread. Since it requires a looper that is probably what 99% of people would do anyway.

I'll create an issue for tracking it as well though.

cmelchior:
@beeender I restricted the access to the UI thread which should solve these concurrency issues for now. There seems to be bugs in Object Store preventing other solutions + I'm guessing pratically no-one will try to open this on multiple handler threads anyway.

realm/realm-library/src/androidTestObjectServer/java/io/realm/SessionTests.java
```diff
@@ -188,6 +191,58 @@ public void onError(SyncSession session, ObjectServerError error) {
...
+                    @Override
+                    public void onError(SyncSession session, ObjectServerError error) {
+                        final ClientResetRequiredError handler = (ClientResetRequiredError) error;
+
+                        // Execute Client Reset
+                        looperThread.closeTestRealms();
+                        handler.executeClientReset();
+
+                        // Validate that files have been moved
+                        assertFalse(handler.getOriginalFile().exists());
```
nhachicha: include these asserts in the Runnable, otherwise, we will only log the assertion failure, without calling `looperThread.testComplete();`

realm/realm-library/src/androidTestObjectServer/java/io/realm/SessionTests.java
```diff
@@ -188,6 +191,58 @@ public void onError(SyncSession session, ObjectServerError error) {
...
+                                looperThread.addTestRealm(instance);
+                                try {
+                                    SyncManager.getSession(config).downloadAllServerChanges();
+                                    looperThread.testComplete();
+                                } catch (InterruptedException e) {
+                                    fail(e.toString());
+                                }
+                            }
+                        });
+
```
nhachicha: remove spaces

realm/realm-library/src/androidTestObjectServer/java/io/realm/SessionTests.java
```diff
@@ -188,6 +191,58 @@ public void onError(SyncSession session, ObjectServerError error) {
...
 
+    // Check that if we manually trigger a Client Reset, then it should be possible to start
+    // downloading the Realm immediately after.
+    @Test
+    @RunTestInLooperThread
+    @Ignore("https://github.com/realm/realm-java/issues/5143")
+    public void clientReset_manualTriggerAllowSessionToRestart() {
+        SyncUser user = createTestUser();
+        String url = "realm://objectserver.realm.io/~/myrealm";
+        final AtomicReference<SyncConfiguration> configRef = new AtomicReference<>(null);
```
nhachicha: why an `AtomicReference` `configRef` being final can be accessed from the runnable

realm/realm-library/src/objectServer/java/io/realm/PermissionManager.java
nhachicha: what about calls to `getPermissionManager` within the same thread the ref count will still be > 1 & closing here will only decrement the cache?

realm/realm-library/src/androidTestObjectServer/java/io/realm/SessionTests.java
cmelchior: Because the `config` is being used inside the builder. You cannot reference a final field while it is being created.

cmelchior:
> what about calls to getPermissionManager within the same thread the ref count will still be > 1 & closing here will only decrement the cache?

The reference count is for the permission manager instance. This closes the underlying Realm instance. As the reference counter ensures that we only ever have 1 PermissionManager pr. thread, there is also only 1 Realm instance.

cmelchior:
All review comments addressed. Merging.

PR #5119, opened 2017-08-18
Update Javadoc to not reference Realm.getDefaultInstanceAsync.

dalinaum:
Fix https://github.com/realm/realm-java/issues/2299#issuecomment-303508956

### DIFF
--- a/realm/realm-library/src/main/java/io/realm/BaseRealm.java
+++ b/realm/realm-library/src/main/java/io/realm/BaseRealm.java
@@ -774,7 +774,11 @@ abstract class BaseRealm implements Closeable {
      *   protected void onCreate(Bundle savedInstanceState) {
      *     super.onCreate(savedInstanceState);
      *     setContentView(R.layout.layout_main);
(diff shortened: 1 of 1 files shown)


realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -774,7 +774,11 @@ protected RealmObjectContext initialValue() {
      *   protected void onCreate(Bundle savedInstanceState) {
      *     super.onCreate(savedInstanceState);
      *     setContentView(R.layout.layout_main);
-     *     realmAsyncTask = Realm.getDefaultInstanceAsync(new Callback() {
+     *     RealmConfiguration config = new RealmConfiguration.Builder()
```
beeender: move the `config` as an static member var?

dalinaum:

dalinaum:
I prefer creating a configuration explicitly to using a default configuration. I think users could get the feel when he/she should use `Realm.getDefaultInstanceAsync()` easily.

PR #5125, opened 2017-08-20
Enable @ParametersAreNonnullByDefault annotation in Kotlin compiler.

zaki50:


zaki50:
@cmelchior Ah, yes. I'll update this PR.

realm/realm-library/src/androidTest/kotlin/io/realm/KotlinSchemaTests.kt
```diff
@@ -49,7 +49,7 @@ class KotlinSchemaTests {
 
     @Test
     fun kotlinTypeNonNull() {
-        val objSchema = realm.schema.get(AllKotlinTypes::class.simpleName)!!
+        val objSchema = realm.schema.get(AllKotlinTypes::class.simpleName!!)!!
 
         // Document current nullability. Ideally all should be non-nullable. This is currently
```
bmeike: This whole test is pretty weird.  Is it testing only something that should not be true?  I looked at the related issue and couldn't find anything that seemed to pertain, so maybe  it would be cool to say a little more about what's going on in there...

zaki50: I guess that this comment means we should respect nullability information in Kotlin's type information.

Currently users need to add `@Required` annotation to define non-null field even if the type of the field is non-null (e.g. `String`, not `String?`).
Once we implement that in our annotation processor, fields below are all non-null.

cmelchior: Yes exactly. This means that we should add support for `@Nonnull` as an alternative way of defining required types. This would make the interop be seamless with Kotlin as this is what they do in their bytecode. Of course, it also means that you could write: `
@NonNull public String name;` instead of `@Required public String name` in Java, but that is IMO not a big problem. We shouldn't encourage it either though.

zaki50: I've update the comment.

zaki50:
Since nullability annotations are merged into `releases` branch, I've changed base branch to `releases`.

zaki50:
I'll merge this PR after CI success.

PR #5126, opened 2017-08-20
Apply kotlin-kapt plugin

zaki50:
Original kapt feature has been deprecated. This PR enables `kotlin-kapt` plugin instead.


cmelchior:
Don't we do this in our plugin?

zaki50:
@cmelchior As long as our plugin works without `kotlin-kapt` plugin, I think that whether applying it or not is user's choice.

zaki50:
Ah, this PR should target `releases`...

PR #5129, opened 2017-08-21
Update realm-sync to 1.10.8

beeender:
And update object-store


zaki50:
A new error code `Client::Error::ssl_server_cert_rejected` seems to be added in `1.10.6`.
Don't we need to add a code to follow that change?

cmelchior:
Yes we do

PR #5133, opened 2017-08-21
Add support for bulk updating fields in query results

### LINKED ISSUE #762: Batch or bulk updates
Feature request following this discussion: https://groups.google.com/forum/#!topic/realm-java/6NeywfWmCA4

Basically it allows setting a property on many objects in one operation just like this SQL:

```
UPDATE Foo SET isDeleted = 1 WHERE name = "123"
```

One option could be

```
RealmResults<Foo> foos = realm.where(Foo.class).equalsTo("name", "123");
realm.beginTransaction();
foos.updateAllObjects("isDeleted", 1);
realm.commitTransaction();
```

bmeike:
Closes #762

**CM:**
This PR adds methods for updating all objects using a dynamic like API.

* The API methods mirrors the ones found in `DynamicRealmObject`, i.e. methods for all non-nullable types like `setLong(fieldName, long)`,  `setNull(fieldName)` and `setValue(fieldName, Object)` that does automatic type conversion.
* It currently depends on functionality found in  #6224, which is why it is using it as base. Rebase once it is merged.
* Only first-level properties are supported in this iteration, i.e you cannot do something like `
realm.where(Person.class).findAll().setString("dog.name", "Fido")`.

TODO
- [x] Add support for Lists using `OsObjectBuilder` + tests
- [x] Hook native code up to `JavaContext`
- [x] Add tests for all types
- [x] Updating linked fields will not be supported in first iteration. Make issue for tracking interest in that: https://github.com/realm/realm-java/issues/6278

----
**BMEIKE:**
OUTDATED:
This is my first guess at an API for the bulk update feature.  I'd like comment, in particular, about the following things:
* Is this the API you are looking for?
* Is this the right way to handle null values?
* Is the handling of exceptions and errors basically correct?
* Is `RealmObject` the right parameter type for 'RealmResults.setObject`
* Is `UncheckedRow` the right parameter type for 'Collection.setObject`
Any other kinda thing ya got to say.


realm/realm-library/src/main/cpp/io_realm_internal_Collection.cpp
```diff
@@ -414,8 +414,46 @@ JNIEXPORT jboolean JNICALL Java_io_realm_internal_Collection_nativeDeleteFirst(J
     return JNI_FALSE;
 }
 
-JNIEXPORT void JNICALL Java_io_realm_internal_Collection_nativeDelete(JNIEnv* env, jclass, jlong native_ptr,
-                                                                      jlong index)
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wunused-parameter"
```
cmelchior: We shouldn't hide this error in general, but it is fine while prototyping.

realm/realm-library/src/main/java/io/realm/RealmResults.java
```diff
@@ -16,19 +16,25 @@
...
+import io.realm.exceptions.RealmException;
 import io.realm.internal.CheckedRow;
 import io.realm.internal.Collection;
+import io.realm.internal.RealmObjectProxy;
 import io.realm.internal.Row;
 import io.realm.internal.SortDescriptor;
 import io.realm.internal.Table;
 import io.realm.internal.UncheckedRow;
+import io.realm.internal.Util;
+import io.realm.internal.android.JsonUtils;
```
cmelchior: This looks like a mistake>

realm/realm-library/src/main/java/io/realm/RealmResults.java
```diff
@@ -38,8 +44,8 @@
  * increases speed.
  * <p>
  * RealmResults are live views, which means that if it is on an {@link Looper} thread, it will automatically
- * update its query results after a transaction has been committed. If on a non-looper thread, {@link Realm#waitForChange()}
- * must be called to update the results.
+ * update its query results after a transaction has been committed. If on a non-looper thread,
+ * {@link Realm#waitForChange()} must be called to update the results.
```
cmelchior: Since we re-added `refresh` `Realm#refresh()` would probably be more accurate.

realm/realm-library/src/main/java/io/realm/RealmResults.java
```diff
@@ -138,6 +143,279 @@ public boolean load() {
     }
 
     /**
+     * Sets the value to {@code null} for the given field in all of the objects in the result collection.
```
cmelchior: result collection -> collection ?

cmelchior: result collection -> collection ?
Same for the rest of the methods

realm/realm-library/src/main/java/io/realm/RealmResults.java
```diff
@@ -138,6 +143,279 @@ public boolean load() {
...
+     *
+     * @param fieldName name of the field to update.
+     * @throws IllegalArgumentException if field name doesn't exist or isn't nullable.
+     * @throws RealmException if the field is a {@link io.realm.annotations.PrimaryKey} field.
```
cmelchior: Not `IllegalArgumentException`?

realm/realm-library/src/main/java/io/realm/RealmResults.java
```diff
@@ -337,4 +619,17 @@ public void removeChangeListener(OrderedRealmCollectionChangeListener<RealmResul
...
     }
+
+    private boolean checkAndSetNullValue(@Nonnull String fieldName, @Nullable Object value) {
+        if (value != null) { return false; }
+
+        setNull(fieldName);
+        return true;
+    }
+
+    private void verifyString(@Nonnull String fieldName) {
```
cmelchior: Historically we used `checkString` for these kinds of asserts.

realm/realm-library/src/main/java/io/realm/RealmResults.java
```diff
@@ -138,6 +143,279 @@ public boolean load() {
...
+    /**
+     * Sets the {@code boolean} value of the given field in all of the objects in the result collection.
+     *
+     * @param fieldName name of the field to update.
+     * @param value new value for the field.
+     * @throws IllegalArgumentException if field name doesn't exist or field isn't boolean.
+     */
+    public void setBoolean(@Nonnull String fieldName, @Nullable Boolean value) {
+        verifyString(fieldName);
+
```
cmelchior: Not sure I understand the need for these empty lines in this and other methods?

realm/realm-library/src/main/java/io/realm/RealmResults.java
```diff
@@ -138,6 +143,279 @@ public boolean load() {
     }
 
     /**
+     * Sets the value to {@code null} for the given field in all of the objects in the result collection.
+     *
+     * @param fieldName name of the field to update.
+     * @throws IllegalArgumentException if field name doesn't exist or isn't nullable.
+     * @throws RealmException if the field is a {@link io.realm.annotations.PrimaryKey} field.
+     */
+    public void setNull(@Nonnull String fieldName) {
```
cmelchior: Since you are handling `null` in the other methods using `checkAndSetNullValue()` it seems a bit odd to have this public. Although there is a symmetry argument with `DynamicRealmObject`, so it is probably fine

realm/realm-library/src/main/java/io/realm/RealmResults.java
```diff
@@ -138,6 +143,279 @@ public boolean load() {
...
+        collection.setDouble(getColumnIndex(fieldName), value);
+    }
+
+    /**
+     * Sets the {@code String} value of the given field in all of the objects in the result collection.
+     *
+     * @param fieldName name of the field to update.
+     * @param value new value for the field.
+     * @throws IllegalArgumentException if field name doesn't exist or field isn't string.
+     * @throws RealmException if the field is a {@link io.realm.annotations.PrimaryKey} field.
```
cmelchior: Same, why not IllegalArgumentException?

realm/realm-library/src/main/java/io/realm/RealmResults.java
```diff
@@ -138,6 +143,279 @@ public boolean load() {
...
+     * <p>
+     * Using the typed setters is faster than using this method.
+     *
+     * @throws IllegalArgumentException if field name doesn't exist or if the input value cannot be converted
+     * to the appropriate input type.
+     * @throws NumberFormatException if a String based number cannot be converted properly.
+     * @throws RealmException if the field is a {@link io.realm.annotations.PrimaryKey} field.
+     */
+    @SuppressWarnings("unchecked")
+    public void set(@Nonnull String fieldName, @Nullable String strValue) {
```
cmelchior: If this method is mirroring the `set` method in `DynamicRealmObject` it should be `@Nullable Object` and it should also handle RealmObjects and Lists.

realm/realm-library/src/main/java/io/realm/RealmResults.java
```diff
@@ -138,6 +143,279 @@ public boolean load() {
...
+        verifyString(fieldName);
+
+        realm.checkIfValid();
+
+        if (checkAndSetNullValue(fieldName, value)) { return; }
+
+        long columnIndex = getColumnIndex(fieldName);
+
+        if (!(value instanceof RealmObjectProxy)) {
+            throw new IllegalArgumentException("A field value must be a managed Realm object.");
```
cmelchior: `'value' must be a managed Realm object.`

realm/realm-library/src/main/java/io/realm/RealmResults.java
```diff
@@ -138,6 +143,279 @@ public boolean load() {
...
+        }
+
+        ProxyState proxyState = ((RealmObjectProxy) value).realmGet$proxyState();
+        if (proxyState.getRealm$realm() != realm) {
+            throw new IllegalArgumentException("A field value must belong to the same Realm as its owner.");
+        }
+
```
cmelchior: It seems odd to use the generic term "a field value" when we are talking about a specific parameter. Using `'value' must` seems more appropriate to me.

cmelchior:
To answer your comments:

1) 
Having the API match the one in `DynamicRealmObject` does seem like a good idea, and personally, I find methods like `setInt(field, intVal)` more readable, but it is quite a few extra methods.

You could also argue that this approach doesn't really scale well if we start supporting a larger number of classes, e.g. this doesn't support `MutableRealmInteger` well. DynamicRealmObject doesn't either though.

Also, the speed advantage of using `setString()/setInt()` is not relevant for the Collection updates as those will never be called in a hot loop compared to the `DynamicRealmObject` methods that might be.

So in short, I'm still torn between this and just having one method `setValue()` (your current `set`). But it doesn't seem totally off base to go for symmetry with `DynamicRealmObject` for now, and then address the problem across all classes if or when we add support for something like TypeAdapters or other Realm classes like geo points, Java 8 time or Sets/Maps.

2)
I do not remember the argument for having an explicit `setNull` on `DynamicRealmObject`, and ind your API it doesn't seem needed, but again. If we aim for symmetry with `DynamicRealmObject` having it is fine.

3) Error handling seems fine
4) No, we should use the `RealmModel` interface, which is the most generic representation we have

realm/realm-library/src/main/java/io/realm/RealmResults.java
```diff
@@ -138,6 +143,279 @@ public boolean load() {
...
+    }
+
+    /**
+     * Sets a reference to another object on the given field in all of the objects in the result collection.
+     *
+     * @param fieldName name of the field to update.
+     * @param value new value for the field.
+     * @throws IllegalArgumentException if field name doesn't exist or is of the wrong type.
+     */
+    public void setObject(@Nonnull String fieldName, @Nullable RealmObject value) {
```
cmelchior: Should be `RealmModel`

realm/realm-library/src/main/cpp/io_realm_internal_Collection.cpp
```diff
@@ -414,8 +414,46 @@ JNIEXPORT jboolean JNICALL Java_io_realm_internal_Collection_nativeDeleteFirst(J
...
+JNIEXPORT void JNICALL Java_io_realm_internal_Collection_nativeSetNull(JNIEnv* env, jclass, jlong native_ptr, jlong index)
```
kneth: Will you implement the functionality here or in Object Store?

cmelchior: The idea was Object Store

cmelchior: The idea was Object Store

kneth: :+1:

beeender: Implement this in the Object Store would be a good idea since almost all checks can be moved to object store where they are supposed to be in the future.

beeender:
Why those methods don't apply to `RealmList`? It feels like they should be part of `RealmCollection` ? 

Please be noticed that I am working on using Object Store `List` to replace `LinkView` in the java as the backend of `RealmList`. So if those things are going to be implemented in the Object Store, both `List` and `Results` should have those.

cmelchior:
Yes, these methods should also apply to RealmList. I missed that on the first parse.

realm/realm-library/src/main/cpp/io_realm_internal_Collection.cpp
bmeike: Exactly.  I just want it to compile.

bmeike:
Note that the API as proposed requires boxing and unboxing native arguments.  Again, since it won't be called in a hot loop, perhaps this isn't a problem.

realm/realm-library/src/main/java/io/realm/RealmResults.java
bmeike: We use `JsonUtils.stringToDate` and `Util.isEmptyString`

realm/realm-library/src/main/java/io/realm/RealmResults.java
bmeike: Ok.  I just formatted this.  Will fix.

realm/realm-library/src/main/java/io/realm/RealmResults.java
bmeike: Done

bmeike: Done

realm/realm-library/src/main/java/io/realm/RealmResults.java
bmeike: You mean instead of the RealmException.  This is just paralleling the the Dynamic API.

realm/realm-library/src/main/java/io/realm/RealmResults.java
bmeike: I almost went with requiring explicitly setting `null`, and using primitive types in the API methods.  That seems significantly more elegant, but it pushes null checks into the client code, which I think is unacceptable.

realm/realm-library/src/main/java/io/realm/RealmResults.java
bmeike: Readability?  Looks nice to me.  Will change if you like.

realm/realm-library/src/main/java/io/realm/RealmResults.java
bmeike: As above.  This is what the Dynamic API does.

realm/realm-library/src/main/java/io/realm/RealmResults.java
bmeike: See below...

realm/realm-library/src/main/java/io/realm/RealmResults.java
bmeike: My thinking is that, what this method does, is to set the value of a field for all of the members of the collection.  From that I abbreviated "the value for a member object's field" to "field value".  How about "A field's value must be..."?

realm/realm-library/src/main/java/io/realm/RealmResults.java
bmeike: I *really* don't like what that method does.  It is a loosey-goosey, untyped horror.  This method has a specific meaning: it translates string data into the correct type for the field.  I would rather fix the other one (yes, yes, I know we won't do that).  I can see making this similar to the one in Dynamic, but I will hate doing it.

realm/realm-library/src/main/java/io/realm/RealmResults.java
bmeike: You just mean change the name?  I don't see a `checkString` around...

realm/realm-library/src/main/java/io/realm/RealmResults.java
bmeike: Ok... but that is kind of a lie.  What is actually required is a `RealmObjectProxy`.  It is, slightly, a perversion of the typing system.
Fixed

realm/realm-library/src/main/cpp/io_realm_internal_Collection.cpp
bmeike: I think there is a pretty good argument for keeping any checks that are reasonable, on this side of the JNI boundary.

bmeike: Will move this stuff to OS.

bmeike:
Christian said: ```
Not sure I understand the question. Checked/UncheckedRow is really constructs used in Java, where CheckedRow is used by the DynamicAPI, while UncheckedRow is used by the typed API. Reason being that in the typed API we have much better guarantees about the types through the annotation processor, so no need to check those, hence Unchecked. Inside the C++ Collection these two classes doesn't make much sense since we just have a rowIndex to a Row in Core
```
Right.  `UncheckedRow` is a candidate because, from it, I can get the type of the value to be inserted into the field `row.getTable().getClassName()`, and the row index of the value to be inserted.

File: realm/realm-library/src/main/java/io/realm/RealmResults.java (hunk 7 above)
Comment (cmelchior) [inline realm/realm-library/src/main/java/io/realm/RealmResults.java]: No a big issue

File: realm/realm-library/src/main/java/io/realm/RealmResults.java (hunk 9 above)
Comment (cmelchior) [inline realm/realm-library/src/main/java/io/realm/RealmResults.java]: Okay, no idea why we did that there, feels like a mistake :)

File: realm/realm-library/src/main/java/io/realm/RealmResults.java (hunk 12 above)
Comment (cmelchior) [inline realm/realm-library/src/main/java/io/realm/RealmResults.java]: Well, in this case you are talking about the input parameter, of which there is only one, not the collection right?

File: realm/realm-library/src/main/java/io/realm/RealmResults.java (hunk 10 above)
Comment (cmelchior) [inline realm/realm-library/src/main/java/io/realm/RealmResults.java]: IMO doing a third variant here would be bad, so I would prefer if it either did what `DynamicRealmObject.set()` does or just remove it altogether?

File: realm/realm-library/src/main/java/io/realm/RealmResults.java (hunk 6 above)
Comment (cmelchior) [inline realm/realm-library/src/main/java/io/realm/RealmResults.java]: Sorry, i meant just change the name here

**Comment by cmelchior** [comment]
> Note that the API as proposed requires boxing and unboxing primitive arguments. Again, since it won't be called in a hot loop, perhaps this isn't a problem.

That might have been the argument for explicit `setNull` in the DynamicRealmObject API, to avoid this.

> but I do not understand whether or not it is the right abstraction.

The `Row` interface has the `getTable` method and since a `RealmResults` can either be typed or dynamic e.g. `RealmResults<DynamicRealmObject>` I don't think you can depend on a specific subtype of `Row`

File: realm/realm-library/src/main/java/io/realm/RealmResults.java [hunk 15]
```diff
@@ -138,6 +143,279 @@ public boolean load() {
...
+     * @throws IllegalArgumentException if field name doesn't exist or field isn't boolean.
+     */
+    public void setBoolean(@Nonnull String fieldName, @Nullable Boolean value) {
+        verifyString(fieldName);
+
+        realm.checkIfValid();
+
+        if (checkAndSetNullValue(fieldName, value)) { return; }
+
+        collection.setBoolean(getColumnIndex(fieldName), value);
```
cmelchior: I'm not 100% sure, but If we are going to use the `Results` class in Object Store, I'm guessing that is going to take the fieldname as input to it's methods instead of a column index, so `Results.setBoolean(fieldName, value)`. I believe OS already have functionality for mapping between a field name and the index.

realm/realm-library/src/main/java/io/realm/RealmResults.java
bmeike: The deciding factor, then, is whether this will be used for JSON.  It appears to me that the main argument for that method is that it makes JSON parsing much easier.  If this bulk update will be used in the context of parsing JSON, I will make this method like the one in Dynamic.  If we don't see a JSON use-case, I will just remove it.

realm/realm-library/src/main/java/io/realm/RealmResults.java
bmeike: Yeah, that's an important consideration that I didn't get.  I will move the implementations to OS, tomorrow, to see how it affects the APIs.

realm/realm-library/src/main/java/io/realm/RealmResults.java
cmelchior: The use case for `DynamicRealmObject.set()` was CSV, not JSON, since JSON is more about whole objects while CSV is about dynamically parsing single lines and fields.

I doubt that would be the case for `RealmResults.setX()`. I see that more used in cases like "Mark all emails as read"-scenarios, aka. UI driven events that operate on the local Realm.

So IMO I would remove it unless you can see a use case?

Zhuinden: Welp, I know that *my* usecase would have been to set a certain boolean field to `false` for *every single item in the Realm*, and that's the most common one I've seen elsewhere too. Setting a boolean or String or Date or stuff like that for each element in the Results, where the Results is kinda big and causes GC overhead by iterating in Java.

Setting a JSON for a field's value directly seems awkward, if that was the route I'd assume that you could only set managed RealmObject.

realm/realm-library/src/main/java/io/realm/internal/Collection.java
```diff
@@ -594,6 +711,16 @@ private static native long nativeCreateResultsFromLinkView(long sharedRealmNativ
 
     private static native void nativeDelete(long nativePtr, long index);
 
+    private static native void nativeSetNull(long nativePtr, long columnIndex);
+    private static native void nativeSetBoolean(long nativePtr, long columnIndex, boolean value);
+    private static native void nativeSetInt(long nativePtr, long columnIndex, long value);
```
nhachicha: this should be name `nativeSetLong` (since core uses long to represent Int, long ...) similar to [setLong](https://github.com/realm/realm-java/blob/19e45278726ecb90f1d37b39146753b46537c4bb/realm/realm-library/src/main/java/io/realm/internal/Table.java#L518)

realm/realm-library/src/main/java/io/realm/RealmResults.java
bmeike: Yeah, thus my question this AM.  I'm looking into it...

realm/realm-library/src/main/java/io/realm/RealmResults.java
bmeike: RIght.  Right.  CSV, not JSON.  I will remove it.  Happy, happy, joy, joy.

realm/realm-library/src/main/java/io/realm/RealmResults.java
bmeike: Fixed

realm/realm-library/src/main/java/io/realm/RealmResults.java
```diff
@@ -138,6 +143,279 @@ public boolean load() {
...
+        }
+        if (!(row instanceof UncheckedRow)) {
+            throw new IllegalArgumentException("WTF?");
+        }
+
+        collection.setObject(columnIndex, (UncheckedRow) row);
+    }
+
+    /**
+     * Sets the value for the given field  in all of the objects in the collection, converting {@code String}
```
bmeike: Gone

realm/realm-library/src/main/java/io/realm/internal/Collection.java
bmeike: mmmm.... We should revisit once this is in SO and has proper signatures.
I believe, though, that core refers to the type used to represent byte, short, int and long, as Int.  I want to use the name of the core type, since I will call it from the `set` methods for all four java types.

realm/realm-library/src/main/java/io/realm/RealmResults.java
```diff
@@ -138,6 +143,279 @@ public boolean load() {
...
+     * @param fieldName name of the field to update.
+     * @param value new value for the field.
+     * @throws IllegalArgumentException if field name doesn't exist or field isn't boolean.
+     */
+    public void setBoolean(@Nonnull String fieldName, @Nullable Boolean value) {
+        verifyString(fieldName);
+
+        realm.checkIfValid();
+
+        if (checkAndSetNullValue(fieldName, value)) { return; }
```
bmeike: Yes.  Totally agree that this is an essential question.  Please see https://github.com/realm/realm-java/pull/5133#discussion_r134476731 for my reasoning for this choice.  We need to just take a vote or something.  I can see either primitive params and an explicit `setNull` or nullable boxed params.  Either one makes sense.

realm/realm-library/src/main/java/io/realm/RealmResults.java
bmeike: @nhachicha wins the fur-lined syrup pitcher!  Both of the relevant packages are `ParametersAreNonnullByDefault` so the method specific annotations are unnecessary. Woo hoo!

bmeike:
@nhachicha argues:

> Having a dedicated `setNull` + primitive setters will be consistent with the schema:

let’s take the example of
`public void setDouble(String fieldName, @Nullable Double value) `

if I have the following schema
```class Foo extends RealmObject {
    @Required
    Double bar;
    
}
```

Then calling `setDouble` with a `@Nullable` value doesn’t make sense and it’s more error prone.
The current implementation will crash at runtime with the `IllegalStateException`
Having an API upfront that thinks about nullability could avoid those scenarios … (can be detected at compile time)

by calling `setNull` we’re explicit about whether the field is required or not (we can even have a linter to check it if it contradicts the schema)

the non primitive `set` is `@Nonnullable` because we either, pass a managed object or required boxed types. (edited)

realm/realm-library/src/main/java/io/realm/RealmResults.java
emanuelez: Are we ok having android specific code in this class?

cmelchior: Not really. It would have to be refactored for JVM support, but if no immediate replacement exist, we can just punt it to we actually need to

Zhuinden:
Just mentioning this to get a reference from the issue https://github.com/realm/realm-java/issues/762

bmeike:
Can we close this?

Zhuinden:
It's a shame this never happened, it would be so much more memory-efficient than a loop :cry:

realm/realm-library/src/main/cpp/io_realm_internal_OsResults.cpp
```diff
@@ -323,6 +324,91 @@ JNIEXPORT jboolean JNICALL Java_io_realm_internal_OsResults_nativeDeleteFirst(JN
...
+        wrapper->collection().set_property_value(ctx, prop_name, value);
+    }
+    CATCH_STD()
+}
+
+
+JNIEXPORT void JNICALL Java_io_realm_internal_OsResults_nativeSetNull(JNIEnv* env, jclass, jlong native_ptr, jstring j_field_name)
+{
+    TR_ENTER_PTR(native_ptr)
+    auto value = JavaValue();
```
kneth: Can it throw?

realm/realm-library/src/main/cpp/io_realm_internal_OsResults.cpp
```diff
@@ -323,6 +324,91 @@ JNIEXPORT jboolean JNICALL Java_io_realm_internal_OsResults_nativeDeleteFirst(JN
...
+{
+    TR_ENTER_PTR(native_ptr)
+    JavaValue value(j_value);
+    update_objects(env, native_ptr, j_field_name, value);
+}
+
+JNIEXPORT void JNICALL Java_io_realm_internal_OsResults_nativeSetString(JNIEnv* env, jclass, jlong native_ptr, jstring j_field_name, jstring j_value)
+{
+    TR_ENTER_PTR(native_ptr)
+    JStringAccessor str(env, j_value);
```
kneth: Can it throw?

realm/realm-library/src/main/cpp/io_realm_internal_OsResults.cpp
```diff
@@ -323,6 +324,91 @@ JNIEXPORT jboolean JNICALL Java_io_realm_internal_OsResults_nativeDeleteFirst(JN
     return JNI_FALSE;
 }
 
+static inline void update_objects(JNIEnv* env, jlong results_ptr, jstring& j_field_name, JavaValue& value) {
+    try {
```
kneth: I think you need to move the `try`/`CATCH_STD()` to _all_ the actual JNI functions.

cmelchior:
Ready for review. I see @kneth already started :)

realm/realm-library/src/main/java/io/realm/RealmResults.java
```diff
@@ -136,6 +146,403 @@ public boolean load() {
...
+        checkType(fieldName, RealmFieldType.DATE);
+        realm.checkIfValid();
+        osResults.setDate(fieldName, value);
+    }
+
+    /**
+     * Sets a reference to another object on the given field in all of the objects in the collection.
+     *
+     * @param fieldName name of the field to update.
+     * @param value new value for the field.
```
kneth: Maybe "object reference" instead of "value"

realm/realm-library/src/main/java/io/realm/RealmResults.java
```diff
@@ -136,6 +146,403 @@ public boolean load() {
...
+        checkType(fieldName, RealmFieldType.OBJECT);
+        realm.checkIfValid();
+        Row row = checkRealmObjectConstraints(fieldName, value);
+        osResults.setObject(fieldName, row);
+    }
+
+    private Row checkRealmObjectConstraints(String fieldName, @Nullable RealmModel value) {
+        if (value != null) {
+            if (!(RealmObject.isManaged(value) && RealmObject.isValid(value))) {
+                throw new IllegalArgumentException("'value' is not a valid a valid, managed Realm object.");
```
kneth: "a valid a valid" -> "a valid"

realm/realm-library/src/main/java/io/realm/RealmResults.java
```diff
@@ -136,6 +146,403 @@ public boolean load() {
...
+        checkNonEmptyFieldName(fieldName);
+        fieldName = mapFieldNameToInternalName(fieldName);
+        realm.checkIfValid();
+
+        //noinspection ConstantConditions
+        if (list == null) {
+            throw new IllegalArgumentException("Non-null 'list' required");
+        }
+
+        // Due to type erasure it is not to have multiple overloaded methods with the same signature.
```
kneth: "not to" -> "not possible to"

realm/realm-library/src/main/java/io/realm/internal/ColumnInfo.java
```diff
@@ -152,6 +154,16 @@ public ColumnDetails getColumnDetails(String javaFieldName) {
         return indicesFromJavaFieldNames.get(javaFieldName);
     }
 
+    /**
+     * Returns the internal field name that corrosponds to the name found in the Java model class.
```
kneth: "corrosponds" -> "corresponds"

realm/realm-library/src/main/java/io/realm/RealmResults.java
```diff
@@ -136,6 +146,403 @@ public boolean load() {
...
+     * @param fieldName field to update
+     * @param value value to update with.
+     * @throws IllegalArgumentException if the field could not be found, could not be updated or
+     * the argument didn't match the field type or could not be converted to match the underlying
+     * field type.
+     */
+    public void setValue(String fieldName, @Nullable Object value) {
+        checkNonEmptyFieldName(fieldName);
+        fieldName = mapFieldNameToInternalName(fieldName);
+        realm.checkIfValid();
```
nhachicha: You should call `checkIfValidAndInTransaction` instead (same for all `set*` below)

realm/realm-library/src/main/java/io/realm/RealmResults.java
```diff
@@ -136,6 +146,403 @@ public boolean load() {
...
+        osResults.setBoolean(fieldName, value);
+    }
+
+    /**
+     * Sets the {@code byte} value of the given field in all of the objects in the collection.
+     *
+     * @param fieldName name of the field to update.
+     * @param value new value for the field.
+     * @throws IllegalArgumentException if field name doesn't exist, is a primary key property or isn't an integer field.
```
nhachicha: isn't a byte value _user doesn't know the underlying implementation_ (same comment for below)

realm/realm-library/src/main/java/io/realm/internal/OsResults.java
```diff
@@ -395,6 +400,147 @@ public boolean deleteLast() {
...
+                rowPtr = ((CheckedRow) row).getNativePtr();
+            } else {
+                // Should never happen, but just in case.
+                throw new UnsupportedOperationException("Unsupported Row type: " + row.getClass().getCanonicalName());
+            }
+            nativeSetObject(nativePtr, fieldName, rowPtr);
+        }
+    }
+
+    // Interface wrapping adding the specific list tye
```
nhachicha: tye  - > type

realm/realm-library/src/main/java/io/realm/internal/OsResults.java
```diff
@@ -395,6 +400,147 @@ public boolean deleteLast() {
...
+            nativeSetList(nativePtr, fieldName, builder.getNativePtr());
+        } finally {
+            builder.close();
+        }
+    }
+
+    public void setStringList(String fieldName, RealmList<String> list) {
+        addTypeSpecificList(fieldName, list, (builder, lst) -> {
+            builder.addStringList(0, lst);
+
```
nhachicha: remove empty line, same for below

realm/realm-library/src/androidTest/java/io/realm/entities/MappedAllJavaTypes.java
```diff
@@ -0,0 +1,73 @@
...
+import io.realm.annotations.Ignore;
+import io.realm.annotations.Index;
+import io.realm.annotations.LinkingObjects;
+import io.realm.annotations.PrimaryKey;
+import io.realm.annotations.RealmClass;
+import io.realm.annotations.RealmNamingPolicy;
+
+
+@RealmClass(fieldNamingPolicy = RealmNamingPolicy.LOWER_CASE_WITH_UNDERSCORES)
+public class MappedAllJavaTypes extends RealmObject {
```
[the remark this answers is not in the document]
cmelchior: Because this class is used in a unit test that verifies how all the fields work if the underlying name is different than the model class name. In this case all internal names are mapped to `RealmNamingPolicy.LOWER_CASE_WITH_UNDERSCORES`

realm/realm-library/src/main/cpp/io_realm_internal_OsResults.cpp
```diff
@@ -323,6 +324,91 @@ JNIEXPORT jboolean JNICALL Java_io_realm_internal_OsResults_nativeDeleteFirst(JN
...
+    JavaValue value(reinterpret_cast<Row*>(row_ptr));
+    update_objects(env, native_ptr, j_field_name, value);
+}
+
+JNIEXPORT void JNICALL Java_io_realm_internal_OsResults_nativeSetList(JNIEnv* env, jclass, jlong native_ptr, jstring j_field_name, jlong builder_ptr)
+{
+    // OsObjectBuilder have been used to build up the list we want to insert. The list is always
+    // assumed to be at index = 0;
+    std::vector<JavaValue> builder = *reinterpret_cast<std::vector<JavaValue>*>(builder_ptr);
+    REALM_ASSERT_DEBUG(builder.size() == 1);
```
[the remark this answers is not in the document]
cmelchior: Expanded the doc comment to clarify this.

cmelchior:
@nhachicha All comments addressed.

CHANGELOG.md
```diff
@@ -28,6 +28,7 @@ This release also contains all changes in 5.8.0-BETA1 and 5.8.0-BETA2.
 * [ObjectServer] Added support for timeouts when uploading or downloading data manually using `SyncSession.downloadAllServerChanges(long timeout, TimeUnit unit)` and `SyncSession.uploadAllLocalChanges(long timeout, TimeUnit unit)`. [#6073](https://github.com/realm/realm-java/pull/6073)
 * [ObjectServer] Added support for timing out when downloading initial data for synchronized Realms using `SyncConfiguration.waitForInitialRemoteData(long timeout, TimeUnit unit)`. [#6247](https://github.com/realm/realm-java/issues/6247)
 * Added support for `ImportFlag`s to `Realm.copyToRealm()` and `Realm.copyToRealmOrUpdate()`. This makes it possible to choose a mode so only fields that actually changed are written to disk. This improves notifications and Object Server performance. [#6224](https://github.com/realm/realm-java/pull/6224).
+* Added support for bulk updating the same property in all objects part of a query result using `RealmResults.setValue(String fieldName, Object value)` or one of the specialized overrides that have been added for all supported types, e.g. `RealmResults.setString(String fieldName, String value)` [#762](https://github.com/realm/realm-java/issues/762).
```
bmunkholm: "objects part" -> "objects that are part"?

realm/realm-library/src/main/java/io/realm/RealmResults.java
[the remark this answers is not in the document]
cmelchior: Fixed

PR #5135, opened 2017-08-22
apply kotlin-kapt plugin since original kapt feature has been deprecated

zaki50:
This PR is applying the same changes with #5126 to `releases` branch.


zaki50:
I'll merge this as soon as CI accept it.

PR #5137, opened 2017-08-22
Update a reference URL of Client Error

zaki50:
@morten-krogh told me that `doc/protocol.md` was not the one we should refer from `ErrorCode.java`.

https://github.com/realm/realm-sync/pull/1561#issuecomment-323847407

I've update the comment in `ErrorCode.java`.

### DIFF
--- a/realm/realm-library/src/objectServer/java/io/realm/ErrorCode.java
+++ b/realm/realm-library/src/objectServer/java/io/realm/ErrorCode.java
@@ -23,7 +23,7 @@ import java.net.ConnectException;
  */
 public enum ErrorCode {
 
-    // See https://github.com/realm/realm-sync/blob/master/doc/protocol_17.md
+    // See Client::Error in https://github.com/realm/realm-sync/blob/master/src/realm/sync/client.hpp
     // See https://github.com/realm/realm-object-server/blob/master/object-server/doc/problems.md
 
     // Realm Java errors (0-49)


realm/realm-library/src/objectServer/java/io/realm/ErrorCode.java
```diff
@@ -23,7 +23,7 @@
  */
 public enum ErrorCode {
 
-    // See https://github.com/realm/realm-sync/blob/master/doc/protocol_17.md
+    // See Client::Error in https://github.com/realm/realm-sync/blob/master/src/realm/sync/client.hpp
```
beeender: I am bit confused, then what does https://github.com/realm/realm-sync/blob/master/doc/protocol_17.md#error-codes mean?

morten-krogh:
Let me try to make it clear how sync error codes work.

When you register the error_callback, you can get two types of errors:

1. Errors that are found by the server and sent back to the client in a special ERROR message where they are reported to the binding. Those error codes are found in protocol.md or protocol.hpp

2. Errors that originate on the client. The server does not know anything about this error. The error might be due to a buggy server, but the error is discovered by the client itself. The error does not come in an ERROR message. It is discovered directly by the client. These error codes are found in
 client.hpp

zaki50:
Thanks @morten-krogh

realm/realm-library/src/objectServer/java/io/realm/ErrorCode.java
cmelchior: From Mortens description it sounds like we need to keep both a reference to `client.hpp` and the protocol docs, since we can get errors reported from both?

zaki50:
@cmelchior `client.hpp` contains both types of errors.

PR #5139, opened 2017-08-22
Added APIs for Primitive List feature(Design 2)

zaki50:
This PR is basically following `Design 2` in RFC https://github.com/realm/realm-java/pull/4165/files

Many classes and methods don't have Javadoc comment for now. I'll add them later.

Differences are:

* This PR is using `ValueList` instead of `Primitive` or `Primitive List` since values in lists are not limited to `primitive`s.
* This PR omits async APIs. I don't think it's necessary, but need to be discussed.
* This PR does not contain APIs for querying `RealmObject`s with values in the list.

Could you review APIs to make sure that I am on the same track? @realm/java 

----
List of tasks (Added by CM, Am I missing something?):

- [ ] New `RealmValueQuery` class + tests
- [ ] New `RealmValueCollection` interface + tests
- [ ] New `OrderedRealmValueCollection` interface + tests
- [ ] New `RealmValueList` class + tests
- [ ] New `RealmValueResults` class + tests
- [ ] New `RealmValueResultsSnapshot` class + tests
- [ ] Add support to `DynamicRealmObject` + tests
- [ ] Add support to `RealmObjectSchema` + tests
- [ ] Extend `RxObservableFactory` interface to support the above
- [ ] Extend `RealmQuery` to allow queries on value lists
- [ ] Extend AP to emit schema validation for `RealmValueList`

### DIFF
--- a/realm/realm-library/src/main/java/io/realm/valuelist/RealmValueListQuery.java
+++ b/realm/realm-library/src/main/java/io/realm/valuelist/RealmValueListQuery.java
@@ -1,0 +1,444 @@
+/*
+ * Copyright 2017 Realm Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.realm.valuelist;
+
+import java.util.Date;
+
+import javax.annotation.Nullable;
+
+import io.realm.Case;
+import io.realm.Sort;
+
+
+public class RealmValueListQuery<E> {
+
+    public boolean isValid() {
+        // TODO implement this
+        return false;
+    }
+
+    public RealmValueListQuery<E> isNull() {
+        // TODO implement this
+        return null;
+    }
+
+    public RealmValueListQuery<E> isNotNull() {
+        // TODO implement this
+        return null;
+    }
+
+    public RealmValueListQuery<E> equalTo(@Nullable String value) {
+        // TODO implement this
+        return null;
+    }
+
+    public RealmValueListQuery<E> equalTo(@Nullable String value, Case casing) {
+        // TODO implement this
+        return null;
+    }
+
+    public RealmValueListQuery<E> equalTo(@Nullable Long value) {
+        // TODO implement this
+        return null;
+    }
+
+    public RealmValueListQuery<E> equalTo(@Nullable Integer value) {
+        // TODO implement this
+        return null;
+    }
+
+    public RealmValueListQuery<E> equalTo(@Nullable Short value) {
+        // TODO implement this
+        return null;
+    }
+
+    public RealmValueListQuery<E> equalTo(@Nullable Byte value) {
+        // TODO implement this
+        return null;
+    }
+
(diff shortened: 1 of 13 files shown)


Zhuinden:
I'm mostly just curious why `RealmValueList` gets its own package.

cmelchior:
@zaki50 I'll go into more detail later (quite late here), but I really, _really_ wish you if you have taken this discussion up front instead of starting an implementation. Especially, it would be nice to hear your arguments for doing it this way instead of the other.

Not saying that this approach is wrong, but it has a quite big impact on queries (you need a new query class) and collection interfaces (none of our existing interfaces will be able to support this).

realm/realm-library/src/main/java/io/realm/valuelist/RealmValueListQuery.java
```diff
@@ -0,0 +1,444 @@
...
+
+import java.util.Date;
+
+import javax.annotation.Nullable;
+
+import io.realm.Case;
+import io.realm.Sort;
+
+
+public class RealmValueListQuery<E> {
```
beeender: Should this extends from `RealmQuery`? Or share the same interface?

realm/realm-library/src/main/java/io/realm/valuelist/RealmValueListQuery.java
```diff
@@ -0,0 +1,444 @@
...
+        // TODO implement this
+        return null;
+    }
+
+    public RealmValueListQuery<E> notEqualTo(@Nullable String value, Case casing) {
+        // TODO implement this
+        return null;
+    }
+
+    public RealmValueListQuery<E> notEqualTo(@Nullable Long value) {
```
beeender: I actually feel supporting `short`/`int`/`byte` was an incorrect design ... it adds too much methods but not much benefit for users :( `long` only would be good enough for all cases. But it seems we need to follow it.

realm/realm-library/src/main/java/io/realm/valuelist/RealmValueListResults.java
```diff
@@ -0,0 +1,222 @@
...
+import java.util.List;
+import java.util.ListIterator;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+import io.realm.Sort;
+
+
+public class RealmValueListResults<E> extends AbstractList<E>
```
beeender: `RealmValueResults` is better i think

realm/realm-library/src/main/java/io/realm/valuelist/RealmValueListSnapshot.java
```diff
@@ -0,0 +1,222 @@
...
+import java.util.List;
+import java.util.ListIterator;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+import io.realm.Sort;
+
+
+public class RealmValueListSnapshot<E> extends AbstractList<E>
```
beeender: `RealmValueSnapshot`?

cmelchior:
**> I was thinking the option 2 is better, but now it feels there are too much trouble users to adapter it. It is kind of nature to have RealmList<Integer> and RealmList<Dog> to use ... except the queries,(use null for field names) does option 1 has other drawbacks?** 

Not really, that is the biggest one, and that we need to relax the generic constraints in many places from `<? extends RealmModel>` to `<Object>`

The list I compiled in the RFC was this: https://github.com/realm/realm-java/pull/4165/files#diff-09d208b07ba64d6abbf155b031836bc1R229

Note that Cocoa (and I believe .NET) are choosing option 1 which also count against us taking this route.

Zhuinden:
@cmelchior going with option 1 makes it easier to go towards type adapters, too.

zaki50:
I'm making a PR for option 1 as well.

zaki50:
I made a PR to `Design 1` #5145

zaki50:
I'm closing this PR since `Design 1` seems to be a better choice.
I'll revisit this when we need to change the way to `Design 2`.

PR #5145, opened 2017-08-24
Primitive List feature(Part1: API changes)

zaki50:
This PR is following Design 1 in RFC https://github.com/realm/realm-java/pull/4165/files

Many classes and methods don't have Javadoc comment for now. I'll add them later.


realm/realm-library/src/main/java/io/realm/DynamicRealmObject.java
```diff
@@ -360,6 +360,18 @@ public DynamicRealmObject getObject(String fieldName) {
     }
 
     /**
+     * Returns the {@link RealmList} of values being linked from the given field.
+     *
+     * @param fieldName the name of the field.
+     * @return the {@link RealmList} data for this field.
+     * @throws IllegalArgumentException if field name doesn't exist or it doesn't contain a list of values.
+     */
+    public <T> RealmList<T> getValueList(String fieldName, Class<T> valueClass) {
```
cmelchior: Doesn't seem needed? We already have `getList(fieldName)` ?

realm/realm-library/src/main/java/io/realm/DynamicRealmObject.java
```diff
@@ -761,6 +793,19 @@ public void setList(String fieldName, RealmList<DynamicRealmObject> list) {
...
     /**
+     * Sets the reference to a {@link RealmList} on the given field.
+     *
+     * @param fieldName field name.
+     * @param list list of references.
+     * @throws IllegalArgumentException if field name doesn't exist, it is not a list field, the type
+     * of the object represented by the DynamicRealmObject doesn't match or any element in the list belongs to a
+     * different Realm.
+     */
+    public void setValueList(String fieldName, RealmList<?> list) {
```
cmelchior: Same, we already have `setList()` ... since RealmList is changed to `RealmList<Object>` that should be fine to use that.

realm/realm-library/src/main/java/io/realm/OrderedRealmCollection.java
```diff
@@ -100,14 +100,15 @@
...
-public interface OrderedRealmCollection<E extends RealmModel> extends List<E>, RealmCollection<E> {
+public interface OrderedRealmCollection<E> extends List<E>, RealmCollection<E> {
 
     /**
      * Gets the first object from the collection.
      *
      * @return the first object.
      * @throws IndexOutOfBoundsException if the collection is empty.
      */
+    @Nullable
```
cmelchior: Okay, this is annoying, but I can see why it is needed.

realm/realm-library/src/main/java/io/realm/RealmFieldType.java
```diff
@@ -45,15 +48,29 @@
     DOUBLE(10),
     OBJECT(12),
     LIST(13),
-    LINKING_OBJECTS(14);
+    LINKING_OBJECTS(14),
+
+    @SuppressWarnings("PointlessArithmeticExpression")
+    INTEGER_LIST(LIST_OFFSET + 0),
```
cmelchior: We do really need all of these? If yes, you are missing `BYTE_LIST` and `BLOB_LIST`

realm/realm-library/src/main/java/io/realm/DynamicRealmObject.java
beeender: at least seems no need to be public?

realm/realm-library/src/main/java/io/realm/RealmList.java
```diff
@@ -170,14 +174,20 @@ private boolean isAttached() {
      * @throws IndexOutOfBoundsException if {@code location < 0 || location > size()}.
      */
     @Override
-    public void add(int location, E object) {
+    public void add(int location, @Nullable E object) {
```
beeender: IMO if the primitive list support null values, the `LinkView` should support it as well. Otherwise it feels so inconsistent.

realm/realm-library/src/main/java/io/realm/DynamicRealmObject.java
zaki50: We need to know which class is actually used as type parameter of `RealmList`.
When the list field is INTEGER_LIST in schema, `T` can be `Long`, `Integer`, `Short` and `Byte`.


```
DynamicRealmObject obj = ...;
RealmList<Short> list = obj.getList("shortList");

Short value = list.get(0); // the implementation of 'get(int)' must return `Short` object.
```

realm/realm-library/src/main/java/io/realm/DynamicRealmObject.java
zaki50: Yes.
I added this because I didn't find a way to avoid `getValueList()`.

If we can omit `getValueList()`, I'll remove this.

realm/realm-library/src/main/java/io/realm/RealmList.java
cmelchior: Very, and I agree that we should push for `LinkView` to support it as well, but otherwise the only alternative right now is a complete API split "RealmValueList vs RealmList".

realm/realm-library/src/main/java/io/realm/DynamicRealmObject.java
cmelchior: Ah yeah, good point 👍

zaki50:
I'll split this PR into smaller parts.

realm/realm-library/src/main/java/io/realm/RealmFieldType.java
zaki50: What's the `BYTE_LIST`? Isn't it the same with `BINARY(4)`?

And `BLOB_LIST ` means we need to support `RealmList<byte[]>`, right?

zaki50:
I've split Original commit into 4 PRs.

#5145: Part1: API changes (This PR)
#5150: Nullability changes
#5151: Solve compilation errors
#5152: Tests

realm/realm-library/src/main/java/io/realm/RealmFieldType.java
cmelchior: I believe `BINARY` is BLOB actually so it should probably be `BINARY_LIST`. I'm not sure but perhaps `byte` is store in `INTEGER` columns. You are right about `RealmList<byte[]>`. I'm not sure if core supports it though.

zaki50: I confirmed that core supported list of BLOB.
I've added `RealmFieldType.BINARY_LIST`.

realm/realm-library/src/main/java/io/realm/OrderedRealmCollection.java
```diff
@@ -100,7 +100,7 @@
  * As you can see, after deletion, the size and elements order of snapshot stay the same as before. But the element at
  * the position becomes invalid.
  */
-public interface OrderedRealmCollection<E extends RealmModel> extends List<E>, RealmCollection<E> {
+public interface OrderedRealmCollection<E> extends List<E>, RealmCollection<E> {
```
bmeike: As you know, I just hate doing this.  Moving these checks out of the type system and into the runtime (I *assume* we'll check at runtime!) just seems like a perversion of the typing system and a misunderstanding of Java.  I wish we could have discussed this approach as a design and considered other approaches before the coding started.  At this point, it seems too late to object.  I cannot shake the feeling, though, that we are headed for a type-less library.

Zhuinden: You will check it in the annotation processor, which is compilation time. Or at least I think that is the plan. 

You can't really avoid this if you ever want to introduce type adapters either, so I think this is a good decision overall.

PR #5150, opened 2017-08-24
Primitive List feature(Part2: Nullability changes)

zaki50:


realm/realm-library/src/main/java/io/realm/RealmList.java
```diff
@@ -380,7 +380,7 @@ public E remove(int location) {
      * @throws NullPointerException if {@code object} is {@code null}.
      */
     @Override
-    public boolean remove(Object object) {
+    public boolean remove(@Nullable Object object) {
```
cmelchior: Does this make sense? Can you remove the `null` element in a standard `ArrayList` ?

zaki50: Yes http://hg.openjdk.java.net/jdk7/jdk7/jdk/file/tip/src/share/classes/java/util/ArrayList.java#l473

cmelchior: Huh, things I learned today 👍

zaki50:
@cmelchior Added more `@Nullable` annotations.

zaki50:
@cmelchior fixed all nullability issues in #5151

PR #5151, opened 2017-08-24
Primitive List feature(Part3: Temporary fix of implementations and tests)

zaki50:


realm/realm-library/src/main/java/io/realm/DynamicRealmObject.java
```diff
@@ -513,9 +513,29 @@ private void setValue(String fieldName, Object value) {
...
-            setList(fieldName, list);
+            RealmList<?> list = (RealmList<?>) value;
+            if (list.className == null && list.clazz == null) {
+                // unmanaged RealmList
+                long columnIndex = proxyState.getRow$realm().getColumnIndex(fieldName);
+                final RealmFieldType columnType = proxyState.getRow$realm().getColumnType(columnIndex);
+                if (columnType == RealmFieldType.LIST) {
+                    //noinspection unchecked
+                    for (Object element : list) {
+                        if (!(element instanceof RealmModel)) {
```
beeender: why add this check? And why doesn't this check apply to line 535?

zaki50: The list is *unmanaged* `RealmList` and that means it may contain wrong elements.

On the other hands, line 535 is handling *managed* `RealmList`.
Managed `RealmList` never contain wrong elements in it.

Issue #5094, opened 2017-08-11
Update objectServerExample to use MutableRealmInteger


cmelchior:
With `MutableRealmInteger` merged, we should use that instead of the homebrew solution in `objectServerExample`. It would simplify it quite a lot, but also gives us a possibility to show how it can be used.

beeender:
#5158 is merged

PR #5158, opened 2017-08-24
Update to use MutableRealmIntegers

bmeike:
Convert to MutableRealmInteger and update the code where necessary.

Addresses #5094


examples/build.gradle
```diff
@@ -23,7 +23,7 @@ allprojects {
             maven { url 'https://jitpack.io' }
         }
         dependencies {
-            classpath 'com.android.tools.build:gradle:3.0.0-alpha4'
+            classpath 'com.android.tools.build:gradle:3.0.0-beta2'
```
beeender: update android gradle together with others in a separated PR

examples/objectServerExample/build.gradle
```diff
@@ -29,7 +29,7 @@ android {
     defaultConfig {
         applicationId 'io.realm.examples.objectserver'
         targetSdkVersion rootProject.sdkVersion
-        minSdkVersion 15
+        minSdkVersion 16
```
beeender: Why update this to 16?

examples/objectServerExample/README.md
```diff
@@ -1,14 +1,20 @@
 # Using this example
 
-This example shows a minimal example on how to connect to and use the
+This example is a minimal demonstration of how to connect to and use the
 Realm Object Server to synchronize changes between devices.
 
-The example will assume that the Object Server is running on the machine
-building the example and the IP address will automatically be injected
-into the build configuration.
+The example assumes that the Object Server is running on the machine
```
cmelchior: Nice rewrite of this section 👏

examples/objectServerExample/src/main/java/io/realm/examples/objectserver/CounterActivity.java
```diff
@@ -45,36 +47,35 @@
...
-    private SyncSession session;
-    private CRDTCounter counter;
-    private SyncUser user;
-    private AtomicBoolean downloadingChanges = new AtomicBoolean(false);
-    private AtomicBoolean uploadingChanges = new AtomicBoolean(false);
-    private ProgressListener downloadListener = new ProgressListener() {
+    private final ProgressListener downloadListener = new ProgressListener() {
         @Override
-        public void onChange(Progress progress) {
+        public void onChange(@Nonnull Progress progress) {
```
cmelchior: Isn't it non-null by default?

examples/objectServerExample/src/main/java/io/realm/examples/objectserver/CounterActivity.java
```diff
@@ -95,38 +96,34 @@ protected void onCreate(Bundle savedInstanceState) {
...
-
-            counter = realm.where(CRDTCounter.class).findFirstAsync();
-            counter.addChangeListener(new RealmChangeListener<CRDTCounter>() {
-                @Override
-                public void onChange(CRDTCounter counter) {
-                    if (counter.isValid()) {
-                        counterView.setText(String.format(Locale.US, "%d", counter.getCount()));
-                    } else {
-                        counterView.setText("-");
+        if (user == null) { return; }
```
cmelchior: I realise this is just the old behaviour in a different way, but if this ever happened users would just stare at a blank screen. In a real app you would probably throw them back to a login screen and log an error so you could fix it, in this example would it make sense to throw `IllegalState` here?

In any case, it is a minor point

examples/objectServerExample/src/main/java/io/realm/examples/objectserver/CounterActivity.java
```diff
@@ -194,18 +186,25 @@ private void updateProgressBar(boolean downloading, boolean uploading) {
 
     private void adjustCounter(final int adjustment) {
         // A synchronized Realm can get written to at any point in time, so doing synchronous writes on the UI
-        // thread is HIGHLY discouraged as it might block longer than intended. Only use async transactions.
+        // thread is HIGHLY discouraged as it might block longer than intended. Use only async transactions.
         realm.executeTransactionAsync(new Realm.Transaction() {
             @Override
-            public void execute(Realm realm) {
+            public void execute(@Nonnull Realm realm) {
```
cmelchior: Same, isn't it nonnull by default?

examples/objectServerExample/src/main/java/io/realm/examples/objectserver/model/CRDTCounter.java
```diff
@@ -13,38 +13,30 @@
...
- *
- * @see <href ref="https://en.wikipedia.org/wiki/Conflict-free_replicated_data_type">Conflict Free Replicated Data Structures</href>
+ * A named, conflict-free replicated data-type.
  */
 public class CRDTCounter extends RealmObject {
-
     @PrimaryKey
-    private long id;
-    private RealmList<CounterOperation> operations;
+    private String name;
```
cmelchior: Since each user has her own Realm, this is strictly not needed as each user Realm would only have 1 object. It would also simplify the query which would just be `where(CRDTCounter.class).findFirstAsync()`

examples/objectServerExample/src/main/java/io/realm/examples/objectserver/CounterActivity.java
```diff
@@ -95,38 +96,34 @@ protected void onCreate(Bundle savedInstanceState) {
...
-                    } else {
-                        counterView.setText("-");
+        if (user == null) { return; }
+
+        // Create a RealmConfiguration for our user
+        SyncConfiguration config = new SyncConfiguration.Builder(user, REALM_URL)
+                .initialData(new Realm.Transaction() {
+                    @Override
+                    public void execute(@Nonnull Realm realm) {
+                        realm.createObject(CRDTCounter.class, user.getIdentity());
```
cmelchior: I'm surprised this works `createObject(Class, Object)` should only work if a primary key is defined?

examples/build.gradle
zaki50: We already have https://github.com/realm/realm-java/pull/4899

examples/objectServerExample/src/main/java/io/realm/examples/objectserver/CounterActivity.java
zaki50: Default nullability of `java.io.realm.examples.objectserver` package is not specified.
I think it's better to set default nullability by adding `package-info.java`.

examples/objectServerExample/src/main/java/io/realm/examples/objectserver/CounterActivity.java
zaki50: @cmelchior `CRDTCounter` seems to have the primaryKey (`name`).

cmelchior: Ups, missed that above the edits

examples/build.gradle
cmelchior: Any reason it cannot be merged?

zaki50: @cmelchior the reason is written in the description of the PR.

zaki50: And beta3 is adding far more critical issue...

examples/objectServerExample/build.gradle
bmeike: because `android:fontFamily="sans-serif-light"` requires API level 16

examples/build.gradle
bmeike: If you'll notice, this is utterly orthogonal. CI builds it with no problem.

examples/objectServerExample/src/main/java/io/realm/examples/objectserver/CounterActivity.java
bmeike: Interesting huh.  No, it doesn't appear to be.  This is not is a package that we've NonNullByDefault-ed.  If this is example code, I think it makes sense to make it look like user code would look and I think it would have these annotations.  I'd be happy to do it by package, if you think that's more typical...

examples/objectServerExample/src/main/java/io/realm/examples/objectserver/CounterActivity.java
bmeike: No, it is a great point.  I felt pretty nervous about it too, but I didn't want to change too much in a single PR.  IMO, it should close the current activity and start the login activity...

examples/objectServerExample/src/main/java/io/realm/examples/objectserver/CounterActivity.java
bmeike: As I said above, I slightly disagree.  Only slightly, though, so I can be persuaded.  I think user code is much more likely to have per-method annotations than to do it with package-info.  ... if we do it with package info, we surely need a comment here, because, when a user copies and pastes the un-annotated version, they will get a warning and it may not be obvious how to fix it.

examples/objectServerExample/src/main/java/io/realm/examples/objectserver/model/CRDTCounter.java
bmeike: Is that right?  Users don't share Realms in this app?

cmelchior: I'm pretty sure they don't since you construct a new `SyncConfiguration` for each user.

examples/build.gradle
bmeike: Reverted...

examples/objectServerExample/src/main/java/io/realm/examples/objectserver/CounterActivity.java
bmeike: Fixed.

examples/objectServerExample/src/main/java/io/realm/examples/objectserver/model/CRDTCounter.java
bmeike: Yep.  That would do it, wouldn't it.

bmeike:
Tests are passing on my machine.  Restarting.

examples/objectServerExample/src/main/java/io/realm/examples/objectserver/CounterActivity.java
cmelchior: Strange, I guess there is something about these annotations I don't understand then. Because the `ProgressListener` interface is in a package covered by a package level annotation. It seems really odd if people implementing this interface in whatever package doesn't inherit that :(

examples/objectServerExample/src/main/java/io/realm/examples/objectserver/CounterActivity.java
cmelchior: If it really is true that package level annotations on interfaces are not inherited in anonymous implementations in another package, then we need to go over all our interfaces and manually add `@Nonnull`, because then I assume it will work without the user having to do it himself? Do you know @zaki50 ?

Issue #4997, opened 2017-07-19
[Sync] ClientReset, using the backup Realm file is not possible


nhachicha:
Following the concerns from this post https://forums.realm.io/t/how-can-i-test-my-apps-error-handling/393/7
it seems we don't expose a way to reuse the backup Realm file (after a ClientReset). The current API returns the full path name, making it impossible to build a new `SyncConfiguration` that targets the saved file, so the user can copy data from it, making the saved Realm nearly unusable.

Zhuinden:
That's interesting, the change log says that this is relevant to #4080

sipersso:
@nhachicha thank you for noticing my forum post. Right now, I have just made an additional copy of the Realm using Realm.writeToFile() before executing the reset, but I am still struggling to get the data over and I haven't found any examples on how to execute a client reset properly.

If there is anything I can do to help, let me know.

Zhuinden:
It's a bit odd though as I would think https://realm.io/docs/java/latest/api/io/realm/ClientResetRequiredError.html should provide the necessary data.

``` java
error.executeClientReset();

// Validate that files have been moved
assertFalse(error.getOriginalFile().exists());
assertTrue(error.getBackupFile().exists());
```

sipersso:
@Zhuinden agreed! And I think a little more clarification on how to actually handle the ClientResets in a proper way would be great.

sipersso:
@nhachicha any info on why this is blocked? Is there any more info on how to handle ClientResets properly?

nhachicha:
The implementation should happen at the [Object Store](https://github.com/realm/realm-object-store) level. The currently exported file has a history that needs a `SyncConfiguration` in order to be opened. We're experimenting a workaround at the ObjectStore level in order to allow the user to open the Realm without specifying SyncConfig. 
stay tuned!

cmelchior:
Implemented in #4759. Closing

PR #5159, opened 2017-08-25
Client reset fixes #4759

### LINKED ISSUE #4759: Sync - Verify certificate using TrustManager
Sync is working on a [PR](https://github.com/realm/realm-sync/pull/1312) that will perform a series of callback to validate the certificate chain, in order to complete the SSL Handshake using the trusted root CA configured with Android.

We need to handle the callback from the Java binding, and use the default `TrustManager` in order to validate the certificate.

The C++ need to pass as an argument the certificate that we should transform into a `X509Certificate` format, so it can be usable with the `JSSE` API

nhachicha:
Exposing a `SyncConfiguration` that allows a user to open the backup Realm after the client reset (#4997).


realm/realm-library/src/androidTestObjectServer/java/io/realm/SessionTests.java
```diff
@@ -125,12 +128,17 @@ public void onError(SyncSession session, ObjectServerError error) {
...
                         final ClientResetRequiredError handler = (ClientResetRequiredError) error;
-                        String filePathFromError = handler.getOriginalFile().getAbsolutePath();
-                        String filePathFromConfig = session.getConfiguration().getPath();
-                        assertEquals(filePathFromError, filePathFromConfig);
-                        assertFalse(handler.getBackupFile().exists());
-                        assertTrue(handler.getOriginalFile().exists());
-                        looperThread.testComplete();
+                        final String filePathFromError = handler.getOriginalFile().getAbsolutePath();
+                        final String filePathFromConfig = session.getConfiguration().getPath();
+                        looperThread.postRunnable(new Runnable() {
```
cmelchior: Not sure I understand why you move this check to the looper thread?

realm/realm-library/src/androidTestObjectServer/java/io/realm/SessionTests.java
```diff
@@ -168,10 +176,15 @@ public void onError(SyncSession session, ObjectServerError error) {
                         looperThread.closeTestRealms();
                         handler.executeClientReset();
 
-                        // Validate that files have been moved
-                        assertFalse(handler.getOriginalFile().exists());
-                        assertTrue(handler.getBackupFile().exists());
-                        looperThread.testComplete();
+                        looperThread.postRunnable(new Runnable() {
```
cmelchior: Same

realm/realm-library/src/objectServer/java/io/realm/SyncConfiguration.java
```diff
@@ -286,6 +304,13 @@ boolean isSyncConfiguration() {
     }
 
     /**
+     * @return {@code true} if this configuration is intended to open a synced Realm offline (example for client reset), {@code false} otherwise.
+     */
+    public boolean isOpenSyncedRealmOffline() {
```
cmelchior: Since this is on a SyncConfiguration calling it `SyncedRealm` seems a bit like overkill, what about `isRealmOffline`, `isRealmForcedOffline` or perhaps just `isForcedOffline` ?

nhachicha:
Ok will add an option to open the file after app restart

realm/realm-library/src/androidTestObjectServer/java/io/realm/SessionTests.java
```diff
@@ -130,12 +131,12 @@ public void onError(SyncSession session, ObjectServerError error) {
...
-                        System.out.println(">>>>>>>>>>>>>>>>> filePathFromError  : " + filePathFromError);
-                        System.out.println(">>>>>>>>>>>>>>>>> filePathFromConfig : " + filePathFromConfig);
-
-                        System.out.println(">>>>>>>>>>>>>>>>> handler.getBackupFile().exists()   : " + handler.getBackupFile().exists());
-                        System.out.println(">>>>>>>>>>>>>>>>> handler.getOriginalFile().exists() : " + handler.getOriginalFile().exists());
+                        Log.e("H4XOR", ">>>>>>>>>>>>>>>>> filePathFromError  : " + filePathFromError);
+                        Log.e("H4XOR", ">>>>>>>>>>>>>>>>> filePathFromConfig : " + filePathFromConfig);
 
+                        Log.e("H4XOR", ">>>>>>>>>>>>>>>>> handler.getBackupFile().exists()   : " + handler.getBackupFile().exists());
+                        Log.e("H4XOR", ">>>>>>>>>>>>>>>>> handler.getOriginalFile().exists() : " + handler.getOriginalFile().exists());
```
Zhuinden: Um... This is getting trickier by the day, what's up?

realm/realm-library/src/androidTestObjectServer/java/io/realm/SessionTests.java
```diff
@@ -182,4 +186,147 @@ public void onError(SyncSession session, ObjectServerError error) {
...
         SyncManager.simulateClientReset(SyncManager.getSession(config));
     }
+
+    // Check that we can use the backup SyncConfiguration to open the Realm.
+    @Test
+    @RunTestInLooperThread
+    public void errorHandler_useBackupSyncConfigurationForClientReset() {
+        SyncUser user = createTestUser();
+        String url = "realm://objectserver.realm.io/default";
+        final SyncConfiguration config = configFactory.createSyncConfigurationBuilder(user , url)
```
dalinaum: `(user, url)

realm/realm-library/src/androidTestObjectServer/java/io/realm/SessionTests.java
nhachicha: I think CI acting up, it's fixed in next commits

zaki50:
Can users use offline `SyncConfiguration` at initial launch and then reopen the same Realm with online `SyncConfiguration later?

We should have test case for the case above.

Chane itsekf is :+1:

nhachicha:
@zaki50 using `SyncConfiguration#forOffline`  will allow the user to open the backup Realm in offline mode, re-opening the Realm again in online mode will trigger a Client reset again (since the sync history will diverge).

nhachicha:
@zaki50 it's not possible giving the current API to re-open the backup file as online using the intial `SyncConfiguration`, since this one set the full path to the Realm using the `uid`. 

Using `directory` and `name` option with `SyncConfiguration` to specify the location of the backup file will not work either since this will still look for a file under `uid`

ex:
`backup file`: `/data/data/io.realm.test/files/realm-object-server/io.realm.object-server-recovered-realms/recovered_realm-20170828-164217-rzYYx8HB`

if you use SyncConfiguration with the following option
```Java
                        String name = handler.getBackupFile().getName();
                        String parent = handler.getBackupFile().getParent();
                        SyncConfiguration syncConfiguration = new SyncConfiguration
                                .Builder(user, Constants.AUTH_URL)
                                .name(name)
                                .directory(new File(parent))
                                .build();

```
 you'll end up trying to open 

`/data/data/io.realm.test/files/realm-object-server/io.realm.object-server-recovered-realms/b26f97533e0b191dd075aca8b6a9cab2/auth:recovered_realm-20170828-164217-rzYYx8HB`

beeender:
since we are going to make a minor release later, so it would be easier to make this PR based on the master branch. th `OsConfigWrapper` has been removed on the master, there would be some conflicts.

realm/realm-library/src/androidTestObjectServer/java/io/realm/SessionTests.java
```diff
@@ -334,4 +335,30 @@ public void onError(SyncSession session, ObjectServerError error) {
...
         SyncManager.simulateClientReset(SyncManager.getSession(config));
     }
+
+    @Test
+    @UiThreadTest
+    public void uploadAllLocalChanges_throwsOnUiThread() throws InterruptedException {
+        SyncUser user = createTestUser();
+        Realm realm = Realm.getInstance(configuration);
+        try {
+            SyncManager.getSession(configuration).uploadAllLocalChanges();
```
Zhuinden: Assertion seems to be missing?

realm/realm-library/src/androidTestObjectServer/java/io/realm/SessionTests.java
```diff
@@ -334,4 +335,30 @@ public void onError(SyncSession session, ObjectServerError error) {
...
+        }
+    }
+
+    @Test
+    @UiThreadTest
+    public void downloadAllServerChanges_throwsOnUiThread() throws InterruptedException {
+        SyncUser user = createTestUser();
+        Realm realm = Realm.getInstance(configuration);
+        try {
+            SyncManager.getSession(configuration).downloadAllServerChanges();
```
Zhuinden: Assertion seems to be missing?

realm/realm-library/src/androidTestObjectServer/java/io/realm/SessionTests.java
nhachicha: which assertion? this test didn't change, but now you mention it I've noticed that we don't need the `createTestUser` in this test (same for `downloadAllServerChanges_throwsOnUiThread`)

Zhuinden: I mean that you don't seem to `fail()` the test if the exception is not thrown.

realm/realm-library/src/main/cpp/io_realm_internal_OsRealmConfig.cpp
```diff
@@ -327,4 +327,13 @@ JNIEXPORT void JNICALL Java_io_realm_internal_OsRealmConfig_nativeSetSyncConfigS
     CATCH_STD()
 }
 
+JNIEXPORT void JNICALL Java_io_realm_internal_OsRealmConfig_nativeSetForceSyncHistory(JNIEnv*, jclass,
+                                                                                      jlong native_ptr)
+{
+    TR_ENTER_PTR(native_ptr);
+
+    auto& config = *reinterpret_cast<Realm::Config*>(native_ptr);
+    config.force_sync_history = true;
```
beeender:
```
        // FIXME: Realm Java manages sync at the Java level, so it needs to create Realms using the sync history
        //        format.
        bool force_sync_history = false;
```

I thought this should be removed? It was added because of we were handling sync in java at the beginning.

realm/realm-library/src/objectServer/java/io/realm/SyncConfiguration.java
```diff
@@ -142,6 +155,39 @@ private SyncConfiguration(File directory,
...
+
+    /**
+     * Returns a SyncConfiguration appropriate to open a synced Realm offline.
+     * This is useful when trying to open a backup Realm (after a Client Reset)
+     *
+     * @param canonicalPath the absolute path to the Realm file defined by this configuration.
+     * @param modules if specified it will restricts Realm schema to the provided module.
+     * @return SyncConfiguration that can be used offline
+     */
+    public static SyncConfiguration forOffline(String canonicalPath, Object... modules) {
```
beeender: wouldn't it return a `RealmConfiguration` instead since it cannot be used for sync?

realm/realm-library/src/main/cpp/io_realm_internal_OsRealmConfig.cpp
beeender:
```
        // FIXME: Realm Java manages sync at the Java level, so it needs to create Realms using the sync history
        //        format.
        bool force_sync_history = false;


            bool server_synchronization_mode = bool(config.sync_config) || config.force_sync_history;
            if (server_synchronization_mode) {
#if REALM_ENABLE_SYNC
                history = realm::sync::make_client_history(config.path);
#else
                REALM_TERMINATE("Realm was not built with sync enabled");
#endif
            }

```

This was added because of java was handling sync without OS at the beginning, but it seems this will be removed later. What is conclusion of supporting offline sync realm in Object Store? are we going to keep this `force_sync_history` in the future?

realm/realm-library/src/androidTestObjectServer/java/io/realm/SessionTests.java
nhachicha: good catch 👍🏼 I thought we were using `@Test(expected = IllegalStateException.class)`

realm/realm-library/src/objectServer/java/io/realm/SyncConfiguration.java
nhachicha: Good point, I think Cocoa POC implementation used a [RLMRealmConfiguration](https://github.com/realm/realm-cocoa/pull/5259/files#diff-c64398ac8b77c37143815b54e83e9427R1121)

realm/realm-library/src/main/cpp/io_realm_internal_OsRealmConfig.cpp
nhachicha: AFAIK there's no plan to remove `force_sync_history` currently https://realmio.slack.com/archives/C0340J37W/p1503545777000115?thread_ts=1503491762.000111&cid=C0340J37W 
plus, [.NET](https://github.com/realm/realm-dotnet/issues/1536) may also use this option

beeender: 👍

realm/realm-library/src/main/java/io/realm/RealmConfiguration.java
```diff
@@ -132,6 +139,39 @@ protected RealmConfiguration(@Nullable File realmDirectory,
         this.initialDataTransaction = initialDataTransaction;
         this.readOnly = readOnly;
         this.compactOnLaunch = compactOnLaunch;
+        this.forceSyncHistory = forceSyncHistory;
+    }
+
+    /**
+     * Returns a SyncConfiguration appropriate to open a synced Realm offline.
```
beeender: `SyncConfiguration` -> `RealmConfiguration`

realm/realm-library/src/main/java/io/realm/RealmConfiguration.java
```diff
@@ -132,6 +139,39 @@ protected RealmConfiguration(@Nullable File realmDirectory,
...
+     * This is useful when trying to open a backup Realm (after a Client Reset)
+     *
+     * @param canonicalPath the absolute path to the Realm file defined by this configuration.
+     * @param modules if specified it will restricts Realm schema to the provided module.
+     * @return SyncConfiguration that can be used offline
```
beeender: `SyncConfiguration` -> `RealmConfiguration`

realm/realm-library/src/main/java/io/realm/RealmConfiguration.java
```diff
@@ -132,6 +139,39 @@ protected RealmConfiguration(@Nullable File realmDirectory,
...
+     */
+    public static RealmConfiguration forOffline(String canonicalPath, Object... modules) {
```
beeender: wouldn't it be enough to have this in `SyncConfiguration`?
Another thing is this doesn't cover all cases.
eg.: The sync Realm is encrypted, the backup realm file is still encrypted. So we still need to give user a chance to add more params to the config.
Would it make sense to return a `Builder` instead?

realm/realm-library/src/main/java/io/realm/RealmConfiguration.java
```diff
@@ -266,6 +306,13 @@ public boolean isReadOnly() {
         return readOnly;
     }
 
+    /**
+     * @return {@code true} if this configuration is intended to open a synced Realm offline (example for client reset), {@code false} otherwise.
+     */
+    public boolean isForceSyncHistory() {
```
beeender: need to be public?

nhachicha: yes since it will be accessed from `OsRealmConfig` (internal)

realm/realm-library/src/main/java/io/realm/RealmConfiguration.java
nhachicha: I think it's consistent to build the offline conf from `RealmConfiguration` otherwise it will be confusing to have an offline SyncConf.

> The sync Realm is encrypted, the backup realm file is still encrypted.
Will add a test to cover this use case

beeender: cocoa has this api in sync config extension https://github.com/realm/realm-cocoa/pull/5259/files#diff-89bce12b01636fffc4e7d1848568b610R72

`RealmConfiguration.forOffLine()` sounds strange to me, but i don't have a better name + i am idiot of naming something ... so if @realm/java is fine with this, i am fine.

realm/realm-library/src/main/java/io/realm/RealmConfiguration.java
cmelchior: Could we call it something else that would make a little more sense to end users in that case? like isForcedOfflineCould we call it something else that would make a little more sense to end users in that case? like isForcedOffline

realm/realm-library/src/main/java/io/realm/RealmConfiguration.java
Zhuinden: It would be odd to see Sync things if Sync is disabled.

realm/realm-library/src/main/java/io/realm/RealmConfiguration.java
Zhuinden: I'm honestly understanding client reset less and less :D but I agree I don't really understand what this does. 

Is it like, `shouldOpenOffline`?

nhachicha: I was trying to avoid anything _offline_ on a RealmConfiguration, since the concept applies to Sync only ...

cmelchior: Good point, but then `Sync` also only applies to sync only :)

I don't know, it will not be ideal no matter what we do, so I guess this is fine.

realm/realm-library/src/main/java/io/realm/RealmConfiguration.java
nhachicha: @beeender Unfortunately we can't use the builder as it is now since we don't have the option like in Cocoa for instance where you can pass directly the location of the Realm (`fileURL`) we can combine `directory` and `name` option on the builder, whereas the ObjectStore return the full path to the backup file

I will revert the option `forOffline` on `SyncConfiguration`.

realm/realm-library/src/main/java/io/realm/RealmConfiguration.java
nhachicha: What about 
```Java
    /**
     * @return {@code true} if this configuration is intended to open the client reset backup Realm
     * {@code false} otherwise.
     *
     * @see <a href="https://realm.io/docs/java/latest/api/io/realm/ClientResetRequiredError.html">ClientResetRequiredError</a>
     */
    public boolean isClientResetBackupRealm() {
    }

```
this is the only use case where we explicitly set the `force_sync_history` to `true` in order to open the Realm without a `SyncConfig`. It will make more sense IMO to the end user, WDYT?

realm/realm-library/src/objectServer/java/io/realm/SyncConfiguration.java
```diff
@@ -144,6 +145,39 @@ private SyncConfiguration(File directory,
         this.waitForInitialData = waitForInitialData;
     }
 
+    /**
+     * Returns a SyncConfiguration appropriate to open a synced Realm offline.
+     * This is useful when trying to open a backup Realm (after a Client Reset)
```
cmelchior: Missing `.` at the end

realm/realm-library/src/objectServer/java/io/realm/SyncConfiguration.java
```diff
@@ -144,6 +145,39 @@ private SyncConfiguration(File directory,
...
+     *
+     * @param canonicalPath the absolute path to the Realm file defined by this configuration.
+     * @param encryptionKey the key used to encrypt/decrypt the Realm file.
+     * @param modules if specified it will restricts Realm schema to the provided module.
+     * @return SyncConfiguration that can be used offline
+     */
+    public static RealmConfiguration forOffline(String canonicalPath, @Nullable byte[] encryptionKey, Object... modules) {
```
cmelchior: Uggh, the signature for this method :( ... I can see why it is needed, but still.

Any reason to specify modules? At worst, we will open the Realm with a superset of the classes available? This will create those in the Realm file, but is that a big problem?

Should `modules` be nullable? Would also require a null check in the code

I know this is really a power feature, but should we also have a simple variant that is just `forOffline(String path)` ?

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/ProgressListenerTests.java
```diff
@@ -190,7 +190,6 @@ public void onChange(Progress progress) {
         userRealm.close();
         userWithData.logout();
         adminUser.logout();
-        worker.join();
```
cmelchior: It concerns me quite a lot that this is a problem. This should only ever be a problem if the progress listeners do not work and the `case 2` is not triggered. It shouldn't be fixed in this PR, but please add a FIXME here instead so we remember to follow up on it.

realm/realm-library/src/objectServer/java/io/realm/SyncConfiguration.java
```diff
@@ -145,6 +145,39 @@ private SyncConfiguration(File directory,
...
+                validatedModules.add(Realm.getDefaultModule());
+            }
+        }
+
+        RealmProxyMediator schemaMediator = createSchemaMediator(validatedModules, Collections.<Class<? extends RealmModel>>emptySet());
+        return forOffline(canonicalPath, encryptionKey, schemaMediator);
+    }
+
+    static RealmConfiguration forOffline(String canonicalPath, byte[] encryptionKey, RealmProxyMediator schemaMediator) {
+        return new RealmConfiguration(null,null, canonicalPath,null, encryptionKey, 0,null, false, OsRealmConfig.Durability.FULL, schemaMediator, null, null, false, null, true);
```
beeender: should we open it with read-only mode?

cmelchior: Would probably be a good idea. I don't see any use cases where writing to it would make sense

cmelchior: It would also mean we could remove the modules parameter

nhachicha: As discussed this morning we should keep the module option (since not related to the Durability option), we will not restrict it to read only as well, the write obviously will not be synchronised, it's a local Realm after all, the user may which to update some state locally (example delete all rows that have been re-integrated into the new Synced Realm ...)

nhachicha:
⚠️ To be merged after https://github.com/realm/realm-java/pull/5249 ⚠️

beeender:
@nhachicha I think we can enable this test in this PR? https://github.com/realm/realm-java/pull/4693/files#diff-646188da3618e09d8aa093e28fed4158R308

PR #5161, opened 2017-08-25
@NotNull as alias for @Required

cmelchior:
This PR adds support for `org.jetbrains.annotations.NotNull` as an alias for `@Required`. This means that Realm now understands the Kotlin type-system in the Realm schema.

Previously we reported an error if you did `@Required RealmList<MyType> list = new RealmList<>();`. It is a bit unclear why, since a RealmList is always non-null on managed objects. Since it made the interop with Kotlin kinda strange, I made a change so you can now do `@Required RealmList<MyType> list = new RealmLis<>()` as well as `val list : RealmList<MyType> = RealmList();`

After this change it is no longer possible to do `val person : Person` either, you have to do `val person : Person?`. This reflects the constraints of Realm, but if people are able to maintain the variant themselves (hint: not possible when using sync), then they can use a custom non-null getter.


realm/realm-annotations-processor/src/main/java/io/realm/processor/ClassMetaData.java
```diff
@@ -441,13 +459,13 @@ private boolean categorizeIndexField(Element element, VariableElement variableEl
     private void categorizeRequiredField(Element element, VariableElement variableElement) {
         if (Utils.isPrimitiveType(variableElement)) {
             Utils.error(String.format(Locale.US,
-                    "@Required annotation is unnecessary for primitive field \"%s\".", element));
+                    "@Required and @NotNull annotation is unnecessary for primitive field \"%s\".", element));
```
nhachicha: @Required or @NotNull

realm/realm-annotations-processor/src/main/java/io/realm/processor/ClassMetaData.java
```diff
@@ -408,6 +409,23 @@ private boolean categorizeField(Element element) {
...
+    private boolean isRequiredField(VariableElement field) {
+        if (field.getAnnotation(Required.class) != null) {
+            return true;
+        }
+
+        // Kotlin uses the `org.jetbrains.annotations.NotNull` annotation to mark non-null fields.
+        // In order to fully support the Kotlin type system we interpret `@NotNull` as an alias
+        // for `@Required`
+        for (AnnotationMirror annotation : field.getAnnotationMirrors()) {
+            if (annotation.getAnnotationType().toString().equals("org.jetbrains.annotations.NotNull")) {
```
nhachicha: why not use `field.getAnnotation(Required.class)` ? you don't want to pull the dependency?

realm/realm-annotations-processor/src/main/java/io/realm/processor/ClassMetaData.java
```diff
@@ -441,13 +459,13 @@ private boolean categorizeIndexField(Element element, VariableElement variableEl
     private void categorizeRequiredField(Element element, VariableElement variableElement) {
         if (Utils.isPrimitiveType(variableElement)) {
             Utils.error(String.format(Locale.US,
-                    "@Required annotation is unnecessary for primitive field \"%s\".", element));
+                    "@Required and @NotNull annotation is unnecessary for primitive field \"%s\".", element));
             return;
         }
 
-        if (Utils.isRealmList(variableElement) || Utils.isRealmModel(variableElement)) {
+        if (Utils.isRealmModel(variableElement)) {
```
nhachicha: why removing the `isRealmList ` check?

cmelchior: Because it conflicted with Kotlin mostly. RealmList is always non-null for managed object. I don't know why we didn't allow this to be set. At worst it would be redundant information.

realm/realm-annotations-processor/src/main/java/io/realm/processor/ClassMetaData.java
cmelchior: You mean `@NotNull`, but yes.

Issue #5163, opened 2017-08-28
Can't cancel a non-existing write transaction


DKKovalev:
Hello, i'm having an issue. I needed to add migration to my project and after that i caught some crashes with the following stacktrace:

```
java.lang.RuntimeException: Unable to create application com.project.app.App: java.lang.IllegalStateException: Can't cancel a non-existing write transaction in /home/cc/repo/realm/release/realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp line 272
	at android.app.ActivityThread.handleBindApplication(ActivityThread.java:6057)
	at android.app.ActivityThread.access$1700(ActivityThread.java:218)
	at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1805)
	at android.os.Handler.dispatchMessage(Handler.java:102)
	at android.os.Looper.loop(Looper.java:145)
	at android.app.ActivityThread.main(ActivityThread.java:7007)
	at java.lang.reflect.Method.invoke(Native Method)
	at java.lang.reflect.Method.invoke(Method.java:372)
	at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:1404)
	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1199)
Caused by: java.lang.IllegalStateException: Can't cancel a non-existing write transaction in /home/cc/repo/realm/release/realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp line 272
	at io.realm.internal.SharedRealm.nativeCancelTransaction(Native Method)
	at io.realm.internal.SharedRealm.cancelTransaction(SharedRealm.java:267)
	at io.realm.BaseRealm.cancelTransaction(BaseRealm.java:372)
	at io.realm.DynamicRealm.cancelTransaction(DynamicRealm.java:48)
	at io.realm.BaseRealm$4.onResult(BaseRealm.java:661)
	at io.realm.RealmCache.doInvokeWithGlobalRefCount(RealmCache.java:489)
	at io.realm.RealmCache.invokeWithGlobalRefCount(RealmCache.java:484)
	at io.realm.BaseRealm.migrateRealm(BaseRealm.java:634)
	at io.realm.Realm.migrateRealm(Realm.java:1692)
	at io.realm.Realm.createInstance(Realm.java:355)
	at io.realm.RealmCache.doCreateRealmOrGetFromCache(RealmCache.java:345)
[...]

PR #5164, opened 2017-08-28
Fix a bug that BaseRealm.migrateRealm cancels a transaction even if it is not in a transaction.

dalinaum:
Address #5163


dalinaum:
Since 3.6.0 has a new implementation, I will close this PR.

zaki50:
I guess the exception is thrown from `realm = DynamicRealm.createInstance(configuration);`.
I think that moving that line out of try {} catch(){} is better.

PR #5168, opened 2017-08-29
Primitive List feature(Part5: Element type check in annotation processor)

zaki50:
Implemented a type check for `RealmList`'s type variable.

If we have a invalid field like `public RealmList<java.math.BigInteger> integers;`, we will get:

```
Error: Element type of RealmList must be a class implementing 'RealmModel' or one of the 'java.lang.String', 'byte[]', 'java.lang.Boolean', 'java.lang.Long', 'java.lang.Integer', 'java.lang.Short', 'java.lang.Byte', 'java.lang.Double', 'java.lang.Float', 'java.util.Date'.
    public RealmList<java.math.BigInteger> integers;
                                           ^
```


realm/realm-annotations-processor/src/main/java/io/realm/processor/ClassMetaData.java
```diff
@@ -301,6 +304,68 @@ private boolean checkListTypes() {
...
+                        "Only concrete Realm classes are allowed in RealmLists. "
+                                + "Neither interfaces nor abstract classes are allowed.",
+                        field);
+                return false;
+            }
+        }
+
+        // check if the actual value class is acceptable
+        if (!validListValueTypes.contains(elementTypeMirror)
+                && !Utils.isRealmModel(elementTypeMirror)) {
```
cmelchior: Couldn't this fit on the line above?

realm/realm-annotations-processor/src/main/java/io/realm/processor/ClassMetaData.java
```diff
@@ -301,6 +304,68 @@ private boolean checkListTypes() {
...
+            final StringBuilder messageBuilder = new StringBuilder(
+                    "Element type of RealmList must be a class implementing 'RealmModel' or one of the ");
+            for (TypeMirror type : validListValueTypes) {
+                messageBuilder.append('\'').append(type.toString()).append("', ");
+            }
+            messageBuilder.setLength(messageBuilder.length() - ", ".length());
```
cmelchior: Instead of having `", "` twice I would move it into a `String seperator = ", ";` or something like that

realm/realm-annotations-processor/src/main/java/io/realm/processor/ClassMetaData.java
```diff
@@ -301,6 +304,67 @@ private boolean checkListTypes() {
...
+            messageBuilder.append('.');
+            Utils.error(messageBuilder.toString(), field);
+            return false;
+        }
+
+        return true;
+    }
+
+    private boolean checkRealmResultsType(VariableElement field) {
```
cmelchior: Perhaps add a comment here about `RealmResults` currently being used for backlinks and this is why primitive types are not allowed here.

zaki50:
@cmelchior all comments are addressed.

PR #5173, opened 2017-08-29
Fix possible leaks with java callback in config

beeender:
Holding strong java refs in the lambda function in the Object Store
config will cause circular refs with SharedRealm/Callbacks/Config. That
will leak some objects.

### DIFF
--- a/realm/realm-library/src/main/cpp/io_realm_internal_OsRealmConfig.cpp
+++ b/realm/realm-library/src/main/cpp/io_realm_internal_OsRealmConfig.cpp
@@ -26,7 +26,7 @@
 #include "util.hpp"
 #include "jni_util/java_method.hpp"
 #include "jni_util/java_class.hpp"
-#include "jni_util/java_global_ref.hpp"
+#include "jni_util/java_global_weak_ref.hpp"
 #include "jni_util/jni_utils.hpp"
 #include "jni_util/java_exception_thrower.hpp"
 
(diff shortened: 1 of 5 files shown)


beeender:
After more checking, this might not be leaking. still checking now.

realm/realm-library/src/main/cpp/io_realm_internal_OsRealmConfig.cpp
```diff
@@ -132,16 +132,26 @@ JNIEXPORT void JNICALL Java_io_realm_internal_OsRealmConfig_nativeSetSchemaConfi
             static JavaMethod run_migration_callback_method(
                 env, get_shared_realm_class(env), "runMigrationCallback",
                 "(JLio/realm/internal/OsRealmConfig;Lio/realm/internal/SharedRealm$MigrationCallback;J)V", true);
-            JavaGlobalRef j_config_ref(env, j_config, true);
-            JavaGlobalRef j_migration_callback_ref(env, j_migration_callback, true);
-            config.migration_function = [j_config_ref, j_migration_callback_ref](SharedRealm old_realm,
+            // weak ref to avoid leaks caused by circular refs.
```
cmelchior: What kind of circular refs? configs cannot reference each other?

beeender: @cmelchior eg.:
you have a `CompactOnLaunch` as below:

```java
class MyCompactOnLaunch implements CompactOnLaunch {
    OsRealmConfig config;
}
```

Then this `compactOnLaunch` instance has been passed to the Object Store in the function block as a global ref.

```c++
struct Config {
//...
        ShouldCompactOnLaunchFunction should_compact_on_launch_function;
//...
}
```

The native `Config` instance won't be freed until the `OsRealmConfig` gets GCed. But since the `should_compact_on_launch_function` holding a global ref to the java `MyCompactOnLaunch` instance, it will never be GCed since java GC cannot detect if the global ref is still necessary in native code.

The real case could be more complicated than this, I think we should always use the weak ref in c++ lambda captures (unless there is a real need, but again, that makes things very complex.)

cmelchior: Good point 👍

PR #5180, opened 2017-09-01
Optimize opening Realm

### LINKED ISSUE #5036: Avoid to call compactOnLaunch if there are Realm instances opened.
This has been fixed in Object Store by https://github.com/realm/realm-object-store/pull/496

However, it doesn't fix the java side problem since we are using dynamic schema mode of the `SharedRealm`.

Revisit this issue after the OS schema integration work finished.

beeender:
- Open realm for checking legacy PK table issue only if 1) the realm
  file doesn't exist, 2) it is not a sync realm config.
- Open realm for downloading initial data only if 1) it is sync realm
  config, 2) the realm file doesn't exist.
- Modify the compactOnLaunch_multipleThread to reflect the fix in
  https://github.com/realm/realm-object-store/pull/496
  Close #5036
- Optimize the compactOnLaunch test case. Populate 300000 objects with
  debug version core will take 3+ minutes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,8 @@ and `SyncUser#retrieveInfoForUserAsync` which returns a `SyncUserInfo` with mode
 
 ### Bug Fixes
 
+* The `compactOnLaunch` callback is no longer invoked if the Realm at that path is already open on other threads.
+
 ### Internal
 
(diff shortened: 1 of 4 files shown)


beeender:
CI failed because of docker clean up.

PR #5181, opened 2017-09-01
Unclosed realm & SyncManager::reset_for_testing

beeender:
See https://github.com/realm/realm-object-store/pull/532
When assertion enabled, those will fail earlier. Without the assertion,
all different kind of native crash will happen.

### DIFF
--- a/realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/SyncSessionTests.java
+++ b/realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/SyncSessionTests.java
@@ -246,6 +246,7 @@ public class SyncSessionTests extends BaseIntegrationTest {
         });
 
         TestHelper.awaitOrFail(testCompleted, 60);
+        realm.close();
     }
 
     // A Realm that was opened before a user logged out should be able to resume uploading if the user logs back in.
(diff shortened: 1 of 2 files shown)


Zhuinden:
https://github.com/realm/realm-java/blob/17cad7c0b46902662aa099697c02bd81ba9e53ed/realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/SyncSessionTests.java#L389

wait what, why does this need `refresh()`?

-------

btw using `try(Realm realm = ....)` would handle the close in most of these cases.

cmelchior:
Yup, but try-with-resources requires min API 19 😞

Zhuinden:
If you are using build tools 3.0.0 then it shouldn't, though. AS tends to lie about this even when you use Retrolambda, they'll probably fix that in AS 3 too.

cmelchior:
Uhh, I completely missed that announcement ... Nice 💃

Zhuinden:
@cmelchior ya it's written down [here](https://developer.android.com/studio/write/java8-support.html#supported_features)

> In addition to the Java 8 language features and APIs above, Android Studio 3.0 Preview 1 and later extends support for try-with-resources to all Android API levels.

I'm mostly surprised that they managed to get `default interfaces` to work.

beeender:
I am going to just merge this. The CI fails because of docker clean up.

PR #5185, opened 2017-09-01
Fix release script

beeender:
- sed for Mac and Linux.
- Update submodule when release from master branch.


beeender:
I stopped CI since there is no point to verify this.

Issue #5195, opened 2017-09-02
Realm 3.7.0: permissionManager.makeOffer: First start error: If you see this error, please send this log to help@realm.io


lmdic:
Gradle 4.1, Android 4.3.

I want to get token by Realm 3.7.0. (by using class **PermissionManager**) and send token to all Android clients.

My Project's buil.gradle:

```
`buildscript {
    ext.kotlin_version = '1.0.5-2'
    repositories {
        jcenter()
    }
    dependencies {
        classpath 'com.android.tools.build:gradle:2.3.3'
        classpath "io.realm:realm-gradle-plugin:3.7.0"
        classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
    }
}

allprojects {
    repositories {
        jcenter()
    }
}
```

My app's build.gradle:

```
apply plugin: 'com.android.application'
apply plugin: 'kotlin-android'
apply plugin: 'realm-android'

realm {
    syncEnabled = true;
}

android {
    compileSdkVersion 26
    buildToolsVersion "26.0.1"

    defaultConfig {
        applicationId "com.myproject"
        minSdkVersion 15
        targetSdkVersion 26
        versionCode 1
        versionName "1.0"
        testInstrumentationRunner "android.support.test.runner.AndroidJUnitRunner"
    }

    buildTypes {
      release {
            minifyEnabled false
            proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
      }
      debug {
          
        }
    }
    configurations.all {
        resolutionStrategy.force 'com.google.code.findbugs:jsr305:1.3.9'
    }
}


dependencies {
    compile fileTree(dir: 'libs', include: ['*.jar'])
    androidTestCompile('com.android.support.test.espresso:espresso-core:2.2.2', {
        exclude group: 'com.android.support', module: 'support-annotations'
    })
    compile 'com.android.support:appcompat-v7:26.+'
    compile 'com.android.support.constraint:constraint-layout:1.0.2'
    compile 'com.google.code.gson:gson:2.8.1'
    compile "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
    compile 'org.nanohttpd:nanohttpd:2.2.0'

    testCompile 'junit:junit:4.12'
}

```

My Android code:`

```
import io.realm.ObjectServerError;
import io.realm.PermissionManager;
import io.realm.Realm;
import io.realm.RealmConfiguration;
[...]

PR #5196, opened 2017-09-02
Report the class missing correctly

cmelchior:
Seen in #5195 that we don't report the class missing correctly

### DIFF
--- a/realm/realm-library/src/main/cpp/jni_util/java_class.cpp
+++ b/realm/realm-library/src/main/cpp/jni_util/java_class.cpp
@@ -47,7 +47,7 @@ JavaClass::JavaClass(JavaClass&& rhs)
 JavaGlobalRef JavaClass::get_jclass(JNIEnv* env, const char* class_name)
 {
     jclass cls = env->FindClass(class_name);
-    REALM_ASSERT_RELEASE(cls);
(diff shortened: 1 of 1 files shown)


realm/realm-library/src/main/cpp/jni_util/java_class.cpp
```diff
@@ -47,7 +47,7 @@ JavaClass::JavaClass(JavaClass&& rhs)
 JavaGlobalRef JavaClass::get_jclass(JNIEnv* env, const char* class_name)
 {
     jclass cls = env->FindClass(class_name);
-    REALM_ASSERT_RELEASE(cls);
+    REALM_ASSERT_RELEASE(class_name);
```
beeender: This actually changes the semantics of the assertion. we want to check if the class can be found or not.

`REALM_ASSERT_RELEASE_EX` should be used in this case.

I am checking the original issue now.

cmelchior: Ups, you are right. The idea was that I wanted to report what class was missing if it wasn't present

cmelchior:
Thanks @beeender ... I cannot approve my own PR, but 👍

PR #5208, opened 2017-09-05
Primitive List feature(Part6: OsList)

zaki50:

PR #5209, opened 2017-09-05
Primitive List feature(Part7: Schema)

zaki50:
This PR is based on Part 6 (#5208)

This PR implements schema related feature required from proxy class.


realm/realm-annotations-processor/src/main/java/io/realm/processor/Constants.java
```diff
@@ -43,29 +43,41 @@
...
+        DOUBLE("DOUBLE", "Double", false),
+        BOOLEAN("BOOLEAN", "Boolean", false),
+        STRING("STRING", "String", false),
+        DATE("DATE", "Date", false),
+        BINARY("BINARY", "BinaryByteArray", false),
+        REALM_INTEGER("INTEGER", "Long", false),
+        OBJECT("OBJECT", "Object", false),
+        LIST("LIST", "List", true),
+
+        BACKLINK("LINKING_OBJECTS", null, true),
```
beeender: will this result https://github.com/realm/realm-java/pull/5209/files#diff-36f5b86acab52385e590ad92a6d7a80cR690 add backlink `RealmResults` as a persisted property? 
Would `isList()` useful for backlinks? maybe change it to `isRealmList` and return false for backlink if it is not useful for it?

realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmProxyClassGenerator.java
```diff
@@ -1943,7 +1952,16 @@ private static int countModelOrListFields(Collection<VariableElement> fields) {
             return Constants.RealmFieldType.OBJECT;
         }
         if (Utils.isRealmList(field)) {
```
beeender: it seems `Field.isList()` is not needed?

realm/realm-library/src/main/cpp/io_realm_internal_Table.cpp
```diff
@@ -165,8 +166,12 @@ JNIEXPORT jboolean JNICALL Java_io_realm_internal_Table_nativeIsColumnNullable(J
         ThrowException(env, UnsupportedOperation, "Not allowed to convert field in subtable.");
         return JNI_FALSE;
     }
+
     size_t column_index = S(columnIndex);
-    return to_jbool(table->is_nullable(column_index));
+    if (table->get_column_type(S(columnIndex)) != type_Table) {
+        return to_jbool(table->is_nullable(column_index)); // noexcept
+    }
+    return to_jbool(table->get_descriptor()->get_subdescriptor(S(columnIndex))->is_nullable(S(0))); // noexcept
```
beeender: why ? would not primitive list be always nullable?

realm/realm-library/src/main/cpp/io_realm_internal_Table.cpp
```diff
@@ -505,7 +510,12 @@ JNIEXPORT jint JNICALL Java_io_realm_internal_Table_nativeGetColumnType(JNIEnv*
         return 0;
     }
 
-    return static_cast<jint>(TBL(nativeTablePtr)->get_column_type(S(columnIndex))); // noexcept
+    auto column_type = TBL(nativeTablePtr)->get_column_type(S(columnIndex)); // noexcept
```
beeender: Is `Table.getColumnType()` used anywhere? I thought we should just delete this.

realm/realm-library/src/main/cpp/io_realm_internal_UncheckedRow.cpp
```diff
@@ -71,7 +72,12 @@ JNIEXPORT jint JNICALL Java_io_realm_internal_UncheckedRow_nativeGetColumnType(J
                                                                                jlong columnIndex)
 {
     TR_ENTER_PTR(nativeRowPtr)
-    return static_cast<jint>(ROW(nativeRowPtr)->get_column_type(S(columnIndex))); // noexcept
+    auto column_type = ROW(nativeRowPtr)->get_column_type(S(columnIndex)); // noexcept
+    if (column_type != type_Table) {
+        return static_cast<jint>(column_type);
+    }
+    return static_cast<jint>(ROW(nativeRowPtr)->get_table()->get_descriptor()->get_subdescriptor(S(columnIndex))->get_column_type(S(0))
```
beeender: This sounds hacky ... please add a FIXME here and i will check -- if Object Store changed the enum, we are quite difficult to figure out what is going wrong.

realm/realm-library/src/main/cpp/io_realm_internal_Table.cpp
cmelchior: We might use it in the Dynamic API. I didn't check though

realm/realm-library/src/main/cpp/io_realm_internal_Table.cpp
zaki50: no.

*nullable* for primitive list means that the list can contain null as value.

zaki50: fixed

realm/realm-library/src/main/cpp/io_realm_internal_Table.cpp
zaki50: Used in `MutableRealmObjectSchema` at least.

PR #5210, opened 2017-09-05
Primitive List feature(Part8: Annotation processor implementation)

zaki50:
This PR is based on Part 10 (#5216)

PR #5213, opened 2017-09-06
Primitive List feature(Part 9, RealmList)

zaki50:
This PR is based on Part 8 (#5210)


zaki50:
@realm/java Part 9 is basically ready for review.

There still exist some tests that expect unmanaged RealmList throws an exception against null value.

realm/realm-library/src/main/java/io/realm/RealmList.java
```diff
@@ -867,46 +837,86 @@ private void checkValidRealm() {
...
     @Override
     public String toString() {
-        if (forValues) {
-            // TODO implement this
-            return "";
-        }
-
-        StringBuilder sb = new StringBuilder();
+        final StringBuilder sb = new StringBuilder("RealmList<");
         if (isManaged()) {
```
cmelchior: Two `if (isManaged())` blocks looks a bit funny. Perhaps add a comment above each block what it does like `// Determines type of List and //Print list values`

realm/realm-library/src/main/java/io/realm/RealmList.java
```diff
@@ -1288,15 +1292,1274 @@ public void add(@Nullable E e) {
...
+                a = (T[]) Array.newInstance(componentType, size);
+            }
+
+            for (int i = 0; i < size; i++) {
+                final E e = osListOperator.get(i);
+                if (e == null || e instanceof Long) {
+                    //noinspection unchecked
+                    a[i] = (T) e;
+                } else {
+                    if (!(e instanceof Number)) {
```
cmelchior: Not sure I understand this else block. When you can enter that? Same for the rest

realm/realm-library/src/main/java/io/realm/RealmList.java
```diff
@@ -1288,15 +1292,1274 @@ public void add(@Nullable E e) {
...
+     * If this list contains {@code null} value, {@code valueForNull} is used instead.
+     *
+     * @return newly allocated {@code long} array.
+     * @throws IllegalStateException if this {@link RealmList} is not managed or element type is not implementing {@link Number}.
+     * @see #toLongArray()
+     * @see #toArray()
+     * @see #toArray(Object[])
+     */
+    @SuppressWarnings("WeakerAccess")
+    public long[] toLongArray(@SuppressWarnings("SameParameterValue") long valueForNull) {
```
cmelchior: Why add these methods. It seems a lot of code for very little benefit compared to the standard `T toArray()` and adds quite a lot of code. ?

realm/realm-library/src/main/java/io/realm/RealmList.java
```diff
@@ -1288,15 +1292,1274 @@ public void add(@Nullable E e) {
...
+        }
+        if (clazz == Date.class) {
+            //noinspection unchecked
+            return (ManagedListOperator<E>) new DateListOperator(realm, osList, (Class<Date>) clazz);
+        }
+        throw new IllegalArgumentException("Unexpected value class: " + clazz.getName());
+    }
+}
+
+abstract class ManagedListOperator<T> {
```
cmelchior: Please add Javadoc describing what this inner class is for

realm/realm-library/src/main/java/io/realm/RealmList.java
beeender: the `get(i)` here should never return a non-number value i think?

realm/realm-library/src/main/java/io/realm/RealmList.java
```diff
@@ -1288,15 +1292,1274 @@ public void add(@Nullable E e) {
...
+        if (!isManaged()) {
+            return super.toArray(a);
+        }
+
+        final Class<?> componentType = a.getClass().getComponentType();
+        final int size = size();
+        if (componentType == String.class) {
+            if (a.length < size) {
+                //noinspection unchecked
+                a = (T[]) Array.newInstance(componentType, size);
```
beeender: Can we alloc the arrays for all `if` cases? this method is really long.

realm/realm-library/src/main/java/io/realm/RealmList.java
```diff
@@ -1288,15 +1292,1274 @@ public void add(@Nullable E e) {
...
+            }
+
+            for (int i = 0; i < size; i++) {
+                final E e = osListOperator.get(i);
+                if (e == null || e instanceof Double) {
+                    //noinspection unchecked
+                    a[i] = (T) e;
+                } else {
+                    if (!(e instanceof Number)) {
```
beeender: for `float` and `double`, the `get()` should always return the right type. right?

realm/realm-library/src/main/java/io/realm/RealmList.java
beeender: I agree ... seems not quite useful + need to write lots of code.

realm/realm-library/src/main/java/io/realm/RealmList.java
```diff
@@ -1288,15 +1292,1274 @@ public void add(@Nullable E e) {
...
+            //noinspection unchecked
+            return (ManagedListOperator<E>) new DateListOperator(realm, osList, (Class<Date>) clazz);
+        }
+        throw new IllegalArgumentException("Unexpected value class: " + clazz.getName());
+    }
+}
+
+abstract class ManagedListOperator<T> {
+    static final String NULL_OBJECTS_NOT_ALLOWED_MESSAGE = "RealmList does not accept null values";
+    static final String INVALID_OBJECT_TYPE_MESSAGE = "Unacceptable value type. Acceptable: %1$s, actual: %2$s";
```
beeender: both `.` at the end.

realm/realm-library/src/main/java/io/realm/RealmList.java
```diff
@@ -1288,15 +1292,1274 @@ public void add(@Nullable E e) {
...
+            osList.addNull();
+        } else {
+            osList.addDate((Date) value);
+        }
+    }
+
+    @Override
+    public void insert(int index, @Nullable Object value) {
+        checkValidValue(value);
+        checkInsertIndex(index);
```
beeender: `checkInsertIndex()` is not needed. Object Store is doing the check.

realm/realm-library/src/main/java/io/realm/RealmList.java
```diff
@@ -1288,15 +1292,1274 @@ public void add(@Nullable E e) {
...
+        } else {
+            osList.insertDate(index, (Date) value);
+        }
+    }
+
+    @Nullable
+    @Override
+    protected Date set(int index, @Nullable Object value) {
+        checkValidValue(value);
+        checkElementExists(index);
```
beeender: Same, object store is doing that.

realm/realm-library/src/main/java/io/realm/RealmList.java
```diff
@@ -1288,15 +1292,1274 @@ public void add(@Nullable E e) {
...
+            osList.insertNull(index);
+        } else {
+            osList.insertDate(index, (Date) value);
+        }
+    }
+
+    @Nullable
+    @Override
+    protected Date set(int index, @Nullable Object value) {
+        checkValidValue(value);
```
beeender: the type needs to be checked any way. So if
```java
abstract Class checkValidValue(value);

T set(int index, @Nullable Object value) {
    Class clazz = checkValidValue(value);
    if (value == null) {
        osList.setNull(index, value);
        return;
    }

    switch(clazz) {
        case Date.class:
            osList.setDate();
            return;
        case ...:
    }
}
```

Then the child class only need to implement `checkValidValue()`.

realm/realm-library/src/main/java/io/realm/RealmList.java
zaki50: `T toArray()` and `T[] toArray(T[])` cannot return primitive array.
it means that users always need to wrapper array to get array of elements even if the list is not nullable.
That is huge overhead and no way for users to avoid that.

realm/realm-library/src/androidTest/java/io/realm/ManagedRealmListForValueTests.java
```diff
@@ -0,0 +1,1505 @@
...
+        //noinspection unchecked
+        list.add(generateValue(listType, 100));
+        realm.commitTransaction();
+
+        assertEquals(2, listenerCalledCount.get());
+        looperThread.testComplete();
+    }
+
+    @Test
+    @Ignore("unexpected fail. Unexpected insertion index.")
```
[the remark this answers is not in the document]
zaki50: These unexpected failures were already fixed and `Ignore` was removed in another PR.
To avoid conflicts I'd like to keep them as it is.

realm/realm-library/src/main/java/io/realm/RealmList.java
cmelchior: Sure, but the collection API's in general have that problem and almost no one are complaining about thoses. Besides, these add quite a lot of corner cases where they fail as well, e..g calling `toDateArray()` on a boolean list or nullable values.

E.g. even Android only have special types for maps like `SparseIntArray`, but nothing for lists. 

Also, the implementation is just doing the `Long -> long` conversion internally so you are not really saving anything on the performance when reading the array out.

beeender: Also, please notice `ArrayList<Long>.toArray()` won't return a primitive array either. So I think it is acceptable and there are common ways to do the converting which is not much slower that this.

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -76,7 +76,7 @@
      * @return {@link RealmQuery} object. After building the query call one of the {@code find*} methods
      * to run it.
      */
-    public static <E extends RealmModel> RealmQuery<E> createQuery(Realm realm, Class<E> clazz) {
```
Zhuinden: Are you sure these can just be hidden? Weren't they part of public API?

https://realm.io/docs/java/latest/api/io/realm/RealmQuery.html#createQueryFromResult-io.realm.RealmResults-

realm/realm-library/src/main/java/io/realm/RealmList.java
zaki50: Hmm, OK. I'll remove these methods.

realm/realm-library/src/main/java/io/realm/RealmQuery.java
zaki50: I think that its being public is mistake.

realm/realm-library/src/main/java/io/realm/RealmList.java
zaki50: removed

realm/realm-library/src/androidTest/java/io/realm/ManagedRealmListForValueTests.java
```diff
@@ -0,0 +1,1505 @@
...
+        object.deleteFromRealm();
+        realm.commitTransaction();
+
+        // RealmList contained in removed object is invalid.
+        assertFalse(list.isValid());
+    }
+
+    private static final int HEADER_SIZE = 8;
+    private static final int MAX_STRING_SIZE = 0xFFFFF8 - HEADER_SIZE - 1;
+    private static final int MAX_BINARY_SIZE = 0xFFFFF8 - HEADER_SIZE;
```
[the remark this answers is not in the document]
zaki50: done

realm/realm-library/src/androidTest/java/io/realm/ManagedRealmListForValueTests.java
```diff
@@ -0,0 +1,1505 @@
...
+                case STRING_LIST: {
+                    assertEquals(generateValue(STRING_LIST, i), list.get(typeIsNullable ? (i * 2) : i));
+                }
+                break;
+                case BOOLEAN_LIST: {
+                    assertEquals(generateValue(BOOLEAN_LIST, i), list.get(typeIsNullable ? (i * 2) : i));
+                }
+                break;
+                case BINARY_LIST: {
+                    assertTrue(Arrays.equals((byte[]) generateValue(BINARY_LIST, i), (byte[]) list.get(typeIsNullable ? (i * 2) : i)));
```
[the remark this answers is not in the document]
zaki50: Good catch!
Fixed.

realm/realm-library/src/main/java/io/realm/RealmQuery.java
cmelchior: I tend to agree. I believe they were public because the original migration/dynamic API had a lot of moving parts and at some point the dynamic parts lived in their own package.

It is a little bad form to just remove them without warning, but all of these should be accessible using normal `realm.where()` methods and the like instead, so probably not a big deal.

As long as it is mentioned in the changelog removing from public is probably fine.

zaki50: Added changelog entry for these.

realm/realm-library/src/androidTest/java/io/realm/ManagedRealmListForValueTests.java
```diff
@@ -0,0 +1,1505 @@
...
+        private final RealmFieldType fieldType;
+        private final String valueTypeName;
+
+        ListType(RealmFieldType fieldType, String valueTypeName) {
+            this.fieldType = fieldType;
+            this.valueTypeName = valueTypeName;
+        }
+
+        @SuppressWarnings("unused")
+        public RealmFieldType getFieldType() {
```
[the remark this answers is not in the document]
zaki50: removed

realm/realm-library/src/main/java/io/realm/RealmList.java
```diff
@@ -215,17 +210,11 @@ public void add(int location, @Nullable E object) {
      */
     @Override
     public boolean add(@Nullable E object) {
-        if (forValues) {
-            // TODO implement this
-            return false;
-        }
-
         //noinspection ConstantConditions
```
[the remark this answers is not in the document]
zaki50: yes. removed.

realm/realm-library/src/main/java/io/realm/RealmList.java
```diff
@@ -215,17 +210,11 @@ public void add(int location, @Nullable E object) {
...
-        checkValidObject(object);
         if (isManaged()) {
             checkValidRealm();
-            RealmObjectProxy proxy = (RealmObjectProxy) copyToRealmIfNeeded((RealmModel) object);
-            osList.addRow(proxy.realmGet$proxyState().getRow$realm().getIndex());
+            //noinspection ConstantConditions
```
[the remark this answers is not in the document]
zaki50: removed

realm/realm-library/src/main/java/io/realm/RealmList.java
```diff
@@ -1288,15 +1292,1274 @@ public void add(@Nullable E e) {
...
+            }
+
+            for (int i = 0; i < size; i++) {
+                final E e = osListOperator.get(i);
+                if (e == null || e instanceof Long) {
+                    //noinspection unchecked
+                    a[i] = (T) e;
+                } else {
+                    if (!(e instanceof Number)) {
+                        throw new ArrayStoreException(String.format(Locale.ENGLISH,
```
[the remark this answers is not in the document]
zaki50: No such a predefined locale. `Location.ENGLISH` is `"US_en"`

realm/realm-library/src/main/java/io/realm/RealmList.java
zaki50: No. User can write following code.

```
RealmList<String> stringList = obj.getStringList();
Double[] doubleArray = stringList.toArray(new Double[0]);
```

Of course it's wrong code, but we need to throw `ArrayStoreException` in that case.

beeender: > Thrown to indicate that an attempt has been made to store the wrong type of object into an array of objects. For example, the following code generates an ArrayStoreException:

```
         Object x[] = new String[3];
         x[0] = new Integer(0);
```

It will throw the `ArrayStoreException` anyway I think?

Implementation for the `AbastractCollection` http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/6-b14/java/util/AbstractCollection.java#173

I think `ArrayStoreException` will be thrown when it tries to set the first element of the array.

realm/realm-library/src/main/java/io/realm/RealmList.java
zaki50: @cmelchior When a user defines the field like `public RealmList<Integer> fooList;`, following call can enter `else` block.

```
Long[] integerList = realmList.toArray(new Long[0]);
```

Should we throw `ArrayStoreException` in this case?

realm/realm-library/src/main/java/io/realm/RealmList.java
```diff
@@ -409,9 +345,26 @@ public boolean remove(@Nullable Object object) {
         if (isManaged() && !realm.isInTransaction()) {
             throw new IllegalStateException(REMOVE_OUTSIDE_TRANSACTION_ERROR);
         }
+
+        if (object instanceof byte[] && clazz == byte[].class) {
+            // we need special handling for class since equals against byte[] never matches.
```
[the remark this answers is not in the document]
cmelchior: I agree. The Java Collection API's are far from perfect, but people are used to working with them.

realm/realm-library/src/androidTest/java/io/realm/ManagedRealmListForValueTests.java
```diff
@@ -0,0 +1,1505 @@
...
+        assertEquals(0, list.size());
+    }
+
+    @Test
+    public void deleteAllFromRealm_outsideTransaction() {
+        try {
+            list.deleteAllFromRealm();
+            fail("deleteAllFromRealm should be called in a transaction.");
+        } catch (IllegalStateException e) {
+            assertThat(e.getMessage(), containsString("Must be in a write transaction "));
```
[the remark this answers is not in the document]
zaki50: done

realm/realm-library/src/main/java/io/realm/RealmList.java
```diff
@@ -59,7 +61,7 @@
 public class RealmList<E> extends AbstractList<E> implements OrderedRealmCollection<E> {
 
     private static final String ONLY_IN_MANAGED_MODE_MESSAGE = "This method is only available in managed mode";
-    private static final String NULL_OBJECTS_NOT_ALLOWED_MESSAGE = "RealmList does not accept null values";
+    private static final String ALLOWED_ONLY_FOR_REALM_MODEL_ELEMENT_MESSAGE = "This feature is available only when the element type is implementing RealmModel";
```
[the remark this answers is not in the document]
zaki50: done

realm/realm-library/src/androidTest/java/io/realm/ManagedRealmListForValueTests.java
```diff
@@ -0,0 +1,1505 @@
...
+                    assertValueEquals(generateValue(listType, i / 2), list.get(i));
+                }
+            } else {
+                assertValueEquals(generateValue(listType, i), list.get(i));
+            }
+        }
+    }
+
+    @Test
+    public void add_atAfterContainerObjectRemoved() {
```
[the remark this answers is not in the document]
zaki50: fixed

realm/realm-library/src/androidTest/java/io/realm/ManagedRealmListForValueTests.java
```diff
@@ -0,0 +1,1505 @@
...
+        final List<Object> unsupportedValues = Arrays.<Object>asList(
+                new int[] {0},
+                new StringBuilder("0")
+        );
+        realm.executeTransaction(new Realm.Transaction() {
+            @Override
+            public void execute(Realm realm) {
+                for (Object unsupportedValue : unsupportedValues) {
+                    //noinspection UseBulkOperation
+                    list.remove(unsupportedValue);
```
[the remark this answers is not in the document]
zaki50: added assertion

realm/realm-library/src/main/java/io/realm/RealmList.java
zaki50: updated.

realm/realm-library/src/main/java/io/realm/RealmList.java
zaki50: done

realm/realm-library/src/main/java/io/realm/RealmList.java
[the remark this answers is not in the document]
zaki50: removed all special handling for `byte[]`.

realm/realm-library/src/main/java/io/realm/RealmList.java
zaki50: removed special implementation of `toArray(T[])`

realm/realm-library/src/main/java/io/realm/RealmList.java
zaki50: removed special implementation of `toArray(T[])`

realm/realm-library/src/main/java/io/realm/RealmList.java
zaki50: removed special implementation of `toArray(T[])`

realm/realm-library/src/main/java/io/realm/RealmList.java
zaki50: removed

realm/realm-library/src/main/java/io/realm/RealmList.java
zaki50: removed

realm/realm-library/src/main/java/io/realm/RealmList.java
zaki50: refactored

realm/realm-library/src/main/java/io/realm/RealmList.java
zaki50: added

zaki50:
I've addressed all comment. Could you give second-round review? @realm/java

CHANGELOG.md
```diff
@@ -7,6 +7,8 @@
 ### Breaking Changes
 
 * `RealmResults.distinct()`/`RealmResults.distinctAsync()` have been removed. Use `RealmQuery.distinct()`/`RealmQuery.distinctAsync()` instead.
+* `RealmQuery.createQuery(Realm, Class)`, `RealmQuery.createDynamicQuery(DynamicRealm, String)`, `RealmQuery.createQueryFromResult(RealmResults)` and `RealmQuery.createQueryFromList(RealmList)` have been removed. Use `Realm.where(Class)`, `DynamicRealm.where(String)`, RealmResults.where()` and `RealmList.where()` instead.
+* Some exception messages are revised.
```
beeender: this is not serious breaking change i guess :P put it to enhancement or just remove?

realm/realm-library/src/main/java/io/realm/RealmList.java
```diff
@@ -1288,15 +1223,664 @@ public void add(@Nullable E e) {
...
+        if (clazz == Date.class) {
+            //noinspection unchecked
+            return (ManagedListOperator<E>) new DateListOperator(realm, osList, (Class<Date>) clazz);
+        }
+        throw new IllegalArgumentException("Unexpected value class: " + clazz.getName());
+    }
+}
+
+/**
+ * This class provides façade for against {@link OsList}. {@link OsList} is used for both {@link RealmModel}s
```
beeender: `façade` -> `facade` `ç` here (I thought my monitor was dirty)

realm/realm-library/src/main/java/io/realm/RealmList.java
```diff
@@ -1277,41 +1277,82 @@ private static boolean isClassForRealmModel(Class<?> clazz) {
         this.osList = osList;
     }
 
-    abstract boolean forRealmModel();
+    public abstract boolean forRealmModel();
 
-    OsList getOsList() {
+    public final OsList getOsList() {
```
Zhuinden: Are you sure?

realm/realm-library/src/main/java/io/realm/RealmList.java
```diff
@@ -1277,41 +1277,82 @@ private static boolean isClassForRealmModel(Class<?> clazz) {
...
-    final void checkInsertIndex(int index) {
-        final int size = size();
-        if (index < 0 || size < index) {
-            throw new IndexOutOfBoundsException("Invalid index " + index + ", size is " + osList.size());
+    protected abstract void checkValidValue(@Nullable Object value);
+
+    @Nullable
+    public abstract T get(int index);
+
+    public final void append(@Nullable Object value) {
```
Zhuinden: Now the question is, what is the difference between `List.add()` and `RealmList.append()`.

realm/realm-library/src/main/java/io/realm/RealmList.java
zaki50: `façade` is correct word, but I don't mind to fix this!

CHANGELOG.md
zaki50: removed.

realm/realm-library/src/main/java/io/realm/RealmList.java
zaki50: fixed

PR #5220, opened 2017-09-07
Primitive List feature(Part 11, Collection, Snapshot)

zaki50:
This PR is based on Part 9 (#5213).


realm/realm-library/src/main/java/io/realm/OrderedRealmCollectionImpl.java
```diff
@@ -60,8 +59,8 @@ Collection getCollection() {
         return collectionOperator.getCollection();
     }
 
-    boolean forValue() {
-        return collectionOperator.forValue();
+    boolean forModelClasses() {
```
Zhuinden: this was renamed to mean its opposite but the values it provides is still `return true`, is that ok?

cmelchior: I changed it because having a method called `forValue()` return `false` for an `int` sounded really wrong 😄

Zhuinden: I agree, that does sound wrong!

cmelchior:
This change introduces a lot of Iterator crashes:
```

io.realm.OrderedRealmCollectionIteratorTests > useCase_forEachIterator_modifyQueryResult_innerTransaction[REALMRESULTS][Nexus 5X - 6.0] FAILED 
	java.lang.AssertionError: expected:<0> but was:<5>
	at org.junit.Assert.fail(Assert.java:88)

io.realm.OrderedRealmCollectionIteratorTests > useCase_forEachIterator_modifyQueryResult_outerTransaction[REALMRESULTS][Nexus 5X - 6.0] FAILED 
	java.lang.AssertionError: expected:<0> but was:<5>
	at org.junit.Assert.fail(Assert.java:88)

io.realm.OrderedRealmCollectionIteratorTests > listIterator_transactionBeforeNextItem[REALMRESULTS][Nexus 5X - 6.0] FAILED 
	java.lang.AssertionError: Failed at index: 1 expected:<1> but was:<0>
	at org.junit.Assert.fail(Assert.java:88)

io.realm.OrderedRealmCollectionIteratorTests > useCase_forEachIterator_modifyQueryResult_outerTransaction_looperThread[REALMRESULTS][Nexus 5X - 6.0] FAILED 
	java.lang.AssertionError: expected:<0> but was:<5>
	at org.junit.Assert.fail(Assert.java:88)

io.realm.OrderedRealmCollectionIteratorTests > listIterator_deleteManagedObjectIndirectly[REALMRESULTS][Nexus 5X - 6.0] FAILED 
	java.lang.AssertionError
	at org.junit.Assert.fail(Assert.java:86)

io.realm.OrderedRealmCollectionIteratorTests > useCase_forEachIterator_modifyQueryResult_innerTransaction_looperThread[REALMRESULTS][Nexus 5X - 6.0] FAILED 
	java.lang.AssertionError: expected:<0> but was:<5>
	at org.junit.Assert.fail(Assert.java:88)

io.realm.OrderedRealmCollectionIteratorTests > iterator_unrelatedTransactionBeforeNextItem[REALMRESULTS][Nexus 5X - 6.0] FAILED 
	java.lang.AssertionError: Failed at index: 1 expected:<1> but was:<0>
	at org.junit.Assert.fail(Assert.java:88)
```

But with our introduction of `createSnapshot()` I'm actually surprised these didn't fail sooner. Also,  I'm not sure why they suddenly start failing now. I suspect it is due to changes in the underlying Results class, but I'm not sure. I hope you know @beeender ?

realm/realm-library/src/main/java/io/realm/OrderedRealmCollectionImpl.java
```diff
@@ -144,23 +143,15 @@ public E first(@Nullable E defaultValue) {
...
-        } else {
+        if (collectionOperator.isEmpty()) {
             if (shouldThrow) {
                 throw new IndexOutOfBoundsException("No results were found.");
             } else {
                 return defaultValue;
             }
         }
+
+        return collectionOperator.get(0);
```
Zhuinden: I'm actually not a fan of this change after having to do optimizations specifically because `.size()` and `.get(0)` were slower than just accessing the first row directly. :|

realm/realm-library/src/main/java/io/realm/OrderedRealmCollectionImpl.java
```diff
@@ -184,23 +175,16 @@ public E last(@Nullable E defaultValue) {
...
+        final int size = collectionOperator.size();
+        if (size == 0) {
             if (shouldThrow) {
                 throw new IndexOutOfBoundsException("No results were found.");
             } else {
                 return defaultValue;
             }
         }
+
+        return collectionOperator.get(size - 1);
```
Zhuinden: Same ( https://github.com/realm/realm-java/pull/5220/files#r142044476 )

cmelchior:
Replaced by https://github.com/realm/realm-java/pull/5364

realm/realm-library/src/main/java/io/realm/OrderedRealmCollectionImpl.java
```diff
@@ -144,23 +143,15 @@ public E first(@Nullable E defaultValue) {
...
-        if (forValues) {
-            // TODO implement this
-            return null;
-        }
-
-        if (row != null) {
-            //noinspection unchecked
-            return (E) realm.get((Class<? extends RealmModel>) classSpec, className, row);
-        } else {
+        if (collectionOperator.isEmpty()) {
```
Zhuinden: This is going to cause performance degradation.

realm/realm-library/src/main/java/io/realm/OrderedRealmCollectionImpl.java
```diff
@@ -184,23 +175,16 @@ public E last(@Nullable E defaultValue) {
...
-        if (forValues) {
-            // TODO implement this
-            return null;
-        }
-
-        if (row != null) {
-            //noinspection unchecked
-            return (E) realm.get((Class<? extends RealmModel>) classSpec, className, row);
-        } else {
+        final int size = collectionOperator.size();
```
Zhuinden: This is going to cause performance degradation.

Issue #5231, opened 2017-09-08
Fix PrimaryKeyTests#removingPrimaryKeyRemovesConstraint_typeSetters


cmelchior:
This unit test surfaces a problem with our schema cache:
```
   /**
     * This test surfaces a bunch of problems, most of them seem to be around caching of the schema
     * during a transaction
     *
     * 1) Removing the primary key do not invalidate the cache in RealmSchema and those cached
     *    are ImmutableRealmObjectSchema so do not change when the primary key is removed.
     *
     * 2) Addding `schema.refresh()` to RealmObjectSchema.removePrimaryKey()` causes
     *    RealmPrimaryKeyConstraintException anyway. Unclear why.
     */
```

beeender:
I am going to fix this by using Object Store PK logic to do related stuff. Caching the PK index in java is not a wise idea. And it is not on a hot path, we benefit very little by caching.

PR #5197, opened 2017-09-02
Fix flaky integration tests

cmelchior:
Work in progress

PR #5232, opened 2017-09-08
Ignore broken unit test

cmelchior:
(for now)

We seem to have some issues with schema cache invalidation during a transaction. Not sure why this was not caught on the `master-4.0` branch. Anyway, disabling for now and tracking fixing it in https://github.com/realm/realm-java/issues/5231

Also, I moved the unit test to the public API as using the internal API missed some things like adding search indexes which broke other checks in the lower levels of the API.

### DIFF
--- a/realm/realm-library/src/androidTest/java/io/realm/internal/PrimaryKeyTests.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/internal/PrimaryKeyTests.java
@@ -21,6 +21,7 @@ import android.support.test.runner.AndroidJUnit4;
 
 import org.junit.After;
 import org.junit.Before;
+import org.junit.Ignore;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.runner.RunWith;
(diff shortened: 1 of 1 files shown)


realm/realm-library/src/androidTest/java/io/realm/internal/PrimaryKeyTests.java
```diff
@@ -83,36 +90,40 @@ private Table getTableWithIntegerPrimaryKey() {
...
+     * during a transaction
+     * 
+     * 1) Removing the primary key do not invalidate the cache in RealmSchema and those cached
+     *    are ImmutableRealmObjectSchema so do not change when the primary key is removed.
+     *
+     * 2) Addding `schema.refresh()` to RealmObjectSchema.removePrimaryKey()` causes
+     *    RealmPrimaryKeyConstraintException anyway. Unclear why.
+     */
     @Test
+    @Ignore
```
zaki50: We should have *ignore comment*. How about `See test case comment`?

cmelchior:
Folding into #5197

PR #5249, opened 2017-09-11
Use NPM to install integration test server

cmelchior:
This PR now lets us install the Realm Object Server used for our Integration tests using NPM instead of package cloud.

Node.js 6.10+ is required. 

Changes
-----
* Upgraded to use latest NPM package (2.0.0-alpha.30)
* Server side token revocation disabled for now.
* Changed `retrieveUserInfo` endpoint to match new behaviour
* Disabled tests for `changePassword`. Functionality is currently broken
* Workaround for bug with default admin user on ROS
* Workaround for bug when stopping the ROS process


cmelchior:
Docker seems to fail on CI with
```
$ docker stop --time=1 030613c06086d15ea2c63eeabad5351195a793713dafe258e3ba8e9116303131
$ docker rm -f 030613c06086d15ea2c63eeabad5351195a793713dafe258e3ba8e9116303131
[Pipeline] // withDockerContainer
[Pipeline] sh
[realm-java] Running shell script
+ docker logs f6b9df4d01948c47be5814129a82c216dcb8bc49f13b86642541b894ab5a0076
info: Integration test server listening on: 127.0.0.1:8888
info: Starting sync server in  /tmp/ros117811-7-fi57l3.6v6fphw7b9
info: 
events.js:160
      throw er; // Unhandled 'error' event
      ^

Error: spawn realm-object-server ENOENT
    at exports._errnoException (util.js:1020:11)
    at Process.ChildProcess._handle.onexit (internal/child_process.js:197:32)
    at onErrorNT (internal/child_process.js:376:16)
    at _combinedTickCallback (internal/process/next_tick.js:80:11)
    at process._tickCallback (internal/process/next_tick.js:104:9)
[Pipeline] sh
```
It works locally on my Mac 🤔

tools/sync_test_server/Dockerfile
```diff
@@ -2,21 +2,21 @@ FROM ubuntu:16.04
...
 RUN apt-get update -qq \
-    && apt-get install -y curl npm \
-    && curl -s https://packagecloud.io/install/repositories/realm/realm-beta/script.deb.sh | bash
-    #&& curl -s https://packagecloud.io/install/repositories/realm/realm/script.deb.sh | bash
-    #&& curl -s https://packagecloud.io/install/repositories/realm/realm-testing/script.deb.sh | bash
+    && apt-get install -y curl
+
+# Install node using nodesource to be able to install desired major version
+# Not doing this will cause apt-get to install from 4.*
+RUN curl -sL https://deb.nodesource.com/setup_6.x | bash - \
```
emanuelez: Why not use the official node image as the base for this Dockerfile? Then you wouldn't have to worry about installing the right version and it would be faster to build the image.

tools/sync_test_server/Dockerfile
```diff
@@ -2,21 +2,21 @@ FROM ubuntu:16.04
...
+    && apt-get install -y nodejs build-essential
```
emanuelez: What do we need build-essential for? We donøt need to build node or ROS in here, do we?

tools/sync_test_server/Dockerfile
```diff
@@ -2,21 +2,21 @@ FROM ubuntu:16.04
...
 
 # ROS npm dependencies
```
emanuelez: I think this section is now obsolete and should not be needed anymore

tools/sync_test_server/Dockerfile
cmelchior: How can you do that?

tools/sync_test_server/Dockerfile
nnrealm: There's a tiny bit of code that seems to need it, unfortunately. URSA, which generates keys, runs gyp and fails if there's no c++ compiler available. I do the same on the packet machines.

nnrealm: I just tried it with the Node image, and URSA does require a C++ compiler, but it's there, so I changed the Dockerfile.

cmelchior:
CI now fails with:
```
rpc error: code = 2 desc = oci runtime error: exec failed: container_linux.go:255: creating new parent process caused "container_linux.go:1462: running lstat on namespace path \"/proc/20451/ns/net\" caused \"lstat /proc/20451/ns/net: no such file or directory\""
```
Anyone have any good ideas what that is about?

nnrealm:
This looks more like a regular error?

cmelchior:
@nnrealm Yes, awesome. Thanks for fixing it. You don't seem to use our custom configuration file though. Is that not supported? (We have some tests that depend on it being a specific way)

nnrealm:
Unfortunately, these options will need to be set in another way. ROS doesn't use a config file at all right now. What do you need it to do?

cmelchior:
This is the file: https://github.com/realm/realm-java/blob/master/tools/sync_test_server/configuration.yml

A list would be:
- Use custom keys for SSH
- Duration of access token is modified
- `debug` provider is enabled (to access admin functionality)
- https is enabled with custom certificates

nhachicha:
@cmelchior this is because the query to check if the auth server is ready `HttpUtils#waitAuthServerReady` is no longer supported. Sending an ` HTTP GET http://127.0.0.1:9080/` will return a `404` 

```
Response{protocol=http/1.1, code=404, message=Not Found, url=http://127.0.0.1:9080/}
```

realm/realm-library/src/objectServer/java/io/realm/SyncUser.java
```diff
@@ -274,31 +274,7 @@ public void logout() {
...
-                protected void onSuccess(LogoutResponse response) {
-                    SyncManager.notifyUserLoggedOut(SyncUser.this);
-                }
-
-                @Override
-                protected void onError(LogoutResponse response) {
-                    RealmLog.error("Failed to log user out.\n" + response.getError().toString());
-                }
-            });
+            SyncManager.notifyUserLoggedOut(SyncUser.this);
```
beeender: Is this related with https://github.com/realm/ros/issues/261 ?
no need to send logout request anymore?

realm/realm-library/src/objectServer/java/io/realm/SyncUserInfo.java
```diff
@@ -24,50 +27,41 @@
...
     /**
-     * @return Whether the user is flagged on the Realm Object Server as an administrator.
+     * @return whether the user is flagged on the Realm Object Server as an administrator.
      */
     public boolean isAdmin() {
         return isAdmin;
     }
 
+    /**
+     * @return the metadata associated with this user.
```
beeender: maybe a example about what kind of map could be returned? i am bit unclear when read this doc.

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/AuthTests.java
```diff
@@ -661,15 +664,6 @@ public void run() {
     }
 
     @Test
-    public void retrieve_AdminUser() {
```
beeender: this test is not needed? why?

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/utils/HttpUtils.java
```diff
@@ -56,62 +61,27 @@ public static void startSyncServer() throws Exception {
...
+    public static void stopSyncServer() throws Exception {
         Request request = new Request.Builder()
-                .url(Constants.AUTH_SERVER_URL)
+                .url(STOP_SERVER)
                 .build();
 
-        while (retryTimes != 0) {
-            Response response = null;
+        Response response;
+        int attempts = 3;
```
beeender: why do you need retries on stop but not start?

realm/realm-library/src/objectServer/java/io/realm/SyncUserInfo.java
```diff
@@ -24,50 +27,41 @@
...
+     */
+    public Map<String, String> getMetadata() {
```
beeender: looks like a breaking change, mention it in changelog?

realm/realm-library/src/objectServer/java/io/realm/SyncUser.java
cmelchior: Yup, it is related to that issue. Until that is resolved in some way sending the logout request isn't supported.

realm/realm-library/src/objectServer/java/io/realm/SyncUserInfo.java
cmelchior: It is just a `key/value` map. I'll expand the description

realm/realm-library/src/objectServer/java/io/realm/SyncUserInfo.java
cmelchior: Not really breaking since we control the `SyncUserInfo` class, but it probably still requires a mention in the changelog.

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/AuthTests.java
cmelchior: The `debug` auth provider is no longer supported, which this test was built around.

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/utils/HttpUtils.java
cmelchior: This was an attempt to fix another issue, but is really the wrong fix. I'll remove the retry again.

PR #5263, opened 2017-09-15
Check and throw when converting String/byte[]

beeender:
Throw IAE if the data length exceeds the limit.


realm/realm-library/src/main/java/io/realm/internal/Table.java
```diff
@@ -47,6 +47,9 @@
     private static final long PRIMARY_KEY_FIELD_COLUMN_INDEX = 1;
     public static final long NO_PRIMARY_KEY = -2;
 
+    public static final int MAX_BINARY_SIZE = 0xFFFFF8 - 8/*array header size*/;
```
cmelchior: Where did you get this information from? Perhaps point to that (I assume Core?)

beeender: @cmelchior See https://github.com/realm/realm-java/pull/5263/files#diff-92306bfcbfe23e84db32c9f0a63c7c62R35

beeender:
Failures are known integration testing failures.

For a single flavour test it takes 17 mins vs 13 mins before which is kind of acceptable.

I will merge this. and there are some other easy things to do to make CI faster. I will do it in other PRs.

PR #5264, opened 2017-09-15
Enable debug core by default and remove deprectad RealmResults.distinct

beeender:
It seems core with assertion is getting much faster on 3.0.
Finished all base tests on my local machine in 3 mintues! Let's try it on CI.


realm/realm-library/src/androidTest/java/io/realm/RealmQueryTests.java
```diff
@@ -3034,8 +3034,8 @@ private void populateForDistinctInvalidTypesLinked(Realm realm) {
 
     @Test
     public void distinct() {
-        final long numberOfBlocks = 25;
-        final long numberOfObjects = 10; // Must be greater than 1
+        final long numberOfBlocks = 3;
```
nhachicha: Performance change ?

beeender: yups, creating rows takes lots of time in core when assertion enabled.

beeender:
Failures are known integration testing failures.

For a single flavour test it takes 17 mins vs 13 mins before which is kind of acceptable.

I will merge this. and there are some other easy things to do to make CI faster. I will do it in other PRs.

PR #5277, opened 2017-09-18
Upgrade to latest Sync RC + ROS 

cmelchior:
* Sync-RC21
* ROS 2.0.0-alpha.34
* Added support for Authorization headers
* Added support for new Sync exception
* Enabled remote revokation again

TODO
- [x] Figure out why some `retrieve` tests are now failing

### DIFF
--- a/realm/realm-library/src/objectServer/java/io/realm/SyncUser.java
+++ b/realm/realm-library/src/objectServer/java/io/realm/SyncUser.java
@@ -131,6 +131,7 @@ public class SyncUser {
         Collection<SyncUser> storedUsers = userStore.allUsers();
         Map<String, SyncUser> map = new HashMap<>();
         for (SyncUser user : storedUsers) {
+            // FIXME: Users might be marked for deletion here
             if (user.isValid()) {
                 map.put(user.getIdentity(), user);
             }
@@ -280,20 +281,17 @@ public class SyncUser {
             // don't reference directly the refreshToken inside the revoke request
             // as it may revoke the newly acquired and refresh_token
             final Token refreshTokenToBeRevoked = refreshToken;
-            RealmLog.error("revoke " + refreshTokenToBeRevoked.value());
 
             ThreadPoolExecutor networkPoolExecutor = SyncManager.NETWORK_POOL_EXECUTOR;
             networkPoolExecutor.submit(new ExponentialBackoffTask<LogoutResponse>() {
 
                 @Override
                 protected LogoutResponse execute() {
-                    RealmLog.error("Executing revoke: " + refreshTokenToBeRevoked.value());
                     return server.logout(refreshTokenToBeRevoked, getAuthenticationUrl());
                 }
 
                 @Override
                 protected void onSuccess(LogoutResponse response) {
-                    RealmLog.error("User logged out success" + SyncUser.this.getAccessToken().value());
                     SyncManager.notifyUserLoggedOut(SyncUser.this);
                 }
 
--- a/realm/realm-library/src/objectServer/java/io/realm/SyncUser.java
+++ b/realm/realm-library/src/objectServer/java/io/realm/SyncUser.java
@@ -528,16 +528,10 @@ public class SyncUser {
         AuthenticationServer authServer = SyncManager.getAuthServer();
         LookupUserIdResponse response = authServer.retrieveUser(refreshToken, provider, providerUserIdentity, getAuthenticationUrl());
         if (!response.isValid()) {
-            // the endpoint returns a 404 if it can't honor the query, either because
-            // - provider is not valid
-            // - provider_id is not valid
-            // - token used is not an admin one
-            // in this case we should return null instead of throwing
-            if (response.getError().getErrorCode() == ErrorCode.UNKNOWN_ACCOUNT) {
-                return null;
-            } else {
-                throw response.getError();
-            }
+            // Right now errors are very inconsistent. See https://github.com/realm/ros/issues/310
+            // Treat them all as "User not existing". This is too broad, and should be revisited
+            // once #310 is fixed.
+            return null;
         } else {
             return SyncUserInfo.fromLookupUserIdResponse(response);
         }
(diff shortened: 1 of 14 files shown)


cmelchior:
Ready for review @realm/java

realm/realm-library/src/objectServer/java/io/realm/SyncUser.java
```diff
@@ -274,7 +275,31 @@ public void logout() {
...
+            ThreadPoolExecutor networkPoolExecutor = SyncManager.NETWORK_POOL_EXECUTOR;
+            networkPoolExecutor.submit(new ExponentialBackoffTask<LogoutResponse>() {
+
+                @Override
+                protected LogoutResponse execute() {
+                    return server.logout(refreshTokenToBeRevoked, getAuthenticationUrl());
+                }
+
+                @Override
+                protected void onSuccess(LogoutResponse response) {
```
beeender: So the ROS solved the logout issue?

It is not related with this PR, but shall we just ignore the response of logout and just notify user that it has been logged out? Could user logout again if the first try failed? I guess user cannot ?

realm/realm-library/src/objectServer/java/io/realm/internal/network/AuthServerResponse.java
```diff
@@ -63,7 +63,14 @@ public static ObjectServerError createError(String response, int httpErrorCode)
             JSONObject obj = new JSONObject(response);
             String title = obj.optString("title", null);
             String hint = obj.optString("hint", null);
-            ErrorCode errorCode = ErrorCode.fromInt(obj.optInt("code", -1));
+            ErrorCode errorCode;
+            if (obj.has("code")) {
+                errorCode = ErrorCode.fromInt(obj.getInt("code"));
+            } else if (obj.has("status")) {
```
beeender: this looks bad :( are `code` and `status` the same thing? they all indicate the error code?

realm/realm-library/src/objectServer/java/io/realm/SyncUser.java
```diff
@@ -503,16 +528,10 @@ public SyncUserInfo retrieveInfoForUser(final String providerUserIdentity, final
...
-            // - provider is not valid
-            // - provider_id is not valid
-            // - token used is not an admin one
-            // in this case we should return null instead of throwing
-            if (response.getError().getErrorCode() == ErrorCode.NOT_FOUND) {
-                return null;
-            } else {
-                throw response.getError();
-            }
+            // Right now errors are very inconsistent. See https://github.com/realm/ros/issues/310
```
bmunkholm: Add "TODO" ?

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/utils/UserFactory.java
```diff
@@ -161,40 +161,12 @@ public static void logoutAllUsers() {
...
-                            // For dummy users, calling `logout()` will never result in the
-                            // authentication listener to trigger since the URL doesn't exist.
-                            // For these cases, we manually trigger the listener.
-                            listener.loggedOut(user);
-                        }
-                    }
+                for (SyncUser user : users.values()) {
+                    user.logout();
                 }
+                SystemClock.sleep(2000); // Remove when https://github.com/realm/ros/issues/304 is fixed
```
beeender: https://github.com/realm/ros/issues/304 has been marked as fixed

tools/sync_test_server/ros-testing-server.js
```diff
@@ -93,8 +96,6 @@ function stopRealmObjectServer(onSuccess, onError) {
         onSuccess();
     });
 
-    // Move back to `SIGTERM` once https://github.com/realm/ros/issues/234
-    // is resolved
```
beeender: NO need to change back to `SIGTERM` ? Actually `SIGKILL` is better I think, it will surely kill the ROS.

realm/realm-library/src/objectServer/java/io/realm/SyncUser.java
cmelchior: Yes, you can now revoke tokens.

Right now the `AuthenticationListener` was supposed to represent the server state, so it shouldn't be triggered for only a local logout (where you just delete the refreshToken).

But that said, there are a lot of cases where these semantics get a bit murky, e.g. it isn't triggered if another device logs you out. Only way we detect this is by having something throw an `invalid access` error somewhere.

The biggest problem as you point out is that if the revoke call fails, the user is not notified, but they can call `logout` again. Only access tokens to Realms are removed and they are marked as logged out, but they keep the refresh token.

I'll create an issue to track this as it probably requires some discussion

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/utils/UserFactory.java
cmelchior: Yes, but not released yet.

realm/realm-library/src/objectServer/java/io/realm/internal/network/AuthServerResponse.java
cmelchior: Yes. https://github.com/realm/ros/issues/311 is tracking it

realm/realm-library/src/objectServer/java/io/realm/SyncUser.java
cmelchior: https://github.com/realm/realm-java/issues/5288

tools/sync_test_server/ros-testing-server.js
cmelchior: I was still having problems with `SIGTERM`. Not sure if SIGKILL is too heavy-handed, but it seems to get the job done.

realm/realm-library/src/objectServer/java/io/realm/SyncUser.java
```diff
@@ -131,6 +131,7 @@ public static SyncUser currentUser() {
         Collection<SyncUser> storedUsers = userStore.allUsers();
         Map<String, SyncUser> map = new HashMap<>();
         for (SyncUser user : storedUsers) {
+            // FIXME: Users might be marked for deletion here
```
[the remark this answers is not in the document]
cmelchior: Sorry, this FIXME no longer apply. It was a wrong conclussion I made when testing. I'll remove it

Issue #2476, opened 2016-03-16
Support mapping from field name in model class to column name in database file.


zaki50:
If other bindings than Java create a field `class`, we can't access to it with typed Realm since we can't define the field `class` in model classes.

One idea to solve this is that introducing `Field` annotation which maps field name in model class and column name in Realm database file.

beeender:
Yep, I think it is something necessary sooner or later.

cmelchior:
With Sync this has become even more important since the default style for fields/properties differ between iOS/Android/.NET

cmelchior:
From internal conversation:

---

From #sync https://realmio.slack.com/archives/sync/p1472132469000678
Christian Melchior
A huge thing I haven’t seen mentioned elsewhere is how to handle differences in class/field names across bindings? All our major languages have widely different conventions like `camelCase`, `PascalCase` and `underscore_style`, not to mention that e.g. C++ is fond of Hungarian notation `m_underscore_style`

Do we have any strategy for handling this both short and long term, because it is definitely something that will annoy people that they have to use another platforms style convention.

[3:55]  
Requiring people to map each and every field is not going to be a viable option long-term IMO, but any alternative doesn’t exactly feel nice either. GSON have an very elaborate mapping schema, but I have seen that getting some hate as well https://google.github.io/gson/apidocs/com/google/gson/FieldNamingPolicy.html

Thomas Goyne [3:57 PM] 
automagic mapping sucks, manual mapping sucks, and no mapping sucks

[3:58]  
so I kinda think we need to do all three and let people pick which one they hate least

---

Zhuinden:
I'm a fan of what LoganSquare does.

If you want to map things, then annotate it.

cmelchior:
Where I think I see this going is (bikeshedding names aside):


1) Top-level Realm config method effecting all classes: 
```
Realm.setNamePolicy(RealmNamePolicy.LOWER_CASE_WITH_DASHES)

// Perhaps
Realm.setClassNamePolicy(RealmNamePolicy.LOWER_CASE_WITH_DASHES)
Realm.setFieldNamePolicy(RealmNamePolicy.LOWER_CASE_WITH_DASHES)
```
Perhaps on RealmConfiguration instead although I have a hard time seeing the same model class being used for two end points with different naming schemes.

2) Realm class annotation:
```
@RealmClassName("person") // Rename class
@RealmFieldNamePolicy(RealmFieldNamePolicy.IDENTITY) // Map all field names
[...]

Issue #1616, opened 2015-10-19
Support multiple model classes with the same name


cmelchior:
Right now it is not possible to create multiple Realm model classes with the same name, but in different packages. This will crash the annotation-processor with:

```
Error:(5, 8) error: Attempt to recreate a file for type io.realm.FooRealmProxy
```

Figure out a strategy for supporting this. Concerns are
1) Core requires a unique name
2) How to support this across multiple platforms

dalinaum:
:+1:

beeender:
With better object, bytecode manipulation will randomly choose one of the objects with same name to process.

Mayur-007:
i am stuck with same issues.
i am working on one big app and it contains lots of model classes with same name in different packages. please provide help as fast as possible.
thanks

mrtysn:
:thumbsup:

olibook:
👍

PR #5280, opened 2017-09-19
Support for overriding internal names using annotations

cmelchior:
Fixes https://github.com/realm/realm-java/issues/2476
Fixes https://github.com/realm/realm-java/issues/1616

**Note:** This PR is very much work in progress and just intended as a backup from a train ride and as mean to facilitate discussion around this feature.

This PR adds support for a special `@RealmName` annotation. This makes it possible to override what the internal name is in the Realm database which e.g. makes it easier to share schemas between languages without compromising the standard naming conventions in Java

it is also possible to apply a `RealmNamePolicy` for a class or module instead of being forced to manually override each and every field.

The basic implementation is a simple mapping scheme in Java before any relevant field or class name is sent to Object Store. As long as Object Store does not support a true dynamic mode, we cannot have this functionality living there.

```
// Example
// "name" just effect the element for which this is applied.
// "policy"  does the same as name, but uses an in-built converter. Default is NO_POLICY
// "recursive" defines if the policy should be applied to child elements. Default is true
@RealmName(name = "people", policy = RealmNamePolicy.LOWER_CASE_WITH_UNDERSCORE, recursive = true)
public class Persons extends RealmObject { // becomes people inside Realm
  public String firstName; // becomes first_name inside Realm
  public String lastName; // becomes last_name inside Realm
}
```

Especially, the following semantics are in place:

* Queries on Realm (typed Java model classes) still uses the names defined in Java
* Queries on DynamicRealm must use the mapped names
* Migrations and `RealmObjectSchema` must use the mapped names. This might seem really annoying, but the primary use case is sync, where migrations are automatic. 
* This does not have any impact on importing from JSON which must still use the Java names. As discussed in https://github.com/realm/realm-java/issues/1470 this is two different concerns.

TODO:
- [ ] Figure out naming for the annotation and related classes
- [ ] Determine the number of policies supported and any restrictions.
- [ ] Figure out if Policy should be recursive by default, which means we can remove the recursive param. It means that some rather exotic use cases become impossible.
- [ ] Implement query support
- [ ] Implement detection of conflicting policies for a class part of different modules.


Zhuinden:
I don't really get the "policy" thing (I know it's like from GSON, but...) I'd just expect something like a `@RealmField` annotation and if I want to set a different name, then set it.

Maybe that's just me though.

cmelchior:
@Zhuinden The problem with `@FieldName` is that it doesn't work classes or modules.

Originally I was playing around with overloading `@RealmClass(name = "custom")` and `@RealmField(name = "custom)`, but it feels awkward to use two different annotations for the same thing. Also, it left the question of what to do on modules.

Modules are a bit of a special case though and allowing `@RealmName` on there complicates things quite a lot, so we could also consider not allowing `@RealmName` there, but that means there is no place to define a naming policy for an entire Realm (as the RealmConfiguration is not available to the  annotation processor).

Zhuinden:
> @RealmClass(name = "custom") and @RealmField(name = "custom), but it feels awkward to use two different annotations for the same thing.

I think it is the distinction between `@Table/@Column` or `@JsonObject/@JsonField`.

cmelchior:
I guess you are right. So there is precedence for that 👍 
Having `@RealmField` would also be nice for other possible future enhancements. I guess we could overload modules the same way:

```
@RealmModule(policy = RealmNamePolicy.IDENTITY, allClasses = true)
public void MyModule {
}

@RealmClass(name = "custom)
public class Persons extends RealmObject {

  @RealmField(name = "bar)
  public String foo;
}
```

I kinda like that. It also means we can remove the `recursive` flag on e.g `@RealmModule` and `@RealmField`

Zhuinden:
@cmelchior The only tricky thing though is that you already have a `@RealmClass` annotation that you inherit from `RealmObject`!

cmelchior:
Yes, but that last one would take priority I would assume, but I haven't tested how the annotation processor actually behaves in that scenario

cmelchior:
Just for reference. This is how LoganSquare does it: https://github.com/bluelinelabs/LoganSquare/commit/41572aa599ec6deba268b3fa3928c1453d7c3b73

realm/realm-annotations-processor/src/test/resources/some/test/NamePoliceConflictingModuleDefinitionsForAllClasses.java
```diff
@@ -18,7 +18,7 @@
 import io.realm.annotations.RealmNamingPolicy;
 import some.test.AllTypes;
 
-public class NamingPolicyConflictingModules {
+public class NamePoliceConflictingModuleDefinitionsForAllClasses {
```
Zhuinden: Police!

realm/realm-annotations-processor/src/test/resources/some/test/NamePoliceConflictingModuleDefinitionsForMixedDefinitions.java
```diff
@@ -0,0 +1,37 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import io.realm.annotations.RealmModule;
+import io.realm.annotations.RealmNamingPolicy;
+import some.test.Simple;
+
+public class NamePoliceConflictingModuleDefinitionsForMixedDefinitions {
```
Zhuinden: Police!

realm/realm-annotations-processor/src/test/resources/some/test/NamePoliceConflictingModuleDefinitionsForNamedClasses.java
```diff
@@ -0,0 +1,37 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import io.realm.annotations.RealmModule;
+import io.realm.annotations.RealmNamingPolicy;
+import some.test.Simple;
+
+public class NamePoliceConflictingModuleDefinitionsForNamedClasses {
```
Zhuinden: Police?

realm-annotations/src/main/java/io/realm/annotations/RealmField.java
```diff
@@ -0,0 +1,41 @@
...
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Inherited;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Annotation used for meta data about a field persisted by Realm.
+ */
+@Retention(RetentionPolicy.CLASS)
```
Zhuinden: Any chance of RUNTIME?

cmelchior:
Ready for review @nhachicha 
It still isn't parsing CI, but any changes to fix that should not effect the semantics.

realm-annotations/src/main/java/io/realm/annotations/RealmNamingPolicy.java
```diff
@@ -0,0 +1,149 @@
...
+ *         Examples are "my😁" and "MY😁" which are both treated as one word.
+ *     </li>
+ *     <li>
+ *         Hungarian notation, i.e. variable names starting with lowercase "m" followed by uppercase
+ *         letter is stripped and not considered part of any word.
+ *         Example is "mFirstName" and "mFIRSTName" which becomes "First" and "Name.
+ *     </li>
+ * </ol>
+ * <p>
+ * Note, that changing the internal name does <i>NOT</i> effect importing data from JSON. The JSON
```
kneth: No comma

realm-annotations/src/main/java/io/realm/annotations/RealmNamingPolicy.java
```diff
@@ -0,0 +1,149 @@
...
+ * @see RealmField
+ */
+public enum RealmNamingPolicy {
+
+    /**
+     * No policy is applied. This policy will not override any policy set on a parent element, e.g.
+     * if set in {@link RealmClass#fieldNamingPolicy}, the module policy will still apply to field
+     * names.
+     * <p>
+     * If two modules disagree on the policy and one of them is {@code NO_POLICY}, the other one
```
kneth: "other one" -> "other"

realm/realm-annotations-processor/src/main/java/io/realm/processor/nameconverter/WordTokenizer.java
```diff
@@ -0,0 +1,145 @@
...
+ * https://docs.oracle.com/javase/specs/jls/se7/html/jls-3.html#jls-3.8
+ *
+ * In this implementation we treat word separators as any of the following:
+ * <ol>
+ *     <li>
+ *         Anytime a {@code _} or {@code $} is encountered.
+ *         Example is "_FooBar" or "_Foo$Bar" which both becomes "Foo" and "Bar".
+ *     </li>
+ *     <li>
+ *         Anytime your switch from a lower case character to a upper case character as
```
kneth: "your" -> "you"

realm/realm-annotations-processor/src/main/java/io/realm/processor/nameconverter/WordTokenizer.java
```diff
@@ -0,0 +1,145 @@
...
+ *         identified by a Character.isUpperCase(codepoint)` and `Character.isLowerCase(codepoint)`.
```
kneth: Missing backquote (after "by a")

realm/realm-annotations-processor/src/main/java/io/realm/processor/nameconverter/WordTokenizer.java
```diff
@@ -0,0 +1,145 @@
...
+ *         identified by `Character.isUpperCase(codepoint)` and `Character.isLowerCase(codepoint)`.
+ *         Example is "FOOBar" which becomes "FOO" and "Bar.
+ *     </li>
+ *     <li>
+ *         Some characters like emojiis are neither uppercase or lowercase characters, so they will
+ *         not trigger any of the above rules.
+ *         Examples are "my😁" and "MY😁" which are both treated as one word.
+ *     </li>
+ *     <li>
+ *         Hungarian notation, i.e. Strings starting with lowercase "m" followed by uppercase letter
```
kneth: "Strings" -> "strings"

realm-annotations/src/main/java/io/realm/annotations/RealmNamingPolicy.java
```diff
@@ -4,7 +4,7 @@
  * This enum defines the possible ways class and field names can be mapped from what is used in Java
  * to the name used internally in the Realm file.
  * <p>
- * Examples where this can be useful is e.g:
+ * Examples where this are useful:
```
Zhuinden: `this are`? That can't be right

realm-annotations/src/main/java/io/realm/annotations/RealmNamingPolicy.java
```diff
@@ -91,7 +91,7 @@
  *     </li>
  * </ol>
  * <p>
- * Note, that changing the internal name does <i>NOT</i> effect importing data from JSON. The JSON
+ * Note that changing the internal name does <i>NOT</i> effect importing data from JSON. The JSON
```
Zhuinden: Now I wonder if this is `affect` or `effect`.

I never know that.

realm-annotations/src/main/java/io/realm/annotations/RealmNamingPolicy.java
```diff
@@ -105,8 +105,8 @@
      * if set in {@link RealmClass#fieldNamingPolicy}, the module policy will still apply to field
      * names.
      * <p>
-     * If two modules disagree on the policy and one of them is {@code NO_POLICY}, the other one
-     * will be chosen without an error being thrown.
+     * If two modules disagree on the policy and one of them is {@code NO_POLICY}, the other will
```
Zhuinden: maybe `the other module's policy`

But what happens if there are 3 `@RealmModule`s and they all have conflicting policy? Which one will override `NO_POLICY`?

realm-annotations/src/main/java/io/realm/annotations/RealmNamingPolicy.java
```diff
@@ -0,0 +1,149 @@
+package io.realm.annotations;
+
+/**
+ * This enum defines the possible ways class and field names can be mapped from what is used in Java
+ * to the name used internally in the Realm file.
+ * <p>
+ * Examples where this can be useful is e.g:
```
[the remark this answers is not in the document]
Zhuinden: This can't be right!! 😄

realm-annotations/src/main/java/io/realm/annotations/RealmNamingPolicy.java
cmelchior: For each class you find all modules it is part of, then you compare the policy between all these 3, so it will throw. The implementation is slightly different, but the effect is the same.

realm-annotations/src/main/java/io/realm/annotations/RealmNamingPolicy.java
cmelchior: https://www.vocabulary.com/articles/chooseyourwords/affect-effect/

It should be affect I think :)

realm/realm-annotations-processor/src/test/resources/some/test/NamePolicyModuleDefaults.java
```diff
@@ -0,0 +1,28 @@
...
+ * limitations under the License.
+ */
+package some.test;
+
+import io.realm.RealmObject;
+import io.realm.annotations.RealmField;
+import io.realm.annotations.RealmNamingPolicy;
+
+/**
+ * Class with only a custom name
```
nhachicha: where's the custom name definition? or do you mean the module that uses this class use a custom name?

realm/realm-library/src/androidTest/java/io/realm/CustomRealmNameTests.java
```diff
@@ -0,0 +1,220 @@
...
+
+        // Backlinks do not exist as internal fields that can be queried
+    }
+
+
+    @Test
+    public void dynamicQueryWithInternalNames() {
+        // Backlink queries not supported on dynamic queries
+        RealmResults<DynamicRealmObject> results = dynamicRealm.where(ClassWithPolicy.CLASS_NAME)
+                .equalTo(ClassWithPolicy.FIELD_CAMEL_CASE, "foo") // Java name in model class
```
nhachicha: Java name? shouldn't be internal name for Dynamic Realm. same for below

realm/realm-library/src/androidTest/java/io/realm/entities/realmname/CustomRealmNamesModule.java
```diff
@@ -0,0 +1,16 @@
+package io.realm.entities.realmname;
```
nhachicha: licence

realm/realm-library/src/androidTest/java/io/realm/entities/realmname/DefaultPolicyFromModule.java
```diff
@@ -0,0 +1,7 @@
+package io.realm.entities.realmname;
```
nhachicha: licence

realm/realm-library/src/main/java/io/realm/internal/ColumnInfo.java
```diff
@@ -165,22 +167,36 @@ public void copyFrom(ColumnInfo src) {
...
                 buf.append(entry.getKey()).append("->").append(entry.getValue());
                 commaNeeded = true;
             }
+            buf.append("]");
+        }
+        if (indicesFromColumnNames != null) {
+            buf.append(", InternalFieldNames=[");
+            boolean commaNeeded = false;
+            for (Map.Entry<String, ColumnDetails> entry : indicesFromColumnNames.entrySet()) {
+                if (commaNeeded) { buf.append(","); }
```
nhachicha: `commaNeeded = true;` should be inside the if?

realm-annotations/src/main/java/io/realm/annotations/RealmField.java
```diff
@@ -0,0 +1,40 @@
...
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Inherited;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Annotation used on fields in Realm model classes. It describes metadata about the field.
+ */
+@Retention(RetentionPolicy.RUNTIME)
```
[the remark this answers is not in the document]
Zhuinden: Because I asked! 😄 

The other annotations like `@RealmModule` and `@Index` and `@Required` were changed to `RUNTIME` so that they can be accessed via reflection, this way you can read this data without having to write your own annotation processor.

Like here https://github.com/realm/realm-java/issues/5344

realm-annotations/src/main/java/io/realm/annotations/RealmNamingPolicy.java
```diff
@@ -0,0 +1,149 @@
...
+ *      <li>
+ *          If applied to {@link RealmClass#fieldNamingPolicy} all fields in that class will be
+ *          affected. This will override any field naming policy specified on a module.
+ *      </li>
+ * </ul>
+ * <p>
+ * An example of this:
+ * <pre>
+ * {@code
+ * \@RealmClass(name = "__person", fieldNamingPolicy = RealmNamingPolicy.LOWER_CASE_WITH_UNDERSCORES)
```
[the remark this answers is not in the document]
Zhuinden: If we already inherit it from `extends RealmObject`, then does it get overwritten by specifying the annotation with different params again?

realm-annotations/src/main/java/io/realm/annotations/RealmNamingPolicy.java
```diff
@@ -0,0 +1,149 @@
...
+ * When automatically converting Java variable names, each variable name is normalized by splitting
+ * it into a list of words that are then joined using the rules of the target format. The following
+ * heuristics are used for determining what constitutes a "word".
+ * <ol>
+ *     <li>
+ *         Anytime a {@code _} or {@code $} is encountered.
+ *         Example is "_FirstName" or "_First$Name" which both becomes "First" and "Name".
+ *     </li>
+ *     <li>
+ *         Anytime your switch from a lower case character to a upper case character as
```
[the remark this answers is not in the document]
Zhuinden: Also, anytime `your` switch -> `you`

realm-annotations/src/main/java/io/realm/annotations/RealmNamingPolicy.java
```diff
@@ -0,0 +1,149 @@
...
+    PASCAL_CASE,
+
+    /**
+     * The name in the Java model class is converted lowercase with each word separated by {@code _}.
+     * This is the default naming scheme in C++.
+     * <p>
+     * Examples: "firstName", "FirstName", "mFirstName", "FIRST_NAME", "First$Name" all becomes
+     * "first_name".
+     */
+    LOWER_CASE_WITH_UNDERSCORES
```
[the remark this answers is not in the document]
Zhuinden: I've only heard it as `UNDER_SCORE` 😮

realm/realm-library/src/androidTest/java/io/realm/CustomRealmNameTests.java
Zhuinden: If the class no longer exists, then you can't obtain any field-specific / class-specific policy configuration for it, though.

realm/realm-library/src/main/java/io/realm/internal/fields/DynamicFieldDescriptor.java
```diff
@@ -34,7 +37,7 @@
      * Build a dynamic field descriptor for the passed field description string.
      *
      * @param table the start table.
-     * @param fieldDescription the field description.
+     * @param fieldDescription the field description using internal column columns.
```
Zhuinden: column columns?

realm-annotations/src/main/java/io/realm/annotations/RealmNamingPolicy.java
```diff
@@ -0,0 +1,149 @@
...
+ * <ul>
+ *      <li>
+ *          To support two model classes with the same simple name but in different packages.
+ *      </li>
+ *      <li>
+ *          To make it easier to work with cross platform schemas as naming conventions are different.
+ *      </li>
+ *      <li>
+ *          To use a Java class name that is longer than the 57 character limit enforced by Realm.
+ *      </li>
```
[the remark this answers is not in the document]
cmelchior: Good point. Adding

realm-annotations/src/main/java/io/realm/annotations/RealmNamingPolicy.java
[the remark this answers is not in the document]
cmelchior: @Zhuinden Yes, the unit tests does that.

@nhachicha Not sure what you mean? Do you mean we should create a new annotation for this? Originally I had a generic `@RealmName` annotation, but that caused issues with defining policies in a sensible manner as e.g. class policy doesn't make sense for field names. Adding 3 separate annotations just for this also seems a bit overkill?

realm-annotations/src/main/java/io/realm/annotations/RealmNamingPolicy.java
```diff
@@ -0,0 +1,149 @@
...
+ * }
+ * }
+ * </pre>
+ * <p>
+ * Choosing an internal name that differs from the name used in the Java model classes has the
+ * following implications:
+ * <ul>
+ *      <li>
+ *          Queries on {@code DynamicRealm} must use the internal name. Queries on normal {@code Realm}
+ *          instances should continue to use the name as it is defined in the Java class.
```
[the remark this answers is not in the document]
cmelchior: It's not optional for typed Realms. You cannot use internal names for those, but I can clarify that.

realm-annotations/src/main/java/io/realm/annotations/RealmNamingPolicy.java
```diff
@@ -0,0 +1,149 @@
...
+ *      </li>
+ *      <li>
+ *          Migrations must use the internal name when creating classes and fields.
+ *      </li>
+ *      <li>
+ *          Schema errors reported will use the internal names.
+ *      </li>
```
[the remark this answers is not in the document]
cmelchior: I suspect that will just be confusing at this point. The annotation processor doesn't allow you to set `@RealmField` on backlink fields so you should never get into that situation.

realm-annotations/src/main/java/io/realm/annotations/RealmNamingPolicy.java
```diff
@@ -0,0 +1,149 @@
...
+ * </ul>
+ * <p>
+ * When automatically converting Java variable names, each variable name is normalized by splitting
+ * it into a list of words that are then joined using the rules of the target format. The following
+ * heuristics are used for determining what constitutes a "word".
+ * <ol>
+ *     <li>
+ *         Anytime a {@code _} or {@code $} is encountered.
+ *         Example is "_FirstName" or "_First$Name" which both becomes "First" and "Name".
```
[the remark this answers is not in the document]
cmelchior: Both _ and $ are replaced, but I'll make an better example

realm-annotations/src/main/java/io/realm/annotations/RealmNamingPolicy.java
```diff
@@ -0,0 +1,149 @@
...
+ *     <li>
+ *         Anytime your switch from more than one uppercase character to a lower case one. The last
+ *         upper case letter is assumed to be part of the next word. This is identified by using
+ *         {@link Character#isUpperCase(int)} and {@link Character#isLowerCase(int)}.
+ *         Example is "FIRSTName" which becomes "FIRST" and "Name.
+ *     </li>
+ *     <li>
+ *         Some characters like emojiis are neither uppercase nor lowercase characters, so they will
+ *         be part of the current word.
+ *         Examples are "my😁" and "MY😁" which are both treated as one word.
```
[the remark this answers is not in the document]
cmelchior: Yes 😛

realm-annotations/src/main/java/io/realm/annotations/RealmNamingPolicy.java
```diff
@@ -0,0 +1,149 @@
...
+ *     </li>
+ *     <li>
+ *         Hungarian notation, i.e. variable names starting with lowercase "m" followed by uppercase
+ *         letter is stripped and not considered part of any word.
+ *         Example is "mFirstName" and "mFIRSTName" which becomes "First" and "Name.
+ *     </li>
+ * </ol>
+ * <p>
+ * Note that changing the internal name does <i>NOT</i> affect importing data from JSON. The JSON
+ * data must still follow the names as defined in the Realm Java class.
```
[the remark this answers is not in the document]
cmelchior: Yes, I will add a code snippet using GSON 👍

realm-annotations/src/main/java/io/realm/annotations/RealmNamingPolicy.java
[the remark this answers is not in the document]
cmelchior: Actually, https://en.wikipedia.org/wiki/Snake_case allows the individual words to be the upper case as well. This one actually comes from GSON: https://google.github.io/gson/apidocs/com/google/gson/FieldNamingPolicy.html and seems like a reasonable cross-platform naming schema with well-defined semantics.

The problem with automatic naming schemes is that there are so many variants that it is impossible to cover them all (which I guess is also why some projects just disallow them all together).

I originally had a few more but removed them to focus on just the few normal patterns you see in Java/Swift/.NET/JS and C++.

realm/realm-annotations-processor/src/main/java/io/realm/processor/ClassMetaData.java
```diff
@@ -508,6 +548,22 @@ private boolean categorizeField(Element element) {
         return true;
     }
 
+    private String getInternalFieldName(VariableElement field, NameConverter defaultConverter) {
+        RealmField nameAnnotation = field.getAnnotation(RealmField.class);
+        if (nameAnnotation != null) {
+            String declaredName = nameAnnotation.name();
+            if (!declaredName.equals("")) {
```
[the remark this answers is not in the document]
cmelchior: We actually moved all of that logic to OS instead of having it in the annotation processor, but having the check will not only improve the error message, but also throw on compile time. I'll add it.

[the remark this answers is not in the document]
cmelchior: Actually, in this place. It is just field names. Field names do not have any restrictions

realm/realm-annotations-processor/src/main/java/io/realm/processor/Utils.java
```diff
@@ -209,7 +216,23 @@ public static boolean isRealmModel(Element field) {
      * @return {@code true} if a given type is {@code RealmModel}, {@code false} otherwise.
      */
     public static boolean isRealmModel(TypeMirror type) {
-        return typeUtils.isAssignable(type, realmModel);
+        // Not sure what is happening here, but typeUtils.isAssignable("Foo", realmModel)
```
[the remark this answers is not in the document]
cmelchior: I don't think this was ever tested before. It started happening after I added a reference to a non-existing class in test file.

realm/realm-annotations-processor/src/main/java/io/realm/processor/Utils.java
```diff
@@ -209,7 +216,23 @@ public static boolean isRealmModel(Element field) {
...
+        // returns true even if Foo doesn't exist. No idea why this is happening.
+        // For now punt on the problem and check the direct supertype which should be either
+        // RealmObject or RealmModel.
+        // Original implementation: `return typeUtils.isAssignable(type, realmModel);`
+        //
+        // Theory: It looks like if `type` has the internal TypeTag.ERROR (internal API) it
+        // automatically translate to being assignable to everything. Possible some Java Specification
+        // rule taking effect. In our case, however we can do better since all Realm classes
+        // must be in the same compilation unit, so we should be able to look the type up.
+        for (TypeMirror typeMirror : typeUtils.directSupertypes(type)) {
```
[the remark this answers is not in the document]
cmelchior: Actually, it should be allowed to do something like `public interface MyInterface extends RealmModel`, and I suspect that this check will not correctly validate that case. I'll take another look

realm/realm-library/src/androidTest/java/io/realm/CustomRealmNameTests.java
cmelchior: DynamicRealms has to work with the assumption that no java model classes exists. That this test case actually references `ClassWithPolicy` is just so it is easier to maintain the test case.

But you are correct, DynamicRealms do not understand the concept of name policies as that concept isn't relevant to them.

realm/realm-library/src/main/java/io/realm/internal/ColumnInfo.java
cmelchior: No, in that case, it would never be set to the value `true`.

realm/realm-library/src/androidTest/java/io/realm/entities/realmname/DefaultPolicyFromModule.java
cmelchior: Done

realm/realm-library/src/androidTest/java/io/realm/entities/realmname/CustomRealmNamesModule.java
cmelchior: Done

realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmProcessor.java
```diff
@@ -242,6 +232,31 @@ private boolean processModules(RoundEnvironment roundEnv) {
...
     }
 
+    private boolean createProxyClassFiles(TypeMirrors typeMirrors) {
+        boolean success = true;
+        for (ClassMetaData metadata : classCollection.getClasses()) {
+            RealmProxyInterfaceGenerator interfaceGenerator = new RealmProxyInterfaceGenerator(processingEnv, metadata);
+            try {
+                interfaceGenerator.generate();
+            } catch (IOException e) {
+                Utils.error(e.getMessage(), metadata.getClassElement());
```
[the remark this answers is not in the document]
cmelchior: Done

realm/realm-annotations-processor/src/main/java/io/realm/processor/ModuleMetaData.java
```diff
@@ -112,19 +240,70 @@ public boolean generate(Set<? extends Element> clazzes) {
...
+                            "They same policy must be used.",
+                    moduleInfo.qualifiedModuleClassName,
+                    otherModuleInfo.qualifiedModuleClassName,
+                    (className != null) ? " for " + className : "",
+                    classPolicy,
+                    otherClassPolicy));
+            foundErrors = true;
+        }
+
+        // Check field naming policy
```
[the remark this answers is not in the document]
cmelchior: In this case I thought it was better to report class + field errors in one.

realm/realm-annotations-processor/src/main/java/io/realm/processor/Utils.java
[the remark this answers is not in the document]
Zhuinden: One day we'll be able to inherit from abstract class that has no fields :smile:

(just rambling heed me no mind)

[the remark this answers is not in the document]
cmelchior: Yep, but I disregarded that for now 😛

realm/realm-annotations-processor/src/main/java/io/realm/processor/Utils.java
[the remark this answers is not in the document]
cmelchior: Rewrote the unit test and reverted this change. It actually showed a bug elsewhere that was fixed.

realm/realm-annotations-processor/src/main/java/io/realm/processor/ModuleMetaData.java
```diff
@@ -69,37 +102,132 @@ public boolean generate(Set<? extends Element> clazzes) {
...
+        ModulePolicyInfo moduleInfo = new ModulePolicyInfo(qualifiedModuleClassName, classNamePolicy, fieldNamePolicy);
+
+        // The difference between `allClasses` and a list of classes is a bit tricky at this stage
+        // as we haven't processed the full list of classes yet. We therefore need to treat
+        // each case specifically :(
+        // We do not compare against the default module as it is always configured correctly
+        // with NO_POLICY, meaning it will not trigger any errors.
+        if (moduleAnnotation.allClasses()) {
+            // Check for conflicts with other modules with `allClasses` set. Only need to
+            // check the first since other conflicts would have been detected in an earlier
```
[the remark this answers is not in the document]
cmelchior: Hmm thinking about this, you might be right. E.g. the following sequence of illegal modules would not be detected. 

module1: NO_POLICY
module2: PASCAL_CASE
module3: NO_POLICY
module4: CAMEL_CASE

I'll compare against all previous modules.

realm/realm-annotations-processor/src/main/java/io/realm/processor/ModuleMetaData.java
```diff
@@ -69,37 +102,132 @@ public boolean generate(Set<? extends Element> clazzes) {
...
+                    ModulePolicyInfo otherModuleInfo = globalModuleInfo.iterator().next();
+                    if (checkAndReportPolicyConflict(moduleInfo, otherModuleInfo)) {
                         return false;
                     }
-                    classes.add(metadata);
                 }
+
+                // Check that this specific class isn't conflicting with another module
+                // specifically mentioning it using `classes = { ... }`
+                ModulePolicyInfo otherModuleInfo = classSpecificModuleInfo.get(qualifiedClassName);
```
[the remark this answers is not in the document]
cmelchior: Yes, given the same reason as above.

realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmProxyMediatorGenerator.java
```diff
@@ -177,7 +180,7 @@ private void emitGetSimpleClassNameMethod(JavaWriter writer) throws IOException
         emitMediatorShortCircuitSwitch(new ProxySwitchStatement() {
             @Override
             public void emitStatement(int i, JavaWriter writer) throws IOException {
-                writer.emitStatement("return %s.getSimpleClassName()", qualifiedProxyClasses.get(i));
+                writer.emitStatement("return \"%s\"", internalClassNames.get(i));
```
Zhuinden: That does seem more Proguard-friendly.

cmelchior:
All comments addressed @nhachicha 
I added a runtime check that throws if conflicts between internal names are detected among different modules (not possible to check at compile time). It isn't really possible to create a unit test for this, but I verified that we correctly throw an exception using our `moduleExample`.

realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmProxyMediatorGenerator.java
cmelchior: Not to mention correct 😉 The previous implementation did not work if you renamed the class name using `@RealmClass(name = "othername")`.

realm-annotations/src/main/java/io/realm/annotations/RealmNamingPolicy.java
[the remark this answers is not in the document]
nhachicha: @cmelchior No I didn't mean to add a new annotation, maybe just for the sake of this example use `RealmModel` instead?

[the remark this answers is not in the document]
cmelchior: Ah like that. Sure, I can change it to that.

PR #5285, opened 2017-09-19
Cache the dowloaded core on CI

beeender:


beeender:
@emanuelez or shall we create a special dir for it on CI?

emanuelez:
that should be fine

PR #5287, opened 2017-09-19
Minimize overhead on CI for PR builds.

cmelchior:
This PR changes PR's so they only build for `armeabi-v7a` and run integration tests against the `ObjectServer` variant. `master` still runs the full test suite.

Nesting was wrong as well so fixed.
Relevant changes are in https://github.com/realm/realm-java/commit/78a2af9af4f3e73a8edec4f57143f88c53cf3479

Build time for PR's have been reduced to ~22 minutes, so at least a 50% reduction. We can probably get that lower.

@emanuelez What would it take to run integration tests on an x86 emulator instead of on a device?

TODO
- [x] Make build work
- [x] Measure savings

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -9,103 +9,103 @@ try {
     timeout(time: 1, unit: 'HOURS') {
       // Allocate a custom workspace to avoid having % in the path (it breaks ld)
       ws('/tmp/realm-java') {
-	stage('SCM') {
-	  checkout([
-		     $class: 'GitSCM',
-		    branches: scm.branches,
-		    gitTool: 'native git',
-		    extensions: scm.extensions + [
-		      [$class: 'CleanCheckout'],
-		      [$class: 'SubmoduleOption', recursiveSubmodules: true]
-		    ],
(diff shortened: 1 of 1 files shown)


emanuelez:
For core and OS, we use an emulator running inside a Docker container. Unfortunately, while we can run an x86 emulator in Docker, it won't be able to use HAVC, making the emulation quite slow. For this reason, core and OS use an ARM emulator, since it covers the most common use-case.

Jenkinsfile
```diff
@@ -9,103 +9,115 @@ try {
...
+                userRemoteConfigs: scm.userRemoteConfigs
+               ])
+        }
+
+        // Toggles for PR vs. Master builds.
+        // For PR's, we just build for arm-v7a and run unit tests for the ObjectServer variant
+        // A full build is done on `master`.
+        // TODO Once Android emulators are available on all nodes, we can switch to x86 builds
+        // on PR's for even more throughput.
+        def ABIs = null
```
zaki50: We must use empty string instead of `null`. `null` is not expanded to empty string in `GString`.

PR #5311, opened 2017-09-22
Fix change listener getting GC'ed

cmelchior:
Fixed example not working correctly


examples/objectServerExample/src/main/java/io/realm/examples/objectserver/CounterActivity.java
```diff
@@ -131,6 +132,7 @@ protected void onStop() {
         }
         closeRealm();
         user = null;
+        counter = null;
```
nhachicha: showing `removeAllChangeListeners` or `removeChangeListener` is also a good idea instead ? also do we want to call `user.logout` as well?

cmelchior: It is on purpose I'm not logging the user out, since there is a manual logout button in the actionbar. It means the next time you open the app you go directly to this Activity.

PR #5329, opened 2017-09-26
Array of Primitives: RealmObjectSchema support

cmelchior:
This adds support for primitive lists to `RealmObjectSchema

- Going around Object Store and directly to Core to dynamically add primitive lists during migrations. Not ideal, but using Object Store constants where possible.
- Refactored `Table.cpp#nativeConvertColumnToNotNullable and` Table.cpp#nativeConvertColumnToNullable` so they work on both normal tables and sub tables. This is the biggest change, so please focus review there.
- Added bunch of tests. 

**Open questions:**
Is `RealmObjectSchema.addRealmListField(String name, Class<?> class)` the best name? We already have `RealmObjectSchema.addRealmListField(String name, RealmObjectSchema class)`, but is the overload weird? Another alternative could be `addPrimitiveRealmListField(name, class)` or `addPrimitiveListField(name, class)`? For the dynamic API, we have `getList(string fieldName)`. This will either need an overload with `getList(String fieldName, Class<?>` or `getPrimitiveList(String fieldName, Class<?> clazz)`.

Since these lists behavior is different from lists with Realm models, I would be tempted to go with:

- `RealmObjectSchema.addPrimitiveListField(String fieldName, Class<?> elementType);
- `DynamicRealmObject.getPrimitiveList(String fieldName, Class<?> elementType);

or

- `RealmObjectSchema.addValueListField(String fieldName, Class<?> elementType);
- `DynamicRealmObject.getValueList(String fieldName, Class<?> elementType);

`ValueList` is probably more technically correct as we also support more types than just the primitives, even though Date technically isn't a value type. On the other hand, I suspect that `primitive type` is more familiar to people than `value type`, 

So 3 proposals:

1) Overload methods:

- `RealmObjectSchema.addRealmListField(String fieldName, Class<?> elementType);`
- `RealmObjectSchema.addRealmListField(String fieldName, RealmObjectSchema objectSchema);`
- `RealmList<E> list = DynamicRealmObject.getList(String fieldName, Class<?> elementType);`
- `RealmList<DynamicRealmObject> list = DynamicRealmObject.getList(String fieldName);`

2) Use `primitive`:

- `RealmObjectSchema.addPrimitiveListField(String fieldName, Class<?> elementType);`
- `RealmObjectSchema.addRealmListField(String fieldName, RealmObjectSchema objectSchema);`
- `RealmList<E> list = DynamicRealmObject.getPrimitiveList(String fieldName, Class<E> elementType);`
- `RealmList<DynamicRealmObject> list = DynamicRealmObject.getList(String fieldName);`

3) Use `value`:

- `RealmObjectSchema.addValueListField(String fieldName, Class<?> elementType);`
- `RealmObjectSchema.addRealmListField(String fieldName, RealmObjectSchema objectSchema);`
- `RealmList<E> list = DynamicRealmObject.getValueList(String fieldName, Class<E> elementType);`
- `RealmList<DynamicRealmObject> list = DynamicRealmObject.getList(String fieldName);`

Thoughts?

TODO
- [x] Fix currently broken unit tests (changing nullability doesn't seem to take effect).
- [x] Update `IndexFieldType` with primitive list types
- [x] Update `InvalidIndexFieldType` with primitive list types
- [x] Update `InvalidPrimaryKeyFieldType` with primitive list types


cmelchior:
Ready for review @realm/java . Disregard the CI failure for now. It looks related to the Core fix incoming and should not be related to the functionality here.

Especially interested in hearing your opinion on the API methods.

Might as well summon @Zhuinden as well 😛

Zhuinden:
I wanted to answer this about 12 hours ago but I was hanging out at the Realm World Tour and snatching T-shirts 😄 

-------

The tricky thing with the names is that if the current method was called `addLinkList` and not `addRealmListField`, then `addValueList` wouldn't be an unprecedented addition to the API.

From the user standpoint, it's just a `RealmList<T>`, so I don't think they think about whether it is a primitive list or a link list.

That said, currently the `addRealmListField` method takes `objectSchema.addRealmListField(field.getName(), linkedObjectSchema);`

In which case you'd have two overloads, one that takes a `Class<?>` and one that takes a `RealmObjectSchema`.

This however works well only if you can provide **both** `Class<? extends RealmModel>` and `Class<?>` where ? is value type.

Vaguely this would mean that the overload should default its behavior over to the `RealmObjectSchema` , although it **does** seem to be possible...

``` java
    /**
     * Adds a new field that references a {@link RealmList}.
     *
     * @param fieldName name of the field to add.
     * @param elementType class for the value type, or Realm type being referenced.
     * @return the updated schema.
     * @throws IllegalArgumentException if the field name is illegal or a field with that name already exists.
     * @throws UnsupportedOperationException if this {@link RealmObjectSchema} is immutable.
     */
    public abstract RealmObjectSchema addRealmListField(String fieldName, Class<?> elementType);
```

and

``` java
    private boolean isValidForValueList(Class<?> fieldType) {
        return isFieldRegularObjectType(fieldType) || isPrimitiveObjectWrapper(fieldType);
    }
    
    private boolean isFieldRegularObjectType(Class<?> fieldType) {
        return fieldType == String.class || fieldType == Date.class || fieldType == byte[].class;
    }

    private boolean isPrimitiveObjectWrapper(Class<?> fieldType) {
        return fieldType == Boolean.class //
                || fieldType == Byte.class || fieldType == Short.class || fieldType == Integer.class || fieldType == Long.class //
                || fieldType == Float.class || fieldType == Double.class;
    }

    @Override
    public RealmObjectSchema addRealmListField(String fieldName, Class<?> elementType) {
        checkLegalName(fieldName);
        checkFieldNameIsAvailable(fieldName);
        if(RealmModel.class.isAssignableFrom(fieldType)) {
            // this is the class that belongs to Realm, so it should be added as a link.
            String tableName = realm.getConfiguration().getSchemaMediator().getSimpleClassName((Class<? extends RealmModel>)elementType); // FIXME is this trickier than needed?
            RealmObjectSchema objectSchema = schema.get(tableName);
            addRealmListField(fieldName, objectSchema);
            return this;
        }
        if(isValidForValueList(elementType)) {
            // this is the class for a value list.
            FieldMetaData metadata = SUPPORTED_SIMPLE_FIELDS.get(elementType);
            table.addColumn(metadata.listType, fieldName, metadata.defaultNullable);
            return this;
        }
        throw new IllegalArgumentException(String.format(Locale.US,
                         "RealmList does not support lists with this type: %s(%s)",
                         fieldName, elementType));
    }
```

cmelchior:
> This however works well only if you can provide both Class<? extends RealmModel> and Class<?> where ? is value type. Vaguely this would mean that the overload should default its behavior over to the RealmObjectSchema , although it does seem to be possible...

It doesn't really work that way.

A migration cannot reference any `RealmModel` class as it might not exist. Which is why these methods take `RealmObjectSchema` directly instead of `Class<? extends RealmModel>`

Zhuinden:
@cmelchior eh, I should have double-checked what `addRealmListField()` does for a `null` RealmObjectSchema, but it'll just crash when calling `objectSchema.getClassName()`.

``` java
    @Override
    public RealmObjectSchema addRealmListField(String fieldName, RealmObjectSchema objectSchema) {
        checkNotNull(objectSchema); // <-- !!!
        checkLegalName(fieldName);
        checkFieldNameIsAvailable(fieldName);
        table.addColumnLink(RealmFieldType.LIST, fieldName, realm.sharedRealm.getTable(Table.getTableNameForClass(objectSchema.getClassName())));
        return this;
    }
```

and

``` java
        if(RealmModel.class.isAssignableFrom(elementType)) {
            // this is a class that belongs to Realm, so it should be added as a link.
            // consider the fact that elementType.getSimpleName() can be obfuscated.
            String tableName = realm.getConfiguration().getSchemaMediator().getSimpleClassName((Class<? extends RealmModel>)elementType); // FIXME is this trickier than needed?
            RealmObjectSchema objectSchema = schema.get(tableName);
            if(objectSchema == null) {
                throw new IllegalArgumentException("The class [" + tableName + "] is not part of the schema."); // <-- !!!
            }
            return addRealmListField(fieldName, objectSchema);
        }
```

would solve this?

cmelchior:
> But I still think if you do have the RealmModel and you provide its class, then Realm should provide the default behavior of trying to look it up, instead of failing immediately. Otherwise, the name addRealmListField is a bit misleading. 

Sorry, that isn't going to happen. The problem in a Migration is that whatever `RealmModel` class you are referencing might not contain the fields you expect. E.g. if you go from V1->V2->V3, then the model class itself is already V3, so whatever logic you are applying will most likely fail. This the reason we only reference RealmSchemaObject's. Because they are a dynamic representation of what the schema looked liked at any given point in time. Only requirement is that it should match the RealmModel class when the migration ends, but in-between it can have any number of states.

Also, I'm not sure it is misleading. You are still adding a `RealmList<X>` field, you are just prevented from referencing the type if it is a RealmModel type.

Zhuinden:
> Because they are a dynamic representation of what the schema looked liked at any given point in time. Only requirement is that it should match the RealmModel class when the migration ends, but in-between it can have any number of states.

@cmelchior oh so you mean that if you are at VERSION 8, and you are trying to do a migration from VERSION 5 to VERSION 6, then if you hard-coded any dependency on the existing RealmModels then there is a possibility that from VERSION 5 to VERSION 6 providing the class would work, but from VERSION 5 to VERSION 8 it won't work anymore! So your migration code will break over time. 😭 

Yeah, now I see why that wouldn't be reliable. I guess I didn't think that through, sorry.

Zhuinden:
Anyways, I think if you name the `elementType` to `primitiveType` on the parameter list, then the usage will be clear.

My initial guess, although I am starting to distrust myself:

> RealmObjectSchema.addRealmListField(String fieldName, Class<?> primitiveType);

> RealmObjectSchema.addRealmListField(String fieldName, RealmObjectSchema objectSchema);

> RealmList<E> list = DynamicRealmObject.getValueList(String fieldName, Class<?> primitiveType);

> RealmList<DynamicRealmObject> list = DynamicRealmObject.getList(String fieldName);

cmelchior:
So you would be okay with the asymmetry between `RealmObjectSchema` and `DynamicRealmObject?  As in only `addRealmListField()` vs `getList/getValueList()` on DynamicRealmObject? Also shouldn't the parameter be `valueType` if the method is `getValueList()` ?

Zhuinden:
@cmelchior Okay so I started brainstorming about this with the following scheme and "what makes most sense to use"

``` java
    @Override
        RealmObjectSchema dogSchema = schema.get("Dog");
        dogSchema.addRealmListField("lovedBy", schema.get("Cat"));
        dogSchema.addRealmListField("names", String.class);
        dogSchema.transform(obj -> {
            String name = obj.getString("name");
            RealmList<DynamicRealmObject> models = obj.getList("lovedBy");
            RealmList<String> names = obj.getList("names", String.class);
            //val names : RealmList<String> = obj.getList<String>("names"); // kotlin reified?

            RealmList<String> names = obj.getPrimitiveList("names", String.class);
            //RealmList<String> names = obj.<String>getPrimitiveList("names");
            RealmList<String> names = obj.getValueList("names", String.class);
            //RealmList<String> names = obj.<String>getValueList("names");
            //RealmList<String> names = obj.getSimpleList("names", String.class); // from external perspective

            names.add(name);
        })
    }
```

With that in mind, I have to retract my vote for `DynamicRealmObject.getValueList(String fieldName, Class<?> primitiveType);` because it is cognitive overhead to need to know that you need a different method for primitives. `getList()` without additional `class` argument already returns `DynamicRealmObject` which is quite specific.

So I revise my vote to be:

``` java
 RealmObjectSchema.addRealmListField(String fieldName, Class<?> primitiveType);

 RealmObjectSchema.addRealmListField(String fieldName, RealmObjectSchema objectSchema);

 RealmList<E> list = DynamicRealmObject.getList(String fieldName, Class<?> primitiveType);

 RealmList<DynamicRealmObject> list = DynamicRealmObject.getList(String fieldName);
```

Zhuinden:
(one thing I noticed is that in the proposal, the overloaded method is

``` java
RealmList<E> list = DynamicRealmObject.getList(String fieldName, Class<?> elementType);
```

but the next versions are

``` java
RealmList<E> list = DynamicRealmObject.getPrimitiveList(String fieldName, Class<E> elementType);
RealmList<E> list = DynamicRealmObject.getValueList(String fieldName, Class<E> elementType);
```

the difference being the wildcard vs the `E` generic parameter. I assume the overload would also use `Class<E>` so I edited my post above accordingly.)

realm/realm-library/src/main/cpp/io_realm_internal_Table.cpp
```diff
@@ -199,7 +217,106 @@ JNIEXPORT jboolean JNICALL Java_io_realm_internal_Table_nativeIsColumnNullable(J
...
+    std::string column_name = table->get_column_name(column_index);
+    DataType column_type = table->get_column_type(column_index);
+    bool is_subtable = table->get_column_type(column_index) == DataType::type_Table;
+    std::string tmp_column_name;
+    size_t j = 0;
+    while (true) {
+        std::ostringstream ss;
+        ss << std::string("__TMP__") << j;
+        std::string str = ss.str();
+        StringData sd(str);
```
beeender: rename this to `tmp_column_name`

realm/realm-library/src/main/cpp/io_realm_internal_Table.cpp
```diff
@@ -199,7 +217,106 @@ JNIEXPORT jboolean JNICALL Java_io_realm_internal_Table_nativeIsColumnNullable(J
...
+    }
+}
+
+// Creates the new column into which all old data is copied when switching between nullable and non-nullable.
+static void create_new_column(Table* table, size_t column_index, bool nullable)
+{
+    std::string column_name = table->get_column_name(column_index);
+    DataType column_type = table->get_column_type(column_index);
+    bool is_subtable = table->get_column_type(column_index) == DataType::type_Table;
+    std::string tmp_column_name;
```
beeender: remove this

realm/realm-library/src/main/cpp/io_realm_internal_Table.cpp
```diff
@@ -199,7 +217,106 @@ JNIEXPORT jboolean JNICALL Java_io_realm_internal_Table_nativeIsColumnNullable(J
...
+            if (is_subtable) {
+                DataType original_type = table->get_subdescriptor(column_index)->get_column_type(0);
+                table->insert_column(column_index, type_Table, sd, true);
+                table->get_subdescriptor(column_index)->add_column(original_type, ObjectStore::ArrayColumnName, nullptr, nullable);
+            }
+            else {
+                table->insert_column(column_index, column_type, sd, nullable);
+            }
+
+            tmp_column_name = ss.str();
```
beeender: remove this, it seems we don't need this anymore.

realm/realm-library/src/main/cpp/io_realm_internal_Table.cpp
```diff
@@ -208,248 +325,213 @@ JNIEXPORT void JNICALL Java_io_realm_internal_Table_nativeConvertColumnToNullabl
...
-                    // checked previously
-                    break;
-                case type_OldDateTime:
-                    // not used
-                    ThrowException(env, UnsupportedOperation, "The old DateTime type is not supported.");
-                    return;
-            }
-        }
+        // 3. Delete old values
+        // TODO Depends on implementation details in create_new_column
```
beeender: What does this TODO mean?

beeender:
@cmelchior @Zhuinden 

> A migration cannot reference any RealmModel class as it might not exist. Which is why these methods take RealmObjectSchema directly instead of Class<? extends RealmModel>

```java
public RealmObjectSchema addRealmListField(String fieldName, Class<?> clazz)
```

This works for `RealmObject` list as well since we can get the class name from the mediator:

```
dynamicRealm.getConfiguration().getSchemaMediator().getSimpleClassName(clazz)`;
```

But it is not ideal since then the API is not "dynamic" enough then.

I tend to like the first proposal (Overload methods)

Zhuinden:
@beeender in that long code I did exactly that, obtain simple class name from schema mediator, but Melchior convinced me that this means if you delete a class later then your migration code would also break, and that'd be very strange.

realm/realm-library/src/main/cpp/io_realm_internal_Table.cpp
```diff
@@ -199,7 +217,106 @@ JNIEXPORT jboolean JNICALL Java_io_realm_internal_Table_nativeIsColumnNullable(J
...
+                }
+                else {
+                    new_table->set_string(new_col_ndx, i, sd);
+                }
+                break;
+            }
+            case type_Binary: {
+                // Payload copy is needed
+                BinaryData bd = old_table->get_binary(old_col_ndx, i);
+                std::vector<char> binary_copy(bd.data(), bd.data() + bd.size());
```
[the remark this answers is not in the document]
kneth: `BinaryData` doesn't maintain ownership/lifetime of the payload so I think we have to do that here. After the `set_binary()` with the copying, I believe you will have *two* cells pointing to the same place in memory.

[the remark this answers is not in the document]
cmelchior: Removed in https://github.com/realm/realm-java/pull/5351

realm/realm-library/src/main/java/io/realm/RealmObjectSchema.java
```diff
@@ -162,6 +164,33 @@ public String getClassName() {
...
+     * {@link #addRealmListField(String, RealmObjectSchema)} instead.
+     *
+     * @param fieldName name of the field to add.
+     * @param elementType simple type of elements in the array.
+     * @return the updated schema.
+     * @throws IllegalArgumentException if the field name is illegal, a field with that name already exists or
+     * the element type isn't supported.
+     * @throws UnsupportedOperationException if this {@link RealmObjectSchema} is immutable.
+     */
+    public abstract RealmObjectSchema addRealmListField(String fieldName, Class<?> elementType);
```
Zhuinden: I think naming `elementType` to be `primitiveType` would make this method easier to understand, that way you don't need to read the javadoc to know that this is for array of primitives only.

realm/realm-library/src/main/java/io/realm/MutableRealmObjectSchema.java
```diff
@@ -132,18 +132,18 @@ public RealmObjectSchema addRealmListField(String fieldName, RealmObjectSchema o
...
-    public RealmObjectSchema addRealmListField(String fieldName, Class<?> elementType) {
+    public RealmObjectSchema addRealmListField(String fieldName, Class<?> primitiveType) {
         checkLegalName(fieldName);
         checkFieldNameIsAvailable(fieldName);
 
-        FieldMetaData metadata = SUPPORTED_SIMPLE_FIELDS.get(elementType);
+        FieldMetaData metadata = SUPPORTED_SIMPLE_FIELDS.get(primitiveType);
         if (metadata == null) {
-            if (elementType.equals(RealmObjectSchema.class)) {
+            if (primitiveType.equals(RealmObjectSchema.class)) {
```
Zhuinden: I wonder if this should check against `if (primitiveType.equals(RealmObjectSchema.class) || RealmModel.class.isAssignableFrom(primitiveType)) {` to ensure that they get this error message if they provide the `Class<?>` that belongs to a RealmModel.

Currently I think they'd get `RealmList does not support lists with this type: "SomeRealmObject"` which could be confusing.

realm/realm-library/src/main/java/io/realm/RealmObjectSchema.java
```diff
@@ -178,7 +181,7 @@ public String getClassName() {
      *     .setRequired("children", true)
      * }
      * </pre>
-     * If the list contains references to other Realm classes, then use
+     * If the list contains references to other Realm classes use
```
Zhuinden: The comma was needed here 😛

realm/realm-library/src/androidTest/java/io/realm/RealmObjectSchemaTests.java
```diff
@@ -354,18 +393,31 @@ public void requiredFieldAttribute() {
...
         }
+        for (FieldListType fieldType : FieldListType.values()) {
+            switch(fieldType) {
+                case LIST:
+                    continue; // Not possible.
+                default:
+                    // All simple list types
+                    schema.addRealmListField(fieldName, fieldType.getType());
+                    if (fieldType.isNullable()) {
+                        schema.setRequired(fieldName, true);
```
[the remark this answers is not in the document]
cmelchior: Yes, changing name of method to increase clarity

realm/realm-library/src/androidTest/java/io/realm/RealmObjectSchemaTests.java
```diff
@@ -604,6 +684,17 @@ public void setRemoveRequired() {
                     } catch (IllegalArgumentException ignored) {
                     }
                     break;
+                default:
```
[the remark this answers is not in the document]
cmelchior: I modified the enum, so it only contains non_list types. The lists are check in the next block

realm/realm-library/src/androidTest/java/io/realm/RealmObjectSchemaTests.java
```diff
@@ -667,6 +757,21 @@ public void setRequired_nullValueBecomesDefaultValue() {
                     break;
             }
         }
+        for (FieldListType fieldType : FieldListType.values()) {
```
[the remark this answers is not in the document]
cmelchior: A test was being added here in #5351 . I would prefer to keep this for now to decrease merge conflicts.

realm/realm-library/src/main/java/io/realm/MutableRealmObjectSchema.java
Zhuinden: @cmelchior please verify this

Zhuinden:
https://github.com/realm/realm-java/pull/5329#discussion_r142172956

realm/realm-library/src/main/java/io/realm/MutableRealmObjectSchema.java
cmelchior: Yes, this would be nice. Fixing

realm/realm-library/src/main/java/io/realm/DynamicRealmObject.java
```diff
@@ -358,15 +361,54 @@ public DynamicRealmObject getObject(String fieldName) {
...
      * @param fieldName the name of the field.
+     * @param elementType the type of elements in the list. Only primitive types are supported.
      * @return the {@link RealmList} data for this field.
-     * @throws IllegalArgumentException if field name doesn't exist or it doesn't contain a list of values.
+     * @throws IllegalArgumentException if field name doesn't exist or it doesn't contain a list of primitive objects.
      */
-    public <T> RealmList<T> getValueList(String fieldName, Class<T> valueClass) {
-        // TODO implement this
-        return null;
+    public <E> RealmList<E> getList(String fieldName, Class<E> elementType) {
```
Zhuinden: Sooo, `elementType` vs `primitiveType`?

realm/realm-library/src/main/java/io/realm/DynamicRealmObject.java
```diff
@@ -727,21 +757,44 @@ public void setObject(String fieldName, @Nullable DynamicRealmObject value) {
...
 
+        // Find type of list in Realm
         long columnIndex = proxyState.getRow$realm().getColumnIndex(fieldName);
-        OsList osList = proxyState.getRow$realm().getList(columnIndex);
+        final RealmFieldType columnType = proxyState.getRow$realm().getColumnType(columnIndex);
+
+        switch (columnType) {
+            case LIST:
+                //noinspection unchecked
+                setModelList(fieldName, (RealmList<DynamicRealmObject>) list);
```
Zhuinden: I think there should be some kind of custom error message for providing a `RealmList<? extends RealmModel>` that is not `RealmList<DynamicRealmObject>`.

realm/realm-library/src/main/java/io/realm/DynamicRealmObject.java
```diff
@@ -727,21 +757,44 @@ public void setObject(String fieldName, @Nullable DynamicRealmObject value) {
      * Sets the reference to a {@link RealmList} on the given field.
      *
      * @param fieldName field name.
-     * @param list list of references.
-     * @throws IllegalArgumentException if field name doesn't exist, it is not a list field, the type
-     * of the object represented by the DynamicRealmObject doesn't match or any element in the list belongs to a
-     * different Realm.
+     * @param list list of objects. Must either be primitive types or Realm objects.
```
Zhuinden: This should be a bit more specific that it requires `{@link DynamicRealmObject}`.

realm/realm-library/src/main/java/io/realm/DynamicRealmObject.java
cmelchior: `primitiveType`.
I'm renaming variables right now 😛

realm/realm-library/src/main/java/io/realm/DynamicRealmObject.java
cmelchior: Done

realm/realm-library/src/main/java/io/realm/DynamicRealmObject.java
cmelchior: Done

realm/realm-library/src/androidTest/java/io/realm/RealmObjectSchemaTests.java
```diff
@@ -1317,6 +1317,21 @@ public void getFieldType_nonLatinName() {
...
+    public void addList_modelClassThrowsWithProperError() {
+        if (type == ObjectSchemaType.IMMUTABLE) {
+            return;
+        }
+
+        try {
+            schema.addRealmListField("field", AllJavaTypes.class);
+            fail();
+        } catch (IllegalArgumentException e) {
+            assertTrue(e.getMessage().contains("Use 'addRealmListField(String name, RealmObjectSchema schema)' instead"));
```
Zhuinden: 👍

realm/realm-library/src/main/java/io/realm/MutableRealmObjectSchema.java
```diff
@@ -138,7 +138,7 @@ public RealmObjectSchema addRealmListField(String fieldName, Class<?> primitiveT
 
         FieldMetaData metadata = SUPPORTED_SIMPLE_FIELDS.get(primitiveType);
         if (metadata == null) {
-            if (primitiveType.equals(RealmObjectSchema.class)) {
+            if (primitiveType.equals(RealmObjectSchema.class) || RealmModel.class.isAssignableFrom(primitiveType)) {
```
Zhuinden: 👍

realm/realm-library/src/main/java/io/realm/DynamicRealmObject.java
```diff
@@ -775,6 +775,16 @@ public void setObject(String fieldName, @Nullable DynamicRealmObject value) {
 
         switch (columnType) {
             case LIST:
+                // Due to type erasure it is not possible to check the generic parameter,
+                // instead we try to see if the first element is of the wrong type in order
+                // to throw a better error message.
+                // Primitive types are checked inside `setModelList`
+                if (!list.isEmpty()) {
+                    E element = list.first();
```
Zhuinden: I think just catching the possible `ClassCastException` and re-throwing it with the custom message would be more reliable and less hacky.

I like the message though!

cmelchior: The classcast exception would not catch the generic parameter, only if `RealmList` was of the wrong type. Type erasure is a pain :(

Zhuinden: I see, I guess it'll do. Thank you! 🙂

Issue #3210, opened 2016-07-24
Support for OrderedRealmCollection as RealmObject field (if possible)


Zhuinden:
> **Feature Request**: Just fill in the first two sections below.

#### Goal

> What do you want to achieve?

Set the elements of a RealmObject to be the objects returned by a query without having to explicitly map the proxies of the RealmResults into a RealmList

Something like this

#### Code Sample

``` java
    public class SomeObject extends RealmObject {
         public OrderedRealmCollection<OtherObject> others;
    }

    realm.executeTransaction((realm) -> {
        SomeObject someObject = new SomeObject();
        someObject.others = realm.where(OtherObject.class).equalTo("someObjectId", someId).findAll();
        realm.insertOrUpdate(someObject);
    });
```

This would avoid having to explicitly map all elements of a RealmResults into a RealmList before saving a one-many (or many-many) relationship.

stk1m1:
@Zhuinden 

I think it would make our API leaner. Since this needs further discussion I'll keep this open until enough opinion gets collected. Thanks for the contribution!

Zhuinden:
Well, the real question is whether it's even possible :) 

But it would definitely greatly simplify mapping one-many and many-many relationships. Currently that's the most memory-intensive operation when using Realm.

stk1m1:
I think it is possible and I'd be glad to see something that reduces lines and make things clear.

What I'm wondering is the point when you get results from Realm. Suppose I have a RealmObject like below.

```java
public class User extends RealmObject {
    @PrimaryKey
    private String name;
    public OrderedRealmCollection<Dog> dogs; 
}
```

Once you saved `dogs` field with `OrderedRealmCollection` it would work great for now. In the future, however, what should I expect for `dogs` field when `RealmSet` API is available? 😅 

It's just one concern, but I think the idea is something we'd like to consider when expanding out API.

Zhuinden:
@stk1m1 I'd expect it to be a `RealmList` at the moment when obtained with `findAll()` :tongue:

cmelchior:
We have talked internally about computed fields which is a broader concept than what is described here, but it would probably look something like this:

```
public class Person extends RealmObject {
  private String name;
  
  @Query("name == 'Fido") // Text based query language shared between all Realm bindings.
  RealmResults<Dog> dogs;

  @Computed("Mr. $name") // Basic math + string manipulation
  private String displayName;
}
```

[...]

PR #5331, opened 2017-09-27
Inverse insert

tulioccalazans:
Hi,

I propose a new feature allows to insert an item that belongs to an entity easily. It will help to write less code lines.

E.g.:

```
public class Task extends RealmObject {

    @Required
    @PrimaryKey
    private String id;

    @Required
    private String name;
    private boolean done;

    private long timestamp;

    @LinkingObjects("tasks")
    private final RealmResults<User> user = null;

}
```

```
public class User extends RealmObject {

    @Required
    @PrimaryKey
    private String id;
    private String firstName;
    private String lastName;
    //private Task task;
    private RealmList<Task> tasks;
}
```

Today the insertion of a new task needs to be like that:

```
User user = realm.where(User.class).equalTo("id", CurrentUserId).findFirst();
 Task task = realm.createObject(Task.class, UUID.randomUUID().toString());
 task.setName(String.valueOf(taskEditText.getText()));
 task.setTimestamp(System.currentTimeMillis());
 user.getTasks().add(task);
```

I propose some changes to simplify this insertion so you do not have to load the user.

```
public class Task extends RealmObject {

    @Required
    @PrimaryKey
    private String id;

    @Required
    private String name;
    private boolean done;

    private long timestamp;

    @LinkingObjects("tasks") 
    private final RealmResults<User> user = null;

   **@LinkingIds("id")
    private final RealmResultIds<String> userId = 0;**
}
```

Then, insertion can be like that:
 ```
 Task task = realm.createObject(Task.class, UUID.randomUUID().toString());
 task.setName(String.valueOf(taskEditText.getText()));
 task.setTimestamp(System.currentTimeMillis());
 task.setUserId(CurrentUserId);
 realm.insert(task)
```

PS.: This is a simply example... this will be very useful when we need to create several relations in different entities to the task.


Zhuinden:
1.) this shouldn't be a PR

2.) I think this is a subset of https://github.com/realm/realm-java/issues/3210#issuecomment-234908853

cmelchior:
I won't rule out that you can do something clever with how to define parent ID's but overloading the `@LinkingObjects` annotation like that is probably not a good idea. Also, it would not work, as your `RealmResultsIds` do not know _who_ the parent is.

While can see your use case and why you might want to make it easier, we also have to consider the impact on complexity, and in this case, I'm not entirely convinced the benefits outway the drawbacks

In any case, this sort of feature request should not be discussed in a PR, but rather an issue. So, I'll close this PR. If you feel strongly about this feature feel free to create an issue where we can discuss further.

PR #5336, opened 2017-09-28
Primitive List feature(Part15: Disable RealmList.createSnapshot() when the elements are not RealmModel)

zaki50:


CHANGELOG.md
```diff
@@ -15,7 +15,7 @@
 
 ## Enhancements
 
-* Support list of values as a field of model class. Now users can use `String`, `byte[]`, `Boolean`, `Long`, `Integer`, `Short`, `Byte`, `Double`, `Float` and `Date` as a type parameter of `RealmList`.
+* Supported list of values as a field of model class. Now users can use `String`, `byte[]`, `Boolean`, `Long`, `Integer`, `Short`, `Byte`, `Double`, `Float` and `Date` as a type parameter of `RealmList`.
```
Zhuinden: I don't think this is right. It's `supported` only if it's something like, `List of values are now supported as a field of model class`.

PR #5339, opened 2017-09-28
Build core from source

beeender:
- Separate core library cmake to RealmCore.cmake
- Build core from source code with the new core's cmake build. This is
  much faster than before, it will only build the needed ABI and
  release/debug variant.

### DIFF
--- a/realm/realm-library/build.gradle
+++ b/realm/realm-library/build.gradle
@@ -61,6 +61,7 @@ android {
                         "-DENABLE_DEBUG_CORE=$project.enableDebugCore"
                 if (project.ccachePath) arguments "-DNDK_CCACHE=$project.ccachePath"
                 if (project.lcachePath) arguments "-DNDK_LCACHE=$project.lcachePath"
+                if (project.coreSourcePath) arguments "-DCORE_SOURCE_PATH=$project.coreSourcePath"
                 if (project.hasProperty('buildTargetABIs') && !project.getProperty('buildTargetABIs').trim().isEmpty()) {
                     abiFilters(*project.getProperty('buildTargetABIs').trim().split('\\s*,\\s*'))
                 } else {
@@ -120,7 +121,7 @@ android {
         abortOnError false
     }
 
-   flavorDimensions 'api'
+    flavorDimensions 'api'
 
     productFlavors {
         base {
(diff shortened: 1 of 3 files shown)


realm/realm-library/build.gradle
```diff
@@ -496,35 +497,9 @@ task downloadCore() {
...
-            into project.coreArchiveFile.parent
-            rename "realm-core-android-${coreVersion}.tar.gz", "realm-sync-android-${coreVersion}.tar.gz"
-        }
-    }
-}
-
 task deployCore(group: 'build setup', description: 'Deploy the latest version of Realm Core') {
     dependsOn {
-        coreSourcePath ? compileCore : downloadCore
+        downloadCore
```
cmelchior: Is this right? How does the local core get deployed then?

realm/realm-library/src/main/cpp/CMake/RealmCore.cmake
```diff
@@ -0,0 +1,125 @@
...
+        set(debug_lib_suffix "-dbg")
+        add_compile_options(-DREALM_DEBUG)
+    else()
+        add_compile_options(-DNDEBUG)
+    endif()
+
+    # Configure import realm core lib
+    set(core_lib_path ${sync_dist_path}/librealm-android-${ANDROID_ABI}${debug_lib_suffix}.a)
+    if (NOT EXISTS ${core_lib_path})
+        if (ARMEABI)
```
cmelchior: What about x86?

realm/realm-library/src/main/cpp/CMake/RealmCore.cmake
```diff
@@ -0,0 +1,125 @@
...
+    set_target_properties(lib_realm_core PROPERTIES IMPORTED_LOCATION ${core_lib_path}
+        IMPORTED_LINK_INTERFACE_LIBRARIES atomic
+        INTERFACE_INCLUDE_DIRECTORIES "${sync_dist_path}/include")
+
+    if (enable_sync)
+        # Sync static library
+        set(sync_lib_path ${sync_dist_path}/librealm-sync-android-${ANDROID_ABI}${debug_lib_suffix}.a)
+        # Workaround for old core's funny ABI nicknames
+        if (NOT EXISTS ${sync_lib_path})
+            if (ARMEABI)
```
cmelchior: Same, what about x86?

realm/realm-library/build.gradle
beeender: local core should be existing already when you pass the `coreSourcePath`. but yeah, we could support clone a new core when build.

realm/realm-library/src/main/cpp/CMake/RealmCore.cmake
beeender: core's ABI name for x86 is correct. This old code moved from `CMakeList.txt`. those thing could be removed later when sync switched to cmake.

cmelchior: 👍

PR #5031, opened 2017-07-27
Support Primitive List feature

zaki50:
This PR implements #4165 and fixes #575.

TODO: 
**Required:**
- [x] Add, modify APIs (@zaki50) #5145 
- [x] Loosen type restriction on `RealmList<RealmModel>` to `RealmList<Object>` (@zaki50) #5145
- [x] Loosen type restriction on `RealmResults<RealmModel>` to `RealmResults<Object>` (@zaki50) #5145
- [x] Implement type parameter check in annotation processor (@zaki50) #5168 
- [x] Implement Schema related feature used on initialization (@zaki50) #5209 (part 7)
- [x] Implement proxy's getters/setters logic in annotation processor (@zaki50) #5210 (part 8)
- [x] Implement support for primitives in `RealmList` (Java side, @zaki50) #5213 (part 9)
- [x] Wire with Core/Object Store (@beeender) #5208 (part 6), #5216 (part10)
- [x] Fix for change listener (@beeender) #5259 (part 13) 
- [x] Fix a bug introduced in this feature branch (@zaki50) #5266 
- [x] If we merge feature branch without optional feature, we need to add a code to throw an exception to some methods which does not support primitive list (ex. `createSnapshot()`) (@zaki50) #5336 (part 15)

**Optional/In prioritized order:**
- [x] `copyToRealm/copyOrUpdate` (@zaki50) #5210 (part 8)
- [x] `insert/insertOrUpdate` (@zaki50) #5210 (part 8)
- [x] `copyFromRealm` (@zaki50) #5210 (part 8)
- [ ] APIs for migration (mainly methods in `RealmObjectSchema`)  (@cmelchior) #5329 
- [ ] `OrderedRealmCollectionImpl`, `OrderedRealmCollectionSnapshot`  (@zaki50) #5220 (part11)
- [ ] Dynamic API (@zaki50) #5221 (part 12)
- [ ] Query APIs for list elements
- [ ] Query APIs for object (by using primitive list value)
- [ ] Implement support for primitives in `RealmResults` (@zaki50) #5220 (part11)
- [ ] JSON APIs

PR #5351, opened 2017-09-29
Array of primitives: DynamicRealm support

cmelchior:
Part of #5031
This PR adds support for primitive arrays in DynamicRealmObject

TODO
- [x] Cleanup 
- [x] Fix RealmSchema unit tests so they use DynamicRealm instead
- [x] Determine naming. See #5329 . I already made several mistakes with the overloaded `dynamicRealm.getList()`, so I'm probably starting to lean towards having a seperate `get<Primitive/Value>List()` method instead ... People seem to like the overloaded methods best, so lets keep those.

### DIFF
--- a/realm/realm-library/src/androidTest/java/io/realm/DynamicRealmObjectTests.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/DynamicRealmObjectTests.java
@@ -1024,6 +1024,21 @@ public class DynamicRealmObjectTests {
     }
 
     @Test
+    public void setList_objectsOwnList() {
+        dynamicRealm.beginTransaction();
+
+        // Test model classes
+        int originalSize = dObjDynamic.getList(AllJavaTypes.FIELD_LIST).size();
+        dObjDynamic.setList(AllJavaTypes.FIELD_LIST, dObjDynamic.getList(AllJavaTypes.FIELD_LIST));
+        assertEquals(originalSize, dObjDynamic.getList(AllJavaTypes.FIELD_LIST).size());
+
+        // Smoke test value lists
+        originalSize = dObjDynamic.getList(AllJavaTypes.FIELD_STRING_LIST, String.class).size();
+        dObjDynamic.setList(AllJavaTypes.FIELD_STRING_LIST, dObjDynamic.getList(AllJavaTypes.FIELD_STRING_LIST, String.class));
+        assertEquals(originalSize, dObjDynamic.getList(AllJavaTypes.FIELD_STRING_LIST, String.class).size());
+    }
+
+    @Test
     public void untypedSetter_listWrongTypeThrows() {
         realm.beginTransaction();
         AllTypes wrongObj = realm.createObject(AllTypes.class);
--- a/realm/realm-library/src/androidTest/java/io/realm/DynamicRealmObjectTests.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/DynamicRealmObjectTests.java
@@ -95,6 +95,13 @@ public class DynamicRealmObjectTests {
         typedObj.setFieldDate(new Date(1000));
         typedObj.setFieldObject(typedObj);
         typedObj.getFieldList().add(typedObj);
+        typedObj.getFieldIntegerList().add(1);
+        typedObj.getFieldStringList().add("str");
+        typedObj.getFieldBooleanList().add(true);
+        typedObj.getFieldFloatList().add(1.23F);
+        typedObj.getFieldDoubleList().add(1.234D);
+        typedObj.getFieldBinaryList().add(new byte[] {1, 2, 3});
+        typedObj.getFieldDateList().add(new Date(1000));
         dObjTyped = new DynamicRealmObject(typedObj);
         realm.commitTransaction();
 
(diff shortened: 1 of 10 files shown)


Zhuinden:
The tricky thing with `getList()` is that it can by default only return `List<DynamicRealmObject>`...

With this setup, `getList()` isn't a very good name. 😑 😟 

While I said the `addRealmListField()` should do the two different things based on the provided `Class<?>`, in this case having two separately named methods is pretty much necessary.

cmelchior:
Ready for review @realm/java

realm/realm-library/src/main/java/io/realm/DynamicRealmObject.java
```diff
@@ -857,11 +858,55 @@ private void setModelList(String fieldName, RealmList<DynamicRealmObject> list)
             default:
                 throw new IllegalArgumentException("Unsupported type: " + elementType);
         }
+        final ManagedListOperator<?> operator = getOperator(proxyState.getRealm$realm(), osList, elementType, elementClass);
+
+        if (list.isManaged() && osList.size() == list.size()) {
```
Zhuinden: When does this happen exactly? Is there a test for it? Is `osList.size == list.size` sufficient to determine that this is the case :question:

cmelchior: No, but if the `size()` is different we know with 100% that the lists are not the same, in which case we can use `clear() + insertAll()` which is much faster than manually replacing each element in the the list. 

If they are the same size() we don't know for sure, so we have to revert to manually replacing each element.

realm/realm-library/src/main/java/io/realm/DynamicRealmObject.java
```diff
@@ -336,6 +337,8 @@ public DynamicRealmObject getObject(String fieldName) {
 
     /**
      * Returns the {@link RealmList} of {@link DynamicRealmObject}s being linked from the given field.
+     * <p>
+     * If the list only contain primitive types, use {@link #getList(String, Class)} instead.
```
beeender: `contain` -> `contains`

realm/realm-library/src/androidTest/java/io/realm/RealmObjectSchemaTests.java
```diff
@@ -758,22 +759,118 @@ public void setRequired_nullValueBecomesDefaultValue() {
...
+        schema.setRequired("foo", true);
+        list = obj.getList("foo", byte[].class);
+        assertEquals(0, list.get(0).length);
+        assertArrayEquals(new byte[] {1, 2, 3}, list.get(1));
+
+        // Transform back to nullable
+        schema.setRequired("foo", false);
+        list = obj.getList("foo", byte[].class);
+        assertEquals(0, list.get(0).length);
+        assertArrayEquals(new byte[] {1, 2, 3}, list.get(1));
```
beeender: Do you want to call 
```
// Cleanup
schema.removeField("foo");
```
here as well?

It is actually no need, the `tearDown` will do that. So https://github.com/realm/realm-java/pull/5351/files#diff-4767794a7ded2dc2e3cc9447650ac1c7R838 can be removed as well.

realm/realm-library/src/main/java/io/realm/DynamicRealmObject.java
```diff
@@ -358,15 +361,54 @@ public DynamicRealmObject getObject(String fieldName) {
...
      * @param fieldName the name of the field.
+     * @param elementType the type of elements in the list. Only primitive types are supported.
      * @return the {@link RealmList} data for this field.
-     * @throws IllegalArgumentException if field name doesn't exist or it doesn't contain a list of values.
+     * @throws IllegalArgumentException if field name doesn't exist or it doesn't contain a list of primitive objects.
      */
-    public <T> RealmList<T> getValueList(String fieldName, Class<T> valueClass) {
-        // TODO implement this
-        return null;
+    public <E> RealmList<E> getList(String fieldName, Class<E> elementType) {
```
Zhuinden: I think https://github.com/realm/realm-java/pull/5329#discussion_r142131879 applies here as well (that `primitiveType` is a better name)

zaki50: Why `primitive` is better? It's not a primitive type at least in Java.

Zhuinden: Technically because the feature is called "array of primitives".

The term `element` does not portray that this is not meant to be used for explicit RealmModel classes.

zaki50: "array of primitives" is just a internal feature name and I don't think we are using it in our documentation.

cmelchior: The problem is that no terminology is 100% correct without also being too vague:

`elementType`: Correct, but doesn't describe the difference between the RealmModel variant and this very well.
`valueType`: `Date` and byte[] isn't value types. Nor may others be in the future
`primitiveType` `Date/byte[]` isn't primitive types. Technically neither are any of the boxed types.

`valueType` is probably the one with the least amount of "wrongness", but IMO it is also a pretty specialized concept which I'm guessing not a lot know about. Primitive types is hopefylly someone even starting out with Java knows about.

Hence `primitiveType`

cmelchior: Damn, writing this up, made me doubt if perhaps `valueType` isn't the correct word after all 🤕

Zhuinden: @cmelchior I used to stand behind `value` but I'm pretty sure the core name and the feature name and the way people refer to it has been `list of primitives` for a long time.

If you check the Core issue https://github.com/realm/realm-core/issues/1881 from which there are many linked issues, they are all called "list of primitives" and `value` is never there.

zaki50: We don't need to use the same word with core.
The most important thing is what is the most natural for Java developers.

Zhuinden: > We don't need to use the same word with core.

You don't *need to*, but I think the other alternatives allow more confusion.

cmelchior: > We don't need to use the same word with core.The most important thing is what is the most natural for Java developers.

True, but what is that? :)

cmelchior: Especially, if there isn't a _clear_ benefit_ to choosing another word, we should probably go with what is otherwise used in Core/Bindings/Issues.

zaki50: @cmelchior  All candidates don't describe the difference between the RealmModel variant and expected ones well. 
I could't find the reason to choose more incorrect one...

But I can accept your decision.

realm/realm-library/src/androidTest/java/io/realm/DynamicRealmObjectTests.java
```diff
@@ -545,6 +578,55 @@ public void setter_null() {
                         } catch (IllegalArgumentException ignored) {
                         }
                         break;
+                    case LIST_INTEGER:
+                        try {
+                            dObj.setNull(NullTypes.FIELD_INTEGER_LIST_NULL);
```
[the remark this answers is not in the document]
cmelchior: Yes, but this is setting the field itself to null, not the values in the list. The field itself isn't allowed to be null.

realm/realm-library/src/main/java/io/realm/DynamicRealmObject.java
Zhuinden: I think it ended up staying `elementType`.

PR #5356, opened 2017-09-30
Expose SyncSessionStopPolicy in SyncConfiguration

cmelchior:
This PR exposes the SyncSessionStopPolicy in SyncConfiguration.
This will hopefully help combat the test assertions we are seeing with active sessions being present when a test ends. This is especially a problem in the PermissionManager tests that rely on internal Realms.

Only downside is that `SyncConfiguration.getSessionStopPolicy()` had to be made public. I put big fat warnings there though, so hopefully, it is okay?

### DIFF
--- a/realm/realm-library/src/objectServer/java/io/realm/SyncConfiguration.java
+++ b/realm/realm-library/src/objectServer/java/io/realm/SyncConfiguration.java
@@ -88,6 +88,7 @@ public class SyncConfiguration extends RealmConfiguration {
     @Nullable
     private final String serverCertificateFilePath;
     private final boolean waitForInitialData;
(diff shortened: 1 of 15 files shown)


realm/realm-library/src/objectServer/java/io/realm/SyncConfiguration.java
```diff
@@ -346,6 +349,17 @@ boolean isSyncConfiguration() {
     }
 
     /**
+     * NOTE: Only for internal usage. Maybe change without warning.
```
nhachicha: > Maybe change without warning. 

?

cmelchior:
I have definitely seen less random crashes

Issue #5354, opened 2017-09-30
Crashes when linking with debug version of Sync


cmelchior:
Moving #5346 to a separate issue, so we can track it.

Observations so far:
- Only crash on armeabi-v7a with `objectServerDebug` builds. x86 and 64 bit variants do not crash.
- Debug builds with `baseDebug` do not crash.
- Release builds for both base and ObjectServer do not crash. This only effect debug builds.
- Using `-O0/01/02` instead of `-Og` does not make a difference.

The problematic method is:

```
void ObjectStore::set_schema_columns(Group const& group, Schema& schema)
{
    for (auto& object_schema : schema) {
        auto table = table_for_object_schema(group, object_schema);
        if (!table) {
            continue;
        }
        for (auto& property : object_schema.persisted_properties) {
            property.table_column = table->get_column_index(property.name);
        }
    }
}
```
From https://github.com/realm/realm-object-store/blob/master/src/object_store.cpp#L777

For some reason Core returns `realm::not_found` (4294967295) for all non-latin names in `get_column_index`, while ascii names are correctly found. The columns have been verified to be there with the correct name (by going around the ObjectStore `Property` and directly querying Cores `Table`

![image](https://user-images.githubusercontent.com/406066/31045953-9a0a9e16-a5ef-11e7-8cb2-fd88281ffcc9.png)

Example of crash:

```
09-30 14:46:42.389 6362-6378/? E/REALM: /home/jenkins/workspace/lm_realm-core_release_4.0.0-LXTALRDDGEXO6MUGNEPKX627PBBN5UZMAYUDJFS3OHOQUQGYVBKA@5/src/realm/table.cpp:3270: [realm-core-4.0.0] Assertion failed: col_ndx < get_column_count() [4294967295, 5]
                                        IMPORTANT: if you see this error, please send this log to help@realm.io.
09-30 14:46:42.390 6362-6378/? A/libc: Fatal signal 6 (SIGABRT), code -6 in tid 6378 (roidJUnitRunner)
09-30 14:46:42.583 6386-6386/? A/DEBUG: *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
09-30 14:46:42.583 6386-6386/? A/DEBUG: Build fingerprint: 'google/angler/angler:8.0.0/OPR6.170623.019/4299446:user/release-keys'
09-30 14:46:42.583 6386-6386/? A/DEBUG: Revision: '0'
[...]

PR #5358, opened 2017-09-30
Use release version of Sync when testing

cmelchior:
From #5354 

I should have listened to @beeender 
Using the debug version is causing all sorts of crashes. Reverting to using the `release` version until we can fix it.


cmelchior:
Fast-merge

PR #5359, opened 2017-10-01
Support Partial Sync

### LINKED ISSUE #5276: Support Partial Sync Preview
Add support in Realm Java once https://github.com/realm/realm-cocoa/pull/5326 is done

nhachicha:
fixes #5276


nhachicha:
@realm/java

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
```diff
@@ -501,3 +506,57 @@ JNIEXPORT void JNICALL Java_io_realm_internal_SharedRealm_nativeRegisterSchemaCh
...
+                catch (const std::exception& e) {
+                    env->CallStaticVoidMethod(shared_realm_class, partial_sync_cb, to_jstring(env, e.what()),
+                                              reinterpret_cast<jlong>(nullptr), nullptr, nullptr,
+                                              j_callback_weak.global_ref(env).get(), nullptr);
+                }
+                return;
+            }
+
+            auto wrapper = new ResultsWrapper(results);
+            env->CallStaticVoidMethod(shared_realm_class, partial_sync_cb, nullptr, reinterpret_cast<jlong>(wrapper),
```
cmelchior: Don't we risk that these weak references have been GC'ed when the callback is invoked?

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -1648,6 +1649,33 @@ public static boolean compactRealm(RealmConfiguration configuration) {
...
+    /**
+     * If the Realm is a partially synchronized Realm, fetch and synchronize the objects of a given
+     * object type that match the given query (in string format).
+     *
+     * The results will be returned asynchronously in the callback.
+     *
+     * @param clazz the class to query.
+     * @param query string query.
+     * @param callback A callback used to vend the results of a partial sync fetch.
+     * @throws IllegalStateException if it is called from a non-Looper or {@link IntentService} thread.
```
cmelchior: Also document the IllegalState for non-synced Realms.

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -1648,6 +1649,33 @@ public static boolean compactRealm(RealmConfiguration configuration) {
...
+     */
+    @Beta
+    public <E extends RealmModel> void subscribeToObjects(Class<E> clazz, String query, PartialSyncCallback<E> callback) {
```
cmelchior: Should this return a `RealmAsyncTask` instead? 
Or maybe we shouldn't worry about this for just a preview API?

realm/realm-library/src/objectServer/java/io/realm/internal/SyncObjectServerFacade.java
```diff
@@ -92,9 +93,9 @@ public void realmClosed(RealmConfiguration configuration) {
             String rosUserIdentity = user.getIdentity();
             String syncRealmAuthUrl = user.getAuthenticationUrl().toString();
             String rosSerializedUser = user.toJson();
-            return new Object[]{rosUserIdentity, rosServerUrl, syncRealmAuthUrl, rosSerializedUser, syncConfig.syncClientValidateSsl(), syncConfig.getServerCertificateFilePath()};
+            return new Object[]{rosUserIdentity, rosServerUrl, syncRealmAuthUrl, rosSerializedUser, syncConfig.syncClientValidateSsl(), syncConfig.getServerCertificateFilePath(), syncConfig.isPartialSync()};
         } else {
-            return new Object[6];
+            return new Object[7];
```
cmelchior: Merge conflict with #5356 .... I'll race you 😛

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/SyncSessionTests.java
```diff
@@ -455,4 +462,121 @@ public void run() {
         assertEquals(3, realm.where(StringOnly.class).count());
         realm.close();
     }
+
+    @Test
+    public void partialSync() throws InterruptedException {
```
cmelchior: Move this to its own class? I suspect we will have to add a lot more tests for this down the line?

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/SyncSessionTests.java
```diff
@@ -455,4 +462,121 @@ public void run() {
...
+                .build();
+        SyncConfiguration adminConfig = configFactory
+                .createSyncConfigurationBuilder(adminUser, partialSyncConfig.getServerUrl().toString())
+                .modules(new PartialSyncModule())
+                .build();
+
+        // Using Admin user, populate the Realm.
+        Realm realm = Realm.getInstance(adminConfig);
+        realm.beginTransaction();
+        PartialSyncObjectA objectA = realm.createObject(PartialSyncObjectA.class);
```
cmelchior: Create these in a loop instead? Would make it much more readable

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/SyncSessionTests.java
```diff
@@ -455,4 +462,121 @@ public void run() {
...
+        HandlerThread handlerThread = new HandlerThread("background");
+        handlerThread.start();
+        Handler handler = new Handler(handlerThread.getLooper());
+        handler.post(new Runnable() {
+            @Override
+            public void run() {
+                final Realm partialSyncRealm = Realm.getInstance(partialSyncConfig);
+                assertTrue(partialSyncRealm.isEmpty());
+
+                partialSyncRealm.subscribeToObjects(PartialSyncObjectA.class, "number > 5", new Realm.PartialSyncCallback<PartialSyncObjectA>() {
```
Zhuinden: This is the first time I see string based query API in Realm-Java, I don't know if its usage is documented anywhere.

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/SyncSessionTests.java
```diff
@@ -455,4 +462,121 @@ public void run() {
...
+                partialSyncRealm.subscribeToObjects(PartialSyncObjectA.class, "invalid_property > 5", new Realm.PartialSyncCallback<PartialSyncObjectA>() {
+
+                    @Override
+                    public void onSuccess(RealmResults<PartialSyncObjectA> results) {
+                        fail("Invalid query should not succeed");
+                    }
+
+                    @Override
+                    public void onError(RealmException error) {
+                        assertNotNull(error);
```
cmelchior: What kind of error is thrown here?

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/model/PartialSyncModule.java
```diff
@@ -0,0 +1,7 @@
+package io.realm.objectserver.model;
+
+import io.realm.annotations.RealmModule;
+
+@RealmModule(classes = { PartialSyncObjectA.class, PartialSyncObjectB.class})
```
cmelchior: Spacing is wrong in the end `class })`

tools/sync_test_server/ros-testing-server.js
```diff
@@ -66,6 +66,7 @@ function startRealmObjectServer(onSuccess, onError) {
             syncServerChildProcess = spawn('ros',
                     ['start',
                         '--data', path,
+			'--loglevel', 'all',
```
cmelchior: Do we want to merge this? Doesn't it make looking at the logs rather hard in general?

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/SyncSessionTests.java
kneth: @Zhuinden This is a preview of a upcoming feature, and we might not have documented completely yet.

CHANGELOG.md
```diff
@@ -34,6 +34,7 @@
 
 * [ObjectServer] `SyncUserInfo` now also exposes a users metadata using `SyncUserInfo.getMetadata()`
 * Minor performance improvement when copy/insert objects into Realm.
+* [ObjectServer] Added support for partial sync (#5276).
```
cmelchior: Make it really explicit that this is just a Tech Preview and the API _will_ change.

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/SyncSessionTests.java
Zhuinden: Not what I meant. I'm just surprised this isn't based on `RealmQuery<T>` which pretty much does the same thing except not as a String.

cmelchior: It probably will be. This is just a really early preview of it in order to facilitate feedback. The API _will_ change.

realm/realm-library/src/main/java/io/realm/Realm.java
Zhuinden: I guess that depends on if it should be cancellable

tools/sync_test_server/ros-testing-server.js
nhachicha: I'll remove it, this is only helpful locally since AFAIK we don't capture the ROS log (within Docker) from CI. we'll revert

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/SyncSessionTests.java
nhachicha: This will wrap the error message we get from the Sync client as a `RealmException` https://github.com/realm/realm-object-store/pull/547/files#diff-854aeb40d09f785f4384eb120ed94eadR106

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/SyncSessionTests.java
nhachicha: I'll refactor the creation of `PartialSyncObjectB` in a loop, `PartialSyncObjectA` string's has different value

realm/realm-library/src/main/java/io/realm/Realm.java
nhachicha: We generally return a `RealmAsyncTask` when Java sends the query (via OkHttp) this will just call the ObjectStore to manage the creation of the `__ResultSets` and listening for changes in the query's status. `RealmAsyncTask` is not appropriate is this case

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
nhachicha: I can add a check in `SharedRealm#runPartialSyncRegistrationCallback` to see if the weak references returned from JNI are not null (i.e not GC'ed)

```Java
if (callback != null && realm != null) {
            if (error != null) {
               // notify error
                callback.onError(new RealmException(error));
            } else {
               // notify success  
                ...
                callback.onSuccess(results);
            }    
        }
```

@beeender is it safe to call `j_callback_weak.global_ref(env).get()` if the `JavaGlobalWeakRef` has been GC'ed ? I suppose it will return `nullptr` to Java? is this assumption correct?

realm/realm-library/src/objectServer/java/io/realm/SyncConfiguration.java
```diff
@@ -793,6 +812,15 @@ public Builder waitForInitialRemoteData() {
             return this;
         }
 
+        /**
+         * Setting this will open a partially synchronized Realm.
+         * @see #isPartialSync()
+         */
+        public SyncConfiguration.Builder withPartialSync() {
```
cmelchior: Not sure what the other bindings call this. `.partialRealm()` feels more accurate to me, but can we revisit this when fleshing out the full API.

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
beeender: It looks quite complex and breaking our architectures layers since now the `SharedRealm` needs to know the information about `Class<?>`/`RealmResults` which are supposed to live in the `io.realm` package. And also the reference relationship becomes quite complex. 

My idea would be have: `SharedRealm.PartialSyncCallback` then create new instance in `Realm.java`.

I will create a PR to this one to fix this.

beeender: address this in #5359

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -1684,6 +1685,34 @@ public static boolean compactRealm(RealmConfiguration configuration) {
...
     }
 
+    /**
+     * If the Realm is a partially synchronized Realm, fetch and synchronize the objects of a given
+     * object type that match the given query (in string format).
+     *
+     * The results will be returned asynchronously in the callback.
+     *
+     * @param clazz the class to query.
+     * @param query string query.
```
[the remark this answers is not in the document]
nhachicha: I guess we should point to the website doc when it's available (not sure we can mention NSPredicat format since it's usually unknown to Java dev).

`Query` should not be null (passing a null query is like saying I want to sync back the entire Realm, which defies the point of using this method)

realm/realm-library/src/objectServer/java/io/realm/SyncConfiguration.java
nhachicha: `partialRealm ` is good 
Cocoa is using `isPartial` as a property of the SyncConfig https://github.com/realm/realm-cocoa/pull/5326/files?diff=unified#diff-4ac32a78649ca5bdd8e0ba38b7006a1eR54

DotNet is using `IsPartial` https://github.com/realm/realm-dotnet/pull/1580/files#diff-bc069b1553b16b22b430e81c0e3b641cR95

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
```diff
@@ -501,3 +509,53 @@ JNIEXPORT void JNICALL Java_io_realm_internal_SharedRealm_nativeRegisterSchemaCh
...
+
+    auto& shared_realm = *(reinterpret_cast<SharedRealm*>(shared_realm_ptr));
+    try {
+        JStringAccessor class_name(env, j_class_name);               // throws
+        JStringAccessor query(env, j_query);                         // throws
+
+        // The lambda will capture the copied reference and it will be unreferenced when the lambda's life cycle is over.
+        // That happens when the Realm is closed or the callback has been triggered once.
+        JavaGlobalRef j_callback_ref(env, j_callback);
+        JavaGlobalWeakRef j_shared_realm_instance_ref(env, j_shared_realm_instance);
```
cmelchior: I assume the comment above also refer to this one, but I wasn't 100% sure?

realm/realm-library/src/main/java/io/realm/RealmResults.java
```diff
@@ -83,7 +83,7 @@
                 srcClassName);
     }
 
-    RealmResults(BaseRealm realm, Collection collection, Class<E> clazz) {
+    public RealmResults(BaseRealm realm, Collection collection, Class<E> clazz) {
```
cmelchior: Is this really needed?

realm/realm-library/src/main/java/io/realm/internal/OsRealmConfig.java
```diff
@@ -232,12 +238,20 @@ private OsRealmConfig(final RealmConfiguration config,
...
+            String resolvedSyncRealmUrl = nativeCreateAndSetSyncConfig(nativePtr, syncRealmUrl, syncRealmAuthUrl, syncUserIdentifier,
+                    syncRefreshToken, isPartial, sessionStopPolicy);
+            try {
+                resolvedRealmURI = new URI(resolvedSyncRealmUrl);
+            } catch (URISyntaxException e) {
+                RealmLog.error(e, "Cannot create a URI from the Realm URL address");
+            }
             nativeSetSyncConfigSslSettings(nativePtr, syncClientValidateSsl, syncSslTrustCertificatePath);
         }
+        this.resolvedRealmURI = resolvedRealmURI;
```
cmelchior: Why is this needed? Doesn't the SyncConfiguration have the full URL?

realm/realm-library/src/objectServer/java/io/realm/SyncConfiguration.java
```diff
@@ -360,6 +363,21 @@ boolean isSyncConfiguration() {
...
     /**
+     * Whether this configuration is for a partial synchronization Realm.
+     * Partial synchronization allows a synchronized Realm to be opened in such a way that
+     * only objects requested by the user are synchronized to the device. You can use it by setting
+     * the {@link Builder#partialRealm()}, opening the Realm, and then calling
+     * {@link Realm#subscribeToObjects(Class, String, Realm.PartialSyncCallback)} with the type of
+     * object you're interested in, a string containing a query determining which objects you want
+     * to subscribe to, and a callback which will report the results.
+     *
+     * @return {@code true} is to open a partial synchronization Realm {@code false} otherwise.
```
cmelchior: remove `is`

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
beeender: the comment still refers to the `j_callback_ref`

realm/realm-library/src/main/java/io/realm/RealmResults.java
nhachicha: Good catch, not with Mulong's refactor, fixing

realm/realm-library/src/main/java/io/realm/internal/OsRealmConfig.java
nhachicha: no, the ObjectStore will resolve the URL differently when using a partial sync https://github.com/realm/realm-object-store/pull/547/files#diff-864223d710d7b7e687e4c562de76f750R45

Zhuinden:
So this uses the same query language as Cocoa/JS?

cmelchior:
As JS, yes, but that is heavily inspired by Cocoa

PR #5363, opened 2017-10-02
Fix partial sync callback refs

beeender:


beeender:
To fix https://github.com/realm/realm-java/pull/5359/files#r142059523

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
```diff
@@ -522,42 +521,37 @@ JNIEXPORT void JNICALL Java_io_realm_internal_SharedRealm_nativeRegisterPartialS
...
         JStringAccessor class_name(env, j_class_name);               // throws
-        std::string table_name = JStringAccessor(env, j_table_name); // throws
         JStringAccessor query(env, j_query);                         // throws
 
-        JavaGlobalWeakRef j_callback_weak(env, j_callback);
-        JavaGlobalWeakRef j_realm_weak(env, j_realm);
-        JavaGlobalWeakRef j_clazz_weak(env, j_clazz);
+        // The lambda will capture the copied reference and it will be unrefed when the lambda's life cycle is over.
+        // That happens when the Realm is closed or the callback has been triggered once.
+        JavaGlobalRef j_callback_ref(env, j_callback);
```
nhachicha: retaining a global reference to the callback will prevent the Realm from being GC'ed, since the callback may outlive the Realm

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
```diff
@@ -522,42 +521,37 @@ JNIEXPORT void JNICALL Java_io_realm_internal_SharedRealm_nativeRegisterPartialS
     auto& shared_realm = *(reinterpret_cast<SharedRealm*>(shared_realm_ptr));
     try {
         JStringAccessor class_name(env, j_class_name);               // throws
-        std::string table_name = JStringAccessor(env, j_table_name); // throws
         JStringAccessor query(env, j_query);                         // throws
 
-        JavaGlobalWeakRef j_callback_weak(env, j_callback);
-        JavaGlobalWeakRef j_realm_weak(env, j_realm);
-        JavaGlobalWeakRef j_clazz_weak(env, j_clazz);
+        // The lambda will capture the copied reference and it will be unrefed when the lambda's life cycle is over.
```
nhachicha: > unrefed

unreferenced ?

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -1698,19 +1698,29 @@ public static boolean compactRealm(RealmConfiguration configuration) {
      * @throws IllegalStateException if called from a non-synchronized (Realm Object Server) Realm.
      */
     @Beta
-    public <E extends RealmModel> void subscribeToObjects(Class<E> clazz, String query, PartialSyncCallback<E> callback) {
+    public <E extends RealmModel> void subscribeToObjects(final Class<E> clazz, String query, final PartialSyncCallback<E> callback) {
```
nhachicha: nice idea 👏

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -1698,19 +1698,29 @@ public static boolean compactRealm(RealmConfiguration configuration) {
...
 
         sharedRealm.capabilities.checkCanDeliverNotification(BaseRealm.LISTENER_NOT_ALLOWED_MESSAGE);
 
-        Table table = getTable(clazz);
-        String className = table.getClassName();
-        String tableName = table.getName();
+        String className = configuration.getSchemaMediator().getSimpleClassName(clazz);
+        SharedRealm.PartialSyncCallback internalCallback = new SharedRealm.PartialSyncCallback(sharedRealm, className) {
+            @Override
+            public void onSuccess(io.realm.internal.Collection osResults) {
```
nhachicha: use import instead of the FQN

realm/realm-library/src/main/java/io/realm/internal/SharedRealm.java
```diff
@@ -486,27 +503,23 @@ private static void runInitializationCallback(long nativeSharedRealmPtr, OsRealm
...
-                RealmResults results = new RealmResults(realm, collection, clazz);
-                callback.onSuccess(results);
-            }
+    private static void runPartialSyncRegistrationCallback(@Nullable String error, long nativeResultsPtr,
+                                                           PartialSyncCallback callback) {
+        if (error != null) {
+            callback.onError(new RealmException(error));
+        } else {
+            @SuppressWarnings("ConstantConditions")
+            Table table = callback.sharedRealm.getTable(Table.getTableNameForClass(callback.className));
```
nhachicha: what about making `runPartialSyncRegistrationCallback` non static, we will no longer need to keep `sharedRealm` in the callback

beeender: I will change `PartialSyncCallback.sharedRealm` to be a type of `WeakReference<SharedRealm>` instead. That is easier to check.

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
beeender: will use `private final WeakReference<SharedRealm> sharedRealm;` in the Callback

realm/realm-library/src/main/java/io/realm/Realm.java
beeender: conflicts with `java.util.Collection` and it is totally my fault ... will rename `io.realm.internal.Collection` to `io.realm.internal.OsResults` in a future PR.

PR #5366, opened 2017-10-02
Using weak ref + removing static for the nativeRegisterPartialSyncQuery

nhachicha:


realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
```diff
@@ -525,29 +525,34 @@ JNIEXPORT void JNICALL Java_io_realm_internal_SharedRealm_nativeRegisterPartialS
 
         // The lambda will capture the copied reference and it will be unreferenced when the lambda's life cycle is over.
         // That happens when the Realm is closed or the callback has been triggered once.
-        JavaGlobalRef j_callback_ref(env, j_callback);
+        JavaGlobalWeakRef j_callback_ref(env, j_callback);
```
beeender: `j_callback_ref` has to be strong here since there is no strong ref to it in java. And it is totally safe to let the lambda capture the global ref. Everything lambda captured will be freed when there is lambda gets freed. In this case, the lambda will be freed when then `NotificationToken` is freed -- that is the time when realm get closed (`Realm.close()`) or the notification gets triggered once.

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
```diff
@@ -525,29 +525,34 @@ JNIEXPORT void JNICALL Java_io_realm_internal_SharedRealm_nativeRegisterPartialS
...
 
             JNIEnv* env = JniUtils::get_env(true);
             if (err) {
                 try {
                     std::rethrow_exception(err);
                 }
                 catch (const std::exception& e) {
-                    env->CallStaticVoidMethod(shared_realm_class, partial_sync_cb, to_jstring(env, e.what()),
-                                              reinterpret_cast<jlong>(nullptr), j_callback_ref.get());
+                    j_shared_realm_instance_ref.call_with_local_ref(env, [&](JNIEnv*, jobject row_obj) {
```
beeender: move `j_shared_realm_instance_ref.call_with_local_ref` one level out:
```
j_shared_realm_instance_ref.call_with_local_ref(env, [&](JNIEnv*, jobject row_obj) {
   if (err) {
   // ...
   } else {
   }
}
```

realm/realm-library/src/main/cpp/io_realm_internal_SharedRealm.cpp
nhachicha: as discussed in Slack we'll keep a weak reference to `j_callback_ref` just to prevent the use case where the user doesn't call `Realm.close` explicitly, and we end up with a strong reference to `Realm.this`, as passed in 
```Java
SharedRealm.PartialSyncCallback internalCallback = new SharedRealm.PartialSyncCallback(className) {
            @Override
            public void onSuccess(io.realm.internal.Collection osResults) {
                RealmResults<E> results = new RealmResults<>(Realm.this, osResults, clazz);
                callback.onSuccess(results);
            }
```

PR #5365, opened 2017-10-02
Add optimized first()/last() to RealmList's operator for models

Zhuinden:
While I was frantically trying to figure out how to re-introduce the optimization of `first()`/`last()` to `OrderedRealmCollectionImpl` (which was never actually merged), I found that there seems to be a place for optimizing RealmList's first/last methods.

I think it should work, but CI should be run on it.

PR #5367, opened 2017-10-02
Added model/value-specific first()/last() to ManagedListOperator

Zhuinden:
Continued https://github.com/realm/realm-java/pull/5365 .

I believe it should be working this time!

### DIFF
--- a/realm/realm-library/src/main/java/io/realm/RealmList.java
+++ b/realm/realm-library/src/main/java/io/realm/RealmList.java
@@ -444,8 +444,9 @@ public class RealmList<E> extends AbstractList<E> implements OrderedRealmCollect
     private E firstImpl(boolean shouldThrow, @Nullable E defaultValue) {
         if (isManaged()) {
             checkValidRealm();
-            if (!osListOperator.isEmpty()) {
-                return get(0);
+            E firstValue = osListOperator.first();
+            if (firstValue != null) {
+                return firstValue;
             }
         } else if (unmanagedList != null && !unmanagedList.isEmpty()) {
             return unmanagedList.get(0);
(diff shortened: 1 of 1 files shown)


Zhuinden:
> This commit cannot be built

....

zaki50:
The fail seems to be caused by our CI issue.

Zhuinden:
Can you re-run this?

zaki50:
@Zhuinden I couldn't find `rerun` button on CI. Could you push empty commit on the branch?

cmelchior:
We had to disable running CI for external commits due to other security concerns. We need to download the PR and run manually :(

Zhuinden:
Well, hopefully I didn't mess anything up this time, so if you find the time, then neat 😄 

But I actually needed `firstUncheckedRow` primarily for `RealmResults` which still has it, so this isn't critical for me right now. Basically an optimization for `RealmList.first()`/`RealmList.last()` if it is a model list. Assuming it works exactly as I expect.

cmelchior:
@Zhuinden 

Not entirely sure whatis wrong, but cloning master from your fork indicate that:
```
   @Nullable
    @Override
    public final T first() {
        UncheckedRow row = collection.firstUncheckedRow();
        if (row == null) {
            return null;
        }
        //noinspection unchecked
        return (T) realm.get((Class<? extends RealmModel>) clazz, className, row);
    }

    @Nullable
    @Override
    public final T last() {
        UncheckedRow row = collection.lastUncheckedRow();
        if (row == null) {
            return null;
        }
        //noinspection unchecked
        return (T) realm.get((Class<? extends RealmModel>) clazz, className, row);
    }
```

`collection` is an unknown variable in `RealmList`?

Zhuinden:
@cmelchior ......uerghh I couldn't get Android Studio to do even the most basics checks for me I guess 😭 

Technically it should be `getCollection()`.

Zhuinden:
the managed list operator talks to an `osList` which is not `internal.Collection` so I think that explains why this isn't a thing.

Maybe some other time

PR #5368, opened 2017-10-02
Prepare CHANGELOG for public release

cmelchior:
Combined `BETA1-3` + unreleased 3.7.3 into one changelog entry called `4.0.0-RC1`
Also added bug fix entries for 

```
+* "Bad File Header" caused by the device running out of space while compacting the Realm (#5011).
+* `RealmQuery.equalTo()` failed to find null values on an indexed field if using Case.INSENSITIVE (#5299).
```
They were solved on the Core side, and we forgot to add them to the changelog when updating Core.


CHANGELOG.md
```diff
@@ -1,98 +1,53 @@
-## 4.0.0 (YYYY-MM-DD)
+## 4.0.0-RC1 (YYYY-MM-DD)
 
 ## Breaking Changes
 
-* Calling `distinct()` on a sorted `RealmResults` no longer clears the sorting (#3503).
+The internal fileformat has been upgraded. Opening an older Realm will upgrade the file automatically, but older versions of Realm will can no longer be able to read the file.
```
beeender: `fileformat` -> `file format`. also please mention the file format version, should be version 9 i think. 
https://github.com/realm/realm-core/commit/a43e8449d3a20e37ae63d9564e825bcbfdad372f#diff-4ac32a78649ca5bdd8e0ba38b7006a1eR13

CHANGELOG.md
```diff
@@ -1,98 +1,53 @@
-## 4.0.0 (YYYY-MM-DD)
+## 4.0.0-RC1 (YYYY-MM-DD)
 
 ## Breaking Changes
 
-* Calling `distinct()` on a sorted `RealmResults` no longer clears the sorting (#3503).
+The internal fileformat has been upgraded. Opening an older Realm will upgrade the file automatically, but older versions of Realm will can no longer be able to read the file.  
+
+* [ObjectServer] Updated protocol version to 22 which is only compatible with Realm Object Server > 2.0.0.
```
beeender: `>` -> `>=`

CHANGELOG.md
cmelchior: I cannot find any mention of the fileformat version in Cores changelog. With that in mind, it seems a bit odd that we put it in ours?

beeender: The link i posted is not correct?

> Changing instruction values for `Table::move_row()` requires a version bump to 9.
>  Version 8 files in read only mode without any history can be opened without 

 
 But yeah, i guess it is OK to not mention it ...

Zhuinden:
The only additional thing is that `SyncUser.Callback to becomes generic.` isn't really valid english, so if you can change that to something lke `SyncUser.Callback is now generic` then that's awesome.

CHANGELOG.md
```diff
@@ -1,101 +1,55 @@
...
+## 4.0.0-RC1 (YYYY-MM-DD)
 
 ## Breaking Changes
 
-* Calling `distinct()` on a sorted `RealmResults` no longer clears the sorting (#3503).
+The internal file format has been upgraded. Opening an older Realm will upgrade the file automatically, but older versions of Realm will no longer be able to read the file.
+
+* [ObjectServer] Updated protocol version to 22 which is only compatible with Realm Object Server >= 2.0.0.
 * [ObjectServer] Removed deprecated APIs `SyncUser.retrieveUser()` and `SyncUser.retrieveUserAsync()`. Use `SyncUser.retrieveInfoForUser()` and `retrieveInfoForUserAsync()` instead.
+* [ObjectServer] `SyncUser.Callback` now accepts an generic parameter indicating type of object returned when `onSuccess` is called.
```
nhachicha: accepts a generic

CHANGELOG.md
```diff
@@ -1,101 +1,55 @@
...
   * `RealmResults.asObservable()` has been renamed to `RealmResults.asFlowable()`.
   * `RealmObject.asObservable()` has been renamed to `RealmObject.asFlowable()`.
   * `RxObservableFactory` now return RxJava2 types instead of RxJava1 types.
+* Removed deprecated APIs `RealmSchema.close()` and `RealmObjectSchema.close()`. Those don't have to be called anymore.
+* Removed deprecated API `RealmResults.removeChangeListeners()`. Use `RealmResults.removeAllChangeListeners()` instead.
+* Removed deprecated API `RealmObject.removeChangeListeners()`. Use `RealmObject.removeAllChangeListeners()` instead.
+* Removed `UNSUPPORTED_TABLE`, `UNSUPPORTED_MIXED` and `UNSUPPORTED_DATE` from `RealmFieldType`.
+* Removed deprecated API `RealmResults.distinct()`/`RealmResults.distinctAsync()`. Use `RealmQuery.distinct()`/`RealmQuery.distinctAsync()` instead.
+* `RealmQuery.createQuery(Realm, Class)`, `RealmQuery.createDynamicQuery(DynamicRealm, String)`, `RealmQuery.createQueryFromResult(RealmResults)` and `RealmQuery.createQueryFromList(RealmList)` have been removed. Use `Realm.where(Class)`, `DynamicRealm.where(String)`, RealmResults.where()` and `RealmList.where()` instead.
```
nhachicha: missing "`" in _RealmResults.where()_

cmelchior: Well spotted :)

Zhuinden:
@cmelchior I've just noticed that the `RealmObjectSchema` method in the change log is `setRealmListField` instead of `addRealmListField`. Kinda wish I had noticed earlier.

cmelchior:
Ups, should be fixed on `master`

PR #5371, opened 2017-10-03
Don't push to releases if it is beta release

beeender:


cmelchior:
Added manually. Closing

PR #5374, opened 2017-10-03
Make Sync Integration Tests more reliable.

cmelchior:
Enable a test that should now work. Ignore another we need to fix.


cmelchior:
Added this here: https://github.com/realm/realm-java/commit/0ab3d135335127c3d56b1f5f379fc05686160c9d

 Closing

PR #5384, opened 2017-10-04
Keep annotations at runtime

### LINKED ISSUE #5344: Annotations `@Ignore`, `@PrimaryKey`, `@Index` and `@Required` are not retained for RUNTIME
#### Goal

I'm tinkering with auto-migration based on reflection.

#### Expected Results

I would hope to be able to access the `@Ignore`, `@PrimaryKey`, `@Index` and `@Required` annotations with `field.getAnnotation(annotationClass)`.

#### Actual Results

The annotations are not retained for runtime, only for class.

#### Code Sample

I'm experimenting with this [here](https://github.com/Zhuinden/realm-auto-migration/blob/18063e72f1b9169953adddbf7232ea6a485a256f/auto-migration/src/main/java/com/zhuinden/realmautomigration/AutoMigration.java#L209) and I was hoping I wouldn't need to write an annotation processor for it 😄 

Of course, this is nothing critical, considering Realm doesn't use it for reflection, only for annotation processing, so `RetentionPolicy.CLASS` makes perfect sense.

So it's like, totally up to you. 

Technically I can just define my own annotations otherwise.

Realm version(s): 3.7.2

Android Studio version: 2.3.3

Which Android version and device: irrelevant

cmelchior:
Fixes #5344 

Thinking about this, we actually also need to keep these if we at some point support inheritance for model classes as the superclass might exist in another jar file. But for now, the primary driver is giving other library developers better tools for interacting with Realm.

The size increase of this in peoples apps should be minimal.

@Zhuinden does this fix your issue?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ## Enhancements
 
-* All Realm annotations are now kept at runtime, allowing runtime tools to access them (#5344).
+* All Realm annotations are now kept at runtime, allowing runtime tools access to them (#5344).
 
 ## Bug Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ## Enhancements
 
-* All Realm annotations are now kept at runtime, allowing runtime tools to access them (#XXXX).
+* All Realm annotations are now kept at runtime, allowing runtime tools to access them (#5344).
 
 ## Bug Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ## Enhancements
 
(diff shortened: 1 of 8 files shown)


Zhuinden:
Not sure if you really need `@Beta`, but yes, it does :slightly_smiling_face:

CHANGELOG.md
```diff
@@ -4,7 +4,7 @@
 
 ## Enhancements
 
-* All Realm annotations are now kept at runtime, allowing runtime tools to access them (#5344).
+* All Realm annotations are now kept at runtime, allowing runtime tools access to them (#5344).
```
Zhuinden: previous wording was better, current wording is actually valid if it's `allowing runtime tools' access to them` :open_mouth: 

Enforcing grammar is so strange though!

Zhuinden:
Thank you! :smile

Issue #5391, opened 2017-10-06
Realm is too slow on Android first call when compared to iOS


abhi-scenedoc:
#### Goal

> What do you want to achieve?
Faster return for realm.getInstance() call the first time it is executed

#### Expected Results

> I am noticing the iOS Realm is much faster at returning the Realm Instance the first time it is called, significantly faster (with the same schema setup). 

#### Actual Results

>  I am noticing it takes upwards of 1 second on the Samsung s8. On some devices it takes up to 8-9 seconds. 

#### Steps & Code to Reproduce

> Describe your current debugging efforts.
Please run this (https://github.com/abhi-scenedoc/realmTest) sample project that I have put together which displays the results (time taken is ms).

#### Code Sample

```java

long startTime = System.currentTimeMillis();
Realm realm = Realm.getDefaultInstance(); //first time it's called after application is opened
long result = (System.currentTimeMillis() - startTime);
Log.d(REALM, "total time = " + result);

```

#### Version of Realm and tooling
Realm version(s): 3.7.2(testing) and 3.3.1 (in our production app)

Realm sync feature enabled:no

Android Studio version: 2.3.3

Which Android version and device: Samsung S8 - Android 7.0

Zhuinden:
Any migration code or compact on launch?

abhi-scenedoc:
None, no migration code or compaction.

cmelchior:
Can you post your `RealmConfiguration`?

abhi-scenedoc:
For the test project or our actual application?

Zhuinden:
@cmelchior there are many classes, maybe it is schema validation? Although I think that is done by OS now.

nhachicha:
Hi @abhi-scenedoc Thanks for the test project I've installed your app locally, and I can confirm your observation. 

I've run the app on a `(Nexus 6P)` and `Nexus 5 Genymotion 5.1.0 API 22`, and I can notice a 3-4X factor for the time to open the default Realm,  between the version `3.7.2` and `3.6.0`.

I tested with [3.6.0](https://github.com/realm/realm-java/blob/master/CHANGELOG.md#360-2017-09-01) since it corresponds to the latest update to Core AFAICT.

We'll try to investigate and see what caused the performance drop. 

_Test results_

```
Realm 3.7.2 (Nexus 6P)
    1439ms
    1540ms
    1379ms
    1383ms
    1376ms

[...]

PR #5404, opened 2017-10-10
Fix some performance issues when initializing the Schema 

cmelchior:
This PR fixes some of the issues found in #5391

**3.7.2 - Nexus 6P**
1445
1397
1345
1350
1359
Avg: 1379

**4.0.0-SNAPSHOT with proxy class fixes**
1129
1067
1054
973
1030
Avg: 1051
Savings ~25%

**4.0.0-SNAPSHOT with proxy/mediator fixes**
1115
974
939
937
950
Avg: 983
Savings ~ 29%

**4.0.0-SNAPSHOT with no-alloc/proxy/mediator fixes**
988
944
966
959
997
Avg: 971
Savings ~ 30%

1 second to open a Realm still feels like too much though, but this was just low hanging fruits.

### DIFF
--- a/realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmProxyClassGenerator.java
+++ b/realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmProxyClassGenerator.java
@@ -723,8 +723,14 @@ public class RealmProxyClassGenerator {
                 "createExpectedObjectSchemaInfo", // Method name
                 EnumSet.of(Modifier.PRIVATE, Modifier.STATIC)); // Modifiers
 
+        // Guess capacity for Arrays used by OsObjectSchemaInfo.
+        // Used to prevent array resizing at runtime
+        int persistedFields = metadata.getFields().size();
+        int computedFields = metadata.getBacklinkFields().size();
+
         writer.emitStatement(
-                "OsObjectSchemaInfo.Builder builder = new OsObjectSchemaInfo.Builder(\"%s\")", this.simpleClassName);
+                "OsObjectSchemaInfo.Builder builder = new OsObjectSchemaInfo.Builder(\"%s\", %s, %s)",
+                this.simpleClassName, persistedFields, computedFields);
 
         // For each field generate corresponding table index constant
         for (VariableElement field : metadata.getFields()) {
--- a/realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmProxyClassGenerator.java
+++ b/realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmProxyClassGenerator.java
@@ -244,7 +244,7 @@ public class RealmProxyClassGenerator {
 
         writer.emitField("List<String>", "FIELD_NAMES", EnumSet.of(Modifier.PRIVATE, Modifier.STATIC, Modifier.FINAL));
         writer.beginInitializer(true)
-            .emitStatement("List<String> fieldNames = new ArrayList<String>()");
+            .emitStatement("List<String> fieldNames = new ArrayList<String>(%s)", metadata.getFields().size());
             for (VariableElement field : metadata.getFields()) {
                 writer.emitStatement("fieldNames.add(\"%s\")", field.getSimpleName().toString());
             }
--- a/realm/realm-library/src/main/cpp/io_realm_internal_OsObjectSchemaInfo.cpp
+++ b/realm/realm-library/src/main/cpp/io_realm_internal_OsObjectSchemaInfo.cpp
@@ -16,11 +16,14 @@
 
 #include "io_realm_internal_OsObjectSchemaInfo.h"
 
+#include <realm/util/assert.hpp>
+
 #include <object_schema.hpp>
 #include <property.hpp>
 
-#include "jni_util/java_exception_thrower.hpp"
+#include "java_accessor.hpp"
 #include "java_exception_def.hpp"
+#include "jni_util/java_exception_thrower.hpp"
 #include "util.hpp"
 
 using namespace realm;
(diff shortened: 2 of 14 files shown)


realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmProxyClassGenerator.java
```diff
@@ -723,8 +723,14 @@ private void emitCreateExpectedObjectSchemaInfo(JavaWriter writer) throws IOExce
...
 
+        // Guess capacity for Arrays used by OsObjectSchemaInfo.
+        // Used to prevent array resizing at runtime
+        int persistedFields = metadata.getFields().size();
+        int computedFields = metadata.getBacklinkFields().size();
+
         writer.emitStatement(
-                "OsObjectSchemaInfo.Builder builder = new OsObjectSchemaInfo.Builder(\"%s\")", this.simpleClassName);
+                "OsObjectSchemaInfo.Builder builder = new OsObjectSchemaInfo.Builder(\"%s\", %s, %s)",
+                this.simpleClassName, persistedFields, computedFields);
```
Zhuinden: I have this weird idea that you could probably just pass in the `metadata` as is, that way you wouldn't need to directly pass `int`s as parameters to the constructor.

cmelchior: Not entirely sure what you mean? `OsObjectSchemaInfo.Builder` contains two internal arrays. You mean that those should be constructed outside and then parsed in?

Zhuinden: Actually never mind, there's a lot more to do with these sizes than I thought (JNI)

cmelchior:
Next release is going to be from `master` anyway. Okay if I just keep it here?

beeender:
Yeah, master is fine to me

realm/realm-library/src/main/java/io/realm/internal/OsObjectSchemaInfo.java
```diff
@@ -57,8 +58,10 @@ public Builder(String className) {
          */
         public Builder addPersistedProperty(String name, RealmFieldType type, boolean isPrimaryKey, boolean isIndexed,
                 boolean isRequired) {
-            final Property property = new Property(name, type, isPrimaryKey, isIndexed, isRequired);
-            persistedPropertyList.add(property);
+            long propertyPtr = Property.nativeCreatePersistedProperty(name,
+                    Property.convertFromRealmFieldType(type, isRequired), isPrimaryKey, isIndexed);
+            persistedPropertyPtrArray[persistedPropertyPtrCurPos] = propertyPtr;
+            persistedPropertyPtrCurPos++;
```
nhachicha: why not `persistedPropertyPtrArray[persistedPropertyPtrCurPos++] = propertyPtr;`

realm/realm-library/src/main/java/io/realm/internal/OsObjectSchemaInfo.java
```diff
@@ -71,8 +74,10 @@ public Builder addPersistedProperty(String name, RealmFieldType type, boolean is
          * @return this {@code OsObjectSchemaInfo}.
          */
         public Builder addPersistedValueListProperty(String name, RealmFieldType type, boolean isRequired) {
-            final Property property = new Property(name, type, !Property.PRIMARY_KEY, !Property.INDEXED, isRequired);
-            persistedPropertyList.add(property);
+            long propertyPtr = Property.nativeCreatePersistedProperty(name,
+                    Property.convertFromRealmFieldType(type, isRequired), !Property.PRIMARY_KEY, !Property.INDEXED);
+            persistedPropertyPtrArray[persistedPropertyPtrCurPos] = propertyPtr;
```
nhachicha: ditto

realm/realm-library/src/main/cpp/io_realm_internal_OsObjectSchemaInfo.cpp
```diff
@@ -53,24 +56,33 @@ JNIEXPORT jlong JNICALL Java_io_realm_internal_OsObjectSchemaInfo_nativeGetFinal
...
+        ObjectSchema& object_schema = *reinterpret_cast<ObjectSchema*>(native_ptr);
+        JLongArrayAccessor persisted_properties(env, j_persisted_properties);
+        for (jsize i = 0; i < persisted_properties.size(); ++i)
+        {
+            Property* prop = reinterpret_cast<Property*>(persisted_properties[i]);
+            REALM_ASSERT_DEBUG(prop != nullptr);
+            if (prop->is_primary) {
+                object_schema.primary_key = prop->name;
             }
+            object_schema.persisted_properties.emplace_back(std::move(*prop));
```
nhachicha: @beeender I assume the container will free the dynamically allocated pointer when destructed (since it took ownership of the pointer when using the move ctor)

beeender: uh, no. the container will "empty" the `Property&&` which is `*prop` here. That means after this line, the `prop` is still a valid pointer which stores an empty `Property`.
The memory of the empty `Property`(pointed by `prop`) will be freed at line 76.
It will be easy to understand in this way i think:

```
Property& prop_ref = *prop; // nothing is alloced here
object_schema.persisted_properties.emplace_back(std::move(prop_ref)); 
// prop_ref has been "emptied" by the move constructor
// but it is still a valid ref since prop is still a valid pointer.
```

realm/realm-library/src/main/java/io/realm/internal/OsObjectSchemaInfo.java
beeender: I'd like this way since i always have problems to understand the operator priority :P

PR #5406, opened 2017-10-10
Dont use ApplicationContext

### LINKED ISSUE #5389: Can't initiate Realm Configuration in AttachBaseContext in Application Class.
#### Expected Results

Realm initialized with the RealmConfiguration.

#### Actual Results

java.lang.IllegalStateException: Call `Realm.init(Context)` before creating a RealmConfiguration

Can't set realm configuration in Application class in attachBaseContext. it is possible in OnCreate.

#### Code Sample

```java

@Override
	protected void attachBaseContext(final Context base) {
		initRealm(base);
}
private void initRealm(Context context) {
		Realm.init(context);

		try {
			RealmConfiguration configuration = new RealmConfiguration.Builder().name(REALM_NAME).schemaVersion(1).deleteRealmIfMigrationNeeded()
					.build();
			Realm.setDefaultConfiguration(configuration);
		} catch (Exception ignore) {
			ignore.printStackTrace();
		}

	}

```

Realm version(s): ?

3.7.2

no.

Android Studio version: ?

3.0 b7

Which Android version and device: ?

Emulator, any android version.

cmelchior:
Fixes #5389

Using the `ApplicationContext` was a safeguard against memory leaks in case people decided to accidentally send in an Activity, but apparently, the `ApplicationContext` is not available in `Application.attachBaseContext(Context)`, so this PR changes it so we just use the Context directly.

Since best practice is to init Realm in the `Application` class anyway and the Google documentation specifically warns against saving references to Activity contexts, making this change shouldn't really impact anyone in a negative way.


realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -246,7 +246,7 @@ public static synchronized void init(Context context) {
             RealmCore.loadLibrary(context);
             setDefaultConfiguration(new RealmConfiguration.Builder(context).build());
             ObjectServerFacade.getSyncFacadeIfPossible().init(context);
-            BaseRealm.applicationContext = context.getApplicationContext();
+            BaseRealm.applicationContext = context;
```
beeender: I'd like still handle the activity context case with 
```java
BaseRealm.applicationContext = context.getApplicationContext();
if (BaseRealm.applicationContext == null) {
    BaseRealm.applicationContext = context;
}
if (BaseRealm.applicationContext == null) {
    throw new IllegalArgumentException("The context is null");
}
```

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -67,6 +67,7 @@
 import io.realm.internal.Table;
 import io.realm.internal.async.RealmAsyncTaskImpl;
 import io.realm.log.RealmLog;
+import io.realm.permissions.Permission;
```
Zhuinden: ???

cmelchior: Ups

Issue #5416, opened 2017-10-13
Need a way to ensure the Realm instance on the sync client thread has been closed


beeender:
`SyncSessionStopPolicy::Immediately` is supposed to close the Realm instance on the sync client thread before the `close()` call on the last Realm instance returned. But it seems it is not doing correctly.

This results:

1. `Realm.deleteRealm()` cannot apply to synced Realm.
1. `RealmCache.doCreateRealmOrGetFromCache()` for `downloadRemoteChanges()` cannot clean the Realm file if the downloading failed.

nganthony:
Hey @beeender, will this issue be fixed soon? I currently have a login screen where the user logs into their Google account to sync all their data from Realm. If the user backs out of the login activity, I cancel the RealmAsyncTask which should close the Sync realm and delete the sync realm. I instead get a crash saying "It's not allowed to delete the file associated with an open Realm. Remember to close() all the instances of the Realm before deleting its file"

mrfaa:
Hello everybody, is there an update on this issue? I'm currently observing this behavior on my Android 9 device.

sebdomdev:
In 2023, his is still an issue. Will this be fixed?

AdelQraft:
This was in 2017. How is this not yet resolved? What a joke this library is this if one can't even delete a file?

nirinchev:
[Atlas Device Sync has been deprecated](https://www.mongodb.com/docs/atlas/app-services/sync/device-sync-deprecation/) so it's probably best to find another product that does synchronization and allows file deletion 😊

PR #5417, opened 2017-10-13
Process safe deleteRealm and assetFile

beeender:
- Use SharedRealm::call_with_lock() to make sure deleteRealm and
  assetFile calls are process safe.
- Also those calls will throw if the Realm instance on the sync client
  thread is still opened.
- Disallow deleteRealm on a synced Realm file. See #5416 .
- Manually delete SharedRealm pointer if the SharedRealm was created for
  callbacks, eg.: initialization_data_callback and migration_callback.


CHANGELOG.md
```diff
@@ -2,15 +2,19 @@
...
 
 * All Realm annotations are now kept at runtime, allowing runtime tools access to them (#5344).
 * Speedup schema initialization when a Realm file is first accessed (#5391).
+* `Realm.deleteRealm()` and `RealmConfiguration.assetFile()` are multi-processes safe now.
 
 ## Bug Fixes
 
 * Assigning a managed object's own list to itself would accidentally clear it (#5395).
 * Don't try to acquire `ApplicationContext` if not available in `Realm.init(Context)` (#5389).
+* Fix some potential database corruption caused by deleting realm file while a Realm instance are still opened on the other process or sync client thread.
```
Zhuinden: in another process?

realm/realm-library/src/androidTest/java/io/realm/internal/CollectionTests.java
```diff
@@ -74,6 +72,18 @@ public void tearDown() {
     }
 
     private SharedRealm getSharedRealm() {
+        RealmConfiguration config;
+        config = configFactory.createConfiguration();
```
nhachicha: why not `RealmConfiguration config = configFactory.createConfiguration();` ?

realm/realm-library/src/androidTest/java/io/realm/internal/CollectionTests.java
```diff
@@ -74,6 +72,18 @@ public void tearDown() {
...
+        return getSharedRealm(config);
+    }
+
+    private SharedRealm getSharedRealmForLooper() {
+        RealmConfiguration config;
+        config = looperThread.createConfiguration();
```
nhachicha: ditto

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -1657,14 +1657,24 @@ public static void migrateRealm(RealmConfiguration configuration, @Nullable Real
     }
 
     /**
-     * Deletes the Realm file specified by the given {@link RealmConfiguration} from the filesystem.
+     * Deletes the Realm file along with the related temporary files specified by the given {@link RealmConfiguration}
+     * from the filesystem. Temporary file with ".lock" extension won't be deleted.
```
nhachicha: is Core responsible for deleting the lock file? and when?

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -492,20 +494,31 @@ synchronized void invokeWithLock(Callback0 callback) {
...
+                    if (realmFileFromAsset != null) {
+                        copyFileIfNeeded(configuration.getAssetFilePath(), realmFileFromAsset);
+                    }
 
-        // Copy Sync Server certificate path if available
-        String syncServerCertificateAssetName = ObjectServerFacade.getFacade(configuration.isSyncConfiguration()).getSyncServerCertificateAssetName(configuration);
-        if (!Util.isEmptyString(syncServerCertificateAssetName)) {
-            String syncServerCertificateFilePath = ObjectServerFacade.getFacade(configuration.isSyncConfiguration()).getSyncServerCertificateFilePath(configuration);
+                    // Copy Sync Server certificate path if available
+                    if (!Util.isEmptyString(syncServerCertificateAssetName)) {
```
nhachicha: use a final `boolean` in  https://github.com/realm/realm-java/pull/5417/files#diff-79e6ca3beee0a83dd76772285b606694R504 to the redundant check again

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -492,20 +494,31 @@ synchronized void invokeWithLock(Callback0 callback) {
...
-        }
+    private static void copyAssetFileIfNeeded(final RealmConfiguration configuration) {
+        final File realmFileFromAsset = configuration.hasAssetFile() ?
+                new File(configuration.getRealmDirectory(), configuration.getRealmFileName())
+                : null;
+        final String syncServerCertificateAssetName = ObjectServerFacade.getFacade(
+                configuration.isSyncConfiguration()).getSyncServerCertificateAssetName(configuration);
+
+        if (realmFileFromAsset!= null || !Util.isEmptyString(syncServerCertificateAssetName)) {
+            OsObjectStore.callWithLock(configuration, new Runnable() {
```
nhachicha: why copying the asset using the lock? this is part of the initialization ... I rather have the entire initialization done within a lock (rather than parts of it) ...

realm/realm-library/src/main/java/io/realm/internal/OsObjectStore.java
```diff
@@ -71,6 +73,19 @@ public static boolean deleteTableForObject(SharedRealm sharedRealm, String class
         return nativeDeleteTableForObject(sharedRealm.getNativePtr(), className);
     }
 
+    /**
+     * Try to grab a exclusive lock on the given Realm file. If the lock can be acquired, the {@code runnable} will
```
nhachicha: an exclusive

realm/realm-library/src/main/java/io/realm/internal/OsObjectStore.java
```diff
@@ -71,6 +73,19 @@ public static boolean deleteTableForObject(SharedRealm sharedRealm, String class
...
+     * executed while the lock is hold. The lock will ensure no one else can read or write the Realm file at the same
```
nhachicha: is held

realm/realm-library/src/main/java/io/realm/internal/OsObjectStore.java
```diff
@@ -71,6 +73,19 @@ public static boolean deleteTableForObject(SharedRealm sharedRealm, String class
...
+     * time.
+     *
+     * @param configuration to specify the realm path.
+     * @param runnable to run with lock.
+     * @return true if the lock can be acquired and the {@code runnable} has been executed.
```
nhachicha: `{@code true}`

realm/realm-library/src/main/java/io/realm/internal/OsObjectStore.java
```diff
@@ -71,6 +73,19 @@ public static boolean deleteTableForObject(SharedRealm sharedRealm, String class
...
+     */
+    public static boolean callWithLock(RealmConfiguration configuration, Runnable runnable) {
```
nhachicha: What happens in case of an exception (within the Runnable) is the lock released? How are we reporting unchecked exception?

realm/realm-library/src/main/java/io/realm/internal/SharedRealm.java
```diff
@@ -208,13 +210,16 @@ private SharedRealm(OsRealmConfig osRealmConfig) {
...
+     * <p>
+     * WARNING: The {@code SharedRealm} pointer holds by this instance is not managed by the
+     * {@link java.lang.ref.PhantomReference}. {@link #delete()} needs to be called to free the memory.
      */
     private SharedRealm(long nativeSharedRealmPtr, OsRealmConfig osRealmConfig) {
         this.nativePtr = nativeSharedRealmPtr;
         this.osRealmConfig = osRealmConfig;
         this.schemaInfo = new OsSchemaInfo(nativeGetSchemaInfo(nativePtr), this);
         this.context = osRealmConfig.getContext();
-        this.context.addReference(this);
```
nhachicha: does it mean we shouldn't `this.context.addReference(this);` when `isPhantomRef` is `true`? 

what's the purpose of keeping the `NativeObject` interface if we don't add the reference(`addReference`)

realm/realm-library/src/main/java/io/realm/internal/Util.java
```diff
@@ -96,46 +98,45 @@ public static boolean isEmptyString(@Nullable String str) {
...
 
         // Deletes files in management directory and the directory.
         // There is no subfolders in the management directory.
         File[] files = managementFolder.listFiles();
         if (files != null) {
             for (File file : files) {
-                realmDeleted = realmDeleted && file.delete();
+                boolean deleteResult = file.delete();
+                if (!deleteResult) {
+                    RealmLog.warn( String.format(Locale.ENGLISH,"Realm temporary file at %s cannot deleted",
```
nhachicha: cannot be deleted

realm/realm-library/src/main/java/io/realm/internal/Util.java
```diff
@@ -96,46 +98,45 @@ public static boolean isEmptyString(@Nullable String str) {
...
-                if (!deleteResult) {
-                    realmDeleted.set(false);
-                    RealmLog.warn("Could not delete the file %s", fileToDelete);
-                }
+        boolean realmDeleted;
+        if (realmFile.exists()) {
+            realmDeleted = realmFile.delete();
+            if (!realmDeleted) {
+                RealmLog.warn(
+                        String.format(Locale.ENGLISH,"Realm file at %s cannot deleted", realmFile.getAbsolutePath()));
```
nhachicha: cannot be deleted

realm/realm-library/src/syncIntegrationTest/java/io/realm/SSLConfigurationTests.java
```diff
@@ -65,12 +65,12 @@ public void trustedRootCA() throws InterruptedException {
         SystemClock.sleep(TimeUnit.SECONDS.toMillis(2));  // FIXME: Replace with Sync Progress Notifications once available.
         realm.close();
         user.logout();
-        Realm.deleteRealm(configOld);
 
         // 2. Local state should now be completely reset. Open the Realm again with a new configuration which should
         // download the uploaded changes.
         user = SyncUser.login(SyncCredentials.usernamePassword(username, password), Constants.AUTH_URL);
         SyncConfiguration config = configurationFactory.createSyncConfigurationBuilder(user, Constants.USER_REALM_SECURE)
+                .name("newRealm")
```
nhachicha: These tests are disabled for now, https://github.com/realm/ros/issues/512 any reason to add the name?

realm/realm-library/src/main/java/io/realm/internal/SharedRealm.java
beeender: See https://github.com/realm/realm-java/pull/5417/files#diff-6e1f09495610b341d4f834a4e13002ccR199

In normal case it will still be managed by phantom reference.
This constructor is a special case and only handled in this file.

realm/realm-library/src/syncIntegrationTest/java/io/realm/SSLConfigurationTests.java
beeender: because of we sadly don't have a reliable way to know when the sync client thread will close the Realm instance.

https://github.com/realm/realm-java/pull/5417/files#diff-c4f8c4a7fed4b6347565a6e587f5da61L68

realm/realm-library/src/main/java/io/realm/internal/OsObjectStore.java
beeender: yes, the lock will be released. 
I planned to make a test case for this but forgot it finally ... will add a test case.

CHANGELOG.md
```diff
@@ -2,15 +2,19 @@
 
 ## Breaking Changes
 
+* `Realm.deleteRealm()` doesn't support synced Realm for now. See issue #5416.
```
cmelchior: Realm -> Realms

CHANGELOG.md
```diff
@@ -2,15 +2,19 @@
...
 
 * All Realm annotations are now kept at runtime, allowing runtime tools access to them (#5344).
 * Speedup schema initialization when a Realm file is first accessed (#5391).
+* `Realm.deleteRealm()` and `RealmConfiguration.assetFile()` are multi-processes safe now.
 
 ## Bug Fixes
 
 * Assigning a managed object's own list to itself would accidentally clear it (#5395).
 * Don't try to acquire `ApplicationContext` if not available in `Realm.init(Context)` (#5389).
+* Fix some potential database corruption caused by deleting realm file while a Realm instance are still opened in another process or the sync client thread.
```
cmelchior: deleting realm file -> deleting the Realm file

realm/realm-library/src/androidTest/java/io/realm/RealmTests.java
```diff
@@ -2302,7 +2302,11 @@ public void run() {
         assertTrue(Realm.deleteRealm(configuration));
 
         // Directory should be empty now.
```
cmelchior: That note doesn't seem correct given that the length is 0. I assume it is the `.lock` file. Clarify this or just move the comment below up instead of this one?

realm/realm-library/src/main/cpp/io_realm_internal_OsObjectStore.cpp
```diff
@@ -140,3 +141,25 @@ JNIEXPORT jboolean JNICALL Java_io_realm_internal_OsObjectStore_nativeDeleteTabl
...
+    try {
+        JStringAccessor path_accessor(env, j_realm_path);
+        std::string realm_path(path_accessor);
+        static JavaClass runnable_class(env, "java/lang/Runnable");
+        static JavaMethod run_method(env, runnable_class, "run", "()V");
+        bool exception_thrown = false;
+        bool result = SharedGroup::call_with_lock(realm_path, [&](std::string path) {
+            REALM_ASSERT_RELEASE_EX(realm_path.compare(path) == 0, realm_path.c_str(), path.c_str());
+            env->CallVoidMethod(j_runnable, run_method);
+            exception_thrown = env->ExceptionCheck();
```
cmelchior: This callback is made on the same thread right, so using the `env` should be safe here?

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -1657,14 +1657,24 @@ public static void migrateRealm(RealmConfiguration configuration, @Nullable Real
...
+     * <p>
      * All Realm instances must be closed before calling this method.
+     * <p>
+     * This doesn't support deleting synced Realm for now.
      *
      * @param configuration a {@link RealmConfiguration}.
-     * @return {@code false} if a file could not be deleted. The failing file will be logged.
+     * @return {@code false} if the realm file could not be deleted. Temporary files deletion failure won't impact on
```
cmelchior: remove `on`

cmelchior: realm -> Realm

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -1657,14 +1657,24 @@ public static void migrateRealm(RealmConfiguration configuration, @Nullable Real
...
+     * the return value. All of the failing file will be logged.
```
cmelchior: Not sure I understand `All of the failing file will be logged` ?

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -1657,14 +1657,24 @@ public static void migrateRealm(RealmConfiguration configuration, @Nullable Real
...
      * @throws IllegalStateException if not all realm instances are closed.
+     * @throws IllegalStateException if the {@code configuration} is a {@link SyncConfiguration}.
      */
     public static boolean deleteRealm(RealmConfiguration configuration) {
+        // FIXME: We don't know when the Realm instance on the sync client thread will be closed. Disable it for now.
+        if (configuration.isSyncConfiguration()) {
+            throw new IllegalStateException("'deleteRealm() is not supported for sync for now.");
```
cmelchior: `not supported for synchronized Realms for now.`

realm/realm-library/src/main/java/io/realm/internal/OsObjectStore.java
cmelchior: willl be

realm/realm-library/src/main/java/io/realm/internal/OsObjectStore.java
cmelchior: to the Realm file

realm/realm-library/src/main/java/io/realm/internal/SharedRealm.java
```diff
@@ -196,6 +197,7 @@ private SharedRealm(OsRealmConfig osRealmConfig) {
         this.schemaInfo = new OsSchemaInfo(nativeGetSchemaInfo(nativePtr), this);
         this.context = osRealmConfig.getContext();
         this.context.addReference(this);
+        this.isPhantomRef = true;
```
cmelchior: Would `isManagedByPhantomRef` be more clear? Right now it feels a bit unclear to me

realm/realm-library/src/main/java/io/realm/internal/SharedRealm.java
```diff
@@ -482,8 +499,13 @@ private void executePendingRowQueries() {
     @SuppressWarnings("unused")
     private static void runMigrationCallback(long nativeSharedRealmPtr, OsRealmConfig osRealmConfig, MigrationCallback callback,
                                              long oldVersion) {
-        callback.onMigrationNeeded(new SharedRealm(nativeSharedRealmPtr, osRealmConfig), oldVersion,
-                osRealmConfig.getRealmConfiguration().getSchemaVersion());
+        SharedRealm sharedRealm = new SharedRealm(nativeSharedRealmPtr, osRealmConfig);
```
cmelchior: Not sure I understand the reason for making this change?

realm/realm-library/src/main/java/io/realm/internal/Util.java
```diff
@@ -96,46 +98,45 @@ public static boolean isEmptyString(@Nullable String str) {
         return str == null || str.length() == 0;
     }
 
+    /**
+     * To delete Realm and related temporary files. This must be called in
+     * {@link OsObjectStore#callWithLock(RealmConfiguration, Runnable)}'s callback.
+     *
+     * @return {@code true} if the realm file is deleted. Temporary file deletion failure will not impact on return
```
cmelchior: on return -> the return

realm/realm-library/src/main/java/io/realm/Realm.java
beeender: no, lock file should never be deleted. There is no reliable way to delete it for multi processes case.

realm/realm-library/src/main/java/io/realm/RealmCache.java
beeender: the problem is, there could be another process/sync client opened the Realm at the same time. so the asset file could be copied already.
When core is holding the lock file, the realm file is not necessarily to be there.

it is like below:

core holding the lock file ->
check if file existence ->
copy the file ->
core release the lock

realm/realm-library/src/main/java/io/realm/internal/Util.java
cmelchior: Also, it will impact the return value right? `false` is returned if they cannot be deleted?

realm/realm-library/src/main/cpp/io_realm_internal_OsObjectStore.cpp
beeender: yes, it is a synchronized callback. It will return before the outer function returns.

realm/realm-library/src/main/java/io/realm/Realm.java
beeender: will change it to `all the failing file deletion`

realm/realm-library/src/main/java/io/realm/internal/SharedRealm.java
beeender: complex story.
take the migration for example, ObjectStore will pass a `SharedRealm` instance to the callback
to make the life control easier, i created a new pointer from that instance and pass it to java, so it can be managed like a normal `SharedRealm`, user will never meet a invalid pointer of `SharedRealm` even when he passed that `SharedRealm` instance out.

But then it is a problem when exception happens in migration block. after exception thrown, `getInstance()` hasn't return anything yet, so user doesn't have to call `close()`. But there is a `SharedRealm` pointer created in the migration callback which is waiting for the GC to be deleted. That means, from OS's RealmCoordinator pointer of view, there is still one `SharedRealm` is not dealloced. -- call `deleteRealm()` at this time will throw.

But i think i still miss one thing in this modification, `SharedRealm.isClosed()` should return true after `delete()` called to avoid use pass the `Realm`/`DyanmicRealm` instance out of the callback and see a deleted pointer.

realm/realm-library/src/main/java/io/realm/internal/Util.java
beeender: no, it will always return true if the realm file has been deleted.

I think it is just good enough to ensure the realm file is deleted since they take most of disk spaces.

Also, for the Realm on external storage, the temporary files are not deleted since we don't know where are they. (the logic of this part is decided by core at runtiem :( where all the fifo permission nightmare come from)

realm/realm-library/src/main/java/io/realm/internal/SharedRealm.java
nhachicha: `*isPhantomRef` is no longer used? ( you removed the `delete` method) ...?

realm/realm-library/src/androidTest/java/io/realm/rule/TestRealmConfigurationFactory.java
```diff
@@ -91,6 +94,23 @@ protected void after() {
         }
     }
 
+    @Override
+    public void create() throws IOException {
+        super.create();
+        tempFolder = new File(super.getRoot(), testName);
```
nhachicha: nice move 👍

realm/realm-library/src/main/cpp/io_realm_internal_OsRealmConfig.cpp
```diff
@@ -142,18 +142,22 @@ JNIEXPORT void JNICALL Java_io_realm_internal_OsRealmConfig_nativeSetSchemaConfi
...
                     env->CallStaticVoidMethod(get_shared_realm_class(env), run_migration_callback_method,
-                                              reinterpret_cast<jlong>(&realm), config_global.get(), obj,
+                                              reinterpret_cast<jlong>(new_shared_realm_ptr), config_global.get(), obj,
                                               old_realm->schema_version());
                 });
-                TERMINATE_JNI_IF_JAVA_EXCEPTION_OCCURRED(env);
+                // Close the SharedRealm. Otherwise it will only be closed when the Java OsSharedRealm gets GCed. And
+                // that will be too late.
+                TERMINATE_JNI_IF_JAVA_EXCEPTION_OCCURRED(
+                    env, [&new_shared_realm_ptr]() { (*new_shared_realm_ptr)->close(); });
```
nhachicha: do we need to `delete new_shared_realm_ptr` as well?

realm/realm-library/src/main/cpp/io_realm_internal_OsRealmConfig.cpp
```diff
@@ -214,17 +218,21 @@ JNIEXPORT void JNICALL Java_io_realm_internal_OsRealmConfig_nativeSetInitializat
...
                     env->CallStaticVoidMethod(get_shared_realm_class(env), run_initialization_callback_method,
-                                              reinterpret_cast<jlong>(&realm), config_global_ref.get(),
+                                              reinterpret_cast<jlong>(new_shared_realm_ptr), config_global_ref.get(),
                                               obj);
                 });
-                TERMINATE_JNI_IF_JAVA_EXCEPTION_OCCURRED(env);
+                // Close the SharedRealm. Otherwise it will only be closed when the Java OsSharedRealm gets GCed. And
+                // that will be too late.
+                TERMINATE_JNI_IF_JAVA_EXCEPTION_OCCURRED(
+                    env, [&new_shared_realm_ptr]() { (*new_shared_realm_ptr)->close(); });
```
nhachicha: ditto

realm/realm-library/src/main/java/io/realm/internal/SharedRealm.java
beeender: I removed both in the yesterday's change.
So the pointer itself will be managed like a normal `SharedRealm` pointer. So there is no chance for users to meet a invalid pointer anymore. (before yeterday's change, there was still a problem.)

See commit https://github.com/realm/realm-java/pull/5417/commits/ed305d2b8525def04887caff6887b2a0d3a3baa2

realm/realm-library/src/main/cpp/io_realm_internal_OsRealmConfig.cpp
beeender: nope~ it will be still handled by the GC and phantom refs.

PR #5427, opened 2017-10-15
Fix wrong Javadoc

### LINKED ISSUE #5422: Error in javadoc of findAllAsync(): "Finds all objects that fulfill the query conditions and sorted by specific field name"
There isn't a fieldName parameter. Someone copypasted the description from findAllSorted which is below.

https://github.com/realm/realm-java/blob/d9e0ae4ac083286c5440cc4bdb0f18d9c70fdb45/realm/realm-library/src/main/java/io/realm/RealmQuery.java#L1805-L1812

I checked and it isn't fixed in 4.0

https://github.com/realm/realm-java/blob/master-4.0/realm/realm-library/src/main/java/io/realm/RealmQuery.java#L1805

cmelchior:
Fixes #5422

### DIFF
--- a/realm/realm-library/src/main/java/io/realm/RealmQuery.java
+++ b/realm/realm-library/src/main/java/io/realm/RealmQuery.java
@@ -1763,8 +1763,7 @@ public class RealmQuery<E extends RealmModel> {
     }
 
     /**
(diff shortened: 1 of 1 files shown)


cmelchior:
Fast merge

PR #5430, opened 2017-10-16
Enable SSL integration tests

beeender:
- Some features were missing which caused ssl tests failures.
  See https://github.com/realm/openssl-android/pull/2
  Linking with pre-build openssl 1.0.2k-1.
- Use a new Realm instead of using deleteRealm()

### DIFF
--- a/realm/realm-library/src/main/cpp/CMakeLists.txt
+++ b/realm/realm-library/src/main/cpp/CMakeLists.txt
@@ -91,7 +91,8 @@ use_realm_core(${build_SYNC} "${REALM_CORE_DIST_DIR}" "${CORE_SOURCE_PATH}")
 set(openssl_build_TYPE "release")
 # FIXME Read the openssl version from core when the core/sync release has that information.
 set(openssl_VERSION "1.0.2k")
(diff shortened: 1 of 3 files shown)


realm/realm-library/src/main/cpp/CMakeLists.txt
```diff
@@ -91,7 +91,8 @@ use_realm_core(${build_SYNC} "${REALM_CORE_DIST_DIR}" "${CORE_SOURCE_PATH}")
 set(openssl_build_TYPE "release")
 # FIXME Read the openssl version from core when the core/sync release has that information.
 set(openssl_VERSION "1.0.2k")
-set(openssl_FILENAME "openssl-${openssl_build_TYPE}-${openssl_VERSION}-Android-${ANDROID_ABI}")
+set(openssl_BUILD_NUMBER "1")
```
cmelchior: Perhaps add a comment that this number should match https://github.com/realm/openssl-android/pull/2/files#diff-5e3269174d70512b79d0b874838be93fR1 ?

beeender: for the next core release, this will be removed since core will ship a openssl lib. no need to download anymore.

PR #5431, opened 2017-10-16
Track down flaky tests

cmelchior:
Attempts to fix flaky integration tests:

* Centralize generation of random values
* Fail if a random value is generated twice.


cmelchior:
Replaced by other PR's.

PR #5433, opened 2017-10-17
Add realm.ignoreKotlinNullability

### LINKED ISSUE #5412: Realm migration between 3.5.0 and 3.7.2 crash
Related to #5270

#### Expected Results

> No migration involved.

#### Actual Results

>  App is crashing during the Realm initialization, all Realm objects were forced to migrate, even if there is no changes on these objects.

> Send privately to help@realm.io.

Realm version(s): 3.7.2

Realm sync feature enabled: No

Android Studio version: 3.0 Beta 7

Which Android version and device: All

beeender:
To disable treating kotlin non-null types as @Required.
From v3.6.0, all the kotlin RealmModel's non-null field will be set as
required in the schema. However it is a breaking change for the kotlin
project which has a Realm file created before 3.6.0.

To disable this behaviour introduced in 3.6.0, add below things to the
project's build.gradle:

kapt {
  arguments {
    arg("realm.ignoreKotlinNullability", true)
  }
}

Close #5412

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## 4.0.1 (YYYY-MM-DD)
+
+## Bug Fixes
+
+* Added `realm.ignoreKotlinNullability` as a kapt argument to disable treating kotlin non-null types as `@Required` (#5412) (introduced in `v3.6.0`).
+
+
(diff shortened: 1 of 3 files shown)


Zhuinden:
@ppanorim will be happy about this

PR #5447, opened 2017-10-20
Increase http timeout

beeender:
Otherwise integration tests would fail when setting the network type to
GSM on emulator.


beeender:
reported by https://stackoverflow.com/questions/46837982/realm-object-server-sync-on-low-connection

Issue #5403, opened 2017-10-10
Test: Reconnect not happening correctly


cmelchior:
Leaving a phone in airplane mode for a couple of minutes, then try to activate it again. The Java layer does not seem to handle this well resulting in the access token not being refreshed

```
10-10 13:58:15.082 20935-21485/io.realm.examples.objectserver E/REALM_SYNC: Connection[2]: Failed to connect to endpoint '192.168.87.111:9080': Network is unreachable
10-10 13:58:15.083 20935-21485/io.realm.examples.objectserver E/REALM_SYNC: Connection[2]: Failed to connect to '192.168.87.111:9080': All endpoints failed
10-10 13:58:15.083 20935-21485/io.realm.examples.objectserver D/REALM_SYNC: Connection[2]: Reconnecting in 55173 milliseconds
10-10 13:58:20.776 20935-20935/io.realm.examples.objectserver D/REALM_JAVA: NetworkListener: Connection available
10-10 13:58:47.485 20935-22048/io.realm.examples.objectserver D/REALM_JAVA: Network request (authenticate): http://192.168.87.111:9080/auth
10-10 13:58:54.420 20935-20935/io.realm.examples.objectserver D/REALM_JAVA: NetworkListener: Connection available
10-10 13:58:54.422 20935-20935/io.realm.examples.objectserver D/REALM_JAVA: NetworkListener: Connection available
10-10 13:58:59.640 20935-22048/io.realm.examples.objectserver D/REALM_JAVA: AuthenticateResponse - Error: UNKNOWN(-1)
                                                                            java.net.SocketException: Connection reset
                                                                                at java.net.SocketInputStream.read(SocketInputStream.java:209)
                                                                                at java.net.SocketInputStream.read(SocketInputStream.java:139)
                                                                                at okio.Okio$2.read(Okio.java:138)
                                                                                at okio.AsyncTimeout$2.read(AsyncTimeout.java:236)
                                                                                at okio.RealBufferedSource.indexOf(RealBufferedSource.java:344)
                                                                                at okio.RealBufferedSource.readUtf8LineStrict(RealBufferedSource.java:216)
                                                                                at okio.RealBufferedSource.readUtf8LineStrict(RealBufferedSource.java:210)
                                                                                at okhttp3.internal.http1.Http1Codec.readResponseHeaders(Http1Codec.java:189)
                                                                                at okhttp3.internal.http.CallServerInterceptor.intercept(CallServerInterceptor.java:75)
[...]

PR #5453, opened 2017-10-23
fix SocketTimeout

nhachicha:
fixes https://github.com/realm/realm-java/issues/5403 now we should retry when encountering any `IOException` error

### DIFF
--- a/realm/realm-library/src/objectServer/java/io/realm/ErrorCode.java
+++ b/realm/realm-library/src/objectServer/java/io/realm/ErrorCode.java
@@ -16,10 +16,8 @@
 
 package io.realm;
 
-import java.net.ConnectException;
-import java.net.SocketTimeoutException;
 
-import io.realm.internal.IOException;
+import java.io.IOException;
 
 /**
  * This class enumerate all potential errors related to using the Object Server or synchronizing data.
--- a/realm/realm-library/src/objectServer/java/io/realm/ErrorCode.java
+++ b/realm/realm-library/src/objectServer/java/io/realm/ErrorCode.java
@@ -17,6 +17,9 @@
 package io.realm;
 
 import java.net.ConnectException;
+import java.net.SocketTimeoutException;
+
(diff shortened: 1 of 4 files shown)


realm/realm-library/src/objectServer/java/io/realm/ErrorCode.java
```diff
@@ -199,7 +200,11 @@ public static ErrorCode fromInt(int errorCode) {
      */
     public static ErrorCode fromException(Exception exception) {
         // ConnectException is recoverable (with exponential backoff)
```
cmelchior: Remove this comment?

cmelchior:
@nhachicha I'm slightly concerned that this times out? This could indicate that there are some cases where it goes into infinite exponential backoff outside our normal kill switches?

nhachicha:
@cmelchior 
A simple use case where this happens is when you have a valid token, then stop ROS, subsequent refresh queries will `SocketTimeout` ...

I don't think it's an issue since for example `ExponentialBackoff` will reach `4h 33m` of sleep after 15 retries... can't imagine an app open for that long anyway, it will either reach a large sleep value or the user will restart it at some point ...

cmelchior:
Well, it wasn't apps I was concerned about, but our CI 😄

realm/realm-library/src/objectServer/java/io/realm/internal/network/ExponentialBackoffTask.java
```diff
@@ -69,7 +78,7 @@ public void run() {
                 onSuccess(response);
                 break;
             } else {
-                if (shouldAbortTask(response)) {
+                if (shouldAbortTask(response) || attempt == maxRetries + 1) {
```
beeender: what about if the network drops and reconnect? is it handled by the other logic?

nhachicha: yes

PR #5466, opened 2017-10-25
Fix a socket timeout issue with test server

beeender:
If startRealmObjectServer was called twice without stop in between,
syncServerChildProcess will be overwritten by a terminated process since
the 2nd call of creating sync server will fail and set the syncServerChildProcess.
This would make the stopRealmObjectServer get stuck forever.

In practice, we need to make sure the server has been stopped before
start it. So just treat the "start" as a "restart" command which will
stop the sync server if needed.

Thus, no need to stop in tear down functions in java side anymore.

### DIFF
--- a/realm/realm-library/src/syncIntegrationTest/java/io/realm/StandardIntegrationTest.java
+++ b/realm/realm-library/src/syncIntegrationTest/java/io/realm/StandardIntegrationTest.java
@@ -35,10 +35,6 @@ public abstract class StandardIntegrationTest extends BaseIntegrationTest {
         startSyncServer();
     }
 
-    @AfterClass
-    public static void tearDownTestClass() throws Exception {
-    }
-
     @Before
     public void setupTest() throws IOException {
         prepareEnvironmentForTest();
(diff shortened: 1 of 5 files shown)


realm/realm-library/src/syncIntegrationTest/java/io/realm/StandardIntegrationTest.java
```diff
@@ -37,7 +37,6 @@ public static void setupTestClass() throws Exception {
 
     @AfterClass
     public static void tearDownTestClass() throws Exception {
-        stopSyncServer();
```
cmelchior: Just remove the method completely?

beeender:
yup, it shouldn't matter ~

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/utils/HttpUtils.java
```diff
@@ -63,6 +66,10 @@ public static void startSyncServer() throws Exception {
         SystemClock.sleep(2000);
```
nhachicha: We should remove the sleep, issue above was resolved https://github.com/realm/ros/issues/389

beeender: Nice! will remove it in another PR

PR #5470, opened 2017-10-26
Enable downloadProgressListener_indefinitely

beeender:
In createRemoteData we try to identify the uploading is finished by two
conditions:

- Check if the newly created data exists in the Realm on the callback thread.
- Check if uploadable == uploaded

This works if the callback is called on the sync client thread.
But actually there is a chance the callback gets called immediately after
registered because of some other data created before just got uploaded. In
this case, the data does exist in the Realm on the caller thread (we just
add on the same thread), and the uploaded could equal to the uploadable.

By only accept the callback on the sync client thread by if (threadId ==
Thread.currentThread().getId()) will make sure the data has been uploaded.

### DIFF
--- a/realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/ProgressListenerTests.java
+++ b/realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/ProgressListenerTests.java
@@ -18,7 +18,6 @@ package io.realm.objectserver;
 
 import android.support.test.runner.AndroidJUnit4;
 
-import org.junit.Ignore;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.runner.RunWith;
@@ -35,6 +34,7 @@ import io.realm.Progress;
 import io.realm.ProgressListener;
 import io.realm.ProgressMode;
 import io.realm.Realm;
+import io.realm.RealmConfiguration;
 import io.realm.StandardIntegrationTest;
 import io.realm.SyncConfiguration;
 import io.realm.SyncManager;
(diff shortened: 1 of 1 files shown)


realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/ProgressListenerTests.java
```diff
@@ -83,19 +83,23 @@ private void assertTransferComplete(Progress progress, boolean nonZeroChange) {
...
         writeSampleData(realm);
 
+        final long threadId = Thread.currentThread().getId();
+
         session.addUploadProgressListener(ProgressMode.INDEFINITELY, new ProgressListener() {
             @Override
             public void onChange(Progress progress) {
+                // FIXME: This check is to make sure before this method returns, all the uploads has been done.
+                // See https://github.com/realm/realm-object-store/issues/581#issuecomment-339353832
+                if (threadId == Thread.currentThread().getId()) {
```
cmelchior: Shouldn't this check be moved into the library implementation? If we have this bug, end users will also run into it?

beeender: I think the API design is actually buggy, moving the checking to implementation will create other issues -- like user wants to get the progress made by other threads.

We really need to address https://github.com/realm/realm-object-store/issues/581#issuecomment-339353832

cmelchior: Good point. This isn't about thread ID's but Realm versions. My bad

PR #5471, opened 2017-10-26
[skip ci][WIP]Simplify copy/insert in Proxy

beeender:
The logic for insert can be just reused by the copy. The only difference
is returning a RealmObject or not, which can be created with a row
index.


realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -1203,11 +1213,14 @@ public void insertOrUpdate(RealmModel object) {
             return new ArrayList<>(0);
         }
 
-        Map<RealmModel, RealmObjectProxy> cache = new HashMap<>();
+        Map<RealmModel, Long> cache = new HashMap<>();
         ArrayList<E> realmObjects = new ArrayList<>();
```
Zhuinden: This could be optimized for larger data sets by receiving a size, or doing

    LinkedList<T> realmObjects = new LinkedList<>();
    for(T t : objects) {
        realmObjects.add(t); // something something
    }
    return new ArrayList<>(realmObjects);

PR #5473, opened 2017-10-26
Add multi-process example

beeender:
All Realm public APIs should be process-safe now.

Encryption is not supported for multi processes due to core
restriction https://github.com/realm/realm-core/issues/1845

Accessing same Realm file in different processes from different apks
is safe, but the notifications won't work.


Zhuinden:
> Accessing same Realm file in different processes from different apks
is safe, but the notifications won't work.

So if another APK writes into Realm, the other APK won't be notified? :confused:

beeender:
> So if another APK writes into Realm, the other APK won't be notified? 😕

Yes. Mostly because of the SeLinux, we cannot create pipe on the external storage. But notifications rely on the pipe. It could be solved by adding a new public API to set up a realm temp directory to one of the apk's internal storage, but then the other apk won't work independently. Also, that requires both apk use the same uid by `sharedUserId` in manifest https://developer.android.com/guide/topics/manifest/manifest-element.html#uid

It will be quite fun to just use Realm to communicate between apks ...

Zhuinden:
Understandable, I guess this really is an issue only if a Realm is exposed through a Content Provider and both APKs write into it simultaneously. Even then, you could manually broadcast change events through Content Provider, at least notifications work in the same APK

examples/multiProcessExample/build.gradle
```diff
@@ -0,0 +1,38 @@
...
+    productFlavors {
+    }
+
+    command {
+        monkey.events 2000
+    }
+}
+
+dependencies {
+    compile 'com.android.support:appcompat-v7:23.1.1'
```
zaki50: We need to use the same major version of `appcompat` with `compileSdkVersion`.

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 ### Enhancements
 
+* Added support for using non-encrypted Realm in multiple processes from the same apk (#1091).
```
cmelchior: This is slightly pessimistic as using `sharedUserId` allows multiple APK's to access it, but I'm not entirely sure how to reword it to make it clear since there is also a bunch of other caveats. Perhaps something likes

```
Added support for using non-encrypted Realms in multiple processes. Some caveats apply. Read <insertLinkToDocs> for more info (#1091).
```

examples/settings.gradle
```diff
@@ -13,6 +13,7 @@ include 'unitTestExample'
 include 'newsreaderExample'
 include 'rxJavaExample'
 include 'objectServerExample'
+include 'multiProcessExample'
```
cmelchior: Multiprocess is one word, so this should be `multiprocessExample`

examples/multiProcessExample/src/main/java/io/realm/examples/realmmultiprocessexample/AnotherProcessService.java
```diff
@@ -0,0 +1,43 @@
+package io.realm.examples.realmmultiprocessexample;
```
cmelchior: Missing license in this file and a bunch of others

examples/multiProcessExample/src/main/java/io/realm/examples/realmmultiprocessexample/MainActivity.java
```diff
@@ -0,0 +1,90 @@
+/*
+ * Copyright 2014 Realm Inc.
```
cmelchior: 2017 in this file and a bunch of others

examples/multiProcessExample/src/main/java/io/realm/examples/realmmultiprocessexample/MainActivity.java
```diff
@@ -0,0 +1,90 @@
...
+        @Override
+        public void onChange(RealmResults<ProcessInfo> results) {
+            StringBuilder stringBuilder = new StringBuilder();
+
+            for (ProcessInfo processInfo : results) {
+                stringBuilder.append(processInfo.getName());
+                stringBuilder.append("\npid: ");
+                stringBuilder.append(processInfo.getPid());
+                stringBuilder.append("\nlast response time: ");
+                DateFormat df = new SimpleDateFormat("MM/dd/yyyy HH:mm:ss", Locale.ENGLISH);
```
cmelchior: Move this to a class variable?

examples/multiProcessExample/src/main/java/io/realm/examples/realmmultiprocessexample/Utils.java
```diff
@@ -0,0 +1,41 @@
...
+    public static String getMyProcessName(Context context) {
+        String processName = "";
+        ActivityManager am = (ActivityManager)context.getSystemService(Context.ACTIVITY_SERVICE);
+        List<RunningAppProcessInfo> infoList= am.getRunningAppProcesses();
+        if (infoList == null) {
+            throw new RuntimeException("getRunningAppProcesses() returns 'null'.");
+        }
+        for (RunningAppProcessInfo info : infoList) {
+            try {
+                if (info.pid == Process.myPid()) {
```
cmelchior: Can this throw?

examples/multiProcessExample/src/main/java/io/realm/examples/realmmultiprocessexample/MainActivity.java
beeender: UH, yeah! This code was written in 2014 actually ! will change it.

examples/multiProcessExample/src/main/java/io/realm/examples/realmmultiprocessexample/Utils.java
```diff
@@ -0,0 +1,41 @@
...
+import java.util.List;
+
+import io.realm.examples.realmmultiprocessexample.models.ProcessInfo;
+
+public class Utils {
+
+    public static String getMyProcessName(Context context) {
+        String processName = "";
+        ActivityManager am = (ActivityManager)context.getSystemService(Context.ACTIVITY_SERVICE);
+        List<RunningAppProcessInfo> infoList= am.getRunningAppProcesses();
```
nhachicha: `infoList=` missing space

examples/multiProcessExample/src/main/java/io/realm/examples/realmmultiprocessexample/Utils.java
```diff
@@ -0,0 +1,41 @@
...
+        if (infoList == null) {
+            throw new RuntimeException("getRunningAppProcesses() returns 'null'.");
+        }
+        for (RunningAppProcessInfo info : infoList) {
+            try {
+                if (info.pid == Process.myPid()) {
+                    processName = info.processName;
```
nhachicha: shouldn't you break from the for loop if you find your process name?

examples/multiProcessExample/src/main/java/io/realm/examples/realmmultiprocessexample/models/ProcessInfo.java
```diff
@@ -0,0 +1,54 @@
+/*
+ * Copyright 2014 Realm Inc.
```
nhachicha: update year

examples/multiProcessExample/src/main/java/io/realm/examples/realmmultiprocessexample/Utils.java
beeender: nope, should never throw. Unless it is called in a terminated process :P

Issue #5192, opened 2017-09-01
release step failed at monkeyDebug


cmelchior:
I got a COULD_NOT_FIND_MATCHING_ABI error when running `monkeyDebug` against our GridViewExample that uses APK Split. We need to figure out why

beeender:
dup to #5184

beeender:
reported to the upstream novoda/gradle-android-command-plugin#122

nhachicha:
fixed in https://github.com/realm/realm-java/pull/5478

PR #5478, opened 2017-10-26
Fix monkey, release no longer need a real device

nhachicha:
As reported by @beeender https://github.com/novoda/gradle-android-command-plugin/issues/122 the `android-command` plugin has an issue when running `monkey`. This PR remove this dependency, also we no longer require having a physical device to run monkey task (and perform a release)

fixes https://github.com/realm/realm-java/issues/5192


nhachicha:
@realm/java

examples/build.gradle
```diff
@@ -38,6 +46,29 @@ allprojects {
         jcenter()
         google()
     }
+
+    if (!project.name.startsWith("realm-examples")
+            && !project.name.startsWith("library")
+            && !project.name.startsWith("moduleExample")) { // exclude root and library project
+        task runMonkey(dependsOn: 'installDebug') {
```
beeender: we still need to run the `monkeyRelease` after uploading to bintrary. This doesn't seem to handle it?

beeender: I don't quite understand this, the task to run monkey for example is still `:monkeyDebug` and `:monkeyRelease`?

nhachicha: updated the script, now it has the same name as the previous tasks, also generating a release & debug tasks

PR #5513, opened 2017-11-06
Fix spelling mistake

### LINKED ISSUE #5512: 'and' instead of 'an'
Hi,

It seems to me there is a mistake in error message `Replacing and element is not supported.`

This message appear in my release at io.realm.internal.Collection$ListIterator.set(Collection.java:257)
```
public void set(T object) {
    throw new UnsupportedOperationException("Replacing and element is not supported.");
}
```

Best regards,
Pierre

cmelchior:
Fixes #5512


cmelchior:
Fast merge

PR #5520, opened 2017-11-09
Added the and() function in RealmQuery

rakshithravi1997:
Added the and() function to `RealmQuery<E>`. Basically does nothing, just returns the query object. This can be really helpful if we have a complex query and need to read the query in a more English-like language. Helps in understanding the query better, especially for newcomers, while it doesn't affect the performance either.

Purely added it for syntactic sugar.
Code can now be written as
```
getRealmObject()
    .where(User.class)
    .equalTo("username", "john-doe")
    .and()
    .equalTo("email" , "john@example.com")
    .findAll();
```

First time contributing to a big open source project.....so do let me know how I can improve my commits.


realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -1524,6 +1524,18 @@ public boolean isValid() {
         this.query.or();
         return this;
     }
+    
+    /**
+     * Logical-and two conditions
+     * Technically, does nothing. Intended purely for syntactic sugar
```
cmelchior: It would rewrite this to: "Realm automatically applies logical-and between all query statements, so this is intended only as a mean to increase readability."

rakshithravi1997:
Glad to hear that!

I just updated the CHANGELOG.md to account for this Enhancement. I also updated the Java Doc to word it better as you suggested.

I have personally never written a test case before. However, I was able to read the existing code, understand the norms and pick it up quickly. I wrote a basic test case for the `and()` function. I am still not so confident about my test code, considering this is my first time with tests. So could you please look into it and let me know if it would be sufficient?

PR #5523, opened 2017-11-10
Fix leaking file handlers.

### LINKED ISSUE #5521: File handles are not released in RealmTransformer
realm-gradle-plugin:4.1.1
Original issue: https://issuetracker.google.com/69114369
Android plugin for Gradle: 3.0.0

JarFile opened in RealmTransform is not closed, which causes issues on Windows. The issue seems to be this line: https://github.com/realm/realm-java/blob/c1302392424a3b8bed323d3164ed5f875551e174/realm-transformer/src/main/groovy/io/realm/transformer/RealmTransformer.groovy#L245

Stacktrace showing where the file got opened (and not closed) ->
[stacktrace.txt](https://github.com/realm/realm-java/files/1460873/stacktrace.txt)

There is another file handle to the same file, captured in this stacktrace ->
[stacktrace_classpool.txt](https://github.com/realm/realm-java/files/1460891/stacktrace_classpool.txt)

This one seems to be caused by not closing the javassist.JarClassPath opened on this line https://github.com/realm/realm-java/blob/c1302392424a3b8bed323d3164ed5f875551e174/realm-transformer/src/main/groovy/io/realm/transformer/RealmTransformer.groovy#L219

Thanks!

cmelchior:
Fixes #5521 

I don't have a Windows machine to verify this on, but by doing code review it does appear that file handlers are being leaked. Unfortunately one of the leaks are inside JavaAssist, so I reported that upstream while implementing a workaround in our end for the time being.

### DIFF
--- a/realm-transformer/src/main/groovy/io/realm/transformer/ManagedClassPool.groovy
+++ b/realm-transformer/src/main/groovy/io/realm/transformer/ManagedClassPool.groovy
@@ -1,0 +1,83 @@
+/*
+ * Copyright 2017 Realm Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.realm.transformer
+
+import com.android.build.api.transform.TransformInput
+import javassist.ClassPath
+import javassist.ClassPool
+
+/**
+ * This class implements the Transform API provided by the Android Gradle plugin.
+ */
+@SuppressWarnings("GroovyUnusedDeclaration")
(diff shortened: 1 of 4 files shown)


realm-transformer/src/main/groovy/io/realm/transformer/ManagedClassPool.groovy
```diff
@@ -0,0 +1,83 @@
...
+ */
+
+package io.realm.transformer
+
+import com.android.build.api.transform.TransformInput
+import javassist.ClassPath
+import javassist.ClassPool
+
+/**
+ * This class implements the Transform API provided by the Android Gradle plugin.
```
Zhuinden: It does?

cmelchior: Well, part of it anyway :)

realm-transformer/src/main/groovy/io/realm/transformer/ManagedClassPool.groovy
```diff
@@ -0,0 +1,81 @@
...
+ */
+@SuppressWarnings("GroovyUnusedDeclaration")
+class ManagedClassPool extends ClassPool implements Closeable {
+
+    def List<ClassPath> pathElements = new ArrayList<ClassPath>()
+
+    /**
+     * Constructor for creating and populating the JavAssist class pool.
+     * Remember to class {@link #close()} when done with it to avoid leaking file resources
```
nhachicha: Remember to `call

PR #5532, opened 2017-11-13
CI fails without exporting the logcat.zip

nhachicha:
The _logcat_ is collected on CI but only archived if the tests fail.

sometimes the CI fails with `Expected X tests, received Y` ([example](https://ci.realm.io/job/realm/job/realm-java/job/PR-5511/21/consoleFull)) which doesn't provide the _logcat_ to investigate what happened. This PR allow archive  the _logcat_ regardless of the test


nhachicha:
@realm/java

Issue #5507, opened 2017-11-03
RealmChangeListener not called for RealmList of RealmObject


jpmcosta:
#### Goal

After adding a `RealmChangeListener` to a `RealmList`, if the number of its elements changes, the listener should be called.

#### Expected Results

`RealmChangeListener` should always be called when the number of elements of a `RealmList` changes.

#### Actual Results

`RealmChangeListener` is not always called when the number of elements of a `RealmList` changes.

#### Steps & Code to Reproduce

I created a simplified Android project to reproduce the issue's workflow:
https://github.com/jpmcosta/RealmTestProject/tree/9375194c255977ae944fde292b7ca5b9ba845f5a

The workflow of the issue:
- `MainActivity` opens a `CreateActivity` that starts `CreateIntentService` to add or remove items from a `RealmList` (then, `CreateActivity` is finished)
- when the `MainActivity` starts again, it adds a `RealmChangeListener` to the `RealmList` and logs its current number of items
- the listener's `onChange()` method is not called, despite the number of items being changed (this is the issue)

To reproduce:
- run the app
- click on the floating action button to open `CreateActivity` and change the number of elements of the list
  - if the list is empty, it should fill the list with 10 items
  - if the list has items, it should clear all items
- logs are created when (log_tag: `ItemListAdapter`)
  - the `RealmChangeListener` is added: `add listener <current number of items>`
  - the `RealmChangeListener` is removed: `remove listener`
  - the `RealmChangeListener.onChange()`: `onChange() <current number of items>`
- to understand the issue:
  - notice the number of items when the listener is added, and compare it to the list's actual number of items
  - notice whether or not `onChange()` was called (usually, it's not called).
- you can also long press the floating action button to add a single item, to make sure the listener is working (it should work and a `onChange()` log should be created)

I'm reproducing the issue consistently, but not 100% of the time (maybe 99%).
If you can't reproduce the issue, try to change the `Don't keep activities` setting in Android's Developer options.

#### Version of Realm and tooling
Realm version: 4.1.0

Android Studio version: 3.0

Which Android version and device: 7.1.2

Zhuinden:
``` java
    @Override
    protected void onStart() {
        super.onStart();

        App app = getRealm().where(App.class).findFirst();
        ItemListAdapter mAdapter = new ItemListAdapter(app.items);
        mRecyclerView.setAdapter(mAdapter);
    }

    @Override
    protected void onStop() {
        super.onStop();

[...]

PR #5537, opened 2017-11-15
Attempt to reproduce missing collection notifications

cmelchior:
Unit test trying to reproduce https://github.com/realm/realm-java/issues/5507. Still not succeeding.

Best guess is that it is related to the objects being GC'ed between MainActivity and the CreateActivity as it seems to work if you just use background threads on the MainActivity. We need to modify this unit test so it also GC results.

https://github.com/jpmcosta/RealmTestProject/tree/9375194c255977ae944fde292b7ca5b9ba845f5a can reproduce the behavior consistently.


beeender:
See test case in #5507 . will be fixed soon. Close this.

PR #5538, opened 2017-11-15
Feature: Add Support for Partial Realms

cmelchior:
This PR is the feature branch for partial Realm support.

Partial Realms are a new Realm type that only download the data you actually listen to. This allows for the server side Realms to be much larger than the on-device Realms, but each device only downloads the data they are interested. "Insterested in" is defined as any data that is queried for.

Limitations in first version:
* No Object Level permissions
* No Progress Listeners on invidiual Queries
* No aggregate Queries
* No automatic removal of data again
* No paging/limit/offset
* No advanced options for controlling subscription lifecycle. In first iteration they are kept around unless manually removed.

### TODO
- [ ] Automatically subscribe queries
-- [ ] Query Mapper in Object Store (from Core Query Object to String and back) 
- [ ] Add subscription meta data to object store
- [ ] Add new ChangeListener state for detecting when remote queries are done
- [ ] Add Subscription class and public API's
- [ ] Add wrapper for progress listeners to the Subscription

### Proposed API (not final)

```
// 1. Configure Realm as partial
       .partialRealm()
       .build();


// 2. Start subscription automatically
RealmResults results = realm.where(Product.class).findAllASync();

results.addChangeListener(new RealmChangeListener() {
  @Override
   public void onChange(RealmResults<Product> results) {
       if (results.isLoaded()) {
           // Local query completed
       }

       if (results.isRemoteLoaded()) {
           // Remote query completed and data integrated onto the device
	
       }
   }
});

// 3. It is possible to work with the underlying Subscription directly
public class Subscription {
   void setId(String id); // Rename subscription
   String getId();
   Date getLastUsed(); // Last timestamp when a query result was either created or deallocated.
   boolean isInUse(); // True if a local query is using this currently
   boolean unsubscribe(); // Unsubscribe, might not delete the data right away
   void registerProgressListener(ProgressMode mode, ProgressListener listener);
   void removeProgressListener(listener);
}

// 4A: Sync only methods on standard Realm API's :( Kotlin Extension functions to the rescue?
Subscription sub = results.getSubscription();
realm.subscribe("main-products", realm.where(Product.class));
boolean hasSubscription = realm.hasSubscription("main-products");
Subscripton sub = realm.getSubscription("main-products");

// 4B: On Session. A little weird as you still need the Realm reference :(
// Still unclear how many methods we want for managing subscriptions. Having more than 2 or 3
// probably means we need to encapsulate them somehow.
SyncSession session = SyncManager.getSession(config);
session.subscribe("main-products", realm.where(Product.class));
Subscription sub = session.hasSubscription(String id); // By id
Subscription sub = session.getSubscription(RealmResults<?> results); // By Query result
Subscription sub = session.getSubscription(RealmModel object); // By findFirst query
List<Subscription> subscriptions = session.getSubscriptions();

// Suggestion: Keep methods on Session in Java, but use Kotlin extension functions to move them
// to Realm/RealmResults in Kotlin?
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### Enhancements
 
-* Added support for partial Realms. Read [here](https://realm.io/docs/java/latest/#partal-realms) for mere information.
+* Added support for partial Realms. Read [here](https://realm.io/docs/java/latest/#partial-realms) for more information.
 
 
 ## 4.2.0 (YYYY-MM-DD)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## 4.3.0 (YYYY-MM-DD)
+
+### Enhancements
+
+* Added support for partial Realms. Read [here](https://realm.io/docs/java/latest/#partal-realms) for mere information.
+
+
 ## 4.2.0 (YYYY-MM-DD)
 
 ### Breaking Changes


CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+## 4.3.0 (YYYY-MM-DD)
+
+### Enhancements
+
+* Added support for partial Realms. Read [here](https://realm.io/docs/java/latest/#partal-realms) for mere information.
```
Zhuinden: `partal` and `mere`? :+1:

cmelchior:
Replaced by other seperate PR's to `next-major`

PR #5552, opened 2017-11-21
Added notification to OsList instead of OsResults

beeender:
- The notification token for RealmList<RealmModel> will be added to the
  underlying OsList instead of the OsResults created from the list. This
  will have a different behavior for a corner case,
  see https://github.com/realm/realm-object-store/pull/602 . Thus the
  test cases needs to be modified accordingly.
- This would be the precondition of using
  https://github.com/realm/realm-object-store/pull/601
- Add toString() for OrderedCollectionChangeSet.

### DIFF
--- a/realm/realm-library/src/androidTest/java/io/realm/OrderedCollectionChangeSetTests.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/OrderedCollectionChangeSetTests.java
@@ -131,6 +131,7 @@ public class OrderedCollectionChangeSetTests {
                 assertNotNull(owner);
                 RealmList<Dog> dogs = owner.getDogs();
                 boolean added = false;
+                // Insert the newly created dog to the RealmList by the order of age.
                 for (int i = 0; i < dogs.size(); i++) {
                     if (dog.getAge() <= dogs.get(i).getAge()) {
                         dogs.add(i, dog);
@@ -143,11 +144,6 @@ public class OrderedCollectionChangeSetTests {
                 }
             }
         }
-        /*
-        if (type == ObservablesType.REALM_LIST) {
-            reorderRealmList(realm);
-        }
-        */
     }
 
     // Modifies Dogs objects which's columnLong is in the indices array.
@@ -161,6 +157,10 @@ public class OrderedCollectionChangeSetTests {
 
     private void moveObjects(Realm realm, int originAge, int newAge) {
         if (type == ObservablesType.REALM_LIST) {
+            // For RealmList we need to:
+            // 1. Find the object by the original age and move it to the new place where it should be with the new age
+            //    set -- the RealmList is sorted by age.
+            // 2. Set the object's age with new value.
             RealmList<Dog> dogs = realm.where(Owner.class).findFirst().getDogs();
             int originIdx = -1;
             int newIdx = -1;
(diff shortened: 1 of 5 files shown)


realm/realm-library/src/androidTest/java/io/realm/OrderedCollectionChangeSetTests.java
```diff
@@ -130,11 +124,30 @@ private void deleteObjects(Realm realm, int... indices) {
...
-            realm.createObject(Dog.class).setAge(index);
+            Dog dog = realm.createObject(Dog.class);
+            dog.setAge(index);
+            if (type == ObservablesType.REALM_LIST) {
+                Owner owner = realm.where(Owner.class).findFirst();
+                assertNotNull(owner);
+                RealmList<Dog> dogs = owner.getDogs();
+                boolean added = false;
+                for (int i = 0; i < dogs.size(); i++) {
+                    if (dog.getAge() <= dogs.get(i).getAge()) {
```
cmelchior: This could use a little description I think. There looks like there is some assumptions about the `index` and `dogs.size()` that I don't fully understand.

realm/realm-library/src/androidTest/java/io/realm/OrderedCollectionChangeSetTests.java
```diff
@@ -130,11 +124,30 @@ private void deleteObjects(Realm realm, int... indices) {
...
+                        dogs.add(i, dog);
+                        added = true;
+                        break;
+                    }
+                }
+                if (!added) {
+                    dogs.add(dog);
```
cmelchior: Not sure I understand this one either?

realm/realm-library/src/androidTest/java/io/realm/OrderedCollectionChangeSetTests.java
```diff
@@ -147,9 +160,38 @@ private void modifyObjects(Realm realm, int... indices) {
     }
 
     private void moveObjects(Realm realm, int originAge, int newAge) {
-        realm.where(Dog.class).equalTo(Dog.FIELD_AGE, originAge).findFirst().setAge(newAge);
         if (type == ObservablesType.REALM_LIST) {
-            reorderRealmList(realm);
+            RealmList<Dog> dogs = realm.where(Owner.class).findFirst().getDogs();
```
cmelchior: Please add a comment about what this method does

realm/realm-library/src/main/java/io/realm/RealmList.java
```diff
@@ -1281,20 +1259,6 @@ private static boolean isClassForRealmModel(Class<?> clazz) {
...
         throw new IllegalArgumentException("Unexpected value class: " + clazz.getName());
     }
-
-    // TODO: Object Store is not able to merge change set for links list. Luckily since we were still using LinkView
-    // when ship the fine grain notifications, the listener on RealmList is actually added to a OS Results which is
-    // created from the link view. OS Results is computing the change set by comparing the old/new collection. So it
-    // will give the right results if you remove all elements from a RealmList then add all them back and add one more
-    // new element. By right results it means the change set only include one insertion. But if the listener is on the
-    // OS List, the change set will include all ranges of th list. So we keep the old behaviour for
-    // RealmList<RealmModel> for now. See https://github.com/realm/realm-object-store/issues/541
```
cmelchior: I tend to agree that this sounds like a mistake, but if we introduce the behaviour (until fixed in OS), then it doesn't seem like a major issue either.

realm/realm-library/src/androidTest/java/io/realm/OrderedCollectionChangeSetTests.java
beeender: added some comments. The assumption is the both `RealmResults` and `RealmList` are sorted by age. See `populateData()`.

realm/realm-library/src/androidTest/java/io/realm/OrderedCollectionChangeSetTests.java
beeender: ditto

realm/realm-library/src/androidTest/java/io/realm/OrderedCollectionChangeSetTests.java
beeender: simplified the code and added some comments.

PR #5553, opened 2017-11-21
Em/kotlin extensions help

ericmaxwell2003:
Tests and KDoc for the Kotlin extensions being added as part of https://github.com/realm/realm-java/pull/4684.  

This PR is setup to merge into @cmelchior's branch which is the main feature branch for (4684).

The KDocs are the prettiest, but not sure how to tidy them up anymore.  Unit tests all pass locally via `../gradlew/connectedBaseDebugAndroidTest`.


examples/kotlinExample/src/main/kotlin/io/realm/examples/kotlin/KotlinExampleActivity.kt
```diff
@@ -113,7 +114,10 @@ class KotlinExampleActivity : Activity() {
         showStatus("\nPerforming basic Query operation...")
         showStatus("Number of persons: ${realm.where<Person>().count()}")
 
-        val results = realm.where<Person>().equalTo(Person::age, 99).findAll()
+        val ageCriteria = 99
+        val results = realm.where<Person>()
+                .equalTo(Person::age.name, ageCriteria)
```
cmelchior: Do we want to do this in our example when this approach doesn't work for linked fields either? Or should we just move back to "age"?

realm/kotlin-extensions/src/androidTest/kotlin/io/realm/KotlinRealmModelTests.kt
```diff
@@ -0,0 +1,233 @@
...
+        // sure we are sure calling deleteFromRealm
+        realm.executeTransaction { realm.where<SimpleClass>().findFirst()!!.deleteFromRealm() }
+        Assert.assertEquals(0, realm.where<SimpleClass>().count())
+
+    }
+
+    @Test
+    fun isValid() {
+        realm.executeTransaction {
+            val simple = it.createObject<SimpleClass>()
```
cmelchior: use `obj` instead would make the test more readable. Same for the rest of the tests.

realm/kotlin-extensions/src/androidTest/kotlin/io/realm/KotlinRealmModelTests.kt
```diff
@@ -0,0 +1,233 @@
...
+            assertFalse("Expected not managed until attached", simple.isManaged())
+
+            simple = it.copyToRealm(simple)
+            assertTrue("Expected managed after attaching", simple.isManaged())
+        }
+    }
+
+    @Test
+    @RunTestInLooperThread
+    fun add_realmObjectChangeListener() {
```
cmelchior: Generally, we name tests after the following pattern: `<methodUnderTest>_<conditions>`, so this would be `addObjectChangeListener_addObject` or something like that.

realm/kotlin-extensions/src/androidTest/kotlin/io/realm/KotlinRealmModelTests.kt
```diff
@@ -0,0 +1,233 @@
...
+        })
+
+        realm.beginTransaction()
+        simple.name = "simple1"
+        realm.commitTransaction()
+    }
+
+    @Test
+    @RunTestInLooperThread
+    fun add_realmChangeListener() {
```
cmelchior: This would e.g. be `addChangeListener_addObject`

realm/kotlin-extensions/src/androidTest/kotlin/io/realm/KotlinRealmModelTests.kt
```diff
@@ -0,0 +1,233 @@
...
+        })
+
+        realm.beginTransaction()
+        simple.name = "simple1"
+        realm.commitTransaction()
+    }
+
+    @Test
+    @RunTestInLooperThread
+    fun remove_realmChangeListener() {
```
cmelchior: This would just be `removeChangeListener`. Similar renames needed for the rest of these methods

realm/kotlin-extensions/src/androidTest/kotlin/io/realm/KotlinRealmTests.kt
```diff
@@ -61,4 +66,88 @@ class KotlinRealmTests {
     fun where() {
         assertEquals(0, realm.where<SimpleClass>().count())
     }
-}
+
+    @Test
+    fun where_getsResult() {
```
cmelchior: Not entirely sure what this tests? Is it `where` ? In which case `assertEquals(0, realm.where<PrimaryKeyClass>().count())` would be enough?

If it is more, then all of that is tested on the Java side and doesn't need to be tested from Kotlin I think?

realm/kotlin-extensions/src/androidTest/kotlin/io/realm/KotlinRealmTests.kt
```diff
@@ -61,4 +66,88 @@ class KotlinRealmTests {
...
+                .equalTo(PrimaryKeyClass::id.name, pk)
+                .findFirst()
+
+        assertNotNull(pkClassObjOut)
+        assertEquals(pk, pkClassObjOut?.id)
+        assertEquals(name, pkClassObjOut?.name)
+    }
+
+    @Test
+    fun callInTransaction_withPairOfRealmModelReturned() {
```
cmelchior: I am not sure I understand the value of the `callInTransaction` extension function?

Is it to return the object being created? Is that use case really important enough to warrant an extension function in our official API?

Also, I'm not sure about the name either? Is it a naming convention from Kotlin?

realm/kotlin-extensions/src/androidTest/kotlin/io/realm/entities/SimpleClass.kt
```diff
@@ -1,9 +1,10 @@
...
+import io.realm.RealmModel
 import io.realm.RealmObject
 import io.realm.annotations.RealmClass
 
 @RealmClass
-open class SimpleClass : RealmObject() {
-    var name: String = "";
+open class SimpleClass : RealmModel {
+    var name: String = ""
 }
```
cmelchior: Missing newline at end of file

realm/kotlin-extensions/src/main/kotlin/io/realm/kotlin/RealmExtensions.kt
```diff
@@ -26,32 +21,58 @@ inline fun <reified T : RealmModel> Realm.where(): RealmQuery<T> {
...
+ * // Andy Kotlin has a 2 year old dog named Tony
+ * println("${person.name} has a ${dog.age} year old dog named ${dog.name}")
+ * ```
+ *
+ * @param action A function that operates on a receiver type Realm that is in a transaction.
+ *               The return value of `action` is returned by this function.
+ * @return T the object returned from the `action`.
+ *
+ */
+inline fun <T> Realm.callInTransaction(crossinline action: Realm.() -> T): T {
```
cmelchior: Is this naming idiomatic in Kotlin? To me, it doesn't convey what is actually happening, but it might just be my inexperience.

realm/kotlin-extensions/src/main/kotlin/io/realm/kotlin/RealmExtensions.kt
```diff
@@ -26,32 +21,58 @@ inline fun <reified T : RealmModel> Realm.where(): RealmQuery<T> {
...
+    executeTransaction {
+        ref.set(action(it))
+    }
+    return ref.get()
+}
 
 /**
- * Instantiates and adds a new object to the Realm.
- * <p>
+ *
```
cmelchior: Remove whitespace

realm/kotlin-extensions/src/main/kotlin/io/realm/kotlin/RealmModelExtensions.kt
```diff
@@ -3,46 +3,202 @@
...
-import kotlin.reflect.KMutableProperty1
 
+/**
+ * Deletes the object from the Realm it is currently associated with.
+ *
+ * After this method is called the object will be invalid and any operation (read or write) performed on it will
+ * fail with an `IllegalStateException`.
+ *
+ * @throws IllegalStateException if the corresponding Realm is closed or in an incorrect thread.
+ * @see #isValid()
```
cmelchior: I think you need to do `[isValid]` in KDoc

realm/kotlin-extensions/src/main/kotlin/io/realm/kotlin/RealmModelExtensions.kt
```diff
@@ -3,46 +3,202 @@
...
+ * notified when changes happen. Managed objects are thread confined so that they cannot be accessed from other threads
+ * than the one that created them.
+ *
+ * If this method returns `false`, the object is unmanaged. An unmanaged object is just a normal Java object,
+ * so it can be parsed freely across threads, but the data in the object is not connected to the underlying Realm,
+ * so it will not be live updated.
+ *
+ * It is possible to create a managed object from an unmanaged object by using
+ * [Realm.copyToRealm(RealmModel)]. An unmanaged object can be created from a managed object by using
+ * [Realm#copyFromRealm(RealmModel)].
```
cmelchior: `# -> .`

examples/kotlinExample/src/main/kotlin/io/realm/examples/kotlin/KotlinExampleActivity.kt
ericmaxwell2003: hmm.. I liked the idea of illustrating that Kotlin gives you the ability to do this, making a safe reference to the property.  I can change it to use a string though if you feel strongly.

realm/kotlin-extensions/src/androidTest/kotlin/io/realm/KotlinRealmModelTests.kt
ericmaxwell2003: As in `val obj = it.createObject<SimpleClass>()`?  Sure, will do.

examples/kotlinExample/src/main/kotlin/io/realm/examples/kotlin/KotlinExampleActivity.kt
cmelchior: I agree it is nice that it works, but since these examples are in many cases copy-pasted, having code in here that doesn't work in all cases is probably a bad idea. We could perhaps just add a comment about it, and just quickly describe advantages and pitfalls of using it, but keep using Strings in actual code?

realm/kotlin-extensions/src/androidTest/kotlin/io/realm/KotlinRealmModelTests.kt
cmelchior: Yes 👍

realm/kotlin-extensions/src/androidTest/kotlin/io/realm/KotlinRealmTests.kt
ericmaxwell2003: My thought is that it's testing that the extension where() method can (setup a RealmQuery) to get data from Realm.  It just seemed like a deeper/more thorough test that the simple where() test method. 

Perhaps too much?  I can remove.   Let me know

cmelchior: `where()` returns a `RealmQuery` which is already pretty well tested from Java and the Kotlin API doesn't change anything once you have the `RealmQuery`, so in the context of Kotlin it is probably a bit too much. I would probably remove it, also to keep the cognitive load down.

examples/kotlinExample/src/main/kotlin/io/realm/examples/kotlin/KotlinExampleActivity.kt
beeender: So this would work even with proguard enabled? That is nice!

realm/kotlin-extensions/src/androidTest/kotlin/io/realm/KotlinRealmTests.kt
ericmaxwell2003: 👍 will do

realm/kotlin-extensions/src/main/kotlin/io/realm/kotlin/RealmExtensions.kt
ericmaxwell2003: Agree..  Open to ideas.  I guess it's similar to `[let](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/let.html)` functionality but transactional and specific to Realm.    Or perhaps, we should just drop it altogether if you feel it's not applicable to all??

It's not a core functionality, more of a convenience function that might not be used by all.  It's a great way to get objects out of a Tx block and makes for a clean API feel IMO.  However, the use case is that you're probably using this in a BG thread (since it's blocking) and presumably you can just do all work inside the transaction anyway (and not need to pass objects back).

examples/kotlinExample/src/main/kotlin/io/realm/examples/kotlin/KotlinExampleActivity.kt
ericmaxwell2003: Yeah, I will change it to strings with a comment about what's possible in kotlin for simple field access.

I also think we may want to point people to `realmfieldnameshelper`.  Maybe I can add that in the comment too?  Any objection to that?

realm/kotlin-extensions/src/main/kotlin/io/realm/kotlin/RealmExtensions.kt
cmelchior: I'm not inherently against it, we should just be careful about overloading our extension function library with all sorts of ideas. Mostly because we are going to need to support them going forward and we need to ensure that they are widely applicable.

At this point in time, I don't really have a feeling for how useful this is. I mean it isn't something I missed in the Java API, but others might feel differently?

Then of course there is an entire naming discussion as well. Good with the `let` reference which also is a horrible name IMO :)

examples/kotlinExample/src/main/kotlin/io/realm/examples/kotlin/KotlinExampleActivity.kt
cmelchior: No, thats fine.

realm/kotlin-extensions/src/main/kotlin/io/realm/kotlin/RealmExtensions.kt
ericmaxwell2003: In light of this discuss, I say let's keep it out for now. Release this first extensions support and revisit later.

cmelchior: Sounds good to me

realm/kotlin-extensions/src/main/kotlin/io/realm/kotlin/RealmModelExtensions.kt
```diff
@@ -2,47 +2,198 @@
...
 }
 
+/**
+ * Checks if this object is managed by Realm. A managed object is just a wrapper around the data in the underlying
+ * Realm file. On Looper threads, a managed object will be live-updated so it always points to the latest data. It
+ * is possible to register a change listener using [addChangeListener] to be
+ * notified when changes happen. Managed objects are thread confined so that they cannot be accessed from other threads
+ * than the one that created them.
+ *
+ * If this method returns `false`, the object is unmanaged. An unmanaged object is just a normal Java object,
```
nhachicha: maybe `a normal Kotlin` instead?

realm/kotlin-extensions/src/main/kotlin/io/realm/kotlin/RealmModelExtensions.kt
```diff
@@ -2,47 +2,198 @@
...
+ * so it can be parsed freely across threads, but the data in the object is not connected to the underlying Realm,
```
nhachicha: it can be passed

realm/kotlin-extensions/src/main/kotlin/io/realm/kotlin/RealmModelExtensions.kt
```diff
@@ -2,47 +2,198 @@
...
+ * Checks if the query used to find this RealmObject has completed.
+ *
+ * Async methods like [io.realm.RealmQuery.findFirstAsync] return an RealmObject that represents the future result
+ * of the RealmQuery. It can be considered similar to a [java.util.concurrent.Future] in this regard.
+ *
+ * Once `isLoaded()` returns `true`, the object represents the query result even if the query
+ * didn't find any object matching the query parameters. In this case the RealmObject will
+ * become a `null` object.
+ *
+ * "Null" objects represents `null`.  An exception is throw if any accessor is called, so it is important to also
```
nhachicha: is thrown

realm/kotlin-extensions/src/main/kotlin/io/realm/kotlin/RealmModelExtensions.kt
```diff
@@ -2,47 +2,198 @@
...
+
+/**
+ * Adds a change listener to a RealmObject that will be triggered if any value field or referenced RealmObject field
+ * is changed, or the RealmList field itself is changed.
+
+ * Registering a change listener will not prevent the underlying RealmObject from being garbage collected.
+ * If the RealmObject is garbage collected, the change listener will stop being triggered. To avoid this, keep a
+ * strong reference for as long as appropriate e.g. in a class variable.
+ *
+ * ``` kotlin
```
nhachicha: space after kotlin

realm/kotlin-extensions/src/main/kotlin/io/realm/kotlin/RealmModelExtensions.kt
```diff
@@ -2,47 +2,198 @@
...
+ *
+ * Once `isLoaded()` returns `true`, the object represents the query result even if the query
+ * didn't find any object matching the query parameters. In this case the RealmObject will
+ * become a `null` object.
+ *
+ * "Null" objects represents `null`.  An exception is throw if any accessor is called, so it is important to also
+ * check isValid before calling any methods. A common pattern is:
+ *
+ *
+ * ```
```
nhachicha: "```kotlin"

realm/kotlin-extensions/src/main/kotlin/io/realm/kotlin/RealmModelExtensions.kt
```diff
@@ -2,47 +2,198 @@
...
+/**
+ * Adds a change listener to a RealmObject to get detailed information about the changes. The listener will be
+ * triggered if any value field or referenced RealmObject field is changed, or the RealmList field itself is
+ * changed.
+
+ * Registering a change listener will not prevent the underlying RealmObject from being garbage collected.
+ * If the RealmObject is garbage collected, the change listener will stop being triggered. To avoid this, keep a
+ * strong reference for as long as appropriate e.g. in a class variable.
+ *
+ * ``` kotlin
```
nhachicha: space before kotlin

realm/kotlin-extensions/src/androidTest/kotlin/io/realm/KotlinRealmModelTests.kt
```diff
@@ -0,0 +1,233 @@
...
+@RunWith(AndroidJUnit4::class)
+class KotlinRealmModelTests {
+
+    @Suppress("MemberVisibilityCanPrivate")
+    @get:Rule
+    val configFactory = TestRealmConfigurationFactory()
+
+    @get:Rule
+    val looperThread = RunInLooperThread()
+
```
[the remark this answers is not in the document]
ericmaxwell2003: Where at?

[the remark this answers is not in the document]
ericmaxwell2003: oh, nvm, I see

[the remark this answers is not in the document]
ericmaxwell2003: This is done, not sure why this comment doesn't show as outdated now.

Zhuinden:
How will you be able to add this as dependency to the project?

cmelchior:
The gradle plugin does it automatically

PR #5556, opened 2017-11-22
Evaluate queries immediately for sync queries

### LINKED ISSUE #5328: Performance regression in query (nativeSize and nativeGetRow)
#### Goal

Upgrade Realm to latest and retain performance

#### Expected Results

No performance drop

#### Actual Results

My query which looks like this became slow when updating from 1.2.0 -> 4.0.0-BETA3-SNAPSHOT.

``` java
        Date startDate = DateUtils.setToMidnight(date);
        Date endDate = DateUtils.setToEndOfDay(startDate);
        RealmResults<Schedule> schedules = query(realm)
                .beginGroup()
                .beginGroup()
                .lessThan(ScheduleFields.START_TIME, endDate)
                .greaterThanOrEqualTo(ScheduleFields.START_TIME, startDate)
                .endGroup()
                .or()
                .beginGroup()
                .greaterThanOrEqualTo(ScheduleFields.END_TIME, startDate)
                .lessThan(ScheduleFields.START_TIME, endDate)
                .endGroup()
                .endGroup()
                .lessThanOrEqualTo(ScheduleFields.START_TIME, date)
                .equalTo(ScheduleFields.CHANNEL_ID, channel.getId())
                .findAllSorted(ScheduleFields.START_TIME);
        if(!schedules.isEmpty()) {
            return schedules.get(schedules.size() - 1);
        } else {
            return null;
        }
```

With a schema that looks like this

``` java
public class Schedule
        extends RealmObject {
    @PrimaryKey
    private long id;

    private String title;

    @Index
    private String titleLowerCase;

    private long categoryId;

    private String category;

    @Index
    private Date startTime;

    @Index
    private Date endTime;

    private int duration;

    private String ageLimit;

    private long programId;

    private String episodeId;

    @Index
    private long channelId;

    private Channel channel;

    private RealmList<TVGroup> tvGroups;
```

The query was then reduced to 

``` java
        Schedule schedule = query(realm)
                .lessThanOrEqualTo(ScheduleFields.START_TIME, date)
                .equalTo(ScheduleFields.CHANNEL_ID, channel.getId())
                .findAllSorted(ScheduleFields.START_TIME, Sort.DESCENDING)
                .first(null); // <-- added this
        if(schedule == null) {
            return null;
        }
        long scheduleStartTime = schedule.getStartTime().getTime();
        long scheduleEndTime = schedule.getEndTime().getTime();

        if((scheduleStartTime < endDate.getTime() && scheduleStartTime >= startDate.getTime())
            || (scheduleEndTime >= startDate.getTime() && scheduleStartTime < endDate.getTime()) ){
            return schedule;
        }
        return null;
```

But it is still slower in `first()` than in Realm 2.3.0 (and then slow in Realm 3.0.0)

Realm version(s): 1.2.0 -> 4.0.0-BETA3-SNAPSHOT

Android Studio version: 2.3.3

Which Android version and device: LG Nexus 5X, 8.0.0

### LINKED ISSUE #5387: Iterating result very slow
Hi,

Here are my java classes : 

```java
public class PublicArticleEntity extends RealmObject {

    /**
     * For fast import no @PrimaryKey defined
     */
    private String id;

    private String type;

    @Index
    private String manufacturerId;

    private Boolean sterilizable;

    private Boolean active;

    private Date importModified;

    private RealmList<PublicArticlePartNumberEntity> partNumbers = new RealmList<>();
    private RealmList<PublicArticleCodificationEntity> codifications = new RealmList<>();

public class PublicArticlePartNumberEntity extends RealmObject {

    @Index
    private String partNumber;

    @Index
    private String normalizedPartNumber;

    private PublicArticleEntity publicArticle;
}

public class PublicArticleCodificationEntity extends RealmObject {

    @Index
    private String type;

    @Index
    private String value;

    private PublicArticleEntity publicArticle;
}
```

And here is the piece of code where I have a problem. The query returns 24 results and I use the `limit` to get the 21 first elements. It tooks 2000ms to do the iteration :o

```java
RealmResults<PublicArticlePartNumberEntity> entities = realm.where(PublicArticlePartNumberEntity.class)
                .equalTo("publicArticle.manufacturerId", manufacturerId.toString())
                .beginGroup()
                .contains("partNumber", search, Case.INSENSITIVE)
                .or()
                .contains("normalizedPartNumber", search, Case.INSENSITIVE)
                .endGroup()
                .findAllSorted("partNumber", Sort.ASCENDING);

        Set<String> items = new HashSet<>();
        int listSize = entities.size();
        for (int i = 0; i < listSize && i < limit; i++) {
            items.add(entities.get(i).getPartNumber());
        }
```

Here the state of each :
screenshot at oct 06 15-04-13

Because Realm is using lazy loading the request if very very fast, but the `for` is taking 2s.
Did I miss something ?

beeender:
- OsResults.load() actually does evaluate queries if needed now by using
  newly exposed OS method Results.evaluate_query_if_needed(). So the
  Results's mode will be changed to TABLEVIEW from QUERY. This matches
  the old async query behaviour and solved the performance issues when
  the query results are huge, the size() method took unnecessary long
  time even the Results accessor will be called at the next line.
  close #5328 close #5387
- Update Object Store to 3eb19c014fdf .
- Refactor the APIs for creating OsResults.

### DIFF
--- a/realm/realm-library/src/main/java/io/realm/internal/OsResults.java
+++ b/realm/realm-library/src/main/java/io/realm/internal/OsResults.java
@@ -85,8 +85,6 @@ public class OsResults implements NativeObject, ObservableCollection {
 
         /**
          * Not supported by Realm collection iterators.
-         *
-         * @throws UnsupportedOperationException
          */
         @Override
         @Deprecated
@@ -138,8 +136,6 @@ public class OsResults implements NativeObject, ObservableCollection {
 
         /**
          * Unsupported by Realm collection iterators.
-         *
-         * @throws UnsupportedOperationException
          */
         @Override
         @Deprecated
@@ -193,8 +189,6 @@ public class OsResults implements NativeObject, ObservableCollection {
 
         /**
          * Unsupported by RealmResults iterators.
-         *
-         * @throws UnsupportedOperationException
          */
         @Override
         @Deprecated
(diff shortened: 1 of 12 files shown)


realm/realm-library/src/main/java/io/realm/internal/OsResults.java
```diff
@@ -85,8 +85,6 @@ public T next() {
 
         /**
          * Not supported by Realm collection iterators.
-         *
-         * @throws UnsupportedOperationException
```
cmelchior: Not sure I understand this? The implementation does throw?

Zhuinden:
I'll be super-happy if this solves my performance issue 😄 

although I'm not sure if it solves `nativeGetRow`?

realm/realm-library/src/main/java/io/realm/internal/OsResults.java
beeender: yeah, but empty description of `@throws` causes lint warnings. And the description is really redundant for this obvious deprecated method. So ...

mgohin:
Yay ! Can't wait to test it ! 👏

PR #5567, opened 2017-11-24
Ensure stable classes order in generated mediators

### LINKED ISSUE #5566: io.realm.DefaultRealmModuleMediator is not stable
#### Goal

If no source changes are made then the generated classes by Realm should be stable, otherwise building is slower due to incremental compilers, transforms and dexers doing extra work to process the changed outputs.

#### Expected Results

Given a simple project with 2 realm objects, `com.example.realm.MyRealmObject` and `com.example.realm.MyAnotherRealmObject` the generated class `io.realm.DefaultRealmModuleMediator` should always have the same contents.

#### Actual Results

The order of modelClasses in mediator varies randomly from build-to-build (code taken from the `static {}` initializer of `DefaultRealmModuleMediator`)

sometimes it is:

```java
Set<Class<? extends RealmModel>> modelClasses = new HashSet<Class<? extends RealmModel>>(2);
modelClasses.add(com.example.realm.MyAnotherRealmObject.class);
modelClasses.add(com.example.realm.MyRealmObject.class);
MODEL_CLASSES = Collections.unmodifiableSet(modelClasses);
```

and other times it is:

```java
Set<Class<? extends RealmModel>> modelClasses = new HashSet<Class<? extends RealmModel>>(2);
modelClasses.add(com.example.realm.MyRealmObject.class);
modelClasses.add(com.example.realm.MyAnotherRealmObject.class);
MODEL_CLASSES = Collections.unmodifiableSet(modelClasses);
```

Any realm project with more than 1 realm object.

Realm version(s): 4.3.0
Android Studio version: 3.0.1

The problem seems to be the `HashSet<ClassMetaData>` sets used in the processor. It does not specify a hashcode so `Object#hashCode` is used which is not stable from run-to-run. Hence the ordering within the set is random. A quick fix would be to use `LinkedHashSet` instead, will send a PR shortly.

madisp:
Fixes #5566.

### DIFF
--- a/realm/realm-annotations-processor/src/main/java/io/realm/processor/ModuleMetaData.java
+++ b/realm/realm-annotations-processor/src/main/java/io/realm/processor/ModuleMetaData.java
@@ -18,6 +18,7 @@ package io.realm.processor;
 
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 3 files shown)


madisp:
@cmelchior CLA done

PR #5568, opened 2017-11-24
Add support for new sort and distinct predicates

cmelchior:
This is part of #5538 

The proposed API for supporting partial Sync is currently aiming at introducing `query.subscribeAsync()`. With our current query API that is problematic as sorting and distinct information is currently inlined in the `findAllSorted` and `distinct()` methods, which will require a ton of annoying overloads to support + there is also some benefit to only have 1 or 2 "execute" methods and delegate all the query set up to predicates. 

With this PR, these methods are now deprecated and are being converted to predicates instead, so:

```
// findAllSorted is being replaced by sort
realm.where(Person.class).findAllSorted("name")
realm.where(Person.class).sort("name").findAll()

// distinct is being moved to a predicate
realm.where(Person.class).distinct("name")
realm.where(Person.class).distinct("name").findAll()
```

This also has the advantage that you can now combine `sort` and `distinct` in a single query, something that wasn't possible before.

Unfortunately, we cannot make this change without it being a breaking change as `distinct` cannot both be a `findAll` method and a predicate at the same time. So as an intermediate step, the `distinct` predicate is `distinctValues` and will be renamed in the next major version. The new API is marked in `@Beta` for that reason. It isn't ideal, but only other way would have been to delay everything until the next major release which would have made releasing incremental Partial Realm Support impossible.

The most relevant parts of this PR is in `RealmQuery`, the rest is just making sure that tests are not using deprecated functions.

TODO
- [x] Update distinct tests
- [x] Verify that sorting on children works.


CHANGELOG.md
```diff
@@ -28,12 +28,12 @@
...
 
 * Use `OsList` instead of `OsResults` to add notification token on for `RealmList<RealmModel>`.
-* Updated Gralde and plugins to support Android Studio `3.0.0` (#5472).
+* Updated Gradle and plugins to support Android Studio `3.0.0` (#5472).
 
 ### Credits
 
 * Thanks to @tbsandee for fixing a typo (#5548).
-* Thanks to @vivekkiran for updating Gralde and plugins to support Android Studio `3.0.0` (#5472).
+* Thanks to @vivekkiran for updating Grale and plugins to support Android Studio `3.0.0` (#5472).
```
Zhuinden: Gradle!

Zhuinden:
I think this is great! Can barely wait for Realm 5.0! 😄

cmelchior:
Ready for review @realm/java

realm/realm-library/src/androidTest/java/io/realm/SortTest.java
```diff
@@ -559,18 +564,70 @@ public void sortByLongDistinctByInt() {
...
+
+        // Case 1: Selecting highest numbers
+        RealmResults<AnnotationIndexTypes> results1a = realm.where(AnnotationIndexTypes.class)
+                .sort(AnnotationIndexTypes.FIELD_INDEX_LONG, Sort.DESCENDING)
+                .distinctValues(AnnotationIndexTypes.FIELD_INDEX_INT)
+                .findAll();
+        assertEquals(1, results1a.size());
+        assertEquals(3, results1a.get(0).getIndexLong());
+
+        RealmResults<AnnotationIndexTypes> results1b = realm.where(AnnotationIndexTypes.class)
```
beeender: This seems to be against my instinct. So the sorting is still evaluated first then distinct?
Another question is how does `or()` and `beginGroup()`/`endGroup()` work with `sort()` and `distinct()`?

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -1865,6 +1873,121 @@ public long count() {
         return createRealmResults(query, sortDescriptor, null, false);
     }
 
+    /**
+     * Sort the query result by the specific field name in ascending order. This predicate can be applied multiple
```
beeender: `Sort` -> `Sorts`. And also other places

beeender: `This predicate can be applied multiple times, but only the last specified sorting order will be used.`
Would it be better we just throw in this case?
Also, this can be called on different fields, what will be the behavior then? like:
`query.sort("A").sort("B").sort("C");`

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -1865,6 +1873,121 @@ public long count() {
...
+     * @param sortOrder how to sort the results.
+     * @throws java.lang.IllegalArgumentException if the field name does not exist.
+     */
+    public RealmQuery<E> sort(String fieldName, Sort sortOrder) {
+        realm.checkIfValid();
+        return sort(new String[] { fieldName}, new Sort[] { sortOrder});
+    }
+
+    /**
+     * Sort the query result by the specific field names in the provided order. {@code fieldName2} is only used
```
beeender: `order` -> `orders`

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -1865,6 +1873,121 @@ public long count() {
...
+     * @param sortOrder2 sort order for second field
+     * @throws java.lang.IllegalArgumentException if the field name does not exist.
+     */
+    public RealmQuery<E> sort(String fieldName1, Sort sortOrder1, String fieldName2, Sort sortOrder2) {
+        realm.checkIfValid();
+        return sort(new String[] { fieldName1, fieldName2 }, new Sort[] { sortOrder1, sortOrder2 });
+    }
+
+    /**
+     * Sort the query result by the specific field names in the provided order. Later fields will only be used
```
beeender: `order` -> `orders`

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -1865,6 +1873,121 @@ public long count() {
...
+        }
+        sortDescriptor = SortDescriptor.getInstanceForSort(getSchemaConnector(), query.getTable(), fieldNames, sortOrders);
+        return this;
+    }
+
+    /**
+     * BETA API: Will be renamed to {@code distinct} in next major release.
+     *
+     * Selects a distinct set of objects of a specific class. If the result is sorted, the first object will be
+     * returned in case of multiple occurrences, otherwise it is undefined which object is returned. This
```
beeender: `is returned` -> `will be returned`.

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -1865,6 +1873,121 @@ public long count() {
...
+     * predicate can be applied multiple times, but only the last applied will be used.
```
beeender: similar like `sort()`, shall we throw?

realm/realm-library/src/androidTest/java/io/realm/SortTest.java
cmelchior: Sort/distinct just saves the descriptor and applies it when you create the `OsResults` object, so the order doesn't matter for that reason.

But semantically I would say that doing distinct before sorting would result in "undefined" in terms of which objects are returned in case of similar value. With sorting first you pick the first encountered after sorting. 

So in this case, "undefined" just means do whatever would have happened if you did sort first. That looks acceptable to me?

What is slightly more annoying is that you can put sort/distinct anywhere, even where it doesn't make a lot of sense:

```
query.equalTo("field", 1).or().distinct("field").equalTo("field", 2).findAll()`
query.equalTo("field", 1).or().beginGroup().sort("field").equalTo("field", 2).endGroup().findAll()`
```

I don't think we can do much about this currently as it would require a more extensive refactoring of the entire query system. On the other side, as long as we don't support subqueries then it isn't directly harmful either from what I can tell?

realm/realm-library/src/main/java/io/realm/RealmQuery.java
cmelchior: In each case a new sort descriptor is created replacing the existing ones, meaning that only `sort("C")` is applied. You might be right that throwing might be a safer route, then we can re-evaluate if people really want to override older sortings (I suspect that no-one will run into it).

realm/realm-library/src/main/java/io/realm/RealmQuery.java
cmelchior: Sounds like a good idea 👍  See above

realm/realm-library/src/androidTest/java/io/realm/SortTest.java
beeender: what about force `sort()`/`distinct()` to be called at the end?
```
query.equalTo().sort().findAll(); // good
query.sort().equalTo() // throws
query.sort().distinct() // good
```

cmelchior: Yes, that was my idea as well, but without any help from the type system, I think that would actually be worse than allowing it anywhere? Especially since it only impacts readability if you have it elsewhere?

beeender: OK ... just document this in the `sort()`/`distinctValues()` then. We can make a proper design in the next major release.

PR #5571, opened 2017-11-27
Improve encryption example

cmelchior:
Based on question here: https://stackoverflow.com/questions/47505171/realm-database-decryption-failed


Aks-4125:
+1 for working example.

examples/encryptionExample/src/main/java/io/realm/examples/encryption/Util.java
```diff
@@ -0,0 +1,33 @@
+/*
+ * Copyright 2016 Realm Inc.
```
beeender: 2017

PR #5576, opened 2017-11-29
Life cycle of temp OsSharedRealm in callbacks

### LINKED ISSUE #5570: Crash on deleting realm after migrating to 4.1.1 from 3.3.1
We are using below code to to open realm instance. It used to work fine for version 3.1.1
Idea is that we have some migrations for last two three versions but for version before that we dont so it throws exception or in case migration fails we delete that current realm and open again.

this used to work fine in 3.1.1 how ever we just upgraded to 4.1.1 last sprint and after it we are seeing a-lot of crashes.
```
            Realm.init(context);
            RealmConfiguration configuration = new RealmConfiguration.Builder()
                    .schemaVersion(<version>)
                    .migration(<migrator>)
                    .build();

            try {
                defaultInstance = Realm.getInstance(configuration );
            } catch (IllegalArgumentException | IllegalStateException | RealmMigrationNeededException ex) {
                Realm.deleteRealm(configuration);
                defaultInstance = Realm.getInstance(configuration);
            }
```

#### Expected Results

>it should not crash

#### Actual Results

This is the exception we are getting
```
Caused by java.lang.IllegalStateException: It's not allowed to delete the file associated with an open Realm. Remember to close() all the instances of the Realm before deleting its file: /data/data/com.x.x/files/default.realm
       at io.realm.Realm.io.realm.BaseRealm.deleteRealm(SourceFile:14624)
       at com.x.x.x.RealmManager.getDefaultInstance(SourceFile:52)
```

Realm version(s): 4.1.1

Android Studio version: 3.0.0

Which Android version and device: no specific version happening on almost every os

beeender:
Every temp OsSharedRealm created during construction for the callbacks
have to be closed before the exception throws to users.
close #5570

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 * Added missing `toString()` for the implementation of `OrderedCollectionChangeSet`.
 * Sync queries are evaluated immediately to solve the performance issue when the query results are huge, `RealmResults.size()` takes too long time (#5387).
+* Close the temporary Realm instances properly during construction to avoid `IllegalStateException` when delete Realm in the catch block (#5570).
 
 ### Internal
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 
 * Added missing `toString()` for the implementation of `OrderedCollectionChangeSet`.
 * Sync queries are evaluated immediately to solve the performance issue when the query results are huge, `RealmResults.size()` takes too long time (#5387).
-* Close the temporary Realm instances properly during construction to avoid `IllegalStateException` when delete Realm in the catch block (#5570).
+* Correctly close the Realm instance if an exception was thrown while opening it. This avoids `IllegalStateException` when deleting the Realm in the catch block (#5570).
 
 ### Internal
 
--- a/realm/realm-library/src/main/java/io/realm/internal/OsSharedRealm.java
+++ b/realm/realm-library/src/main/java/io/realm/internal/OsSharedRealm.java
@@ -160,6 +160,14 @@ public final class OsSharedRealm implements Closeable, NativeObject {
     // JNI will only hold a weak global ref to this.
     public final RealmNotifier realmNotifier;
     public final Capabilities capabilities;
+    // For the Java callbacks during constructing in Object Store, some temporary OsSharedRealm objects need to be
+    // created as the parameter of the callback. The native pointers of those temp OsSharedRealm objects have to be
+    // valid during the whole life cycle of the Java object. The living native pointers still hold a ref-count to the
+    // SharedRealm which means the SharedRealm won't be closed automatically if there is any exception throws during
+    // construction. GC will clear them later, but that would be too late. So we are tracking the temp OsSharedRealm
+    // during the construction stage and manually close them if exception throws.
+    private final static List<OsSharedRealm> sharedRealmsUnderConstruction = new CopyOnWriteArrayList<OsSharedRealm>();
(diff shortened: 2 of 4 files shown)


realm/realm-library/src/main/java/io/realm/internal/OsSharedRealm.java
```diff
@@ -198,6 +221,18 @@ private OsSharedRealm(long nativeSharedRealmPtr, OsRealmConfig osRealmConfig) {
         // This instance should never need notifications.
         this.realmNotifier = null;
         nativeSetAutoRefresh(nativePtr, false);
+
+        boolean foundParentSharedRealm = false;
+        for (OsSharedRealm sharedRealm : sharedRealmsUnderConstruction) {
```
cmelchior: Not sure I fully understand this fix? This will only work if _all_ the temporary `SharedRealm` instances are being created from the Java side? I cannot tell from this PR if this is the case or not?

realm/realm-library/src/main/java/io/realm/internal/OsSharedRealm.java
```diff
@@ -160,6 +160,14 @@ protected PartialSyncCallback(String className) {
     // JNI will only hold a weak global ref to this.
     public final RealmNotifier realmNotifier;
     public final Capabilities capabilities;
+    // For the Java callbacks during constructing in Object Store, some temporary OsSharedRealm objects need to be
+    // created as the parameter of the callback. The native pointers of those temp OsSharedRealm objects have to be
+    // valid during the whole life cycle of the Java object. The living native pointers still hold a ref-count to the
+    // SharedRealm which means the SharedRealm won't be closed automatically if there is any exception throws during
+    // construction. GC will clear them later, but that would be too late. So we are tracking the temp OsSharedRealm
+    // during the construction stage and manually close them if exception throws.
+    private final static List<OsSharedRealm> sharedRealmsUnderConstruction = new CopyOnWriteArrayList<OsSharedRealm>();
```
cmelchior: Does this need to be a `CopyOnWriteArrayList` ? If I remember correctly are all `Realm.getInstance()` methods under a synchronized lock before we hit this part of the code?

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 * Added missing `toString()` for the implementation of `OrderedCollectionChangeSet`.
 * Sync queries are evaluated immediately to solve the performance issue when the query results are huge, `RealmResults.size()` takes too long time (#5387).
+* Close the temporary Realm instances properly during construction to avoid `IllegalStateException` when delete Realm in the catch block (#5570).
```
cmelchior: I would rephrase this slightly so it becomes more clear to users. Something like:

"Correctly close the Realm instance if an exception was thrown while opening it. This avoids `IllegalStateException` when deleting the Realm in the catch block (#5570)."

realm/realm-library/src/main/java/io/realm/internal/OsSharedRealm.java
beeender: nope, we only lock on the same configuration now . see https://github.com/realm/realm-java/pull/4551

realm/realm-library/src/main/java/io/realm/internal/OsSharedRealm.java
beeender: The temp instances created in native code will be released immediately and guaranteed by RAII (not the stupid GC :) )

So we only need to worry about the instances created by java, see `runMigrationCallback` and `runInitializationCallback`. Also this constructor is designed for this purpose only and it is private method.

PR #5592, opened 2017-12-05
Fixes Kotlin extension functions always being added

### LINKED ISSUE #5587: 4.3.0 adds kotlin stdlib into project
After update from 4.2.0, I found that my project had a dependency on kotlin stdlib.

Well, I'm glad that kotlin became official language for Android Development, but isn't it too soon? :)
I guess, that is a bug and kotlin + extensions must be optional.

Realm version(s): 4.3.0
Android Studio version: 3.0.1

cmelchior:
Closes #5587 

Big ups. Could have sworn I tested this 😞

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## 4.3.1 (YYYY-MM-DD)
+
+### Bug Fixes
+
+* Fixed kotlin standard library being added to both Java and Kotlin projects (#5587).
+
+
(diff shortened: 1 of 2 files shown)


bmunkholm:
Could this be tested by test cases?

cmelchior:
No, it cannot be tested, or well, our kotlinExample uses it, but it isn't really feasible to test the negative variant.

PR #5610, opened 2017-12-11
Better exception message for addField()

### LINKED ISSUE #3388: Throwing wrong exception when trying to add new field
```
schema.get("Person")
  .addField("field_tmp", Dog.class)
```

Should throw `Use addRealmObjectField() instead to add fields that link to other RealmObjects:`
but instead throws `Realm doesn't support this field type:`

Also this is not documented in the Javadoc either.

beeender:
Close #3388

### DIFF
--- a/realm/realm-library/src/main/java/io/realm/MutableRealmObjectSchema.java
+++ b/realm/realm-library/src/main/java/io/realm/MutableRealmObjectSchema.java
@@ -87,6 +87,10 @@ class MutableRealmObjectSchema extends RealmObjectSchema {
         if (metadata == null) {
             if (SUPPORTED_LINKED_FIELDS.containsKey(fieldType)) {
                 throw new IllegalArgumentException("Use addRealmObjectField() instead to add fields that link to other RealmObjects: " + fieldName);
+            } else if (RealmModel.class.isAssignableFrom(fieldType)) {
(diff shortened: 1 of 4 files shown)


realm/realm-library/src/main/java/io/realm/MutableRealmObjectSchema.java
```diff
@@ -87,6 +87,10 @@ public RealmObjectSchema addField(String fieldName, Class<?> fieldType, FieldAtt
         if (metadata == null) {
             if (SUPPORTED_LINKED_FIELDS.containsKey(fieldType)) {
                 throw new IllegalArgumentException("Use addRealmObjectField() instead to add fields that link to other RealmObjects: " + fieldName);
+            } else if (RealmModel.class.isAssignableFrom(fieldType)) {
```
cmelchior: The issue also mention that we should document this in the Javadoc

beeender: it has been documented already:
```
* To add fields that reference other RealmObjects or RealmLists use
     * {@link #addRealmObjectField(String, RealmObjectSchema)} or {@link #addRealmListField(String, RealmObjectSchema)}
     * instead.
```

how ever, there is a bug in the java doc, will fix it.

beeender:
CI failure is not caused by this. merge.

PR #5611, opened 2017-12-11
Partial Sync States in OrderedChangeSet

### LINKED ISSUE #5619: Partial Sync: Unify Notification System
See https://github.com/realm/product/issues/392 for details.

Note that this introduce a small breaking change to the Java API's as `OrderedChangeSet` will no longer be `@Nullable` in callbacks.

cmelchior:
Closes #5619 
Part of #5359 

We need to figure out what to do with this branch. Either

a) Merge to some new `master-5.0` and releases from that or easier:
b) Merge to master as we don't expect any minor releases before this is released.


cmelchior:
While there still are a few outstanding issues around exactly which states to expose and how, the general concepts in this PR should be ready for review @realm/java

realm/realm-library/src/main/cpp/collection_changeset_wrapper.hpp
```diff
@@ -0,0 +1,89 @@
...
+    CollectionChangeSetWrapper& operator=(CollectionChangeSetWrapper&&) = delete;
+    CollectionChangeSetWrapper(CollectionChangeSetWrapper const&) = delete;
+    CollectionChangeSetWrapper& operator=(CollectionChangeSetWrapper const&) = delete;
+
+    CollectionChangeSet& get()
+    {
+        return m_changeset;
+    };
+
+    jthrowable get_error() {
```
beeender: Return a `JavaGlobalRef` instead to make the life cycle clearer

realm/realm-library/src/main/cpp/collection_changeset_wrapper.hpp
```diff
@@ -0,0 +1,89 @@
...
+        JNIEnv* env = JniUtils::get_env(false);
+        if (m_error_message != "") {
+            static JavaClass realm_exception_class(env, "io/realm/exceptions/RealmException");
+            static JavaMethod realm_exception_constructor(env, realm_exception_class, "<init>", "(Ljava/lang/String;)V");
+            return (jthrowable) env->NewObject(realm_exception_class, realm_exception_constructor, to_jstring(env, m_error_message));
```
beeender: use below code in `java_exception_def.hpp`/`.cpp`

```c++
class JavaExceptiondef {
// ...
static const char* "
}
```

beeender: use below code in `java_exception_def.hpp`/`.cpp`

```c++
class JavaExceptiondef {
// ...
static const char* RealmException;

// ...

static JavaGlobalRef create_realm_exception(JNIEnv* env, const std::string& error_message) {
     static JavaClass realm_exception_class(env, "io/realm/exceptions/RealmException");
     static JavaMethod realm_exception_constructor(env, realm_exception_class, "<init>", "(Ljava/lang/String;)V");
     return JavaGlobalRef(env, env->NewObject(realm_exception_class, realm_exception_constructor, to_jstring(env, m_error_message)));
}
}
```

realm/realm-library/src/main/cpp/collection_changeset_wrapper.hpp
```diff
@@ -0,0 +1,89 @@
...
+        } else if (m_changeset.partial_sync_error_message != "") {
+            // Indicates a soft error, i.e. illegal name of query.
+            static JavaClass illegal_argument_class(env, "java/lang/IllegalArgumentException");
+            static JavaMethod illegal_argument_constructor(env, illegal_argument_class, "<init>", "(Ljava/lang/String;)V");
+            return (jthrowable) env->NewObject(illegal_argument_class, illegal_argument_constructor, to_jstring(env, m_changeset.partial_sync_error_message));
```
beeender: ditto

realm/realm-library/src/main/cpp/collection_changeset_wrapper.hpp
```diff
@@ -0,0 +1,89 @@
...
+        } else {
+            return nullptr;
```
beeender: `return JavaGlobalRef();`

realm/realm-library/src/main/java/io/realm/internal/OsList.java
```diff
@@ -18,6 +18,7 @@
     private static final long nativeFinalizerPtr = nativeGetFinalizerPtr();
     private final ObserverPairList<CollectionObserverPair> observerPairs =
             new ObserverPairList<CollectionObserverPair>();
+    private int notificationCount = 0; // Keeps track of how many times a notification has been
```
beeender: doesn't seem to be needed?

realm/realm-library/src/main/java/io/realm/internal/ObservableCollection.java
```diff
@@ -9,22 +9,23 @@
 // Helper class for supporting add change listeners on OsResults & OsList.
 @Keep
 interface ObservableCollection {
-
     class CollectionObserverPair<T> extends ObserverPairList.ObserverPair<T, Object> {
+        public int called = 0;
```
beeender: Is this for debugging only?

realm/realm-library/src/main/java/io/realm/OrderedCollectionChangeSet.java
```diff
@@ -28,6 +30,64 @@
...
+
+    /**
+     * State describing the nature of the changeset.
+     */
+    public enum State {
+        /**
+         * This state is used first time the callback is invoked. The query will have completed and
+         * data is ready for the UI.
+         */
+        INITIAL,
```
beeender: So this presents the state where the query returns the first time and null change set passed to user before? 

Please notice there might be some error cases for the sync query -- Object Store don't distinguish the sync & async queries. So we had some hacks in the java to simulate our old sync query behaviour -- the first time callback called for sync query, the change set is not null or empty, it is supposed to contain the changeset between the data at the sync query created and the latest. Which means, to match that behavior, the state should be `UPDATE` here . So sync query doesn't have `INITIAL` state.

But it is actually quite a lot headache for us, see https://github.com/realm/realm-object-store/issues/605

I would suggest that we totally follow the cocoa API, merge `findAll` & `findAllAsync` into one API, all the queries are async by default (and they are actually already.). So every query callback will have an `INITIAL` state. Since we have a chance to make breaking changes, we'd better do it now to save lots of time in the future.

cmelchior: Good point about synchronous queries not really having an `INITIAL` state, but only `UPDATES`.

But I'm not sure that following Cocoas API is actually the right choice, to be honest. I mean, I can see it will help _us_ on the internal side, but it is a quite severe breaking change that also causes a more unsafe and confusing API, as if you by accident do the wrong thing, then you switched an async query to be synchronous. This is what annoys me most about having a single API, nothing prevents you from shooting yourself in the foot, and Android developers have historically wanted more control over their threads than Cocoa devs.

With regard to naming  then:
If we combine them in `findAll()`: It breaks our naming pattern used everywhere else that suffixes `Async` to API's that uses a worker thread.

If we combine them in `findAllAsync()`: Then we have to explain to people that touching any property outside a change listener will cause it to evaluate the query immediately.

Part of this issue is also that the Java API returns a `RealmResult` immediately instead of a token, which makes it a lot easier to do wrong things by accident.

If we followed the Cocoa API, our API would look like this:

```
NotificationToken token = realm.where(Person.class).findAllAsync(new OrderedRealmCollectionChangeListener() {
    @Override
    public voic onChange(RealmResults<Person> persons, OrderedColllectionChangeSet changeSet) { 
     // ... 
    }
});
```

No Android/Java API I have ever seen does it like this.

beeender: No need to expose the token to users. We still return the `RealmResults`.

If we can solve 
https://github.com/realm/realm-object-store/issues/605 , the current API is also good and easy to maintain.

realm/realm-library/src/main/java/io/realm/OrderedCollectionChangeSet.java
```diff
@@ -73,6 +133,16 @@
      */
     Range[] getChangeRanges();
 
+    /**
+     * Returns any error that happened. If an error has happened, the state of the collection and other
+     * changeset information is undefined. It is is possible for a collection to go into an error state
```
nhachicha: > It is is

realm/realm-library/src/main/java/io/realm/OrderedCollectionChangeSet.java
```diff
@@ -28,6 +30,64 @@
  * change, or an array of {@link Range}s.
  */
 public interface OrderedCollectionChangeSet {
+
+    /**
+     * State describing the nature of the changeset.
+     */
+    public enum State {
```
nhachicha: Some of the descriptions of the states apply mainly for Sync use case. It might be confusing to read/introduce this to non Sync usage ...

realm/realm-library/src/main/java/io/realm/internal/OsCollectionChangeSet.java
```diff
@@ -152,4 +153,39 @@ public String toString() {
...
+    @Nullable
+    private native Object nativeGetError(long nativePtr);
+
+    private native int nativeGetOldStatusCode(long nativePtr);
+
+    private native int nativeGetNewStatusCode(long nativePtr);
+
+    // Returns true if the data described by the subscription has been downloaded to the device,
+    // false if not. In either case, the query is run against the local dataset.
+    private native boolean nativeIsRemoteDataLoaded(long nativePtr);
```
nhachicha: I might have misunderstood the part saying _query is run against the local dataset_ 
The name suggested a check to see if the data was loaded, I'm not sure why the query should run locally as well?

realm/realm-library/src/main/java/io/realm/internal/OsCollectionChangeSet.java
```diff
@@ -152,4 +153,39 @@ public String toString() {
...
+
+    private native static long nativeGetFinalizerPtr();
+
+    // Returns the ranges as an long array. eg.: [startIndex1, length1, startIndex2, length2, ...]
```
nhachicha: as a long

realm/realm-library/src/main/java/io/realm/internal/OsList.java
nhachicha: related to the `called` attribute for debugging?

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/PartialSyncTests.java
```diff
@@ -34,20 +40,89 @@
...
+
+        final Realm realm = Realm.getInstance(partialSyncConfig);
+        looperThread.closeAfterTest(realm);
+
+        // FIXME: Currently all queries are invalid
+        RealmResults<AllTypes> query = realm.where(AllTypes.class).findAllAsync();
+        query.addChangeListener(new OrderedRealmCollectionChangeListener<RealmResults<AllTypes>>() {
+            @Override
+            public void onChange(RealmResults<AllTypes> allTypes, OrderedCollectionChangeSet changeSet) {
+                // TODO It is a bit undeterministic what is returned. If Sync is fast enough,
```
nhachicha: indeterministic

realm/realm-library/src/main/java/io/realm/OrderedCollectionChangeSet.java
Zhuinden: @cmelchior

    Disposable disposable = observable.subscribeWith((newValue) -> {
        //...
    });

cmelchior: @Zhuinden You just had to prove me wrong 😉 

But that is actually a great example. Thanks 👍

realm/realm-library/src/main/cpp/io_realm_internal_OsCollectionChangeSet.cpp
```diff
@@ -140,11 +140,11 @@ JNIEXPORT jboolean Java_io_realm_internal_OsCollectionChangeSet_nativeIsRemoteDa
 JNIEXPORT jint JNICALL Java_io_realm_internal_OsCollectionChangeSet_nativeGetOldStatusCode(JNIEnv*, jobject, jlong native_ptr) {
     TR_ENTER_PTR(native_ptr)
     auto& change_set = *reinterpret_cast<CollectionChangeSetWrapper*>(native_ptr);
-    return realm::partial_sync::state_to_status_code(change_set.get().partial_sync_old_state);
+    return (int8_t) change_set.get().partial_sync_old_state;
```
kneth: `static_cast<int8_t>(change_set.get().partial_sync_old_state)` will be consistent with the rest of the code.

realm/realm-library/src/main/cpp/collection_changeset_wrapper.hpp
cmelchior: Trying to do this resulted in `JNI DETECTED ERROR IN APPLICATION: use of deleted global`. After spending a few hours on it, I reverted it back to this as it didn't seem important enough to spend more time on.

realm/realm-library/src/main/java/io/realm/internal/ForcedLoadChangeset.java
```diff
@@ -0,0 +1,117 @@
...
+ * limitations under the License.
+ */
+package io.realm.internal;
+
+import io.realm.RealmResults;
+
+/**
+ * Fake changeset used if {@link RealmResults#load()} is called manually.
+ */
+public class ForcedLoadChangeset extends OsCollectionChangeSet {
```
nhachicha: Why not called `EmptyLoadChangeset` since it's used in this context https://github.com/realm/realm-java/pull/5611/files/ff07d57f17539b9365949465abbda2c85378cb07..65a392c552ce267e726e8fd64ba1f9fb0022740d#diff-013603dffe5ba56b83771eaefb17ce9cR435 
the I feel `ForceLoad` represents the action that triggers the use of this class not it's state

realm/realm-library/src/main/java/io/realm/internal/StatefulCollectionChangeSet.java
```diff
@@ -55,52 +50,37 @@ public State getState() {
...
 
     @Nullable
     @Override
     public Throwable getError() {
         return error;
     }
+
+    @Override
+    public boolean isCompleteResult() {
+        return remoteDataSynchronized;
```
nhachicha: Why is this idempotent? shouldn't we call `changeset.isRemoteDataLoaded()` again to check if the data download finished?

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/PartialSyncTests.java
```diff
@@ -53,26 +49,24 @@ public void invalidQuery() {
...
-                        assertTrue(changeSet.getError() instanceof IllegalArgumentException);
-                        Throwable iae = changeSet.getError();
-                        assertTrue(iae.getMessage().contains("ERROR: realm::QueryParser: Comparison operator missing"));
-                        looperThread.testComplete();
-                        break;
-
-                    default:
-                        fail("Unexpected state: " + changeSet.getState());
-                }
+        // Backlinks not yet supported: https://github.com/realm/realm-core/pull/2947
```
nhachicha: FYI PR is merged in Core, is this still true? ...

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/PartialSyncTests.java
```diff
@@ -53,26 +49,24 @@ public void invalidQuery() {
...
+        RealmResults<AllJavaTypes> query = realm.where(AllJavaTypes.class).equalTo("objectParents.fieldString", "Foo").findAllAsync();
+        query.addChangeListener((results, changeSet) -> {
+            switch (callbacks.incrementAndGet()) {
+                case 1:
+                    assertEquals(OrderedCollectionChangeSet.State.INITIAL, changeSet.getState());
```
nhachicha: Do we have the guarantee that we `INITIAL` state will be triggered and not directly `ERROR`?

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/utils/HttpUtils.java
```diff
@@ -56,6 +56,10 @@ public static void startSyncServer() throws Exception {
 
         Response response = client.newCall(request).execute();
         if (!response.isSuccessful()) throw new IOException("Unexpected code " + response);
+
+        // Work around race condition between starting ROS and logging in first user
+        // See https://github.com/realm/ros/issues/389
```
nhachicha: you should reopen the issue on ROS repo if this still happens

realm/realm-library/src/androidTest/java/io/realm/OrderedCollectionChangeSetTests.java
```diff
@@ -469,14 +469,42 @@ public void emptyChangeSet_findAllAsync() {
...
+        Realm realm = looperThread.getRealm();
+        populateData(realm, 10);
+        final RealmResults<Dog> results = realm.where(Dog.class).sort(Dog.FIELD_AGE).findAll();
+        looperThread.keepStrongReference(results);
+        results.addChangeListener((collection, changeSet) -> {
+            assertSame(collection, results);
+            assertEquals(11, collection.size());
+            assertTrue(changeSet.isCompleteResult());
+            assertEquals(OrderedCollectionChangeSet.State.UPDATE, changeSet.getState());
+            assertEquals(1, changeSet.getInsertions().length);
```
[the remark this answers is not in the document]
cmelchior: Actually, this test is for the synchronous query. So no empty initial changeset is sent.

realm/realm-library/src/main/java/io/realm/OrderedCollectionChangeSet.java
```diff
@@ -135,14 +115,44 @@
...
+     * fully synchronized Realm.
+     * <p>
+     * Consider the following case:
+     * <ol>
+     *   <li>An app is online and makes a query for all messages containing the word "Realm".</li>
+     *   <li>Partial synchronization downloads all those messages.</li>
+     *   <li>The app goes offline.</li>
+     *   <li>The app makes an offline query against all messages containing the word "Database".</li>
+     * </ol>
+     * In this case, the query result will only contain the subset of messages that contains both
```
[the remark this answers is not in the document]
cmelchior: Yes, I can rephrase. But all queries always run locally always, it is just that they _also_ start downloading data in the background.

realm/realm-library/src/main/java/io/realm/internal/ForcedLoadChangeset.java
```diff
@@ -0,0 +1,117 @@
...
+
+    private static final int[] NO_INDEX_CHANGES = new int[0];
+    private static final Range[] NO_RANGE_CHANGES = new Range[0];
+
+    public ForcedLoadChangeset() {
+        super(0, true);
+        // FIXME Read partial sync status from Realm when creating this
```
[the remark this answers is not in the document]
cmelchior: This is because otherwise, it might report the wrong value for `IsCompleteResult`.

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/PartialSyncTests.java
cmelchior: Yes, it is based on the Object Store implementation which should be deterministic about this

cmelchior:
Comments addressed @nhachicha

realm/realm-library/src/main/java/io/realm/internal/StatefulCollectionChangeSet.java
cmelchior: Each changeset is a value object, so if this value changes a new ChangeSet object is created.

realm/realm-library/src/main/cpp/io_realm_internal_OsCollectionChangeSet.cpp
```diff
@@ -140,11 +140,17 @@ JNIEXPORT jboolean Java_io_realm_internal_OsCollectionChangeSet_nativeIsRemoteDa
...
     auto& change_set = *reinterpret_cast<CollectionChangeSetWrapper*>(native_ptr);
-    return (int8_t) change_set.get().partial_sync_old_state;
+    return static_cast<int8_t>(change_set.get().partial_sync_old_state);
 }
 
 JNIEXPORT jint JNICALL Java_io_realm_internal_OsCollectionChangeSet_nativeGetNewStatusCode(JNIEnv*, jobject, jlong native_ptr) {
     TR_ENTER_PTR(native_ptr)
     auto& change_set = *reinterpret_cast<CollectionChangeSetWrapper*>(native_ptr);
-    return (int8_t) change_set.get().partial_sync_new_state;
+    return static_cast<int8_t>(change_set.get().partial_sync_new_state);
```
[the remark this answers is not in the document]
cmelchior: Expanding the size should be safe, but I can change it if you think it would make it more clear. I did this because the ObjectStore type is `int8_t` so it seemed more explicit to use it all the way through?

realm/realm-library/src/main/java/io/realm/OrderedCollectionChangeSet.java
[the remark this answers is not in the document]
cmelchior: Rephrased, but I would like to leave out the diagram as it would be neigh impossible to translate to HTML.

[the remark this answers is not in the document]
cmelchior: Reworded

Zhuinden:
```
    @Override
    public State getState() {
        throw new UnsupportedOperationException("This method should be overridden in a subclass");
    }

    @Override
    public boolean isCompleteResult() {
        throw new UnsupportedOperationException("This method should be overridden in a subclass");
    }
```

Why not make the class `abstract` and make this method abstract?

realm/realm-library/src/main/java/io/realm/internal/OsCollectionChangeSet.java
```diff
@@ -45,12 +45,19 @@
...
-    public OsCollectionChangeSet(long nativePtr) {
+    public OsCollectionChangeSet(long nativePtr, boolean firstAsyncCallback) {
         this.nativePtr = nativePtr;
+        this.firstAsyncCallback = firstAsyncCallback;
         NativeContext.dummyContext.addReference(this);
     }
 
+    @Override
+    public State getState() {
+        throw new UnsupportedOperationException("This method should be overridden in a subclass");
```
Zhuinden: Why not make the class abstract instead, and not implement this method here?

realm/realm-library/src/main/java/io/realm/internal/OsCollectionChangeSet.java
```diff
@@ -99,17 +106,41 @@ public OsCollectionChangeSet(long nativePtr) {
...
-        return nativePtr;
+    public Throwable getError() {
+        return (Throwable) nativeGetError(nativePtr);
     }
 
     @Override
-    public long getNativeFinalizerPtr() {
-        return finalizerPtr;
+    public boolean isCompleteResult() {
+        throw new UnsupportedOperationException("This method should be overridden in a subclass");
```
Zhuinden: Why not make the class abstract instead, and not implement this method here?

realm/realm-library/src/main/java/io/realm/internal/OsCollectionChangeSet.java
cmelchior: The query only ever runs locally, this boolean indicates how "trustworthy" the result is. If data hasn't been synchronized with the server, you can get query results that would not be possible in a fully synchronized Realms.

Would renaming to `nativeIsRemoteDataSynchronized()` be a better name?

realm/realm-library/src/main/java/io/realm/internal/OsCollectionChangeSet.java
cmelchior: I agree this is not pretty. The reason for this is mostly because `OsX` sort of indicates a wrapper around an ObjectStore class. This also implies that we probably shouldn't add extra responsibility to it. Which is where the `StatefulCollectionChangeSet` comes from. These methods are only here because `OsCollectionChangeset` currently need to implement the public interface `OrderedCollectionChangeSet`, but these methods don't make sense here.

I should probably take a second look at this to see if it could be made cleaner.

Zhuinden:
Supercedes https://github.com/realm/realm-java/pull/5538

PR #5614, opened 2017-12-12
Fix CI failure caused by ROS docker

beeender:
Health checking without waiting is too frequent. Use 0.5 second as the
interval between every attempt.

NOTE: CI failure caused by this usually cannot be seen the exact test
case in the gradle log. But it will give you something like:
"Tests on Nexus 5X - 6.0 failed: Test run failed to complete. Expected
3762 tests, received 3"

### DIFF
--- a/tools/sync_test_server/ros-testing-server.js
+++ b/tools/sync_test_server/ros-testing-server.js
@@ -51,7 +51,9 @@ function waitForRosToInitialize(attempts, onSuccess, onError, startSequence) {
     http.get("http://0.0.0.0:9080/health", function(res) {
         if (res.statusCode != 200) {
             winston.info("ROS /health/ returned: " + res.statusCode)
-            waitForRosToInitialize(attempts - 1, onSuccess, onError, startSequence)
+            setTimeout(function() {
+                waitForRosToInitialize(attempts - 1, onSuccess, onError, startSequence);
+            }, 500);
         } else {
             onSuccess(startSequence);
         }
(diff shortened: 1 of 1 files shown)


tools/sync_test_server/ros-testing-server.js
```diff
@@ -51,7 +51,9 @@ function waitForRosToInitialize(attempts, onSuccess, onError, startSequence) {
     http.get("http://0.0.0.0:9080/health", function(res) {
         if (res.statusCode != 200) {
             winston.info("ROS /health/ returned: " + res.statusCode)
-            waitForRosToInitialize(attempts - 1, onSuccess, onError, startSequence)
+            setTimeout(function() {
+                waitForRosToInitialize(attempts - 1, onSuccess, onError, startSequence);
+            }, 500);
```
nhachicha: Note that in reality, when ROS is not started you'll get the `ECONNREFUSED` error which is caught by the `.on('error')` block, which actually does the wait. This is just a double check when somehow the Webserver (Express) is up but the health endpoint is not successful (`200`)

tools/sync_test_server/ros-testing-server.js
```diff
@@ -112,7 +114,8 @@ function doStartRealmObjectServer(onSuccess, onError) {
                 winston.info(`${data}`);
             });
 
-            waitForRosToInitialize(100, onSuccess, onError, Date.now());
+            // The interval between every health check is 0.5 second. Give the ROS 15 seconds to get fully initialized.
+            waitForRosToInitialize(30, onSuccess, onError, Date.now());
```
nhachicha: make sure this works for both docker images, I had longer startup times when using `docker-02` if I remember (with the Shield)

PR #5646, opened 2017-12-27
javadoc for Required annotation

beeender:
It caused confusion like:
https://stackoverflow.com/questions/47940219/realmmigrationneededexception-when-adding-realmlistint-kotlin


Zhuinden:
hmm

Issue #761, opened 2015-01-19
Inheritance / Polymorphism


cmelchior:
A highly request feature is polymorphism. Mostly for use in ListAdapters, but a lot of other scenarios exists.

We need support for the following:
- Abstract RealmObject classes
- Query support

Example
```
public abstract Animal extends RealmObject {
  private String name;
}

public class Dog extends Animal {
  private boolean willPlayCatch;
}

public class Spider extends Animal {
  private boolean isScary;
}

// Following should be possible
RealmResults<Animal> animals = realm.where(Animal.class).equalsTo("name","Foo").findAll();
String name = animals.get(0).getName();

RealmResults<Dog> dogs = realm.where(Dog.class).equalsTo("name","Foo").and().equalsTo("willPlayCatch", true).findAll();
String name = dogs.get(0).getName();
```

T-Spoon:
Any estimate on a time frame for this?  I'm hoping to use Realm in an upcoming project, but without support for Model inheritance I'll have to hold off on trying it out.

Looking forward to where you guys take this project though!

cmelchior:
Hi @T-Spoon 
We have a couple of other major features we would like to look at first, so we don't have a timeframe yet, sorry. If inheritance is a dealbreaker for you, you should probably not use us this time around. But we do hope you will come back :)

caroth92:
:+1:

msegers:
Still no estimates? Going to write a lot of duplicate code since I want to use Realm anyways. Works fine in swift already. Monitoring this.

cmelchior:
Unfortunately not. The support for this in Objective C and Swift is actually just sharing fields. It is not proper polymorphism. This is good enough for some use cases, but we feel that without being able to query on abstract super classes it will be less useful on Android. Right now we are focusing on migrations, null support and async querying, but hopefully we will be able to tackle this afterwards.

nolanamy:
So is there any way to share fields on Android? We have 5 models that all share the same synchronization-related fields and code. We use inheritance in our current SQLite models; if we switch to Realm, will we have to duplicate the synchronization fields across each of the 5 model classes?

As a workaround, I'm thinking about having those classes implement an interface with getters and setters for the shared fields, which at least let me share the synchronization code... is there a better way?

AliSoftware:
While waiting for this inheritance feature to be available, is there any workaround, like a good practice to design our RLMObjects and model to make such use cases possible?

[...]

PR #5678, opened 2018-01-12
Add support for Weak/Strong links aka. Cascading Deletes

### LINKED ISSUE #1104: Support for cascading deletes
It should be possible to delete linked objects as well using some defined policy.

Possible policies:
- NEVER 
- IF_LAST_LINK
- ALWAYS

Cocoa issue: https://github.com/realm/realm-cocoa/issues/1186

cmelchior:
Closes #1104

This PR adds support for the Core concepts of Weak/Strong links or effectively Cascading deletes.

The semantics are:
- Default relationship  are "weak"
- If A has a strong relationship with B, then B is automatically deleted if A removes it's reference or is deleted itself.
- Multiple A's can hold a strong reference to B. B is only deleted when the final A is.

In particular, this solves the most common cascading delete problems we have seen:

- Nested hierarchies of objects where parents own their children.
- Updating list problem: I.e. how to delete objects no longer part of a list after the list has been modified due to a network request.

The public API is a new annotation:

```
// Model classes
@StrongRelationship
public Person person;

@StrongRelationship
public RealmList<Dog> dogs;

// DynamicRealmObject
boolean DynamicRealmObject.isStrongRelationship("linkField");

// RealmObjectSchema
boolean RealmObjectSchema.isStrongRelationship(String fieldName);
RealmObjectSchema.setStrongRelationship(String fieldName, boolean isStrongRelationship);

```
**API Thoughts:**
Cycled through a few ideas before ending on a single `@StrongRelationship`. This mimics the current behavior of e.g `@Required` where we only annotate things if different from default behavior. It restricts the number of "types" of relationships we can support, but I don't think that is a problem for the foreseeable future.

Other things attempted:
```
// More descriptive, can also set it explicitly to weak
@Relationship(RelationshipType.STRONG) 

// Reducing the number of annotations by adding it to other annotation proposed in PR about naming
@RealmField(relationship = Relationship.STRONG)
```
Should we add a `@WeakRelationship`? Nice for symmetry ala `@Nullable/@NonNull` but our own API does not have the precedence. 

TODO:
- [ ] Flesh out annotation processor unit tests
- [ ] Add DynamicRealm API + tests
- [ ] Add Migration API + tests
- [ ] Merge Sync upgrade from `master`


Zhuinden:
Oooo this is great 😄  I wonder if this will be in 5.0 along with all the other goodies

MarkOSullivan94:
Where can we find what release this will be in or what will be in the next release?

Zhuinden:
¯\_(ツ)_/¯

cmelchior:
Sorry. This was put on hold due as we discovered we needed a few more internal changes before it made sense to release this. I don't have a new timeline unfortunately 😢

Zhuinden:
Oh I see now :(

https://github.com/realm/realm-object-store/pull/622#issuecomment-385618340

lmdic:
Realm 5.8.0.  - not found cascade delete.
When you implement cascade delete in Realm?
Thanks.

ChinmayAcharya28:
Hi, 
we are using realm for database. We have created Realm classes as per our application database design. In this, There are some models which are having inter related dependancy. But they won't have primary and foreign key relation. 

They add row as per defined with db structure but when deletion comes we need to manage that for every model. does realm provides cascasde deletion ?

Zhuinden:
> They add row as per defined with db structure but when deletion comes we need to manage that for every model. 

yep

> does realm provides cascasde deletion ?

nope

ashughes:
According to @roberhofer, this is coming. See https://github.com/realm/realm-java/issues/761#issuecomment-499296327.

rovkinmax:
any progress?

bmunkholm:
No progress on that yet. We will however release "Embedded Objects" very soon. They will automatically "cascade delete". That will however not cover all needs.

PR #5684, opened 2018-01-15
fixes unknown error code

nhachicha:
The ObjectStore is now reportings unrecognized sync error to the binding https://github.com/realm/realm-object-store/commit/d36a7b603c256b9df5e9b3acc4645c770d630a31

Previously `ErrorCode#fromInt` was throwing an exception when it can't map the error to the known ones. This PR fixed this behaviour by logging a warning instead, since the Sync client may report operating system errors ex https://stackoverflow.com/questions/48217332/realm-sync-not-working-after-reconnecting-to-the-internet

### DIFF
--- a/realm/realm-library/src/androidTestObjectServer/java/io/realm/SessionTests.java
+++ b/realm/realm-library/src/androidTestObjectServer/java/io/realm/SessionTests.java
@@ -30,6 +30,7 @@ import java.util.concurrent.atomic.AtomicBoolean;
 import io.realm.entities.StringOnly;
 import io.realm.exceptions.RealmFileException;
 import io.realm.exceptions.RealmMigrationNeededException;
+import io.realm.log.RealmLog;
 import io.realm.objectserver.utils.StringOnlyModule;
 import io.realm.rule.RunInLooperThread;
 import io.realm.rule.RunTestInLooperThread;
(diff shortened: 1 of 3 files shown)


realm/realm-library/src/androidTestObjectServer/java/io/realm/SessionTests.java
```diff
@@ -125,23 +122,20 @@ public void errorHandler_clientResetReported() {
...
-                        String filePathFromError = handler.getOriginalFile().getAbsolutePath();
-                        String filePathFromConfig = session.getConfiguration().getPath();
-                        assertEquals(filePathFromError, filePathFromConfig);
-                        assertFalse(handler.getBackupFile().exists());
-                        assertTrue(handler.getOriginalFile().exists());
-
-                        looperThread.testComplete();
+                .errorHandler((session, error) -> {
+                    if (error.getErrorCode() != ErrorCode.CLIENT_RESET) {
+                        fail("Wrong error " + error.toString());
```
bmunkholm: What's a "Wrong" error? Perhaps "Unexpected"?

bmunkholm: nevermind - it's justs tests... :-)

nhachicha:
@cmelchior true `notifyErrorHandler` should catch the previously exception anyway ... this is will help rule out any side effect, and _confirm_ the issue with `Sync 2.2.9` once the snapshot is released

PR #5685, opened 2018-01-16
Remove deprecated methods

cmelchior:
Remove deprecated query and partial sync methods for 5.0


realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -1698,6 +1700,71 @@ public static boolean compactRealm(RealmConfiguration configuration) {
...
+            @Override
+            public void execute(Realm realm) {
+
+                // Need to manually run a dynamic query here.
+                // TODO Add support for DynamicRealm.executeTransactionAsync()
+                Table table = realm.sharedRealm.getTable("class___ResultSets");
+                TableQuery query = table.where()
+                        .equalTo(new long[]{table.getColumnIndex("name")}, new long[]{NativeObject.NULLPTR}, subscriptionName);
+
+                OsResults result = OsResults.createFromQuery(realm.sharedRealm, query);
```
Zhuinden: Wow I didn't know you could do that 😄

cmelchior: It is also kinda cheating :)

cmelchior:
Ready for review @nhachicha

PR #5700, opened 2018-01-19
Remove getFieldNames from proxy classes

cmelchior:
These are no longer used since we moved to the Object Store schema concept. I found this while working on #5280


cmelchior:
Ready for review @nhachicha

PR #5725, opened 2018-01-29
Fix oneOf() generics

PhilGlass:
This change allows passing an array of non-null values to the various `oneOf()` overloads. Kotlin's arrays are invariant by default, so `Array<T>` is not a subtype of `Array<T?>`.


Zhuinden:
Wasn't this already done in https://github.com/realm/realm-java/pull/5723 ?

PhilGlass:
Oh, nice. Probably should have searched the issues first. 🙈

cmelchior:
Actually, your solution is much better than what I came up with. We should probably use your approach instead of duplicating a bunch of methods which was what I did

PR #5729, opened 2018-01-30
Add support for Object Level Permissions

cmelchior:
Add support for object level permissions for partially synchronized Realms. This PR is still very rough and is mostly here to gather feedback on the public API.

### Public API

**New classes**

In package `io.realm.sync.permissions`. The old permission system lives in `io.realm.permissions`.

* `RealmPermissions`
* `ClassPermissions`
* `Permission` (conflicts with `Permission` from the old permission system :( )
* `Role`
* `User` (overlaps with `SyncUser`)
* `RealmPrivileges` 

**New methods**

```java
// in BaseRealm.java (thus both in Realm/DynamicRealm
RealmPrivileges getPrivileges();
RealmPrivileges getPrivileges(RealmModel obj);
RealmPermissions getPermissions();
RealmResults<Role> getRoles();
RealmResults<User> getUsers();
```

```java
// In Realm.java
RealmPrivileges getPrivileges(Class<? extends RealmModel> clazz); 
ClassPermissions getPermissions(Class<? extends RealmModel> clazz);
```

```java
// In DynamicRealm.java
RealmPrivileges getPrivileges(String className);
ClassPermissions getPermissions(String className);
```

### Examples

```java
// Check for write access for UI purposes
Person person = realm.where(Person.class).equalTo("name", "Jane").findFirst();
if (!realm.getPrivileges(person).canWrite()) {
    showReadOnlyLabel()
}

// Add a new user to a role
realm.getRoles().where("name", "my-role").findFirst().getMembers().add(new User(syncUser.getIdentity()));

// Listen to role changes
RealmResults<Role> roles = realm.getRoles();
roles.addChangeListener((result, changeSet) -> {
   // Something changed
})

// Change the permissions of a class
realm.executeTransactionAsync(r -> {
	ClassPermissions classPermissions = r.getPermissions(Person.class);
	if (classPermissions == null) {
		// ClassPermissions does not exist. It might just be because it is not synchronized yet.
		// Need to create it here which will override any potential server version.
                Permission permissions = new Permission.Builder().allPrivileges().canWrite(false).build();

                // Annoying that ClassPermissions have a String primary key. `createObject` doesn't work well here.
                realm.copyToRealmOrUpdate(new ClassPermissions(Person.class, permissions);
            } else {
		classPermissions.role.getPermissions().setWriteAccess(false);
	}
});

// Setup a Realm with initial permissions. The default is no-permissions
SyncUser syncUser = getUser();
SyncConfiguration config  = new SyncConfiguration(syncUser, url)
    .partialRealm()
    .initialData(realm -> {
        Role role = realm.createObject(Role.class, "my-role");
        role.addMembers().add(new User(user.getIdentity()));
        Permission appPermissions = new Permission.Builder(role).allPrivileges().build();
        realm.createObject(RealmPermissions.class).getPermissions().add(appPermissions)
     })
    .build()

// Change the permissions for a class
// Quite a lot of safe guards is needed to guard against data missing :(
realm.executeTransactionAsync(r -> {
    ClassPermissions classPermissions = r.getPermissions(Person.class);
    if (classPermissions == null) {
        // Annoying that ClassPermissions have a String primary key. `createObject` doesn't work well here.
        classPermissions = realm.copyToRealm(new ClassPermissions(Person.class));
    }

    Permission perm = classPermissions.getPermissions().where()
        .equalTo("name", "my-role")
        .findFirst();
    if (perm == null) {
        // Create default Role and Permission if not existing
        Role role = realm.getRoles().where().equalTo("name", "my-role").findFirst();
        if (role == null) {
            role = realm.copyToRealm(new Role("my-role", new User(syncUser.getIdentity())));
        }
        perm = new Permission.Builder(role).allPrivileges().build();
    }
    
    // Modify permissions for the particular class
    perm.setCanRead(false);
});
```

**Concerns**

* Multiple permission concepts that overlap in naming? How to resolve? How quickly is old system deprecated? 
* Handling missing data is cumbersome? Any way to make it easier?
* RealmPermissions is a singleton object. ObjectStore should probably create this object?

TODO:
- [ ] Determine naming / package of classes: Especially `Permission` (which conflict with `io.realm.permissions.Permission`, `User` (which conflict with `SyncUser`) and `Role`.
- [ ] Should `getPrivileges()` (overloaded), be split into `realm.getRealmPrivileges(), realm.getClassPrivileges(MyClass.class), realm.getObjectPrivileges(myObj)`
- [ ] Should `getPermissions` be split into `realm.getRealmPermissions(), realm.getClassPermissions`?
- [ ] Should we add a `getPrivileges()` method directly on `RealmObject`?
- [ ] Determine the exact methods/constructors for `Permission.Builder`, especially around no/all.
- Should we have async variants of `getPermissions()/getRoles/getUsers`?
- [ ] Creating `ClassPermissions` requires copyToRealm/insert as the primary key is the String based name for the model class, which is not easy to get to so `createObject` cannot be used easily. Is this a problem?
- [ ] Judging from examples above we need a bunch of helper constructors 
-  [x] Depend on the proper version of Sync
- [ ] Upgrade ROS
- [ ] Depend on a proper version of Object Store
- [ ] Tests, tests, tests


Zhuinden:
`role = realm.copyFromRealm(new Role("my-role", new User(syncUser.getIdentity())));`

copy FROM realm?

nhachicha:
I have read the design document as well as the sync PR & OS PR, my feeling is that the current Java proposal seems too low level, it has the correct mapping to Sync & OS concepts (Permission, User ClassPermission etc) but it doesn't mean we need to expose this low-level API to the user, as I find it a bit challenging to grasp.

What I suggest is to start from the use case point of view, we need basically two things: 

1. **Know** privileges 
1. **Modify** privileges 

### Know privileges 

Permissions are set for different levels: Realm, Class (type) and Object(row), we need a symmetric API that allows you to query for privileges in a consistent way, example
```Java
RealmPrivileges p = Privileges.of(realm);
RealmPrivileges p = Privileges.of(realm, class);
RealmPrivileges p = Privileges.of(managedObj);
```

### Modifying privileges

Permission table is the cornerstone of this scenario, we should articulate the API around it.

```Java
Permissions.add(privileges).to(realm).with("role").forUser("userId");
Permissions.add(privileges).to(realm, class).with("role").forUser("userId");
Permissions.add(privileges).to(managedObj).with("role").forUser("userId");
```
The `privileges` in this case is an `EnumSet`, we can also use it as a bit mask, to modify the existing permission in order to give `read` priviledge to a user. This could be done by setting the bit to `1` and then perform an `OR` to keep whatever existing privileges the user had before.

_Note: to remove a privilege, reverse a process by setting the bit to `0`, then perform an `AND` to keep the existing privileges but remove the selected one._

#### Roles

We can imagine a similar API based on the concept we want to interact with (here `Role`), then make it the focal point of the API.
```Java
Roles.add("user_id").to("role").in(realm);
Roles.del("user_id").to("role").in(realm);
```

Internally all these new APIs (`Privileges`, `Permissions` and `Roles`) uses the underlining concept defined in the ObjectStore, i.e have access to the `PermissionCache` perform the CRUD on the Permission Schema tables (_Permission, PermissionUser, PermissionClass, PermissionRole_).

They exist merely to simplify the API to the user and hide the internals 

*Disclaimer:* 
This proposal doesn't cover all the use cases, but it gives a direction about where we should be heading in terms of API design.

simonask:
@nhachicha Note that the underlying implementation no longer uses bitsets to denote privileges, but rather columns of booleans. Of course you are free to do so at the API level if that is more ergonomic for you. :-)

realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -492,6 +497,54 @@ public long getVersion() {
...
+
+    /**
+     * Returns all {@link Role} objects available in this Realm. Attach a change listener
+     * using {@link Role#addChangeListener(RealmChangeListener)} to be notified about
+     * any future changes.
+     *
+     * @return all roles available in the current Realm.
+     */
+    @Beta
+    abstract public RealmResults<Role> getRoles();
```
nhachicha: should we add similar method for `Class` & `Object`

realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -492,6 +497,54 @@ public long getVersion() {
...
+
+    /**
+     * Returns all {@link User} objects available in this Realm. Attach a change listener
+     * using {@link User#addChangeListener(RealmChangeListener)} to be notified about
+     * any future changes.
+     *
+     * @return all roles available in the current Realm.
+     */
+    @Beta
+    abstract public RealmResults<User> getUsers();
```
nhachicha: I suggest we work only with "user_id" as a string, this eliminates the need for the conflicting `User` class, 
example
```Java
realm.getRoles().where("name", "my-role").findFirst().getMembers().add(new User(syncUser.getIdentity()));
```
becomes 
```Java
realm.getRoles().where("name", "my-role").findFirst().getMembers().add(syncUser.getIdentity());
```

realm/realm-library/src/objectServer/java/io/realm/sync/permissions/RealmPrivileges.java
```diff
@@ -0,0 +1,118 @@
...
+import io.realm.RealmModel;
+import io.realm.internal.ManagableObject;
+
+/**
+ * FIXME: Javadoc
+ * Thread confined object representing the current level of access to either the Realm, Class
+ * or individual objects.
+ *
+ */
+public class RealmPrivileges implements ManagableObject {
```
nhachicha: this needs a `shared_realm` pointer so we can delegate to the ObjectStore method https://github.com/realm/realm-object-store/commit/c80348445d7e355ba385d2119952020df03d16af#diff-4d344943bb20e4505258929b6f4707f3R318

realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -540,16 +539,16 @@ public long getVersion() {
...
+//    /**
+//     * Returns all users (i.e {@link SyncUser#identity}) available in this Realm. Attach a change listener
+//     * using {@link RealmResults#addChangeListener(RealmChangeListener)} to be notified about
+//     * any future changes.
+//     *
+//     * @return all roles available in the current Realm.
+//     */
+//    @Beta
+//    @ObjectServer
+//    abstract public RealmResults<String> getPermissionUsers();
```
Zhuinden: `RealmResults<String>`? Subtable queries are coming? 😮

realm/realm-library/src/main/java/io/realm/sync/permissions/Role.java
```diff
@@ -59,12 +60,35 @@ public String getName() {
...
+                Role role = realm.where(Role.class).equalTo("name", name).findFirst();
+                if (user != null && role != null) {
+                    role.members.remove(user);
+                }
+            }
+        });
+    }
+
+
+    // Relatively small list, it should be ok to query from main
```
Zhuinden: Memeber

realm/realm-library/src/objectServer/java/io/realm/internal/sync/permissions/ObjectPermissionsModule.java
```diff
@@ -4,15 +4,15 @@
...
 import io.realm.sync.permissions.RealmPermissions;
+import io.realm.sync.permissions.PermissionUser;
 import io.realm.sync.permissions.Role;
-import io.realm.sync.permissions.User;
 
 @RealmModule(library = true, classes = {
         ClassPermissions.class,
         RealmPermission.class,
         RealmPermissions.class,
         Role.class,
```
Zhuinden: Makes me wonder why Role is still `Role` if `User` is now `PermissionUser` 😛

realm/realm-library/src/main/java/io/realm/BaseRealm.java
cmelchior: Not yet unfortunately.

realm/realm-library/src/main/java/io/realm/sync/permissions/Role.java
```diff
@@ -60,35 +60,40 @@ public String getName() {
         return name;
     }
 
+    /**
+     * Adds a member to this Role. Must be done from within a write transaction.
```
Zhuinden: :+1:

nhachicha:
@cmelchior last commit has an issue with `:realm-library:kaptGenerateStubsObjectServerDebugAndroidTestKotlin` task, it fails with
```
Caused by: java.lang.NullPointerException: Class RealmPermission was not found
        at io.realm.processor.ClassCollection.getClassFromSimpleName(ClassCollection.java:50)
        at io.realm.processor.RealmProxyClassGenerator.emitCreateExpectedObjectSchemaInfo(RealmProxyClassGenerator.java:755)
        at io.realm.processor.RealmProxyClassGenerator.generate(RealmProxyClassGenerator.java:147)
        at io.realm.processor.RealmProcessor.createProxyClassFiles(RealmProcessor.java:247)
        at io.realm.processor.RealmProcessor.process(RealmProcessor.java:169)
        at com.sun.tools.javac.processing.JavacProcessingEnvironment.callProcessor(JavacProcessingEnvironment.java:794)
```
can it be realted to the new internal naming change?

cmelchior:
Something doesn't look right at least. Because the internal name there should be `__Permission`. I suspect this is fixed with https://github.com/realm/realm-java/pull/5737

realm/realm-annotations-processor/src/test/resources/io/realm/some_test_AllTypesRealmProxy.java
```diff
@@ -868,6 +868,10 @@ public static String getSimpleClassName() {
         return "AllTypes";
     }
 
+    public static final class ClassNameHelper {
```
nhachicha: why using an inner class can't you store it/access using a static field only instead?

realm/realm-library/src/main/java/io/realm/internal/ObjectServerFacade.java
```diff
@@ -119,4 +120,9 @@ public boolean wasDownloadInterrupted(Throwable throwable) {
     public boolean isPartialRealm(RealmConfiguration configuration) {
         return false;
     }
+
+    public void addSupportForObjectLevelPermissions(RealmConfiguration.Builder builder) {
+        builder.addModule(new ObjectPermissionsModule());
+        // Do nothing
```
nhachicha: remove

cmelchior:
Ready for review. Tests parse locally.
A lof changes in the one, sorry :(, but most are test related. Also a lof the Javadoc for `canRead/canCreate/etc.` are copied around so should be safe to only review in one place.

realm/realm-library/src/main/java/io/realm/sync/permissions/Permission.java
```diff
@@ -0,0 +1,562 @@
...
+         *
+         * <ol>
+         *     <li>
+         *         <b>Realm:</b>
+         *         The role is allowed to read all objects from the Realm. If {@code false}, the
+         *         Realm will appear completely empty to the role, effectively making it inaccessible.
+         *     </li>
+         *     <li>
+         *         <b>Class:</b>
+         *         The role is allowed to read the objects of this type and all referenced objects.
```
simonask: "All referenced objects" are visible even if `canRead` is false for the class of those objects.

realm/realm-library/src/main/java/io/realm/sync/permissions/Permission.java
```diff
@@ -0,0 +1,562 @@
...
+            return this;
+        }
+
+        /**
+         * Defines if this role can update the given resource or not.
+         *
+         * <ol>
+         *     <li>
+         *         <b>Realm:</b>
+         *         Not applicable.
```
simonask: `canUpdate` is very applicable to the Realm. It determines whether any writes will be accepted at all!

realm/realm-library/src/main/java/io/realm/sync/permissions/Permission.java
```diff
@@ -0,0 +1,562 @@
...
+         *     </li>
+         *     <li>
+         *         <b>Class:</b>
+         *         Not applicable.
```
simonask: Also very applicable to the class. It determines whether any object is modifiable in that class!

realm/realm-library/src/main/java/io/realm/sync/permissions/Permission.java
```diff
@@ -0,0 +1,562 @@
...
+         * Defines if this role can delete the given resource or not.
+         *
+         * <ol>
+         *     <li>
+         *         <b>Realm:</b>
+         *         Not applicable.
+         *     </li>
+         *     <li>
+         *         <b>Class:</b>
+         *         Not applicable.
```
simonask: Theoretically applicable, since it determines whether the user can remove the class from the Realm. But in practice not applicable because destructive schema changes are disallowed.

realm/realm-library/src/main/java/io/realm/sync/permissions/Permission.java
```diff
@@ -0,0 +1,562 @@
...
+            return this;
+        }
+
+        /**
+         * Defines if this role can delete the given resource or not.
+         *
+         * <ol>
+         *     <li>
+         *         <b>Realm:</b>
+         *         Not applicable.
```
simonask: In the future, it may denote whether the user can delete the Realm file on the server.

realm/realm-library/src/main/java/io/realm/sync/permissions/Permission.java
```diff
@@ -0,0 +1,562 @@
...
+         *     </li>
+         *     <li>
+         *         <b>Class:</b>
+         *         Not applicable.
+         *     </li>
+         *     <li>
+         *         <b>Object:</b>
+         *         If {@code true}, the role is allowed to update properties on the object, but not
+         *         set or change permissions.
```
simonask: This is wrong. If true, the user can delete the object.

realm/realm-library/src/main/java/io/realm/sync/permissions/Permission.java
```diff
@@ -0,0 +1,562 @@
...
+         *
+         * @param canDelete {@code true} if the role is allowed to delete this resource, {@code false} if not.
+         */
+        public Builder canDelete(boolean canDelete) {
+            this.canDelete = canDelete;
+            return this;
+        }
+
+        /**
+         * Defines if this role is allowed to change permissions on the given resource.
```
simonask: Maybe add a note about the fact that users can only give permissions to others at or below their own level of permissions.

realm/realm-library/src/main/java/io/realm/sync/permissions/Permission.java
```diff
@@ -0,0 +1,562 @@
...
+        return canRead;
+    }
+
+    /**
+     * Defines if this role can read from given resource or not.
+     *
+     * <ol>
+     *     <li>
+     *         <b>Realm:</b>
+     *         The role is allowed to read all objects from the Realm. If {@code false}, the
```
simonask: Not "all objects", but any object at all. If `canRead` is false, nothing is visible. If it is true, everything in the Realm may still happen to be inaccessible to that user.

`canRead` doesn't determine whether you can read everything, but determines whether you can read anything at all.

realm/realm-library/src/main/java/io/realm/sync/permissions/RealmPermissions.java
```diff
@@ -0,0 +1,52 @@
...
+    private int id = 0; // Singleton object for the Realm file
+    private RealmList<Permission> permissions = new RealmList<>();
+
+    public RealmPermissions() {
+        // Required by Realm
+    }
+
+    /**
+     * Returns all Realm level permissions. This is the default set of permissions if not
+     * otherwise defined by {@link ClassPermissions} and Object level permissions.
```
simonask: No, Realm-level permissions are not applied as default permissions in any sense.

realm/realm-library/src/main/java/io/realm/sync/permissions/RealmPrivileges.java
```diff
@@ -0,0 +1,271 @@
...
+        this.canRead = (privileges & (1 << 0)) != 0;
+        this.canUpdate = (privileges & (1 << 1)) != 0;
+        this.canDelete = (privileges & (1 << 2)) != 0;
+        this.canSetPermissions = (privileges & (1 << 3)) != 0;
+        this.canQuery = (privileges & (1 << 4)) != 0;
+        this.canCreate = (privileges & (1 << 5)) != 0;
+        this.canModifySchema = (privileges & (1 << 6)) != 0;
+    }
+
+    /**
```
simonask: Doc comments in this file seem copy-pasted — I won't repeat the review feedback here, but the same comments apply. :)

realm/realm-library/src/syncIntegrationTest/java/io/realm/PermissionManagerTests.java
```diff
@@ -54,6 +55,7 @@
 import static org.junit.Assert.fail;
 
 @RunWith(AndroidJUnit4.class)
+@Ignore // FIXME: Temporary disable unit tests due to lates (3.0.0-alpha.2) ROS having issues
```
simonask: Probably shouldn't be committed, I'm guessing?

realm/realm-library/src/main/java/io/realm/sync/permissions/Permission.java
```diff
@@ -0,0 +1,562 @@
...
+import io.realm.annotations.RealmClass;
+import io.realm.internal.annotations.ObjectServer;
+
+/**
+ * This class encapsulates the privileges granted a given {@link Role}. These privileges can be
+ * applied to either the entire Realm, Classes or individual objects.
+ * <p>
+ * If no privileges are defined for an individual object, the values for {@link ClassPermissions}
+ * will be inherited, if no values are defined there, the ones from {@link RealmPermissions} will
+ * be used. If no values can be found there, no privileges are granted.
```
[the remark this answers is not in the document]
cmelchior: How would you describe it then? I assume the issue is that the high-level privileges might map to other low-level privileges? But not sure how to describe that accurately

realm/realm-library/src/main/java/io/realm/sync/permissions/Permission.java
cmelchior: The cheat sheet said `N/A`. Can you check if it is correct now: https://docs.google.com/document/d/1E5W3sQMfWa_wjvG2u43IDsBpeTbhi3g_wrIV9eCKRjc/edit#

realm/realm-library/src/main/java/io/realm/sync/permissions/Permission.java
cmelchior: I'll rather leave it as N/A for now. No reason to complicate things with possible future enhancements.

realm/realm-library/src/main/java/io/realm/sync/permissions/Permission.java
cmelchior: Same as above.

realm/realm-library/src/syncIntegrationTest/java/io/realm/PermissionManagerTests.java
cmelchior: Actually, this is intentional to unblock CI, but we probably shouldn't release anything until we determined what is causing this.

realm/realm-library/src/main/java/io/realm/sync/permissions/Permission.java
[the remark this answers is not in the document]
simonask: They don't commute at all. You can't change objects in any way (Update, Delete, SetPermissions, etc.) in a Realm that you don't have Update permissions to change. The individual privilege type does not "trickle down". This would quickly create some nonsensical situations that would make things hard to model.

realm/realm-library/src/main/java/io/realm/sync/permissions/Permission.java
simonask: I have updated the cheat sheet. :)

nhachicha:
@cmelchior I think we still have two issues with the custom naming feature

1- Currently adding a module that includes `Permission` class will break the annotation processor 
```Java
@RealmModule(classes = {PermissionObject.class, Permission.class})
public class OLPermissionModule {
}

SyncConfiguration syncConfig = configurationFactory
                        .createSyncConfigurationBuilder(user_0, Constants.GLOBAL_REALM)
                        .modules(new OLPermissionModule())
                        .partialRealm().build();
         
```

will break with 

```
error: Permission could not be added to the module. Only classes extending RealmObject or implementing RealmModel, which are part of this project, can be added.

```

2- Similarly using `schema` to specify the classes you want to include as part of the configuration will also break

```Java
.schema(Permission.class, PermissionObject.class);// I made schema public for the sake of the example
```
will fail with 
```
02-10 17:48:00.297 29703 29771 I TestRunner: failed: syncPartialPermissionSet(io.realm.objectserver.ObjectLevelPermissionIntegrationTests)
02-10 17:48:00.297 29703 29771 I TestRunner: ----- begin exception -----
02-10 17:48:00.299 29703 29771 I TestRunner: java.lang.IllegalStateException: Schema validation failed due to the following errors:
02-10 17:48:00.299 29703 29771 I TestRunner: - Property 'PermissionObject.permissions' of type 'array' has unknown object type '__Permission'
02-10 17:48:00.299 29703 29771 I TestRunner: 	at io.realm.internal.OsSharedRealm.nativeGetSharedRealm(Native Method)
02-10 17:48:00.299 29703 29771 I TestRunner: 	at io.realm.internal.OsSharedRealm.<init>(OsSharedRealm.java:171)

```

nhachicha:
@cmelchior is the code in OS tracking this branch `tg/object-level-permissions`? it has a slightly different logic for `create_default_permissions` compared to the current commit.

Zhuinden:
@nhachicha there is a good chance that if `Permission` class comes in from `realm-android` library then Realm-Android must expose `Permission` class with a `@RealmModule(library = true, allClasses = false)`

cmelchior:
@nhachicha Are you making that module in the unit tests or in the sample project. If you are doing it in the sample project it will not work as `schema()` is just filtering the default module, so does not know `Permission` which comes from the library. If that is the case it would look like this: `modules(new ObjectPermissionsModule(), Realm.getDefaultModule())`

The OS branch in this PR is probably a little behind. I'll update it.

nhachicha:
@cmelchior @Zhuinden this module is intended to be used within the `syncIntegrationTest`, 
`Permission` should be part of the default module and thus filterable with `schema`

realm/realm-library/src/main/java/io/realm/internal/OsCollectionChangeSet.java
```diff
@@ -140,7 +151,9 @@ public boolean isFirstAsyncCallback() {
      * Returns {@code true} if this changeset is empty, and doesn't contain any relevant changes.
      */
     public boolean isEmpty() {
-        return nativeIsEmpty(nativePtr);
+        // Since this wrap a Object Store changeset, it will always contains changes if an
+        // Object Store changeset exists.
+        return nativePtr == 0;
```
Zhuinden: Is it guaranteed that object store change set cannot be `delete`d without us knowing?

cmelchior: Yes, because we create a copy with the changeset here: https://github.com/realm/realm-java/blob/master/realm/realm-library/src/main/cpp/observable_collection_wrapper.hpp#L89

And that is only deleted when the finalizer is called.

PR #5737, opened 2018-02-02
Correctly support multiple classes with the same name

cmelchior:
When #5280 was merged, it was stipulated that it would mean having two model classes with the same name would be supported.

It turned out this was not fully tested and thus didn't work. In particular we had problems with multiple proxy classes being created with the same name.

This was fixed by changing the naming scheme for proxy classes, so they now include the package name, i.e. from `PersonRealmProxy` to `some_package_PersonRealmProxy`. We cannot use the internal name as that might be an illegal Java class like "!Person". After this changed it also turned out that backlinks didn't work correctly in the case of conflicting classes. This was fixed as well.   

So in summary, this PR consists of:

1) Proxy classes renamed from `<SimpleName>RealmProxy` to <my_package_SimpleName>RealmProxy`. Proxy interfaces were changed in a similar way. 

2) Referencing classes by simple names can lead to subtle bugs, so removed that functionality from the Realm annotation processor where I could.

3) Backlinks use the terms Target and Source, where Source is the object linking normally to Target, so the backlink goes from Target to Source. In some places this terminology was mixed. This has been cleaned up as well.

4) No CHANGELOG entry as this is just fixing bugs in unreleased functionality.


Zhuinden:
Hey, this resolves the limitation in the docs

> It is not possible to have two model classes with the same name in different packages.

That's useful! :+1: :confetti_ball:

cmelchior:
Yes, but it only works if you provide a custom name using `@RealmClass(name = "nonConflictingName") in one of the classes. The internal names are still not allowed to conflict.

Zhuinden:
@cmelchior technically I was thinking about the use-case of having a `v1` folder and a `v2` folder for RealmObjects, in two different Realm files.

cmelchior:
Yes, that should also work. You probably want to separate them into modules in that case, so you don't have to think about changing the underlying name, like:

```
@RealmModule(classes = { my.package.v1.User.class })
public class ApiV1 {} 

@RealmModule(classes = { my.package.v2.User.class })
public class ApiV1 {} 
```

Zhuinden:
Yep, that's exactly what I had in mind! :slightly_smiling_face: :

realm/realm-library/src/androidTest/java/io/realm/entities/conflict/AllJavaTypes.java
```diff
@@ -0,0 +1,26 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.realm.entities.conflict;
+
+import io.realm.RealmObject;
+import io.realm.annotations.RealmClass;
+
+// Potential conflict with `io.realm.entities.AllJavaTypes` but proxy classes should be generated
+// using the internal name (Which do not conflict)
```
nhachicha: is there a test for this?

cmelchior: Yes, both the annotation processor and the test suite now have classes that conflict with the Simple name, but not the internal one.

Issue #5758, opened 2018-02-21
clearTable is not supported on Partial Sync


cmelchior:
`table.clear()` is not supported for Partial Sync, so we should change any internal implementation that uses this. 

APIs:

* `Realm.delete(Class)` -> Replace with `findAll().deleteAllFromRealm()`. 

For efficiency we should probably only do this if Partial Sync is enabled as deleting the query result is slower.

cmelchior:
Fixed

PR #5759, opened 2018-02-21
Refactor Partial Sync to use Object Store master

cmelchior:
This PR refactors the Partial Sync support so it uses the API available on OS/master:

* Upgrade to Sync 3.0.0-beta.6
* Add a wrapper for Subscription class in Object Store
* Add support for `will_send_notifications/did_send_notifications` from Object Sdtore
* Combine Subscription/Results notifications in Java after they are done sending.
* Some of these classes, unfortunately, have to live in the `main` flavour for the time being.


realm/realm-library/src/main/cpp/observable_collection_wrapper.hpp
```diff
@@ -53,57 +52,48 @@ class ObservableCollectionWrapper {
...
-        std::string error_message = "";
         if (err) {
             try {
                 std::rethrow_exception(err);
             }
-            catch (const std::exception &e) {
-                error_message = e.what();
+            catch (const std::exception& e) {
+                realm::jni_util::Log::e("Caught exception in collection change callback %1", e.what());
+                return;
```
Zhuinden: Can exceptions really be swallowed here?

cmelchior:
@nhachicha  Ready for review. CI is currently having issues with the test devices, but all relevant tests parse locally on my machine.

The only test failing is currently:

```
io.realm.ManagedRealmListForValueTests > changeListener_forDeleteAll[0: Type: STRING_LIST, Nullable?: true][Custom Phone - 7.1.0 - API 25 - 768x1280_1 - 7.1.1] FAILED 
Test failed to run to completion. Reason: 'Instrumentation run failed due to 'Process crashed.''. Check device logcat for details
Tests on Custom Phone - 7.1.0 - API 25 - 768x1280_1 - 7.1.1 failed: Instrumentation run failed due to 'Process crashed.'
```

Which might be related to https://github.com/realm/realm-java/issues/5758, but I'm looking into it now

realm/realm-library/src/main/cpp/io_realm_internal_sync_OsSubscription.cpp
```diff
@@ -0,0 +1,120 @@
...
+
+JNIEXPORT jobject JNICALL Java_io_realm_internal_sync_OsSubscription_nativeGetError(JNIEnv* env, jclass, jlong native_ptr)
+{
+    TR_ENTER()
+    try {
+        auto wrapper = reinterpret_cast<SubscriptionWrapper*>(native_ptr);
+        auto err = wrapper->subscription().error();
+        if (err) {
+            std::string error_message = "";
+            if (err) {
```
nhachicha: why the double check? this is already `true`/defined?

realm/realm-library/src/main/cpp/io_realm_internal_sync_OsSubscription.cpp
```diff
@@ -0,0 +1,120 @@
...
+                try {
+                    std::rethrow_exception(err);
+                }
+                catch (const std::exception &e) {
+                    error_message = e.what();
+                }
+            }
+
+            static JavaClass illegal_argument_class(env, "java/lang/IllegalArgumentException");
```
nhachicha: Why `IllegalArgumentException`? are you assuming the error is related to the subscription parameters? we deprecated the dynamic query for partial sync AFAIK, so this exception cannot be something the user did wrong (besides the naming perhaps)

realm/realm-library/src/main/cpp/io_realm_internal_sync_OsSubscription.cpp
```diff
@@ -0,0 +1,120 @@
...
+            static JavaMethod illegal_argument_constructor(env, illegal_argument_class, "<init>", "(Ljava/lang/String;)V");
+            return (jthrowable) env->NewObject(illegal_argument_class, illegal_argument_constructor, to_jstring(env, error_message));
```
nhachicha: should you check for a pending Java exception before throwing? ` if (env->ExceptionCheck()){return;}`

realm/realm-library/src/main/cpp/subscription_wrapper.hpp
```diff
@@ -0,0 +1,93 @@
...
+    static jni_util::JavaClass os_results_class(env, "io/realm/internal/sync/OsSubscription");
+    static jni_util::JavaMethod notify_change_listeners(env, os_results_class, "notifyChangeListeners", "()V");
+
+    if (!m_subscription_weak_ref) {
+        m_subscription_weak_ref = jni_util::JavaGlobalWeakRef(env, j_subscription_object);
+    }
+
+    auto cb = [=]() {
+        // OS will call all notifiers' callback in one run, so check the Java exception first!!
+        if (env->ExceptionCheck())
```
nhachicha: is this callback invoked from the caller thread? if not be careful when capturing in the lambda the `env` by copy `[=]`

realm/realm-library/src/main/java/io/realm/internal/EmptyLoadChangeSet.java
```diff
@@ -15,19 +15,22 @@
...
 package io.realm.internal;
 
+import javax.annotation.Nullable;
+
 import io.realm.RealmResults;
+import io.realm.internal.sync.OsSubscription;
 
 /**
- * Fake changeset used if {@link RealmResults#load()} is called manually.
+ * Fake changeset used if {@link RealmResults#load()} is called manually or if no collection
```
nhachicha: I prefer _empty_

realm/realm-library/src/main/java/io/realm/internal/OsCollectionChangeSet.java
```diff
@@ -117,15 +126,14 @@ public boolean isCompleteResult() {
...
-    }
-
-    public int getOldStatusCode() {
-        return nativeGetOldStatusCode(nativePtr);
-    }
-
-    public int getNewStatusCode() {
-        return nativeGetNewStatusCode(nativePtr);
+        // If no subscription status exist, it means that either the Realm is not a partial Realm
+        // or the the query result is a sub query of some kind, which means all data already
```
nhachicha: remove _the_

realm/realm-library/src/main/java/io/realm/internal/SubscriptionAwareOsResults.java
```diff
@@ -0,0 +1,109 @@
...
+import io.realm.internal.sync.OsSubscription;
+
+/**
+ * Wrapper around Object Stores Results class that is capable of combining partial sync Subscription
+ * state updates and collection change updates.
+ */
+public class SubscriptionAwareOsResults extends OsResults {
+
+    private final String subscriptionName;
+    // The native ptr to a delayed notification. Since Java group all notifications for each each
```
nhachicha: remove _each_

realm/realm-library/src/main/java/io/realm/internal/SubscriptionAwareOsResults.java
```diff
@@ -0,0 +1,109 @@
...
+                subscriptionChanged = true;
+            }
+        });
+        RealmNotifier notifier = sharedRealm.realmNotifier;
+        notifier.addBeginSendingNotificationsCallback(new Runnable() {
+            @Override
+            public void run() {
+                subscriptionChanged = false;
+                collectionChanged = false;
+                delayedNotificationPtr = 0;
```
nhachicha: this can't have a previous pointer address? otherwise, you're leaking the dynamic allocation?

realm/realm-library/src/main/cpp/io_realm_internal_sync_OsSubscription.cpp
```diff
@@ -0,0 +1,120 @@
+/*
+ * Copyright 2016 Realm Inc.
```
kneth: 2018

realm/realm-library/src/main/cpp/io_realm_internal_sync_OsSubscription.cpp
kneth: C style type casting?

realm/realm-library/src/main/cpp/observable_collection_wrapper.hpp
kneth: Exceptions in callbacks are really hard to debug since the stack trace will not make much sense.

realm/realm-library/src/main/cpp/subscription_wrapper.hpp
```diff
@@ -0,0 +1,93 @@
...
+    partial_sync::Subscription& subscription() {
+        return m_subscription;
+    };
+
+    void start_listening(JNIEnv* env, jobject j_subscription_object);
+    void stop_listening();
+
+private:
+    jni_util::JavaGlobalWeakRef m_subscription_weak_ref;
+    partial_sync::SubscriptionNotificationToken m_notification_token;
```
kneth: So only one listener per subscription?

realm/realm-library/src/main/cpp/io_realm_internal_sync_OsSubscription.cpp
cmelchior: Some queries are still disallowed, like backlink queries, but the most common error is probably re-using a subscription name for a different query. 

Theoretically, a few very rare exceptions might happen here that are not considered `IllegalArgument`,  but I don't even know where to begin to filter for those, so I just left everything as `IllegalArgument` instead of the generic `RealmException`. I think that is a sensible trade-off?

realm/realm-library/src/main/cpp/io_realm_internal_sync_OsSubscription.cpp
cmelchior: I don't think that will be needed since this code isn't triggered by a callback, but from a direct call from Java, so if a Java exception existed the program would already have crashed.

realm/realm-library/src/main/cpp/subscription_wrapper.hpp
cmelchior: Yes, it should be as this is just a specialized notifier callback, we use the same approach in `observable_collection_wrapper.hpp`

realm/realm-library/src/main/java/io/realm/internal/SubscriptionAwareOsResults.java
cmelchior: It can, but if `!= 0` it will be sent to an `OsCollectionChangeSet` which will clean it up when no longer needed.

realm/realm-library/src/main/cpp/subscription_wrapper.hpp
cmelchior: Yes, you can register multiple listeners in Java, but only one is ever created in JNI. `Results` observables work the same way.

realm/realm-library/src/syncIntegrationTest/java/io/realm/PermissionManagerTests.java
```diff
@@ -54,6 +54,7 @@
 import static org.junit.Assert.fail;
 
 @RunWith(AndroidJUnit4.class)
+@Ignore
```
nhachicha: add a FIXME to re-enable when ROS permissions are stable again

PR #5763, opened 2018-02-23
Update arch example

Zhuinden:
This PR updates the branch `my/arch_example` in the PR https://github.com/realm/realm-java/pull/4685 .

Adds support for:

- `LiveResults extends LiveData<List<T>>` wrapping a RealmResults

- a list fragment and a detail fragment

- makes the example prettier and stuff

I initially tried to push back to `realm/realm-java` and I don't have permission for that, so I forked and added changes with `git add` / `git commit` instead of through the IDE, but I hope I got everything in here.

I promised I'd do this about 18 days ago, finally got some spare time to do it.


Zhuinden:
@cmelchior btw I think I've never signed the CLA, so that might be a required thingy

cmelchior:
Hey @Zhuinden Thanks a ton for this  🎉. I'll see if I can manage to review this today. Otherwise, it will be on Monday.

examples/architectureComponentsExample/src/main/java/io/realm/examples/arch/ArchExampleActivity.java
```diff
@@ -0,0 +1,90 @@
...
+
+    @Override
+    protected void onCreate(Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+
+        setContentView(R.layout.activity_arch_example);
+        setupViews();
+
+        backgroundTask = (BackgroundTask) getLastCustomNonConfigurationInstance();
+        if (backgroundTask == null) { // this could also live inside a ViewModel, a singleton job queue, etc.
```
cmelchior: I managed to get into a state where two background updater threads were running at the same time. I'm not 100% how it happened. 

But doing:

1) Click details for Christian
2) Press back
3) Press back (exit app)
4) Click app (opens it again)
5) Increments now happen at double speed.

The thread isn't stopped in `onBackPressed` so my guess is that it keeps running somehow.

examples/architectureComponentsExample/src/main/java/io/realm/examples/arch/BackgroundTask.java
```diff
@@ -16,66 +16,73 @@
...
 
     private volatile Thread thread;
 
+    @MainThread
+    public boolean isStarted() {
+        return isStarted;
+    }
+
     @MainThread
     public void start() {
```
cmelchior: All of these methods depend on `start` only being called from a single thread. It would be more clear and safer if start/stop was synchronized.

examples/architectureComponentsExample/src/main/java/io/realm/examples/arch/ArchExampleActivity.java
```diff
@@ -0,0 +1,90 @@
...
+            super.onBackPressed();
+        }
+    }
+
+    @MainThread
+    private void setupViews() {
+        backgroundJobStartStop = findViewById(R.id.backgroundJobStartStop);
+        backgroundJobStartStop.setOnClickListener(new View.OnClickListener() {
+            @Override
+            public void onClick(View v) {
```
cmelchior: We support Java 8 in the examples, so feel free to use lamdas.

examples/architectureComponentsExample/src/main/java/io/realm/examples/arch/helper/Helper.java
```diff
@@ -0,0 +1,40 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.realm.examples.arch.helper;
+
+import android.app.Activity;
+import android.content.Context;
+import android.content.ContextWrapper;
+
+public class Helper {
```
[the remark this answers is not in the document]
Zhuinden: I can rename it to `ContextUtils`.

examples/architectureComponentsExample/src/main/java/io/realm/examples/arch/livemodel/LiveRealmObject.java
```diff
@@ -26,27 +27,44 @@
 
 
 public class LiveRealmObject<T extends RealmModel> extends LiveData<T> {
```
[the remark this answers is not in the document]
Zhuinden: Good point! :smile: 

And I even added some comments in other places, too.. :roll_eyes:

examples/architectureComponentsExample/src/main/java/io/realm/examples/arch/livemodel/LiveRealmObject.java
```diff
@@ -26,27 +27,44 @@
...
+            RealmObject.addChangeListener(object, listener);
+        }
     }
 
     @Override
     protected void onInactive() {
         super.onInactive();
-        RealmObject.removeChangeListener(getValue(), listener);
+        T object = getValue();
+        if (object != null && RealmObject.isValid(object)) {
```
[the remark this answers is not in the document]
Zhuinden: `onInactive` is bound to observing the LiveData, and can contain null if the object was deleted. Also, object could have its Realm instance invalidated if we're unlucky.

A listener should not be removed from an invalid object, because invalid objects throw when accessed even if you try to remove listeners from them, if I recall correctly.

The intent here is that this can only be null if the object was deleted, which mimics the behavior of `findFirst().addChangeListener` (calls with `isDeleted == true` change set once more, and then never again).

Now that you mention it, I should probably add the valid check to the RealmResults' live data too, its Realm instance can be closed, after all.

examples/architectureComponentsExample/src/main/java/io/realm/examples/arch/ArchExampleActivity.java
Zhuinden: Gah, I didn't notice this! Thanks!

Zhuinden:
@cmelchior comments addressed

Zhuinden:
@cmelchior comments addressed

Zhuinden:
Hmm, I wonder if I should go through some of the other examples too later 😄

cmelchior:
You will not hear any complaints from us 😉 
Nabil promised to review this either today or early next week, so we have not forgotten this PR.

examples/architectureComponentsExample/src/main/java/io/realm/examples/arch/CustomApplication.java
```diff
@@ -0,0 +1,58 @@
...
+                        person.setName("Christian Melchior");
+                        person.setAge(34);
+
+                        person = realm.createObject(Person.class);
+                        person.setName("Chen Mulong");
+                        person.setAge(28);
+
+                        person = realm.createObject(Person.class);
+                        person.setName("Nabil Hachicha");
+                        person.setAge(29);
```
nhachicha: I wish 👴🏻

examples/architectureComponentsExample/src/main/java/io/realm/examples/arch/BackgroundTask.java
```diff
@@ -0,0 +1,92 @@
...
+                final RealmResults<Person> persons = realm.where(Person.class).findAll();
+                Realm.Transaction transaction = (Realm r) -> {
+                    for (Person person : persons) {
+                        person.setAge(person.getAge() + 1); // updates the Persons in the Realm.
+                    }
+                };
+
+                while (!isInterrupted()) {
+                    realm.executeTransaction(transaction);
+                    Thread.sleep(1000L);
```
nhachicha: Minor: You can sleep with `SystemClock.sleep(1000);` without the need of the try-catch

examples/architectureComponentsExample/src/main/java/io/realm/examples/arch/PersonFragment.java
```diff
@@ -0,0 +1,85 @@
...
+
+    public static PersonFragment create(String personName) {
+        PersonFragment personFragment = new PersonFragment();
+        Bundle bundle = new Bundle();
+        bundle.putString(ARG_PERSON_NAME, personName);
+        personFragment.setArguments(bundle);
+        return personFragment;
+    }
+
+    TextView name;
```
nhachicha: private? same for below

Zhuinden:
@nhachicha PR comments addressed

PR #5764, opened 2018-02-23
Support types in other compilation units

cmelchior:
With the support for changing the internal names, we can no longer infer the name from the reference type itself. This turned out to cause problems if the referenced type was in another compilation unit. This PR splits out the fix from https://github.com/realm/realm-java/pull/5729 as the functionality was originally merged to master (but has not yet been released).

Not Changelog as the functionality has not yet been released.

The functionality is being tested in the `moduleExample` where the app now has a reference to a library module class.


cmelchior:
Ready for review @nhachicha

PR #5781, opened 2018-02-28
Support clearing data using Partial Sync

cmelchior:
Fixes https://github.com/realm/realm-java/issues/5758

Partial Sync does not support the `clearTable` instruction. This fixes it. Due to performance concerns, this code path is only triggered for partially synchronized Realms.

### DIFF
--- a/realm/realm-library/src/main/java/io/realm/internal/Table.java
+++ b/realm/realm-library/src/main/java/io/realm/internal/Table.java
@@ -261,10 +261,13 @@ public class Table implements NativeObject {
 
     /**
      * Clears the table i.e., deleting all rows in the table.
+     *
+     * WARNING: This method is not supported by Partial Sync, use `findAll().deleteFromRealm`
+     * instead.
      */
-    public void clear() {
+    public void clear(boolean partialRealm) {
         checkImmutable();
-        nativeClear(nativePtr);
+        nativeClear(nativePtr, partialRealm);
     }
 
     // Column Information.
@@ -724,7 +727,7 @@ public class Table implements NativeObject {
 
     private native long nativeSize(long nativeTablePtr);
 
-    private native void nativeClear(long nativeTablePtr);
+    private native void nativeClear(long nativeTablePtr, boolean partialRealm);
 
     private native long nativeGetColumnCount(long nativeTablePtr);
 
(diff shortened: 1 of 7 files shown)


realm/realm-library/src/main/java/io/realm/internal/Table.java
```diff
@@ -261,10 +261,13 @@ public boolean isEmpty() {
 
     /**
      * Clears the table i.e., deleting all rows in the table.
+     *
+     * WARNING: This method is not supported by Partial Sync, use `findAll().deleteFromRealm`
```
Zhuinden: It kind of is supported now, no? With magic 😄

cmelchior: The server isn't enforcing it right now. I believe it just ignores the instruction.

Zhuinden: No I'm asking because you seem to have just added the means to clear via `isPartial == true` in JNI side

cmelchior: There is a subtle difference `table->where().find_all().clear()` creates a query result and delete the individual objects in it. Resulting in a lot of `delete_row` instructions. Which is also why it is slower.

`table->clear()` just results in one `clearTable` instruction, which isn't supported by partial sync right now (mostly because "the whole table" is a loose concept in a world with partial sync.

nhachicha: I think  @Zhuinden point was that the comment doesn't makes sense since you implemented internally a fallback to `table->where().find_all().clear()`

I think it should rephrase it to something like 
> WARNING: In the context of partial sync, this method will behave similarly to a `findAll().deleteFromRealm()` which might have a performance impact.

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/PartialSyncTests.java
```diff
@@ -274,6 +269,30 @@ public void onError(String subscriptionName, Throwable error) {
...
     }
 
+    @Test
+    @RunTestInLooperThread
+    public void clearTable() throws InterruptedException {
+        SyncUser user = UserFactory.createUniqueUser(Constants.AUTH_URL);
+        Realm realm = getPartialRealm(user);
+        looperThread.closeAfterTest(realm);
+
+        // Create test data and make it is uploaded to the server
```
nhachicha: typo? 

Create test data and wait for it to be uploaded to the server

PR #5791, opened 2018-03-02
fix wait_for_download behaviour with permission schema

nhachicha:
After the fix in [Sync](https://github.com/realm/realm-sync/pull/2051) we should no longer need the hack that registers a subscription and wait for `Update` to be triggered in order to start locking the schema. 
It should only require an `asyncOpen` with a `waitForInitialData` for the Realm to synchronise the permission schema from Sync (the one that includes the correction without duplicates), we can then start modifying it to lock the schema for instance.

However, this was not working in Android as opposed to iOS. Basically we were doing a `waitForDownload` on an empty schema, this causes the local Realm to not upload the local Realm with the default permissions populated from the ObjectStore to Sync which will apply the permissions corrections. (i.e `waitForInitialData` was useless after at this level.)

This PR, moved waitForInitialData to be performed on the Realm opened with the full schema.

Since this is the `next-major` I removed the workarounds we had on different old versions 0.8x & 2.x (PK migration, index etc ..)


bdash:
Isn't it necessary to wait for the initial download with an empty schema in order to support opening synced Realms that the user has read-only access to?

Zhuinden:
Does that mean that if you follow one of those silly tutorials from the internet that tells you to use 0.82.2 then Realm no longer automatically upgrades the Realm to a compatible format from 5.0.0+?

cmelchior:
@Zhuinden Yes, but most of the bugs where only relevant if you attempted to read a Realm file created by Cocoa, so I doubt it will affect anyone.

Zhuinden:
Ah, that's super-rare at this point, you are right.

nhachicha:
This will be revisited once https://github.com/realm/realm-java/issues/5812 is implemented to solve the Readonly use case highlighted in https://github.com/realm/realm-java/pull/5791#issuecomment-370124297

bmunkholm:
@nhachicha I imagine this can be closed now?

PR #5796, opened 2018-03-04
Better exception message if non model class is provided

### LINKED ISSUE #5779: Better error when calling where with RealmModel
In Kotlin I sometimes end up in a situation where a type that I pass to `where` is treated as a `RealmModel` instead of a `RealmModel` subclass, but it's not always obvious from looking at the code. The problem is `where` eventually calls `Util.getOriginalModelClass` which get a superclass of null, causing a runtime NPE exception (`superclass.equals` because superclass is null).

It would be nice if there was a more informative message so that I could tell what was happening right away.

cmelchior:
Fixes #5779


realm/realm-library/src/main/java/io/realm/internal/Util.java
```diff
@@ -47,6 +47,14 @@ public static String getTablePrefix() {
      * was a RealmProxy class.
      */
     public static Class<? extends RealmModel> getOriginalModelClass(Class<? extends RealmModel> clazz) {
+
+        // The compiler would allow these classes to be parsed as arguments, but they are never
```
nhachicha: to be passed

realm/realm-library/src/main/java/io/realm/internal/Util.java
```diff
@@ -47,6 +47,14 @@ public static String getTablePrefix() {
...
+        // valid as a Realm model class
+        if (clazz.equals(RealmModel.class) || clazz.equals(RealmObject.class)) {
+            throw new IllegalArgumentException("RealmModel or RealmObject was parsed as an argument. " +
```
nhachicha: ditto

PR #5813, opened 2018-03-07
getDefaultInstance with support for Sync

### LINKED ISSUE #5806: Default Sync Realm API
As part of the upcoming release of partial sync, Realm Platform Self-Hosted and Cloud will offer the concept of a default synced Realm. This is designed to simplify how developers start working with synced data. Specifically it will fix or improve the following:

1. Simplify how developers think about their data model. Previously developers needed to split their data into multiple Realms to control data access. Now with partial sync and the new permission system this is not necessary.
2. Mirror the ease of use of Realm Database which offers a similar default Realm convenience API
3. Limit the code needed to open a synced Realm - now the user simply needs to login with just the auth URL instead of needing to worry about Realm URLs and paths
4. The default synced Realm will include the master User list of the application allowing developers to incorporate relationships to users simplifying how to think about user metadata.

API Proposal:

```
/*
  Open the default synced Realm
  - Has a dedicated API so that we can have a default configuration using:
    - SyncUser.current
    - Default Realm URL based off SyncUser auth URL (supplied within the refresh token)
    - Hardcoded path
    - Setup the Realm with a sync history type

  - If opened before there is a SyncUser.current one of two things can happen depending on the implementation:
    - The simple implementation is to throw if there is no SyncUser.current
    - The advanced implementation is to open the Realm without a sync session allowing offline access
*/

// Java
Realm realm = Realm.getDefaultSyncInstance();

// Swift
let realm = try! Realm.defaultSync()

// Objective-C
RLMRealm *realm = [RLMRealm defaultSyncRealm];

// JS
// Currently lacks a default Realm, so we should add
var realm = new Realm.defaultSync()

// .Net
var realm = Realm.GetDefeaultSyncInstance();

/*
  Login Example
*/

let credentials = SyncCredentials.nickname(nickname: "Adam", isAdmin: false)
let serverURL = NSURL(“http://testapp.us1.cloud.realm.io”)
SyncUser.logIn(with: credentials,
               server: serverURL) { user, error in
    if let user = user {
        // can now open the default synchronized Realm with this user
        let realm = try! Realm.defaultSync()
    } else if let error = error {
        // handle error
    }
}

```

cmelchior:
Fixes #5806 

This PR changes the meaning of calling `Realm.getDefaultInstance()` when sync is enabled. Instead of creating a local Realm, which is probably not what you want if sync is enabled, it infers the URL from the currently logged in user and creates a default configuration based on that.

This means that calling `Realm.getDefaultConfiguration()` can now fail in new ways and that it potentially can return different instances depending on what user is currently logged in. There should hopefully be tests documenting this behavior.

This also had some consequences for `Realm.getDefaultConfiguration()` and `Realm.removeDefaultConfiguration()`, but essentially they continue to work as before, until you call `getDefaultInstance()`. If the configuration is `null` when doing that it will create a new default sync configuration.

### DIFF
--- a/realm/realm-library/src/main/java/io/realm/Realm.java
+++ b/realm/realm-library/src/main/java/io/realm/Realm.java
@@ -149,7 +149,6 @@ public class Realm extends BaseRealm {
     private final RealmSchema schema;
     private static final Object defaultConfigurationLock = new Object();
     // guarded by `defaultConfigurationLock`
-    private static volatile boolean userClearedConfiguration = false;
     private static volatile RealmConfiguration userDefinedDefaultConfiguration;
 
     /**
@@ -332,7 +331,7 @@ public class Realm extends BaseRealm {
      * If sync is enabled, a sync configuration will be used that automatically infer the URL to the
      * default Realm on the server on which the current user is logged in.
      * <p>
-     * if sync is not enabled, a local realm called {@code default.realm} will be created in
+     * if sync is not enabled, a local Realm called {@code default.realm} will be created in
      * {@link Context#getFilesDir()}.
      *
      * @return an instance of the Realm class.
@@ -357,7 +356,6 @@ public class Realm extends BaseRealm {
                     // create it lazily instead. The Facade is responsible for caching it.
                     // This also re-creates it if `removeDefaultConfiguration` was manually called.
                     configuration = ObjectServerFacade.getSyncFacadeIfPossible().getSystemDefaultSyncConfiguration();
-                    userClearedConfiguration = false;
                 } else {
                     throw new IllegalStateException("The default configuration was manually cleared " +
                             "using 'Realm.removeDefaultConfiguration()'. Add a new default configuration " +
@@ -426,7 +424,6 @@ public class Realm extends BaseRealm {
         }
         synchronized (defaultConfigurationLock) {
             userDefinedDefaultConfiguration = configuration;
-            userClearedConfiguration = false;
         }
     }
 
@@ -455,7 +452,6 @@ public class Realm extends BaseRealm {
         synchronized (defaultConfigurationLock) {
             userDefinedDefaultConfiguration = null;
             ObjectServerFacade.getSyncFacadeIfPossible().removeCachedDefaultSyncConfiguration();
-            userClearedConfiguration = true;
         }
     }
 
--- a/realm/realm-library/src/main/java/io/realm/Realm.java
+++ b/realm/realm-library/src/main/java/io/realm/Realm.java
@@ -50,7 +50,6 @@ import javax.annotation.Nullable;
 
 import io.reactivex.Flowable;
 import io.realm.annotations.Beta;
-import io.realm.exceptions.DownloadingRealmInterruptedException;
 import io.realm.exceptions.RealmException;
 import io.realm.exceptions.RealmFileException;
 import io.realm.exceptions.RealmMigrationNeededException;
@@ -144,7 +143,6 @@ import io.realm.sync.permissions.Role;
 public class Realm extends BaseRealm {
 
     private static final String NULL_CONFIG_MSG = "A non-null RealmConfiguration must be provided";
-    private static final String FATAL_ASSERT_MESSAGE = "This should not happen. Please report this to help@realm.io";
 
     public static final String DEFAULT_REALM_NAME = RealmConfiguration.DEFAULT_REALM_NAME;
 
@@ -341,7 +339,7 @@ public class Realm extends BaseRealm {
      * @throws RealmMigrationNeededException if no migration has been provided by the default configuration and the
      * RealmObject classes or version has has changed so a migration is required.
      * @throws RealmFileException if an error happened when accessing the underlying Realm file.
-     * @throws DownloadingRealmInterruptedException if {@link SyncConfiguration.Builder#waitForInitialRemoteData()}
+     * @throws io.realm.exceptions.DownloadingRealmInterruptedException if {@link SyncConfiguration.Builder#waitForInitialRemoteData()}
      * was set and the thread opening the Realm was interrupted while the download was in progress.
      * @throws IllegalStateException if no default configuration is defined or if the default sync
      * configuration was used, but no user is logged in.
--- a/realm/realm-library/src/main/java/io/realm/Realm.java
+++ b/realm/realm-library/src/main/java/io/realm/Realm.java
@@ -50,6 +50,7 @@ import javax.annotation.Nullable;
 
 import io.reactivex.Flowable;
 import io.realm.annotations.Beta;
+import io.realm.exceptions.DownloadingRealmInterruptedException;
 import io.realm.exceptions.RealmException;
 import io.realm.exceptions.RealmFileException;
 import io.realm.exceptions.RealmMigrationNeededException;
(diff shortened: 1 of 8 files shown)


realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -149,10 +146,11 @@
 
     public static final String DEFAULT_REALM_NAME = RealmConfiguration.DEFAULT_REALM_NAME;
 
+    private final RealmSchema schema;
     private static final Object defaultConfigurationLock = new Object();
     // guarded by `defaultConfigurationLock`
-    private static RealmConfiguration defaultConfiguration;
-    private final RealmSchema schema;
+    private static volatile boolean userClearedConfiguration = false;
```
nhachicha: Why do we need this? AFAIK this is only set...

realm/realm-library/src/objectServer/java/io/realm/internal/SyncObjectServerFacade.java
```diff
@@ -191,4 +194,56 @@ public boolean isPartialRealm(RealmConfiguration configuration) {
...
+
+    // Only call this method while under the lock of `Realm.defaultConfigurationLock`
+    @Override
+    public RealmConfiguration getSystemDefaultSyncConfiguration() {
+        SyncUser currentUser;
+        try {
+            currentUser = SyncUser.currentUser();
+        } catch (IllegalStateException e ) {
+            throw new IllegalStateException("Could not create a default sync configuration as " +
+                    "multiple users are logged in. Use Realm.getInstance() instead of se", e);
```
nhachicha: >  instead of se

realm/realm-library/src/objectServer/java/io/realm/internal/SyncObjectServerFacade.java
```diff
@@ -191,4 +194,56 @@ public boolean isPartialRealm(RealmConfiguration configuration) {
...
+        }
+
+        if (currentUser == null) {
+            throw new IllegalStateException("No user is logged in. Login a user using SyncUser.login() first.");
+        }
+
+        // Check if a new user have become the "default" user. In that case, invalidate any
```
nhachicha: has become

bdash:
> This also had some consequences for Realm.getDefaultConfiguration() and Realm.removeDefaultConfiguration(), but essentially they continue to work as before, until you call getDefaultInstance() which then creates a new default sync configuration.

This seems like surprising behavior. What if a user wanted to do something involving the default configuration prior to first opening the Realm? It also seems surprising that opening a Realm using the default configuration would have the side-effect of changing the default configuration.

bdash:
Is there an asynchronous equivalent to `getDefaultInstance`? If not, how would I open the default Realm asynchronously rather than synchronously?

cmelchior:
@bdash It is probably poorly worded. If you manually set the default configuration, we will not override it. The only exception is that if you remove the default config again by setting it to null, we will replace it with a new default sync config. That was what I meant.

The reason being that trying to track _why_ it was null was probably more trouble than it was worth.

cmelchior:
@nhachicha No plans to add tests because the functionality is entirely local. I guess we could add tests for the properties of the default sync Realm, like the URL and it being partial, but I'm not sure that is needed?

Zhuinden:
I assume `Realm.getInstanceAsync(Realm.getDefaultConfiguration(), new Realm.Callback`

cmelchior:
Async callbacks are currently categorized as "advanced" usage ad falls outside the scope of `getDefaultInstance`. We actually talked about that specific use case when adding it in the first place and is the reason there isn't a `getDefaultInstanceAsync()`

bdash:
> @bdash It is probably poorly worded. If you manually set the default configuration, we will not override it. The only exception is that if you remove the default config again by setting it to null, we will replace it with a new default sync config. That was what I meant.

My point is that it seems incredibly surprising to have this fail (pseudocode):
```
let config1 = getDefaultConfiguration()
getDefaultInstance()
let config2 = getDefaultConfiguration()
ASSERT(config1 == config2);
```

 Having retrieving the default instance change the default configuration seems bizarre.

> Async callbacks are currently categorized as "advanced" usage ad falls outside the scope of getDefaultInstance.

Ok, so now if the user wants to open the Realm asynchronously they have to resort to manually constructing the entire sync configuration?

Is there a reason not to follow Cocoa's lead and have it be the _configuration_ that's automatically set up for the default synced Realm, rather than the function that retrieves the default synced Realm? That'd trivially allow opening the default synced Realm either synchronously or asynchronously using existing API. As it stands this API appears to force you to do things entirely manually if you want to do anything but the trivial. The configuration approach seems like a good middle ground that makes the easy things easy, while also making the more advanced cases easier.

cmelchior:
Yes, we could easily just have a factory method like `SyncConfiguration.automatic()`, but it seemed to me that you only added it because you have no other way of determining if it was a sync build or not (which Java does).

Java's approach cuts down on the number of lines of code you need to write, at the expense of few oddities in, what looks like, corner cases. 

Having something like
```
Realm.init(context); // Initialize Realm, always required.

// Setup Realm
Realm.setDefaultConfiguration(SyncConfiguration.automatic());
```

Would mirror what you have and having it consistent across bindings is definitely a benefit.
I guess your approach has the advantage of locking down the user being used as well, which in Java's case is also a little weird as the default instance returned can change depending on the user.

My only concern was that we were optimizing for the very simple starting out case, but it depends on how much we value a super simple "tutorial" flow vs. something that is slightly more verbose?

@nhachicha What do you think. Adding factory methods to `SyncConfiguration` seems to fix the concerns you had about the API as well we discussed earlier right? We do give up a little on the onboarding experience, but the API gets a lot more clear + we align with Swift.

bdash:
> Yes, we could easily just have a factory method like SyncConfiguration.automatic(), but it seemed to me that you only added it because you have no other way of determining if it was a sync build or not (which Java does).

There's two different aspects at play here:
1. Should you have to explicitly opt into this new default synced Realm.
2. At what level should the new default synced Realm stuff operate (configuration, or Realm).

You're right that 1 makes it so Cocoa cannot change our default behavior, while Java could. That's independent from the question of whether the thing whose behavior changes is the configuration or the Realm.

If you'd proposed having `getDefaultConfiguration()` be the thing whose behavior changes (so the default configuration is now for the default synced Realm) then it'd match Cocoa's approach. It'd make it possible for the default synced Realm to be easily used for both synchronous and asynchronous opening of Realms, as well as allowing the user to modify the configuration if they want to use something similar to but not the same as the default configuration (a per-user Realm path, for instance).

Using the factory method does have a second advantage, though. It allows use of the automatic configuration even if there's more than one sync user logged in, since it allows the user to explicitly specify which user the configuration should be for. This again allows for customizing the behavior without having to fall back to creating the entire configuration from scratch.

Zhuinden:
I still think this would be much easier if instead of `getInstance(RealmConfig)` and `getInstanceAsync(RealmConfig)` and `getDefaultInstance()`, there would be only `open(RealmConfig)`, `openAsync(RealmConfig)` and `open()`.

But I think I'm late with that proposal by about 3 years.

cmelchior:
> But I think I'm late with that proposal by about 3 years.

Yes 😉

nhachicha:
Is the scope of `Realm.setDefaultConfiguration(SyncConfiguration.automatic());` limited to the use of the `default` Realm URL, or does it assume a logged in user to be used (similar to https://github.com/realm/realm-java/pull/5813/files#diff-c13707dd883d2901ac94ca3cbea87820R221) 

I have [one concern](https://github.com/realm/realm-cocoa/issues/5653#issuecomment-371640484) with `SyncConfiguration.automatic()` as used with Cocoa

cmelchior:
Replaced by #5822

PR #5820, opened 2018-03-09
logout logging not resuming sync

nhachicha:
Currently when the user login, the Object Store will try to revive its inactive sessions
(stored previously after a logout). This will cause the OS to call `bindSession` to obtain an access token, however since the Realm might not be open yet, the `wrapObjectStoreSessionIfRequired`will not be invoked to wrap the OS store session with the Java session, which causes the Sync client to not resume syncing.

### DIFF
--- a/realm/realm-library/src/objectServer/java/io/realm/SyncManager.java
+++ b/realm/realm-library/src/objectServer/java/io/realm/SyncManager.java
@@ -43,7 +43,6 @@ import javax.net.ssl.TrustManagerFactory;
 import javax.net.ssl.X509TrustManager;
 
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
-import io.realm.exceptions.RealmException;
 import io.realm.internal.Keep;
 import io.realm.internal.Util;
 import io.realm.internal.network.AuthenticationServer;
@@ -206,16 +205,14 @@ public class SyncManager {
     /**
      * Gets a cached {@link SyncSession} for the given {@link SyncConfiguration} or throw if no one exists yet.
      *
-     * A session should exists after your open a Realm with a {@link SyncConfiguration}.
-     *
-     * Note: This is mainly for internal usage, consider using {@link #getSession(SyncConfiguration)} instead.
+     * A session should exists after you open a Realm with a {@link SyncConfiguration}.
      *
      * @param syncConfiguration configuration object for the synchronized Realm.
      * @return the {@link SyncSession} for the specified Realm.
      * @throws IllegalArgumentException if syncConfiguration is {@code null}.
-     * @throws RealmException if the session could not be found using the provided {@code SyncConfiguration}.
+     * @throws IllegalStateException if the session could not be found using the provided {@code SyncConfiguration}.
      */
-    public static synchronized SyncSession getSession(SyncConfiguration syncConfiguration) throws RealmException {
+    public static synchronized SyncSession getSession(SyncConfiguration syncConfiguration) throws IllegalStateException {
         //noinspection ConstantConditions
         if (syncConfiguration == null) {
             throw new IllegalArgumentException("A non-empty 'syncConfiguration' is required.");
@@ -223,7 +220,7 @@ public class SyncManager {
 
         SyncSession session = sessions.get(syncConfiguration.getPath());
         if (session == null) {
-            throw new RealmException("No SyncSession found using the path : " + syncConfiguration.getPath()
+            throw new IllegalStateException("No SyncSession found using the path : " + syncConfiguration.getPath()
             + "\nplease ensure to call this method after you've open the Realm");
         }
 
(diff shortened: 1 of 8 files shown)


realm/realm-library/src/objectServer/java/io/realm/SyncManager.java
```diff
@@ -209,7 +210,7 @@ public static void setDefaultSessionErrorHandler(@Nullable SyncSession.ErrorHand
      * @return the {@link SyncSession} for the specified Realm.
      * @throws IllegalArgumentException if syncConfiguration is {@code null}.
      */
-    public static synchronized SyncSession getSession(SyncConfiguration syncConfiguration) {
+    public static synchronized SyncSession getSession(SyncConfiguration syncConfiguration, URI... resolvedRealmURL) {
```
cmelchior: This API is public. I don't assume that was the intent? 

I can see the issue though, the `SyncManager` is in `io.realm` while the facade is in `io.realm.internal`. I'm trying to remember why the facade isn't just package private in `io.realm`? I think I tried at some point, but something else failed. In any case, there is a lot of TODO/FIXME's in the Facade about it.

realm/realm-library/src/objectServer/java/io/realm/SyncManager.java
```diff
@@ -226,6 +227,15 @@ public static synchronized SyncSession getSession(SyncConfiguration syncConfigur
             sessions.put(syncConfiguration.getPath(), session);
             if (sessions.size() == 1) {
                 RealmLog.debug("first session created add network listener");
+                if(resolvedRealmURL.length > 0) {
```
cmelchior: Can it be more than 1, why use varargs?

realm/realm-library/src/syncIntegrationTest/java/io/realm/SyncedRealmTests.java
```diff
@@ -49,6 +49,36 @@
...
+
+        Realm realm = Realm.getInstance(config);
+        realm.beginTransaction();
+        realm.createObject(StringOnly.class).setChars("Foo");
+        realm.commitTransaction();
+        SyncManager.getSession(config).uploadAllLocalChanges();
+        user.logout();
+        realm.close();
+
+        user = SyncUser.login(SyncCredentials.usernamePassword(username, password, false), Constants.AUTH_URL);
```
cmelchior: You are logging in the same user So the path on disk would be the same. I don't see how this actually test that syncing works again as you just uploaded all changes and nothing is being created on the server in the meantime?

realm/realm-library/src/objectServer/java/io/realm/SyncManager.java
nhachicha: it's a variadic function, so shouldn't break the API

realm/realm-library/src/objectServer/java/io/realm/SyncManager.java
nhachicha: see https://github.com/realm/realm-java/pull/5820#discussion_r173489662

realm/realm-library/src/objectServer/java/io/realm/SyncManager.java
cmelchior: It doesn't break it, but it is totally non-obvious what this paramter do. I would still pull this signature out in its own method, and then document it as internal so we can remove it later if at all possible?

realm/realm-library/src/objectServer/java/io/realm/SyncManager.java
```diff
@@ -201,15 +203,44 @@ public static void setDefaultSessionErrorHandler(@Nullable SyncSession.ErrorHand
         }
     }
 
+    /**
+     * Gets a cached {@link SyncSession} for the given {@link SyncConfiguration} or throw if no one exists yet.
+     *
+     * A session should exists after your open a Realm with a {@link SyncConfiguration}.
+     *
+     * Note: This is mainly for internal usage, consider using {@link #getSession(SyncConfiguration)} instead.
```
cmelchior: Wrong copy/paste?

realm/realm-library/src/objectServer/java/io/realm/SyncManager.java
```diff
@@ -201,15 +203,44 @@ public static void setDefaultSessionErrorHandler(@Nullable SyncSession.ErrorHand
...
      * no one exists.
      *
+     * Note: This is mainly for internal usage, consider using {@link #getSession(SyncConfiguration)} instead.
+     *
      * @param syncConfiguration configuration object for the synchronized Realm.
      * @return the {@link SyncSession} for the specified Realm.
      * @throws IllegalArgumentException if syncConfiguration is {@code null}.
      */
-    public static synchronized SyncSession getSession(SyncConfiguration syncConfiguration) {
+    public static synchronized SyncSession getOrCreateSession(SyncConfiguration syncConfiguration, URI... resolvedRealmURL) {
```
cmelchior: Javadoc missing for `resolvedRealmURL`

cmelchior: Also, there is no need to make this varadic now, right?

realm/realm-library/src/objectServer/java/io/realm/SyncManager.java
cmelchior: See above. I assume this can be removed now?

realm/realm-library/src/objectServer/java/io/realm/SyncUser.java
```diff
@@ -229,9 +230,11 @@ public void logout() {
 
             // invalidate all pending refresh_token queries
             for (SyncConfiguration syncConfiguration : realms.keySet()) {
-                SyncSession session = SyncManager.getSession(syncConfiguration);
-                if (session != null) {
+                try {
+                    SyncSession session = SyncManager.getSession(syncConfiguration);
                     session.clearScheduledAccessTokenRefresh();
+                } catch (RealmException e) {
+                    // No session, either the Realm was not opened or session was removed.
```
cmelchior: RealmException is kinda generic. It would feel safer if we did more checking here, like checking the error message before just throwing the error away?

realm/realm-library/src/objectServer/java/io/realm/SyncManager.java
Zhuinden: also: `should exists` => `exist`

nhachicha:
merging, unrelated CI error 
> Error response from daemon: cannot join network of a non running container

PR #5822, opened 2018-03-10
Add support for automatic configurations

### LINKED ISSUE #5806: Default Sync Realm API

cmelchior:
Fixes #5806 
Replaces #5813

See #5813 for the discussion.

### DIFF
--- a/realm/realm-library/src/objectServer/java/io/realm/SyncConfiguration.java
+++ b/realm/realm-library/src/objectServer/java/io/realm/SyncConfiguration.java
@@ -22,6 +22,7 @@ import java.io.File;
 import java.io.UnsupportedEncodingException;
 import java.net.URI;
 import java.net.URISyntaxException;
+import java.net.URL;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
 import java.util.Arrays;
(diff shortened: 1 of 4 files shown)


realm/realm-library/src/objectServer/java/io/realm/SyncConfiguration.java
```diff
@@ -183,6 +184,64 @@ public static RealmConfiguration forRecovery(String canonicalPath, @Nullable byt
...
+     * Creates an automatic default configuration based on the the currently logged in user.
+     * <p>
+     * This configuration will point to the default Realm on the server where the user was
+     * authenticated.
+     *
+     * @throws IllegalStateException if no user are logged in, or multiple users have. Only one should
+     * be logged in when calling this method.
+     * @return The constructed {@link SyncConfiguration}.
+     */
+    public static SyncConfiguration automatic() {
```
cmelchior: I'm not sure why do not call this `default()` instead. Somehow that feels more natural to me.

bdash: I'm not sure about the default vs automatic naming. To me "default" feels a little more static, while "automatic" is more active. Given that the returned configuration can change between calls due to outside actions (changing the active user), the more active term felt like a better fit. This perception is very subjective, though.

One thing worth noting is that Cocoa's `SyncConfiguration.automatic()` returns a `Realm.Configuration`, not a `SyncConfiguration`. That was done to minimize the amount of code required to use the returned configuration.

cmelchior: In Java `SyncConfiguration` is a subclass of `RealmConfiguration` so it can be used everywhere a RealmConfiguration is used.

Good point about `default` feeling more static vs. `automatic` that feels more active. Lets just stick to `automatic()`

cmelchior:
Waiting for a release of ROS with support for the default Realm before merging this.

bdash:
That support is in v3.0.0-beta.3.

realm/realm-library/src/syncIntegrationTest/java/io/realm/SyncedRealmIntegrationTests.java
```diff
@@ -276,4 +276,21 @@ public void waitForInitialRemoteData_readOnlyFalse_upgradeSchema() {
...
+    public void defaultRealm() throws InterruptedException {
+        SyncCredentials credentials = SyncCredentials.nickname("test", true);
+        SyncUser user = SyncUser.login(credentials, Constants.AUTH_URL);
+        SyncConfiguration config = SyncConfiguration.automatic();
+        Realm realm = Realm.getInstance(config);
+        SyncManager.getSession(config).downloadAllServerChanges();
+        realm.refresh();
+
+        try {
+            assertFalse(realm.isEmpty());
```
nhachicha: I assume it's `false` because you expect the Permission schema to be populated?

cmelchior: Yes, I need to merge in your branch to verify though

PR #5830, opened 2018-03-13
Fix module example

cmelchior:
This fixes the app module example.

It, unfortunately, does this by reverting the code that tests that app modules can link to library module model classes. The reason for this is that testing this is impossible without seriously complicating the current example (e.g. just using the default configuration is no longer possible).

Creating another example just for this seems a bit much and the chance of the functionality breaking is probably low.

I verified that it worked manually before removing the code, by using the following default config `RealmConfiguration defaultConfig = new RealmConfiguration.Builder().modules(Realm.getDefaultModule(), new DomesticAnimalsModule()).build();`

/cc @nhachicha

### DIFF
--- a/examples/moduleExample/app/src/main/java/io/realm/examples/appmodules/model/Pig.java
+++ b/examples/moduleExample/app/src/main/java/io/realm/examples/appmodules/model/Pig.java
@@ -16,17 +16,11 @@
 
 package io.realm.examples.appmodules.model;
 
-import io.realm.RealmList;
 import io.realm.RealmObject;
-import io.realm.examples.librarymodules.model.Dog;
 
 public class Pig extends RealmObject {
 
     private String name;
-
-    // It is possible for model classes to to reference library model classes as long
-    // as they all are included in the schema when opening the Realm.
-    private RealmList<Dog> afraidOf = new RealmList<>();
 
     public String getName() {
         return name;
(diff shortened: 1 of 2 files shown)


examples/moduleExample/app/src/main/java/io/realm/examples/appmodules/model/Pig.java
```diff
@@ -16,18 +16,12 @@
...
 import io.realm.RealmObject;
-import io.realm.examples.librarymodules.model.Dog;
 
 public class Pig extends RealmObject {
 
     private String name;
 
-    // It is possible for model classes to to reference library model classes as long
-    // as they all are included in the schema when opening the Realm.
-    private RealmList<Dog> afraidOf = new RealmList<>();
```
Zhuinden: It is no longer possible to use library model classes in classes of the default module?

cmelchior: It is, but it is just impossible to show in this example at the same time as showing how the default config works.

cmelchior: So instead of having an example that is only showing a pretty small cornercase, I thought it was better to show how the default settings work.

nhachicha: You can, but you should include the library module as well as the default. since the default scope doesn’t include implicitly the library models

PR #5834, opened 2018-03-13
Update dependencies

nhachicha:


cmelchior:
Some of the Object Level Tests are crashing. Restarting to check if they are flaky or not: https://ci.realm.io/blue/organizations/jenkins/realm%2Frealm-java/detail/PR-5834/2/pipeline/

nhachicha:
strange there's no `delete` in `restrictAccessToOwner` test?, usually as soon as one test fails and one session remains, the OS sync_manger assertion will fail hence the native crash on the second test

nhachicha:
I'm getting elsewhere `AuthenticateResponse - Error: ACCESS_DENIED(614)
03-13 21:01:08.931 12833 12906 D REALM_JAVA: The path is invalid or current user has no access.` which can cause a test timeout. When the test tear down it tries to delete the Realm in `TestRealmConfigurationFactory` which may cause the failure the CI is experiencing

PR #5840, opened 2018-03-14
Update GridView example

Zhuinden:
Base: Master. You guys probably know which branch this is supposed to go on.

Currently the GridView example was written by rlwaldrop in 2014.09, and it was never updated beyond adding `Realm.init(Context)`. So it had some pretty bad practices regarding current-day Realm usage.

So I updated it.

### DIFF
--- a/examples/gridViewExample/src/main/java/io/realm/examples/realmgridview/CityAdapter.java
+++ b/examples/gridViewExample/src/main/java/io/realm/examples/realmgridview/CityAdapter.java
@@ -48,7 +48,7 @@ public class CityAdapter extends BaseAdapter {
 
     @Override
     public int getCount() {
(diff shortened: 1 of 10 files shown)


examples/gridViewExample/src/main/java/io/realm/examples/realmgridview/CityAdapter.java
```diff
@@ -33,32 +33,26 @@
...
+        notifyDataSetChanged();
     }
 
     @Override
     public int getCount() {
-        if (cities == null) {
-            return 0;
-        }
-        return cities.size();
+        return cities == null ? 0 : cities.size();
```
cmelchior: This can never be null given the code in `setData` above?

Zhuinden:
PR comment addressed

PR #5841, opened 2018-03-14
Update Intro example

Zhuinden:
Base: Master, you guys know what this branch should go on.

The intro example was kinda old, so I did some magic to it. Added a RealmChangeListener, a RealmList<String>, an `@Index`, a `@PrimaryKey`, things like that.

I did preserve its initial scenarios, though.


examples/introExample/src/main/java/io/realm/examples/intro/IntroExampleActivity.java
```diff
@@ -77,54 +69,74 @@ protected void onDestroy() {
...
-            public void execute(Realm realm) {
-                // Add a person
-                Person person = realm.createObject(Person.class);
-                person.setId(1);
-                person.setName("Young Person");
-                person.setAge(14);
+        // Results obtained from a Realm are live, and can be observed on looper threads (like the UI thread).
+        // Note that if you want to observe the RealmResults for a long time, then it should be a field reference.
+        // Otherwise, the RealmResults can no longer be notified if the GC has cleared the reference to it.
+        RealmResults<Person> persons = realm.where(Person.class).findAll();
```
cmelchior: Maybe we should make this a field reference. I know it isn't needed, but it reads a bit funny to have the above disclaimer and then proceeding not doing it.

examples/introExample/src/main/java/io/realm/examples/intro/IntroExampleActivity.java
```diff
@@ -77,54 +69,74 @@ protected void onDestroy() {
...
 
+        // The change listener will be notified when the Realm is written to from any threads.
+        persons.addChangeListener((people, changeSet) -> {
+            if(changeSet == null) {
```
cmelchior: With 5.0, this can never be `null` anymore. Instead check for `changeSet.getState() == State.INITITAL`

examples/introExample/src/main/java/io/realm/examples/intro/MyApplication.java
```diff
@@ -27,5 +27,9 @@ public void onCreate() {
         super.onCreate();
         // Initialize Realm. Should only be done once when the application starts.
         Realm.init(this);
+
+        // In this example, no default configuration is set, so by default, `RealmConfiguration.Builder().build()` is used.
+        // Please note that while this helps when starting out, you might want to set your own custom default configuration.
+        // Generally, either add a migration to handle schema changes, or specify `deleteIfMigrationNeeded()`.
```
cmelchior: I would just remove this sentence. IMO the above is fine and this just goes into too much detail for an intro example

examples/introExample/src/main/java/io/realm/examples/intro/IntroExampleActivity.java
Zhuinden: I was using `master` branch so this didn't have `changeSet.getState()` yet ._.

examples/introExample/src/main/java/io/realm/examples/intro/IntroExampleActivity.java
Zhuinden: Makes sense

Zhuinden:
PR comments addressed, although the submodule update upstream kinda got in the way of actually testing how I changed the change set code.

It should work?

PR #5843, opened 2018-03-14
Update Kotlin sample

Zhuinden:
Base: Master, and this is also eerily familiar.

Although I made less changes to it than I did do the intro version.


Zhuinden:
Agh I'm conflicting about something

Zhuinden:
`findAllSorted()` vs `sort().findAll()` apparently. Fixed

PR #5848, opened 2018-03-19
Expose bugs with findFirst

### LINKED ISSUE #5714: findFirst() not working with sort()
Using DynamicRealm 4.3.3 

A) 
`DynamicRealmObject lastTrack = realm.where(Model.TRACK).sort("timeStart", Sort.DESCENDING).findAll().first(null);`
and 
B) 
`DynamicRealmObject lastTrack = realm.where(Model.TRACK).sort("timeStart", Sort.DESCENDING).findFirst();`

A) and B) have different results. A) is correct working, B) not working

cmelchior:
Fixes #5714 

`findFirst` was broken in a lot of scenarios involving sorting which apparently wasn't tested.
The issue was that going directly to the `Table` API wasn't the right choice when sorting is involved, so in those cases, we need to construct a TableView first.

The CI failures are unrelated and will hopefully be fixed in another PR.


realm/realm-library/src/androidTest/java/io/realm/RealmQueryTests.java
```diff
@@ -1366,6 +1366,60 @@ public void findFirst() {
...
+    public void findFirst_subQuery() {
+        realm.beginTransaction();
+        realm.insert(new Dog("Milo"));
+        realm.insert(new Dog("Fido"));
+        realm.insert(new Dog("Bella"));
+        realm.commitTransaction();
+
+        RealmResults<Dog> dogs = realm.where(Dog.class).in("name", new String[] { "Fido", "Bella" }).findAll();
+        Dog dog = dogs.where().findFirst();
+        assertEquals("Fido", dog.getName());
```
nhachicha: This may break when moving to stable keys

PR #5853, opened 2018-03-20
Fix SyncSession tests

cmelchior:
**Changes:**

* `SyncManager.NETWORK_THREAD_POOL` is now public
* Will now shutdown and close the network thread pool before exiting tests that log all users out (which result in network requests)
* Object Level Permission tests now depend on the global default Realm instead of trying to create one themselves (which often fail because of the `PermissionManager` being laky).

### DIFF
--- a/realm/realm-library/src/objectServer/java/io/realm/internal/SyncObjectServerFacade.java
+++ b/realm/realm-library/src/objectServer/java/io/realm/internal/SyncObjectServerFacade.java
@@ -195,12 +195,14 @@ public class SyncObjectServerFacade extends ObjectServerFacade {
         builder.addModule(new ObjectPermissionsModule());
     }
 
+    // Only exposed for testing purposes
     @Override
     @SuppressFBWarnings("ST_WRITE_TO_STATIC_FROM_INSTANCE_METHOD")
     public void waitForNetworkThreadExecutorToFinish() {
-        // Since the network pool should only consist of remote logout calls at the point where this
-        // is called. These can be safely interrupted, so just shutdown the pool and create a new
-        // that can be used by future tests.
+        // Since this method should only be called when exiting a test, it should be safe to just
+        // cancel all ongoing network requests and shut down the pool as soon as possible.
+        // When shut down we replace it with a new, now empty, pool that can be used by future
+        // tests
         SyncManager.NETWORK_POOL_EXECUTOR.shutdownNow();
         try {
             SyncManager.NETWORK_POOL_EXECUTOR.awaitTermination(5, TimeUnit.SECONDS);
--- a/realm/realm-library/src/objectServer/java/io/realm/internal/SyncObjectServerFacade.java
+++ b/realm/realm-library/src/objectServer/java/io/realm/internal/SyncObjectServerFacade.java
@@ -23,7 +23,11 @@ import android.net.ConnectivityManager;
 
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
+import java.util.concurrent.ArrayBlockingQueue;
+import java.util.concurrent.ThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
 
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import io.realm.RealmConfiguration;
 import io.realm.SyncConfiguration;
 import io.realm.SyncManager;
@@ -189,5 +193,24 @@ public class SyncObjectServerFacade extends ObjectServerFacade {
     @Override
     public void addSupportForObjectLevelPermissions(RealmConfiguration.Builder builder) {
         builder.addModule(new ObjectPermissionsModule());
+    }
+
+    @Override
+    @SuppressFBWarnings("ST_WRITE_TO_STATIC_FROM_INSTANCE_METHOD")
+    public void waitForNetworkThreadExecutorToFinish() {
+        // Since the network pool should only consist of remote logout calls at the point where this
+        // is called. These can be safely interrupted, so just shutdown the pool and create a new
+        // that can be used by future tests.
+        SyncManager.NETWORK_POOL_EXECUTOR.shutdownNow();
+        try {
+            SyncManager.NETWORK_POOL_EXECUTOR.awaitTermination(5, TimeUnit.SECONDS);
+        } catch (InterruptedException e) {
+            throw new AssertionError("NetworkPoolExecutor was not shut down in time:\n" + Util.getStackTrace(e));
+        } finally {
+            // Replace the executor, since the old one is now dead.
+            // The setup of this should mirror what is done in SyncManager.
+            SyncManager.NETWORK_POOL_EXECUTOR = new ThreadPoolExecutor(
+                    10, 10, 0, TimeUnit.MILLISECONDS, new ArrayBlockingQueue<Runnable>(100));
+        }
     }
 }
--- a/realm/realm-library/src/objectServer/java/io/realm/internal/SyncObjectServerFacade.java
+++ b/realm/realm-library/src/objectServer/java/io/realm/internal/SyncObjectServerFacade.java
@@ -23,11 +23,7 @@ import android.net.ConnectivityManager;
 
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
-import java.util.concurrent.ArrayBlockingQueue;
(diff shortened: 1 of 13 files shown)


realm/realm-library/src/syncIntegrationTest/java/io/realm/SyncedRealmIntegrationTests.java
```diff
@@ -324,15 +324,15 @@ public void waitForInitialRemoteData_readOnlyFalse_upgradeSchema() {
...
+        SyncCredentials credentials = SyncCredentials.nickname("test", false);
         SyncUser user = SyncUser.logIn(credentials, Constants.AUTH_URL);
         SyncConfiguration config = SyncConfiguration.automatic();
         Realm realm = Realm.getInstance(config);
         SyncManager.getSession(config).downloadAllServerChanges();
         realm.refresh();
 
         try {
-            assertFalse(realm.isEmpty());
+            assertTrue(realm.isEmpty());
```
nhachicha: FYI the `default` Realm is not completely empty on ROS, it has a default table `_DefaultRealmVersion` which maybe synchronised if you're not using partial Sync

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/ObjectLevelPermissionIntegrationTests.java
```diff
@@ -192,17 +174,10 @@ public void restrictAccessToOwner() throws InterruptedException {
...
-                    break;
-                case UPDATE:
-                    assertEquals(0, permissionObjects2.size());
-                    looperThread.testComplete();
-                    break;
-                case ERROR:
-                    fail("Unexpected error callback");
-                    break;
+            RealmLog.info("State: " + changeSet.getState().toString() + ", complete: " + changeSet.isCompleteResult());
+            if (changeSet.isCompleteResult()) {
```
nhachicha: nice 👍🏼

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/utils/UserFactory.java
```diff
@@ -145,18 +145,20 @@ public static void logoutAllUsers() {
...
+            for (SyncUser user : users.values()) {
+                user.logOut();
             }
+            ObjectServerFacade.getSyncFacadeIfPossible().waitForNetworkThreadExecutorToFinish();
+            allUsersLoggedOut.countDown();
         });
         TestHelper.awaitOrFail(allUsersLoggedOut);
         ht.quit();
+        try {
+            ht.join(5000);
```
nhachicha: use `TestHelper.SHORT_WAIT_SECS` with `TimeUnit`

realm/realm-library/src/objectServer/java/io/realm/internal/SyncObjectServerFacade.java
```diff
@@ -190,4 +194,23 @@ public boolean isPartialRealm(RealmConfiguration configuration) {
     public void addSupportForObjectLevelPermissions(RealmConfiguration.Builder builder) {
         builder.addModule(new ObjectPermissionsModule());
     }
+
+    @Override
+    @SuppressFBWarnings("ST_WRITE_TO_STATIC_FROM_INSTANCE_METHOD")
+    public void waitForNetworkThreadExecutorToFinish() {
```
nhachicha: I like the idea, however, this is used only in tests, I prefer using this logic inside `SyncTestUtils` (similar to what we did with accessing the refresh token with reflection)

realm/realm-library/src/objectServer/java/io/realm/internal/SyncObjectServerFacade.java
```diff
@@ -190,4 +194,23 @@ public boolean isPartialRealm(RealmConfiguration configuration) {
...
+        // Since the network pool should only consist of remote logout calls at the point where this
```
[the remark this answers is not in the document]
cmelchior: No guarantees as such, but since the test is finishing when this is called, most likely that should be the case (since ending the test also calls `user.logOut()` for all users). 

In any case, no matter what is in it, the result is no longer relevant because the test either already succeeded or failed. I'll reword the description

realm/realm-library/src/objectServer/java/io/realm/internal/SyncObjectServerFacade.java
cmelchior: Fair enough. I'll move the implementation

realm/realm-library/src/objectServer/java/io/realm/internal/network/ExponentialBackoffTask.java
```diff
@@ -66,6 +66,9 @@ protected boolean shouldAbortTask(T response) {
     public void run() {
         int attempt = 0;
         while (true) {
+            if (Thread.interrupted()) {
```
nhachicha: you can simplify the expression by using the condition in the while 
```
while(!Thread.interrupted()) {
//...
}
```

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/utils/UserFactory.java
```diff
@@ -150,13 +151,13 @@ public static void logoutAllUsers() {
...
             }
-            ObjectServerFacade.getSyncFacadeIfPossible().waitForNetworkThreadExecutorToFinish();
+            TestHelper.waitForNetworkThreadExecutorToFinish();
             allUsersLoggedOut.countDown();
         });
         TestHelper.awaitOrFail(allUsersLoggedOut);
         ht.quit();
         try {
-            ht.join(5000);
+            ht.join(TimeUnit.MILLISECONDS.convert(TestHelper.SHORT_WAIT_SECS, TimeUnit.SECONDS));
```
nhachicha: I think `TimeUnit.SECONDS.toMillis(TestHelper.SHORT_WAIT_SECS)` is more concise

realm/realm-library/src/testUtils/java/io/realm/TestHelper.java
```diff
@@ -1249,4 +1252,39 @@ public static void populateLinkedDataSet(Realm realm) {
...
+     * This method will kill all tasks then shutdown and replace the SyncManager.NETWORK_POOL_EXECUTOR
+     * with a fresh and empty instance. This should only be called when exiting tests.
+     *
+     * If the build does not support Sync, this method will do nothing
+     */
+    public static void waitForNetworkThreadExecutorToFinish() {
+        try {
+            Class syncManager = Class.forName("io.realm.SyncManager");
+            Field field = syncManager.getDeclaredField("NETWORK_POOL_EXECUTOR");
+            ThreadPoolExecutor pool = (ThreadPoolExecutor) field.get(null);
```
nhachicha: minor: for performance reason, you can make the reflection code static for caching, since this is called more often for integration tests

realm/realm-library/src/testUtils/java/io/realm/TestHelper.java
```diff
@@ -1258,11 +1258,22 @@ public static void populateLinkedDataSet(Realm realm) {
...
-    public static void waitForNetworkThreadExecutorToFinish() {
+    private static final Field networkPoolExecutorField;
+    static {
         try {
             Class syncManager = Class.forName("io.realm.SyncManager");
-            Field field = syncManager.getDeclaredField("NETWORK_POOL_EXECUTOR");
-            ThreadPoolExecutor pool = (ThreadPoolExecutor) field.get(null);
+            networkPoolExecutorField = syncManager.getDeclaredField("NETWORK_POOL_EXECUTOR");
+        } catch (ClassNotFoundException | NoSuchFieldException e) {
+            networkPoolExecutorField = null;
```
nhachicha: this won't compile :) 
You should throw an assertion error like we do in `SyncTestUtils`. 
Btw I think `waitForNetworkThreadExecutorToFinish` should be on the `objectServer` flavour i.e inside `SyncTestUtils` not `TestHelper` which is a helper for the base flavour mainly

cmelchior: Yes, I agree it would have been better if this could be in `SyncTestUtils`, but I cannot move it there since this method is also called from `RunInLooperThread` which is also used by our base variant tests which do not have any Sync classes compiled in.

PR #5858, opened 2018-03-24
Expose the per-user private role

nhachicha:
Sync is adding a private per-user role in realm/realm-sync#2068. We'll need to update our schema to accommodate.

### DIFF
--- a/realm/realm-library/src/androidTestObjectServer/java/io/realm/ObjectLevelPermissionsTest.java
+++ b/realm/realm-library/src/androidTestObjectServer/java/io/realm/ObjectLevelPermissionsTest.java
@@ -32,6 +32,7 @@ import io.realm.sync.permissions.ClassPermissions;
 import io.realm.sync.permissions.ClassPrivileges;
 import io.realm.sync.permissions.ObjectPrivileges;
 import io.realm.sync.permissions.Permission;
+import io.realm.sync.permissions.PermissionUser;
 import io.realm.sync.permissions.RealmPermissions;
 import io.realm.sync.permissions.RealmPrivileges;
 import io.realm.sync.permissions.Role;
@@ -39,6 +40,7 @@ import io.realm.sync.permissions.Role;
 import static io.realm.util.SyncTestUtils.createTestUser;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 
@@ -428,6 +430,20 @@ public class ObjectLevelPermissionsTest {
 //            fail();
 //        } catch (IllegalStateException ignore) {
 //        }
+    }
+
+    @Test
+    public void userPrivateRole() {
+        RealmResults<PermissionUser> permissionUsers = realm.where(PermissionUser.class).findAll();
+        assertEquals(1, permissionUsers.size());
+
+        PermissionUser permissionUser = permissionUsers.get(0);
+        assertNotNull(permissionUser);
+        Role role = permissionUser.getRole();
+        assertNotNull(role);
+
+        assertEquals("__User:" + user.getIdentity(), role.getName());
+        assertTrue(role.hasMember(user.getIdentity()));
     }
 
     @Test
--- a/realm/realm-library/src/androidTestObjectServer/java/io/realm/ObjectLevelPermissionsTest.java
+++ b/realm/realm-library/src/androidTestObjectServer/java/io/realm/ObjectLevelPermissionsTest.java
@@ -457,7 +457,7 @@ public class ObjectLevelPermissionsTest {
         assertNull(builtInRole);
         permissionUser = realm.where(PermissionUser.class).equalTo("id", "id123").findFirst();
         assertNull(permissionUser.getRole());
-        assertNull(permissionUser.getRoles());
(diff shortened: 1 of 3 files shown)


realm/realm-library/src/main/java/io/realm/sync/permissions/PermissionUser.java
```diff
@@ -72,4 +74,18 @@ public String getId() {
     public @Nullable  RealmResults<Role> getRoles() {
         return roles;
     }
+
+    /**
+     * The user's private role. This will be initialized to a role named for the user's
```
cmelchior: I think this description is slightly misleading since the user can set it to anything after the fact. This actually works more like `tag` in a ViewHolder. I pushed for the semantics described here, but currently, that is not what is implemented.

Perhaps something like this instead.

```
Returns the role set by {@link #setRole(Role)}. This will be initialized to a role named for the user's identity that contains the user as its only member.
```

realm/realm-library/src/main/java/io/realm/sync/permissions/PermissionUser.java
```diff
@@ -72,4 +74,18 @@ public String getId() {
...
+     * identity that contains this user as its only member.
+     *
+     * @return User private {@link Role}.
+     */
+    public Role getRole() {
+        return role;
+    }
+
+    public void setRole(Role role) {
```
cmelchior: Add javadoc. Perhaps something like this:

```
Sets a role for this user that can be accessed later. When the PermissionUser object is created this will be initialized with a role named for the user's identity that contains the user as its only member.

```

realm/realm-library/src/androidTestObjectServer/java/io/realm/ObjectLevelPermissionsTest.java
```diff
@@ -430,6 +432,20 @@ public void getClassPermissions_closedRealmThrows() {
 //        }
     }
 
+    @Test
+    public void userPrivateRole() {
+        RealmResults<PermissionUser> permissionUsers = realm.where(PermissionUser.class).findAll();
+        assertEquals(1, permissionUsers.size());
+
+        PermissionUser permissionUser = permissionUsers.get(0);
```
cmelchior: It isn't clear to me where this user is created?

cmelchior: Perhaps also create a test for what happens if `createObject(PermissionUser.class)` is called as I suspect that the `role` isn't created until you talked to the server?

E.g. what happens if you do 
```
Role role = getExistingRole();
role.addMember(newUserId)`
realm.where(PermissionUser.class).equalTo("id", newUserId).findFirst().getRole(); // What is returned here?
```

realm/realm-library/src/main/java/io/realm/sync/permissions/PermissionUser.java
nhachicha: This method shouldn't exist, the `role` is built-in automatically at the Sync client, you can not set  it since `Role` is immutable as far as permission is concerned

cmelchior: Not sure what you mean? `Role` isn't immutable since you can/remove members all the time. If you mean `PermissionUser` then you are correct the name of the user is a key and hence immutable, but apart from that, it is just a standard Object.

There are many things about this property I find problematic. See https://github.com/realm/realm-sync/pull/2068#discussion_r173429225

I agree that exposing `setRole()` is just confusing and having it means that people can accidentally set it to what-ever. In any case, how it works, should be aligned across bindings.

Personally, I think having a `getPrivateRole()` only getter would be much more in line with the intended behaviour.

realm/realm-library/src/androidTestObjectServer/java/io/realm/ObjectLevelPermissionsTest.java
nhachicha: @cmelchior user is created during the `setUp` (`user = createTestUser();`)

realm/realm-library/src/main/java/io/realm/sync/permissions/PermissionUser.java
nhachicha: I agree `setRole` should be removed, furthermore, the user should not create `PermissionUser` directly it's more of an internal API, it's less likely that you ever want to create user directly, let alone setting the `PermissionUser` built-in Role.

realm/realm-library/src/main/java/io/realm/sync/permissions/PermissionUser.java
nhachicha: We'll keep the original description since the `setRole` is removed now

realm/realm-library/src/main/java/io/realm/sync/permissions/PermissionUser.java
```diff
@@ -72,4 +74,14 @@ public String getId() {
...
         return roles;
     }
+
+    /**
+     * The user's private role. This will be initialized to a role named for the user's
+     * identity that contains this user as its only member.
+     *
+     * @return User private {@link Role}.
+     */
+    public Role getRole() {
```
cmelchior: Any reason for calling this `getRole` vs. e.g. `getPrivateRole()` which seems more descriptive?

PR #5893, opened 2018-04-11
Query-based sync as the default mode

cmelchior:
In 6.0 Partial Sync will become the default mode instead of Full Synchronization.

Partial Sync has only been an internal working name for this feature, but the public name will be "Query-based Synchronization" or "Query-based Sync". This PR makes the changes to support that.

Most of the changes are in the unit tests as we can continue to use the old `isPartial` parameter internally. However, there is some value in also refactoring the internal API's just to avoid confusion (like with Backlinks/LinkingObjects). Thoughts @nhachicha ?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## 6.0.0 (YYYY-MM-DD)
+
+### Breaking Changes
+
+* [ObjectServer] Partial sync is now the default mode of synchronization. `SyncConfiguration.Builder.partialRealm()` has been replaced by `SyncConfiguration.Builder.fullSynchronization()`
+
+
 ## 5.0.1 (2018-04-09)
 
 ### Enhancements
--- a/realm/realm-library/src/androidTestObjectServer/java/io/realm/SyncConfigurationTests.java
+++ b/realm/realm-library/src/androidTestObjectServer/java/io/realm/SyncConfigurationTests.java
@@ -20,6 +20,7 @@ import android.support.test.InstrumentationRegistry;
 import android.support.test.runner.AndroidJUnit4;
 
 import org.junit.After;
+import org.junit.Before;
 import org.junit.Ignore;
 import org.junit.Rule;
 import org.junit.Test;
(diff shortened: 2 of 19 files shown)


realm/realm-library/src/objectServer/java/io/realm/SyncConfiguration.java
```diff
@@ -56,13 +56,32 @@
...
  *
+ * Synchronized Realms come in two forms:
+ * <ul>
+ *     <li>
+ *         <b>Query-based synchronization:</b>
+ *         This is the default mode. The Realm will only synchronize data you have queried for.
+ *         This means the Realm on the device is initially empty and will gradually fill up as
+ *         you start to query for data. This is useful if the server side Realm is too large
+ *         to fit on the device or contains data from multiple users. Data synchronized this way
+ *         can also be removed from the device again without being deleted on the server.
```
nhachicha: add a link to the old partial sync doc, so user have a full picture about the query based approach

realm/realm-library/src/androidTestObjectServer/java/io/realm/SyncConfigurationTests.java
```diff
@@ -61,12 +62,17 @@
...
+
     @After
     public void tearDown() {
-        for (SyncUser syncUser : SyncUser.all().values()) {
-            syncUser.logOut();
+        UserStore userStore = SyncManager.getUserStore();
+        for (SyncUser syncUser : userStore.allUsers()) {
+            userStore.remove(syncUser.getIdentity(), syncUser.getAuthenticationUrl().toString());
         }
-        SyncManager.reset();
```
nhachicha: why removing this? are you trying to avoid the `no_active_sessions` assertion in OS?

CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+## 6.0.0 (YYYY-MM-DD)
```
Zhuinden: `5.0.0 (2018-03-15)`

I'm curious when the next major release is :smile:

realm/realm-library/src/androidTestObjectServer/java/io/realm/SyncConfigurationTests.java
cmelchior: Yes, The `SyncConfigurationTests` don't actually log users in but fake it by creating a fake server response without touching the server. This means Object Store isn't initialized correctly resulting in a lot of weird errors.

cmelchior:
Ready for review.

nhachicha:
still flaky :/

cmelchior:
The CI gods are finally happy @nhachicha 😉

Issue #3238, opened 2016-08-02
Allow 0-length values array on RealmQuery.in() 


dalinaum:
Allow 0-length values array on RealmQuery.in() 

> cmelchior: Should we allow the 0-length values array? It would basically default to the query always returning the empty set, but I would imagine that this values array in most cases will be generated in a loop based on some other input, so I think that this array in a lot of cases will actually be empty, making it a hassle for the user to work around.
> 
> bdash: In Cocoa we have [`TrueExpression` and `FalseExpression`](https://github.com/realm/realm-cocoa/blob/30947f35a115db9e62e208a2ee2c2b75a3009e9b/Realm/RLMQueryUtil.mm#L88-L114) that we use for cases like [handling an `or` group containing zero elements](https://github.com/realm/realm-cocoa/blob/30947f35a115db9e62e208a2ee2c2b75a3009e9b/Realm/RLMQueryUtil.mm#L727-L732).

from [previous discussion](https://github.com/realm/realm-java/pull/3133#discussion_r73099630)

cmelchior:
Folding this into https://github.com/realm/realm-java/issues/4011

Zhuinden:
Also I'd be surprised if it just started not returning anything.

I'd rather expect it to do a no-op.

PR #5898, opened 2018-04-13
Support empty input to in and alwaysTrue and alwaysFalse

### LINKED ISSUE #4011: Why not return an empty realmResults when using "in" query with empty values instead of throwing an exception?
#### Goal

Obtain a realmResults using the "in" query operator on where the values list is empty.
Currently this throws an exception that has to be handled.

#### Expected Results

In set theory, the empty set already is comply by the full set or the empty set, so instead of throwing an exception return an empty realmresults

#### Actual Results
An exception

cmelchior:
Closes #4011

This PR will make our `in()` operator mirror Cocoas equivalent where parsing in no values and treat it as "always false".

At the same time I added support for two new predicates: `alwaysTrue()` and `alwaysFalse()`
`alwaysFalse()` is used by `in()`. `alwaysTrue()` isn't terrible useful outside generic code going down a branch in a 'or()` expression, but it also provides symmetri. See https://github.com/realm/realm-java/issues/3238#issue-168877473

Naming the operators wasn't that easy, but in the end, I opted for the naming also used in Guava https://google.github.io/guava/releases/snapshot/api/docs/com/google/common/base/Predicates.html

It seems more descriptive in our query language than both `.trueExpression()` or `.truePredicate()`

### DIFF
--- a/realm/realm-library/src/main/java/io/realm/RealmQuery.java
+++ b/realm/realm-library/src/main/java/io/realm/RealmQuery.java
@@ -521,8 +521,8 @@ public class RealmQuery<E> {
      * In comparison. This allows you to test if objects match any value in an array of values.
      *
      * @param fieldName the field to compare.
-     * @param values array of values to compare with. If {@code null} or the empty is parsed in the query will never
-     *               match any results. and it cannot be null or empty.
+     * @param values array of values to compare with. If {@code null} or the empty array is provided the query will never
+     *               match any results.
      * @return the query object.
      * @throws java.lang.IllegalArgumentException if the field isn't a String field.
      */
@@ -534,8 +534,8 @@ public class RealmQuery<E> {
      * In comparison. This allows you to test if objects match any value in an array of values.
      *
      * @param fieldName the field to compare.
-     * @param values array of values to compare with. If {@code null} or the empty is parsed in the query will never
-     *               match any results. and it cannot be null or empty.
+     * @param values array of values to compare with. If {@code null} or the empty array is provided the query will never
+     *               match any results.
      * @param casing how casing is handled. {@link Case#INSENSITIVE} works only for the Latin-1 characters.
      * @return the query object.
      * @throws java.lang.IllegalArgumentException if the field isn't a String field.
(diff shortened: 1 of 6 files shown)


realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -522,31 +521,31 @@ public boolean isValid() {
      * In comparison. This allows you to test if objects match any value in an array of values.
      *
      * @param fieldName the field to compare.
-     * @param values array of values to compare with and it cannot be null or empty.
+     * @param values array of values to compare with. If {@code null} or the empty is parsed in the query will never
+     *               match any results. and it cannot be null or empty.
```
nhachicha: remove old description 
> and it cannot be null or empty.

_same  remark for other methods_

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -522,31 +521,31 @@ public boolean isValid() {
      * In comparison. This allows you to test if objects match any value in an array of values.
      *
      * @param fieldName the field to compare.
-     * @param values array of values to compare with and it cannot be null or empty.
+     * @param values array of values to compare with. If {@code null} or the empty is parsed in the query will never
```
nhachicha: If {@ null} or empty is passed, the query will never ...

_same  remark for other methods_

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -1925,6 +1936,24 @@ public long count() {
         return this;
     }
 
+    /**
+     * This predicate will always match.
+     */
+    public RealmQuery<E> alwaysTrue() {
```
nhachicha: is it necessary to expose these as public API? what's the potential use case?

nhachicha: I assume this returns all parents when used with `LinkingObjects`?

cmelchior: Yes. From what I can tell from the NSPredicate docs, this is equivalent to `1 == 1` or something like that, and is only really useful in a subquery where you need to have _something_ to be semantically correct, i.e.

```
<query> OR _something_
```

PR #5902, opened 2018-04-17
Migration that changes primary key deletes rows

### LINKED ISSUE #5899: Primary key migration from optional to required deleting rows
Hi guys, i just encountered a weird problem and don't know if is a bug or i just don't know how to
migrate things. 
I found a workaround for that, create new field with required attribute, transform to get the value of the old primary key, remove old primary key and set primary key to the newly created field.

#### Goal
Change primary key name and required to true (Kotlin non-null but not @Required)

#### Expected Results
The primary key with the new name and the field to be required

#### Actual Results
Migration removing 4 of 5 rows in the table after seting setRequired(Y, true)
I checked the operation before and all the rows have the primary key with a value and are not null

Didn't check with another project but in my case was:
Have a realm object schema with nullable primary key.
Multiple objects of that type with values in the primary key.
Make a migration changing the name and required to true
The new table contains only one row and not the 5 before the migration

#### Code Sample

```java

```

Realm version(s): ?
4.3.3
No
Android Studio version: ?
3.1.1
Which Android version and device: ?
7.0

cmelchior:
Closes #5899 

Root cause:
When changing nullability status on a field, we actually do it by creating a new column, move all data there and then delete the old column. This is a problem in the context of primary keys for the following reason:

Use case: Change primary key int column from nullable to required.

1) create tmp int column: All rows will have the 0 value default value (which actually violates the pk constraint)

2) Start moving values from the old column to the new row by row. This is done using a special `set_unique` instruction. If 0 was present in the old column, we will now do a `set_unique_int(col, row, 0)`, but since all rows have the 0 value, the result is that the rows are collapsed causing data loss.

Solution:
Since changing nullability is only supported for non-sync Realms anyway, there shouldn't be any harm i in not using the `set_unique` method, since the pk invariant is already is already guaranteed for the old column. We just need to be 100% that this method is never called when using Sync.

### DIFF
--- a/realm/realm-library/src/main/cpp/io_realm_internal_Table.cpp
+++ b/realm/realm-library/src/main/cpp/io_realm_internal_Table.cpp
@@ -215,10 +215,13 @@ JNIEXPORT jboolean JNICALL Java_io_realm_internal_Table_nativeIsColumnNullable(J
 // 4d. the column to be converted will index shifted one place to column_index + 1
 // 5. search indexing must be preserved
 // 6. removing the original column and renaming the temporary column will make it look like original is being modified
+//
+// WARNING: These methods do NOT work on primary key columns if the Realm is synchronized.
+//
 
 // Converts a table to allow for nullable values
 // Works on both normal table columns and sub tables
-static void convert_column_to_nullable(JNIEnv* env, Table* old_table, size_t old_col_ndx, Table* new_table, size_t new_col_ndx, bool is_primary_key)
+static void convert_column_to_nullable(JNIEnv* env, Table* old_table, size_t old_col_ndx, Table* new_table, size_t new_col_ndx)
 {
     DataType column_type = old_table->get_column_type(old_col_ndx);
     if (old_table != new_table) {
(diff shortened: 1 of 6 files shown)


cmelchior:
Ready for review @nhachicha

realm/realm-library/src/main/cpp/io_realm_internal_Table.cpp
```diff
@@ -215,10 +215,13 @@ JNIEXPORT jboolean JNICALL Java_io_realm_internal_Table_nativeIsColumnNullable(J
 // 4d. the column to be converted will index shifted one place to column_index + 1
 // 5. search indexing must be preserved
 // 6. removing the original column and renaming the temporary column will make it look like original is being modified
+//
+// WARNING: These methods do NOT work on primary key columns if the Realm is synchronized.
```
nhachicha: maybe add an assertion that throws if called from sync
```C++
#if REALM_ENABLE_SYNC
REALM_ASSERT(false);
#endif
```

cmelchior: Good idea 👍

PR #5907, opened 2018-04-19
Add support for resetting password and email confirmation.

### LINKED ISSUE #5821: Add wrappers for new password provider methods
We’ve added a few additional methods in ROS that allow people to implement password reset and email confirmation flows.
While these are simple HTTP API, it’d be nice to create wrappers for them. All of these rely on the idea that the
username/password provider is actually used as email/password (which we probably have to highlight in the method docs).

The url for all actions is: `POST https://server-url/auth/password/updateAccount`.

The body looks like:

```js
{
    "provider_id":"user's email - for some actions it's optional",
    "data": {
        "action":"%ACTION%",
        // Other fields that depend on action
    }
}
```

All return an empty response and will not provide indication if the user doesn't exist, but will return errors if
some of the parameters are invalid.

1. Request password reset: a reset password email is sent to the user's email with some opaque token.
    - `provider_id` is **required**
    - `action` is `reset_password`
    - **No** other fields in `data`
2. Confirm password reset: after obtaining the token from the email, the user can complete the password reset flow.
By default, ROS will send an email that will redirect to `http://ros-url/some-page?token=%TOKEN%` where a generic
UI will be shown to complete the password reset. However, it's possible to override this URL and e.g. change it to
`myapp://reset-password/%TOKEN%` so that it's handled by the app itself. It's the developer's responsibility to obtain
the token if they want to complete the reset flow within the app.
    - `provider_id` should not be provided
    - `action` is `complete_reset`
    - `data` contains the following additional fields:
        - `token`: the token sent to the user's email address.
        - `new_password`: the new password of the user.
3. Initiate email confirmation: for a lot of use cases, people may want to know that the user has confirmed their email
before allowing them to take certain actions in the app. An email will be sent to the user's email address with an opaque token.
    - `provider_id` is **required**
    - `action` is `request_email_confirmation`
    - **No** other fields in `data`
4. Confirm email: similarly to the password reset flow, the user obtains the token from the email somehow any must pass
it to the server. By default, we'll redirect to a generic page hosted in ROS.
    - `provider_id` should not be provided
    - `action` is `confirm_email`
    - `data` contains the following additional fields:
        - `token`: the token sent to the user's email address.
    - **Note**: After the email is confirmed, the json response after logging in with username/password will contain `isEmailConfirmed` field set to `true`. We should probably surface that on the `User` object.

Implementation/testing notes: This depends on https://github.com/realm/realm-object-server-private/pull/920 being merged and released.

cmelchior:
Closes #5821 

TODO
- [x] Tests
- [x] Public Javadoc
- [x] Run tests

Tests are currently disabled and waiting for #5909 to complete before we can enable them. They have all been run manually and succesfully against a cloud instance.

### DIFF
--- a/realm/realm-library/src/objectServer/java/io/realm/internal/network/OkHttpAuthenticationServer.java
+++ b/realm/realm-library/src/objectServer/java/io/realm/internal/network/OkHttpAuthenticationServer.java
@@ -39,6 +39,7 @@ public class OkHttpAuthenticationServer implements AuthenticationServer {
     private static final String ACTION_LOGOUT = "revoke"; // Auth end point for logging out users
     private static final String ACTION_CHANGE_PASSWORD = "password"; // Auth end point for changing passwords
     private static final String ACTION_LOOKUP_USER_ID = "users/:provider:/:providerId:"; // Auth end point for looking up user id
+    private static final String ACTION_UPDATE_ACCOUNT = "auth/password/updateAccount"; // Password reset and email confirmation
 
     private final OkHttpClient client = new OkHttpClient.Builder()
             .connectTimeout(15, TimeUnit.SECONDS)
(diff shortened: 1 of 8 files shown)


realm/realm-library/src/objectServer/java/io/realm/internal/network/OkHttpAuthenticationServer.java
```diff
@@ -176,6 +217,18 @@ private LookupUserIdResponse lookupUserId(URL lookupUserIdUrl, String authToken)
         return LookupUserIdResponse.from(response);
     }
 
+    private UpdateAccountResponse updateAccount(URL updateAccountUrl, String requestBody) throws Exception {
+        RealmLog.debug("Network request (updateAccount): " + updateAccountUrl);
+        Request request = newAuthRequest(updateAccountUrl)
+                .post(RequestBody.create(JSON, requestBody))
+                .build();
+        RealmLog.error(updateAccountUrl.toString());
```
nhachicha: debugging logging? if not the level should be `log` not `error`

realm/realm-library/src/objectServer/java/io/realm/internal/network/OkHttpAuthenticationServer.java
```diff
@@ -176,6 +217,18 @@ private LookupUserIdResponse lookupUserId(URL lookupUserIdUrl, String authToken)
...
+        RealmLog.error(requestBody);
```
nhachicha: ditto

realm/realm-library/src/objectServer/java/io/realm/internal/network/UpdateAccountRequest.java
```diff
@@ -0,0 +1,80 @@
+/*
+ * Copyright 2016 Realm Inc.
```
nhachicha: 2018

realm/realm-library/src/objectServer/java/io/realm/internal/network/UpdateAccountResponse.java
```diff
@@ -0,0 +1,44 @@
+package io.realm.internal.network;
```
nhachicha: missing licence

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/AuthTests.java
```diff
@@ -824,4 +824,120 @@ public void onError(ObjectServerError error) {
...
     }
+
+    @Test
+    @RunTestInLooperThread
+    @Ignore("Depends on https://github.com/realm/realm-java/pull/5909")
+    public void requestPasswordResetAsync() {
+        String email = "foo@bar.baz";
+        UserFactory.createUser(email).logOut();
+
+        // Currently no easy way to see if we actually get an email.
```
nhachicha: In theory, we can if we configure ROS with the emailHandler to point to `https://ethereal.email` then query the SMTP server to fetch the different emails to retrieve the different tokens, but I guess this is out of scope ...

PR #5950, opened 2018-05-14
Post size metrics to our own InfluxDB server

emanuelez:
Data will be transferred from the old instance as well


cmelchior:
Merged @emanuelez

PR #5952, opened 2018-05-15
Improved performance when parsing field descriptions

### LINKED ISSUE #5547: Optimization: Realm doing regex string split and String[] allocation on every .equalTo()
For a simple query such as `r.where(My.class).equalTo("id", 1).findfirst() `there is a cost of a regex string causing a String[] allocation even if the field in question doesn't contain any "." separators. 

```java
    private List<String> parseFieldDescription(String fieldDescription) {
        if (fieldDescription == null || fieldDescription.equals("")) {
            throw new IllegalArgumentException("Invalid query: field name is empty");
        }
        if (fieldDescription.endsWith(".")) {
            throw new IllegalArgumentException("Invalid query: field name must not end with a period ('.')");
        }
        return Arrays.asList(fieldDescription.split("\\."));
    }
```
```java
parseFieldDescription:272, FileDescriptor (regex split + String[] allocation)
...
getColumnIndices:428, RealmObjectSchema
...
equalTo:385 RealmQuery
```

As it is there is no apparent issue but imagine if the small query is in a scrolling RecyclerView done on every onBindView() pass. Now, we have an unnecessary allocation problem in a tight loop. I would suggest:

1) avoid any parsing if we know ahead of time this field cannot contain ".". I believe the transformer can keep a mapped cache of ALL mapped fields that do contain a "." and any negative match against it will skip this parsing code.

2) Cache the parseFieldDescription in a map and only dynamically parse on first attempt for this Realm.class field.

cmelchior:
Fixes #5547

### DIFF
--- a/realm/realm-library/src/main/java/io/realm/internal/fields/FieldDescriptor.java
+++ b/realm/realm-library/src/main/java/io/realm/internal/fields/FieldDescriptor.java
@@ -21,6 +21,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Locale;
 import java.util.Set;
+import java.util.regex.Pattern;
 
 import io.realm.RealmFieldType;
 import io.realm.internal.ColumnInfo;
@@ -38,6 +39,9 @@ import io.realm.internal.Table;
  * </ul>
  */
 public abstract class FieldDescriptor {
+
+    private static final Pattern SEPERATOR = Pattern.compile("\\.");
+
     public interface SchemaProxy {
         boolean hasCache();
 
@@ -273,10 +277,17 @@ public abstract class FieldDescriptor {
         if (fieldDescription == null || fieldDescription.equals("")) {
             throw new IllegalArgumentException("Invalid query: field name is empty");
         }
-        if (fieldDescription.endsWith(".")) {
+
+        int lastDotIndex = fieldDescription.lastIndexOf(".");
+        if (lastDotIndex == fieldDescription.length() - 1) {
             throw new IllegalArgumentException("Invalid query: field name must not end with a period ('.')");
         }
-        return Arrays.asList(fieldDescription.split("\\."));
+
+        if (lastDotIndex > -1) {
+            return Arrays.asList(SEPERATOR.split(fieldDescription));
+        } else {
+            return Collections.emptyList();
+        }
     }
 
     private void verifyColumnType(String className, String columnName, RealmFieldType columnType, Set<RealmFieldType> validTypes) {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 5.1.1 (YYYY-MM-DD)
+
+### Enhancements
+
+* Improved speed and allocations when parsing field descriptions in queries (#5547).
+
 ## 5.1.0 (2018-04-25)
 
 ### Enhancements


bmunkholm:
Please add measurement to demonstrate the effects.

cmelchior:
This code compares the two approaches:

```
package io.realm.benchmarks;

import org.junit.runner.RunWith;

import java.util.ArrayList;
import java.util.Arrays;
import java.util.List;
import java.util.regex.Pattern;

import dk.ilios.spanner.AfterExperiment;
import dk.ilios.spanner.BeforeExperiment;
import dk.ilios.spanner.Benchmark;
import dk.ilios.spanner.BenchmarkConfiguration;
import dk.ilios.spanner.SpannerConfig;
import dk.ilios.spanner.junit.SpannerRunner;
import io.realm.benchmarks.config.BenchmarkConfig;

@RunWith(SpannerRunner.class)
public class ParseBenchmarks {

    @BenchmarkConfiguration
    public SpannerConfig configuration = BenchmarkConfig.getConfiguration(this.getClass().getCanonicalName());
    private OriginalParser oldParser;
    private NewParser newParser;
    private String shortField;
    private String mediumField;
    private String longField;


    @BeforeExperiment
    public void before() {
        oldParser = new OriginalParser();
        newParser = new NewParser();
        shortField = "foo";
        mediumField = "name.id";
        longField = "verylongfield.withverylong.linked.field.names";
    }

    @AfterExperiment
    public void after() {
    }

    @Benchmark
    public void old_short(long reps) {
        List<String> result;
        for (long i = 0; i < reps; i++) {
            result = oldParser.parse(shortField);
        }
    }

    @Benchmark
    public void old_medium(long reps) {
        List<String> result;
        for (long i = 0; i < reps; i++) {
            result = oldParser.parse(mediumField);
        }
    }

    @Benchmark
    public void old_long(long reps) {
        List<String> result;
        for (long i = 0; i < reps; i++) {
            result = oldParser.parse(longField);
        }
    }

    @Benchmark
    public void new_short(long reps) {
        List<String> result;
        for (long i = 0; i < reps; i++) {
            result = newParser.parse(shortField);
        }
    }

    @Benchmark
    public void new_medium(long reps) {
        List<String> result;
        for (long i = 0; i < reps; i++) {
            result = newParser.parse(mediumField);
        }
    }

    @Benchmark
    public void new_long(long reps) {
        List<String> result;
        for (long i = 0; i < reps; i++) {
            result = newParser.parse(longField);
        }
    }

    private interface Parser {
        List<String> parse(String input);
    }

    public static class OriginalParser implements Parser {
        @Override
        public List<String> parse(String fieldDescription) {
            if (fieldDescription == null || fieldDescription.equals("")) {
                throw new IllegalArgumentException("Invalid query: field name is empty");
            }
            if (fieldDescription.endsWith(".")) {
                throw new IllegalArgumentException("Invalid query: field name must not end with a period ('.')");
            }
            return Arrays.asList(fieldDescription.split("\\."));
        }
    }

    public static class NewParser implements Parser {

        private static final Pattern FIELD_SEPARATOR = Pattern.compile("\\.");

        @Override
        public List<String> parse(String fieldDescription) {
            if (fieldDescription == null || fieldDescription.equals("")) {
                throw new IllegalArgumentException("Invalid query: field name is empty");
            }

            int lastDotIndex = fieldDescription.lastIndexOf(".");
            if (lastDotIndex == fieldDescription.length() - 1) {
                throw new IllegalArgumentException("Invalid query: field name must not end with a period ('.')");
            }

            if (lastDotIndex > -1) {
                return Arrays.asList(FIELD_SEPARATOR.split(fieldDescription));
            } else {
                ArrayList<String> result = new ArrayList<>(1);
                result.add(fieldDescription);
                return result;
            }
        }
    }
}
```

image

Note that for medium/large strings they have similar performance but for the most common short fields there is a big difference.

realm/realm-library/src/main/java/io/realm/internal/fields/FieldDescriptor.java
```diff
@@ -286,7 +287,9 @@ protected final void setCompilationResults(
         if (lastDotIndex > -1) {
             return Arrays.asList(FIELD_SEPARATOR.split(fieldDescription));
         } else {
-            return Collections.emptyList();
+            ArrayList<String> result = new ArrayList<>(1);
```
Zhuinden: Is it possible to use `return Collections.singletonList(fieldDescription)` here? Or does this have to be mutable?

cmelchior: No, it doesn't. I didn't know that class existed, nice 👍

bmunkholm:
Since it's not making things slower, I don't see a problem merging it. But since it's only getting twice as fast in the optimal case, I'm not sure if it actually fixes the users problem. Perhaps it does as you are not measuring gc impact here I assume.

PR #5957, opened 2018-05-16
Add executeTransactionCompletable method

### LINKED ISSUE #2126: async executeTransaction with Oberservables
It would be nice to have an async executeTransaction method in Realm-class which returns an Observable since we RxJava-users don't like callables ;)
Maybe something like:
```
Observable<void> executeTransaction(Realm.Transaction transaction)
```
Usage would be:
```
Observable<void> transaction = realm.executeTransaction(new Realm.Transaction() {
            @Override
            public void execute(Realm bgRealm) {
                User user = bgRealm.createObject(User.class);
                user.setName("John");
                user.setEmail("john@corporation.com");
            }
        });
transaction.subscribe(new rx.Subscriber<T> {
           @Override public void onCompleted() {
                 //do your stuff here, if transaction was successful
            }

           @Override public void onError(Throwable e) {
                 //cleanup
            }

            @Override public void onNext(T t) {
                 //nothing.
            }
        });
```

jereksel:
fixes #2126

DynamicRealm's transactions are synchronous only, so I've only added Completable support for Realm.

### DIFF
--- a/realm/realm-library/src/main/java/io/realm/Realm.java
+++ b/realm/realm-library/src/main/java/io/realm/Realm.java
@@ -48,6 +48,7 @@ import java.util.concurrent.atomic.AtomicInteger;
 
 import javax.annotation.Nullable;
 
+import io.reactivex.Completable;
 import io.reactivex.Flowable;
 import io.realm.annotations.Beta;
 import io.realm.exceptions.RealmException;
@@ -74,6 +75,7 @@ import io.realm.internal.Util;
 import io.realm.internal.annotations.ObjectServer;
 import io.realm.internal.async.RealmAsyncTaskImpl;
 import io.realm.log.RealmLog;
+import io.realm.rx.RxObservableFactory;
 import io.realm.sync.permissions.ClassPermissions;
 import io.realm.sync.permissions.ClassPrivileges;
 import io.realm.sync.permissions.RealmPermissions;
@@ -1584,6 +1586,18 @@ public class Realm extends BaseRealm {
         });
 
         return new RealmAsyncTaskImpl(pendingTransaction, asyncTaskExecutor);
+    }
+
+    /**
+     * Invokes transaction on worker thread and returns Completable
+     *
+     * @param transaction {@link io.realm.Realm.Transaction} to execute.
+     * @return {@link Completable} representing asynchronous transaction
+     * @throws UnsupportedOperationException if the required RxJava framework is not on the classpath.
+     */
+    public Completable executeTransactionCompletable(final Transaction transaction) {
+        RxObservableFactory factory = configuration.getRxFactory();
+        return factory.from(this, transaction);
     }
 
     /**
--- a/realm/realm-library/src/main/java/io/realm/Realm.java
+++ b/realm/realm-library/src/main/java/io/realm/Realm.java
@@ -1589,13 +1589,13 @@ public class Realm extends BaseRealm {
     }
 
     /**
-     * Invokes transaction on worker thread and returns Completable
+     * Invokes the transaction on a worker thread wrapping the result as a {@link Completable}.
      *
      * @param transaction {@link io.realm.Realm.Transaction} to execute.
-     * @return {@link Completable} representing asynchronous transaction
+     * @return {@link Completable} representing the asynchronous transaction
      * @throws UnsupportedOperationException if the required RxJava framework is not on the classpath.
      */
-    public Completable executeTransactionCompletable(final Transaction transaction) {
+    public Completable asCompletable(final Transaction transaction) {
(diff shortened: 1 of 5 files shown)


realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -1586,6 +1588,18 @@ public void run() {
...
     }
 
+    /**
+     * Invokes transaction on worker thread and returns Completable
+     *
+     * @param transaction {@link io.realm.Realm.Transaction} to execute.
+     * @return {@link Completable} representing asynchronous transaction
+     * @throws UnsupportedOperationException if the required RxJava framework is not on the classpath.
+     */
+    public Completable executeTransactionCompletable(final Transaction transaction) {
```
cmelchior: I'm not 100% sure about this name. From the discussion in https://github.com/realm/realm-java/issues/2126#issuecomment-196844283 it seems that this should be renamed to `asCompletable(final Transaction transaction). This also seems to be more in line with the naming from e.g `RealmResults.asFlowable()`.

Or did you have some other reasoning for naming it this?

@nhachicha do you have any thoughts on this?

If we indeed end up wrapping an async transaction `asCompleteteAsync` would propably fit our current naming scheme better as well...Naming is hard :smile:

realm/realm-library/src/main/java/io/realm/rx/RealmObservableFactory.java
```diff
@@ -592,6 +593,11 @@ public void run() {
         throw new RuntimeException("RealmQuery not supported yet.");
     }
 
+    @Override
+    public Completable from(Realm realm, Realm.Transaction transaction) {
+        return Completable.create(e -> realm.executeTransactionAsync(transaction, e::onComplete, e::onError));
```
cmelchior: I'm a bit torn whether we should use `executeTransaction` or `executeTransactionAsync` here. If we use `executeTransaction()` we force people to manually use `subscribeOn`, but somehow that feels more idiomatic. The problem with using Realms internal async system is that it e.g. becomes impossible to swap schedulers when testing.

Did you have any reason for choosing this over the synchronous variant?

In any case, if we keep this, we need to update the doc to say it throws if called from a non-looper thread.

@nhachicha / @Zhuinden Do you have any thoughts?

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -1586,6 +1588,18 @@ public void run() {
         return new RealmAsyncTaskImpl(pendingTransaction, asyncTaskExecutor);
     }
 
+    /**
+     * Invokes transaction on worker thread and returns Completable
+     *
+     * @param transaction {@link io.realm.Realm.Transaction} to execute.
+     * @return {@link Completable} representing asynchronous transaction
```
cmelchior: representing the

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -1586,6 +1588,18 @@ public void run() {
         return new RealmAsyncTaskImpl(pendingTransaction, asyncTaskExecutor);
     }
 
+    /**
+     * Invokes transaction on worker thread and returns Completable
```
cmelchior: `Invokes the transaction on a worker thread wrapping the result as a {@link Completable}.`

realm/realm-library/src/main/java/io/realm/rx/RealmObservableFactory.java
Zhuinden: My thought is that this is a breaking change for anyone who implemented their custom `RxObservableFactory`.

But then again, it follows the https://github.com/realm/realm-java/issues/2126#issuecomment-196844283 proposition directly.

I think it should be possible to use `executeTransaction` vs `executeTransactionAsync` depending on whether the Realm is on a looper thread, which you can verify with `realm.isAutoRefresh()` or more-so the internal "capabilities" class should know whether this thread is able to use async transactions.

nhachicha: If you use `executeTransaction ` then you need to make sure the Realm you're using will be on the same scheduler's Thread (`subscribeOn`). I think `executeTransactionAsync` is appropriate here

realm/realm-library/src/main/java/io/realm/Realm.java
nhachicha: `asCompletable` feels more aligned with current API

jereksel:
@cmelchior I've fixed all comments

idok595:
Hey There is no handling for the RealmAsyncTask! i think that when the Disoposable dispoded before it complete it is require to cancel the RealmAsyncTask have been created by the call "executeTransactionAsync"

PR #5965, opened 2018-05-22
Upgrade build tool dependencies.

cmelchior:


cmelchior:
For some reason the projects doesn't compile with `3.1.2`. I'm not sure why 😒

cmelchior:
Same problem exists with 3.1.3:
```
* What went wrong:

attempt to unlock read lock, not locked by current thread
```

cmelchior:
The above was caused by a bug in the Android plugin: https://issuetracker.google.com/issues/80464216 . I added a workaround and upgraded to latest stable version of the Android plugin and Gradle. Upgrading Gradle caused a few other minor issues, so had to completely remove Groovy from the Realm-transformer (not a big loss since it was just used for tests).

PR #6000, opened 2018-06-07
Add getRealm methods

### LINKED ISSUE #5997: Feature Request: RealmQuery.getRealm()
#### Goal

Write an extension function `fun <T: RealmModel> RealmQuery<T>.findAllCopied(depth: Int = 4) = realm.copyFromRealm(this, depth)`.

#### Expected Results

This should be possible

#### Actual Results

`fun <T: RealmModel> RealmQuery<T>.findAllCopied(realm: Realm, depth: Int = 4) = realm.copyFromRealm(this, depth)`

which results in

    realm.where<Dog>().findAllCopied(realm)

Instead of 

    realm.where<Dog>().findAllCopied()

Realm version(s): 5.2.0

cmelchior:
Closes #5997 

We already had `RealmObject.getRealm()` this adds the same methods to `RealmList`, `RealmResults`, `RealmQuery` and `OrderedRealmCollectionSnapshot`.

### DIFF
--- a/realm/realm-library/src/main/java/io/realm/RealmQuery.java
+++ b/realm/realm-library/src/main/java/io/realm/RealmQuery.java
@@ -1951,6 +1951,28 @@ public class RealmQuery<E> {
         return this;
     }
 
+    /**
+     * Returns the {@link Realm} instance to which this query belongs.
+     * <p>
+     * Calling {@link Realm#close()} on the returned instance is discouraged as it is the same as
+     * calling it on the original Realm instance causing the Realm to fully close invalidating the
+     * query.
+     *
+     * @return {@link Realm} instance this query belongs to.
+     * @throws IllegalStateException if the Realm is an instance of {@link DynamicRealm} or the
+     * {@link Realm} was already closed.
+     */
+    public Realm getRealm() {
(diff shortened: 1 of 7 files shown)


Zhuinden:
Thanks!

Although it might also make sense to add it for `OrderedRealmCollectionSnapshot`.

cmelchior:
Actually, its being added there as well since `OrderedRealmCollectionImpl.java` is the base class for that one as well. I should clarify the docs about it though 😄

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -1951,6 +1951,28 @@ public long count() {
         return this;
     }
 
+    /**
+     * Returns the {@link Realm} instance to which this query belongs.
+     * <p>
+     * Calling {@link Realm#close()} on the returned instance is discouraged as it is the same as
+     * calling it on the original Realm instance causing the Realm to fully close invalidating the
```
nhachicha: > causing the Realm

which may cause the Realm ...

_same comment for the other impl (RealmList.java etc)_

PR #6013, opened 2018-06-10
Removed volatile keyword

ZherebtsovAlexandr:
Why volatile variable is here?

Look like the variable using only in synchronized block. Synchronized block ensure atomicity and visibility for the variable in this case https://docs.oracle.com/javase/specs/jls/se7/html/jls-17.html#jls-17.4.5 .


Zhuinden:
Probably to ensure that even if you call `Realm.init()` on multiple threads, it is still loaded only once

ZherebtsovAlexandr:
Synchronized can ensure thread-safe modifications for the variable. Synchronized is stronger synchronization than volatile. Or I don't see cases?

Zhuinden:
There could easily be no cases, `RealmCore` class is quite old :smile:

PR #6025, opened 2018-06-15
Remove reliance on groovy in realm-transformer

kageiit:
Ref #3971


realm-transformer/src/main/java/io/realm/transformer/Utils.java
```diff
@@ -57,4 +64,24 @@ public static String hexStringify(byte[] data) {
...
+    public static String getTargetSdk(Project project) {
+        return getAndroidExtension(project).getDefaultConfig().getTargetSdkVersion().getApiString();
+    }
+
+    public static String getMinSdk(Project project) {
+        return getAndroidExtension(project).getDefaultConfig().getMinSdkVersion().getApiString();
+    }
+
+    public static boolean isSyncEnabled(@Nonnull Project project) {
+        return ((RealmPluginExtension) project.getExtensions().getByName("realm")).getSyncEnabled();
```
kageiit: This is the only compilation issue right now. `RealmPluginExtension` needs to be moved to a new gradle module so it can be a shared dependency between the transformer and the plugin. It can also be moved to the transformer module as the plugin depends on it already.

@cmelchior Whats the preferred option?

cmelchior: It would feel cleaner to move it to a new module, but I doubt it is worth it, so putting it in the transformer is fine I think 👍

realm-transformer/src/main/java/io/realm/transformer/Utils.java
```diff
@@ -57,4 +64,24 @@ public static String hexStringify(byte[] data) {
 
         return stringBuilder.toString();
     }
+
+    public static String getTargetSdk(Project project) {
+        return getAndroidExtension(project).getDefaultConfig().getTargetSdkVersion().getApiString();
```
cmelchior: Wow, I didn't realise it was this easy. Cool 👏 💯

cmelchior:
Running the tests locally seem to cause failures in two places:

```
~/Realm/kageiit-realm-java/gradle-plugin (cm)$ 
 => gw test
Using gradle at '/Users/cm/Realm/kageiit-realm-java/gradle-plugin/gradlew' to run buildfile '/Users/cm/Realm/kageiit-realm-java/gradle-plugin/build.gradle':

Build cache is an incubating feature.

> Task :test 

io.realm.gradle.PluginTest > pluginAddsRightDependencies FAILED
    org.gradle.api.internal.plugins.PluginApplicationException at PluginTest.groovy:64
        Caused by: org.gradle.api.GradleException at PluginTest.groovy:64
            Caused by: java.lang.VerifyError at PluginTest.groovy:64

2 tests completed, 1 failed
```

and 

```
~/Realm/kageiit-realm-java/examples (cm)$ 
 => gw assemble

* Where:
Build file '/Users/cm/Realm/kageiit-realm-java/examples/architectureComponentsExample/build.gradle' line: 2

* What went wrong:
A problem occurred evaluating project ':realm-examples:architectureComponentsExample'.
> Failed to apply plugin [id 'realm-android']
   > Could not generate a proxy class for class io.realm.gradle.RealmPluginExtension.
```

I'll see if I can find the time in the next couple of days to figure out exactly why these fail this way.

kageiit:
@cmelchior Fixed tests now. Let me know if anything else needs fixing!

cmelchior:
Thanks a ton for doing this. Merged.

Zhuinden:
Does this officially allow Realm to work with OkBuck? :smile:

kageiit:
> Does this officially allow Realm to work with OkBuck?

It should. Feel free to give it a try on the next release :)

PR #6043, opened 2018-06-26
Fixes #5970

### LINKED ISSUE #5970: Test root certificates have expired
The test certificates used in https://github.com/realm/realm-java/blob/master/realm/realm-library/src/androidTestObjectServer/java/io/realm/TrustManagerCertificateValidationTests.java
have expired causing the tests to fail. They need to be replaced

https://github.com/realm/realm-java/commit/1ae66530162b965e3b00588d02560c1cf530068f

nhachicha:
Using Android Network Security Configuration to install the custom root CA for tests, since user-added CAs for apps targeting API level 24 will—by design—not honor such CAs unless the app explicitly opts in.
https://android-developers.googleblog.com/2016/07/changes-to-trusted-certificate.html


nhachicha:
@emanuelez cc/ we can now revert to use the other Android devices (>=6.0) for CI

Issue #6056, opened 2018-07-07
Critical bug in Realm 5.3.1, instances returned by Realm (by using findFirst(), findAll(), etc.) change to null


m-Bilal:
Okay, so my guess is that this one bug is causing a lot of unexpected behaviour in Realm 5.3.1. The following is what I observed:
1) realm.createObject(<Class Name>.class, int id) saves all values except for the id as null or 0. (realm.copyToRealm() works.)
2) Any query (such as findAll(), findFrist(), etc) returns an object(s) with the correct id but all other fields as null.
3) When using a static method to return RealmResults or an Object, the object fields have their correct values inside the function, but change to null once you try using them anywhere else.
Eg:
```Java
    public static RealmResults<SubscriptionTypeModel> getAllSubscriptionTypeModelsFromRealm() {
        Realm realm = Realm.getDefaultInstance();
        RealmResults<SubscriptionTypeModel> realmResults =
                        realm.where(SubscriptionTypeModel.class).findAll();
        for (SubscriptionTypeModel i: realmResults) {
            Log.d(TAG,"getAllSubscriptionModelsFromRealm, id: " + i.id);
            Log.d(TAG,"getAllSubscriptionModelsFromRealm, name: " + i.name);
            Log.d(TAG,"getAllSubscriptionModelsFromRealm, minage: " + i.minAge);
            Log.d(TAG,"getAllSubscriptionModelsFromRealm, maxage: " + i.maxAge);
        }
        return realmResults;
    }
```
The logs in this function will print the correct values. However if you log the returned results in another function of another class, all objects will have the correct id, but all other fields will be either null or 0.

NOTE: To anyone who is facing the same issues, you can downgrade your Realm version to 5.0.1 and Realm will start working fine.

Zhuinden:
I'm guessing you use direct field access instead of getters/setters?

Just checking

m-Bilal:
Yes

nhachicha:
@m-Bilal 
So basically calling `getAllSubscriptionTypeModelsFromRealm` in one class/function works (i.e print the correct values) whereas calling it in another class/method doesn't?

If that's the case, the only thing I suspect is that the byte code changes are not applied correctly for some classes, causing it to return default values. This might be related to the incremental build change introduced recently...

Can you please check what happen if you:
1- Use a clean build? (you still experience the issue?)
2- Using explicitly the getter/setter

m-Bilal:
[...]

PR #6058, opened 2018-07-10
Incremental build causing direct access to model without accessor to fail

nhachicha:
https://github.com/realm/realm-java/issues/6056 revealed an issue with the incremental build that causes the bytecode to not replace access to fields with the generated accessor. 

The issue was in `FieldAccessToAccessorConverterUsingClassPool` which excluded by error a legitmate fields from being processed since `isRealmModelClass(fieldAccess.enclosingClass)` was returing `false`, which is wrong since the `enclosingClass` is not the class that we want to make sure it's a valid Realm model but rather `fieldAccess.className`. In the example [project](https://github.com/realm/realm-java/issues/6056#issuecomment-403865389) this translate into excluding changes to `com.bilal.realmerrorrecreate.MyRealmModel.minAge` field for instance since its enclosing class is `com.bilal.realmerrorrecreate.MainActivity` which is indeed not a Realm model.


softwarejoint:
when will this get merged?

nhachicha:
@softwarejoint as soon as it gets reviewed by @cmelchior probably next week, in the meantime, you can work around the issue by performing a full build.

realm-transformer/src/main/kotlin/io/realm/transformer/ByteCodeModifier.kt
```diff
@@ -162,9 +162,16 @@ class BytecodeModifier {
         @Throws(CannotCompileException::class)
         override fun edit(fieldAccess: FieldAccess) {
             logger.debug("      Field being accessed: ${fieldAccess.className}.${fieldAccess.fieldName}")
-            if (isRealmModelClass(fieldAccess.enclosingClass) && isModelField(fieldAccess.field)) {
+
+            val fieldAccessCtClass: CtClass? = try { classPool.get(fieldAccess.className) } catch (e: NotFoundException) { null }
```
cmelchior: Nice catch 👍

realm-transformer/src/main/kotlin/io/realm/transformer/ByteCodeModifier.kt
```diff
@@ -162,9 +162,16 @@ class BytecodeModifier {
...
+            if (fieldAccessCtClass != null && isRealmModelClass(fieldAccessCtClass) && isModelField(fieldAccess.field)) {
                 logger.debug("        Realm: Manipulating ${ctClass.simpleName}.${behaviour.name}(): ${fieldAccess.fieldName}")
                 logger.debug("        Methods: ${ctClass.declaredMethods}")
+
+                // make sure accessors are added, otherwise javassist will fail with
+                // javassist.CannotCompileException: [source error] realmGet$id() not found in 'foo.Model'
+                addRealmAccessors(fieldAccessCtClass)
```
cmelchior: It concerns me that this needed?
`addRealmAccessors` should already be called by `BuildTemplate.transformModelClasses()`, so if they are missing it seems to indicate we have other flaws in the transformer logic.

The thinking behind the implementation was:
1) For a full build, all model classes will correctly be modified.
2) For a incremental build, the model class might not be modified, but the accessors would have been added when the model class was compiled for the first time.

cmelchior: Also won't this create lots of duplicate method definitions as you are adding Realm accessors for every field access?

nhachicha: > addRealmAccessors should already be called by BuildTemplate.transformModelClasses(),

I think that if the model didn't change, then the transformer will not add the model as part of the classes to be modified, this implies that `outputClassNames` will not contain your model so `transformModelClasses` will not generate the accessors again 


The implementation of [addRealmAccessors](https://github.com/realm/realm-java/blob/8aac7d7dab0ddf35624ca2af7aa987995fcceb1e/realm-transformer/src/main/kotlin/io/realm/transformer/ByteCodeModifier.kt#L45) do check if the accessor was already added, so there shouldn't be a duplicate issue

cmelchior: I did a little debugging because the above explanation seemed wrong somehow. 
It turns out that on subsequent incremental builds you don't see the transformed output of earlier builds but only the original class without any transforms applied. This is the reason we artificially need to add the accessor again, even though it was already done on a previous build.

You can verify this by first doing a full build (which should add the accessors to all model classes) and then do an incremental build which then fails to find the accessor method.

I also verified it by looking at what classes gets passed to the transformer:

```
// This is the directory containing the non-transformed classes
09:50:12.930 [DEBUG] [realm-logger] Adding: /Users/cm/RealmCustomerProjects/RealmErrorRecreate/app/build/intermediates/classes/debug
```

It is rather annoying we have to add the accessors twice, but the accessors are not written to disk the 2nd time around as the model class is only part of the referenced input.

So, for now, the above solution seems to be the right one, but I'll reach out to Google to get clarification on whether or not this is intended behaviour because somehow it doesn't feel completely right to me.

PR #6062, opened 2018-07-16
Fix bug when using distinct() and count()

### LINKED ISSUE #5958: realmQuery.distinct().count() might return incorrect value
A mapping of a Stack Overflow question that is most likely a bug.

#### Goal

Evaluate number of distinct items without having to use `findAll().size()`

#### Expected Results

`realm.where(Blah.class).distinct("birthday").count()` should work

#### Actual Results

Count returns the size of the full table data set.

See https://stackoverflow.com/questions/50363904/realm-giving-wrong-result-count-on-distinct

Realm version(s): 5.1.0

cmelchior:
Fixes #5958

### DIFF
--- a/realm/realm-library/src/androidTest/java/io/realm/RealmQueryTests.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/RealmQueryTests.java
@@ -2279,7 +2279,7 @@ public class RealmQueryTests extends QueryTests {
         assertEquals(TEST_DATA_SIZE, realm.where(AllTypes.class).count());
     }
 
-    // Verify that count works together with using distinct.
+    // Verify that count correctly when using distinct.
     // See https://github.com/realm/realm-java/issues/5958
     @Test
     public void distinctCount() {
--- a/realm/realm-library/src/androidTest/java/io/realm/RealmQueryTests.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/RealmQueryTests.java
@@ -24,6 +24,7 @@ import org.junit.runner.RunWith;
 import java.lang.reflect.Field;
 import java.util.Date;
 import java.util.Locale;
+import java.util.UUID;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.atomic.AtomicInteger;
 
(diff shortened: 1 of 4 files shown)


realm/realm-library/src/androidTest/java/io/realm/RealmQueryTests.java
```diff
@@ -2278,6 +2279,20 @@ public void count() {
         assertEquals(TEST_DATA_SIZE, realm.where(AllTypes.class).count());
     }
 
+    // Verify that count works together with using distinct.
```
nhachicha: with->when?

realm/realm-library/src/main/java/io/realm/internal/TableQuery.java
```diff
@@ -614,6 +614,13 @@ public long count(long start, long end, long limit) {
         return nativeCount(nativePtr, start, end, limit);
     }
 
+    /**
+     * Returns only the number of matching objects.
+     * This method is very fast compared to evaluating a query completely, but it does not
+     * goes around any logic implemented in Object Store and other parts of the API that works
+     * on query results. So it the primary use case for this method is testing.
```
nhachicha: So it the primary -> So the primary

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -1749,8 +1749,12 @@ public Date maximumDate(String fieldName) {
      */
     public long count() {
         realm.checkIfValid();
-
-        return this.query.count();
+        // The fastest way of doing `count()` is going through `TableQuery.count()`. Unfortunately
+        // doing this does not correctly apply all side effects of queries (like subscriptions). Also
```
nhachicha: > ... does not correctly apply all side effects of queries (like subscriptions)

the new implementation doesn't create a subscription either?

cmelchior: You are right, but if we decided to change that in the future going directly to `TableQuery` would not work correctly.

cmelchior:
> Isn't it a bug in Core if Query#count() doesn't return the correct size when used with distinct?

Not really. At the core level, `distinct` is not part of the query API, but is a method on `TableView`. Which is also why it is a lot easier to go through ObjectStore `Results` which will mask that difference.

PR #6064, opened 2018-07-17
Fix analytics not being sent

cmelchior:
This also fixed an invalid cast that wasn't detected before because the code path was never touched.

### DIFF
--- a/realm-transformer/src/main/groovy/io/realm/transformer/GroovyUtil.groovy
+++ b/realm-transformer/src/main/groovy/io/realm/transformer/GroovyUtil.groovy
@@ -38,7 +38,7 @@ class GroovyUtil {
 
     @Nonnull
     static boolean isSyncEnabled(@Nonnull Project project) {
-        return project?.realm?.syncEnabled == true
+        return (project?.hasProperty("realm")) ? project.realm.syncEnabled : false
     }
 
     @Nonnull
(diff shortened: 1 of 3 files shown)


realm-transformer/src/main/kotlin/io/realm/transformer/RealmTransformer.kt
```diff
@@ -147,16 +147,16 @@ class RealmTransformer(val project: Project) : Transform() {
...
+        val packages: Set<String> = outputModelClasses.map {
             it.packageName
-        }
+        }.toSet()
 
         val targetSdk: String? = GroovyUtil.getTargetSdk(project)
         val minSdk: String?  = GroovyUtil.getMinSdk(project)
 
-        if (disableAnalytics) {
+        if (!disableAnalytics) {
```
Zhuinden: Ouch, did that happen before or after https://realm.io/blog/realm-report-q4-2017-rise-of-kotlin/ i wonder

cmelchior: Way after... this happened when the incremental transformer was added, which was in 5.2.0 (June this year).

realm-transformer/src/main/groovy/io/realm/transformer/GroovyUtil.groovy
```diff
@@ -39,7 +38,7 @@ class GroovyUtil {
 
     @Nonnull
     static boolean isSyncEnabled(@Nonnull Project project) {
-        return project.realm?.syncEnabled != null && project.realm.syncEnabled
+        return (project?.hasProperty("realm")) ? project.realm.syncEnabled : false
```
MGaetan89: Why not simply `project.hasProperty("realm")` since `project` is marked as `Nonnull`?

cmelchior: Yes, good point

Issue #5799, opened 2018-03-05
Use Transformer to remove unsupported methods from non-sync builds


cmelchior:
Currently in `5.0.0-BETA` we have a number of methods and classes annotated with the `@ObjectServer`

These methods are kinda like extension methods only relevant for sync builds and should not be there when using non-sync Realm.

I attempted to use Proguard to remove these, but all attempts at doing so have failed so far, so plan B it is.

We should be able to copy our current RealmTransformer and only run it when building release builds. Then we can use JavaAssist to remove any code with a given annotation.

cmelchior:
Methods marked `@ObjectServer` are now being stripped when building the library.

PR #6038, opened 2018-06-22
Add support for Realm.syncSession as a extension method

cmelchior:
Makes it easier in Kotlin to get to the SyncSession.

PR #6069, opened 2018-07-18
Add Build Transformer stripping Sync methods

cmelchior:
Fixes https://github.com/realm/realm-java/issues/5799
Replaces https://github.com/realm/realm-java/pull/6038. We should move the `getSession` method back to Java and annotate it with `@ObjectServer` once this is merged.

This adds a transformer only used when building the Realm library. It will remove all traces of any class/method/field annotated with the `@ObjectServer` annotation when building the `base` variant while keeping them when building the `objectServer` variant. It does so using ASM which is more low level than JavaAssist, but also faster because we don't have to load all classes into a shared classpool.

This allows us to emulate Kotlin extension functions at the Java layer which can make it a lot easier to introduce new sync API's. It should not matter that there are bytecode differences between `base` and `objectServer` libraries as trying to include them both today would already result in `Duplicate class definition` errors.

The transformer does not need to be published on BinTray since it is only used internally when building Realm.

TODO:
- [x] Cleanup
- [x] Verify that Android Studio works correctly
- [x] Verify the byte code
- [x] Add unit test to base checking that classes and methods are stripped
- [x] Proxy class interfaces are not stripped, i.e: `io_realm_sync_permissions_ClassPermissionsRealmProxyInterface`

**Result**

You cannot autocomplete stripped methods:
image

Searching for stripped classes will show they exist in the `sources` jar, but not as a normal class. This is a bit unfortunate, but preventing that will mean modifying the source files as well which is probably a bit out of scope for this PR
image

Navigating to a modified class using sources will show a warning saying the bytecode doesn't match. While not ideal, it seems acceptable. It is the same problem as above. In the sources jar you can search for stripped members and find them, but trying to use them in real code will not compile.
image


cmelchior:
Ready for review

cmelchior:
Bump @nhachicha

library-build-transformer/src/main/kotlin/io/realm/buildtransformer/asm/visitors/AnnotationVisitor.kt
```diff
@@ -0,0 +1,68 @@
...
+import io.realm.buildtransformer.ByteCodeTypeDescriptor
+import io.realm.buildtransformer.logger
+import org.objectweb.asm.AnnotationVisitor
+import org.objectweb.asm.ClassVisitor
+import org.objectweb.asm.MethodVisitor
+import org.objectweb.asm.Opcodes
+
+/**
+ * ClassVisitor that gather all classes and methods with the given annotation. This is the first
+ * parse and is required for correctly identifying them in the 2nd parse before any byte code is
```
nhachicha: parse > pass

library-build-transformer/src/test/java/io/realm/buildtransformer/testclasses/NestedTestClass.java
```diff
@@ -0,0 +1,43 @@
...
+
+
+    @ObjectServer
+    public class InnerClass {
+        public String foo;
+    }
+
+    @ObjectServer
+    public enum Enum {
+        FOO
```
nhachicha: Can we apply `@ObjectServer` to a specific enum value?

library-build-transformer/build.gradle
```diff
@@ -0,0 +1,144 @@
...
+                root.appendNode('description', 'Transform library for Realm Java that will strip unwanted files at build time.')
+                root.appendNode('url', 'http://realm.io')
+                root.children().last() + commonPom
+            }
+        }
+    }
+    repositories {
+        maven {
+            credentials(AwsCredentials) {
+                accessKey project.hasProperty('s3AccessKey') ? s3AccessKey : 'noAccessKey'
```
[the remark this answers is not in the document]
cmelchior: I believe this is the setup required if we depend on some SNAPSHOT artifacts only available on S3 which has happened before in a few cases. The logic for Bintray and OJO can be removed though.

build.gradle
```diff
@@ -47,11 +47,19 @@ task installTransformer(type:GradleBuild) {
...
+    buildFile = file('library-build-transformer/build.gradle')
+    tasks = ['publishToMavenLocal']
+}
+
 task assembleRealm(type:GradleBuild) {
     group = 'Build'
     description = 'Assemble the Realm project'
     dependsOn installAnnotations
     dependsOn installTransformer
+    dependsOn installBuildTransformer
```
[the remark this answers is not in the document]
cmelchior: After internal discussion, we are keeping this. Reasons being that we want to detect errors early, also most development happen using the `objectServer` flavor, which skips this step anyway (internally in the transformer).

library-build-transformer/src/test/java/io/realm/buildtransformer/testclasses/NestedTestClass.java
cmelchior: I would have to check the bytecode to be sure and add unit tests for it. For now, I suspect that something would break. I'll add it to the list of limitations

cmelchior:
Feedback addressed: @nhachicha

nhachicha:
not sure if https://github.com/realm/realm-java/pull/6069#discussion_r209037368 was addressed?

cmelchior:
> not sure if #6069 (comment) was addressed?
It was, but not by adding support for it (since we don't have any use now or in the foreseeable future).  I added it to the list of unsupported features in the README  for now.

PR #6073, opened 2018-07-20
Add support for uploading and downloading changes with a timeout

cmelchior:
Closes https://github.com/realm/realm-java/issues/6247

This PR adds two convenience methods overrides for `SyncSession.downloadAllServerChanges()` and `SyncSession.uploadAllLocalChanges()` so they will now terminate if a timeout is hit. This will make it a lot easier to implement sync logic that is guaranteed to proceed after a given period without having to resort to external synchronization

This PR also add support for `SyncConfiguration.waitForInitialRemoteData(timeout, unit)` as it uses the above helper methods.


realm/realm-library/src/objectServer/java/io/realm/SyncSession.java
```diff
@@ -438,6 +499,16 @@ private void checkIfNotOnMainThread(String errorMessage) {
         }
     }
 
+    private void checkTimeout(long timeout, TimeUnit unit) {
+        if (timeout < 0) {
+            throw new IllegalArgumentException("'timeout' most be >= 0. It was: " + timeout);
```
nhachicha: most -> must

realm/realm-library/src/objectServer/java/io/realm/SyncSession.java
```diff
@@ -349,7 +351,34 @@ public void downloadAllServerChanges() throws InterruptedException {
...
+        checkIfNotOnMainThread("downloadAllServerChanges() cannot be called from the main thread.");
+        checkTimeout(timeout, unit);
+
+        // Blocking only happens at the Java layer. To prevent deadlocking the underlying SyncSession we register
+        // an async listener there and let it callback to the Java Session when done. This feels icky at best, but
+        // since all operations on the SyncSession operate under a shared mutex, we would prevent all other actions on the
+        // session, including trying to stop it.
+        // In Java we cannot lock on the Session object either since it will prevent any attempt at modifying the
+        // lifecycle while it is in a waiting state. Thus we use a specialised mutex.
+        synchronized (waitForChangesMutex) {
```
nhachicha: Since the lock is at the Java level only, then I question the usefulness of these overloads, since the `m_state_mutex` will still be acquired from the call to the OS's `session->wait_for_download_completion`, by returning you give the user the false impression that the underlying operation has been cancelled/timed out which  is clearly not the case. Imagine this use case 

```Java
SyncSession sync;
sync.downloadAllServerChanges(1, TimeUnit.MINUTES)
// do some commits 
sync.uploadAllServerChanges(1, TimeUnit.MINUTES);
// by returning here you have no guarantee that the actcual upload was triggered (at OS level), since we might still be holding the lock at the C++ level, from the downloadAll call
```

cmelchior: We are using the asynchronous methods in the C++ layer, so no locks are actually being held. We are just using the Java lock to wait for the callback to happen. The reason for this that is that it is otherwise impossible to interrupt the Java thread at all, which would be highly undesirable.

IMO the semantics we have is that if `downloadAllServerChanges` completes, then changes have been downloaded. If it fails or times out, the behavior is undefined, i.e. it doesn't mean that downloading is canceled, it just means it didn't complete in time. I think that is acceptable. Behaving otherwise indicate a level of control over the syncing process that we don't offer (as everything happens more or less automatically).

cmelchior:
Updated this PR as we had new use cases appear for it. It has been updated to support timeouts for downloading initial data and is ready for review.

realm/realm-library/src/objectServer/java/io/realm/SyncSession.java
cmelchior: I have updated the JavaDoc to indicate that downloading/uploading might still happen.

CHANGELOG.md
```diff
@@ -22,6 +22,9 @@
 * [ObjectServer] Added `Realm.getSubscriptions()`, `Realm.getSubscriptions(String pattern)` and `Realm.getSubscription` to make it easier to find existing subscriptions. These API's are in beta. [#6231](https://github.com/realm/realm-java/pull/6231).
 * [ObjectServer] Added `RealmQuery.subscribe()` and `RealmQuery.subscribe(String name)` to subscribe immediately inside a transaction. These API's are in beta. [#6231](https://github.com/realm/realm-java/pull/6231).
 * [ObjectServer] Added support for subscribing directly inside `SyncConfiguration.initialData()`. This can be coupled with `SyncConfiguration.waitForInitialRemoteData()` in order to block a Realm from opening until the initial subscriptions are ready and have downloaded data. This API are in beta. [#6231](https://github.com/realm/realm-java/pull/6231).
+* [ObjectServer] Added support timeouts when uploading or downloading data manually using `SyncSession.downloadAllServerChanges(long timeout, TimeUnit unit)` and `SyncSession.uploadAllLocalChanges(long timeout, TimeUnit unit)`. [#6073](https://github.com/realm/realm-java/pull/6073).
```
kneth: Remove `.` at the end.

CHANGELOG.md
```diff
@@ -22,6 +22,9 @@
...
+* [ObjectServer] Added support for timing out when downloading initial data for synchronized Realms using `SyncConfiguration.waitForInitialRemoteData(long timeout, TimeUnit unit)`. [#6247](https://github.com/realm/realm-java/issues/6247).
```
kneth: Remove `.` at the end.

realm/realm-library/src/objectServer/java/io/realm/SyncSession.java
```diff
@@ -481,7 +481,35 @@ public void downloadAllServerChanges() throws InterruptedException {
         // In Java we cannot lock on the Session object either since it will prevent any attempt at modifying the
         // lifecycle while it is in a waiting state. Thus we use a specialised mutex.
         synchronized (waitForChangesMutex) {
-            waitForChanges(DIRECTION_DOWNLOAD);
+            waitForChanges(DIRECTION_DOWNLOAD, Long.MAX_VALUE, TimeUnit.MILLISECONDS);
+        }
+    }
+
+    /**
+     * Calling this method will block until all known remote changes have been downloaded and applied to the Realm
```
kneth: "was" -> "is" ???

realm/realm-library/src/objectServer/java/io/realm/SyncSession.java
```diff
@@ -481,7 +481,35 @@ public void downloadAllServerChanges() throws InterruptedException {
...
+     * or the specified timeout was hit. This will involve network access, so calling this method should only be done
+     * from a non-UI thread.
+     * <p>
+     * This method cannot be called before the Realm has been opened.
+     *
+     * @throws IllegalStateException if called on the Android main thread.
+     * @throws InterruptedException if the timeout was hit or the thread was interrupted while downloading was in progress.
```
kneth: "was hit": you mean "if the download took longer than the specified timeout setting".

realm/realm-library/src/objectServer/java/io/realm/SyncSession.java
```diff
@@ -481,7 +481,35 @@ public void downloadAllServerChanges() throws InterruptedException {
...
+     * The download will continue in the background even after this exception is thrown.
+     * @throws IllegalArgumentException if {@code timeout} is less than {@code 0} or {@code unit} is {@code null}.
```
kneth: "is less than" -> is less than or equal to"

realm/realm-library/src/objectServer/java/io/realm/SyncSession.java
```diff
@@ -506,7 +534,35 @@ public void uploadAllLocalChanges() throws InterruptedException {
...
+     * Calling this method will block until all known local changes have been uploaded to the server or the specified
+     * timeout was hit. This will involve network access, so calling this method should only be done from a non-UI
+     * thread.
+     * <p>
+     * This method cannot be called before the Realm has been opened.
+     *
+     * @throws IllegalStateException if called on the Android main thread.
+     * @throws InterruptedException if the timeout was hit or the thread was interrupted while downloading was in progress.
+     * The upload will continue in the background even after this exception is thrown.
+     * @throws IllegalArgumentException if {@code timeout} is less than {@code 0} or {@code unit} is {@code null}.
```
kneth: "less than" -> less than or equal to"

realm/realm-library/src/objectServer/java/io/realm/SyncSession.java
```diff
@@ -635,7 +637,9 @@ private boolean waitForChanges(int direction, long timeout, @Nullable TimeUnit u
             }
             try {
                 result = wrapper.waitForServerChanges(timeout, unit);
+                RealmLog.error("Success: " + result);
```
nhachicha: remove or change the log level to debug/info

realm/realm-library/src/syncIntegrationTest/java/io/realm/SyncSessionTests.java
```diff
@@ -642,4 +644,26 @@ public void stop_multipleTimes() {
...
+    public void waitForInitialRemoteData_throwsOnTimeout() {
+        SyncUser user = UserFactory.createUniqueUser(Constants.AUTH_URL);
+        SyncConfiguration syncConfiguration = configFactory
+                .createSyncConfigurationBuilder(user, Constants.SYNC_SERVER_URL)
+                .initialData(bgRealm -> {
+                    for (int i = 0; i < 100; i++) {
+                        bgRealm.createObject(AllTypes.class);
+                    }
+                })
+                .waitForInitialRemoteData(1, TimeUnit.MILLISECONDS)
```
nhachicha: `NANOSECONDS`?  to be completely sure 🤓

PR #6110, opened 2018-08-16
Correctly keep members of OsSubscription

cmelchior:
Closes https://github.com/realm/realm-java-private/issues/234

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### Bug Fixes
 
-* [ObjectServer] ProGuard was not configered correctly when working with Subscriptions for Query-based Realms.
+* [ObjectServer] ProGuard was not configured correctly when working with Subscriptions for Query-based Realms.
 
 
 ## 5.4.2 (2018-08-09)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## 5.4.3 (2018-08-09)
+
+### Bug Fixes
+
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+## 5.4.3 (2018-08-09)
+
+### Bug Fixes
+
+* [ObjectServer] ProGuard was not configered correctly when working with Subscriptions for Query-based Realms.
```
Zhuinden: config***u***red

realm/realm-library/src/main/java/io/realm/internal/sync/OsSubscription.java
```diff
@@ -24,6 +24,7 @@
 import io.realm.internal.ObserverPairList;
 import io.realm.internal.OsResults;
 
+@KeepMember
```
nhachicha: Use `@Keep` the class (like for `SyncManager` & `SyncSession`)  `@KeepMember` is for methods

cmelchior: Yes, but `@Keep` keeps all class members, we are only interested in one method.
`@KeepMethod` is also required on the class to make the Proguard rules work. They are:
```
-keep class io.realm.internal.KeepMember
-keep @io.realm.internal.KeepMember class * { @io.realm.internal.KeepMember *; }
```

Where as this is the rules for `@Keep`:
```
-keep class io.realm.internal.Keep
-keep @io.realm.internal.Keep class * { *; }
```

PR #6111, opened 2018-08-16
Make ObjectServerExample more accessible

cmelchior:
Ran into this while testing https://github.com/realm/realm-java/pull/6110

Previously the example depended on a local version of ROS running, but testing using Cloud is much simpler, so the example was modified to work when you just copy the instance URL from the web UI.

The actual demo actually also didn't work correctly since it used `findFirstAsync()` instead of `findAllAsync()` meaning that no subscription was ever created.

### DIFF
--- a/examples/objectServerExample/build.gradle
+++ b/examples/objectServerExample/build.gradle
@@ -16,7 +16,7 @@ android {
     buildTypes {
         // Go to https://cloud.realm.io and copy the URL to your instance. Insert it below.
         // It will look something like "https://test.us1.cloud.realm.io"
-        def cloudUrl = "https://cmelchior.us1.cloud.realm.io"
+        def cloudUrl = "<INSERT_CLOUD_URL>"
         def realmAuthUrl = "\"${cloudUrl}/auth\""
         def realmUrl = "\"${cloudUrl.replace("https", "realms")}/default\""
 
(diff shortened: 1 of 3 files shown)


examples/objectServerExample/build.gradle
```diff
@@ -34,18 +14,20 @@ android {
...
 
     buildTypes {
-        // This will automatically try to detect the IP address of the machine
-        // building the example. It is assumed that this machine is also running
-        // the Object Server. If not, replace 'host' with the IP of the machine
-        // hosting the server. In some cases the wrong IP address will also
-        // be detected. In that case also insert the IP address manually.
-        def host = getIP()
+        // Go to https://cloud.realm.io and copy the URL to your instance. Insert it below.
+        // It will look something like "https://test.us1.cloud.realm.io"
```
nhachicha: Add a section about self-hosted:

If you're running a self-hosted version, use the hostname/IP address of the Realm Object Server (example: `http://192.168.1.64:9080`

examples/objectServerExample/build.gradle
```diff
@@ -34,18 +14,20 @@ android {
...
+        def cloudUrl = "<INSERT_CLOUD_URL>"
```
nhachicha: the example is also used for self-hosted, maybe rename to `rosURL` or just `url`?

Issue #5066, opened 2017-08-03
Detach Realm from RealmQuery


VenomVendor:
## Problem statement.
As described in https://stackoverflow.com/q/44035272/1008278
Create a **RealmQuery** in Thread A & pass it across to be executed conditionally in different Thread.
This triggers `IllegalStateException(BaseRealm.INCORRECT_THREAD_MESSAGE);`.
[Solution provided cannot be used, as listeners can be added only if Looper is available](https://stackoverflow.com/a/44043733/1008278) & it is not necessary the result will always update UI.

[RealmQuery.createQuery(this, clazz);](https://github.com/realm/realm-java/blob/5e20952decab2a2f4dda826a06a21a3b93115665/realm/realm-library/src/main/java/io/realm/Realm.java#L1346)
Creates a query binding realm Object tightly to RealmQuery, but the actual usage of realm is required only when query is executed. [createRealmResults](https://github.com/realm/realm-java/blob/5e20952decab2a2f4dda826a06a21a3b93115665/realm/realm-library/src/main/java/io/realm/RealmQuery.java#L1972), [findFirstAsync](https://github.com/realm/realm-java/blob/5e20952decab2a2f4dda826a06a21a3b93115665/realm/realm-library/src/main/java/io/realm/RealmQuery.java#L1936) or similay query. 
As mentioned in doc, RealmQuery is built on builder pattern. Clearly there is no dependency of Realm until execution.

RealmQuery should be detached from Realm & should be able to execute after attaching to Realm.

Ex:
``` java
Query query = Realm.where(clazz); // Query is detached from realm.
 (or)
Query query = Realm.createQuery(clazz); // Realm.prepareQuery();
```
> query[.equalTo(String fieldName, String value)](https://github.com/realm/realm-java/blob/5e20952decab2a2f4dda826a06a21a3b93115665/realm/realm-library/src/main/java/io/realm/RealmQuery.java#L238)[.equalTo(String fieldName, Byte value)](https://github.com/realm/realm-java/blob/5e20952decab2a2f4dda826a06a21a3b93115665/realm/realm-library/src/main/java/io/realm/RealmQuery.java#L271)...

Since the above query is detached from Realm, this can be passed across threads & doesn't cause any harm. This is still a plain Object ready to be executed.

Before execution, Query can attach itself to Realm returning RealmQuery.
``` java
// Below cannot be transmitted across threads.
RealmQuery realmQuery = query.attachWith(realm); // Attaches to realm
[...]

Issue #6092, opened 2018-08-01
Query child objects with multiple queries like SUBQUERY on iOS


kevinvanmierlo:
I want to be able to add multiple queries on child objects. For example I have a `Member` object which has multiple `MemberOther` objects. `MemberOther` is actually a collection for numerous things. I'm trying to query a `RealmList` which contains members to retrieve `MemberOther` objects using a filter. The problem that I'm having is that I need to query multiple things on `MemberOther` objects. For example a simple query would be that a `MemberOther` object should be a specific type AND lower than a specific number. So I thought I would do this query:

```Kotlin
members.equalTo("others.type", type).lessThanOrEqualTo("others.startYear", yearFilter.startYear)
```

So from the first `equalTo` retrieves the members which have memberOthers of a specific type. But the second one retrieves the members which have a startYear lower than my filter BUT doesn't take into account the specific type I used for the first query. 

On iOS this is possible using a SUBQUERY:

```objective-c
[NSPredicate predicateWithFormat:@"SUBQUERY(memberOthers, $memberOther, $memberOther.type == %@ AND $memberOther.startYear <= %@) .@count > 0", type, @(yearFilter.startYear)];
```

I've read other issues like https://github.com/realm/realm-java/issues/1237 and https://github.com/realm/realm-java/issues/1598. But one is already closed and the other is about the aggregate queries and not specifically about multiple queries on child objects. 

Do you have anything on the timeline that will support these kind of queries?

Thanks in advance!

cmelchior:
Unfortunately `SUBQUERY` isn't supported in the Java SDK yet but you can read https://realm.io/docs/java/latest/#link-queries which can give you some insight in to how to build up a query doing something similar.

kevinvanmierlo:
I have looked at the documentation, but unfortunately it isn't able to do this kind of query. What I have done now is query all the objects and filter them by using Kotlin collection functions. While this would suffice for small lists, it could be a problem when the list grows. Because the thing I want to do is filter the list by two conditions of the child objects at the same time. So I can write the conditions, but Realm will only execute one at the time. 

[...]

Issue #5730, opened 2018-01-30
Support string based queries


cmelchior:
With https://github.com/realm/realm-core/pull/2979 nearing completion the query serializer is soon in a state where can consider exposing it in the public API since it supports use cases not yet covered by the static query API like #1615 and #1598.  

Suggestions:
```

// 1) `filter` as a new predicate that takes a string. 
RealmResults<Person> results = realm.where(Person.class).filter("field1 > field2").findAll();

// 2) `filter` as an alternative to `where`. Makes the difference more "important" 
      plus aligns better with Swift naming
RealmResults<Person> results = realm.filter(Person.class, "field1 > field2").findAll();

// 3) Let filter be fully lazy, like e.g. Cocoa. A step toward unifying semantics 
      across bindings. It will probably massively complicate the internals though, 
      also means we cannot combine typed/stringbased predicates.
RealmResults<Person> results = realm.filter("Person.class, "field1 > field2");
```

It is still missing support for backlink queries and sorting/distinct, but that should be acceptable since it should be possible to combine the string-based and function based API's: 

```
realm.where(Person.class)
    .filter( "field1 > field2")
    .equalTo("name", "Jane")
    .sort("name")
    .findAll()
```

Some caveats:
- An initial simple implementation will only work with cores internal names and not the names defined in Java (this could be a problem when using `@RealmField` to change names). This is especially problematic if combining typed and string based input

ironage:
I like this idea :) Note that the parser is built independently from core as `realm-core-parser.a` because it adds about 1.4MB to the binary. Would this be a concern to Android users? Maybe users could opt-in to add the parser if they don't mind the extra space?

By the way, sort and distinct in the query string is also supported (in https://github.com/realm/realm-core/pull/2979) but you can also chain it to the typed API if preferred like the example above.

cmelchior:
If it is 1.4 MB pr. architecture, that seems excessive, and something we would need to think extremely hard about how to include because I doubt we can make that the default in that case, but it should be possible as some sort of extension I guess.

Zhuinden:
@ironage but there [used to be a query parser in object-store too, so I'd guess there must have been some size gain from moving it into a core module?](https://github.com/realm/realm-object-store/commit/21e671a155989f4e34e279fc8a7ace5817495de0)

I can't really tell how much size difference it makes overall, though :D that's a bit bigger than I expected, but still, this feature would be super-powerful.

[...]

Issue #1615, opened 2015-10-19
Add support for multi-field comparisons


ghost:
How to compare two fields in android realm.io ?

emanuelez:
We currently we don't support multi-column comparisons in the Java API, mostly because it's hard to make a nice API for it. We are thinking about some possible options.

cmelchior:
Just adding `equalToField` would probably be easy, but the list is also a bit bigger which complicates it:

- notEqualTo
- Calculations on integer fields, eg. `2 * field1 > field2`
- does contains/startsWith/endsWith work on multiple fields?

brozot:
+1

Hello I need to use this solution to make this comparison  : updateDate > than syncDate (these two dates are in the database). I want to fetch only the updated entities or there is another solution for this?

Saketme:
Please consider adding support for this. I've been running `equalTo` and other operators in a loop :(

eyalfein:
Came here looking for solution to this problem (now missing feature) too.
Looks like there is significant demand for it.
I am hoping it will be in the form of:
.equalTo(string, string), notEqualTo(string, string), etc - to compare compare two properties of the same entity.

papagno:
No workaround at the moment? I can't select comparing two columns in any way?

bmunkholm:
One workaround is to add a third boolean which is the result of the comparison you wan to make. You would then need to update that whenever the properties you compare changes.

The Core does support this, so anyone with time can make a PR. The first task is to propose a proper API, then it is trivial to implement (I'll be happy to provide pointers to the core API).

nachogarrone:
I would like to have this feature too!

I want to filter my RealmResults in my adapter when the user types in a search text view. I want to filter by all the fields in RealmResults.

Example:
firstName -- lastName
John -- Johnnes

`constraint = "John J"
mRealmObjectList.where().contains("firstName", String.valueOf(constraint), Case.INSENSITIVE).or().contains("lastName", String.valueOf(constraint), Case.INSENSITIVE) `

This wouldn't receive any record. 

My workaround is to get all the fields in a temporal string and do the contains() in there. But i wonder how much it will cost if my table has a lot of rows...

bdpham:
@bmunkholm I'd like some pointers on the core API please. When referring to the core API, are you talking about the cpp files that provide JNI functions?

letronje:
+1 for this

acib708:
[...]

PR #6116, opened 2018-08-17
Add support for text-based query predicates

cmelchior:
This PR adds support for Cores query parser, making it possible to convert string-based query input to Query objects.

PR Features:
* Full query support including subqueries, aggregates, multifield comparison and other things.
* Immediate support for new query features in Core.
* It is possible to mix and match raw and typed predicates.

It has the potential to close the following issues (and possibly more I could not find):
https://github.com/realm/realm-java/issues/1573
https://github.com/realm/realm-java/issues/1598
https://github.com/realm/realm-java/issues/1615
https://github.com/realm/realm-java/issues/5730
https://github.com/realm/realm-java/issues/6092

Example:
```
// Simple
RealmResults<Person> results = realm.where(Person.class)
    .rawPredicate("name = 'Jane'")
    .findAll();

// Mixed predicates
RealmResults<Person> results = realm.where(Person.class)
    .rawPredicate("name = 'Jane'")
    .rawPredicate("children@count > 2")
    .equalTo("lastName", "Doe") 
    .findAll();

// With arguments
RealmResults<Person> results = realm.where(Person.class)
    .rawPredicate("name = '%s'", "Jane")
    .findAll();
```

**Size (non-sync):**

Current 10.3.1 release:

```
AAR file: 9.473.250 bytes
4346732 ./armeabi-v7a/librealm-jni.so
6328504 ./x86/librealm-jni.so
5829216 ./arm64-v8a/librealm-jni.so
6705992 ./x86_64/librealm-jni.so
```

Current master (NDK 22)
```
AAR file: 9.307.694 bytes
4307436 ./jni/armeabi-v7a/librealm-jni.so
6371600 ./jni/x86/librealm-jni.so
5760808 ./jni/arm64-v8a/librealm-jni.so
6620064 ./jni/x86_64/librealm-jni.so
```

This PR:
```
AAR file: 10.047.742

4698424 ./obj/armeabi-v7a/librealm-jni.so
7050428 ./obj/x86/librealm-jni.so
6360080 ./obj/arm64-v8a/librealm-jni.so
7311176 ./obj/x86_64/librealm-jni.so
```

TODO:
- [x] Figure out if we can minize the size further just with compiler flags. I need help from a CMake / low-level expert for this.

- [x] If we cannot get the parser down to a reasonable size. Figure out a way to make the parser opt-in. Ideally we want to deploy this a a seperate dependency configurable using the Gradle plugin. This means releasing the parser as a independent `.so` file. A bit unclear how to do this. Will require further research. Worst-case we need a matrix deployment of (base, base+parser, sync, sync+parser). 

- [x] Figure out the exact API name for text-based queries. `filter(..)`, `rawPredicate(...)`, `predicate(...)`, `<other>?`. Personally, I like `rawPredicate(..)` best right now.

- [x] Figure out how to handle input parameters. E.g. `rawPredicate("foo = '$s'", "bar")`. IMO we should just reuse the format specification from `String.format()` since people probably know this as it is also used by Androids resource system: https://docs.oracle.com/javase/7/docs/api/java/util/Formatter.html

- [x] Figure out a long-time roadmap for query features. Do we eventually want all features as type-safe methods or only the "simple" ones. If so, where do we draw the line? Personally I think that all the advanced features will probably be fine as `rawPredicates` only since making nice type safe API's for them is near impossible.

- [x] Figure out how to handle column name aliases. For the typed API we require that people use the java names. Doing that here, might be a bit complicated, but if the api is called `rawX` we could probably get away with requiering the name written in the Realm file, just like we do for DynamicObject queries. It is unclear to me how annoying it would be.

- [x] Figure out how the ordering in predicates are used by the low level API.

- [x] Smoke tests for Dynamic Realms

- [x] Change exception type from IllegalState to IllegalArgument when something is wrong with the query.

- [x] Add support for class name remapping

- [x] Refactor how KeyPathMapping is created. Move it out of the constructor and after SharedRealm creation.

- [x] Wait for fix for issues with KeyPathMapping in backlink queries: https://github.com/realm/realm-core/pull/4401


realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -1973,6 +1973,19 @@ private OsResults lazyFindAll() {
...
     }
 
+    /**
+     * FIXME Add description
+     * FIXME Figure out an API for providing arguments, probably just re-use `String.format()` arguments
+     *
+     * @param filter
+     * @return
+     */
+    public RealmQuery<E> rawPredicate(String filter) {
```
kneth: I suggested `filtered()` - like Realm JS.

cmelchior: The issue I have with `filter` or `filtered` is that it is too generic. I mean it is a great keyword if is the only keyword you have when querying, but in our case, you can interleave with a lot of other statements and there it feels a little weird (at least to me):

```
realm.where(Person.class)
  .equalTo("name", "Jane")
  .filtered("children@count > 2")
  .lessThan("age", 35)
  .findAll();
```


```

**Comment by roberhofer** [comment]
I have a backlog item for full-text search: https://realmio.aha.io/features/SDK-8
This is something I see further out. And if we support it, then it has to be E2E (end-to-end) - including sync.

File: realm/realm-library/src/main/cpp/io_realm_internal_TableQuery.cpp [hunk 2]
```diff
@@ -1828,3 +1829,22 @@ JNIEXPORT jlong JNICALL Java_io_realm_internal_TableQuery_nativeGetFinalizerPtr(
...
+        Query* query = reinterpret_cast<Query *>(nativeQueryPtr);
+        JStringAccessor filter(env, j_filter); // throws
+        parser::ParserResult parser_result = realm::parser::parse(filter);
+
+        query_builder::NoArguments no_args;
+        query_builder::apply_predicate(*query, parser_result.predicate, no_args);
+
+        // TODO What about ordering. What does this mean?
+        // DescriptorOrdering ordering;
+        // realm::parser::query_builder::apply_ordering(ordering, table, parser_result.ordering);
```
ironage: The ordering handles sort/distinct (and soon limit) parts of the query. I think to handle that here, you'd want to save the ordering produced by `apply_ordering` and then apply it when the user calls `find_all()` or `count()` etc. I'm guessing there is already a ordering state somewhere since it is supported through Java's typed query API, so you could append the sort/distinct descriptors of this result with the existing state (the API's for this aren't super friendly right now but it is possible).

ironage:
And for the column aliasing, there is support for this using `parser::KeyPathMapping` we use it in the [js binding](https://github.com/realm/realm-js/pull/1746) to alias user defined backlink column names which core doesn't know about.

realm/realm-library/src/main/cpp/io_realm_internal_TableQuery.cpp
cmelchior: Thanks 👍 Yes, we are storing a Descriptor object in Java right now. It will be a bit fiddly trying having to send this one back from this and then decide which to use if people later override it, but it isn't impossible. Thanks for the clarification 😄

ironage: It should be possible to merge orderings, so that all of them are applied. This should let users combine sort/distinct across the typed and untyped API correctly and both the sort/distinct parts be applied in the requested order.

```
realm.where(Person.class)
  .equalTo("name", "Jane")
  .filtered("children.@count > 2 SORT(name ASC) DISTINCT(name)")
  .lessThan("age", 35)
  .sort("age", ASCENDING)
  .distinct("age")
  .findAll();
```

It would be a bit tricky to do this currently, but we could easily add a `DescriptorOrdering::appendOther(DescriptorOrdering other)` method in core.

realm/realm-library/src/androidTest/java/io/realm/RealmQueryTests.java
```diff
@@ -3518,4 +3519,40 @@ public void getRealm_throwsIfRealmClosed() {
...
+    public void rawPredicate() {
+        populateTestRealm();
+        RealmResults<AllTypes> result = realm.where(AllTypes.class).rawPredicate("columnString = 'test data 0'").findAll();
+        assertEquals(1, result.size());
+    }
+
+    @Test
+    public void rawPredicate_invalidFieldName() {
+        try {
+            realm.where(AllTypes.class).rawPredicate(("foo = 'test data 0'");
```
ironage: syntax error with an extra `(`

realm/realm-library/src/main/cpp/io_realm_internal_TableQuery.cpp
cmelchior: Right now we are storing a separate `sortDescriptor` and `distinctDescriptor` in Java, keeping the latest one defined and when you call `findAll()` it merges them using `append_sort` and `append_distinct` on a new Descriptor object.

I assume the same approach would work for the `ParseResult`? Although it doesn't seem possible to pull out the sort/distinct parts of the Descriptor individually?

If that isn't the case, having a method that merges them seems required.  Then we could also simplify the logic in the Java query class and only have one Descriptor.

realm/realm-library/src/androidTest/java/io/realm/QueryTests.java
```diff
@@ -80,6 +82,8 @@
 
     @Before
     public void setUp() throws Exception {
+        Realm.init(InstrumentationRegistry.getTargetContext());
```
nhachicha: @cmelchior 
I tested it locally & it looks like `TestRealmConfigurationFactory#apply` is not called (which is responsible to call `Realm.init(InstrumentationRegistry.getTargetContext());`) when using `base` flavour. If you switch to `objectServer` flavour the rule is applied correctly.  

It looks related to the `io.realm.buildtransformer.RealmBuildTransformer`, I disabled the transformer & run the test again using the `base` flavour and it succeeded. It seems like we have a bug with the Transformer that needs investigation

cmelchior: Nice catch. Thanks 👍

Zhuinden:
> Figure out a long-time roadmap for query features. Do we eventually want all features as type-safe methods or only the "simple" ones. If so, where do we draw the line?

I vote for the line being all those things that are still pending (aggregate queries, subqueries) and what's not even really mentioned but kinda necessary in some way or any (ALL link queries instead of ANY)

---

btw I hope this gets through eventually because it's going to make life much nicer :+1:

Zhuinden:
I don't want typed apis for this. I'd be happy to have the core parser handle it. You guys already have your hands full with everything else, it's easier to just say "use this raw predicate and it will work like magic" :smile:

realm/realm-library/src/main/java/io/realm/RealmQuery.java
Zhuinden: Not to mention, then `sort()` should be called `sorted()`. And it only makes sense if you don't have *any* of the other keywords (lessThan, equalTo, ...)

realm/realm-library/src/androidTest/java/io/realm/RealmQueryTests.java
```diff
@@ -3530,24 +3530,26 @@ public void rawPredicate() {
...
-    }
+//    @Test
+//    public void rawPredicate_mixedWithTypedPredicates() {
+//        populateTestRealm();
+//        // FIXME We might need to track `or()` in cases such as this so we can add a TRUE_PREDICATE
+//        //
+//        RealmResults<AllTypes> result = realm.where(AllTypes.class)
+//                .equalTo("columnString", "test data 0")
+//                .or()
+//                .rawPredicate("columnString = 'test data 1'") // This throws `Missing right-hand side of OR`
```
Zhuinden: This sounds like a fairly complex problem especially if you consider raw predicates that have their own AND/OR... but what if it is enclosed in beginGroup/endGroup?

kneth:
A very dense list of the query capabilities can be found at https://realm.io/docs/javascript/latest/#filtering

realm/realm-library/src/androidTest/java/io/realm/RealmQueryTests.java
cmelchior: Yes, I suspect that us adding `beginGroup/endGroup` around the predicate might actually solve it. Worth a shot at least 😄

cmelchior: It turned out the problem I was validating the query too soon. The above is supported by Core.

VenomVendor:
Can we get an api to convert RealmQuery to equivalent `rawPredicate`, like
`String rawPredicate = realmQuery.toRawPredicate();`
or 
`String rawPredicate = realmQuery.toString();`
This would resolve #5066

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -1973,6 +1973,23 @@ private OsResults lazyFindAll() {
...
+     *
+     * @param filter
+     * @return
+     */
+    public RealmQuery<E> rawPredicate(String filter, Object... arguments) {
+        realm.checkIfValid();
+        if (Util.isEmptyString(filter)) {
+            throw new IllegalArgumentException("Non-null 'filter' required.");
+        }
+        String formattedFilter = String.format(filter, arguments);
```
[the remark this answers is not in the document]
cmelchior: Good point. Thanks 👍

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -1973,6 +1973,23 @@ private OsResults lazyFindAll() {
...
     }
 
+    /**
+     * FIXME Add description
+     * FIXME Figure out an API for providing arguments, probably just re-use `String.format()` arguments
+     *
+     * @param filter
+     * @return
+     */
+    public RealmQuery<E> rawPredicate(String filter, Object... arguments) {
```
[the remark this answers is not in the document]
Zhuinden: Are you sure you need to pass explicit `null` for varargs?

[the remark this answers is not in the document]
cmelchior: No, you can just ignore it. No override is needed.

Zhuinden:
¯\_(ツ)_/¯

cmelchior:
It hasn't been forgotten, unfortunately a bit delayed. But this is definitely something we want to do somehow.

cmelchior:
Latest stats for Core 10 is:

```
// Unzip AAR and run 
// find . -name "*.so" -print0 | xargs -0 stat -f '%z %N'

Base without Parser (10.0.0-BETA.6)
AAR file: 8.813.770 bytes
4.215.952 ./jni/armeabi-v7a/librealm-jni.so
6.328.776 ./jni/x86/librealm-jni.so
6.025.880 ./jni/arm64-v8a/librealm-jni.so
6.657.104 ./jni/x86_64/librealm-jni.so

Base with Parser (10.0.0-BETA.6)
AAR file: 10.735.381 bytes (+22%)
5.358736 ./jni/armeabi-v7a/librealm-jni.so
8.335816 ./jni/x86/librealm-jni.so
7.574168 ./jni/arm64-v8a/librealm-jni.so
8.471632 ./jni/x86_64/librealm-jni.so

ObjectServer without Parser (10.0.0-BETA.6)
AAR file: 12.524.448 bytes 
5.800.200 ./jni/armeabi-v7a/librealm-jni.so
8.883.856 ./jni/x86/librealm-jni.so
8.375.176 ./jni/arm64-v8a/librealm-jni.so
9.186.672 ./jni/x86_64/librealm-jni.so

ObjectServer with Parser (10.0.0-BETA.6)
AAR file: 14.447.120 bytes (+15%)
6.938.888 ./jni/armeabi-v7a/librealm-jni.so
10.886.800 ./jni/x86/librealm-jni.so
9.911.176 ./jni/arm64-v8a/librealm-jni.so
10.997.104 ./jni/x86_64/librealm-jni.so

```

cmelchior:
This PR is now ready for review. Most of the changes are smaller changes required to support the KeyPathMapping in the query parser so you can choose to use either internal or java defined names in the queries. Since a name has to be unique across both sets, this shouldn't cause any problems.

The only outstanding issue is that, currently, Core is not throwing the correct exception type in all cases. This means that some syntax errors will be reported as `RuntimeException` instead of `IllegalArgumentException`. I don't think this is critical enough to hold back review and releasing this though, so I have added tests for the some of the cases I ran into. Then we will discover when Core fixes this.

There is a small size increase associated with this, +6% for an AAR with all ABI's. However I think that is within acceptable bounds given the massive increase in usability this provides. Also, Core is still working on reducing the size.

realm/realm-library/src/main/cpp/io_realm_internal_Property.cpp
```diff
@@ -66,15 +69,17 @@ JNIEXPORT jlong JNICALL Java_io_realm_internal_Property_nativeCreatePersistedPro
...
                                                                                           jstring j_target_class_name)
 {
     try {
-        JStringAccessor name(env, j_name_str);
+        JStringAccessor public_name(env, j_public_name);
+        JStringAccessor internal_name(env, j_internal_name);
         JStringAccessor link_name(env, j_target_class_name);
         PropertyType p_type = static_cast<PropertyType>(static_cast<int>(type));
-        return reinterpret_cast<jlong>(new Property(name, p_type, link_name));
+        return reinterpret_cast<jlong>(new Property(internal_name, p_type, link_name, "", public_name));
```
clementetb, line 82: Are the parameters for the Property constructor, ok?:
`inline Property::Property(std::string name, PropertyType type, std::string object_type, std::string link_origin_property_name, std::string public_name)`
```suggestion
        return reinterpret_cast<jlong>(new Property(internal_name, p_type, "", link_name, public_name));
```

realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmProxyClassGenerator.kt
```diff
@@ -679,12 +679,13 @@ class RealmProxyClassGenerator(private val processingEnvironment: ProcessingEnvi
                 val persistedFields = metadata.fields.size
                 val computedFields = metadata.backlinkFields.size
                 val embeddedClass = if (metadata.embedded) "true" else "false"
-
-                emitStatement("OsObjectSchemaInfo.Builder builder = new OsObjectSchemaInfo.Builder(\"%s\", %s, %s, %s)", internalClassName, embeddedClass, persistedFields, computedFields)
+                val publicClassName = if (simpleJavaClassName.name != internalClassName) simpleJavaClassName.name else ""
+                emitStatement("OsObjectSchemaInfo.Builder builder = new OsObjectSchemaInfo.Builder(\"%s\", \"%s\", %s, %s, %s)", publicClassName, internalClassName, embeddedClass, persistedFields, computedFields)
```
clementetb, line 683: Shouldn't we use string templating?
```suggestion
                emitStatement("OsObjectSchemaInfo.Builder builder = new OsObjectSchemaInfo.Builder(\"${publicClassName}\", \"${internalClassName}\", ${embeddedClass}, ${persistedFields}, ${computedFields})")
```

realm/realm-library/src/androidTestObjectServer/kotlin/io/realm/SyncedRealmMigrationTests.kt
```diff
@@ -130,8 +130,8 @@ class SyncedRealmMigrationTests {
 
         // Setup initial Realm schema (with a different primary key)
         val expectedObjectSchema = OsObjectSchemaInfo.Builder(PrimaryKeyAsString.CLASS_NAME, false, 2, 0)
-                .addPersistedProperty(PrimaryKeyAsString.FIELD_PRIMARY_KEY, RealmFieldType.STRING, false, true, false)
-                .addPersistedProperty("_id", RealmFieldType.INTEGER, true, true, true)
+                .addPersistedProperty("", PrimaryKeyAsString.FIELD_PRIMARY_KEY, RealmFieldType.STRING, false, true, false)
```
clementetb, line 133: We could replace the `""` empty string with a constant like EMPTY_NAME or NO_NAME.

realm/realm-annotations-processor/src/test/resources/io/realm/some_test_BooleansRealmProxy.java
```diff
@@ -185,11 +185,11 @@ protected final void copy(ColumnInfo rawSrc, ColumnInfo rawDst) {
     }
 
     private static OsObjectSchemaInfo createExpectedObjectSchemaInfo() {
-        OsObjectSchemaInfo.Builder builder = new OsObjectSchemaInfo.Builder("Booleans", false, 4, 0);
-        builder.addPersistedProperty("done", RealmFieldType.BOOLEAN, !Property.PRIMARY_KEY, !Property.INDEXED, Property.REQUIRED);
-        builder.addPersistedProperty("isReady", RealmFieldType.BOOLEAN, !Property.PRIMARY_KEY, !Property.INDEXED, Property.REQUIRED);
-        builder.addPersistedProperty("mCompleted", RealmFieldType.BOOLEAN, !Property.PRIMARY_KEY, !Property.INDEXED, Property.REQUIRED);
-        builder.addPersistedProperty("anotherBoolean", RealmFieldType.BOOLEAN, !Property.PRIMARY_KEY, !Property.INDEXED, Property.REQUIRED);
+        OsObjectSchemaInfo.Builder builder = new OsObjectSchemaInfo.Builder("", "Booleans", false, 4, 0);
+        builder.addPersistedProperty("", "done", RealmFieldType.BOOLEAN, !Property.PRIMARY_KEY, !Property.INDEXED, Property.REQUIRED);
```
clementetb, line 189: We could replace the " empty string with a constant like EMPTY_NAME or NO_NAME.

clementetb (changes requested):
Looks great 💎 !!! Added some minor comments.

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -2216,6 +2217,56 @@ private OsResults lazyFindAll() {
...
+     * for a more detailed description of the Realm Query Language.
+     * <p>
+     * <pre>
+     * Examples:
+     * {@code
+     * RealmQuery<Person> query = realm.where(Person.class);
+     *
+     * // Simple query
+     * query.rawPredicate("name = 'Jane'");
+     * query.rawPredicate("name = '%s'", getName());
```
ironage, line 2241: You will find syntax errors if the string being substituted here contains any single quote character without escaping it. The recommended way is to use the query parser's arguments (`name = $1`) and pass the string variable in to the query parser in an argument list, then you don't have to do any special character escaping. The `Table::query` method accepts a `vector<Mixed>` for an argument list for this purpose. Not blocking though if you want to get this in for now.

cmelchior: Hmm, could point. I used the formatting from Java because it was easier instead of trying to marshell parameters across JNI, but that was probably a bad idea. By using the parser substitution it will also make it easier to document.

realm/realm-library/src/main/java/io/realm/RealmSchema.java
```diff
@@ -246,12 +248,23 @@ final RealmObjectSchema removeFromClassNameToSchemaMap(String name) {
         return dynamicClassToSchema.remove(name);
     }
 
+    final OsKeyPathMapping getKeyPathMapping() {
```
rorbech, line 251: Should this trigger creation first time it is called? Or it is intentionally left as a _manual_ step?

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -2216,6 +2217,56 @@ private OsResults lazyFindAll() {
...
+     * // Sort, Distinct, Limit
+     * query.rawPredicate("name = 'Jane' SORT(lastName) DISTINCT(city) LIMIT(5)");
+     * }
+     * </pre>
+     *
+     * @param predicate A Realm Query Language predicate.
+     */
+    public RealmQuery<E> rawPredicate(String predicate, Object... arguments) {
+        realm.checkIfValid();
+        if (Util.isEmptyString(predicate)) {
```
rorbech, line 2262: You can query on the `"TRUE_PREDICATE"` (AFAIR) if you want all the objects. Don't know if it makes sense to expose that though.

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -2216,6 +2217,56 @@ private OsResults lazyFindAll() {
...
+     * query.rawPredicate("SUBQUERY(children, $child, $child.age > 21 AND $child.gender = 'male'").@count > 0');
+     *
+     * // Sort, Distinct, Limit
+     * query.rawPredicate("name = 'Jane' SORT(lastName) DISTINCT(city) LIMIT(5)");
+     * }
+     * </pre>
+     *
+     * @param predicate A Realm Query Language predicate.
+     */
+    public RealmQuery<E> rawPredicate(String predicate, Object... arguments) {
```
rorbech, line 2260: Missing `@throws` on syntax errors?

rorbech (approved):
Only some minor comments 🚀

rorbech:
Maybe adding support for printing queries (#3351) would be nice to include in the same go, as it it would assist in migrating existing queries if one needs to tweak it with the options from specifying queries from strings.

cmelchior:
> Maybe adding support for printing queries (#3351) would be nice to include in the same go, as it it would assist in migrating existing queries if one needs to tweak it with the options from specifying queries from strings.

That support is already there through `RealmQuery.getDescription()` https://docs.mongodb.com/realm-sdks/java/latest/io/realm/RealmQuery.html#getDescription--

realm/realm-library/src/main/java/io/realm/RealmQuery.java
cmelchior: Yes, I have been thinking about the same thing. "TRUEPREDICATE" does work as the "select all" option. I don't have strong opinions, but I assumed that if you provided an empty string it was more likely a mistake than not.

realm/realm-library/src/main/cpp/io_realm_internal_Property.cpp
cmelchior: Yes, they should be fine (at least I think so, and all tests pass). The only constructor that accepts the public_name is also used to support backlinks, so the number of arguments is a bit weird.

realm/realm-library/src/main/java/io/realm/RealmSchema.java
cmelchior: It is a manual step on purpose since I can then make sure it is called as part of opening the Realm. While I haven't measured the performance, it is slightly heavyweight as it as to loop through all classes and properties, so it seemed safer to have it as part of the startup cost of opening a Realm.

realm/realm-library/src/main/java/io/realm/RealmQuery.java
cmelchior: It turns out that supporting the Mixed argument list isn't trivial yet until we have the full Mixed support currently being added. Instead of spending time adding work-arounds, I'm removing support for argument substitution until we have full Mixed support. This shouldn't take long anyway and the work-arounds are pretty clear and simple, especially in Kotlin with already support in-place substitution of variables.

cmelchior:
Ready for final reviews. I removed the support for replaced arguments until we can use the Mixed support. This shouldn't be super problematic as we can add the support without breaking the API and the workarounds are trivial.

clementetb (approved):
🚀 great work!

PR #6121, opened 2018-08-22
Performance detach results

lucasdornelasv:
Added:

ArrayList<E> realmObjects;
if (objects instanceof Collection) {
        realmObjects = new ArrayList<>(((Collection) objects).size());
} else {
        realmObjects = new ArrayList<>();
}


lucasdornelasv:
Ok, 👍

PR #6126, opened 2018-08-24
Support LIMIT

### LINKED ISSUE #544: LIMIT
It would be very useful to be able to specify the maximum number of objects to return in a query, similar to LIMIT in SQL. Otherwise the RealmResults list must be manually trimmed with subList.

cmelchior:
Closes #544 

This PR adds support for the keyword LIMIT when making queries. Note, that OFFSET is still not supported. Core 5.8.0 has not yet been released, so this cannot be completed until that has happend

- [x] Refactor descriptors to use Cores DescriptorOrdering
- [x] Refactor Java `SortDescriptor` to better named `FieldDescriptor` or similar (FieldDescriptor was already taken, renamed to QueryDescriptor).
- [x] Update to Sync/Core release with support for LIMIT
- [x] Extend DescriptorOrdering class with correct API calls for specifying the limit
- [x] Add public LIMIT API
- [x] Add query tests
- [x] Add notification tests
- [x] Add subscription tests
- [x] Update ObjectStore to support notifications


realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -1951,6 +1949,21 @@ private OsResults lazyFindAll() {
         return this;
     }
 
+    /**
+     * Limits the number of objects returned in case the query matched more objects.
+     *
+     * @param limit a limit that is {@code &gt; 0}.
```
Zhuinden: Is limit(0) allowed?

cmelchior: I think technically yes, it would just evaluate to a FALSE predicate.

cmelchior:
Ready for review

realm/realm-library/src/main/java/io/realm/internal/core/package-info.java
```diff
@@ -0,0 +1,18 @@
+/*
+ * Copyright 2017 Realm Inc.
```
nhachicha: 2018

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/QueryBasedSyncTests.java
```diff
@@ -296,6 +300,42 @@ public void clearTable() {
         looperThread.testComplete();
     }
 
+    @Test
+    @RunTestInLooperThread
+    @Ignore("FIXME: Enable when https://github.com/realm/realm-js/issues/1971 is fixed")
```
nhachicha: I think you're referencing the wrong issue, the one used concerns custom SSL conf

realm/realm-library/src/androidTest/java/io/realm/RealmQueryTests.java
```diff
@@ -3518,4 +3517,78 @@ public void getRealm_throwsIfRealmClosed() {
...
+                .sort(AllTypes.FIELD_LONG, Sort.DESCENDING) // [4, 4, 3, 3, 2, 2, 1, 1, 0, 0]
+                .distinct(AllTypes.FIELD_LONG) // [4, 3, 2, 1, 0]
+                .limit(2) // [4, 3]
+                .findAll();
+        assertEquals(2, results.size());
+        assertEquals(4, results.first().getColumnLong());
+        assertEquals(3, results.last().getColumnLong());
+
+        results = realm.where(AllTypes.class)
+                .limit(2) // [0, 1]
```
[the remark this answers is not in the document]
cmelchior: Good point. I'll change it so the query is run on a list instead which will guarantee the order at the beginning

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -1938,20 +1933,37 @@ private OsResults lazyFindAll() {
...
+    }
+
+    /**
+     * Limits the number of objects returned in case the query matched more objects.
+     * <p>
+     * Note that when using this method in combination with {@link #sort(String)} and
+     * {@link #distinct(String)} they will be executed in the order they where added which can
+     * effect the end result.
+     *
+     * @param limit a limit that is {@code &gt; 0}.
```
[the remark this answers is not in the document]
cmelchior: I did consider disallowing 0, because like you I had a hard time coming up with a sensible use case. The only thing I could come up with is making sure an empty list is returned, but you can also use `.alwaysFalse()` for that. I'll change it to 1 instead.

Zhuinden:
> keyword LIMIT when making queries. Note, ***that OFFSET is still not supported.***

You won't be able to use this for pagination yet.

SahajRana:
Ohh okay, If pagination would be there, that would have been really great.

PR #6135, opened 2018-08-29
Added support for stopping and starting a session

cmelchior:
Closes: https://github.com/realm/realm-java-private/issues/235

This PR adds support for manually stopping and starting a session. 
The session lifecycle is still tied to the Realm though. Changing that will be a major refactor of the SyncSession in Object Store.

This approach just uses existing methods on the underlying session that already have the semantics we desire. Calling `log_out()` is probably a bit heavy-handed since it will also clear the access token, but caching the token in ObjectStore didn't seem worth it as we probably need to do a full refactor of the Session anyway. 

This means that using this approach there is currently the overhead of an extra HTTP request when calling `stop/start()`.

**TODO:**
- [x] Public API
- [x] Integration tests
- [x] Unit tests
- [x] Cleanup

### DIFF
--- a/realm/realm-library/src/objectServer/java/io/realm/SyncSession.java
+++ b/realm/realm-library/src/objectServer/java/io/realm/SyncSession.java
@@ -510,6 +510,38 @@ public class SyncSession {
         }
     }
 
+    /**
+     * Attempts to start the session and enable synchronization with the Realm Object Server.
+     * <p>
+     * This happens automatically when opening the Realm instance, so doing it manually should only
+     * be needed if the session was stopped using {@link #stop()}.
+     * <p>
+     * If the session was already started, calling this method will do nothing.
+     * <p>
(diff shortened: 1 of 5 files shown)


realm/realm-library/src/objectServer/java/io/realm/SyncSession.java
```diff
@@ -510,6 +510,37 @@ public void uploadAllLocalChanges() throws InterruptedException {
         }
     }
 
+    /**
+     * Attempts to start the session and enable synchronization with the Realm Object Server.
+     * <p>
+     * This happens automatically when opening the Realm instance, so doing it manually should only
+     * be needed if the session was stopped using {@link #stop()}.
+     * <p>
+     * If the session was already started, calling this method will do nothing.
```
kneth: idempotent

PR #6149, opened 2018-09-04
Add support for using the default parameter when defining custom names

### LINKED ISSUE #6145: @RealmClass, @RealmField value name
#### Goal

I would like to be able to write `@RealmClass("name_of_the_class")` or `@RealmField("name_of_field")` instead of  `@RealmClass(name="name_of_the_class")` or `@RealmField(name="name_of_field")` which is unnecesarilly verbose I think.

#### Expected Results

The annotations classes should use `value()` for specifying name of the class/field.
The same concept use e.g. libraries parsing from JSON (e.g. GSON's `@SerializedName`)

#### Actual Results

Annotation classes use method `name()` and therefore it must be specified when using the annotations.

Realm version: 5.5.0

Android Studio version: 3.1.4

cmelchior:
Fixes #6145

### DIFF
--- a/realm-annotations/src/main/java/io/realm/annotations/RealmField.java
+++ b/realm-annotations/src/main/java/io/realm/annotations/RealmField.java
@@ -31,7 +31,6 @@ import java.lang.annotation.Target;
 public @interface RealmField {
 
     /**
-     *
      * Manually set the internal name used by Realm for this field. This will override any
      * {@link RealmNamingPolicy} set on the class or the module.
      *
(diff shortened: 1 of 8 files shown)


realm-annotations/src/main/java/io/realm/annotations/RealmField.java
```diff
@@ -31,10 +31,22 @@
...
+     */
+    String value() default "";
+
+    /**
+     *
      * Manually set the internal name used by Realm for this field. This will override any
      * {@link RealmNamingPolicy} set on the class or the module.
      *
      * @see io.realm.annotations.RealmNamingPolicy for more information about what setting the name means.
+     * @see #value()
```
nhachicha: In `RealmClass.java` the Javadoc is on the `name` field, we should probably follow it for consistency

PR #6162, opened 2018-09-15
Fix base unit tests

cmelchior:
The build transformer did not correctly keep annotations for unit test files. Most importantly it removed the `@Rule` annotation which broke pretty much all base unit tests. This was also what we saw in the Core 6 branch.


library-build-transformer/src/main/kotlin/io/realm/buildtransformer/asm/visitors/AnnotatedCodeStripVisitor.kt
```diff
@@ -54,21 +58,12 @@ class AnnotatedCodeStripVisitor(private val annotationDescriptor: String,
...
-            override fun visitEnd() {
-                if (!ignoreField) {
-                    // Call super ClassVisitor directly
-                    this@AnnotatedCodeStripVisitor.cv.visitField(access, name, descriptor, signature, value)
-                } else {
-                    logger.debug("Removing field: $name")
-                }
-            }
+    override fun visitField(access: Int, name: String?, descriptor: String?, signature: String?, value: Any?): FieldVisitor? {
+        return if (!markedFieldsInClass.contains(name)) {
```
nhachicha: Where do you update `markedFieldsInClass` with the visited name?

PR #6167, opened 2018-09-18
Update CHANGELOG.md

bmunkholm:
@cmelchior: Missing when the bug that was fixed was introduced. Merge when fixed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,17 @@
 ## 5.5.1 (YYYY-MM-DD)
 
-### Bug Fixes
+### Enhancements
+* None
 
-* Building with Android App Bundle enabled should now work correctly (#5977).
+### Fixes
+* Building with Android App Bundle enabled should now work correctly [#5977](https://github.com/realm/realm-java/issues/5977, since v?.?.?).
+
+### Compatibility
+* File format: ver. 7 (upgrades automatically from previous formats)
+* Realm Object Server: 3.0.0 or later.
+* APIs are backwards compatible with all previous release of realm-java in the 5.x.y series.
 
 ### Internal
-
 * Updated ReLinker to 1.3.0.
 
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 * None
 
 ### Fixes
-* Building with Android App Bundle enabled should now work correctly [#5977](https://github.com/realm/realm-java/issues/5977, since v?.?.?).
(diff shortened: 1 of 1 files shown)


cmelchior:
It never worked correctly. It was a new feature added by Google that wasn't supported by the 3rd party library we used to load native code.

cmelchior:
Also shouldn't the template have sections for `### Breaking changes` and `### Credits` as these have been used previously? Also why `### Fixes` instead of the much more common `### Bug Fixes` ?

CHANGELOG.md
```diff
@@ -1,11 +1,17 @@
...
 
-* Building with Android App Bundle enabled should now work correctly (#5977).
+### Fixes
+* Building with Android App Bundle enabled should now work correctly [#5977](https://github.com/realm/realm-java/issues/5977, since v?.?.?).
 
-### Internal
+### Compatibility
+* File format: ver. 7 (upgrades automatically from previous formats)
+* Realm Object Server: 3.0.0 or later.
+* APIs are backwards compatible with all previous release of realm-java in the 5.x.y series.
```
cmelchior: Isn't this redundant? I mean this is what semantic versioning means?

CHANGELOG.md
```diff
@@ -1,11 +1,17 @@
 ## 5.5.1 (YYYY-MM-DD)
 
-### Bug Fixes
+### Enhancements
+* None
 
-* Building with Android App Bundle enabled should now work correctly (#5977).
+### Fixes
```
cmelchior: Why not the much more common `Bug Fixes` which is also used by other repos

CHANGELOG.md
bmunkholm: We removed Breaking Changes as it's only used for the Major version releases (rare), so when we do a Major version it's included there. We added this comment under Compatibility to be explicit as we don't mention anywhere that we use Semantic Versioning :-)

CHANGELOG.md
bmunkholm: In some cases you can around back and forth if a small "fix" is a bug or not. It will be the same across all repos.

CHANGELOG.md
```diff
@@ -1,11 +1,17 @@
...
+* Building with Android App Bundle enabled should now work correctly [#5977](https://github.com/realm/realm-java/issues/5977, since 1.0.0).
```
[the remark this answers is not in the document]
bmunkholm: ups. good catch - fixed the link.
The value of the version is that people can see if they are affected.

PR #6183, opened 2018-09-24
Support ROS protocol v25

cmelchior:
_DO NOT MERGE UNTIL 5.6.0 HAS BEEN RELEASED_

Release PR for supporting protocol v25.

### DIFF
--- a/realm/realm-library/src/main/cpp/util.cpp
+++ b/realm/realm-library/src/main/cpp/util.cpp
@@ -50,7 +50,7 @@ void ConvertException(JNIEnv* env, const char* file, int line)
     catch (JavaExceptionThrower& e) {
         e.throw_java_exception(env);
     }
-    catch (bad_alloc& e) {
+    catch (std::bad_alloc& e) {
         ss << e.what() << " in " << file << " line " << line;
         ThrowException(env, OutOfMemory, ss.str());
     }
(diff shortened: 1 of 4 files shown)


realm/realm-library/src/main/cpp/util.cpp
```diff
@@ -50,7 +50,7 @@ void ConvertException(JNIEnv* env, const char* file, int line)
     catch (JavaExceptionThrower& e) {
         e.throw_java_exception(env);
     }
-    catch (bad_alloc& e) {
+    catch (std::bad_alloc& e) {
```
kneth: Why the need for `std::`? Line 36 is including the name space.

cmelchior: Core made some changes to exception handling, which seems to require this, but I am also unsure if this is good enough to catch the exceptions now thrown from Core

bmunkholm: @finnschiermer or @simonask ? ^

simonask: If you have both `using namespace std` and `using namespace realm::util`, `bad_alloc` is ambiguous since I put a `util::bad_alloc` in Core. It captures a backtrace from the throw-site and puts it in the `what()` description.

You can safely use `std::bad_alloc` and expect to catch both `std::bad_alloc` and `util::bad_alloc` (the latter inherits from the former). :-)

PR #6224, opened 2018-10-08
Ignore same values when using copyToRealm

cmelchior:
This PR changes the implementation of `copyToRealm` to use Object Stores `Object::create` method.
This allows us to ignore fields with the same value. Currently this generates a lot of extra set instructions when syncing Realms which has a negative performance impact on the server.

A side-effect of this is also that using this method fine-grained notifications will not report any changes for this field. However this is only true if using `copyToRealm` or `copyToRealmOrUpdate`

Implementation is done by adding a new `OsObjectBuilder` class that will build up a list of data on the C++ side before creating the whole thing. From the user API side, we have a new `ImportFlag` enum. 

Enums are being used instead of booleans or int flags since that mirros the API used by `RealmObjectSchema` for specifying field properties. It also has the upside of being more typesafe (plus I strongly suspect more flags in the future).

```
// New enum
enum ImportFlag {
  DO_NOT_SET_SAME_VALUES
}

// Usage
realm.copyToRealm(object, ImportFlag.DO_NOT_SET_SAME_VALUES)`
realm.copyToRealmOrUpdate(object, ImportFlag.DO_NOT_SET_SAME_VALUES)`
```

For the first iteration only `copyToRealm/copyToRealmOrUpdate` will use this method. We should consider adding the same flags to other methods like and `createOrUpdateFromJSon*`. `insertOrUpdate` is designed to be super performant, so it is unclear if we can support ImportFlags in this method without a lot of extra effort.

## Performance
Performance in the new implementation is pretty much on par with the current performance. When diffing is enabled performance decrease about 20% which is unavoidable as all properties needs to be compared with what is on disk.

5.8.0
image

5.7.0
image

Diffed vs. non-diffed performance in 5.8
image

## Relevant files to review
* `java_object_accessor.hpp` contains the majority of the C++ code
*  `io_realm_internal_objectstore_OsObjectBuilder.cpp` contains the JNI boundary
* `OsObjectBuilder` contains most of the Java code, but isn't really very interesting.
* See all the proxy classes for the change to the output from the annotation processor

## TODO
- [x] Cleanup Changelog
- [x] Cleanup Javadoc
- [x] Javadoc in `OsObjectBuilder`
- [x] Finish unit tests of `OsObjectBuilder`
- [x] Cleanup `OsObjectBuilder`
- [x] Modify annotation processor to use `OsObjectBuilder`
- [x] Update annotation processor unit tests.
- [x] Performance test
- [x] Add unit test for notifications triggering correctly
- [x] Any input to  name of flag? Other suggestions: `DO_NOT_WRITE_SAME_VALUES`, `IGNORE_SAME_VALUES`? Others.. The current choice is somewhat arbitrary
- [x] Send flags to C++
- [x] Hook up into Object Store (using a WIP branch)
- [x] Fix `JavaContext::print()` crashing on CI
- [x] Remove use of `util::Any` or figure another way to get performance back to normal.


realm/realm-library/src/main/cpp/java_object_accessor.hpp
```diff
@@ -0,0 +1,320 @@
...
+    // Get a string representation of the given value for use in error messages.
+    // This method should only be used when printing warnings about primary keys
+    // which means the input should only be valid types for primary keys:
+    // StringData, int64_t and Optional<int64_t>
+    std::string print(util::Any const& val) const {
+        if (!val.has_value()) {
+            return "null";
+        }
+
+        if (val.type() == typeid(StringData)) {
```
cmelchior: Isn't there a better way to just print the value than trying to anticipate what kind of types are being sent?
Also, `typeid` doesn't seem to work the same way on different architectures. I'm seeing deterministic crashes don CI that I cannot reproduce locally (which are about these checks all failing so the final logic error is thrown).

kneth: maybe you need to use `hash_code()` to get a guarantee?

realm/realm-library/src/main/cpp/java_object_accessor.hpp
```diff
@@ -0,0 +1,320 @@
...
+            auto str = any_cast<StringData>(val);
+            if (str.is_null()) {
+                return "null";
```
kneth: Consider `(null)` instead.

realm/realm-library/src/main/cpp/java_object_accessor.hpp
```diff
@@ -0,0 +1,320 @@
...
+    }
+    auto& value = any_cast<JStringAccessor&>(v);
+    StringData str = StringData(value);
+    return str;
+}
+
+template <>
+inline BinaryData JavaContext::unbox(util::Any& v, bool, bool) const
+{
+    if (!v.has_value())
```
kneth: Please use `{ ... }`

realm/realm-library/src/main/cpp/java_object_accessor.hpp
cmelchior: How would that work?

realm/realm-library/src/main/cpp/io_realm_internal_objectstore_OsObjectBuilder.cpp
```diff
@@ -0,0 +1,255 @@
...
+    (JNIEnv* env, jclass, jlong data_ptr, jstring j_key, jstring j_value)
+{
+    auto value = util::Any(JStringAccessor(env, j_value));
+    add_property(env, data_ptr, j_key, value);
+}
+
+JNIEXPORT void JNICALL Java_io_realm_internal_objectstore_OsObjectBuilder_nativeAddInteger
+        (JNIEnv* env, jclass, jlong data_ptr, jstring j_key, jlong j_value)
+{
+    auto value = util::Any(j_value);
```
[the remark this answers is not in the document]
cmelchior: Yes, but that is done here: https://github.com/realm/realm-java/pull/6224/files#diff-51482567bf1c8917ea34c6436bd2309aR224

The boundary between JNI and how the JavaContext works is a but unspecified. Currently, I try to wrap the JNI types with as little overhead as possible and then leave the heavy lifting to `JavaContext::unbox`. I should probably document that at the top of this file and somewhere in the `java_object_accessor` file

realm/realm-library/src/main/cpp/java_object_accessor.hpp
cmelchior: This method should return the string representation of whatever values is parsed in. I'm not sure I follow why `(null)` is better since I assume it involves some automatic conversion somewhere?

realm/realm-library/src/main/cpp/java_object_accessor.hpp
kneth: From https://en.cppreference.com/w/cpp/language/typeid:

"There is no guarantee that the same std::type_info instance will be referred to by all evaluations of the typeid expression on the same type, although std::type_info::hash_code of those type_info objects would be identical"

cmelchior: I'm not 100% sure, but https://stackoverflow.com/a/40357969/1389357 seems to indicate that it might not always be the case?

realm/realm-library/src/main/cpp/io_realm_internal_objectstore_OsObjectBuilder.cpp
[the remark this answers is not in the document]
cmelchior: Yes, but that is done here: https://github.com/realm/realm-java/pull/6224/files#diff-51482567bf1c8917ea34c6436bd2309aR224

The boundary between JNI and how the JavaContext works is a but unspecified. Currently, I try to wrap the JNI types with as little overhead as possible and then leave the heavy lifting to JavaContext::unbox. I should probably document that at the top of this file and somewhere in the java_object_accessor file

realm/realm-library/src/main/cpp/java_object_accessor.hpp
cmelchior: Good point. Right now we wrap all values in `''` which makes it hard to see what is strings and what is not.

CHANGELOG.md
```diff
@@ -1,3 +1,15 @@
+## 5.8.0-BETA1 (2018-10-11)
+
+### Enhancements
+* Added new `ImportFlag` class that is used to specify additional behaviour when importing
```
bmunkholm: This could be incorporated in the next section and described slightly more on what it does instead of "additional behaviour" :-) What can a user do with this now?

bmunkholm: Otherwise fine.

cmelchior:
Ready for review @nhachicha @kneth

realm/realm-library/src/main/cpp/java_object_accessor.hpp
```diff
@@ -0,0 +1,543 @@
...
+
+template <JavaValueType> struct JavaValueTypeRepr;
+template <> struct JavaValueTypeRepr<JavaValueType::Integer> { using Type = jlong; };
+template <> struct JavaValueTypeRepr<JavaValueType::String>  { using Type = std::string; };
+template <> struct JavaValueTypeRepr<JavaValueType::Boolean> { using Type = jboolean; };
+template <> struct JavaValueTypeRepr<JavaValueType::Float>   { using Type = jfloat; };
+template <> struct JavaValueTypeRepr<JavaValueType::Double>  { using Type = jdouble; };
+template <> struct JavaValueTypeRepr<JavaValueType::Date>    { using Type = Timestamp; };
+template <> struct JavaValueTypeRepr<JavaValueType::Binary>  { using Type = OwnedBinaryData; };
+template <> struct JavaValueTypeRepr<JavaValueType::Object>  { using Type = Row*; };
```
jedelbo: Why is this a Row* and not a RowExpr?

realm/realm-library/src/main/cpp/io_realm_internal_objectstore_OsObjectBuilder.cpp
```diff
@@ -0,0 +1,260 @@
...
+JNIEXPORT void JNICALL Java_io_realm_internal_objectstore_OsObjectBuilder_nativeDestroyBuilder(JNIEnv*, jclass, jlong data_ptr)
+{
+    TR_ENTER()
+    delete reinterpret_cast<OsObjectData*>(data_ptr);
+}
+
+JNIEXPORT jlong JNICALL Java_io_realm_internal_objectstore_OsObjectBuilder_nativeCreateBuilder(JNIEnv*, jclass, jlong size)
+{
+    TR_ENTER()
+    auto list = new std::vector<JavaValue>(size);
```
kneth: Can throw.

realm/realm-library/src/main/cpp/io_realm_internal_objectstore_OsObjectBuilder.cpp
```diff
@@ -0,0 +1,260 @@
...
+    return reinterpret_cast<jlong>(list);
+}
+
+static inline void add_property(JNIEnv* env, jlong data_ptr, jlong column_index, JavaValue const& value)
+{
+    try {
```
kneth: Consider move `try`/`CATCH_STD` to the actual JNI function.

realm/realm-library/src/main/cpp/io_realm_internal_objectstore_OsObjectBuilder.cpp
```diff
@@ -0,0 +1,260 @@
...
+        auto item = JavaValue(reinterpret_cast<Row*>(rows[i]));
+        list.push_back(item);
+    }
+    JavaValue value(list);
+    add_property(env, data_ptr, column_index, value);
+}
+
+static inline void add_list_element(JNIEnv* env, jlong list_ptr, JavaValue const& value)
+{
+    try {
```
kneth: Consider to move `try`/`CATCH_STD` to actual JNI functions.

realm/realm-library/src/main/java/io/realm/ImportFlag.java
```diff
@@ -0,0 +1,67 @@
...
+     * value as the value already present in the Realm.
+     * <p>
+     * For local Realms this only has an impact on change listeners which will not report changes to
+     * those fields that was not written.
+     * <p>
+     * For synchronized Realms this also impacts the server, which will see improved performance as
+     * there is less changes to upload and merge into the server Realm.
+     * <p>
+     * It also impact how the server merges changes from different devices. Realm uses a
+     * last-write-wins approach when merging individual fields in an object, so if a field is not
```
nhachicha: it could be nice to point to the blog post Simon wrote to explain the merge algorithm

realm/realm-library/src/main/java/io/realm/ImportFlag.java
```diff
@@ -0,0 +1,67 @@
...
+     *         Then Device 2 uploads its changes. Due to last-write-wins, the server version now
+     *         becomes (Field A = 3, Field B = 2).
+     *     </li>
+     * </ol>
+     * This is normally the desired behaviour as the final object is the merged result of the latest
+     * changes from both devices, however if all the fields in an object are considered an atomic
+     * unit, then this flag should not be set as it will ensure that all fields are set and thus have
+     * the same "age" when data are sent to the server.
+     */
+    DO_NOT_SET_SAME_VALUES,
```
nhachicha: I suggest to rename it to `CHECK_SAME_VALUES_BEFORE_SET` which explain clearly the cost incurred by using the option. Keeping the name `DO_NOT_SET_SAME_VALUES` hints that don't using the option is like setting `DO_SET_SAME_VALUES` which is absurd for the user perspective since it doesn't add any value

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -1630,9 +1640,21 @@ public void delete(Class<? extends RealmModel> clazz) {
 
 
     @SuppressWarnings("unchecked")
-    private <E extends RealmModel> E copyOrUpdate(E object, boolean update, Map<RealmModel, RealmObjectProxy> cache) {
+    private <E extends RealmModel> E copyOrUpdate(E object, boolean update, Map<RealmModel, RealmObjectProxy> cache, Set<ImportFlag> flags) {
         checkIfValid();
-        return configuration.getSchemaMediator().copyOrUpdate(this, object, update, cache);
+        if (!isInTransaction()) {
```
nhachicha: are you sure this check is not done in the public/caller methods?

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -1630,9 +1640,21 @@ public void delete(Class<? extends RealmModel> clazz) {
...
+            throw new IllegalStateException("`copyOrUpdate` can only be called inside a write transaction.");
+        }
+        try {
+            return configuration.getSchemaMediator().copyOrUpdate(this, object, update, cache, flags);
+        } catch (IllegalStateException e) {
+            // Convert OS exception to match old API
```
nhachicha: We need a TODO referencing an enhancement issue, we can't rely on matching strings for error code behaviour

realm/realm-library/src/main/java/io/realm/internal/NativeContext.java
```diff
@@ -30,14 +30,14 @@
     private static final ReferenceQueue<NativeObject> referenceQueue = new ReferenceQueue<NativeObject>();
     private static final Thread finalizingThread = new Thread(new FinalizerRunnable(referenceQueue));
     // Dummy context which will be used by native objects which's destructors are always thread safe.
-    static final NativeContext dummyContext = new NativeContext();
+    public static final NativeContext dummyContext = new NativeContext();
```
nhachicha: why public?

realm/realm-library/src/main/java/io/realm/internal/NativeContext.java
```diff
@@ -30,14 +30,14 @@
...
 
     static {
         finalizingThread.setName("RealmFinalizingDaemon");
         finalizingThread.start();
     }
 
-    void addReference(NativeObject referent) {
+    public void addReference(NativeObject referent) {
```
nhachicha: why public?

realm/realm-library/src/main/java/io/realm/internal/OsSharedRealm.java
```diff
@@ -141,7 +141,7 @@ public int hashCode() {
     private static final long nativeFinalizerPtr = nativeGetFinalizerPtr();
     private final long nativePtr;
     private final OsRealmConfig osRealmConfig;
-    final NativeContext context;
+    public final NativeContext context;
```
nhachicha: why public?

realm/realm-library/src/main/java/io/realm/internal/Table.java
```diff
@@ -354,7 +354,7 @@ public static void throwDuplicatePrimaryKeyException(Object value) {
     // Getters
     //
 
-    OsSharedRealm getSharedRealm() {
+    public OsSharedRealm getSharedRealm() {
```
nhachicha: public because of `OsObjectBuilder`?

realm/realm-library/src/main/java/io/realm/internal/objectstore/OsObjectBuilder.java
```diff
@@ -0,0 +1,415 @@
...
+ * The upside of making a JNI call for each property is that we do minimal allocations on the Java
+ * side. Also each method call is fairly lightweight as no checks are performed compared to using
+ * Proxy setters and {@link io.realm.internal.UncheckedRow}. The only downside is the current need for
+ * sending the key as well. Hopefully we can change that to schema indices at some point.
+ * <p>
+ * There is quite a few variants we can attempt to optimize this, but at this point we lack data
+ * that can guide any architectural design and the only way to really find out is to build out each
+ * solution and benchmark it.
+ */
+public class OsObjectBuilder {
```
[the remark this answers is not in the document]
cmelchior: I actually did that originally, but that put reclaiming resources to the mercy of the GC. This can potentially be problematic if you are copying in a lot of objects so I opted for the pattern where you manually reclaim the native resources. This is possible here since we always know precisely when we are done with the class.

CHANGELOG.md
```diff
@@ -17,14 +17,63 @@
...
+## 5.8.0-BETA2 (2018-10-19)
+
+### Enhancements
+* None
+
+### Fixed
+* `RealmResults` listeners not triggering the initial callback for Query-based Realm when the device is offline [#6235](https://github.com/realm/realm-java/issues/6235). 
+
+### Known Bugs
+* `Realm.copyToRealm()` and `Realm.copyToRealmOrUpdate` has been rewritten to support import flags. It is currently ~30% slower than in 5.7.0.
```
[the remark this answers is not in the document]
cmelchior: This is the changelog for one of the BETA's. At that point this was true even when not using the import flag. It was fixed in the final version though.

realm/realm-library/src/main/cpp/io_realm_internal_objectstore_OsObjectBuilder.cpp
```diff
@@ -0,0 +1,315 @@
...
+        (JNIEnv* env, jclass, jlong data_ptr, jlong column_index, jlong row_ptr)
+{
+    try {
+        const JavaValue value(reinterpret_cast<RowExpr*>(row_ptr));
+        add_property(data_ptr, column_index, value);
+    }
+    CATCH_STD()
+}
+
+static inline const ObjectSchema& get_schema(const Schema& schema, Table* table)
```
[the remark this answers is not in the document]
cmelchior: The `find` method currently works with ObjectStore names, which makes kinda sense.

realm/realm-library/src/main/java/io/realm/internal/NativeContext.java
cmelchior: Last year when discussing internal architecture we decided that to have a cleaner separation of concerns and make it easier for new people to come to the code base we should move relevant code from `internal` into `internal.objectstore` and `internal.core`. `OsObjectBuilder` lives in `internal.objectstore` now which means that some of the methods in `internal` are forced to become public.

realm/realm-library/src/main/java/io/realm/internal/Table.java
cmelchior: Yes.

Last year when discussing internal architecture we decided that to have a cleaner separation of concerns and make it easier for new people to come to the code base we should move relevant code from internal into `internal.objectstore` and `internal.core`. OsObjectBuilder lives in `internal.objectstore` which means that some of the methods in internal are forced to become public.

realm/realm-library/src/main/java/io/realm/internal/Util.java
```diff
@@ -147,4 +152,23 @@ public static boolean deleteRealm(String canonicalPath, File realmFolder, String
         }
         return realmDeleted;
     }
+
+    /**
+     * Converts a var arg argument list to a set ignoring any duplicates and null values.
+     */
+    public static <T> Set<T> toSet(T... items) {
```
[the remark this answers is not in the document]
cmelchior: https://github.com/realm/realm-java/pull/6224/files#diff-b77f99deee1ca52ca3dc3e47554e4e50R1040

We use varars as input parameters (similar to `RealmObjectSchema`), this is just used to convert to correct semantics for internal use

realm/realm-library/src/main/java/io/realm/Realm.java
cmelchior: Added a reference to https://github.com/realm/realm-java/issues/6262

cmelchior:
All comments addressed @nhachicha.

While it is true that ideally this should be added to all ways of importing data, it isn't clear it is worth the effort currently. It is, no matter what, outside the scope of this PR. I have created https://github.com/realm/realm-java/issues/6263 that tracks adding it to the rest of our API's

realm/realm-library/src/main/java/io/realm/ImportFlag.java
cmelchior: Good idea. Done

realm/realm-library/src/main/java/io/realm/Realm.java
cmelchior: Not sure exactly what you mean. This is the public method? And this isn't verified at the lower levels anymore as we are no longer using the setters of the `Row`.

cmelchior:
Ups. Some other comments where hiding. All should be fixed now

PR #6231, opened 2018-10-12
Expose subscriptions more directly

cmelchior:
Expose Subscriptions more directly, so it is easier to create, find and remove them. This mirrors the API's in https://github.com/realm/realm-js/pull/2060 but exposes the Subscription class more directly as it makes it easier to query, list and manipulate the subscriptions.

* Added new `Subscription` model class that represents `__ResultSets` objects.
* Added `Realm.getSubscriptions()`, `Realm.getSubscriptions(String pattern)`, `Realm.getSubscription(String name)`.
• Added `RealmQuery.subscribe()` that creates the subscription directly (requires a write transaction). 
* If `SyncConfiguration.waitForInitialRemoteData()` is set and `SyncConfiguration.initialData()` creates subscriptions, the Realm isn't opened until the subscriptions  have  downloaded data.

TODO:
- [x] Changelog
- [x] Merge https://github.com/realm/realm-object-store/pull/701 and switch back to using OS master
- [x] Fix https://github.com/realm/realm-object-store/pull/698
- [x] Add tests for initialData and getInstanceAsync
- [x] Added tests for Subscription class.


realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -361,6 +371,57 @@ private static RealmCache getCache(String realmPath, boolean createIfNotExist) {
         return (E) refAndCount.localRealm.get();
     }
 
+    /**
+     * Synchronize all initial subscriptions to disk (if needed).
+     *
+     * If this fails for a new Realm file, the file will be deleted so a new attempt can be done later.
```
kneth: Does "this" refer to the synchronization?

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -361,6 +371,57 @@ private static RealmCache getCache(String realmPath, boolean createIfNotExist) {
...
+     * Old Realm files will be left alone.
+     *
+     * This method is not threadsafe. Synchronization should happen outside it.
```
kneth: "outside" of what?

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -2007,6 +2015,63 @@ public Realm getRealm() {
         return (Realm) realm;
     }
 
+    /**
+     * Creates an anonymous subscription from this query or returns the existing Subscription if
+     * one already existed.
```
kneth: "existed" -> "exists"

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -2007,6 +2015,63 @@ public Realm getRealm() {
...
+     */
+    @ObjectServer
+    @Beta
+    public Subscription subscribe(String name) {
+        realm.checkIfValid();
+        if (realm instanceof DynamicRealm) {
+            throw new IllegalStateException("'subscribe' is not available for queries on Dynamic Realms.");
+        }
+        if (osList != null) {
+            throw new IllegalStateException("Cannot create subscriptions for queries based on a 'RealmList. Subscribe to the object holding the list instead.'");
```
kneth: Why the single quotes (`'`)?

realm/realm-library/src/main/java/io/realm/sync/Subscription.java
```diff
@@ -0,0 +1,228 @@
...
+import io.realm.RealmQuery;
+import io.realm.annotations.Beta;
+import io.realm.annotations.Index;
+import io.realm.annotations.RealmClass;
+import io.realm.annotations.RealmField;
+import io.realm.annotations.Required;
+import io.realm.internal.annotations.ObjectServer;
+
+/**
+ * Subscriptions represent the data that a device is interested in from the server when using
```
kneth: Maybe move "from the server" to right after "data"

realm/realm-library/src/main/java/io/realm/sync/Subscription.java
```diff
@@ -0,0 +1,228 @@
...
+ * Query-based Realms.
+ * <p>
+ * They are created automatically when using {@link RealmQuery#findAllAsync()} or {@link RealmQuery#findAllAsync(String)}
+ * on those Realms, but can also be created manually.
+ * <p>
+ * As long as a any subscription exists that include a Object, that object will be present on the
```
kneth: remove first "a" - "exists" -> "exist"
second "a" -> "an"

realm/realm-library/src/main/java/io/realm/sync/Subscription.java
```diff
@@ -0,0 +1,228 @@
...
+ * device. If an object is not covered by any subscription it will be removed from the device,
+ * but not the server.
+ * <p>
+ * Subscriptions are Realm objects, so deleting them by e.g. calling {@link RealmObject#deleteFromRealm()},
```
kneth: "by e.g." -> "e.g. by"

realm/realm-library/src/main/java/io/realm/sync/Subscription.java
```diff
@@ -0,0 +1,228 @@
...
+
+    /**
+     * Returns the internal name of the Class being queried.
+     *
+     * @return the internal name of the of the class being queried.
+     */
+    public String getQueryClassName() {
+        // Strip the __matches suffix to end up with the class being queried.
+        String classQueried = matchesProperty;
+        return classQueried.substring(0, classQueried.length() - "_matches".length());
```
kneth: This might not always give the expected result. Please take a look at https://github.com/realm/realm-js/pull/2075

realm/realm-library/src/objectServer/java/io/realm/internal/SyncObjectServerFacade.java
```diff
@@ -173,12 +178,19 @@ private void invokeRemoveSession(SyncConfiguration syncConfig) {
...
+    public void downloadInitialRemoteChanges(RealmConfiguration config) {
         if (config instanceof SyncConfiguration) {
             SyncConfiguration syncConfig = (SyncConfiguration) config;
             if (syncConfig.shouldWaitForInitialRemoteData()) {
                 SyncSession session = SyncManager.getSession(syncConfig);
                 try {
+                    if (!syncConfig.isFullySynchronizedRealm()) {
+                        // For Query-based Realms we want to upload all our local changes
+                        // first since those might include subscriptions the server need to process.
+                        // This means that once `downloadAllServerChanges` complete all
```
kneth: "complete" -> "completes"

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -340,6 +339,17 @@ private static RealmCache getCache(String realmPath, boolean createIfNotExist) {
...
                 // RealmMigrationNeededException might be thrown here.
                 realm = Realm.createInstance(this);
+
+                // If `waitForInitialRemoteData` data is set, we also want to ensure that all subscriptions
+                // are fully ready before proceeding. The `initialData` block is only executed for
+                // normal Realms, so initial subscriptions can only be created for those as well. Most
+                // of the Realm is initialized during a write (for atomicity). So we cannot download
+                // subscription data until all other initializers have run. At this point we also have access
+                // to all normal APIs as the schema is fully initialized.
+                // We the Realm file is new also need to reset it at this point as from the perspective of the
```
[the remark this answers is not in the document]
Zhuinden: > // We the Realm file is new also need to reset it at this point as

I'm actually not sure what is being said here! :confused:

realm/realm-library/src/main/java/io/realm/sync/Subscription.java
cmelchior: Actually, this should always be safe as we created the `*_matches` property so even if the user had a class named `foo_matches` it would become `foo_matches_matches` here and the above way only remove the last `_matches`

realm/realm-library/src/main/java/io/realm/sync/Subscription.java
```diff
@@ -0,0 +1,228 @@
...
+
+        /**
+         * The subscription has been removed. Data is no longer being synchronized from the Realm
+         * Object Server, and the objects covered by this subscription might be deleted from the
+         * device if no other subscriptions include them.
+         */
+        INVALIDATED(null);
+
+
+        private final Byte nativeValue;
```
nhachicha: why not just an int/long?

realm/realm-library/src/main/java/io/realm/sync/Subscription.java
```diff
@@ -0,0 +1,228 @@
...
+            return nativeValue;
+        }
+    }
+
+    public Subscription() {
+        // Required by Realm.
+    }
+
+    /**
+     * Creates a unmanaged named subscription from a {@link RealmQuery}.
```
nhachicha: an unmanaged

realm/realm-library/src/main/java/io/realm/sync/Subscription.java
```diff
@@ -0,0 +1,228 @@
...
+     * @return the name of the subscription.
+     */
+    public String getName() {
+        return name;
+    }
+
+    /**
+     * Returns a textual description of the query that created this subscription.
+     *
+     * @return a textual description of the query
```
nhachicha: .

realm/realm-library/src/objectServer/java/io/realm/internal/SyncObjectServerFacade.java
```diff
@@ -173,12 +177,19 @@ private void invokeRemoveSession(SyncConfiguration syncConfig) {
...
-    public void downloadRemoteChanges(RealmConfiguration config) {
+    public void downloadInitialRemoteChanges(RealmConfiguration config) {
         if (config instanceof SyncConfiguration) {
             SyncConfiguration syncConfig = (SyncConfiguration) config;
             if (syncConfig.shouldWaitForInitialRemoteData()) {
                 SyncSession session = SyncManager.getSession(syncConfig);
                 try {
+                    if (!syncConfig.isFullySynchronizedRealm()) {
+                        // For Query-based Realms we want to upload all our local changes
+                        // first since those might include subscriptions the server need to process.
```
nhachicha: needs

realm/realm-library/src/objectServer/java/io/realm/internal/SyncObjectServerFacade.java
```diff
@@ -206,4 +217,40 @@ public boolean isPartialRealm(RealmConfiguration configuration) {
     public void addSupportForObjectLevelPermissions(RealmConfiguration.Builder builder) {
         builder.addModule(new ObjectPermissionsModule());
     }
+
+    @Override
+    public void downloadInitialSubscriptions(Realm realm) {
```
nhachicha: Should we include an optional timeout parameter to cancel the wait, in case all subscriptions could not be retrieved for some reason (error/latency etc.)

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/QueryBasedSyncTests.java
```diff
@@ -336,6 +335,69 @@ public void downloadLimitedData() throws InterruptedException {
...
+                    r.where(PartialSyncObjectA.class).greaterThan("number", 5).subscribe("my-sub");
+                })
+                .waitForInitialRemoteData()
+                .modules(new PartialSyncModule())
+                .build();
+        Realm realm = Realm.getInstance(partialSyncConfig);
+        looperThread.closeAfterTest(realm);
+
+        // Check the state of subscriptions
+        assertEquals(6, realm.getSubscriptions().size());
```
nhachicha: We have 1 subscription in `initialData` where does the 6 come from?

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/QueryBasedSyncTests.java
```diff
@@ -336,6 +335,69 @@ public void downloadLimitedData() throws InterruptedException {
...
+                .waitForInitialRemoteData()
+                .addModule(new PartialSyncModule())
+                .build();
+        Realm realm = Realm.getInstance(partialSyncConfig);
+        looperThread.closeAfterTest(realm);
+
+        realm.executeTransaction(r -> {
+            Subscription sub = r.getSubscription("my-sub");
+            assertEquals(Subscription.State.ACTIVE, sub.getState());
+            sub.unsubscribe();
```
nhachicha: Can you test also that deleting the subscription will unsubscribe & make the objects disappear locally?

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -361,6 +371,57 @@ private static RealmCache getCache(String realmPath, boolean createIfNotExist) {
...
+     * If this fails for a new Realm file, the file will be deleted so a new attempt can be done later.
+     * Old Realm files will be left alone.
+     *
+     * This method is not threadsafe. Synchronization should happen outside it.
+     *
+     * @param realm Realm instance to synchronize instances for. It is safe to close this Realm if an exception is thrown.
+     * @param realmFileExistedOnDisk {@code true} if the file existed on disk before trying to open the Realm.
+     */
+    private static void synchronizeInitialSubscriptionsIfNeeded(Realm realm, boolean realmFileExistedOnDisk) {
+        if (!realmFileExistedOnDisk) {
```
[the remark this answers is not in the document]
cmelchior: `realmFileExistedOnDisk` is set before creating anything at the entry of the function, but I agree the name isn't very descriptive at this point in time. I changed the name to `realmFileIsBeingCreated` and reversed the boolean logic in the places required.

The check is here because `initialData` is only ever called for typed Realms, not DynamicRealm.

realm/realm-library/src/objectServer/java/io/realm/internal/SyncObjectServerFacade.java
cmelchior: We probably should, but that should be on the `SyncConfiguration.waitForInitialRemoteData`. I think that is outside the scope of this PR, but I'll create an issue tracking it.

realm/realm-library/src/syncIntegrationTest/java/io/realm/objectserver/QueryBasedSyncTests.java
cmelchior: Sync creates them automatically for the permission classes. I'll add a note a about it.

tgoyne:
Maybe the query API is sufficiently different to begin with that it isn't a big deal, but it's potentially confusing to add `.subscribe()` and `.unsubscribe()` methods with different semantics from what the the methods with those names do in all of the other SDKs.

Nothing else really stands out to me.

CHANGELOG.md
```diff
@@ -29,7 +49,7 @@
 * APIs are backwards compatible with all previous release of realm-java in the 5.x.y series.
 
 ### Internal
-* None
+* Updated to Object Store commit: 362b886628b3aefc5b7a0bc32293d794dc1d4ad5
```
bmunkholm: Does that solve any issues in Java?

bmunkholm:
Regarding the comment from Thomas:
> Maybe the query API is sufficiently different to begin with that it isn't a big deal, but it's potentially confusing to add `.subscribe()` and `.unsubscribe()` methods with different semantics from what the methods with those names do in all of the other SDKs.

I think it's a problem if we make naming the same but semantics different. We should avoid that if at all possible.

cmelchior:
> I think it's a problem if we make naming the same but semantics different. We should avoid that if at all possible.

Yes, we should. But in this case, it isn't possible.

It is virtually impossible to make these API's similar as the way queries are performed and run are so very different between Java and Cocoa/JS and changing that would be a massive, massive change for all existing users.

The name `subscribe()` encapsulates the meaning of actually creating the subscription on both platforms, so even though the semantics are different, the end result is more or less the same.

PR #6234, opened 2018-10-17
Fix.

HTopete:
In this way the context is obtained.


Zhuinden:
Application is also a `context`, and `application.getApplicationContext()` can be null on some devices I think.

So what does this solve?

HTopete:
Kotlin could not find the context crashing, that's how I was able to find it.

Issue #6235, opened 2018-10-18
FindAllAsync doesn't trigger the INITIAL callback for Query-based Realms when offline


cmelchior:
To reproduce:

1) Use the `objectServerExample`
2) Log into the app to create a user
3) See that the changelistener for the async query is correctly called.
4) Kill the app
5) Go into offline mode
6) Open the app again
7) See that the listener for the async query isn't called when the query completes but is called when a write is made.

cmelchior:
Fixed in #6236

PR #6236, opened 2018-10-18
Fix initial listener not triggering for query-based listeners

cmelchior:
Closes https://github.com/realm/realm-java/issues/6235

This bug caused the INITIAL callback from Object Store to never trigger for Query-based Realms. For devices that only created a query, this could mean they never saw the callback until the device came online as that caused a write from the server triggering the listener again.

From the perspective of the user they still got the INITIAL callback, but only for the 2nd callback from Object Store.


CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## 5.7.1 (YYYY-MM-DD)
+
+### Fixes
+* `RealmResults` listeners not triggering the initial callback for Query-based Realm when the device is offline [#6235](https://github.com/realm/realm-java/issues/6235).
```
bmunkholm: Since when?

PR #6243, opened 2018-10-23
Fixed changelog year

daewin:


cmelchior:
Thanks for doing this, but it was also fixed in in https://github.com/realm/realm-java/pull/6245, so closing this one.

PR #6266, opened 2018-11-01
Add support for latest version of Sync

cmelchior:
This upgrades all dependencies to latest version of Sync.
This will also enable setting the user agent. This API will be added in another API shortly.


realm/realm-library/src/main/cpp/io_realm_SyncManager.cpp
```diff
@@ -97,7 +97,8 @@ JNIEXPORT void JNICALL Java_io_realm_SyncManager_nativeInitializeSyncManager(JNI
     TR_ENTER()
     try {
         JStringAccessor base_file_path(env, sync_base_dir); // throws
-        SyncManager::shared().configure_file_system(base_file_path, SyncManager::MetadataMode::NoEncryption);
+        std::string user_agent_info(""); // TODO Add support for this
```
nhachicha: is there an issue number we can reference?

PR #6270, opened 2018-11-02
Add support for setting a custom User-Agent

### LINKED ISSUE #6267: Add support for custom user agent strings
The latest Sync/ROS release added support for defining a custom user agent string that is sent when creating a session. This can help debugging and identifying problematic clients.

The user agent string consists of 3 parts:
1) A part defined by Sync
2) A part defined by us (Realm version / Device / OS version)
3) A part defined by the user

We need to implement 2) and expose 3) to end users.

Suggested API is:

```
Realm.init(Context context, String userAgent)
```

As the UserAgent can only be defined up to the point we the C++ `SyncManager` this will fit both that and the semantics that the user agent cannot be changed once the `Client` is started.

Internal ref: https://github.com/realm/realm-sync/issues/2455#issuecomment-432447520

cmelchior:
Closes #6267 

This PR adds an optional parameter when calling `Realm.init()`. This will set the user part of the User-Agent string.

I'm using `Realm.init()` as people are only calling this once and we also initialize the SyncManager as part of calling it. The extra constructor is only visible when using a build supporting the Object Server.

```
// Usage
Realm.init(context, "MyApp/" + BuildConfig.VERSION_NAME);
```

The String constructed by RealmJava will look like this for e.g. OnePlus:
```
// Example of String for OnePlus
RealmJava/5.8.0-SNAPSHOT (OnePlus5T, ONEPLUS A5010, v27) ObjectServerExample/1.0
```

The server will output something like this:
```
// Server connection example from emulator
2018-11-02T10:52:30.103Z - info: [sync] Sync Connection[29]: Session[1]: Client info: (path='/9e896ae0c1ff90402bc978cab085659f/tests', user='9e896ae0c1ff90402bc978cab085659f', from=[127.0.0.1]:46980, protocol=25) RealmSync/3.13.1 (Android Linux 4.4.34-genymotion #1 SMP PREEMPT Mon Dec 5 15:49:16 UTC 2016 i686) RealmJava/5.8.0-SNAPSHOT (vbox86p, Custom Phone - 7.1.0 - API 25 - 768x1280_1, v25) ObjectServerExample/1.0
```

Specifically, we are sending 

* Version name for `RealmJava`.
* Build.DEVICE
* Build.MODEL
* Build.VERSION.SDK_INT (named SDK is deprecated)

Build.PRODUCT sees to contain the same information as DEVICE/MODEL so doesn't look relevant.

@roberhofer Is this sufficient?


realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -261,6 +261,56 @@ public RealmSchema getSchema() {
...
+     * <manifest xmlns:android="http://schemas.android.com/apk/res/android" package="io.realm.example">
+     * <application android:name=".MyApplication">
+     *   // ...
+     * </application>
+     * </manifest>
+     * }
+     * </pre>
+     *
+     * @param context the Application Context.
+     * @param userAgent optional user defined string that will be sent to the Realm Object Server
```
kneth: Maybe it should be clear that if your app doesn't sync, `userAgent` is ignored.

realm/realm-library/src/main/java/io/realm/Realm.java
```diff
@@ -261,6 +261,56 @@ public RealmSchema getSchema() {
...
+     *                  as part of a {@code User-Agent} header when a session is established.
```
nhachicha: remove extra spaces

realm/realm-library/src/objectServer/java/io/realm/ObjectServer.java
```diff
@@ -33,7 +36,7 @@
 @Keep
 class ObjectServer {
 
-    public static void init(Context context) {
+    public static void init(Context context, String appDefinedUserAgent) {
```
nhachicha: use an overload or varargs to make this parameter optional, so: 
1- you don't break current API
2- don't force the user to think about this parameter (especially in the context of non sync)

realm/realm-library/src/objectServer/java/io/realm/ObjectServer.java
```diff
@@ -42,6 +45,31 @@ public static void init(Context context) {
...
+            sb.append(" (");
+            sb.append(Util.isEmptyString(Build.DEVICE) ? "unknown-device" : Build.DEVICE);
+            sb.append(", ");
+            sb.append(Util.isEmptyString(Build.MODEL) ? "unknown-model" : Build.MODEL);
+            sb.append(", v");
+            sb.append(Build.VERSION.SDK_INT);
+            sb.append(")");
+
+            // Setup User part of User-Agent string
+            if (!appDefinedUserAgent.isEmpty()) {
```
nhachicha: this could be null optional, use `Util.isEmpty` instead

realm/realm-library/src/objectServer/java/io/realm/SyncManager.java
```diff
@@ -92,6 +92,11 @@
      */
     public static String APP_ID = null;
 
+    /**
+     * User-Agent string sent to the Realm Object Server when a session is established.
+     */
+    public static String USER_AGENT = "";
```
nhachicha: this looks unused, remove

realm/realm-library/src/objectServer/java/io/realm/ObjectServer.java
cmelchior: I think there is a misunderstanding. This is a briding class only used by internal code (the class is package private). In the public API specifying the user agent is optional

PR #6286, opened 2018-11-06
Transformer strips similar named methods

cmelchior:
The library-transformer strips all methods marked `@ObjectServer`. Unfortunately, it seems it also strips similar named methods even though they don't have this annotation. In this case `ObjectServerFacade.init(Context, String)`. Renaming the internal method is the quickest way to make 5.8.0 ready for release.

I'll be making a fix for the behaviour after the release.


cmelchior:
Fast-merge

PR #6301, opened 2018-11-15
Update LiveData value when becoming active

RemiLatapy:
Allow LiveData to stay up to date, especially when resuming an activity

It solves this issue: https://stackoverflow.com/questions/49981734/observer-for-android-livedata-not-called-but-it-is-with-observeforever


bmunkholm:
@cmelchior Looks like this needs another review

Issue #6364, opened 2018-12-07
ErrorCode is not correct


cmelchior:
The enum names for the Realm Object Server error codes in `ErrorCode.java` seems to have diverged from their current meaning. 

This is unfortunate as it is most likely a breaking change to fix it, but we need to update them regardless so they are accurate.

Example is `BAD_SERVER_VERSION_DOWNLOAD(111)` which is `bad_changeset_header_syntax(111)`

----

EDIT: This is actually more severe than that. Core reports errors using `(category, code)`, but only the code is sent to Java, this means that some errors are not being mapped correctly.  We need to fix this so error mapping are done using both the error code and category

PR #6379, opened 2019-01-03
Add support for native error category and code

cmelchior:
Closes https://github.com/realm/realm-java/issues/6364
Closes https://github.com/realm/realm-java/issues/6381

Before this PR, native errors where only mapped to Java error codes using their integer value. However, C++ errors are uniquely identified using the pair of `(category, int_value)` and not `int_value` alone. This resulted in some errors being reported as the wrong type. This PR fixes this so Java now uses both values to identify native errors.

Java already has a catch-all error type named `UNKNOWN`. This will now report the native values when encountered which should ease debugging of errors we did not anticipate.

The PR also add support for a few common connection issues thrown by Linux as there is a high chance users will see them in real world apps, and it is easier to handle them this way instead of trying to map the UNKNOWN type to them. 

Java already had a category concept for errors, which makes the terminology a bit icky. For now, the term "Type" is used when we are talking about the C++ error category. We should consider fixing this in the next major release as it will be a breaking change.

Also, for a naming scheme, I have started prefixing the new enum values so we can have similar errors from different categories without having name clashes.

TODO:
- [x] Define the correct error category for all supported types
- [x] Verify that errors are being reported correctly from Sync and Client Reset
  - [x] Protocol level errors
  - [x] Session level errors
  - [x] Object store level errors, especially Client Reset
  - [x] Java level errors (auth)


realm/realm-library/src/objectServer/java/io/realm/ObjectServerError.java
```diff
@@ -56,6 +77,17 @@ public ObjectServerError(ErrorCode errorCode, Throwable exception) {
         this(errorCode, null, exception);
     }
 
+    /**
+     * Errors happening while trying to authenticate a user.
+     *
+     * @param errorCode error code for this type of error.
+     * @param title Title for this type of error.
```
kneth:
```suggestion
     * @param title title for this type of error.
```

realm/realm-library/src/objectServer/java/io/realm/ErrorCode.java
```diff
@@ -212,8 +291,22 @@ public static ErrorCode fromException(Exception exception) {
         }
     }
 
+    public static class Type {
+        public static final String AUTH = "auth"; // Errors from the Realm Object Server
+        public static final String CONNECTION = "realm.basic_system"; // Connection/System errors from the native Sync Client
+        public static final String DEPRECATED = "deprecated"; // Deprecated errors
+        public static final String HTTP = "http"; // Errors from the HTTP layer
```
nhachicha: does it include https ssl errors (like invalid certificat/PEM etc)?

nhachicha:
`permissionManagerAsyncTask_handleTwoErrorsDifferentErrorCode` is failing on CI

realm/realm-library/src/main/cpp/io_realm_internal_OsRealmConfig.cpp
```diff
@@ -281,10 +287,39 @@ JNIEXPORT jstring JNICALL Java_io_realm_internal_OsRealmConfig_nativeCreateAndSe
...
+            // For this reason we manually map the constants to the error integer values defined in Java.
+            // For simplicity Java re-use the values currently defined in errno.h.
+            if (std::strcmp(error_category, "realm.basic_system") == 0) {
+                switch(error_code) {
+                    case ECONNRESET: error_code = 104; break;
+                    case ESHUTDOWN: error_code = 110; break;
+                    case ECONNREFUSED: error_code = 111; break;
+                    case EADDRINUSE: error_code = 112; break;
+                    case ECONNABORTED: error_code = 113; break;
+                    default:
```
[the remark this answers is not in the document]
cmelchior: We can see a lot more error codes here. I'm just mapping the few ones we care especially about. The default case just parses on the other one and they end up being mapped to `UNKNOWN` in Java. The ones with cases here have specific enums in Java.

PR #6413, opened 2019-01-27
Convert ObjectServerExample to Kotlin and use data binding

cmelchior:
Done in order to have a example using Sync written in Kotlin

Internal: See https://realmio.slack.com/archives/C0E1CGS03/p1548449974015500?thread_ts=1548196688.011800&cid=C0E1CGS03


examples/objectServerExample/src/main/java/io/realm/examples/objectserver/CounterActivity.kt
```diff
@@ -0,0 +1,178 @@
...
+        binding.lower.setOnClickListener { adjustCounter(-1) }
+    }
+
+    override fun onStart() {
+        super.onStart()
+        user = loggedInUser
+        user?.let {
+
+            // Create a RealmConfiguration for our user
+            val config = user!!.createConfiguration(BuildConfig.REALM_URL)
```
Zhuinden: prefer `val user = loggedInUser; if(user != null) {` to have safe-casting.

examples/objectServerExample/src/main/java/io/realm/examples/objectserver/CounterActivity.kt
```diff
@@ -0,0 +1,178 @@
...
+                    .initialData { realm -> realm.createObject(CRDTCounter::class.java, it.identity) }
```
Zhuinden: I think you have an inline reified function in `io.realm.kotlin` for this.

examples/objectServerExample/src/main/java/io/realm/examples/objectserver/CounterActivity.kt
```diff
@@ -0,0 +1,178 @@
...
+    override fun onCreateOptionsMenu(menu: Menu): Boolean {
+        menuInflater.inflate(R.menu.menu_counter, menu)
+        return true
+    }
+
+    override fun onOptionsItemSelected(item: MenuItem): Boolean {
+        return when (item.itemId) {
+            R.id.action_logout -> {
+                realm.close()
+                user?.let  {
```
Zhuinden: Prefer `val user = user; if(user != null) {`

examples/objectServerExample/src/main/java/io/realm/examples/objectserver/CounterActivity.kt
```diff
@@ -0,0 +1,178 @@
...
+                }
+                true
+            }
+
+            else -> super.onOptionsItemSelected(item)
+        }
+    }
+
+    private fun updateProgressBar(downloading: Boolean, uploading: Boolean) {
+        @ColorRes var color = android.R.color.black
```
Zhuinden: You can use 

``` kotlin
val color = when {
    downloading && uploading -> R.color.progress_both
    downloading -> R.color.progress_download
    uploading -> R.color.progress_upload
    else -> android.R.color.black
}
```

examples/objectServerExample/src/main/java/io/realm/examples/objectserver/CounterActivity.kt
```diff
@@ -0,0 +1,178 @@
...
+        }
+        progressBar.indeterminateDrawable.setColorFilter(resources.getColor(color), PorterDuff.Mode.SRC_IN)
+        progressBar.visibility = visibility
+    }
+
+    private fun adjustCounter(adjustment: Int) {
+        // A synchronized Realm can get written to at any point in time, so doing synchronous writes on the UI
+        // thread is HIGHLY discouraged as it might block longer than intended. Use only async transactions.
+        realm.executeTransactionAsync { realm ->
+            val counter = realm.where(CRDTCounter::class.java).findFirst()
```
Zhuinden: I think you have an inline reified function for this

examples/objectServerExample/src/main/java/io/realm/examples/objectserver/LoginActivity.kt
```diff
@@ -0,0 +1,124 @@
...
+        finish()
+    }
+
+    private fun onLoginFailed(errorMsg: String) {
+        loginButton.isEnabled = true
+        createUserButton.isEnabled = true
+        Toast.makeText(baseContext, errorMsg, Toast.LENGTH_LONG).show()
+    }
+
+    private fun validate(): Boolean {
```
Zhuinden: I think you can reduce this to:

``` kotlin
private fun validate(): Boolean {
    username.text.toString().takeIf { it.isNotEmpty() } ?: return false

    password.text.toString().takeIf { it.isNotEmpty() } ?: return false

    return true
}
```

or

``` kotlin
private fun validate(): Boolean = when {
    username.text.toString().isEmpty() -> false
    password.text.toString().isEmpty() -> false
    else -> true
}
```

whichever seems nicer.

examples/objectServerExample/src/main/java/io/realm/examples/objectserver/model/CRDTCounter.kt
```diff
@@ -0,0 +1,41 @@
...
+ */
+package io.realm.examples.objectserver.model
+
+import io.realm.MutableRealmInteger
+import io.realm.RealmObject
+import io.realm.annotations.PrimaryKey
+import io.realm.annotations.Required
+
+/**
+ * A named, conflict-free replicated data-type.
```
Zhuinden: I know this is inherited, but this comment isn't that helpful, tbh. :smile:

examples/objectServerExample/src/main/java/io/realm/examples/objectserver/model/CRDTCounter.kt
```diff
@@ -0,0 +1,41 @@
...
+ */
+open class CRDTCounter(@PrimaryKey var name: String): RealmObject() {
+
+    @Required
+    private val counter = MutableRealmInteger.valueOf(0L)
+
+    val count: Long
+        get() = this.counter.get()!!.toLong()
+
+    constructor() : this("") {
```
Zhuinden: Why not just do:

``` kotlin
open class CRDTCounter: RealmObject() {
    @PrimaryKey
    var name: String

    @Required
    private val counter = MutableRealmInteger.valueOf(0L)
}
```

that way you don't need to juggle a secondary constructor.

Otherwise you could do `@JvmOverloads constructor(@PrimaryKey var name: String = "")` anyways, but I think that is worse.

examples/objectServerExample/src/main/java/io/realm/examples/objectserver/model/CRDTCounter.kt
```diff
@@ -0,0 +1,41 @@
...
+        // Required by Realm
+    }
+
+    fun incrementCounter(delta: Long) {
+        counter.increment(delta)
```
Zhuinden: Now I wonder why `MutableRealmInteger` doesn't get a `+=` extension operator from `io.realm.kotlin` :wink:

examples/objectServerExample/src/main/java/io/realm/examples/objectserver/CounterActivity.kt
cmelchior: I'm not 100% sure why you feel this way, because it feels more verbose? Is it because `let` actually returns a value or...?

examples/objectServerExample/src/main/java/io/realm/examples/objectserver/CounterActivity.kt
cmelchior: `it` is actually safe here.

examples/objectServerExample/src/main/java/io/realm/examples/objectserver/model/CRDTCounter.kt
cmelchior: Yes, we should do that 😄

examples/objectServerExample/src/main/java/io/realm/examples/objectserver/CounterActivity.kt
```diff
@@ -151,26 +152,20 @@ class CounterActivity : AppCompatActivity() {
...
-        } else {
-            visibility = View.GONE
+        @ColorRes val color = when {
+            downloading && uploading -> R.color.progress_both
+            downloading -> R.color.progress_download
+            uploading -> R.color.progress_upload
+            else -> android.R.color.black
         }
         progressBar.indeterminateDrawable.setColorFilter(resources.getColor(color), PorterDuff.Mode.SRC_IN)
-        progressBar.visibility = visibility
```
Zhuinden: But did you intend to remove this completely? :thinking:

examples/objectServerExample/src/main/java/io/realm/examples/objectserver/CounterActivity.kt
```diff
@@ -92,16 +93,16 @@ class CounterActivity : AppCompatActivity() {
         user?.let {
 
             // Create a RealmConfiguration for our user
-            val config = user!!.createConfiguration(BuildConfig.REALM_URL)
-                    .initialData { realm -> realm.createObject(CRDTCounter::class.java, it.identity) }
+            val config = it.createConfiguration(BuildConfig.REALM_URL)
```
Zhuinden: I really think the usage of `it` here makes it hard to read. I think either my initial proposition, or `user?.let { syncUser ->` would be better.

cmelchior: Yeah, I could agree with that. In larger code blocks it can be a bit hard to reason about (at a glance) what `it` actually is. A named parameter, in that case, is probably preferable.

examples/objectServerExample/src/main/java/io/realm/examples/objectserver/CounterActivity.kt
Zhuinden: The idea is that you can avoid `?.let { it ->` with it. But `it` can be named too if you want. It's just that if you name it the same thing, then the lint starts yelling at you that it's "name-shadowing"

examples/objectServerExample/src/main/java/io/realm/examples/objectserver/LoginActivity.kt
```diff
@@ -0,0 +1,114 @@
...
+            override fun onSuccess(user: SyncUser) {
+                progressDialog.dismiss()
+                onLoginSuccess()
+            }
+
+            override fun onError(error: ObjectServerError) {
+                progressDialog.dismiss()
+                val errorMsg: String
+                when (error.errorCode) {
+                    ErrorCode.UNKNOWN_ACCOUNT -> errorMsg = "Account does not exists."
```
Zhuinden: I know this is inherited, but it should be `does not exist.`

examples/objectServerExample/build.gradle
```diff
@@ -13,13 +28,17 @@ android {
...
+    }
+
     buildTypes {
         // Go to https://cloud.realm.io and copy the URL to your instance. Insert it below.
         // It will look something like "https://test.us1.cloud.realm.io"
         //
         // If you're running a self-hosted version, use the hostname/IP address of the Realm Object
         // Server, e.g "http://127.0.0.1:9080".
-        def rosUrl = "<INSERT_REALM_OBJECT_SERVER_URL>"
+        def rosUrl = "https://cmelchior.us1.cloud.realm.io"
```
cmelchior:
```suggestion
        def rosUrl = "<INSERT_REALM_OBJECT_SERVER_URL>"
```

examples/objectServerExample/src/main/java/io/realm/examples/objectserver/CounterActivity.kt
```diff
@@ -0,0 +1,175 @@
...
+        counterView = binding.textCounter
+        progressBar = binding.progressbar
+        binding.upper.setOnClickListener { adjustCounter(1) }
+        binding.lower.setOnClickListener { adjustCounter(-1) }
+    }
+
+    override fun onStart() {
+        super.onStart()
+        user = loggedInUser
+        val user = user
```
darrinps: Not sure the reason for making a function local copy of the user is here.

examples/objectServerExample/src/main/java/io/realm/examples/objectserver/MyApplication.kt
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016 Realm Inc.
+ * Copyright 2018 Realm Inc.
```
darrinps: Shouldn't this be 2019?

examples/objectServerExample/src/main/java/io/realm/examples/objectserver/CounterActivity.kt
```diff
@@ -0,0 +1,175 @@
+/*
+ * Copyright 2018 Realm Inc.
```
darrinps: Shouldn't this be 2019?

examples/objectServerExample/src/main/java/io/realm/examples/objectserver/CounterActivity.kt
```diff
@@ -0,0 +1,175 @@
...
+    private lateinit var progressBar: MaterialProgressBar
+    private lateinit var counters: RealmResults<CRDTCounter> // Keep strong reference to counter to keep change listeners alive.
+
+    private val loggedInUser: SyncUser?
+        get() {
+            var user: SyncUser? = null
+
+            try {
+                user = SyncUser.current()
+            } catch (ignore: IllegalStateException) {
```
darrinps: Shouldn't we at least log when this happens so it doesn't go unnoticed?

examples/objectServerExample/src/main/java/io/realm/examples/objectserver/CounterActivity.kt
```diff
@@ -0,0 +1,175 @@
...
+        menuInflater.inflate(R.menu.menu_counter, menu)
+        return true
+    }
+
+    override fun onOptionsItemSelected(item: MenuItem): Boolean {
+        return when (item.itemId) {
+            R.id.action_logout -> {
+                realm.close()
+                val user = user
+                if (user != null) {
```
darrinps: The != null isn't so much "the Kotlin way". Have a look at ?.let instead (https://discuss.kotlinlang.org/t/let-vs-if-not-null/3542)

examples/objectServerExample/src/main/java/io/realm/examples/objectserver/CounterActivity.kt
Zhuinden: To make it final for the next line to enable smart-casting 😉

examples/objectServerExample/src/main/java/io/realm/examples/objectserver/CounterActivity.kt
Zhuinden: Actually it depends on context. `user?.let { user ->` works too, but then the IDE nags you about name shadowing.

darrinps: You don't have to call it user. Renaming it might make it more clear and get around the nag without using the != maybe?

Anyway, it's not a big deal. Just not what I'm used to seeing.

cmelchior: I suspect this is one of those small details the Kotlin community is still deciding on what is considered "best practice". I started out using `user?.let { it }` as well, but you can then argue that it can be hard to deduce what `it` is. Giving it a name obviously fixes that, but then the name is different from the variable.

I don't have strong opinions either way and as you point out there might be different trade-offs. I suspect this is just one of those things that each team has to decide on their own until the community gravitates towards a "default" way of doing this.

Issue #6416, opened 2019-01-30
Support Realms on external storage that are using FAT32


cmelchior:
Originally discovered on React Native (Android): https://github.com/realm/realm-js/issues/2062
Realm Java is affected by the same problem

With https://github.com/realm/realm-object-store/pull/765 merged we should automatically set the `Realm::Config::fifoFallbackPath` to `Context.getCacheDir()`. There is probably no reason to expose the setting in the public API as there are a very limited set of places a user could put these files and I'm hard pressed to find any use cases for not just using the internal cache dir.

cmelchior:
Ups, looks like we already support this through `OsSharedRealm.getTemporaryDirectory()`. Closing.

PR #6433, opened 2019-02-18
Add support for fifo fallback dir

cmelchior:
Closes https://github.com/realm/realm-java/issues/6416

Keeping this as an internal API for now as I'm hard pressed to find any use cases for having a public API for this.


cmelchior:
We already support this though `OsSharedRealm.getTemporaryDirectory()`. Closing

Issue #6426, opened 2019-02-12
Selected backlink closures for subscriptions


kneth:
See Epic: [realm/engineering#13](https://github.com/realm/engineering/issues/13)
Helps unblock a adapter performance issue.

PR #6489, opened 2019-04-10
Including LinkingObjects in subscriptions

cmelchior:
Closes #https://github.com/realm/realm-java/issues/6426

API:
```
// Throws if called on non-query-based Realm
RealmQuery.includeLinkingObjects(String property)

// Usage: simple case
realm.where(Person.class)
  .includeLinkingObjects("parents")
  .subscribe("my-subscription");

// Usage: Advanced case with implicit subscription case
realm.where(Person.class)
  .equalTo("name", "Jane")
  .sort("age", Sort.ASCENDING)
  .includeLinkingObjects("parents", "other_parents")
  .findAllAsync();
```

I opted for the `RealmQuery.includeLinkingObjects()` method instead of e.g `RealmQuery.subscribe()` since this method already has 4 parameters in the worst case and we are having a combinatorial explosion trying to accommodate the many variants.  Also both `Clean Code` and `Effective Java` recommend either splitting up methods or introducing builders in that case. The method is stripped out when building without sync as well. 

TODO
- [x] Update changelog
- [x] Add new API
- [x] Tests


ironage:
Couldn't the inclusions be part of the subscribe options along with the subscription name? That way there's no ambiguity about having inclusions on a local (non subscription) query. The js implementation does it this way.

cmelchior:
Ready for review

realm/realm-library/src/main/cpp/io_realm_internal_core_IncludeDescriptor.cpp
```diff
@@ -0,0 +1,82 @@
...
+#include "java_accessor.hpp"
+#include "java_query_descriptor.hpp"
+#include "util.hpp"
+
+
+using namespace realm;
+using namespace realm::util;
+using namespace realm::_impl;
+
+static void finalize_descriptor(jlong ptr);
```
kneth: Why do you have a prototype and an implementation?

realm/realm-library/src/main/cpp/io_realm_internal_core_IncludeDescriptor.cpp
```diff
@@ -0,0 +1,82 @@
...
+{
+    TR_ENTER()
+    try {
+        JLongArrayAccessor table_arr(env, table_pointers);
+        JLongArrayAccessor index_arr(env, column_indexes);
+        auto starting_table = reinterpret_cast<Table*>(starting_table_ptr);
+        std::vector<LinkPathPart> parts;
+        parts.reserve(index_arr.size());
+        for (int i = 0; i < index_arr.size(); ++i) {
+            auto col_index = size_t(index_arr[i]);
```
kneth: `static_cast<size_t>(...)`

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -2054,6 +2054,42 @@ private OsResults lazyFindAll() {
...
     }
 
+    /**
+     * This predicate is only relevant for Query-based Realms.
+     * <p>
+     * Objects referenced through fields marked with {@link io.realm.annotations.LinkingObjects} are normally not downloaded
+     * as part of the subscription in Query-based Realms, but by using this predicate, it is possible to specify which linking
+     * objects relationships should also be included in the subscription as well.
+     * <p>
+     * Note that all "forward" object references like object references and lists are always included in the subscription
```
kneth: Is it worth to mention that links and lists are followed eagerly?

realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -2054,6 +2054,42 @@ private OsResults lazyFindAll() {
         return this;
     }
 
+    /**
+     * This predicate is only relevant for Query-based Realms.
+     * <p>
+     * Objects referenced through fields marked with {@link io.realm.annotations.LinkingObjects} are normally not downloaded
```
kneth: If we have two classes `A` and `B`, and three objects `a1`, `a2`, and `b1`. If `a1` has a linking object to `b1`, and `b1` has a link to `a2`, all three objects will be included, when the query is on `A`, right?

cmelchior: Good question. I would assume yes.

kneth: I think it is important to document the behaviour. Maybe the API doc isn't the right place.

cmelchior: We should probably explain it in more details here: https://docs.realm.io/sync/using-synced-realms/syncing-data#subscribing-to-data-reading-data

PR #6496, opened 2019-04-19
Add support for incremental annotation processors

### LINKED ISSUE #5906: Add support for incremental annotation with Gradle 4.7
As Gradle 4.7 is out, there had to be one issue for this :)
The documentation to handle this is [here](https://docs.gradle.org/4.7/userguide/java_plugin.html#sec:incremental_annotation_processing).
I think Realm-java could implement this ? (but I might be wrong as I don't fully understand all the works behind annotation processors).

cmelchior:
Closes #5906 

I don't think there is any way to add automated tests for this so I did smoke tests doing the following:

- [x] Adding / Removing simple String fields
- [x] Adding / Removing Object reference fields
- [x] Adding / Removing model classes
- [x] Modifying non-Realm files.
- [x] Check log output. Especially looking for something like `Gradle may disable incremental compilation as the following annotation processors are not incremental: realm-annotations-processor-5.9.1.jar (io.realm:realm-annotations-processor:5.9.1).`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,36 +1,53 @@
-## 5.10.1(YYYY-MM-DD)
+## 5.11.0(YYYY-MM-DD)
 
-## Enhancements 
-* None
+### Enhancements 
+* Added support for incremental annotation processing added in Gradle 4.7. (Issue [#5906](https://github.com/realm/realm-java/issues/5906)).
 
-## Fixed
-*  Native crash happening if bulk updating a field in a `RealmResult` would cause the object to no longer be part of the query result. (Issue [#6478](https://github.com/realm/realm-java/issues/6478), since 5.8.0).
+### Fixed
+* None.
 
-## Compatibility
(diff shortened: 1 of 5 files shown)


bingranl:
Hi Christian @cmelchior , thanks for adding support for incremental annotation processing! Do we have a release plan for this feature?

bingranl:
FYI If you are interested in end to end testing for incremental feature, this is an example for Lifecycle INCAP support. https://android.googlesource.com/platform/frameworks/support/+/refs/heads/androidx-master-dev/lifecycle/integration-tests/incrementality/src/test/kotlin/androidx/lifecycle/IncrementalAnnotationProcessingTest.kt

Zhuinden:
@bingranl it was released in 5.11.0 (2019-05-01)

bingranl:
Thanks, @Zhuinden I just realized it was released! And I found the commit in 5.10.0. :)

PR #6499, opened 2019-04-24
Querying the existence of data

anggrayudi:
I am very often using this query to check if Realm contains specific data in a table. We could prevent code repetition using these two methods.

In many lines, I usually write the following code:
```java
if (realm.where(Purchase.class).equalTo("sku", "pro").equalTo("purchased", true).exists()) {
    // activate paid items
}
```

### DIFF
--- a/realm/realm-library/src/main/java/io/realm/RealmQuery.java
+++ b/realm/realm-library/src/main/java/io/realm/RealmQuery.java
@@ -1500,6 +1500,25 @@ public class RealmQuery<E> {
         return this;
     }
 
+    /**
+     * Equal to {@code query.count() > 0}
+     *
+     * @return {@code true} if data found in this query condition.
+     * @see #count()
+     */
+    public boolean exists() {
+        return count() > 0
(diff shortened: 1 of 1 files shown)


Zhuinden:
Hmm, I'm not an official Realm member, but I think while `exists()` sounds nice (although also easily possible to make into an extension function if you use Kotlin), the `noData()` name isn't that great for the other case.

I'd just prefer `!realmQuery.exists()`.

anggrayudi:
I forgot to mention that we need to optimize `exists()` performance by adding `limit(1)`. So it will look like this:
```java
public boolean exists() {
    return limit(1).count() > 0
}
```

cmelchior:
Given that `exists()` is just a shorthand for `.count() > 0` or, as you point out `.limit(1).count() > 0`, then I'm not sure it is worth adding it to the API surface. It is just 1 extra method sure, but we also need to factor in the API surface of `RealmQuery` which is already rather large.

PR #6504, opened 2019-05-01
Improve compatibility section in changelog and API docs

cmelchior:


realm/realm-library/src/main/java/io/realm/RealmQuery.java
```diff
@@ -2065,7 +2065,9 @@ private OsResults lazyFindAll() {
      * subscription by default.
      * <p>
      * This predicate can be called multiple times, in which case all fields will be added to the subscription.
-     * 
+     * <p>
+     * NOTE: This method version is only available with Realm Object Server 3.21.1 or later.
```
bmunkholm:
```suggestion
     * NOTE: This method is only supported when connecting to Realm Object Server 3.21.0 or later. If you use it with previous versions of Realm Object Server, an 'InvalidSyntaxException' will be observable via the ChangeListeners getError() method on the changeset returned.
```
Perhaps rephrase the error returned?

CHANGELOG.md
```diff
@@ -16,7 +18,7 @@
 *  Native crash happening if bulk updating a field in a `RealmResult` would cause the object to no longer be part of the query result. (Issue [#6478](https://github.com/realm/realm-java/issues/6478), since 5.8.0).
 
 ### Compatibility
-* Realm Object Server: 3.21.0-rc1 or later.
+* Realm Object Server: 3.21.1 or later.
```
bmunkholm:
```suggestion
* Realm Object Server: 3.21.0 or later.
```

PR #6516, opened 2019-05-12
Use AndroidX Benchmark library

cmelchior:
Replaces our homebrewn benchmark solution with the new official one from Google.
It is still in alpha, but seems to work for our use case: https://developer.android.com/jetpack/androidx/releases/benchmark

It also converts the project to Kotlin as the benchmark API is nicer there.


library-benchmarks/src/androidTest/java/io/realm/benchmarks/RealmAllocBenchmarks.kt
```diff
@@ -0,0 +1,87 @@
+/*
+ * Copyright 2016 Realm Inc.
```
nhachicha: 2019 same for the other new files

Issue #2530, opened 2016-03-31
Android Migration Like ios


bsmn99:
Hi,

I just have a request, 

why migration code is different from ios to android? in ios they just increasing the schema version and the altered table structure is updated automatically, but in android we have to write the migrate code for altered table as well as the schema version to be increased. Can u make in android too just increasing the schema number is able to alter the table? now its horrible to write migrate code for changes, and that too we have to make a track for what columns added, deleted and updated. Pls consider this..

cmelchior:
Hi @bsmn99 

It is also an discussion we had internally and there are good arguments for both types of migrations. 

Be aware that the iOS migrations are not automatic in all cases and those corner cases are non-trivial to solve and will have to be maintained even for future migrations.

The Android migrations require more boilerplate code for sure, but they are write-once. After that you can forget about them and they will continue to work. Also we have plans for creating an IntelliJ plugin that can help with creating the migration code, making it less painful. 

Note that if you do not want to keep your data across schema changes, you can always use `RealmConfiguration.deleteRealmIfMigrationRequired()`. For some use cases like offline caches, that is a lot easier as you don't have to write any migration code.

For the current future we plan to let iOS and Android behave differently in this area. That also enable us to gather more realworld input about both approaches.

bsmn99:
Hi @cmelchior 

Thanks for the clarification, expecting that IntelliJ plugin!..

bsmn99:
> _Be aware that the iOS migrations are not automatic in all cases and those corner cases are non-trivial to solve and will have to be maintained even for future migrations._

Can u explain what kind of corner cases for iOS?

cmelchior:
A couple of cases:

a) 
v1: Add new field `name`
v2: Remove field `name`.
v3: Add field `name` again. `name` should be empty here, not contain all the v1 data.

b) 
v1: Add field `email`
v2: Rename `email` to `contact`
v3: Add field `email` again. This should now be empty while contact contains all the data

A good way to think about is that iOS always go from `currentVersion -> latestVersion` skipping any intermediate calculations while Android always go `v1 -> v2 -> latestVersion`, which means that all steps are applied. So if you are changing the same field across multiple migrations on iOS you usually have to be a bit careful. This is also described in their docs btw: https://realm.io/docs/objc/latest/#linear-migrations

vonox7:
> For the current future we plan to let iOS and Android behave differently in this area.
[...]

PR #6529, opened 2019-05-29
Add an abstract RealmMigration class

wlxyhy:
User can extend this class to simplify realm migration

### DIFF
--- a/realm/realm-library/src/main/java/io/realm/AbsMigration.java
+++ b/realm/realm-library/src/main/java/io/realm/AbsMigration.java
@@ -1,0 +1,119 @@
+package io.realm
+
+import androidx.annotation.NonNull;
+
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Field;
+import java.lang.reflect.Modifier;
+import java.lang.reflect.ParameterizedType;
+import java.lang.reflect.Type;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Set;
+/**
+ * Created by wlxyhy on 2019/05/29.
+ * Extends this to simplify RealmMigration.
+ * Add "-keep public class * extends io.realm.RealmObject {*;}" in proguard-rule.pro if minifyEnabled.
+ * Do not extends this if there's name-difference between schema fields and your RealmObject fields in your project(eg:field with'@RealmField(name = "person_name")')! 
+ */
+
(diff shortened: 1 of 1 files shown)


cmelchior:
Can you give an example of how such a class could look like because I don't see what an abstract class could do better than an interface with a single method?

Zhuinden:
It seems to be similar to what I tried to accomplish with [AutoMigration](https://github.com/Zhuinden/realm-auto-migration/blob/cbf0e73c2b1921ca59d179d8b05f1ea91218b3f3/auto-migration/src/main/java/com/zhuinden/realmautomigration/AutoMigration.java#L76).

In fact, I'm quite surprised that https://github.com/realm/realm-java/pull/6529/files#diff-7db08e5a4da97d5e99415e36d8b08743R96 works, that's pretty neat.

wlxyhy:
> It seems to be similar to what I tried to accomplish with [AutoMigration](https://github.com/Zhuinden/realm-auto-migration/blob/cbf0e73c2b1921ca59d179d8b05f1ea91218b3f3/auto-migration/src/main/java/com/zhuinden/realmautomigration/AutoMigration.java#L76).
> 
> In fact, I'm quite surprised that https://github.com/realm/realm-java/pull/6529/files#diff-7db08e5a4da97d5e99415e36d8b08743R96 works, that's pretty neat. That would mean I can remove the `MigratedList` annotation I was hacking around with... 🤔

Yes, it's similar, I use abstract class just because
1. in some cases there's field to be renamed, so I add method renameFieldIfNeeded();
2.If you have more than 1 databases in your project, you can choose needed classes to add supported by getRealmClasses(). Can work with @RealmModule(classes = {x.class, xx.class})

wlxyhy:
> Can you give an example of how such a class could look like because I don't see what an abstract class could do better than an interface with a single method?

In most cases, you don't have to add or remove fields when your realm needs migration, just update version of your realm is enough.

cmelchior:
Hi @wlxyhy Thank you for this PR. 

While it probably does make sense to support some kind of auto-migration, I don't think adding this class is the right solution. The reason is that we actually already have support for this in [Object Store](https://github.com/realm/realm-object-store), it just hasn't been exposed in Java yet. And we wouldn't want to have this logic in multiple places.

You can see https://github.com/realm/realm-java/issues/2530 for more info.

PR #6540, opened 2019-06-19
exposing to_json from Core

nhachicha:
This is used mainly for debugging https://github.com/realm/realm-java-private/issues/262

- [x] waiting for Core to release `fsa/enhance-tableview-tojson`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,10 @@
 * APIs are backwards compatible with all previous release of realm-java in the 5.x.y series.
 
 ### Internal
-* None.
+* Updated to Realm Core 5.22.0.
+* Updated to Relm Sync 4.6.1.
+* Updated to Object Store commit 7c3ff8235579550a3e3c6060c47140b2005174f5
+
 
 
 ## 5.11.0(2019-05-01)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ### Enhancements
 * [ObjectServer] Added `SyncManager.refreshConnections()` that can be used to manually trigger a reconnect for all sessions. This is useful if the device has been offline for a long time or fail to detect that it regained connectivity. (Issue [#259](https://github.com/realm/realm-java-private/issues/259))
+* Added `RealmResults.asJson()` in `@Beta` that returns the result of the query as a JSON payload (#6540).
+
 
 ### Fixed
 * [ObjectServer] `PermissionManager` stopped working if an intermittent network error was reported. (Issue [#6492](https://github.com/realm/realm-java/issues/6492), since 3.7.0)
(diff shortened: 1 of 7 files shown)


realm/realm-library/src/androidTest/java/io/realm/RealmResultsTests.java
```diff
@@ -1685,4 +1691,240 @@ public void setValue_specificType_internalNameOnDynamicRealms() {
...
+                "                        \"name\": \"Dog owner 1\",\n" +
+                "                        \"dogs\": [],\n" +
+                "                        \"cat\": []\n" +
+                "                    }\n" +
+                "                ]\n" +
+                "            }\n" +
+                "        ],\n" +
+                "        \"columnStringList\": [\n" +
+                "            {\n" +
+                "                \"!ARRAY_VALUE\": \"Foo\"\n" +
```
cmelchior: It seems strange that primitive lists are modeled like `[{ "!ARRAY_VALUE": "Foo"}]` instead of `["Foo"]`. It would be much more natural to do the later?

I do realize there is a time constraint here, but I would much prefer to have the best possible output if we are adding this as a public API. 

Alternatively, we could expose it as a package protected method until we get a chance to fix this?

realm/realm-library/src/androidTest/java/io/realm/RealmResultsTests.java
```diff
@@ -1685,4 +1691,240 @@ public void setValue_specificType_internalNameOnDynamicRealms() {
...
+                "    {\n" +
+                "        \"id\": 0,\n" +
+                "        \"name\": \"One\",\n" +
+                "        \"date\": \"" + now + "\",\n" +
+                "        \"object\": [\n" +
+                "            {\n" +
+                "                \"id\": 0,\n" +
+                "                \"name\": \"Two\",\n" +
+                "                \"date\": \"" + now + "\",\n" +
+                "                \"object\": \"0\",\n" +
```
cmelchior: Why is the row index returned as a `string` instead of `number`?

nhachicha:
@cmelchior I created https://github.com/realm/realm-core/issues/3305 to list the enhancements Core needs to implement before we release the method as public, currently, it will have a `@Beta` annotation (subject to change)

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ### Enhancements
 * [ObjectServer] Added `SyncManager.refreshConnections()` that can be used to manually trigger a reconnect for all sessions. This is useful if the device has been offline for a long time or fail to detect that it regained connectivity. (Issue [#259](https://github.com/realm/realm-java-private/issues/259))
+* Added `RealmResults.asJson()` in `@Beta` that returns the result of the query as a JSON payload (#6540).
```
cmelchior: Also add the new Sync/Core version to the internal section

PR #6547, opened 2019-06-25
Fix queries on indexed linked properties

### LINKED ISSUE #6546: Realm using TimeZone of Date object in greaterThanOrEqualsTo(Date)
#### Goal
I want to query using date value independently from the timezone of Date object has.

#### Actual Results
When using greaterThanOrEqualTo(String fieldName, Date value) the method recognizes the timezone date object has, and it is not making the query properly.

I want to receive inputs within the beginning and the end of the day. I get the today's date, and set time to 00:00:00.000 when getting de date from the calendar instance, it uses the device timezone, and when the method is executed is not using UTC.

Calendar calendar = new GregorianCalendar();
        calendar.set(Calendar.HOUR_OF_DAY, 0);
        calendar.set(Calendar.MINUTE, 0);
        calendar.set(Calendar.SECOND, 0);
        calendar.set(Calendar.MILLISECOND, 0);
        Date today = calendar.getTime();
        calendar.add(Calendar.DAY_OF_MONTH, 1);
        Date tomorrow = calendar.getTime();

        result= realm.where(OrderRealm.class)
                .beginGroup()
                .greaterThanOrEqualTo("time", today)
                .lessThan("time", tomorrow)
                .endGroup()
                .findAll();

_Alternatively I set UTC to calendar instance_
_calendar.setTimeZone(TimeZone.getTimeZone("UTC"));_

Realm version(s): 361

Realm Sync feature enabled: Yes

Android Studio version: 3.4

Android Build Tools version: 28

Which Android version and device(s): 6

cmelchior:
Closes #6546 

Looks like a bug in Core, which needs to be fixed there. Currently, this PR just exposes the bug from the Java side.

I have not been able to create a repo case directly in Core yet, and playing around with the unit test also seems to indicate that the order of operations has to be really specific to trigger it.

What is clear is that removing the `@Index` on the object in the list being queried fixes it. So most likely this is related to the newly added optimizations for indexes added in Core. This is further supported by the fact that downgrading Realm fixes it.

### DIFF
--- a/realm/realm-library/src/androidTest/java/io/realm/RealmQueryTests.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/RealmQueryTests.java
@@ -22,7 +22,9 @@ import org.junit.Test;
 import org.junit.runner.RunWith;
 
 import java.lang.reflect.Field;
+import java.util.ArrayList;
 import java.util.Date;
+import java.util.List;
 import java.util.Locale;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.atomic.AtomicInteger;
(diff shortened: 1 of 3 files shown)


cmelchior:
Replaced by #6571

Issue #6522, opened 2019-05-20
Query with in() returns an empty RealmResults


mick1418:
#### Goal
I want to query a table and retrieve objects by their IDS

#### Actual Results
The RealmResults returns an empty list, even if the objects are in the table (checked with Realm Studio).

#### Steps & Code to Reproduce
Requesting the table with multiples IDs always return an empty list, but querying the IDs separately returns a list with the object.

`realm.where("mytable").equalTo("uid", "firstUID").findAll()`-> returns a list with the object
`realm.where("mytable").equalTo("uid", "secondUID").findAll()`-> returns a list with the object

`realm.where("mytable").beginGroup().equalTo("uid", "firstUID").or().equalTo("uid", "secondUID").endGroup().findAll()`-> returns an empty list

`realm.where("mytable").in("uid", new String[]{"firstUID", "secondUID"}).findAll()`-> returns an empty list

Maybe it has something to do with the following PR: (Realm Core PR #3250).
Reverting to 5.9.1 solved the issue.

#### Version of Realm and tooling

Realm version(s): ? 5.11.0

Realm Sync feature enabled: No

Android Studio version: ? 3.4 stable

Android Build Tools version: ? 28.0.3

Gradle version: ? 5.1.1

Which Android version and device(s): ? Android API 28, Emulator

Zhuinden:
Are we talking about https://github.com/realm/realm-core/pull/3250?

> `realm.where("mytable").beginGroup().equalTo("uid", "firstUID").or().equalTo("uid", "secondUID").endGroup().findAll()`-> returns an empty list

This is supposed to work. 😞

Zhuinden:
Although it says:

> This change does not try to optimise indexed columns which should be running O(log(N)*C). The benchmark with indexes turned on runs in 3.5 seconds.

If the `uid` column is Primary Key, then it is Indexed, and therefore should have been unaffected by this change. 

------

I was wondering if https://github.com/realm/realm-core/pull/3271 is related, but it is for `int` columns...

mick1418:
I should have mentioned my "uid" field is of type String, and not the Primary key in this given table

Zhuinden:
It is also not marked with `@Index`, correct?

mick1418:
Correct

Zhuinden:
@cmelchior

bmunkholm:
@mick1418 Could you please make a minimal self-contained repro case? Or a PR with a test-case that shows this?
[...]

PR #6571, opened 2019-07-29
Updates Sync/Core to fix query bug on indexed integer properties

cmelchior:
Adds unit test capturing https://github.com/realm/realm-java/issues/6522


cmelchior:
The release was made without the unit test in this PR. Only the changelog entry was added. This PR should be modified accordingly.

cmelchior:
This should be tested in Core. So closing this PR.

Issue #6531, opened 2019-05-31
Realm Gradle plugin should process R classes from jar


gavra0:
Starting with Android Gradle plugin 3.6.0-alpha01, R.java is not generated any more, and R clasess are compiled separately from the rest of the sources to R.jar e.g. `./app/build/intermediates/compile_and_runtime_not_namespaced_r_class_jar/debug/R.jar`.

This means that the `RealmTransformer` should be able to process project classes coming from both directories and jars. Looking at implementation of `BuildTemplate.categorizeClassNames` it seems that classes coming from jars are in not processed.

Related issue: https://issuetracker.google.com/130965921#comment14

cmelchior:
Thanks. Yes, we should fix this.

cmelchior:
I can reproduce this with `3.6.0-alpha05` and our `/examples/moduleExample`. It crashes with:

```
2019-08-02 14:49:43.548 11045-11045/? E/AndroidRuntime: FATAL EXCEPTION: main
    Process: io.realm.examples.realmmultiprocessexample, PID: 11045
    java.lang.NoClassDefFoundError: Failed resolution of: Landroid/support/v7/appcompat/R$drawable;
        at android.support.v7.widget.k.<clinit>(:107)
        at android.support.v7.widget.k.a(:84)
        at android.support.v7.widget.v0.c(:85)
        at android.support.v7.app.f.<init>(:128)
        at android.support.v7.app.j.<init>(:149)
        at android.support.v7.app.h.<init>(:56)
        at android.support.v7.app.i.<init>(:31)
        at android.support.v7.app.g.<init>(:31)
        at android.support.v7.app.e.a(:198)
        at android.support.v7.app.e.a(:183)
        at android.support.v7.app.c.h(:519)
        at android.support.v7.app.c.onCreate(:70)
        at io.realm.examples.realmmultiprocessexample.MainActivity.onCreate(:60)
        at android.app.Activity.performCreate(Activity.java:7136)
        at android.app.Activity.performCreate(Activity.java:7127)
        at android.app.Instrumentation.callActivityOnCreate(Instrumentation.java:1271)
        at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2893)
        at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:3048)
        at android.app.servertransaction.LaunchActivityItem.execute(LaunchActivityItem.java:78)
[...]

PR #6575, opened 2019-08-05
Correctly copy JAR files in the Realm transformer

cmelchior:
Closes https://github.com/realm/realm-java/issues/6531

Test by compilling and running all all examples in the repo with `3.6.0-alpha05` (latest release). Before this PR some of the examples would crash at runtime, now they all run.

Reason being that in 3.6.0 the R file is now generated inside a seperate JAR file that needs to be copied as well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,20 @@
+## 5.13.1(YYYY-MM-DD)
+
+### Enhancements
+* None.
+
+### Fixed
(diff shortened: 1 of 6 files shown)


cmelchior:
Fast-merging due to an imminent release, but since this change should only affect people using Android Gradle plugin 3.6.0-alpha01 and above it should be relatively safe.

PR #6590, opened 2019-08-23
Add support for frozen objects

### LINKED ISSUE #1208: Thread handoff for RealmObject and RealmResults / Frozen state

### LINKED ISSUE #4291: Better RxJava support

cmelchior:
Closes #4291
Closes #1208 ( most of it. There is still an open question about schedulers)

Blocked by: 
* https://github.com/realm/realm-java/pull/6590/files#diff-4e0e1e80f7c3471fb1ff091a810a021dR140

This PR adds support for frozen Realm objects. The major benefit of frozen objects is that they are not thread confined and can thus be accessed on any thread. This is especially important for most kinds of stream-based architectures (which is most of them at this point).

The trade-off is that the objects are then immutable (which is actually considered a positive for most reactive architectures on Android)

### High level API
```
// Adding two new methods to all top level Realm objects: .freeze() and .isFrozen()
// Anything can be frozen: Realm, Results, Lists, Objects
Realm frozenRealm = realm.freeze();
RealmResults<Person> persons = realm.where(Person.class).findAll().freeze();
Person p = realm.where(Person.class).findFirst().freeze();
RealmList<Child> children = realm.where(Person.class).findFirst().getChildren().freeze();

// The RxJava factory will be changed so observables now emit frozen objects by default
realm.where(Person.class).findAllAsync().asObservable()
  .map((result, changeSet) -> {
    result.isFrozen() == true;    
  })

// It is still possible to navigate and query a frozen Realm just like a live one, 
// but all objects will be frozen
Realm frozenRealm = realm.freeze();
frozenRealm.where(Person.class).findAll().isFrozen() == true;

// Because freezing objects increase the chance of leaking versions and thus 
// disk space we should expose some logic around Cores tracking of versions. 
// I suggest throwing an exception if the number of active versions goes over 
// a certain number.
  .maxNumberOfActiveVersions(20);
  .build();
```
### Semantics

Freezing an object will freeze the entire object graph at that version, effectively turning it read-only. Doing this will also lift the thread-confined restriction on it.

These are still to be decided, but so far I'm envisioning the following

* [x] Starting a write transaction on a frozen Realm will throw an exception
* [x] Trying to add a change listener will throw an exception.
* [x] Trying to refresh a Realm in any way will throw an exception
* [x] Trying to write to a frozen object inside the live Realms write transaction will throw. 
* [x] Doing `liveRealm.copyToRealmOrUpdate(frozenObject)` should work the same way as using unmanaged objects.

### Open design questions

**`Realm.getLocalInstanceCount(config)`:**
What are the semantics of `Realm.getLocalInstanceCount(config)` for frozen Realms? 

For now, globalCount == LocalThreadCount for frozen Realms, since these Realms are not thread constrained.

**What are the semantics if the live Realm spawning frozen Realms is closed?**
There are conflicting interests in terms of making it easy to reason about a Realm files life cycle vs. the Java objects. I.e. it becomes difficult to close a Realm file if frozen objects cannot easily close their Realm. 
]
I can think of two solutions: 1) When all live Realms across all thread are closed, this will also automatically close all frozen Realms or 2) Introduce something like `Realm.closeFrozenRealms(config)` which will do it manually. It is unclear what is better.

For now, frozen Realms will all be closed when the live Realm spawning them are. This is primarely driven by making RxJava support easier.

**Freeze existing object or return frozen copy?**
Should freezing objects return a frozen copy or freeze the object in place? 

There are performance benefits to freezing current object, but it seems to conflict with e.g. RxJava support and people need access to the live object in order to get notifications. Kotlin Multiplatform freeze in place. 

For now we return frozen copies. I suggest we leave it up to user feedback to see if we should change it. We can consider having an internal method that freezes in-place if it will improve performance a lot. 

**Equality and hashcode**: 
How should object equality work for frozen objects? Current RealmLists and RealmLists compare objects on pr. object basis. How should frozen Realms and objects be compared? 

For now Frozen objects with the same values are not considered equal to a live object. Same for Realms. Frozen objects at different versions (but equal values) are also not considered equal.

Use cases are unclear. The suggestion is to leave it up to user feedback to see if we should change it.

**Generic parameter on `RealmModel`?**
So it becomes: `RealmModel<MyImplementation>`, e.g `public class Dog extends RealmObject<Dog>`. 

This enables better return types on `freeze()` as well as the current `asFlowable()` (which is kinda annoying to use). Setting the parameter is optional in Java (non-breaking), but seems to be required in Kotlin if present.

Unclear if we want to break everyone using Kotlin for this, but it will remove a lot of annoying casts. I will defer changing this to a later PR so we can discuss further (and to prevent cluttering this PR too much).

### Benchmark results (on Emulator)
Results where frozen with 10.000 objects. It looks like freezing TableViews are fast enough that we don't need to be too concerned about performance. For a more scalable solution we should look into freezing TableViews off the UI thread (but doing it on the notifier thread also seems like a bad idea).

```
benchmark:       149,500 ns EMULATOR_UNLOCKED_FrozenObjectsBenchmarks.freezeResults
benchmark:        27,400 ns EMULATOR_UNLOCKED_FrozenObjectsBenchmarks.freezeList
benchmark:        24,454 ns EMULATOR_UNLOCKED_FrozenObjectsBenchmarks.freezeObject
benchmark:        14,375 ns EMULATOR_UNLOCKED_FrozenObjectsBenchmarks.freezeRealm
```

### TODO:
* [x] Changelog
* [x] Cleanup
* [x] Finalize https://github.com/realm/realm-object-store/pull/856 and merge it to master
* [x] Implement `RealmObject.freeze()` and `RealmObject.isFrozen()`
* [x] Implement `RealmModel.freeze()` and `RealmModel.isFrozen()`
* [x] Implement `RealmList.freeze()` and `RealmList.isFrozen()`
Implement `RealmResults.freeze()` and `RealmResults.isFrozen()`
* [x] Implement `Realm.freeze()` and `Realm.isFrozen()`
* [x] Implement `DynamicRealm.freeze()` and `DynamicRealm.isFrozen()`
* [x] It should be possible to freeze deleted objects (which might seem a bit weird, but is needed for us to support event streams that want to notify about deleted objects).
* [x] Fully closing the live Realm should also close all frozen instances
* [x] Add benchmarks for freezing Realm
* [x] Add benchmarks for freezing large RealmResults
* [x] Add benchmark for freezing large RealmList
* [x] Add benchmark for freezing RealmObject
* [x] Javadoc
* [x] Change `RxObservableFactory` so it returns frozen objects.
* [x] Add option to `RealmConfiguration` to throw at certain number of active versions. Should be checked when we start a write transaction.
* [x] All mutator methods should throw on frozen Realms
* [x] Tests: Starting write transactions fail
* [x] Tests: Adding Realm listeners fail
* [x] Tests: Adding Results listeners fail
* [x] Tests: Adding List listeners fail
* [x] Tests: Adding Object listeners fail
* [x] Tests: refreshing fail
* [x] Tests: Equality of frozen Realms
* [x] Tests: Equality of frozen RealmObject
* [x] Tests: `copyOrUpdate` frozen Realm data
* [x] Tests: `insertOrUpdate` frozen Realm data
* [x] Other tests


cmelchior:
The PR is ready for review even though there are two outstanding issues in Core that need to be resolved and the ObjectStore PR is still work-in-progress.

The most important bits in this PR are:

* RealmCache.java
* RealmObservableFactory.java
* FrozenObjectsTests.java

Also, please read and verify that you agree with the design as described in the initial post. Especially the section about open design questions.

realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -433,6 +438,36 @@ public void cancelTransaction() {
...
+     * @return a frozen copy of this Realm.
+     * @throws IllegalStateException if this method is called from inside a write transaction.
+     */
+    public abstract BaseRealm freeze();
+
+    /**
+     * Returns whether or not this Realm is frozen.
+     *
+     * @return {@code true} if the Realm is frozen, {@code false} if it is not.
+     * @see #freeze()
```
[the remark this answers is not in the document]
cmelchior: `isFrozen` can only throw if the Realm if the Realm is closed. That case isn't documented anywhere else either.

realm/realm-library/src/androidTest/java/io/realm/RealmAsyncQueryTests.java
```diff
@@ -1377,6 +1377,38 @@ public void execute(Realm realm) {
...
+        looperThread.keepStrongReference(results);
+        assertFalse(results.isLoaded());
+        assertTrue(results.isValid());
+        assertEquals(0, results.size());
+
+        RealmResults<AllTypes> frozenResults = results.freeze();
+        assertFalse(frozenResults.isLoaded());
+        assertTrue(frozenResults.isValid());
+        assertEquals(0, frozenResults.size());
+
```
kneth: `assertTrue(frozenResults.isFrozen())` maybe

realm/realm-library/src/main/cpp/io_realm_internal_OsList.cpp
```diff
@@ -489,3 +489,16 @@ JNIEXPORT jobject JNICALL Java_io_realm_internal_OsList_nativeGetValue(JNIEnv* e
...
+JNIEXPORT jlong JNICALL Java_io_realm_internal_OsList_nativeFreeze(JNIEnv* env, jclass, jlong native_list_ptr, jlong frozen_realm_native_ptr)
+{
+    try {
+        auto& wrapper = *reinterpret_cast<ListWrapper*>(native_list_ptr);
+        auto frozen_realm = *(reinterpret_cast<SharedRealm*>(frozen_realm_native_ptr));
+        List list = wrapper.collection().freeze(frozen_realm);
+        return reinterpret_cast<jlong>(new ListWrapper(list));
+    }
+    CATCH_STD()
+    return -1;
```
kneth: Why not 0 or `reinterpret_cast<jlong>(nullptr)`?

realm/realm-library/src/main/cpp/io_realm_internal_OsResults.cpp
```diff
@@ -474,3 +476,16 @@ JNIEXPORT void JNICALL Java_io_realm_internal_OsResults_nativeEvaluateQueryIfNee
...
+JNIEXPORT jlong JNICALL Java_io_realm_internal_OsResults_nativeFreeze(JNIEnv* env, jclass, jlong native_ptr, jlong frozen_realm_native_ptr)
+{
+    try {
+        auto wrapper = reinterpret_cast<ResultsWrapper*>(native_ptr);
+        auto frozen_realm = *(reinterpret_cast<SharedRealm*>(frozen_realm_native_ptr));
+        Results results = wrapper->collection().freeze(frozen_realm);
+        return reinterpret_cast<jlong>(new ResultsWrapper(results));
+    }
+    CATCH_STD()
+    return -1;
```
kneth: why not `reinterpret_cast<jlong>(nullptr)`?

realm/realm-library/src/main/java/io/realm/RealmConfiguration.java
```diff
@@ -765,6 +783,28 @@ public Builder compactOnLaunch(CompactOnLaunchCallback compactOnLaunch) {
...
         }
 
+        /**
+         * Sets the maximum number of live versions in the Realm file before an {@link IllegalStateException} is thrown when
+         * attempting to write more data.
+         * <p>
+         * Realm is capable of concurrently handling many different versions of Realm objects. This can e.g. happen if you
+         * have a Realm open on many different threads or are freezing objects while data is being written to the file.
+         * <p>
+         * Under normal circumstances this is not a problem, but if the number of active versions grow to large, it will
```
kneth:
```suggestion
         * Under normal circumstances this is not a problem, but if the number of active versions grow too large, it will
```

realm/realm-library/src/objectServer/java/io/realm/SyncConfiguration.java
```diff
@@ -1158,6 +1162,25 @@ public Builder clientResyncMode(ClientResyncMode mode) {
             return this;
         }
 
+        /**
+         * Sets the maximum number of live versions in the Realm file before an {@link IllegalStateException} is thrown when
+         * attempting to write more data.
+         * <p>
+         * Realm is capable of concurrently handling many different versions of Realm objects. This can e.g. happen if you
```
kneth:
```suggestion
         * Realm is capable of concurrently handling many different versions of Realm objects. This can happen if you
```

realm/realm-library/src/objectServer/java/io/realm/SyncConfiguration.java
```diff
@@ -1158,6 +1162,25 @@ public Builder clientResyncMode(ClientResyncMode mode) {
...
+         * have a Realm open on many different threads or are freezing objects while data is being written to the file.
+         * <p>
+         * Under normal circumstances this is not a problem, but if the number of active versions grow to large, it will
```
kneth:
```suggestion
         * Under normal circumstances this is not a problem, but if the number of active versions grow too large, it will
```

CHANGELOG.md
```diff
@@ -1,3 +1,24 @@
...
+* All RxJava Flowables and Observables now return frozen objects instead of live objects. This can be configured using `RealmConfiguration.Builder.rxFactory(new RealmObservableFactory(true|false))`. By using frozen objects, it is possible to send RealmObjects across threads, which means that all RxJava operators should now be supported without the need to copy Realm data into unmanaged objects. 
+
+### Enhancements
+* Added `Realm.freeze()`, `RealmObject.freeze()`, `RealmResults.freeze()` and `RealmList.freeze()`. These methods will return a frozen snapshot of the current Realm data. This data can be read from any thread without throwing an `IllegalStateException`. All frozen Realms and data can be closed by calling `Realm.close()` on the frozen Realm, but fully closing all live Realms will also close the frozen ones. Frozen data can be queried as normal, but trying to mutate it in any way will throw an `IllegalStateException`. This includes all methods that attempt to refresh or add changelisteners. (Issue [#6590](https://github.com/realm/realm-java/pull/6590))
```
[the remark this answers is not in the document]
kneth: "changelisteners" -> "change listeners"

realm/realm-library/src/main/cpp/io_realm_internal_OsSharedRealm.cpp
```diff
@@ -516,3 +526,23 @@ JNIEXPORT jboolean JNICALL Java_io_realm_internal_OsSharedRealm_nativeIsPartial(
...
+}
+
+JNIEXPORT jlong JNICALL Java_io_realm_internal_OsSharedRealm_nativeFreeze(JNIEnv* env, jclass, jlong shared_realm_ptr)
+{
+    try {
+        auto& shared_realm = *(reinterpret_cast<SharedRealm*>(shared_realm_ptr));
+        return reinterpret_cast<jlong>(new SharedRealm(std::move(shared_realm->freeze())));
+    }
+    CATCH_STD()
+    return -1;
```
kneth: Shouldn't it be `reinterpret_cast<jlong>(nullptr)`?

realm/realm-library/src/main/cpp/io_realm_internal_UncheckedRow.cpp
```diff
@@ -389,6 +389,20 @@ JNIEXPORT void JNICALL Java_io_realm_internal_UncheckedRow_nativeSetNull(JNIEnv*
...
+                                                                         jlong j_frozen_realm_native_ptr)
+{
+    try {
+        Obj* obj = reinterpret_cast<Obj*>(j_native_row_ptr);
+        auto frozen_realm = *(reinterpret_cast<SharedRealm*>(j_frozen_realm_native_ptr));
+        auto frozen_obj = new Obj(frozen_realm->transaction().import_copy_of(*obj));
+        return reinterpret_cast<jlong>(frozen_obj);
+    }
+    CATCH_STD()
+    return -1;
```
kneth: Shouldn't it be `reinterpret_cast<jlong>(nullptr)`?

realm/realm-library/src/main/java/io/realm/DynamicRealm.java
```diff
@@ -394,6 +394,24 @@ public RealmSchema getSchema() {
         return schema;
     }
 
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public DynamicRealm freeze() {
+        // In some cases a Read transaction has not been begun for the the Realm, which means
```
kneth:
```suggestion
        // In some cases a Read transaction has not been begun for the Realm, which means
```

library-benchmarks/src/androidTest/java/io/realm/benchmarks/FrozenObjectsBenchmarks.kt
```diff
@@ -0,0 +1,119 @@
...
+    val benchmarkRule = BenchmarkRule()
+
+    private lateinit var realm: Realm
+    private lateinit var readObject: AllTypes
+    private lateinit var realmConfig: RealmConfiguration
+
+    @Before
+    fun setUp() {
+        Realm.init(InstrumentationRegistry.getInstrumentation().context)
+        RealmLog.error("SETUP");
```
bmunkholm:
```suggestion
        RealmLog.error("SETUP")
```

realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -433,6 +438,36 @@ public void cancelTransaction() {
...
+    /**
+     * Returns a frozen snapshot of the current Realm. This Realm can be read and queried from any thread without throwing
+     * an {@link IllegalStateException}. A frozen Realm has its own lifecycle and can be closed by calling {@link #close()},
+     * but fully closing the Realm that spawned the frozen copy will also close the frozen Realm.
+     * <p>
+     * Frozen data can be queried as normal, but trying to mutate it in any way or attempting to register any listener will
+     * throw an {@link IllegalStateException}.
+     * <p>
+     * Note: Keeping a large number of Realms with different versions alive can have a negative impact on the filesize
+     * of the Realm. In order to catch such use patterns it is possible to set {@link RealmConfiguration.Builder#maxNumberOfActiveVersions(long)}.
```
bmunkholm: Shouldn't we make a default for maxNumberOfActiveVersions? This seems like something people are bound to accidentally do.

bmunkholm:
Awesome work @cmelchior !

realm/realm-library/src/androidTest/java/io/realm/RxJavaTests.java
```diff
@@ -1058,29 +844,264 @@ public void dynamicRealmObject_gcStressTest() {
...
+            realm.createObject(AllTypes.class).setColumnLong(1);
+        }
+        realm.commitTransaction();
+
+        subscription = realm.where(AllTypes.class).sort(AllTypes.FIELD_LONG).findAllAsync().asFlowable()
+                .subscribeOn(Schedulers.io())
+                .filter(RealmResults::isLoaded)
+                .map(results -> {
+                    RealmLog.error(Thread.currentThread().getName());
+                    return new Pair<>(results.size(), results);
```
nhachicha: can you pass also one or multiple frozen objects so we can assert that observing them from another thread is ok

CHANGELOG.md
```diff
@@ -1,3 +1,24 @@
+## 7.0.0(YYYY-MM-DD)
+
+### Breaking Changes
+* RxJava Flowables and Observables are now subscribed to and unsubscribed to asynchronously on the thread holding the live Realm, instead of previously where this was done synchronously.
+* All RxJava Flowables and Observables now return frozen objects instead of live objects. This can be configured using `RealmConfiguration.Builder.rxFactory(new RealmObservableFactory(true|false))`. By using frozen objects, it is possible to send RealmObjects across threads, which means that all RxJava operators should now be supported without the need to copy Realm data into unmanaged objects.
```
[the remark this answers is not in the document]
cmelchior: I expand on the definition in the next sentence, so think it is acceptable to use it here.

realm/realm-library/src/main/java/io/realm/BaseRealm.java
cmelchior: The default is `Long.MAX_VALUE`, which is essentially the same as disabling. Finding a good default might be difficult as it very much depends on what else the app is doing, but I guess we cant set it to something relatively high and the make the exception message really clear about how to change the value if people hit it and think it is a mistake.

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -446,34 +590,37 @@ synchronized void release(BaseRealm realm) {
...
                 // We keep the cache in the caches list even when its global counter reaches 0. It will be reused when
                 // next time a Realm instance with the same path is opened. By not removing it, the lock on
                 // cachesList is not needed here.
                 configuration = null;
+
+                // TODO: Close all frozen Realms. This can introduce race conditions on other
+                // threads if the lifecyle of using Realm data is not correctly controlled.
+                // We need to run experiments to verify weather we should use this approach
+                // or require that people manually close Frozen Realms.
+                for (ReferenceCounter counter : refAndCountMap.values()) {
```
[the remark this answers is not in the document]
Zhuinden: `whether`

realm/realm-library/src/main/java/io/realm/RealmCache.java
```diff
@@ -66,13 +67,140 @@
...
-        // The Realm instance in this thread.
-        private final ThreadLocal<BaseRealm> localRealm = new ThreadLocal<>();
+    private static abstract class ReferenceCounter {
+
         // How many references to this Realm instance in this thread.
-        private final ThreadLocal<Integer> localCount = new ThreadLocal<>();
+        protected final ThreadLocal<Integer> localCount = new ThreadLocal<>();
         // How many threads have instances refer to this configuration.
-        private int globalCount = 0;
+        protected int globalCount = 0;
```
nhachicha: This is accessed potentially from various Threads, make it `volatile` or use atomic (`AtomicInteger`).

realm/realm-library/src/androidTest/java/io/realm/FrozenObjectsTests.java
```diff
@@ -0,0 +1,753 @@
...
+        realm.beginTransaction();
+        try {
+            frozenRealm.copyToRealm(new AllTypes());
+            fail();
+        } catch (IllegalStateException ignore) {
+        } finally {
+            realm.cancelTransaction();
+        }
+
+        realm.executeTransaction(r -> {
```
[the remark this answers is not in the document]
cmelchior: Good catch. I actually forgot to guard async transactions

realm/realm-library/src/androidTest/java/io/realm/FrozenObjectsTests.java
```diff
@@ -0,0 +1,754 @@
...
+    }
+
+    @Test
+    public void insertOrUpdateList() throws InterruptedException {
+        realm.executeTransaction(r -> {
+            r.createObject(AllJavaTypes.class, 42);
+            r.createObject(AllJavaTypes.class, 43);
+        });
+        AllJavaTypes frozenObject1 = realm.where(AllJavaTypes.class).equalTo(AllJavaTypes.FIELD_ID, 42).findFirst().freeze();
+        AllJavaTypes frozenObject2 = realm.where(AllJavaTypes.class).equalTo(AllJavaTypes.FIELD_ID, 42).findFirst().freeze();
```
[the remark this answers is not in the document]
cmelchior: Actually no. The tests needs to have to Java objects pointing to the same underlying Realm object in order to hit some codepaths in the insert methods that check for equal objects, but I can see it being confusing. I'll add a comment.

realm/realm-library/build.gradle
```diff
@@ -208,14 +208,14 @@ dependencies {
     compileOnly 'com.google.code.findbugs:findbugs-annotations:3.0.1'
 
     api "io.realm:realm-annotations:${version}"
+    implementation 'io.reactivex.rxjava2:rxandroid:2.1.1' // Does this also pull in RxJava
```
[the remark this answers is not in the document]
cmelchior: This actually pulls in RxJava, which we wanted to avoid by having it `compileOnly`. I made this `compileOnly` as well.

bmunkholm:
Have you measured test coverage?

realm/realm-library/src/main/java/io/realm/RealmCache.java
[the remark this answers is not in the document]
cmelchior: Yes, there is is a test for `getLocalInstanceCount()` and `getGlobalInstanceCount()` that verifies this.

realm/realm-annotations-processor/src/test/resources/io/realm/some_test_BooleansRealmProxy.java
```diff
@@ -502,9 +502,15 @@ public boolean equals(Object o) {
...
 
-        String path = proxyState.getRealm$realm().getPath();
-        String otherPath = aBooleans.proxyState.getRealm$realm().getPath();
+        BaseRealm realm = proxyState.getRealm$realm();
+        BaseRealm otherRealm = aBooleans.proxyState.getRealm$realm();
+        String path = realm.getPath();
+        String otherPath = otherRealm.getPath();
         if (path != null ? !path.equals(otherPath) : otherPath != null) return false;
+        if (realm.isFrozen() != otherRealm.isFrozen()) return false;
+        if (!realm.sharedRealm.getVersionID().equals(otherRealm.sharedRealm.getVersionID())) {
```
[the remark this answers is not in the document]
cmelchior: IMO Yes, but it is still an open question exactly what equality means in the context of a Realm object.

I.e. Is `Object 1:V1` equal to `Object: V2`?

- if the version is different, but all properties are the same (note, checking this might be extremely expensive if there are links involved)?
- If the version is different and properties have changed?
- What if objects are frozen?

Since this is the auto-generated equals method I opted for the conservative approach, i.e. the objects are only equal if they have the same "state", which involves checking both version and frozen state, even if they have the same property values they are still considered "not-equal" if the version is different.

If people want to loosen this, they can implement equals themselves, which will override this implementation.

But this was also one of the open questions I raised in the original post about this PR, so any input to these semantics this would be great.

CHANGELOG.md
```diff
@@ -1,3 +1,24 @@
+## 7.0.0(YYYY-MM-DD)
+
+### Breaking Changes
+* RxJava Flowables and Observables are now subscribed to and unsubscribed to asynchronously on the thread holding the live Realm, instead of previously where this was done synchronously.
```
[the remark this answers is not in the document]
cmelchior: Ideally yes. Our current RxJava example goes to great length showing the pitfalls involving live objects. These are still relevant, but I think we should adjust the project so it more highlights how it works when using the new default frozen objects. I would rather do this in another PR though as this one is already pretty big

realm/realm-library/src/main/java/io/realm/RealmConfiguration.java
```diff
@@ -765,6 +783,28 @@ public Builder compactOnLaunch(CompactOnLaunchCallback compactOnLaunch) {
...
+         * have a Realm open on many different threads or are freezing objects while data is being written to the file.
+         * <p>
+         * Under normal circumstances this is not a problem, but if the number of active versions grow too large, it will
+         * have a negative effect on the filesize on disk. Setting this parameters can therefore be used to prevent uses of
+         * Realm that can result in very large Realms.
+         *
+         * @param number the maximum number of active versions before an exception is thrown.
+         * @see <a href="https://realm.io/docs/java/latest/#faq-large-realm-file-size">FAQ</a>
+         */
+        public Builder maxNumberOfActiveVersions(long number) {
```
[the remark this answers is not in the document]
cmelchior: Yes, the number includes versions produced by Sync. I agree it can get a bit misleading in that case. The problem is that the number of versions is really just a bit poor proxy for what we really care about. Namely that the file doesn't grow out of control.

IMO, this is still useful, but I will update the docs to be more clear about what it is.

realm/realm-library/src/main/java/io/realm/internal/OsSharedRealm.java
```diff
@@ -37,6 +37,11 @@
 public final class OsSharedRealm implements Closeable, NativeObject {
 
     public static class VersionID implements Comparable<VersionID> {
+        // Realm Core uses unsigned integers to represent versions. This means
+        // they could theoretically hit this value (maximum value of unsigned + overflow)
```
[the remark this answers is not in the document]
cmelchior: You are probably right 😄

realm/realm-annotations-processor/src/test/resources/io/realm/some_test_BooleansRealmProxy.java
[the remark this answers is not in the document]
nhachicha: I don't think there's a right decision here since it's subject to the user intention, we should stick with a coherent default implementation and let the user customize it as necessary.

From my perspective, two managed objects are equals if they reference the same `ObjKey` and have the same version ID, this is quick to check (even with links). Implementing this as the default behaviour will allow the user to override it with a version of `equal` based on the content if he wants.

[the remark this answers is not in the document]
cmelchior: So you would say that a frozen object is equal to the live version if they have the same version?

I guess I would just worry that one object accidentally being equal to another could cause problems, e.g. if used as a key in a HashMap or similar:

```
// Example
Person person = realm.where(Person.class).findFirst();
Person frozenPerson = person.freeze();
person.equals(frosenPerson); // true
realm.refresh();
person.equals(frosenPerson); // potentially false
```

realm/realm-library/src/androidTest/java/io/realm/RxJavaTests.java
```diff
@@ -68,381 +67,290 @@
...
-        final AtomicBoolean subscribedNotified = new AtomicBoolean(false);
-        subscription = obj.<AllTypes>asFlowable().subscribe(new Consumer <AllTypes>() {
-            @Override
-            public void accept(AllTypes rxObject) throws Exception {
-                assertTrue(rxObject == obj);
-                subscribedNotified.set(true);
-            }
+        subscription = obj.<AllJavaTypes>asFlowable().subscribe(rxObject -> {
+            assertTrue(rxObject.isFrozen());
+            assertNotEquals(rxObject, obj); // Frozen objects are not equal to their live counter parts.
```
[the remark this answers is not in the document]
cmelchior: Actually not, in the current semantics of `equals()`, frozen object are never equal to their live counterpart. But probably better to keep the discussion in: https://github.com/realm/realm-java/pull/6590#discussion_r346961188

realm/realm-library/src/androidTest/java/io/realm/RxJavaTests.java
```diff
@@ -454,91 +362,70 @@ public void accept(RealmResults<AllTypes> allTypes) throws Exception {
...
-                    assertEquals(1, change.getChangeset().getInsertions().length);
-                    looperThread.testComplete();
-                }
+        subscription = results.asChangesetObservable().subscribe(change -> {
+            RealmResults<AllTypes> rxResults = change.getCollection();
+            assertTrue(rxResults.isFrozen());
+            if (rxResults.isEmpty()) {
+                realm.executeTransaction(r -> r.createObject(AllTypes.class));
+            } else if (rxResults.size() == 1) {
+                assertEquals(1, change.getChangeset().getInsertions().length);
```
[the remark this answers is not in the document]
cmelchior: This is just a smoke test for the RxJava implementation. All of this should be thoroughly tested by our existing change listener tests.

cmelchior:
All comments addressed. Note that I now included a copy of `RxAndroid` in Realm to avoid a lot of hassle. 

There are still some outstanding questions around `equals` for RealmObjects.

All unit tests failing are due to other changes in Core6 that should be addressed in that PR.

realm/realm-annotations-processor/src/test/resources/io/realm/some_test_BooleansRealmProxy.java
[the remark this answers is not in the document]
nhachicha: but that's the usage/choice of a mutable key for a Map which is a bad practice (since the hash code changes ...)

realm/realm-library/src/main/java/io/realm/internal/rxjava/RealmAndroidSchedulers.java
```diff
@@ -0,0 +1,88 @@
...
+import android.os.Handler;
+import android.os.Looper;
+import android.os.Message;
+import io.reactivex.Scheduler;
+import java.util.concurrent.Callable;
+
+/**
+ * Original source: https://raw.githubusercontent.com/ReactiveX/RxAndroid/2.x/rxandroid/src/main/java/io/reactivex/android/schedulers/AndroidSchedulers.java
+ *
+ * Class is imported here in order to issues that will arise if we depend directly on RxAndroid since
```
nhachicha: to avoid issues

realm/realm-library/src/main/java/io/realm/internal/rxjava/RealmAndroidSchedulers.java
```diff
@@ -0,0 +1,88 @@
...
+ * limitations under the License.
+ */
+package io.realm.internal.rxjava;
+
+import android.annotation.SuppressLint;
+import android.os.Build;
+import android.os.Handler;
+import android.os.Looper;
+import android.os.Message;
+import io.reactivex.Scheduler;
```
nhachicha: you should import `RealmHandlerScheduler` instead

realm/realm-library/src/testUtils/java/io/realm/rule/RunInLooperThread.java
```diff
@@ -408,7 +408,7 @@ public boolean isRuleUsed() {
         private final RunTestInLooperThread annotation;
         private final Statement base;
 
-        RunInLooperThreadStatement(RunTestInLooperThread annotation, Statement base) {
+         RunInLooperThreadStatement(RunTestInLooperThread annotation, Statement base) {
```
nhachicha: formatting

realm/realm-library/src/testUtils/java/io/realm/rule/TestRealmConfigurationFactory.java
```diff
@@ -82,7 +82,8 @@ protected void before() throws Throwable {
     protected void after() {
         try {
             for (RealmConfiguration configuration : configurations) {
```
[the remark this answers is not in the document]
nhachicha: actually the issue still persists so re-enable the commented code once Core issue is fixed

realm/realm-library/src/main/java/io/realm/internal/rxjava/AndroidSchedulers.java
```diff
@@ -0,0 +1,86 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
```
kneth: Just to be on the safe side, I think you should mention who the copyright holder is.

cmelchior: These files are removed again. Nabil had a better idea for including that I completely forgot was possible.

realm/realm-annotations-processor/src/test/resources/io/realm/some_test_BooleansRealmProxy.java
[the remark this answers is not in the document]
cmelchior: With Core 6 I would think we would change `RealmObject.hashCode()` to use the `ObjKey` since hash codes are allowed to be the same even if `equals` returns false

realm/realm-library/src/androidTest/java/io/realm/RealmAnnotationTests.java
```diff
@@ -127,10 +127,10 @@ public void primaryKey_errorOnInsertingSameObject() {
     }
 
     @Test
-    public void primaryKey_isIndexed() {
+    public void string_primaryKey_isNotIndexed() {
         Table table = realm.getTable(PrimaryKeyAsString.class);
         assertNotNull(OsObjectStore.getPrimaryKeyForObject(realm.getSharedRealm(), PrimaryKeyAsString.CLASS_NAME));
-        assertTrue(table.hasSearchIndex(table.getColumnKey("name")));
+        assertFalse(table.hasSearchIndex(table.getColumnKey("name")));
```
Zhuinden: Primary keys are not indexed by default anymore? :thinking:

cmelchior: With Core 6 we are mapping user defined String primary keys to our internal ObjectId which is faster so the index isn't strictly needed for looking up single objects.

That said, it is possible that we are forgetting about some use cases or that we are missing some upgrade tests. Thanks for pointing it out 👏

cmelchior:
The branch should now be up to date with the Core-6 branch and all test parses. @nhachicha

realm/realm-library/src/androidTest/java/io/realm/RealmAnnotationTests.java
nhachicha: There is no migration needed, there's a test covering this use case in https://github.com/realm/realm-java/pull/6107/commits/d113014e9506a49e41fcd290df2d10e634aa7c6f#diff-fe5ca9f6e0254b639a10bcc000406fdbR1439

realm/realm-library/src/androidTest/java/io/realm/RxJavaTests.java
```diff
@@ -1105,13 +1105,26 @@ public void realmResults_readableAcrossThreads() {
...
 
+        AtomicLong startingThread = new AtomicLong(Thread.currentThread().getId());
+        AtomicLong subscriberThread = new AtomicLong();
         subscription = realm.where(AllTypes.class).sort(AllTypes.FIELD_LONG).findAllAsync().asFlowable()
+                .doOnSubscribe((r) -> {
+                    subscriberThread.set(Thread.currentThread().getId());
+                })
+                // Note, that Realm automatically subscribes on the thread with the live Realm
+                // so calling subscribeOn has very little effect.
+                // In most cases you probably want to call `observeOn` directly after `ad
```
nhachicha: incomplete sentence

realm/realm-library/src/androidTest/java/io/realm/RxJavaTests.java
```diff
@@ -1105,13 +1105,26 @@ public void realmResults_readableAcrossThreads() {
...
                 .subscribeOn(Schedulers.io())
-                .filter(RealmResults::isLoaded)
+                .filter(results -> {
+                    assertNotEquals(startingThread, subscriberThread);
```
nhachicha: you should check the value `.get()` not the references

realm/realm-library/src/main/java/io/realm/BaseRealm.java
```diff
@@ -245,17 +245,33 @@ public boolean isInTransaction() {
      * when subscribed to. Items will continually be emitted as the Realm is updated -
      * {@code onComplete} will never be called.
      * <p>
+     * Items emitted from Realm Flowables are frozen (See {@link #freeze()}. This means that they
+     * are immutable and can be be read on any thread.
```
nhachicha: remove `be`

realm/realm-library/src/main/java/io/realm/RealmList.java
```diff
@@ -906,6 +906,21 @@ public String toString() {
      * subscribed to. RealmList will continually be emitted as the RealmList is updated -
      * {@code onComplete} will never be called.
      * <p>
+     * Items emitted from Realm Flowables are frozen (See {@link #freeze()}. This means that they
+     * are immutable and can be be read on any thread.
```
nhachicha: remove `be`

realm/realm-library/src/main/java/io/realm/RealmList.java
```diff
@@ -947,14 +959,25 @@ public String toString() {
      * <p>
      * RealmList will continually be emitted as the RealmList is updated - {@code onComplete} will never be called.
      * <p>
-     * * Note that when the {@link Realm} is accessed from threads other than where it was created,
-     * {@link IllegalStateException} will be thrown. Care should be taken when using different schedulers
-     * with {@code subscribeOn()} and {@code observeOn()}. Consider using {@code Realm.where().find*Async()}
-     * instead.
+     * Items emitted from Realm Observables are frozen (See {@link #freeze()}. This means that they
+     * are immutable and can be be read on any thread.
```
nhachicha: remove `be`

realm/realm-library/src/main/java/io/realm/RealmObject.java
```diff
@@ -736,6 +736,21 @@ public final void removeAllChangeListeners() {
      * When chaining a RealmObject flowable use {@code obj.<MyRealmObjectClass>asFlowable()} to pass on
      * type information, otherwise the type of the following observables will be {@code RealmObject}.
      * <p>
+     * Items emitted from Realm Flowables are frozen (See {@link #freeze()}. This means that they
+     * are immutable and can be be read on any thread.
```
nhachicha: remove `be`

realm/realm-library/src/main/java/io/realm/RealmObject.java
```diff
@@ -773,14 +784,25 @@ public final void removeAllChangeListeners() {
      * <p>
      * The RealmObject will continually be emitted as it is updated - {@code onComplete} will never be called.
      * <p>
-     * Note that when the {@link Realm} is accessed from threads other than where it was created,
-     * {@link IllegalStateException} will be thrown. Care should be taken when using different schedulers
-     * with {@code subscribeOn()} and {@code observeOn()}. Consider using {@code Realm.where().find*Async()}
-     * instead.
+     * Items emitted from Realm Observables are frozen (See {@link #freeze()}. This means that they
+     * are immutable and can be be read on any thread.
```
nhachicha: remove `be`

realm/realm-library/src/main/java/io/realm/RealmObject.java
```diff
@@ -795,6 +817,21 @@ public final void removeAllChangeListeners() {
      * When chaining a RealmObject observable use {@code obj.<MyRealmObjectClass>asFlowable()} to pass on
      * type information, otherwise the type of the following observables will be {@code RealmObject}.
      * <p>
+     * Items emitted from Realm Flowables are frozen (See {@link #freeze()}. This means that they
+     * are immutable and can be be read on any thread.
```
nhachicha: remove `be`

realm/realm-library/src/main/java/io/realm/RealmObject.java
```diff
@@ -843,15 +881,26 @@ public final void removeAllChangeListeners() {
      * <p>
      * The RealmObject will continually be emitted as it is updated - {@code onComplete} will never be called.
      * <p>
-     * Note that when the {@link Realm} is accessed from threads other than where it was created,
-     * {@link IllegalStateException} will be thrown. Care should be taken when using different schedulers
-     * with {@code subscribeOn()} and {@code observeOn()}. Consider using {@code Realm.where().find*Async()}
-     * instead.
+     * Items emitted from Realm Observables are frozen (See {@link #freeze()}. This means that they
+     * are immutable and can be be read on any thread.
```
nhachicha: remove `be`

realm/realm-library/src/main/java/io/realm/RealmResults.java
```diff
@@ -703,6 +703,21 @@ public void removeChangeListener(OrderedRealmCollectionChangeListener<RealmResul
      * subscribed to. RealmResults will continually be emitted as the RealmResults are updated -
      * {@code onComplete} will never be called.
      * <p>
+     * Items emitted from Realm Flowables are frozen (See {@link #freeze()}. This means that they
+     * are immutable and can be be read on any thread.
```
nhachicha: remove `be`

realm/realm-library/src/main/java/io/realm/RealmResults.java
```diff
@@ -750,14 +762,26 @@ public void removeChangeListener(OrderedRealmCollectionChangeListener<RealmResul
      * time an RealmResults is emitted.
      * <p>
      * RealmResults will continually be emitted as the RealmResults are updated - {@code onComplete} will never be called.
-     * <p>Note that when the {@link Realm} is accessed from threads other than where it was created,
-     * {@link IllegalStateException} will be thrown. Care should be taken when using different schedulers
-     * with {@code subscribeOn()} and {@code observeOn()}. Consider using {@code Realm.where().find*Async()}
-     * instead.
+     * <p>
+     * Items emitted from Realm Observables are frozen (See {@link #freeze()}. This means that they
+     * are immutable and can be be read on any thread.
```
nhachicha: remove `be`

PR #6598, opened 2019-08-28
Fix #6597 - Don't enforce jCenter repository

### LINKED ISSUE #6597: Don't enforce jCenter repository
#### Describe your problem or use case

My project is building on a Jenkins instance, which does not have access to the external network. All our dependencies are downloaded through our local repository:

```groovy
buildscript {
    repositories {
        maven {
            url 'https://maven.my-company.com/'
        }
    }
}

allprojects {
    repositories {
        maven {
            url 'https://maven.my-company.com/'
        }
    }
}
```

In [Realm.groovy#L75](https://github.com/realm/realm-java/blob/789740d8050ba020c8990fd7983fb709b748f7b3/gradle-plugin/src/main/groovy/io/realm/gradle/Realm.groovy#L75), I can see that you are explicitly adding the `jCenter` repository.
This causes my build to take way longer than necessary, because Gradle is trying to connect to an unreachable repository for each of the declared dependency.

#### Describe the solution you'd like

Don't enforce the `jCenter` repository if it is not necessary.

#### Additional context

In https://github.com/gradle/gradle/issues/10376#issuecomment-525336635, @lacasseio provided some tips on how this issue could be addressed: only adding `jCenter` if no other repositories are defined.

MGaetan89:


gradle-plugin/src/test/groovy/io/realm/gradle/PluginTest.groovy
```diff
@@ -108,6 +108,71 @@ class PluginTest {
...
+        }
+
+        project.evaluate()
+
+        assertTrue(project.repositories.size() == 1)
+        assertTrue(project.repositories.contains(project.getRepositories().jcenter()))
+    }
+
+    @Test
+    void pluginAddsRightRepositories_withRepositoriesSet() {
```
cmelchior: Nice tests 👏

MGaetan89:
@cmelchior I didn't have time to run the tests to see if my changes are correct, that's why I put it as draft. If you think it's correct like that, feel free to merge it :)

cmelchior:
I ran the tests locally. They seem to pass 👍

gradle-plugin/src/test/groovy/io/realm/gradle/PluginTest.groovy
```diff
@@ -108,6 +108,71 @@ class PluginTest {
...
+
+            defaultConfig {
+                minSdkVersion 16
+                targetSdkVersion 27
+            }
+        }
+
+        project.evaluate()
+
+        assertTrue(project.repositories.size() == 1)
```
lacasseio: This doesn't assert that a repository was added to the `project.buildscript` bloc. It actually tests for the `project.repositories`. Although plugins could, in theory, change the `project.buildscript.repositories` container, it has no effect as the `project.buildscript` DSL is only used for building the classpath for Gradle to run not the compile classpath. There is no change to this code sample required but the `pluginAddsRightRepositories_withRepositoriesSet` test will need to be fixed.

gradle-plugin/src/test/groovy/io/realm/gradle/PluginTest.groovy
```diff
@@ -108,6 +108,71 @@ class PluginTest {
...
+        assertTrue(project.repositories.contains(project.getRepositories().jcenter()))
```
lacasseio: ❌ This doesn't test anything. The fact of calling `project.repositories.jcenter()` returns a `jcenter` `ArtifactRepository` instance but also adds it to the container. This assertion is a no-op in this case. A better way to assert is looking that the host of the URL of the first repository is `jcenter.bintray.com`. Something like this would work: `assertTrue(project.repositories.first().url.host == 'jcenter.bintray.com')`.

gradle-plugin/src/test/groovy/io/realm/gradle/PluginTest.groovy
```diff
@@ -108,6 +108,71 @@ class PluginTest {
...
+            dependencies {
+                classpath "com.android.tools.build:gradle:${projectDependencies.get("GRADLE_BUILD_TOOLS")}"
+                classpath 'com.jakewharton.sdkmanager:gradle-plugin:0.12.0'
+            }
+        }
+
+        def manifest = project.file("src/main/AndroidManifest.xml")
+        manifest.parentFile.mkdirs()
+        manifest.text = '<manifest xmlns:android="http://schemas.android.com/apk/res/android"  package="com.realm.test"></manifest>'
+
```
lacasseio: ❌ You need to add the following before the plugin is applied to correctly assert for this test:
```
repositories {
    google()
}
```

gradle-plugin/src/test/groovy/io/realm/gradle/PluginTest.groovy
```diff
@@ -108,6 +108,71 @@ class PluginTest {
...
+            defaultConfig {
+                minSdkVersion 16
+                targetSdkVersion 27
+            }
+        }
+
+        project.evaluate()
+
+        assertTrue(project.getRepositories().size() == 1)
+        assertTrue(project.repositories.contains(project.getRepositories().google()))
```
lacasseio: ❌ As explained earlier, this will result in two repositories after this executes. A similar assertion as the previous one would fix the assertion.

gradle-plugin/src/test/groovy/io/realm/gradle/PluginTest.groovy
```diff
@@ -108,6 +108,71 @@ class PluginTest {
...
+                classpath 'com.jakewharton.sdkmanager:gradle-plugin:0.12.0'
+            }
+        }
+
+        def manifest = project.file("src/main/AndroidManifest.xml")
+        manifest.parentFile.mkdirs()
+        manifest.text = '<manifest xmlns:android="http://schemas.android.com/apk/res/android"  package="com.realm.test"></manifest>'
+
+        project.apply plugin: 'com.android.application'
+        project.apply plugin: 'realm-android'
```
lacasseio: [nicpick] Relying on the apply order of plugins is considered a bad practice. It would be better to be reactive to the other plugins instead. This way a user can apply the plugin using the plugin bloc or in whatever order they wish while maintaining the same behaviour. This has repercussions on how the plugin is architectured which would require some change. It's something to consider changing to improve user happiness.

gradle-plugin/src/main/groovy/io/realm/gradle/Realm.groovy
```diff
@@ -72,7 +72,11 @@ class Realm implements Plugin<Project> {
 
         project.android.registerTransform(new RealmTransformer(project))
 
-        project.repositories.add(project.getRepositories().jcenter())
+        if (project.repositories.isEmpty()) {
```
lacasseio: ❌ This only account for repositories added before the plugin is applied. If, for example, a user would have something like the following:
```
apply plugin: 'realm-android'

repositories {
    maven {
        url 'https://jcenter.my-company.com'
    }
}
```

There would be two repositories added where `jcenter.bintray.com` would still be the first one to resolve. A better solution, unfortunately using the dredded `afterEvaluate` would be something like this:
```
project.afterEvaluate {
    if (project.repositories.isEmpty()) {
        project.repositories.jcenter()
    }
}
```

This will react by waiting at the end of the evaluation of the script before checking the if no repositories are added. There are still issues when a user would also add repositories during an `afterEvaluate` bloc, but it should be quite limited. The Gradle team will hopefully provide an API to default repositories when none is specified without relying on `afterEvaluate` but I can't say when that will happen.

gradle-plugin/src/test/groovy/io/realm/gradle/PluginTest.groovy
cmelchior: Cool. Thanks for clarifying

MGaetan89:
@lacasseio thank you for your inputs! I will submit a new PR to improve these changes.

PR #6599, opened 2019-08-28
[RJAVA-4] Bypass the sync proxy on Cloud

fealebenpae:
See https://github.com/realm/realm-object-store/pull/824


realm/realm-library/src/objectServer/java/io/realm/internal/objectserver/SyncWorker.java
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    public SyncWorker(String path) {
+        this.path = path;
+    }
+
+    public String path() {
+        return path;
+    }
+}
```
cmelchior:
```suggestion
}

```

realm/realm-library/src/objectServer/java/io/realm/SyncSession.java
```diff
@@ -971,4 +978,5 @@ public void throwExceptionIfNeeded() {
     private static native byte nativeGetConnectionState(String localRealmPath);
     private static native void nativeStart(String localRealmPath);
     private static native void nativeStop(String localRealmPath);
+    private static native void nativeSetUrlPrefix(String localRealmPath, String urlPrefix);
```
cmelchior: So it is safe to call this method at any point during the sessions lifecycle?

fealebenpae: We can call this method at any point, but it will only have an effect before the sync session is bound with an access token. Because the server always returns the same value this method is effectively idempotent.

fealebenpae:
@cmelchior I'm getting build failures in the gradle plugin tests, but I'm not sure if it's caused by my changes or not. Is it safe to ignore?

cmelchior:
Yes, they are safe to ignore. Weird, I tested those locally, but I need to take a second look anyway due to feedback on how they are written.

PR #6602, opened 2019-08-30
Fix #6597 - Don't enforce jCenter repository (follow up #6598)

### LINKED ISSUE #6597: Don't enforce jCenter repository

MGaetan89:
Follow up PR for #6598 to address @lacasseio comments

### DIFF
--- a/gradle-plugin/src/main/groovy/io/realm/gradle/Realm.groovy
+++ b/gradle-plugin/src/main/groovy/io/realm/gradle/Realm.groovy
@@ -72,9 +72,12 @@ class Realm implements Plugin<Project> {
 
         project.android.registerTransform(new RealmTransformer(project))
 
-        if (project.repositories.isEmpty()) {
-            // If no repository was defined, we add jCenter
-            project.repositories.add(project.getRepositories().jcenter())
+        project.afterEvaluate {
+            if (project.repositories.isEmpty()) {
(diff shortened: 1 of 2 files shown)


MGaetan89:
@lacasseio I followed what the Realm's team had already in place. I don't know if they have a plan to move to TestKit.

PR #6615, opened 2019-09-10
Replace Permissions with REST API

cmelchior:
Closes https://github.com/realm/realm-java-private/issues/264

This PR replaces the Realm backed `PermissionManager` with a REST API instead.
This also means that all the methods on the PermissionManager are now folded into the `SyncUser` just like the other SDK's. Previously it was in a separate class to better manage the Realm lifecycle.

This is a breaking change so it will be shipped as a "minor" 6.0 before other planned changes which will now go into 7.0 instead.

**New API:**
Uses the same naming as https://github.com/realm/realm-js/pull/2410/files with a few exceptions:

* `getX` methods have been replaced with `retrieveX` as `getX` in Java indicates something cheap and fast, which a network request definitely is not. Also `retrieve` is already used by the `retrieveUserInfo` method, so there is a precedent for it already.

* Two variants of all methods are provided, a normal synchronous blocking method and a `XAsync()` variant. This is already the pattern for other methods on SyncUser.

This means the new API looks like this:
```
// SyncUser.java
List<Permission> retrieveGrantedPermissions();
RealmAsyncTask retrieveGrantedPermissions(Callback<List<Permission> callback);

List<Permission> applyPermissions(PermissionRequest request);
RealmAsyncTask applyPermissions(PermissionRequest request, Callback<Void> callback);

String makePermissionsOffer(PermissionOffer offer);
RealmAsyncTask makePermissionsOffer(PermissionOffer offer, Callback<String> callback);

String acceptPermissionsOffer(String offerToken);
RealmAsyncTask acceptPermissionsOffer(String offerToken, Callback<String> callback);

void invalidatePermissionsOffer(String offerToken);
RealmAsyncTask invalidatePermissionsOffer(String offerToken, Callback<Void> callback))

List<PermissionOffer> retrieveCreatedPermissionsOffers();
RealmAsyncTask retrieveCreatedPermissionsOffers(Callback<List<PermissionOffer>> callback);
```

**Open questions:**
* Right now I'm just using the one (existing) callback interface. In the PermissionManager we had multiple. There are tradeoffs to both approaches, especially having `Callback<Void>` is a bit annoying, but at the same time having 6-7 different callback interfaces are also hard to maintain.

TODO:
- [x] Cleanup public Javadocs
- [x] Add support for Http redirects anticipated as part of the migration to Stitch backend.
- [x] Fix Test server not correctly shutting down.
- [x] Changelog


realm/realm-library/src/objectServer/java/io/realm/internal/network/AcceptPermissionsOfferResponse.java
```diff
@@ -0,0 +1,91 @@
...
+
+import java.io.IOException;
+
+import io.realm.ErrorCode;
+import io.realm.ObjectServerError;
+import io.realm.log.RealmLog;
+import okhttp3.Response;
+
+/**
+ * FIXME
```
blagoev: is this fixed or something else should happen ?

realm/realm-library/src/objectServer/java/io/realm/internal/network/ApplyPermissionsRequest.java
```diff
@@ -0,0 +1,72 @@
...
+
+import org.json.JSONException;
+import org.json.JSONObject;
+
+import io.realm.permissions.AccessLevel;
+import io.realm.permissions.PermissionRequest;
+import io.realm.permissions.UserCondition;
+
+/**
+ * FIXME
```
blagoev: i feel these are leftovers

realm/realm-library/src/objectServer/java/io/realm/permissions/AccessLevel.java
```diff
@@ -16,28 +16,27 @@
...
- * {@link PermissionRequest}.
- * <p>
- * Note that each access level guarantees all allowed actions provided by less permissive access levels.
- * Specifically, users with write access to a Realm can always read from that Realm, and users with administrative
- * access can always read or write from the Realm. This means that {@code NONE < READ < WRITE < ADMIN}.
- *
- * @see PermissionRequest
- * @see io.realm.PermissionManager#applyPermissions(PermissionRequest, PermissionManager.ApplyPermissionsCallback)
- */
+///**
```
blagoev: are the double comments intentional

realm/realm-library/src/syncIntegrationTest/java/io/realm/PermissionManagerTests.java
```diff
@@ -1095,14 +433,63 @@ public void onError(ObjectServerError error) {
...
     }
-
-    @Test
-    @RunTestInLooperThread(emulateMainThread = true)
-    @Ignore("Figure out why clocks on server/emulator on CI seem to differ")
-    public void revokeOffer_alreadyExpired() {
-        fail("Implement this");
-    }
-
+//
```
blagoev: if these tests are not needed why not just remove them

CHANGELOG.md
```diff
@@ -1,3 +1,23 @@
+## 6.0.0(YYYY-MM-DD)
+
+### Breaking Changes
+* [ObjectServer] The `PermissionManager` is no longer backed by Realms but instead a REST API. This means that the `PermissionManager` class has been removed and all methods have been moved to `SyncUser`. Some method names have been renamed slightly and return values for methods have changed from `RealmResults<Permission>` to `List<Permission>`. This should only have an impact if change listeners were used to listen for changes. In these cases, you must now manually retry the request.
```
bmunkholm: Perhaps list here how to upgrade?

realm/realm-library/src/syncIntegrationTest/java/io/realm/PathLevelPermissionsTests.java
```diff
@@ -0,0 +1,507 @@
...
+            user3.acceptPermissionsOffer(offerToken);
+            fail();
+        } catch (ObjectServerError error) {
+            assertEquals(ErrorCode.EXPIRED_PERMISSION_OFFER, error.getErrorCode());
+            looperThread.testComplete();
+        }
+    }
+
+    /**
+     * Creates a offer for a newly created Realm.
```
nhachicha:
```suggestion
     * Creates an offer for a newly created Realm.
```

tools/sync_test_server/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:6.11.4
+FROM node:10
```
nhachicha: why the update to 10?

nirinchev: We dropped support for node 6 some time ago.

nirinchev:
@cmelchior seems like everyone's happy with this - should we just merge and release?

cmelchior:
Yes, I just want to merge and release https://github.com/realm/realm-java/pull/6621 first in a patch release. After that  I'll merge this.

PR #6617, opened 2019-09-12
Make setList Javadoc more accurate.

### LINKED ISSUE #6616: DynamicRealmObject.setList documentation is misleading
#### Goal
Setting a list-field of a DynamicRealmObject by reference and adding values to that list afterwards should result in the DynamicRealmObject's list to contain the added value.

#### Actual Results
The list is not set by reference, although the documention implies so. It only copies the values of the list.

```
DynamicRealmObject handle = realm.createObject(Handle.class.getSimpleName(), "testhandle");
RealmList<DynamicRealmObject> tokens = new RealmList<>();
handle.setList("tokens", tokens);
DynamicRealmObject token = realm.createObject(Token.class.getSimpleName(), "testtoken");
tokens.add(token);
```
The handle is created in the DB, the token is also created in the DB, but the token is not included in the handle's token-list. According to the documention, I would expect that it is included.

Realm version(s): 5.13.1

cmelchior:
Closes #6616

### DIFF
--- a/realm/realm-library/src/main/java/io/realm/DynamicRealmObject.java
+++ b/realm/realm-library/src/main/java/io/realm/DynamicRealmObject.java
@@ -757,7 +757,7 @@ public class DynamicRealmObject extends RealmObject implements RealmObjectProxy 
      * Sets the reference to a {@link RealmList} on the given field.
      * <p>
      * This will copy all the elements in the list into Realm, but any further changes to the list
-     * will not be reflected in the Realm. Use {@link #getList(String)} in order to a reference to
+     * will not be reflected in the Realm. Use {@link #getList(String)} in order to get a reference to
      * the managed list.
      *
      * @param fieldName field name.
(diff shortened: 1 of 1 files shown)


realm/realm-library/src/main/java/io/realm/DynamicRealmObject.java
```diff
@@ -755,6 +755,10 @@ public void setObject(String fieldName, @Nullable DynamicRealmObject value) {
 
     /**
      * Sets the reference to a {@link RealmList} on the given field.
+     * <p>
+     * This will copy all the elements in the list into Realm, but any further changes to the list
+     * will not be reflected in the Realm. Use {@link #getList(String)} in order to a reference to
```
bmunkholm:
```suggestion
     * will not be reflected in the Realm. Use {@link #getList(String)} in order to get a reference to
```
Was that what you meant?

Issue #6619, opened 2019-09-18
Android updated realm-gradle-plugin from 5.12 to 5.15 throws exception


tloshi:
#### Goal
  I updated realm from 5.12 to 5.15 and without any changes on the models I got this exception when updating the app, but if I install as a fresh apk it will work.
 
java.lang.IllegalStateException: The following changes cannot be made in additive-only schema mode:
    - Property '__ResultSets.created_at' has been made required.
    - Property '__ResultSets.updated_at' has been made required.
        at io.realm.internal.OsSharedRealm.nativeGetSharedRealm(Native Method)
        at io.realm.internal.OsSharedRealm.<init>(OsSharedRealm.java:171)
        at io.realm.internal.OsSharedRealm.getInstance(OsSharedRealm.java:241)
 

#### Steps & Code to Reproduce
 Calling like this:
          Realm.setDefaultConfiguration(SyncConfiguration.automatic())
          mRealm = Realm.getDefaultInstance()
 
#### Version of Realm and tooling

Realm version(s): 5.12 <-> 5.15.1

Realm Sync feature enabled: Yes 

Android Studio version: 3.5

Android Build Tools version: 29

Gradle version: 3.5.0

Which Android version and device(s): All versions

cmelchior:
Hmm, this could look like something we might have caused by accident. I'll look into it. Thanks for reporting.

cmelchior:
This should be fixed in Realm Java 5.15.2

PR #6456, opened 2019-03-07
Better Subscription lifecycle management

cmelchior:
Closes #6453
Closes #2722

This PR adds support for new timestamps which enables better tracking of Subscriptions. It also makes it possible to set a TTL for subscriptions as well as updating the underlying query.

**New API's:**
````
// Uses the same naming conventions as copyToRealmOrUpdate or createObjectOrUpdate
RealmQuery.subscribe(String name, long timeToLive, TimeUnit timeUnit);
RealmQuery.subscribeOrUpdate(String name, long timeToLive, TimeUnit timeUnit);
RealmQuery.subscribeOrUpdate(String name);

// These methods use an update boolean since the creation of subscriptions 
// are a side-effect and calling them findAllAsyncOrUpdate feels really awkward. 
// It does make it a little inconsistent though. 
RealmQuery.findAllAsync(String name, boolean update);
RealmQuery.findAllAsync(String name, long timeToLive, TimeUnit timeUnit);
RealmQuery.findAllAsync(String name, long timeToLive, TimeUnit timeUnit, boolean update);
````

TODO
- [x] Finalize OS PR and direct to master. Will need a few adjustments when this is happening.
- [x] Add `findAllAsync()` API's + tests
- [x] Add `RealmQuery.subscribe()` API's + tests
- [x] Extend Subscription + tests
- [ ] Document new API's on website

PR #6621, opened 2019-09-18
IllegalStateException when opening old synchronized Realm

cmelchior:
Closes https://github.com/realm/realm-java/issues/6619

The fields added in Realm Java 5.10 for better subscription management were made optional by mistake. This wasn't caught in testing and due to how OS initialized its code, wasn't caught there either.

However, after 5.13.0 OS changed how it initialized __ResultSets, which now caused schema mismatches for older Java clients upgrading. 

This PR attempts to fix this with minimal hassle to users. Basically, we silently rewrite the schema, before Object Store uses it. This means that if you introspect the schema it will report something different than what is defined in the `Subscription.java` class, which given the alternatives seems acceptable.

TODO:
- Merge https://github.com/realm/realm-object-store/pull/849 and update changelog and OS commit


CHANGELOG.md
```diff
@@ -1,3 +1,20 @@
+## 5.15.2(YYYY-MM-DD)
+
+### Enhancements
+* None.
+
+### Fixed
+* [ObjectServer] Fixed crash with `java.lang.IllegalStateException: The following changes cannot be made in additive-only schema mode` when opening an old Realm. (Issue [#6619](https://github.com/realm/realm-java/issues/6619), since 5.13.0).
```
bmunkholm: Want to define "old"?

cmelchior:
Looks like this might require more work. Observations so far:

- In #6456 `createdAt` and `updatedAt` was added as optional properties (Realm Java 5.10.0)
- In #6548  `createdAt` and `updatedAt` was made required (Realm Java 5.13.0)
- From what I can tell, the code in Object Store creating these properties always made them required: https://github.com/realm/realm-object-store/blob/master/src/sync/partial_sync.cpp#L121
- But this code is never called on Java since the Java schema is used first: https://github.com/realm/realm-object-store/blob/master/src/impl/realm_coordinator.cpp#L287

This wasn't caught in testing as we don't have any tests opening previous synchronized Realms.

However, it is a big headache, as no matter what we do, someone can risk getting a schema mismatch error.

cmelchior:
I had to rework the PR, so it now needs a 2nd round of eyes.

PR #6652, opened 2019-11-01
Realm plugin: consider kotlin-multiplatform plugin

sellmair:
Using the Kotlin multiplatform plugin instead of the "kotlin-android" one currently, lead to the kapt dependencies not being applied.

### DIFF
--- a/gradle-plugin/src/main/groovy/io/realm/gradle/Realm.groovy
+++ b/gradle-plugin/src/main/groovy/io/realm/gradle/Realm.groovy
@@ -43,7 +43,7 @@ class Realm implements Plugin<Project> {
 
         def syncEnabledDefault = false
         def usesAptPlugin = project.plugins.findPlugin('com.neenbedankt.android-apt') != null
-        def isKotlinProject = project.plugins.findPlugin('kotlin-android') != null
+        def isKotlinProject = project.plugins.findPlugin('kotlin-android') != null || project.plugins.findPlugin("kotlin-multiplatform") != null
         def useKotlinExtensionsDefault = isKotlinProject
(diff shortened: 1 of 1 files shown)


cmelchior:
Hi @sellmair Sorry for the late reply.

I'm not inherently against merging this, but Realm doesn't work across both iOS and Android this way, so I assume you are trying to run Realm in the Android part of the project?

In that case, we just need to make it really clear what works and what doens't

sellmair:
Absolutely: We are using Realm as the "platform implementation detail" for Android inside a multiplatform project. Obviously, this will only make the "realm" Gradle plugin compatible with the "kotlin-multiplatform"  plugin (when used instead of "kotlin-android").  This will just fix the issue of the missing kapt dependency. It's still only applied for Android only.

It will just consider the multiplatform plugin when defining `def isKotlinProject`

sellmair:
Any updates on that? I am sure there are some people that would like to use Realm inside a multiplatform project already! It would be very cool if this would work out of the box 😇

cmelchior:
@sellmair Yes, I think we can merge this, but it looks like you haven't signed the CLA: https://github.com/realm/realm-java/blob/master/CONTRIBUTING.md#cla

sellmair:
Signed 👍

cmelchior:
Thank you 👏

SebOh:
I would like to use the fix already :) ... When is the next release planed, including this update?

sellmair:
Hey @SebOh, 
you can use the following workaround until this fix is released:

> Workaround
> Add the kapt dependency
> 
> kapt("io.realm:realm-annotations-processor:${realmVersion}")

PR #6695, opened 2019-12-17
Update CHANGELOG.md

bmunkholm:
Isn't the beta release versioned with a number?  "-beta.0"?
The date of the release needs update.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,10 @@
 ## 7.0.0-beta (YYYY-MM-DD)
 
-### Breaking Changes
+Based on v6.0.2.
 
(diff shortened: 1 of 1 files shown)


bmunkholm:
Ok. So perhaps we should mention that in the release?
Also warn not to use this in production?
It makes it harder to handle bug reports, as we have no clue which version they saw the issue in :-)

bmunkholm:
@cmelchior This also means we need to update the docs as we refer to "beta-0" there.

nhachicha:
@bmunkholm fair point on the bug report, we can add `beta-0` as long as we suffix it with `-SNAPSHOT`, the warning could be added in the doc ...

cmelchior:
Not sure it makes sense to add numbers to the beta label if we release as SNAPSHOT. 
I'll update the website docs to include the SNAPSHOT setup.

PR #6723, opened 2020-01-16
Upgrade to latest version of Sync and Core

cmelchior:


realm/realm-library/src/objectServer/java/io/realm/ObjectServer.java
```diff
@@ -87,12 +81,12 @@ public static void init(Context context, String appDefinedUserAgent) {
                             "Directory '%s' for SyncManager cannot be created. ",
                             dir.getPath()));
                 }
-                SyncManager.nativeInitializeSyncManager(dir.getPath(), userAgent);
+                SyncManager.nativeInitializeSyncManager(dir.getPath(), userAgentBindingInfo, appDefinedUserAgent);
```
nhachicha: where is `appUserAgentInfo` populated?

realm/realm-library/src/androidTest/java/io/realm/RealmResultsTests.java
```diff
@@ -1765,36 +1765,40 @@ public void asJSON() throws JSONException {
...
                 "    {\n" +
+                "        \"_key\":100," +
                 "        \"columnString\": \"alltypes1\",\n" +
                 "        \"columnLong\": 1337,\n" +
                 "        \"columnFloat\": 3.1400001,\n" +
                 "        \"columnDouble\": 0.89122999999999997,\n" +
                 "        \"columnBoolean\": false,\n" +
                 "        \"columnDate\": \"" + now + "\",\n" +
-                "        \"columnBinary\": \"010203\",\n" +
+                "        \"columnBinary\": \"AQID\",\n" +
```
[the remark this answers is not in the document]
cmelchior: No idea, it concerned me a bit as well and I meant to ask Core about it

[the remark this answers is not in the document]
cmelchior: No idea, but it concerned me a bit as well. I'll ask Core about it

realm/realm-library/src/objectServer/java/io/realm/ObjectServer.java
cmelchior: It is an optional setting set through `Realm.init(context, userAgent)`

realm/realm-library/src/androidTest/java/io/realm/RealmResultsTests.java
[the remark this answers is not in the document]
cmelchior: Binary data is now Base64 encoded apparently. I'll add a mention in the changelog. Normally it would be considered a breaking change as you removed the `@Beta` label on the `asJson()` method, but since it is also mentioned that the method is only for debugging it is probably acceptable? WDYT?

realm/realm-library/src/objectServer/java/io/realm/ObjectServer.java
nhachicha: is this exposed for test? who needed this override?

cmelchior: It was added in this PR: https://github.com/realm/realm-java/pull/6270
It looks like we don't have automatic tests for it, but only verified the behavior manually.

PR #6736, opened 2020-01-31
Remove Permissions and ObjectLevel permissions

cmelchior:
Permissions are now managed through MongoDB Realm and Stitch functions, not through dedicated SDK API's. 

This PR remove just remove all current API's and tests for Permissions and Object Level Permissions. Stich functionality will be added later.


cmelchior:
Ready for review

realm/realm-library/src/androidTestObjectServer/java/io/realm/SyncedRealmTests.java
```diff
@@ -92,6 +93,7 @@ private Realm getFullySyncRealm() {
 
     // Test for https://github.com/realm/realm-java/issues/6619
     @Test
+    @Ignore("Going to be removed anyway")
     public void testUpgragendingOptionalSubscriptionFields() throws IOException {
```
nhachicha:
```suggestion
    public void testUpgradingOptionalSubscriptionFields() throws IOException {
```

realm/realm-library/src/main/java/io/realm/internal/sync/permissions/ObjectPermissionsModule.java
```diff
@@ -0,0 +1,13 @@
+package io.realm.internal.sync.permissions;
+
+import io.realm.annotations.RealmModule;
+import io.realm.sync.Subscription;
+
+/**
+ * Realm model classses that are always part of Query-based Realms
```
nhachicha:
```suggestion
 * Realm model classes that are always part of Query-based Realms
```

realm/realm-library/src/main/java/io/realm/internal/sync/permissions/ObjectPermissionsModule.java
```diff
@@ -0,0 +1,13 @@
...
+ */
+@RealmModule(library = true, classes = {
+        Subscription.class
+})
+public class ObjectPermissionsModule {
```
nhachicha: Are we still keeping this permission model for QBS though? I thought all permission including object-level permission are going to be on the Stitch side as well?

cmelchior: No, I'm going to remove this when I remove subscriptions. I just went with minimal changes when removing the permissions.

PR #6745, opened 2020-02-07
Remove Query-based-sync API's

cmelchior:
This PR removes all API's and mentions of Query-based Sync. "Query-based-sync" in MongoDB will initially be supported by providing a "partionKey" to the `SyncConfiguration`. This will come in a later PR.

In particular this means:
* Removing the Subscription class
* Removing all support for Subscriptions in Realm and RealmQuery
* Remove `fullSynchronization` as a parameter on SyncConfiguration
* Cleanup internal API's
* Remove all Query-based sync tests


cmelchior:
Ready for review.

This PR is first step towards the SDK merge with Stitch. It removes all Query-based API's and unit tests related to that.

A few of the changes to unit tests are temporary until we can test against a proper test server.

Once this PR is merged. The next will remove all permission API's related to full Realms.

PR #6753, opened 2020-02-21
Disable building debug builds for release targets

cmelchior:
Calling `assemble` builds both debug and release targets. No reason for that when building an artifact that is deployed.

TODO:
- [ ] Measure difference


cmelchior:
Hmm, this doesn't work for some reason. I still think there is room for optimization. But I don't have time to take a closer look. So closing this for now.

PR #6758, opened 2020-02-27
Remove Query-based Sync API's

cmelchior:
This was accidentially closed when renaming branches. See https://github.com/realm/realm-java/pull/6745

----

This PR removes all API's and mentions of Query-based Sync. "Query-based-sync" in MongoDB will initially be supported by providing a "partionKey" to the SyncConfiguration. This will come in a later PR.

In particular this means:

- Removing the Subscription class
- Removing all support for Subscriptions in Realm and RealmQuery
- Remove fullSynchronization as a parameter on SyncConfiguration
- Cleanup internal API's
- Remove all Query-based sync tests


realm/realm-library/src/main/java/io/realm/internal/RealmObjectProxy.java
```diff
@@ -34,6 +35,7 @@
      * Tuple class for saving meta data about a cached RealmObject.
      */
     class CacheData<E extends RealmModel> {
+        @SuppressFBWarnings("URF_UNREAD_PUBLIC_OR_PROTECTED_FIELD")
         public int minDepth;
```
nhachicha: Why silencing FindBug? is this used from the generated class?

realm/realm-library/src/syncIntegrationTest/java/io/realm/SyncSessionTests.java
```diff
@@ -652,7 +642,9 @@ public void stop_multipleTimes() {
 
     @Test
     @RunTestInLooperThread
+    @Ignore("FIXME https://github.com/realm/realm-sync/issues/3234")
     public void waitForInitialRemoteData_throwsOnTimeout() {
+        RealmLog.setLevel(LogLevel.DEBUG);
```
nhachicha: Leftover for the FIXME?

realm/realm-library/src/syncIntegrationTest/java/io/realm/SyncSessionTests.java
```diff
@@ -652,7 +642,9 @@ public void stop_multipleTimes() {
 
     @Test
     @RunTestInLooperThread
+    @Ignore("FIXME https://github.com/realm/realm-sync/issues/3234")
```
nhachicha: The Java & Core issue related to this seems closed, do you want to enable this again?

realm/realm-library/src/syncIntegrationTest/java/io/realm/SyncedRealmIntegrationTests.java
```diff
@@ -299,7 +293,7 @@ public void waitForInitialRemoteData_readOnlyTrue_throwsIfWrongServerSchema() {
             // schema.
             realm = Realm.getInstance(configNew);
             fail();
-        } catch (IllegalStateException ignored) {
+        } catch (RealmMigrationNeededException ignored) {
```
nhachicha: Hmm. this is confusing `RealmMigrationNeededException` doesn't make sense in an additive Sync schema scenario...

cmelchior: Good point. I'll fix it.

realm/realm-library/src/syncIntegrationTest/java/io/realm/SyncSessionTests.java
cmelchior: Yup, reenabling.

realm/realm-library/src/main/java/io/realm/internal/RealmObjectProxy.java
cmelchior: Yes, it is referenced by our proxy classes. Not entirely sure why it suddenly starts complaining.

realm/realm-library/src/syncIntegrationTest/java/io/realm/SyncedRealmIntegrationTests.java
cmelchior: There is something odd here. But changing the exception to `IllegalState` causes other errors. It looks like a difference between query-based and fully synchronized Realms (since this PR changes the default back to fully synchronized). I have created https://github.com/realm/realm-java/issues/6759 to track it.

PR #6760, opened 2020-02-27
Upgrade to Sync 10.0.0-alpha.1. Disable all Sync builds

cmelchior:
This upgrade to latest Core + Sync relase, but due to a bug in the Sync release it doesn't work. So all Sync builds have been disabled.

Merging this PR should unblock embedded objects (#6730) and the new datatypes though (#6721)

Note there are also a few changes to how AccessTokens are exposed. This is a temporary solution that will be fixed in the RealmApp PR (#6757 )


realm/realm-annotations-processor/src/main/java/io/realm/processor/ClassMetaData.kt
```diff
@@ -698,7 +698,9 @@ class ClassMetaData(env: ProcessingEnvironment, typeMirrors: TypeMirrors, privat
     }
 
     // The field has the @PrimaryKey annotation. It is only valid for
-    // String, short, int, long and must only be present one time
+    // String, short, int, long and must only be present one time.
+    // From Core 6 primary keys are not indexed pr default anymore as
```
nhachicha:
```suggestion
    // From Core 6 primary keys are not indexed by default anymore as
```

nhachicha: Actually in Core 6 String primary keys don’t need an Index, and in Core 10 none of the primary keys do

realm/realm-library/src/main/cpp/io_realm_RealmFileUserStore.cpp
```diff
@@ -68,12 +68,13 @@ JNIEXPORT jstring JNICALL Java_io_realm_RealmFileUserStore_nativeGetUser(JNIEnv*
...
-                                                                                 jstring j_user_id, jstring json_token,
+                                                                                 jstring j_user_id,
+                                                                                 jstring j_refresh_json_token,
                                                                                  jstring j_auth_url)
 {
     try {
-        JStringAccessor user_json_token(env, json_token); // throws
-        SyncManager::shared().get_user(create_sync_user_identifier(env, j_user_id, j_auth_url), user_json_token);
+        JStringAccessor refresh_json_token(env, j_refresh_json_token); // throws
+        SyncManager::shared().get_user(create_sync_user_identifier(env, j_user_id, j_auth_url), refresh_json_token, refresh_json_token);
```
nhachicha: we're passing twice the refresh token as a parameter? ☹️  shouldn't the second parameter be the `access_token`?

realm/realm-library/src/main/cpp/io_realm_internal_OsRealmConfig.cpp
```diff
@@ -346,7 +346,8 @@ JNIEXPORT jstring JNICALL Java_io_realm_internal_OsRealmConfig_nativeCreateAndSe
         if (!user) {
             JStringAccessor realm_auth_url(env, j_auth_url);
             JStringAccessor refresh_token(env, j_refresh_token);
-            user = SyncManager::shared().get_user(sync_user_identifier, refresh_token);
+            JStringAccessor access_token(env, j_access_token);
+            user = SyncManager::shared().get_user(sync_user_identifier, refresh_token, access_token);
```
nhachicha: what happens if the access_token is expired?

realm/realm-library/src/main/cpp/io_realm_RealmFileUserStore.cpp
cmelchior: Yes, I discussed this with Jason. This is just a temporary workaround until we use Stitch for Authentication.

realm/realm-library/src/main/cpp/io_realm_internal_OsRealmConfig.cpp
cmelchior: This logic will change once we connect using Stitch. Right now this is just a temporary workaround.

PR #6767, opened 2020-03-05
Add support for RealmProxy#toString() to print length information for binary field

### LINKED ISSUE #1650: RealmProxy#toString() should print length information for binary field
Current implementation simply call `byte[].toString()` and `RealmProxy.toString()` generates something like `"Foo = [{foo:[I@23fc4bec}]"`.

We should print for `binary(25)`, `25 bytes binary`, etc.

joxon:
Fixes #1650

### DIFF
--- a/realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmProxyClassGenerator.kt
+++ b/realm/realm-annotations-processor/src/main/java/io/realm/processor/RealmProxyClassGenerator.kt
@@ -1632,6 +1632,9 @@ class RealmProxyClassGenerator(private val processingEnvironment: ProcessingEnvi
                         Utils.isMutableRealmInteger(field) -> {
                             emitStatement("stringBuilder.append(%s().get())", metadata.getInternalGetter(fieldName))
                         }
+                        field is ByteArray -> {
+                            emitStatement("stringBuilder.append(\"binary(\" + field.length + \")\")")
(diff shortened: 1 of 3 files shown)


cmelchior:
Hi @joxon Thank you very much for adding this 👏 . The change itself looks like, but it would be nice with a unit test as well to ensure that the functionality both works now and doesn't break in the future.

Also, it looks you already signed our CLA 👍

bmunkholm:
An update in the changelog.md would also be needed.

joxon:
> The change itself looks fine, but it would be nice with a unit test as well to ensure that the functionality both works now and doesn't break in the future.

Hi. I noticed there is an existing test case that could verify the correctness, Can I update that instead of writing new test case?

bmunkholm:
@joxon Absolutely! That would be perfect!

joxon:
Okay, the test cases are done. How should I update the changelog.md?

cmelchior:
@joxon Thanks for doing this. I'll update the CHANGELOG you don't have to worry about it.

Raynor999:
NPE： 
```
 field is ByteArray -> {
                            emitStatement("stringBuilder.append(\"binary(\" + field.length + \")\")")
                        }
```
https://github.com/realm/realm-java/pull/6767

Caused by: java.lang.NullPointerException: Attempt to get length of null array

PR #6770, opened 2020-03-07
Temporarely disable broken unit tests

cmelchior:
Disable all tests using the SyncUser. The new token format is breaking a lot of tests. These should be re-enabled as part of #6757


cmelchior:
Unclear why it is marked as failed. CI should be green. Merging and adding to https://jira.mongodb.org/browse/RJAVA-520

Issue #6764, opened 2020-02-28
Tests: Flaky RxJava tests?


cmelchior:
With the upgrade to Core 6, they appear to have become more unstable.

This is a bit anecdotal, so this issue is trying to track the cases where it happens:

findFirstAsync_emittedOnUpdate – io.realm.RxJavaTests
https://ci.realm.io/blue/organizations/jenkins/realm%2Frealm-java/detail/PR-6760/7/tests 
https://ci.realm.io/blue/organizations/jenkins/realm%2Frealm-java/detail/PR-6760/9/pipeline/

findFirstAsync_emittedOnDelete – io.realm.RxJavaTests
https://ci.realm.io/blue/organizations/jenkins/realm%2Frealm-java/detail/PR-6760/10/tests

cmelchior:
Fixed in #6776

PR #6776, opened 2020-03-14
Fix ProgressListeners for Core 6

cmelchior:
Closes https://github.com/realm/realm-java/issues/6764

- This should remove the last known bug for Core 6.
- Cleaned up the changelog preparing for release.
- Also found the bug with our RxJava tests. It turned out to reveal a true issue. Apparently, the async queries are so fast on my emulator and device that it resolved correctly before the subscription triggered, but on CI, we attempted to subscribe to a still-pending object which failed.

### DIFF
--- a/realm/realm-library/src/androidTest/java/io/realm/RxJavaTests.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/RxJavaTests.java
@@ -247,8 +247,17 @@ public class RxJavaTests {
         final AllTypes asyncObj = realm.where(AllTypes.class).findFirstAsync();
         subscription = asyncObj.<AllTypes>asFlowable().subscribe(rxObject -> {
             assertTrue(rxObject.isFrozen());
+            // Because the subscription is run asynchronously. There is a chance
+            // the query resolved before the subscription triggers.
+            // This means it is not deterministic what state is first emitted here.
+            // It can either be a fully loaded object or one that is still loading.
             assertFalse(rxObject.isValid());
-            assertFalse(rxObject.isLoaded());
+            if (rxObject.isLoaded()) {
+                assertTrue(rxObject.isValid());
+                assertEquals(42, rxObject.getColumnLong());
+            } else {
+                assertFalse(rxObject.isValid());
+            }
             disposeSuccessfulTest(realm);
         });
     }
--- a/realm/realm-library/src/androidTest/java/io/realm/RxJavaTests.java
+++ b/realm/realm-library/src/androidTest/java/io/realm/RxJavaTests.java
@@ -247,7 +247,8 @@ public class RxJavaTests {
         final AllTypes asyncObj = realm.where(AllTypes.class).findFirstAsync();
         subscription = asyncObj.<AllTypes>asFlowable().subscribe(rxObject -> {
             assertTrue(rxObject.isFrozen());
-            assertEquals(42, rxObject.getColumnLong());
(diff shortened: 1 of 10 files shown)


realm/realm-library/src/androidTest/java/io/realm/RxJavaTests.java
```diff
@@ -99,6 +99,8 @@ private void disposeSuccessfulTest(BaseRealm testRealm) {
                 @Override
                 public void run() {
                     // Wait for Subscription to dispose of external resources
+                    RealmLog.error("Global: " + Realm.getGlobalInstanceCount(testRealm.getConfiguration()));
+                    RealmLog.error("Local: " + Realm.getGlobalInstanceCount(testRealm.getConfiguration()));
```
Zhuinden: Should be local

cmelchior: Yeah, I already fixed it. Trying to locate a bug that only manifest on CI. 20 minute roundtrip times when debugging is fun...or something 😆

realm/realm-library/src/androidTest/java/io/realm/RxJavaTests.java
```diff
@@ -247,8 +247,17 @@ public void findFirstAsync_emittedOnSubscribe() {
         final AllTypes asyncObj = realm.where(AllTypes.class).findFirstAsync();
         subscription = asyncObj.<AllTypes>asFlowable().subscribe(rxObject -> {
             assertTrue(rxObject.isFrozen());
+            // Because the subscription is run asynchronously. There is a chance
+            // the query resolved before the subscription triggers.
+            // This means it is not deterministic what state is first emitted here.
+            // It can either be a fully loaded object or one that is still loading.
             assertFalse(rxObject.isValid());
-            assertFalse(rxObject.isLoaded());
+            if (rxObject.isLoaded()) {
```
Zhuinden: Is it not possible to use Realm.refresh() here?

cmelchior: Yes, that would actually also have worked. Might make for a cleaner test 👍

cmelchior: Ups, to fast there. `refresh()` doesn't work on frozen objects which we emit as the default now.

cmelchior:
```
restrictAccessToOwner – io.realm.objectserver.ObjectLevelPermissionIntegrationTests
getPrivileges_serverDefaults – io.realm.objectserver.ObjectLevelPermissionIntegrationTests
```

These two also seem to fail pretty consistently. Needs investigation

realm/realm-library/src/androidTest/java/io/realm/RxJavaTests.java
Zhuinden: Maybe `realmObject.load()`? :thinking:

cmelchior: No, that doesn't work either. It isn't possible to change the state of frozen objects.
The only thing we can do is not emitting unloaded objects, which has to be filtered in the `RealmObservableFactory`.

The issue is that both `findFirstAsync` and `subscribe` are now asynchronous. Previously, `subscribe` was synchronous so you always knew that the first emission was an unloaded object.

PR #6779, opened 2020-03-18
Add support for RealmApp.logOut + associated helper methods

cmelchior:
The primary purpose of this PR was to add support for `RealmApp.logOut()`, but it pulled in a few other things in order to enable testing.

So this PR adds:

**"done done"**
- Support for `RealmApp.logOut()` and `RealmApp.logOutAsync(callback)`
- Support for `RealmUser.logOut()` and `RealmUser.logOutAsync(callback)`
- Support for `RealmUser.getState()`
- Support for `RealmUser.equals()`

**In addition**
- The network transport has been refactored to be more generalized and ease adoption of more functions. In particular, it now supports two types of Callbacks. Those that return a result and those that do not.

- A new helper class `BlockingLooperThread`. This is a replacement for the `@RunInLooperThread` annotation which had a number of problems with lifecycle and combination with other rules. This new class behaves in a similar way as `runBlockingTest` from Googles Coroutine testing library. All of the logic is a copy/paste from the `@RunInLooperThread` rule, but has been massively simplified (so only a light review should be needed). I would like to fully deprecate the annotation based rule in favor of this, but we can do that gradually. I expect all new Kotlin-based unit tests will be using this class. It should still be considered somewhat WIP.

- Refactored how test enabled `RealmApp` is configured and set up. `TestRealmApp` is now a subclass of `RealmApp` with a `close()` method injected. I'm still not sure which approach is best for setting this up. So further iteration for finding the best pattern is probably needed.


cmelchior:
Ready for review @nhachicha

realm/realm-library/src/main/cpp/io_realm_RealmApp.cpp
```diff
@@ -27,6 +28,58 @@ using namespace realm::app;
 using namespace realm::jni_util;
 using namespace realm::_impl;
 
+// Helper method for constructing callbacks for REST calls that must return an actual result to Java
+template<typename T>
+std::function<void(T, Optional<app::AppError>)> create_result_callback(JNIEnv* env, jobject j_callback, const std::function<jobject (JNIEnv*, T)>& success_mapper) {
+    jobject callback = env->NewGlobalRef(j_callback);
+    return [callback, success_mapper](T result, Optional<app::AppError> error) {
```
nhachicha:
```suggestion
    return [&](T result, Optional<app::AppError> error) {
```
why not capture by reference?

realm/realm-library/src/main/cpp/io_realm_RealmApp.cpp
```diff
@@ -35,66 +88,71 @@ JNIEXPORT jlong JNICALL Java_io_realm_RealmApp_nativeCreate(JNIEnv* env, jobject
...
-        if (ret != 0) {
-            throw std::runtime_error(util::format("Failed to get Java VM. Error: %d", ret));
-        }
         jobject java_app_obj = env->NewGlobalRef(obj); // FIXME: Leaking the app object
-        std::function<std::unique_ptr<GenericNetworkTransport>()> transport_generator = [jvm, java_app_obj] {
-            JNIEnv* env;
-            if (jvm->GetEnv((void**)&env, JNI_VERSION_1_6) != JNI_OK) {
-                jvm->AttachCurrentThread(&env, nullptr); // Should never fail
-            }
+        std::function<std::unique_ptr<GenericNetworkTransport>()> transport_generator = [java_app_obj] {
```
nhachicha:
```suggestion
        std::function<std::unique_ptr<GenericNetworkTransport>()> transport_generator = [&] {
```

realm/realm-library/src/main/cpp/io_realm_internal_objectstore_OsSyncUser.cpp
```diff
@@ -20,11 +20,13 @@
 #include "util.hpp"
 #include "jni_util/java_class.hpp"
 
+#include <realm/util/to_string.hpp>
```
nhachicha:
```suggestion
```

realm/realm-library/src/objectServer/java/io/realm/RealmApp.java
```diff
@@ -201,11 +181,67 @@ public RealmUser run() throws ObjectServerError {
...
+     public void logOut() {
+        RealmUser user = currentUser();
+        if (user == null) {
+            throw new IllegalStateException("No current user was found.");
+        }
+        logOut(user);
+     }
 
+    /**
+     * Log the current user out of the Realm App, destroying their server state, unregistering them from the
```
nhachicha:
```suggestion
     * Log the current user out of the Realm App asynchronously, destroying their server state, unregistering them from the
```

realm/realm-library/src/objectServer/java/io/realm/RealmApp.java
```diff
@@ -201,11 +181,67 @@ public RealmUser run() throws ObjectServerError {
...
     }
-    public RealmAsyncTask logoutAsync(RealmUser user, Callback<Void> callback) {
-        return null;
+
+    RealmAsyncTask logOutAsync(RealmUser user, Callback<RealmUser> callback) {
+        checkLooperThread("Asynchronous log out is only possible from looper threads.");
+        return new Request<RealmUser>(SyncManager.NETWORK_POOL_EXECUTOR, callback) {
+            @Override
+            public RealmUser run() throws ObjectServerError {
+                RealmLog.error("Log out user");
```
nhachicha:
```suggestion
                RealmLog.info("Log out user " + user.getEmail());
```

PR #6783, opened 2020-03-24
Add support for EmailPasswordAuthProvider

cmelchior:
Add support for the EmailPasswordAuthProvider and associated functions. 

- registerUser
- confirmUser
- resendConfirmationEmail
- sendResetPasswordEmail
- callResetPasswordFunction
- resetPassword

Integration tests for `confirmUser` and `resetPassword` are not possible currently. Added them as TODO items in #6756 

### Callbacks

Getting great callback support between Java and Kotlin is turning out to be fairly problematic.

Currently, we have this interface in Java
```
    public interface Callback<T> {
        void onSuccess(T t);
        void onError(ObjectServerError error);
    }
```

This actually works pretty well in Java even with Nullability annotations. The problem is that we have a really common use case of `T` being `Void` and that causes problems in Kotlin because you cannot create instances of Void, which means that null is usually used which breaks the type system in Kotlin.

```
val callback = object: RealmApp.Callback<Void> {
  override fun onSuccess(t: Void) { / * Crash with null pointer exception */ }
  override fun onError(error: Throwable) {  }
}
```

Fixing this through annotations has turned out to be impossible due to where these annotations can be placed. I'll be happy to elaborate on the details if needed.

This leaves us with 4 options as far as I can tell

1) Make `onSuccess` always nullable: `public void onSuccess(@Nullable T t)`. This turns into `T?` on the Kotlin side, which gets really annoying for consumers when the type isn't `Void`.

2) Create two interfaces `Callback` and `VoidCallback` with the later having no arguments for onSuccess. This just feels extremely annoying API wise.

3) Split Callback into 3 SAM interfaces. `OnSuccess`, `OnVoidSuccess` and `OnError`. We actually do something similar for transactions right now. This approach looks fairly nice in Kotlin and with lamdas in Java. Without Lambdas it is a lot of boilerplate. I also have concerns with regards to how understandable the API is. It is pretty important that people handle the onError case since it involves a network, so both should always be present. Thus splitting into two interfaces feel a bit wrong. Also still having `OnSuccess` and `OnVoidSuccess` is annoying.

4) The currently implemented solution. I'm using reflection to hack around not being able to create an instance of Void. But with an instance of Void Kotlin support is no longer broken. It does mean that `onSuccess(Void v)` in Java will actually return an instance of Void instead of `null`, but no-body should be checking that anyway. 

### TODO
- [x] Integration tests
- [x] Describe the tradeoffs of the Callback using reflection for Void
- [x] Async implementation


cmelchior:
@nhachicha More reviewing fun for you

realm/realm-library/src/objectServer/java/io/realm/EmailPasswordAuthProvider.java
```diff
@@ -0,0 +1,294 @@
...
+        RealmApp.handleResult(null, error);
+    }
+
+    /**
+     * Call the reset password function configured to the
+     * {@link RealmCredentials.IdentityProvider#EMAIL_PASSWORD} provider.
+     *
+     * @param email the email of the user.
+     * @param newPassword the new password of the user.
+     * @param args any additional arguments provided to to the reset function. All arguments must
```
nhachicha:
```suggestion
     * @param args any additional arguments provided to the reset function. All arguments must
```

realm/realm-library/src/objectServer/java/io/realm/EmailPasswordAuthProvider.java
```diff
@@ -0,0 +1,294 @@
...
+        }.start();
+    }
+
+    /**
+     * Call the reset password function configured to the
+     * {@link RealmCredentials.IdentityProvider#EMAIL_PASSWORD} provider.
+     *
+     * @param email the email of the user.
+     * @param newPassword the new password of the user.
+     * @param args any additional arguments provided to to the reset function. All arguments must
```
nhachicha:
```suggestion
     * @param args any additional arguments provided to the reset function. All arguments must
```

realm/realm-library/src/objectServer/java/io/realm/RealmApp.java
```diff
@@ -225,15 +228,25 @@ public RealmAsyncTask logOutAsync(Callback<RealmUser> callback) {
          return logOutAsync(user, callback);
      }
 
+    /**
+     * Returns a wrapper for interacting with functionality related to users either being created or
+     * logging in using the {@link RealmCredentials.IdentityProvider#EMAIL_PASSWORD} identity provider.
```
nhachicha:
```suggestion
     * login using the {@link RealmCredentials.IdentityProvider#EMAIL_PASSWORD} identity provider.
```

realm/realm-library/src/androidTestObjectServer/kotlin/io/realm/EmailPasswordAuthProviderTests.kt
```diff
@@ -0,0 +1,573 @@
...
+            provider.confirmUserAsync(TestHelper.getNullString(), "token-id", checkNullArgCallback)
+        }
+        looperThread.runBlocking {
+            provider.confirmUserAsync("token", TestHelper.getNullString(), checkNullArgCallback)
+        }
+    }
+
+    @Test
+    fun resendConfirmationEmail() {
+        // We only test that the server successfully accepts the request. We have no way of knowing
```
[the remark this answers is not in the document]
cmelchior: I don't think it is possible to verify this using the Admin API. We might be able to use a custom SMTP server, but it is very unclear to me how much work that is. But I'll add a FIXME here about it.

realm/realm-library/src/androidTestObjectServer/kotlin/io/realm/EmailPasswordAuthProviderTests.kt
```diff
@@ -0,0 +1,573 @@
...
+            fail()
+        } catch (ignore: IllegalArgumentException) {
+        }
+    }
+
+    @Test
+    fun registerUser() {
+        val email = TestHelper.getRandomEmail()
+        val password = "password1234"
+        app.emailPasswordAuthProvider.registerUser(email, password)
```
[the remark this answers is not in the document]
cmelchior: No, pending users and registered users are two different things. Right now I automatically delete all registered users in `app.close()`. There was an undocumented API for doing the same for pending users, so I'll add that to `app.close()` as well.

realm/realm-library/src/androidTestObjectServer/kotlin/io/realm/EmailPasswordAuthProviderTests.kt
```diff
@@ -0,0 +1,573 @@
...
+    }
+
+    @After
+    fun tearDown() {
+        app.close()
+        admin.deleteAllUsers()
+        RealmLog.setLevel(LogLevel.WARN)
+    }
+
+    inline fun testNullArg(method: () -> Unit) {
```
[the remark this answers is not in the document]
cmelchior: Nice 🎉

nhachicha:
@cmelchior regarding the callback we should not rely on a hack since `Void` is not meant to be instantiated. We could envisage an approach similar to [Kotlin Result pattern ](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-result/). Alternatively, the first option is still better than the existing one IMO

_Possible implementation of Kotlin `Result` class in Java_ 
```Java
    public static class Result<T> {
        private T data;
        private Exception exp;

        private Result(@Nullable T d, @Nullable Exception exp) {
            this.data = d;
            this.exp = exp;
        }

        public static <T> Result<T> success() {
            return new Result<T>(null, null);
        }

        public static <T> Result<T> withResult(T res) {
            return new Result<>(res, null);
        }

        public static <T> Result<T> error(Exception exp) {
            return new Result<>(null, exp);
        }

        public boolean isSuccess() {
            return exp != null;
        }

        public T get() {
            return data;
        }

        public Exception getError() {
            return exp;
        }
    }

// usage 
    public interface Callback<T> {
        void onResult(Result<T> r); 
    }

  private void postError(final ObjectServerError error) {
       callback.onResult(Result.error(error));
}

  private void postSuccess(final T result) {
           if (result == null) {
               callback.onResult(Result.success());
           } else {
               callback.onResult(Result.withResult(result)); 
          }
}

```

```Kotlin
// registering callback
provider.resendConfirmationEmailAsync(email, object: RealmApp.Callback<Void> {
                    override fun onResult(r: Result<Void>) {
                        if (r.isSuccess) {
                            // no data to retrieve since it's Void, (otherweise call r.get())  
                        } else {
                            fail(r.error.toString())
                        }
                    }
                })


// this could be simplified to 
provider.resendConfirmationEmailAsync(email) { r ->
                    if (r.isSuccess) {
                        // no data to retrieve since it's Void
                    } else {
                        fail(r.error.toString())
                    }
                }
```

cmelchior:
Hmm, having a result wrapper is actually not a bad idea. It would also allow us to turn the callback into a SAM interface which makes it a lot easier to consume than interfaces with multiple methods.

cmelchior:
I introduced a Result class like you proposed and removed the Void hack. I also copied a few helper methods from the Kotlin one that seemed useful: `getOrThrow()` and `getOrDefault()`

realm/realm-library/src/objectServer/java/io/realm/RealmApp.java
```diff
@@ -448,30 +448,106 @@ private void postSuccess(final T result) {
...
+         * Returns whether or not request was successful
+         *
+         * @return {@code true} if the request was a success, {@code false} if not.
+         */
+        public boolean isSuccess() {
+            return error == null;
+        }
+
+        /**
+         * Returns the response in case the request as a success.
```
nhachicha:
```suggestion
         * Returns the response in case the request was a success.
```

realm/realm-library/src/objectServer/java/io/realm/RealmApp.java
```diff
@@ -448,30 +448,106 @@ private void postSuccess(final T result) {
                 handler.post(new Runnable() {
                     @Override
                     public void run() {
-                        callback.onSuccess(result);
+                        callback.onResult(Result.withResult(result));
```
nhachicha: You're not using `Result.success` is the result always non-null? it doesn't seem to be the case since you're returning null in a couple of callbacks ...

cmelchior: True. I actually didn't think of it, but in this case `Results.withResult(null)` is actually equal to `Results.success()`. The Result class will end up the correct state either way. Is it worth adding the null check here?

nhachicha: I think it does make sense if the implementation was in Kotlin, since `withResult` will not accept nullable values. We can annotate it though in Java with `Nullable`

cmelchior: Fair point. I'll change it

PR #6788, opened 2020-03-30
Refactor support for Sync to work with MongoDB Realm

cmelchior:
This PR refactors the old Sync API's to work against MongoDB Realm:

Noticeable changes:

* SyncManager has been refactored. All functionality related to connections or auth has moved to RealmApp or RealmAppConfiguration. So only remaining functionality is purely Sync specific. It is now something you access under `RealmApp`. Note: This API is still very much work-in-progress, and it hasn't been aligned with the other SDK's yet. So any input on it would be nice. ObjectStore still only allows one Sync Client to exist, so right now there is a bit of a hacky check to make sure this doesn't happen. We should probably evaluate exactly how to do this.

* I would like to rename SyncManager to SyncService to better fit with the service concept in Stitch, but name is still up in the air. The renaming has been left out of this PR to avoid it growing to big.

* We no longer need callbacks to get the access token. The one provided by the user object is sufficient. It only needs to be valid when creating the sync configuration. After that, the server handles it. This cleans up a lot of code in SyncManager/SyncSession.

* Creating Sync configurations have been reverted to the old pattern of manually creating the config through a builder. This makes it yet again more similar to how local Realms are created. The other SDK's are reverting to this as well.

* Added a `SyncConfiguration.default(user, partionValue)` to make it as easy to create a new config. This API is still very much experiemental.

* Defining proxy settings (headers, authorization header) have moved to RealmAppConfiguration.

### Example of API usage

```
// Configure App
val config = RealmAppConfiguration.Builder("app-id")
  .authorizationHeader("customHeaderName")
  .addCustomHeader("customName", "customValue")
  .appName("MyApp")
  .appVersion("1.0") 
  .build()
val app = RealmApp(config)

// Open Realm
val user = app.login(RealmCredentials.anonymous())
val syncConfig = SyncConfiguration.Builder(user, "partion-value").build()
val defaultConfig = SyncConfiguration.defaultConfig(user, "partion-value")

val realm = Realm.getInstance(synConfig)

// Kotlin
val session: SyncSession = realm.syncSession; 
// Java
SyncSession  session = app.getSyncService().getSession(realmConfig);

// Other Sync functionality
val syncService: SyncService = app.syncService
val sessions: List<SyncSession> = syncService.allSessions // Get all sessions
syncService.refreshConnections() // Reset all sessions
```

### Open Question

* Name of `SyncManager`?  Should it be public at all or should functionality be elsewhere?
* Should ConnectionListeners remain on SyncSession or move to `SyncManager`?

### TODO (in other PR's)

(Copied to #6756 )


cmelchior:
Moving this PR to review. All tests should pass now. Unfortunately, the PR got a lot bigger and a lot messier than I would have liked.

- It is mostly moving things around. Most of the functionality in the old SyncManager has moved to other classes.

- Right now we only allow one `RealmApp/SyncManager` but this restriction is going to be lifted soon. Right now there are some odd checks preventing misuse. Not sure if there is a better way?

- Some tests have been migrated, others have not. I mostly migrated the ones I depended on the most. I'll move the rest in other PR's. The only new tests are in `RealmAppConfigurationTests` and `KotlinSyncedRealmTests`. The rest have only been modified as part of moving to Kotlin.

- Hopefully, I added FIXME's for all known things that will be fixed later.

realm/realm-library/src/objectServer/java/io/realm/RealmApp.java
```diff
@@ -98,14 +85,95 @@ public RealmApp(String appId) {
...
+        // exception if you try to create it twice. This is a really hacky way to do this
+        // Figure out a better API that is always forward compatible
+        synchronized (SyncManager.class) {
+            String appId = config.getAppId();
+            if (CREATED_APP_ID != null && !appId.equals(CREATED_APP_ID)) {
+                throw new IllegalStateException("Only one RealmApp is currently supported. " + CREATED_APP_ID + " already exists.");
+            }
+            CREATED_APP_ID = appId;
+        }
+
```
rorbech, line 98: Yes. I guess we should somehow manage them in Realm, but don't what differentiates multiple apps when we want to support more than one. Is it only _app id_? And are they never a need to clearing them?

rorbech: And is it safe to have multiple instance? Seems like we can create multiple instances as long as it just have the same _app id_.

realm/realm-library/src/objectServer/java/io/realm/RealmAppConfiguration.java
```diff
@@ -219,25 +294,97 @@ public Builder appVersion(String appVersion) {
...
+
+        /**
+         * Adds extra HTTP headers to append to every request to a Realm Object Server.
+         *
+         * @param headers map of (headerName, headerValue) pairs.
+         * @throws IllegalArgumentException If any of the headers provided are illegal.
+         */
+        public Builder addCustomRequestHeaders(@Nullable Map<String, String> headers) {
+            if (headers != null) {
+                customHeaders.putAll(headers);
```
rorbech, line 349: Not really any check for illegal headers

cmelchior: Good point. It mostly isn't there because I would expect the lower layers to validate as needed, but we need to add tests for it. I'll add a FIXME

realm/realm-library/src/objectServer/java/io/realm/SyncConfiguration.java
```diff
@@ -182,6 +196,55 @@ public static RealmConfiguration forRecovery(String canonicalPath, @Nullable byt
...
 
+    /**
+     * FIXME
+     *
+     * @param user
+     * @param partitionValue
+     * @return
+     */
+    @Beta
+    public static SyncConfiguration defaultConfig(RealmUser user, String partitionValue) {
```
rorbech, line 207: Any reason for all these config methods not being on `RealmUser`?

realm/realm-library/src/objectServer/java/io/realm/RealmApp.java
```diff
@@ -49,30 +57,7 @@
...
-                    break;
-                case RECOVERABLE:
-                    RealmLog.info(errorMsg);
-                    break;
-                default:
-                    throw new IllegalArgumentException("Unsupported error category: " + error.getErrorCode().getCategory());
-            }
-        }
-    };
+    private volatile static String CREATED_APP_ID = null;
```
edualonso, line 60: Just a generic thought about this:
I acknowledge the comment written just above this line but using static variables to controlling instantiation logic is kind of ugly. I don't think the class itself should be responsible for knowing whether it has been instantiated or not. Would it not make sense to hide it behind something else?

realm/realm-library/src/objectServer/java/io/realm/internal/SyncObjectServerFacade.java
```diff
@@ -90,16 +74,32 @@ public void realmClosed(RealmConfiguration configuration) {
...
             String urlPrefix = syncConfig.getUrlPrefix();
-            String customAuthorizationHeaderName = SyncManager.getAuthorizationHeaderName(syncConfig.getServerUrl());
-            Map<String, String> customHeaders = SyncManager.getCustomRequestHeaders(syncConfig.getServerUrl());
+            String customAuthorizationHeaderName = app.getConfiguration().getAuthorizationHeaderName();
+            Map<String, String> customHeaders = app.getConfiguration().getCustomRequestHeaders();
+
+            // Temporary work-around for serializing supported bson values
+            BsonValue val = syncConfig.getPartitionValue();
+            String partitionValue = null;
+            if (val.isString()) {
```
rorbech, line 92: Unfortunately I still have not found a nice way to serialize to json without adding json keys using org.bson-library

realm/realm-library/src/syncTestUtils/java/io/realm/TestRealmApp.kt
```diff
@@ -26,7 +26,7 @@ import io.realm.log.LogLevel
  *
  * NOTE: This class must remain in the [io.realm] package in order to work.
  */
-class TestRealmApp(networkTransport: OsJavaNetworkTransport? = null) : RealmApp(createConfiguration()) {
+class TestRealmApp(networkTransport: OsJavaNetworkTransport? = null, customizeConfig: (RealmAppConfiguration.Builder) -> Unit = {}) : RealmApp(createConfiguration()) {
```
rorbech, line 29: `customizeConfig` is not used...but assume it is for future use.

realm/realm-library/src/objectServer/java/io/realm/RealmApp.java
```diff
@@ -98,14 +85,95 @@ public RealmApp(String appId) {
...
+                            dir.getPath()));
+                }
+                syncDir = dir.getPath();
+            } catch (IOException e) {
+                throw new IllegalStateException(e);
+            }
+        } else {
+            syncDir = context.getFilesDir().getPath();
+        }
+
```
edualonso, line 167: We could improve the readability in this method by hiding the computation of `syncDir`, `userAgentBindingInfo` and `appDefinedUserAgent` behind helper methods/classes. These could be tested in isolation making it less error-prone.

realm/realm-library/src/main/java/io/realm/internal/ObjectServerFacade.java
```diff
@@ -66,7 +65,7 @@ public void realmClosed(RealmConfiguration configuration) {
     }
 
     public Object[] getSyncConfigurationOptions(RealmConfiguration config) {
-        return new Object[12];
+        return new Object[14];
```
edualonso, line 68: Is this change intended? Also, avoid using magic numbers.

realm/realm-library/src/main/java/io/realm/internal/OsRealmConfig.java
```diff
@@ -218,6 +218,8 @@ private OsRealmConfig(final RealmConfiguration config,
         String urlPrefix = (String)(syncConfigurationOptions[8]);
         String customAuthorizationHeaderName = (String)(syncConfigurationOptions[9]);
         Byte clientResyncMode = (Byte) syncConfigurationOptions[11];
+        String partionKeyValue = (String) syncConfigurationOptions[12];
```
edualonso, line 221: Is there a typo in the variable name?

```suggestion
        String partitionKeyValue = (String) syncConfigurationOptions[12];
```

realm/realm-library/src/androidTestObjectServer/kotlin/io/realm/util/KotlinTestUtils.kt
```diff
@@ -0,0 +1,34 @@
...
+import io.realm.ObjectServerError
+import org.junit.Assert.assertEquals
+import org.junit.Assert.fail
+
+// Helper methods for improving Kotlin unit tests.
+
+/**
+ * Verify that a specific exception is thrown
+ */
+inline fun <reified T : Exception> expectException(method: () -> Unit) {
```
rorbech, line 13: As already mentioned `org.jetbrains.kotlin:kotlin-test` has an `assertFailsWith` which we could use in Kotlin-land.

realm/realm-library/src/objectServer/java/io/realm/RealmUser.java
```diff
@@ -432,8 +442,9 @@ public RealmMongoDBService getMongoDBService() {
         return null;
     }
 
+    @SuppressFBWarnings("NP_METHOD_PARAMETER_TIGHTENS_ANNOTATION")
```
edualonso, line 445: What does this annotation do?

realm/realm-library/src/objectServer/java/io/realm/SyncConfiguration.java
```diff
@@ -200,29 +263,33 @@ static RealmConfiguration forRecovery(String canonicalPath, @Nullable byte[] enc
         return new RealmConfiguration(null,null, canonicalPath,null, encryptionKey, 0,null, false, OsRealmConfig.Durability.FULL, schemaMediator, null, null, true, null, true, Long.MAX_VALUE);
     }
 
-    static URI resolveServerUrl(URI serverUrl, String userIdentifier) {
+    static URI resolveServerUrl(URI baseUrl, RealmUser user) {
+        // Configure the full path to the Realm
+        String path = String.format("/api/client/v2.0/app/%s/realm-sync", user.getApp().getConfiguration().getAppId());
```
nhachicha, line 268: I think all these endpoints (+ Stitch) should be in the dependencies file (built via `BuildConfig`)

realm/realm-library/src/objectServer/java/io/realm/SyncConfiguration.java
```diff
@@ -411,19 +477,27 @@ public String getUrlPrefix() {
...
         return clientResyncMode;
     }
 
+    /**
+     * Returns the value this Realm is partitioned on. The partition key is a property defined in
+     * MongoDB Realm. All classes with a property with this value will be synchronized to the
+     * Realm.
+     *
+     * @return the value being used by MongoDB Realm to partition the server side MongoDB Database
+     * into smaller Realms that can be synchronized independently.
```
nhachicha, line 490: maybe remove _smaller_ since it's subjective 
```suggestion
     * into Realms that can be synchronized independently.
```

realm/realm-library/src/objectServer/java/io/realm/SyncConfiguration.java
```diff
@@ -456,65 +530,101 @@ public ClientResyncMode getClientResyncMode() {
...
+         */
+        public Builder(RealmUser user, long partitionValue) {
+            this(user, new BsonInt64(partitionValue));
+        }
+
+        /**
+         * Builder used to construct instances of a SyncConfiguration in a fluent manner.
+         *
+         * @param user the user opening the Realm on the server.
+         * @param partitionValue the partition value specifying which subset of data to include in the Realm.
```
nhachicha, line 579: this description is a bit ambiguous, maybe reformulate into something similar to https://github.com/realm/realm-java/pull/6788/files#diff-f19583eb27d992ab7655df81251f244cR489

realm/realm-library/src/objectServer/java/io/realm/SyncConfiguration.java
```diff
@@ -1052,48 +1086,35 @@ public SyncConfiguration build() {
...
-                throw new IllegalStateException("The serverUrl contains a /~/, but the user does not have an identity." +
-                        " Most likely it hasn't been authenticated yet or has been created directly from an" +
-                        " access token. Use a path without /~/.");
-            }
-
             // Set the default Client Resync Mode based on the current type of Realm.
             // Eventually RECOVER_LOCAL_REALM should be the default for all types.
             if (clientResyncMode == null) {
-                clientResyncMode = ClientResyncMode.RECOVER_LOCAL_REALM;
+                clientResyncMode = ClientResyncMode.MANUAL;
```
nhachicha, line 1092: can we add a FIXME if the sync  team is willing to re-enable `RECOVER_LOCAL_REALM`?

realm/realm-library/src/objectServer/java/io/realm/SyncConfiguration.java
```diff
@@ -1052,48 +1086,35 @@ public SyncConfiguration build() {
...
-            // Use the serverUrl + user to create a unique filepath unless it has been explicitly overridden.
-            // <rootDir>/<userIdentifier>/<serverPath>/<serverFileNameOrOverriddenFileName>
-            URI resolvedServerUrl = resolveServerUrl(serverUrl, user.getIdentity());
-            File rootDir = overrideDefaultFolder ? directory : defaultFolder;
-            String realmPathFromRootDir = user.getIdentity() + "/" + getServerPath(resolvedServerUrl);
-            File realmFileDirectory = new File(rootDir, realmPathFromRootDir);
-
-            String realmFileName = overrideDefaultLocalFileName ? fileName : defaultLocalFileName;
+            // Use the serverUrl + user to create a unique filepath.
+            // It is possible for users to change the root directory or realm file name.
```
nhachicha, line 1102: it is possible or not? if yes we need to add a test for it

realm/realm-library/src/objectServer/java/io/realm/internal/SyncObjectServerFacade.java
```diff
@@ -112,10 +112,12 @@ public void realmClosed(RealmConfiguration configuration) {
...
                     customAuthorizationHeaderName,
                     customHeaders,
-                    syncConfig.getClientResyncMode().getNativeValue()
+                    OsRealmConfig.CLIENT_RESYNC_MODE_MANUAL, // Client Resync is no longer supported in v10, but might be re-added later.
+                    partitionValue,
+                    app.getSyncService()
             };
         } else {
-            return new Object[12];
+            return new Object[15];
```
nhachicha, line 120: I think you used 14 elsewhere (since we added two parameters only)

realm/realm-library/src/syncTestUtils/java/io/realm/SyncTestUtils.kt
```diff
@@ -0,0 +1,148 @@
...
+         */
+        fun restoreEnvironmentAfterTest() {
+            // Block until all users are logged out
+            UserFactory.logoutAllUsers()
+
+            // Reset log level
+            RealmLog.setLevel(originalLogLevel)
+            if (BaseRealm.applicationContext != null) {
+                // Realm was already initialized. Reset all internal state
+                // in order to be able fully re-initialize.
```
nhachicha, line 51:
```suggestion
                // in order to be able to fully re-initialize.
```

realm/realm-library/src/androidTestObjectServer/java/io/realm/SessionTests.java
```diff
@@ -149,16 +153,17 @@ public void errorHandler_clientResetReported() {
         looperThread.addTestRealm(realm);
 
         // Trigger error
-        SyncManager.simulateClientReset(SyncManager.getOrCreateSession(config, null));
+        SyncManager syncService = user.getApp().getSyncService();
```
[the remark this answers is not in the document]
cmelchior: After yesterday's design meeting with the other SDK's the consensus landed on just renaming this to `RealmSync` in a similar way to `RealmFunctions`. 

We are probably going to go through another round of naming review when we get closer to official launch.

realm/realm-library/src/objectServer/java/io/realm/RealmApp.java
cmelchior: I modified the check, so only one instance is created. Also improved the error message.

realm/realm-library/src/objectServer/java/io/realm/RealmAppConfiguration.java
cmelchior: Added a FIXME in the tests

realm/realm-library/src/objectServer/java/io/realm/SyncConfiguration.java
cmelchior: Yes and no.... The original design intent was to keep the same pattern for local and synced Realms, so:

```
val localConfig = RealmConfiguration.Builder()....build()
val localRealm = Realm.getInstance(localConfig)

val syncConfig = SyncConfiguration.Builder(user, "partitionValue")....build()
val syncConfig = Realm.getInstance(syncConfig)
```

That doesn't really explain why these helper methods for creating the simple config are here though and if we do this, we should probably also add `RealmConfiguration.defaultConfig()` for symmetry.

Would it make sense to instead have?

```
val user = app.login(RealmCredentials.anonymous())
val syncConfig = user.defaultSyncConfiguration("partitionValue")
```

It might, but we wanted to move people away from having to go through users for configurations, which the above doesn't really do.

For this PR, I would like to keep them where they are...but it is definitely an area of the API we need to keep a close eye one.

realm/realm-library/src/objectServer/java/io/realm/RealmApp.java
cmelchior: Yes, this is super ugly and a hack. But it is one that is going to be removed shortly so this was the smallest amount of code needed to check it. Note, I improved the check and description around these variables. Hopefully, that makes it acceptable as a temp solution.

realm/realm-library/src/objectServer/java/io/realm/RealmApp.java
cmelchior: Yes, I split the class into smaller methods

realm/realm-library/src/main/java/io/realm/internal/ObjectServerFacade.java
cmelchior: Yes, good point. Added a constant with this that is shared.

realm/realm-library/src/objectServer/java/io/realm/RealmUser.java
cmelchior: It disables a FindBugs false positive. You can use this if what FindBugs complains about isn't accurate.

realm/realm-library/src/androidTestObjectServer/java/io/realm/SyncConfigurationTests.kt
```diff
@@ -0,0 +1,263 @@
...
+        val user: RealmUser = createTestUser(app)
+        val builder = SyncConfiguration.Builder(user, DEFAULT_PARTITION)
+        expectException<java.lang.IllegalArgumentException> { builder.errorHandler(TestHelper.getNull())  }
+    }
+
+    @Test
+    fun equals() {
+        val user: RealmUser = createTestUser(app)
+        val config: SyncConfiguration = SyncConfiguration.defaultConfig(user, DEFAULT_PARTITION)
+        Assert.assertTrue(config == config)
```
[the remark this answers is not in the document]
cmelchior: In Kotlin `==` is the same as using `.equals(...)` while `===` is identity.

realm/realm-library/src/main/cpp/io_realm_internal_OsRealmConfig.cpp
```diff
@@ -255,12 +258,11 @@ JNIEXPORT jstring JNICALL Java_io_realm_internal_OsRealmConfig_nativeCreateAndSe
...
         // https://developer.android.com/training/articles/perf-jni.html#faq_FindClass
         static JavaMethod java_error_callback_method(env, sync_manager_class, "notifyErrorHandler",
-                                                     "(Ljava/lang/String;ILjava/lang/String;Ljava/lang/String;)V", true);
-        static JavaMethod java_bind_session_method(env, sync_manager_class, "bindSessionWithConfig",
-                                                   "(Ljava/lang/String;Ljava/lang/String;)Ljava/lang/String;", true);
+                                                     "(Ljava/lang/String;ILjava/lang/String;Ljava/lang/String;)V");
 
         // error handler will be called form the sync client thread
-        auto error_handler = [](std::shared_ptr<SyncSession> session, SyncError error) {
+        auto sync_service_object = env->NewGlobalRef(j_java_sync_service);
```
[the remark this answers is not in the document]
cmelchior: Yes. Good catch. We need to modify the JavaGlobalRef class so it can be passed in lamdas. I'm adding a FIXME for now.

realm/realm-library/src/objectServer/java/io/realm/SyncConfiguration.java
cmelchior: This is being handled by ObjectStore now. So this method can actually be deleted.

realm/realm-library/src/objectServer/java/io/realm/SyncConfiguration.java
cmelchior: I removed the option for changing the realm file name. The Sync root dir is now being defined as part of the RealmAppconfiguration. There should be a test for it there.

cmelchior:
I addressed all raised issues by either fixing them or adding FIXME's in the code. 
Normally this PR would need a few more rounds of review, but we need to unblock external testing  as well, so I'll be merging this now. Then fix remaining issues in other (hopefully smaller) PR's.

PR #6796, opened 2020-04-03
RealmUser API Cleanup

cmelchior:
After a discussion with the other SDK's we agreed on the following:

RealmApp should be considered a user factory instead of "user" in itself, so all methods that required a `currentUser` should be moved to RealmUser, while the methods for joggling between users and login stay on RealmApp. This is a mix of the old Realm approach with everything on SyncUser and the old Stitch approach where everything was on `StitchApp`, but we felt it strikes a good balance between having multiple "active" users, and the current capabilities Stitch has. 

**RealmApp:**
 - Login
 - EmailPasswordAuthProvider
 - CurrentUser
 - AllUSers
 - Everything related to configuring the network

**RealmUser:**
 - ApiKeyAuthProvider
 - linkUser
 - removeUser
 - Push Notifications
 - MongoDB Services
 - Functions
 - Custom Data
 - logOut

Coming in the Sync PR, we also agreed to move back to the old way of creating SyncConfigurations, so it will now become:

```
// New approach and "oldest" approach for synced Ralms
val realm = Realm(SyncConfiguration.Builder(user, "partion-value").build())

// Current approach, which was a compromise to reduce mental overhead
// of the URL weirdness
val realm = Realm(user.createConfiguration("partion-key").build())
```

This also means that `getDefaultConfiguration` and `createConfiguration` will not be added to `RealmUser` (they are not there right now).


cmelchior:
@nhachicha Moving this to "Ready for review" to expedite the process. CI is still not up and running, but I don't expect it to find anything as this PR is just moving things around and not adding new functionality

realm/realm-library/src/objectServer/java/io/realm/RealmUser.java
```diff
@@ -218,42 +225,211 @@ public State getState() {
...
+    public RealmFunctions getFunctions() {
+        return null;
+    }
+
+    /**
+     * FIXME Add support for push notifications. Name of Class and method still TBD.
      */
-    public RealmAsyncTask logOutAsync(RealmApp.Callback callback) {
-        return app.logOutAsync(this, callback);
+    public RealmPushNotifications getFSMPushNotifications() {
```
nhachicha, line 424: I thought we renamed this in another PR to lose the prefix?

realm/realm-library/src/objectServer/java/io/realm/RealmUser.java
```diff
@@ -218,42 +225,211 @@ public State getState() {
...
-     * If the user is already logged out, this method does nothing.
+     * Linking a user with more credentials, mean the user can login either of these credentials.
+     * It also makes it possible to "upgrade" an anonymous user by linking it with e.g.
+     * Email/Password credentials.
+     * <pre>
+     * {@code
+     * // Example
+     * RealmApp app = new RealmApp("app-id")
+     * RealmUser user = app.login(RealmCredentials.anonymous());
+     * app.linkUser(RealmCredentials.emailPassword("email", "password"));
```
edualonso, line 247: Is this correct? Should `linkUser` not be called on `user` instead of `app`?

edualonso: There are some other instances of this one below as well.

realm/realm-library/src/objectServer/java/io/realm/RealmUser.java
```diff
@@ -218,42 +225,211 @@ public State getState() {
...
+            @Override
+            protected RealmUser mapSuccess(Object result) {
+                return RealmUser.this;
+            }
+        });
+        return handleResult(success, error);
+    }
+
+    /**
+     * Removes a users credentials from this device. If the user was currently logged in, they
```
edualonso, line 328:
```suggestion
     * Removes a user's credentials from this device. If the user was currently logged in, they
```

cmelchior:
> Question totally outside the scope of this PR: are there any plans to incorporate Rx versions of the user async operations, e.g. removeUserAsync, logOutAsync and the like? I was thinking it would be pretty cool to have something like public Single<RealmUser> logOut().

That is actually a good idea. There wasn't any immediate plans to add it, but given that we have native RxJava support for other things I could see it making sense.

Can you create an issue about it?

realm/realm-library/src/androidTestObjectServer/kotlin/io/realm/RealmUserTests.kt
```diff
@@ -66,17 +69,193 @@ class RealmUserTests {
...
+    fun logOutAsync_throwsOnNonLooperThread() {
+        val user: RealmUser = app.login(RealmCredentials.anonymous())
+        try {
+            user.logOutAsync { fail() }
+            fail()
+        } catch (ignore: IllegalStateException) {
+        }
+    }
+
+    @Ignore("FIXME: Wait for linkUser support in ObjectStore")
```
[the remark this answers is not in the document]
cmelchior: Yes, there are apparently still issues with linkUser, I need to investigate that further.

realm/realm-library/src/objectServer/java/io/realm/RealmUser.java
```diff
@@ -218,42 +225,211 @@ public State getState() {
...
+     * <p>
+     * Note: It is not possible to link two existing users of MongoDB Realm. The provided credentials
+     * must not have been used by another user.
+     *
+     * @param credentials the credentials to link with the current user.
+     * @param callback callback when user identities has been linked or it failed. The callback will
+     * always happen on the same thread as this method is called on.
+     * @throws IllegalStateException if called from a non-looper thread.
+     */
+    public RealmAsyncTask linkUserAsync(RealmCredentials credentials, RealmApp.Callback<RealmUser> callback) {
```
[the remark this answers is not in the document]
cmelchior: You can and no, we don't have a test for it. I can add it.

realm/realm-library/src/objectServer/java/io/realm/RealmUser.java
cmelchior: I think you are right. I'll rename here.

cmelchior:
Ready for round 2

realm/realm-library/src/objectServer/java/io/realm/RealmUser.java
```diff
@@ -281,7 +281,7 @@ protected RealmUser mapSuccess(Object result) {
      * // Example
      * RealmApp app = new RealmApp("app-id")
      * RealmUser user = app.login(RealmCredentials.anonymous());
-     * app.linkUser(RealmCredentials.emailPassword("email", "password"));
+     * userg.linkUser(RealmCredentials.emailPassword("email", "password"));
```
nhachicha, line 284:
```suggestion
     * user.linkUser(RealmCredentials.emailPassword("email", "password"));
```

realm/realm-library/src/objectServer/java/io/realm/RealmUser.java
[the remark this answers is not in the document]
nhachicha: I believe the test added was for the same credentials type (email) consider adding a test with other types once they're implemented

PR #6806, opened 2020-04-17
Release SNAPSHOT builds from v10

cmelchior:
We should start releasing SNAPSHOT builds from the v10 branch so other teams can more easily consume the merged Stitch/Realm SDK.

This PR changes CI so the `v10` branch is now treated like our other "major" branches, i.e. we run all tests for all architectures on it, and SNAPSHOTs are being released.

The version is `10.0.0-SNAPSHOT` and can be found here once deployed: http://oss.jfrog.org/oss-snapshot-local/io/realm

Howto for using SNAPSHOTs are here: https://github.com/realm/realm-java/blob/master/README.md#using-snapshots

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -6,6 +6,7 @@ def buildSuccess = false
 def mongoDbRealmContainer = null
 def mongoDbRealmCommandServerContainer = null
 def dockerNetworkId = UUID.randomUUID().toString()
+def releaseBranches = ['master', 'next-major', 'v10'] // Branches from which we release SNAPSHOT's
 try {
   node('android') {
     timeout(time: 90, unit: 'MINUTES') {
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -140,7 +140,7 @@ try {
                 }
               }
 
-              if (['master', 'next-major'].contains(env.BRANCH_NAME)) {
+              if (['master', 'next-major', 'v10'].contains(env.BRANCH_NAME)) {
```
rorbech, line 143: Does not look like a common scenario, but always dangerous with things that need to be added multiple places. I get that we could do additional testing on non-published builds, but wouldn't it most often follow each other. Further, could maybe be filtered on generic branch naming conventions, e.g. `['master', 'next-major', 'ojo/']` if not ruining other CI/branching naming conventions.

cmelchior: Good point. We should probably extract this list into a shared one.
Generally, we only rigorously test `master` and `next-major`. Releasing from `v10` should probably be considered a hack we will revert once we can merge the branch to `next-major`

PR #6808, opened 2020-04-20
Fix detecting current branch

cmelchior:
#6806  didn't actually work, because detecting the branch name logic was broken. This should fix it.

See https://ci.eclipse.org/webtools/env-vars.html/


cmelchior:
Verified on CI that the branch name is now correctly detected: https://ci.realm.io/job/realm/job/realm-java/job/PR-6808/1/console

PR #6810, opened 2020-04-22
Add support for Realm Functions

### LINKED ISSUE #6807: Add support for Realm Functions
See [https://docs.google.com/document/d/1iD9ujY76BWVUOdN7rRH4XyPF9hHxx48RcGc-ZaQPyII/edit#heading=h.wt1zjxqe6taa](https://docs.google.com/document/d/1iD9ujY76BWVUOdN7rRH4XyPF9hHxx48RcGc-ZaQPyII/edit#heading=h.wt1zjxqe6taa)

Part of [https://github.com/realm/realm-java/issues/6801](https://github.com/realm/realm-java/issues/6801)

From PR #6810

* [x] Argument/result encoding/decoding support. Defaults to codec registry for primitive types, `BsonValue`s, and `List` and `Map` types
* [x] Backend support
* [x] Tests
* [x] Test of invoking function only authorized to specific user
* [x] Adobt final package/naming scheme
* [ ] Align async API across other MongoDB related APIs (will be done in separate task when set of API's are known)
* [ ] Ignored test due to C++ Parser issues (issue created and tracked in ROS-127)

rorbech:
Functions API with BSON encoding/decoding:

Adds _RealmFunctions_ with support for calling Stitch functions [(legacy api)](https://docs.mongodb.com/stitch-sdks/java/4/com/mongodb/stitch/android/core/StitchAppClient.html#callFunction-java.lang.String-java.util.List-java.lang.Class-) synchronously and asynchronously and with default or explicit codec registry for encoding and decoding objects.
```
public <T> T callFunction(String name, List<?> args, Class<T> resultClass)
public <T> T callFunction(String name, List<?> args, Class<T> resultClass, CodecRegistry codecRegistry)
public <T> RealmAsyncTask callFunctionAsync(String name, List<?> args, Class<T> resultClass, RealmApp.Callback<T> callback)
public <T> RealmAsyncTask callFunctionAsync(String name, List<?> args, Class<T> resultClass, CodecRegistry codecRegistry, RealmApp.Callback<T> callback)
```

The _RealmFunctions_ instance is retrieved from _RealmUser_ with default or explicit codec registry through:
```
public RealmFunctions getFunctions() 
public RealmFunctions getFunctions(CodecRegistry codecRegistry)
```

Arguments are encoded with the 
- [x] Argument/result encoding/decoding support. Defaults to codec registry for primitive types, `BsonValue`s, and `List` and `Map` types
- [x] Backend support
- [x] Tests
- [x] Test of invoking function only authorized to specific user
- [x] Adobt final package/naming scheme
- [ ] Align async API across other MongoDB related APIs
- [ ] Ignored test due to C++ Parser issues

Closes #6807


realm/realm-library/src/testObjectServer/kotlin/io/realm/BsonTest.kt
```diff
@@ -0,0 +1,86 @@
...
+import org.junit.Test
+import kotlin.test.assertFailsWith
+
+class BsonTest {
+
+    /**
+     * Simple test to verify semantics of org.bson JSON encoding and decoding.
+     */
+    @Test
+    fun test_bson_roundtrip() {
```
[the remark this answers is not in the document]
cmelchior: Normally we don't prefix tests with `test` since the `@Test` annotation already convey that

realm/realm-library/build.gradle
```diff
@@ -97,6 +97,9 @@ android {
     }
 
     sourceSets {
+        testObjectServer {
```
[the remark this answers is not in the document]
rorbech: Seems to be executed by `./gradlew check` (at least locally), but test results were not picked up. Added that to `Jenkinsfile`. Will try to inspect CI artifacts.

realm/realm-library/src/androidTestObjectServer/kotlin/io/realm/RealmFunctionTests.kt
```diff
@@ -0,0 +1,74 @@
...
+                    looperThread.testComplete()
+                },
+                BsonInt32(32)
+        )
+    }
+
+
+    
+}
+
```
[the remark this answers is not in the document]
rorbech: Tried configuring AS to ensure proper line feed at end of file.

realm/realm-library/src/objectServer/java/io/realm/RealmFunctions.java
```diff
@@ -15,5 +15,107 @@
...
+
+        String resultString = invoke(name, document.toJson());
+
+        BsonDocument resultDocument = BsonDocument.parse(resultString);
+        // FIXME Guard if no values, etc. when conventions are clarified
+        BsonValue result = resultDocument.values().iterator().next();
+        return result;
+    }
+
+    <T extends BsonValue> T callFunctionTyped(String name, Class<T> clz, BsonValue... args) {
```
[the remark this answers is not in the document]
rorbech: Reworked public interface.

realm/realm-library/src/objectServer/java/io/realm/RealmFunctions.java
```diff
@@ -15,5 +15,107 @@
...
+     * Call a Stitch function synchronously.
+     *
+     * @param name Name of the Stitch function to call.
+     * @param args Arguments to the Stitch function.
+     * @return Result of the Stitch function.
+     *
+     * // FIXME Document possible exceptions
+     * @throws
+     */
+    public BsonValue callFunction(String name, BsonValue... args) {
```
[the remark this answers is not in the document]
rorbech: Intention was to also let people have an API without too much automatism, but have now updated interface to accept Objects doing automatic conversion for primitive types. Added 'ObjectServcerError' 👍

realm/realm-library/src/objectServer/java/io/realm/internal/util/BsonConverter.java
```diff
@@ -0,0 +1,163 @@
...
+        if (value instanceof BsonValue) {
+            return (BsonValue) value;
+        } else if (value instanceof Integer) {
+            return new BsonInt32((Integer) value);
+        } else if (value instanceof Long) {
+            return new BsonInt64((Long) value);
+        } else if (value instanceof String){
+            return new BsonString((String) value);
+        }
+        throw new UnsupportedOperationException("Conversion to BSON value not supported for " + value.getClass().getSimpleName() );
```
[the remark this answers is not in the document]
rorbech: 👍

realm/realm-library/src/testObjectServer/kotlin/io/realm/BsonTest.kt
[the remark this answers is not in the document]
rorbech: Updated naming. This test is actually not testing any Realm specific code, so ignoring it, but left to future programmatic experiments with the BSON api.

realm/realm-library/src/testObjectServer/kotlin/io/realm/BsonTest.kt
```diff
@@ -0,0 +1,86 @@
...
+            BsonConverter.from(Long::class.java, BsonInt32(32))
+        }
+        assertFailsWith<java.lang.UnsupportedOperationException> {
+            BsonConverter.from(Long::class.java, BsonInt32(32))
+        }
+
+        // FIXME Add tests for BsonConverter.bsontype
+    }
+
+}
```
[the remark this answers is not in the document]
rorbech: @cmelchior Is it right now?

realm/realm-library/src/objectServer/java/io/realm/RealmFunctions.java
```diff
@@ -15,5 +15,107 @@
...
+        List<BsonElement> elements = new ArrayList<>(args.length);
+        int i = 0;
+        for (BsonValue arg : args) {
+            // FIXME Bson argument naming
+            elements.add(new BsonElement("arg" + i, arg));
```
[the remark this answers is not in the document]
rorbech: Contract seems to be passing all arguments as one BSONArray with key `arguments`.

realm/realm-library/src/testObjectServer/kotlin/io/realm/BsonTest.kt
```diff
@@ -63,24 +65,46 @@ class BsonTest {
...
+
+        // FIXME Do we actually want to unwrap all of the BsonValues?
+        //assertEquals(oid, BsonConverter.from(ObjectId::class.java, bOid))
 
-        // FIXME Add tests for BsonConverter.bsontype
     }
 
-}
\ No newline at end of file
+}
```
rorbech, line 110: @cmelchior Right now?

realm/realm-library/src/testObjectServer/kotlin/io/realm/BsonTest.kt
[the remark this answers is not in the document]
rorbech: 👍

realm/realm-library/src/testObjectServer/kotlin/io/realm/BsonTest.kt
cmelchior: Yes

realm/realm-library/src/androidTestObjectServer/kotlin/io/realm/RealmFunctionTests.kt
```diff
@@ -0,0 +1,74 @@
...
+    @After
+    fun tearDown() {
+        app.close() 
+    }
+
+    @Test
+    fun test_basic() = looperThread.runBlocking {
+        val functions = anonUser.functions
+        val result1a: BsonValue = functions.callFunction("sum", BsonInt32(32))
```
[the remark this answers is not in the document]
rorbech: Tried to follow the principle. I guess it should be combined with https://junit.org/junit4/javadoc/4.12/org/junit/rules/ErrorCollector.html to collect all error across enum values and report at the end.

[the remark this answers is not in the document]
cmelchior: Ah, we never used that one. We just let the test fail after the first error, but actually sounds like a good idea. But doesn't have to be part of this PR.

realm/realm-library/src/androidTestObjectServer/kotlin/io/realm/RealmFunctionsTests.kt
```diff
@@ -116,36 +162,205 @@ class RealmFunctionsTests {
...
+    @Test
+    fun instanceCodecRegistry() {
+        val input = Dog("PojoFido")
+        val functionsWithCodecRegistry = anonUser.getFunctions(pojoRegistry)
+        assertEquals(input, functionsWithCodecRegistry.callFunction(FIRST_ARG_FUNCTION, listOf(input), Dog::class.java))
+    }
+
+    @Test
+    fun unknownFunction() {
+        assertFailsWith<ObjectServerError> {
```
cmelchior, line 269: We have a better `KotlinTestUtils.expectErrorCode` for this. The name might need a revisit after moving to `assertFailsWith` though.

realm/realm-library/src/androidTestObjectServer/kotlin/io/realm/RealmFunctionsTests.kt
```diff
@@ -116,36 +162,205 @@ class RealmFunctionsTests {
...
+
+    @Test
+    fun asyncUnknownFunction() = looperThread.runBlocking {
         val input = Dog("PojoFido")
+        functions.callFunctionAsync("unknown", listOf(input), Dog::class.java, pojoRegistry) { result ->
+            if (result.isSuccess) {
+                fail()
+            } else  {
+                // FIXME How verify exact error. NativeErrorIntValue? Or error message?
+                assertTrue(result.error is ObjectServerError)
```
cmelchior, line 282: See above. Normally we assert on a specific the error code, so `result.error.errorCode`

realm/realm-library/src/androidTestObjectServer/kotlin/io/realm/RealmFunctionsTests.kt
```diff
@@ -116,36 +162,205 @@ class RealmFunctionsTests {
...
+    fun callFunction_sum() {
+        val numbers = listOf(1, 2, 3, 4)
+        assertEquals(10, functions.callFunction("sum", numbers, Integer::class.java).toInt())
+    }
 
-        val actual: Dog = functions.invoke(input, Dog::class.java, pojoRegistry)
+    @Test
+    fun callFunction_remoteError() {
+        assertFailsWith<ObjectServerError> {
+            // FIXME Do we need to assert more about the error
```
cmelchior, line 306: Yes, check the error code. See above.

realm/realm-library/src/androidTestObjectServer/kotlin/io/realm/RealmFunctionTests.kt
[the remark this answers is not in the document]
rorbech: No, somehow also breaks initialization of the Androidx runner, so I will be for some other time. Seems like there is a fix for the BSON parser already, so everything should hopefully pass after upgrading.

realm/realm-library/src/androidTestObjectServer/kotlin/io/realm/RealmFunctionsTests.kt
```diff
@@ -116,36 +162,205 @@ class RealmFunctionsTests {
...
+    fun callFunction_null() {
+        assertTrue(functions.callFunction("null", emptyList<Any>(), BsonNull::class.java).isNull)
+    }
+
+    @Test
+    fun callFunction_empty() {
+        assertEquals(BsonType.UNDEFINED, functions.callFunction("empty", emptyList<Any>(), BsonUndefined::class.java).bsonType)
+    }
+
+    // FIXME Do we need to test connectivity issues
```
cmelchior, line 321: No, we normally don't do that. The general network transport is tested elsewhere and as long as there are tests that ensure an exception/error handler can be correctly called we assume that connection errors will also be reported correctly.

realm/realm-library/src/androidTestObjectServer/kotlin/io/realm/RealmFunctionsTests.kt
```diff
@@ -116,36 +162,205 @@ class RealmFunctionsTests {
...
+
+    @Test
+    fun defaultCodecRegistry() {
+        // TODO Maybe we should test that setting configuration specific would propagate all the way
```
cmelchior, line 325: The `TestRealmApp` actually allows you to define a custom builder for this exact case: 
```
TestRealmApp(networkTransport: OsJavaNetworkTransport? = null, customizeConfig: (RealmAppConfiguration.Builder) -> Unit = {})
```

realm/realm-library/src/androidTestObjectServer/kotlin/io/realm/RealmFunctionsTests.kt
```diff
@@ -116,36 +162,205 @@ class RealmFunctionsTests {
...
-        assertEquals(input, actual)
+    @Test
+    fun customCodecRegistry() {
+        val configCodecRegistry = CodecRegistries.fromCodecs(StringCodec())
+        val customCodecRegistryFunctions = anonUser.getFunctions(configCodecRegistry)
+        assertEquals(configCodecRegistry, customCodecRegistryFunctions.defaultCodecRegistry)
+    }
+
+    @Test
+    // FIXME JNI Parsing crashes
```
cmelchior: Is this still true?

realm/realm-library/src/androidTestObjectServer/kotlin/io/realm/RealmFunctionsTests.kt
```diff
@@ -116,36 +162,205 @@ class RealmFunctionsTests {
...
+        val valueOk = listOf(1, 2, 3)
+        assertEquals(valueOk, functions.callFunction(FIRST_ARG_FUNCTION, listOf(valueOk), List::class.java))
+
+        // Actual failing call
+        // {"value": [[{"$numberInt": "1"}, true, {"$numberInt": "3"}]]}
+        val value = listOf(1, true, 3)
+        assertEquals(value, functions.callFunction(FIRST_ARG_FUNCTION, listOf(value), List::class.java))
+    }
+    @Test
+    // FIXME JNI Parsing only returns part of array
```
cmelchior: Same, this was fixed, right?

realm/realm-library/src/main/cpp/io_realm_RealmFunctions.cpp
```diff
@@ -18,19 +18,40 @@
...
 
-// FIXME This is just a basic round trip test for passing bson back and forth. Proper implementation
-//  will come with actual Function implementation.
-JNIEXPORT jstring JNICALL Java_io_realm_RealmFunctions_nativeCallFunction
-        (JNIEnv* env, jclass, jstring j_args) {
+static std::function<jobject(JNIEnv*, Optional<bson::Bson> )> response_mapper = [](JNIEnv* env, Optional<bson::Bson> response) {
+    if (response) {
+        return JniBsonProtocol::bson_to_jstring(env, *response);
+    } else {
+        // FIXME How to raise errors here
```
cmelchior, line 32: If you throw a `std::logic_error` or `std:runtime_error` they will be caught by the `CATCH()` macro in the JNI layer and converted into a Java exception. 

The logic for how the conversion is defined here: https://github.com/realm/realm-java/blob/v10/realm/realm-library/src/main/cpp/util.cpp#L156

If those exceptions do not match the error you want to return to Java, we normally create a custom exception class in C++ that can be caught explicitly and converted.

realm/realm-library/src/main/cpp/io_realm_RealmFunctions.cpp
```diff
@@ -18,19 +18,40 @@
...
+JNIEXPORT void JNICALL
+Java_io_realm_RealmFunctions_nativeCallFunction(JNIEnv* env, jclass , jlong j_app_ptr, jlong j_user_ptr, jstring j_name,
+                                                jstring j_args_json , jobject j_callback) {
     try {
-        bson::Bson bson = JniBsonProtocol::jstring_to_bson(env, j_args);
-        return JniBsonProtocol::bson_to_jstring(env, bson);
+        auto app = *reinterpret_cast<std::shared_ptr<App>*>(j_app_ptr);
+        auto user = *reinterpret_cast<std::shared_ptr<SyncUser>*>(j_user_ptr);
+
+        std::function<void(Optional<bson::Bson>, Optional<app::AppError>)> callback;
```
cmelchior, line 44: I assume the reason for doing this in two lines is for readability? But there is a slight performance penalty as you are creating a dummy callback that is being thrown away straight away.

realm/realm-library/src/objectServer/java/io/realm/RealmFunctions.java
```diff
@@ -17,30 +17,173 @@
...
+import java.util.concurrent.atomic.AtomicReference;
+
+import io.realm.internal.Util;
 import io.realm.internal.jni.JniBsonProtocol;
+import io.realm.internal.jni.OsJNIResultCallback;
+import io.realm.internal.objectstore.OsJavaNetworkTransport;
+
+/**
+ * A <i>Realm functions<i> manager to call MongoDB Realm functions.
+ *
```
cmelchior, line 30: We need to manually add `<p>` tags for empty lines to make the generated Javadoc render correctly. This is also missing in a bunch of other places in this PR.

```suggestion
 * <p>
```

realm/realm-library/src/objectServer/java/io/realm/RealmFunctions.java
```diff
@@ -17,30 +17,173 @@
...
+ * Arguments and results are encoded/decoded with the <i>Realm functions'</i> codec registry either
+ * inherited from the {@link RealmAppConfiguration#getDefaultCodecRegistry()} or set explicitly
+ * when creating the <i>Realm functions</i>-instance through {@link RealmUser#getFunctions(CodecRegistry)}
+ * or through the individual calls to {@link #callFunction(String, List, Class, CodecRegistry)}.
+ */
+// TODO Timeout is currently handled uniformly through OkHttpNetworkTransport configured through RealmAppConfig
+// FIXME Application wide invocation; like StitchAppClient appClient = Stitch.getDefaultAppClient();
```
cmelchior, line 37: Not 100% sure what these FIXME's mean or if they are still relevant?

realm/realm-library/src/objectServer/java/io/realm/RealmFunctions.java
```diff
@@ -17,30 +17,173 @@
...
+// FIXME Encoder/Decoder specific variants
+// FIXME Service?
+public class RealmFunctions {
+
+    // FIXME Review memory allocation
```
cmelchior, line 42: What is the concern here?

realm/realm-library/src/objectServer/java/io/realm/RealmFunctions.java
```diff
@@ -17,30 +17,173 @@
...
+ * Arguments and results are encoded/decoded with the <i>Realm functions'</i> codec registry either
+ * inherited from the {@link RealmAppConfiguration#getDefaultCodecRegistry()} or set explicitly
+ * when creating the <i>Realm functions</i>-instance through {@link RealmUser#getFunctions(CodecRegistry)}
+ * or through the individual calls to {@link #callFunction(String, List, Class, CodecRegistry)}.
+ */
+// TODO Timeout is currently handled uniformly through OkHttpNetworkTransport configured through RealmAppConfig
+// FIXME Application wide invocation; like StitchAppClient appClient = Stitch.getDefaultAppClient();
+// FIXME Encoder/Decoder specific variants
+// FIXME Service?
+public class RealmFunctions {
```
cmelchior, line 40: We should probably also expose `getApp()` and `getUser()` functions from this class. This is useful if the class is passed around by itself. We are doing something similar in `ApiKeyAuth`. This would also require tests.

realm/realm-library/src/androidTestObjectServer/kotlin/io/realm/RealmFunctionsTests.kt
```diff
@@ -16,28 +16,64 @@
...
+                        PojoCodecProvider.builder()
+                                .register(Dog::class.java)
+                                .build()
+                )
+        )
+    }
 
     @Before
     fun setup() {
-        Realm.init(InstrumentationRegistry.getInstrumentation().targetContext)
```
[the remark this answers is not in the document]
rorbech: Previously I did not have it, as I did not use have a `TestRealmApp`, so I am piggybacking on the hidden one now. I forgot it in the first place and forgetting that this was the root of my missing JNI symbols, so I guess some pattern for avoiding this pitfall in the future would be nice. 

I guess we can re-adopt to pattern if updating `TestRealmApp`.

realm/realm-library/src/androidTestObjectServer/kotlin/io/realm/RealmAppConfigurationTests.kt
```diff
@@ -152,4 +155,29 @@ class RealmAppConfigurationTests {
...
+        assertFailsWith<IllegalArgumentException> {
+            builder.codecRegistry(TestHelper.getNull())
+        }
+    }
+
+    @Test
+    fun defaultFunctionsCodecRegistry() {
+        val config: RealmAppConfiguration = RealmAppConfiguration.Builder("app-id")
+                .build()
+        assertEquals(RealmAppConfiguration.Builder.DEFAULT_BSON_CODEC_REGISTRY, config.defaultCodecRegistry)
```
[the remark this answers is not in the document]
rorbech: I guess it makes sense to allow people to retrieve to build on top of it. Exposing it in `RealmAppConfiguration`.

realm/realm-library/src/objectServer/java/io/realm/RealmFunctions.java
rorbech: 👍

realm/realm-library/src/main/cpp/io_realm_RealmFunctions.cpp
rorbech: 👍

realm/realm-library/src/objectServer/java/io/realm/RealmFunctions.java
rorbech: Initially I though it was possible to execute a function without a user. Apparently it was due to implicitly resolving a _current user_. Since we want to avoid this implicit resolving of users this is no longer relevant.

realm/realm-library/src/objectServer/java/io/realm/RealmFunctions.java
rorbech: From before I had an overview of how we managed the underlying C++ objects. Just a reminder for me to check up on it.

realm/realm-library/src/objectServer/java/io/realm/RealmFunctions.java
rorbech: 👍

realm/realm-library/src/androidTestObjectServer/kotlin/io/realm/RealmFunctionsTests.kt
```diff
@@ -16,28 +16,64 @@
...
         app = TestRealmApp()
-        functions = RealmFunctions(app.configuration.defaultCodecRegistry)
+        admin = ServerAdmin()
+        anonUser = app.login(RealmCredentials.anonymous())
```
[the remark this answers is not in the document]
rorbech: Execution is controlled by [Authorization Expressions](https://docs.mongodb.com/stitch/functions/define-a-function/#id3), so technically it has not necessarily anything to do with user, but could be any scriptable check. I added a function that requires a specific user email address, and added a test showing that the anonymous user cannot execute it while the _authorized user_ can.

realm/realm-library/src/androidTestObjectServer/kotlin/io/realm/RealmFunctionsTests.kt
rorbech: The configuration options for settings the default codec registry is already covered in the `RealmAppConfigurationTests` and we test that we get the same default as the configuration. So, as the customization options is not fully implemented, I will consider this customization update to be part of another PR.

realm/realm-library/src/androidTestObjectServer/kotlin/io/realm/RealmFunctionsTests.kt
rorbech: Yes, there are still issues with the C++ parser.

realm/realm-library/src/androidTestObjectServer/kotlin/io/realm/RealmFunctionsTests.kt
rorbech: I have separated the current C++ parser issues into separate ignored test cases and added a todo element to follow up on it.

realm/realm-library/src/main/cpp/io_realm_RealmFunctions.cpp
rorbech: Seem like we should never reach here, as we would not reach here if we do not have a valid Bson response. I added a `std::logic_error` to propagate it nicely should it ever happen.

realm/realm-library/src/objectServer/java/io/realm/InternalFunctions.java
```diff
@@ -0,0 +1,85 @@
...
+        this(user, user.getApp().getConfiguration().getDefaultCodecRegistry());
+    }
+
+    InternalFunctions(RealmUser user, CodecRegistry codecRegistry) {
+        super(codecRegistry);
+        this.user = user;
+    }
+
+    @Override
+    public RealmApp getApp() {
```
cmelchior, line 50: Should live directly in Functions I think?

realm/realm-library/src/objectServer/java/io/realm/InternalFunctions.java
```diff
@@ -0,0 +1,85 @@
...
+        return user.getApp();
+    }
+
+    @Override
+    public RealmUser getUser() {
```
cmelchior, line 55: Same

realm/realm-library/src/objectServer/java/io/realm/InternalFunctions.java
```diff
@@ -0,0 +1,85 @@
...
+import io.realm.internal.objectstore.OsJavaNetworkTransport;
+import io.realm.log.RealmLog;
+import io.realm.mongodb.functions.Functions;
+
+
+/**
+ * Internal implementation of Functions invoking the actual OS function in the context of the
+ * {@link RealmUser}/{@link RealmApp}.
+ */
+class InternalFunctions extends Functions {
```
cmelchior, line 36: I guess we should come up with a common name if we are to implement this pattern elsewhere.

Nitpick: Not sure how I feel about this compared to `FunctionsImpl`. Although `FunctionsImpl` is normally used when interfaces, it does feel more descriptive in my mind. But I don't have strong reasons and it is entirely subjective.

realm/realm-library/src/objectServer/java/io/realm/RealmApp.java
```diff
@@ -369,6 +371,25 @@ public RealmSync getSync() {
         return syncManager;
     }
 
+    /**
+     * Returns a <i>Realm Functions</i> manager for invoking MongoDB Realm Functions.
+     * <p>
+     * This will use the associated app's default codec registry to encode and decode arguments and
+     * results.
+     */
+    public InternalFunctions getFunctions(RealmUser user) {
```
cmelchior, line 380:
```suggestion
    public Functions getFunctions(RealmUser user) {
```

realm/realm-library/src/objectServer/java/io/realm/RealmApp.java
```diff
@@ -369,6 +371,25 @@ public RealmSync getSync() {
...
+        return new InternalFunctions(user);
+    }
+
+    /**
+     * Returns a <i>Realm Functions</i> manager for invoking MongoDB Realm Functions with custom
+     * codec registry for encoding and decoding arguments and results.
+     */
+    public InternalFunctions getFunctions(RealmUser user, CodecRegistry codecRegistry) {
```
cmelchior, line 388:
```suggestion
    public Functions getFunctions(RealmUser user, CodecRegistry codecRegistry) {
```

realm/realm-library/src/objectServer/java/io/realm/InternalFunctions.java
rorbech: I kept it here to have all App/User related separately, but as none of it uses internals it can be in Functions. Will move it.

realm/realm-library/src/objectServer/java/io/realm/InternalFunctions.java
rorbech: Ok. Will go for `FunctionsImpl`

realm/realm-library/src/objectServer/java/io/realm/InternalFunctions.java
rorbech: Moved

PR #6811, opened 2020-04-22
Fix Kotlin Extensions Base tests

cmelchior:
@nhachicha Not entirely sure how this happened as your datatype PR was green. Anyway. This should fix the build failure on `v10`


cmelchior:
Actually, I know why this wasn't caught on CI. Our PR's only run tests against the objectServer variant. So this PR will not actually detect anything. 

I ran the tests locally and the pass. So I'll just merge this.

PR #6815, opened 2020-04-27
Update ObjectServerExample

cmelchior:
This PR
- The ObjectServer example has been updated to work. This required updating to AndroidX


cmelchior:
Ready for review. This PR bumps the ObjectServer so it works with latest v10 branch. 
It still contains some "bugs", like being forced to use `waitForInitialRemoteData` because we only support ObjectId primary keys, so it isn't possible for two devices to create the same counter object offline.

bmunkholm:
I guess it makes sense to rename the example now.

examples/objectServerExample/gradle.properties
```diff
@@ -0,0 +1,2 @@
+# FIXME: Required as long as we depend on PlayServices for RemoteMongDB API's
+android.useAndroidX=true
```
rorbech, line 2: The issue is just that we do not want to force it from Realm, but I guess it should be OK to have test apps using it?

If you really want to keep test apps pre-AndroidX, we could consider adding the AndroidX as an `api` dependency in Realm until eliminated again. This way dependent projects where we added did not need the explicit dependency and would fail when we removed it again. This way we would not forget to remove it again.

cmelchior: Since the rest of the Android ecosystem have more or less moved to AndroidX since it is a requirement for a lot of Google libraries, I think we should probably migrate all of our examples to that as well. Maybe besides one, just to make sure we don't depend on it.

But we should do that in another PR

Issue #6814, opened 2020-04-27
Consider letting SyncSession.downloadAllServerChanges() also refresh the Realm


cmelchior:
Consider the following test code:

```
        val realm = Realm.getInstance(config2)
        realm.syncSession.downloadAllServerChanges()
        assertEquals(10, realm.where<SyncDog>().count())
        assertEquals(1, realm.where<SyncPerson>().count())
```

By reading it you would assume it would work, but it doesn't. The reason is that `downloadAllServerChanges` run on the Sync Client worker thread and when it completes the caller thread has not been updated yet.

Instead, you manually need to add a `realm.refresh()` after calling `download()`.

We should consider doing that automatically. Realms normally using this API are not looper threads, so we don't need to be particularly worried about callbacks.

The only annoying thing is that the Session and Realm are largely uncoupled. The session is e.g. not thread confined, while the Realm is. So we need a clever way of finding any open Realms on the current thread, and there might be multiple (think typed and Dynamic Realm).

But maybe this is just too confusing...something to ponder.

PR #6820, opened 2020-04-30
Upgrade to Sync 10.0.0-alpha.12

cmelchior:
Upgrade to latest Sync release + ObjectStore v10 branch. 

This will, in particular, fix the SNI bug towards realm-dev.mongodb.com

----

- This PR will also fix CI, by restricting it to run on the only working device right now
- Also fixes annotation processor tests. Which for some reason have become broken as well 🤔 
- Also uncovered that the PlayServices has a strict requirement on AndroidX, which broke a bunch of things. It should only be applied to the objectServer variant.
- Add support for the new `/location` endpoint used during login to setup connection endpoints.

### DIFF
--- a/realm/realm-library/src/androidTestObjectServer/kotlin/io/realm/KotlinSyncedRealmTests.kt
+++ b/realm/realm-library/src/androidTestObjectServer/kotlin/io/realm/KotlinSyncedRealmTests.kt
@@ -68,12 +68,11 @@ class KotlinSyncedRealmTests { // FIXME: Rename to SyncedRealmTests once remaini
     }
 
     // Smoke test for Sync
-    @Ignore("Dev Mode doesn't work fully yet on the server")
     @Test
     fun roundTripObjectsNotInServerSchemaObject() {
         // User 1 creates an object an uploads it to MongoDB Realm
         val user1: RealmUser = createNewUser()
-        val config1: SyncConfiguration = createDefaultConfig(user1, partitionValue)
+        val config1: SyncConfiguration = createCustomConfig(user1, partitionValue)
         realm = Realm.getInstance(config1)
         realm.executeTransaction {
             for (i in 1..10) {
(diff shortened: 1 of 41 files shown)


cmelchior:
There are still 2 failing unit tests. These are being investigated here: https://github.com/realm/realm-core/issues/3707, but I still think we should merge this if possible since we need to keep up with the Sync/ObjectStore changes 

Since the rest of the tests pass, I don't expect the failing unit tests to have a widespread impact.

In particular this PR also "fixes" CI so our tests now run on the only working Android device.

cmelchior:
@edualonso I discovered some more issues after you approved. It looks like Annotation Tests became broken at some point and the play service dependency was accidentally applied to all flavors. This was a problem because it depends on AndroidX, so it broke all our examples.

That the PlayService API depends on AndroidX is a problem by itself, and we need to take that into account when evaluating it.

realm/realm-library/src/androidTestObjectServer/kotlin/io/realm/KotlinSyncedRealmTests.kt
```diff
@@ -122,6 +122,7 @@ class KotlinSyncedRealmTests { // FIXME: Rename to SyncedRealmTests once remaini
         val config2 = createDefaultConfig(user2, partitionValue)
         realm = Realm.getInstance(config2)
         realm.syncSession.downloadAllServerChanges()
+        realm.refresh()
```
rorbech, line 125: Is this normal way of doing it. If so, seems like something others could also forget and it could be worth considering adding an argument to `downloadAllServerChanges` to force people to think of it.

cmelchior: Yes, the problem is that `downloadAllServerChanges` block until the background thread doing the download is done, but the Realm on the caller thread is not automatically updated. I created this issue: https://github.com/realm/realm-java/issues/6814 to track it. Because it is not entirely trivial.

PR #6821, opened 2020-04-30
Scheme version/migration configuration clarifications

rorbech:
Scheme version/migration configuration clarifications:

Added additional version/migration configuration test to clarify existing semantics:
- `RealmMigration` will never be called unless specifying a `schemeVersion` too
- Increasing the `schemeVersion` will trigger a `RealmMigration` if present or will otherwise silently bump the version.

Proposing deprecation of old `schemeVersion(long)` and `migration(RealmMigration)` in favour of `schemeVersion(long, RealmMigration)` to highlight semantically connection.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,6 +15,7 @@ ENV ANDROID_NDK /opt/android-ndk
 ENV PATH ${PATH}:${ANDROID_HOME}/tools:${ANDROID_HOME}/tools/bin:${ANDROID_HOME}/platform-tools
 ENV PATH ${PATH}:${NDK_HOME}
 ENV NDK_CCACHE /usr/bin/ccache
+ENV CCACHE_CPP2 yes
 
 # The 32 bit binaries because aapt requires it
 # `file` is need by the script that creates NDK toolchains
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -9,7 +9,7 @@ def dockerNetworkId = UUID.randomUUID().toString()
 def releaseBranches = ['master', 'next-major', 'v10'] // Branches from which we release SNAPSHOT's
 def currentBranch = env.CHANGE_BRANCH
 try {
-  node('android') {
+  node('docker-cph-01') { // FIXME: Only working Slave
     timeout(time: 90, unit: 'MINUTES') {
       // Allocate a custom workspace to avoid having % in the path (it breaks ld)
       ws('/tmp/realm-java') {
--- a/dependencies.list
+++ b/dependencies.list
@@ -1,7 +1,7 @@
 # Realm Sync release used by Realm Java (This includes Realm Core)
 # https://github.com/realm/realm-sync/releases
-REALM_SYNC_VERSION=10.0.0-alpha.8
-REALM_SYNC_SHA256=2cdccdd43f1cb0c0d7297e9616824efa34be62deefd311d722ecee88d8a6e44e
+REALM_SYNC_VERSION=10.0.0-alpha.12
+REALM_SYNC_SHA256=d9ab40f7b73b4438c811adfbbb22ed8da1461d922e77bb46f555f49ae1921748
 
 # Object Server Release used by Integration tests. Installed using NPM.
 # Use `npm view realm-object-server versions` to get a list of available versions.
@@ -9,7 +9,7 @@ REALM_OBJECT_SERVER_VERSION=3.28.2
 
 # Version of MongoDB Realm used by integration tests
 # See https://github.com/realm/ci/packages/147854 for available versions
-MONGODB_REALM_SERVER_VERSION=2020-04-26
+MONGODB_REALM_SERVER_VERSION=2020-04-30
 
 # Common Android settings across projects
 GRADLE_BUILD_TOOLS=3.6.1
@@ -24,4 +24,3 @@ GRADLE_BINTRAY_PLUGIN=1.8.4
 
 # Bson dependency version
 BSON_DEPENDENCY_VERSION=3.12.1
-
--- a/examples/build.gradle
+++ b/examples/build.gradle
@@ -1,7 +1,7 @@
 def projectDependencies = new Properties()
 projectDependencies.load(new FileInputStream("${rootDir}/../dependencies.list"))
-project.ext.sdkVersion = 27
-project.ext.minSdkVersion = 21 // FIXME: Should be 16. Figure out how to enable MultiDex for ObjectServer tests
+project.ext.sdkVersion = 29
+project.ext.minSdkVersion = 16
 project.ext.buildTools = projectDependencies.get("ANDROID_BUILD_TOOLS")
 
 // Don't cache SNAPSHOT (changing) dependencies.
--- a/realm/realm-annotations-processor/build.gradle
+++ b/realm/realm-annotations-processor/build.gradle
@@ -19,7 +19,7 @@ dependencies {
     testImplementation files("${System.properties['java.home']}/../lib/tools.jar") // This is needed otherwise compile-testing won't be able to find it
     testImplementation group:'junit', name:'junit', version:'4.12'
     testImplementation group:'com.google.testing.compile', name:'compile-testing', version:'0.6'
-    testImplementation files(file("${System.env.ANDROID_HOME}/platforms/android-27/android.jar"))
+    testImplementation files(file("${System.env.ANDROID_HOME}/platforms/android-29/android.jar"))
 }
 
 // for Ant filter
--- a/realm/realm-library/build.gradle
+++ b/realm/realm-library/build.gradle
@@ -211,6 +211,8 @@ dependencies {
     implementation('io.reactivex.rxjava2:rxandroid:2.1.1') {
         exclude group: 'io.reactivex.rxjava2', module: 'rxjava'
     }
+    // FIXME: Attempt to find a way to remove this dependency
+    objectServerImplementation "com.google.android.gms:play-services-tasks:17.0.2"      // added to support mongo client's asynchronous nature without breaking Stitch's API
 
     // TODO: investigate why we can't use the latest multidex version
     // check baseDebugAndroidTestRuntimeClasspath and objectServerDebugAndroidTestRuntimeClasspath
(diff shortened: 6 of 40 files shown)


realm/realm-library/src/main/java/io/realm/RealmConfiguration.java
```diff
@@ -594,11 +610,46 @@ public Builder schemaVersion(long schemaVersion) {
...
          * Sets the {@link io.realm.RealmMigration} to be run if a migration is needed. If this migration fails to
          * upgrade the on-disc schema to the runtime schema, a {@link io.realm.exceptions.RealmMigrationNeededException}
          * will be thrown.
+         * <p>
+         * FIXME Also triggered on version update even though no schema change
+         * FIXME Not triggered if no version number is specified.
+         *
+         * @deprecated Use {@link #schemaVersion(long, RealmMigration)} instead.
          */
+        @Deprecated
```
cmelchior, line 652: Normally this would be correct, but since is this being merged into a branch that will be released as a major (breaking) version. It is fine to just remove this outright. But we do need to add a line about in the CHANGELOG.md Breaking Changes section.

realm/realm-library/src/main/java/io/realm/RealmConfiguration.java
```diff
@@ -594,11 +610,46 @@ public Builder schemaVersion(long schemaVersion) {
             return this;
         }
 
+        /**
+         * Update the schema version of the Realm and apply the provided migration.
+         * <p>
+         * The schema version must be equal to or higher than the schema version of the existing
```
cmelchior, line 616: What happens if the version is less?

realm/realm-library/src/main/java/io/realm/RealmConfiguration.java
```diff
@@ -468,6 +468,8 @@ boolean isSyncConfiguration() {
      * RealmConfiguration.Builder used to construct instances of a RealmConfiguration in a fluent manner.
      */
     public static class Builder {
+        private static long DEFAULT_SCHEMA_VERSION = 0;
```
cmelchior, line 471: 👏  We should have named constants for all magic numbers.

realm/realm-library/src/androidTest/kotlin/io/realm/RealmMigrationTestsKotlin.kt
```diff
@@ -0,0 +1,163 @@
...
+import org.junit.After
+import org.junit.Before
+import org.junit.Rule
+import org.junit.Test
+import org.junit.runner.RunWith
+import kotlin.test.assertFailsWith
+import kotlin.test.assertTrue
+
+@RunWith(AndroidJUnit4::class)
+class RealmMigrationTestsKotlin {
```
cmelchior, line 35: If you don't want to migrate the existing RealmMigration tests to Kotlin in this PR, add a FIXME here that that the existing Java tests should be moved here and the file renamed to loose the Kotlin suffix once done.

realm/realm-library/src/androidTest/kotlin/io/realm/RealmMigrationTestsKotlin.kt
```diff
@@ -0,0 +1,163 @@
...
+
+    @get:Rule
+    val configFactory = TestRealmConfigurationFactory()
+
+    @Before
```
cmelchior, line 40: No need for these setup/teardown methods if they are just empty.

realm/realm-library/src/androidTest/kotlin/io/realm/RealmMigrationTestsKotlin.kt
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+    // Test that
+    // - supplying updated version and migration for schema change we actually trigger migration
+    //   with out errors
+    @Test
+    fun standardMigration() {
+        val configV1: RealmConfiguration = configFactory.createConfigurationBuilder()
+                .schema(StringAndInt::class.java)
+                .schemaVersion(1).build()
```
cmelchior, line 66: build() should come on the next line

realm/realm-library/src/androidTest/kotlin/io/realm/RealmMigrationTestsKotlin.kt
```diff
@@ -0,0 +1,163 @@
...
+            this.oldVersion = oldVersion
+            this.newVersion = newVersion
+        }
+    }
+
+    // Test that
+    // - supplying updated version and migration for schema change we actually trigger migration
+    //   with out errors
+    @Test
+    fun standardMigration() {
```
cmelchior, line 63: I really hope we have an existing test for this? 😄

realm/realm-library/src/androidTest/kotlin/io/realm/RealmMigrationTestsKotlin.kt
```diff
@@ -0,0 +1,163 @@
...
+                .schema(StringAndInt::class.java, StringOnly::class.java)
+                .migration(migration)
+                .build()
+        Realm.getInstance(configV2).close()
+
+        assertTrue(migration.triggered)
+    }
+
+    // Test that
+    // - we never reach migration f version is not specified, not even when triggering
```
cmelchior, line 87:
```suggestion
    // - we never reach migration if version is not specified, not even when triggering
```

realm/realm-library/src/androidTest/kotlin/io/realm/RealmMigrationTestsKotlin.kt
```diff
@@ -0,0 +1,163 @@
...
+    //   RealmMigrationNeededException by scheme updates
+    @Test
+    fun noMigrationWithoutVersion() {
```
cmelchior, line 90: This test can be removed if we remove `migration()` outright.

realm/realm-library/src/main/java/io/realm/RealmConfiguration.java
```diff
@@ -594,11 +610,46 @@ public Builder schemaVersion(long schemaVersion) {
...
+         * If the schema has changed without increasing the version or failing to resolve schema
+         * migration through the supplied {@link RealmMigration},, Realm will throw a
+         * {@link io.realm.exceptions.RealmMigrationNeededException}.
+         *
+         * @throws IllegalArgumentException If {@code schemaVersion} is negative or {@code migration} is null.
+         *
+         * @see #schemaVersion(long)
+         * @see #migration(RealmMigration)
+         */
+        public Builder schemaVersion(long schemaVersion, RealmMigration migration) {
```
cmelchior, line 630: Did you consider alternatives to this name? Would just `schema(long schemaVersion, RealmMigration migration)` be more accurate? I'm not sure. Maybe it makes sense making "schemaVersion" the primary concept and the migration just a result of that.

cmelchior: Also. Since we are in the process of making breaking changes. Would it make sense to prepare the API for https://github.com/realm/realm-java/issues/2530 ?

Should we already now rename `RealmMigration` to perhaps `RealmManualMigration` so in the future we can add `ReamAutomaticMigration` (just throwing ideas at the wall, we could also postpone and decide later. E.g. we can introduce both of these interfaces in a new update and just let RealmMigration be an alias for RealmManualMigration until the next major release).

realm/realm-library/src/main/java/io/realm/RealmConfiguration.java
rorbech: The conventional `IllegalArgumentException` with message like `Provided schema version 1 is less than last set version 3`. I will update the docs.

Don't know if it would make sense to differentiate errors a bit. Also if changing scheme without bumping version number the migration is not triggered and we just emit the usual `RealmMigrationNeededException`.

realm/realm-library/src/androidTest/kotlin/io/realm/RealmMigrationTestsKotlin.kt
rorbech: 👍

realm/realm-library/src/androidTest/kotlin/io/realm/RealmMigrationTestsKotlin.kt
rorbech: 👍

realm/realm-library/src/main/java/io/realm/RealmConfiguration.java
rorbech: Since the migration would never be fully automated for all cases, I guess a `RealmAutomaticMigration` would also need some customization points, so maybe better off separating the automatism into a helper and allowing people to take advantage of this in other scenarios. Thus, keeping the `RealmMigration` interface for customization but supplying a `DefaultRealmMigration` implementation for simple cases where the automatic migration helper is assumed to handle everything (simple additions/deletions, etc).

cmelchior: I think I need to see some code to understand that 😄

rorbech: > Did you consider alternatives to this name? Would just `schema(long schemaVersion, RealmMigration migration)` be more accurate? I'm not sure. Maybe it makes sense making "schemaVersion" the primary concept and the migration just a result of that.

Shorter names are good, but maybe some keeping it to align with other platforms. If not bound by that I would go for `version` but maybe that is too overloaded?

cmelchior: JS has separate: `schema`, `schemaVersion` and `migration` blocks
Cocao has: `schemaVersion` and `migrationBlock`
.NET has `SchemaVersion` and `MigrationCallback`

That probably hints to us to keep `schemaVersion(version, migration)`. Thoughts?

rorbech: Yep. Lets just keep it, to avoid too may different names.

realm/realm-library/src/main/java/io/realm/RealmConfiguration.java
rorbech: 👍

realm/realm-library/src/androidTest/kotlin/io/realm/RealmMigrationTestsKotlin.kt
rorbech: 👍

PR #6826, opened 2020-05-01
Merge Stitch and Realm SDKs - 2: add task framework from Stitch

edualonso:
Added task solution from Stitch to support GMS Tasks.


realm/realm-library/src/objectServer/java/io/realm/internal/common/Callback.java
```diff
@@ -0,0 +1,23 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.realm.internal.common;
+
+import javax.annotation.Nonnull;
+
+public interface Callback<T, U> {
+    void onComplete(@Nonnull OperationResult<T, U> result);
```
cmelchior, line 22: Instead of using `@Nonnull` you should use the package level annotations. Ideally non-null is the default and @Nullable the exception.

PR #6827, opened 2020-05-01
Merge Stitch and Realm SDKs - 3: collection count

edualonso:
First steps towards adding Stitch's RemoteMongoCollection API to Realm. First operation is count.


edualonso:
Will bundle count and insert in the same PR.

PR #6830, opened 2020-05-04
Fix broken tests on CI and expose bug in Core

cmelchior:
Fixes our own tests and add a test for https://github.com/realm/realm-core/issues/3707


realm/realm-library/src/androidTest/java/io/realm/internal/OsSharedRealmTests.java
```diff
@@ -178,6 +178,33 @@ public void onSchemaChanged() {
...
+                listenerCalled.set(true);
+            }
+        });
+
+        // Change schema using another Realm
+        // Classes not starting with class_ were treated differently by Sync
+        OsSharedRealm bgRealm = OsSharedRealm.getInstance(config, OsSharedRealm.VersionID.LIVE);
+        bgRealm.beginTransaction();
+        bgRealm.createTable("NewTable");
+        bgRealm.commitTransaction();
```
edualonso, line 199: I assume the callback above will be immediately executed after committing the transaction, right? Otherwise, we could face problems with the test finishing before the callback kicks in.

cmelchior: Yes, it does.

PR #6831, opened 2020-05-04
JNI BSON Protocol

rorbech:
Passing BSON to JNI by wrapping it in a document with key `value`. 

Java side: `OSJNIBsonProtocol`:
- `public static String encode(BsonValue bsonValue)`
- `public static BsonValue decode(String string)`

JNI side:
- `realm::bson::Bson jstring_to_bson(JNIEnv* env, jstring arg);`
- `jstring bson_to_jstring(JNIEnv* env, realm::bson::Bson bson);`

- 
Conversion does currently not handle the below types as mapping from primitive types is not obvious: 
```
                BsonType.DOCUMENT,
                BsonType.UNDEFINED,
                BsonType.DATE_TIME,
                BsonType.NULL,
                BsonType.REGULAR_EXPRESSION,
                BsonType.SYMBOL,
                BsonType.DB_POINTER,
                BsonType.JAVASCRIPT,
                BsonType.JAVASCRIPT_WITH_SCOPE,
                BsonType.TIMESTAMP,
                BsonType.END_OF_DOCUMENT,
                BsonType.MIN_KEY,
                BsonType.MAX_KEY
```
- [x] Verify format
- [x] Full implementation
- [x] Converter tests (no JNI interaction)
- [x] Basic JNI roundtrip (not covering all BSON types)

### DIFF
--- a/realm/realm-library/src/androidTestObjectServer/kotlin/io/realm/RealmFunctionsTest.kt
+++ b/realm/realm-library/src/androidTestObjectServer/kotlin/io/realm/RealmFunctionsTest.kt
@@ -19,16 +19,20 @@ package io.realm
 import androidx.test.platform.app.InstrumentationRegistry
 import io.realm.internal.util.BsonConverter
 import org.bson.*
+import org.junit.Before
 import org.junit.Test
 import java.util.stream.Collectors
 import kotlin.test.assertEquals
 
 class RealmFunctionsTest {
 
-    @Test
-    operator fun invoke() {
+    @Before
+    fun setup() {
         Realm.init(InstrumentationRegistry.getInstrumentation().targetContext)
+    }
 
+    @Test
+    fun jniBsonOnlyRoundtrip() {
         val functions = RealmFunctions()
         val i32 = 42
         val i64 = 42L
--- a/realm/realm-library/src/androidTestObjectServer/kotlin/io/realm/RealmFunctionsTest.kt
+++ b/realm/realm-library/src/androidTestObjectServer/kotlin/io/realm/RealmFunctionsTest.kt
@@ -17,11 +17,10 @@
 package io.realm
 
 import androidx.test.platform.app.InstrumentationRegistry
-import org.bson.BsonArray
-import org.bson.BsonInt32
-import org.bson.BsonInt64
-import org.bson.BsonString
+import io.realm.internal.util.BsonConverter
+import org.bson.*
 import org.junit.Test
+import java.util.stream.Collectors
 import kotlin.test.assertEquals
 
 class RealmFunctionsTest {
@@ -35,12 +34,13 @@ class RealmFunctionsTest {
         val i64 = 42L
         val s = "Realm"
 
-        // Return 64 bit integer
-//        assertEquals(i32, functions.invoke(BsonInt32(i32)).asInt32().value)
+        assertEquals(i32, functions.invoke(BsonInt32(i32)).asInt32().value)
         assertEquals(i64, functions.invoke(BsonInt64(i64)).asInt64().value)
         assertEquals(s, functions.invoke(BsonString(s)).asString().value)
-        // Crashes
-//        val invoke = functions.invoke(BsonArray(listOf(BsonInt32(i32), BsonInt64(i64), BsonString(s))))
+        
+        val values = listOf<Any>(BsonInt32(i32), BsonInt64(i64), BsonString(s))
+        val invoke: BsonValue = functions.invoke(BsonConverter.to(values))
+        assertEquals(values, invoke.asArray().values)
     }
 
 }
--- a/realm/realm-library/src/androidTestObjectServer/kotlin/io/realm/RealmFunctionsTest.kt
+++ b/realm/realm-library/src/androidTestObjectServer/kotlin/io/realm/RealmFunctionsTest.kt
@@ -1,0 +1,46 @@
+/*
+ * Copyright 2020 Realm Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.realm
+
+import androidx.test.platform.app.InstrumentationRegistry
+import org.bson.BsonArray
+import org.bson.BsonInt32
+import org.bson.BsonInt64
+import org.bson.BsonString
+import org.junit.Test
+import kotlin.test.assertEquals
+
+class RealmFunctionsTest {
+
+    @Test
+    operator fun invoke() {
+        Realm.init(InstrumentationRegistry.getInstrumentation().targetContext)
+
+        val functions = RealmFunctions()
+        val i32 = 42
+        val i64 = 42L
+        val s = "Realm"
+
+        // Return 64 bit integer
+//        assertEquals(i32, functions.invoke(BsonInt32(i32)).asInt32().value)
+        assertEquals(i64, functions.invoke(BsonInt64(i64)).asInt64().value)
+        assertEquals(s, functions.invoke(BsonString(s)).asString().value)
+        // Crashes
+//        val invoke = functions.invoke(BsonArray(listOf(BsonInt32(i32), BsonInt64(i64), BsonString(s))))
+    }
+
+}
--- a/realm/realm-library/src/main/cpp/util_sync.cpp
+++ b/realm/realm-library/src/main/cpp/util_sync.cpp
@@ -13,20 +13,25 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package io.realm;
 
-import org.bson.BsonValue;
+#include "util.hpp"
+#include "util_sync.hpp"
 
-import io.realm.internal.jni.OSJNIBsonProtocol;
+// Must match OSJNIBsonProtocol.VALUE
+static const std::string VALUE("value");
 
-public class RealmFunctions {
+using namespace realm::bson;
 
-    // FIXME Prelimiry implementation to be able to test passing BsonValues through JNI
-    BsonValue invoke(BsonValue arg) {
-        String response = nativeCallFunction(OSJNIBsonProtocol.encode(arg));
-        return OSJNIBsonProtocol.decode(response);
-    }
-
-    private static native String nativeCallFunction(String arg);
-
+Bson jstring_to_bson(JNIEnv* env, jstring arg) {
+    JStringAccessor args_json(env, arg);
+    BsonDocument document(parse(args_json));
+    return document[VALUE];
 }
+
+jstring bson_to_jstring(JNIEnv* env, Bson bson) {
+    BsonDocument document{{VALUE, bson}};
+    std::stringstream buffer;
+    buffer << document;
+    std::string r = buffer.str();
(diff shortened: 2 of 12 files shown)


realm/realm-library/src/main/cpp/util_sync.cpp
```diff
@@ -0,0 +1,37 @@
...
+
+// Must match OSJNIBsonProtocol.VALUE
+static const std::string VALUE("value");
+
+using namespace realm::bson;
+
+Bson jstring_to_bson(JNIEnv* env, jstring arg) {
+    JStringAccessor args_json(env, arg);
+    BsonDocument document(parse(args_json));
+    return document[VALUE];
```
edualonso, line 28: Is this the best approach? Why not return the `BsonDocument` instance instead? You could have an additional method that provides a `Bson` too. There are multiple methods receiving a document in `remote_mongo_collection.hpp`, e.g.

`void count(const bson::BsonDocument& filter_bson ...`

rorbech: The reason for wrapping them is that I cannot find a proper way to parse non-document values, which is also argument/response to a number of the C++ calls. The `BsonDocument` is however itself a `BsonValue` so there should be no problem embedding these in the convention (besides the problem that current C++ parser yields different types than the original document and I cannot pass `BsonArray`...but I am working on resolving those issues)

rorbech: The idea is also to provide a wrapper for easier wrapping/unwrapping correctly typed values, as in https://github.com/realm/realm-java/pull/6810/files#diff-84db7612944e18ff25865af1cb6daa17

realm/realm-library/src/androidTestObjectServer/kotlin/io/realm/RealmFunctionsTest.kt
```diff
@@ -0,0 +1,46 @@
...
+import org.bson.*
+import org.junit.Test
+import java.util.stream.Collectors
+import kotlin.test.assertEquals
+
+class RealmFunctionsTest {
+
+    @Test
+    operator fun invoke() {
+        Realm.init(InstrumentationRegistry.getInstrumentation().targetContext)
```
cmelchior, line 30: Normally this would be in the setup method:

```
@Before
public void setUp() {
  Realm.init(InstrumentationRegistry.getInstrumentation().targetContext)
}
```

realm/realm-library/src/androidTestObjectServer/kotlin/io/realm/RealmFunctionsTest.kt
```diff
@@ -0,0 +1,46 @@
...
+
+        val functions = RealmFunctions()
+        val i32 = 42
```
cmelchior, line 33: In other tests of this kind we normally either create a custom enum or switch on an existing one. That way it becomes really clear if you are missing tests for a specific type. In this case there is already a `BsonType` enum, right?

realm/realm-library/src/objectServer/java/io/realm/internal/jni/OSJNIBsonProtocol.java
```diff
@@ -0,0 +1,48 @@
...
+import org.bson.json.JsonMode;
+import org.bson.json.JsonWriterSettings;
+
+/**
+ * Protocol for passing {@link BsonValue}s to JNI.
+ *
+ * For now this just encapsulated the BSON value in a document with key {@value VALUE}. This
+ * overcomes the shortcoming of {@code org.bson.JsonWrite} not being able to serialize single values.
+ */
+public class OSJNIBsonProtocol {
```
cmelchior, line 30: Normally we use the `Os` name to signal that the class is wrapping an existing C++ class, so using that name here is a bit confusing. I would probably just rename to `JniBsonProtocol`

realm/realm-library/src/objectServer/java/io/realm/internal/util/BsonConverter.java
```diff
@@ -0,0 +1,224 @@
...
+    /**
+     *
+     *
+     * @param clz
+     * @param value
+     * @param <T>
+     * @return
+     *
+     */
+    // FIXME Would we rather return null? Would maybe make it cleaner to use with
```
cmelchior, line 126: Isn't there a specific null type in Bson? If yes, we should disallow null IMO, otherwise we should allow it.

realm/realm-library/src/objectServer/java/io/realm/internal/util/BsonConverter.java
```diff
@@ -0,0 +1,224 @@
...
+    //  functions.callFunctionTyped("sum", BsonString::class.java, "Realm")?.value
+    //  but would also silently hide if expectations of type is wrong
+    private static <T extends BsonValue> T fromToBson(Class<T> clz, BsonValue value) {
```
cmelchior, line 129: Not sure I understand the naming of this method?

realm/realm-library/src/testObjectServer/kotlin/io/realm/BsonTest.kt
```diff
@@ -0,0 +1,131 @@
...
+import org.junit.Ignore
+import org.junit.Test
+import kotlin.test.assertFailsWith
+
+class BsonTest {
+
+    /**
+     * Simple test to verify semantics of org.bson JSON encoding and decoding.
+     */
+    @Ignore("Only for Bson API evaluation, not testing Realm functionality")
```
cmelchior, line 32: Why ignore it?

realm/realm-library/src/androidTestObjectServer/kotlin/io/realm/RealmFunctionsTest.kt
rorbech: Yep. You commented earlier on another PR, and just before pushing I got the point :)

realm/realm-library/src/objectServer/java/io/realm/internal/util/BsonConverter.java
rorbech: Yes, there is a `BSONNull`. Removing FIXME.

realm/realm-library/src/objectServer/java/io/realm/internal/util/BsonConverter.java
rorbech: No. Leftover from before the interface settled. Removing it completely, as `from` now handles it.

realm/realm-library/src/objectServer/java/io/realm/internal/jni/OSJNIBsonProtocol.java
rorbech: 👍

realm/realm-library/src/androidTestObjectServer/kotlin/io/realm/RealmFunctionsTest.kt
rorbech: 👍

realm/realm-library/src/testObjectServer/kotlin/io/realm/BsonTest.kt
rorbech: I wasn't even sure if it ends up representing the way we use their API, so did not really see a reason for having it. But I can included it.

cmelchior: I was just wondering. I didn't look closely at the implementation. If it isn't going to be used it should probably just be removed instead.

realm/realm-library/src/main/cpp/util_sync.hpp
```diff
@@ -0,0 +1,26 @@
...
+ * limitations under the License.
+ */
+
+#ifndef REALM_UTIL_SYNC_HPP
+#define REALM_UTIL_SYNC_HPP
+
+#include <jni.h>
+#include <util/bson/bson.hpp>
+
+realm::bson::Bson jstring_to_bson(JNIEnv* env, jstring arg);
```
cmelchior, line 23: These functions are currently free functions. I would probably put them inside a class. `BsonUtil` perhaps.

realm/realm-library/src/objectServer/java/io/realm/internal/jni/JniBsonProtocol.java
```diff
@@ -0,0 +1,48 @@
...
+
+import org.bson.BsonDocument;
+import org.bson.BsonValue;
+import org.bson.json.JsonMode;
+import org.bson.json.JsonWriterSettings;
+
+/**
+ * Protocol for passing {@link BsonValue}s to JNI.
+ *
+ * For now this just encapsulated the BSON value in a document with key {@value VALUE}. This
```
cmelchior, line 27: Good with this kind of explanation 👏

realm/realm-library/src/testObjectServer/kotlin/io/realm/BsonTest.kt
```diff
@@ -0,0 +1,175 @@
...
+        assertFailsWith<RuntimeException> {
+            roundtrip.getInt64("arg1");
+        }
+    }
+
+    /**
+     * Simple test of type conversion between native Java object types and BSON types.
+     */
+    @Test
+    fun bsonConversion() {
```
cmelchior, line 70: This test seems to cover a lot, both to/from/exceptions. Isn't it possible to split it apart. I can see why to/from are in the same test, but the exception could be moved to its own test

realm/realm-library/src/objectServer/java/io/realm/RealmFunctions.java
```diff
@@ -15,5 +15,18 @@
  */
 package io.realm;
 
-class RealmFunctions {
+import org.bson.BsonValue;
+
+import io.realm.internal.jni.JniBsonProtocol;
+
+public class RealmFunctions {
```
cmelchior, line 22: Add FIXME for the Javadoc

realm/realm-library/src/androidTestObjectServer/kotlin/io/realm/RealmFunctionsTest.kt
```diff
@@ -0,0 +1,50 @@
...
+
+class RealmFunctionsTest {
+
+    @Before
+    fun setup() {
+        Realm.init(InstrumentationRegistry.getInstrumentation().targetContext)
+    }
+
+    @Test
+    fun jniBsonOnlyRoundtrip() {
```
cmelchior, line 35: This test is currently a bit weird. 

Normally we would test all cases through the public API, e.g. all the `BsonTest` cases would be expressed in terms of the public API.

This is a bit different though as the helper class is supposed to be used elsewhere. So, in this case, it is probably acceptable to just smoke test the public API that the serialization works.

rorbech: Yep. This is a spinout from #6810, as we suddently needed the BSON passing across multiple PRs. As we did not have a mechanism for doing JNI methods for test only, I had to put a round trip somewhere in the actual code. It will be superseeded by tests in #6810. Should of course have noted that.

cmelchior: Ok 👍

realm/realm-library/src/main/cpp/util_sync.hpp
rorbech: Ok. Will take it into account for later updates.

realm/realm-library/src/objectServer/java/io/realm/RealmFunctions.java
rorbech: Same a for the test. This was only included to have some place for doing a round trip test to JNI. This will be superseeded by #6810.

realm/realm-library/src/testObjectServer/kotlin/io/realm/BsonTest.kt
rorbech: Will separate on future updates.

PR #6837, opened 2020-05-05
Add ability to use CodecRegistry for passing objects as BSON to JNI

rorbech:
This add support for passing a `CodecRegistry` to the `JniBsonProtocol` enabling support for using MongoDB codecs for converting values. For example using the `PojoCodecProvider` to send and construct custom java objects as in https://github.com/realm/realm-java/compare/v10...cr/bson-jni-interaction?expand=1#diff-1af7924ad8b59c18ae9d822f88479d3dR143. 

Adding the below methos to `JniBsonProtocol`:
```
public static <T> String encode(T value, CodecRegistry registry)
public static <T> T decode(String string, Class<T> clz, CodecRegistry registry)
```

The `org.bson`'s `ValueCodecProvider`, `BsonValueCodecProvider` and `IterableCodecProvider` can be combined to offer similar convert to the manually implemented `BsonConverter` methods, but cannot, as far as I can see, be used to do conversion without serializing.  

- [ ] Settle API
- [ ] Settle on default CodecRegistry as substitution for `BsonConverter`
- [ ] Test

### DIFF
--- a/realm/realm-library/src/objectServer/java/io/realm/RealmFunctions.java
+++ b/realm/realm-library/src/objectServer/java/io/realm/RealmFunctions.java
@@ -16,15 +16,25 @@
 package io.realm;
 
 import org.bson.BsonValue;
+import org.bson.codecs.configuration.CodecRegistry;
 
 import io.realm.internal.jni.JniBsonProtocol;
 
-public class RealmFunctions {
+// FIXME This class is only a placeholder for JNI round trip as  until actual RealmFunctions
+//  implementation supersedes it.
+class RealmFunctions {
 
     // FIXME Prelimiry implementation to be able to test passing BsonValues through JNI
     BsonValue invoke(BsonValue arg) {
         String response = nativeCallFunction(JniBsonProtocol.encode(arg));
         return JniBsonProtocol.decode(response);
+    }
+
+    // FIXME Prelimiry implementation to be able to test passing BsonValues through JNI
+    <T> T invoke(Object arg, Class<T> resultClass, CodecRegistry registry) {
+        String a = JniBsonProtocol.encode(arg, registry);
+        String s = nativeCallFunction(a);
+        return JniBsonProtocol.decode(s, resultClass, registry);
     }
 
     private static native String nativeCallFunction(String arg);
--- a/realm/realm-library/src/objectServer/java/io/realm/internal/jni/JniBsonProtocol.java
+++ b/realm/realm-library/src/objectServer/java/io/realm/internal/jni/JniBsonProtocol.java
@@ -18,8 +18,19 @@ package io.realm.internal.jni;
 
 import org.bson.BsonDocument;
 import org.bson.BsonValue;
+import org.bson.codecs.BsonDocumentCodec;
+import org.bson.codecs.Decoder;
+import org.bson.codecs.DecoderContext;
+import org.bson.codecs.Encoder;
+import org.bson.codecs.EncoderContext;
+import org.bson.codecs.configuration.CodecRegistry;
 import org.bson.json.JsonMode;
+import org.bson.json.JsonReader;
+import org.bson.json.JsonWriter;
 import org.bson.json.JsonWriterSettings;
+
+import java.io.StringReader;
+import java.io.StringWriter;
 
 /**
  * Protocol for passing {@link BsonValue}s to JNI.
@@ -43,6 +54,32 @@ public class JniBsonProtocol {
     public static BsonValue decode(String string) {
         BsonDocument document = BsonDocument.parse(string);
         return document.get(VALUE);
+    }
+
+    public static <T> String encode(T value, CodecRegistry registry) {
+        return encode(value, (Encoder<T>)registry.get(value.getClass()));
+    }
+    private static <T> String encode(T value, Encoder<T> encoder) {
+        StringWriter stringWriter = new StringWriter();
+        JsonWriter jsonWriter = new JsonWriter(stringWriter, writerSettings);
+        jsonWriter.writeStartDocument();
+        jsonWriter.writeName(VALUE);
+        encoder.encode(jsonWriter, value, EncoderContext.builder().build());
+        jsonWriter.writeEndDocument();
+        return stringWriter.toString();
+    }
+
+    public static <T> T decode(String string, Class<T> clz, CodecRegistry registry) {
+        return decode(string, registry.get(clz));
+    }
+    private static <T> T decode(String string, Decoder<T> decoder) {
+        StringReader stringReader = new StringReader(string);
+        JsonReader jsonReader = new JsonReader(stringReader);
+        jsonReader.readStartDocument();
+        jsonReader.readName(VALUE);
+        T value = decoder.decode(jsonReader, DecoderContext.builder().build());
+        jsonReader.readEndDocument();
+        return value;
     }
 
 }
--- a/realm/realm-library/src/androidTestObjectServer/kotlin/io/realm/RealmFunctionsTest.kt
+++ b/realm/realm-library/src/androidTestObjectServer/kotlin/io/realm/RealmFunctionsTest.kt
@@ -17,11 +17,21 @@
 package io.realm
 
 import androidx.test.platform.app.InstrumentationRegistry
+import io.realm.internal.jni.JniBsonProtocol
 import io.realm.internal.util.BsonConverter
 import org.bson.*
+import org.bson.codecs.BsonValueCodecProvider
+import org.bson.codecs.IterableCodecProvider
+import org.bson.codecs.StringCodec
+import org.bson.codecs.ValueCodecProvider
+import org.bson.codecs.configuration.CodecRegistries
+import org.bson.codecs.pojo.PojoCodecProvider
+import org.bson.types.Decimal128
+import org.bson.types.ObjectId
 import org.junit.Before
 import org.junit.Test
-import java.util.stream.Collectors
+import java.time.Instant
+import java.util.*
 import kotlin.test.assertEquals
 
 class RealmFunctionsTest {
@@ -31,6 +41,7 @@ class RealmFunctionsTest {
         Realm.init(InstrumentationRegistry.getInstrumentation().targetContext)
     }
 
+    // Test of BSON JNI round trip until superseded with actual public api tests are added.
     @Test
     fun jniBsonOnlyRoundtrip() {
         val functions = RealmFunctions()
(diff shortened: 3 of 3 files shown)


cmelchior:
Looking at https://docs.mongodb.com/stitch-sdks/java/4/com/mongodb/stitch/android/core/services/StitchServiceClient.html

It seems the old API could be created with default registry, and the call function could then be invoked with a specific Decoder.

In our API think that would translate to:

```
val app = RealmApp("my-id")
val user = app.login(RealmCredentials.emailPassword("email", "password"))

// Default registry
val functions1 = user.getFunctions()

// With custom Registry
val registry = createCustomRegistry()
val functions2 = user.getFunctions(customRegistry)

// Custom decoder
Dog dog = functions2.callFunction(Dog::class.java, listOf, DogDecoder())
```

Unless you had something else in mind?

rorbech:
> Looking at https://docs.mongodb.com/stitch-sdks/java/4/com/mongodb/stitch/android/core/services/StitchServiceClient.html
> 
> It seems the old API could be created with default registry, and the call function could then be invoked with a specific Decoder.
> 
> In our API think that would translate to:
> 
> ```
> val app = RealmApp("my-id")
> val user = app.login(RealmCredentials.emailPassword("email", "password"))
> 
> // Default registry
> val functions1 = user.getFunctions()
> 
> // With custom Registry
> val registry = createCustomRegistry()
> val functions2 = user.getFunctions(customRegistry)
> 
> // Custom decoder
> Dog dog = functions2.callFunction(Dog::class.java, listOf(), DogDecoder())
> ```
> 
> Unless you had something else in mind?

No, I think that makes sense. Also adding the option to the `RealmAppConfiguration.Builder` to set a `RealmApp`-wide one,  and allowing override options granularly down the hierarchy.

realm/realm-library/src/objectServer/java/io/realm/RealmFunctions.java
```diff
@@ -15,16 +15,30 @@
...
+class RealmFunctions {
+
+    private CodecRegistry codecRegistry;
+
+    RealmFunctions(CodecRegistry codecRegistry) {
+        this.codecRegistry = codecRegistry;
+    }
+
+    // FIXME Prelimiry implementation to be able to test passing BsonValues through JNI
+    <T> T invoke(Object arg, Class<T> resultClass) {
```
edualonso, line 33: Rather than deleting these when the functions are ready, I would actually put them in their own `JniBsonProtocolTest` file. I think these two functions are super useful in and of themselves to make sure we never brake the conversion.

realm/realm-library/src/androidTestObjectServer/kotlin/io/realm/RealmFunctionsTests.kt
```diff
@@ -0,0 +1,150 @@
...
+class RealmFunctionsTests {
+
+    private lateinit var app: TestRealmApp
+    private lateinit var functions : RealmFunctions
+
+    @Before
+    fun setup() {
+        Realm.init(InstrumentationRegistry.getInstrumentation().targetContext)
+        app = TestRealmApp()
+        functions = RealmFunctions(app.configuration.codecRegistry)
```
cmelchior, line 40: The functions should be created independently but come from them the user `user.getFunctions()` which I guess becomes `user.functions` in Kotlin. We should make the RealmFunctions constructor package-private if possible.

cmelchior: The functions need to be tied to a user. This isn't happening in ObjectStore right now, so we need to fix it there as well and pass in the user as a constructor argument from Java

realm/realm-library/src/androidTestObjectServer/kotlin/io/realm/RealmFunctionsTests.kt
```diff
@@ -0,0 +1,150 @@
...
+                    assertTypedEcho(java.lang.Float(1.4), java.lang.Float::class.java)
+                    assertTypedEcho(java.lang.Double(1.4), java.lang.Double::class.java)
+                    assertTypedEcho(BsonDouble(1.4), BsonDouble::class.java)
+                }
+                BsonType.STRING -> {
+                    assertTypedEcho("Realm", String::class.java)
+                    assertTypedEcho(BsonString("Realm"), BsonString::class.java)
+                }
+                BsonType.ARRAY -> {
+                    // FIXME Fails in C++ parsing when boolean values are added...needs investigation
```
cmelchior, line 68: Seems to have been fixed in OS

realm/realm-library/src/androidTestObjectServer/kotlin/io/realm/RealmFunctionsTests.kt
```diff
@@ -0,0 +1,150 @@
...
+                BsonType.END_OF_DOCUMENT,
+                BsonType.MIN_KEY,
+                BsonType.MAX_KEY -> {
+                    // No conversion is implemented for these types yet
+                }
+            }
+        }
+    }
+
+    private fun <T: Any> assertTypedEcho(value: T, returnClass: Class<T>) : T {
```
cmelchior, line 124: minor: Not sure I understand the Echo suffix. Am I missing some commonly known convention here?

realm/realm-library/src/androidTestObjectServer/kotlin/io/realm/RealmFunctionsTests.kt
```diff
@@ -0,0 +1,150 @@
...
+    // Test of BSON JNI round trip until superseded with actual public api tests are added.
+    @Test
+    fun jniRoundTripForDefaultCodecRegistry() {
+        val i32 = 42
+        val i64 = 42L
+
+        for (type in BsonType.values()) {
+            when (type) {
+                BsonType.DOUBLE -> {
+                    assertTypedEcho(java.lang.Float(1.4), java.lang.Float::class.java)
```
cmelchior, line 59: What about the primitive types? This is the boxed types if I remember correctly? Same for the other tests

realm/realm-library/src/androidTestObjectServer/kotlin/io/realm/RealmFunctionsTests.kt
```diff
@@ -0,0 +1,150 @@
...
+                    //val listValues = listOf<Any>(true, i32, i64)
+                    val listValues = listOf<Any>(i32, i64)
+                    assertTypedEcho(listValues, List::class.java)
+                }
+                BsonType.BINARY -> {
+                    val value = byteArrayOf(1, 2, 3)
+                    val actual = functions.invoke(value, ByteArray::class.java)
+                    assertEquals(value.toList(), actual.toList())
+                    // FIXME Does not seem to preserve type
+                    // assertTypedEcho(BsonBinary(UUID.randomUUID()), BsonBinary::class.java)
```
cmelchior, line 79: UUID is a custom type. Maybe because a CodeRegistry is missing?

realm/realm-library/src/objectServer/java/io/realm/RealmAppConfiguration.java
```diff
@@ -175,10 +184,25 @@ public File getSyncRootDirectory() {
         return syncRootDir;
     }
 
+    // FIXME Doc
+    public CodecRegistry getCodecRegistry() { return codecRegistry; }
```
cmelchior, line 188: Should probably be `getDefaultCodecRegistry`

realm/realm-library/src/objectServer/java/io/realm/RealmFunctions.java
```diff
@@ -15,16 +15,30 @@
  */
 package io.realm;
 
-import org.bson.BsonValue;
+import org.bson.codecs.configuration.CodecRegistry;
 
 import io.realm.internal.jni.JniBsonProtocol;
 
-public class RealmFunctions {
+// FIXME This class is only a placeholder for JNI round trip as  until actual RealmFunctions
```
cmelchior, line 22: It is fine to keep this public as long as there is a Javadoc stating it is still work-in-progress. The `v10` branch is considered under development anyway, so merging "half-finshed" work should be okay as long as it is marked as such and doesn't break existing functionality.

realm/realm-library/src/objectServer/java/io/realm/internal/jni/JniBsonProtocol.java
```diff
@@ -35,14 +44,30 @@
                 .outputMode(JsonMode.EXTENDED)
                 .build();
 
-    public static String encode(BsonValue bsonValue) {
-        BsonDocument document = new BsonDocument(VALUE, bsonValue);
-        return document.toJson(writerSettings);
+    public static <T> String encode(T value, CodecRegistry registry) {
+        return encode(value, (Encoder<T>)registry.get(value.getClass()));
+    }
+    public static <T> String encode(T value, Encoder<T> encoder) {
```
cmelchior, line 50: Nitpick: Newlines are missing between these methods

realm/realm-library/src/androidTestObjectServer/kotlin/io/realm/RealmFunctionsTests.kt
cmelchior: Looking at the RealmFunctions class it was marked as "in-progress" I guess this is fine for now.

rorbech: Yep. All actual function-implementation (creating it from users, etc.) is in #6810, as it requires OS bson stuff, and I did not want to pull that in yet.

realm/realm-library/src/androidTestObjectServer/kotlin/io/realm/RealmFunctionsTests.kt
rorbech: No. Just because the underlying JNI implementation is just echo'ing (returning the arguments directly) as it is not yet hitting the action Stitch functions .

realm/realm-library/src/objectServer/java/io/realm/RealmAppConfiguration.java
rorbech: 👍

realm/realm-library/src/androidTestObjectServer/kotlin/io/realm/RealmFunctionsTests.kt
rorbech: Seems to be that the C++ roundtrip actually reset subtube. I will raise it with OS.

realm/realm-library/src/androidTestObjectServer/kotlin/io/realm/RealmFunctionsTests.kt
rorbech: Yes. There were some constraints around typing for the `asertTypedEcho` call to match the result. I have updated it to direct `invoke`-calls, to be able to keep the primitive types.

realm/realm-library/src/objectServer/java/io/realm/RealmAppConfiguration.java
```diff
@@ -175,10 +184,25 @@ public File getSyncRootDirectory() {
...
 
+    // FIXME Doc
+    public CodecRegistry getDefaultCodecRegistry() { return codecRegistry; }
+
     /**
      * FIXME
      */
     public static class Builder {
+        // Default BSON codec for passing BSON to/from JNI
+        static CodecRegistry DEFAULT_BSON_CODEC_REGISTRY = CodecRegistries.fromRegistries(
```
edualonso, line 195: Just stumbled upon the `BsonUtils.java` class in the Stitch SDK. The use this:
```
  public static final CodecRegistry DEFAULT_CODEC_REGISTRY =
      fromProviders(
          asList(
              new ValueCodecProvider(),
              new BsonValueCodecProvider(),
              new DocumentCodecProvider(),
              new IterableCodecProvider(),
              new MapCodecProvider()));
```

cmelchior: @rorbech Any thoughts on the above ☝️ ? I would assume we would go with the broadest possible set of defaults, so whatever type people bring there is a fair good chance it will work?

realm/realm-library/src/objectServer/java/io/realm/RealmAppConfiguration.java
```diff
@@ -388,6 +413,14 @@ public Builder syncRootDirectory(File rootDir) {
             return this;
         }
 
+        // FIXME Naming: Does it clash with Realm sync concepts and set up wrong expectations?
```
cmelchior, line 416: We don't have a similar name. I find this name really exotic and not very readable (as a non-native english speaker), but given that I don't really have a good alternative we might as well adopt this I guess. It does has the benefit that it will be familiar to anyone coming to it from the Mongo side.

realm/realm-library/src/objectServer/java/io/realm/RealmAppConfiguration.java
rorbech: Did not evaluate all providers/codecs yet, but just added what I immediately spotted as userful. 
The `MapCodecProvider` sounds like an obvious candidate, and from @edualonso it sounded like the `DocumentCodecProvider` would be useful for the collections. 
So the above `DEFAULT_CODEC_REGISTRY` sound like a good update, but given the parser issues lower down the stack I would prefer awaiting usage/tests and add them in another PR to avoid pushing this one further.

cmelchior: 👍

realm/realm-library/src/objectServer/java/io/realm/RealmAppConfiguration.java
rorbech: Ok, will keep it then.

realm/realm-library/src/objectServer/java/io/realm/RealmAppConfiguration.java
rorbech: I added it as TODO on #6810

realm/realm-library/src/objectServer/java/io/realm/RealmFunctions.java
rorbech: Yes, it already pinpointed some errors in C++ parsing. I migrated the test to work across the actual functions by _echoing_  from the Stitch server in #6810 , as it seems like there is currently no way to easily do _test only_ JNI implementation.

PR #6846, opened 2020-05-08
Reduce build time

cmelchior:
This PR separates some changes discovered while working on https://github.com/realm/realm-java/pull/6822

- We accidentally still built all debug ABI's on PR's instead of only armeabi-v7a. (this should save quite some time)
- Quicker exit if the device is not available. Currently, we wait 1.5 hours before timing out the build which would block other builds because we held a resource lock.
- Fixing a bug when tearing down the build. If it was terminated early it would result in the wrong kind of error being reported.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -117,7 +117,7 @@ try {
                 try {
                   backgroundPid = startLogCatCollector()
                   forwardAdbPorts()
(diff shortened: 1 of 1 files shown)


cmelchior:
That is actually a good idea. We should talk with CI about that to see if it is possible

